PR #4706, opened 2020-06-04
[Monit] Use the string "/usr/bin/syncd\s" to monitor the syncd process

yozhao101:
**- Why I did it**
After discussed with Joe, we use the string "/usr/bin/syncd\s" in Monit configuration file to monitor 
syncd process on Broadcom and Mellanox. Due to my careless, I did not find this bug during the 
previous testing. If we use the string "/usr/bin/syncd" in Monit configuration file to monitor the 
syncd process, Monit will not detect whether syncd process is running or not. 

If we ran the command  `sudo monit procmactch “/usr/bin/syncd”` on Broadcom, there will be three 
processes in syncd container which matched this "/usr/bin/syncd": `/bin/bash /usr/bin/syncd.sh
wait`, `/usr/bin/dsserve /usr/bin/syncd –diag -u -p /etc/sai.d/sai.profile` and `/usr/bin/syncd –diag -
u -p /etc/sai.d/said.profile`. Monit will select the processes with the highest uptime (at there 
`/bin/bash /usr/bin/syncd.sh wait`) to match and did not select `/usr/bin/syncd –diag -u -p
/etc/sai.d/said.profile` to match. 

Similarly, On Mellanox Monit will also select the process with the highest uptime (at there 
`/bin/bash /usr/bin/syncd.sh wait`) to match and did not select `/usr/bin/syncd –diag -u -p
/etc/sai.d/said.profile` to match.

That is why Monit is unable to detect whether syncd process is running or not if we use the string “/usr/bin/syncd” in Monit configuration file. If we use the string "/usr/bin/syncd\s" in Monit configuration file, Monit can filter out the process `/bin/bash /usr/bin/syncd.sh wait` and thus can correctly monitor the syncd process.

**- How I did it**

**- How to verify it**

PR #5153, opened 2020-08-11
[Monit] Unmonitor the processes in containers which are disabled.

yozhao101:
**- Why I did it**
We want to let Monit to unmonitor the processes in containers which are disabled in `FEATURE` table such that
Monit will not generate false alerting messages into the syslog.

**- How I did it**
Monit will periodically run a script which accepts three parameters: `<container_name>`, `<process_name>` and
`<process_cmdline>`. This script will first check whether the container is disabled in the `FEATURE` table or not.
If it is disabled, Monit will skip monitoring the processes. Otherwise, this script will leverage psutil library to inspect
the process tree in host to look for the processes. If the process is not found, then an alerting message will be written
into syslog.

**- How to verify it**
We can change the `state` field of a container in `FEATURE` table from `enabled` to `disabled` and then kill a critical 
process in it to see whether Monit can generate the alerting message in syslog or not. The message format in syslog is:
`<process_name> is not running.`

- [ ] 201811
- [ ] 201911
- [ ] 202006


dockers/docker-database/base_image_files/monit_database
```diff
@@ -3,5 +3,5 @@
 ## process list:
 ##  redis_server
 ###############################################################################
-check process redis_server matching "/usr/bin/redis-server"
-    if does not exist for 5 times within 5 cycles then alert
+check program container_process_redis_server with path "/usr/bin/process_checker database redis-server /usr/bin/redis-server"
```
jleveque, line 6: Does Monit accept colons in the program name? If so, I suggest changing the names for readability to the format, `<container>:<program>`. E.g., this one would be `database:redis-server`

Edit: I found that Monit does NOT accept colons. Instead, use `|` as the separator. E.g., `database|redis-server`

yozhao101: Great suggestion! Will update.

abdosi:
@yozhao101 @jleveque 

While looking into fix for Issue#https://github.com/Azure/sonic-buildimage/issues/5292 I came across this PR.  I am wondering it will be easier we can use 'group' concept of monit.  We can control monitor/unmonitor of this group from hostcfgd same as how we start/unmask and stop/mask the service. I verified locally and looks good.

Same Telemtry monit file with group

check process telemetry matching "/usr/sbin/telemetry"
    #depends on telemetry.unmonitor
    if does not exist for 5 times within 5 cycles then alert
    group telemetry.service

check process dialout_client matching "/usr/sbin/dialout_client_cli"
    #depends on telemetry.unmonitor
    if does not exist for 5 times within 5 cycles then alert
    group telemetry.service

Now when Feature is disabled we can execute sudo monit -g telemetry.service unmonitor from hostcfgd.
And when Feature is enabled we can execute sudo monit -g telemetry.service monitor from hostcfgd.

Any concern in this approach ?

jleveque:
@abdosi: Thank you for this suggestion. I think it is a feasable solution. The only concern I have is we would then spread the need for knowledge of Monit to other components (in this case, hostcfgd). I would prefer to encapsulate and abstract this knowledge as much as possible, but I wouldn't say it's a dealbreaker.

I think we can go ahead with this PR for the time being, and @yozhao101 can investigate the 'group' approach that you propose as a future cleanup/enhancement.

abdosi:
> @abdosi: Thank you for this suggestion. I think it is a feasable solution. The only concern I have is we would then spread the need for knowledge of Monit to other components (in this case, hostcfgd). I would prefer to encapsulate and abstract this knowledge as much as possible, but I wouldn't say it's a dealbreaker.
> 
> I think we can go ahead with this PR for the time being, and @yozhao101 can investigate the 'group' approach that you propose as a future cleanup/enhancement.

@jleveque Thanks. I felt since hostcfgd already has awareness with Feature Table and Services Enable/Disable monit going there makes it more logical.

yozhao101:
> @abdosi: Thank you for this suggestion. I think it is a feasable solution. The only concern I have is we would then spread the need for knowledge of Monit to other components (in this case, hostcfgd). I would prefer to encapsulate and abstract this knowledge as much as possible, but I wouldn't say it's a dealbreaker.
> 
> I think we can go ahead with this PR for the time being, and @yozhao101 can investigate the 'group' approach that you propose as a future cleanup/enhancement.

@jleveque Thanks so much for helping me answering this question. I will investigate the group approach and give the feedback.

platform/barefoot/docker-syncd-bfn/base_image_files/monit_syncd
```diff
@@ -3,5 +3,5 @@
 ## process list:
 ##  syncd
 ###############################################################################
-check process syncd matching "/usr/bin/syncd\s"
-    if does not exist for 5 times within 5 cycles then alert
+check program syncd|syncd with path "/usr/bin/process_checker syncd syncd /usr/bin/syncd --diag"
```
jleveque, line 6: Why did you go back to adding the `--diag` flag? The reason we originally changed to `"/usr/bin/syncd\s"` was so that we wouldn't need to worry about which flags were passed to the process.

platform/mellanox/docker-syncd-mlnx/base_image_files/monit_syncd
```diff
@@ -3,5 +3,5 @@
 ## process list:
 ##  syncd
 ###############################################################################
-check process syncd matching "/usr/bin/syncd\s"
-    if does not exist for 5 times within 5 cycles then alert
+check program syncd|syncd with path "/usr/bin/process_checker syncd syncd /usr/bin/syncd --diag"
```
jleveque, line 6: Is this tested? It seems like this reintroduces the issue that we fixed with https://github.com/Azure/sonic-buildimage/pull/4706. Not all vendors pass the same flags to syncd

platform/barefoot/docker-syncd-bfn/base_image_files/monit_syncd
yozhao101: Good catch! Initially I used the string `/usr/bin/syncd\s` as a parameter of the script `process_checker`. In the line 38 of the script `process_checker`, I checked whether the string `/usr/bin/syncd\s` is `in` the process command line. If I used `/usr/bin/syncd\s` then it will did not match the syncd process, then I went back to use the string `/usr/bin/syncd --diag`.

After you asked me this question, I thought again and found I can use the regular expression in python to do this check. For example, I can still use the string `/usr/bin/syncd\s` as a parameter of the script. In the line 38, `re.search(process_cmdline, ' '.join(process.cmdline()))` can be used to do this check.

jleveque: I'm not sure we even need to use regex here. The original problem was that with `check process`, Monit would search for the string anywhere in the process command line. Now that we're using /usr/bin/process_checker, we have control over the logic. You can adjust the logic to simply ensure that the command starts with the string, then we don't need the regex.

yozhao101: OK, Let me give an example to double check what I am understanding is correct. For example, for `syncd` process, the script `/usr/bin/process_checker` will accept the string `/usr/bin/sync\s` as a parameter. In the line 38, we can use the `' '.join(process.cmdline()).startswith(process_cmd.replace("\s", " "))` method to check whether that process is `syncd` process or not, right?

jleveque: No. I'm saying that `/usr/bin/process_checker` can take `/usr/bin/syncd` as a parameter, and it will only try to find a process whose command line _starts with_ that string. Unlike Monit, which would try to find any process whose command line _contains_ that string.

yozhao101: Got it. That's the difference between Monit and `/usr/bin/process_checker` to find a process. For example, for `syncd` process, `/usr/bin/process_checker` will accept '/usr/bin/syncd' as the parameter and then this script can use the method `(' '.join(proceess.cmdline)).startswith(process_cmdline)` to search any process whose command line start with '/usr/bin/syncd', right?

jleveque: Right.

yozhao101: Joe, Thanks for your great suggestions such that I can polish this script.

files/image_config/monit/process_checker
```diff
@@ -0,0 +1,64 @@
...
+            # If the process entity is found in process tree and it is also in the 'running' or 'sleeping'
+            # state, then it will be marked as 'running'.
+            is_running = False
+            for process in psutil.process_iter(["name", "cmdline", "status"]):
+                # The script process_checker has the command line format '/usr/bin/process_checker <container_name>
+                # <process_name> <process_cmdline>' such as '/usr/bin/process_checker bgp fpmsyncd fpmsyncd'. So
+                # when using psutil to search process 'fpmsyncd', we should skip the process which ran process_checker
+                # since it is not 'fpmsyncd' process although 'fpmsyncd' is a sustring of its cmdline.
+                if process.name() == "process_checker":
+                    continue
```
jleveque, line 36: Is this check necessary?

files/image_config/monit/process_checker
```diff
@@ -0,0 +1,64 @@
...
+
+                if ((process_name == process.name() or (' '.join(process.cmdline())).startswith(process_cmdline))
```
jleveque, line 38: I believe you should be able to simplify this to:

```python
if ' '.join(process.cmdline())).startswith(process_cmdline)
```

If the two strings are equal, `startswith()` should return True.

jleveque (changes requested):

files/image_config/monit/process_checker
yozhao101: Since we already used the `startswith(...)` method to check whether the process command line starts with the string, this check will be unnecessary.

files/image_config/monit/process_checker
yozhao101: Will update.

dockers/docker-fpm-frr/base_image_files/monit_bgp
```diff
@@ -7,17 +7,17 @@
...
+    if status != 0 for 5 times within 5 cycles then alert
 
-check process staticd matching "/usr/lib/frr/staticd"
-    if does not exist for 5 times within 5 cycles then alert
+check program bgp|staticd with path "/usr/bin/process_checker bgp /usr/lib/frr/staticd"
+    if status != 0 for 5 times within 5 cycles then alert
 
-check process bgpcfgd matching "python /usr/local/bin/bgpcfgd"
-    if does not exist for 5 times within 5 cycles then alert
+check program bgp|bgpcfgd with path "/usr/bin/process_checker bgp /usr/bin/python /usr/local/bin/bgpcfgd"
```
jleveque, line 22: This command `/usr/bin/python /usr/local/bin/bgpcfgd` has a space in it. Is this tested? I believe you may need to wrap it in quotes.

yozhao101: Yes, I tested this pattern on DUT and it did work. Because the string `/usr/bin/python /usr/local/bin/bgpcfgd` will be as a single parameter passed to the script `process_checker`. In other words, the content of variable `process_cmdline` will be `/usr/bin/python /usr/local/bin/bgpcfgd`.

jleveque: I'm a bit surprised by this, but as long as it's tested and working as expected, it's OK.

jleveque (changes requested):
Check builds are failing and there are conflicts. Please address.

dockers/docker-fpm-frr/base_image_files/monit_bgp
yozhao101: Yes, that's the reason why it worked. I just wanted to give the detailed reason and found you already got it.

yozhao101:
> Check builds are failing and there are conflicts. Please address.

Fixed.

jleveque:
Retest baseimage please

jleveque:
Retest vsimage please

jleveque:
Retest vsimage please

files/image_config/monit/process_checker
```diff
@@ -0,0 +1,55 @@
...
+            # If the process entity is found in process tree and it is also in the 'running' or 'sleeping'
+            # state, then it will be marked as 'running'.
+            is_running = False
+            for process in psutil.process_iter(["cmdline", "status"]):
+                if ((' '.join(process.cmdline())).startswith(process_cmdline) and process.status() in ["running", "sleeping"]):
+                    is_running = True
+                    break
+
+            if not is_running:
+                print("'{}' is not running.".format(process_cmdline))
```
jleveque, line 36: This script will not be run from the command line. Suggest removing this print statement.

jleveque (changes requested):

jleveque:
@yozhao101: Will this PR cherry-pick cleanly to 201911, or will you need to open a separate PR?

files/image_config/monit/process_checker
yozhao101: Good catch! Initially at here I used the `syslog.syslog(...)` to print the error message if the process was not running. The error message in the syslog will be like the following one:

`ERR monit[22646]: 'lldp|lldpmgrd' status failed (1) -- no output`
`ERR process_checker: 'python /usr/bin/lldpmgrd' is not running`

If I used the `print(...)` statement, the error message will be:

`ERR monit[22050]: 'lldp|lldpmgrd' status failed (1) -- 'python /usr/bin/lldpmgrd' is not running.`

Can you give me some suggestions which format of error message we should choose please?

yozhao101:
> @yozhao101: Will this PR cherry-pick cleanly to 201911, or will you need to open a separate PR?

I will try cherry-pick this PR to 201911.

files/image_config/monit/process_checker
jleveque: Oh, so Monit captures this stdout and appends it to its log message?

yozhao101: Yes,

jleveque: OK. Maybe add a comment above the print statement declaring that if this script is run by Monit, it will append the line to its log message.

yozhao101: Updated.

files/image_config/monit/process_checker
```diff
@@ -0,0 +1,57 @@
...
+            # If the process entity is found in process tree and it is also in the 'running' or 'sleeping'
+            # state, then it will be marked as 'running'.
+            is_running = False
+            for process in psutil.process_iter(["cmdline", "status"]):
+                if ((' '.join(process.cmdline())).startswith(process_cmdline) and process.status() in ["running", "sleeping"]):
+                    is_running = True
+                    break
+
+            if not is_running:
+                # If this script is run by Monit, then the following output will be appneded to
```
jleveque, line 36: Typo: s/appneded/appended/

jleveque (changes requested):

files/image_config/monit/process_checker
yozhao101: Fixed.

jleveque (approved):

yozhao101:
Retest vsimage please

yozhao101:
Retest vsimage please

yozhao101:
Retest vsimage please

jleveque (approved):

PR #5155, opened 2020-08-12
[sonic-yang-models]: Fix vlan name pattern incorrect issue

kuanyu99:
* Original pattern can't accept Vlan1000~Vlan3999. Fix it.

**- Why I did it**

Try to fix dynamic breakout port configuration function when there is Vlan1000~Vlan3999 exist.

**- How I did it**

Modify the yang model pattern of the vlan name string

**- How to verify it**

1. Create a vlan with name, Vlan1000
```config vlan add 1000```

2. Breakout the port
```config interface breakout Ethernet0 4x25G```

3. It should able to breakout the port


jleveque:
@praveen-li: Please review.

praveen-li:
@kuanyu99 : I raised https://github.com/Azure/sonic-buildimage/pull/5165, which has the fix and the test case. If that works for you. We can commit from only PR 5165.
@jleveque

kuanyu99:
> @kuanyu99 : I raised #5165, which has the fix and the test case. If that works for you. We can commit from only PR 5165.
> @jleveque

Yes, it works correctly.

jleveque:
Closing in favor of https://github.com/Azure/sonic-buildimage/pull/5165

PR #5174, opened 2020-08-13
[interfaces] Reduce Calls to SONiC Cfggen

tahmed-dev:
Calls to sonic-cfggen is CPU expensive. This PR reduces calls to
sonic-cfggen to one call during startup when running interfaces-
config.

**- Why I did it**
Reduce time required when invoking interfaces-config

**- How I did it**
Used template batch mode to batch together three calls into one call sonic-cfggen

**- How to verify it**
```
admin@str-s6000-acs-14:~$ time sudo ./interfaces-config-old.sh

real	0m4.801s
user	0m4.011s
sys	0m0.639s
admin@str-s6000-acs-14:~$ time sudo ./interfaces-config-new.sh

real	0m1.617s
user	0m1.360s
sys	0m0.224s
admin@str-s6000-acs-14:~$ diff /etc/network/interfaces.old /etc/network/interfaces.new
admin@str-s6000-acs-14:~$ diff /etc/sysctl.d/90-dhcp6-systcl.conf.old /etc/sysctl.d/90-dhcp6-systcl.conf.new
admin@str-s6000-acs-14:~$ diff /etc/dhcp/dhclient.conf.old /etc/dhcp/dhclient.conf.new
```

- [ ] 201811
- [x] 201911
- [x] 202006

### DIFF
--- a/files/image_config/interfaces/interfaces-config.sh
+++ b/files/image_config/interfaces/interfaces-config.sh
@@ -18,8 +18,14 @@ else
     echo "{ \"ZTP_DHCP_DISABLED\" : \"true\" }" > /tmp/ztp_input.json
 fi
 
-# Create /e/n/i file for existing and active interfaces
-sonic-cfggen -d -j /tmp/ztp_input.json -t /usr/share/sonic/templates/interfaces.j2 > /etc/network/interfaces
+# Create /e/n/i file for existing and active interfaces, dhcp6 sytcl.conf and dhclient.conf
+CFGGEN_PARAMS=" \
+    -d -j /tmp/ztp_input.json \
+    -t /usr/share/sonic/templates/interfaces.j2,/etc/network/interfaces \
+    -t /usr/share/sonic/templates/90-dhcp6-systcl.conf.j2,/etc/sysctl.d/90-dhcp6-systcl.conf \
+    -t /usr/share/sonic/templates/dhclient.conf.j2,/etc/dhcp/dhclient.conf \
+"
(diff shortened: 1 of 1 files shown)


files/image_config/interfaces/interfaces-config.sh
```diff
@@ -18,8 +18,14 @@ else
     echo "{ \"ZTP_DHCP_DISABLED\" : \"true\" }" > /tmp/ztp_input.json
 fi
 
-# Create /e/n/i file for existing and active interfaces
-sonic-cfggen -d -j /tmp/ztp_input.json -t /usr/share/sonic/templates/interfaces.j2 > /etc/network/interfaces
+# Create /e/n/i file for existing and active interfaces, dhcp6 sytcl.conf and dhclient.conf
+CFGGEN_PARAMS=" \
+    -d -j /tmp/ztp_input.json \
+    -t /usr/share/sonic/templates/interfaces.j2,/etc/network/interfaces \
+    -t /usr/share/sonic/templates/90-dhcp6-systcl.conf.j2,/etc/sysctl.d/90-dhcp6-systcl.conf \
```
prsunny, line 25: looks like a typo in systcl (sysctl) but its from the old code.. :)

tahmed-dev: It is a type. This will a bigger change

tahmed-dev: Just searched and it is limited to this container (script, and copy constructs)

PR #5179, opened 2020-08-14
[For test only] Enable sync mode

shi-su:
This PR only intends to test if enabling sync mode could pass the PR checkers. Please do not review or merge it.

**- Why I did it**
This PR intends to test if enabling sync mode could pass the PR checkers.

**- How I did it**
Enable sync mode by default

**- How to verify it**

- [ ] 201811
- [ ] 201911
- [ ] 202006

### DIFF
--- a/platform/barefoot/docker-syncd-bfn/syncd_init_common.sh
+++ b/platform/barefoot/docker-syncd-bfn/syncd_init_common.sh
@@ -1,0 +1,212 @@
+#!/usr/bin/env bash
+
+CMD_SYNCD=/usr/bin/syncd
+
+# dsserve: domain socket server for stdio
+CMD_DSSERVE=/usr/bin/dsserve
+CMD_DSSERVE_ARGS="$CMD_SYNCD --diag"
+
+ENABLE_SAITHRIFT=0
+
+PLATFORM_DIR=/usr/share/sonic/platform
+HWSKU_DIR=/usr/share/sonic/hwsku
+
+SONIC_ASIC_TYPE=$(sonic-cfggen -y /etc/sonic/sonic_version.yml -v asic_type)
+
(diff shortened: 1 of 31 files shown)


platform/barefoot/docker-syncd-bfn/syncd_init_common.sh
```diff
@@ -0,0 +1,212 @@
+#!/usr/bin/env bash
+
+CMD_SYNCD=/usr/bin/syncd
```
lguohan, line 3: not sure what these changes are.

shi-su: This syncd_init_common.sh configures the default async/sync mode for syncd. Since this file is located in sairedis and installed with syncd packages, I copied a file with sync mode enabled here and let it cover the original syncd_init_common.sh file after it is installed with the syncd packages. As such, the sync mode can be enabled by default for both orchagent and syncd. The aim of these changes  is only to test if enabling sync mode can pass the PR checkers.

lguohan: where is this syncd_init_common.sh? can you submit pr to properly change the original file?

shi-su: The file is in sonic-sairedis. I thought we are not going to enable sync mode by default soon, so I made the changes in this PR as a temporary solution only to test sync mode with minimal changes in the submodules. I will create another PR that properly changes the original file when we want to enable sync mode.

shi-su: I am closing this PR since we already confirmed that it could pass all PR checkers with sync mode enabled. Will create another PR that enables sync mode more properly in the future when we want to enable it by default.

PR #5188, opened 2020-08-14
[build] Build/install remaining platform daemons as Python wheel packages

jleveque:
As part of migrating all Python-based package installers to wheel format rather than Debian packages. Also to allow for easily building a Python 3 version of the package in the near future. ledd and psud were converted in earlier PRs. This PR converts the remainder:

- pcied
- syseepromd
- thermalctld
- xcvrd


jleveque:
Retest this please

PR #4549, opened 2020-05-06
[MultiDB] daemon base should use multiDB DBConnector

dzhangalibaba:
* Should use new multiDB Connector in daemon base
* When merging this this PR, need to make sure https://github.com/Azure/sonic-platform-daemons/pull/57 is already updated to the latest.  

* update submodule in this PR as well to avoid dependency between  https://github.com/Azure/sonic-platform-daemons/pull/57 and this PR.
  * [MultiDB] shoule use multiDB API (#57)
  * [thermalctld] Fix invalid warning status (#58)
  * [syseepromd] Prevent the syseepromd from termination (#56)

Signed-off-by: Dong Zhang d.zhang@alibaba-inc.com

Issue #4999, opened 2020-07-17
Create a Python library to house common general SONiC-related APIs


jleveque:
There is currently a lot of duplicated code in many SONiC repos. We need to centralize this common functionality in a Python library which resides in sonic-buildimage that we can import from sonic-daemon-base, sonic-utilities, sonic-platform-common, sonic-platform-daemons, etc. Examples of this functionality include:

- get_platform_and_hwsku()
- get_path_to_platform_and_hwsku()
- get_path_to_port_config_file()
- Logging
- ...

jleveque:
First step toward resolving this: https://github.com/Azure/sonic-buildimage/pull/5003

PR #5191, opened 2020-08-15
[201911] Fix issues with sonic-py-common package

abdosi:
What I did:
Two fixes/change were made with new sonic-py-common package:

a) we should use get_platform() with new sonic_py-common package
instead of get_platform_info()

b) In 201911 DB Connector is still using db id based constructor
as following PR#https://github.com/Azure/sonic-buildimage/pull/4549
is not cherry-picked yet. 
So revert the change to use db is instead of db_name for now. This will be updated when we move to db name 

How I verified:
Because of (b) PMON process were bailing out. After fix things are fine.

### DIFF
--- a/src/sonic-py-common/sonic_py_common/daemon_base.py
+++ b/src/sonic-py-common/sonic_py_common/daemon_base.py
@@ -8,7 +8,11 @@ from .logger import Logger
 #
 # Constants ====================================================================
 #
+# Redis DB information
+REDIS_HOSTNAME = 'localhost'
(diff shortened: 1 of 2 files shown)


lguohan:
what kind of test can catch this?

jleveque:
As the final step of the transition to sonic-py-common, I indend to add unit tests (see: https://github.com/Azure/sonic-buildimage/issues/4999) which can be run at package build time. Whenever anyone adds new functions like these to the library, we should ensure 100% test coverage.

PR #5196, opened 2020-08-15
[build] Clean up any '.pc' directories left by `quilt`

jleveque:
`quilt` may leave behind a `.pc/` directory, which will show up in `git status`. Delete it if it is present as part of cleaning up after `quilt` operations.


lguohan:
@qiluo-msft , can you review?

lguohan:
retest mellanox please

PR #5198, opened 2020-08-17
Minigraph resource type changes

anish-n:
**- Why I did it**
Add resource_type from minigraph to DEVICE_METADATA section

**- How I did it**
Add parsing for field in minigraph metadata section

**- How to verify it**
Unit testing in sonic-config-engine

- [ ] 201811
- [X] 201911
- [ ] 202006

Add minigraph resource_type to DEVICE_METADATA section

### DIFF
--- a/src/sonic-config-engine/minigraph.py
+++ b/src/sonic-config-engine/minigraph.py
@@ -918,7 +918,9 @@ def parse_xml(filename, platform=None, port_config_file=None, asic_name=None, hw
     if sub_role is not None:
         current_device['sub_role'] = sub_role
         results['DEVICE_METADATA']['localhost']['sub_role'] =  sub_role
+    if asic_name is not None:
         results['DEVICE_METADATA']['localhost']['asic_name'] =  asic_name
+
     results['BGP_NEIGHBOR'] = bgp_sessions
     results['BGP_MONITORS'] = bgp_monitors
     results['BGP_PEER_RANGE'] = bgp_peers_with_range
--- a/src/sonic-config-engine/minigraph.py
+++ b/src/sonic-config-engine/minigraph.py
@@ -546,6 +546,7 @@ def parse_meta(meta, hname):
     deployment_id = None
     region = None
     cloudtype = None
+    sub_role = None
     device_metas = meta.find(str(QName(ns, "Devices")))
     for device in device_metas.findall(str(QName(ns1, "DeviceMetadata"))):
         if device.find(str(QName(ns1, "Name"))).text.lower() == hname.lower():
@@ -572,7 +573,9 @@ def parse_meta(meta, hname):
                     region = value
                 elif name == "CloudType":
                     cloudtype = value
-    return syslog_servers, dhcp_servers, ntp_servers, tacacs_servers, mgmt_routes, erspan_dst, deployment_id, region, cloudtype
+                elif name == "SubRole":
+                    sub_role = value
+    return syslog_servers, dhcp_servers, ntp_servers, tacacs_servers, mgmt_routes, erspan_dst, deployment_id, region, cloudtype, sub_role
 
 
 def parse_linkmeta(meta, hname):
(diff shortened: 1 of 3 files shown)


src/sonic-config-engine/minigraph.py
```diff
@@ -915,7 +918,9 @@ def parse_xml(filename, platform=None, port_config_file=None, asic_name=None, hw
     if sub_role is not None:
         current_device['sub_role'] = sub_role
         results['DEVICE_METADATA']['localhost']['sub_role'] =  sub_role
+    if asic_name is not None:
```
prsunny, line 921: asic_name was previously added under sub_role. Was it intentional? @SuvarnaMeenakshi , @abdosi ?

prsunny (dismissed):

prsunny:
retest this please

src/sonic-config-engine/minigraph.py
SuvarnaMeenakshi: asic_name is passed to create configuration for asic and only for asic configuration, sub_role is defined.
So if sub_role isNone, asic_name is also none, hence specific check was not added.

src/sonic-config-engine/minigraph.py
```diff
@@ -572,7 +573,9 @@ def parse_meta(meta, hname):
                     region = value
                 elif name == "CloudType":
                     cloudtype = value
-    return syslog_servers, dhcp_servers, ntp_servers, tacacs_servers, mgmt_routes, erspan_dst, deployment_id, region, cloudtype
+                elif name == "SubRole":
```
arlakshm, line 576: The SubRole is present only when generating the configuration for individual ASICs in a multi ASIC platform
It is handled seperately in this [function ](https://github.com/Azure/sonic-buildimage/blob/eb76a63b09573ca3adb199fb85a9ac4ad35a8833/src/sonic-config-engine/minigraph.py#L613)

Why do we need introduce SubRole in DEVICE_METADATA for all the devices ?

src/sonic-config-engine/tests/simple-sample-graph-metadata.xml
```diff
@@ -253,6 +253,11 @@
      <a:Reference i:nil="true"/>
      <a:Value>10.0.10.7;10.0.10.8</a:Value>
     </a:DeviceProperty>
+    <a:DeviceProperty>
+     <a:Name>SubRole</a:Name>
+     <a:Reference i:nil="true"/>
+     <a:Value>sub_role_1</a:Value>
```
arlakshm, line 259: Can you please elaborate what this Property signifies ?

src/sonic-config-engine/minigraph.py
anish-n: This is so that SubRole can be used for policy enforcement. We will be using it for the internal policies.

src/sonic-config-engine/tests/simple-sample-graph-metadata.xml
anish-n: We will receive specific SubRole DeviceProperty values and use the same for policy enforcement.

pavel-shirshov: So now we have two subroles. One of them was introduced by multiasic efforts and another one is part of the minigraph.
I'd suggest to rename multinpu subrole to something else, otherwise we will always confuse both of them.

arlakshm: Since the new property for the whole device, can rename this to device_role. The sub_role can remain the roles of each asic in a device

abdosi: we already have "type" (Leaf/Tor/Spine)  as part of DEVICE_METADATA. May be have sub_type as part of DEVICE_METADATA

lguohan: we cannot have two sub_role

anish-n: Just to be most precise we are looking at what to name the final DEVICE_METADATA value? We are not looking at changing SubRole to some other element in the minigraph. 

I see 2 suggestions here, one is changing the sub_role for multi_asic, this could be multi_asic_sub_role to be most explicit. The other suggestion is changing sub_role for minigraph to something like what @arlakshm suggested: device_subrole/device_role.

abdosi (dismissed):
LGTM. Why we need this back port to 201911 ?

anish-n:
retest this please

pavel-shirshov (approved):

abdosi (approved):

prsunny (approved):

abdosi:
@anish-n will not take it for 201911. I assume it's ok ?

anish-n:
> @anish-n will not take it for 201911. I assume it's ok ?

Yes we don't need it in 201911

PR #5203, opened 2020-08-17
[cfggen] Allow Write To Redis DB With Template/Batch Mode

tahmed-dev:
Argument to write to config-db is not allowed when using template.
This PR allows cfggen to write to redis db when using template
mode.

signed-off-by: Tamer Ahmed <tamer.ahmed@microsoft.com>

**- Why I did it**
Reduce CPU util during bot

**- How I did it**
relocate write-to-db to new section such that it is not mutually exclusive with new batch mode

**- How to verify it**
I did load minigraph with and without this change and the resulting config_db is the same
```
admin@str-s6000-acs-14:~$ diff config_db.json.new config_db.json.old
admin@str-s6000-acs-14:~$ 
```

During boot profiling, there was about 1~1.5% CPU reduction when batching 4 cfggen calls into one call.

- [ ] 201811
- [x] 201911
- [x] 202006

### DIFF
--- a/src/sonic-config-engine/sonic-cfggen
+++ b/src/sonic-config-engine/sonic-cfggen
@@ -266,10 +266,10 @@ def main():
     parser.add_argument("-T", "--template_dir", help="search base for the template files", action='store')
     group.add_argument("-v", "--var", help="print the value of a variable, support jinja2 expression")
     group.add_argument("--var-json", help="print the value of a variable, in json format")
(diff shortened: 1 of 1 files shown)


tahmed-dev:
retest mellanox please
retest baseimage please

tahmed-dev:
retest mellanox please

tahmed-dev:
retest baseimage please

PR #5081, opened 2020-07-31
[build]: combine feature and container feature table

lguohan:
**- Why I did it**
1. have both feature and container feature table is confusing.
2. a feature should only be available in config db if the feature is built into the image

**- How I did it**
1. remove container feature table
2. do not generate feature entry if the feature is not included
   in the image
3. rename ENABLE_* to INCLUDE_* for better clarity

This requires Azure/sonic-utilities#1015

**- How to verify it**

**- Which release branch to backport (provide reason below if seleted)**

- [ ] 201811
- [x] 201911
- [x] 202006

**- Description for the changelog**

**- A picture of a cute animal (not mandatory but encouraged)**

PR #5210, opened 2020-08-18
Enable NAT feature

kperumalbfn:
NAT feature is not enabled in the latest Sonic image after the recent change 
https://github.com/Azure/sonic-buildimage/pull/5081/files

admin@sonic:~$ show feature status
Feature         State     AutoRestart
--------------  --------  -------------
bgp             enabled   enabled
database        enabled   disabled
dhcp_relay      enabled   enabled
lldp            enabled   enabled
mgmt-framework  enabled   enabled
nat             disabled  enabled
pmon            enabled   enabled

Verified by checking NAT docker on bootup and validated by configuring few static NAT entries.

Backport required to 201911

### DIFF
--- a/files/build_templates/init_cfg.json.j2
+++ b/files/build_templates/init_cfg.json.j2
@@ -29,7 +29,7 @@
                    ("teamd", "enabled", "enabled")] %}
 {%- if include_iccpd == "y" %}{% do features.append(("iccpd", "disabled", "enabled")) %}{% endif %}
 {%- if include_mgmt_framework == "y" %}{% do features.append(("mgmt-framework", "enabled", "enabled")) %}{% endif %}
-{%- if include_nat == "y" %}{% do features.append(("nat", "disabled", "enabled")) %}{% endif %}
+{%- if include_nat == "y" %}{% do features.append(("nat", "enabled", "enabled")) %}{% endif %}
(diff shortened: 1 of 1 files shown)


files/build_templates/init_cfg.json.j2
```diff
@@ -29,7 +29,7 @@
                    ("teamd", "enabled", "enabled")] %}
 {%- if include_iccpd == "y" %}{% do features.append(("iccpd", "disabled", "enabled")) %}{% endif %}
 {%- if include_mgmt_framework == "y" %}{% do features.append(("mgmt-framework", "enabled", "enabled")) %}{% endif %}
-{%- if include_nat == "y" %}{% do features.append(("nat", "disabled", "enabled")) %}{% endif %}
+{%- if include_nat == "y" %}{% do features.append(("nat", "enabled", "enabled")) %}{% endif %}
```
lguohan, line 32: nat is an optional feature, it should remind the disable state until user explicitly enable this feature.

kperumalbfn: Thanks @lguohan, these optional features can be enabled using 'sudo config feature state nat enabled'??

lguohan: yes

lguohan:
nat is an optional feature, it should remain the disable state until user explicitly enable this feature.

PR #5214, opened 2020-08-19
[iccpd] Fix the bug of iccp cmd show.

sdddean:
Signed-off-by: Sun Dandan <sundandan@asterfusion.com>

**- Why I did it**
The cmd "mclagdctl dump state" would goes wrong when there are two or more mclag_groups configured. The field "MCLAG Interface" can not be displayed in some group.

**- How I did it**
Reset the len in each csm.

**- How to verify it**
When 4 mclag_groups were configured, running cmd "mclagdctl dump state" display all mclag_group correctly.

- [ ] 201811
- [ ] 201911
- [x] 202006

### DIFF
--- a/src/iccpd/src/iccp_cmd_show.c
+++ b/src/iccpd/src/iccp_cmd_show.c
@@ -99,6 +99,7 @@ int iccp_mclag_config_dump(char * *buf,  int *num, int mclag_id)
         state_info.role = csm->role_type;
 
         str_size = MCLAGDCTL_PORT_MEMBER_BUF_LEN;
+        len = 0;
 
         LIST_FOREACH(lif_po, &(MLACP(csm).lif_list), mlacp_next)
         {


src/iccpd/src/iccp_cmd_show.c
```diff
@@ -99,6 +99,7 @@ int iccp_mclag_config_dump(char * *buf,  int *num, int mclag_id)
         state_info.role = csm->role_type;
 
         str_size = MCLAGDCTL_PORT_MEMBER_BUF_LEN;
+        len = 0;
```
jleveque, line 102: How does this fix the issue? `len` is initialized to 0 on line 48 and doesn't appear to be modified before this line.

sdddean: In line 61, we list for each `csm`, if we get 4 `enabled_po` in the first `csm`, and then the `len` is modified in line 115, the `len` has not been reinitialized to zero before we turn to the sceond `csm`, and then the second `csm's lif_po->name` could not be correctly copied to `state_info.enabled_po` in line 111, the field "MCLAG Interface" could not be displayed in the second `csm`. So, the len should be reinitialized to 0 in each `csm`.

jleveque: Thanks for the explanation. I missed that this is inside the `LIST_FOREACH` macro/loop.

jleveque:
Retest broadcom please

jleveque:
Retest baseimage please

jleveque:
Retest vs please

jleveque:
Retest vsimage please

PR #5218, opened 2020-08-19
[Mellanox] Add installation of click-default-group package

noaOrMlnx:
**- Why I did it**
Add an installation of click-default-group to mellanox platforms

**- How I did it**
Add the installation in sonic_debian_extension.j2 file. 
**- How to verify it**
After building the image check that switch had click-default-group package. 
``` pip list | grep click-default-group```

- [ ] 201811
- [x] 201911
- [x] 202006


lguohan:
why is this mellanox specific platform?

jleveque:
@noaOrMlnx: We purposefully removed click-default-group, as newer versions of Click support this concept natively (see invoke_without_command). I suggest refactoring your code to utilize this instead.

noaOrMlnx:
@jleveque Can you please provide an example of how to use invoke_without_command and also call to a default command (if not command is passed)?
there is an option like receiving `ctx` as a first parameter, and then execute `ctx.invoke(default_command_name)` but if the default command should receive parameters, calling it with parameters won't work. 
An example would be great. Thanks.

jleveque:
@noaOrMlnx: You can search the sonic-utilities repo for examples, like this one: https://github.com/Azure/sonic-utilities/blob/master/show/main.py#L242

Issue #3161, opened 2019-07-16
radvd out of range MTU error


rajendra-dendukuri:
**Description**

radvd out of range MTU error

As part of "PR#2795 Build radvd from source; Patch so as not to treat out-of-range MTU as an error", the error message was reduced to warning
and failure was bypassed. 

This issue is created to track a final fix. Following fixes have been commited to 4.18 kernel to support setting of bridge MTU to a value > 9100. When SONiC kernrel is upgraded to a kernel > 4.18, the patch done in radvd needs to be removed and Bridge MTU needs to be set to 9100.

https://github.com/torvalds/linux/commit/3341d9170228d521fb38afae14b449308f3e4018#diff-221ed728b95e6fa6475069eed594287c
https://github.com/torvalds/linux/commit/804b854d374e39f5f8bff9638fd274b9a9ca7d33#diff-221ed728b95e6fa6475069eed594287c
https://github.com/torvalds/linux/commit/f40aa23339e2d06b1a8daaece5a511bb1c4f704c#diff-221ed728b95e6fa6475069eed594287c
https://github.com/torvalds/linux/commit/419d14af9e07fb5ca32b1b1614793c6b1e242152#diff-221ed728b95e6fa6475069eed594287c

**Steps to reproduce the issue:**
1. Make sure that DEVICE_METADATA|localhost|type="ToRRouter"

2. Create a VLAN and add an active link to it
config vlan add 100
config vlan member add 100 Ethernet0
   
3. Assing IPv6 address to the Vlan interface
config interface ip add Vlan100 2002::1/64

4. Restart radv container
systemctl restart radv

5. Check syslog

**Describe the results you received:**

Following error message is seen
Jul 16 00:41:19.823049 sonic WARNING radv#radvd[60]: AdvLinkMTU for Vlan1 (9100) must be zero or between 1280 and 1500

**Describe the results you expected:**
No error/warnings need to be observed

jleveque:
Thanks for creating this issue to track this.

jleveque:
Master branch is now running 4.19 kernel, so the patches in the description should be present.

@prsunny: Do you know if there is a plan to set the bridge MTU to 9100?

tahmed-dev:
@jleveque: @prsunny and I verified that we can set the bridge MTU to 9100. [PR:1393](https://github.com/Azure/sonic-swss/pull/1393) should close this issue.

jleveque:
[...]

PR #5242, opened 2020-08-24
[radvd] No longer build from source; Install vanilla Debian package once again

jleveque:
**- Why I did it**

Resolves https://github.com/Azure/sonic-buildimage/issues/3161

- In PR https://github.com/Azure/sonic-buildimage/pull/2795, we started building radvd from source and patching it to prevent it from erroring out when advertising an MTU of 9100 which was greater than the MTU size configured on the bridge interface (1500), which was due to a limitation in the 4.9 Linux kernel.
- Master branch is now using Linux kernel 4.19. As of 4.18, the kernel supports setting a bridge MTU to a value > 1500.
- PR https://github.com/Azure/sonic-swss/pull/1393 modified vlanmgrd to take advantage of this and now configures the MTU of bridge interfaces in SONiC to the proper size of 9100. Therefore, we no longer need to patch radvd. Since we no longer need to patch radvd, we no longer need to build it from source, so we can save build time by going back to simply installing the vanilla radvd Debian package in the router-advertiser container.

**- How I did it**

- Remove radvd Makefile and patch, change docker-router-advertiser Dockerfile template to simply install the vanilla radvd package using apt-get.

**- How to verify it**

- Ensure that radvd starts and runs in the container and that it does *not* emit log messages like the following:
```
ERR radv#radvd[38]: AdvLinkMTU for Vlan1000 (9100) must be zero or between 1280 and 1500
```

- [ ] 201811
- [ ] 201911
- [x] 202006


dockers/docker-router-advertiser/Dockerfile.j2
```diff
@@ -10,6 +10,9 @@ ENV DEBIAN_FRONTEND=noninteractive
 # Update apt's cache of available packages
 RUN apt-get update
 
+# Install radvd Debian package
+RUN apt-get -y install radvd
```
prsunny, line 14: just a thought, should we be always installing the latest or do we need to do some versioning?

jleveque: radvd doesn't seem to update frequently. However, it probably wouldn't hurt to add more versioning throughout SONiC so as to prevent surprises as pacakages get upgraded. I'll look into it here.

jleveque: Locked down the version in the latest commit.

jleveque:
Retest vsimage please

jleveque:
Retest vsimage please

Issue #5241, opened 2020-08-24
supervisord-dependent-startup print out error log when stop daemons


keboliu:
**Description**

supervisord-dependent-startup print out error logs when stopping daemons:

```
Aug 23 22:50:43.789760 sonic-dut INFO swss#supervisord 2020-08-23 22:50:10,061 ERRO pool dependent-startup event buffer overflowed, discarding event 34
Aug 23 22:50:43.789760 sonic-dut INFO swss#supervisord 2020-08-23 22:50:10,063 ERRO pool dependent-startup event buffer overflowed, discarding event 35
Aug 23 22:50:43.789760 sonic-dut INFO swss#supervisord 2020-08-23 22:50:10,064 ERRO pool dependent-startup event buffer overflowed, discarding event 36
Aug 23 22:50:43.789760 sonic-dut INFO swss#supervisord 2020-08-23 22:50:10,066 ERRO pool dependent-startup event buffer overflowed, discarding event 37
Aug 23 22:50:43.789760 sonic-dut INFO swss#supervisord 2020-08-23 22:50:10,067 ERRO pool dependent-startup event buffer overflowed, discarding event 38
Aug 23 22:50:43.789760 sonic-dut INFO swss#supervisord 2020-08-23 22:50:10,069 ERRO pool dependent-startup event buffer overflowed, discarding event 39
Aug 23 22:50:43.789760 sonic-dut INFO swss#supervisord 2020-08-23 22:50:10,070 ERRO pool dependent-startup event buffer overflowed, discarding event 40
Aug 23 22:50:43.789760 sonic-dut INFO swss#supervisord 2020-08-23 22:50:10,071 ERRO pool dependent-startup event buffer overflowed, discarding event 41
Aug 23 22:50:43.789760 sonic-dut INFO swss#supervisord 2020-08-23 22:50:12,076 ERRO pool dependent-startup event buffer overflowed, discarding event 42
Aug 23 22:50:43.789760 sonic-dut INFO swss#supervisord 2020-08-23 22:50:12,080 ERRO pool dependent-startup event buffer overflowed, discarding event 43
Aug 23 22:50:43.789760 sonic-dut INFO swss#supervisord 2020-08-23 22:50:12,082 ERRO pool dependent-startup event buffer overflowed, discarding event 44
[...]

PR #5247, opened 2020-08-25
[dockers][supervisor] Increase event buffer size for dependent-startup

jleveque:
When stopping the swss, pmon or bgp containers, log messages like the following can be seen:

```
Aug 23 22:50:43.789760 sonic-dut INFO swss#supervisord 2020-08-23 22:50:10,061 ERRO pool dependent-startup event buffer overflowed, discarding event 34
Aug 23 22:50:43.789760 sonic-dut INFO swss#supervisord 2020-08-23 22:50:10,063 ERRO pool dependent-startup event buffer overflowed, discarding event 35
Aug 23 22:50:43.789760 sonic-dut INFO swss#supervisord 2020-08-23 22:50:10,064 ERRO pool dependent-startup event buffer overflowed, discarding event 36
Aug 23 22:50:43.789760 sonic-dut INFO swss#supervisord 2020-08-23 22:50:10,066 ERRO pool dependent-startup event buffer overflowed, discarding event 37
```

This is due to the number of programs in the container managed by supervisor, all generating events at the same time. The default event queue buffer size in supervisor is 10. This patch increases that value in all containers in order to eliminate these errors. As more programs are added to the containers, we may need to further adjust these values. I increased all buffer sizes to 25 except for containers with more programs or templated supervisor.conf files which allow for a variable number of programs. In these cases I increased the buffer size to 50. One final exception is the swss container, where the buffer fills up to ~50, so I increased this buffer to 100.

Resolves https://github.com/Azure/sonic-buildimage/issues/5241

### DIFF
--- a/dockers/docker-dhcp-relay/docker-dhcp-relay.supervisord.conf.j2
+++ b/dockers/docker-dhcp-relay/docker-dhcp-relay.supervisord.conf.j2
@@ -10,6 +10,7 @@ autorestart=unexpected
 startretries=0
 exitcodes=0,3
 events=PROCESS_STATE
+buffer_size=50
 
 [eventlistener:supervisor-proc-exit-listener]
 command=/usr/bin/supervisor-proc-exit-listener --container-name dhcp_relay
--- a/dockers/docker-fpm-frr/frr/supervisord/supervisord.conf.j2
+++ b/dockers/docker-fpm-frr/frr/supervisord/supervisord.conf.j2
@@ -10,6 +10,7 @@ autorestart=unexpected
 startretries=0
 exitcodes=0,3
 events=PROCESS_STATE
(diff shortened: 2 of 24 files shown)


jleveque:
Retest broadcom please

jleveque:
Retest vsimage please

keboliu (approved):

rlhui:
@jleveque  - is this needed in 201911?

jleveque:
@rlhui: No, this is only applicable to master.

lguohan:
@jleveque , yes we did as part of improving the fast reboot time.

jleveque:
@lguohan: Thanks. I was unaware that we backported it.

@rlhui: In that case, we should probably try to cherry-pick this to 201911.

PR #5260, opened 2020-08-27
Change update_feature_state call to pass False as default if feature has no 'has_timer' field

noaOrMlnx:
**- Why I did it**
In order to make hostcfgd not to fail if feature has no 'has_timer' field
**- How I did it**
Change the update_feature_state call to pass False by default.
**- How to verify it**
Enable feature that has no 'has timer' field and make sure hostcfgd not failing. 

- [ ] 201811
- [x] 201911
- [x] 202006

### DIFF
--- a/files/image_config/hostcfgd/hostcfgd
+++ b/files/image_config/hostcfgd/hostcfgd
@@ -291,7 +291,7 @@ class HostConfigDaemon:
                 syslog.syslog(syslog.LOG_WARNING, "Eanble state of feature '{}' is None".format(feature_name))
                 continue
 
-            update_feature_state(feature_name, state, feature_table[feature_name]['has_timer'])
+            update_feature_state(feature_name, state, feature_table[feature_name].get('has_timer', False))
 
     def aaa_handler(self, key, data):
         self.aaacfg.aaa_update(key, data)
--- a/files/image_config/hostcfgd/hostcfgd
+++ b/files/image_config/hostcfgd/hostcfgd
@@ -291,7 +291,7 @@ class HostConfigDaemon:
                 syslog.syslog(syslog.LOG_WARNING, "Eanble state of feature '{}' is None".format(feature_name))
                 continue
 
-            update_feature_state(feature_name, state, feature_table[feature_name].get('has_timer', False))
+            update_feature_state(feature_name, state, feature_table[feature_name].get('has_timer', 'False'))
 
     def aaa_handler(self, key, data):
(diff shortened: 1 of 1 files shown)


noaOrMlnx:
retest vsimage please

liat-grozovik (approved):

liat-grozovik:
retest vsimage please

tahmed-dev:
@noaOrMlnx, can you please check with latest master. There was an issue where has_timer was passed in as a string and so it was evaluating to `True` all the time. That issues was fixed via [PR:5241](https://github.com/Azure/sonic-buildimage/issues/5241)

jleveque:
retest vsimage please

noaOrMlnx:
@tahmed-dev I checked an image with the merged PR. 
The problem is caused when calling `update_feature_state` function, and trying to receive the has_timer field, even though there is no such field. 
`update_feature_state(feature_name, state, feature_table[feature_name]['has_timer'])` -> this call makes the feature initialization to fail if the feature doesn't have the field. 
This PRs purpose is to prevent the failure.

noaOrMlnx:
retest vsimage please

noaOrMlnx:
retest vsimage please

liat-grozovik:
retest bsimage please

liat-grozovik:
retest baseimage please

tahmed-dev:
> @tahmed-dev I checked an image with the merged PR.
> The problem is caused when calling `update_feature_state` function, and trying to receive the has_timer field, even though there is no such field.
> `update_feature_state(feature_name, state, feature_table[feature_name]['has_timer'])` -> this call makes the feature initialization to fail if the feature doesn't have the field.
> This PRs purpose is to prevent the failure.

@noaOrMlnx what is the exact scenario where this new field `has_timer` would go missing? The reason is I see all feature have their `has_timer` fields populated in [init_cfg.json.j:21](https://github.com/Azure/sonic-buildimage/blob/master/files/build_templates/init_cfg.json.j2#L21). So, it is not clear to me which scenario requires this change.

noaOrMlnx:
> @noaOrMlnx what is the exact scenario where this new field has_timer would go missing? The reason is I see all feature have their has_timer fields populated in init_cfg.json.j:21. So, it is not clear to me which scenario requires this change.

@tahmed-dev This change is required for external features we have. Instead of changing the feature structure, this change will prevent the system to crush if the external feature doesn't have this field.

tahmed-dev (approved):

PR #5265, opened 2020-08-27
[sonic-py-common] Add function to get hostname

renukamanavalan:
Added one more common function to retrieve hostname.

### DIFF
--- a/src/sonic-py-common/sonic_py_common/device_info.py
+++ b/src/sonic-py-common/sonic_py_common/device_info.py
@@ -29,8 +29,23 @@ FRONTEND_ASIC_SUB_ROLE = "FrontEnd"
 BACKEND_ASIC_SUB_ROLE = "BackEnd"
 
 
+def get_localhost_info(field):
+    try:
+        config_db = ConfigDBConnector()
+        config_db.connect()
+
+        metadata = config_db.get_table('DEVICE_METADATA')
+
+        if 'localhost' in metadata and field in metadata['localhost']:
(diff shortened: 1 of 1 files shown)


src/sonic-py-common/sonic_py_common/device_info.py
```diff
@@ -29,6 +29,10 @@
 BACKEND_ASIC_SUB_ROLE = "BackEnd"
 
 
+def get_hostname():
+    return os.uname()[1]
```
prsunny, line 33: Should this be from DEVICE_METADATA? Not sure if we recently had any case of this one different from `os.uname` and/or this uname output change in any kernel version. Just a thought!
`"DEVICE_METADATA": "localhost": "hostname"`

jleveque: It depends on whether this function will ever be called before the database is ready.

prsunny: Agree with function. I'm was only referring, if this function should fetch the hostname from DEVICE_METADATA. Fine with this as well.

jleveque: Technically, the OS host name will not be configured until the Config DB is populated. Thus, maybe we *should* always read it from Config DB here...

renukamanavalan: Changed it. 
Thank you guys, for looking at this in-depth.

PR #5264, opened 2020-08-27
[sonic-utilities][sonic-py-common] Move logic to get port config file path to sonic-py-common and update sonic-utilities to comply

vdahiya12:
This PR addresses fixes in sonic-py-common to imitate the behavior inside 
sonic-cfggen. Essentially this is a fix for accessing the port-config file. 
First check if there is a platform.json file for config generation 
and then for legacy port_config.ini. 
Also updating the sub-module sonic-utilities. 

sonic-utilities update submodule includes all these PR's 

Fix pfcwd stats crash with invalid queue name (#1077)
[show][bgp]Display the Total number of neighbors in the show ip bgp(v6) summary. (#1079)
[config] Update SONiC Environment Vars When Loading Minigraph (#1073)
Multi asic platform changes for interface, portchannel commands (#878)
Update Command-Reference.md (#1075)
[filter-fdb] Fix Filter FDB With IPv6 Present in Config DB (#1059)
[config] Remove _get_breakout_cfg_file_name helper function (#1069)
[SHOW][BGP] support show ip(v6) bgp summary for multi asic platform (#1064)
[fanshow] Display other fan status, such as Updating (#1014)
Add ip_prefix len based on proxy_arp status (#1046)
Enable the platform specific ssd firmware upgrade during reboot (#954)
[show][cli[show interface portchannel support for Multi ASIC (#1005)
support show interface commands for multi ASIC platforms (#1006)

Signed-off-by: vaibhav-dahiya <vdahiya@microsoft.com>

PR #5268, opened 2020-08-28
[docker-sonic-vs] Create /usr/share/sonic/platform symlink

jleveque:
**- Why I did it**
Copying platform.json file into an empty /usr/share/sonic/platform directory does not mimic an actual device. A more correct approach is to create a /usr/share/sonic/platform symlink which links to the actual platform directory; this is more like what is done inside SONiC containers. Then, we only need to copy the platform.json file into the actual platform directory; the symlink takes care of the alternative path, and also exposes all the other files in the platform directory.

One question: Since the SONiC VS container mimics the host OS of a physical switch, do we even need to create the /usr/share/sonic/platform and /usr/share/sonic/hwsku symlinks? Maybe we can remove these in a future patch.

**- How I did it**
Create /usr/share/sonic/platform symlink in start.sh

**- How to verify it**
Start a SONiC VS container and ensure the symlink exists.

- [ ] 201811
- [ ] 201911
- [ ] 202006

### DIFF
--- a/platform/vs/docker-sonic-vs/Dockerfile.j2
+++ b/platform/vs/docker-sonic-vs/Dockerfile.j2
@@ -136,7 +136,6 @@ COPY ["chassis_db.py", "/usr/bin/"]
 
 COPY ["platform.json", "/usr/share/sonic/device/x86_64-kvm_x86_64-r0/"]
 COPY ["hwsku.json", "/usr/share/sonic/device/x86_64-kvm_x86_64-r0/Force10-S6000/"]
-COPY ["platform.json", "/usr/share/sonic/platform/"]
 
 # Workaround the tcpdump issue
 RUN mv /usr/sbin/tcpdump /usr/bin/tcpdump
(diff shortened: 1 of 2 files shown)


jleveque:
@samaity: Please review.

jleveque:
Retest vsimage please

samaity:
LGTM! @jleveque. Just one question is there any place where we try to find the port configuration file like `portconfig.ini` under `/usr/share/sonic/platform/` folder. During implementation, I was thinking of doing something like this. But then I was just being extra cautious so, didn't change anything.

jleveque:
@samaity:

> Just one question is there any place where we try to find the port configuration file like `portconfig.ini` under `/usr/share/sonic/platform/` folder. During implementation, I was thinking of doing something like this. But then I was just being extra cautious so, didn't change anything. 

I'm not clear what your question is here. Is this in regard to my question @ https://github.com/Azure/sonic-buildimage/pull/5264#discussion_r478753566? If so, can you please respond there? If it's not related to that, can you please add some more details here?

jleveque:
Retest vsimage please

PR #5269, opened 2020-08-28
[sonic-py-common] Relocate some functions from sonic-utilities

judyjoseph:
Move the common multi-asic/general functions from sonic-utilities to py-common 
validate_namespace()
get_port_table_name()
get_interface_table_name()

### DIFF
--- a/src/sonic-py-common/tests/interface_test.py
+++ b/src/sonic-py-common/tests/interface_test.py
@@ -1,22 +1,7 @@
 import os
 import sys
 
-# TODO: Remove this if/else block once we no longer support Python 2
-if sys.version_info.major == 3:
-    from unittest import mock
-else:
-    # Expect the 'mock' package for python 2
-    # https://pypi.python.org/pypi/mock
-    import mock
-
 from sonic_py_common import interface
-
-
-# TODO: Remove this if/else block once we no longer support Python 2
-if sys.version_info.major == 3:
-    BUILTINS = "builtins"
-else:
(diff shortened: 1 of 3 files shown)


src/sonic-py-common/tests/interface_test.py
```diff
@@ -0,0 +1,56 @@
+import os
+import sys
+
+# TODO: Remove this if/else block once we no longer support Python 2
+if sys.version_info.major == 3:
+    from unittest import mock
+else:
+    # Expect the 'mock' package for python 2
+    # https://pypi.python.org/pypi/mock
+    import mock
```
jleveque, line 10: It appears this is not being used in this file. Suggest to remove. We can add it back if we need to do mocking in this file at a later time.

src/sonic-py-common/tests/interface_test.py
```diff
@@ -0,0 +1,56 @@
...
+
+from sonic_py_common import interface
+
+
+# TODO: Remove this if/else block once we no longer support Python 2
+if sys.version_info.major == 3:
+    BUILTINS = "builtins"
+else:
+    BUILTINS = "__builtin__"
```
jleveque, line 19: It appears this is not being used in this file. Suggest to remove. We can add it back if we need to do mocking in this file at a later time.

judyjoseph: Sure, removed unused imports from the interface_test.

judyjoseph:
retest this please

PR #5282, opened 2020-08-31
Secure boot

kelly-yeh-ms:
- Why I did it
To make SONiC boot with signed shim and grub when secure boot is enabled
Allow secure boot key signing and secure boot verification

- How I did it
Added signed grub and shim binary packages
Boot with shim instead of grub
Added efitools and mokutil 

- How to verify it
Use sbsign to sign binaries and mokutil to check secure boot state
Check if the following exists:
/usr/lib/shim/shimx64.efi.signed
/usr/lib/grub/x86_64-efi-signed/grubx64.efi.signed

### DIFF
--- a/build_debian.sh
+++ b/build_debian.sh
@@ -147,6 +147,8 @@ sudo openssl req -new -x509 -newkey rsa:2048 -subj "/CN=db/" -keyout kernel_db.k
 sudo openssl x509 -in kernel_db.crt -outform der -out kernel_db.der
 sudo sbsign --key kernel_db.key --cert kernel_db.crt --output fsroot/boot/vmlinuz-4.19.0-9-2-amd64 fsroot/boot/vmlinuz-4.19.0-9-2-amd64
 
+sudo apt-get -y install mokutil
+
 ## Update initramfs for booting with squashfs+overlay
 cat files/initramfs-tools/modules | sudo tee -a $FILESYSTEM_ROOT/etc/initramfs-tools/modules > /dev/null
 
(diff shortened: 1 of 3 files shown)


installer/x86_64/install.sh
```diff
@@ -412,7 +412,7 @@ demo_install_uefi_grub()
     efibootmgr --quiet --create \
         --label "$demo_volume_label" \
         --disk $blk_dev --part $uefi_part \
-        --loader "/EFI/$demo_volume_label/grubx64.efi" || {
+        --loader "/EFI/$demo_volume_label/shimx64.efi" || {
```
lguohan, line 415: does this affect non secure boot?

jleveque:
Please add more detail to the PR title

build_debian.sh
```diff
@@ -142,6 +142,12 @@ sudo LANG=C DEBIAN_FRONTEND=noninteractive chroot $FILESYSTEM_ROOT apt-get -y in
 if [[ $CONFIGURED_ARCH == amd64 ]]; then
     sudo LANG=C DEBIAN_FRONTEND=noninteractive chroot $FILESYSTEM_ROOT apt-get -y install dmidecode hdparm
 fi
+sudo apt-get -y install efitools
+sudo openssl req -new -x509 -newkey rsa:2048 -subj "/CN=db/" -keyout kernel_db.key -out kernel_db.crt -days 365 -nodes -sha256
+sudo openssl x509 -in kernel_db.crt -outform der -out kernel_db.der
+sudo sbsign --key kernel_db.key --cert kernel_db.crt --output fsroot/boot/vmlinuz-4.19.0-9-2-amd64 fsroot/boot/vmlinuz-4.19.0-9-2-amd64
```
lguohan, line 148: use LINUX_KERNEL_VERSION

installer/x86_64/install.sh
kelly-yeh-ms: Yes, but not if  we add shimx64.efi to the EFI partition. Still working on this.

PR #5283, opened 2020-08-31
[ChassisDB]: bring up ChassisDB service

BrynXu:
bring up chassisdb service on sonic switch according to the design in
Distributed Forwarding in VoQ Arch HLD

Signed-off-by: Honggang Xu <hxu@arista.com>

**- Why I did it**
To bring up new ChassisDB service in sonic as designed in ['Distributed forwarding in a VOQ architecture HLD' ](https://github.com/Azure/SONiC/blob/90c1289eaf89a70939e7c81e042e261947c6a850/doc/chassis/architecture.md). 

**- How I did it**
Implement the section 2.3.1 Global DB Organization of the VOQ architecture HLD.

**- How to verify it**
ChassisDB service won't start without chassisdb.conf file on the existing platforms.
ChassisDB service is accessible with global.conf file in the distributed arichitecture.

- [ ] 201811
- [ ] 201911
- [ ] 202006


dockers/docker-database/ping_pong_db_insts
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/python
+""" this tool is only used in docker-database-global container post init script to
+wait for docker-database container up running. docker-database-gloabl container
```
eswaranb, line 3: typo docker-database-gloabl -> docker-database-global

BrynXu:
retest braodcom please, retest vsimage please

BrynXu:
retest broadcom please

BrynXu:
retest vsimage please

dockers/docker-database-global/database_config.json
```diff
@@ -0,0 +1,18 @@
...
+            "hostname" : "redis_global.server",
+            "port": 6385,
+            "unix_socket_path": "/var/run/redis/redis_global.sock",
+            "unix_socket_perm" : 777
+        }
+    },
+    "DATABASES" : {
+        "GLOBAL_APP_DB" : {
+            "id" : 8,
+            "separator": "|",
```
eswaranb, line 13: This separator must be :

dockers/docker-database-global/database_config.json
```diff
@@ -0,0 +1,18 @@
...
+        "redis_global":{
+            "hostname" : "redis_global.server",
+            "port": 6385,
+            "unix_socket_path": "/var/run/redis/redis_global.sock",
+            "unix_socket_perm" : 777
+        }
+    },
+    "DATABASES" : {
+        "GLOBAL_APP_DB" : {
+            "id" : 8,
```
eswaranb, line 12: Double check to make sure this ID doesn't conflict with the latest code.

files/build_templates/docker_image_ctl.j2
```diff
@@ -154,6 +154,10 @@ function postStartAction()
             docker cp $PSENSOR pmon:/usr/bin/
         fi
     fi
+{%- elif docker_container_name == "database-global" %}
+    # Wait until redis starts
+    /usr/bin/docker exec database-global$DEV ping_pong_db_insts supervisor
```
eswaranb, line 159: Check if swss startup has a ping-pong to validate connectivity

dockers/docker-database/database_config.json.j2
```diff
@@ -5,6 +5,12 @@
             "port" : 6379,
             "unix_socket_path" : "/var/run/redis{{NAMESPACE_ID}}/redis.sock",
             "persistence_for_warm_boot" : "yes"
+        },
+        "redis_global":{
+            "hostname" : "redis_global.server",
+            "port": 6385,
+            "unix_socket_path": "/var/run/redis-global/redis_global.sock",
```
eswaranb, line 12: This is not used. Maybe get rid of it?

files/build_templates/docker_image_ctl.j2
```diff
@@ -154,6 +154,10 @@ function postStartAction()
...
+    /usr/bin/docker exec database-global$DEV python /usr/bin/global_db.py
```
eswaranb, line 160: For now, the first connection is necessary to set the AKE mode correctly in redis server. We could avoid this dependency by populating redis.conf correctly.

dockers/docker-database/database_config.json.j2
```diff
@@ -52,6 +58,11 @@
             "id" : 7,
             "separator": "|",
             "instance" : "redis"
+        },
+        "GLOBAL_APP_DB" : {
+            "id" : 8,
```
lguohan, line 63: since this is a new redis instance, should it start from 0?

files/build_templates/docker_image_ctl.j2
```diff
@@ -234,6 +245,19 @@ start() {
...
 
+    global_server="redis_global.server"
+    global_db_address="127.100.0.1"
+    redis_global_server="$( grep $global_server /etc/hosts )"
+    if [ -n "$redis_global_server" ]; then
+        global_db_address="$( grep "redis_global.server" /etc/hosts | cut -f 1 -d ' ' )"
+    fi
+    global_db_opt="--add-host=$global_server:$global_db_address"
+    start_globaldb=$(sonic-cfggen -G -v GLOBALDB_CONFIG.start_global_db)
+    globaldb_address=$(sonic-cfggen -G -v GLOBALDB_CONFIG.global_db_address)
```
lguohan, line 256: @tahmed-dev , what is the way to optimize this sonic-cfggen, we'd like to reduce the sonic-cfggen call in the startup.

files/build_templates/per_namespace/database.service.j2
```diff
@@ -3,6 +3,9 @@ Description=Database container
 {% if multi_instance == 'true' %}
 Requires=database.service
 After=database.service
+{% else %}
```
eswaranb, line 6: Can we have a variable for chassis_system and then use it here?

dockers/docker-database-global/global_db.py
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env python
+
+import swsssdk
+import json
+globaldb = swsssdk.SonicV2Connector(host='redis_global.server', port='6385')
```
lguohan, line 5: how is this 6385 decided?

files/build_templates/per_namespace/database.service.j2
eswaranb: Either chassis_system or switch type voq will need database-global.service

files/build_templates/docker_image_ctl.j2
lguohan: @qiluo-msft, how is AKE mode is set here?

dockers/docker-database-global/Dockerfile.j2
```diff
@@ -0,0 +1,10 @@
+FROM docker-database
```
eswaranb, line 1: Can we even avoid creating the docker-database-global image altogether? Just use docker-database image and when the docker-database-global container is started, it is started with appropriate env variables.

src/sonic-config-engine/sonic-cfggen
```diff
@@ -371,6 +373,12 @@ def main():
             hardware_data['DEVICE_METADATA']['localhost'].update(asic_id=asic_id)
         deep_update(data, hardware_data)
 
+    # look for globaldb.config at /usr/share/sonic/<platform>/
```
tahmed-dev, line 376: What does `get_globaldb_config(..)` do? Can the search for `globaldb.config` be done outside this module and pass the info as  parameter to this module.

platform/vs/docker-sonic-vs/start.sh
```diff
@@ -59,34 +59,34 @@ rm -f /var/run/rsyslogd.pid
...
-   python /usr/bin/chassis_db.py
+start_global_db=`sonic-cfggen -v DEVICE_METADATA.localhost.start_global_db -y $globaldb_cfg_file`
+if [[ "$HOSTNAME" == *"supervisor"* ]] || [ "$start_global_db" == "1" ]; then
+   supervisorctl start redis-global
+   python /usr/bin/global_db.py
 fi
 
-conn_chassis_db=`sonic-cfggen -v DEVICE_METADATA.localhost.connect_to_chassis_db -y $chassis_cfg_file`
-if [ "$start_chassis_db" != "1" ] && [ "$conn_chassis_db" != "1" ]; then
+conn_global_db=`sonic-cfggen -v DEVICE_METADATA.localhost.connect_to_global_db -y $globaldb_cfg_file`
```
tahmed-dev, line 86: can we combine the calls to `sonic-cfggen` into one call using a template file? Here is an [example](https://github.com/Azure/sonic-buildimage/blob/master/dockers/docker-sonic-telemetry/telemetry_vars.j2).

dockers/docker-database-global/database_config.json
eswaranb: Does the ID even need to be unique? Check if DbConnector actually has a bug here that requires conflicts.

dockers/docker-database-global/global_db.py
eswaranb, line 5: Can we avoid hardcoding the port?

platform/vs/docker-sonic-vs/start.sh
tahmed-dev: I'd do something like:
```
SONIC_CFGGEN_PARAMS="
    -v DEVICE_METADATA.localhost.connect_to_global_db \
    -y $globaldb_cfg_file \
    -t globaldb_vars.json.j2,/tmp/globaldb.json
"
if [ -r "$globaldb_cfg_file" ]; then
￼    $SONIC_CFGGEN_PARAMS+=" -j $globaldb_cfg_file -t $host_template,/etc/hosts"
else
.
fi
$(sonic-cfggen $SONIC_CFGGEN_PARAMS)

￼

File: dockers/docker-database-global/global_db.py [hunk 14]
```diff
@@ -0,0 +1,13 @@
...
+globaldb.connect( globaldb.GLOBAL_APP_DB )
+fname='/etc/sonic/global_db.json'
```
eswaranb, line 7: Add a comment: This is temporary and will go away eventually once all the contents of GLOBAL_APP_DB come from each FSI.

device/arista/x86_64-arista_7800_sup/globaldb.conf
```diff
@@ -0,0 +1,2 @@
+start_global_db=1
```
eswaranb, line 1: How can the global db entries be flushed by swss on startup. (swss.sh for reference)

dockers/docker-database/docker-database-init.sh
```diff
@@ -38,7 +53,14 @@ then
     fi
 fi
 
-# generate all redis server supervisord configuration file
-sonic-cfggen -j /var/run/redis/sonic-db/database_config.json -t /usr/share/sonic/templates/supervisord.conf.j2 > /etc/supervisor/conf.d/supervisord.conf
+sonic-cfggen -j $db_cfg_file_tmp -t /usr/share/sonic/templates/supervisord.conf.j2 > /etc/supervisor/conf.d/supervisord.conf
+
+start_globaldb=$(sonic-cfggen -G -v GLOBALDB_CONFIG.start_global_db)
```
tahmed-dev, line 58: similar comments here regarding the use of `sonic-cfggen`

files/build_templates/docker_image_ctl.j2
tahmed-dev: ACK. @BrynXu can you please json template file with `start_global_db` and `global_db_address`? Here is an [example](https://github.com/Azure/sonic-buildimage/blob/master/dockers/docker-orchagent/swss_vars.j2)

files/build_templates/docker_image_ctl.j2
qiluo-msft: Seems it is set by [https://github.com/Azure/sonic-py-swsssdk/blob/c25d492c4cbb356b1fdd4502f75eda7db9e48aea/src/swsssdk/interface.py#L183](https://github.com/Azure/sonic-py-swsssdk/blob/c25d492c4cbb356b1fdd4502f75eda7db9e48aea/src/swsssdk/interface.py#L183)
We should move it into /etc/redis/redis.conf instread of runtime set.


---
In reply to: [485785869](https://github.com/Azure/sonic-buildimage/pull/5283#discussion_r485785869) [](ancestors = 485785869)

eswaranb: Is the right next step to open an issue to track this work?

dockers/docker-database/database_config.json.j2
BrynXu: There is an issue on handling multi-instances in [table.cpp](https://github.com/Azure/sonic-swss-common/blob/master/common/table.cpp),  const TableNameSeparatorMap TableBase::tableNameSeparatorMap doesn't consider multi-instance case, I opened a new issue to track it - [5351](https://github.com/Azure/sonic-buildimage/issues/5351)

src/sonic-config-engine/sonic-cfggen
BrynXu: To locate the globaldb.config, platform info is required, but the platform info is obtained through callng sonic-cfggen.
The globaldb.config file is key-value pairs text file, get_lobaldb_config() parses it and return them in json. There are multiple shell scripts need parse the file, I feel it is better to centralize them in one place inside sonic-cfggen like sonic-cfggen parses machine.conf and asic.conf. 
passing globaldb.config info to sonic-cfggen will lost benefit to use sonic-cfggen. If it is not proper to wrapper globaldb.config parsing logic inside sonic-cfggen, we can just hardcode the parsing logic in different places, I feel that is hard to maintain in future.

platform/vs/docker-sonic-vs/start.sh
BrynXu: I can the two calls into one call as the example does, one concern is that will require jq tool, is jq available on host and database container?

src/sonic-config-engine/sonic-cfggen
tahmed-dev: A counter example from existing code is [here](https://github.com/Azure/sonic-buildimage/blob/master/files/image_config/config-setup/config-setup#L212) Please notice that the platform is available as an envvar in `/etc/sonic/sonic-evironment` file.`sonic-cfggen` is heavy weight process and it does delay the switch start time.

platform/vs/docker-sonic-vs/start.sh
tahmed-dev: `jq` is available in dockers' nut not on host yet. There was no need to add this support then. please feel free to add it here.

files/build_templates/docker_image_ctl.j2
BrynXu: @tahmed-dev I will follow your counter example of parsing config file [example](https://github.com/Azure/sonic-buildimage/blob/master/files/image_config/config-setup/config-setup#L212) to parse globaldb.conf directly in code instead of calling sonic-cfggen.

files/build_templates/docker_image_ctl.j2
BrynXu: swss container startup doesn't have a ping-pong to validate db connection.

files/build_templates/docker_image_ctl.j2
BrynXu: this is generic redis issue, it should be addressed out of this PR.

files/build_templates/per_namespace/database.service.j2
BrynXu: switch type is not published in make file, I don't have clue how we will generate it at compiling time. when it is ready, we use it here, for now, database.service is started by condition, it won't run without globaldb.conf present.

device/arista/x86_64-arista_7800_sup/globaldb.conf
BrynXu: we have switch-id as part of key for entries in global db, that info can be used to flush the db, but the flush function should be handle by swss code not in the scope of this PR.

dockers/docker-database-global/Dockerfile.j2
BrynXu: current slave.mk doesn't support sharing image for new container, and per-instance container doesn't fit globaldb case. I made the change to support using docker-database image, it appears to work in initial test, but some issues with per-instance database container are still under investigation.

dockers/docker-database-global/database_config.json
BrynXu: 8 current is not conflict with other. Yes, now db id has to be unique even crossing instances, this is a bug in sonic:
Duplicated dbid in redis multi-instances causes error #5351

dockers/docker-database-global/global_db.py
BrynXu: it is a IANA unassigned port, should be good for internal use. Also I made the global db port number configurable through globaldb.conf

dockers/docker-database/database_config.json.j2
BrynXu: unic_soucket_path has to be valid when start redis server.

BrynXu: sonic-db-cl -n asic0 PING will use domain socket to ping instead of TCP, we have to make it valid so that PING won't fail in multi-asic env.

dockers/docker-database/docker-database-init.sh
```diff
@@ -17,16 +20,45 @@ then
...
 if [ -f /etc/sonic/database_config$NAMESPACE_ID.json ]; then
     cp /etc/sonic/database_config$NAMESPACE_ID.json $REDIS_DIR/sonic-db/database_config.json
 else
     HOST_IP=$host_ip j2 /usr/share/sonic/templates/database_config.json.j2 > $REDIS_DIR/sonic-db/database_config.json
 fi
 
-mkdir -p /etc/supervisor/conf.d/
+# on modular system, we only publish redis_chassis instance and CHASSIS_APP_DB when
+# either chassisdb.conf indicates starts chassis_db or connect to chassis_db,
+# and redis_chassis instance is started in differnt container.
```
mlorrillere, line 33: differnt -> different

BrynXu:
retest baseimage please

BrynXu:
retest vs please
retest vsimage please

BrynXu:
retest vs please

BrynXu:
retest mellanox please

BrynXu:
retest vs please

dockers/docker-database/database_config.json.j2
```diff
@@ -72,6 +78,11 @@
             "id" : 11,
             "separator": "|",
             "instance" : "redis"
+        },
+        "CHASSIS_APP_DB" : {
+            "id" : 12,
+            "separator": ":",
```
eswaranb, line 84: Change this back to VBAR

dockers/docker-database/database_config.json.j2
```diff
@@ -5,6 +5,12 @@
             "port" : 6379,
             "unix_socket_path" : "/var/run/redis{{NAMESPACE_ID}}/redis.sock",
             "persistence_for_warm_boot" : "yes"
+        },
+        "redis_chassis":{
+            "hostname" : "redis_chassis.server",
+            "port": 6385,
```
eswaranb, line 11: Can this be 6380? Why 6385?

lguohan (dismissed):

BrynXu:
retest vsimage please

lguohan (approved):

Issue #5275, opened 2020-08-30
COPP with ~350 rules take more than 10 min to install at iptables 


Ezrickd:
Sonic swith running 201911 build commit f6a8678d
switch has 4 tables 2 for IPv4 and 2 for IPv6 for each there are two protocols SSH & SNMP.
when config reload and check iptables -L -v -n --line-number we see the list of rules go up and down, as if the table is flushed for every rule that is installed. I can see the same on syslog file that rules are insterted to iptales again and again....

it takes the device to install all rules more than 10 Min when at this time you can access the device though the rule applyed if Drop.
untill rule table is syncd with iptbels.

i have used an old caclmgrd config file and the issue is resolved, so ~350 rules are installed instanlly.
the file was taken from old post which is updated by now..

https://github.com/Azure/sonic-buildimage/blob/master/files/image_config/caclmgrd/caclmgrd 

attached show from terminal display the Iptables number going up and down, also syslog file that dispaly the insertion of the rules.

attached rule & tables from config_db.json 
attched old caclmgrd file that resolve the issue.

[ACL_Rule & Table .txt](https://github.com/Azure/sonic-buildimage/files/5146747/ACL_Rule.Table.txt)
[iptables_line number.txt](https://github.com/Azure/sonic-buildimage/files/5146748/iptables_line.number.txt)
[log_iptables .txt](https://github.com/Azure/sonic-buildimage/files/5146749/log_iptables.txt)
[old caclmgrd file taken from git .txt](https://github.com/Azure/sonic-buildimage/files/5146750/old.caclmgrd.file.taken.from.git.txt)
[caclmgrd_comes with onic hash f6a8678d.txt](https://github.com/Azure/sonic-buildimage/files/5146751/caclmgrd_comes.with.onic.hash.f6a8678d.txt)

anshuv-mfst:
Issue in 2019 release, causing degradation in performance.

jleveque:
@abdosi: It appears this regression occurred with the introduction of multi-ASIC support in caclmgrd. Can you please investigate?

abdosi:
@jleveque It is not with multi-asic changes. Issue is coming with "caclmgrd_comes with onic hash f6a8678d.txt" which does not have multi-asic changes.

It is working with "old caclmgrd file taken from git .txt" which is pretty old colde where we have different thread to do processing.

jleveque:
[...]

PR #5312, opened 2020-09-03
[caclmgrd] Prevent unnecessary iptables updates

jleveque:
**- Why I did it**

When a large number of changes occur to the ACL table of Config DB, caclmgrd will get flooded with notifications, and previously, it would regenerate and apply the iptables rules for each change, which is unnecessary, as the iptables rules should only get applied once after the last change notification is received. If the ACL table contains a large number of control plane ACL rules, this could cause a large delay in caclmgrd getting the rules applied.

This patch causes caclmgrd to delay updating the iptables rules until it has not received a change notification for at least 0.5 seconds.

Resolves https://github.com/Azure/sonic-buildimage/issues/5275

**- How I did it**

When an ACL table change notification is received for a particular namespace, we spawn an update thread for that namespace, which sits in a loop, sleeps 0.5 seconds and checks if any new change notifications were received while it was asleep. If new notification were received, it goes back to sleep and repeats the process until no new notifications were received while it was asleep. It then generates and applies the iptables rules.

**- How to verify it**

Load a large set of control plane ACL rules (see https://github.com/Azure/sonic-buildimage/issues/5275 for an example)

- [ ] 201811
- [ ] 201911
- [ ] 202006

This PR will not cleanly cherry-pick backwards. Separate PRs will be necessary.

### DIFF
--- a/files/image_config/caclmgrd/caclmgrd
+++ b/files/image_config/caclmgrd/caclmgrd
@@ -15,6 +15,8 @@ try:
     import os
     import subprocess
     import sys
+    import threading
+    import time
 
     from sonic_py_common import daemon_base, device_info
     from swsscommon import swsscommon
@@ -75,8 +77,20 @@ class ControlPlaneAclManager(daemon_base.DaemonBase):
         }
     }
 
+    UPDATE_DELAY_SECS = 0.5
+
     def __init__(self, log_identifier):
         super(ControlPlaneAclManager, self).__init__(log_identifier)
+
+        # Update-thread-specific data per namespace
+        self.update_thread = {}
+        self.lock = {}
+        self.num_changes = {}
+
+        # Initialize update-thread-specific data for default namespace
+        self.update_thread[''] = None
+        self.lock[''] = threading.Lock()
+        self.num_changes[''] = 0
 
         SonicDBConfig.load_sonic_global_db_config()
         self.config_db_map = {}
@@ -88,8 +102,13 @@ class ControlPlaneAclManager(daemon_base.DaemonBase):
         self.namespace_mgmt_ipv6 = self.get_namespace_mgmt_ipv6(self.iptables_cmd_ns_prefix[''], '')
         self.namespace_docker_mgmt_ip = {}
         self.namespace_docker_mgmt_ipv6 = {}
+
         namespaces = device_info.get_all_namespaces()
         for front_asic_namespace in namespaces['front_ns']:
+            self.update_thread[front_asic_namespace] = None
+            self.lock[front_asic_namespace] = threading.Lock()
+            self.num_changes[front_asic_namespace] = 0
+
             self.config_db_map[front_asic_namespace] = ConfigDBConnector(use_unix_socket_path=True, namespace=front_asic_namespace)
             self.config_db_map[front_asic_namespace].connect()
             self.iptables_cmd_ns_prefix[front_asic_namespace] = "ip netns exec " + front_asic_namespace + " "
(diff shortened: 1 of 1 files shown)


jleveque:
@Ezrickd: Please test this patch and verify that it resolves https://github.com/Azure/sonic-buildimage/issues/5275

jleveque:
Retest mellanox please

Ezrickd:
Hi Joe.
I have tested the new caclmgrd config, it creates additional rules which is expected, but it takes ~45 sec for the rules to be applied.
true the issue of fluctuated number is solved by installation of 360 rules take 45 sec ... 

when I run the old caclmgrd which I verify with, the install take ~ 5 sec for 345 Rules. 

where does the long delay is inserted, this create a security as every time you will run config reload the switch is vulnerable for long seconds.

I have supplied the old caclmgrd config i am using today.

please advise.

/Ezrick

Ezrickd:
Hi Joe.
here is the device full config_db.json file 
thank you for the help.
/Ezrick 


[4700-2-config_db.zip](https://github.com/Azure/sonic-buildimage/files/5214437/4700-2-config_db.zip)

jleveque:
Retest vsimage please

jleveque:
Retest vsimage please

liat-grozovik (changes requested):
the fix has a major issue with performance as can be seen with the comments by Ezrik. 
we dont know which version it started but as can be seen the acl manager version of working/not working has been provided.

solution should be something that a user can use with reasonable time. 45 seconds for 350 ACLs is not something that we should have.

jleveque:
@liat-grozovik:  According to @Ezrickd, this patch will help eliminate fluctuation in the time. This fix helps prevent unnecessary iptables updates, so I believe it should be merged. However, there is apparently another underlying cause for the delay that @Ezrickd reported. It appears that notifications are "trickling in" to the application from the underlying Redis connection (~10 per second), rather that being received in a large batch. The root cause of the delay appears to be at a lower layer here.

jleveque:
Retest vsimage please

lguohan:
> It appears that notifications are "trickling in" to the application from the underlying Redis connection (~10 per second), rather that being received in a large batch. The root cause of the delay appears to be at a lower layer here.

do we have this issue opened somewhere?

jleveque:
@lguohan: I have opened an issue here: https://github.com/Azure/sonic-swss-common/issues/393

jleveque:
Retest vsimage please

jleveque:
Retest vsimage please

jleveque:
Retest mellanox please

jleveque:
Retest this please

jleveque:
Retest mellanox please

jleveque:
Retest mellanox please

files/image_config/caclmgrd/caclmgrd
```diff
@@ -495,9 +520,44 @@ class ControlPlaneAclManager(daemon_base.DaemonBase):
...
+            with self.lock[namespace]:
+                if self.num_changes[namespace] > num_changes:
+                    # More ACL table changes occurred since this thread was spawned
+                    # spawn a new thread with the current number of changes
+                    new_changes = self.num_changes[namespace] - num_changes
+                    self.log_info("ACL config not stable for namespace '{}': {} changes detected in the past {} seconds. Skipping update ..."
+                             .format(namespace, new_changes, self.UPDATE_DELAY_SECS))
+                    num_changes = self.num_changes[namespace]
+                else:
+                    if num_changes == self.num_changes[namespace] and num_changes > 0:
```
abdosi, line 546: @jleveque do we need num_changes > 0 check ?

jleveque: It's basically a safeguard, in case `num_changes == 0` and  `self.num_changes[namespace] == 0`. We shouldn't get here in this situation, but it we do we will hit the else case and log an error message. It could be helpful for debugging.

files/image_config/caclmgrd/caclmgrd
```diff
@@ -540,14 +600,23 @@ class ControlPlaneAclManager(daemon_base.DaemonBase):
         # Loop on select to see if any event happen on config db of any namespace
         while True:
             ctrl_plane_acl_notification = set()
+
+            # A brief sleep appears necessary in this loop or any spawned
+            # update threads will get stuck. Appears to be due to the sel.select() call.
+            # TODO: Eliminate the need for this sleep.
+            time.sleep(0.1)
```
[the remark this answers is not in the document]
jleveque: I cannot explain the root cause of the need for this sleep. Adjusting SELECT_TIMEOUT_MS doesn't make a difference. It seems to be related to some internal Python context switching in conjunction with the `sel.select()` call. If we don't sleep a tiny bit here, any spawned update threads will not make progress. It's very strange.

abdosi (approved):
LGTM

PR #5314, opened 2020-09-04
Update bcm soc property bcm_num_cos from 8 to 10

abdosi:
Why I did:
Update bcm soc property bcm_num_cos from 8 to 10
as needed by SAI 3.7 and above. Without this change
Warmboot fails from 3.5 to 3.7 as Braodcom Data structure
gets corrupted after warm-boot with in 3.5 memory allocation happening based
on 10 queue(hard-coded in 3.5 SAI) bit in 3.7+ using this property which if 8 causes corruption.

How I verify:-
Verified after this change Warmboot from 3.5 to 3.7 is fine
Also TD2 based device basic sanity looks good.


yxieca (approved):

abdosi:
Sorry by mistake created PR directly . Will delete the branch after merge

PR #5327, opened 2020-09-05
Allow user to silence variable dump during make

Staphylo:
**- Why I did it**

Most of the devs know how they configured their build.
Instead of generating a huge header for every make operation, allow the
user to silence it via an environment variable.

Also add SONIC_BUILDER_EXTRA_ENV variable which allows the user to
inject arbitrary options to the docker run cmdline for the builder as part of tooling.

**- How I did it**

Just added a few ifdef in some Makefiles.

**- How to verify it**

Set the SONIC_BUILD_QUIETER environment variable with any value to prevent
make from displaying all the build options.
```
SONIC_BUILD_QUIETER=y make ...

bash% SONIC_BUILD_QUIETER=1 NOJESSIE=1 NOSTRETCH=1 make configure PLATFORM=broadcom
+++ Making configure +++
BLDENV=buster make -f Makefile.work configure
make[1]: Entering directory `/home/staphylo/sonic-buildimage'
SONiC Build System for broadcom:amd64
make[1]: Leaving directory `/home/staphylo/sonic-buildimage'
```

To export 2 environment variables with this mechanism, one can run the following.
```
SONIC_BUILDER_EXTRA_CMDLINE="-e test=var -e var='with spaces'" make ...
```
This is especially useful for tooling.

- [ ] 201811
- [ ] 201911
- [ ] 202006

Allow user to silence variable dump during make and to inject docker run variables

### DIFF
--- a/slave.mk
+++ b/slave.mk
@@ -189,6 +189,7 @@ export FRR_USER_GID
 ## Dumping key config attributes associated to current building exercise
 ###############################################################################
 
+ifndef SONIC_BUILD_QUIETER
 $(info SONiC Build System)
 $(info )
 $(info Build Configuration)
@@ -232,6 +233,9 @@ $(info "INCLUDE_KUBERNETES"              : "$(INCLUDE_KUBERNETES)")
 $(info "TELEMETRY_WRITABLE"              : "$(TELEMETRY_WRITABLE)")
 $(info "ENABLE_SYNCHRONOUS_MODE"         : "$(ENABLE_SYNCHRONOUS_MODE)")
 $(info )
+else
+$(info SONiC Build System for $(CONFIGURED_PLATFORM):$(CONFIGURED_ARCH))
+endif
 
 include Makefile.cache
 
--- a/Makefile.cache
+++ b/Makefile.cache
@@ -89,11 +89,13 @@ endif
 -include $(PLATFORM_PATH)/rules.dep
 endif
 
+ifndef SONIC_BUILD_QUIETER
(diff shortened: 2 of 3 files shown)


jleveque:
Looks good to me.

slave.mk
```diff
@@ -232,6 +233,9 @@ $(info "INCLUDE_KUBERNETES"              : "$(INCLUDE_KUBERNETES)")
 $(info "TELEMETRY_WRITABLE"              : "$(TELEMETRY_WRITABLE)")
 $(info "ENABLE_SYNCHRONOUS_MODE"         : "$(ENABLE_SYNCHRONOUS_MODE)")
 $(info )
+else
+$(info SONiC Build System for $(CONFIGURED_PLATFORM):$(CONFIGURED_ARCH))
```
yxieca, line 237: Why is this line in else not outside if block?

yxieca (approved):
Minor suggestion/question

slave.mk
Staphylo: I felt it was still useful for SONIC_BUILD_QUIETER to print that it entered SONiC build system for debug purposes.
The information I added to the simple `SONiC Build System` in this line is redundant with what is in the ifdef block above. Since I introduced this one liner I felt it was meaningful to print the targeted platform and arch.
Both `CONFIGURED_PLATFORM` and `CONFIGURED_ARCH` have their dedicated line in the configuration dump above.
I would prefer keeping things as is but I'm open to suggestions.

PR #5338, opened 2020-09-08
[ssh limit] add infrastructure to set per user login limit

yxieca:
**- Why I did it**
We have an use case to limit max number of login per user.

**- How I did it**
Specifying per user login limit via USER_LOGIN_LIMIT build argument.

When not specified, the per user login is unlimited by default.

**- How to verify it**
Verified in 2 aspects:
1. limit working on DUT: set the limit to 2 and validated that the 3rd login is rejected from ssh and/or console.
2. build test: without specifying the limit, /etc/security/limits.conf didn't change. Build with option SSH_USER_LOGIN_LIMIT=2, and verified that the configuration changed accordingly.

- [x] 201811
- [x] 201911
- [x] 202006

### DIFF
--- a/rules/config
+++ b/rules/config
@@ -112,7 +112,6 @@ SONIC_DPKG_CACHE_SOURCE ?= /var/cache/sonic/artifacts
 # Default VS build memory preparation
 DEFAULT_VS_PREPARE_MEM = yes
 
-
 # INCLUDE_SYSTEM_TELEMETRY - build docker-sonic-telemetry for system telemetry support
 INCLUDE_SYSTEM_TELEMETRY = y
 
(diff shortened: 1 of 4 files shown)


rules/config
```diff
@@ -161,3 +160,6 @@ SONIC_ENABLE_IMAGE_SIGNATURE ?= n
 
 # ENABLE_SYNCHRONOUS_MODE - enable synchronous mode between orchagent and syncd
 ENABLE_SYNCHRONOUS_MODE = n
+
+# Default SSH per user login limit (unlimited)
+DEFAULT_USER_LOGIN_LIMIT = 0
```
jleveque, line 165: Is this tested and working? According to the [man page](https://man7.org/linux/man-pages/man5/limits.conf.5.html), it appears that `-1` is the numerical representation of unlimited:

> All items support the values -1, unlimited or infinity indicating no limit, except for priority and nice.

PR #5340, opened 2020-09-08
[redis] Use redis-server and redis-tools in blob storage to prevent upstream link broken

qiluo-msft:
**- Why I did it**
Upstream link broken and the build is failed

**- How I did it**
Use redis-server and redis-tools in blob storage

**- How to verify it**

- [ ] 201811
- [ ] 201911
- [x] 202006

### DIFF
--- a/dockers/docker-base-buster/Dockerfile.j2
+++ b/dockers/docker-base-buster/Dockerfile.j2
@@ -53,6 +53,12 @@ RUN apt-get update &&        \
 # Install dependencies of supervisor
         python-pkg-resources \
         python-meld3         \
+# Install dependencies of redis-tools
+        libatomic1           \
+        libjemalloc2         \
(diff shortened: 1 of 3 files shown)


yxieca:
Is this change compatible with 201911/201811 branch? Do we need to change these 2 branches?

qiluo-msft:
No. I added proper label.

jleveque (approved):
Should we consider upgrading to 6.0.7, which is the latest release available for Buster backports? Maybe missing some bugfixes otherwise?

yxieca (approved):

qiluo-msft:
@jleveque that will be a long term goal. Here I am trying to fix the build.

renukamanavalan (commented):
Would it be feasible to declare this redis version in one place and refer from everywhere ?

PR #5359, opened 2020-09-11
[docker_img_ctl][multi-asic]: Remove existing database docker upon reboot if database_global.json does not exist

SuvarnaMeenakshi:
**- Why I did it**
Currently multi-asic image is built as a single asic image. To convert single asic VS to multi-asic, asic.conf file is updated with number of asics and vs is rebooted. Upon reboot, vs will come back up as multi-asic VS.
In single asic image, database_global.json file is not created in the host(or global) database docker. Hence during conversion, to avoid using the previous single asic database docker, the host database docker should be removed.

**- How I did it**
For multi-asic image, database_global.json file does not exist, remove existing database docker so that new docker gets created with database_global.json.

**- How to verify it**
This will not affect single asic image.
This change will take effect only if the first boot was like a single asic and next boot was done like multi-asic, which currently is done only on multi-asic vs. 
Therefore, verified that single asic to multi-asic VS image is successfully done and host database docker and N database dockers come up where N is the number of asics.

- [ ] 201811
- [X] 201911
- [ ] 202006

### DIFF
--- a/files/build_templates/docker_image_ctl.j2
+++ b/files/build_templates/docker_image_ctl.j2
@@ -184,8 +184,12 @@ start() {
     # If it does not exist, remove the current docker so that the file is created.
     # This is mainly to ensure that database_global.json file is created when converting
     # single asic VS to multi-asic VS.
-    if [[ ($NUM_ASIC -gt 1) && (-z $DEV) && (! -f "/var/run/redis/sonic-db/database_global.json") ]]; then
-        docker rm -f {{docker_container_name}}$DEV
+    SONIC_DB_GLOBAL_JSON="/var/run/redis/sonic-db/database_global.json"
+    if [[ ($NUM_ASIC -gt 1) && (-z $DEV) && (! -f $SONIC_DB_GLOBAL_JSON) ]]; then
+       DOCKERCHECK=`docker inspect --type container {{docker_container_name}}$DEV 2>/dev/null`
+       if [ "$?" -eq "0" ]; then
+           docker rm -f {{docker_container_name}}$DEV
+       fi
     fi
     {%- endif %}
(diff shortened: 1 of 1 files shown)


files/build_templates/docker_image_ctl.j2
```diff
@@ -179,7 +179,15 @@ start() {
     # Obtain our HWSKU as we will mount directories with these names in each docker
     HWSKU=${HWSKU:-`$SONIC_CFGGEN -d -v 'DEVICE_METADATA["localhost"]["hwsku"]'`}
     {%- endif %}
-
+    {%- if docker_container_name == "database" %}
+    # Ensure that database_global.json file exists in database docker for multi-asic platform.
+    # If it does not exist, remove the current docker so that the file is created.
+    # This is mainly to ensure that database_global.json file is created when converting
+    # single asic VS to multi-asic VS.
+    if [[ ($NUM_ASIC -gt 1) && (-z $DEV) && (! -f "/var/run/redis/sonic-db/database_global.json") ]]; then
```
arlakshm, line 187: Can we used the  constant SONIC_DB_GLOBAL_JSON instead of the filename. SONIC_DB_GLOBAL_JSON used in other places in this file

arlakshm (changes requested):

files/build_templates/docker_image_ctl.j2
SuvarnaMeenakshi: Updated as suggessted.

judyjoseph:
LGTM, please resolve the conflicts.

SuvarnaMeenakshi:
This change is only required to convert single asic vs to multi-asic vs, this is not required if the sonic-vs image is pre-built with NUM_ASIC updated in asic.conf. Hence, closing this PR as we will go with the pre-built approach to match with an actual hardware scenario.

PR #5364, opened 2020-09-12
[Multi-Asic] Fix for multi-asic where we should allow namespace docker local communication on its docker/eth0 ip

abdosi:
Why/What i did:
Fix for multi-asic where we should allow namespace docker local communication on its docker/eth0 ip
. Without this TCP Connection to Redis Server running on docker eth0 ip does not happen in namespace if there is Catch all DROP Rule.

Found this issue as part of debugging
a)  lldpmgrd was not able to start in namespace (uses Redis TCP Connection)
b) Feature auto restart not working for namespace docker (supervisor-proc-exit-listener also  uses Redis TCP Connection)

How I verify:

After fix both issue (a) and (b) are working fine.

### DIFF
--- a/files/image_config/caclmgrd/caclmgrd
+++ b/files/image_config/caclmgrd/caclmgrd
@@ -200,6 +200,10 @@ class ControlPlaneAclManager(object):
         allow_internal_docker_ip_cmds = []
 
         if namespace:
+            # For namespace docker allow local communication on docker management ip for all proto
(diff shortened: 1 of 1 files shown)


jleveque (approved):

judyjoseph (approved):

lguohan:
do we need to modify the control plane acl test to adapt this change?

abdosi:
> do we need to modify the control plane acl test to adapt this change?

Yes @lguohan will update(to-do work item)  caclmgr test case as of part of multi-asic test case's enhancement.

lguohan (approved):

PR #5375, opened 2020-09-15
[Mellanox] Always restart thermalctld on Mellanox platform when it exits

Junchao-Mellanox:
**- Why I did it**

On mellanox paltform, part of thermalctld function is to handle user space thermal policies for events like fan/PSU removing, it works together with kernel thermal algorithm to make sure the switch won't be overheat.

Recently, we found that commit https://github.com/Azure/sonic-buildimage/commit/cbc75fe4c85fd01aded0bacb283c8bbf62adf2f8 changes its autorestart configuration in supervisord, and it won't be auto restarted after being killed. This PR is to make sure that thermalctld will be always restarted on mellanox platform when it is killed.

**- How I did it**

1. Add a variable "always_restart_thermalctld" in pmon_daemon_control.json
2. In docker-pmon.supervisord.conf.j2, it checks variable "always_restart_thermalctld" and set autorestart configuration for thermalctld accordingly.

**- How to verify it**

Manual test

Depends on where https://github.com/Azure/sonic-buildimage/commit/cbc75fe4c85fd01aded0bacb283c8bbf62adf2f8 is going to merge to. 

- [ ] 201811
- [ ] 201911
- [ ] 202006

### DIFF
--- a/device/mellanox/x86_64-mlnx_msn2700-r0/pmon_daemon_control.json
+++ b/device/mellanox/x86_64-mlnx_msn2700-r0/pmon_daemon_control.json
@@ -1,5 +1,6 @@
 {
     "skip_ledd": true,
-    "skip_fancontrol": true
+    "skip_fancontrol": true,
+    "always_restart_thermalctld": true
 }
 
--- a/dockers/docker-platform-monitor/docker-pmon.supervisord.conf.j2
+++ b/dockers/docker-platform-monitor/docker-pmon.supervisord.conf.j2
@@ -120,7 +120,11 @@ dependent_startup_wait_for=start:exited
 command=/usr/local/bin/thermalctld
 priority=9
 autostart=false
+{% if not always_restart_thermalctld %}
 autorestart=unexpected
+{% else %}
+autorestart=true
+{% endif %}
 stdout_logfile=syslog
 stderr_logfile=syslog
 startsecs=10


jleveque:
What is wrong with `autorestart=unexpected`? Under what circumstances the daemon will exit "expectedly" according to supervisor and thus not get restarted?

Junchao-Mellanox:
For example, user issue an kill command in pmon docker, it will send a terminate signal to thermalctld, thermalctld handls such signal and will clear all resources and exit with code 0. This won't trigger an auto restart.

jleveque:
> For example, user issue an kill command in pmon docker, it will send a terminate signal to thermalctld, thermalctld handls such signal and will clear all resources and exit with code 0. This won't trigger an auto restart.

Certainly, this makes sense. However, this is not "normal operation". If we are this concerned, then I would prefer reverting back to the previous solution (i.e., not creating a new variable/option -- make it `autorestart=true` always. However, I must ask, if we are concerned about user error, what would prevent a user from stopping the PMon container altogether?

Junchao-Mellanox:
Actually, we have regression test case which will kill thermalctld and wait for it to auto restart, and this case failed on master. We can also change the test case.

jleveque (changes requested):
I prefer not introducing a new variable here. We should decide on a universal solution.

Junchao-Mellanox:
How about we put a exit(-1) to the end of thermalctld?

jleveque:
> How about we put a exit(-1) to the end of thermalctld?

As long as we have a non-zero exit code at every exit path, there should be no problem. FYI, exit codes in Linux are non-negative short integers, so `-1` will get converted to `255`. Prefer using a positive value so that it's easy to track down.

Junchao-Mellanox:
@jleveque Could you please take a look at https://github.com/Azure/sonic-platform-daemons/pull/94? With this new PR, we can drop the current PR.

jleveque:
Closing in favor of https://github.com/Azure/sonic-platform-daemons/pull/94

PR #5400, opened 2020-09-18
Updated sai.profile for Broadcom platform to pass SAI_KEY_NUM_ECMP_MEMBERS

abdosi:
Why I did it:-
In Broadcom SAI 3.5 by default we are supporting 256 Group with 64 Member each.
However in SAI 3.7 default behavior got changed to 128 Group and 128
Member each.

This change is to make sure we are using same ECMP Group/Member Per
Group for 3.7 also so that behavior is consistent. 

This change applies only to platforms where SAI_KEY_NUM_ECMP_MEMBERS was not passed explicitly.

How I did:-
Using SAI Key,Value Profile SAI_KEY_NUM_ECMP_MEMBERS to pass number of ecmp members per group.

How I verify:-
Verified on Arista 7260cx3

admin@str-7260cx3-acs-1:~$ crm show resources all
Resource Name           Used Count    Available Count
--------------------  ------------  -----------------
nexthop_group                    2                254

admin@str-7260cx3-acs-1:~$ redis-cli -n 6
127.0.0.1:6379[6]> hgetall "SWITCH_CAPABILITY|switch"
 1) "MAX_NEXTHOP_GROUP_COUNT"
 2) "256"


prsunny (approved):

prsunny:
lgtm, @lguohan to confirm

abdosi:
> lgtm, @lguohan to confirm

@lguohan Can you please confirm

abdosi:
@lguohan can you please confirm.

lguohan (approved):

PR #5407, opened 2020-09-18
[mgmt] Fix Azure CLI install behind proxy

daall:

**- Why I did it**
`sudo bash` doesn't work properly behind a proxy because the environment variables are lost, and `apt` is used in the installation script, so we can't run it in user mode.

**- How I did it**
Move the installation before we switch context to $user so that we have access to the proxy environment variables and can run the install script without sudo.

**- How to verify it**
Verified that `az` commands work inside the container. Ran w/ proxy configured and build succeeds.

- [ ] 201811
- [ ] 201911
- [ ] 202006


lguohan (approved):

yxieca (approved):

lguohan:
if sudo bash is the issue, you can use sudo -E bash which will preserve the environment variables.

PR #5438, opened 2020-09-23
[bash.bashrc] Add reverse SSH script to bash.bashrc

Blueve:
**- Why I did it**

To support reverse SSH for console connection, we need to add the script to attach to console line automatically after user login with reverse SSH.
(see HLD for more detail: https://github.com/Azure/SONiC/blob/master/doc/console/SONiC-Console-Switch-High-Level-Design.md#313-reverse-ssh)

The script will be executed only when this feature is available in SONiC: https://github.com/Blueve/openssh-portable/pull/1

**- How I did it**

Add script to `/etc/bash.bashrc` to support reverse ssh function.

**- How to verify it**

**Functional Test**
- [x] Test similar script on my WSL ubuntu (`sudo picocom` instead of `sudo connect`)
- [x] Test on Physical DUT with dummy environment variable
- [x] Build image with this change and test image on Physical DUT

- [ ] 201811
- [ ] 201911
- [ ] 202006

image
_My first pet. My cutest boy. My favorite friend, My family. My cat, the cat I'll never to be seen again... R.I.P Cloud·Cake (2020-09-24)_

### DIFF
--- a/files/image_config/bash/bash.bashrc
+++ b/files/image_config/bash/bash.bashrc
@@ -58,9 +58,9 @@ fi
 tty | grep ttyS >/dev/null && TMOUT=900
 
 # if SSH_TARGET_CONSOLE_LINE was set, attach to console line interactive cli directly
-if [ ! -z "$SSH_TARGET_CONSOLE_LINE" ]; then
+if [ -n "$SSH_TARGET_CONSOLE_LINE" ]; then
     # enter the interactive cli
-    sudo connect $SSH_TARGET_CONSOLE_LINE
+    sudo connect line $SSH_TARGET_CONSOLE_LINE
 
     # exit after console session ended
     exit
--- a/files/image_config/bash/bash.bashrc
+++ b/files/image_config/bash/bash.bashrc
@@ -56,3 +56,12 @@ fi
 
 # Automatically log out console ttyS* sessions after 15 minutes of inactivity
 tty | grep ttyS >/dev/null && TMOUT=900
+
+# if SSH_TARGET_CONSOLE_LINE was set, attach to console line interactive cli directly
+if [ ! -z "$SSH_TARGET_CONSOLE_LINE" ]; then
+    # enter the interactive cli
+    sudo connect $SSH_TARGET_CONSOLE_LINE
+
+    # exit after console session ended
+    exit
+fi
--- a/files/image_config/bash/bash.bashrc
+++ b/files/image_config/bash/bash.bashrc
@@ -59,9 +59,17 @@ tty | grep ttyS >/dev/null && TMOUT=900
 
 # if SSH_TARGET_CONSOLE_LINE was set, attach to console line interactive cli directly
 if [ -n "$SSH_TARGET_CONSOLE_LINE" ]; then
-    # enter the interactive cli
-    sudo connect line $SSH_TARGET_CONSOLE_LINE
+    if [ $SSH_TARGET_CONSOLE_LINE -eq $SSH_TARGET_CONSOLE_LINE 2>/dev/null ]; then
+        # enter the interactive cli
+        connect line $SSH_TARGET_CONSOLE_LINE
 
-    # exit after console session ended
-    exit
+        # test exit code, 1 means the console switch feature not enabled
+        if [ $? -ne 1 ]; then
+            # exit after console session ended
+            exit
+        fi
+    else
+        # exit directly when target console line variable is invalid
+        exit
+    fi
 fi


jleveque (changes requested):

Blueve:
retest please

files/image_config/bash/bash.bashrc
```diff
@@ -56,3 +56,13 @@ fi
 
 # Automatically log out console ttyS* sessions after 15 minutes of inactivity
 tty | grep ttyS >/dev/null && TMOUT=900
+
+# if SSH_TARGET_CONSOLE_LINE was set, attach to console line interactive cli directly
+if [ ! -z "$SSH_TARGET_CONSOLE_LINE" ]; then
```
Blueve, line 61: use `-n` instead

yxieca (dismissed):

Blueve:
Hi @jleveque, I have updated the PR. Could you please take a look?

files/image_config/bash/bash.bashrc
```diff
@@ -56,3 +56,12 @@ fi
 
 # Automatically log out console ttyS* sessions after 15 minutes of inactivity
 tty | grep ttyS >/dev/null && TMOUT=900
+
+# if SSH_TARGET_CONSOLE_LINE was set, attach to console line interactive cli directly
+if [ -n "$SSH_TARGET_CONSOLE_LINE" ]; then
```
lguohan, line 61: i think we need to make it optional. for example, if the device is not c0 device, then we must not act on this variable even if it is not empty.

also, we need to verify the content this variable to make sure it is valid. this potentially allow hack to modify the variable on a shell.

files/image_config/bash/bash.bashrc
```diff
@@ -56,3 +56,12 @@ fi
...
+    # enter the interactive cli
+    sudo connect line $SSH_TARGET_CONSOLE_LINE
```
lguohan, line 63: what is the user does not have sudo privilege? what will happen here?

Blueve: A short error message will display:
```
<user_name> is not in the sudoers file.  This incident will be reported.
```
And then the session will close.

Do we need test if the user is in sudoer list before run this command?

files/image_config/bash/bash.bashrc
Blueve: For the first comment:
I think we can made a build feature flag for console switch feature. 
In my view, this feature is not platform dependent. e.g. A E1031 can be both a bare management switch (without console hub) and console switch(plugged a console hub).
Only user can decide if they want to use it or not. Let me know your thought...

For the second comment:
Agree. We should only accept integer here. I will update.

lguohan: it is hard to make it as a build flag. imo, it should depends on the device role, for example mgmttorrouter, or in config db, we can add a switch metadata to see where console switch is enabled or not.

files/image_config/bash/bash.bashrc
lguohan: i think normal user should be able to login via the console line, should not need sudo user.

Blueve: No. The connect command require admin privilege because it need to talk with \dev\ttyUSB*

files/image_config/bash/bash.bashrc
Blueve: I prefer use config db. Add a key to enable/disable console switch features. And user can also use cli `config console {on/off}` to config that.
For minigraph user, we can simply enable console switch by judging if there is a console port.

If we do so, we can keep this code for a while. I will raise a new PR to update the script after config part done.

files/image_config/bash/bash.bashrc
yxieca: Can we fix that somehow? Current console access doesn't require elevated privilege. I think.

Blueve: We need to add user to group `dialout` and full reboot.
`sudo adduser <user> dialout`

Or apply UDEV rule:
https://askubuntu.com/questions/112568/how-do-i-allow-a-non-default-user-to-use-serial-device-ttyusb0
And set the usb device file to 666

I prefer udev rule because we already have one for console ports and we can control which device can expose without admin privilege need

Blueve:
@yxieca @lguohan I've update the script.

Now it will exit directly when:
* The SSH_TARGET_CONSOLE_LINE is not a number
* The connect command returns a failure that the user input is invalid

And it will enter the session when:
* The console switch feature is not enabled

yxieca (approved):

PR #5450, opened 2020-09-24
[Centec-ARM64] Update docker-syncd-centec critical_processes file to use new syntax

jleveque:
This file was added recently, but was created using the old syntax. Update to the new syntax.


jleveque:
@taocy001 for awareness.

lguohan (approved):

PR #5463, opened 2020-09-24
[caclmgrd] Provide empty list of destination ports for 'ANY' protocol

jleveque:
**- Why I did it**

For the new "ANY" protocol, instead of using the reserved port "0" as a destination port, simply provide an empty list of destination ports. This way, no `--dport` arguments will be added to the command, effectively applying the rule to all ports without the need for a "magic number".

- [ ] 201811
- [ ] 201911
- [ ] 202006


jleveque:
@venkatmahalingam: Please review.

jleveque:
Closing, as it doesn't work. :(

PR #5468, opened 2020-09-25
[Mellanox] Refactor platform API to remove dependency on database

keboliu:
***This PR has a dependency on https://github.com/Azure/sonic-platform-common/pull/121***

**- Why I did it**
- Platform API implementation using sonic-cfggen to get platform name and SKU name, which will fail when the database is not available.
- Chassis name is not correctly assigned, it shall be assigned with EEPROM TLV "Product Name", instead of SKU name  
- Chassis model is not implemented, it shall be assigned with EEPROM TLV "Part Number"

**- How I did it**

1. Chassis

> - Get platform name from /host/machine.conf
> - Remove get SKU name with sonic-cfggen
> - Get Chassis name and model from EEPROM TLV "Product Name" and "Part Number" 
> - Add function to return model

2. EEPROM

> - Add function to return product name and part number

3. Platform

> - Init EEPROM on the host side, so also can get the Chassis name model from EEPROM on the host side.

**- How to verify it**

- Check whether platform API can still work when stop database service
- Check whether daemons calling platform API are still work well.
- Check whether Chassis name and model are as expected.

- [ ] 201811
- [x] 201911
- [ ] 202006

### DIFF
--- a/platform/mellanox/mlnx-platform-api/sonic_platform/chassis.py
+++ b/platform/mellanox/mlnx-platform-api/sonic_platform/chassis.py
@@ -18,7 +18,6 @@ try:
     import sys
     import io
     import re
-    import subprocess
     import syslog
 except ImportError as e:
     raise ImportError (str(e) + "- required module not found")
@@ -26,8 +25,6 @@ except ImportError as e:
 MAX_SELECT_DELAY = 3600
 
 MLNX_NUM_PSU = 2
-
-GET_PLATFORM_CMD = "sonic-cfggen -d -v DEVICE_METADATA.localhost.platform"
 
 EEPROM_CACHE_ROOT = '/var/cache/sonic/decode-syseeprom'
 EEPROM_CACHE_FILE = 'syseeprom_cache'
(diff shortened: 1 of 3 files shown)


jleveque:
https://github.com/Azure/sonic-platform-common/pull/121 has merged. Submodule update PR here: https://github.com/Azure/sonic-buildimage/pull/5470

platform/mellanox/mlnx-platform-api/sonic_platform/chassis.py
```diff
@@ -27,7 +27,6 @@
 
 MLNX_NUM_PSU = 2
 
-GET_HWSKU_CMD = "sonic-cfggen -d -v DEVICE_METADATA.localhost.hwsku"
 GET_PLATFORM_CMD = "sonic-cfggen -d -v DEVICE_METADATA.localhost.platform"
```
sujinmkang, line 30: You can remove this line too, right? Since this command is not used any more either.

keboliu: fixed.

jleveque (approved):

PR #5471, opened 2020-09-27
[Mellanox] Skip pcied on mellanox platform

Junchao-Mellanox:
**- Why I did it**

pcied is not supported on mellanox platform and may cause regression failure.

**- How I did it**

Skip pcied in pmon_daemon_control.json

**- How to verify it**

Manual test

- [ ] 201811
- [x] 201911
- [ ] 202006

### DIFF
--- a/device/mellanox/x86_64-mlnx_msn2700-r0/pmon_daemon_control.json
+++ b/device/mellanox/x86_64-mlnx_msn2700-r0/pmon_daemon_control.json
@@ -1,5 +1,6 @@
 {
     "skip_ledd": true,
-    "skip_fancontrol": true
+    "skip_fancontrol": true,
+    "skip_pcied": true
 }
 


lguohan:
please supply data to support pcied.

lguohan (changes requested):
i do not think we should skip, instead we should add needed data to enable the check.

liat-grozovik:
@lguohan as this is new req which is not yet supported and should be added to all Mellanox switches we should skip it for now and once we deliver the info for all supported platforms we should enable it again.

sujinmkang:
@Junchao-Mellanox Since the pcie configuration file is not located in the expected platform directory, then pcied will just exit. Did you see any problem with the pcied?

Junchao-Mellanox:
> @Junchao-Mellanox Since the pcie configuration file is not located in the expected platform directory, then pcied will just exit. Did you see any problem with the pcied?

It fails the reboot test due to check_pmon_daemon_status report that pcied is not there.

Junchao-Mellanox:
Close as Kebo will add a configuration file for pcied.

PR #5474, opened 2020-09-28
[cfggen] Build Python 2 And Python 3 Wheel Packages

tahmed-dev:
This builds Python 2&3 wheel packages for sonic-cfggen script.
It also removes the natural sorting of the output as natsorted
did not work when dict has tuple object as key. The sole use of
natsorted was to make test comparison easier and so, the natsorting
logic is now relocated to the test suite. As a pleasurable
side effect, sonic-cfggen gained nearly 1 sec per call since we no
longer import natsorted module nor run!

**- Why I did it**
Help SONiC move to Python 3

**- How I did it**
Provided packaging for both Python 2 and Python 3

**- How to verify it**
All unit test passed
Also, loaded SONiC on real DUT and no issues observed.

- [ ] 201811
- [x] 201911
- [x] 202006


src/sonic-bgpcfgd/tests/data/sonic-cfggen/bgpd.conf.j2/all.conf
```diff
@@ -29,14 +33,15 @@ ip prefix-list PL_LoopbackV4 permit 55.55.55.55/32
 !
 ipv6 prefix-list PL_LoopbackV6 permit fc00::/64
 !
-ip prefix-list LOCAL_VLAN_IPV4_PREFIX seq 5 permit 10.10.10.1/24
+ip prefix-list LOCAL_VLAN_IPV4_PREFIX seq 10 permit 10.10.10.1/24
```
lguohan, line 36: is this intended?

src/sonic-bgpcfgd/tests/data/sonic-cfggen/bgpd.conf.j2/all.conf
```diff
@@ -56,10 +61,11 @@ router bgp 55555
     network fc00::1/64
   exit-address-family
 !
-  network 10.10.10.1/24
   address-family ipv6
    network fc01::1/64
   exit-address-family
+  network 10.10.10.1/24
```
lguohan, line 67: is this intended?

src/sonic-bgpcfgd/tests/data/sonic-cfggen/bgpd.spine_chassis_frontend_router.conf.j2/base.conf
```diff
@@ -1,36 +1,36 @@
 !
 ! Vnet BGP instance
-router bgp 555 vrf First
+router bgp 555 vrf Second
```
lguohan, line 3: is this intended?

src/sonic-bgpcfgd/tests/data/sonic-cfggen/bgpd.conf.j2/all.conf
tahmed-dev: The it is an artefact of the order change due to the elemination of natsorted. The [bgpd.conf.j2](https://github.com/Azure/sonic-buildimage/blob/master/dockers/docker-fpm-frr/frr/bgpd/bgpd.main.conf.j2#L22) user the loop index and multiplies it by 5 and so the seq would depend on the index of the IP address

src/sonic-bgpcfgd/tests/data/sonic-cfggen/bgpd.conf.j2/all.conf
tahmed-dev: this is an artefact of order change. Also, dict insertion order would depend on dict implementation on Python 2/3.

src/sonic-bgpcfgd/tests/data/sonic-cfggen/bgpd.spine_chassis_frontend_router.conf.j2/base.conf
tahmed-dev: same this due to order shuffling.

lguohan:
i can see many test data are changed, not sure why.

tahmed-dev:
> i can see many test data are changed, not sure why.

Changes in `src/sonic-bgpcfgd/tests/data/sonic-cfggen` data dir is due to the change in the sonic-cfggen not returning `natsorted` dict anymore. This causes some entries to change index/order in Python list/dict. The files changed under data directory are used to validate `sonic-cfggen` output and so, the order change of `sonic-cfggen` output required changing those files.

@pavel-shirshov are there requirements to have those sections in specific order when configuring frr/bgpd/etc.? same question to `seq` field?

src/sonic-config-engine/tests/common_utils.py
```diff
@@ -0,0 +1,33 @@
...
+        return re.sub(r"( *, *)", '|', tupleobj).replace("'", '')
+
+    return re.sub(r"(\(.*?\))", to_str, tuplestr)
+
+def to_dict(dictstr):
+    """ Convert string represention of Python dict into dict """
+    # handle tuple elements
+    dictstr = tuple_to_str(dictstr)
+
+    return json.loads(dictstr.replace("'", '"'))
```
pavel-shirshov, line 22: will python `eval()` fucntion does the job?

src/sonic-config-engine/tests/common_utils.py
```diff
@@ -0,0 +1,33 @@
...
+
+def liststr_to_dict(liststr):
+    """ Convert string represention of Python list into dict with list key and sorted liststr as ots value """
```
pavel-shirshov, line 25: as ??? value? I didn't understand it

src/sonic-config-engine/tests/multi_npu_data/py3/ipinip.json
```diff
@@ -0,0 +1,23 @@
+[
```
pavel-shirshov, line 1: Why do we have this file created here? Is it new?

src/sonic-config-engine/tests/sample_output/py2/bgpd_quagga.conf
```diff
@@ -23,19 +23,11 @@ router bgp 65100
   bgp graceful-restart restart-time 240
```
pavel-shirshov, line 23: It is hard to check this kind of changes. Is it possible to preserve the previous file order?

src/sonic-config-engine/tests/sample_output/py2/buffers-dell6100.json
```diff
@@ -0,0 +1,640 @@
+
```
pavel-shirshov, line 1: New file?

src/sonic-config-engine/tests/sample_output/py2/docker-dhcp-relay.supervisord.conf
```diff
@@ -0,0 +1,69 @@
+[supervisord]
```
pavel-shirshov, line 1: New file?

src/sonic-config-engine/tests/sample_output/py2/frr.conf
```diff
@@ -0,0 +1,87 @@
+!
```
pavel-shirshov, line 1: New file?

src/sonic-config-engine/tests/sample_output/py2/ipinip.json
```diff
@@ -0,0 +1,23 @@
+[
```
pavel-shirshov, line 1: New file?

src/sonic-config-engine/tests/sample_output/py2/l2switch.json
```diff
@@ -0,0 +1,268 @@
+{
```
pavel-shirshov, line 1: New file?

src/sonic-config-engine/tests/sample_output/py2/ports.json
```diff
@@ -0,0 +1,30 @@
+[
```
pavel-shirshov, line 1: New file?

src/sonic-config-engine/tests/sample_output/py2/t2-chassis-fe-vni-zebra.conf
```diff
@@ -0,0 +1,32 @@
+!
```
pavel-shirshov, line 1: New file?

src/sonic-config-engine/tests/sample_output/py2/t2-chassis-fe-zebra.conf
```diff
@@ -0,0 +1,32 @@
+!
```
pavel-shirshov, line 1: New file?

src/sonic-config-engine/tests/sample_output/py2/wait_for_intf.sh
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bash
```
pavel-shirshov, line 1: New file?

src/sonic-config-engine/tests/test_cfggen.py
```diff
@@ -55,7 +62,7 @@ def test_device_desc(self):
         self.assertEqual(output.strip(), 'ACS-MSN2700')
 
     def test_device_desc_mgmt_ip(self):
-        argument = '-v "MGMT_INTERFACE.keys()[0]" -M "' + self.sample_device_desc + '"'
+        argument = '-v "(MGMT_INTERFACE.keys()|list)[0]" -M "' + self.sample_device_desc + '"'
```
pavel-shirshov, line 65: Why do you need list here?

pavel-shirshov (commented):
As comments.
My opinion: This PR should be split into many. Currently it is hard to check some sonic-cfggen results because you moved the files from one place to another and the history was lost.
Also the order has been changed, it is very hard to check that nothing was broken with this change.
I'd suggest:
1. Remove natsorted and review changes
2. Add Py2+Py3 code. The code will be changed, but target files must be unchanged
3. Move Py2 target files to its own directory and add Py3 target files (why d0 we need two different set of target files at first?)
4. Make the tests better (I'd introduce pytest here)

src/sonic-config-engine/tests/test_cfggen.py
tahmed-dev: In Python 3, Jinja2 template will return dict_keys iterable object (and not index-able) and so in order to maintain the existing behavior, we cast it into list.

src/sonic-config-engine/sonic-cfggen
```diff
@@ -166,17 +168,16 @@ TODO(taoyl): Current version of config db only supports BGP admin states.
...
+            if lookup_key is not None:
                 newData = {}
                 for key in data.keys():
                     if ((type(key) is STR_TYPE and lookup_key == key) or (type(key) is tuple and lookup_key in key)):
                         newData[ConfigDBConnector.serialize_key(key)] = data.pop(key)
                         break
                 return newData
 
-            for key in data.keys():
+            current_keys = list(data.keys())
```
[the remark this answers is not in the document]
tahmed-dev: This is caching the current keys as in Python 3 when we insert into the same dict, the new keys gets appended to the end and we process them again. This cause it to crash as the new format is not expected.

src/sonic-config-engine/minigraph.py
```diff
@@ -1097,7 +1097,7 @@ def parse_xml(filename, platform=None, port_config_file=None, asic_name=None, hw
     results['VLAN'] = vlans
     results['VLAN_MEMBER'] = vlan_members
 
-    for nghbr in neighbors.keys():
+    for nghbr in list(neighbors.keys()):
```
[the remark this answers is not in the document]
tahmed-dev: It needs to be indexed and it is failing on Python 3

src/sonic-config-engine/minigraph.py
```diff
@@ -1059,7 +1059,7 @@ def parse_xml(filename, platform=None, port_config_file=None, asic_name=None, hw
     results['PORTCHANNEL'] = pcs
     results['PORTCHANNEL_MEMBER'] = pc_members
 
-    for pc_intf in pc_intfs.keys():
+    for pc_intf in list(pc_intfs.keys()):
```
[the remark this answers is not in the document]
tahmed-dev: It needs to be indexed and it is failing on Python 3

src/sonic-config-engine/minigraph.py
```diff
@@ -1044,7 +1044,7 @@ def parse_xml(filename, platform=None, port_config_file=None, asic_name=None, hw
 
     if port_config_file:
         port_set = set(ports.keys())
-        for (pc_name, mbr_map) in pcs.items():
+        for (pc_name, mbr_map) in list(pcs.items()):
```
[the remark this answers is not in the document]
tahmed-dev: It needs to be indexed and it is failing on Python 3

src/sonic-config-engine/minigraph.py
```diff
@@ -922,7 +922,7 @@ def parse_xml(filename, platform=None, port_config_file=None, asic_name=None, hw
     results['BGP_PEER_RANGE'] = bgp_peers_with_range
     if mgmt_routes:
         # TODO: differentiate v4 and v6
-        iter(mgmt_intf.values()).next()['forced_mgmt_routes'] = mgmt_routes
+        next(iter(mgmt_intf.values()))['forced_mgmt_routes'] = mgmt_routes
```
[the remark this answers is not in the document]
tahmed-dev: This is existing code. it is getting the first value.

src/sonic-config-engine/tests/sample_output/py2/wait_for_intf.sh
tahmed-dev: No, it is copied from sample_output dir

src/sonic-config-engine/minigraph.py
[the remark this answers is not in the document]
pavel-shirshov: should it better to keep the code as it is?

src/sonic-config-engine/minigraph.py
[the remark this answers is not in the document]
pavel-shirshov: but for doesn't do indexing here. pcs is a special type here?

tahmed-dev:
> As comments.
> My opinion: This PR should be split into many. Currently it is hard to check some sonic-cfggen results because you moved the files from one place to another and the history was lost.
> Also the order has been changed, it is very hard to check that nothing was broken with this change.
> I'd suggest:
> 
> 1. Remove natsorted and review changes
> 2. Add Py2+Py3 code. The code will be changed, but target files must be unchanged
> 3. Move Py2 target files to its own directory and add Py3 target files (why d0 we need two different set of target files at first?)
> 4. Make the tests better (I'd introduce pytest here)

@pavel-shirshov the PR is now broken into 6 commits. As per offline discussion, the rendering result depends on the Python and so our rendered templates differ since it relies on the order of the output. This is a template issue and is not a cfggen issue.

src/sonic-config-engine/minigraph.py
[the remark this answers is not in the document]
tahmed-dev: the .next() is not a method of the iterable object. it works in Python 2 but not Python 3

src/sonic-config-engine/minigraph.py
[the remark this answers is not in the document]
tahmed-dev: On line [1052](https://github.com/Azure/sonic-buildimage/pull/5474/files#diff-18b32e4dc045fa3d1c5941a732124c89R1052) below the pcs is index

src/sonic-config-engine/sonic-cfggen
```diff
@@ -187,22 +188,27 @@ TODO(taoyl): Current version of config db only supports BGP admin states.
     def to_deserialized(data):
         for table in data:
             if type(data[table]) is dict:
-                for key in data[table].keys():
+                current_keys = list(data[table].keys())
+                for key in current_keys:
```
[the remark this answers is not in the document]
tahmed-dev: same as above

src/sonic-config-engine/tests/common_utils.py
tahmed-dev: typo for its

src/sonic-config-engine/tests/multi_npu_data/py3/ipinip.json
tahmed-dev: This is an issue with git diff when viewed all at once. The respective commit will show the history and that it is moved. This why, I'd avoid squash merge

src/sonic-config-engine/tests/sample_output/py2/bgpd_quagga.conf
tahmed-dev: I think this issue could be resolve in the template itself.

src/sonic-config-engine/tests/sample_output/py2/docker-dhcp-relay.supervisord.conf
tahmed-dev: Nope

src/sonic-config-engine/tests/sample_output/py2/frr.conf
tahmed-dev: Nope

src/sonic-config-engine/tests/sample_output/py2/ipinip.json
tahmed-dev: Nope

src/sonic-config-engine/tests/sample_output/py2/l2switch.json
tahmed-dev: Nope

src/sonic-config-engine/tests/sample_output/py2/ports.json
tahmed-dev: Nope

src/sonic-config-engine/tests/sample_output/py2/t2-chassis-fe-vni-zebra.conf
tahmed-dev: Nope

src/sonic-config-engine/tests/sample_output/py2/t2-chassis-fe-zebra.conf
tahmed-dev: Nope

src/sonic-config-engine/tests/common_utils.py
tahmed-dev: No, it ill not as tuple were also failing when using AssertEqual from pytest.

src/sonic-config-engine/sonic-cfggen
```diff
@@ -187,22 +188,27 @@ TODO(taoyl): Current version of config db only supports BGP admin states.
...
 
 
 def deep_update(dst, src):
-    for key, value in src.items():
-        if isinstance(value, dict):
-             node = dst.setdefault(key, {})
-             deep_update(node, value)
-        else:
-             dst[key] = value
+    pending_nodes = [(dst, src)]
```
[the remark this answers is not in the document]
tahmed-dev: Done. I'll stick with the current implementation as the type of the values is not unique.

tahmed-dev:
retest broadcom please

lguohan (approved):

src/sonic-config-engine/tests/sample_output/py2/l2switch.json
```diff
@@ -0,0 +1,268 @@
...
+    "DEVICE_METADATA": {"localhost": {"hwsku": "Mellanox-SN2700"}},
```
qiluo-msft, line 2: >DEVICE_METADATA": {"localhost": {"hwsku": "Mellanox-SN2700"}}, [](start = 5, length = 62)

I suggest only keep one sample_output for both py2 and py3. You can reorder items in the sonic-cfggen output, then you can compare the output against sample_output order-insensitive in the test script.
It does not make sense to degrade the sample_output to fit your change.

tahmed-dev: > > DEVICE_METADATA": {"localhost": {"hwsku": "Mellanox-SN2700"}}, [](start = 5, length = 62)
> 
> I suggest only keep one sample_output for both py2 and py3. You can reorder items in the sonic-cfggen output, then you can compare the output against sample_output order-insensitive in the test script.
> It does not make sense to degrade the sample_output to fit your change.

I agree with you that one sample_output should be the way forward. This is a long running task and few issues are coming from the templates using loop index. Also, this requires knowledge of how to sort out configuration files in a way that does not spoil its syntax such that we do not end up with passing test cases while element in the files are shuffled. Configuration files span multiple domains and the effort will delay conversion to Python 3.

Also, another point to add here is that we are going to deprecate Python 2 packaging and so the folders that has Python 2 will go away soon.

abdosi:
@tahmed-dev since 201911 will be python 2 I am removing to request for 201911.

tahmed-dev:
> @tahmed-dev since 201911 will be python 2 I am removing to request for 201911.

@abdosi makes sense. The remove natsorted which is important for 201911 is ported via separate PR to 201911-t0 branch.

PR #5476, opened 2020-09-28
[Mellanox] Optimize SFP Platform API implementation

keboliu:
**- Why I did it**

Each SFP object inside Chassis will open an SDK client, this is not necessary and SDK client can be shared between SFP objects.

**- How I did it**

Open one SDK client in Chassis init phase, pass this SDK client handle to all the SFP objects when init them.
Remove SDK client open/close code from the SFP object init/delete process.

**- How to verify it**

Verify current pmon daemons work well with these change.
Call related platform API and test the result.

- [ ] 201811
- [ ] 201911
- [ ] 202006


stephenxs (approved):

stephenxs:
retest this please

platform/mellanox/mlnx-platform-api/sonic_platform/sfp.py
```diff
@@ -283,59 +282,41 @@
...
+def deinitialize_sdk_handle(sdk_handle):
+    if sdk_handle is not None:
+        rc = sx_api_close(sdk_handle)
+        if (rc != SX_STATUS_SUCCESS):
+            logger.log_warning("Failed to close api handle.")
+
+        return rc == SXD_STATUS_SUCCESS
+    else:
+         logger.log_warning("Sdk handle is none")
+         return False
```
jleveque, line 304: Since these functions are not called from within this module and are only called from chassis.py, why define them here instead of in chassis.py? It seems like chassis.py is responsible for managing the SDK handle and passing it around so these functions feel like they belong there.

keboliu:
retest mellanox please

keboliu:
retest mellanox please

keboliu:
retest mellanox please

liat-grozovik:
retest mellanox please

platform/mellanox/mlnx-platform-api/sonic_platform/sfp.py
keboliu: @jleveque  after did the change found that it will break the uni-test, to move the SFP init/deinit functions to chassis then we need to import some SDK specific lib, in the uni-test environment, these libs are not available, so uni-test will be broken. Currently, we don't have a good solution to resolve the import issue for uni-test cases,  so would like to keep it as is, what's your view?

jleveque: We can keep it as-is for now, then. This can be saved for a future refactor.

jleveque (approved):

PR #5498, opened 2020-09-29
fix build behind proxy

jon-nokia:
Signed-off-by: Jon Goldberg <jon.goldberg@nokia.com>

The problem was proxy was missing on "pip install".  This is to fix the build behind the proxy.


vganesan-nokia:
This PR fixes #5551

PR #5505, opened 2020-09-29
Fix config qos reload on Multi-asic platforms.

abdosi:
Why I did:
config qos reload on Multi-asic platforms not working 
The sort Function on Portname is not able to understand
Backplane Interface name.

How I did:
Fix is to to remove sorting from j2 template as this is not needed
as all Ports are dervie from port_config.ini where ports are listed
in fixed static sort order.

Also we do not sort port in buffer_config.j2 so made logic similar to
that and keep it simple instead of enhancing sort function.

How I verified:
config load_minigraph -y is fine which internally invoke config qos reload.

### DIFF
--- a/files/build_templates/qos_config.j2
+++ b/files/build_templates/qos_config.j2
@@ -2,7 +2,6 @@
 {%- for port in PORT %}
     {%- if PORT_ALL.append(port) %}{% endif %}
 {%- endfor %}
-{%- if PORT_ALL | sort_by_port_index %}{% endif %}
 
 {%- set port_names_list_all = [] %}
 {%- for port in PORT_ALL %}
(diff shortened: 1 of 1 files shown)


smaheshm:
Can you add more info about where is it breaking. This change is there in both master and 201911.

https://github.com/Azure/sonic-buildimage/blob/master/src/sonic-config-engine/sonic-cfggen#L58

abdosi:
retest broadcom please

abdosi:
retest broadcom please

abdosi:
retest broadcom please

smaheshm:
retest broadcom please

smaheshm:
@abdosi Is this still relevant?

smaheshm:
@abdosi Can you also update sonic-cfggen to remove the sort_by_index. This change is required for chassis.

abdosi:
> @abdosi Can you also update sonic-cfggen to remove the sort_by_index. This change is required for chassis.

@smaheshm updated UT output accordingly.

smaheshm:
can you also remove the API from sonic-cfggen

PR #5517, opened 2020-10-01
[docker-orchagent]: Add NDP Proxy Daemon

theasianpianist:

Add NDP Proxy Daemon to the `docker-orchagent` image, and manage it with `supervisord`.

**- Why I did it**
Native proxy NDP in Linux requires explicit neighbor entries in the kernel neighbor table, i.e. it's not possible to configure a subnet for proxy NDP. NDP Proxy Daemon (`ndppd`) bypasses the kernel neighbor table entirely and allows for subnets when configuring proxy NDP.

**- How I did it**
* Install `ndppd` during image build, and copy config files to image
* Configure proxy settings based on config DB at container start
* Pipe `ndppd` output to `logger` inside container to log output in syslog

**- How to verify it**
1. Copy and load `docker-orchagent` image to a device that supports `proxy_arp`.
2. Enable `proxy_arp` for a VLAN interface in config_db.
3. Stop and delete the existing `swss` container.
4. `config reload`
5. Select an unused IPv6 address in the same subnet as one of the VLAN interfaces on the device from step 1.
6. On a neighbor connected to the device from step 1, choose an interface connected to the VLAN used in step 5, and add the IPv6 address from step 5 to this interface.
7. From the neighbor, attempt to ping a new, unused IPv6 address in the same subnet as the VLAN interface from step 5 (the ping should fail).
8. Check the output of `ip -6 neigh`. The IPv6 address pinged in step 7 should be `REACHABLE`.

- [ ] 201811
- [ ] 201911
- [ ] 202006

Add support for proxy NDP


dockers/docker-orchagent/ndppd.conf.j2
```diff
@@ -0,0 +1,38 @@
+{% block banner %}
```
lguohan, line 1: can we have a test to validate the configuration? i think there is some example in sonic-cfggen test cases.

dockers/docker-orchagent/ndppd.conf.j2
```diff
@@ -0,0 +1,38 @@
...
+{%      endfor -%}
+
+{%      for intf, prefix_list in proxy_interfaces.items() %}
+{%          if prefix_list %}
+
+proxy {{ intf }} {
+{%              for prefix in prefix_list %}
+    rule {{ prefix | network }}/{{ prefix | prefixlen }} {
+        static
+    }
```
prsunny, line 33: If there are multiple IPv6 prefixes, are there any separators required or just appending rule is enough?

theasianpianist: Just appending the rule works (no commas needed between rules), unfortunately the format is not standard JSON.

dockers/docker-orchagent/ndppd.conf.j2
theasianpianist: Added test

prsunny (approved):

theasianpianist:
retest baseimage please

lguohan (approved):

lguohan:
there are still build failures

theasianpianist:
@lguohan in the process of debugging

theasianpianist:
@lguohan build failures fixed

lguohan (approved):

PR #5534, opened 2020-10-02
[make init] update submodule's remotes

yxieca:
**- Why I did it**
I am using a long standing checked out repo to build dev images. From time to time the repo is too out-sync'ed with remotes and some submodule stay behind even after make reset and make init call. This causes build to fail unexpectedly.

**- How I did it**
With further tests, I moved the recursive "git remote update' to make reset code block and placed in front of recursive --init. The original change has the ordering wrong but that was hidden by the fact that make init was executed multiple times in master branch and 201911 branch.

The summary is:
- for a newly cloned repo. existing 'make init' is enough.
- for long standing repo, when things are getting too out dated, or when we switching branch, we usually need to do a make reset before build again. Place the fix in make reset makes more sense.

**- How to verify it**
Build a dev image. Tested it with a newly cloned repo as well.

- [ ] 201811
- [ ] 201911
- [ ] 202006

### DIFF
--- a/Makefile.work
+++ b/Makefile.work
@@ -283,7 +283,6 @@ showtag:
 
 init :
 	@git submodule update --init --recursive
-	@git submodule foreach --recursive git remote update &>/dev/null
 	@git submodule foreach --recursive '[ -f .git ] && echo "gitdir: $$(realpath --relative-to=. $$(cut -d" " -f2 .git))" > .git'
 
 .ONESHELL : reset
@@ -302,6 +301,7 @@ reset :
 	        git reset --hard;
 	        git submodule foreach --recursive 'git clean -xfdf || true';
 	        git submodule foreach --recursive 'git reset --hard || true';
+	        git submodule foreach --recursive 'git remote update || true';
 	        git submodule update --init --recursive;
 	        echo "Reset complete!";
 	    else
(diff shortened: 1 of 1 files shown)


yxieca:
retest broadcom please

Makefile.work
```diff
@@ -283,6 +283,7 @@ showtag:
 
 init :
 	@git submodule update --init --recursive
+	@git submodule foreach --recursive git remote update &>/dev/null
```
jleveque, line 286: Maybe we should also add this to `make reset` below?

yxieca: possibly, I did see make reset didn't cleanup everything once. we could try this on that too.

yxieca: Just to clarify @jleveque, are you requesting a change to be added to this PR or suggesting experiment with make reset separately?

qiluo-msft (dismissed):

Makefile.work
jleveque: @yxieca: Either way is fine. If you want to test it and add it you can, otherwise we can take note for a future PR.

yxieca: Thanks Joe, per our discussion offline and my test results. I moved the recursive "git remote update" to make reset and placed in front of --init.

jleveque (approved):

PR #5549, opened 2020-10-06
[python-click] No longer build our own package, let pip/setuptools install vanilla

jleveque:
**- Why I did it**
We were building our own python-click package because we needed features/bug fixes available as of version 7.0.0, but the most recent version available from Debian was in the 6.x range.

"Click" is needed for building/testing and installing sonic-utilities. Now that we are building sonic-utilities as a wheel, with Click specified as a dependency in the setup.py file, setuptools will install a more recent version of Click in the sonic-slave-buster container when building the package, and pip will install a more recent version of Click in the host OS of SONiC when installing the sonic-utilities package. Also, we don't need to worry about installing the Python 2 or 3 version of the package, as the proper one will be installed as necessary.

**- How I did it**
Remove makefiles and related files for building python-click package from source, no longer install package in host OS or slave containers.

Also clean up the installation of some other Python dependencies which should now implicitly get installed by setuptools at build time and pip at install time.

**- How to verify it**

Build and install an image. Ensure sonic-utilities package builds and installs OK and that the SONiC CLI works properly. You can also verify via `pip list` that 'click' is installed.

- [ ] 201811
- [ ] 201911
- [ ] 202006


jleveque:
Check builds are failing, as this PR depends on https://github.com/Azure/sonic-utilities/pull/1151

jleveque:
Retest this please

jleveque:
Retest vs please

jleveque:
Retest baseimage please

lguohan (approved):

xumia:
The python3-click=7.1.2-1 in bullseye and python3-click=7.0-1 in buster are supported, do we still need it?

PR #5532, opened 2020-10-02
[BCMSAI] Update BCM SAI debian package to 4.2.1.3 (6.5.19 hsdk)

smaheshm:
**- Why I did it**

Update BCM SAI debian package to 4.2.1.3 (6.5.19 hsdk)

**- How I did it**

Build debain package with BCMSAI 4.2.1.3 and 6.5.19 hsdk.

**- How to verify it**

Manual tests
```
 admin@str-s6000-acs-8:~$ show version

SONiC Software Version: SONiC.bcmsai4.2-nogpl.0-dirty-20201001.191202
Distribution: Debian 10.6
Kernel: 4.19.0-9-2-amd64
Build commit: 79f21728
Build date: Thu Oct  1 19:45:40 UTC 2020
Built by: samaddik@samaddik-vm-01

Platform: x86_64-dell_s6000_s1220-r0
HwSKU: Force10-S6000
ASIC: broadcom
Serial Number: 1QBRX42
Uptime: 18:30:19 up  2:03,  1 user,  load average: 1.05, 0.92, 0.82

Docker images:
REPOSITORY                    TAG                                       IMAGE ID            SIZE
docker-teamd                  bcmsai4.2-nogpl.0-dirty-20201001.191202   ea43f1f2af6d        391MB
docker-teamd                  latest                                    ea43f1f2af6d        391MB
docker-sflow                  bcmsai4.2-nogpl.0-dirty-20201001.191202   8f7610eeb99a        395MB
docker-sflow                  latest                                    8f7610eeb99a        395MB
docker-orchagent              bcmsai4.2-nogpl.0-dirty-20201001.191202   fe9b0db5ff0e        405MB
docker-orchagent              latest                                    fe9b0db5ff0e        405MB
docker-nat                    bcmsai4.2-nogpl.0-dirty-20201001.191202   4286e9655aff        394MB
docker-nat                    latest                                    4286e9655aff        394MB
docker-fpm-frr                bcmsai4.2-nogpl.0-dirty-20201001.191202   49b027a3aed7        407MB
docker-fpm-frr                latest                                    49b027a3aed7        407MB
docker-snmp                   bcmsai4.2-nogpl.0-dirty-20201001.191202   1c8494bcbc8a        399MB
docker-snmp                   latest                                    1c8494bcbc8a        399MB
docker-router-advertiser      bcmsai4.2-nogpl.0-dirty-20201001.191202   4ca6adf3a696        359MB
[...]

PR #5550, opened 2020-10-06
[201911] Fix the error message for Tx Drop happening when netdev link is down

abdosi:
Why/How I did:
The error message is updated to provide correct information
of Netdevice link being down for Tx Drop
Taken the change from BRCM SAI 4.2.1.3 (6.5.19 hsdk) (https://github.com/Azure/sonic-buildimage/pull/5532/files#diff-9787d8debde44bb0dd3ce502443bf25cR5362)

### DIFF
--- a/platform/broadcom/saibcm-modules/systems/linux/kernel/modules/bcm-knet/bcm-knet.c
+++ b/platform/broadcom/saibcm-modules/systems/linux/kernel/modules/bcm-knet/bcm-knet.c
@@ -5279,7 +5279,7 @@ bkn_tx(struct sk_buff *skb, struct net_device *dev)
     }
 
     if (!netif_carrier_ok(dev)) {
-        DBG_WARN(("Tx drop: Invalid RCPU encapsulation\n"));
+        DBG_WARN(("Tx drop: Netif link is down.\n"));
         priv->stats.tx_dropped++;
         sinfo->tx.pkts_d_no_link++;
         dev_kfree_skb_any(skb);


platform/broadcom/saibcm-modules/systems/linux/kernel/modules/bcm-knet/bcm-knet.c
```diff
@@ -5279,7 +5279,7 @@ bkn_tx(struct sk_buff *skb, struct net_device *dev)
     }
 
     if (!netif_carrier_ok(dev)) {
-        DBG_WARN(("Tx drop: Invalid RCPU encapsulation\n"));
+        DBG_WARN(("Tx drop: Netif link is down.\n"));
```
lguohan, line 5282: is there way to get the device name in the log?

lguohan (approved):

platform/broadcom/saibcm-modules/systems/linux/kernel/modules/bcm-knet/bcm-knet.c
prsunny: May be print `ifalias `or `name `field?

judyjoseph: **dev->name** looks ok. But as a general comment, most of the debugs in this file don't have device name printed, may be to suggest brcm to improve debugs.

abdosi: @prsunny and @judyjoseph Sure, will ask Broadcom to enhance debug messages. For now since this change same as what is in BRCM-SAI 4.2 will keep same so to avoid local changes.

Issue #390, opened 2017-03-09
bgp container not started due to database 


lguohan:
```
admin@str-s6000-acs-7:~$ sudo systemctl status bgp
â— bgp.service - BGP container
   Loaded: loaded (/etc/systemd/system/bgp.service; enabled)
   Active: failed (Result: exit-code) since Wed 2017-03-08 10:55:45 UTC; 21h ago
 Main PID: 802 (code=exited, status=125)

Mar 08 10:55:39 str-s6000-acs-7 systemd[1]: Started BGP container.
Mar 08 10:55:42 str-s6000-acs-7 bgp.sh[802]: docker: Error response from daemon: No such container: database.
Mar 08 10:55:42 str-s6000-acs-7 bgp.sh[802]: See 'docker run --help'.
Mar 08 10:55:42 str-s6000-acs-7 systemd[1]: bgp.service: main process exited, code=exited, status=125/n/a
Mar 08 10:55:45 str-s6000-acs-7 bgp.sh[903]: Error response from daemon: No such container: bgp
Mar 08 10:55:45 str-s6000-acs-7 systemd[1]: bgp.service: control process exited, code=exited status=1
Mar 08 10:55:45 str-s6000-acs-7 systemd[1]: Unit bgp.service entered failed state.
```

stcheng:
```
syslog:998:Mar 21 20:42:51 str-s6000-on-2 INFO bash[739]: Error response from daemon: No such container: database
syslog:1001:Mar 21 20:42:52 str-s6000-on-2 INFO bash[739]: Error response from daemon: No such container: database
syslog:1006:Mar 21 20:42:54 str-s6000-on-2 INFO bash[739]: Error response from daemon: No such container: database
syslog:1021:Mar 21 20:42:55 str-s6000-on-2 INFO bgp.sh[731]: docker: Error response from daemon: No such container: database.
syslog:1026:Mar 21 20:42:55 str-s6000-on-2 INFO bash[739]: Error response from daemon: No such container: database
syslog:1087:Mar 21 20:42:58 str-s6000-on-2 INFO bgp.sh[832]: Error response from daemon: No such container: bgp
```
can be reproduced.

taoyl-ms:
Mar 21 20:42:50 str-s6000-on-2 INFO systemd[1]: Starting Database container...
Mar 21 20:42:50 str-s6000-on-2 INFO systemd[1]: Started Database container.
[...]

PR #5553, opened 2020-10-06
Add Host Account Management (HAM) project

martin-belanger:
**- Dependencies**
It wasn't clear how to put a dependency between two pull requests, hence I added this section.
This pull request depends on [sonic-swss-common #390](https://github.com/Azure/sonic-swss-common/pull/390) being merged first. 

**- Why I did it**
As described in [AAA Improvements](https://github.com/Azure/SONiC/blob/master/doc/aaa/AAA%20Improvements/AAA%20Improvements.md), there is a need for a central location to manage user accounts. The solution is "Host Account Management" or HAM, which acts as the single source of truth for user accounts. Some of the features of HAM include:
  - Automatically assign Linux user credentials (UID, GID) to RADIUS/TACACS+ user accounts.
  - Allow applications running in containers to retrieve Linux user credentials (UID, GID) from the host.
  - Allow applications running in containers (such as a REST server) to create/modify/delete local user accounts on the host. This is only allowed for users with admin privileges.
  - Fixes several problems related to RADIUS/TACACS+ login (for the full list refer to the document cited above). 

**- How I did it**
Not sure what "How I did it" means. I wrote C/C++ code? 
For a description of how to use HAM, refer to the src/ham/README.md. 

**- How to verify it**
This pull request adds a new building block called HAM. In this pull request we do not hook HAM up to any other components in SONiC. In a second phase, we will hook up RADIUS/TACACS+ to use HAM. In a third phase, the sonic-management-framework will also use HAM. Since HAM is not hooked up to anything in this first phase, the only way to verify that things are in place is through the debug utility "**`hamctl`**". This can be invoked at the bash shell.

No backport needed. This is a new feature.

New Host Account Management feature. A detailed description of HAM can be found in: https://github.com/Azure/SONiC/blob/master/doc/aaa/AAA%20Improvements/AAA%20Improvements.md


renukamanavalan:
The new guidelines as I understand is to associate test with every new code to be checked in. This is indeed a *big* new code for a new feature.

What is the plan for tests?

martin-belanger:
Hi @renukamanavalan ,

Agreed, tests for HAM need to be added. I will find some time to add a test framework for HAM (hopefully soon). By the way, are there any test guidelines? I assume tests are run in a VM? In other words, if a test fails in such a way that it leaves the system in a wonky state, it's only the test VM that will be affected. Any pointers would be greatly appreciated.

NOTE: Currently, hamd is not used by anything. In other words, the risk level is very low. Regardless, I will start working on tests as soon as I can. 

Regards,
Martin

rules/ham.mk
```diff
@@ -0,0 +1,21 @@
+HAM_VERSION := 1.0.0
+
+# Main package
+HAM = ham_$(HAM_VERSION)_$(CONFIGURED_ARCH).deb
+$(HAM)_SRC_PATH = $(SRC_PATH)/ham
+$(HAM)_DEPENDS  = $(LIBSWSSCOMMON_DEV)
+$(HAM)_RDEPENDS = $(LIBSWSSCOMMON)
+
+SONIC_DPKG_DEBS += $(HAM)
```
renukamanavalan, line 9: Can we make this conditional include using a build time flag?

davidpil2002:
Hi,
Few Q:
1.Its this AAA enhancement going to create some backward compatibility with the existing SONiC commands ?
2.its this flow going to affect the existing hostcfgd flow?
3. there is some estimation time when the feature will be merged?
@martin-belanger  @lguohan 
Thanks

PR #5560, opened 2020-10-07
Optimized caclmgrd ACL Rule Table Notification handling. 

abdosi:
Why I did:
Optimized caclmgrd ACL Rule Table Notification handling. 
Previously any event happening on ACL Rule Table (eg DATAACL rules
programmed) caused control plane default action to be triggered
and invoking iptables commands

Now Control Place ACTION will be trigger only
for ACL Rule belonging to Control ACL Table

**- How I did it**
Check for Table Type for ACL Rule notification
**- How to verify it**
Run crm/test_crm.py which populate DATAACL Rules for ACL Counters and made sure DATACL ACL
rules prgrammed do not trigger Control plane Action


jleveque:
@abdosi: This PR should be complimentary to https://github.com/Azure/sonic-buildimage/pull/5312, correct?

jleveque (approved):

PR #5569, opened 2020-10-09
[fabric] Disable unnecessary processes in swss and the orchagent-portsyncd dependency for fabric asic

ngoc-do:
Signed-off-by: ngocdo <ngocdo@arista.com>

**- Why I did it**

Fabric asic just needs very few processes in swss to run. Many of them should be disabled. The following should be disabled.

program:portsyncd
program:neighsyncd
program:fdbsyncd
program:vlanmgrd
program:intfmgrd
program:portmgrd
program:buffermgrd
program:vrfmgrd
program:nbrmgrd
program:vxlanmgrd
program:coppmgrd
program:tunnelmgrd

Fabric asic doesn't run portsyncd to turn its ports on because it doesn't have any ports. But orchagent is currently configured to not start if portsyncd. We need to break this dependency.

**- How I did it**

Creating critical_processes.j2 and supervisord.conf.j2 for different asics. Also disabled dependency of orchagent and portsyncd for fabric asic by modifying /etc/supervisor/conf.d/supervisord.conf.j2. We now can know asic type from CONFIG_DB. So it's possible to have custom critical_processes and supervisord.conf based on chip type.

**- How to verify it**

Tested on a fixed system with NPUs (BCM88690) and fabric asics (BCM88790) - worked great.


anshuv-mfst:
@Srikanth (Nokia) could you please take a look, thanks.

skeesara-nokia:
> 
> 
> @srikanth (Nokia) could you please take a look, thanks.

Will do

dockers/docker-orchagent/supervisord.conf.j2
```diff
@@ -152,7 +178,9 @@ stdout_logfile=syslog
 stderr_logfile=syslog
 dependent_startup=true
 dependent_startup_wait_for=swssconfig:exited
+{% endif %}
 
+{% if is_fabric_asic == 0 %}
```
prsunny, line 183: The concern with this approach is, if someone have to add a new process in the end, he may ignore this `if is_fabric_asic`. My suggestion is lets have one section for `is_fabric_asic == 0` and else section for fabric processes (similar to critical_process.j2). This way priority can also be adjusted accordingly. What do you think?

prsunny:
Can you also modify the description and title. From the changes, it doesn't look like only the portsyncd dependency changes but also some process are not even started in fabric.

dockers/docker-orchagent/supervisord.conf.j2
ngoc-do: Sure, that makes sense.

ngoc-do:
> Can you also modify the description and title. From the changes, it doesn't look like only the portsyncd dependency changes but also some process are not even started in fabric.

That makes sense too.

dockers/docker-orchagent/supervisord.conf.j2
ngoc-do: @prsunny I addressed your comments. Thanks for reviewing. The new commit 1 overrided the previous commit when I resolved conflicts. Hope that it's clear to you. Also updated the title + description.

prsunny: I would prefer to have it properly listed in if-else. Currently the order/priority is not in order. This would be more confusing if we have to add new process later. please keep the original order for `if not-fabric` and add the fabric process in `else `part

ngoc-do: Sounds reasonable.  But some processes like `orchagent` is available in both `fabric` and `not fabric` part. So if we are doing that:

`if not-fabric:`
`    orchagent`
`    portsyncd`
`else:`
`    orchagent`
`end-if`

`orchagent` will appear both, and seems redundant. Is this OK? All processes for fabric are also in `not fabric`

ngoc-do: My main concern that having `if` and `else` requires to update `orchagent` at both parts if someone wants to update `orchagent`'s configuration. That seems harder to manage and redundant. So having `if` may be better.

anshuv-mfst:
@skeesara-nokia - please review/approve the PR, thanks.

skeesara-nokia (dismissed):

ngoc-do:
Could we please merge this PR? It could cause conflicts if we keep it here for too long.

ngoc-do:
@prsunny Please review it.

ngoc-do:
Hello, could some one please approve and merge this PR?

anshuv-mfst:
Hi @prsunny , @skeesara-nokia - Could you please review and approve.

dockers/docker-orchagent/docker-init.sh
```diff
@@ -13,6 +13,12 @@ CFGGEN_PARAMS=" \
 "
 VLAN=$(sonic-cfggen $CFGGEN_PARAMS)
 
+mkdir -p /etc/supervisor/
+sonic-cfggen -d -t /usr/share/sonic/templates/critical_processes.j2 > /etc/supervisor/critical_processes
```
prsunny, line 17: @tahmed-dev for feedback here.

dockers/docker-orchagent/supervisord.conf.j2
```diff
@@ -28,49 +28,69 @@ stdout_logfile=syslog
...
+{% set is_fabric_asic = 0 %}
+{% set orchagent_dependent_startup_wait_for = "portsyncd:running" %}
+{% if DEVICE_METADATA.localhost.switch_type %}
+{% if DEVICE_METADATA.localhost.switch_type == "fabric" %}
+{% set is_fabric_asic = 1 %}
+{% set orchagent_dependent_startup_wait_for = "rsyslogd:running" %}
+{%- endif %}
+{%- endif %}
+
+[program:orchagent]
```
prsunny, line 40: My concern is, priorities are all jumbled up here. IMO, its harder to add a process. If you dont prefer to have duplicate, please have it in the order of priority and add the skip the ones for fabric within "if" case

ngoc-do: Sorting based on priority makes sense.

For clarification, I will add the processes which do NOT run on fabric in "if" case.

ngoc-do:
@prsunny Could you please review the updated code? Fabric code in swss merged, and so we need this one merge asap. Thanks.

prsunny:
> @prsunny Could you please review the updated code? Fabric code in swss merged, and so we need this one merge asap. Thanks.

@ngoc-do , Sure will review. Kindly avoid force-push, thereby deleting commit history.

prsunny (dismissed):

dockers/docker-orchagent/docker-init.sh
tahmed-dev: Thanks @prsunny. Can you please combine the above calls into a single call to sonic-cggen.

```
mkdir -p /etc/supervisor/
mkdir -p /etc/supervisor/conf.d/
sonic-cfggen -d -t /usr/share/sonic/templates/critical_processes.j2,/etc/supervisor/critical_processes \
    -t /usr/share/sonic/templates/supervisord.conf.j2,/etc/supervisor/conf.d/supervisord.conf
```

tahmed-dev (changes requested):

dockers/docker-orchagent/docker-init.sh
tahmed-dev: Indeed, those new `sonic-cfggen` calls need to be part of the above calls at line 12 above.

```
CFGGEN_PARAMS=" \
    -d \
    -y /etc/sonic/constants.yml \
    -t /usr/share/sonic/templates/switch.json.j2,/etc/swss/config.d/switch.json \
    -t /usr/share/sonic/templates/ipinip.json.j2,/etc/swss/config.d/ipinip.json \
    -t /usr/share/sonic/templates/ports.json.j2,/etc/swss/config.d/ports.json \
    -t /usr/share/sonic/templates/vlan_vars.j2 \
    -t /usr/share/sonic/templates/ndppd.conf.j2,/etc/ndppd.conf \
    -t /usr/share/sonic/templates/critical_processes.j2,/etc/supervisor/critical_processes \
    -t /usr/share/sonic/templates/supervisord.conf.j2,/etc/supervisor/conf.d/supervisord.conf
"
```

ngoc-do: Thanks @tahmed-dev. That's a great suggestion. I updated with a new commit.

ngoc-do:
Addressed reviewer comments with latest commit.

prsunny (approved):

anshuv-mfst:
@tahmed-dev - can you please take a look at the earliest, review comments addressed.

tahmed-dev (approved):

ngoc-do:
Could we merge this PR? Seems it got approved and no conflicts.

anshuv-mfst:
@tahmed-dev, @prsunny - Can you please help with merge.

saravanansv:
on a supervisor dockers/docker-orchagent/base_image_files/monit_swss needs to be a smaller version of what it is right ?

PR #5600, opened 2020-10-12
[bgp]: Enable next-hop-tracking through default

pavel-shirshov:
**- Why I did it**
FRR introduced [next hop tracking](http://docs.frrouting.org/projects/dev-guide/en/latest/next-hop-tracking.html) functionality. 
That functionality requires resolving BGP neighbors before setting BGP connection (or explicit ebgp-multihop command). Sometimes (BGP MONITORS) our neighbors are not directly connected and sessions are IBGP. In this case current configuration prevents FRR to establish BGP connections.  Reason would be "waiting for NHT". To fix that we need either add static routes for each not-directly connected ibgp neighbor, or enable command `ip nht resolve-via-default`

**- How I did it**
Put `ip nht resolve-via-default` into the config

**- How to verify it**
Build an image. Enable BGP_MONITOR entry and check that entry is Established or Connecting in FRR

- [ ] 201811
- [x] 201911
- [x] 202006

### DIFF
--- a/dockers/docker-fpm-frr/frr/zebra/zebra.interfaces.conf.j2
+++ b/dockers/docker-fpm-frr/frr/zebra/zebra.interfaces.conf.j2
@@ -10,6 +10,8 @@ vni {{ vnet_metadata['vni'] }}
 {% endblock vrf %}
 !
 {% block interfaces %}
+! Enable nht through default route
+ip nht resolve-via-default
 ! Enable link-detect (default disabled)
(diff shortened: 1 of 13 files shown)


lguohan (dismissed):

lguohan:
can you fix the test?

lguohan:
i can see the test is still failing.

rlhui:
retest broadcom please

lguohan (approved):

PR #5601, opened 2020-10-12
[cfggen] Remove NatSorted

tahmed-dev:
Natural sorting of SONiC config gen output consumes lot of CPU cycles.
The sole use of natsorted was to make test comparison easier and so,
the natsorting logic is now relocated to the test suite. As a result
sonic-cfggen gained nearly 1 sec per call since we no longer import
natsorted module!


lguohan (approved):

lguohan:
good improvement. is there a way to make sure the natsorted is not introduced again in the future accidentally? it is easier missed. especially, I am a little concern that if one of the dependency packages that sonic-cfggen uses will introduce natsorted. this is not a real concern for this pr. it is about future-proofing.

tahmed-dev:
> good improvement. is there a way to make sure the natsorted is not introduced again in the future accidentally? it is easier missed. especially, I am a little concern that if one of the dependency packages that sonic-cfggen uses will introduce natsorted. this is not a real concern for this pr. it is about future-proofing.

Thanks! I am aware of putting a package on hold using apt `sudo apt-mark hold <package-name>` not sure about pip though. @jleveque is there something similart in pip? Or can we whitelist packages?

tahmed-dev:
retest vsimage please

tahmed-dev:
retest vsimage please

PR #5621, opened 2020-10-13
Optimze ACL Table/Rule notification handling

abdosi:
What I did:
Optimize ACL Table/Rule notification handling
to loop pop() until empty to consume all the data in a batch
This way we prevent multiple call to iptable updates

**- How to verify it**

Loaded (sudo acl-loader update full acl.json) with 100 rules and verified all the set notification are batch together in one update to iptables .

Similarly (sudo acl-loader delete) batch all del operation in one update.

### DIFF
--- a/files/image_config/caclmgrd/caclmgrd
+++ b/files/image_config/caclmgrd/caclmgrd
@@ -563,7 +563,7 @@ class ControlPlaneAclManager(daemon_base.DaemonBase):
                     # Check ACL Rule notification and make sure Rule point to ACL Table which is Controlplane
                     else:
                         acl_table = key.split(acl_rule_table_seprator)[0]
-                        if self.config_db_map[namespace].get_table(self.ACL_TABLE)[acl_table]["type"]  == self.ACL_TABLE_TYPE_CTRLPLANE:
+                        if self.config_db_map[namespace].get_table(self.ACL_TABLE)[acl_table]["type"] == self.ACL_TABLE_TYPE_CTRLPLANE:
                             ctrl_plane_acl_notification.add(namespace)
 
             # Update the Control Plane ACL of the namespace that got config db acl table event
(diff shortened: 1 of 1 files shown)


jleveque (approved):
Looks good to me. @qiluo-msft to review, as well.

files/image_config/caclmgrd/caclmgrd
```diff
@@ -550,23 +550,24 @@ class ControlPlaneAclManager(daemon_base.DaemonBase):
...
+                        break
+                    # ACL Table notification. We will take Control Plane ACTION for any ACL Table Event
+                    # This can be optimize further but we should not have many acl table set/del events in normal
+                    # scenario
+                    if acl_rule_table_seprator not in key:
+                        ctrl_plane_acl_notification.add(namespace)
+                    # Check ACL Rule notification and make sure Rule point to ACL Table which is Controlplane
+                    else:
+                        acl_table = key.split(acl_rule_table_seprator)[0]
+                        if self.config_db_map[namespace].get_table(self.ACL_TABLE)[acl_table]["type"]  == self.ACL_TABLE_TYPE_CTRLPLANE:
```
qiluo-msft, line 566: >  [](start = 102, length = 1)

Use only one space

abdosi: @ Done

---
In reply to: [504353871](https://github.com/Azure/sonic-buildimage/pull/5621#discussion_r504353871) [](ancestors = 504353871)

jleveque (approved):

qiluo-msft (approved):

qiluo-msft:
Do you have performance test result?

abdosi:
> Do you have performance test result?

@qiluo-msft 
I verified the number of update happening to iptable (function update_control_plane_acls) for 100 rule is only 1 time. Previously it was called 100 times.

I did not measure in time perspective.

PR #5142, opened 2020-08-10
[bgpcfgd] Support for FRR configuration based on config DB events

zhaozhenhong:
**- Why I did it**
1. Support for non-template based FRR configurations (BGP, route-map, OSPF, static route..etc) using config DB schema.
1. Support for save & restore - Jinja template based config-DB data read and apply to FRR during startup

**- How I did it**

Added a new script and corresponding service. Based on the settings in DEVICE_METADATA table, new service will be started along with the bgp container start. And when user changed the BGP or other related table entries in config DB, the daemon will run corresponding VTYSH commands to program on FRR.
Also added jinja template to generate FRR config file to be used by FRR daemons while bgp container restarted

**- How to verify it**
1. Add/delete data on config DB and then run VTYSH "show running-config" command to check if FRR configuration changed.
1. Restart bgp container and check if generated FRR config file is correct and run VTYSH "show running-config" command to check if FRR configuration is consistent with attributes in config DB

**- Description for the changelog**

**- A picture of a cute animal (not mandatory but encouraged)**

PR #5622, opened 2020-10-13
Add templates and code to support VoQ chassis iBGP peers

jmmikkel:
* Add support to convert a new VoQChassisInternal element in the
  BGPSession element of the minigraph to a new BGP_VOQ_CHASSIS_NEIGHBOR
  table in CONFIG_DB.
* Add a new set of "voq_chassis" templates to docker-fpm-frr
* Add a new BGP peer manager to bgpcfgd to add neighbors from the
  BGP_VOQ_CHASSIS_NEIGHBOR table using the voq_chassis templates.
* Add a test case for minigraph.py, making sure the VoQChassisInternal
  element creates a BGP_VOQ_CHASSIS_NEIGHBOR entry, but not if its
  value is "false".
* Add a set of test cases for the new voq_chassis templates in
  sonic-bgpcfgd tests.

Note that the templates expect the new
"bgp bestpath peer-type multipath-relax" bgpd configuration to be
available.

Signed-off-by: Joanne Mikkelson <jmmikkel@arista.com>

**- Why I did it**

For the VoQ chassis system, the iBGP sessions between the ASIC Instances have substantially different configuration needs from the usual eBGP neighbors. Pushing those differences into the "general" templates would involve a lot of if-statements and the like, while the bgpcfgd system looks designed around keeping a set of templates per kind of neighbor. So this change adds a new set of templates, keeping the VoQ chassis peers out of the other templates.

See the VoQ chassis BGP HLD for additional details, particularly of the contents of the new templates: https://github.com/Azure/SONiC/pull/674

**- How I did it**

Added new voq_chassis directory in the fpm-frr docker image. In order to make these templates selectable, changed bgpcfgd to look at a new BGP_VOQ_CHASSIS_NEIGHBOR table in config_db. To get those entries into config_db, added a new element to BGPSession in the minigraph and code to honor it when parsing the minigraph.

**- How to verify it**

Boot a VoQ chassis system with VoQChassisInternal BGPSessions (minigraph) or BGP_VOQ_CHASSIS_NEIGHBOR config_db entries, and check that all the peerings in the iBGP mesh are up. Advertise routes from an eBGP neighbor and confirm they are learned by the iBGP peers.

- [ ] 201811
- [ ] 201911
- [ ] 202006

Add support for configuring VoQ chassis iBGP peers


jmmikkel:
retest baseimage please

src/sonic-bgpcfgd/bgpcfgd/managers_bgp.py
```diff
@@ -170,6 +172,10 @@ def add_peer(self, vrf, nbr, data):
         #
         if "local_addr" not in data:
             log_warn("Peer %s. Missing attribute 'local_addr'" % nbr)
+        elif not self.require_intf_in_config_db:
```
jmmikkel, line 175: When the inband interfaces have been added to CONFIG_DB, we may be able to add them with another InterfaceMgr instead of doing this.

src/sonic-bgpcfgd/tests/data/voq_chassis/peer-group.conf/param_all.json
```diff
@@ -0,0 +1,7 @@
+{
+    "CONFIG_DB__DEVICE_METADATA": {
+        "localhost": {
+            "type": "ToRRouter"
```
minionatwork, line 4: Dont we need T2 router here?.

minionatwork (approved):

src/sonic-bgpcfgd/tests/data/voq_chassis/instance.conf/param_all.json
```diff
@@ -0,0 +1,24 @@
+{
+    "CONFIG_DB__DEVICE_METADATA": {
+        "localhost": {
+            "sub_role": "BackEnd"
```
vganesan-nokia, line 4: In "voq" based systems, all asics (except fabric asics) are considered "FrontEnd" and unless we run BGP in fabric asics, we would not need a "BackEnd" sub-role in voq systems.

vganesan-nokia, line 4: In "voq" based systems, all asics (except fabric asics) are considered "FrontEnd" and unless we run BGP in fabric asics, we would not need a "BackEnd" sub-role in voq systems.

jmmikkel: Nothing was looking at the sub_role so it didn't hurt the test. But I have removed it now.

vganesan-nokia (dismissed):

dockers/docker-fpm-frr/frr/bgpd/templates/voq_chassis/instance.conf.j2
```diff
@@ -0,0 +1,35 @@
...
+!
+  neighbor {{ neighbor_addr }} peer-group VOQ_CHASSIS_PEER
+  neighbor {{ neighbor_addr }} remote-as {{ bgp_session['asn'] }}
+  neighbor {{ neighbor_addr }} description {{ bgp_session['name'] }}
+{# set the bgp neighbor timers if they have not default values #}
+{% if     (bgp_session['keepalive'] is defined and bgp_session['keepalive'] | int != 60)
+      or  (bgp_session['holdtime'] is defined  and bgp_session['holdtime']  | int != 180) %}
+  neighbor {{ neighbor_addr }} timers {{ bgp_session['keepalive'] | default("60") }} {{ bgp_session['holdtime'] | default("180") }}
+{% endif %}
+!
```
venkatmahalingam, line 15: Can we start using the generic BGP schemas defined to suit any BGP use-cases instead of defining multiple new config-DB schemas per use-case? the below HLD captures all possible BGP configs for a typical DC deployment, this will be complex if we start defining the YANG models for various config-DB schemas defined for different use-cases e.g VOQ chassis.

https://github.com/Azure/SONiC/blob/e311a870a74d311bea1f3ece4a0d7042ed84d0dc/doc/mgmt/SONiC_Design_Doc_Unified_FRR_Mgmt_Interface.md

The backend handling for generic config DB schemas is already merged.
https://github.com/Azure/sonic-buildimage/pull/5142

dockers/docker-fpm-frr/frr/bgpd/templates/voq_chassis/peer-group.conf.j2
```diff
@@ -0,0 +1,23 @@
+!
+! template: bgpd/templates/voq_chassis/peer-group.conf.j2
+!
+  neighbor VOQ_CHASSIS_PEER peer-group
+  address-family ipv4
+{% if CONFIG_DB__DEVICE_METADATA['localhost']['type'] == 'ToRRouter' %}
+    neighbor VOQ_CHASSIS_PEER allowas-in 1
```
arlakshm, line 7: This check will not be true for internal peers in VoQ chassis, can we remove this ?

dockers/docker-fpm-frr/frr/bgpd/templates/voq_chassis/peer-group.conf.j2
```diff
@@ -0,0 +1,23 @@
...
+{% endif %}
+    neighbor VOQ_CHASSIS_PEER activate
+    neighbor VOQ_CHASSIS_PEER addpath-tx-all-paths
+    neighbor VOQ_CHASSIS_PEER soft-reconfiguration inbound
+  exit-address-family
+  address-family ipv6
+{% if CONFIG_DB__DEVICE_METADATA['localhost']['type'] == 'ToRRouter' %}
```
arlakshm, line 14: This check will not be true for internal peers in VoQ chassis, can we remove this ?

src/sonic-config-engine/tests/simple-sample-graph.xml
```diff
@@ -105,6 +132,11 @@
         <a:Hostname>ARISTA04T1</a:Hostname>
         <a:RouteMaps/>
       </a:BGPRouterDeclaration>
+      <a:BGPRouterDeclaration>
+        <a:ASN>65100</a:ASN>
+        <a:Hostname>CHASSIS_PEER</a:Hostname>
```
arlakshm, line 137: This hostname name is of remote Linecard ? What will be the hostname for multi ASIC Linecard.

src/sonic-config-engine/tests/simple-sample-graph.xml
```diff
@@ -20,6 +20,7 @@
         <Multihop>1</Multihop>
         <HoldTime>180</HoldTime>
         <KeepAliveTime>60</KeepAliveTime>
+        <VoQChassisInternal>false</VoQChassisInternal>
```
arlakshm, line 23: we don't need to set the VoQChassisInternal as false. if VoQChassisInternal  attribute is not present then we assume as False.

src/sonic-config-engine/minigraph.py
```diff
@@ -696,46 +697,40 @@ def parse_cpg(cpg, hname, local_devices=[]):
                     keepalive = 60
                 nhopself = 1 if session.find(str(QName(ns, "NextHopSelf"))) is not None else 0
 
+                # choose the right table and admin_status for the peer
+                voq_chassis = session.find(str(QName(ns, "VoQChassisInternal")))
+                if voq_chassis is not None and voq_chassis.text == "true":
+                    table = bgp_voq_chassis_sessions
+                    admin_status = None
```
arlakshm, line 704: should the internal sessions be up by default, if not the admin status will be changed by CLI ?

dockers/docker-fpm-frr/frr/bgpd/templates/voq_chassis/instance.conf.j2
eswaranb: Hi Venkat, when we discussed this at the design review for this feature, we settled on adding a new config DB schema for this (https://github.com/Azure/SONiC/pull/674/files)

The main reason for this was that the voq_chassis peering instances are meant to be 'internal' to the chassis and had specific configurations that won't appear in the other BGP configurations. If we use the generic BGP schema, how would we specify the configs that are only applicable for the voq_chassis instances?

venkatmahalingam: >> The main reason for this was that the voq_chassis peering instances are meant to be 'internal' to the chassis and had specific configurations that won't appear in the other BGP configurations.

I assume VoQ_ chassis configs will be present in config-DB and will be converted to FRR configs with the help of template files, when you say 'internal' what exactly the behaviour and how it wont appear in the other BGP configurations, I believe, FRR treats all BGP configurations including internal or external as same, I'm trying to understand the importance of defining new schemas, does it help to filter anyways? or it can be generic as well.

eswaranb: Yes, the voq_chassis configs will be present in Config DB and will be converted to FRR configs with the help of the template files. 

By 'internal', I mean that they won't be displayed in some of the Bgp related show commands or be affected by some of the bgp config commands. From FRR point of view, there is no difference, correct. Some FRR configs like "bgp bestpath as-path multipath-relax" is specific to the voq_chassis peers, for example. And we don't expect this config to be changed by the operator an runtime. 

If I understood you right, you are proposing to update the generic table to map to these new commands and have the voq_chassis specific flags/configs defined in the Sonic DB schema? I suppose that could be done, do you have an example of this? We spent quite a bit of time debating this at the design review and followed the existing model where each BGP peer type defined a new template.

venkatmahalingam: I did some of the VoQ BGP configs using generic BGP models and observed the following, see if this helps for VoQ.

Note: In order for the generic BGP model configs to handle using "src/sonic-frr-mgmt-framework/frrcfgd", we should set the field  "frr_mgmt_framework_config" to true in DEVICE_METADATA table, this was introduced to keep config handling by bgpcfgd as default in the community.

Also, config restore is supported, we read configs from config-DB and push to FRR during restore.

```

admin@sonic:~$ redis-cli -n 4
127.0.0.1:6379[4]> keys BGP*
1) "BGP_PEER_GROUP_AF|default|VOQ_CHASSIS_PEER|ipv6_unicast"
2) "BGP_GLOBALS_AF|default|ipv4_unicast"
3) "BGP_NEIGHBOR|default|10.10.10.10"
4) "BGP_GLOBALS_AF|default|ipv6_unicast"
5) "BGP_GLOBALS|default"
6) "BGP_PEER_GROUP|default|VOQ_CHASSIS_PEER"
7) "BGP_PEER_GROUP_AF|default|VOQ_CHASSIS_PEER|ipv4_unicast"
127.0.0.1:6379[4]> hgetall BGP_GLOBALS|default
 1) "as_path_mp_as_set"
 2) "false"
 3) "local_asn"
 4) "65100"
 5) "log_nbr_state_changes"
 6) "true"
 7) "always_compare_med"
 8) "false"
 9) "external_compare_router_id"
10) "false"
11) "fast_external_failover"
12) "true"
13) "network_import_check"
14) "true"
15) "load_balance_mp_relax"
16) "true"
17) "holdtime"
18) "30"
19) "ignore_as_path_length"
20) "false"
21) "keepalive"
22) "5"
127.0.0.1:6379[4]> hgetall BGP_GLOBALS_AF|default|ipv4_unicast
1) "route_flap_dampen"
2) "false"
3) "ibgp_equal_cluster_length"
4) "false"
5) "max_ebgp_paths"
6) "1"
7) "max_ibgp_paths"
8) "64"
127.0.0.1:6379[4]> hgetall BGP_GLOBALS_AF|default|ipv6_unicast
1) "ibgp_equal_cluster_length"
2) "false"
3) "max_ebgp_paths"
4) "1"
5) "max_ibgp_paths"
6) "64"
127.0.0.1:6379[4]> hgetall BGP_NEIGHBOR|default|10.10.10.10
1) "holdtime"
2) "240"
3) "peer_group_name"
4) "VOQ_CHASSIS_PEER"
5) "name"
6) "internal1"
7) "keepalive"
8) "60"
127.0.0.1:6379[4]> hgetall BGP_PEER_GROUP|default|VOQ_CHASSIS_PEER
 1) "min_adv_interval"
 2) "0"
 3) "ebgp_multihop"
 4) "false"
 5) "passive_mode"
 6) "false"
 7) "admin_status"
 8) "true"
 9) "conn_retry"
10) "30"
11) "asn"
127.0.0.1:6379[4]> hgetall BGP_PEER_GROUP_AF|default|VOQ_CHASSIS_PEER|ipv4_unicast
 1) "rrclient"
 2) "false"
 3) "admin_status"
 4) "true"
 5) "send_default_route"
 6) "false"
 7) "send_community"
 8) "both"
 9) "max_prefix_warning_only"
10) "false"
11) "NULL"
12) "NULL"
127.0.0.1:6379[4]>
127.0.0.1:6379[4]> hgetall BGP_PEER_GROUP_AF|default|VOQ_CHASSIS_PEER|ipv6_unicast
 1) "NULL"
 2) "NULL"
 3) "admin_status"
 4) "true"
 5) "rrclient"
 6) "false"
 7) "send_default_route"
 8) "false"
 9) "max_prefix_warning_only"
10) "false"
11) "send_community"
12) "both"
127.0.0.1:6379[4]>


admin@sonic:~$ vtysh
Hello, this is FRRouting (version 7.2-sonic).
Copyright 1996-2005 Kunihiro Ishiguro, et al.

sonic#
sonic#
sonic# show running-config bgpd
Building configuration...

Current configuration:
!
frr version 7.2-sonic
frr defaults traditional
hostname sonic
log syslog informational
log facility local4
agentx
no service integrated-vtysh-config
!
enable password zebra
password zebra
!
router bgp 65100
 bgp log-neighbor-changes
 no bgp default ipv4-unicast
 bgp bestpath as-path multipath-relax
 bgp bestpath compare-routerid
 bgp network import-check
 timers bgp 5 30
 neighbor VOQ_CHASSIS_PEER peer-group
 neighbor VOQ_CHASSIS_PEER remote-as 555
 neighbor VOQ_CHASSIS_PEER advertisement-interval 0
 neighbor VOQ_CHASSIS_PEER timers connect 30
 neighbor 10.10.10.10 peer-group VOQ_CHASSIS_PEER
 neighbor 10.10.10.10 description internal1
 neighbor 10.10.10.10 timers 60 240
 !
 address-family ipv4 unicast
  neighbor VOQ_CHASSIS_PEER activate
  no neighbor VOQ_CHASSIS_PEER send-community large
  no neighbor 10.10.10.10 send-community large
  maximum-paths 1
  maximum-paths ibgp 64
 exit-address-family
 !
 address-family ipv6 unicast
  neighbor VOQ_CHASSIS_PEER activate
  no neighbor VOQ_CHASSIS_PEER send-community large
  maximum-paths 1
  maximum-paths ibgp 64
 exit-address-family
!
line vty
!
end
sonic# show bgp ipv4 unicast summary
BGP router identifier 100.94.134.22, local AS number 65100 vrf-id 0
BGP table version 0
RIB entries 0, using 0 bytes of memory
Peers 1, using 20 KiB of memory
Peer groups 1, using 64 bytes of memory

Neighbor        V         AS MsgRcvd MsgSent   TblVer  InQ OutQ  Up/Down State/PfxRcd
10.10.10.10     4        555       0       0        0    0    0    never       Active

Total number of neighbors 1
Total number of neighbors established 0
sonic# show bgp ipv6 unicast summary
BGP router identifier 100.94.134.22, local AS number 65100 vrf-id 0
BGP table version 0
RIB entries 0, using 0 bytes of memory
Peers 1, using 20 KiB of memory
Peer groups 1, using 64 bytes of memory

Neighbor        V         AS MsgRcvd MsgSent   TblVer  InQ OutQ  Up/Down State/PfxRcd
10.10.10.10     4        555       0       0        0    0    0    never       Active

Total number of neighbors 1
Total number of neighbors established 0
sonic#


```

lguohan: @venkatmahalingam, this feature is designed to work with bgpcfgd, not frrcfgd.

venkatmahalingam: Okay, no problem.

eswaranb: Aah, I did not realize the distinction earlier. @venkatmahalingam please let us know if there is any change required here.

venkatmahalingam: @eschwabe There are two approaches we follow in the community for FRR configurations 1) bgpcfgd 2) frrcfgd, @lguohan mentioned that VoQ chassis is designed to work with bgpcfgd, so, we can go ahead with that, changing to adapt to frrcfgd might be a lot of work at this point.

dockers/docker-fpm-frr/frr/bgpd/templates/voq_chassis/peer-group.conf.j2
jmmikkel: If one puts the VoQ chassis in the TOR role, then other templates would set up allowas-in on the eBGP connections. If it's not also set on the iBGP connections, any routes that do have a loop would be dropped by the iBGP peers. This was intended to avoid that routing asymmetry if ToRRouter is set. If it's not, there is no harm.

src/sonic-config-engine/tests/simple-sample-graph.xml
jmmikkel: I don't fully understand the question. The hostname would be what someone picked for it. Are you saying you want this name changed?

src/sonic-config-engine/tests/simple-sample-graph.xml
jmmikkel: There are three logical states: 1) VoQChassisInternal is not present, 2) VoQChassisInternal is present and true, 3) VoQChassisInternal is present and false.

While it may not be intended to set it to false, it seemed to me we should test all three cases. However, if that is a problem I guess I can remove it.

src/sonic-config-engine/minigraph.py
jmmikkel: It was agreed the sessions would be up by default so that routing works as intended as soon as any eBGP session comes up. The admin status should be changeable with the CLI though this PR does not cover the CLI. Are you saying admin_status should be 'up', rather than not-specified? Not-specified is not 'down' so the sessions would be configured up either way.

minionatwork (dismissed):

eswaranb (approved):

src/sonic-config-engine/minigraph.py
arlakshm: My point was should we set the admin_status to up here as we want these session to up by default.

arlakshm (approved):

src/sonic-config-engine/minigraph.py
jmmikkel: Thanks. I have changed it to 'up'. Please take a look!

arlakshm:
@jmmikkel, the unit tests are failing. Can you take a look 
Log file
https://dev.azure.com/mssonic/build/_build/results?buildId=8902&view=logs&j=83516c17-6666-5250-abde-63983ce72a49&t=c10d5f44-55ce-55d7-7975-407ed75d9a96&l=1861

jmmikkel:
> @jmmikkel, the unit tests are failing. Can you take a look
> Log file
> https://dev.azure.com/mssonic/build/_build/results?buildId=8902&view=logs&j=83516c17-6666-5250-abde-63983ce72a49&t=c10d5f44-55ce-55d7-7975-407ed75d9a96&l=1861

Ugh, how embarrassing, I had tested it to work in our local build system, then copied over the change wrong.

jmmikkel:
@arlakshm I don't think this failure is due to my changes.
"FAIL: advanced-reboot.ReloadTest",
"FAILED:dut:Error collecting current date from DUT: empty value returned",

Is there a way to retry this test or does it need a full rebuild? Thanks

arlakshm (approved):

minionatwork (approved):

vganesan-nokia (approved):

PR #5628, opened 2020-10-14
Add teamd as a depedent service to swss

judyjoseph:
**- Why I did it**
On teamd docker restart, the swss and syncd needs to be restarted as there are dependent resources present.

**- How I did it**
Add the teamd as a dependent service for swss
Updated the docker-wait script to handle service and dependent services separately.
Handle the case of warm-restart for the dependent service   

**- How to verify it**

Verified the following scenario's with the following testbed 
VM1 ----------------------------[DUT 6100] -----------------------VM2,  ping traffic continuous between VMs

1. Stop teamd docker alone  
      >  swss, syncd dockers seen going away
      >  The LAG reference count error messages seen for a while till swss docker stops.
      >  Dockers back up.

2. Enable WR mode for teamd. Stop teamd docker alone  
      >  swss, syncd dockers not removed.
      >  The LAG reference count error messages not seen
      >  Repeated stop teamd docker test - same result, no effect on swss/syncd.

3. Stop swss docker. 
      >  swss, teamd, syncd goes off - dockers comes back correctly, interfaces up

4. Enable WR mode for swss . Stop swss docker 
      >  swss goes off not affecting syncd/teamd dockers.

5. Config reload 
      > no reference counter error seen, dockers comes back correctly, with interfaces up

6. Warm reboot, observations below
	 > swss docker goes off first 
	 > teamd + syncd goes off to the end of WR process.
 	 > dockers comes back up fine.
	 > ping traffic between VM's was NOT HIT

7. Fast reboot, observations below
	 > teamd goes off first ( **confirmed swss don't exit here** )
	 > swss goes off next 
	 > syncd goes away at the end of the FR process
	 > dockers comes back up fine.
	 > there is a traffic HIT as per fast-reboot

8. Verified in multi-asic platform, the tests above other than WR/FB scenarios

- [ ] 201811
- [ ] 201911
- [ ] 202006

### DIFF
--- a/files/image_config/misc/docker-wait-any
+++ b/files/image_config/misc/docker-wait-any
@@ -44,8 +44,8 @@ def wait_for_container(docker_client, container_name):
 
         # If this is a dependent service and warm restart is enabled for the system/container,
         # OR if the system is going through a fast-reboot, DON'T signal main thread to exit
-        if container_name in g_dep_services and
-           (device_info.is_warm_restart_enabled(container_name) or device_info.is_fast_reboot_enabled()):
+        if (container_name in g_dep_services and
+            (device_info.is_warm_restart_enabled(container_name) or device_info.is_fast_reboot_enabled())):
             continue
 
         # Signal the main thread to exit
--- a/files/image_config/misc/docker-wait-any
+++ b/files/image_config/misc/docker-wait-any
@@ -17,28 +17,36 @@
 import sys
 import time
 import argparse
+import subprocess
 import threading
 from docker import APIClient
 from swsssdk import SonicV2Connector
+from sonic_py_common import logger
+
+SYSLOG_IDENTIFIER = 'docker-wait-any'
+
+# Global logger instance
+log = logger.Logger(SYSLOG_IDENTIFIER)
 
 # Instantiate a global event to share among our threads
 g_thread_exit_event = threading.Event()
 g_service = []
 g_dep_services = []
 
-def check_warm_restart_state(container_name):
+# Check if System warm reboot, or Container warm restart is enabled.
+def is_warm_restart_enabled(container_name):
     state_db = SonicV2Connector(host='127.0.0.1')
     state_db.connect(state_db.STATE_DB, False)
 
-    # Get the system warm reboot enable state
     TABLE_NAME_SEPARATOR = '|'
     prefix = 'WARM_RESTART_ENABLE_TABLE' + TABLE_NAME_SEPARATOR
+
+    # Get the system warm reboot enable state
     _hash = '{}{}'.format(prefix, 'system')
     wr_system_state = state_db.get(state_db.STATE_DB, _hash, "enable")
     wr_enable_state = True if wr_system_state == "true" else False
 
     # Get the container warm reboot enable state
-    prefix = 'WARM_RESTART_ENABLE_TABLE' + TABLE_NAME_SEPARATOR
     _hash = '{}{}'.format(prefix, container_name)
     wr_container_state = state_db.get(state_db.STATE_DB, _hash, "enable")
     wr_enable_state |= True if wr_container_state == "true" else False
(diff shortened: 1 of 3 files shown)


files/image_config/misc/docker-wait-any
```diff
@@ -14,39 +14,81 @@
...
+    parser  = argparse.ArgumentParser(description='Wait for dependent docker services',
+                                      version='1.0.0',
+                                      formatter_class=argparse.RawTextHelpFormatter,
+                                      epilog="""
+Examples:
+  docker-wait-any -s swss -d syncd teamd
+""")
+
+    parser.add_argument('-s','--service', nargs='+', default=None, help='The service which is waiting for dependent services')
+    parser.add_argument('-d','--dependent', nargs='*', default=None, help='The dependent services')
```
jleveque, line 78: Curious: why the need to differentiate main service from dependencies?

judyjoseph: As per my current understanding, It is needed for warm boot scenarios. In the function **wait_for_container()**

In case of the main service(here swss)
``` 
  irrespective of whether it in WR or non-WR mode) we signal main thread with g_thread_exit_event.set(). 
```
In case of the dependent service( eg: teamd)
```
if dep service is in WR mode:
   don't signal main thread, instead go back and wait for the same container when it is back up after warm-boot. 
   We need to keep track of that container in the current "/usr/bin/docker-wait" context itself.
else:
   signal main thread with g_thread_exit_event.set()
```

files/image_config/misc/docker-wait-any
```diff
@@ -17,49 +17,72 @@
 import sys
 import time
 import argparse
+import subprocess
 import threading
 from docker import APIClient
 from swsssdk import SonicV2Connector
+from sonic_py_common import logger
```
jleveque: Suggest reorganizing imports as follows:

```python
import argparse
import subprocess
import sys
import threading
import time

from docker import APIClient
from sonic_py_common import logger
from swsssdk import SonicV2Connector
```

files/image_config/misc/docker-wait-any
```diff
@@ -17,49 +17,72 @@
...
     wr_container_state = state_db.get(state_db.STATE_DB, _hash, "enable")
     wr_enable_state |= True if wr_container_state == "true" else False
 
     state_db.close(state_db.STATE_DB)
     return wr_enable_state
 
+# Check if System fast reboot is enabled.
+def is_fast_reboot_enabled():
+    fb_system_state = 0
+    cmd='sonic-db-cli STATE_DB get "FAST_REBOOT|system"'
```
jleveque, line 60: Add spaces around `=`

jleveque (changes requested):

judyjoseph:
Updated the tests done to verify this fix in PR discussion

jleveque (dismissed):
Looks good to me. Please wait for other reviewers.

files/image_config/misc/docker-wait-any
```diff
@@ -14,39 +14,108 @@
...
-    print("Usage: {} <container_name> [<container_name> ...]".format(sys.argv[0]))
-    sys.exit(1)
+    state_db.close(state_db.STATE_DB)
+    return wr_enable_state
 
+# Check if System fast reboot is enabled.
+def is_fast_reboot_enabled():
+    fb_system_state = 0
+    cmd = 'sonic-db-cli STATE_DB get "FAST_REBOOT|system"'
+    proc = subprocess.Popen(cmd, shell=True, stdout=subprocess.PIPE)
```
abdosi, line 62: @judyjoseph why can't we just read from State DB same as is_warm_restart_enabled function. Do we need to do subprocess ?

files/image_config/misc/docker-wait-any
```diff
@@ -14,39 +14,108 @@
...
+log = logger.Logger(SYSLOG_IDENTIFIER)
 
 # Instantiate a global event to share among our threads
 g_thread_exit_event = threading.Event()
+g_service = []
+g_dep_services = []
+
+# Check if System warm reboot, or Container warm restart is enabled.
+def is_warm_restart_enabled(container_name):
+    state_db = SonicV2Connector(host='127.0.0.1')
```
abdosi, line 39: @judyjoseph I am wondering should we move this function to some common place as this looks to be generic and can be used by any service/application if needed.

files/image_config/misc/docker-wait-any
judyjoseph: The way we write the FAST_REBOOT state is using the SET redis API **redis-cli -n 6 SET "FAST_REBOOT|system" "1"**. The dbconnector/interface.py API's available are all using redis **HGET** API's which  returns the value associated with field in the hash stored at key.

Hence had to use the different approach with sonic-db-cli to get FAST_REBOOT|system

Options, we can change FAST_REBOOT to also start using HGET/HSET, may be as a separate change ?

files/image_config/misc/docker-wait-any
judyjoseph: I could move this to sonic_py_common/device_info.py -- if that is a good place

judyjoseph: taken care of

yxieca (approved):

judyjoseph:
retest vs please

abdosi (dismissed):

src/sonic-py-common/sonic_py_common/device_info.py
```diff
@@ -428,3 +428,38 @@ def get_system_routing_stack():
...
+    wr_container_state = state_db.get(state_db.STATE_DB, _hash, "enable")
+    wr_enable_state |= True if wr_container_state == "true" else False
+
+    state_db.close(state_db.STATE_DB)
+    return wr_enable_state
+
+# Check if System fast reboot is enabled.
+def is_fast_reboot_enabled():
+    fb_system_state = 0
+    cmd = 'sonic-db-cli STATE_DB get "FAST_REBOOT|system"'
```
jleveque, line 456: Maybe add a TODO or FIXME comment here as a reminder to clean this up once the entry is changed to a hash value.

jleveque (approved):

src/sonic-py-common/sonic_py_common/device_info.py
judyjoseph: @jleveque  I have added this code reference in the **issue** which Ying raised to track this cleanup item (https://github.com/Azure/sonic-buildimage/issues/5690#issuecomment-714258980). 
Is that ok ?

judyjoseph:
retest mellanox please

src/sonic-py-common/sonic_py_common/device_info.py
jleveque: That's fine. Thanks.

files/image_config/misc/docker-wait-any
judyjoseph: I have added this request in the issue which Ying raised to track this cleanup item https://github.com/Azure/sonic-buildimage/issues/5690#issuecomment-714258980

judyjoseph:
retest mellanox please

PR #5641, opened 2020-10-15
[Host OS][docker-vs] No longer install 'wheel' package explicitly

jleveque:
**- Why I did it**

It should no longer be necessary to explicitly install the 'wheel' package, as SONiC packages built as wheels should specify 'wheel' as a dependency in their setup.py files. Therefore, pip[3] should check for the presence of 'wheel' and install it if it isn't present before attempting to call 'setup.py bdist_wheel' to install the package.

- [ ] 201811
- [ ] 201911
- [ ] 202006


jleveque:
Retest vs please

daall (approved):

PR #5643, opened 2020-10-16
[Mellanox] Add a configuration to delay start xcvrd for fast-reboot

Junchao-Mellanox:
**- Why I did it**

Delay start xcvrd on mellanox platform to save CPU cost during fast-reboot. With this change, fast-reboot saves about 8~9 seconds.

**- How I did it**

Add a sleep before xcvrd start in pmon supervisor.conf

**- How to verify it**

Manual test

- [ ] 201811
- [x] 201911
- [x] 202006

### DIFF
--- a/device/mellanox/x86_64-mlnx_msn2700-r0/pmon_daemon_control.json
+++ b/device/mellanox/x86_64-mlnx_msn2700-r0/pmon_daemon_control.json
@@ -1,5 +1,6 @@
 {
     "skip_ledd": true,
-    "skip_fancontrol": true
+    "skip_fancontrol": true,
+    "delay_xcvrd": true
 }
 
--- a/dockers/docker-platform-monitor/docker-pmon.supervisord.conf.j2
+++ b/dockers/docker-platform-monitor/docker-pmon.supervisord.conf.j2
@@ -79,7 +79,11 @@ dependent_startup_wait_for=start:exited
 
 {% if not skip_xcvrd %}
 [program:xcvrd]
+{% if delay_xcvrd %}
+command=bash -c "sleep 30 && /usr/local/bin/xcvrd"
+{% else %}
 command=/usr/local/bin/xcvrd
+{% endif %}
 priority=6
 autostart=false
 autorestart=false


keboliu (approved):

Junchao-Mellanox:
retest vs please

stepanblyschak (approved):

device/mellanox/x86_64-mlnx_msn2700-r0/pmon_daemon_control.json
```diff
@@ -1,5 +1,6 @@
 {
     "skip_ledd": true,
-    "skip_fancontrol": true
+    "skip_fancontrol": true,
+    "delay_xcvrd": true
```
jleveque, line 4: Should this be configurable, or should we simply always delay the xcvrd start?

liat-grozovik: IMO we should simply always delay, like it is done with counters.

liat-grozovik (approved):

device/mellanox/x86_64-mlnx_msn2700-r0/pmon_daemon_control.json
Junchao-Mellanox: I heard that there are vendors that depend on the xcvrd to finish their fast-reboot, I am not so sure about that, I was not in that loop before. But if there is no vendor need that, we can simply always delay it. @jleveque any suggestion?

jleveque: I am not familiar with the requirements of fast-reboot on various platforms. @yxieca, @lguohan: are you aware of any platforms which depend on xcvrd to finish fast-reboot?

stepanblyschak: @jleveque As we discussed last time, besides of just monitoring xcvrd dynamically tunes transceivers on some platforms. Thus, if not, on fast-reboot there might be port operationally down longer time.
Guohan can correct me if this is the case.

Junchao-Mellanox: @lguohan could you please share your view on this? Thanks!

liat-grozovik:
@lguohan could you please refer to the open issue above?
this fix is important to fastboot in order to be able to meet the req and make the cpu faster at that period same as we do for counters.

dockers/docker-platform-monitor/docker-pmon.supervisord.conf.j2
```diff
@@ -79,7 +79,11 @@ dependent_startup_wait_for=start:exited
 
 {% if not skip_xcvrd %}
 [program:xcvrd]
+{% if delay_xcvrd %}
```
lguohan, line 82: what if delay_xrcvd is not defined? will this throw exception?

device/mellanox/x86_64-mlnx_msn2700-r0/pmon_daemon_control.json
lguohan: @jleveque , it is customizable per platform, therefore it is ok. for mostly for the platform i think delay is not used.

jleveque: @lguohan: My question is do we *need* to make this customizable, or should we simply delay it on all platforms? If delaying on all platforms will cause no harm with fast-reboot, then I think it should be delayed on all platforms to make things more universal and also to prevent adding extra complexity.

dockers/docker-platform-monitor/docker-pmon.supervisord.conf.j2
Junchao-Mellanox: Just did a quick check. It won't throw exception if delay_xcvrd is not defined.

liat-grozovik:
@lguohan how can we proceed with this PR? we would like to have it in 201911 to ensure fastboot req are met.

liat-grozovik:
@Junchao-Mellanox lets have the delay by default for all platforms and add an option for platform specific to disable the delay?

device/mellanox/x86_64-mlnx_msn2700-r0/pmon_daemon_control.json
dgsudharsan: @jleveque In our platforms xcvrd is used to program pre-emphasis and thus a delay would might even lead to ports not coming up. Moreover ins 400G platforms complex initializations need to be done to bring port up and delaying xcvrd there would affect fastboot. Hence we prefer having a platform knob and by default there should be no delay.

jleveque: Thanks, @dgsudharsan. Then we will proceed with this approach, adding a new `delay_xcvrd` parameter.

liat-grozovik: does this means we can sign off it and have it taken to 201911 as well?

jleveque (approved):

device/mellanox/x86_64-mlnx_msn2700-r0/pmon_daemon_control.json
jleveque: I have signed off.

liat-grozovik:
Thanks!

PR #5644, opened 2020-10-16
[synchronous-mode] Add template file for synchronous mode

shi-su:
**- Why I did it**
The orchagent and syncd need to have the same default synchronous mode configuration. This PR adds a template file to translate the default value in CONFIG_DB (empty field) to an explicit mode so that the orchagent and syncd could have the same default mode.

**- How I did it**
Add a template file to /etc/sonic/ directory so that both syncd and orchagent could access the same template file to determine the default mode.

**- How to verify it**
Verified with kvm and dvs locally

- [ ] 201811
- [ ] 201911
- [ ] 202006


platform/p4/docker-sonic-p4/orchagent.sh
```diff
@@ -14,7 +14,7 @@ ORCHAGENT_ARGS="-d /var/log/swss "
 ORCHAGENT_ARGS+="-b 8192 "
 
 # Set synchronous mode if it is enabled in CONFIG_DB
-SYNC_MODE=$(sonic-cfggen -d -v DEVICE_METADATA.localhost.synchronous_mode)
+SYNC_MODE=$(sonic-cfggen -d -t /etc/sonic/synchronous_mode.j2 | jq -r '.synchronous_mode')
```
qiluo-msft, line 17: >sonic-cfggen [](start = 12, length = 12)

sonic-cfggen is slow in the warm/fast reboot critical timeline. So we optimize them aggressively.
It is possible to limit only one call in each script to generate everything into a big variable, and extract each config from the variable.

shi-su: Reduced `sonic-cfggen` usage.

shi-su:
retest vsimage please

rules/scripts.mk
```diff
@@ -23,11 +23,15 @@ $(SYSCTL_NET_CONFIG)_PATH = files/image_config/sysctl
 UPDATE_CHASSISDB_CONFIG_SCRIPT = update_chassisdb_config
 $(UPDATE_CHASSISDB_CONFIG_SCRIPT)_PATH = files/scripts
 
+SWSS_SYNCD_VARS_TEMPLATE = swss_syncd_vars.j2
```
qiluo-msft, line 26: >SWSS_SYNCD_VARS_TEMPLATE [](start = 0, length = 24)

Wondering why original code don't need it.

platform/vs/docker-sonic-vs/orchagent.sh
```diff
@@ -6,8 +6,18 @@ else
...
+EXIT_SWSS_VARS_FILE_NOT_FOUND=1
+SWSS_VARS_FILE=/etc/sonic/swss_syncd_vars.j2
+
+if [ ! -f "$SWSS_VARS_FILE" ]; then
+    echo "SWSS vars template file not found"
+    exit $EXIT_SWSS_VARS_FILE_NOT_FOUND
+fi
+
+# Retrieve SWSS vars from sonic-cfggen
+SWSS_VARS=$(sonic-cfggen -d -y /etc/sonic/sonic_version.yml -t $SWSS_VARS_FILE)
```
qiluo-msft, line 18: >$SWSS_VARS_FILE [](start = 63, length = 15)

If SWSS vars template file not found, the command substitution will fail, so you can catch the error and exit. It is too verbose to check every input files.

dockers/docker-orchagent/orchagent.sh
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 EXIT_SWSS_VARS_FILE_NOT_FOUND=1
-SWSS_VARS_FILE=/usr/share/sonic/templates/swss_vars.j2
+SWSS_VARS_FILE=/etc/sonic/swss_syncd_vars.j2
```
qiluo-msft, line 4: >/etc/sonic/swss_syncd_vars.j2 [](start = 15, length = 29)

Does it make more sense to keep it in original directory? User are not expected to modify this file.
BTW, keep the original filename still make sense to me.

rules/scripts.mk
shi-su: It is needed here to make the file available to be copied to dvs and p4 docker. The template file was not available in the platforms.

platform/vs/docker-sonic-vs/orchagent.sh
shi-su: Removed the verbose check for the template file.

dockers/docker-orchagent/orchagent.sh
shi-su: The original directory `/usr/share/sonic/templates` is only available in swss docker. The reason to move it to `\etc\sonic` is to make the same template file accessible to both swss and syncd. Also, as this template is going to be accessed by syncd as well, I renamed to reflect this.

qiluo-msft (dismissed):
LGTM. Please wait @tahmed-dev review.

shi-su:
retest mellanox please

lguohan:
@qiluo-msft, what do you need @tahmed-dev to review? can you mark it?

qiluo-msft:
@tahmed-dev Could you help review `sonic-cfggen` related code changes?

tahmed-dev:
> @tahmed-dev Could you help review `sonic-cfggen` related code changes?

sure.

tahmed-dev:
The change LGTM, however I just find the case for the `swss_vars.j2` template files residing in `/etc/sonic` to be an outlier.

tahmed-dev:
@shi-su wonder if symlink-ing the file in the source code base would be a viable solution.

qiluo-msft:
> FROM docker-base

Do you need to change for jessie or stretch?

---
Refers to: dockers/docker-config-engine/Dockerfile.j2:1 in b7eaa05. [](commit_id = b7eaa053021a00e886c54767ca6020bd9142fd42, deletion_comment = False)

rules/scripts.mk
```diff
@@ -23,11 +23,15 @@ $(SYSCTL_NET_CONFIG)_PATH = files/image_config/sysctl
 UPDATE_CHASSISDB_CONFIG_SCRIPT = update_chassisdb_config
 $(UPDATE_CHASSISDB_CONFIG_SCRIPT)_PATH = files/scripts
 
+SWSS_VARS_TEMPLATE = swss_vars.j2
+$(SWSS_VARS_TEMPLATE)_PATH = files/build_templates
+
 SONIC_COPY_FILES += $(CONFIGDB_LOAD_SCRIPT) \
```
qiluo-msft, line 29: >SONIC_COPY_FILES [](start = 0, length = 16)

Why we don't need to change this before?

rules/scripts.mk
```diff
@@ -23,11 +23,15 @@ $(SYSCTL_NET_CONFIG)_PATH = files/image_config/sysctl
 UPDATE_CHASSISDB_CONFIG_SCRIPT = update_chassisdb_config
 $(UPDATE_CHASSISDB_CONFIG_SCRIPT)_PATH = files/scripts
 
+SWSS_VARS_TEMPLATE = swss_vars.j2
+$(SWSS_VARS_TEMPLATE)_PATH = files/build_templates
```
tahmed-dev, line 27: Instead of copying the file to `files/build_templates`, we can instead create symlink in the new dockers requiring it in the source code base and let the regular docker creation process copy the file.

rules/scripts.mk
shi-su: The template file was previously put in the docker-orchagent directory, so it could be seen by orchagent docker. Since we want to put this file into more than one docker, I moved it into a shared place and added this change.

rules/scripts.mk
shi-su: The builder has a staging directory when building the docker. As the symlink only contains the relative path of in the source code base, it cannot find the linked file in build.

shi-su:
retest vs please

qiluo-msft (approved):

shi-su:
retest broadcom please

shi-su:
retest broadcom please

rules/scripts.mk
lguohan: agree with @shi-su

PR #5651, opened 2020-10-16
Fixes the DPKG cache break because of SAI package content changes wit…

Kalimuthu-Velappan:
…the docker rebuild

- The issue is that the SAI package content is changed without changing its version. The DPKG caches the wrong version of SAI package. 
- The fix is to include the SAI package content header for SHA calcaulation. This will detect if there is any change in the SAI package.

**- Why I did it**

**- How I did it**

**- How to verify it**

- [ ] 201811
- [ ] 201911
- [ ] 202006

### DIFF
--- a/platform/broadcom/sai.dep
+++ b/platform/broadcom/sai.dep
@@ -3,12 +3,16 @@
 SPATH       := $($(BRCM_SAI)_SRC_PATH)
 DEP_FILES   := $(SONIC_COMMON_FILES_LIST) platform/broadcom/sai.mk platform/broadcom/sai.dep   
 DEP_FILES   += $(SONIC_COMMON_BASE_FILES_LIST)
+# Get the Latest HTTP Header and calculate the SHA value as it is a softlink that always points to LATEST_INT_OCP_SAI_X.X.X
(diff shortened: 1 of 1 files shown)


platform/broadcom/sai.dep
```diff
@@ -3,12 +3,16 @@
 SPATH       := $($(BRCM_SAI)_SRC_PATH)
 DEP_FILES   := $(SONIC_COMMON_FILES_LIST) platform/broadcom/sai.mk platform/broadcom/sai.dep   
 DEP_FILES   += $(SONIC_COMMON_BASE_FILES_LIST)
+# Get the Latest HTTP Header and calculate the SHA value as it is a softlink that always points to LATEST_INT_OCP_SAI_X.X.X
```
qiluo-msft, line 6: >HTTP Header [](start = 17, length = 11)

Is it better to download the whole file and calc the SHA value? It is tens of MB so not a big issues.

Kalimuthu-Velappan: Adding the complete file content for the SHA calculation might cause a delay in the initial build phase especifically in the slow network as these files are included in the begining of the make process.  The HTTP header itself is sufficient and it is always reflects the change if there is any modification in the content.  I feel this is sufficient, your thoughts ?

qiluo-msft (approved):

PR #5657, opened 2020-10-17
[procdockerstatsd] Convert to Python 3

jleveque:
**- Why I did it**

Python 2 is deprecated and no longer supported. This is part of moving all Python code in SONiC to Python 3.

**- How I did it**

Make procdockerstatsd Python 3-compliant and set interpreter to python3 in shebang. Also some other cleanup to improve code reuse.

**- How to verify it**

Run procdockerstatsd, ensure it functions correctly.

- [ ] 201811
- [ ] 201911
- [ ] 202006

### DIFF
--- a/files/image_config/procdockerstatsd/procdockerstatsd
+++ b/files/image_config/procdockerstatsd/procdockerstatsd
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 '''
 procdockerstatsd
 Daemon which periodically gathers process and docker statistics and pushes the data to STATE_DB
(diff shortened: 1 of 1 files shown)


lguohan (approved):

lguohan:
i think it is better to put hostcfgd, caclmgrd, and this one into a host-service wheel package so that later we can add unit test framework to that. it can be a separate pr.

jleveque:
> i think it is better to put hostcfgd, caclmgrd, and this one into a host-service wheel package so that later we can add unit test framework to that. it can be a separate pr.

Agreed. Once they are all converted to Python 3 we should do this.

pra-moh (approved):

Issue #5570, opened 2020-10-09
orchagent continuously prints out "set: setting attribute 0x10000004 status: SAI_STATUS_SUCCESS"


daall:
**Description**
Orchagent continually sprays this message to the syslog:
```
set: setting attribute 0x10000004 status: SAI_STATUS_SUCCESS
```

Which causes this to occur:
```
Oct  9 00:33:00.374116 sonic INFO swss#rsyslogd: imuxsock[pid: 42, name: /usr/bin/orchagent] from <sonic:orchagent>: begin to drop messages due to rate-limiting
Oct  9 00:36:57.412514 sonic INFO swss#rsyslogd: imuxsock[pid: 42, name: /usr/bin/orchagent]: 32511 messages lost due to rate-limiting
```

Which masks legitimate syslog messages from orchagent (e.g. "create ACL table foo").

**Steps to reproduce the issue:**
1. Start the DUT.
2. Check the syslog.

**Describe the results you received:**
```
Oct  9 00:33:00.374116 sonic INFO swss#rsyslogd: imuxsock[pid: 42, name: /usr/bin/orchagent] from <sonic:orchagent>: begin to drop messages due to rate-limiting
Oct  9 00:36:57.412514 sonic INFO swss#rsyslogd: imuxsock[pid: 42, name: /usr/bin/orchagent]: 32511 messages lost due to rate-limiting
```

**Describe the results you expected:**
There shouldn't be any throttling/need to throttle while orchagent is idle and no updates are being received from neighbors/bgp/etc.

theasianpianist:
From what I can tell, the SAI_STATUS_SUCCESS message does not seem to be causing the rate limiting. Will continue investigating.

@daall next time you notice this, can you `grep` for "begin to drop messages" in your syslogs?

PR #5666, opened 2020-10-19
[docker-base]: Rate limit priority INFO and lower in syslog

theasianpianist:

Temporary fix for #5570 

**- Why I did it**
There is currently a bug where messages from swss with priority lower than the current log level are still being counted against the syslog rate limiting threshhold. This leads to rate-limiting in syslog when the rate-limiting conditions have not been met, which causes several sonic-mgmt tests to fail since they are dependent on LogAnalyzer. It also omits potentially useful information from the syslog. Only rate-limiting messages of level INFO and lower allows these tests to pass successfully.

**- How I did it**
Add `$SystemLogRateLimitSeverity` option to `rsyslog.conf` in the base Dockerfiles

**- How to verify it**
Build a new orchagent image and install it. Run the ACL tests, which should now pass. No rate limiting should occur in the syslogs.

- [ ] 201811
- [ ] 201911
- [ ] 202006

### DIFF
--- a/dockers/docker-base-buster/etc/rsyslog.conf
+++ b/dockers/docker-base-buster/etc/rsyslog.conf
@@ -12,7 +12,7 @@
 $ModLoad imuxsock # provides support for local system logging
 
 #
-# Set a rate limit on messages from the container
+# Set a rate limit on messages from the container of priority INFO or lower (level 6 and above)
 #
 $SystemLogRateLimitInterval 300
 $SystemLogRateLimitBurst 20000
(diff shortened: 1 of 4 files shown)


jleveque (changes requested):
Can you also please update the related comments to read:

```
# Set a rate limit on messages from the container of INFO level (level 6) or lower
```

jleveque:
You may also want to make the same change here: files/image_config/rsyslog/rsyslog-container.conf.j2

jleveque (approved):

lguohan (approved):

PR #5668, opened 2020-10-19
[201911][caclmgrd] Prevent unnecessary iptables updates

jleveque:
*This PR is a backport of https://github.com/Azure/sonic-buildimage/pull/5312 to the 201911 branch*

**- Why I did it**

When a large number of changes occur to the ACL table of Config DB, caclmgrd will get flooded with notifications, and previously, it would regenerate and apply the iptables rules for each change, which is unnecessary, as the iptables rules should only get applied once after the last change notification is received. If the ACL table contains a large number of control plane ACL rules, this could cause a large delay in caclmgrd getting the rules applied.

This patch causes caclmgrd to delay updating the iptables rules until it has not received a change notification for at least 0.5 seconds.

**- How I did it**

When an ACL table change notification is received for a particular namespace, we spawn an update thread for that namespace, which sits in a loop, sleeps 0.5 seconds and checks if any new change notifications were received while it was asleep. If new notification were received, it goes back to sleep and repeats the process until no new notifications were received while it was asleep. It then generates and applies the iptables rules.

**- How to verify it**

Load a large set of control plane ACL rules (see https://github.com/Azure/sonic-buildimage/issues/5275 for an example)

### DIFF
--- a/files/image_config/caclmgrd/caclmgrd
+++ b/files/image_config/caclmgrd/caclmgrd
@@ -15,6 +15,8 @@ try:
     import os
     import subprocess
     import sys
+    import threading
+    import time
 
     from sonic_py_common import daemon_base, device_info
     from swsscommon import swsscommon
@@ -25,6 +27,8 @@ except ImportError as err:
 VERSION = "1.0"
 
 SYSLOG_IDENTIFIER = "caclmgrd"
+
+DEFAULT_NAMESPACE = ''
 
 
 # ========================== Helper Functions =========================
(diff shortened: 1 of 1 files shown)


jleveque:
Retest mellanox please

abdosi:
@jleveque I think cherry-pick should be fine for this. Will try to cherry-pick first then we can decide to merge/close this PR

abdosi:
@jleveque Cherry-pick is clean. You can close this PR

jleveque:
Closing as unnecessary, as master branch PR was successfully cherry-picked into 201911 branch.

PR #5686, opened 2020-10-21
[Arista] Update driver submodules

Staphylo:
- Enable thermalctld support for our platforms
 - Fix Chassis.get_num_sfp which had an off by one
 - Implement read_eeprom and write_eeprom in SfpBase
 - Refactor of Psus and PsuSlots. Psus they are now detected and metadata reported
 - Improvements to modular support


rules/docker-platform-monitor.mk
```diff
@@ -43,6 +43,7 @@ $(DOCKER_PLATFORM_MONITOR)_RUN_OPT += --privileged -t
 $(DOCKER_PLATFORM_MONITOR)_RUN_OPT += -v /etc/sonic:/etc/sonic:ro
 
 # Mount Arista python library on Aboot images to be used by plugins
+$(DOCKER_PLATFORM_MONITOR)_aboot_RUN_OPT += -v /run/arista:/run/arista:ro
```
lguohan, line 46: why do we need this?

Staphylo: As part of the platform initialization, our platform library detects the hardware by performing some IO and stores this information in temporary cache files.
As of https://github.com/aristanetworks/sonic/commit/b29d50b1764413131c8c5232bf38d960de642586 all of them are stored under /run/arista.
This currently cover platform eeproms as well as PSU metadata and essentially avoids performing IO and detection again in pmon processes. The detection mechanism being tied to the library initialization does not happen pmon which would be missing some component without the cache files. (e.g: psu temp sensors not showing up)
Codepath used by pmon processes and platform initialization: https://github.com/aristanetworks/sonic/blob/05a02d74c91fb1b91fbb4602bf8c77067da3ecf2/arista/core/psu.py#L39
Codepath used by platform initialization: https://github.com/aristanetworks/sonic/blob/05a02d74c91fb1b91fbb4602bf8c77067da3ecf2/arista/core/psu.py#L154


The idea is to later have a process that monitors for fru insertion/removal to load/unload the appropriate drivers as well as refreshing the cache files (e.g not all power supply use the same kernel driver, but a system can support multiple variants).
The bit that is still under thinking is how to restart the appropriate pmon processes that have an outdated state.

lguohan: i suggest to use /var/run/platform_cache/arista to store you cache data, when you mount, just mount /var/run/platform_cache so that if other vendor has similiar ask later, we do not need to map again.

Staphylo: I can do that.
For now this was kept specific to the swi images with `_aboot_RUN_OPT` which means it was not polluting the ONIE image. I'll make the necessary changes in our driver code to use `/var/run/platform_cache/<vendor>` and mount `/var/run/platform_cache:/var/run/platform_cache` for all SONiC images.

lguohan (approved):

rules/docker-platform-monitor.mk
lguohan: approved for now. let you work on it later.

PR #5689, opened 2020-10-21
[hostcfgd] Enable/disable the container service only when the feature state was changed.

yozhao101:

**- Why I did it**
If we ran the CLI commands `sudo config feature autorestart snmp disabled/enabled` or `sudo config feature autorestart swss disabled/enabled`, then SNMP container will be stopped and started. This behavior was not expected since we updated the
`auto_restart` field not update `state` field in `FEATURE` table. The reason behind this issue is that either `state` field or `auto_restart` field was updated, the function `update_feature_state(...)` will be invoked which then starts snmp.timer service.
The snmp.timer service will first stop snmp.service and later start snmp.service. 

In order to solve this issue, the function `update_feature_state(...)` will be only invoked if `state` field in `FEATURE` table was
updated.

**- How I did it**
When the demon `hostcfgd` was activated, all the values of `state` field in `FEATURE` table of each container will be
cached. Each time the function `feature_state_handler(...)` is invoked, it will determine whether the `state` field of a
container was changed or not. If it was changed, function `update_feature_state(...)` will be invoked and the cached
value will also be updated. Otherwise, nothing will be done.

**- How to verify it**
We can run the CLI commands `sudo config feature autorestart snmp disabled/enabled` or `sudo config feature autorestart swss disabled/enabled` to check whether SNMP container is stopped and started. We also can run the CLI commands  `sudo config feature state snmp disabled/enabled` or `sudo config feature state swss disabled/enabled` to check whether the container is stopped and restarted.

- [ ] 201811
- [x ] 201911
- [ ] 202006

### DIFF
--- a/files/image_config/hostcfgd/hostcfgd
+++ b/files/image_config/hostcfgd/hostcfgd
@@ -316,7 +316,7 @@ class HostConfigDaemon:
                 syslog.syslog(syslog.LOG_WARNING, "Eanble state of feature '{}' is None".format(feature_name))
                 continue
 
-            # Store the initial value of 'state' field in 'FEATURE' table of each container
+            # Store the initial value of 'state' field in 'FEATURE' table of a specific container
             self.cached_feature_states[feature_name] = state
 
             self.update_feature_state(feature_name, state, feature_table)
@@ -365,10 +365,10 @@ class HostConfigDaemon:
         # 'update_feature_state(...)' will be invoked to enable/disable the corresponding container service.
         # At the same time, the cached value will be updated with current value of 'state' field. The reason
         # to cache the previous value of 'state' field and do the comparison at here is that if we issue the 
-        # 'config' command to update the value of 'AutoRestart' field in 'FEATURE' table, the function 
+        # 'config' command to update the value of 'auto_restart' field in 'FEATURE' table, the function 
         # 'update_feature_state(...)' will also be invoked and this behavior is not expected.
         if self.cached_feature_states[feature_name] != state:
-            self.cached_feature_states = state
+            self.cached_feature_states[feature_name] = state
             self.update_feature_state(feature_name, state, feature_table)
 
     def start(self):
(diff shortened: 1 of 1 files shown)


jleveque (changes requested):
The PR title is a bit lengthy yet does not accurately describe the change being made. It instead describes one side effect of the change. The title, which will subsequently become the commit message, should summarize the actual change being made. "[hostcfgd] Fix the SNMP container stopped and started if 'AutoRestart' state of snmp/swss was changed" does not summarize the change; SNMP/SwSS should not be mentioned in the title -- it can be mentioned in the description. The actual change here is to invoke systemd commands only if the feature state has changed from its previous state, and *not* if other data in the FEATURE table has changed. Please try to condense what I wrote above down into a succinct and brief title.

files/image_config/hostcfgd/hostcfgd
```diff
@@ -356,7 +361,15 @@ class HostConfigDaemon:
             syslog.syslog(syslog.LOG_WARNING, "Enable state of feature '{}' is None".format(feature_name))
             return
 
-        self.update_feature_state(feature_name, state, feature_table)
+        # Compare the current value of 'state' field with cached value, if they are not equal, the function
+        # 'update_feature_state(...)' will be invoked to enable/disable the corresponding container service.
+        # At the same time, the cached value will be updated with current value of 'state' field. The reason
+        # to cache the previous value of 'state' field and do the comparison at here is that if we issue the 
+        # 'config' command to update the value of 'auto_restart' field in 'FEATURE' table, the function 
+        # 'update_feature_state(...)' will also be invoked and this behavior is not expected.
```
jleveque, line 369: This comment is a bit overly-verbose. It could be simplified into a single sentence

jleveque (changes requested):

yxieca (dismissed):

files/image_config/hostcfgd/hostcfgd
yozhao101: Great suggestion! Will summarize with a single sentence.

lguohan (approved):

jleveque (approved):

abdosi (approved):

yxieca (approved):

lguohan:
retest vsimage please

PR #5691, opened 2020-10-21
Everflow and Dataacl should not get binded to Backplane port-channels.

abdosi:
Why I did:
Everflow and Dataacl should not get binded to Backplane port-channels.
However with updated port alias mapping it was getting binded.
Reason is because we were relying on port_alias_asic_map dictionary
but that dictionary can't be used as alias name format has changed.

How I verify:

After updating alias name in *.ini  uni-test were failing.
With updating logic in minigraph.py test are passing.
Also verified on Multi-asic platforms.


lguohan (approved):

abdosi:
@SuvarnaMeenakshi Can you please review this

SuvarnaMeenakshi (approved):

PR #5694, opened 2020-10-22
Add sonic-host-services and sonic-host-services-data packages

jleveque:
**- Why I did it**

Install all host services and their data files in package format rather than file-by-file

**- How I did it**

- Create sonic-host-services Python wheel package, currently including procdockerstatsd
  - Also add the framework for unit tests by adding one simple procdockerstatsd test case
- Create sonic-host-services-data Debian package which is responsible for installing the related systemd unit files to control the services in the Python wheel. This package will also be responsible for installing any Jinja2 templates and other data files needed by the host services.

**- How to verify it**

Build and install an image. Ensure the procdockerstatsd service is installed, enabled and running using `systemctl status procdockerstatsd.service`

- [ ] 201811
- [ ] 201911
- [ ] 202006


jleveque:
Retest vs please

lguohan (approved):

PR #5699, opened 2020-10-22
Add verification IPv4 address on LLDP conf Jinja2 Template

bratashX:
Signed-off-by: Petro Bratash <petrox.bratash@intel.com>

**- Why I did it**

LLDP conf Jinja2 Template does not verify IPv4 address and can use IPv6 version. This issue does not effect control LLDP daemon. Issue can be reproduced via `test_snmp_lldp` test. LLDP conf Jinja2 Template selects first item from the list of mgmt interfaces.

TESTBED_1 LLDP conf

```
# cat /etc/lldpd.conf 
configure ports eth0 lldp portidsubtype local eth0
configure system ip management pattern FC00:3::32
configure system hostname dut-1
```
TESTBED_2  LLDP conf

```
# cat /etc/lldpd.conf
configure ports eth0 lldp portidsubtype local eth0
configure system ip management pattern 10.22.24.61
configure system hostname dut-2
```
TESTBED_1  MGMT_INTERFACE

```
$ redis-cli -n 4 keys "*" | grep MGMT_INTERFACE
MGMT_INTERFACE|eth0|10.22.24.53/23
MGMT_INTERFACE|eth0|FC00:3::32/64
```
TESTBED_2  MGMT_INTERFACE

```
$ redis-cli -n 4 keys "*" | grep MGMT_INTERFACE
MGMT_INTERFACE|eth0|FC00:3::32/64
MGMT_INTERFACE|eth0|10.22.24.61/23

```

**- How I did it**

Skip IP address if it includes ":" characters

**- How to verify it**

1. Run:
`docker exec lldp cat /etc/lldpd.conf`
2. Check the type of  IP-address

- [ ] 201811
- [ ] 201911
- [ ] 202006


bratashX:
@akokhan, please review

liat-grozovik:
@shlomibitton can you please review as well?

akokhan (dismissed):

pavel-shirshov (changes requested):
Thank you for the contrinbution
Can you please
1. Describe more what is the problem is? The way to reproduce the issue and so on. Currently it is not clear from the PR description
2. Add tests for your change,

dockers/docker-lldp/lldpd.conf.j2
```diff
@@ -1,12 +1,19 @@
+{% set mgmt_if = {} %}
 {% if MGMT_INTERFACE %}
+{% for interface in MGMT_INTERFACE.keys()|list %}
+{% if not ':' in interface[1] %}
```
tahmed-dev, line 4: would the following work?
```
{% if interface[1] | pfx_filter | ipv4 %}
```

akokhan:
> Thank you for the contrinbution
> Can you please
> 
> 1. Describe more what is the problem is? The way to reproduce the issue and so on. Currently it is not clear from the PR description
> 2. Add tests for your change,

https://github.com/Azure/sonic-buildimage/issues/5603

lguohan:
can you add test in sonic-cfggen to validate the template results.

dockers/docker-lldp/lldpd.conf.j2
bratashX: Thanks, improved

shlomibitton:
LGTM

bratashX:
@lguohan  @pavel-shirshov
Currently test for testing **_lldp.conf_**  exists
https://github.com/Azure/sonic-buildimage/blob/ad2001f972e27034b876a79c6108f47420544bf5/src/sonic-config-engine/tests/test_j2files.py#L69
Test take last interface in **_ManagementIPInterfaces_**  from [t0-sample-graph.xml](https://github.com/Azure/sonic-buildimage/blob/ad2001f972e27034b876a79c6108f47420544bf5/src/sonic-config-engine/tests/t0-sample-graph.xml) file.
Previously test passed because the last interface was IPv4 https://github.com/Azure/sonic-buildimage/blob/ad2001f972e27034b876a79c6108f47420544bf5/src/sonic-config-engine/tests/t0-sample-graph.xml#L202

I think this commit will fix the test(change the order of the interfaces): https://github.com/Azure/sonic-buildimage/pull/5699/commits/ad2001f972e27034b876a79c6108f47420544bf5

pavel-shirshov:
Hi @bratashX

I'd suggest you revert the last commit. It breaks other tests. Usually we have ipv6 as second address on the interface in the minigraph.

How to run the tests:
1. cd sonic-buildimage/src/sonic-config-engine
2. pytest tests

qiluo-msft:
Retest mellanox please

qiluo-msft:
Retest vsimage please

dockers/docker-lldp/lldpd.conf.j2
```diff
@@ -1,12 +1,19 @@
+{% set mgmt_if = {} %}
 {% if MGMT_INTERFACE %}
+{% for interface in MGMT_INTERFACE.keys()|list %}
+{% if interface[1]|ipv4 %}
+{% if mgmt_if.update({'port_name' : interface[0]}) %} {% endif %}
+{% if mgmt_if.update({'ipv4' : interface[1].split('/')[0]}) %} {% endif %}
```
tahmed-dev, line 6: suggestion:

`interface[1].split('/')[0]`-->`interface[1]|pfx_filter`

dockers/docker-lldp/lldpd.conf.j2
```diff
@@ -1,12 +1,19 @@
+{% set mgmt_if = {} %}
 {% if MGMT_INTERFACE %}
+{% for interface in MGMT_INTERFACE.keys()|list %}
+{% if interface[1]|ipv4 %}
```
qiluo-msft, line 4: >ipv4 [](start = 19, length = 4)

Could you add a test case similar to [https://github.com/Azure/sonic-buildimage/blob/781188f54941a2eb9e4a23a96f05986ec51ff106/src/sonic-config-engine/tests/test_j2files.py#L69](https://github.com/Azure/sonic-buildimage/blob/781188f54941a2eb9e4a23a96f05986ec51ff106/src/sonic-config-engine/tests/test_j2files.py#L69)? So it will capture regression like this.

qiluo-msft (changes requested):
As comment

bratashX:
> 2\. pytest tests

@pavel-shirshov Previously I have never been run SONiC test. Could you please say what environment needs use before test running? Maybe is some docker for testing or some README file?

bratashX:
retest broadcom please

bratashX:
Retest vsimage please

bratashX:
retest broadcom please

lguohan:
retest vsimage please

bratashX:
@lguohan  Can I duplicate **_src/sonic-config-engine/tests/t0-sample-graph.xml_**  with  changes from these commit: https://github.com/Azure/sonic-buildimage/pull/5699/commits/ad2001f972e27034b876a79c6108f47420544bf5
(it fix **_test_lldp_** test and doesn't  affect other tests)?   Is it good solution?

dockers/docker-lldp/lldpd.conf.j2
bratashX: Updated, please review

dockers/docker-lldp/lldpd.conf.j2
bratashX: The test was added, please review

qiluo-msft: Thanks for adding test! I think another useful test case is 1 ipv4 + 1 ipv6 mgmt addresses, which are common to in normal use case.

---
In reply to: [515413789](https://github.com/Azure/sonic-buildimage/pull/5699#discussion_r515413789) [](ancestors = 515413789)

pavel-shirshov:
@bratashX 
Currently I don't have such documentation.
What I did to run the tests: manually installed all dependency which is required by using pip.
I'd suggest start with
```
pip install pytest
pip install pytest-runner
```
Also you need to install sonic-py-common and swsssdk for that
```
git clone https://github.com/Azure/sonic-buildimage.git
cd sonic-buildimage/src/sonic-py-common
sudo python setup.py install
git clone https://github.com/Azure/sonic-py-swsssdk.git
sido python setup.py install
```

src/sonic-config-engine/tests/test_j2files.py
```diff
@@ -70,12 +71,17 @@ def test_dhcp_relay(self):
...
 
-        # Test generation of lldpd.conf if management interface IPv4 exist
+        # Test generation of lldpd.conf if IPv4 and IPv6 management interfaces exist
         argument = '-m ' + self.t0_minigraph + ' -p ' + self.t0_port_config + ' -t ' + lldpd_conf_template + ' > ' + self.output_file
         self.run_script(argument)
         self.assertTrue(filecmp.cmp(os.path.join(self.test_dir, 'sample_output', utils.PYvX_DIR, 'lldp_conf', 'lldpd-ipv4-iface.conf'), self.output_file))
 
-        # Test generation of lldpd.conf if Management interface IPv4 doesn't exist (IPv6 only)
+        # Test generation of lldpd.conf if management interface IPv4 only exist
+        argument = '-m ' + self.t0_minigraph_ipv4_mgmt_iface + ' -p ' + self.t0_port_config + ' -t ' + lldpd_conf_template + ' > ' + self.output_file
```
lguohan, line 80: instead of using minigraph directly, i think you can use your own sample json file, it could make the test easier to read and improve.

you can use example here. construct your own json file and generate the template.

 def test_ndppd_conf(self):

lguohan:
> @lguohan Can I duplicate **_src/sonic-config-engine/tests/t0-sample-graph.xml_** with changes from these commit: [ad2001f](https://github.com/Azure/sonic-buildimage/commit/ad2001f972e27034b876a79c6108f47420544bf5)
> (it fix **_test_lldp_** test and doesn't affect other tests)? Is it good solution?

a good solution is to use test_ndppd_conf test as an example, you do not need to introduce a new minigraph.

dockers/docker-lldp/lldpd.conf.j2
bratashX: I think the test must verify  2 cases:

- IPv4 mgmt interface exists
- IPv4 mgmt interface doesn't  exist

But now I add 3 cases, when:
- IPv4 mgmt interface only exists
- IPv6 mgmt interface only exists
- IPv4 + IPv6 mgmt interface exists

Please review

src/sonic-config-engine/tests/test_j2files.py
bratashX: Thank you, fixed.

lguohan (approved):

lguohan:
retest mellanox please

pavel-shirshov (approved):

qiluo-msft (approved):

pavel-shirshov:
retest vsimage please

PR #5714, opened 2020-10-25
Export libyang API "lyd_check_mandatory_tree" for Management framework (CVL)

dutta-partha:
**- Why I did it**
Management framework (CVL) needs to call lyd_check_mandatory_tree() for validation and hence exported lyd_check_mandatory_tree() as an API.

**- How I did it**
Added "API" keyword  before lyd_check_mandatory_tree() definition.

**- How to verify it**
There is no functionality code change here and no specific steps to verify it. Management framework (CVL) should be able to call this function and no patching and compilation error should be seen.

- [ ] 201811
- [ ] 201911
- [ ] 202006

Added "API" keyword before export lyd_check_mandatory_tree() function definition. 


renukamanavalan:
Can you please help review? 
FYI: This PR has  been pending for a while.

@zhenggen-xu , @li-pingmao , @vasant17 , @samaity

Thanks a lot.

praveen-li:
Sure, I will have a look today.


Thanks,

Regards
Praveen

src/libyang/patch/libyang_mgmt_framework.patch
```diff
@@ -0,0 +1,13 @@
...
+index 04653a46..65dca211 100644
+--- a/src/tree_data.c
++++ b/src/tree_data.c
+@@ -842,7 +842,7 @@ error:
+     return ret;
+ }
+ 
+-int
++API int
+ lyd_check_mandatory_tree(struct lyd_node *root, struct ly_ctx *ctx, const struct lys_module **modules, int mod_count,
```
lguohan, line 11: can this go to upstream?

dutta-partha: Yes, this patch is just to export lyd_check_mandatory_tree() as API so that CVL can validate mandatory YANG nodes.

sachinholla:
retest vs vsimage please

sachinholla:
retest vs please

sachinholla:
retest vsimage please

src/libyang/patch/libyang_mgmt_framework.patch
praveen-li: Yeah good to upstream with reason of use. Though no harm in making it an API in our code till then.

src/libyang/patch/series
```diff
@@ -1,2 +1,3 @@
 libyang.patch
+libyang_mgmt_framework.patch
```
praveen-li, line 2: In the Future, are we planning to put all patches related to a framework in this single file?,
 If not, I feel it is better to have a file name based on the patch, for example, "API_lyd_check_mandatory_tree.patch"

praveen-li (commented):
Overall LGTM.

src/libyang/patch/series
dutta-partha: Since we are not expecting many changes for libyang library, we have kept a single patch file for libyang changes related to management framework. Please suggest if current filename should be changed to any other generic name e.g. libyang_extension.patch or libyang_ext.patch.

praveen-li: ext is not good,  libyang_mgmt_framework.patch is fine, but if we had to add in this file in the future then we need to break it down into multiple patches with description. 
That will be inline with other sonic repos, how we maintain patches. That will also help if some patches are accepted by upstream and we have to remove them in the future.

praveen-li (approved):
I think we can go ahead with this patch.

renukamanavalan (approved):

PR #5726, opened 2020-10-27
[minigraph.py]: Parse VLAN MAC address from minigraph when present

theasianpianist:

**- Why I did it**
Allow setting VLAN MAC address from minigraph information
**- How I did it**
While parsing DpgDec, search for `<MacAddress>` tags in VLAN interfaces.
**- How to verify it**
Run the update sonic-config-engine tests.

- [ ] 201811
- [ ] 201911
- [ ] 202006

### DIFF
--- a/src/sonic-config-engine/minigraph.py
+++ b/src/sonic-config-engine/minigraph.py
@@ -362,7 +362,6 @@ def parse_dpg(dpg, hname):
             vlanid = vintf.find(str(QName(ns, "VlanID"))).text
             vintfmbr = vintf.find(str(QName(ns, "AttachTo"))).text
             vlantype = vintf.find(str(QName(ns, "Type")))
-            vlanmac = vintf.find(str(QName(ns, "MacAddress")))
             if vlantype != None:
                 vlantype_name = vintf.find(str(QName(ns, "Type"))).text
             vmbr_list = vintfmbr.split(';')
@@ -384,6 +383,7 @@ def parse_dpg(dpg, hname):
                 vdhcpserver_list = vintfdhcpservers.split(';')
                 vlan_attributes['dhcp_servers'] = vdhcpserver_list
 
+            vlanmac = vintf.find(str(QName(ns, "MacAddress")))
             if vlanmac != None:
                 vlan_attributes['mac'] = vlanmac.text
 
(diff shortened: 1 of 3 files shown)


src/sonic-config-engine/minigraph.py
```diff
@@ -362,6 +362,7 @@ def parse_dpg(dpg, hname):
             vlanid = vintf.find(str(QName(ns, "VlanID"))).text
             vintfmbr = vintf.find(str(QName(ns, "AttachTo"))).text
             vlantype = vintf.find(str(QName(ns, "Type")))
+            vlanmac = vintf.find(str(QName(ns, "MacAddress")))
```
tahmed-dev, line 365: nit: I'd move line 365 down right before line 387 for consistency with current retrieval and if checks.

src/sonic-config-engine/tests/test_minigraph_case.py
```diff
@@ -79,11 +79,10 @@ def test_minigraph_interfaces(self):
         self.assertEqual(output.strip(), "[('Ethernet0', '10.0.0.58/31'), 'Ethernet0', ('Ethernet0', 'FC00::75/126')]")
 
     def test_minigraph_vlans(self):
-        argument = '-m "' + self.sample_graph + '" -p "' + self.port_config + '" -v VLAN'
-        output = self.run_script(argument)
+        vlan_table = minigraph.parse_xml(self.sample_graph, port_config_file=self.port_config)["VLAN"]
```
tahmed-dev, line 82: Hmm, with sonic-cfggen, we are testing more layers. I think the name might be confusing as one might think this is a unit test for minigraph module. I'd prefer to use the sonic-cfggen command for the reason mentioned before and maybe we can unit test for specific modules.

src/sonic-config-engine/minigraph.py
theasianpianist: Not sure I understand, it seems like almost all the top-level `.find()` calls (for direct children of the `VlanInterface` tag) are made at the start of the loop.

src/sonic-config-engine/tests/test_minigraph_case.py
theasianpianist: Good point, I'll put the `sonic-cfggen` call back. Thanks!

src/sonic-config-engine/minigraph.py
tahmed-dev: I view it differently as `vlantype` was followed by `if vlantype != None:`, `vintf_node` is followed by `if vintf_node` and etc.

theasianpianist: Oh I see. Changed!

prsunny (approved):

PR #5731, opened 2020-10-28
Added new method get_back_end_interface_set() to speed up back-end in…

gechiang:
…terface check by cache the back-end intf set instead of each time need to read from redis DB

**- Why I did it**
For Multi-Asic if the caller needs to determine whether  an interface (port or LAG) involved is one of the back-end ASIC interface, it calls "skip_display()" which access the config DB in order to determine the interface's role.  This can not scale if the caller is repeatedly doing this in a scale environment such as handling "show ip route" that needs to filter out all back-end interfaces for each name spaces that contains many thousands of routes.  This new method was implemented to cache the output of interface's role into a set so that the user can refer to the set to do its filtering operation thus able to handle scale cases efficiently.

**- How I did it**
The new method read out from configDB based on the namespace given (default is all name spaces) and cache all backe-end interfaces into a set.  It also walk through all the LAGs of the given namespace and add to the same set if any of its member port is a back-end interface.

**- How to verify it**
Manual testing was done to dumpout the set content to validate that all members of the set are back-end interfaces and back-end LAGs.

- [ ] 201811
- [x] 201911
- [ ] 202006

Added new method to get the back-end interfaces to support multi-asic functionalities. This method will be used by "show ip(ipv6) route" multi-asic support

### DIFF
--- a/src/sonic-py-common/sonic_py_common/multi_asic.py
+++ b/src/sonic-py-common/sonic_py_common/multi_asic.py
@@ -333,16 +333,17 @@ def get_back_end_interface_set(namespace=None):
         if PORT_ROLE in info and info[PORT_ROLE] == INTERNAL_PORT:
             bk_end_intf_list.append(port)
 
-    ns_list = get_namespace_list(namespace)
-    for ns in ns_list:
-        config_db = connect_config_db_for_ns(ns)
-        port_channels = config_db.get_table(PORT_CHANNEL_CFG_DB_TABLE)
-        # a back-end LAG must be configured with all of its member from back-end interfaces.
-        # mixing back-end and front-end interfaces is miss configuration and not allowed.
-        # To determine if a LAG is back-end LAG, just need to check its first member is back-end or not
-        # is sufficient. Note that a user defined LAG may have empty members so the list expansion logic
-        # need to ensure there are members before inspecting member[0].
-        bk_end_intf_list.extend([port_channel for port_channel, lag_info in port_channels.items()\
+    if len(bk_end_intf_list):
+        ns_list = get_namespace_list(namespace)
+        for ns in ns_list:
+            config_db = connect_config_db_for_ns(ns)
+            port_channels = config_db.get_table(PORT_CHANNEL_CFG_DB_TABLE)
+            # a back-end LAG must be configured with all of its member from back-end interfaces.
+            # mixing back-end and front-end interfaces is miss configuration and not allowed.
+            # To determine if a LAG is back-end LAG, just need to check its first member is back-end or not
+            # is sufficient. Note that a user defined LAG may have empty members so the list expansion logic
+            # need to ensure there are members before inspecting member[0].
+            bk_end_intf_list.extend([port_channel for port_channel, lag_info in port_channels.items()\
                                 if 'members' in lag_info and lag_info['members'][0] in bk_end_intf_list])
     a = set()
     a.update(bk_end_intf_list)
--- a/src/sonic-py-common/sonic_py_common/multi_asic.py
+++ b/src/sonic-py-common/sonic_py_common/multi_asic.py
@@ -333,15 +333,16 @@ def get_back_end_interface_set(namespace=None):
         if PORT_ROLE in info and info[PORT_ROLE] == INTERNAL_PORT:
             bk_end_intf_list.append(port)
 
-        ns_list = get_namespace_list(namespace)
-        for ns in ns_list:
-            config_db = connect_config_db_for_ns(ns)
-            port_channels = config_db.get_table(PORT_CHANNEL_CFG_DB_TABLE)
-            for port_channel, lag_info in port_channels.items():
-                if 'members' in lag_info:
-                    members = lag_info['members']
-                    if members[0] in bk_end_intf_list:
-                        bk_end_intf_list.append(port_channel)
+    ns_list = get_namespace_list(namespace)
+    for ns in ns_list:
+        config_db = connect_config_db_for_ns(ns)
+        port_channels = config_db.get_table(PORT_CHANNEL_CFG_DB_TABLE)
+        for port_channel, lag_info in port_channels.items():
+            if 'members' in lag_info:
+                members = lag_info['members']
+                if members[0] in bk_end_intf_list:
+                    bk_end_intf_list.append(port_channel)
+
     a = set()
     a.update(bk_end_intf_list)
     return a
--- a/src/sonic-py-common/sonic_py_common/multi_asic.py
+++ b/src/sonic-py-common/sonic_py_common/multi_asic.py
@@ -321,6 +321,30 @@ def is_port_channel_internal(port_channel, namespace=None):
 
     return False
 
+# Allow user to get a set() of back-end interface and back-end LAG per namespace
+# default is getting it for all name spaces if no namespace is specified
+def get_back_end_interface_set(namespace=None):
+    bk_end_intf_list =[]
+    if not is_multi_asic():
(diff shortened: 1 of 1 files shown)


src/sonic-py-common/sonic_py_common/multi_asic.py
```diff
@@ -321,6 +321,30 @@ def is_port_channel_internal(port_channel, namespace=None):
...
+
+        ns_list = get_namespace_list(namespace)
+        for ns in ns_list:
+            config_db = connect_config_db_for_ns(ns)
+            port_channels = config_db.get_table(PORT_CHANNEL_CFG_DB_TABLE)
+            for port_channel, lag_info in port_channels.items():
+                if 'members' in lag_info:
+                    members = lag_info['members']
+                    if members[0] in bk_end_intf_list:
+                        bk_end_intf_list.append(port_channel)
```
smaheshm, line 344: is this intended to be inside for loop. This will get called for every port in the port_table.

src/sonic-py-common/sonic_py_common/multi_asic.py
```diff
@@ -321,6 +321,30 @@ def is_port_channel_internal(port_channel, namespace=None):
...
+    a = set()
+    a.update(bk_end_intf_list)
```
smaheshm, line 346: a = set(bk_end_intf_list)

src/sonic-py-common/sonic_py_common/multi_asic.py
gechiang: @smaheshm Yes this is intended. for each namespace this needs to be performed if user use default option which is for all name spaces.  each namespace has its own LAG DB.

src/sonic-py-common/sonic_py_common/multi_asic.py
gechiang: @smaheshm according to some research using .update(xx_list) has better performance when compared to set(xx_list).

src/sonic-py-common/sonic_py_common/multi_asic.py
smaheshm: I meant this for loop is inside the port for loop.

gechiang: @smaheshm Thanks for spotting it.  When I copied over it added extra spacing.  Let me fix that.

arlakshm (dismissed):
lgtm

src/sonic-py-common/sonic_py_common/multi_asic.py
smaheshm: samaddik@samaddik-vm-01:/data/sonic$ python -m timeit 'x = [x for x in range(0, 1000000)]; y = set(); y.update(x)'
10 loops, best of 3: 104 msec per loop
samaddik@samaddik-vm-01:/data/sonic$ python -m timeit 'x = [x for x in range(0, 1000000)]; y = set(); y.update(x)'
10 loops, best of 3: 104 msec per loop
samaddik@samaddik-vm-01:/data/sonic$ python -m timeit 'x = [x for x in range(0, 1000000)]; y = set(); y.update(x)'
10 loops, best of 3: 105 msec per loop
samaddik@samaddik-vm-01:/data/sonic$ python -m timeit 'x = [x for x in range(0, 1000000)]; set(x)'
10 loops, best of 3: 104 msec per loop
samaddik@samaddik-vm-01:/data/sonic$ python -m timeit 'x = [x for x in range(0, 1000000)]; set(x)'
10 loops, best of 3: 105 msec per loop

No noticeable diff for million elements. Either is fine.

src/sonic-py-common/sonic_py_common/multi_asic.py
abdosi: @gechiang we can use any function. I got that as one of review comment .
Refer: https://github.com/Azure/sonic-buildimage/blob/master/src/sonic-config-engine/minigraph.py#L756

Also we can consolidate both going forward however not in this PR.

gechiang: @abdosi I took a look at the link you provided.  I see the logic is to check if there are any LAG members being a back-end interface then the LAG is a back-end port channel.  I have a question.  Can we have a case where the port channel is formed with front-end ports and back-end ports mixed as LAG member?  I do not think this can ever be a valid usage of such port channel right?  If that is the case, then my change which is also same as what is already done in the method " is_port_channel_internal()" should be faster right?  It just have to check the very first member and result can already be determined. Think about the case where it is a true front-end port channel with many lag members.  The method you pointed out will require compare ALL of the lag members in order to determine that it is NOT a back-end port channel. performance wise it is not as efficient as what my check can achieve...  Also, that method assumed there are alwasy some members for a port channel.  that may not be always true as some one may have configured an empty port channel which in that case that code may crash as it assumed "member" is present...

gechiang:
retest vsimage please

gechiang:
retest broadcom please

src/sonic-py-common/sonic_py_common/multi_asic.py
abdosi: @gechiang any function should return immediately like first member check. Basically it is not from optimized perspective but better python readability.

gechiang: @abdosi I can replace that block of code: 

            for port_channel, lag_info in port_channels.items():
                if 'members' in lag_info:
                    members = lag_info['members']
                    if members[0] in bk_end_intf_list:
                        bk_end_intf_list.append(port_channel)


with the following one line code:
```
bk_end_intf_list.extend([port_channel for port_channel, lag_info in port_channels.items()\
                                if 'members' in lag_info and lag_info['members'][0] in bk_end_intf_list])
```
If you feel this is more readable...  I actually prefer the original "c" style for clarity although few more lines that does the same thing...  

As for the any() function it is not efficient for the case when LAG contains all front-end ports where it will have to evaluate all the lag members in order to determine that LAG should not be included so I will not use that.

smaheshm: Code should be idiomatic. And in general readability is preferred over efficiency especially when the efficiency doesn't make a big difference. Easier to maintain code.

src/sonic-py-common/sonic_py_common/multi_asic.py
gechiang: I have updated the code to use this one line expansion method.

src/sonic-py-common/sonic_py_common/multi_asic.py
gechiang: I have updated the code to use this one line list expansion and added comments so it should allow reader to understand what are the criteria that determine a LAG is a back-end LAG...

gechiang:
retest broadcom please

gechiang:
retest mellanox please

gechiang:
retest vsimage please

gechiang:
retest mellanox please

gechiang:
retest vsimage please

gechiang:
@arlakshm Thanks for the offline comment in regard to the case when the bk-end-interface after from port table is empty. Added the check to skip the LAG checking so it does not unnecessary waste time there.

arlakshm (approved):

smaheshm:
:shipit: Just wait for the tests to finish

smaheshm (approved):

gechiang:
retest broadcom please

gechiang:
retest baseimage please

gechiang:
retest mellanox please

PR #5740, opened 2020-10-29
Load config after subscribe

renukamanavalan:
**- Why I did it**
The update_all_feature_states can run in the range of 20+ seconds to one minute. With load of AAA & Tacacs preceding it, any DB updates in AAA/TACACS during the long running feature updates would get missed. To avoid, switch the order.

**- How I did it**
Do a load after after updating all feature states. 

**- How to verify it**
Not a easy one
Have a script that 
   restart hostcfgd
   sleep 2s
   run redis-cli/config command to update AAA/TACACS table

Run the script above and watch the file /etc/pam.d/common-auth-sonic for a minute.

***_When it repro:
The updates will not reflect in /etc/pam.d/common-auth-sonic 

- [ ] 201811
- [x] 201911
- [x] 202006

### DIFF
--- a/files/image_config/hostcfgd/hostcfgd
+++ b/files/image_config/hostcfgd/hostcfgd
@@ -239,6 +239,8 @@ class HostConfigDaemon:
         # Cache the values of 'state' field in 'FEATURE' table of each container
         self.cached_feature_states = {}
 
+        self.is_multi_npu = device_info.is_multi_npu()
+
 
     def load(self):
         aaa = self.config_db.get_table('AAA')
@@ -249,7 +251,6 @@ class HostConfigDaemon:
         lpbk_table = self.config_db.get_table('LOOPBACK_INTERFACE')
         self.iptables.load(lpbk_table)
 
-        self.is_multi_npu = device_info.is_multi_npu()
 
     def update_feature_state(self, feature_name, state, feature_table):
         has_timer = ast.literal_eval(feature_table[feature_name].get('has_timer', 'False'))
--- a/files/image_config/hostcfgd/hostcfgd
+++ b/files/image_config/hostcfgd/hostcfgd
@@ -233,17 +233,23 @@ class HostConfigDaemon:
         self.config_db = ConfigDBConnector()
         self.config_db.connect(wait_for_init=True, retry_on=True)
         syslog.syslog(syslog.LOG_INFO, 'ConfigDB connect success')
+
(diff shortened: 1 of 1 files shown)


prsunny (approved):

files/image_config/hostcfgd/hostcfgd
```diff
@@ -367,14 +374,19 @@ class HostConfigDaemon:
...
         self.config_db.subscribe('LOOPBACK_INTERFACE', lambda table, key, data: self.lpbk_handler(key, data))
         self.config_db.subscribe('FEATURE', lambda table, key, data: self.feature_state_handler(key, data))
+
+        # Update all feature states once upon starting
+        self.update_all_feature_states()
+
+        # Defer load until subscribe
+        self.load()
+
         self.config_db.listen()
```
judyjoseph, line 390: I see the code, find that self.config_db.subscribe() is a misnomer .. it doesn't actually do a pubsub.psubscribe(). It just assigns a handler and comes out. It is this API self.config_db.listen() which actually does pubsub.psubscribe()

So we could still miss the events since self.update_all_feature_states() is before listen call ?  
Just wondering if this fix will take care of all cases still ?

abdosi: @judyjoseph   self.load() should take care  care of that at least for TACACS/AAA
I think one window is config-db write happening in between self.load() and self.config_db.listen(). Might be very corner case.

renukamanavalan: Still this 23 seconds long call -- taken to start all services, even when all are running, is way too expensive. We would need to take a second look at it. Better would be to run it on a different thread using timer.

renukamanavalan:
retest baseimage please

renukamanavalan:
retest baseimage please.

renukamanavalan:
retest baseimage please

files/image_config/hostcfgd/hostcfgd
judyjoseph: yes !! It may be ok for the user to wait for 23 secs to get login access in case of single ASIC. On multi-asic wait will be longer, as we have more  processes per ASIC and the total wait could be more than 1min.

renukamanavalan:
retest baseimage please

renukamanavalan:
retest mellanox please

renukamanavalan:
retest baseimage please

PR #5713, opened 2020-10-24
[hostcfgd] Convert to Python 3; Add to sonic-host-services package

jleveque:
**- Why I did it**

To consolidate host services and install via packages instead of file-by-file, also as part of migrating all of SONiC to Python 3, as Python 2 is no longer supported.

**- How to verify it**

Ensure hostcfgd is installed and runs

**- Which release branch to backport (provide reason below if selected)**

- [ ] 201811
- [ ] 201911
- [ ] 202006

PR #5744, opened 2020-10-29
[hostcfgd] Configure service auto-restart in hostcfgd.

stepanblyschak:
Before this change, a process running inside every SONiC container dealt with FEATURE table 'auto_restart' field and depending on the value decided whether a container has to be killed or not. 
If killed service auto restart mechanism restarts the container. 
This change moves the logic from container to the host daemon - hostcfgd. 
The 'auto_restart' handling is kept in supervisor-proc-exit-listener but now it is not required for container that wants to support auto restart feature.

* hostcfgd refactoring - move feature handling in another class.
* override systemd service Restart= setting from hostcfgd.
* remove default systemd Restart=always.

**- Why I did it**

Remove the need to deal with container orchestration logic from the container itself. Leave this logic to the orchestrator - host OS.

**- How I did it**

hostcfgd configures 'Restart=' value for systemd service.

**- How to verify it**

```
root@r-tigon-11:/home/admin# sudo config feature autorestart lldp enabled
root@r-tigon-11:/home/admin# show feature status | grep lldp
lldp            enabled   enabled
root@r-tigon-11:/home/admin# docker exec -it lldp pkill -9 lldpd
root@r-tigon-11:/home/admin# docker ps -a | grep lldp
65058396277c        docker-lldp:latest                   "/usr/bin/docker-lld…"   2 days ago          Exited (0) 20 seconds ago                       lldp
root@r-tigon-11:/home/admin# docker ps -a | grep lldp
65058396277c        docker-lldp:latest                   "/usr/bin/docker-lld…"   2 days ago          Up 5 seconds                            lldp
root@r-tigon-11:/home/admin# sudo config feature autorestart lldp disabled
root@r-tigon-11:/home/admin# docker exec -it lldp pkill -9 lldpd
root@r-tigon-11:/home/admin# docker ps -a | grep lldp
65058396277c        docker-lldp:latest                   "/usr/bin/docker-lld…"   2 days ago          Up 35 seconds                           lldp
root@r-tigon-11:/home/admin# docker ps -a | grep lldp
65058396277c        docker-lldp:latest                   "/usr/bin/docker-lld…"   2 days ago          Exited (0) 3 seconds ago                       lldp
root@r-tigon-11:/home/admin# docker ps -a | grep lldp
65058396277c        docker-lldp:latest                   "/usr/bin/docker-lld…"   2 days ago          Exited (0) 39 seconds ago                       lldp
root@r-tigon-11:/home/admin#

```

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012


stepanblyschak:
retest mellanox please

stepanblyschak:
retest vsimage please

files/image_config/hostcfgd/hostcfgd
```diff
@@ -43,6 +42,139 @@ def obfuscate(data):
         return data
 
 
+class Feature(object):
```
lguohan, line 45: can you add unit test for this?

stepanblyschak: I think it can be added after https://github.com/Azure/sonic-buildimage/pull/5713 which moves hostcfgd to a python package with tests and I can write test in https://github.com/Azure/sonic-buildimage/tree/f4083627f163863f1c2b9e4038c382e3941db218/src/sonic-host-services/tests

stepanblyschak: Done

src/sonic-host-services/scripts/hostcfgd
```diff
@@ -57,6 +57,157 @@ def run_cmd(cmd, log_err = True):
...
+            for feature_name_suffix in feature_name_suffix_list:
+                for suffix in feature_suffixes:
+                    start_cmds.append("sudo systemctl unmask {}.{}".format(feature_name_suffix, suffix))
+                # If feature has timer associated with it, start/enable corresponding systemd .timer unit
+                # otherwise, start/enable corresponding systemd .service unit
+                start_cmds.append("sudo systemctl enable {}.{}".format(feature_name_suffix, feature_suffixes[-1]))
+                start_cmds.append("sudo systemctl start {}.{}".format(feature_name_suffix, feature_suffixes[-1]))
+                for cmd in start_cmds:
+                    syslog.syslog(syslog.LOG_INFO, "Running cmd: '{}'".format(cmd))
+                    try:
```
rajendra-dendukuri, line 143: Can we enhance run_cmd() use it to return error code as well as log the error read from the exception

src/sonic-host-services/scripts/hostcfgd
```diff
@@ -57,6 +57,157 @@ def run_cmd(cmd, log_err = True):
...
+        elif feature.state == "disabled":
+            stop_cmds = []
+            for feature_name_suffix in feature_name_suffix_list:
+                for suffix in reversed(feature_suffixes):
+                    stop_cmds.append("sudo systemctl stop {}.{}".format(feature_name_suffix, suffix))
+                    stop_cmds.append("sudo systemctl disable {}.{}".format(feature_name_suffix, suffix))
+                    stop_cmds.append("sudo systemctl mask {}.{}".format(feature_name_suffix, suffix))
+                for cmd in stop_cmds:
+                    syslog.syslog(syslog.LOG_INFO, "Running cmd: '{}'".format(cmd))
+                    try:
```
rajendra-dendukuri, line 162: Same comment as before to use run_cmd()

rajendra-dendukuri (changes requested):

src/sonic-host-services/scripts/hostcfgd
stepanblyschak: Thanks, yes, please check the enhanced run_cmd()

stepanblyschak:
@lguohan @jleveque

jleveque:
@stepanblyschak: This PR appears to change the container behavior upon critical process exit. Currently, `supervisor-proc-exit-listener` will only stop the container if auto-restart is enabled. With this PR, it appears the container will always stop, but it will only be restarted if auto-restart is enabled. Can you explain why you believe this behavior is preferred?

stepanblyschak:
@jleveque 
This has been discussed before over emails and with App.Ext sub-group community.
Considering the possibility to run 3rd party dockerized applications on SONiC the auto-restart functionality is highly benefitial, although the current approach limits the use of this feature only to SONiC containers which developers explicitely coded support for it. Thus, the choice was to make auto-restart functionality support for free for any container that is managed by systemd. On the other hand there is a change in container lifetime when critical process dies. Consider running an arbitrary container the user or container orchestrator has no control over its entrypoint process exits unexpectedly. Usually in microservices there is only one process running in container and if this process dies so does the container. Given that, I do not see a benefit from having the container running while one or more critical daemons have exited other than debug purpose, but for that a simple tweak in critical_processes file is enough to make the container still running. If you still have a requirement to control this behaviour from CONFIG DB I suggest that we have a CONTAINER_FEATURE table to control this behaviour. We can declare that to support this functionality 3rd party dockers have to explicitly add support for it, but not for auto-restart.

src/sonic-host-services/scripts/hostcfgd
```diff
@@ -537,23 +589,8 @@ class HostConfigDaemon:
...
-            syslog.syslog(syslog.LOG_WARNING, "Enable state of feature '{}' is None".format(feature_name))
-            return
-
-        self.cached_feature_states.setdefault(feature_name, 'disabled')
-
-        # Enable/disable the container service if the feature state was changed from its previous state.
-        if self.cached_feature_states[feature_name] != state:
-            self.update_feature_state(feature_name, state, feature_table)
+    def feature_handler(self, key, data):
+        self.feature.handle(key, data)
```
jleveque, line 593: Is this function necessary? Is it possible to rename the instantiated object `feature_hander` and pass `self.feature_handler.handle(key, data)` as the callback on line 621? Can you see if that works?

jleveque (changes requested):

src/sonic-host-services/scripts/hostcfgd
stepanblyschak: Not necessary, thanks for suggestion.

jleveque:
@lguohan: A few Azure Pipelines check builds are stuck in the "Expected — Waiting for status to be reported" state, and I cannot re-trigger these tests. There are a few PRs like this and I've even tried closing/reopening the PRs to no avail. Can you please help here?

liat-grozovik:
@lguohan tests are stuck for 2weeks. can you reset it? AzurepPipelines run is not working.
@jleveque any further comments or this can be approved?

stepanblyschak:
@rajendra-dendukuri, @jleveque Could you please rereview after conflict resolving?

src/sonic-host-services/scripts/hostcfgd
```diff
@@ -49,13 +53,188 @@ def obfuscate(data):
...
+        else:
+            syslog.syslog(syslog.LOG_ERR, "Unexpected state value '{}' for feature {}"
+                          .format(feature.state, feature.name))
+
+    def update_feature_auto_restart(self, feature):
+        restart_config = "always" if feature.auto_restart == "enabled" else "no"
+        service_conf = "[Service]\nRestart={}\n".format(restart_config)
+        feature_names, feature_suffixes = self.get_feature_attribute(feature)
+
+        for service_name in feature_names:
```
yozhao101, line 167: Do we need use the `feature_name` instead of `service_name` at here?

src/sonic-host-services/scripts/hostcfgd
```diff
@@ -49,13 +53,188 @@ def obfuscate(data):
...
+
+        if not feature_names:
+            syslog.syslog(syslog.LOG_ERR, "Feature '{}' service not available"
+                          .format(feature.name))
+
+        feature_suffixes = ["service"] + (["timer"] if feature.has_timer else [])
+
+        return feature_names, feature_suffixes
+
+    def is_invariant_feature_state(self, state):
```
yozhao101, line 196: Is it necessary to create a separate function which just has one line code?

src/sonic-host-services/scripts/hostcfgd
```diff
@@ -49,13 +53,188 @@ def obfuscate(data):
...
+
+            self.update_feature_state(feature)
+            self.resync_feature_state(feature)
+
+    def update_feature_state(self, feature):
+        cached_feature = self._cached_config[feature.name]
+
+        if self.is_invariant_feature_state(cached_feature.state):
+            if cached_feature.state != feature.state:
+                syslog.syslog(syslog.LOG_INFO, "Feature {} service is {}".format(feature.name, feature.state))
```
yozhao101, line 145: Since the `state` of service/feature is changed, do we need reorganize this message like:
`syslog.syslog(syslog.LOG_INFO, "The state of '{}' is changed from '{}' to '{}'".format(feature.name, cached_feature.state, feature.state))`?

src/sonic-host-services/scripts/hostcfgd
```diff
@@ -49,13 +53,188 @@ def obfuscate(data):
...
+                self.resync_feature_state(cached_feature)
```
yozhao101, line 146: Since the state of service/feature is updated to the new state `feature`, do we need pass `feature` to the function `resync_feature_state(feature)`? Can you check whether my understanding is correct please?

yozhao101: Since the function `resync_feature_state(...)` is used to update the `state` of service/feature and has one line code, is it necessary to create a separate function to do this?

src/sonic-host-services/scripts/hostcfgd
liat-grozovik: @stepanblyschak could you please refer to this comment?

src/sonic-host-services/scripts/hostcfgd
liat-grozovik: is this really needed? what is the ROI of that?

src/sonic-host-services/scripts/hostcfgd
stepanblyschak: > Since the function `resync_feature_state(...)` is used to update the `state` of service/feature and has one line code, is it necessary to create a separate function to do this?

For readability purpose

src/sonic-host-services/scripts/hostcfgd
stepanblyschak: I had no intent to change the logging. This log message migrated from old code - https://github.com/Azure/sonic-buildimage/pull/5744/files#diff-27d10088e7a11c3ae41f16e5272d96926179ae179c61d6fc32a25d0cfa40907eL500.

src/sonic-host-services/scripts/hostcfgd
stepanblyschak: For readability purposes. Comparing:

```python
if cached_feature.state in ('always_enabled', 'always_disabled'):
```

vs

```python
if self.is_invariant_feature_state(cached_feature.state):
```

Reflects the purpose of the condition.

stepanblyschak:
@yozhao101 could you please check if I addressed your comments?

liat-grozovik:
@rajendra-dendukuri and @yozhao101 this PR is ready for approval in the last 2 weeks. kindly reminder to review the fixes following your comments.

src/sonic-host-services/scripts/hostcfgd
renukamanavalan: @yozhao101, Please resolve or add your comments.
We need to get this PR in.

renukamanavalan:
@rajendra-dendukuri,  @yozhao101 

Please resolve or add your comments.
We need to get this PR in.

src/sonic-host-services/scripts/hostcfgd
```diff
@@ -57,15 +71,188 @@ def obfuscate(data):
...
+        if target_state not in ('enabled', 'disabled', 'always_enabled', 'always_disabled'):
+            raise ValueError('Invalid state rendered for feature {}: {}'.format(self.name, target_state))
+        return target_state
+
+
+class FeatureHandler(object):
+    """ Handles FEATURE table updates. """
+
+    SYSTEMD_SYSTEM_DIR = '/etc/systemd/system/'
+    SYSTEMD_SERVICE_CONF_D = os.path.join(SYSTEMD_SYSTEM_DIR, '{}.service.d/')
```
yozhao101, line 117: Do we need rename the variable `SYSTEMD_SERVICE_CONF_D` to `SYSTEMD_SERVICE_CONF_DIR`?

src/sonic-host-services/scripts/hostcfgd
```diff
@@ -57,15 +71,188 @@ def obfuscate(data):
...
-    try:
-        subprocess.check_call(cmd, shell = True)
-    except Exception as err:
-        if log_err:
-            syslog.syslog(syslog.LOG_ERR, "{} - failed: return code - {}, output:\n{}"
-                  .format(err.cmd, err.returncode, err.output))
-        return err.returncode
-    return 0
+class Feature(object):
+    """ Represents a feature configuration from CONFIG DB data. """
```
yozhao101, line 75: Do we need use `CONFIG_DB` to replace `CONFIG DB`?

src/sonic-host-services/scripts/hostcfgd
yozhao101: > > Since the state of service/feature is updated to the new state `feature`, do we need pass `feature` to the function `resync_feature_state(feature)`? Can you check whether my understanding is correct or not please?
> 
> if current state is an invariant state and the new one is different then current one we resync with the current one. Please refer to original code: https://github.com/Azure/sonic-buildimage/pull/5744/files#diff-27d10088e7a11c3ae41f16e5272d96926179ae179c61d6fc32a25d0cfa40907eL497-L504

Please review the comment on the original code.

src/sonic-host-services/scripts/hostcfgd
```diff
@@ -680,125 +866,6 @@ class HostConfigDaemon:
...
-    def is_invariant_feature(self, feature_name, state, feature_table):
-        invariant_feature = self.cached_feature_states[feature_name] == "always_enabled" or \
-                            self.cached_feature_states[feature_name] == "always_disabled"
-        if invariant_feature:
-            invariant_state = self.cached_feature_states[feature_name]
-            if state != invariant_state:
-                syslog.syslog(syslog.LOG_INFO, "Feature '{}' service is '{}'"
-                              .format(feature_name, invariant_state))
-                entry = self.config_db.get_entry('FEATURE', feature_name)
-                entry['state'] = invariant_state
```
yozhao101, line 758: @stepanblyschak Following my previous comment, if the `state` at here is `always_disabled` and `invariant_state` is `always_enabled`, the code at here will update the state field of feature to `invariant_state`. However, the code from line 761 ~ 764 will disable this feature. So I think the code at line 758 should be `entry['state'] = state`, right?

src/sonic-host-services/scripts/hostcfgd
stepanblyschak: Ok

src/sonic-host-services/scripts/hostcfgd
stepanblyschak: Ok

src/sonic-host-services/scripts/hostcfgd
stepanblyschak: No sure I get this comment, are you saying there is a bug in original code?
Could you please point me to a document describing feature state transitions?

stepanblyschak: @yozhao101 Do you have this comment still?

stepanblyschak: @yozhao101 could you please check if the last commit addresses your comment?

src/sonic-host-services/scripts/hostcfgd
```diff
@@ -57,15 +71,210 @@ def obfuscate(data):
...
+        #                 -> always_disabled
+        #                 -> enabled
+        #                 -> disabled
+        #  always_enabled -> always_disabled
+        #  enabled        -> disabled
+        #  disabled       -> enabled
+        if cached_feature.state is None:
+            enable = feature.state in ("always_enabled", "enabled")
+            disable = feature.state in ("always_disabled", "disabled")
+        elif cached_feature.state == ("always_enabled", "always_disabled"):
```
tahmed-dev, line 181: `in` instead of `==`: `elif cached_feature.state in ("always_enabled", "always_disabled"):`

src/sonic-host-services/scripts/hostcfgd
```diff
@@ -57,15 +71,210 @@ def obfuscate(data):
...
+        cached_feature = self._cached_config[feature.name]
+        enable = False
+        disable = False
+
+        # Allowed transitions:
+        #  None           -> always_enabled
+        #                 -> always_disabled
+        #                 -> enabled
+        #                 -> disabled
+        #  always_enabled -> always_disabled
```
tahmed-dev, line 175: Did you mean?
`always_disabled -> always_disabled`
`always_enabled -> always_enabled`

src/sonic-host-services/scripts/hostcfgd
stepanblyschak: Thanks for noticing this!

src/sonic-host-services/scripts/hostcfgd
stepanblyschak: This is something I was wondering and related to comments from yozhao101.

Is a transition always_enabled -> always_disabled - allowed? In the current hostcfgd from master it is.
I also noticed few issues with current hostcfgd with regards to these transitions that I am not sure if they are expected or not, didn't found much info about these new "always*" states.
Transition to the same state is noop. If you could point me to some documents/PR describing the desired behaviour it would help me a lot to understand this.

stepanblyschak: @tahmed-dev

tahmed-dev: Thanks @stepanblyschak! The always_enabled was introduced by @abdosi via [PR:6000](https://github.com/Azure/sonic-buildimage/pull/6000) It is meant to be immutable. It got enhance by @yxieca vi [PR:6067](https://github.com/Azure/sonic-buildimage/pull/6067) to allow the feature to be disabled. Currently `always_disabled` is used to disable feature on deployments that don't support this feature.

stepanblyschak: I see, so I think it works as expected with my changes:
```
admin@r-bulldog-04:~$ show feature status | grep teamd
teamd           always_enabled  enabled
admin@r-bulldog-04:~$ sudo config feature state teamd disabled
Feature 'teamd' state is always enabled and can not be modified
admin@r-bulldog-04:~$ show feature status | grep teamd
teamd           always_enabled  enabled
admin@r-bulldog-04:~$ redis-cli -n 4 hset "FEATURE|teamd" state disabled
(integer) 0
admin@r-bulldog-04:~$ show feature status | grep teamd
teamd           always_enabled  enabled
admin@r-bulldog-04:~$ redis-cli -n 4 hset "FEATURE|teamd" state always_disabled
(integer) 0
admin@r-bulldog-04:~$ show feature status | grep teamd
teamd           always_disabled  enabled
admin@r-bulldog-04:~$ redis-cli -n 4 hset "FEATURE|teamd" state always_enabled
(integer) 0
admin@r-bulldog-04:~$ show feature status | grep teamd
teamd           always_disabled  enabled
admin@r-bulldog-04:~$ redis-cli -n 4 hset "FEATURE|teamd" state enabled
(integer) 0
admin@r-bulldog-04:~$ show feature status | grep teamd
teamd           always_disabled  enabled

```

renukamanavalan:
@tahmed-dev can you please provide your approval?

renukamanavalan:
@yozhao101 can you please provide your review/approval ASAP?

renukamanavalan:
@jleveque, can you please provide your review or approval?

src/sonic-host-services/scripts/hostcfgd
```diff
@@ -643,9 +852,8 @@ class HostConfigDaemon:
         self.hostname_cache=""
         self.aaacfg = AaaCfg()
         self.iptables = Iptables()
+        self.feature = FeatureHandler(self.config_db, self.device_config)
```
jleveque, line 855: Maybe rename this `self.feature_handler`?

stepanblyschak: Done

jleveque (approved):
Looks good from my perspective. @yozhao101: Can you please review again to make sure all your concerns have been addressed?

yozhao101:
> Looks good from my perspective. @yozhao101: Can you please review again to make sure all your concerns have been addressed?

Thanks, Joe and Renuka ! I am checking ...

src/sonic-host-services/scripts/hostcfgd
```diff
@@ -36,16 +36,30 @@ RADIUS_SERVER_TIMEOUT_DEFAULT = "5"
...
 RADIUS_PAM_AUTH_CONF_DIR = "/etc/pam_radius_auth.d/"
 
+
+def run_cmd(cmd, log_err=True, raise_exception=False):
+    try:
+        subprocess.check_call(cmd, shell=True)
+    except Exception as err:
+        if log_err:
+            syslog.syslog(syslog.LOG_ERR, "{} - failed: return code - {}, output:\n{}"
+                  .format(err.cmd, err.returncode, err.output))
```
yozhao101, line 46: It looks like we only have `output` from child process if it was captured by `run()` or `check_output()`. Otherwise, `None`. Please see: https://docs.python.org/3/library/subprocess.html#subprocess.CalledProcessError.output

stepanblyschak: This issue is relevant for existing code as well - https://github.com/Azure/sonic-buildimage/blob/master/src/sonic-host-services/scripts/hostcfgd#L66. This PR has no intend to fix this issue.

src/sonic-host-services/scripts/hostcfgd
```diff
@@ -57,15 +71,210 @@ def obfuscate(data):
...
+
+        if disable:
+            self.disable_feature(feature)
+            syslog.syslog(syslog.LOG_INFO, "Feature {} is stopped and disabled".format(feature.name))
+
+        return True
+
+    def update_feature_auto_restart(self, feature):
+        restart_config = "always" if feature.auto_restart == "enabled" else "no"
+        service_conf = "[Service]\nRestart={}\n".format(restart_config)
```
yozhao101, line 207: Do we consider put the service configuration entry `RestartSec=30` in the `service_conf` since `RestartSec` is usually configured with `Restart`? Please review this: https://www.freedesktop.org/software/systemd/man/systemd.service.html#RestartSec=

tahmed-dev (approved):

src/sonic-host-services/scripts/hostcfgd
stepanblyschak: We have default RestartSec=30 in all service files and there is no way to override it via FEATURE table configuration, so it is not put in service_conf.

yozhao101: What I suggested is not to override the `RestartSec=30` configuration. Would it be better to do as following:

`service_conf = "[Service]\nRestart={}\nRestartSec=30\n".format(restart_config)`

stepanblyschak: But then user will not be able to set this value on a per service basis (like changing systemd serivce file) - he must change this hardcoded value in hostcfgd but then it will apply to all services.
I would like to wait till we introduce this configuration via FEATURE table.

yozhao101: I am just wondering if we want to provide the user the ability to manually configure the `RestartSec=` in each service configuration file `like changing systemd service file`, do we need also provide such ability for users to manually change the `Restart=` value `like changing systemd service file`? The user may want to change `Restart=always` to `Restart=on-failure` or `Restart=on-abnormal`?

stepanblyschak: This means not a regular user, but developer. How to change restart sec for a particular service at build time in service file if it is hardcoded to be 30 sec for all of them? I would like not to introduce such limitation.
A restart configuration change is provided to the user by "config feature auto-restart". A more granular restart options are not exposed by SONiC.

yozhao101: I agreed with you we should not introduce the limitation to hardcode the `RestartSec=30`.

I know the user can use the CLI `sudo config feature autorestart <container_name> <disabled |
enabled>` to disable or enable the auto-restart feature of each container. 

What I am thinking is  whether it is a good way to hardcode `Restart=always` if the developer wants to introduce `Restart=on-failure` or `Restart=on-abnormal` in future?

stepanblyschak: It is different for Restart= because it is exposed to user with "auto_restart" field in SONiC. Lets say today user knows he can configure auto restart via config db, the container auto restarts type is "always". If someone changes it to "on-abnormal" in next version - that will mean that despite auto restart is enabled in config db service will not restart in case service exited with failure. This will be confusing, if developer wants to introduce these restart options it should now be done through CONFIG DB and auto_restart will have more faules e.g: - <enabled|disabled|on-failure|on-abnormal>.

stepanblyschak: @yozhao101 Kindly reminder

yozhao101: Thanks so much for your response. This discussion is resolved.

renukamanavalan:
@yozhao101 can you please resolve/provide your reviews and approval ?

yozhao101 (approved):

renukamanavalan:
Waiting for build to succeed. @stepanblyschak, if you can, please ping me, when build succeeds.

stepanblyschak:
```
tests/hostcfgd/hostcfgd_radius_test.py:10: in <module>
    from parameterized import parameterized
E   ModuleNotFoundError: No module named 'parameterized
```

Same error appears again - https://dev.azure.com/mssonic/build/_build/results?buildId=19697&view=logs&j=88ce9a53-729c-5fa9-7b6e-3d98f2488e3f&t=8d99be27-49d0-54d0-99b1-cfc0d47f0318&l=527

PR #5739, opened 2020-10-28
 [caclmgrd] Convert to Python 3; Add to sonic-host-services package

jleveque:
**- Why I did it**

To consolidate host services and install via packages instead of file-by-file, also as part of migrating all of SONiC to Python 3, as Python 2 is no longer supported.

**- How to verify it**

Ensure caclmgrd is installed and runs

**- Which release branch to backport (provide reason below if selected)**

- [ ] 201811
- [ ] 201911
- [ ] 202006

PR #5748, opened 2020-10-29
Accomadating lower case ACL rules for Control Plane ACLs

madhanmellanox:
- Why I did it
To make Control plane ACLs handle case insensitive ACL rules. Currently, it handles only upper case ACL rules.

- How I did it
I set the rule prop dictionary to be all upper case before reading the rule props for processing.

- How to verify it
Configure ACL rules with any combination of case of ACL rules, either it be upper case or lower case. All rules should be created properly in the switch and working properly.

- [ ] 201811
- [x] 201911
- [ ] 202006

The change is a one line change in caclmgrd script.

### DIFF
--- a/src/sonic-host-services/scripts/caclmgrd
+++ b/src/sonic-host-services/scripts/caclmgrd
@@ -400,7 +400,6 @@ class ControlPlaneAclManager(daemon_base.DaemonBase):
                 acl_rules = {}
 
                 for ((rule_table_name, rule_id), rule_props) in self._rules_db_info.items():
-                for ((rule_table_name, rule_id), rule_props) in self._rules_db_info.iteritems():
                     rule_props = {k.upper(): v for k,v in rule_props.iteritems()}
                     if rule_table_name == table_name:
                         if not rule_props:
--- a/src/sonic-host-services/scripts/caclmgrd
+++ b/src/sonic-host-services/scripts/caclmgrd
@@ -400,7 +400,7 @@ class ControlPlaneAclManager(daemon_base.DaemonBase):
                 acl_rules = {}
 
                 for ((rule_table_name, rule_id), rule_props) in self._rules_db_info.iteritems():
-                    rule_props = {k.upper(): v for k,v in rule_props.iteritems()}
+                    #rule_props = {k.upper(): v for k,v in rule_props.iteritems()}
                     if rule_table_name == table_name:
                         if not rule_props:
                             self.log_warning("rule_props for rule_id {} empty or null!".format(rule_id))
--- a/src/sonic-host-services/scripts/caclmgrd
+++ b/src/sonic-host-services/scripts/caclmgrd
@@ -400,7 +400,7 @@ class ControlPlaneAclManager(daemon_base.DaemonBase):
                 acl_rules = {}
 
                 for ((rule_table_name, rule_id), rule_props) in self._rules_db_info.items():
-                    rule_props = {k.upper(): v for k,v in rule_props.iteritems()}
+                    #rule_props = {k.upper(): v for k,v in rule_props.iteritems()}
                     if rule_table_name == table_name:
                         if not rule_props:
                             self.log_warning("rule_props for rule_id {} empty or null!".format(rule_id))
--- a/src/sonic-host-services/scripts/caclmgrd
+++ b/src/sonic-host-services/scripts/caclmgrd
@@ -400,6 +400,8 @@ class ControlPlaneAclManager(daemon_base.DaemonBase):
                 acl_rules = {}
 
                 for ((rule_table_name, rule_id), rule_props) in self._rules_db_info.items():
+                for ((rule_table_name, rule_id), rule_props) in self._rules_db_info.iteritems():
+                    rule_props = {k.upper(): v for k,v in rule_props.iteritems()}
                     if rule_table_name == table_name:
                         if not rule_props:
                             self.log_warning("rule_props for rule_id {} empty or null!".format(rule_id))
--- a/src/sonic-host-services/scripts/caclmgrd
+++ b/src/sonic-host-services/scripts/caclmgrd
@@ -399,7 +399,7 @@ class ControlPlaneAclManager(daemon_base.DaemonBase):
 
                 acl_rules = {}
 
-                for ((rule_table_name, rule_id), rule_props) in self._rules_db_info.items():
+                for ((rule_table_name, rule_id), rule_props) in self._rules_db_info.iteritems():
                     rule_props = {k.upper(): v for k,v in rule_props.iteritems()}
                     if rule_table_name == table_name:
                         if not rule_props:
--- a/src/sonic-host-services/scripts/caclmgrd
+++ b/src/sonic-host-services/scripts/caclmgrd
@@ -399,8 +399,8 @@ class ControlPlaneAclManager(daemon_base.DaemonBase):
 
                 acl_rules = {}
 
-                for ((rule_table_name, rule_id), rule_props) in self._rules_db_info.iteritems():
-                    #rule_props = {k.upper(): v for k,v in rule_props.iteritems()}
+                for ((rule_table_name, rule_id), rule_props) in self._rules_db_info.items():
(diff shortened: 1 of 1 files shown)


madhanmellanox:
@jleveque can you please review this PR?

jleveque:
@madhanmellanox: Can you please explain the need for this? Why can't we continue relying on the rule prop keys to be all caps? And why the need for this now? What changed?

madhanmellanox:
@jleveque because we check for SRC_IPV6 / DST_IPV6 / TCP_FLAGs in caclmgrd code, the checks are becoming case sensitive and when we configure using lower case letters, the ACL rule is not working. For example, in my ACL rule if I have SRC_IPv6 instead of SRC_IPV6, the rule in ip6tables is not right. This is because of cases sensitive checks we make in the code. It is a Day-1 issue. It is not new.

jleveque:
@madhanmellanox: The acl-loader application configures these to be all-caps before inserting them into config DB. The only way I can see this being a problem is if you're adding them manually, in which case, why not add them as all caps?

I'm not opposed to this change, but I'm just not convinced that it's necessary.

madhanmellanox:
Yes, this is our QA defect and they individually configure the ACLs manually. In any case it should not fail.

jleveque:
I see. Technically, this is human error, as it doesn't follow the spec we had defined, where these keys should always be uppercase in config DB. I feel like we should firm up a spec here, but as this change seems safe, I think we can go ahead with it.

However, please note that I have an open PR to migrate caclmgrd to Python 3 and move it to a new package in the master branch: https://github.com/Azure/sonic-buildimage/pull/5739. That PR should be merging soon, in which case, this PR will have conflicts, and will need to be rebased, but then it will not be cherry-pickable to 201911, so a separate PR would be needed.

madhanmellanox:
retest this please

madhanmellanox:
@jleveque I discussed with QA and they said ACL High level design, has these fields in lower case when defining config db schemas. That is the reason for ambiguity to enter config in config_db.json in lower case letters.
https://github.com/Azure/SONiC/wiki/ACL-High-Level-Design#31212-acl-rules-table

Can schema be defined in Upper case letters? Can I change the ACL HLD document?

jleveque:
@madhanmellanox: Thank you for pointing to the root cause of the ambiguity. I think the crux of the issue is that if the ACLs are defined in an acl.json file and loaded using acl-loader, it will convert to all caps and also replace hyphens with underscores before pushing to Config DB. I originally wrote caclmgrd assuming that all ACLs will be loaded this way, and didn't consider the case where they would be manually entered into a config_db.json file and loaded. This, combined with the HLD you provided makes the case that caclmgrd should be able to handle both upper-case and lower-case.

For now, I think we can go ahead and add your change. However, the HLD might also need some clarification, as it states things like "acl_table_name must be unique" -- we might want to clarify that "acl_table_name must be unique, regardless of case" or similar.

Can you please fix the conflicts here, since my other PR has merged to move the file and convert it to Python 3?

madhanmellanox:
retest this please

madhanmellanox:
retest this please

madhanmellanox:
retest vsimage

madhanmellanox:
retest this please

madhanmellanox:
@jleveque vstest cacl.test_cacl_application.test_cacl_application is failing for me consistently. Not sure because of converting the case of rules. There is a list of Missing IPtables rules listed in the output. Can you please check.

jleveque:
@madhanmellanox: I can only assume that this patch is causing the test to fail. Do you have a testbed you can apply your change to and run the test to reproduce?

madhanmellanox:
@jleveque I don't find the test case tests/cacl/test_cacl_application.py in sonic-buildimage repository or sonic-swss repository. Where can I find it. I thought it is a vstest and I will find it in sonic-swss/tests. But, I don't find it there.

madhanmellanox:
@jleveque I found the test case part of sonic-mgmt framework.

madhanmellanox:
retest this please

madhanmellanox:
@jleveque for testing purpose, I commented my code of making all rules Upper case, even with this change commented out, I am seeing vsimage cacl test case failing. 

I also tried checking with my team here regarding running vstest with kvm sonic image. But, they are not familiar. That is why I tried commenting out my change and still seeing the same test case failure.

jleveque:
Retest vs please

jleveque:
Retest vsimage please

jleveque:
Retest vsimage please

src/sonic-host-services/scripts/caclmgrd
```diff
@@ -399,7 +399,8 @@ class ControlPlaneAclManager(daemon_base.DaemonBase):
 
                 acl_rules = {}
 
-                for ((rule_table_name, rule_id), rule_props) in self._rules_db_info.items():
+                for ((rule_table_name, rule_id), rule_props) in self._rules_db_info.iteritems():
```
jleveque, line 402: Why change `.items()` to `.iteritems()`? This file is now run with Python 3; `.iteritems()` no longer exists. This may be the root of the test failure?

madhanmellanox: The change was not intentional. When trying to resolve merge conflict, the code got changed mistakenly. I will correct it and submit.

madhanmellanox:
@jleveque with changing iteritems to items to traverse the rule props, the CACL test case is still failing, do you want me to comment my code and run?

jleveque:
@madhanmellanox: This is a different test failure. You can try commenting out your change and see if it passes.

madhanmellanox:
@jleveque commenting out my code the vsimage - all test cases passed. So, I am uncommenting out my code and submitting it again.

madhanmellanox:
@jleveque the vsimage is failing again in test_cacl_function with error log "SSH did not time out when expected". I am not sure what my change has to do with this error.

src/sonic-host-services/scripts/caclmgrd
```diff
@@ -400,6 +400,7 @@ class ControlPlaneAclManager(daemon_base.DaemonBase):
                 acl_rules = {}
 
                 for ((rule_table_name, rule_id), rule_props) in self._rules_db_info.items():
+                    rule_props = {k.upper(): v for k,v in rule_props.iteritems()}
```
jleveque, line 403: As with the previous problem, I you also have `.iteritems()` here, which is no longer available in Python 3. It should be replaced with `.items()`.

jleveque (changes requested):

src/sonic-host-services/scripts/caclmgrd
madhanmellanox: Addressed in next commit

madhanmellanox:
retest broadcom

madhanmellanox:
retest this please

madhanmellanox:
@jleveque although VSimage tests are passing now, baseimage is always failing. This is the build error log.
16:22:48  Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
16:22:49  Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
16:22:50  make: *** [docker-start] Error 1
16:22:50  slave.mk:616: recipe for target 'docker-start' failed
16:22:51  Makefile.work:235: recipe for target 'stretch' failed
16:22:51  make[1]: *** [stretch] Error 2
16:22:51  make[1]: Leaving directory '/data2/johnar/workspace/generic/buildimage-baseimage-pr'
16:22:51  Makefile:7: recipe for target 'target/docker-ptf.gz' failed
16:22:51  make: *** [target/docker-ptf.gz] Error 2

jleveque:
Looks like a transient Jenkins issue. Retesting should fix. Might take a couple tries, unfortunately.

jleveque:
Retest baseimage please

madhanmellanox:
@jleveque can you please approve this PR?

jleveque (approved):

jleveque:
@madhanmellanox: Since caclmgrd has moved to a new location, this will not cherry-pick cleanly to 201911. Can you please open a separate PR against the 201911 branch?

PR #5763, opened 2020-10-30
Revert "[docker-base]: Rate limit priority INFO and lower in syslog"

theasianpianist:
Reverts Azure/sonic-buildimage#5666

Since Azure/sonic-sairedis#680 has been merged, this is no longer necessary.

### DIFF
--- a/dockers/docker-base-buster/etc/rsyslog.conf
+++ b/dockers/docker-base-buster/etc/rsyslog.conf
@@ -12,11 +12,10 @@
 $ModLoad imuxsock # provides support for local system logging
 
 #
-# Set a rate limit on messages from the container of priority INFO or lower (level 6 and above)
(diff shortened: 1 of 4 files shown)


jleveque:
Are we sure we don't want to keep this change? If some other info-level log message starts spamming, we could rate-limit and lose more important messages, correct?

 Or should we revert it for now and keep it in our back pocket in case we decide it's needed in the future?

yxieca (approved):

lguohan:
retest this baseimage please

lguohan:
retest baseimage please

theasianpianist:
@jleveque We've only been seeing this rate limiting from orchagent, and the default log level for that is `NOTICE` anyways. I'd say it's better to revert now but keep this in our back pocket like you said.

PR #5764, opened 2020-10-31
[docker-platform-monitor] Check if sonic_platform package is available before checking if installed

jleveque:
**- Why I did it**

On Arista platforms, sonic_platform packages are not installed in the PMon container, but are rather mounted into the container from the host OS. Therefore, `pip show sonic_platform` will fail in the PMon container.  This change will first check if we can import sonic_platform. If this fails, it will then fall back to checking if the package is installed. If both fail, it will attempt to install the package.

### DIFF
--- a/dockers/docker-platform-monitor/start.sh
+++ b/dockers/docker-platform-monitor/start.sh
@@ -16,7 +16,7 @@ if [ -e /usr/share/sonic/platform/platform_wait ]; then
 fi
 
 # If the Python 2 sonic-platform package is not installed, try to install it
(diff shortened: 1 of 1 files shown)


jleveque:
@Staphylo: Please review.

jleveque:
Retest broadcom please

jleveque:
Retest mellanox please

jleveque:
Retest vsimage please

lguohan (approved):

Staphylo:
A bit late to the party but this will work perfectly with our current approach.
Thanks for the heads up Joe.

PR #5773, opened 2020-11-02
[MultiDB] add mutidb warmboot support - restoring database

dzhangalibaba:
* restoring each database  with all data before warmboot and then flush unused data in each instance, following the multiDB warmboot design at https://github.com/Azure/SONiC/blob/master/doc/database/multi_database_instances.md
  * restore needs to be done in database docker since we need to know the database_config.json in new version
  * copy all data rdb file into each instance restoration location andthen flush unused database
  * other logic is the same as before
*  backing up database part is in another PR at sonic-utilities https://github.com/Azure/sonic-utilities/pull/1205, they depend on each other
Signed-off-by: Dong Zhang d.zhang@alibaba-inc.com

### DIFF
--- a/dockers/docker-database/docker-database-init.sh
+++ b/dockers/docker-database/docker-database-init.sh
@@ -87,7 +87,7 @@ do
     mkdir -p /var/lib/$inst
     if [[ -f $DUMPFILE ]]; then
         # copy warmboot rdb file into each new instance location
-        cp $DUMPFILE /var/lib/$inst/dump.rdb
+        cp -n $DUMPFILE /var/lib/$inst/dump.rdb
     else
         echo -n > /var/lib/$inst/dump.rdb
     fi
--- a/dockers/docker-database/docker-database-init.sh
+++ b/dockers/docker-database/docker-database-init.sh
@@ -79,4 +79,18 @@ else
 fi
 rm $db_cfg_file_tmp
 
+# flush unused database after warmboot restoration
+DUMPFILE=/var/lib/redis/dump.rdb
+redis_inst_list=`/usr/bin/python -c "import swsssdk; print(' '.join(swsssdk.SonicDBConfig.get_instancelist().keys()))"`
+for inst in $redis_inst_list
+do
+    mkdir -p /var/lib/$inst
+    if [[ -f $DUMPFILE ]]; then
+        # copy warmboot rdb file into each new instance location
+        cp $DUMPFILE /var/lib/$inst/dump.rdb
+    else
+        echo -n > /var/lib/$inst/dump.rdb
+    fi
+done
+
 exec /usr/local/bin/supervisord
--- a/dockers/docker-database/docker-database-init.sh
+++ b/dockers/docker-database/docker-database-init.sh
@@ -79,7 +79,7 @@ else
 fi
 rm $db_cfg_file_tmp
 
-# flush unused database after warmboot restoration
+# copy dump.rdb file to each instance for restoration
 DUMPFILE=/var/lib/redis/dump.rdb
 redis_inst_list=`/usr/bin/python -c "import swsssdk; print(' '.join(swsssdk.SonicDBConfig.get_instancelist().keys()))"`
 for inst in $redis_inst_list
--- a/dockers/docker-database/docker-database-init.sh
+++ b/dockers/docker-database/docker-database-init.sh
@@ -87,7 +87,7 @@ do
     mkdir -p /var/lib/$inst
     if [[ -f $DUMPFILE ]]; then
         # copy warmboot rdb file into each new instance location
-        cp -n $DUMPFILE /var/lib/$inst/dump.rdb
+        cp -u $DUMPFILE /var/lib/$inst/dump.rdb
     else
         echo -n > /var/lib/$inst/dump.rdb
     fi
--- a/dockers/docker-database/docker-database-init.sh
+++ b/dockers/docker-database/docker-database-init.sh
@@ -87,7 +87,9 @@ do
     mkdir -p /var/lib/$inst
     if [[ -f $DUMPFILE ]]; then
         # copy warmboot rdb file into each new instance location
-        cp -u $DUMPFILE /var/lib/$inst/dump.rdb
+        if [[ "$DUMPFILE" != "/var/lib/$inst/dump.rdb" ]]; then
+            cp $DUMPFILE /var/lib/$inst/dump.rdb
+        fi
     else
         echo -n > /var/lib/$inst/dump.rdb
     fi
--- a/dockers/docker-database/flush_unused_database
+++ b/dockers/docker-database/flush_unused_database
@@ -4,9 +4,10 @@ import redis
 import subprocess
 import time
 
-output = ""
-while('PONG' not in output):
+while(True):
     output = subprocess.Popen(['sonic-db-cli', 'PING'], stdout=subprocess.PIPE).communicate()[0]
+    if 'PONG' in output:
+        break
     time.sleep(1)
 
 instlists = swsssdk.SonicDBConfig.get_instancelist()
--- a/dockers/docker-database/flush_unused_database
+++ b/dockers/docker-database/flush_unused_database
@@ -1,0 +1,26 @@
+#!/usr/bin/python
+import swsssdk
+import redis
+import subprocess
+import time
+
+output = subprocess.check_output('sonic-db-cli PING', stderr=subprocess.STDOUT, shell=True)
+while('PONG' not in output):
+    time.sleep(1)
+
+instlists = swsssdk.SonicDBConfig.get_instancelist()
+for instname, v in instlists.items():
+    insthost = v['hostname']
+    instsocket = v['unix_socket_path']
+
+    dblists = swsssdk.SonicDBConfig.get_dblist()
+    for dbname in dblists:
+        dbid = swsssdk.SonicDBConfig.get_dbid(dbname)
+        dbinst = swsssdk.SonicDBConfig.get_instancename(dbname)
+
+        # this DB is on current instance, skip flush
+        if dbinst == instname:
+            continue
+
+        r = redis.Redis(host=insthost, unix_socket_path=instsocket, db=dbid)
+        r.flushdb()
--- a/dockers/docker-database/flush_unused_database
+++ b/dockers/docker-database/flush_unused_database
@@ -6,7 +6,7 @@ import time
 
 output = ""
 while('PONG' not in output):
-    output = subprocess.check_output('sonic-db-cli PING', stderr=subprocess.STDOUT, shell=True)
+    output = subprocess.Popen(['sonic-db-cli', 'PING'], stdout=subprocess.PIPE).communicate()[0]
     time.sleep(1)
 
 instlists = swsssdk.SonicDBConfig.get_instancelist()
--- a/dockers/docker-database/flush_unused_database
+++ b/dockers/docker-database/flush_unused_database
@@ -4,8 +4,9 @@ import redis
 import subprocess
 import time
 
-output = subprocess.check_output('sonic-db-cli PING', stderr=subprocess.STDOUT, shell=True)
+output = ""
 while('PONG' not in output):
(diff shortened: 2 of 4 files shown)


dzhangalibaba:
@qiluo-msft

dockers/docker-database/docker-database-init.sh
```diff
@@ -79,4 +79,18 @@ else
 fi
 rm $db_cfg_file_tmp
 
+# cleanup database after warmboot
```
qiluo-msft, line 82: >cleanup [](start = 2, length = 7)

Seems not 'cleanup' ?

files/build_templates/docker_image_ctl.j2
```diff
@@ -74,11 +74,7 @@ function preStartAction()
     if [ "$DATABASE_TYPE" != "chassisdb" ]; then
         if [[ ("$BOOT_TYPE" == "warm" || "$BOOT_TYPE" == "fastfast")  && -f $WARM_DIR/dump.rdb ]]; then
             # Load redis content from /host/warmboot/dump.rdb
-            docker cp $WARM_DIR/dump.rdb database$DEV:/var/lib/redis/dump.rdb
-        else
-            # Create an emtpy file and overwrite any RDB if already there
-            echo -n > /tmp/dump.rdb
-            docker cp /tmp/dump.rdb database$DEV:/var/lib/redis/
+            docker cp $WARM_DIR/dump.rdb database$DEV:/var/lib/dump.rdb
```
qiluo-msft, line 77: >/var/lib/ [](start = 54, length = 9)

This directory seems strange.

dockers/docker-database/supervisord.conf.j2
```diff
@@ -33,3 +33,11 @@ stdout_logfile=syslog
 stderr_logfile=syslog
 {%     endfor %}
 {% endif %}
+
+[program:flushdb]
```
qiluo-msft, line 37: >flushdb [](start = 9, length = 7)

Can we delay it like 5 min? so it will not impact warm-reboot time.

dockers/docker-database/flush_db.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+# Wait until redis starts
+until [[ $(sonic-db-cli PING | grep -c PONG) -gt 0 ]]; do
+  sleep 1;
+done
+
+DUMPFILE=/var/lib/dump.rdb
+if [[ -f $DUMPFILE ]]; then
+    flush_warmboot_database
```
qiluo-msft, line 10: >flush_warmboot_database [](start = 4, length = 23)

why merge it into this script as a function?

qiluo-msft (changes requested):
As comments

dockers/docker-database/docker-database-init.sh
dzhangalibaba: uopdated

files/build_templates/docker_image_ctl.j2
dzhangalibaba: redis instance rdb file is at /var/lib/instance_name/, so I thought /var/lib is OK for temporary global location to hold rdb file. I will be deleted later when flushing db. 
BTW, what directory you think is good, /etc/sonic/ or  some directory originally exising in database docker?

dockers/docker-database/supervisord.conf.j2
dzhangalibaba: we don't need to wait. This process is not in warmboot sequence. warmboot go ahead once database docker is started.

dockers/docker-database/flush_db.sh
dzhangalibaba: flush_warmboot_database is a python script codes, we don't want to mix it with shell cmds. If we rewrite those shell cmds into python codes, it is a redundancy， for example, sonic-db-cli is already a python script.

dzhangalibaba:
@qiluo-msft  I saw another pr is merged https://github.com/Azure/sonic-utilities/pull/1205

We need this PR ready as well if we want to update sonic-utilities sub module, otherwise warm boot will be broken. Please take a look on this as well

files/build_templates/docker_image_ctl.j2
```diff
@@ -92,6 +88,7 @@ function preStartAction()
 function postStartAction()
 {
 {%- if docker_container_name == "database" %}
+    WARM_DIR=/host/warmboot
```
qiluo-msft, line 91: >WARM_DIR [](start = 4, length = 8)

Why add here?

files/build_templates/docker_image_ctl.j2
qiluo-msft: Better to keep old code, ie, still using `/var/lib/redis/dump.rdb`. Inside docker, you can copy it on-demand. If there is only 1 instance, no need to copy at all.

---
In reply to: [520934863](https://github.com/Azure/sonic-buildimage/pull/5773#discussion_r520934863) [](ancestors = 520934863)

dockers/docker-database/supervisord.conf.j2
qiluo-msft: But it still burns CPU.

---
In reply to: [520939272](https://github.com/Azure/sonic-buildimage/pull/5773#discussion_r520939272) [](ancestors = 520939272)

dockers/docker-database/flush_db.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+# Wait until redis starts
+until [[ $(sonic-db-cli PING | grep -c PONG) -gt 0 ]]; do
+  sleep 1;
```
qiluo-msft, line 5: >   [](start = 0, length = 2)

4 blanks

dockers/docker-database/flush_db.sh
qiluo-msft: Better to rewrite this in python, so merge should be easier.

---
In reply to: [520939574](https://github.com/Azure/sonic-buildimage/pull/5773#discussion_r520939574) [](ancestors = 520939574)

files/build_templates/docker_image_ctl.j2
qiluo-msft: If this is a true bug, move it to file global?

---
In reply to: [535743785](https://github.com/Azure/sonic-buildimage/pull/5773#discussion_r535743785) [](ancestors = 535743785)

qiluo-msft:
Offline discussed.
Azure/sonic-utilities#1205 is backward-compatible. So it is safe to merge that first.

files/build_templates/docker_image_ctl.j2
dzhangalibaba: kept the old directory as dicussed

dockers/docker-database/supervisord.conf.j2
dzhangalibaba: add sleep 300

dockers/docker-database/flush_db.sh
dzhangalibaba: merge into on python script

files/build_templates/docker_image_ctl.j2
dzhangalibaba: It is not a true BUG, WARM_DIR is set as global when preStart is invoked first and postStart can use its value later.

dockers/docker-database/flush_db.sh
dzhangalibaba: DONE

dockers/docker-database/flush_unused_database
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/python
+import swsssdk
+import redis
+import subprocess
+import time
+
+output = subprocess.check_output('sonic-db-cli PING', stderr=subprocess.STDOUT, shell=True)
+while('PONG' not in output):
+    time.sleep(1)
```
qiluo-msft, line 9: If PONG is not in output, you will sleep forever.

dockers/docker-database/docker-database-init.sh
```diff
@@ -79,4 +79,18 @@ else
 fi
 rm $db_cfg_file_tmp
 
+# flush unused database after warmboot restoration
```
qiluo-msft, line 82: >flush unused database after warmboot restoration [](start = 2, length = 48)

Not related to code

dockers/docker-database/flush_unused_database
dzhangalibaba: updated

dockers/docker-database/docker-database-init.sh
dzhangalibaba: updated

dockers/docker-database/flush_unused_database
qiluo-msft: It is improved, but you can improve even more:
1. if any output is good, no need to sleep and you can immediately exit loop
2. no need to check output at the beginning.

---
In reply to: [536364747](https://github.com/Azure/sonic-buildimage/pull/5773#discussion_r536364747) [](ancestors = 536364747)

dzhangalibaba: It should be OK now

qiluo-msft (dismissed):

dockers/docker-database/docker-database-init.sh
```diff
@@ -79,4 +79,18 @@ else
...
 
+# copy dump.rdb file to each instance for restoration
+DUMPFILE=/var/lib/redis/dump.rdb
+redis_inst_list=`/usr/bin/python -c "import swsssdk; print(' '.join(swsssdk.SonicDBConfig.get_instancelist().keys()))"`
+for inst in $redis_inst_list
+do
+    mkdir -p /var/lib/$inst
+    if [[ -f $DUMPFILE ]]; then
+        # copy warmboot rdb file into each new instance location
+        cp $DUMPFILE /var/lib/$inst/dump.rdb
```
qiluo-msft, line 90: If `cp` source file is same as the destination, there will be error message and exit code.

qiluo-msft (commented):
One more comment

dockers/docker-database/docker-database-init.sh
dzhangalibaba: I tried it on DUT earlier, it just shows a msg telling it is the same file, it didn't affect the function.

qiluo-msft: `cp -n` fits this use case better.

---
In reply to: [536449280](https://github.com/Azure/sonic-buildimage/pull/5773#discussion_r536449280) [](ancestors = 536449280)

dzhangalibaba: agree. Updated

qiluo-msft (dismissed):

dockers/docker-database/docker-database-init.sh
dzhangalibaba: I revisit the behavior, to not change today’s behavior,  cp -u is correct instead of -n. There is a case we need to overwrite. For example, after warm boot, we do a cold boot, an empty rdb file wil overwrite the previous warm boot rdb file. So -u is correct, only newer source will be copied.

qiluo-msft: Thanks for the insight and sorry for the misleading. Even `cp -u` is not exactly what you truly need, considering the extreme the case the time adjusted backward. So either option is ok:
1. revert back to bare `cp`, and keep error messages
2. check $DUMPFILE != /var/lib/$inst/dump.rdb before `cp`

---
In reply to: [536868596](https://github.com/Azure/sonic-buildimage/pull/5773#discussion_r536868596) [](ancestors = 536868596,536449280)

dzhangalibaba: I am checking if src and dst file is the same. Updated

qiluo-msft (approved):

yxieca (approved):

qiluo-msft:
Retest vsimage please

dzhangalibaba:
@qiluo-msft looks everything passed. Could we merg this and at the same time , I'll open another for updating sonic-utilities

PR #5783, opened 2020-11-02
[generate_asic_config_checksum.py] Convert to Python 3

jleveque:
**- How I did it**

- Convert script to Python 3
    - Need to open file in binary mode before hashing due to new string data type in Python 3 being unicode by default. This should probably have been done regardless.
- Reorganize imports alphabetically
- When running the script, don't explicitly call `python`. Instead let the program loader use the interpreter specified in the shebang (which is now `python3`).

**- How to verify it**

Ensure script continues to work as expected

- [ ] 201811
- [ ] 201911
- [ ] 202006


daall (approved):

daall (commented):
Should we switch to using the common `Logger` utility in this file as well?

daall:
> Should we switch to using the common `Logger` utility in this file as well?

Oh wait this is a build script, so we can't use the sonic-py-common, right?

daall (dismissed):

jleveque:
Retest vs please

daall (approved):

PR #5805, opened 2020-11-04
Per-switching silicon Common config for Broadcom Supported Platforms

geans-pin:
**- Why I did it**
We added support for a per-switching silicon common config.bcm file to assist with silicon-wide application configuration settings for all affected Broadcom platforms. The infrastructure saves development time by allowing applications to refactor common settings in a single location, and allows for platform-specific overrides. 

**- How I did it**
Modify the SYNCD script syncd_init_common.sh ( https://github.com/Azure/sonic-sairedis/pull/693 ) 

**- How to verify it**
The finial config.bcm and sai.profile will copy to the shared folder /var/run/sswsyncd/ in the syncd docker for 'show tech' debugging purpose. Developer can check the final config.bcm file for debugging from the show tech dump /sai/xxx.config.bcm.


geans-pin:
close and re-submit

PR #5842, opened 2020-11-06
[fast-reboot] Fix arp entries will be cleared after swssconfig restores.

dennis0113:
**- Why I did it**
Fix the issue that arp entries restored by swssconfig will be cleared by neighsyncd.
**- How I did it**
Let swssconfig be executed after than neighsyncd 
**- How to verify it**
After executing fast-reboot command, the arp entries in arp.json can be restored to redis. 

- [ ] 201811
- [ ] 201911
- [x] 202006


dockers/docker-orchagent/supervisord.conf
```diff
@@ -59,35 +59,35 @@ stderr_logfile=syslog
...
-autorestart=false
-startsecs=0
+autorestart=unexpected
 startretries=0
+startsecs=0
 stdout_logfile=syslog
 stderr_logfile=syslog
 dependent_startup=true
-dependent_startup_wait_for=swssconfig:exited
+dependent_startup_wait_for=neighsyncd:running
```
stepanblyschak, line 82: Not sure that neighsyncd:running state does mean that neighsyncd already called AppRestartAssist::registerAppTable().
Potentially there is still a risk for this race condition to happen.
Probably AppRestartAssist::registerAppTable should be fixed - instead of clearing pending data let the orch consume it?
The AppRestartAssist is related to warm logic so such change may affect the warm reboot/restart.
@yxieca @prsunny @tahmed-dev @lguohan

CharlieChenEC: This issue is caused by the race condition between the init table and start writing table on two threads. I propose to put the init operation into swssconfig. It does not need to care what the detailed implementation of the underlying 'running' state.

bingwang-ms: I agreed with @stepanblyschak. Could we add a check for warm-reboot at the beginning of AppRestartAssist if it's only needed for warm-reboot?

bingwang-ms:
I attempted to fix it in sonic-swss https://github.com/Azure/sonic-swss/pull/1498. Could you help to have a look at?

PR #5848, opened 2020-11-06
[multi-asic] Enhanced the getMount() API and  docker mount point check 

abdosi:
Why I did:
API getMount() API was not updated to handle multi-asic platforms
because of which mount point comparison always return FALSE
and we were always creating new container image when doing config load_minigraph/reload/reboot.

How I did:
Updated API getMount() to return abspath() for Docker Mount Point
and use that one for mount point comparison

How I Verify:

- Verify on Single ASIC platforms behavior is same as present

- Verify on Multi-asic platforms we do not create new container when doing config load_minigraph/reload/reboot.

- Also sonic-mgmt copp test were failing on Multi-asic platform 
since test case update copp.json in swss docker and restart docker but that
change did not took impact since we were creating new container. 
After this changes test case is working fine

### DIFF
--- a/files/build_templates/docker_image_ctl.j2
+++ b/files/build_templates/docker_image_ctl.j2
@@ -45,7 +45,7 @@ function updateSyslogConf()
 
 function getMountPoint()
 {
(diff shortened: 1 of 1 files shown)


abdosi:
retest broadcom please

abdosi:
retest broadcom please

judyjoseph (approved):
LGTM

abdosi:
@SuvarnaMeenakshi Please review this.

SuvarnaMeenakshi (approved):

PR #5860, opened 2020-11-08
[sonic-config-engine/sonic-cfggen]: Fix natsort import.

praveen-li:
natsorted is used but not imported in sonic-cfggen.

Signed-off-by: Praveen Chaudhary<pchaudhary@linkedin.com>

**- Why I did it**

**- How I did it**

**- How to verify it**

- [ ] 201811
- [ ] 201911
- [ ] 202006

### DIFF
--- a/src/sonic-config-engine/sonic-cfggen
+++ b/src/sonic-config-engine/sonic-cfggen
@@ -44,7 +44,7 @@ from redis_bcc import RedisBytecodeCache
 from sonic_py_common.multi_asic import get_asic_id_from_name, is_multi_asic
 from sonic_py_common import device_info
 from swsssdk import SonicV2Connector, ConfigDBConnector, SonicDBConfig, ConfigDBPipeConnector
-
+from natsort import natsorted
 
 PY3x = sys.version_info >= (3, 0)
 


src/sonic-config-engine/sonic-cfggen
```diff
@@ -44,7 +44,7 @@ from redis_bcc import RedisBytecodeCache
 from sonic_py_common.multi_asic import get_asic_id_from_name, is_multi_asic
 from sonic_py_common import device_info
 from swsssdk import SonicV2Connector, ConfigDBConnector, SonicDBConfig, ConfigDBPipeConnector
-
+from natsort import natsorted
```
lguohan, line 47: we explicitly removed natsort since natsort makes sonic-cfggen very slow.

lguohan: check out this one. https://github.com/Azure/sonic-buildimage/commit/9122d16dd28c36d10a05ce0b30d81ff6273f1c63#diff-748e2f017a5a2f6d8980786a9bd426d7412aeb43ea7f561b1ae934be91f9fc22

please add reason why natsorted needs to be added, where does it get used.

lguohan: @tahmed-dev fyi

praveen-li: @Iguohan, In line 216, it is used in the below function. Without this import, sonic-cfggen fails, when sort_data is called. So the question comes, why sort_data function is not fixed when import for natsort is removed.

```
sort_data(data):
    for table in data:
        if type(data[table]) is dict:
            data[table] = OrderedDict(natsorted(data[table].items()))
    return data
```

jleveque: I'll let @tahmed-dev comment on this. However, if we do need this import, you should also add `natsort==6.2.1` to the common dependencies in setup.py, as 6.2.1 is the last version which supports Python 2 and is tested to be working. We can upgrade it later once we no longer support Python 2.

praveen-li: @jleveque: Yeah we can make sort_data work without natsorted, we do not need it. Once we finalize that, I can do other changes. Thx.

tahmed-dev: @praveen-li , @jleveque `sort_data` is not used anywhere in sonic cfggen and I wanted to removed as part of the PR that removed the natsort, however one of the module was explicitly using this method and so I left in as I did not want to break them. I left a comment to that effect above it in [line:212](https://github.com/Azure/sonic-buildimage/pull/5860/files#diff-748e2f017a5a2f6d8980786a9bd426d7412aeb43ea7f561b1ae934be91f9fc22L212). I am not sure if the platform module is still there, if not, please remove the `sort_data` please.

tahmed-dev: Update, the comment indeed read sonic-utilities. Here is the [line](https://github.com/Azure/sonic-utilities/blob/master/config/config_mgmt.py#L23) that prevented removing it.

tahmed-dev (changes requested):
I'll mark this PR with changes requested as we sure do not need natsort back in again

src/sonic-config-engine/sonic-cfggen
praveen-li: @tahmed-dev : So Dy Port Breakout is also broken after the changes of removing natsort. See, keeping a function does not help if it is broken anyway :) . 
This way, it will regress in some other unexpected command and will be hard to debug. I saw Mellanox struggled with "global name 'natsorted' not defined" error already. 
About the solution, if we are sure that sort_data is not used anywhere else I will try to fix it in config_mgmt.py

tahmed-dev: @praveen-li sorry about that. That was an oversight. To answer your question, the `sort_data` us not used in sonic-cfggen. I'd suggest to relocate the sort logic to `config_mgmt.py` or completely remove such logic if not required.

tahmed-dev: I'd also suggest adding test case to the sonic_mgmt to capture such failures.

praveen-li: @tahmed-dev, no issues, 
Let me try, If I can bring sort_data logic to only config_mgmt. First, I need to again check, if it is absolutely necessary to use sort_data. If yes, I will probably import natsorted at the function level. Thx.

praveen-li: > I'd also suggest adding test case to the sonic_mgmt to capture such failures.

Yes Agree. Actually, I am also wondering, how VS tests passed with this change. Because atleast DPB command should fail.

tahmed-dev: @praveen-li can you please check this [PR:1264](https://github.com/Azure/sonic-utilities/pull/1264)

PR #5873, opened 2020-11-10
[sonic-config-engine] openconfig_acl.py: Make compliant with Python 3

jleveque:
**- Why I did it**

Python 3 is more strict with `__slots__`. As per the [documentation](https://docs.python.org/3/reference/datamodel.html#notes-on-using-slots):

> \_\_slots\_\_ are implemented at the class level by creating descriptors (Implementing Descriptors) for each variable name. As a result, class attributes cannot be used to set default values for instance variables defined by \_\_slots\_\_; otherwise, the class attribute would overwrite the descriptor assignment.

This was apparently missed when making sonic-config-engine compliant with Python 3, and errors like the following would be seen:

```
tests/acl_loader_test.py:10: in <module>
    from acl_loader.main import *
acl_loader/main.py:8: in <module>
    import openconfig_acl
/usr/local/lib/python3.7/dist-packages/openconfig_acl.py:24: in <module>
    class yc_state_openconfig_acl__acl_state(PybindBase):
E   ValueError: '_pybind_generated_by' in __slots__ conflicts with class variable
```

**- How I did it**

Set `_yang_name` and `_pybind_generated_by` members in the constructors rather than in the class bodies.

**- How to verify it**

Run tests in tests/acl_loader_test.py (sonic-utilities) and ensure no errors are seen.

- [ ] 201811
- [ ] 201911
- [ ] 202006

### DIFF
--- a/src/sonic-config-engine/openconfig_acl.py
+++ b/src/sonic-config-engine/openconfig_acl.py
@@ -32,14 +32,10 @@ class yc_state_openconfig_acl__acl_state(PybindBase):
   """
   __slots__ = ('_pybind_generated_by', '_path_helper', '_yang_name', '_extmethods', '__counter_capability',)
 
-  _yang_name = 'state'
-
-  _pybind_generated_by = 'container'
-
   def __init__(self, *args, **kwargs):
-
+    self._yang_name = 'state'
+    self._pybind_generated_by = 'container'
     self._path_helper = False
-
     self._extmethods = False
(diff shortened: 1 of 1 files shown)


src/sonic-config-engine/openconfig_acl.py
```diff
@@ -32,14 +32,10 @@ class yc_state_openconfig_acl__acl_state(PybindBase):
...
   __slots__ = ('_pybind_generated_by', '_path_helper', '_yang_name', '_extmethods', '__counter_capability',)
 
-  _yang_name = 'state'
-
-  _pybind_generated_by = 'container'
-
   def __init__(self, *args, **kwargs):
-
+    self._yang_name = 'state'
+    self._pybind_generated_by = 'container'
```
lguohan, line 37: is this generated use the command on line 13?

jleveque: No. I did not regenerate the file, I simply fixed Python 3 errors, since we currently need this file to be compatible with both Python 2 and 3. Once we remove support for Python 2 completely, we can regenerate the file using Python 3.

lguohan (approved):

PR #5905, opened 2020-11-11
[buffers_config.j2]: Use correct cable lengths for backend devices

theasianpianist:

**- Why I did it**
Backend devices do not have cable lengths properly configured because the device role is not recognized in the buffers_config template.

**- How I did it**
Remove 'backend' from the device roles so that backend devices are configured the same as non-backend devices

**- How to verify it**
Render the template on a device with type 'BackEndTorRouter' or 'BackEndLeafRouter' and confirm that the cable lengths are set correctly.

- [ ] 201811
- [ ] 201911
- [ ] 202006

### DIFF
--- a/files/build_templates/buffers_config.j2
+++ b/files/build_templates/buffers_config.j2
@@ -57,6 +57,8 @@ def
                          {%- set roles2 = neighbor_role + '_' + switch_role %}
                          {%- set roles1 = roles1 | lower %}
                          {%- set roles2 = roles2 | lower %}
+                         {%- set roles1 = roles1.replace('backend', '') %}
+                         {%- set roles2 = roles2.replace('backend', '') %}
                 {%- endif %}
                 {%- if roles1 in ports2cable %}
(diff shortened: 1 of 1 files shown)


files/build_templates/buffers_config.j2
```diff
@@ -57,6 +57,8 @@ def
                          {%- set roles2 = neighbor_role + '_' + switch_role %}
                          {%- set roles1 = roles1 | lower %}
                          {%- set roles2 = roles2 | lower %}
+                         {%- set roles1 = roles1.replace('backend', '') %}
```
prsunny, line 60: So its no-op if it is regular torrouter, correct?

prsunny (approved):

files/build_templates/buffers_config.j2
theasianpianist: correct, will just return the original string if 'backend' is not present.

PR #5906, opened 2020-11-11
[bgp]: Update TSA functionality

pavel-shirshov:
**- Why I did it**
Fixed TSA bugs:
1. TSA didn't advertise Loopback ipv6 address
2. TSA and TSB changed BGP dynamic and BGP monitors sessions

**- How I did it**
1. Use 'match ipv6 prefix-list' to match ipv6 loopback
2.  Skip all route-maps, which doesn't have V4 or V6 in it. Both dynamic and monitors sessions doesn't have such suffixes because the sessions works for both ipv4 and ipv6. In the future it is better to re develop this feature inside of bgpcfgd. 

**- How to verify it**
Build an image and run on your DUT.
```
admin@str-s6100-acs-1:~$ TSA
System Mode: Normal -> Maintenance
admin@str-s6100-acs-1:~$ vtysh -c 'show bgp ipv4 neighbors 10.0.0.1 advertised-routes'
BGP table version is 6, local router ID is 10.1.0.32, vrf id 0
Default local pref 100, local AS 64601
Status codes:  s suppressed, d damped, h history, * valid, > best, = multipath,
               i internal, r RIB-failure, S Stale, R Removed
Nexthop codes: @NNN nexthop's vrf id, < announce-nh-self
Origin codes:  i - IGP, e - EGP, ? - incomplete

   Network          Next Hop            Metric LocPrf Weight Path
*> 10.1.0.32/32     0.0.0.0                  0         32768 i

Total number of prefixes 1
admin@str-s6100-acs-1:~$ vtysh -c 'show bgp ipv6 neighbors fc00::a advertised-routes'
BGP table version is 6, local router ID is 10.1.0.32, vrf id 0
Default local pref 100, local AS 64601
Status codes:  s suppressed, d damped, h history, * valid, > best, = multipath,
               i internal, r RIB-failure, S Stale, R Removed
Nexthop codes: @NNN nexthop's vrf id, < announce-nh-self
Origin codes:  i - IGP, e - EGP, ? - incomplete

   Network          Next Hop            Metric LocPrf Weight Path
*> fc00:1::/64      ::                       0         32768 i

Total number of prefixes 1
admin@str-s6100-acs-1:~$ TSB
System Mode: Maintenance -> Normal
```

- [ ] 201811
- [x] 201911
- [x] 202006


dockers/docker-fpm-frr/total_portstat.py
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/env python3
```
lguohan, line 1: why is file in frr docker? can we put it into the sonic-utilities and then add unit test?

pavel-shirshov: Because I made a mistake and put it into the Dockerfile, and base_image_files.
I wanted to put it into the base image.
Now I remove the file completely
I wanted to use this short utility as a shortcut, because we need it for 201911. After that I wanted to implement this change as part of portstat.

lguohan (approved):

PR #5918, opened 2020-11-13
[201911][caclmgrd] Accomadating case insensitive rule props for Control plane ACLs

madhanmellanox:
**- Why I did it**
To make Control plane ACLs handle case insensitive ACL rules. Currently, it handles only upper case ACL rules.

**- How I did it**
I set the rule prop dictionary to be all upper case before reading the rule props for processing.

**- How to verify it**
Configure ACL rules with any combination of case of ACL rules, either it be upper case or lower case. All rules should be created properly in the switch and working properly.

- [ ] 201811
- [ ] 201911
- [ ] 202006

The change is a one line change in caclmgrd script.

### DIFF
--- a/files/image_config/caclmgrd/caclmgrd
+++ b/files/image_config/caclmgrd/caclmgrd
@@ -396,6 +396,7 @@ class ControlPlaneAclManager(daemon_base.DaemonBase):
                 acl_rules = {}
 
                 for ((rule_table_name, rule_id), rule_props) in self._rules_db_info.iteritems():
+                    rule_props = {k.upper(): v for k,v in rule_props.items()}
                     if rule_table_name == table_name:
                         if not rule_props:
                             self.log_warning("rule_props for rule_id {} empty or null!".format(rule_id))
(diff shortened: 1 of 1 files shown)


madhanmellanox:
@jleveque can you please review this PR? It is porting of case insensitive rule props PR from Master branch.

files/image_config/caclmgrd/caclmgrd
```diff
@@ -396,6 +396,7 @@ class ControlPlaneAclManager(daemon_base.DaemonBase):
                 acl_rules = {}
 
                 for ((rule_table_name, rule_id), rule_props) in self._rules_db_info.iteritems():
+                    rule_props = {k.upper(): v for k,v in rule_props.items()}
```
jleveque, line 399: Since this script is still using Python 2 in 201911, I think you should change this to `rule_props.iteritems()`, as `.items()` in Python 2 can consume more memory if the dict is large.

jleveque (changes requested):

files/image_config/caclmgrd/caclmgrd
madhanmellanox: addressed in next commit.

jleveque (approved):

rlhui:
@jleveque, @madhanmellanox   - are there nightly tests/pytests coverage for this change and was there any existing nightly tests done on the change?  Thanks.

jleveque:
@rlhui: Nightly tests for caclmgrd are found under the tests/cacl/ directory of sonic-mgmt.

PR #5919, opened 2020-11-13
[sonic-py-common] Add 'universal_newlines=True' arg to all Popen calls to properly support Python 3

jleveque:
**- Why I did it**

The behavior of `subprocess.Popen()` changed in Python 3 such that stdin, stdout and stderr are treated as bytes by default. Adding the `universal_newlines=True` argument changes this behavior to return strings, matching the behavior of Python 2.

**- How to verify it**'

Call the affected functions using both Python 2 and Python 3 and ensure both provide the same, correct results.

- [ ] 201811
- [ ] 201911
- [ ] 202006

### DIFF
--- a/src/sonic-py-common/sonic_py_common/device_info.py
+++ b/src/sonic-py-common/sonic_py_common/device_info.py
@@ -407,7 +407,7 @@ def get_system_mac(namespace=None):
         hw_mac_entry_cmds = [mac_address_cmd]
 
     for get_mac_cmd in hw_mac_entry_cmds:
-        proc = subprocess.Popen(get_mac_cmd, shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        proc = subprocess.Popen(get_mac_cmd, shell=True, universal_newlines=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
         (mac, err) = proc.communicate()
         if err:
             continue
@@ -439,6 +439,7 @@ def get_system_routing_stack():
         proc = subprocess.Popen(command,
                                 stdout=subprocess.PIPE,
                                 shell=True,
+                                universal_newlines=True,
                                 stderr=subprocess.STDOUT)
         stdout = proc.communicate()[0]
         proc.wait()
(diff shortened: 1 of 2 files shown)


pavel-shirshov:
I'd suggest use encoding='utf-8'. This will convert the results to strings.
universal-newlines sounds like to support Windows and linux newlines

jleveque:
> I'd suggest use encoding='utf-8'. This will convert the results to strings.
> universal-newlines sounds like to support Windows and linux newlines

I looked into this, and it seems like `universal_newlines` is the proper solution:

- https://stackoverflow.com/questions/38162400/python3-popen-do-i-have-to-use-encode-every-time

In fact, it appears as though as of Python 3.7, they have added an new `text` alias for `universal_newlines` because the parameter was unfortunately named and confusing:

> Changed in version 3.7: Added the text parameter, as a more understandable alias of universal_newlines

I can change these to read `text=True` to take advantage of the new alias. Would you prefer that?

- https://python.readthedocs.io/en/latest/library/subprocess.html?highlight=universal_newlines

jleveque:
Retest mellanox please

jleveque:
Retest mellanox please

jleveque:
Retest vsimage please

pavel-shirshov:
Hi Joe. Thank you for the investigation..

jleveque:
Retest mellanox please

jleveque:
Retest vsimage please

src/sonic-py-common/sonic_py_common/device_info.py
```diff
@@ -407,7 +407,7 @@ def get_system_mac(namespace=None):
         hw_mac_entry_cmds = [mac_address_cmd]
 
     for get_mac_cmd in hw_mac_entry_cmds:
-        proc = subprocess.Popen(get_mac_cmd, shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        proc = subprocess.Popen(get_mac_cmd, shell=True, universal_newlines=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
qiluo-msft, line 410: Thanks for the detailed investigation! So the best method should be:
1. No extra argument for python 2
2. Add `text=True` for python 3

I see you tried hard to maintain same statement for different version, actually it is not that important. Instead, I suggest you use a `kwargs` variables, add options based on python version, and pass it to `Popen` constructor.

jleveque: I don't think it's work this much effort and refactoring, as we will be removing support for Python 2 soon. I think we should start with this solution as it's backward-compatible, then change `universal_newlines=True` to `text=True` once we support Python 3 only.

pavel-shirshov (approved):

PR #5935, opened 2020-11-16
[sonic_debian_extension] add docker script to SONiC filesystem

stepanblyschak:
Later this script will be used to start dockerd in chroot environment on SONiC

This PR is part of [SONiC Application Extension](https://github.com/stepanblyschak/SONiC/blob/sonic-app-ext-3/doc/sonic-application-extention/sonic-application-extention-hld.md)

**- Why I did it**
To allow SONiC Package Migration during SONiC-2-SONiC upgrade we need to start docker daemon in chroot-ed environment in new SONiC filesystem.

**- How I did it**
Install a docker service script into /usr/lib/docker/ in SONiC filesystem.

**- How to verify it**
Install SONiC image on the switch, mount squashfs to some directory, mount overlay rw layer over squashfs, mount procfs and sysfs, mount docker library. Start the docker using:
```
root@sonic:~$ /usr/lib/docker/docker.sh start
```

- [ ] 201811
- [ ] 201911
- [ ] 202006


stepanblyschak:
retest broadcom please

stepanblyschak:
@qiluo-msft Could you please review?

jleveque (dismissed):

files/build_templates/sonic_debian_extension.j2
```diff
@@ -512,6 +514,10 @@ sudo LANG=C chroot $FILESYSTEM_ROOT docker $SONIC_NATIVE_DOCKERD_FOR_DOCKERFS ta
 {% endif %}
 {% endfor %}
 
+# Copy docker start script to be able to start docker in chroot
+sudo mkdir -p $FILESYSTEM_ROOT_USR_LIB/docker
+sudo cp $DOCKER_SCRIPTS_DIR/docker $FILESYSTEM_ROOT_USR_LIB/docker/docker.sh
```
qiluo-msft, line 519: I notice it is already copied. Could you use this one?
```
sudo cp files/docker/docker $FILESYSTEM_ROOT/etc/init.d/
```

qiluo-msft (changes requested):
As comment

files/build_templates/sonic_debian_extension.j2
stepanblyschak: It is later removed by:
```
sudo rm $FILESYSTEM_ROOT/etc/init.d/docker
```
And I don't want to put it in /etc/init.d/ as this is not expected to be used as init.d script, docker is a systemd service.

qiluo-msft: Thanks for the insight. That change is introduced by [https://github.com/Azure/sonic-buildimage/pull/2417/files](https://github.com/Azure/sonic-buildimage/pull/2417/files)

The purpose of the that PR and this PR is similar, and I see that PR overwrite the vanilla file and later remove it, so the vanilla `/etc/init.d/docker` file is missing. I suggest unify the 2 effort and I agree `/etc/init.d/docker` may be not a good place.

Your argument "docker is a systemd service" does not relating because you are not using `systemctl` to start service in your use case
```
root@sonic:~$ /usr/lib/docker/docker.sh start
```


---
In reply to: [570580898](https://github.com/Azure/sonic-buildimage/pull/5935#discussion_r570580898) [](ancestors = 570580898)

qiluo-msft:
@lguohan could you also help review this PR?

files/build_templates/sonic_debian_extension.j2
stepanblyschak: Thanks, I see what you mean. I made a change to copy it once and not remove.
My comment was an argument against /etc/init.d/, as later docker debian packages removed this script and putting it back to /etc/init.d/ gives SONiC users wrong idea that user do: "service docker restart" but I doubt that will work correctly, at least won't consider systemd dependencies, so I choose /usr/lib/docker to hide this script from regular user.

qiluo-msft (dismissed):

liat-grozovik:
@jleveque and @qiluo-msft  comments were addressed. can you please review and approve? 
@lguohan  kindly reminder as your feedback was also asked.
I would like to move forward and merge it (as no change effect on existing behaviour)

qiluo-msft (approved):

liat-grozovik:
@jleveque  as your previous approval was dismissed following the feedback changes requested by Qi. Can you please approve so we can move forward and merge?

jleveque (approved):

PR #5938, opened 2020-11-16
[dockers] Tag all docker images with a version number

stepanblyschak:
Previously all docker images built into SONiC binary were tagged with SONiC version string. With the introduction of container upgrades and SONiC extension framework all docker images should have their own version numbers that do not relate with SONiC version. 
All dockers are tagged with version 1.0.0 in this PR. Whenever changes are introduced to a docker image that require a version change, the docker image maintainer should update the version number.

This PR is part of [SONiC Application Extension](https://github.com/stepanblyschak/SONiC/blob/sonic-app-ext-3/doc/sonic-application-extention/sonic-application-extention-hld.md)

**- Why I did it**
SONiC Docker have to follow semantic versioning spec. Thus, Docker images have to be tagged with a semantic version string.

**- How I did it**
For each Docker image defined version variable. This variable is passed to sonic_debian_extension.sh script which loads and tags images.

**- How to verify it**
Build an image and do "show version", all Docker images now are tagged with string "1.0.0".

- [ ] 201811
- [ ] 201911
- [ ] 202006


rajendra-dendukuri (dismissed):

files/build_templates/sonic_debian_extension.j2
```diff
@@ -579,13 +579,15 @@ sudo mkdir $FILESYSTEM_ROOT/target
 sudo mount --bind target $FILESYSTEM_ROOT/target
 sudo chroot $FILESYSTEM_ROOT docker $SONIC_NATIVE_DOCKERD_FOR_DOCKERFS info
 {% for image in installer_images.strip().split(' ') -%}
-{% set imagefilename = image.split('/')|last -%}
+{% set imagefilepath = image.split(':')|first -%}
+{% set imageversion = image.split(':')|last -%}
```
renukamanavalan, line 583: We might need a way to control as remain in current mode (_use image_version_) or adopt new. If not this could break some.

Does this PR allow for a way to stay in current format?

platform/barefoot/docker-syncd-bfn-rpc.mk
```diff
@@ -17,6 +17,7 @@ SONIC_INSTALL_DOCKER_IMAGES += $(DOCKER_SYNCD_BFN_RPC)
 endif
 
 $(DOCKER_SYNCD_BFN_RPC)_CONTAINER_NAME = syncd
+$(DOCKER_SYNCD_BFN_RPC)_VERSION = 1.0.0
```
renukamanavalan, line 20: When multiple users are making updates to syncd code base, who/how/when controls this version?
Even a major change in underlying common libs used could affect version of this docker,

We need a system to enforce before committing this.

files/build_templates/sonic_debian_extension.j2
stepanblyschak: Actually I do not require it to be tagged with a version, a version is in the manifest, so if you're saying it can break someone I return the old tag

platform/barefoot/docker-syncd-bfn-rpc.mk
stepanblyschak: As discussed, currently the version increment is done by developer in PR. The version is incremented in according to semver rules (breaking change, new functionality introduced, bug fix). When the version changes - ideally on every change related to corresponding component, but to be more realistic let's consider the following. Let's put aside master branch, since regular user does not want to run master branch. It only matters for release branches, but in release branches only bug fixes are allowed, so only last digit changes. It is unusual that a package will depend on particular bug fix, it will rather depends on functionality, thus, a version can be changed once on release branch out. From that moment no new functionality or breaking changes are allowed so all app.ext. packages can still work with this release. 
So, it is Ok if we update major, minor version once a release and then don't update patch version. This is one time operation for every package per release so it is realistic to do. So, I don't think that this cannot be committed before we introduce automatic version increment. Besides, if automatic version bump can be done till next release in phase 2 we even won't be required to do it manually for next release.
Besides, in this particular example, since no one can really depend on syncd, because SAI programming is planned for phase 2 (as well as automatic version update), does it really matter for syncd?

files/build_templates/sonic_debian_extension.j2
renukamanavalan: Let us keep the current behavior, as much possible, unless we need to change.

platform/barefoot/docker-syncd-bfn-rpc.mk
renukamanavalan: Wish it is as simple as you suggest. 
Can you please create a write up on your plan for this versioning that includes when/how/who updates the following
1) Base OS version updates
2) Base OSversion compatibility in each docker
3) Individual docker versions

We would need to pass it around and have a dedicated meeting for this?

stepanblyschak: @renukamanavalan Versioning guidelines were added to HLD https://github.com/Azure/SONiC/blob/b154820484fa297e4cb3a010c8f5709e891edc43/doc/sonic-application-extention/sonic-versioning-strategy.md. and updated based on discussion last week.

stepanblyschak:
@bandaru-viswanath Could you please take a look as well?

slave.mk
```diff
@@ -906,7 +906,7 @@ $(addprefix $(TARGET_PATH)/, $(SONIC_INSTALLERS)) : $(TARGET_PATH)/% : \
 	export enable_pfcwd_on_start="$(ENABLE_PFCWD_ON_START)"
 	export installer_debs="$(addprefix $(IMAGE_DISTRO_DEBS_PATH)/,$($*_INSTALLS))"
 	export lazy_installer_debs="$(foreach deb, $($*_LAZY_INSTALLS),$(foreach device, $($(deb)_PLATFORM),$(addprefix $(device)@, $(IMAGE_DISTRO_DEBS_PATH)/$(deb))))"
-	export installer_images="$(addprefix $(TARGET_PATH)/,$($*_DOCKERS))"
+	export installer_images="$(foreach docker, $($*_DOCKERS),$(addprefix $(TARGET_PATH)/,$(addsuffix :$($(docker)_VERSION),$(docker))))"
```
renukamanavalan, line 909: Why do we need this, when version is already added to docker image as label ?

renukamanavalan (commented):
Which PR has the code for labelling dockers with its version and versions of other common modules it used?

Ans: PR #5939 
[Saw that after raising the comments]

files/build_templates/sonic_debian_extension.j2
```diff
@@ -598,9 +598,10 @@ sudo mkdir $FILESYSTEM_ROOT/target
 sudo mount --bind target $FILESYSTEM_ROOT/target
 sudo chroot $FILESYSTEM_ROOT docker $SONIC_NATIVE_DOCKERD_FOR_DOCKERFS info
 {% for image in installer_images.strip().split(' ') -%}
-{% set imagefilename = image.split('/')|last -%}
+{% set imagefilepath = image.split(':')|first -%}
```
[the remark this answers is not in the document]
stepanblyschak: Can be changed back to original. Discussed on community meeting

platform/barefoot/docker-syncd-bfn.mk
```diff
@@ -10,5 +10,8 @@ $(DOCKER_SYNCD_BASE)_DBG_DEPENDS += $(SYNCD_DBG) \
                                $(LIBSAIMETADATA_DBG) \
                                $(LIBSAIREDIS_DBG)
 
+
+$(DOCKER_SYNCD_BASE)_VERSION = 1.0.0
```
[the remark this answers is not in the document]
stepanblyschak: 1. yes
2. recorded in manifest, repo tag may be different
3. yes

stepanblyschak:
This is already part of https://github.com/Azure/sonic-buildimage/pull/5939, thus closing

Issue #5914, opened 2020-11-12
Edge-core as7312-54X SFP tx_power issue


joraff:
**Description**

On accton/edge-core as7312-54X platform, we are unable to get a link with any Q/SFP/+ optical module we have tried. Passive DAC cables are fine. We are able to read the eeprom of the modules.

The i2c nodes for sfp control seem to have changed. Possibly in the CPLD driver? I observe this by trying to use the platform-specific script:

```bash
# accton_as7312_util.py set sfp 1 0
Failed :echo 0 >/sys/bus/i2c/devices/18-0050/sfp_tx_disable

# ls -la /sys/bus/i2c/devices/18-0050/
total 0
drwxr-xr-x 3 root root     0 Feb 16 11:05 .
drwxr-xr-x 6 root root     0 Feb 14 10:12 ..
-rw-r--r-- 1 root root  4096 Feb 14 14:19 dev_class
lrwxrwxrwx 1 root root     0 Feb 14 14:19 driver -> ../../../../../../bus/i2c/drivers/optoe
-rw-r--r-- 1 root root 33152 Feb 14 10:18 eeprom
-r--r--r-- 1 root root  4096 Feb 14 14:19 modalias
-r--r--r-- 1 root root  4096 Feb 14 14:19 name
-rw-r--r-- 1 root root  4096 Feb 14 14:19 port_name
drwxr-xr-x 2 root root     0 Feb 14 14:19 power
lrwxrwxrwx 1 root root     0 Feb 14 10:12 subsystem -> ../../../../../../bus/i2c
-rw-r--r-- 1 root root  4096 Feb 14 10:12 ueven
```

**Steps to reproduce the issue:**
1. Insert optical SFP (any form factor) module into an enabled port
2. Double check port speed and admin enabled status
3. Observe link status

**Describe the results you received:**
No link achieved. TX_disable pin possibly not being disabled

**Describe the results you expected:**
Link achieved, TX_disable status being disabled on plugin, or by using platform-specific control script `accton_as7312_util.py `.

**Output of `show version`:**

```
$ show version

SONiC Software Version: SONiC.master.447-bef76b10
Distribution: Debian 10.6
Kernel: 4.19.0-9-2-amd64
Build commit: bef76b10
Build date: Sat Oct 17 09:50:26 UTC 2020
Built by: johnar@jenkins-worker-8

Platform: x86_64-accton_as7312_54x-r0
HwSKU: Accton-AS7312-54X
[...]

PR #5941, opened 2020-11-17
[device/accton] AS****_54X, validate accton util to set sfp's tx_disable

roylee123:
bug fix: https://github.com/Azure/sonic-buildimage/issues/5914

Validated for tx_disable function of SFP+ on AS7312-54X, AS5812-54X, AS5712-54x, and AS5812-54x.

Signed-off-by: roy_lee <roy_lee@edge-core.com>

**- Why I did it**
For controlling SFP's tx_disable, accton has its own script to do so. 
But it's not working.

**- How I did it**
Fix accton custom script on controlling tx_disable function of SFP ports.
- to correct sysfs names
- map to correct i2c bus.

**- How to verify it**
All tx_disable are asserted in default, which makes link parnter cannot link up.
After disable it, the port is up.   
Some kinds of transceivers can have the red light visible. (if the wavelength of light is 850nm or shorter)

- [ ] 201811
- [ ] 201911
- [ ] 202006

Correct the sysfs path to set tx_disable function. 
BTW, I also refined the code of fan monitor of as7312_54x.


roylee123:
Temporary closed, for there are other models have same issue.
It will reopen after they're all committed.

lguohan (approved):

PR #5954, opened 2020-11-18
[Monit] Enable/disable Monit when starting/stopping the service.

yozhao101:
**- Why I did it**
When we ran the command `sudo config load`, `sudo config reload` or `sudo config load_minigraph`, the containers `swss`, `snmp`, `lldp`, `teamd`, `syncd`, `snmp`, `bgp`, `radv`, `pmon`, `dhcp_relay` and `telemetry` and `restapi` would be stopped and then started. Monit will generate false alerting messages into syslog to indicate some critical processes were not running during restarting these containers. So this PR aims to solve this problem. 

**- How I did it**
We borrowed the concept of `group` statement of Monit. Specifically the Monit configuration entries of each container were grouped together with the help of this `group` statement. When the container was stopped, the command `sudo monit -g <group_name> unmonitor` will stop monitoring the running status of critical processes in this container. When the container was started, the command `sudo monit -g <group_name> monitor` will start monitoring the running status of critical processes. In order to achieve monitoring and unmonitoring automatically, these two commands are placed in the systemd service file of each container.

In the systemd service file of each container, the command `sudo monit -g <group_name> monitor` will be run before the command to start the container. The reason is that if we put the command to start container first and this command failed to start container, then the rest commands would not execute. That means Monit could not generate the alerting message into syslog.

Another thing is I put a special character `-` before the command `-/usr/bin/monit -g <group_name> monitor` and `-/usr/bin/monit -g <group_name> unmonitor` to ignore the failure of these commands. Since we delayed Monit to do the monitoring 5 minutes when the device was up, these commands will fail to read the socket file until Monit created it 5 minutes later. 

Since Monit can only be ran by `root` privilege, I changed the `User=root` in the `[Service]` section of systemd service files.

**- How to verify it**
I tested and verified this method on two lab devices: `str-msn2700-04` and `str-a7050-acs-3`.

- [ ] 201811
- [x ] 201911
- [ ] 202006

### DIFF
--- a/files/build_templates/per_namespace/bgp.service.j2
+++ b/files/build_templates/per_namespace/bgp.service.j2
@@ -10,11 +10,11 @@ StartLimitBurst=3
 
 [Service]
 User=root
-ExecStartPre=/usr/bin/monit -g monit_group_bgp monitor
+ExecStartPre=-/usr/bin/monit -g monit_group_bgp monitor
 ExecStartPre=/usr/local/bin/{{docker_container_name}}.sh start{% if multi_instance == 'true' %} %i{% endif %}
 ExecStart=/usr/local/bin/{{docker_container_name}}.sh wait{% if multi_instance == 'true' %} %i{% endif %}
 ExecStop=/usr/local/bin/{{docker_container_name}}.sh stop{% if multi_instance == 'true' %} %i{% endif %}
-ExecStopPost=/usr/bin/monit -g monit_group_bgp unmonitor
+ExecStopPost=-/usr/bin/monit -g monit_group_bgp unmonitor
 Restart=always
 RestartSec=30
 
--- a/files/build_templates/per_namespace/bgp.service.j2
+++ b/files/build_templates/per_namespace/bgp.service.j2
@@ -9,11 +9,12 @@ StartLimitIntervalSec=1200
 StartLimitBurst=3
 
 [Service]
-User={{ sonicadmin_user }}
+User=root
+ExecStartPre=/usr/bin/monit -g monit_group_bgp monitor
 ExecStartPre=/usr/local/bin/{{docker_container_name}}.sh start{% if multi_instance == 'true' %} %i{% endif %}
 ExecStart=/usr/local/bin/{{docker_container_name}}.sh wait{% if multi_instance == 'true' %} %i{% endif %}
 ExecStop=/usr/local/bin/{{docker_container_name}}.sh stop{% if multi_instance == 'true' %} %i{% endif %}
-
+ExecStopPost=/usr/bin/monit -g monit_group_bgp unmonitor
 Restart=always
 RestartSec=30
 
--- a/files/build_templates/per_namespace/bgp.service.j2
+++ b/files/build_templates/per_namespace/bgp.service.j2
@@ -13,8 +13,8 @@ User=root
 ExecStartPre=-/usr/bin/monit -g monit_group_bgp monitor
 ExecStartPre=/usr/local/bin/{{docker_container_name}}.sh start{% if multi_instance == 'true' %} %i{% endif %}
 ExecStart=/usr/local/bin/{{docker_container_name}}.sh wait{% if multi_instance == 'true' %} %i{% endif %}
+ExecStop=-/usr/bin/monit -g monit_group_bgp unmonitor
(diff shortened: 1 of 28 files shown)


jleveque:
The PR title is a bit misleading and not direct. While this is intended to unmonitor during `config reload` type operations, the actual change is to control Monit for processes in a container based on the corresponding service status. I think you can rephrase this into a concise title.

Also, I think this should be applied to all services which spawn containers. It appears you didn't apply this to a few, including database and sflow. Is there a reason why not?

yozhao101:
> The PR title is a bit misleading and not direct. While this is intended to unmonitor during `config reload` type operations, the actual change is to control Monit for processes in a container based on the corresponding service status. I think you can rephrase this into a concise title.
> 
> Also, I think this should be applied to all services which spawn containers. It appears you didn't apply this to a few, including database and sflow. Is there a reason why not?

Yeah, I think you asked me a very good question. Initially I was wondering this was a general issue not specifically for the commands `sudo config load/reload/load_minigraph`. I will add the same change for containers `acms`, `sflow` and `database`.

files/build_templates/per_namespace/bgp.service.j2
```diff
@@ -9,11 +9,12 @@ StartLimitIntervalSec=1200
 StartLimitBurst=3
 
 [Service]
-User={{ sonicadmin_user }}
```
yozhao101, line 12: @jleveque @lguohan @yxieca For this directive `User=`, I changed it to `User=root` since Monit configuration file is only accessible by root. Do you have any concerns for this change?

yxieca: Please add the reason to the PR comments.

yozhao101: Updated.

lguohan: better to make monit to work with sonicadmin_user, remove that restriction of monit.

yozhao101: Will update.

yozhao101: @lguohan @jleveque @yxieca I did the investigation to let Monit run by `sonicadmin_user`. 

The result is we can remove the restriction to let other users run Monit. However it will involve several modifications and I am not sure such change will introduce other issues or not:

1.We need change ownership of Monit control file `monitrc` from `root` to `admin`.  The reason is that Monit will check whether permission of Monit control file was `0700` or less `0700` when Monit was started to run. If the permission of control file was not, Monit will fail to start.

2.Once the ownership of Monit control file was changed from `root` to `admin`, we have to modify the init script of Monit `/etc/init.d/monit`. The reason is Monit will check whether the euid (effective uid ) of Monit process is euqal to uid of Monit control file. If we change the ownership of Monit control file from `root` to `admin`. The uid of this file will be `1000`, but the euid of Monit process is `0`. `0` represents the uid of root user.  why? The reason is that Monit process was started by `start-stop-daemon` in the init script of Monit `/etc/init.d/monitand`, it will inherit the uid of `start-stop-daemon`.  

In order to change the euid of Monit process to `1000`, we can add option `--chuid admin:admin` in the `start-stop-daemon` command to solve this issue.

3.We need change the location of Monit `pidfile`. Initially this file will be created in the directory `/run/monit.pid` .  After uid of Monit was changed from `0` to `1000`, it can not access `/run`.

3.We also need change the location of httpd socket file in Monit control file `monitrc`. Initially the file was created in the directory `/var/run/monit.sock` once Monit was started. After we change euid of Monit process from 0 to 1000, then Monit can not access the directory `/var/run`.

yxieca: Does this link help? https://serverfault.com/questions/50628/how-i-configure-monit-to-start-a-process-with-a-specific-user

yozhao101: Thanks, Ying! This link introduced a method how should we start a process not as `root` in Monit configuration file. Currently what we want to do is to run Monit not as `root`.

yxieca:
@yozhao101 do you still intend to check in this PR? If so, can you solve t he merge conflict?

yozhao101:
This 

> @yozhao101 do you still intend to check in this PR? If so, can you solve t he merge conflict?

This PR should be closed.

PR #5963, opened 2020-11-18
[yang] update yang model, add autoneg to sonic-port

dmytroxshevchuk:
**- Why I did it**
Dynamic Port Breakout fall in case "autoneg" field exist in config_db.

**- How I did it**
Added "autoneg" field in sonic-port yang model.

**- How to verify it**
Add "autoneg" field into config_db like this:
```
"Ethernet8": {
    "index": "2", 
    "lanes": "8,9,10,11", 
    "fec": "rs", 
    "pfc_asym": "off", 
    "mtu": "9100", 
    "alias": "Ethernet8", 
    "admin_status": "up", 
    "autoneg": "0", 
    "speed": "100000",
},
```
Run `sudo config interface breakout`

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012


dmytroxshevchuk:
@praveen-li please check

src/sonic-yang-models/yang-models/sonic-port.yang
```diff
@@ -102,6 +102,12 @@ module sonic-port{
 						pattern "on|off";
 					}
 				}
+
```
praveen-li, line 105: Kindly copy the relevant config example, or the link to config here, thx.

dmytroxshevchuk: @praveen-li, for example from `config_db.json`

```
"Ethernet8": {
    "index": "2", 
    "lanes": "8,9,10,11", 
    "fec": "rs", 
    "pfc_asym": "off", 
    "mtu": "9100", 
    "alias": "Ethernet8", 
    "admin_status": "up", 
    "autoneg": "0", 
    "speed": "100000",
},
```

praveen-li (approved):
LGTM, You shd add this in tests too. 
https://github.com/Azure/sonic-buildimage/tree/master/src/sonic-yang-models/tests/yang_model_tests

dmytroxshevchuk:
> LGTM, You shd add this in tests too.
> https://github.com/Azure/sonic-buildimage/tree/master/src/sonic-yang-models/tests/yang_model_tests

I tried to add tests for autoneg but I see a lot of exceptions in tests:
First was with try to read data by key from `self.defaultYANGFailure`
https://github.com/Azure/sonic-buildimage/blob/962e838b16a1ee23586d24d33c82fb6d15c982a4/src/sonic-yang-models/tests/yang_model_tests/test_yang_model.py#L183-L184
but we can see initialization of `self.defaultYANGFailure`:
https://github.com/Azure/sonic-buildimage/blob/962e838b16a1ee23586d24d33c82fb6d15c982a4/src/sonic-yang-models/tests/yang_model_tests/test_yang_model.py#L40-L49
There are no such key so we have exception.

I tried to workaround and change key but after that I see next exception:
`Exception >module 'yang' has no attribute 'Context'< in /sonic-yang-models/tests/yang_model_tests/test_yang_model.py:246`

I can try to fix it too but I think its not a true way so.. could you please provide any information about this test. Is the test work? How run it?

praveen-li:
@dmytroxshevchuk 

I fixed this in PR https://github.com/Azure/sonic-buildimage/pull/6029
Yes, an exception was going through unnoticed after the last PR 4442 merge.
Let's wait for the results on PR 6029 .

lguohan:
PR 6029 is merged.

dmytroxshevchuk:
@praveen-li added autoneg to PORT_TEST, please review
Let me know if need add this parameter in other places

praveen-li:
PR #6029 is merged, please add tests now.

dmytroxshevchuk:
@praveen-li done, please check

src/sonic-yang-models/tests/yang_model_tests/test_yang_model.py
```diff
@@ -177,6 +177,18 @@ def initTest(self):
...
+                'desc': 'PORT_VALID_AUTONEG_TEST no failure.',
+                'eStr': self.defaultYANGFailure['None']
+            },
+            'PORT_INVALID_AUTONEG_TEST_1': {
+                'desc': 'PORT_INVALID_AUTONEG_TEST_1 must condition failure.',
+                'eStr': self.defaultYANGFailure['InvalidValue']
+            },
+            'PORT_INVALID_AUTONEG_TEST_2': {
+                'desc': 'PORT_INVALID_AUTONEG_TEST_2 must condition failure.',
+                'eStr': self.defaultYANGFailure['Pattern']
```
praveen-li, line 190: Does this one fall under pattern or InvalidValue?

dmytroxshevchuk: I think in this case must be 'InvalidValue' but I have test failure.
Is it problem with tests?

INFO:YANG-TEST:
------------------- Test 12: PORT_INVALID_AUTONEG_TEST_2 must condition failure.---------------------
libyang[0]: Value "260" does not satisfy the constraint "0..256" (range, length, or pattern). (path: /sonic-port:sonic-port/PORT/PORT_LIST[port_name='Ethernet8']/autoneg)
INFO:YANG-TEST:Value "260" does not satisfy the constraint "0..256" (range, length, or pattern).
ERROR:YANG-TEST: Exception >Mismatch ['Invalid value'] and Value "260" does not satisfy the constraint "0..256" (range, length, or pattern).< in /sonic/src/sonic-yang-models/tests/yang_model_tests/test_yang_model.py:370
INFO:YANG-TEST:PORT_INVALID_AUTONEG_TEST_2 must condition failure. Failed

praveen-li: I guess, for Range, libyang throws a different error, We can add a new error type: 

```
self.defaultYANGFailure = {
            'Must': ['Must condition', 'not satisfied'],
            'InvalidValue': ['Invalid value'],
            'LeafRef': ['Leafref', 'non-existing'],
            'When': ['When condition', 'not satisfied'],
            'Pattern': ['pattern', 'does not satisfy'],
            'Mandatory': ['required element', 'Missing'],
            'Verify': ['verified'],
            'Range': ['does not satisfy', 'range'],
            'None': []
        }
```

And in the test case:
```
'eStr': self.defaultYANGFailure['Range'] + ['0..256']
```

dmytroxshevchuk: Thanks for the hint. Added 'Range', please check.

src/sonic-yang-models/tests/yang_model_tests/yangTest.json
```diff
@@ -642,6 +642,63 @@
...
+					"port_name": "Ethernet8",
+					"alias": "eth8",
+					"lanes": "65",
+					"description": "Ethernet8",
+					"speed": 25000,
+					"fec": "rs",
+					"mtu": 9000,
+					"admin_status": "up",
+					"pfc_asym": "off",
+					"autoneg": -1
```
praveen-li, line 677: Maybe this is an invalid value because the type is unsigned int. And for +ve range is checked.

src/sonic-yang-models/tests/yang_model_tests/yangTest.json
```diff
@@ -642,6 +642,63 @@
 		}
 	},
 
+	"PORT_VALID_AUTONEG_TEST": {
+		"sonic-port:sonic-port": {
+			"sonic-port:PORT": {
+				"PORT_LIST": [{
+					"port_name": "Ethernet8",
```
praveen-li, line 649: We use ijson to read this file, but still for readability, file should not be very long. Kindly remove description, admin_status, MTU, fec and pfc_asym from config of all new tests. They are already tested with previous tests.

praveen-li (dismissed):
LGTM, after removing extra fields from the config of new tests.

praveen-li (approved):
LGTM now. Thx.

dmytroxshevchuk:
@lguohan , please merge

src/sonic-yang-models/yang-models/sonic-port.yang
venkatmahalingam: Please update the revision, also, add a description for the new field.

src/sonic-yang-models/yang-models/sonic-port.yang
```diff
@@ -94,6 +94,12 @@ module sonic-port{
...
+				leaf autoneg {
+					type uint16 {
+						range 0..256;
```
lguohan, line 100: why is autoneg is a bool? on/off?

praveen-li: Question is: why autoneg is not the bool? also if you have a pointer towards the design document of this feature, which explains the limit of 256, that will help. Thx.

dmytroxshevchuk: @lguohan @praveen-li 
Yang model depends on config_db, so if in config_db we have integer, than it must be in yang model in another case we will have error and crush.
We can see in portsorch code that string value from config_db convert to integer:
https://github.com/Azure/sonic-swss/blob/189a9641cf47e81c7c90d2501e70b73b74b082e1/orchagent/portsorch.cpp#L2215-L2219
```
    /* Set autoneg and ignore the port speed setting */
    if (fvField(i) == "autoneg")
    {
        an = (int)stoul(fvValue(i));
    }
```

In next step, we can see how this integer value converts to bool type and use as sai attribute:
https://github.com/Azure/sonic-swss/blob/189a9641cf47e81c7c90d2501e70b73b74b082e1/orchagent/portsorch.cpp#L1752-L1767
```
bool PortsOrch::setPortAutoNeg(sai_object_id_t id, int an)
{
    SWSS_LOG_ENTER();

    sai_attribute_t attr;
    attr.id = SAI_PORT_ATTR_AUTO_NEG_MODE;
    switch(an) {
      case 1:
        attr.value.booldata = true;
        break;
      default:
        attr.value.booldata = false;
        break;
    }

    sai_status_t status = sai_port_api->set_port_attribute(id, &attr);
```


So in config_db we need use integer for valid conversion to int and after that int to bool. So in yang models we need use int value too.

@praveen-li , I agree with you that we no need use range 0..255. We must use only 0 or 1. Thanks for the correction.

anshuv-mfst:
Hi @dmytroxshevchuk - Could you please join YANG subgroup meeting 2/25 (10-11am pst) to review with the YANG community, thanks.

src/sonic-yang-models/yang-models/sonic-port.yang
```diff
@@ -94,6 +94,12 @@ module sonic-port{
 						pattern "on|off";
 					}
 				}
+
+				leaf autoneg {
+					type uint8 {
+						range 0..1;
```
lguohan, line 100: can you change this to on|off?

venkatmahalingam: how about the backward compatibility? we use 0 and 1 in the portsorch.cpp (swss container) today, after making on/off changes, hope 0 and 1 will also continue to work for autoneg, @dmytroxshevchuk please confirm.

praveen-li: As per the below code, we need to change the backend first.
```
  /* Set autoneg and ignore the port speed setting */
                else if (fvField(i) == "autoneg")
                {
                    an = (int)stoul(fvValue(i));
                }
```

dmytroxshevchuk:
@lguohan @praveen-li @venkatmahalingam please review.
A tried add commit from swss to this PR but looks like LGTM fall with error:
```
ERROR: Unable to update submodule: src/sonic-swss
(eventual cause: UserError "Commit does not exist: 33feea40fad80e5c98a46a64040880dcf3e1e1a3")
```
So I created another PR in swss:
https://github.com/Azure/sonic-swss/pull/1658
with changes in backend.
Keep in mind that we need changes from this PR and PR from swss at one time otherwise will be problems.

praveen-li:
What is the number of previous PR if it is merged already you should use the commit id of that PR to update the submodule (i.e. One more PR just to update submodule).
 and then the tests of current PR will pass I guess.



Thanks,

Regards
Praveen

dmytroxshevchuk:
@praveen-li I can update swss submodule to master from this PR, but my own changes not exist in swss submodule and its cause of error in lgtm. I can`t add my own commit into swss submodule from this PR. We need merge PR into swss submodule and after that I can update submodule from this PR, only after that tests will pass. 
https://github.com/Azure/sonic-swss/pull/1658

lguohan (approved):

praveen-li (dismissed):

lguohan (approved):

Issue #5965, opened 2020-11-18
FRR generates log /var/log/frr/frr.log inside bgp docker and taking lots of harddrive space


yxieca:
**Description**

**Steps to reproduce the issue:**
1. Run nightly test
2.
3.

**Describe the results you received:**
Unable to upgrade to next image because BGP docker took too much hard drive space

**Describe the results you expected:**
Upgrade for nightly test goes smooth.

**Additional information you deem important (e.g. issue happens only occasionally):**

    **Output of `show version`:**

    ```
SONiC Software Version: SONiC.20191130.52
    ```

    ```
root@str-a7060cx-acs-10:/home/admin# docker exec -it bgp bash
root@str-a7060cx-acs-10:/# ls
bin  boot  dev  etc  home  lib  lib64  media  mnt  opt  proc  root  run  sbin  srv  sys  tmp  usr  var
root@str-a7060cx-acs-10:/# cd var/log/
root@str-a7060cx-acs-10:/var/log# ls
alternatives.log  apt  btmp  dpkg.log  faillog  frr  lastlog  redis  supervisor  wtmp
root@str-a7060cx-acs-10:/var/log# cd frr         
root@str-a7060cx-acs-10:/var/log/frr# ls
frr.log
root@str-a7060cx-acs-10:/var/log/frr# du -sh .
2.1G    .
    ```

PR #5966, opened 2020-11-18
[frr] mount /var/log/frr onto frr docker and configure logrotate

yxieca:
**- Why I did it**
frr generates log at /var/log/frr/frr.log inside SONiC BGP docker. This log takes too much space. On a small hard drive platform, the log file took all space on harddrive and causing SONiC upgrade to fail.

to fix issue #5965 

**- How I did it**
Mount /var/log/frr path from base image to BGP docker and setup log rotate from the base image.

**- How to verify it**
To be updated

- [ ] 201811
- [x] 201911
- [ ] 202006

### DIFF
--- a/files/image_config/logrotate/logrotate.d/rsyslog
+++ b/files/image_config/logrotate/logrotate.d/rsyslog
@@ -30,6 +30,7 @@
 /var/log/telemetry.log
 /var/log/quagga/bgpd.log
 /var/log/quagga/zebra.log
+/var/log/frr/frr.log
 /var/log/swss/sairedis.rec
 /var/log/swss/swss.rec
 {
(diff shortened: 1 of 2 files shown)


judyjoseph:
@ying the below change will make it work for multi-asic to send logs to /var/log/quagga/*log files on the host.  Since this change is not critical, I can do this as a different PR as well ( need to changes the filenames from quagga --> frr also )

```
index 455fe89f..52c33791 100644
--- a/files/image_config/rsyslog/rsyslog.d/00-sonic.conf
+++ b/files/image_config/rsyslog/rsyslog.d/00-sonic.conf
@@ -1,14 +1,11 @@
 ## Quagga rules
 
-if $programname == ["bgp#quagga",
-                    "bgp#watchquagga",
-                    "bgp#zebra"]
-    then {
+if re_match($programname, "bgp[0-9]*#(quagga|watchquagga|zebra)") then {
     /var/log/quagga/zebra.log
     stop
 }
 
-if $programname == "bgp#bgpd" then {
+if re_match($programname, "bgp[0-9]*#bgpd") then {
     /var/log/quagga/bgpd.log
     stop
 }


```

yxieca:
Unfortunately, this change will cause issue for multi-ASIC BGP logs, where all BGP dockers will write to the same /var/log/frr/frr.log, and in turn, give bgpmon a hard time in each docker. Closing this PR. The issue is still outstanding and will be fixed with another proper change.

PR #5981, opened 2020-11-19
[L2 switch mode] Update l2switch.j2 template

shi-su:
**- Why I did it**
The l2switch.j2 template does not include all fields for PORT. This could be incompatible with the 201911 image or later. 

**- How I did it**
Update l2switch.j2 template and add a unit test.

**- How to verify it**

- [ ] 201811
- [ ] 201911
- [ ] 202006

### DIFF
--- a/src/sonic-config-engine/data/l2switch.j2
+++ b/src/sonic-config-engine/data/l2switch.j2
@@ -2,16 +2,16 @@
     "DEVICE_METADATA": {{ DEVICE_METADATA | tojson }},
     {% set ns = {'firstPrinted': False} -%}
     "PORT": {
-        {%- for key,value in PORT.items() -%}
+        {%- for key, value in PORT.items() -%}
         {%- if ns.firstPrinted %},{% endif %}
 
         "{{ key }}": {
-            {%- for keyPort,valuePort in value.items() %}
+            {%- for keyPort, valuePort in value.items() %}
 
-            "{{ keyPort }}": "{{ valuePort }}",
+            {% if keyPort != "admin_status" %}"{{ keyPort }}": "{{ valuePort }}",{% endif %}
             {%- endfor %}
 
-            "admin_status": "up"
+            "admin_status": "{{ value.admin_status|default("up") }}"
         }
         {%- if ns.update({'firstPrinted': True}) %}{% endif -%}
         {%- endfor %}
--- a/src/sonic-config-engine/data/l2switch.j2
+++ b/src/sonic-config-engine/data/l2switch.j2
@@ -6,11 +6,11 @@
         {%- if ns.firstPrinted %},{% endif %}
 
         "{{ key }}": {
-            {%- for k,v in value.items() %}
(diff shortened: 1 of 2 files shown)


src/sonic-config-engine/data/l2switch.j2
```diff
@@ -6,8 +6,11 @@
         {%- if ns.firstPrinted %},{% endif %}
 
         "{{ key }}": {
-            "alias": "{{ value.alias }}",
-            "lanes": "{{ value.lanes }}",
+            {%- for keyPort,valuePort in value.items() %}
```
qiluo-msft, line 9: >keyPort [](start = 20, length = 7)

If keyPort == 'admin_status', is it a problem?

qiluo-msft, line 9: >, [](start = 27, length = 1)

Add a blank after `,`. Also some more occurrences in the whole file.

qiluo-msft:
>         argument = '-k Mellanox-SN2700 --preset l2 -p ' + self.t0_port_config

Could you add another test case for Mellanox-SN3800-D112C8 in the same way?

---
Refers to: src/sonic-config-engine/tests/test_j2files.py:99 in dd1f4d9. [](commit_id = dd1f4d96b213668c166d05a8ae3e1652dd0cb4f2, deletion_comment = False)

qiluo-msft (changes requested):
As comments

src/sonic-config-engine/data/l2switch.j2
shi-su: Added check to resolve the scenario with `admin_status` exists.

shi-su: Fixed

shi-su:
> > ```
> >     argument = '-k Mellanox-SN2700 --preset l2 -p ' + self.t0_port_config
> > ```
> 
> Could you add another test case for Mellanox-SN3800-D112C8 in the same way?
> 
> Refers to: src/sonic-config-engine/tests/test_j2files.py:99 in [dd1f4d9](https://github.com/Azure/sonic-buildimage/commit/dd1f4d96b213668c166d05a8ae3e1652dd0cb4f2). [](commit_id = [dd1f4d9](https://github.com/Azure/sonic-buildimage/commit/dd1f4d96b213668c166d05a8ae3e1652dd0cb4f2), deletion_comment = False)

I think I understand your point: different device types have different sources of port configuration data, and it would be better to capture the difference. However, in this particular test, the data source is explicitly specified (tests/t0-sample-port-config.ini). So only changing the device type cannot test out the sources of port configuration data, and in this test setup, it does not seem the source of port configuration data could follow the `-k <device>` option. In using the command in SONiC, the `-k <device>` option should specify the data, and rendering the template should not be dependent on what kind of device it is.

qiluo-msft (approved):

shi-su:
retest vsimage please

PR #5990, opened 2020-11-20
Fix bind-mount of /var/log/swss directory in multi-asic setup

mlorrillere:
This change is to make sure that each asic in multi-asic devices has its
/var/log/swss directory mounted at a different mount point on the host.

**- Why I did it**
On single-asic setups, /var/log/swss directory of the swss container is bind-mounted onto the host at /var/log/swss to get easy access to swss logs. On multi-asic setups, the current model doesn't work because the host's directory becomes shared across all swss containers, thus swss logs of one container gets overwritten by the logs of other containers.

**- How I did it**
/var/log/swss of each swss container is now mounted at /var/log/swss$DEV on the host.

**- How to verify it**
Verified manually on a multi-asic device.

### DIFF
--- a/files/build_templates/docker_image_ctl.j2
+++ b/files/build_templates/docker_image_ctl.j2
@@ -387,6 +387,9 @@ start() {
         -v /var/run/redis-chassis:/var/run/redis-chassis:ro \
         -v /usr/share/sonic/device/$PLATFORM/$HWSKU/$DEV:/usr/share/sonic/hwsku:ro \
 {%- endif %}
+{%- if docker_container_name == "swss" %}
+        -v /var/log/swss$DEV:/var/log/swss:rw \
+{%- endif %}
         $REDIS_MNT \
         -v /usr/share/sonic/device/$PLATFORM:/usr/share/sonic/platform:ro \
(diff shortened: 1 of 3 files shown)


files/build_templates/docker_image_ctl.j2
```diff
@@ -364,6 +364,9 @@ start() {
         -v /var/run/redis$DEV:/var/run/redis:rw \
         -v /var/run/redis-chassis:/var/run/redis-chassis:ro \
         -v /usr/share/sonic/device/$PLATFORM/$HWSKU/$DEV:/usr/share/sonic/hwsku:ro \
+{%- endif %}
+{%- if docker_container_name == "swss" %}
+        -v /var/log/swss$DEV:/var/log/swss:rw \
```
lguohan, line 369: this changes the folder even for single asic where we are trying to make single asic exactly the same as if there is no multi-asic setup.

mlorrillere: The path doesn't change for single instance containers because `$DEV` is unset in this case.

lguohan:
obseleted by #6310

PR #5991, opened 2020-11-20
VoQ configuration using minigraph.xml file

mlorrillere:
Contains the following changes:
- Add support for system ports configuration to minigraph
- Add support for SwitchId, SwitchType and MaxCores to minigraph
- Add support for inband vlan configuration in minigraph
- Make `asic_name` in `CONFIG_DB` mandatory for VoQ switches
- Added `sample-voq-graph.xml` and updated `test_cfggen.py` to test for VoQ switches config attributes

**- Why I did it**

Adding support for configuring a VoQ switch using a minigraph.xml file.

**- How I did it**

Added the following new sections to the minigraph.xml file:

- `VoqInbandInterfaces` to `DeviceDataPlaneInfo`
   Contains a list of `VoQInbandInterface` for this device. Example:
```xml
        <a:VoqInbandInterface>
          <Name>Vlan3094</Name>
          <Type>vlan</Type>
          <a:PrefixStr>1.1.1.1/24</a:PrefixStr>
        </a:VoqInbandInterface>
```
- `SystemPorts` in `DeviceInfo`
   Contains the list of all system ports present in the system as `SystemPort` elements. Example:
```xml
        <SystemPort>
          <Name>Ethernet1/6</Name>
          <Hostname>linecard-2</Hostname>
          <AsicName>ASIC0</AsicName>
          <Speed>40000</Speed>
          <SystemPortId>258</SystemPortId>
          <SwitchId>2</SwitchId>
          <CoreId>1</CoreId>
          <CorePortId>2</CorePortId>
          <NumVoq>8</NumVoq>
        </SystemPort>
```

- `SwitchId` to `DeviceMetadata`
   This is used to set the switch ID of a device. Example:
```xml
      <a:DeviceMetadata>
        <a:Name>ASIC0</a:Name>
        <a:Properties>
          <a:DeviceProperty>
            <a:Name>SwitchId</a:Name>
            <a:Reference i:nil="true"/>
            <a:Value>2</a:Value>
          </a:DeviceProperty>
        </a:Properties>
      </a:DeviceMetadata>
```

**- How to verify it**
- Updated tests in sonic-config-engine
- Manually tested on single and multi-asic VoQ switches

### DIFF
--- a/src/sonic-config-engine/minigraph.py
+++ b/src/sonic-config-engine/minigraph.py
@@ -456,6 +456,17 @@ def parse_dpg(dpg, hname):
             gwaddr = ipaddress.ip_address(next(mgmtipn.hosts()))
             mgmt_intf[(intfname, ipprefix)] = {'gwaddr': gwaddr}
 
+        voqinbandintfs = child.find(str(QName(ns, "VoqInbandInterfaces")))
+        voq_inband_intfs = {}
+        if voqinbandintfs:
+            for voqintf in voqinbandintfs.findall(str(QName(ns1, "VoqInbandInterface"))):
+                intfname = voqintf.find(str(QName(ns, "Name"))).text
+                intftype = voqintf.find(str(QName(ns, "Type"))).text
+                ipprefix = voqintf.find(str(QName(ns1, "PrefixStr"))).text
+                if intfname not in voq_inband_intfs:
+                   voq_inband_intfs[intfname] = {'inband_type': intftype}
+                voq_inband_intfs["%s|%s" % (intfname, ipprefix)] = {}
+
         pcintfs = child.find(str(QName(ns, "PortChannelInterfaces")))
         pc_intfs = []
         pcs = {}
@@ -667,8 +678,8 @@ def parse_dpg(dpg, hname):
                     if mg_key in mg_tunnel.attrib:
                         tunnelintfs[tunnel_type][tunnel_name][table_key] = mg_tunnel.attrib[mg_key]
 
-        return intfs, lo_intfs, mvrf, mgmt_intf, vlans, vlan_members, pcs, pc_members, acls, vni, tunnelintfs, dpg_ecmp_content
-    return None, None, None, None, None, None, None, None, None, None
+        return intfs, lo_intfs, mvrf, mgmt_intf, voq_inband_intfs, vlans, vlan_members, pcs, pc_members, acls, vni, tunnelintfs, dpg_ecmp_content
+    return None, None, None, None, None, None, None, None, None, None, None, None, None
 
 def parse_host_loopback(dpg, hname):
     for child in dpg:
@@ -794,6 +805,9 @@ def parse_meta(meta, hname):
     cloudtype = None
     resource_type = None
     downstream_subrole = None
+    switch_id = None
+    switch_type = None
+    max_cores = None
     kube_data = {}
     device_metas = meta.find(str(QName(ns, "Devices")))
     for device in device_metas.findall(str(QName(ns1, "DeviceMetadata"))):
@@ -825,11 +839,17 @@ def parse_meta(meta, hname):
                     resource_type = value
                 elif name == "DownStreamSubRole":
                     downstream_subrole = value
+                elif name == "SwitchId":
+                    switch_id = value
+                elif name == "SwitchType":
+                    switch_type = value
+                elif name == "MaxCores":
+                    max_cores = value
                 elif name == "KubernetesEnabled":
                     kube_data["enable"] = value
                 elif name == "KubernetesServerIp":
                     kube_data["ip"] = value
-    return syslog_servers, dhcp_servers, ntp_servers, tacacs_servers, mgmt_routes, erspan_dst, deployment_id, region, cloudtype, resource_type, downstream_subrole, kube_data
+    return syslog_servers, dhcp_servers, ntp_servers, tacacs_servers, mgmt_routes, erspan_dst, deployment_id, region, cloudtype, resource_type, downstream_subrole, switch_id, switch_type, max_cores, kube_data
 
 
 def parse_linkmeta(meta, hname):
@@ -886,6 +906,9 @@ def parse_linkmeta(meta, hname):
 
 def parse_asic_meta(meta, hname):
     sub_role = None
+    switch_id = None
+    switch_type = None
+    max_cores = None
     device_metas = meta.find(str(QName(ns, "Devices")))
     for device in device_metas.findall(str(QName(ns1, "DeviceMetadata"))):
         if device.find(str(QName(ns1, "Name"))).text.lower() == hname.lower():
@@ -895,7 +918,13 @@ def parse_asic_meta(meta, hname):
                 value = device_property.find(str(QName(ns1, "Value"))).text
                 if name == "SubRole":
                     sub_role = value
-    return sub_role
+                elif name == "SwitchId":
+                    switch_id = value
+                elif name == "SwitchType":
+                    switch_type = value
+                elif name == "MaxCores":
+                    max_cores = value
+    return sub_role, switch_id, switch_type, max_cores
 
 def parse_deviceinfo(meta, hwsku):
     port_speeds = {}
@@ -912,7 +941,28 @@ def parse_deviceinfo(meta, hwsku):
                 if desc != None:
                     port_descriptions[port_alias_map.get(alias, alias)] = desc.text
                 port_speeds[port_alias_map.get(alias, alias)] = speed
-    return port_speeds, port_descriptions
+
+            sysports = device_info.find(str(QName(ns, "SystemPorts")))
+            sys_ports = {}
+            if sysports is not None:
+                for sysport in sysports.findall(str(QName(ns, "SystemPort"))):
+                    portname = sysport.find(str(QName(ns, "Name"))).text
+                    hostname = sysport.find(str(QName(ns, "Hostname")))
+                    asic_name = sysport.find(str(QName(ns, "AsicName")))
+                    system_port_id = sysport.find(str(QName(ns, "SystemPortId"))).text
+                    switch_id = sysport.find(str(QName(ns, "SwitchId"))).text
+                    core_id = sysport.find(str(QName(ns, "CoreId"))).text
+                    core_port_id = sysport.find(str(QName(ns, "CorePortId"))).text
+                    speed = sysport.find(str(QName(ns, "Speed"))).text
+                    num_voq = sysport.find(str(QName(ns, "NumVoq"))).text
+                    key = portname
+                    if asic_name is not None:
+                       key = "%s|%s" % (asic_name.text, key)
+                    if hostname is not None:
+                       key = "%s|%s" % (hostname.text, key)
+                    sys_ports[key] = {"system_port_id": system_port_id, "switch_id": switch_id, "core_index": core_id, "core_port_index": core_port_id, "speed": speed, "num_voq": num_voq}
+
+    return port_speeds, port_descriptions, sys_ports
 
 # Function to check if IP address is present in the key. 
 # If it is present, then the key would be a tuple.
@@ -1097,6 +1147,7 @@ def parse_xml(filename, platform=None, port_config_file=None, asic_name=None, hw
     vlan_members = None
     pcs = None
     mgmt_intf = None
+    voq_inband_intfs = None
     lo_intfs = None
     neighbors = None
     devices = None
@@ -1107,6 +1158,7 @@ def parse_xml(filename, platform=None, port_config_file=None, asic_name=None, hw
     port_speeds_default = {}
     port_speed_png = {}
     port_descriptions = {}
+    sys_ports = {}
     console_ports = {}
     mux_cable_ports = {}
     syslog_servers = []
@@ -1119,6 +1171,9 @@ def parse_xml(filename, platform=None, port_config_file=None, asic_name=None, hw
     deployment_id = None
     region = None
     cloudtype = None
+    switch_id = None
+    switch_type = None
+    max_cores = None
     hostname = None
     linkmetas = {}
     host_lo_intfs = None
@@ -1153,7 +1208,7 @@ def parse_xml(filename, platform=None, port_config_file=None, asic_name=None, hw
     for child in root:
         if asic_name is None:
             if child.tag == str(QName(ns, "DpgDec")):
-                (intfs, lo_intfs, mvrf, mgmt_intf, vlans, vlan_members, pcs, pc_members, acls, vni, tunnel_intfs, dpg_ecmp_content) = parse_dpg(child, hostname)
+                (intfs, lo_intfs, mvrf, mgmt_intf, voq_inband_intfs, vlans, vlan_members, pcs, pc_members, acls, vni, tunnel_intfs, dpg_ecmp_content) = parse_dpg(child, hostname)
             elif child.tag == str(QName(ns, "CpgDec")):
                 (bgp_sessions, bgp_internal_sessions, bgp_voq_chassis_sessions, bgp_asn, bgp_peers_with_range, bgp_monitors) = parse_cpg(child, hostname)
             elif child.tag == str(QName(ns, "PngDec")):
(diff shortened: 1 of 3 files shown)


src/sonic-config-engine/minigraph.py
```diff
@@ -698,7 +715,11 @@ def parse_asic_meta(meta, hname):
                 value = device_property.find(str(QName(ns1, "Value"))).text
                 if name == "SubRole":
                     sub_role = value
-    return sub_role
+                elif name == "SwitchId":
+                    switch_id = value
+                elif name == "InstanceName":
+                    instance_name = value
+    return sub_role, switch_id, instance_name
```
vganesan-nokia, line 722: We also need **switch_type** and **max_cores** attributes in device metadata (Ref: https://github.com/Azure/SONiC/blob/master/doc/voq/voq_hld.md#231-device_metadata)

src/sonic-config-engine/minigraph.py
```diff
@@ -715,7 +736,26 @@ def parse_deviceinfo(meta, hwsku):
...
+                    device = sysport.find(str(QName(ns, "Device"))).text
+                    system_port_id = sysport.find(str(QName(ns, "SystemPortId"))).text
+                    switch_id = sysport.find(str(QName(ns, "SwitchId"))).text
+                    core_id = sysport.find(str(QName(ns, "CoreId"))).text
+                    core_port_id = sysport.find(str(QName(ns, "CorePortId"))).text
+                    speed = sysport.find(str(QName(ns, "Speed"))).text
+                    num_voq = sysport.find(str(QName(ns, "NumVoq"))).text
+                    key = "%s|%s" % (device, portname)
+                    if slot_id is not None:
+                       key = slot_id.text + '-' + key
```
vganesan-nokia, line 755: Need to use | as separator between all Slot, device and name. The suggested naming format is: 
Slot|Device|Name. 
Examples: 
```
multi_npu_platform_01|ASIC0|Cpu0
multi_npu_platform_01|ASIC0|Ethernet1/1
Slot1|Asic0|Cpu0
LineCard1|Asic0|Ethernet1
```

vganesan-nokia (changes requested):

src/sonic-config-engine/minigraph.py
```diff
@@ -1125,33 +1185,33 @@ def parse_xml(filename, platform=None, port_config_file=None, asic_name=None, hw
...
             elif child.tag == str(QName(ns, "LinkMetadataDeclaration")):
                 linkmetas = parse_linkmeta(child, hostname)
             elif child.tag == str(QName(ns, "DeviceInfos")):
-                (port_speeds_default, port_descriptions) = parse_deviceinfo(child, hwsku)
+                (port_speeds_default, port_descriptions, sys_ports) = parse_deviceinfo(child, hwsku)
         else:
             if child.tag == str(QName(ns, "DpgDec")):
-                (intfs, lo_intfs, mvrf, mgmt_intf, vlans, vlan_members, pcs, pc_members, acls, vni, tunnel_intfs, dpg_ecmp_content) = parse_dpg(child, asic_name)
+                (intfs, lo_intfs, mvrf, mgmt_intf, voq_inband_intfs, vlans, vlan_members, pcs, pc_members, acls, vni, tunnel_intfs, dpg_ecmp_content) = parse_dpg(child, asic_name)
                 host_lo_intfs = parse_host_loopback(child, hostname)
```
arlakshm, line 1204: In the minigraph.xml this is defined under each asic.. We don't need this here.

src/sonic-config-engine/minigraph.py
```diff
@@ -1125,33 +1185,33 @@ def parse_xml(filename, platform=None, port_config_file=None, asic_name=None, hw
...
             elif child.tag == str(QName(ns, "PngDec")):
                 (neighbors, devices, port_speed_png) = parse_asic_png(child, asic_name, hostname)
             elif child.tag == str(QName(ns, "MetadataDeclaration")):
-                (sub_role) = parse_asic_meta(child, asic_name)
+                (sub_role, switch_id, switch_type, max_cores, instance_name) = parse_asic_meta(child, asic_name)
             elif child.tag == str(QName(ns, "LinkMetadataDeclaration")):
                 linkmetas = parse_linkmeta(child, hostname)
             elif child.tag == str(QName(ns, "DeviceInfos")):
-                (port_speeds_default, port_descriptions) = parse_deviceinfo(child, hwsku)
+                (port_speeds_default, port_descriptions, sys_ports) = parse_deviceinfo(child, hwsku)
```
arlakshm, line 1214: The sysports are also for each asic. I dont think we need it here.

src/sonic-config-engine/minigraph.py
```diff
@@ -1268,6 +1347,9 @@ def parse_xml(filename, platform=None, port_config_file=None, asic_name=None, hw
     results['INTERFACE'] = phyport_intfs
     results['VLAN_INTERFACE'] = vlan_intfs
 
+    if sys_ports:
```
arlakshm, line 1350: Can you add a unit tests for this ?

src/sonic-config-engine/tests/multi_npu_data/sample-minigraph.xml
```diff
@@ -1115,6 +1143,141 @@
           <Speed>40000</Speed>
         </a:EthernetInterface>
       </EthernetInterfaces>
+      <SystemPorts>
```
arlakshm, line 1146: I think the system ports needs to be defined inside each ASIC, not at the line card level.

src/sonic-config-engine/minigraph.py
```diff
@@ -886,7 +921,26 @@ def parse_deviceinfo(meta, hwsku):
...
+                    portname = sysport.find(str(QName(ns, "Name"))).text
+                    device = sysport.find(str(QName(ns, "Device"))).text
+                    system_port_id = sysport.find(str(QName(ns, "SystemPortId"))).text
+                    switch_id = sysport.find(str(QName(ns, "SwitchId"))).text
+                    core_id = sysport.find(str(QName(ns, "CoreId"))).text
+                    core_port_id = sysport.find(str(QName(ns, "CorePortId"))).text
+                    speed = sysport.find(str(QName(ns, "Speed"))).text
+                    num_voq = sysport.find(str(QName(ns, "NumVoq"))).text
+                    key = "%s|%s" % (device, portname)
+                    if slot_id is not None:
```
arlakshm, line 939: Isn't slot_id a required attribute ? will handling of system ports in swss work, if slot-id is not there in the key ?

src/sonic-config-engine/minigraph.py
```diff
@@ -808,7 +823,15 @@ def parse_meta(meta, hname):
                     cloudtype = value
                 elif name == "ResourceType":
                     resource_type = value
-    return syslog_servers, dhcp_servers, ntp_servers, tacacs_servers, mgmt_routes, erspan_dst, deployment_id, region, cloudtype, resource_type
+                elif name == "SwitchId":
```
arlakshm, line 826: This is configuration is for the Line card host not asic, do we need these attributes ?

arlakshm:
Hi @mlorrillere, The multi_npu minigraph.xml has configuration which is not applicable to the VoQ chassis and vice-versa.  Can you please add a new sample minigraph.xml and unit tests for VoQ chassis, instead of modifying the multi_npu ones.

arlakshm (changes requested):
as comments

mlorrillere:
> Hi @mlorrillere, The multi_npu minigraph.xml has configuration which is not applicable to the VoQ chassis and vice-versa. Can you please add a new sample minigraph.xml and unit tests for VoQ chassis, instead of modifying the multi_npu ones.

Sure, I'll add a dedicated unit test for VoQ minigraph.xml

src/sonic-config-engine/minigraph.py
mlorrillere: These are required in the case of single asic linecards. In this case the linecard is a single host device. For multi-asic linecards or multi-asic fixed systems these attributes needs to be parsed for each asic in `parse_asic_meta`.

src/sonic-config-engine/minigraph.py
mlorrillere: `slot_id` is required to identify each linecard in a chassis. For a fixed system with multiple asics this is not required since the `device` is enough to identify the each asic of the system.

In the end `slot_id` is only necessary to build a unique key to identify each asic in the system. If that helps we can make it mandatory for single-asic systems even if this is not required.

src/sonic-config-engine/minigraph.py
mlorrillere: For single-asic linecards this is defined under the host's `DpgDec` section.

src/sonic-config-engine/minigraph.py
mlorrillere: For single-asic linecards this is defined under the host's `DeviceInfos` section.

src/sonic-config-engine/tests/multi_npu_data/sample-minigraph.xml
mlorrillere: Each asic needs to know about all the system ports in the system. Currently system ports are initialized during orchagent init so this makes it easier to have them defined in `DeviceInfo` and parsed by each asic (either in multi-asic setup or in single-asic linecards) than having them duplicated for each asic in the minigraph.xml

src/sonic-config-engine/minigraph.py
```diff
@@ -1192,6 +1252,25 @@ def parse_xml(filename, platform=None, port_config_file=None, asic_name=None, hw
...
+        results['DEVICE_METADATA']['localhost']['switch_id'] = switch_id
+    # on Voq system each asic has a switch_type
+    if switch_type is not None:
+        results['DEVICE_METADATA']['localhost']['switch_type'] = switch_type
+    # on Voq system each asic has a max_cores
+    if max_cores is not None:
+        results['DEVICE_METADATA']['localhost']['max_cores'] = max_cores
+    # on Voq system each asic has a instance_name
+    if instance_name is not None:
+        results['DEVICE_METADATA']['localhost']['instance_name'] = instance_name
```
vganesan-nokia, line 1266: We do not use the "instance_name" anywhere in the voq implementation. If this is not used anywhere we can remove this.
But we need "slot_id" and "device" in DEVICE_METADATA to derive the system lag name in the same way how system port name is derived (Refer: Voq LAG HLD https://github.com/Azure/SONiC/pull/697)

src/sonic-config-engine/minigraph.py
```diff
@@ -1192,6 +1252,25 @@ def parse_xml(filename, platform=None, port_config_file=None, asic_name=None, hw
         current_device['sub_role'] = sub_role
         results['DEVICE_METADATA']['localhost']['sub_role'] =  sub_role
         results['DEVICE_METADATA']['localhost']['asic_name'] =  asic_name
+    # on Voq system each asic has a switch_id
```
vganesan-nokia, line 1255: VoQ system needs "asic_id" also. The format is either simple sequential number like 0,1,2, etc., identifying the asic or PCI device id format 00:00.0. (Refer multi-asic HLD https://github.com/Azure/SONiC/blob/master/doc/multi_asic/SONiC_multi_asic_hld.md)

vganesan-nokia (changes requested):

src/sonic-config-engine/minigraph.py
mlorrillere: I think "instance_name" here has the same meaning as "asic_name" in the VoQ LAG HLD, so I will rename "instance_name" -> "asic_name".

vganesan-nokia: I noticed that there is an api in minigraph.py that extracts asic_id from asic_name. This indicates that asic_name will be either of the two forms like asic0, asic1, etc. or asic<PCI device id> form.
If we have to have the asic_id in the format PCI device id, I do not think it will be good idea to use it for system lag name.

Therefore, I recommend to change the voq lag hld to use "slot_id" and "device" and add these attributes in device meta data so that both system port name and system lag name will be derived in similar way (using slot_id and device). We can get rid of "instance_name".

If this is fine with all, I'll take care of updating voq lag hld.

mlorrillere: I'm not sure that `asic_id` in minigraph.py has the same meaning as `asic_id` from CONFIG_DB. The former is used to extract the instance ID as described by systemd services, while the later is intended to be used internally by the platform to identify a particular asic (like PCI device ID, see also https://github.com/Azure/sonic-buildimage/blob/master/dockers/docker-orchagent/orchagent.sh#L28-L42).

I think `asic_id` in minigraph.py is misleading as it really is the "instance name" as defined by systemd.unit manual.

In any cases, using `slot_id` and `device` looks reasonable to avoid the confusion.

mlorrillere: Hi @vganesan-nokia 

It looks like `asic_name` in `minigraph.py` is actually the asic's hostname. Adding `device` to minigraph.xml will basically replicate asic's `Hostname`. I think it would be fair here to simply do:
```python3
        results['DEVICE_METADATA']['localhost']['asic_name'] = asic_name
```
or, if it makes more sense to name it `device_name` in CONFIG_DB:
```python3
        results['DEVICE_METADATA']['localhost']['device_name'] = asic_name
```

Let me know what you think.

vganesan-nokia: > I'm not sure that `asic_id` in minigraph.py has the same meaning as `asic_id` from CONFIG_DB. The former is used to extract the instance ID as described by systemd services, while the later is intended to be used internally by the platform to identify a particular asic (like PCI device ID, see also https://github.com/Azure/sonic-buildimage/blob/master/dockers/docker-orchagent/orchagent.sh#L28-L42).
> 
> I think `asic_id` in minigraph.py is misleading as it really is the "instance name" as defined by systemd.unit manual.

Instead of saying misleading, I would say, the minigraph.py does not support PCI device id format for asic_id. Currently minigraph.py takes the asic_id from asic_name, which is assumed to be in the format asic0, asic1, and so on. DEVICE_METATDATA['localhost']['asic_id'] is required attribute for multi-asic. This is actually passed to orchangent and also to SAI for device identification in SAI/sdk. As you mentioned this is also used for instance id (namespace) in systemd services. So as long the asic_id has the name in the format "asic0", "asic1" etc we are good.  But since minigraph.py does not have support for handling and passing asic_id in PCI device id format, If we have to pass the asic_id in PCI device id format in config_db, we'll have to over-write this already filled asic_id by value we get from default_config.json (as being discussed in https://github.com/Azure/sonic-utilities/pull/1228). If this is going to be the case, we do not have to worry about asic_id in minigraph.

vganesan-nokia: > Hi @vganesan-nokia
> 
> It looks like `asic_name` in `minigraph.py` is actually the asic's hostname. Adding `device` to minigraph.xml will basically replicate asic's `Hostname`. I think it would be fair here to simply do:
> 
> ```python
>         results['DEVICE_METADATA']['localhost']['asic_name'] = asic_name
> ```
> 
> or, if it makes more sense to name it `device_name` in CONFIG_DB:
> 
> ```python
>         results['DEVICE_METADATA']['localhost']['device_name'] = asic_name
> ```
> 
> Let me know what you think.

I think both in minigraph and config_db "asic_name" takes the value "asic0", "asic1" and so on. Also note that asic_name is nothing but the namespace passed to parse_xml() by sonic-cfggen. And we have a separate config_db attribute for "hostname" which is "Hostname" in minigraph. So I guess "asic_name" cannot be hostname but should be considered as asic instance name. Therefore, we do not need "InstanceName" and "Device" in minigraph instead use the "asic_name".
 
The "hostname" which is used for both single asic and multi-asic should be equivalent to the line card in chassis context. So we can consider the "Hostname" for "slot_id" and avoid introducing "SlotId" in minigraph.

So my recommendations are:
(1) Set config db ['DEVICE_METADATA']['localhost']['asic_name'] = asic_name, as you suggested.
(2) Set config_db ['DEVICE_METADATA']['localhost']['hostname'] = hostname
(3) Remove "InstanceName" and use "asic_name" for instance name if instance name is needed
(4) For system port table: Remove "SlotId" and "Device". Use "hostname" and "asic_name" instead and derive system port key using the format "hostname|asic_name|portname"

I think (1) and (2) are already there as part of multi-asic. I feel that it is better to extend the use of existing multi-asic attributes for voq chassis systems also instead of introducing new attributes in config db and minigraph.

vganesan-nokia: Correct. We need "hostname" and "asic_name" in the DEVICE_METADATA and should use these for deriving the system port name something like  "<hostname>|<asic_name>|Ethernet1". The recommendation (4) is to say that we do not need to define "SlotId", 'Device" and "InstanceName" in minigraph. We already have "hostname" and "asic_name" and "asic_name" is same as instance name.

vganesan-nokia: This is sounds good to me. It will be good to have the values of HostName and AsicName consistent with values in DEVICE_METADATA "hostname" and "asic_name". The reason is: the values of  ['DEVICE_METADATA']['localhost']['hostname']  and  ['DEVICE_METADATA']['localhost']['asic_name'] are used for deriving the system lag name. It will be good to have naming convention consistent for system port also. So for here, if we make sure HostName and AsicName, use the same corresponding values (the values of  ['DEVICE_METADATA']['localhost']['hostname']  and  ['DEVICE_METADATA']['localhost']['asic_name'], it will be great.

mlorrillere:
Hi @vganesan-nokia and @arlakshm ,

Last changes include updated unit test with a dedicated sample minigraph XML file and address other comments related to InstanceName and SystemPorts .

src/sonic-config-engine/tests/test_cfggen.py
```diff
@@ -647,3 +648,38 @@ def test_show_run_interfaces(self):
         argument = '-a \'{"key1":"value"}\' --var-json INTERFACE'
         output = self.run_script(argument)
         self.assertEqual(output, '')
+
+    def test_minigraph_voq_metadata(self):
+        argument = "-m {} --var-json DEVICE_METADATA".format(self.sample_graph_voq)
+        output = json.loads(self.run_script(argument))
+        self.assertEqual(output['localhost']['switch_id'], '0')
+        self.assertEqual(output['localhost']['switch_type'], 'voq')
+        self.assertEqual(output['localhost']['max_cores'], '16')
```
vganesan-nokia, line 657: Are we checking for the "asic_id", "asic_name", and "hostname" in any other test case? If not would you please add verification (assertEqual) for ['localhost']['asic_id'], ['localhost']['hostname'] and ['localhost']['asic_name'] here? These are mandatory attributes if "switch_type" is "voq".

vganesan-nokia (changes requested):

vganesan-nokia:
> Hi @vganesan-nokia and @arlakshm ,
> 
> Last changes include updated unit test with a dedicated sample minigraph XML file and address other comments related to InstanceName and SystemPorts .

@mlorrillere, thanks for addressing comments. Things look good. I have left a minor comment in test_cfggen.py.

src/sonic-config-engine/tests/test_cfggen.py
mlorrillere: `asic_id` and `asic_name` are multi-asic attributes, not specific to VoQ switches. I don't think we need either of them on single-asic if `switch_type` is `"voq"`. On single-asic devices where `asic_name` is not specified the system port key is "hostname|portname".

For checks:
- `asic_name` is tested in test_multinpu_cfggen.py but not `asic_id`.
- `hostname` is not specific to VoQ switches but I didn't see any check for this one.

I'll update test_cfggen.py to add a check for `hostname` and `test_multinpu_cfggen.py` for `asic_id`.

vganesan-nokia (dismissed):

mlorrillere:
Hi @arlakshm and @lguohan 

Can you approve and merge this review if there are no further comments?

Thanks,
Maxime

arlakshm (approved):
lgtm

minionatwork (approved):

PR #5993, opened 2020-11-21
[bgpcfgd]: Batch bgp updates.

pavel-shirshov:
**- Why I did it**
vtysh -f command is slow. It sometimes takes about 3 seconds.
When we need to run many vtysh -f commands that slows down the system.

**- How I did it**
1. Read as many messages as possible from the database
2. Run through bgpcfgd all the messages, collecting  frr updates;
3. commit all updates as a batch

**- How to verify it**
Build an image and run on your dut. All tests must be working

- [ ] 201811
- [x] 201911
- [x] 202006

### DIFF
--- a/src/sonic-bgpcfgd/bgpcfgd/managers_bgp.py
+++ b/src/sonic-bgpcfgd/bgpcfgd/managers_bgp.py
@@ -7,7 +7,7 @@ import netaddr
 from .log import log_warn, log_err, log_info, log_debug, log_crit
 from .manager import Manager
 from .template import TemplateFabric
-from .utils import run_command
+from .frr import run_command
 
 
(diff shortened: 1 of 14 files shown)


src/sonic-bgpcfgd/bgpcfgd/managers_bgp.py
```diff
@@ -7,7 +7,7 @@
 from .log import log_warn, log_err, log_info, log_debug, log_crit
 from .manager import Manager
 from .template import TemplateFabric
-from .utils import run_command
+from .frr import run_command
```
pavel-shirshov, line 10: Is this correct?

lguohan:
please use your own repo

PR #6011, opened 2020-11-24
[bgpcfgd]: Add on-match next rule for set ipv6 next-hop prefer-global

pavel-shirshov:
**- Why I did it**
Without the rule, IPv6 route-map entries were skipped.
**- How I did it**
Added 'on-match next' rule after 'set ipv6 next-hop prefer-global.
Updated a test to check that we have 'on-match next' rule with the 'set ipv6 next-hop prefer-global'

**- How to verify it**
Build an image. Run on your dut. Add a route-map entry after 'set ipv6 next-hop prefer-global'. Check that the route-map entry works  

- [ ] 201811
- [ ] 201911
- [x] 202006

### DIFF
--- a/src/sonic-bgpcfgd/tests/test_ipv6_nexthop_global.py
+++ b/src/sonic-bgpcfgd/tests/test_ipv6_nexthop_global.py
@@ -79,6 +79,7 @@ def extract_rm_from_peer_group(path, peer_group_name):
 def check_routemap_in_file(filename, route_map_name):
     route_map_re = re.compile(r'^route-map\s+%s\s+permit\s+(\d+)' % route_map_name)
     set_re = re.compile(r'set ipv6 next-hop prefer-global')
+    next_re = re.compile(r'on-match next')
     with open(filename) as fp:
         lines = [line.strip() for line in fp if not line.strip().startswith('!') and line.strip() != '']
     found_entry = False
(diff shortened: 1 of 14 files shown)


src/sonic-bgpcfgd/tests/test_ipv6_nexthop_global.py
```diff
@@ -79,28 +79,34 @@ def extract_rm_from_peer_group(path, peer_group_name):
...
             assert found_seq_no not in route_map_entries, "Route-map has duplicate entries: %s - %d" % (route_map_name, found_seq_no)
             found_entry = True
+            route_map_entries[found_seq_no] = [False, False]
     results = [route_map_entries[seq] for seq in sorted(route_map_entries.keys())]
-    if (len(results)):
-        err_msg = "route-map %s doesn't have mandatory permit entry for 'set ipv6 next-hop prefer-global" % route_map_name
-        assert results[0], err_msg
-    return len(results) > 0
+    err_msg = "route-map %s doesn't have mandatory permit entry for 'set ipv6 next-hop prefer-global" % route_map_name
+    assert len(results), err_msg
```
lguohan, line 107: the err_msg here does not match exactly the assertion.

lguohan (approved):

src/sonic-bgpcfgd/tests/test_ipv6_nexthop_global.py
pavel-shirshov: the results arrays contains a list of pairs.
Each pair corresponds to the permit entry in ipv6 route-map
First permit entry in ipv6 route-map should contains 'set ipv6 ....' (first element in the result element) and 'on-match next' (second element in result element). So if the result array empty - we don't have any permit elelement in ipv6 route-map

Issue #6009, opened 2020-11-23
determine-reboot-cause fails due to TypeError seen during during boot up


vaibhavhd:
**Description**

Reboot cause determination fails and errors are seen in syslog.
The errors seem to be coming due to Python3 script is using Python2 based library from different vendors at 
https://github.com/Azure/sonic-buildimage/tree/master/platform/broadcom

The Python3 jump for `determine-reboot-cause` and `process-reboot-cause` has been made with the platform specific lib still based out in Python2.

**Steps to reproduce the issue:**
1. Install latest SONiC master image on Arista device. I hit the issue in 500
2. Issue is seen in syslog at the beginning stage where DUT is booting into the new image.

**Describe the results you received:**

```
Nov 23 17:10:34 sonic root: SONiC version master.500-e0781f46 starting up...
Nov 23 17:10:34 sonic root: SONiC version master.500-e0781f46 starting up...
Nov 23 17:10:34 sonic rc.local[469]: + cut -d , -f2
Nov 23 17:10:34 sonic rc.local[469]: + cut -d , -f2
Nov 23 17:10:34 sonic determine-reboot-cause[599]: Traceback (most recent call last):
Nov 23 17:10:34 sonic determine-reboot-cause[599]: Traceback (most recent call last):
Nov 23 17:10:34 sonic determine-reboot-cause[599]:   File "/usr/local/bin/determine-reboot-cause", line 236, in <module>
Nov 23 17:10:34 sonic determine-reboot-cause[599]:   File "/usr/local/bin/determine-reboot-cause", line 236, in <module>
Nov 23 17:10:34 sonic determine-reboot-cause[599]:     main()
Nov 23 17:10:34 sonic determine-reboot-cause[599]:     main()
Nov 23 17:10:34 sonic determine-reboot-cause[599]:   File "/usr/local/bin/determine-reboot-cause", line 185, in main
Nov 23 17:10:34 sonic determine-reboot-cause[599]:   File "/usr/local/bin/determine-reboot-cause", line 185, in main
Nov 23 17:10:34 sonic determine-reboot-cause[599]:     (hardware_reboot_cause, additional_reboot_info) = find_hardware_reboot_cause()
Nov 23 17:10:34 sonic determine-reboot-cause[599]:     (hardware_reboot_cause, additional_reboot_info) = find_hardware_reboot_cause()
[...]

PR #6044, opened 2020-11-25
Fix the ExecStart path for process-reboot-cause

sujinmkang:
**- Why I did it**
The path should be /usr/local/bin not /usr/bin
**- How I did it**
Change the ExecStart path
**- How to verify it**
Check if it starts

- [ ] 201811
- [ ] 201911
- [ ] 202006

Addresses an issue mentioned in https://github.com/Azure/sonic-buildimage/issues/6009, but does not close the issue.

### DIFF
--- a/src/sonic-host-services-data/debian/sonic-host-services-data.process-reboot-cause.service
+++ b/src/sonic-host-services-data/debian/sonic-host-services-data.process-reboot-cause.service
@@ -1,7 +1,8 @@
 [Unit]
-Description=Reboot cause determination service
-After=rc-local.service
+Description=Retrieve the reboot cause from the history files and save them to StateDB
+Requires=database.service
+After=database.service
 
 [Service]
 Type=simple
-ExecStart=/usr/bin/process-reboot-cause
+ExecStart=/usr/local/bin/process-reboot-cause


jleveque (approved):

jleveque:
Retest vs please

vaibhavhd:
Closing this PR will auto-close the issue https://github.com/Azure/sonic-buildimage/issues/6009
I believe that this PR is definitely needed, but does not close/solve the issue. Instead of marking as a fix, can we just relate this PR to the issue.

The issue still needs work from Arista to update the dependency to determine-reboot-cause:
https://github.com/aristanetworks/sonic/blob/3943e69510b471e65797d6054dd5f47cff3012c3/systemd/platform-arista-reboot-cause.service#L5

src/sonic-host-services-data/debian/sonic-host-services-data.process-reboot-cause.service
```diff
@@ -1,7 +1,8 @@
 [Unit]
-Description=Reboot cause determination service
-After=rc-local.service
+Description=Retrieve the reboot cause from the history files and save them to StateDB
+Requires=database.service
```
vaibhavhd, line 3: Shouldn't this change (`Requires=database.service`) be added to `determine-reboot-cause.service` too?

https://github.com/Azure/sonic-buildimage/blob/master/src/sonic-host-services-data/debian/sonic-host-services-data.determine-reboot-cause.service#L3

jleveque: `determine-reboot-cause.service` Does not require the database to be ready. It is intended to run as early at boot as possible in order to capture the reboot reason. The reason we need two services is because `process-reboot-cause.service` must run after the DB is ready in order to push the data to the DB.

jleveque:
> Closing this PR will auto-close the issue #6009
> I believe that this PR is definitely needed, but does not close/solve the issue. Instead of marking as a fix, can we just relate this PR to the issue.
> 
> The issue still needs work from Arista to update the dependency to determine-reboot-cause:
> https://github.com/aristanetworks/sonic/blob/3943e69510b471e65797d6054dd5f47cff3012c3/systemd/platform-arista-reboot-cause.service#L5

Agreed. I have removed the linkage.

vaibhavhd (approved):

Issue #6043, opened 2020-11-25
FRR log disabling code caused rsyslog error message when BGP container starting


yxieca:
**Description**
This issue is caused by PR #5962

**Steps to reproduce the issue:**
1. Run nightly test
2.
3.

**Describe the results you received:**

Whenever BGP container starts, following error message comes out:
syslog.104.gz:Nov 25 11:53:20.411386 str-dx010-acs-1 INFO systemd[1]: Starting BGP container...
syslog.104.gz:Nov 25 11:53:38.884340 str-dx010-acs-1 ERR bgp#rsyslogd: file './.': open error: Is a directory [v8.1901.0 try https://www.rsyslog.com/e/2433 ]
syslog.104.gz:Nov 25 11:53:38.885730 str-dx010-acs-1 INFO bgp#supervisord: rsyslogd rsyslogd: file './.': open error: Is a directory [v8.1901.0 try https://www.rsyslog.com/e/2433 ]
syslog.105.gz:Nov 25 11:00:06.351692 str-dx010-acs-1 ERR bgp#rsyslogd: file './.': open error: Is a directory [v8.1901.0 try https://www.rsyslog.com/e/2433 ]
syslog.105.gz:Nov 25 11:00:06.355682 str-dx010-acs-1 INFO bgp#supervisord: rsyslogd rsyslogd: file './.': open error: Is a directory [v8.1901.0 try https://www.rsyslog.com/e/2433 ]

**Describe the results you expected:**
BGP container start without error messages.

**Additional information you deem important (e.g. issue happens only occasionally):**

    **Output of `show version`:**

    ```
SONiC Software Version: SONiC.master.502-80bf8691
Distribution: Debian 10.6
Kernel: 4.19.0-9-2-amd64
Build commit: 80bf8691
Build date: Tue Nov 24 19:47:21 UTC 2020
Built by: johnar@jenkins-worker-8
    ```

    ```
syslog.104.gz:Nov 25 11:53:20.411386 str-dx010-acs-1 INFO systemd[1]: Starting BGP container...
syslog.104.gz:Nov 25 11:53:38.884340 str-dx010-acs-1 ERR bgp#rsyslogd: file './.': open error: Is a directory [v8.1901.0 try https://www.rsyslog.com/e/2433 ]
[...]

PR #6045, opened 2020-11-25
[FRR] remove the whole block of outchannel properly

yxieca:
**- Why I did it**
Fix issue #6043 

**- How I did it**
We are disabling in container frr log. The log entries are sent to base image and are logged in /var/log/quagga/bgpd.log.

However, we need to remove the whole outchannel config block to avoid an error message raised by rsyslogd.

**- How to verify it**
Without the change, test_autorestart bgp container will fail on loganalyer errors. With the change, restarting bgp container is no longer generating error message and the test will pass.

The log generated by frr continued appearing in /var/log/quagga/bgpd.log

- [ ] 201811
- [x] 201911
- [ ] 202006


judyjoseph (approved):
LGTM .. verified that logs are still coming in log files in host /var/log in multi-asic

yxieca:
retest vs please

abdosi:
retest vsimage please

PR #6048, opened 2020-11-26
[bgpcfgd]: Use peer commands for BBR, not peer-group

pavel-shirshov:
**- Why I did it**
FRR 7.2.1 has a bug: when 'allowas-in' is configured on peer-group, and we remove that configuration, the neighbors, which had that configuration, keeps that configuration.

**- How I did it**
To fix the bug I apply the configuration to all neighbors, which are configured with the peer-groups, which should support BBR.

**- How to verify it**
1. Build an image
2. Run on your dut
3. Check that the test https://github.com/Azure/sonic-mgmt/blob/master/tests/bgp/test_bgp_bbr.py not fails

- [ ] 201811
- [ ] 201911
- [ ] 202006

### DIFF
--- a/src/sonic-bgpcfgd/bgpcfgd/managers_bbr.py
+++ b/src/sonic-bgpcfgd/bgpcfgd/managers_bbr.py
@@ -1,4 +1,5 @@
 import re
+from collections import defaultdict
 
 from swsscommon import swsscommon
 
@@ -106,7 +107,9 @@ class BBRMgr(Manager):
         :return: list of commands prepared for FRR
         """
         bgp_asn = self.directory.get_slot("CONFIG_DB", swsscommon.CFG_DEVICE_METADATA_TABLE_NAME)["localhost"]["bgp_asn"]
+        self.cfg_mgr.update()
         available_peer_groups = self.__get_available_peer_groups()
+        available_peers_per_pg = self.__get_available_peers_per_peer_group(available_peer_groups)
         cmds = ["router bgp %s" % bgp_asn]
         prefix_of_commands = "" if status == "enabled" else "no "
         peer_groups_to_restart = set()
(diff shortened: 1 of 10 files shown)


pavel-shirshov:
retest mellanox please

pavel-shirshov:
retest broadcom please

pavel-shirshov:
retest broadcom please

pavel-shirshov:
retest mellanox please

src/sonic-bgpcfgd/bgpcfgd/managers_bbr.py
```diff
@@ -106,15 +107,18 @@ def __set_prepare_config(self, status):
         :return: list of commands prepared for FRR
         """
         bgp_asn = self.directory.get_slot("CONFIG_DB", swsscommon.CFG_DEVICE_METADATA_TABLE_NAME)["localhost"]["bgp_asn"]
+        self.cfg_mgr.update()
```
abdosi, line 110: @pavel-shirshov what is the use case of update() here ?

xumia:
retest mellanox please

lguohan:
retest broadcom please

lguohan:
retest mellanox please

wangxin (approved):
LGTM

src/sonic-bgpcfgd/bgpcfgd/managers_bbr.py
pavel-shirshov: To download latest 'show run' from frr.
Otherwise it would be None.

P.S.: It is better to have this call in the runner, to update the configuration at the start of update cycle. and then we need to check both 'show run' records and proposed records changes, in case managers works with configuration which is changed in many places.

PR #6051, opened 2020-11-26
[kernel] Change grub cmdline to set c-states to 0 for "Intel" CPUs

shlomibitton:

**- Why I did it**
Usually for a use case like networking - should not be configured to reach c6, the maximum used is c1e – due to the added latency getting in & out of states (bad for networking).

Following a recommendation by Intel, networking system should avoid getting in & out of states which introduce latency. The recommended state is c1e and no state change enabling.

In addition, c-state sole purpose is to save power and when inside a networking switch its really negligent being such a tiny consumer vs. the whole cluster.

**- How I did it**
Added a new argument to the grub cmdline to set c-states to 0.

**- How to verify it**
* install 'linux-cpupower' tool.
* Run 'turbostat --debug' command
* Observe the c-states is constantly on state 1.

- [ ] 201811
- [x] 201911
- [ ] 202006

### DIFF
--- a/installer/x86_64/install.sh
+++ b/installer/x86_64/install.sh
@@ -538,8 +538,10 @@ trap_push "rm $grub_cfg || true"
 # Check if the CPU vendor is 'Intel' and disable c-states if True
 CPUVENDOR=$(cat /proc/cpuinfo | grep -m 1 vendor_id | awk '{print $3}')
 echo "Switch CPU vendor is: $CPUVENDOR"
-if [[ $(echo $CPUVENDOR | grep "Intel") ]] ; then
+if [[ $(echo $CPUVENDOR | grep -i "Intel") ]] ; then
     CSTATES="intel_idle.max_cstate=0"
+else
+    CSTATES=""
 fi
 
 DEFAULT_GRUB_SERIAL_COMMAND="serial --port=${CONSOLE_PORT} --speed=${CONSOLE_SPEED} --word=8 --parity=no --stop=1"
--- a/installer/x86_64/install.sh
+++ b/installer/x86_64/install.sh
@@ -536,7 +536,7 @@ trap_push "rm $grub_cfg || true"
 [ -r ./platform.conf ] && . ./platform.conf
 
 DEFAULT_GRUB_SERIAL_COMMAND="serial --port=${CONSOLE_PORT} --speed=${CONSOLE_SPEED} --word=8 --parity=no --stop=1"
-DEFAULT_GRUB_CMDLINE_LINUX="console=tty0 console=ttyS${CONSOLE_DEV},${CONSOLE_SPEED}n8 quiet"
+DEFAULT_GRUB_CMDLINE_LINUX="console=tty0 console=ttyS${CONSOLE_DEV},${CONSOLE_SPEED}n8 quiet intel_idle.max_cstate=0"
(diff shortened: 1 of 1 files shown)


lguohan:
retest vsimage please

lguohan:
what if the platform uses non intel cpu? like amd cpu or arm cpu?

liat-grozovik:
@shlomibitton please check if we can read a parameter which identify this is intel cpu and add it only for it.

shlomibitton:
@lguohan I added a condition to change it only for 'Intel' CPUs.
can you please check it?

installer/x86_64/install.sh
```diff
@@ -535,8 +535,15 @@ trap_push "rm $grub_cfg || true"
 
 [ -r ./platform.conf ] && . ./platform.conf
 
+# Check if the CPU vendor is 'Intel' and disable c-states if True
+CPUVENDOR=$(cat /proc/cpuinfo | grep -m 1 vendor_id | awk '{print $3}')
+echo "Switch CPU vendor is: $CPUVENDOR"
+if [[ $(echo $CPUVENDOR | grep "Intel") ]] ; then
```
lguohan, line 541: use -i make it case insensitive

installer/x86_64/install.sh
```diff
@@ -535,8 +535,15 @@ trap_push "rm $grub_cfg || true"
...
+    CSTATES="intel_idle.max_cstate=0"
+fi
```
qiluo-msft, line 543: else, assign CSTATES to empty string.

qiluo-msft (dismissed):
As comment

lguohan:
> Observe the c-states is constantly on state 1.

is this incorrect? state should be 0, right?

lguohan:
check out this one. why do they set both processor.max-cstate=1 and intel_idle.max_cstate=0? why do you only set intel_idle.max_cstate?

https://gist.github.com/wmealing/2dd2b543c4d3cff6cab7

shlomibitton:
@lguohan 

> > Observe the c-states is constantly on state 1.
> 
> is this incorrect? state should be 0, right?

No, state 0 means the CPU is currently active.
by the turbostat tool man page:
Busy% percent of the measurement interval that the CPU executes instructions, aka. % of time in "C0" state.
This is an example of an output when running the tool with c-states = 0.
As you can see when it is not "busy" i.e. "idle" it is in state C0, which is expected and desired.
image


> check out this one. why do they set both processor.max-cstate=1 and intel_idle.max_cstate=0? why do you only set intel_idle.max_cstate?
> 
> https://gist.github.com/wmealing/2dd2b543c4d3cff6cab7

by using "processor.max-cstate=1" we limit the c-states level to a certain level, but this is not needed when using "intel_idle.max_cstate=0" because this is disables the driver, forcing it to stay active all the time.
you can check in the kernel code base the file "drivers/idle/intel_idle.c" there is an indication for it:
/* intel_idle.max_cstate=0 disables driver */

installer/x86_64/install.sh
shlomibitton: Fixed.

installer/x86_64/install.sh
shlomibitton: Fixed.

lguohan (approved):

qiluo-msft (approved):

PR #6066, opened 2020-11-30
DHCP relay: support for dual ToR

trzhang-msft:
DHCP relay agent: support for dual ToR scenario. In the new DHCP packet flow if enabled, DHCP option 82 suboption 5 will be enabled to carry VLAN interface address, while the switch loopback address will be used as giaddr in DHCP Discover / Request packets destined to DHCP servers.

This PR also fixes the bug of incorrect netmask setup in the original isc-dhcp open source code.

The DHCP docker template will be updated in a separate PR to consume the new option [-U Loopback0 -dt].

The update has been manually verified on a VS-based testbed through checking the DHCP packet flow traces.


src/isc-dhcp/patch/0009-Support-for-dual-tor-scenario.patch
```diff
@@ -0,0 +1,437 @@
...
++static void *receive_pkt_from_loopback(void *);
++static void free_downstream_intfs(void);
++
+ static const char copyright[] =
+ "Copyright 2004-2018 Internet Systems Consortium.";
+ static const char arr[] = "All rights reserved.";
+@@ -189,6 +204,7 @@ char *progname;
+ "                     [-iu interface0 [ ... -iu interfaceN]\n" \
+ "                     [-id interface0 [ ... -id interfaceN]\n" \
+ "                     [-U interface]\n" \
```
tahmed-dev, line 71: I came to learn about -U option recently. Can we instead of having new option, provide the -U with the loopback interface? this will do the following:
```
		/* RFC3527: Use the inbound packet's interface address in
		 * the link selection suboption and set the outbound giaddr
		 * to the uplink address. */
		if (adding_link_select) {
			*sp++ = RAI_LINK_SELECT;
			*sp++ = 4u;
			memcpy(sp, &giaddr.s_addr, 4);
			sp += 4;
			packet->giaddr = uplink->addresses[0];
			log_debug ("Adding link selection suboption"
				   " with addr: %s", inet_ntoa(giaddr));
		}
```
which sending the `giaddr` as option 82 sub-option 5 and update `giaddr` with the loopback interface address.

trzhang-msft: no, 1) loopback on sonic is a 'virtual' interface 2) we use different packet flow logic, e.g., we use yiaddr to filter vlan 3) two options for different scenarios, better not mix them up.

tahmed-dev: This is not to say we are going to eliminate the filter logic. We would still be filtering the packets based on the `yiaddr` since the the loopback interface will be shared among different relay instances. Can you please elaborate on 1 and 3?

tahmed-dev: dug a bit deeper in the code and it looks like -U would mark the interface as UPSTREAM and DOWNSTREAM (INTERFACE_STREAMS)
```
			interface_snorf(uplink, (INTERFACE_REQUESTED |
						INTERFACE_STREAMS));

```

We could use this information know where the packed is coming from, or we could following similar approach as `-U` option and mark the interface as UPSTREAM only. This way, there is no need to introduces a thread to receive packet. We will still be using the common lib to receive packets. let's sync up offline...

src/isc-dhcp/patch/0009-Support-for-dual-tor-scenario.patch
```diff
@@ -0,0 +1,437 @@
...
++	int result;
++	struct iaddr ifrom;
++	int flag = 1;
++	int retry = 0;
++	const int max_retry = 10;
++
++	strcpy(loopback_intf.name, "loopback_intf");
++	loopback_intf.addresses = &loopback_addr;
++	loopback_intf.address_count = 1;
++	loopback_intf.flags = 0 | INTERFACE_UPSTREAM;
```
jleveque, line 392: No need to `or` with `0`. `loopback_intf.flags = INTERFACE_UPSTREAM;` should have the same result.

src/isc-dhcp/patch/0009-Support-for-dual-tor-scenario.patch
```diff
@@ -0,0 +1,437 @@
...
++		free(cdi);
++	}
++}
++
++static void *
++receive_pkt_from_loopback(void *arg) {
++	struct interface_info loopback_intf;
++	struct sockaddr_in address;
++	union {
++		unsigned char buffer [4095];
```
jleveque, line 380: Suggest changing name to `packbuf` and adding a comment for clarity to align with rest of codebase, e.g.,

```c
unsigned char packbuf [4095]; /* Packet input buffer.
                               * Must be as large as largest
                               * possible MTU. */
```

jleveque (changes requested):

src/isc-dhcp/patch/series
```diff
@@ -7,3 +7,5 @@
 0006-Bugfix-Ensure-HAVE_SO_BINDTODEVICE-has-a-chance-to-b.patch
 0007-If-destination-of-BOOTREQUEST-is-directed-broadcast-.patch
 0008-Don-t-skip-down-interfaces-when-discovering-interfac.patch
+0009-Support-for-dual-tor-scenario.patch
+0010-Bugfix-correctly-set-interface-netmask.patch
```
jleveque, line 11: Add newline at EOF

jleveque (changes requested):

src/isc-dhcp/patch/0009-Support-for-dual-tor-scenario.patch
jleveque: > We could use this information know where the packed is coming from, or we could following similar approach as -U option and mark the interface as UPSTREAM only. This way, there is no need to introduces a thread to receive packet. We will still be using the common lib to receive packets.

I would prefer eliminating the need to spawn a new thread, if possible.

src/isc-dhcp/patch/0009-Support-for-dual-tor-scenario.patch
tahmed-dev: the code is using the same underlying package's [nomenclature](https://salsa.debian.org/dhcp-team/isc-dhcp/-/blob/master/common/discover.c#L1094). When we use `-U` or similar approach, this will go away.

src/isc-dhcp/patch/0009-Support-for-dual-tor-scenario.patch
```diff
@@ -0,0 +1,308 @@
...
+ 			return;
+ 		}
+ 
+@@ -987,8 +1044,14 @@ do_relay4(struct interface_info *ip, struct dhcp_packet *packet,
+ 	   forward the response to the correct net.    If it's already
+ 	   set, the response will be sent directly to the relay agent
+ 	   that set giaddr, so we won't see it. */
+-	if (!packet->giaddr.s_addr)
+-		packet->giaddr = ip->addresses[0];
++	if (!packet->giaddr.s_addr) {
```
tahmed-dev, line 210: Is this code snippet required? looks like the `-U` option will update the both `giaddr` and Option 82 sub-option 5, right?

trzhang-msft: this if clause should not be triggered at all, but I follow the existing code logic in case we miss some cases.

tahmed-dev: It is still not required as this is applicable for the generic case. With `-U` option, the giaddr is updated elsewhere per the following excerpt:

```
		/* RFC3527: Use the inbound packet's interface address in
		 * the link selection suboption and set the outbound giaddr
		 * to the uplink address. */
		if (adding_link_select) {
			*sp++ = RAI_LINK_SELECT;
			*sp++ = 4u;
			memcpy(sp, &giaddr.s_addr, 4);
			sp += 4;
			packet->giaddr = uplink->addresses[0];
			log_debug ("Adding link selection suboption"
				   " with addr: %s", inet_ntoa(giaddr));
		}
```

tahmed-dev (changes requested):

src/isc-dhcp/patch/0009-Support-for-dual-tor-scenario.patch
```diff
@@ -0,0 +1,308 @@
...
+ 
+-	/* Scan the interface list looking for an interface whose
+-	   name matches the one specified in circuit_id. */
+-
+-	for (ip = interfaces; ip; ip = ip->next) {
+-		if (ip->circuit_id &&
+-		    ip->circuit_id_len == circuit_id_len &&
+-		    !memcmp(ip->circuit_id, circuit_id, circuit_id_len))
+-			break;
++	if (enable_support_for_dual_tor) {
```
tahmed-dev, line 242: is this piece of code an optimization? The interface would still be found, right?

src/isc-dhcp/patch/0009-Support-for-dual-tor-scenario.patch
trzhang-msft: updated

src/isc-dhcp/patch/0009-Support-for-dual-tor-scenario.patch
trzhang-msft: the code is not necessary there, updated

src/isc-dhcp/patch/0009-Support-for-dual-tor-scenario.patch
```diff
@@ -0,0 +1,269 @@
...
++static void
++free_downstream_intfs(void) {
++	struct downstream_intf_list *cdi;
++	while (downstream_intfs) {
++		cdi = downstream_intfs;
++		downstream_intfs = downstream_intfs->next;
++		free(cdi);
++	}
++}
+\ No newline at end of file
```
tahmed-dev, line 266: nit: new line at end of file

src/isc-dhcp/patch/0009-Support-for-dual-tor-scenario.patch
```diff
@@ -0,0 +1,269 @@
...
+
+diff --git a/relay/dhcrelay.c b/relay/dhcrelay.c
+index e158efe..ddde6b6 100644
+--- a/relay/dhcrelay.c
++++ b/relay/dhcrelay.c
+@@ -56,6 +56,7 @@ int bogus_agent_drops = 0;	/* Packets dropped because agent option
+ 				   specified. */
+ int bogus_giaddr_drops = 0;	/* Packets sent to us to relay back to a
+ 				   client, but with a bogus giaddr. */
++int bogus_yiaddr_drops = 0;
```
tahmed-dev, line 18: nit: comment for `bogus_yiaddr_drops`

tahmed-dev (approved):

src/isc-dhcp/patch/0009-Support-for-dual-tor-scenario.patch
trzhang-msft: updated

src/isc-dhcp/patch/0009-Support-for-dual-tor-scenario.patch
trzhang-msft: updated

tahmed-dev (approved):
Thanks @trzhang-msft lgtm.

tahmed-dev (approved):

jleveque (approved):

PR #6095, opened 2020-12-02
[vmcore] Add more kernel panic conditions for vmcore dump

shlomibitton:

**- Why I did it**
Add more kernel panic conditions to allow vmcore dump in additional situations to have better debugging information.

**- How I did it**
Create new file to "sysctl.d" with desired panic conditions.
It will trigger a vmcore dump using kdump-tools on these situations.

**- How to verify it**

- [ ] 201811
- [ ] 201911
- [ ] 202006


lguohan:
@rajendra-dendukuri , can you help to review this to see if it is align with the kdump design?

build_debian.sh
```diff
@@ -318,6 +318,11 @@ sudo LANG=C DEBIAN_FRONTEND=noninteractive chroot $FILESYSTEM_ROOT apt-get -y in
     haveged                 \
     jq
 
+## Configure kdump-tools with additional flags for vmcore dump
+sudo LANG=C chroot $FILESYSTEM_ROOT /bin/bash -c 'cat >> /etc/default/kdump-tools <<"EOF"
+KDUMP_SYSCTL="kernel.panic_on_oops=1 kernel.panic_on_unrecovered_nmi=1 kernel.panic_on_io_nmi=1 kernel.panic_on_stackoverflow=1 kernel.hung_task_panic=1 kernel.softlockup_panic=1 kernel.unknown_nmi_panic=1 vm.panic_on_oom=1"
```
lguohan, line 323: please check pr https://github.com/Azure/sonic-buildimage/pull/6113 this one does not seem to be align with the design.

shlomibitton: @lguohan  can you please check it now?

shlomibitton:
merge this after https://github.com/Azure/sonic-buildimage/pull/6113
to align with patches numbers

build_debian.sh
rajendra-dendukuri: @shlomibitton 

The below is part of the description of "KDUMP_SYSCTL"
    Each time a kdump kernel is loaded, “sysctl -w $KDUMP_SYSCTL” is also executed

In my opinion, we should not tie the possible causes of kernel panic with kdump being enabled. For example, certain users may not want to panic on oom or on soft lockup but still would want to panic and take kdump on other possibilities. So it is best to define these parameters in a file placed in /etc/sysctl.d/ or in /etc/sysctl.conf instead of tying them with the default kdump config.

installer/x86_64/install.sh
```diff
@@ -536,7 +536,7 @@ trap_push "rm $grub_cfg || true"
 [ -r ./platform.conf ] && . ./platform.conf
 
 DEFAULT_GRUB_SERIAL_COMMAND="serial --port=${CONSOLE_PORT} --speed=${CONSOLE_SPEED} --word=8 --parity=no --stop=1"
-DEFAULT_GRUB_CMDLINE_LINUX="console=tty0 console=ttyS${CONSOLE_DEV},${CONSOLE_SPEED}n8 quiet"
+DEFAULT_GRUB_CMDLINE_LINUX="console=tty0 console=ttyS${CONSOLE_DEV},${CONSOLE_SPEED}n8 quiet crashkernel=256M"
```
rajendra-dendukuri, line 539: This will reserve 256M of RAM by default. The idea is to set crashkernel=xxx value only if kdump is enabled. Can you please elaborate why we need this?

files/vmcore_dump/vmcore-sysctl.conf
```diff
@@ -0,0 +1,7 @@
+kernel.panic_on_oops=1
```
rajendra-dendukuri, line 1: Can we move this file to "files/image_config/kdump"

rajendra-dendukuri (changes requested):

installer/x86_64/install.sh
nazariig: @shlomibitton how about `crashkernel=auto`?

rajendra-dendukuri: The below value is observed to be working on a lot of platforms that I tested. 
crashkernel="0M-2G:256M,2G-4G:320M,4G-8G:384M,8G-:448M"

This https://github.com/Azure/sonic-buildimage/pull/6122 implements logic to set kdump configuration according to the default configuration. Ideally, we will not require setting crashkernel value explicitly here.

build_debian.sh
shlomibitton: I agree, I have added a new file to apply these changes.

installer/x86_64/install.sh
shlomibitton: @rajendra-dendukuri you are correct, this is not required since it is handled in the PR you referenced, I removed it.

files/vmcore_dump/vmcore-sysctl.conf
shlomibitton: Fixed.

lguohan:
@rajendra-dendukuri , can you check and sign-off if possible?

rajendra-dendukuri (approved):

lguohan (approved):

lguohan:
what is the proper description of this pr? it seems it add a few more conditional to instruction when to panic? Not neccessary "Allow vmcore dump in additional situations to have better debugging information." 

is my understanding correct?

shlomibitton:
> what is the proper description of this pr? it seems it add a few more conditional to instruction when to panic? Not neccessary "Allow vmcore dump in additional situations to have better debugging information."
> 
> is my understanding correct?

@lguohan I changed the description, this is more clear now?

Issue #6072, opened 2020-12-01
SONiC system health monitor fails to restart after warm reboot.


vaibhavhd:
**Description**

SONiC systeam health monitor service restarts 21 times before giving up.

**Steps to reproduce the issue:**
1. Install latest of master - the issue is seen on 507.
2. SONiC systeam health monitor service restarts 21 times before giving up.
3. Error message is seen on syslog

**Describe the results you received:**

```
Nov 30 16:27:08.175436 str-7260cx3-acs-7 INFO systemd[1]: Started SONiC system health monitor.
Nov 30 16:27:09.105377 str-7260cx3-acs-7 NOTICE healthd[980]: Starting up...
Nov 30 16:27:09.790022 str-7260cx3-acs-7 WARNING healthd[980]: System health configuration file not found, exit...
Nov 30 16:27:09.827706 str-7260cx3-acs-7 INFO systemd[1]: system-health.service: Succeeded.
Nov 30 16:27:09.930272 str-7260cx3-acs-7 INFO systemd[1]: system-health.service: Service RestartSec=100ms expired, scheduling restart.
Nov 30 16:27:09.930462 str-7260cx3-acs-7 INFO systemd[1]: system-health.service: Scheduled restart job, restart counter is at 1.
Nov 30 16:27:09.930615 str-7260cx3-acs-7 INFO systemd[1]: Stopped SONiC system health monitor.
Nov 30 16:27:09.932198 str-7260cx3-acs-7 INFO systemd[1]: Started SONiC system health monitor.

Nov 30 16:27:25.418807 str-7260cx3-acs-7 WARNING healthd[3434]: System health configuration file not found, exit...
Nov 30 16:27:25.451956 str-7260cx3-acs-7 INFO systemd[1]: system-health.service: Succeeded.
Nov 30 16:27:25.804127 str-7260cx3-acs-7 INFO systemd[1]: system-health.service: Service RestartSec=100ms expired, scheduling restart.
Nov 30 16:27:25.804626 str-7260cx3-acs-7 INFO systemd[1]: system-health.service: Scheduled restart job, restart counter is at 21.
Nov 30 16:27:25.805026 str-7260cx3-acs-7 INFO systemd[1]: Stopped SONiC system health monitor.
Nov 30 16:27:25.805461 str-7260cx3-acs-7 WARNING systemd[1]: system-health.service: Start request repeated too quickly.
[...]

PR #6096, opened 2020-12-02
Use default values for healthd if config file is missing on the device.

vaibhavhd:
Fixes https://github.com/Azure/sonic-buildimage/issues/6072
Use default values for health monitoring config if config file is missing in the device.
Some platforms have the health monitoring config: https://github.com/Azure/sonic-buildimage/blob/master/device/mellanox/x86_64-mlnx_msn2700-r0/system_health_monitoring_config.json
While others don't have it:
https://github.com/Azure/sonic-buildimage/tree/master/device/dell/x86_64-dell_s6100_c2538-r0

**- Why I did it**
To fix the error seen during starting SONiC system health monitor
```
Nov 30 16:27:25.418807 str-7260cx3-acs-7 WARNING healthd[3434]: System health configuration file not found, exit...
Nov 30 16:27:25.451956 str-7260cx3-acs-7 INFO systemd[1]: system-health.service: Succeeded.
Nov 30 16:27:25.804127 str-7260cx3-acs-7 INFO systemd[1]: system-health.service: Service RestartSec=100ms expired, scheduling restart.
Nov 30 16:27:25.804626 str-7260cx3-acs-7 INFO systemd[1]: system-health.service: Scheduled restart job, restart counter is at 21.
Nov 30 16:27:25.805026 str-7260cx3-acs-7 INFO systemd[1]: Stopped SONiC system health monitor.
Nov 30 16:27:25.805461 str-7260cx3-acs-7 WARNING systemd[1]: system-health.service: Start request repeated too quickly.
Nov 30 16:27:25.805786 str-7260cx3-acs-7 WARNING systemd[1]: system-health.service: Failed with result 'start-limit-hit'.
Nov 30 16:27:25.806091 str-7260cx3-acs-7 ERR systemd[1]: Failed to start SONiC system health monitor.
```

**- How I did it**
Removed the return value, and let the infinite loop continue with default values for the config.

**- How to verify it**
Added the fix locally in the Python file on the device.
Before fix:
```
root@str-s6100-acs-4:~# systemctl status system-health
â— system-health.service - SONiC system health monitor
   Loaded: loaded (/lib/systemd/system/system-health.service; enabled-runtime; vendor preset: enabled)
   Active: failed (Result: start-limit-hit) since Wed 2020-12-02 09:12:18 UTC; 2min 21s ago
  Process: 4027 ExecStart=/usr/local/bin/healthd (code=exited, status=0/SUCCESS)
 Main PID: 4027 (code=exited, status=0/SUCCESS)

Dec 02 09:12:18 str-s6100-acs-4 systemd[1]: system-health.service: Service RestartSec=100ms expired, scheduling restart.
Dec 02 09:12:18 str-s6100-acs-4 systemd[1]: system-health.service: Scheduled restart job, restart counter is at 16.
Dec 02 09:12:18 str-s6100-acs-4 systemd[1]: Stopped SONiC system health monitor.
Dec 02 09:12:18 str-s6100-acs-4 systemd[1]: system-health.service: Start request repeated too quickly.
Dec 02 09:12:18 str-s6100-acs-4 systemd[1]: system-health.service: Failed with result 'start-limit-hit'.
Dec 02 09:12:18 str-s6100-acs-4 systemd[1]: Failed to start SONiC system health monitor.
root@str-s6100-acs-4:~# 
```
After fix:
```
root@str-s6100-acs-4:~# systemctl status system-health
â— system-health.service - SONiC system health monitor
   Loaded: loaded (/lib/systemd/system/system-health.service; enabled-runtime; vendor preset: enabled)
   Active: active (running) since Wed 2020-12-02 09:17:22 UTC; 1min 16s ago
 Main PID: 1391 (healthd)
    Tasks: 1 (limit: 4915)
   Memory: 43.2M
   CGroup: /system.slice/system-health.service
           â””â”€1391 /usr/bin/python /usr/local/bin/healthd

Dec 02 09:17:22 str-s6100-acs-4 systemd[1]: Started SONiC system health monitor.
Dec 02 09:17:23 str-s6100-acs-4 healthd[1391]: Starting up...
Dec 02 09:17:25 str-s6100-acs-4 healthd[1391]: System health configuration file not found, using default health monitoring config...
root@str-s6100-acs-4:~#
```

- [ ] 201811
- [ ] 201911
- [ ] 202006

### DIFF
--- a/src/system-health/scripts/healthd
+++ b/src/system-health/scripts/healthd
@@ -72,8 +72,7 @@ class HealthDaemon(DaemonBase):
         chassis = sonic_platform.platform.Platform().get_chassis()
         manager = HealthCheckerManager()
         if not manager.config.config_file_exists():
-            self.log_warning("System health configuration file not found, exit...")
-            return
+            self.log_warning("System health configuration file not found, using default health monitoring config...")
         while 1:
             state, stat = manager.check(chassis)
             if state == HealthCheckerManager.STATE_RUNNING:


jleveque:
@Junchao-Mellanox: Please review.

Staphylo:
FWIW I tested your change on an Arista platform and the system-health service now runs properly.
However you should expect the service to set the status led to red on our platforms for the time being.
We are still missing thermals on the PsuBase objects in the platform API.
Also Broadcom ASICs don't report their temperature in STATE_DB which is a second reason why the health check would fail.

On the other hand we have a status led state machine in our platform library that will set the led to green every 60s. So you should expect the led to go from red to green until we have time to address this.

vaibhavhd:
Retest baseimage please

Junchao-Mellanox:
Hi @jleveque, the code change looks fine. But I remember we discussed this before. The problem is that if we use default value in healthd, the command 'show system-health summary' will treat something as error. For example, say there is one platform who does not support asic temperature, 'show system-health summary' will report something like **"Failed to get ASIC temperature"**. It might confuse user.

vaibhavhd:
> Hi @jleveque, the code change looks fine. But I remember we discussed this before. The problem is that if we use default value in healthd, the command `show system-health summary` will treat something as error. For example, say there is one platform who does not support asic temperature, `show system-health summary` will report something like **"Failed to get ASIC temperature"**. It might confuse user.

That is correct. To add to the same, below is an example:
```
root@str-s6100-acs-4:~# redis-cli -n 6 hgetall SYSTEM_HEALTH_INFO
 1) "PSU 1"
 2) "Failed to get temperature data for PSU 1"
 3) "PSU 2"
 4) "Failed to get temperature data for PSU 2"
 5) "ASIC"
 6) "Failed to get ASIC temperature"
 7) "PSU2 Fan"
 8) "PSU2 Fan is missing"
 9) "PSU1 Fan"
10) "PSU1 Fan is missing"
11) "summary"
12) "Not OK"
root@str-s6100-acs-4:~# 
```

jleveque:
I agree with @Junchao-Mellanox that some of those messages are technically incorrect and deceiving, like "PSU2 Fan is missing" and a summary of "Not OK". So we have a choice to make here:

1. Go with this approach which will cause all platforms which do not have a configuration file to appear as unhealthy until the vendor provides a configuration file for that platform
2. Live with the "Failed to start SONiC system health monitor." error message on platforms which do not have a configuration file until the vendor provides a configuration file for that platform

Any other suggestions?

Junchao-Mellanox:
Another simple way is to provide a configuration for healthd in the platform, it is simple:

```json
{
    "services_to_ignore": [],
    "devices_to_ignore": ["psu.temperature", "asic", "PSU2 Fan", "PSU1 Fan"],
    "user_defined_checkers": [],
    "polling_interval": 60,
    "led_color": {
        "fault": "amber",
        "normal": "green",
        "booting": "orange_blink"
    }
}
```

vaibhavhd:
> Another simple way is to provide a configuration for healthd in the platform, it is simple:
> 

That definitely seems simple and takes away systemd errors from syslog. Two challenges:

1. If we are disabling most devices health monitoring, we are wasting CPU cycles without any merit with that infinite loop.
2. It should be added to every platform directory with platform-specific tweaks. For example this is what Arista shows right now:

```
root@str2-7260cx3-acs-9:~# redis-cli -n 6 hgetall SYSTEM_HEALTH_INFO
1) "PSU 1"
2) "Failed to get temperature data for PSU 1"
3) "PSU 2"
4) "Failed to get temperature data for PSU 2"
5) "ASIC"
6) "Failed to get ASIC temperature"
7) "summary"
8) "Not OK"
root@str2-7260cx3-acs-9:~# 
```

When a new platform is added without a config file, the default values for config will continue to cause these failures in log and deceiving show output.
Can we disable the `system-health` service if the platform don't have the health monitoring config in the device?
This is because we may not have any global default config values that work for all platforms.

jleveque:
> Can we disable the system-health service if the platform don't have the health monitoring config in the device?

I believe via systemd we can make the service start conditionally only if the system health configuration file is present. However, this would also potentially hide the fact that a vendor's implementation is incomplete, as they would not be made aware that the service did not start because they haven't provided a config file.

jleveque:
@Staphylo: Is there anything preventing you from providing healthd configuration files for your platforms, as per Junchao's example above?

Staphylo:
@jleveque we could do it right away, though we prefer completing a few more platform API before we do so.
We need to reparent our power supply temperature sensors to the PsuBase rather than ChassisBase.

If you have a strong preference, we could always go ahead and merge configurations with an ignore list sooner.

jleveque:
> @jleveque we could do it right away, though we prefer completing a few more platform API before we do so.
> We need to reparent our power supply temperature sensors to the PsuBase rather than ChassisBase.
> 
> If you have a strong preference, we could always go ahead and merge configurations with an ignore list sooner.

@Staphylo: Do you have a rough ETA for the full completion?

Staphylo:
Probably around mid-January given the holiday season

jleveque:
@Staphylo: Given the timeframe and the holidays, would it be possible to create placeholder config files for your platforms with temporary ignore lists, and clean up the ignore lists when you complete more platform APIs before the end of the year?

Staphylo:
@jleveque sure, we can go about it this way. I will keep you posted.

Staphylo:
@jleveque I just published https://github.com/Azure/sonic-buildimage/pull/6233

PR #6100, opened 2020-12-02
[multi-asic] Enhancing monit process checker for multi-asic.

abdosi:
Why I did:
Enhanced the monit process checked to support multi-asic .

How I did:

- For given feature we get the host and network namespace instances it's processes should be running
based on it's scope (global/asic) and add it to expected set.
 
- From psutil we get number of running instances of the processes and add where it is running (host/network namespace) using "ip netns identify"  to the the actual set

- Difference between expected and actual set provides instances where the processes are not running
and will be logged as syslog message by monit

How I verify:

- O/P from Multi-asic platform

```
# Stop lldpmgrd in host lldp docker
admin@str2-xxxxx-acs-3:/var/log$ /usr/bin/process_checker lldp python /usr/bin/lldpmgrd
'python /usr/bin/lldpmgrd' is not running in host

# Stop ldpmgrd in lldp0 (asic0 namespace) docker
admin@str2-xxxx-acs-3:/var/log$ /usr/bin/process_checker lldp python /usr/bin/lldpmgrd
'python /usr/bin/lldpmgrd' is not running in host and in namespace asic0

# Stop ldpmgrd in lldp1 (asic1 namespace) docker
admin@str2-xxxxx-acs-3:/var/log$ /usr/bin/process_checker lldp python /usr/bin/lldpmgrd
'python /usr/bin/lldpmgrd' is not running in host and in namespace asic1, asic0
```
-  Verified on Single Asic platforms also.

### DIFF
--- a/files/image_config/monit/process_checker
+++ b/files/image_config/monit/process_checker
@@ -6,7 +6,7 @@ import sys
 import syslog
 
 import psutil
-from sonic_py_common import device_info, multi_asic
+from sonic_py_common import multi_asic
 import swsssdk
 
 
--- a/files/image_config/monit/process_checker
+++ b/files/image_config/monit/process_checker
@@ -1,11 +1,13 @@
 #!/usr/bin/python3
 
 import argparse
+import ast
 import sys
 import syslog
 
 import psutil
 import swsssdk
+from sonic_py_common import device_info, multi_asic
 
 
 def check_process_existence(container_name, process_cmdline):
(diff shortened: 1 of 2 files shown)


src/sonic-py-common/sonic_py_common/multi_asic.py
```diff
@@ -138,14 +138,14 @@ def get_asic_id_from_name(asic_name):
...
-def get_current_namespace():
+def get_current_namespace(pid=None):
     """
     This API returns the network namespace in which it is
     invoked. In case of global namepace the API returns None
     """
 
     net_namespace = None
-    command = ["/bin/ip netns identify", str(os.getpid())]
+    command = ["sudo /bin/ip netns identify {}".format(str(os.getpid()) if not pid else str(pid))]
```
jleveque, line 148: I don't think you need the `str()` conversions here.

abdosi: @jleveque thanks updated.  Works fine and do not need str() conversion

abdosi:
@jleveque and @yozhao101 can you please review/approve it.

yozhao101:
> @jleveque and @yozhao101 can you please review/approve it.

Will leave a comment later.

files/image_config/monit/process_checker
```diff
@@ -27,20 +29,61 @@ def check_process_existence(container_name, process_cmdline):
...
+            process_namespace_found_set = set()
+
+            has_global_scope = ast.literal_eval(feature_table[container_name].get('has_global_scope', 'True'))
+            has_per_asic_scope = ast.literal_eval(feature_table[container_name].get('has_per_asic_scope', 'False'))
+
+            if has_global_scope:
+                process_namespace_expected_set.add(multi_asic.DEFAULT_NAMESPACE)
+
+            if has_per_asic_scope:
+               process_namespace_expected_set.update(multi_asic.get_namespace_list())
```
jleveque, line 51: Is it possible for both to be True, or both to be False?

yozhao101: One more question is the value of `multi_asic.DEFAULT_NAMESPACE` is `''` string and the function `get_namespace_list()` will return a list which includes each ASIC name such as `asic1`, right?

abdosi: > Is it possible for both to be True, or both to be False?

@jleveque Yes both can be true . For now lldp is the use-case. Both can not be False there is no use-case as of now for that.

abdosi: > One more question is the value of `multi_asic.DEFAULT_NAMESPACE` is `''` string and the function `get_namespace_list()` will return a list which includes each ASIC name such as `asic1`, right?

@yozhao101 yes, this is correct.

yozhao101: Thanks @abdosi!

yozhao101: Does the line 51 have a indentation issue?

files/image_config/monit/process_checker
```diff
@@ -27,20 +29,61 @@ def check_process_existence(container_name, process_cmdline):
...
+                         process_namespace_found_set.add(multi_asic.get_current_namespace(process.info['pid']))
+                 except psutil.NoSuchProcess:
+                     pass
+
+            process_namespace_diff_set = process_namespace_expected_set.difference(process_namespace_found_set)
+
+            if not process_namespace_diff_set:
+                is_running = True
 
             if not is_running:
```
yozhao101, line 67: Probably we did not need the variable `is_running` any more since the `process_namespace_diff_set` will indicate whether the process was not running in the host or a namespace?

abdosi: @yozhao101 sure I also wanted to remove the extra flag. Updated now

files/image_config/monit/process_checker
abdosi: @yozhao101 Updated.

files/image_config/monit/process_checker
```diff
@@ -27,20 +29,56 @@ def check_process_existence(container_name, process_cmdline):
...
-                # If this script is run by Monit, then the following output will be appended to
-                # Monit's syslog message.
-                print("'{}' is not running.".format(process_cmdline))
-                sys.exit(1)
+            process_namespace_diff_set = process_namespace_expected_set.difference(process_namespace_found_set)
+
+            if process_namespace_diff_set:
+                 # If this script is run by Monit, then the following output will be appended to
+                 # Monit's syslog message.
+
```
yozhao101, line 65: It would be better to put the comments at line 63 and 64 before the line 80 since these comments are used to explain the `print(...)` statement.

abdosi: @yozhao101 Updated

yozhao101:
The changes look good to me. @jleveque Can you review it again please?

jleveque (approved):

yozhao101 (approved):

PR #6115, opened 2020-12-03
NTP service ordering

PrabhuSreenivasan:
Signed-off-by: Prabhu Sreenivasan <prabhu.sreenivasan@broadcom.com>

**- Why I did it**
Make sure ntp-config service is executed before ntpd

**- How I did it**
Updated ntp-config service files to force dependency with ntp service. Also resolved circular dependency with --no-block flag. (needed as ntp-config service internally invokes systemd to restart ntp which in turn waits for ntp-config to complete)

**- How to verify it**
Existing ntp pytest.

Adjusted the ordering of ntp-config service to run before ntp service. 
Avoided circular dependency due to above change with --no-block flag

### DIFF
--- a/files/image_config/ntp/ntp-config.service
+++ b/files/image_config/ntp/ntp-config.service
@@ -2,6 +2,7 @@
 Description=Update NTP configuration
 Requires=updategraph.service
 After=updategraph.service
+Before=ntp.service
 
 [Service]
 Type=oneshot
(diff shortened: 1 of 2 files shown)


lguohan (approved):

jleveque (approved):

jleveque:
Retest vsimage please

PR #6120, opened 2020-12-03
[Mellanox] Add phy debug tools for SN3800

noaOrMlnx:
**- Why I did it**
Add debug tools scripts for Mellanox SN3800 in order to be able to debug INPHY issues.
* gearbox_dump_rules: shows the configured rules on the vega. e.g. speed, fec, width.
* gearbox_link_status: shows the link status. e.g. signal detect, dsp ready, FW lock. 
* gearbox_mdio_check: checks if there is an mdio connection. (mdio is a protocol that helps to communicate with the vega)
* flash_protected_fix: script to help in a case flash goes to write protected mode
* gearbox_erase_eeprom: deletes the flash / the image in RAM
* gearbox_read_eeprom: reads the flash / the image in RAM

**- How I did it**
Add the scripts as part of sonic-buildimage.
**- How to verify it**
* Build an image and make sure the scripts are under /usr/bin/mlnx-phy-debug-tools
* execute the scripts - each script will have to receive the device and vega number. 
for example:
```./gearbox_link_status -d /dev/mst/<conf device> -e 0```

NOTE: In order to execute the scripts you must have root permissions. 
NOTE: If the vega number is not valid, or if you don't have vega on the switch (non-SN3800 setups), the output will show an error with accessing to vega. 

- [ ] 201811
- [x] 201911
- [x] 202006


liat-grozovik:
@noaOrMlnx suggest to close this PR. We need to think of a different way to add it.

PR #6142, opened 2020-12-07
[NVMe] Add NVMe SSD disc type support to installer.sh script

shlomibitton:

**- Why I did it**
In order to install a SONiC image on top of a NVMe SSD disc properly with ONIE we must configure it properly on the installer.sh script.

**- How I did it**
Added relevant data to 'installer.sh' script to support nvme.

**- How to verify it**
Try to install a SONiC image including this PR on a system with a NVMe disc installed and a compatible ONIE.

- [ ] 201811
- [ ] 201911
- [ ] 202006

### DIFF
--- a/installer/x86_64/install.sh
+++ b/installer/x86_64/install.sh
@@ -107,8 +107,8 @@ if [ "$install_env" != "build" ]; then
     onie_dev=$(blkid | grep ONIE-BOOT | head -n 1 | awk '{print $1}' |  sed -e 's/:.*$//')
     blk_dev=$(echo $onie_dev | sed -e 's/[1-9][0-9]*$//' | sed -e 's/\([0-9]\)\(p\)/\1/')
 
+    # check if we have an nvme device
     blk_suffix=
-    echo $blk_dev | grep -q mmcblk && blk_suffix="p"
     echo $blk_dev | grep -q nvme0 && blk_suffix="p"
 
     # Note: ONIE has no mount setting for / with device node, so below will be empty string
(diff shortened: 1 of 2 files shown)


installer/x86_64/install.sh
```diff
@@ -106,6 +106,11 @@ fi
 if [ "$install_env" != "build" ]; then
     onie_dev=$(blkid | grep ONIE-BOOT | head -n 1 | awk '{print $1}' |  sed -e 's/:.*$//')
     blk_dev=$(echo $onie_dev | sed -e 's/[1-9][0-9]*$//' | sed -e 's/\([0-9]\)\(p\)/\1/')
+
+    # check if we have an nvme device
+    blk_suffix=
+    echo $blk_dev | grep -q nvme0 && blk_suffix="p"
```
qiluo-msft, line 112: >nvme0 [](start = 28, length = 5)

Suggest grep regex `^nvme\d`

installer/x86_64/install.sh
```diff
@@ -424,6 +429,7 @@ image_dir="image-$image_version"
 if [ "$install_env" = "onie" ]; then
     eval $create_demo_partition $blk_dev
     demo_dev=$(echo $blk_dev | sed -e 's/\(mmcblk[0-9]\)/\1p/')$demo_part
+    echo $blk_dev | grep -q nvme0 && demo_dev=$(echo $blk_dev | sed -e 's/\(nvme[0-9]n[0-9]\)/\1p/')$demo_part
```
qiluo-msft, line 432: >nvme0 [](start = 28, length = 5)

Is it better to use `nvme\d` ?

qiluo-msft, line 432: >nvme[0-9]n[0-9] [](start = 76, length = 15)

Is it better to use `nvmv\d\+n\d\+` ?

qiluo-msft (commented):
As comment

installer/x86_64/install.sh
shlomibitton: $blk_dev is starting with "/dev" so the ^ is not correct in this case.
as for the "\d" it is not working in POSIX regular expressions, it is possible to use [[:digit:]]
I think the current one is better don't you think?
The same approach used in ONIE, you can check this PR:
https://github.com/opencomputeproject/onie/commit/ed5108d023d23a876767f0311a8bc43d67d34bd6

qiluo-msft (approved):

shlomibitton:
retest vs please

liat-grozovik:
retest vs please

PR #6174, opened 2020-12-10
[minigraph.py]: Remove prefix length from peer switch loopback address

theasianpianist:

**- Why I did it**
PEER_SWITCH table in config DB expects a standalone IP address w/o a prefix length
**- How I did it**

**- How to verify it**

- [ ] 201811
- [ ] 201911
- [ ] 202006

image

### DIFF
--- a/src/sonic-config-engine/minigraph.py
+++ b/src/sonic-config-engine/minigraph.py
@@ -66,8 +66,10 @@ def get_peer_switch_info(link_metadata, devices):
         if "PeerSwitch" in data:
             peer_hostname = data["PeerSwitch"]
             peer_lo_addr = devices[peer_hostname]["lo_addr"] 
+            peer_lo_addr = ipaddress.ip_network(UNICODE_TYPE(peer_lo_addr))
+
             peer_switch_table[peer_hostname] = {
-                'address_ipv4': peer_lo_addr
+                'address_ipv4': str(peer_lo_addr.network_address)
             }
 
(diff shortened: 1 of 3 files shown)


tahmed-dev (approved):

prsunny (approved):

src/sonic-config-engine/minigraph.py
```diff
@@ -66,8 +66,10 @@ def get_peer_switch_info(link_metadata, devices):
         if "PeerSwitch" in data:
             peer_hostname = data["PeerSwitch"]
             peer_lo_addr = devices[peer_hostname]["lo_addr"] 
+            peer_lo_addr = ipaddress.ip_network(UNICODE_TYPE(peer_lo_addr))
```
tahmed-dev, line 69: if the prefix change for any reason in the minigraph, this will zero out the host section of the IP address. I think it is better to check for if the prefix length is 32. if not, you strip `/<prefix len>` section. Or err out if it is not `/32`

theasianpianist: I thought that we needed to remove the prefix length completely because our config DB schema is expecting a plain address without the prefix length. @prsunny can you confirm?

prsunny: @tahmed-dev , this is ip address field, so we can safely ignore the subnet part. Ideally we expect a `/32` here, however it is still ok to extract the address part even if not `/32`

PR #6190, opened 2020-12-11
[as4630-54pe] Fix issue for management port eth2 change to eth0

ec-michael-shih:
**- Why I did it**
A model names: as4630-54pe is exited.
Our management port is using eth2, but now need to change to using eth0.
**- How I did it**
I add a service in the platform, and when bootup the system, this service will call a shell script to let eth2 to eth0.
**- How to verify it**
After the modify, I set eth0 new ipaddress in my local side:
using "ifconfig eth0 192.168.1.12"
and ping 192.168.1.37 is success。

Test log:
root@sonic:~# ifconfig eth0 192.168.1.12
root@sonic:~#
root@sonic:~# ping 192.168.1.37
PING 192.168.1.37 (192.168.1.37) 56(84) bytes of data.
64 bytes from 192.168.1.37: icmp_seq=1 ttl=128 time=2.52 ms
64 bytes from 192.168.1.37: icmp_seq=2 ttl=128 time=1.49 ms
64 bytes from 192.168.1.37: icmp_seq=3 ttl=128 time=1.92 ms

- [ ] 201811
- [x] 201911
- [x] 202006

Add to two file:
sonic-buildimage/platform/broadcom/sonic-platform-modules-accton/as4630-54pe/service/as4630-54pe-platform-handle-mgmt-interface.service
sonic-buildimage/platform/broadcom/sonic-platform-modules-accton/as4630-54pe/utils/handle_mgmt_interface.sh

### DIFF
--- a/platform/broadcom/sonic-platform-modules-accton/as4630-54pe/utils/handle_mgmt_interface.sh
+++ b/platform/broadcom/sonic-platform-modules-accton/as4630-54pe/utils/handle_mgmt_interface.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+
+#Due to the hardware design, as4630-54pe use "eth2" instead of "eth0" as management interface.
+#Rename netdev "eth0" and "eth2" to swap original "eth2" to "eth0".
+ifconfig eth0 down
+ip link set eth0 name eth3
+ip link set eth2 name eth0
+ifconfig eth0 up
--- a/platform/broadcom/sonic-platform-modules-accton/as4630-54pe/service/as4630-54pe-platform-handle-mgmt-interface.service
+++ b/platform/broadcom/sonic-platform-modules-accton/as4630-54pe/service/as4630-54pe-platform-handle-mgmt-interface.service
@@ -1,0 +1,11 @@
+[Unit]
+Description=Accton AS4630-54PE Platform handle management interface service
+After=sysinit.target
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStart=/usr/local/bin/handle_mgmt_interface.sh
+
+[Install]
+WantedBy=multi-user.target


platform/broadcom/sonic-platform-modules-accton/as4630-54pe/utils/handle_mgmt_interface.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+#Due to the hardware design, as4630-54pe use "eth2" instead of "eth0" as management interface.
+#Rename netdev "eth0" and "eth2" to swap original "eth2" to "eth0".
```
lguohan, line 4: better to use u-dev rules to name the management interface.

lguohan (changes requested):
see comments.

ec-michael-shih:
Dear lguohan:
Thank you for your suggestion,  and the method of udev is work will after the testing.
But with the current code base，I need to add the rules for udev to create the path about /etc/udev/rules.d in "rules"(common code.)
It’s a bit difficult to modify the common place in accton project. So I need this commit to merge into this project. 

BR.
Thank you

platform/broadcom/sonic-platform-modules-accton/as4630-54pe/utils/handle_mgmt_interface.sh
ec-michael-shih: Dear lguohan:
Thank you for your suggestion, and the method of udev is work will after the testing.
But with the current code base，I need to add the rules for udev to create the path about /etc/udev/rules.d in "rules"(common code.)
It’s a bit difficult to modify the common place in accton project. So I need this commit to merge into this project.

BR.
Thank you

PR #1338, opened 2018-01-25
OOM support, Sfputil not showing qsfp devices after powercycle bugfix

Balaselvi:
Added the OOM (Open optical monitoring) support. Fixed a bug on D7054 device, where qsfp devices are not listed by the sfputil or the oom after a powercycle.

Added the optoe driver and changed the init script accordingly. Added bugfix in inventec swps driver.

The oom git can be cloned on the device and the results can be seen both on D7032/D7054 device. After powercycle both sfputil/oom will show the qsfp and the sfp devices on the D7054 device.

**- A picture of a cute animal (not mandatory but encouraged)**

PR #361, opened 2017-03-02
[rc.local]: Put variables into quotes in conditionals

marian-pritsak:
Signed-off-by: marian-pritsak <marianp@mellanox.com>

PR #6194, opened 2020-12-11
[Dynamic buffer calc] Support dynamic buffer calculation

stephenxs:
**- Why I did it**
To support dynamic buffer calculation.
This PR also depends on the following PRs for sub modules
- [sonic-swss: [buffermgr/bufferorch] Support dynamic buffer calculation #1338](https://github.com/Azure/sonic-swss/pull/1338)
- [sonic-swss-common: Dynamic buffer calculation #361](https://github.com/Azure/sonic-swss-common/pull/361)
- [sonic-utilities: Support dynamic buffer calculation #973](https://github.com/Azure/sonic-utilities/pull/973)

**- How I did it**
1. Introduce field `buffer_model` in `DEVICE_METADATA|localhost` to represent which buffer model is running in the system currently:
    - `dynamic` for the dynamic buffer calculation model
    - `traditional` for the traditional model in which the `pg_profile_lookup.ini` is used
2. Add the tables required for the feature:
   - ASIC_TABLE in platform/\<vendor\>/asic_table.j2
   - PERIPHERAL_TABLE in platform/\<vendor\>/peripheral_table.j2
   - PORT_PERIPHERAL_TABLE on a per-platform basis in device/\<vendor\>/\<platform\>/port_peripheral_config.j2 for each platform with gearbox installed.
   - DEFAULT_LOSSLESS_BUFFER_PARAMETER and LOSSLESS_TRAFFIC_PATTERN in files/build_templates/buffers_config.j2
   - Add lossless PGs (3-4) for each port in files/build_templates/buffers_config.j2
3. Copy the newly introduced j2 files into the image and rendering them when the system starts
4. Update the CLI options for buffermgrd so that it can start with dynamic mode
5. Fetches the ASIC vendor name in orchagent:
   - fetch the vendor name when creates the docker and pass it as a docker environment variable
   - `buffermgrd` can use this passed-in variable
6. Clear buffer related tables from STATE_DB when swss docker starts
7. Update the src/sonic-config-engine/tests/sample_output/buffers-dell6100.json according to the buffer_config.j2
8. Remove buffer pool sizes for ingress pools and egress_lossy_pool
   Update the buffer settings for dynamic buffer calculation

**- How to verify it**


files/build_templates/buffers_config.j2
```diff
@@ -172,4 +177,16 @@ def
 {% endfor %}
     }
 {% endif %}
+,
```
qiluo-msft, line 180: Indent `,`?

qiluo-msft: Or better place it in above both `if` and `else` block.

---
In reply to: [541480018](https://github.com/Azure/sonic-buildimage/pull/6194#discussion_r541480018) [](ancestors = 541480018)

files/build_templates/docker_image_ctl.j2
```diff
@@ -83,6 +83,13 @@ function preStartAction()
     fi
 {%- elif docker_container_name == "snmp" %}
     $SONIC_DB_CLI STATE_DB HSET 'DEVICE_METADATA|localhost' chassis_serial_number $(decode-syseeprom -s)
+{%- elif docker_container_name == "swss" %}
+    if [ ! -f /etc/sonic/asic_table.json ] && [ -f /usr/share/sonic/templates/asic_table.j2 ]; then
+        sonic-cfggen -d -t /usr/share/sonic/templates/asic_table.j2 > /etc/sonic/asic_table.json
+    fi
+    if [ ! -f /etc/sonic/peripheral_table.json ] && [ -f /usr/share/sonic/device/$PLATFORM/port_peripheral_config.j2 ]; then
+        sonic-cfggen -d -t /usr/share/sonic/device/$PLATFORM/port_peripheral_config.j2 > /etc/sonic/peripheral_table.json
+    fi
```
qiluo-msft, line 92: Move this block into inside swss container?

files/build_templates/docker_image_ctl.j2
```diff
@@ -243,6 +250,11 @@ start() {
     echo "Creating new ${DOCKERNAME} container with HWSKU $HWSKU"
     {%- endif %}
 
+    {%- if docker_container_name == "swss" %}
+    # Obtain the vendor name
+    ASIC_VENDOR=`$SONIC_CFGGEN -y /etc/sonic/sonic_version.yml -v asic_type`
```
qiluo-msft, line 255: Move this into inside swss container?

files/build_templates/init_cfg.json.j2
```diff
@@ -1,6 +1,7 @@
 {
     "DEVICE_METADATA": {
         "localhost": {
+            "buffer_model": {% if default_buffer_model == "dynamic" %}"dynamic"{% else %}"traditional"{% endif %},
```
qiluo-msft, line 4: >traditional [](start = 90, length = 11)

dynamic vs static?

platform/vs/asic_table.j2
```diff
@@ -0,0 +1 @@
+../mellanox/asic_table.j2
```
qiluo-msft, line 1: Could you cut this dependency, and just use
```
    {
        "ASIC_TABLE:VS-ASIC": {
            "cell_size": "128",
            "pipeline_latency": "19",
            "mac_phy_delay": "0.8",
            "peer_response_time": "3.8"
        },
        "OP": "SET"
    }
```

files/build_templates/sonic_debian_extension.j2
```diff
@@ -392,6 +392,19 @@ sudo cp $BUILD_TEMPLATES/buffers_config.j2 $FILESYSTEM_ROOT_USR_SHARE_SONIC_TEMP
 # Copy the qos configuration template
 sudo cp $BUILD_TEMPLATES/qos_config.j2 $FILESYSTEM_ROOT_USR_SHARE_SONIC_TEMPLATES/
 
+# Copy the templates for dynamically buffer calculation
+{% if sonic_asic_platform == "mellanox" or sonic_asic_platform == "vs" %}
```
qiluo-msft, line 396: >sonic_asic_platform [](start = 43, length = 19)

What is the future expectation on other platform?
If nothing, why need it for vs?

platform/vs/peripheral_table.j2
```diff
@@ -0,0 +1 @@
+../mellanox/peripheral_table.j2
```
qiluo-msft, line 1: Could you cut this dependency?

platform/vs/docker-sonic-vs/asic_table.json
```diff
@@ -0,0 +1,11 @@
+[
+{
```
qiluo-msft, line 2: Please well format this json file.

src/sonic-config-engine/tests/sample_output/py2/buffers-dell6100.json
```diff
@@ -637,4 +637,16 @@
             "profile" : "[BUFFER_PROFILE|egress_lossy_profile]"
         }
     }
+,
```
qiluo-msft, line 640: Indent? and move to above line.

src/sonic-config-engine/tests/sample_output/py3/buffers-dell6100.json
```diff
@@ -637,4 +637,16 @@
             "profile" : "[BUFFER_PROFILE|egress_lossy_profile]"
         }
     }
+,
```
qiluo-msft, line 640: The same

src/sonic-config-engine/tests/sample_output/py2/buffers-dell6100.json
```diff
@@ -637,4 +637,16 @@
...
+    "DEFAULT_LOSSLESS_BUFFER_PARAMETER": {
```
qiluo-msft, line 641: >DEFAULT_LOSSLESS_BUFFER_PARAMETER [](start = 5, length = 33)

Is it possible to keep other vendor's file intact?

qiluo-msft (changes requested):
As comments

files/build_templates/buffers_config.j2
stephenxs: Let's add an indent.
To move it into the `if` block may introduce more complexity because
- the case the comma needs to be added to both `if` and `else` branch
- in the `if` branch, the case  `defs.generate_queue_buffers(port_names_active)` not defined should also been handled, which makes the code complex and hard to understand.

files/build_templates/docker_image_ctl.j2
stephenxs: Given that the asic table and peripheral table won't be changed on-the-fly, I'd like to have them called once only when the docker is created instead of called each time the swss docker starts.
Let me know if there are other concerns.

files/build_templates/docker_image_ctl.j2
stephenxs: The same as the previous one. I'd like call this once during docker creation instead of many times.

platform/vs/asic_table.j2
stephenxs: ok. will fix this.

files/build_templates/init_cfg.json.j2
stephenxs: `dynamic` / `traditional` is the term we already agreed on, isn't it?

files/build_templates/sonic_debian_extension.j2
stephenxs: I believe other vendors will support the dynamic model in the future.
Meanwhile, we need this to verify this feature on vs platform.
By default, the dynamic model is disabled on vs platform and is enabled when the test cases of this feature are executed.

platform/vs/peripheral_table.j2
stephenxs: ok. will fix it.

src/sonic-config-engine/tests/sample_output/py2/buffers-dell6100.json
stephenxs: will add an indent. As [it won't be moved to the if block](https://github.com/Azure/sonic-buildimage/pull/6194#discussion_r541484608), we don't need to move to the above line.

src/sonic-config-engine/tests/sample_output/py2/buffers-dell6100.json
stephenxs: ok. this needs to be done in `files/build_templates/buffers_config.j2`

platform/vs/asic_table.j2
qiluo-msft: Sorry, I mean move the section out of `platform/mellanox/asic_table.j2` into `platform/vs/asic_table.j2`
Don't mix them in one file.

---
In reply to: [541482964](https://github.com/Azure/sonic-buildimage/pull/6194#discussion_r541482964) [](ancestors = 541482964)

files/build_templates/docker_image_ctl.j2
qiluo-msft: Seems here is called when the docker swss is started, not created.
You already check the files existing, so not too much time wasted after file created.

---
In reply to: [541481038](https://github.com/Azure/sonic-buildimage/pull/6194#discussion_r541481038) [](ancestors = 541481038)

platform/vs/asic_table.j2
stephenxs: Ok. The vs part in Mellanox/asic table.j2 is no longer used and can be removed.
For the vs, we use asic table.json and asic table.j2 can be removed.

files/build_templates/docker_image_ctl.j2
stephenxs: Looks like you are correct. This snippet of code runs during swss docker starts.
But as it works correctly and to put it here doesn't introduce any negative consequence, can we keep it as it is? To move it to swss docker needs to change the code in several places.

stephenxs: ok.
Let's move to `buffermgrd.sh`. Put the code generating `asic_table.json` and `peripheral_table.json` just ahead of the place where it is used.

stephenxs: Just investigated and realized it’s will be very difficult to move this part into swss docker because it requires asic_table.j2 to be copied in docker swss. However, it’s defined on a per vendor basis and in platform/<vendor> dir, which makes it difficult to copy into docker in the Dockerfile.j2.
So we have to have this part in docker_image_control.

platform/mellanox/peripheral_table.j2
```diff
@@ -0,0 +1,20 @@
+{%- if DEVICE_METADATA is defined and DEVICE_METADATA['localhost']['platform'] is defined %}
+{%-     set platform = DEVICE_METADATA['localhost']['platform'] %}
+{%- else -%}
+{%-     set platform = "vs-platform" %}
```
qiluo-msft, line 4: The same issue, move `vs` related part into `platform/vs/peripheral_table.j2`. Please don't mix platforms into `mellanox` folder.

files/build_templates/docker_image_ctl.j2
qiluo-msft: Regarding "it’s defined on a per vendor basis and in platform/<vendor> dir, which makes it difficult to copy into docker in the Dockerfile.j2"
I checked `docker inspect swss` and found `/usr/share/sonic/device/$PLATFORM/` is already mapped into swss container.
```
            {
                "Type": "bind",
                "Source": "/usr/share/sonic/device/x86_64-mlnx_msn2700-r0",
                "Destination": "/usr/share/sonic/platform",
                "Mode": "ro",
                "RW": false,
                "Propagation": "rprivate"
            },
```

Mounting is happened after running a container, so I did not propose modify `Dockerfile.j2`, instead, modify `dockers/docker-orchagent/docker-init.sh`.

I think you can treat `asic_table.j2` the same way as `port_peripheral_config.j2`. Why they are copied into different folders?

---
In reply to: [541513109](https://github.com/Azure/sonic-buildimage/pull/6194#discussion_r541513109) [](ancestors = 541513109,541481038)

stephenxs: The port peripheral config is defined on a per platform basis because each platform has its unique configuration. But the asic_table.j2 is shared among all platforms of the same vendor. It should not be copied to each of the platform folders.

qiluo-msft (approved):

stephenxs:
retest vs please

platform/mellanox/rules.mk
```diff
@@ -31,3 +31,5 @@ $(SYNCD)_RDEPENDS += $(MLNX_SAI)
 
 # Inject mlnx sdk libs to platform monitor
 $(DOCKER_PLATFORM_MONITOR)_DEPENDS += $(APPLIBS) $(SX_COMPLIB) $(SXD_LIBS) $(SX_GEN_UTILS) $(PYTHON_SDK_API) $(APPLIBS_DEV) $(SX_COMPLIB_DEV) $(SXD_LIBS_DEV) $(SX_GEN_UTILS_DEV)
+
+export SONIC_BUFFER_MODEL=dynamic
```
qiluo-msft, line 35: Do not change build time config in component makefile, if you need, change the pipeline command line.

stephenxs: Hi Qi,
Yes, it works to change the CLI option but what's the benefit to change a working code?

PR #6196, opened 2020-12-12
[baseimage]: No need to `apt-mark manual`

qiluo-msft:
It makes no difference during build.


lguohan (approved):

build_debian.sh
```diff
@@ -516,7 +516,6 @@ sudo LANG=C DEBIAN_FRONTEND=noninteractive chroot $FILESYSTEM_ROOT apt-get -y re
 
 ## Add mtd and uboot firmware tools package
 sudo LANG=C chroot $FILESYSTEM_ROOT apt-get -y install u-boot-tools mtd-utils device-tree-compiler
-sudo LANG=C chroot $FILESYSTEM_ROOT apt-mark manual u-boot-tools mtd-utils device-tree-compiler
```
lguohan, line 519: wonder if it make difference in arm build, can you check who added this line?

Issue #6183, opened 2020-12-10
Orchagent crash in warm-reboot upgrade between master kvm images


shi-su:
**Description**
Orchagent crash in warm-reboot upgrade between master kvm images

**Steps to reproduce the issue:**
1. Install new sonic kvm image from master branch using sonic-installer
2. Warm reboot
3.

**Describe the results you received:**
```
Dec 10 18:50:56.048499 vlab-01 ERR swss#orchagent: :- removeDefaultBridgePorts: Failed to get bridge port list in default 1Q bridge, rv:-8
Dec 10 18:50:56.049958 vlab-01 INFO swss#/supervisord: orchagent terminate called after throwing an instance of 'std::runtime_error'
Dec 10 18:50:56.052166 vlab-01 INFO swss#/supervisord: orchagent   what():  PortsOrch initialization failure
```

**Describe the results you expected:**
Expect the image to boot up without any problem

**Additional information you deem important (e.g. issue happens only occasionally):**

    
    **Output of `show version`:**

```
admin@vlab-01:~$ show version

SONiC Software Version: SONiC.master.579-3e717210
Distribution: Debian 10.7
Kernel: 4.19.0-9-2-amd64
Build commit: 3e717210
Build date: Mon Dec  7 07:06:57 UTC 2020
Built by: johnar@jenkins-worker-11

Platform: x86_64-kvm_x86_64-r0
HwSKU: Force10-S6000
ASIC: vs
ASIC Count: 1
Serial Number: 000000
Uptime: 21:38:13 up  2:48,  1 user,  load average: 0.22, 0.53, 0.63

Docker images:
REPOSITORY                    TAG                   IMAGE ID            SIZE
docker-gbsyncd-vs             latest                6d8adb4ba152        453MB
docker-gbsyncd-vs             master.579-3e717210   6d8adb4ba152        453MB
docker-syncd-vs               latest                d55558665164        453MB
docker-syncd-vs               master.579-3e717210   d55558665164        453MB
docker-teamd                  latest                8260e7155f83        491MB
docker-teamd                  master.579-3e717210   8260e7155f83        491MB
docker-nat                    latest                725c18f57a4b        494MB
docker-nat                    master.579-3e717210   725c18f57a4b        494MB
docker-router-advertiser      latest                1c9d78360f97        448MB
docker-router-advertiser      master.579-3e717210   1c9d78360f97        448MB
docker-platform-monitor       latest                c6e3a68ed523        572MB
[...]

PR #6198, opened 2020-12-13
[warm-reboot] Remove warmboot file path that overrides the default path

shi-su:
**- Why I did it**
The sai.profile file in kvm images overrides the warmboot file with path `/var/cache/sai_warmboot.bin`. Since the directory `/var/cache` is not mounted in syncd, it will be cleared in an image upgrade, the warm-reboot image upgrade will fail if the file is put in the directory.

Fix https://github.com/Azure/sonic-buildimage/issues/6183

**- How I did it**
Remove the path that overrides the default path. The warmboot file path will then be the default value `/var/warmboot/sai-warmboot.bin`. Since `/var/warmboot/` is mounted by `/host/warmboot/` in the host, it could survive an image upgrade.

**- How to verify it**
Tested warm reboot upgrading kvm image locally.

- [ ] 201811
- [ ] 201911
- [ ] 202006


lguohan (approved):

kcudnik:
@shi-su @lguohan Just to clarify, if we have right now in production image running this profile, it will write on warm shutdown to /var/cache but new image with this PR will not pick up that file, since it will use default value which is "/var/warmboot/sai-warmboot.bin"

kcudnik (approved):

shi-su:
> @shi-su @lguohan Just to clarify, if we have right now in production image running this profile, it will write on warm shutdown to /var/cache but new image with this PR will not pick up that file, since it will use default value which is "/var/warmboot/sai-warmboot.bin"
> 
> from the other hand, i see that this changes are only for VS ?

I only found this profile for VS, real devices should just use the default path. So devices in production should not be affected.

qiluo-msft (approved):

PR #6251, opened 2020-12-18
[Monit] Monitoring the running status of containers.

yozhao101:
**- Why I did it**
This PR aims to monitor the running status of each container. Currently the auto-restart feature was enabled. If a critical process exited unexpected, the container will be restarted. If the container was restarted 3 times during 20 minutes, then it will not run anymore unless we cleared the flag using the command `sudo systemctl reset-failed <container_name>` manually. 

**- How I did it**
We will employ Monit to monitor a script. This script will generate the expected running container list and compare it with the current running containers. If there are containers which were expected to run but were not running, then an alerting message will be written into syslog.

**- How to verify it**
I tested this feature on a lab device `str-a7050-acs-3` which has single ASIC and `str2-n3164-acs-3` which has a Multi-ASIC. First I manually stopped a container by running the command `sudo systemctl stop <container_name>`, then I checked whether there was an alerting message in the syslog.

- [ ] 201811
- [ ] 201911
- [ x] 202006

### DIFF
--- a/files/image_config/monit/container_checker
+++ b/files/image_config/monit/container_checker
@@ -5,6 +5,7 @@ import sys
 
 import swsssdk
 from sonic_py_common import multi_asic
+from sonic_py_common.logger import Logger
 
 SYSLOG_IDENTIFIER = "container_checker"
 
@@ -94,7 +95,6 @@ def main():
     current_running_containers = get_current_running_containers(logger)
 
     not_running_containers = expected_running_containers.difference(current_running_containers)
-
     if not_running_containers:
         logger.log_error("Containers not running: " + ", ".join(not_running_containers))
         sys.exit(4)
--- a/files/image_config/monit/container_checker
+++ b/files/image_config/monit/container_checker
@@ -4,11 +4,12 @@
 container_checker
 
 This script is intended to be run by Monit. It will write an alerting message into
-syslog if it found containers which were expected to run but were not running. Note
-that if print(...) statement in this script was executed, the string in it will be 
-appended to Monit syslog messages.
+syslog if it found containers which were expected to run but were not running. At
+the same time, if some containers were unexpected to run, it also writes an alerting
+syslog message. Note that if print(...) statement in this script was executed, the
+string in it will be appended to Monit syslog messages.
 
-The following is an example in Monit configuration file to show how Monit will run 
+The following is an example in Monit configuration file to show how Monit will run
 this script:
 
 check program container_checker with path "/usr/bin/container_checker"
--- a/files/image_config/monit/container_checker
+++ b/files/image_config/monit/container_checker
@@ -11,7 +11,7 @@ The following is an example in Monit configuration file to show how Monit will r
 this script:
 
 check program container_checker with path "/usr/bin/container_checker"
-    if status != 0 for 5 times within 5 cycles then alert repeat every
+    if status != 0 for 5 times within 5 cycles then alert repeat every 1 cycles
 
 """
 import subprocess
--- a/files/image_config/monit/container_checker
+++ b/files/image_config/monit/container_checker
@@ -42,7 +42,7 @@ def get_command_result(command):
         print("Failed to execute the command '{}'. Error: '{}'".format(command, err))
         sys.exit(2)
 
-    return command_stdout.split("\n")[:-1]
+    return command_stdout.rstrip().split("\n")
 
 
 def get_expected_running_containers():
--- a/files/image_config/monit/container_checker
+++ b/files/image_config/monit/container_checker
@@ -5,11 +5,6 @@ import sys
 
 import swsssdk
 from sonic_py_common import multi_asic
-from sonic_py_common.logger import Logger
-
-SYSLOG_IDENTIFIER = "container_checker"
-logger = Logger(SYSLOG_IDENTIFIER)
-logger.set_min_log_priority_info()
 
 
 def get_command_result(command):
@@ -24,14 +19,14 @@ def get_command_result(command):
                                          shell=True, universal_newlines=True)
         command_stdout, command_stderr = proc_instance.communicate()
         if proc_instance.returncode != 0:
-            logger.log_error("Failed to execute the command '{}'. Return code: '{}'".format(
+            print("Failed to execute the command '{}'. Return code: '{}'".format(
                 command, proc_instance.returncode))
             sys.exit(1)
     except OSError as os_err:
-        logger.log_error("Failed to execute the command '{}'. Error: '{}'".format(command, os_err))
+        print("Failed to execute the command '{}'. Error: '{}'".format(command, os_err))
         sys.exit(2)
     except ValueError as val_err:
-        logger.log_error("Failed to execute the command '{}'. Error: '{}'".format(command, val_err))
+        print("Failed to execute the command '{}'. Error: '{}'".format(command, val_err))
         sys.exit(3)
 
     return command_stdout.split("\n")
@@ -96,7 +91,7 @@ def main():
 
     not_running_containers = expected_running_containers.difference(current_running_containers)
     if not_running_containers:
-        logger.log_error("Containers not running: " + ", ".join(not_running_containers))
+        print("Containers not running: " + ", ".join(not_running_containers))
         sys.exit(4)
 
 
--- a/files/image_config/monit/container_checker
+++ b/files/image_config/monit/container_checker
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 
-import sys
 import subprocess
+import sys
 
 import swsssdk
 from sonic_py_common import multi_asic
--- a/files/image_config/monit/container_checker
+++ b/files/image_config/monit/container_checker
@@ -24,7 +24,7 @@ from sonic_py_common import multi_asic
 
 def get_command_result(command):
     """
-    @summary: This function will execture the command and return the result.
+    @summary: This function will execute the command and return the resulting output.
     @return: A string which contains the output of command.
     """
     command_stdout = ""
@@ -37,12 +37,9 @@ def get_command_result(command):
             print("Failed to execute the command '{}'. Return code: '{}'".format(
                 command, proc_instance.returncode))
             sys.exit(1)
-    except OSError as os_err:
-        print("Failed to execute the command '{}'. Error: '{}'".format(command, os_err))
+    except (OSError, ValueError) as err:
+        print("Failed to execute the command '{}'. Error: '{}'".format(command, err))
         sys.exit(2)
-    except ValueError as val_err:
-        print("Failed to execute the command '{}'. Error: '{}'".format(command, val_err))
-        sys.exit(3)
 
     return command_stdout.split("\n")
 
@@ -107,6 +104,11 @@ def main():
     not_running_containers = expected_running_containers.difference(current_running_containers)
     if not_running_containers:
         print("Containers not running: " + ", ".join(not_running_containers))
+        sys.exit(3)
+
+    unexpected_running_containers = current_running_containers.difference(expected_running_containers)
+    if unexpected_running_containers:
+        print("Unexpected running containers: " + ", ".join(unexpected_running_containers))
         sys.exit(4)
 
 
--- a/files/image_config/monit/container_checker
+++ b/files/image_config/monit/container_checker
@@ -8,8 +8,11 @@ from sonic_py_common import multi_asic
 from sonic_py_common.logger import Logger
 
 SYSLOG_IDENTIFIER = "container_checker"
+logger = Logger(SYSLOG_IDENTIFIER)
+logger.set_min_log_priority_info()
 
-def get_command_result(command, logger):
+
+def get_command_result(command):
     """
     @summary: This function will execture the command and return the result.
     @return: A string which contains the output of command.
@@ -65,7 +68,7 @@ def get_expected_running_containers():
     return expected_running_containers
 
 
-def get_current_running_containers(logger):
+def get_current_running_containers():
     """
     @summary: This function will get the current running container list by analyzing the
               output of command `docker ps`.
@@ -74,7 +77,7 @@ def get_current_running_containers(logger):
     running_containers = set()
 
     command = "docker ps"
-    command_stdout = get_command_result(command, logger)
+    command_stdout = get_command_result(command)
     for line in command_stdout:
         if line and "CONTAINER ID" not in line:
             running_containers.add(line.split()[-1].strip())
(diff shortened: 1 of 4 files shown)


files/image_config/monit/monitoring_containers
```diff
@@ -0,0 +1,116 @@
+#!/usr/bin/python
```
jleveque, line 1: We are migrating SONiC from Python 2 to Python 3. Please change this shebang to `#!/usr/bin/env python3` and make sure your script is Python 3-compliant.

files/image_config/monit/monitoring_containers
```diff
@@ -0,0 +1,116 @@
...
+
+import sys
+import subprocess
+
+import swsssdk
+
```
jleveque, line 7: Delete this blank line, grouping swsssdk and sonic_py_common together, as both are third-party libraries in relation to this script.

files/image_config/monit/monitoring_containers
```diff
@@ -0,0 +1,116 @@
...
+        if len(not_running_containers) > 1:
+            print("Containers '{}' were not running.".format(log_message))
+        else:
+            print("Container '{}' was not running.".format(log_message))
+
+        sys.exit(4)
+
+
+def main():
+    monitoring_containers_and_alerting()
```
jleveque, line 112: This is pointless. Suggest removing this and simply renaming `monitoring_containers_and_alerting()` to `main()`

jleveque (changes requested):
As comments.

Also, I suggest renaming the `monitoring_containers` script to `container_checker`

files/image_config/monit/monitoring_containers
yozhao101: Updated! I tested the script with Python 3 on DuTs (single ASIC and multiple ASICs).

files/image_config/monit/monitoring_containers
yozhao101: Great suggestions! Updated.

files/image_config/monit/monitoring_containers
yozhao101: Great suggestions! Updated.

files/image_config/monit/container_checker
```diff
@@ -0,0 +1,111 @@
+#!/usr/bin/env python3
+
+import sys
+import subprocess
```
jleveque, line 4: Organize alphabetically.

jleveque (changes requested):

files/image_config/monit/container_checker
yozhao101: Updated!

files/image_config/monit/container_checker
```diff
@@ -0,0 +1,111 @@
...
+                log_message = container_name
+            else:
+                log_message += ", " + container_name
+
+        if len(not_running_containers) > 1:
+            print("Containers '{}' were not running.".format(log_message))
+        else:
+            print("Container '{}' was not running.".format(log_message))
+
+        sys.exit(4)
```
jleveque, line 107: This can be greatly simplified to something like:

```python3
    if not_running_containers:
        print("Containers not running: " + ", ".join(not_running_containers))
        sys.exit(4)
```

Also, suggest logging to syslog rather than simply printing to stdout. You can use the Logger class in the sonic_py_common package. There are multiple examples in the codebase.

jleveque (changes requested):

files/image_config/monit/container_checker
yozhao101: Updated! The string in the `print(...)` statement will be appended to Monit syslog message. But I like idea to simplify the logic about generating the alerting message.

jleveque: So the string in the `print(...)` statement will get appended to the monit syslog message? Is that the behavior we want? If so, then maybe we shouldn't use the Logger class. It just depends if we expect the message to be logged by Monit or not.

yozhao101: Actually in the script `process_checker` we used the `print(...)` statement to write the alerting messages into syslog. 

Log messages were generated by `print(...)` statement:
`Jan  5 02:49:39.432635 str-a7050-acs-3 ERR monit[19018]: 'container_checker' status failed (4) -- Containers not running: lldp, pmon`
`Jan  5 02:50:39.917866 str-a7050-acs-3 ERR monit[19018]: 'container_checker' status failed (4) -- Containers not running: pmon, lldp`
`Jan  5 02:51:40.413938 str-a7050-acs-3 ERR monit[19018]: 'container_checker' status failed (4) -- Containers not running: lldp, pmon`
`Jan  5 02:52:40.948582 str-a7050-acs-3 ERR monit[19018]: 'container_checker' status failed (4) -- Containers not running: pmon, lldp`
`Jan  5 02:53:41.420400 str-a7050-acs-3 ERR monit[19018]: 'container_checker' status failed (4) -- Containers not running: pmon, lldp`

Log messages were generated by Logger class:
`Jan  5 02:58:43.926697 str-a7050-acs-3 ERR monit[19018]: 'container_checker' status failed (4) -- no output`
`Jan  5 02:58:46.635607 str-a7050-acs-3 ERR container_checker.py: Containers not running: pmon, lldp`
`Jan  5 02:59:44.390015 str-a7050-acs-3 ERR monit[19018]: 'container_checker' status failed (4) -- no output`
`Jan  5 02:59:47.194435 str-a7050-acs-3 ERR container_checker: Containers not running: lldp, pmon`
`Jan  5 03:00:44.863422 str-a7050-acs-3 ERR monit[19018]: 'container_checker' status failed (4) -- no output`
`Jan  5 03:00:47.784087 str-a7050-acs-3 ERR container_checker: Containers not running: lldp, pmon`
`Jan  5 03:01:45.394784 str-a7050-acs-3 ERR monit[19018]: 'container_checker' status failed (4) -- no output`
`Jan  5 03:01:48.015010 str-a7050-acs-3 ERR container_checker: Containers not running: pmon, lldp`

jleveque: Hmm. Then maybe we should revert back to the `print(...)` statements to consolidate all the output into the same line and also prevent the `no output` messages.

yozhao101: I agree and will revert back to use the `print(...)` statement.

yozhao101: Updated!

files/image_config/monit/container_checker
```diff
@@ -0,0 +1,99 @@
+#!/usr/bin/env python3
+
```
jleveque, line 2: Can you add a header comment/docstring here describing that this script is intended to be run by Monit, what it does and note that all `print(...)` statements will be appended to the Monit log messages? Something like,

```python3
"""
container_checker 

This script is intended to be run by Monit ...
"""
```

Maybe also add an example of how to run it in a Monit config file?

jleveque (changes requested):

files/image_config/monit/container_checker
```diff
@@ -0,0 +1,99 @@
...
+    """
+    @summary: This function will compare the difference between the current running containers
+              and the containers which were expected to run. If containers which were exepcted
+              to run were not running, then an alerting message will be written into syslog.
+    """
+    expected_running_containers = get_expected_running_containers()
+    current_running_containers = get_current_running_containers()
+
+    not_running_containers = expected_running_containers.difference(current_running_containers)
+    if not_running_containers:
```
abdosi, line 93: can we enhanced to this to see if we are not running extra containers. For eg:  in multi-asic we expect bgp0 and bgp1 but if docker ps O/P return bgp0, bgp1, bgp2. Can we catch this scenario also ?

files/image_config/monit/container_checker
```diff
@@ -0,0 +1,113 @@
...
+that if print(...) statement in this script was executed, the string in it will be 
+appended to Monit syslog messages.
+
+The following is an example in Monit configuration file to show how Monit will run 
+this script:
+
+check program container_checker with path "/usr/bin/container_checker"
+    if status != 0 for 5 times within 5 cycles then alert repeat every
+
+"""
```
jleveque, line 16: Please add a blank line above and below this docstring.

jleveque (changes requested):

files/image_config/monit/container_checker
yozhao101: Great suggestion, Updated!

files/image_config/monit/container_checker
```diff
@@ -0,0 +1,114 @@
...
+import subprocess
+import sys
+
+import swsssdk
+from sonic_py_common import multi_asic
+
+
+def get_command_result(command):
+    """
+    @summary: This function will execture the command and return the result.
```
jleveque, line 27: Typo: s/execture/execute/

Also suggest replacing "result" with "resulting output"

files/image_config/monit/container_checker
```diff
@@ -0,0 +1,114 @@
...
+        if proc_instance.returncode != 0:
+            print("Failed to execute the command '{}'. Return code: '{}'".format(
+                command, proc_instance.returncode))
+            sys.exit(1)
+    except OSError as os_err:
+        print("Failed to execute the command '{}'. Error: '{}'".format(command, os_err))
+        sys.exit(2)
+    except ValueError as val_err:
+        print("Failed to execute the command '{}'. Error: '{}'".format(command, val_err))
+        sys.exit(3)
```
jleveque, line 45: Suggest simplifying this, since the behavior is essentially the same. E.g.,

```python3
except (OSError, ValueError) as err:
        print("Failed to execute the command '{}'. Error: '{}'".format(command, err))
        sys.exit(2)
```

jleveque (changes requested):

files/image_config/monit/container_checker
yozhao101: You mean that initially the expected running containers for example BGP container were BGP0 and BGP1, However, BGP0, BGP1, BGP2 and BGP3 are actually running. We should write an syslog message to show that extra containers BGP2 and BGP3 are running, right?

files/image_config/monit/container_checker
```diff
@@ -0,0 +1,117 @@
...
+
+    not_running_containers = expected_running_containers.difference(current_running_containers)
+    if not_running_containers:
+        print("Containers not running: " + ", ".join(not_running_containers))
+        sys.exit(3)
+
+    unexpected_running_containers = current_running_containers.difference(expected_running_containers)
+    if unexpected_running_containers:
+        print("Unexpected running containers: " + ", ".join(unexpected_running_containers))
+        sys.exit(4)
```
yozhao101, line 113: @abdosi @jleveque Please review this code segment which was used to write a syslog message showing the unexpected running containers.

files/image_config/monit/container_checker
```diff
@@ -0,0 +1,117 @@
...
+              output of command `docker ps`.
+    @return:  A set which contains the current running contianers.
+    """
+    running_containers = set()
+
+    command = "docker ps"
+    command_stdout = get_command_result(command)
+    for line in command_stdout:
+        if line and "CONTAINER ID" not in line:
+            running_containers.add(line.split()[-1].strip())
```
jleveque, line 91: `"CONTAINER ID"` is only found in the header line. It might be a bit more straightforward to simply ignore the first line, like:

```python

    # Ignore the first line of output which contains the headings
    for line in command_stdout[1:]:
        running_containers.add(line.split()[-1].strip())
```

jleveque (changes requested):

files/image_config/monit/container_checker
```diff
@@ -0,0 +1,116 @@
...
+        command_stdout, command_stderr = proc_instance.communicate()
+        if proc_instance.returncode != 0:
+            print("Failed to execute the command '{}'. Return code: '{}'".format(
+                command, proc_instance.returncode))
+            sys.exit(1)
+    except (OSError, ValueError) as err:
+        print("Failed to execute the command '{}'. Error: '{}'".format(command, err))
+        sys.exit(2)
+
+    return command_stdout.split("\n")[:-1]
```
jleveque, line 45: Why remove the last line here (`[:-1]`)?

jleveque: I think it makes more sense to `.rstrip()` before splitting rather than blindly delete the last line. This is written as a general-purpose function. Not all command outputs may end with a blank line, in which case you would lose the last line of output.

yozhao101: Great suggestion!

jleveque: My point was that not all command output is guaranteed to end in a `\n`; in those cases `command_stdout.split("\n")[:-1]` would lose the last line.

files/image_config/monit/container_checker
jleveque: You removed the "CONTAINER ID" check but did not add the `[1:]` to `command_stdout`, so you are no longer skipping the header line.

files/image_config/monit/container_checker
yozhao101: Ohh, Got it. Thanks!

files/image_config/monit/container_checker
yozhao101: Sorry, I missed that after testing. Updated!

files/image_config/monit/container_checker
```diff
@@ -0,0 +1,116 @@
...
+    @summary: This function will compare the difference between the current running containers
+              and the containers which were expected to run. If containers which were exepcted
+              to run were not running, then an alerting message will be written into syslog.
+    """
+    expected_running_containers = get_expected_running_containers()
+    current_running_containers = get_current_running_containers()
+
+    not_running_containers = expected_running_containers.difference(current_running_containers)
+    if not_running_containers:
+        print("Containers not running: " + ", ".join(not_running_containers))
```
abdosi, line 106: wondering should we start with Expected container not running:.
This would make it consistent with the below message

files/image_config/monit/container_checker
```diff
@@ -0,0 +1,116 @@
...
+
+def get_expected_running_containers():
+    """
+    @summary: This function will get the expected running containers by following the rule:
+              The 'state' field of container in 'FEATURE' table should not be 'disabled'. Then
+              if the device has Multi-ASIC, this function will get container list by determining the
+              value of field 'has_global_scope', the number of ASICs and the value of field
+              'has_per_asic_scope'. If the device has single ASIC, the container name was put into
+              the list.
+    @return:  A set which contains the expected running containers.
```
abdosi, line 56: need to update this description

files/image_config/monit/container_checker
abdosi: yes

files/image_config/monit/container_checker
abdosi: please update this

abdosi (approved):
LGTM

jleveque (approved):

PR #6285, opened 2020-12-23
Take a copy of existing TACACS credentials and restore it during upgrade

renukamanavalan:
In scenario where upgrade gets config from minigraph, it could miss tacacs credentials as they are not in minigraph. Hence restore explicitly upon load-minigraph, if  present.

**- Why I did it**
Upon boot, when config migration is required, the switch could load config from minigraph. The config-load from minigraph would wipe off TACACS key and disable login via TACACS, which would disable all remote user access. This change, would re-configure the TACACS if there is a saved copy available.

**- How I did it**
When config is loaded from minigraph, look for a TACACS credentials back up (tacacs.json) under /etc/sonic/old_config. If present, load the credentials into running config, before config-save is called.

**- How to verify it**
Remove /etc/sonic/config_db.json and do an image update. Upon reboot, w/o this change, you would not be able ssh in as remote user. You may login as admin and check out, "show tacacs" & "show aaa" to verify that tacacs-key is missing and login is not enabled for tacacs.
With this change applied, remove /etc/sonic/config_db.json, but save tacacs & aaa credentials as tacacs.json in /etc/sonic/. Upon reboot, you should see remote user access possible.

- [x] 201811
- [x] 201911
- [x] 202006
- [x] 202012

### DIFF
--- a/files/image_config/config-setup/config-setup
+++ b/files/image_config/config-setup/config-setup
@@ -362,7 +362,6 @@ do_config_backup()
     echo "Taking backup of curent configuration"
     rm -rf /host/old_config
     cp -ar /etc/sonic /host/old_config
-
     [ -d ${CONFIG_POST_MIGRATION_HOOKS} ] && cp -arL ${CONFIG_POST_MIGRATION_HOOKS} /host/old_config
 
     # Execute custom hooks if present
--- a/files/image_config/config-setup/config-setup
+++ b/files/image_config/config-setup/config-setup
@@ -363,9 +363,6 @@ do_config_backup()
     rm -rf /host/old_config
     cp -ar /etc/sonic /host/old_config
 
-    # Save TACACS
-    sonic-cfggen -d --preset tac > /host/old_config/${TACACS_JSON_BACKUP}
-
     [ -d ${CONFIG_POST_MIGRATION_HOOKS} ] && cp -arL ${CONFIG_POST_MIGRATION_HOOKS} /host/old_config
 
     # Execute custom hooks if present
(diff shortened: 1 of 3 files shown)


files/image_config/config-setup/config-setup
```diff
@@ -355,6 +362,10 @@ do_config_backup()
     echo "Taking backup of curent configuration"
     rm -rf /host/old_config
     cp -ar /etc/sonic /host/old_config
+
+    # Save TACACS
+    sonic-cfggen -d --preset tac > /host/old_config/${TACACS_JSON_BACKUP}
```
lguohan, line 367: i am not sure if we need this as the config_db has the tacacs cofiguration.

renukamanavalan: dropped

lguohan:
@renukamanavalan , can you fill the questionaire?

renukamanavalan:
> @renukamanavalan , can you fill the questionaire?

Done

files/image_config/config-setup/config-setup
```diff
@@ -108,6 +110,11 @@ reload_minigraph()
 {
     echo "Reloading minigraph..."
     config load_minigraph -y -n 
+    if [ -f /etc/sonic/old_config/${TACACS_JSON_BACKUP} ]; then
```
yxieca, line 113: Minor improvement: you could test -r instead of -f.

yxieca (approved):

renukamanavalan:
retest vs please

renukamanavalan:
retest vsimage please

renukamanavalan:
retest vsimage please

yxieca:
@renukamanavalan can you open an PR for 201811 branch? This change cannot be cherry-picked cleanly.

PR #6298, opened 2020-12-25
Enable rsyslogd notify mode for supervisord logs in docker container

abdosi:
What I did:
Enable the notify mode of rsyslogd imfile module used for supervisord logs in docker container

How I did:
As suggested in https://rsyslog-doc.readthedocs.io/en/latest/configuration/modules/imfile.html updated to use "innotify mode" which is by default when using new semantics of imfile parameters.

Why I did:
Without this change because of polling mode the supervisord logs in docker container could take upto 10 seconds (Default Polling Interval) to get logged in host. 

This change will fix the issue where any critical process gets exited the supervisord log for same was not coming in host syslog immediately.

Below is one of  the sequence to hit the issue:

•	Telemetry Service entered failed state during boot up as no certificate present .
•	After sometime certificates are pushed and makes telemetry feature disable/enable 
•	When we enable the telemetry feature again the supervisor message that got buffered when it entered failed state in 
        previous iteration comes now. See the Below Log for difference in time in host log (Dec 25 01:33:46) vs when it got 
        generated in docker (2020-12-25 01:29:29)

`
Dec 25 01:33:46.128364 str-xxxx-on-2 INFO telemetry#supervisord 2020-12-25 01:29:29,686 INFO exited: telemetry (exit status 1; not expected)
`

How I Verify:

After the fix the syslog message is coming immediately.

`
Dec 30 00:07:50.406483 str-xxxx-on-2 INFO telemetry#supervisord 2020-12-30 00:07:50,404 INFO exited: telemetry (exit status 1; not expected)
`


yxieca (approved):

dockers/docker-base/etc/rsyslog.d/supervisor.conf
```diff
@@ -1,9 +1,9 @@
-$ModLoad imfile
-
-$InputFileName /var/log/supervisor/supervisord.log
-$InputFileTag supervisord
-$InputFileStateFile state-supervisor
-$InputFileSeverity info
-$InputFileFacility local0
-$InputFilePersistStateInterval 1
-$InputRunFileMonitor
+module(load="imfile") #By default notify mode
```
hui-ma, line 1: could you please explain why this file format/style is changed? And why did the previous specify polling mode? Not seeing it in the config. It is default in the old format as well? Thanks!

abdosi:
Retest mellanox please

dockers/docker-base/etc/rsyslog.d/supervisor.conf
abdosi: @hui-ma yes old semantics of rsyslogd default to polling mode. Rsyslogd document recommends to use notify mode and new semantic going forward

hui-ma: Thanks!

hui-ma (approved):

abdosi:
retest mellanox please

dockers/docker-base/etc/rsyslog.d/supervisor.conf
jleveque: Sorry for the late reply due to the holidays, but I wonder if we should explicitly set `mode="inotify"` just in case the default mode were ever to change.

abdosi: Thanks @jleveque . PR https://github.com/Azure/sonic-buildimage/pull/6351 has the change.

PR #6058, opened 2020-11-29
[master][build] fix compile error when enable notrip

tim-rj:
Signed-off-by: tim-rj <sonic_rd@ruijie.com.cn>

**- Why I did it**
fix https://github.com/Azure/sonic-buildimage/issues/5982
**- How I did it**
check notrip option before handle dbgsym debs
**- How to verify it**
set SONIC_DEBUGGING_ON=y / SONIC_PROFING_ON and compile the whole image
**- Which release branch to backport (provide reason below if selected)**

- [ ] 201811
- [ ] 201911
- [ ] 202006

**- Description for the changelog**

If SONIC_DEBUG_ON or SONIC_PROFILING_ON set to y, no dbgsym debs will trip from the origin deb, so mv dbgsym deb to some other folders will fail and compiling will stop.

**- A picture of a cute animal (not mandatory but encouraged)**

Issue #5982, opened 2020-11-20
[build] [debug] [201911] compile failed when enable SONIC_DEBUGGING_ON or SONIC_PROFILING_ON


tim-rj:
**Description**

compile failed locally when enable SONIC_DEBUGGING_ON or SONIC_PROFILING_ON

**Steps to reproduce the issue:**
1. set SONIC_DEBUGGING_ON = y (or SONIC_PROFILING_ON = y )
2. make target/sonic-broadcom.bin
3. compile failed

**Describe the results you received:**
```
/sonic/src/monit
mv: cannot stat 'monit-dbgsym_5.20.0-6_amd64.deb': No such file or directory
Makefile:9: recipe for target '/sonic/target/debs/stretch/monit_5.20.0-6_amd64.d                                                                                                          eb' failed
make[1]: *** [/sonic/target/debs/stretch/monit_5.20.0-6_amd64.deb] Error 1
make[1]: Leaving directory '/sonic/src/monit'
[  FAIL LOG END  ] [ target/debs/stretch/monit_5.20.0-6_amd64.deb ]
slave.mk:335: recipe for target 'target/debs/stretch/monit_5.20.0-6_amd64.deb' f                                                                                                          ailed
make: *** [target/debs/stretch/monit_5.20.0-6_amd64.deb] Error 1
make[1]: *** [Makefile.work:211: target/sonic-broadcom.bin] Error 2
make[1]: Leaving directory '/home/sdk/ssp/sonic-sym'
make: *** [Makefile:7: target/sonic-broadcom.bin] Error 2


```

**Describe the results you expected:**
```
ONE MACRO to enable symbol for all components and compile successful
```

**Additional information you deem important (e.g. issue happens only occasionally):**
```
If I run 'make target/debs/stretch/monit-dbgsym_5.20.0-6_amd64.deb ' after compile failed, it will success, but 'make target/sonic-broadcom.bin' will fall again for other component noting the linkly error 'cant stat xxx-dbgsym_xxx.deb'.

All the codes are open source ,we change nothing. what can we do to fix such problem?

```

tim-rj:
I found that in issue log, no dbgsym deb is built:
```
1023    dh_md5sums
1024    dh_builddeb
1025 dpkg-deb: building package 'monit' in '../monit_5.20.0-6_amd64.deb'.
1026 make[2]: Leaving directory '/sonic/src/monit/monit'
1027  dpkg-genbuildinfo --build=binary
[...]

PR #6299, opened 2020-12-25
Enable the whole SONiC to be built with SONIC_DEBUGGING_ON=y.

Str1ker17:
Signed-off-by: Alexey Suslov <alexey.s+git@ultibot.ru>

**- Why I did it**
I wanted test & developement builds of SONiC to contain all the debug info. However if I build SONiC with `SONIC_DEBUGGING_ON=y`, the build stops as not all dbgsym packages are generated.

**- How I did it**
Replaced absent dbgsym packages with a dummy .deb file.

**- How to verify it**
Open `rules/config` in text editor and uncomment `SONIC_DEBUGGING_ON=y`.
Then build as usual:
```make init
make configure PLATFORM=broadcom
make target/sonic-broadcom.bin
```
The reulting image is somewhat bigger, but not much.

- [ ] 201811
- [ ] 201911
- [ ] 202006

Enabled the whole SONiC image to be built with debug info.


lguohan:
can you check this pr? seems trying to solve the same issue? https://github.com/Azure/sonic-buildimage/pull/6058

Str1ker17:
Yes, it seems the issue is the same.
However we are both doing the wrong way; the right way is to make debhelper generate an empty dbgsym.deb file.
@tim-rj what do you think about it?

tim-rj:
hi Str1ker17, I think since there are no dbgsym debs when option is nostrip, there is no need to touch empty dbgsym debs just for compiling.

Str1ker17:
@tim-rj But I want the full SONiC image containing all possible debug symbols. Why it is not possible to build it out of the box?
If some package is poorly configured and does not export symbols from itself, I see few reasons to break the build completely in this case.

P.S. I am responding after a year, so the situation has possibly changed. So please tell me if it's possible to build SONiC with all debug symbols.

raphaelt-nvidia:
> @tim-rj But I want the full SONiC image containing all possible debug symbols. Why it is not possible to build it out of the box? If some package is poorly configured and does not export symbols from itself, I see few reasons to break the build completely in this case.
> 
> P.S. I am responding after a year, so the situation has possibly changed. So please tell me if it's possible to build SONiC with all debug symbols.

It hasn't changed.  The problems described in https://github.com/Azure/sonic-buildimage/issues/6413 andhttps://github.com/Azure/sonic-buildimage/issues/5982 still exist.

PR #6303, opened 2020-12-27
[docker-frr]: add static ipv6 loopback route for bgp to advertise

lguohan:
**- Why I did it**
frr does not advertise route if local route is not reachable, as a result
loopback route /64 is not advertised to the neighbors. Add static route
allows frr to advertise the route to its peers

**- How I did it**
- change staticd template
- add swss orchagent code to ignore the loopback route

**- How to verify it**
in t0 topology, check the neighbor advertisement
also checked warm reboot

- [ ] 201811
- [ ] 201911
- [ ] 202006

### DIFF
--- a/dockers/docker-fpm-frr/frr/frr.conf.j2
+++ b/dockers/docker-fpm-frr/frr/frr.conf.j2
@@ -15,5 +15,7 @@ agentx
 !
 {% include "staticd/staticd.default_route.conf.j2" %}
 !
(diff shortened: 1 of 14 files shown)


shi-su (approved):
Looks good to me. Not sure if the test failure is related

shi-su:
retest vsimage please

lguohan:
the test failure is related to the routecheck in sonic-utilities repo, need to change that.

shi-su (approved):

PR #6351, opened 2021-01-05
Explicitly set the notify mode for ryslog imfile module 

abdosi:
Why I did:

To address the review comment for PR https://github.com/Azure/sonic-buildimage/pull/6298

How I did:

Setup the mode="inotify" when loading imfile.

How I verify:

Made sure we are are getting supervisord logs in host immediately.


dockers/docker-base/etc/rsyslog.d/supervisor.conf
```diff
@@ -1,4 +1,4 @@
-module(load="imfile") #By default notify mode
+module(load="imfile" mode="inotify") # Set up notify mode
```
jleveque, line 1: Nit-pick: Suggest changing the comment to `# Ensure "inotify" mode is used`

abdosi: @jleveque Done

jleveque (approved):

jleveque:
Retest vsimage please

PR #6356, opened 2021-01-05
[build]: add branch and release tag in sonic_version.yml

lguohan:
**- Why I did it**
the branch refers the branch name, for example
master, 202012, 201911, ...

**- How I did it**
will modify 'show version' command to output this as well.

**- How to verify it**
example for master branch
```
build_version: 'master.602-6efc0a88'
debian_version: '10.7'
kernel_version: '4.19.0-9-2-amd64'
asic_type: vs
commit_id: '6efc0a88'
branch: 'master'
release: 'none'
build_date: Tue Dec 29 06:54:02 UTC 2020
build_number: 602
built_by: johnar@jenkins-worker-23
```

example for 202012 release branch
```
build_version: '202012.602-6efc0a88'
debian_version: '10.7'
kernel_version: '4.19.0-9-2-amd64'
asic_type: vs
commit_id: '6efc0a88'
branch: '202012'
release: '202012'
build_date: Tue Dec 29 06:54:02 UTC 2020
build_number: 602
built_by: johnar@jenkins-worker-23
```

- [ ] 201811
- [x] 201911
- [x] 202006
- [x] 202012

### DIFF
--- a/build_debian.sh
+++ b/build_debian.sh
@@ -472,11 +472,16 @@ fi
 
 ## Version file
 sudo mkdir -p $FILESYSTEM_ROOT/etc/sonic
+if [ -f files/image_config/sonic_release ]; then
+    sudo cp files/image_config/sonic_release $FILESYSTEM_ROOT/etc/sonic/
+fi
 export build_version="${SONIC_IMAGE_VERSION}"
 export debian_version="$(cat $FILESYSTEM_ROOT/etc/debian_version)"
 export kernel_version="${kversion}"
 export asic_type="${sonic_asic_platform}"
 export commit_id="$(git rev-parse --short HEAD)"
+export branch="$(git rev-parse --abbrev-ref HEAD)"
+export release="$(if [ -f $FILESYSTEM_ROOT/etc/sonic/sonic_release ]; then cat $FILESYSTEM_ROOT/etc/sonic/sonic_release; fi)"
 export build_date="$(date -u)"
 export build_number="${BUILD_NUMBER:-0}"
 export built_by="$USER@$BUILD_HOSTNAME"
--- a/files/build_templates/sonic_version.yml.j2
+++ b/files/build_templates/sonic_version.yml.j2
@@ -8,6 +8,12 @@ kernel_version: '{{ kernel_version }}'
 {% endif -%}
 asic_type: {{ asic_type }}
 commit_id: '{{ commit_id }}'
+branch: '{{ branch }}'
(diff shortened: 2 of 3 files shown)


yxieca (commented):
The tag will contain the branch name from where the image was built. This works for release images but not for dev builds.

For instance, if I do "git checkout -b foo master", and build from branch foo, the release tag will show 'foo'.

$(basename $(git config branch.$(git rev-parse --abbrev-ref HEAD).merge)) will show upstream branch name, this one has a better chance of being correct.

lguohan:
then, we should explicitly check-in a file and use that to derive the release tag. then, as long as user on this branch, they will be consistent on which branch they are in. the master branch will always have release tag 'master' and for each release branch, we will change that file to add a tag name there.

jleveque:
> The tag will contain the local branch name from where the image was built. This works for release images but not for dev builds.
> 
> For instance, if I do "git checkout -b foo master", and build from branch foo, the release tag will show 'foo'.

But is this a bad thing? Doesn't it make sense that if you're working on a dev branch that you renamed locally, you would probably want that branch name to show up in the build, no?

yxieca:
> > The tag will contain the local branch name from where the image was built. This works for release images but not for dev builds.
> > For instance, if I do "git checkout -b foo master", and build from branch foo, the release tag will show 'foo'.
> 
> But is this a bad thing? Doesn't it make sense that if you're working on a dev branch that you renamed locally, you would probably want that branch name to show up in the build, no?

Depending on the purpose of this tag. I believe this tag is intended for people to distinguish what upsteam branch it is built from so that the test behavior could be tailored to accordingly.

yxieca:
> > The tag will contain the local branch name from where the image was built. This works for release images but not for dev builds.
> > For instance, if I do "git checkout -b foo master", and build from branch foo, the release tag will show 'foo'.
> 
> But is this a bad thing? Doesn't it make sense that if you're working on a dev branch that you renamed locally, you would probably want that branch name to show up in the build, no?

Generally, when I name a branch, I name it according to what I intended to do, not where it come from (e.g. 201911, 202012). But what decides the test behavior is where it come from (201911, 202012).

lguohan:
ok, here is what I am going to do, I am going to rename it as branch, and then I am going to add a tag file called release.

yxieca (dismissed):

yxieca (approved):

dgsudharsan:
@lguohan This is required for 202106

PR #6368, opened 2021-01-06
Updated imfile configuration for supervisord logs 

abdosi:
Why I did:
Adding change for PR https://github.com/Azure/sonic-buildimage/pull/6298 for buster/stretch. Previous change was for docker-base and though that will be used for buster/stretch. However both buster and stretch has their own config file.

How I verify:
Verified on local build.


jleveque (approved):

jleveque:
@abdosi: I assume this needs to be cherry-picked into 201911 and 202012? If so, please add the appropriate labels.

PR #6416, opened 2021-01-11
[Platform][nokia]: python3-smbus package add with python3 and jinja fixes 

carl-nokia:
**- Why I did it**
fix platform driver breakage due to python3 upgrade and fix load minigraph errors with config load_minigraph -y

**- How I did it**
added python3-smbus to the pmon docker template since the previous was python2 specific 
fixed additional "ord" python2 specific code 
fixed the jinja templates used by qos reload - the template logic required data to be parsed 

**- How to verify it**
run "show platform XXX" commands and verify output
run "sudo config load_minigraph -y" and verify configuration 
run "show interfaces XXX" and verify output 

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012

### DIFF
--- a/device/nokia/armhf-nokia_ixs7215_52x-r0/Nokia-7215/buffers.json.j2
+++ b/device/nokia/armhf-nokia_ixs7215_52x-r0/Nokia-7215/buffers.json.j2
@@ -2,7 +2,9 @@
 
 {% set default_cable = '40m' %}
 {% set default_ports_num = 54 -%}
+{}
 
 {# Port configuration to cable length look-up table #}
 {# Each record describes mapping of DUT (DUT port) role and neighbor role to cable length #}
 {# Roles described in the minigraph #}
+
--- a/device/nokia/armhf-nokia_ixs7215_52x-r0/Nokia-7215/buffers.json.j2
+++ b/device/nokia/armhf-nokia_ixs7215_52x-r0/Nokia-7215/buffers.json.j2
@@ -2,9 +2,9 @@
 
 {% set default_cable = '40m' %}
 {% set default_ports_num = 54 -%}
-{}
 
 {# Port configuration to cable length look-up table #}
 {# Each record describes mapping of DUT (DUT port) role and neighbor role to cable length #}
 {# Roles described in the minigraph #}
+{}
 
--- a/dockers/docker-platform-monitor/Dockerfile.j2
+++ b/dockers/docker-platform-monitor/Dockerfile.j2
@@ -21,6 +21,7 @@ RUN apt-get update &&   \
         librrd-dev      \
         rrdtool         \
         python-smbus    \
+        python3-smbus   \
         ethtool         \
         dmidecode       \
         i2c-tools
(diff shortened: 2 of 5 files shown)


device/nokia/armhf-nokia_ixs7215_52x-r0/Nokia-7215/buffers.json.j2
```diff
@@ -2,7 +2,9 @@
 
 {% set default_cable = '40m' %}
 {% set default_ports_num = 54 -%}
+{}
```
jleveque, line 5: Why is this needed?

jleveque (changes requested):
Please fill out the PR description field

device/nokia/armhf-nokia_ixs7215_52x-r0/Nokia-7215/buffers.json.j2
carl-nokia: the rendering code required some data to operate on when processing the jinja templates used by qos reload - the template I originally submitted had some small logic but did not return data to be parsed

jleveque: I see. Nitpicking here, but I think these braces should come after the comments below, since they should technically contain the lookup table defined in the comments.

carl-nokia:
No problem - I see what you are saying - I changed the order

dockers/docker-platform-monitor/Dockerfile.j2
```diff
@@ -21,6 +21,7 @@ RUN apt-get update &&   \
         librrd-dev      \
         rrdtool         \
         python-smbus    \
+        python3-smbus   \
```
jleveque, line 24: Is this needed by the Nokia sonic-platform package? If so , I would prefer that you add it in the `install_requires` list of your setup.py file. This way it will get installed along with your package, and will not be installed if not needed.

carl-nokia: Yes, it is needed by the Nokia sonic-platform package and we see Arista using it also btw.    I attempted adding to install_requires list of setup.py file as you suggested, but it appears that mechanism is at runtime install time and did not work.  See error below:
WARNING: Retrying (Retry(total=4, connect=None, read=None, redirect=None, status=None)) after connection broken by 'NewConnectionError('<pip._vendor.urllib3.connection.HTTPSConnection object at 0xb5731730>: Failed to establish a new connection: [Errno -3] Temporary failure in name resolution')': /simple/python3-smbus/
Build time install solution is preferred I think  -  putting the package in our debian another way - still working on alternatives other than the current Arista method multiple debians

jleveque: @carl-nokia: Ah yes. The vendor's sonic-platform package gets installed in the PMon container at runtime, thus you are correct, my suggestion will not work. I am OK with this solution for now. Sorry to trouble you.

jleveque (approved):

PR #6418, opened 2021-01-12
[minigraph.py]: Force /32 prefix for mux cable server IPv4 loopbacks

theasianpianist:

**- Why I did it**
Server IPv4 loopbacks do not always arrive with /32 prefix, which is a requirement for the MUX_CABLE table in config DB
**- How I did it**

**- How to verify it**

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012

### DIFF
--- a/src/sonic-config-engine/minigraph.py
+++ b/src/sonic-config-engine/minigraph.py
@@ -1505,7 +1505,12 @@ def get_mux_cable_entries(mux_cable_ports, neighbors, devices):
             entry = {}
             neighbor = neighbors[intf]['name']
             entry['state'] = 'auto'
-            entry['server_ipv4'] = devices[neighbor]['lo_addr']
+
+            # Always force a /32 prefix for server IPv4 loopbacks
+            server_ipv4_lo_addr = devices[neighbor]['lo_addr'].split("/")[0]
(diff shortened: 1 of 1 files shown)


src/sonic-config-engine/minigraph.py
```diff
@@ -1505,7 +1505,12 @@ def get_mux_cable_entries(mux_cable_ports, neighbors, devices):
             entry = {}
             neighbor = neighbors[intf]['name']
             entry['state'] = 'auto'
-            entry['server_ipv4'] = devices[neighbor]['lo_addr']
+
+            # Always force a /32 prefix for server IPv4 loopbacks
+            server_ipv4_lo_addr = devices[neighbor]['lo_addr'].split("/")[0]
```
tahmed-dev, line 1510: can we check if there is prefix length? Are we not converting it to `/32`?

theasianpianist: This code is agnostic of existing prefix length and behaves the same with or without. The call to `ip_network` below implicitly adds a prefix length of /32, since we have removed any existing prefix here.

prsunny (approved):

tahmed-dev (approved):

PR #6420, opened 2021-01-12
Anchor "libprotobuf-dev" version to a specific version by using debian control file dependency

gechiang:
**- Why I did it**
From time to time the BRCM SAI debian built may be using one version of libprotobuf-dev while a different version being picked during installation.  This causes the installation failure.  We should anchor to a specific version during SAI debian build via the debian control file dependency specification to eliminate this type of failures.

**- How I did it**
Change the SAI debian control file to specify a specific dependency version:
       Depends: libprotobuf-dev, libprotobuf-lite17, libprotobuf17
Remove the use of "sudo apt install -y libprotobuf-dev" at build script.

**- How to verify it**
Incorporate this new SAI debian on a private sonic-buildimage and build for BRCM binary and installed to a BRCM SKU DUT to validate the entire process did not encounter any issue and that DUT boots up fine as previous version.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012

Anchor "libprotobuf-dev" version to a specific version by using debian control file dependency


lguohan (approved):

gechiang:
@lguohan Should I be concerned about the "Azure.sonic-buildimage (Test kvmtest)" and "Azure.sonic-buildimage" both failed?
or I can go ahead Squash and merge?

lguohan:
it is ok, I sent email on this subject.

PR #6447, opened 2021-01-14
[kdump] Fix OOM events in crashkernel

Staphylo:
**- Why I did it**

A few issues where discovered with crashkernel on Arista platforms.

1) platforms using `docker_inram=on` would end up OOM in kdump environment.
This happens because the same initramfs is used by SONiC and the crashkernel.
With `docker_inram=on` the `dockerfs.tar.gz` is extracted in a `tmpfs` created for the occasion.
Since `dockerfs.tar.gz` weights more than 1.5G, it doesn't fit into the kdump environment and ends up OOM.
This OOM event can in turn trigger a panic.

2) Arista platforms with `secureboot` enabled would fail to load the crashkernel because the kernel parameter would be discarded on boot.
This happens because the `boot0` in secureboot mode is strict about kernel parameter injection.

3) The secureboot path allowlist would remove kernel crash reports.

4) The kdump service would fail on Arista products since `/boot/` is empty in `secureboot`

**- How I did it**

1) To prevent an OOM event in the crashkernel the fix is to avoid the codepaths in `union-mount` that create tmpfs and populate them. Some more codepath specific to Arista devices are also skipped to make the kdump process faster.
This relies on detecting that the initramfs is starting in a kdump environment and skipping some initialization.
The `/usr/sbin/kdump-config` tool appends a few kernel cmdline arguments when loading the crashkernel.
The most unique one is `systemd.unit=kdump-tools.service` which is used in a few initramfs hooks to set `in_kdump`.

2) To allow `kdump` to work in `secureboot` environment the cmdline generation in boot0 was slightly modified.
The codepath to load kernel parameters changed by SONiC is now running for booting in secure mode.
It was altered to prevent an append only behavior which would grow the `kernel-cmdline` at every reboot.
This ever growing behavior would lead `kexec` to fail to load the kernel due to a too long cmdline.

3) To get the kernel crash under /var/crash this path has to be added to `allowlist_paths`

4) The `/host/image-XXX/boot` folder is now populated in `secureboot` mode but not used.

**- How to verify it**

Regular boot:
 - enable kdump
 - enable docker_inram=on via kernel-params
 - reboot
 - generate a crash `echo c > /proc/sysrq-trigger`
 - before: witness OOM events on the console
 - after: crash kernel works and crash available under /var/crash

Secure boot:
 - enable kdump
 - reboot
 - generate a crash `echo c > /proc/sysrq-trigger`
 - before: witness no kdump
 - after: crash kernel works and crash available under /var/crash

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [X] 202012

This bug is present in 202012 and needs to be addressed.

Fix kdump crashkernel errors seen on Arista devices

### DIFF
--- a/files/Aboot/boot0.j2
+++ b/files/Aboot/boot0.j2
@@ -86,7 +86,7 @@ installer_image_path="$image_path/$installer_image"
 
 boot_config="$target_path/boot-config"
 
-cmdline_allowlist='crashkernel'
+cmdline_allowlist="crashkernel"
 
 # for backward compatibility with the sonic_upgrade= behavior
 install="${install:-${sonic_upgrade:-}}"
@@ -392,7 +392,8 @@ extract_image() {
 
 extract_image_secureboot() {
     info "Extracting necessary swi content"
-    unzip -oq "$swipath" platform/firsttime .imagehash -d "$image_path"
+    # NOTE: boot/ is not used by the boot process but only extracted for kdump
+    unzip -oq "$swipath" 'boot/*' platform/firsttime .imagehash -d "$image_path"
 
     info "Installing image as $installer_image_path"
     mv "$swipath" "$installer_image_path"
@@ -650,6 +651,32 @@ write_default_cmdline() {
     cmdline_add "$delimiter"
 }
 
+write_cmdline() {
+    # use extra parameters from kernel-params hook if the file exists
+    if [ -f "$target_path/$kernel_params" ]; then
+        if $secureboot && $debug; then
+            warn "Unsafe: Loading extra kernel parameters from $kernel_params"
+            cat "$target_path/$kernel_params" | cmdline_append
+        elif ! $secureboot; then
+            info "Loading extra kernel parameters from $kernel_params"
+            cat "$target_path/$kernel_params" | cmdline_append
+        fi
+    fi
+
+    # FIXME: sonic sometimes adds extra kernel parameters from user space
+    #        this is unsafe but some will be kept as part of the regular boot
+    if [ -f "$image_path/kernel-cmdline" ]; then
+        for field in $cmdline_allowlist; do
+           cat "$image_path/kernel-cmdline" | tr ' ' '\n' | grep -E "$field" | tail -n 1 | cmdline_append
+        done
+    fi
+
+    # FIXME: legacy configuration files used by fast-reboot and eos2sonic
+    #        these should be deprecated over time.
+    cmdline_echo > "$image_path/kernel-cmdline"
+    cmdline_echo | sed 's/ cmdline-aboot-end.*$//' > "$target_path/kernel-params-base"
+}
+
 write_common_configs() {
     write_default_cmdline
     write_platform_specific_cmdline
@@ -667,28 +694,14 @@ write_secureboot_configs() {
     cmdline_add aboot.secureboot=enabled
     # setting panic= has the side effect of disabling the initrd shell on error
     cmdline_add panic=0
+    write_cmdline
 }
 
 write_regular_configs() {
     write_common_configs
     cmdline_add "loop=$image_name/fs.squashfs"
     cmdline_add loopfstype=squashfs
-
-    # use extra parameters from kernel-params hook if the file exists
-    if [ -f "$target_path/$kernel_params" ]; then
-        cat "$target_path/$kernel_params" | cmdline_append
-    fi
-
-    # FIXME: sonic sometimes adds extra kernel parameters from user space
-    #        this is unsafe but some will be kept as part of the regular boot
-    if [ -f "$image_path/kernel-cmdline" ]; then
-        cat "$image_path/kernel-cmdline" | tr ' ' '\n' | grep -E "$cmdline_allowlist" | cmdline_append
-    fi
-
-    # FIXME: legacy configuration files used by fast-reboot and eos2sonic
-    #        these should be deprecated over time.
-    cmdline_echo > "$image_path/kernel-cmdline"
-    cmdline_echo | sed 's/ cmdline-aboot-end.*$//' > "$target_path/kernel-params-base"
+    write_cmdline
 }
 
 run_kexec() {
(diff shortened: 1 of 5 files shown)


Staphylo:
retest vsimage please

Staphylo:
@jleveque @qiluo-msft

files/Aboot/boot0.j2
```diff
@@ -392,7 +392,8 @@ extract_image() {
 
 extract_image_secureboot() {
     info "Extracting necessary swi content"
-    unzip -oq "$swipath" platform/firsttime .imagehash -d "$image_path"
+    # NOTE: boot/ is not used by the boot process but only extracted for kdump
```
qiluo-msft, line 395: >boot/ is not used by the boot process but only extracted for kdump [](start = 12, length = 66)

Why this is made specially for secureboot? How about the normal boot?

files/Aboot/boot0.j2
```diff
@@ -650,6 +651,32 @@ write_default_cmdline() {
     cmdline_add "$delimiter"
 }
 
+write_cmdline() {
+    # use extra parameters from kernel-params hook if the file exists
+    if [ -f "$target_path/$kernel_params" ]; then
+        if $secureboot && $debug; then
```
qiluo-msft, line 657: >if $secureboot && $debug; then [](start = 8, length = 30)

If secureboot and not debug, you will not use extra parameter. This is surprising different than the case of secureboot and debug.

files/initramfs-tools/arista-convertfs.j2
```diff
@@ -141,14 +142,20 @@ for x in "$@"; do
...
+            in_kdump=true
+            ;;
     esac
 done
 
 # Check aboot
 [ -z "$aboot_flag" ] && exit 0
 
+# Check kdump
+$in_kdump && exit 0
```
qiluo-msft, line 155: >$in_kdump [](start = 0, length = 9)

Could you compare strings instead of run it?

files/initramfs-tools/arista-convertfs.j2
```diff
@@ -141,14 +142,20 @@ for x in "$@"; do
...
+
 # Skip this script for warm-reboot/fast-reboot from sonic
-[ "$sonic_fast_reboot" == true ] && [ "$prev_os" != eos ] && exit 0
+$sonic_fast_reboot && [ "$prev_os" != eos ] && exit 0
```
qiluo-msft, line 158: >$sonic_fast_reboot [](start = 0, length = 18)

compare strings?

files/initramfs-tools/union-mount.j2
```diff
@@ -86,7 +90,7 @@ mkdir -p "$rw_dir"
 mkdir -p "$work_dir"
 
 ## Remove the files not in allowlist in the rw folder
-if $secureboot; then
+if $secureboot && ! $in_kdump; then
```
qiluo-msft, line 93: >$secureboot [](start = 3, length = 11)

compare strings

files/initramfs-tools/union-mount.j2
```diff
@@ -120,23 +124,25 @@ case "${ROOT}" in
 esac
 
 mkdir -p ${rootmnt}/var/lib/docker
-if $secureboot; then
-    mount -t tmpfs -o rw,nodev,size={{ DOCKER_RAMFS_SIZE }} tmpfs ${rootmnt}/var/lib/docker
-    if [ "$bootloader" = "aboot" ]; then
-        unzip -qp "$swi_path" dockerfs.tar.gz | tar xz --numeric-owner -C ${rootmnt}/var/lib/docker
-        ## Boot folder is not extracted during secureboot since content would inherently become unsafe
-        mkdir -p ${rootmnt}/host/$image_dir/boot
+if ! $in_kdump; then
```
qiluo-msft, line 127: >$in_kdump [](start = 5, length = 9)

compare strings

files/Aboot/boot0.j2
Staphylo: In secureboot we don't want to reuse the extracted kernel on the flash because it could have been tampered with.
Instead we extract kernel/initrd from the SWI which is verified at boot time and kexec these which removes this attack vector.

files/initramfs-tools/arista-convertfs.j2
Staphylo: Sure I'll change that.

files/initramfs-tools/arista-convertfs.j2
Staphylo: Will change

files/initramfs-tools/union-mount.j2
Staphylo: Will change too

files/Aboot/boot0.j2
Staphylo: That's correct.
The `$target_path/$kernel_params` file which is `/host/kernel-params` allows the user to inject more kernel parameters on the cmdline. It's especially useful during development to figure out which set of parameters should be used on a platform. But it can also be used to change some runtime behavior. Unfortunately some could have unfortunate behavior and end up compromising a system. To remove the risk of an attacker injecting harmful parameter we disable this feature in secureboot and print a message. It is still possible to boot the box in debug mode and inject parameters but that will show at runtime.
If you prefer I can disable this mechanism for secureboot entirely, including `secureboot && debug`.

files/Aboot/boot0.j2
qiluo-msft: Is it true that normal build extracts this 'boot/*'?

---
In reply to: [562265581](https://github.com/Azure/sonic-buildimage/pull/6447#discussion_r562265581) [](ancestors = 562265581)

Staphylo: Yes they do.
Now secureboot do to. Not doing so would make `kdump-tools` report an error during boot after a crash has occured.
Hence the reason I now exctract kernel and initrd. Technically the kernel would be enough, but there's no harm in extracting both.

Staphylo:
retest vsimage please

qiluo-msft (approved):

lguohan:
is this needed for 202012?

Staphylo:
Yes it's needed for 202012

PR #6479, opened 2021-01-18
Way to skip over TACACS when there is system level error.

renukamanavalan:
**- Why I did it**
Fact: With TACACS configured, only remote users could login. 
When there is a serious system level error, for example Hard-disk becoming read-only, no new users could login, as that requires passwd and other files update along with creation of home dir. The "admin" user would have all entries pre-created, but TACACS would block local user login.

**- How I did it**
Added a script to run at the top of PAM stack, when login=tacacs+. This script would check the system and if there is any serious error, it would return system_err and the corresponding action is to skip the TACACS servers to land in pam_unix.so which would authenticate local users

**- How to verify it**
As it is tough to simulate read-only Hard disk, I hacked the check_system script to return "system_err" if a file /var/log/read_only exists. 
0) Update with following, if you need to test w/o upgrading image
    sudo cp src/sonic-host-services-data/templates/common-auth-sonic.j2 /usr/share/sonic/templates/common-auth-sonic.j2; cp 
    sudo cp src/sonic-host-services/scripts/check_system /usr/local/bin
    sudo chmod +x src/sonic-host-services/scripts/check_system
1) Configure TACACS with fail through not enabled
2) Try login as admin
    Expect: Fail
    Note: If you are running ssh, do Ctrl-C and get out
3) touch /var/log/read_only
4) Try login as admin
     Expect: Success

- [ ] 201811
- [x ] 201911
- [x ] 202006
- [x ] 202012


src/sonic-host-services/scripts/check_system
```diff
@@ -0,0 +1,23 @@
+#! /bin/bash
+
+echo "check_system called"
+SYSTEM_ERR=4
+
+# touch /var/log/syslog
+# ret=$?
+
+# to help simulate
+if [ -f /var/log/read_only ]; then
```
lguohan, line 10: how can this check system read-only?

src/sonic-host-services/setup.py
```diff
@@ -12,6 +12,7 @@
     maintainer_email = 'jolevequ@microsoft.com',
     scripts = [
         'scripts/caclmgrd',
+        'scripts/check_system',
```
lguohan, line 15: -> check_readonly_disk

src/sonic-host-services-data/templates/common-auth-sonic.j2
```diff
@@ -23,6 +23,7 @@ auth	[success=1 default=ignore]	pam_tacplus.so server={{ last_server.ip }}:{{ la
 
 {% endif %}
 {% elif auth['login'] == 'tacacs+' or auth['login'] == 'tacacs+,local' %}
+auth    [default=ignore system_err={{servers | count}}] pam_exec.so debug stdout type=auth /usr/local/bin/check_system
```
lguohan, line 26: better to introduce a new mode, not enable this by default for public.

src/sonic-host-services/scripts/check_system
lguohan: if the disk become read-only, how can this file be generated?

renukamanavalan: This is a hack for immediate testing. The commented lines above are the actual code.

renukamanavalan: Removed hack.

src/sonic-host-services/setup.py
renukamanavalan: done

src/sonic-host-services/scripts/check_readonly_disk
```diff
@@ -0,0 +1,16 @@
+#! /bin/bash
+
+echo "checking for read-only disk"
+SYSTEM_ERR=4
+
+touch /var/log/syslog
```
lguohan, line 6: sometime, the /var/log is mounted as tmpfs ramdisk, so we need to check some other file in the file system, maybe we should touch /home/admin/.test_disk_readwrite file?

src/sonic-host-services/scripts/check_readonly_disk
```diff
@@ -0,0 +1,16 @@
...
+ret=$?
+
+echo "ret=${ret}"
+
+if [ ${ret} -eq 0 ]; then
+    exit 0
+else
+    echo "Read-Only system"
```
lguohan, line 14: logger

src/sonic-host-services-data/templates/common-auth-sonic.j2
```diff
@@ -23,10 +23,23 @@ auth	[success=1 default=ignore]	pam_tacplus.so server={{ last_server.ip }}:{{ la
 
 {% endif %}
 {% elif auth['login'] == 'tacacs+' or auth['login'] == 'tacacs+,local' %}
+{% if not auth['failthrough'] %}
+{% if auth['check_readonly_disk'] %}
```
lguohan, line 27: the feature is probably better named as failthrough_on_readonly_disk?

lguohan:
can you update the testing section?

src/sonic-host-services/scripts/check_readonly_disk
renukamanavalan: The echo will show this info to the user logging in as below. Thought this might be useful.
Outside this, the monitor process could watch the disk status & log required error for alerts.

**Read-Only system**
Linux str-s6000-acs-13 4.19.0-9-2-amd64 #1 SMP Debian 4.19.118-2+deb10u1 (2020-06-07) x86_64
You are on
  ____   ___  _   _ _  ____
 / ___| / _ \| \ | (_)/ ___|
 \___ \| | | |  \| | | |
  ___) | |_| | |\  | | |___
 |____/ \___/|_| \_|_|\____|

-- Software for Open Networking in the Cloud --

Unauthorized access and/or use are prohibited.
All access and/or use are subject to monitoring.

src/sonic-host-services-data/templates/common-auth-sonic.j2
renukamanavalan: ok

src/sonic-host-services/scripts/check_readonly_disk
renukamanavalan: ok

src/sonic-host-services-data/templates/common-auth-sonic.j2
renukamanavalan: done

src/sonic-host-services/scripts/check_readonly_disk
lguohan: /tmp can also be tmpfs which is not on the disk.

renukamanavalan:
Instead,  monit does periodic disk_check, which will mount /home/ & /etc/ on top of tmpfs, if disk has become read-only. This way /etc & /home are assured to be RW anytime and hence TACACS login will not be blocked.

PR #6488, opened 2021-01-18
[voq/systemlag] Lag id boundary setting for system lag functionality

vganesan-nokia:
**- Why I did it**

System lag in VOQ based chassis systems requires unique id for system port aggregator id attribute in LAG object.
This id is generated by lag id generator within set boundary. The boundaries are system capability specific and should
be set during system boot. 

Ref: https://github.com/Azure/SONiC/pull/697

**- How I did it**

Changes for setting platform specific lag id boundary id in the chassis
app db. The platform specific lag id boundaries are supplied via
chassisdb.conf in the supervisor card. The lag_id_start and lag_id_end boundary values sourced
from this file are set in chassis app db which will be used by lag id
allocator to allocate unique lag id in atomic fashion

**- How to verify it**

- Update chassisdb.conf in supervisor card with two parameters "lag_id_start" and "lag_id_end"
- Reboot the system. 
- Connect to redis-chassis server. There should be "SYSTEM_LAG_ID_START" and "SYSTEM_LAG_ID_END" keys set with value as given in the chassisdb.conf.
- Verify that LAG id objects are created in ASIC DB.

- doker_image_ctl.j2 modified to create database.sh with script modification to write lag id keys in chassis app db

### DIFF
--- a/files/build_templates/docker_image_ctl.j2
+++ b/files/build_templates/docker_image_ctl.j2
@@ -101,6 +101,19 @@ function preStartAction()
     updateSyslogConf
 }
 
+{%- if docker_container_name == "database" %}
+
+function setPlatformLagIdBoundaries()
+{
+    CHASSIS_CONF=/usr/share/sonic/device/$PLATFORM/chassisdb.conf
+    if [ -f "$CHASSIS_CONF" ]; then
+        source $CHASSIS_CONF
+        $SONIC_DB_CLI CHASSIS_APP_DB SET "SYSTEM_LAG_ID_START" "$lag_id_start"
+        $SONIC_DB_CLI CHASSIS_APP_DB SET "SYSTEM_LAG_ID_END" "$lag_id_end"
+    fi
+}
+{%- endif %}
+
 function postStartAction()
 {
 {%- if docker_container_name == "database" %}
@@ -162,6 +175,7 @@ function postStartAction()
                  ($(docker exec -i ${DOCKERNAME} $SONIC_DB_CLI CHASSIS_APP_DB PING | grep -c True) -gt 0) ]]; do
            sleep 1
         done
+        setPlatformLagIdBoundaries
         REDIS_SOCK="/var/run/redis-chassis/redis_chassis.sock"
     fi
     chgrp -f redis $REDIS_SOCK && chmod -f 0760 $REDIS_SOCK


anshuv-mfst:
@judyjoseph - could you please review this PR, thanks.

files/build_templates/docker_image_ctl.j2
```diff
@@ -89,6 +89,19 @@ function preStartAction()
...
 }
 
+{%- if docker_container_name == "database" %}
+
+function setPlatformLagIdBoundaries()
+{
+    CHASSIS_CONF=/usr/share/sonic/device/$PLATFORM/chassisdb.conf
+    if [ -f "$CHASSIS_CONF" ]; then
+        source $CHASSIS_CONF
+        $SONIC_DB_CLI CHASSIS_APP_DB SET "SYSTEM_LAG_ID_START" "$lag_id_start"
```
judyjoseph, line 99: Could you add these parameters viz lag_id_start, lag_id_end as needed in the chassisdb.conf in the VOQ LAG HLD if it is not present ?

vganesan-nokia: Yes. Updated section 3.3 System LAG ID management section of "Distributed VOQ HLD" Azure/SONiC#697  to mention the file chassisdb.conf and these parameters.

judyjoseph: Looks ok thanks ! Also what happens in case the user didn't specify the START and END limits in chassis_db.conf file. Should we define a default set of values here, just before we source the chassis conf ? OR we expect the user to always specify it in chassis db for it to work correctly.

vganesan-nokia: If chassisdb.conf does not specify lag_id_start and lag_id_end, chassis db will not have these set. So lag id allocator will return error and port channel creation will not proceed further without lag id (aka system port aggregator id)  since system port aggregator id is a mandatory attribute expected by SAI in voq chassis systems. Therefore, with voq chassis system these two parameters are mandatory to be specified by the user based on platform capability. Since this is platform dependent, we'll not be able to choose a default unless all vendors agree on some default. Until now we have not thought about a default starting and ending lag id.

judyjoseph (approved):

files/build_templates/docker_image_ctl.j2
```diff
@@ -101,6 +101,19 @@ function preStartAction()
...
```
ysmanman, line 111: @vganesan-nokia We tested this change and it's not working at our side.  $SONIC_DB_CLI CHASSIS_APP_DB SET "SYSTEM_LAG_ID_START" "$lag_id_start"  failed with this error: Invalid database name input : 'CHASSIS_APP_DB'. However,  it worked fine if we do this inside chassisDb docker. E.g.,

docker exec -i ${DOCKERNAME} $SONIC_DB_CLI CHASSIS_APP_DB SET "SYSTEM_LAG_ID_START" "$lag_id_start"

So it seems sonic-db-cli is not aware of CHASSIS_APP_DB when databash.sh is executed.  Any thought on this?

vganesan-nokia: @ysmanman, This is supposed to be run only in database-chassis docker in supervisor card only. Assuming you are running this in database-chassis docker in supervisor, please check if you have the CHASSIS_APP_DB defined in database_config.json for redis-chassis instance (i.e. in file /var/run/redis-chassis/sonic-db/database_config.json)

ysmanman: @vganesan-nokia I am not suer that this is actually run in database-chassis docker based on the following code:

https://github.com/Azure/sonic-buildimage/blob/master/files/build_templates/docker_image_ctl.j2#L174-L178

Line 174-175 are running outside the docker, right?

vganesan-nokia: @ysmanman: you are right. It is run outside docker. We can modify the function setPlatfromLagIdBoundaries() to do "docker exec ...". However we do not have to do this or even to run this inside docker. The error you are seeing is because of this CHASSIS_APP_DB not defined in /var/run/redis/sonic-db/database_config.json. We are supposed to have the same content of database_config.json in all instances of /var/run/redis\<instance\>/sonic-db. If CHASSIS_APP_DB is defined in database_config.json with correct instance and if the instance info is specified, sonic-cfggen will connect to correct redis server.

ysmanman: @vganesan-nokia We do have CHASSIS_APP_DB defined in database_config.json once switch boots up. However, it appears that /var/run/redis/sonic-db/database_config.json didn't exist yet when setPlatformLagIdBoundaries was invoked.  The file was installed after chassisDb was up. Do you know if SONiC ensures  /var/run/redis/sonic-db/database_config.json is installed before bringing up chassisDb?

vganesan-nokia: @ysmanman, The file /var/run/redis\<instance\>/sonic-db/database_config.json is made available before starting the redis server inside the database docker (ref. docker-database-init.sh).  Even though we call the range setting function setPlatformLagIdBoundaries in the script  (ref. database.sh) that starts the database docker, we wait and call this function only after "ping" to the redis-server is successful. So the /var/run/redis\<instance\>/sonic-db/database_config.json should be available when we do this.

ysmanman: @vganesan-nokia  Unfortunately, we were observing different from what you explained. The way we set lag id range successfully is to modify setPlatformLagIdBoundaries to set it directly inside chassis database docker like this:

docker exec -i ${DOCKERNAME} $SONIC_DB_CLI CHASSIS_APP_DB SET "SYSTEM_LAG_ID_START" "$lag_id_start"

When setPlatformLagIdBoundaries wass called in the script  (ref. database.sh) that starts the database docker, it seems only  /var/run/redis-chassis/sonic-db/database_config.json was available but not  /var/run/redis/sonic-db/database_config.json.

ysmanman: @vganesan-nokia Thanks for checking and confirming!

vganesan-nokia: > 
> 
> @vganesan-nokia Thanks for checking and confirming!

@ysmanman, raised issue https://github.com/Azure/sonic-buildimage/issues/7912 and fixed the issue by PR https://github.com/Azure/sonic-buildimage/pull/7911

ysmanman: @vganesan-nokia thanks for the update!

PR #6517, opened 2021-01-21
[device/celestica]: Add xcvrd event support for Haliburton

mudsut4ke:
**- Why I did it**
- The xcvrd service requires an event detection function, unplug or plug in the transceiver.

**- How I did it**
- Add sysfs interrupt to notify userspace app of external interrupt
- Implement get_change_event() in chassis api.
- Also begin installing Python 3 sonic-platform package for Celestica platforms

**- How to verify it**
- Verify xcvrd running with error in syslog
- Verify port status and port information with this following command
  - `show interfaces status`
  - `show interfaces transceiver eeprom`
- Log :

**Signed-off-by**: Wirut Getbamrung [wgetbumr@celestica.com]

### DIFF
--- a/device/celestica/x86_64-cel_e1031-r0/sonic_platform/chassis.py
+++ b/device/celestica/x86_64-cel_e1031-r0/sonic_platform/chassis.py
@@ -13,13 +13,11 @@ import subprocess
 import json
 
 try:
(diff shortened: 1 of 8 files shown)


jleveque (changes requested):
Please fix LTGM warning: `get_serial()` method is defined twice in chassis.py

jleveque (approved):

PR #6519, opened 2021-01-21
[FRR] Create a separate script to wait zebra to be ready to receive connections

shi-su:
**- Why I did it**
The requirement for zebra to be ready to accept connections is a generic problem that is not specific to bgpd. Making the script to wait for zebra socket a separate script and let bgpd and staticd to wait for zebra socket.

**- How I did it**
Pull the script to wait for the zebra socket for bgpd as a separate one.

**- How to verify it**

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012


lguohan (approved):

lguohan:
@sudhanshukumar22, can you take a look to see if this address your issue on losing staticd routes?

dockers/docker-fpm-frr/frr/supervisord/supervisord.conf.j2
```diff
@@ -60,7 +71,7 @@ startsecs=0
 stdout_logfile=syslog
 stderr_logfile=syslog
 dependent_startup=true
-dependent_startup_wait_for=zebra:running
+dependent_startup_wait_for=zsocket:exited
```
sudhanshukumar22, line 74: Please change it to "staticd:running". This will make sure that bgp daemon starts after static daemon. In one case, static route had to be sent to BGP peer. Static route is config dependent and should come up before bgp.

dockers/docker-fpm-frr/zsocket.sh
```diff
@@ -23,11 +23,10 @@ while getopts ":a:o:h" opt; do
             ;;
     esac
 done
-shift $((OPTIND-1))
 
 timeout 5s bash -c -- "until </dev/tcp/${addr}/${port}; do sleep 0.1;done"
```
sudhanshukumar22, line 27: Can you please explain me what this does ? Does it wait only for 5seconds. Do you think it will be sufficient. In my case, I was waiting more because there may be a lot of configurations. What is port 2601. My command was
zebra_ready=$(netstat -tulpn | grep LISTEN | grep zebra)

sudhanshukumar22: There is a script src/sonic-bgpcfgd/bgpcfgd/main.py where we wait for 20 seconds for all daemons to come up. We are creating a dependency here which will increase the time for daemons to come up. Hence, we should increase the time for all daemons to come up there in frr.wait_for_daemons().

sudhanshukumar22: what is the meaning of line shift $((OPTIND-1))

dockers/docker-fpm-frr/frr/supervisord/supervisord.conf.j2
lguohan: does this matter? how does static route send route to bgp peer? i assume it is via zebra, so staticd connect to zebra and send route to zebra, later as soon as bgp connects, zebra will send those routes to bgpd. why bgpd has dependency on staticd?

dockers/docker-fpm-frr/zsocket.sh
```diff
@@ -23,11 +23,10 @@ while getopts ":a:o:h" opt; do
...
 if [ "$?" != "0" ]; then
     logger -p error "Error: zebra is not ready to accept connections"
 fi
```
lguohan, line 30: even if return value is zero, suggest to logger -p info for the time spend on waiting, it will be helpful for future debugging purpose.

dockers/docker-fpm-frr/zsocket.sh
shi-su: This line checks if the port is opened. Port 2601 is the default port for zebra connections. And it waits for at most 5 seconds for the port to be open. From my observation, the wait time for the port is about 0.2 seconds. So I think the 5 seconds limit should be enough. As this dependency only takes a limited amount of time, I do not think it is very necessary to extend the 20-second wait time for bgpcfgd (my observation of the wait time is about 2 seconds).

shi-su: For the line `shift $((OPTIND-1))`, the script was originally written for bgpd only. The options to this script were passed in along with the ones for bgpd. Therefore, we needed to shift off the processed options and only pass the ones for bgpd to bgpd. Since the script is now pulled out as an independent one and it is no longer necessary to pass the options, I removed the line.

dockers/docker-fpm-frr/zsocket.sh
shi-su: Added an info logger to log the time spent in waiting.

shi-su:
retest vsimage please

shi-su:
retest vsimage please

dockers/docker-fpm-frr/frr/supervisord/supervisord.conf.j2
sudhanshukumar22: @shi-su , can you please test this scenario. Create a 2 node topology and "redistribute static" in BGP. have a static route, save the FRR config and reboot. we should see the static route in the BGP neighbor.

lguohan: hi @sudhanshukumar22 , we are not familiar with that scenario, can you help to test?

sudhanshukumar22: @shi-su , please checkin your changes. Once we take the changes, if there are issues, I will file another PR.

lguohan (approved):

shi-su:
retest this please

shi-su:
retest this please

shi-su:
It seems that quite a few warm-reboot tests conducted on Jenkins worker-w16 are failing. Need to find out what is going on there.

shi-su:
retest vsimage please

lguohan:
let's not warm about the vsimage test on jenkins anymore, we are switching to azure pipeline which has more stable environment.

lguohan:
can you rebase and push, I am not sure why Test vstest is always failing on jenkins pipeline.

lguohan (approved):

PR #6527, opened 2021-01-21
[dhcp-relay]: Launch DHCP Relay On L3 Vlan

tahmed-dev:
Recent changes brought L2 vlan concept which do not have DHCP
clients behind them and so DHCP relay is not required. Also,
dhcpmon fails to launch on those vlans as their interfaces
lack IP addresses. This PR limit launch of both DHCP relay
and dhcpmon to L3 vlans only.

**- Why I did it**
Stop non functioning DHCP relay from being launched and also disable error being thrown by dhcpmon

**- How I did it**
Added extra check for presence of vlan in the vlan_interface table

**- How to verify it**
PR check passes

- [ ] 201811
- [x] 201911
- [x] 202006
- [x] 202012

### DIFF
--- a/dockers/docker-dhcp-relay/docker-dhcp-relay.supervisord.conf.j2
+++ b/dockers/docker-dhcp-relay/docker-dhcp-relay.supervisord.conf.j2
@@ -43,7 +43,7 @@ dependent_startup_wait_for=rsyslogd:running
 {# Count how many VLANs require a DHCP relay agent... #}
 {% set num_relays = { 'count': 0 } %}
 {% for vlan_name in VLAN %}
-{% if VLAN[vlan_name]['dhcp_servers'] %}
+{% if VLAN[vlan_name]['dhcp_servers'] and vlan_name in VLAN_INTERFACE %}
 {% set _dummy = num_relays.update({'count': num_relays.count + 1}) %}
 {% endif %}
 {% endfor %}
@@ -53,7 +53,7 @@ dependent_startup_wait_for=rsyslogd:running
 programs=
 {%- set add_preceding_comma = { 'flag': False } %}
 {% for vlan_name in VLAN %}
-{% if VLAN[vlan_name]['dhcp_servers'] %}
+{% if VLAN[vlan_name]['dhcp_servers'] and vlan_name in VLAN_INTERFACE %}
 {% if add_preceding_comma.flag %},{% endif %}
 {% set _dummy = add_preceding_comma.update({'flag': True}) %}
 isc-dhcp-relay-{{ vlan_name }}
@@ -64,7 +64,7 @@ isc-dhcp-relay-{{ vlan_name }}
 {# Create a program entry for each DHCP relay agent instance #}
 {% set relay_for_ipv4 = { 'flag': False } %}
 {% for vlan_name in VLAN %}
-{% if VLAN[vlan_name]['dhcp_servers'] %}
+{% if VLAN[vlan_name]['dhcp_servers'] and vlan_name in VLAN_INTERFACE %}
 {% for dhcp_server in VLAN[vlan_name]['dhcp_servers'] %}
 {% if dhcp_server | ipv4 %}
 {% set _dummy = relay_for_ipv4.update({'flag': True}) %}
@@ -107,7 +107,7 @@ dependent_startup_wait_for=start:exited
 programs=
 {%- set add_preceding_comma = { 'flag': False } %}
 {% for vlan_name in VLAN %}
-{% if VLAN[vlan_name]['dhcp_servers'] %}
+{% if VLAN[vlan_name]['dhcp_servers'] and vlan_name in VLAN_INTERFACE %}
 {% if add_preceding_comma.flag %},{% endif %}
 {% set _dummy = add_preceding_comma.update({'flag': True}) %}
 dhcpmon-{{ vlan_name }}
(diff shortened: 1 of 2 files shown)


abdosi:
we don't need this for 201811 ?

dockers/docker-dhcp-relay/docker-dhcp-relay.supervisord.conf.j2
```diff
@@ -118,7 +118,7 @@ dhcpmon-{{ vlan_name }}
 {# Create a program entry for each DHCP MONitor instance #}
 {% set relay_for_ipv4 = { 'flag': False } %}
 {% for vlan_name in VLAN %}
-{% if VLAN[vlan_name]['dhcp_servers'] %}
+{% if VLAN[vlan_name]['dhcp_servers'] and vlan_name in VLAN_INTERFACE %}
```
lguohan, line 121: i think you need to add unit test for this one.

dockers/docker-dhcp-relay/docker-dhcp-relay.supervisord.conf.j2
```diff
@@ -43,7 +43,7 @@ dependent_startup_wait_for=rsyslogd:running
 {# Count how many VLANs require a DHCP relay agent... #}
 {% set num_relays = { 'count': 0 } %}
 {% for vlan_name in VLAN %}
-{% if VLAN[vlan_name]['dhcp_servers'] %}
+{% if VLAN[vlan_name]['dhcp_servers'] and vlan_name in VLAN_INTERFACE %}
```
jleveque, line 46: Should we instead change this to:

```python
{% if VLAN_INTERFACE %}
{# Count how many VLANs require a DHCP relay agent... #}
{% set num_relays = { 'count': 0 } %}
{% for vlan_name in VLAN_INTERFACE %}
{% if VLAN and vlan_name in VLAN and VLAN[vlan_name]['dhcp_servers'] %}
{% set _dummy = num_relays.update({'count': num_relays.count + 1}) %}
{% endif %}
{% endfor %}
```

prsunny (dismissed):
lgtm.

dockers/docker-dhcp-relay/docker-dhcp-relay.supervisord.conf.j2
tahmed-dev: The `VLAN_INTERFACE` has other entries that are not vlan maps. here is an example:
```
    "VLAN_INTERFACE": {
        "Vlan701": {
            "proxy_arp": "enabled",
            "grat_arp": "enabled"
        },
        "Vlan701|2603:10e2:3b0:8000::1/64": {},
        "Vlan701|10.138.12.1/26": {}
    },
```

tahmed-dev:
> we don't need this for 201811 ?

No, L2 Vlans are not 201811

dockers/docker-dhcp-relay/docker-dhcp-relay.supervisord.conf.j2
jleveque: Ah, right. Technically, I think my proposal would still work, but `{% for vlan_name in VLAN_INTERFACE %}` would not be technically correct, because `vlan_name` would not be guaranteed to be an actual VLAN name.

tahmed-dev: I see, the `vlan_name in VLAN` should guard against this.

jleveque: Right. But I do think it feels a bit "wrong", as I described above. Thoughts?

dockers/docker-dhcp-relay/docker-dhcp-relay.supervisord.conf.j2
tahmed-dev: done. extended current test case to have L2 vlan

dockers/docker-dhcp-relay/docker-dhcp-relay.supervisord.conf.j2
tahmed-dev: I guess it boils down to nomenclature issue. The other way is to check for `|` and then do splitting to extract the vlan. The presence redis kind keys is strange here where we have `json` luxury.

jleveque: Yeah, great point. Why don't we simply store the IPv4 and IPv6 IP addresses in lists in the one VLAN_INTERFACE entry? @prsunny : Is there a reason?

prsunny: AFAIK, its a legacy design. One advantage is, it helps in deletion by having it as single key.

tahmed-dev:
retest vsimage please

lguohan:
where is the expected results?

lguohan (approved):

tahmed-dev:
> where is the expected results?

It is still one L3 vlan, however new L2 vlan does not have dhcprelay launched. It is a -ve test where without the change in the supervisord, the test fails with the following logs:
```
======================================================================
FAIL: test_dhcp_relay (tests.test_j2files.TestJ2Files)
----------------------------------------------------------------------
  File "/sonic/src/sonic-config-engine/tests/test_j2files.py", line 72, in test_dhcp_relay
    self.assertTrue(filecmp.cmp(os.path.join(self.test_dir, 'sample_output', utils.PYvX_DIR, 'docker-dhcp-relay.supervisord.conf'), self.output_file))
AssertionError: False is not true

----------------------------------------------------------------------
Ran 146 tests in 40.672s

FAILED (failures=1)
```

jleveque (approved):

PR #6528, opened 2021-01-21
[pmon] Run ledd using python3 unless excluded

Staphylo:
**- Why I did it**

Ledd is the last daemon that is not enabled to run in python3.
Even though there is a plan to deprecate this daemon and to replace it by something else it's one simple step toward python2 deprecation.

**- How I did it**

Changed the `command=` line for `ledd` in the `supervisord` configuration of `pmon`.
Copied what was done for other daemons.

**- How to verify it**

Booting a product that has a `led_control.py` should now show the ledd running in python3.
I ran `python3 -m pylint` on all `led_control.py` plugin which means that most of them should be python3 compliant.
There is however still a risk that some might not work.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [X] 202012

Run ledd in python3 unless excluded by the platform

### DIFF
--- a/dockers/docker-platform-monitor/docker-pmon.supervisord.conf.j2
+++ b/dockers/docker-platform-monitor/docker-pmon.supervisord.conf.j2
@@ -68,7 +68,7 @@ dependent_startup_wait_for=rsyslogd:running
 
 {% if not skip_ledd %}
 [program:ledd]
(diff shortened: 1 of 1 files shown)


jleveque (approved):

lguohan:
az pipeline test passed.

Issue #6483, opened 2021-01-18
Incorrect route for 0.0.0.0 in APPL_DB


bingwang-ms:
**Description**
The test case ```test_hash``` will extract route from ```APPL_DB``` by ```redis-dump -d 0 -k 'ROUTE*' -y```. 
Recently, ```test_hash``` was consistently failing with exception
```
E                   "Traceback (most recent call last):", 
E                   "  File \"ptftests/hash_test.py\", line 264, in runTest", 
E                   "    self.check_hash(hash_key)", 
E                   "  File \"ptftests/hash_test.py\", line 83, in check_hash", 
E                   "    (matched_index, _) = self.check_ip_route(hash_key, in_port, dst_ip, exp_port_list)", 
E                   "  File \"ptftests/hash_test.py\", line 91, in check_ip_route", 
E                   "    (matched_index, received) = self.check_ipv4_route(hash_key, in_port, dst_port_list)", 
E                   "  File \"ptftests/hash_test.py\", line 161, in check_ipv4_route", 
E                   "    return verify_packet_any_port(self, masked_exp_pkt, dst_port_list)", 
E                   "  File \"/usr/lib/python2.7/dist-packages/ptf/testutils.py\", line 2538, in verify_packet_any_port", 
E                   "    % (ports, device_number, result.format()))", 
E                   "AssertionError: Did not receive expected packet on any of ports [] for device 0.", 
```
The exception was because of incorrect default route, which results in empty ```exp_port_list```.
```
"ROUTE_TABLE:0.0.0.0/0": {
    "expireat": 1610968205.649774, 
    "ttl": -0.001, 
    "type": "hash", 
    "value": {
      "ifname": "eth0", 
      "nexthop": "10.3.146.1"
    }
  }
```
The most bizarre thing is the route in kernel is correct.
```
root@str2-7260cx3-acs-9:~# show ip route 0.0.0.0
Routing entry for 0.0.0.0/0
  Known via "bgp", distance 20, metric 0, best
  Last update 02:06:27 ago
  * 10.0.0.33, via PortChannel0001, weight 1
  * 10.0.0.35, via PortChannel0002, weight 1
  * 10.0.0.37, via PortChannel0003, weight 1
  * 10.0.0.39, via PortChannel0004, weight 1
```

**Still not sure if this issue is caused by previously running test cases or triggered by some operations.**

**Steps to reproduce the issue:**
The issue is not easy to repro when running manually. But was consistently failing on regression test.
[...]

PR #6543, opened 2021-01-24
[vstest]: add default route vs test

lguohan:
Check #6483

add test to make sure default route change in eth0 does not
affect the default route in the default vrf

**- Why I did it**
add regression test for #6483 

**- How I did it**
add vstest

**- How to verify it**
run vstest and failed. will make sure it success after issued fixed.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012

### DIFF
--- a/platform/vs/tests/bgp/test_default_route.py
+++ b/platform/vs/tests/bgp/test_default_route.py
@@ -1,0 +1,48 @@
+from swsscommon import swsscommon
+import os
+import re
+import time
+import json
+import pytest
+
+def test_DefaultRoute(dvs, testlog):
+
+    dvs.copy_file("/etc/frr/", "bgp/files/default_route/bgpd.conf")
+    dvs.runcmd("supervisorctl start bgpd")
+    dvs.runcmd("ip addr add 10.10.10.0/31 dev Ethernet0")
+    dvs.runcmd("config interface startup Ethernet0")
+    dvs.runcmd("ip route del 0.0.0.0/0")
+    dvs.runcmd("vtysh -c \"confgure terminal\" -c \"ip route 0.0.0.0/0 via 172.17.0.1 200\"")
+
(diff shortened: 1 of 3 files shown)


shi-su (approved):

lguohan:
az pipeline passed.

platform/vs/tests/bgp/test_default_route.py
```diff
@@ -0,0 +1,48 @@
...
+    time.sleep(10)
+
+    (exit_code, output) = dvs.runcmd(["redis-cli", "hgetall", "ROUTE_TABLE:0.0.0.0/0"])
+    print(exit_code, output)
+
+    # make sure 10.10.10.1 is the correct next hop for default route
+    assert "10.10.10.1" in output
+
+    # insert default route for table default
+    dvs.runcmd("ip route add default via 172.17.0.1 table default")
```
imzyxwvu, line 35: `ip route add` installs a kernel route to zebra. The administrative distance of the kernel route is 0 since the metric value is 0 and zebra uses the upper byte of kernel route metric as the distance. So the kernel route wins the BGP route and the only nexthop will stay eth0 until the kernel route is deleted. It's an expected behaviour and will prevent all subsequent BGP updates from taking effect in swss, and the ASIC. Filtering routes with eth0 nexthop just hides the problem and won't fix it.

prsunny: Just to clarify, these are two different tables - `default `and `main`. In Sonic, eth0 routes are part of `default route table` and we would want to give preference to BGP routes in swss and ASIC (which is part of main table) and not the eth0 routes (kernel). I think your point is valid if we have all routes in a single routing table. Did we miss anything?

imzyxwvu: Thanks. The two difference tables you pointed out are what I've missed. And I've successfully reproduced it in our environment.

PR #6546, opened 2021-01-25
[bgpmon]: Fix exception in bgpmon caused by duplicate bgp neighbor ID

bingwang-ms:
Signed-off-by: bingwang <bingwang@microsoft.com>

**- Why I did it**
It is possible that BGP neighbors in IPv4 and IPv6 address families share the same name. 
For example, if belowing config is applied to creare a bgp monitor:
```
{
    "BGP_MONITORS": {
        "11.0.0.1": {
            "admin_status": "up",
                "asn": "4200065100",
                "holdtime": "10",
                "keepalive": "3",
                "local_addr": "10.1.0.32",
                "name": "bgp_monitor",
                "nhopself": "0",
                "rrclient": "0"
        }
    }
}
```
Then a bgp neighbor named ```11.0.0.1``` will exist in both ipv4 and ipv6 neighbors.
```
root@str-7260cx3-acs-2:/var/log# show ip bgp sum

IPv4 Unicast Summary:
BGP router identifier 10.1.0.32, local AS number 4200065100 vrf-id 0
BGP table version 45133
RIB entries 12807, using 2458944 bytes of memory
Peers 5, using 109080 KiB of memory
Peer groups 5, using 320 bytes of memory


Neighbhor      V          AS    MsgRcvd    MsgSent    TblVer    InQ    OutQ  Up/Down    State/PfxRcd    NeighborName
-----------  ---  ----------  ---------  ---------  --------  -----  ------  ---------  --------------  --------------
10.0.0.33      4  4200064600       9616      18386         0      0       0  02:33:56   6400            ARISTA01T1
10.0.0.35      4  4200064600       9616      19205         0      0       0  02:33:56   6400            ARISTA02T1
10.0.0.37      4  4200064600       9618      20598         0      0       0  02:33:55   6400            ARISTA03T1
10.0.0.39      4  4200064600       9616      17468         0      0       0  02:33:55   6400            ARISTA04T1
11.0.0.1       4  4200065100          0          0         0      0       0  never      Active          bgp_monitor

Total number of neighbors 5
root@str-7260cx3-acs-2:/var/log# show ipv6 bgp sum

IPv6 Unicast Summary:
BGP router identifier 10.1.0.32, local AS number 4200065100 vrf-id 0
BGP table version 36748
RIB entries 12807, using 2458944 bytes of memory
Peers 5, using 109080 KiB of memory
Peer groups 5, using 320 bytes of memory


Neighbhor      V          AS    MsgRcvd    MsgSent    TblVer    InQ    OutQ  Up/Down    State/PfxRcd    NeighborName
-----------  ---  ----------  ---------  ---------  --------  -----  ------  ---------  --------------  --------------
11.0.0.1       4  4200065100          0          0         0      0       0  never      Active          bgp_monitor
fc00::2a       4  4200064600       9617      21462         0      0       0  02:33:59   6400            ARISTA03T1
fc00::2e       4  4200064600       9618      20863         0      0       0  02:33:59   6400            ARISTA04T1
fc00::22       4  4200064600       9615      19880         0      0       0  02:34:00   6400            ARISTA01T1
fc00::26       4  4200064600       9616      20582         0      0       0  02:34:00   6400            ARISTA02T1

Total number of neighbors 5
```
However, such case is not handled in ```bgpmon.py```, and an Exception will be raised because the same key is deleted twice. 
```
Jan 25 06:40:56.540613 str-7260cx3-acs-2 INFO bgp#/supervisord: bgpmon Traceback (most recent call last):
Jan 25 06:40:56.540971 str-7260cx3-acs-2 INFO bgp#/supervisord: bgpmon   File "/usr/local/bin/bgpmon", line 8, in <module>
Jan 25 06:40:56.540971 str-7260cx3-acs-2 INFO bgp#/supervisord: bgpmon     sys.exit(main())
Jan 25 06:40:56.540971 str-7260cx3-acs-2 INFO bgp#/supervisord: bgpmon   File "/usr/local/lib/python3.7/dist-packages/bgpmon/bgpmon.py", line 167, in main
Jan 25 06:40:56.540971 str-7260cx3-acs-2 INFO bgp#/supervisord: bgpmon     bgp_state_get.update_neigh_states()
Jan 25 06:40:56.540971 str-7260cx3-acs-2 INFO bgp#/supervisord: bgpmon   File "/usr/local/lib/python3.7/dist-packages/bgpmon/bgpmon.py", line 143, in update_neigh_states
Jan 25 06:40:56.540971 str-7260cx3-acs-2 INFO bgp#/supervisord: bgpmon     del self.peer_state[peer]
Jan 25 06:40:56.540971 str-7260cx3-acs-2 INFO bgp#/supervisord: bgpmon KeyError: '11.0.0.1'
```
This commit will address the issue.

**- How I did it**
Use ```set``` instead of ```list``` to store neighbors' ID to avoid duplicate keys.

**Please be noted that this PR only fix exception in ```bgpmon.py```. Similar issue exists in ```STATE_DB``` because ```NEIGH_STATE_TABLE|x.x.x.x``` is used as keys in ```STATE_DB```. If neighbors have duplicated keys, the value will be overwriten.** 

**- How to verify it**
Verified on A7260. No exception is observed after update.

- [ ] 201811
- [x] 201911
- [x] 202006
- [x] 202012

No.

### DIFF
--- a/src/sonic-bgpcfgd/bgpmon/bgpmon.py
+++ b/src/sonic-bgpcfgd/bgpmon/bgpmon.py
@@ -145,7 +145,7 @@ class BgpStateGet:
         # If anything in the pipeline not yet flushed, flush them now
         if len(data) > 0:
             self.flush_pipe(data)
-        # Save the new dict
+        # Save the new set
         self.peer_l = self.new_peer_l.copy()
 
 def main():
--- a/src/sonic-bgpcfgd/bgpmon/bgpmon.py
+++ b/src/sonic-bgpcfgd/bgpmon/bgpmon.py
@@ -4,20 +4,20 @@
 Description: bgpmon.py -- populating bgp related information in stateDB.
     script is started by supervisord in bgp docker when the docker is started.
 
-    Initial creation of this daemon is to assist SNMP agent in obtaining the 
+    Initial creation of this daemon is to assist SNMP agent in obtaining the
     BGP related information for its MIB support. The MIB that this daemon is
     assisting is for the CiscoBgp4MIB (Neighbor state only). If there are other
-    BGP related items that needs to be updated in a periodic manner in the 
+    BGP related items that needs to be updated in a periodic manner in the
     future, then more can be added into this process.
 
     The script check if there are any bgp activities by monitoring the bgp
     frr.log file timestamp.  If activity is detected, then it will request bgp
-    neighbor state via vtysh cli interface. This bgp activity monitoring is 
+    neighbor state via vtysh cli interface. This bgp activity monitoring is
     done periodically (every 15 second). When triggered, it looks specifically
     for the neighbor state in the json output of show ip bgp neighbors json
     and update the state DB for each neighbor accordingly.
     In order to not disturb and hold on to the State DB access too long and
-    removal of the stale neighbors (neighbors that was there previously on 
+    removal of the stale neighbors (neighbors that was there previously on
     previous get request but no longer there in the current get request), a
     "previous" neighbor dictionary will be kept and used to determine if there
     is a need to perform update or the peer is stale to be removed from the
@@ -65,11 +65,15 @@ class BgpStateGet:
         except (IOError, OSError):
             return True
 
-    def update_new_peer_states(self, peer_dict):
+    def update_new_peer_states(self, peer_dict, ip_ver):
         peer_l = peer_dict["peers"].keys()
-        self.new_peer_l.extend(peer_l)
         for peer in peer_l:
-            self.new_peer_state[peer] = peer_dict["peers"][peer]["state"]
+            if ip_ver == 4:
+                peer_key = peer + "_ipv4"
+            else:
+                peer_key = peer + "_ipv6"
+            self.new_peer_l.append(peer_key)
+            self.new_peer_state[peer_key] = peer_dict["peers"][peer]["state"]
 
     # Get a new snapshot of BGP neighbors and store them in the "new" location
     def get_all_neigh_states(self):
@@ -84,8 +88,10 @@ class BgpStateGet:
         del self.new_peer_l[:]
         self.new_peer_state.clear()
         for key, value in peer_info.items():
-            if key == "ipv4Unicast" or key == "ipv6Unicast":
-                self.update_new_peer_states(value)
+            if key == "ipv4Unicast":
+                self.update_new_peer_states(value, 4)
+            elif key == "ipv6Unicast":
+                self.update_new_peer_states(value, 6)
 
     # This method will take the caller's dictionary which contains the peer state operation
     # That need to be updated in StateDB using Redis pipeline.
@@ -117,7 +123,7 @@ class BgpStateGet:
         data = {}
         for i in range (0, len(self.new_peer_l)):
             peer = self.new_peer_l[i]
-            key = "NEIGH_STATE_TABLE|%s" % peer
+            key = "NEIGH_STATE_TABLE|%s" % peer.split('_')[0]
             if peer in self.peer_l:
                 # only update the entry if state changed
                 if self.peer_state[peer] != self.new_peer_state[peer]:
@@ -138,9 +144,10 @@ class BgpStateGet:
         while len(self.peer_l) > 0:
             # remove this from the stateDB and the current neighbor state entry
             peer = self.peer_l.pop(0)
-            del_key = "NEIGH_STATE_TABLE|%s" % peer
+            del_key = "NEIGH_STATE_TABLE|%s" % peer.split('_')[0]
             data[del_key] = None
-            del self.peer_state[peer]
+            if peer in self.peer_state:
+                del self.peer_state[peer]
             if len(data) > PIPE_BATCH_MAX_COUNT:
                 self.flush_pipe(data)
         # If anything in the pipeline not yet flushed, flush them now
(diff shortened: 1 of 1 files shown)


lguohan:
@gechiang , I think bgpmon still missing unit test. please add unit test to protect such behavior.

bingwang-ms:
@lguohan Please help to take a look at another related issue https://github.com/Azure/sonic-buildimage/issues/6547 . Thanks

gechiang:
@bingwang-ms Thanks for triaging this issue.  I have a very basic question.  Why is this "11.0.0.1" Neighbor being reported by FRR as part of the IPV6 unicast neighbor?  Isn't this not a valid behavior that needs to be addressed first by FRR?  Will this be investigated or raised as an issue with BGP FRR community?  But in any case we need to beef up our code to prevent this from causing an issue.

As for the fix I am thinking we should convert the "peer list" to a "peer set".  this way the duplicated entry(ies) will only be in the set ONCE.  If we do this it will also avoid causing the STATE DB issue since we no longer have duplicated peers.  Let's not try to append ip version to the key to solve the initial issue.  Let me know if my suggestion is sound?
Thanks again!

Let me know if you agree with my suggested approach...
Thanks!

bingwang-ms:
> @bingwang-ms Thanks for triaging this issue. I have a very basic question. Why is this "11.0.0.1" Neighbor being reported by FRR as part of the IPV6 unicast neighbor? Isn't this not a valid behavior that needs to be addressed first by FRR? Will this be investigated or raised as an issue with BGP FRR community? But in any case we need to beef up our code to prevent this from causing an issue.
> 
> As for the fix I am thinking we should convert the "peer list" to a "peer set". this way the duplicated entry(ies) will only be in the set ONCE. If we do this it will also avoid causing the STATE DB issue since we no longer have duplicated peers. Let's not try to append ip version to the key to solve the initial issue. Let me know if my suggestion is sound?
> Thanks again!
> 
> Let me know if you agree with my suggested approach...
> Thanks!

Thanks @gechiang . I'm not sure if the behavior is triggered by FRR or our scripts. So I filed a issue here. 
For the second question, I think using a set instead of dict is not a good option. We may lose some neighbors if ipv4 and ipv6 neigbors share the same ID. In other words, if we are asure that duplicated keys should never exist in ipv4 and ipv6 neighbors, we don't need this fix.

abdosi:
> @bingwang-ms Thanks for triaging this issue. I have a very basic question. Why is this "11.0.0.1" Neighbor being reported by FRR as part of the IPV6 unicast neighbor? Isn't this not a valid behavior that needs to be addressed first by FRR? Will this be investigated or raised as an issue with BGP FRR community? But in any case we need to beef up our code to prevent this from causing an issue.
> 
> As for the fix I am thinking we should convert the "peer list" to a "peer set". this way the duplicated entry(ies) will only be in the set ONCE. If we do this it will also avoid causing the STATE DB issue since we no longer have duplicated peers. Let's not try to append ip version to the key to solve the initial issue. Let me know if my suggestion is sound?
> Thanks again!
> 
> Let me know if you agree with my suggested approach...
> Thanks!

@gechiang we use v4 neighbor peering to advertise v6 routes.

gechiang:
@abdosi  Thanks for responding.  So this is the same peer that got reported via IPV4 unicast as well as IPV6 unicast.  Therefore one of this should be ignored.  This come back to the suggestion on how we should address this duplication.
@bingwang-ms Since it is the same peer if we use the way you intent to solve the exception issue it will still result to having the same peer in the list.  Which one should we use to update the state DB?  The purpose of this STATE DB is to report the state of the peer's state.  If we were to change the handling to convert the "peer list" to a "peer set" then this duplication will be resolved as we will only have one single peer in the "peer set" and that is the one we can use for the State DB update operation.
let me know if you also agree?

bingwang-ms:
Thanks @abdosi and @gechiang for clarification. If the same ID in IPv4 neigh and IPv6 neigh refer to the same one, I think it's correct to ignore one. I will update the PR to address the issue. And items in STATE_DB are expecetd. Correct?

bingwang-ms:
Updated. Thanks

src/sonic-bgpcfgd/bgpmon/bgpmon.py
```diff
@@ -135,19 +134,19 @@ def update_neigh_states(self):
...
+            if peer in self.peer_state:
+                del self.peer_state[peer]
             if len(data) > PIPE_BATCH_MAX_COUNT:
                 self.flush_pipe(data)
         # If anything in the pipeline not yet flushed, flush them now
         if len(data) > 0:
             self.flush_pipe(data)
-        # Save the new List
-        self.peer_l = self.new_peer_l[:]
+        # Save the new dict
```
gechiang, line 148: Overall the change looks great.  Just the comment here needs to be changed from "dict" to "set" to avoid confusion.  Nice work overall!  Thanks for the fix!

bingwang-ms: Thanks. Updated.
So I'll close #6547. I think it's not an issue. Correct?

gechiang (approved):

src/sonic-bgpcfgd/bgpmon/bgpmon.py
gechiang: @bingwang-ms Yes. #6547 is not an issue. Your PR fix will not allow duplicated entry in STATE DB to occur.
Thanks!

bingwang-ms:
Retest this, please

yxieca (approved):

gechiang:
retest vsimage please

PR #6560, opened 2021-01-26
[ci]: add more branches into azure-pipelines trigger

qiluo-msft:
**- Why I did it**
Currently we have an empty branch view on Azure pipeline webpage. I added all branches into trigger.

**- How I did it**

**- How to verify it**

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -4,7 +4,9 @@
 # https://aka.ms/yaml
 
 trigger:
-- main
+  branches:
+    include:
+      - "*"
 
 pr:
 - master
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -6,7 +6,8 @@
 trigger:
   branches:
     include:
-      - "*"
+      - master
+      - 202012
 
 pr:
 - master


azure-pipelines.yml
```diff
@@ -4,7 +4,9 @@
 # https://aka.ms/yaml
 
 trigger:
-- main
+  branches:
+    include:
+      - "*"
```
lguohan, line 9: not all branches support all builds listed below, and kvmtest.

lguohan: we should consider to do batch as well. https://docs.microsoft.com/en-us/azure/devops/pipelines/repos/azure-repos-git?view=azure-devops&tabs=yaml#batching-ci-runs

lguohan: it is better just enable master and 202012 for now. 201911 will have kvm test failure.

lguohan (changes requested):
as comments.

lguohan (approved):

PR #6570, opened 2021-01-27
[ci]: Download artifact instead of using nfs storage

qiluo-msft:
**- Why I did it**
I notice that I rerun a failed job (not the stages), the nfs store is already cleaned by previous failed jobs.

**- How I did it**
Use artifacts instead.

**- How to verify it**

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -123,14 +123,15 @@ stages:
     - task: DownloadPipelineArtifact@2
       inputs:
         artifact: sonic-buildimage.kvm
+      displayName: "Download sonic-buildimage.kvm artifact"
 
     - script: |
+        set -x
         sudo dpkg -i --force-confask,confnew ../sonic-swss-common.amd64.ubuntu20_04/libswsscommon_1.0.0_amd64.deb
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -149,9 +146,13 @@ stages:
     displayName: "kvmtest"
     timeoutInMinutes: 240
     steps:
+    - task: DownloadPipelineArtifact@2
+      inputs:
+        artifact: sonic-buildimage.kvm
```
lguohan, line 151: i am not sure it will download the artifact from the current job?

qiluo-msft: I learned on https://docs.microsoft.com/en-us/azure/devops/pipelines/artifacts/pipeline-artifacts?view=azure-devops&tabs=yaml

> Pipeline artifacts provide a way to share files between stages in a pipeline or between different pipelines.

In this case, we share files between stages.

lguohan (approved):

lguohan:
kvmtest passed.

PR #6577, opened 2021-01-27
Add pipeline file on branch 202012, supporting PR build & official build

liushilongbuaa:
**- Why I did it**
Add pipeline file for Azure pipeline. PR build & official build
**- How I did it**

**- How to verify it**

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012


pipelines/official_build.yml
```diff
@@ -0,0 +1,32 @@
+# Starter pipeline
+# Start with a minimal pipeline that you can customize to build and deploy your code.
+# Add steps that build, run tests, deploy, and more:
+# https://aka.ms/yaml
+
+stages:
+- stage: One
```
lguohan, line 7: what is stage One?

pipelines/official_build.yml
```diff
@@ -0,0 +1,32 @@
+# Starter pipeline
+# Start with a minimal pipeline that you can customize to build and deploy your code.
+# Add steps that build, run tests, deploy, and more:
+# https://aka.ms/yaml
+
+stages:
```
lguohan, line 6: what is the trigger?

lguohan:
i suggest to start form official build on master branch first, then we can start doing other branches.

pipelines/official_build.yml
liushilongbuaa: trigger can be overide by UI config. And I have tried yaml format trigger: [schedules](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/scheduled-triggers?view=azure-devops&tabs=yaml), which not work as expected.
So I perfer [UI config.](https://dev.azure.com/sonicswitch/build/_apps/hub/ms.vss-ciworkflow.build-ci-hub?_a=edit-build-definition&id=33&view=Tab_Triggers)

pipelines/official_build.yml
liushilongbuaa: I worried that use different stages maybe run on different node. 
This worry comes from log, and haven't confirmed by ms doc.
So I prefer to use one stage. it is a non-sense name

liushilongbuaa: changed to Build

PR #6585, opened 2021-01-28
Stop prompt during build

qiluo-msft:
**- Why I did it**
Some commands used during build will prompt user interactively, but this is not expected during build. Since most output is collected into log file, user could not see the prompt and feel the build process hangs.

**- How I did it**
1. Use `mv` command in non interactive mode
2. Redirect stdin to null if command output is collected into log file.

**- How to verify it**
Manual test in local build

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012


lguohan:
never experienced such issue. how to repro?

qiluo-msft:
Somehow build a binary by root, and then build by normal user again.
```
-rw-r--r-- 1 root  root    205468 Dec 29 18:48 ../../target/debs/buster/libswsscommon_1.0.0_amd64.deb
```

lguohan (approved):

PR #6586, opened 2021-01-28
[system-health] Check fan speed before check fan status

Junchao-Mellanox:
**- Why I did it**

In thermalctd, when speed of fan exceeds threashold, the fan status will be saved as "bad". So in system health, it is better to check fan speed before fan status. In this case, if fan speed exceeds threashold, we get more deatil information.

**- How I did it**

Move fan speed check logic before fan status check

**- How to verify it**

Manual test

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012


jleveque:
@Junchao-Mellanox: Can you please also add a unit test to ensure this behavior doesn't regress?

liat-grozovik (approved):

PR #6600, opened 2021-01-29
[installer] Fix variable inside machine.conf caused install.sh error

kuanyu99:
**- Why I did it**

Encounter error when install SONiC image if there are some onie_discovery variables assigned with multiple values inside machine.conf

**- How I did it**

Replace original ". /machine.conf" method and add another function to do the same thing.

**- How to verify it**

1. Add a item inside /host/machine.conf like `onie_disco_ntpsrv=10.254.141.10 10.254.141.131`
2. Do `sonic_installer install` to check if any error occurs

- [ ] 201811
- [x] 201911
- [x] 202006
- [x] 202012

Add a new function 'read_conf_file' to read the settings from machine.conf when the value in machine.conf are separated by space without double quotes between them.

### DIFF
--- a/installer/x86_64/install.sh
+++ b/installer/x86_64/install.sh
@@ -24,8 +24,12 @@ read_conf_file() {
     local conf_file=$1
     while IFS='=' read -r var value
     do
-        f_char=$(echo $var | cut -c1)
-        [ "$f_char" = "#" ] && continue
+        # remove comment string
+        var=${var%#*}
+        value=${value%#*}
+        # trim the space
+        var="$(echo $var)"
+        # skip blank line
         [ -z "$var" ] && continue
         # remove double quote in the beginning
         tmp_val=${value#\"}
--- a/installer/x86_64/install.sh
+++ b/installer/x86_64/install.sh
@@ -24,6 +24,9 @@ read_conf_file() {
     local conf_file=$1
     while IFS='=' read -r var value
     do
+        f_char=$(echo $var | cut -c1)
+        [ "$f_char" = "#" ] && continue
+        [ -z "$var" ] && continue
         # remove double quote in the beginning
         tmp_val=${value#\"}
         # remove double quote in the end
--- a/installer/x86_64/install.sh
+++ b/installer/x86_64/install.sh
@@ -20,6 +20,18 @@ _trap_push() {
 }
 _trap_push true
 
+read_conf_file() {
+    local conf_file=$1
+    while IFS='=' read -r var value
+    do
+        # remove double quote in the beginning
+        tmp_val=${value#\"}
+        # remove double quote in the end
+        value=${tmp_val%\"}
+        eval "$var=\"$value\""
+    done < "$conf_file"
+}
+
 # Main
 set -e
 cd $(dirname $0)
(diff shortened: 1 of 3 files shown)


installer/x86_64/install.sh
```diff
@@ -20,6 +20,21 @@ _trap_push() {
 }
 _trap_push true
 
+read_conf_file() {
+    local conf_file=$1
+    while IFS='=' read -r var value
+    do
+        f_char=$(echo $var | cut -c1)
```
qiluo-msft, line 27: Do you assume the `#` appear as first char of a line? If yes, can you assume it could appear anywhere on a line?

installer/x86_64/install.sh
```diff
@@ -20,6 +20,21 @@ _trap_push() {
...
+        [ "$f_char" = "#" ] && continue
+        [ -z "$var" ] && continue
+        # remove double quote in the beginning
+        tmp_val=${value#\"}
+        # remove double quote in the end
+        value=${tmp_val%\"}
+        eval "$var=\"$value\""
```
qiluo-msft, line 34: Is it possible to use `declare` instead of `eval` ?

qiluo-msft (commented):
As comments

installer/x86_64/install.sh
kuanyu99: Your suggestion is helpful. I modify this part to remove the string behind `#`.

installer/x86_64/install.sh
kuanyu99: I am not familiar with `declare`. What is the advantage to use `declare` instead of `eval`?

qiluo-msft: `declare` is safer than `eval`. I know it in bash, not sure about this environment (sh).

---
In reply to: [568218112](https://github.com/Azure/sonic-buildimage/pull/6600#discussion_r568218112) [](ancestors = 568218112)

installer/x86_64/install.sh
```diff
@@ -20,6 +20,25 @@ _trap_push() {
 }
 _trap_push true
 
+read_conf_file() {
```
qiluo-msft, line 23: >read_conf_file [](start = 0, length = 14)

Could you write a unit test to cover old cases and your new feature?

installer/x86_64/install.sh
kuanyu99: I tried it in sh environment. There is no `declare` command.

installer/x86_64/install.sh
kuanyu99: Sure. But what kind of the test will be better and suitable?
Add a test inside sonic-mgmt and be independent from building. Or, add additional process to test this during building.
I haven't found similar case of this type of unit test.

qiluo-msft: `read_conf_file()` is complex. I expect you have tested it locally. Could you just automate the test in another shell script or a python test runner.
In future we may either improve the code or the testcases.

---
In reply to: [568342160](https://github.com/Azure/sonic-buildimage/pull/6600#discussion_r568342160) [](ancestors = 568342160)

kuanyu99: I just copied the function to another script and test it with a sample `machine.conf`.
The test can only be triggered manually for now.

qiluo-msft (approved):

PR #6644, opened 2021-02-02
[platform][barefoot] Install sonic_platform to host

vboykox:

**- Why I did it**
SONiC design requires sonic_platform package to be installed in SONiC host environment, not only in docker containers.
**- How I did it**
For now, sonic_platform python wheel package, that is used by pmon, is provided via device-specific platform modules deb packages that unpacks the wheel package file into specific device's directory on lazy-install.
The PR makes deb packages' postinst script also install these unpacked wheel packages to host.
**- How to verify it**
python3 -c "import sonic_platform" should succeed when running on host.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [X] 202012

### DIFF
--- a/platform/barefoot/sonic-platform-modules-bfn-montara/debian/postinst
+++ b/platform/barefoot/sonic-platform-modules-bfn-montara/debian/postinst
@@ -1,0 +1,10 @@
+#!/bin/sh
+set -e
+
+PLATFORM_NAME=x86_64-accton_wedge100bf_32x-r0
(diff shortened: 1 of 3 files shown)


vboykox:
@jleveque

jleveque (approved):
Approving with a reminder that the Python 2 sonic-platform package should no longer be necessary. Suggest attempting to removing the Python 2 packages in a future PR to confirm and help advance the efforts to remove Python 2 from the SONiC codebase.

PR #6537, opened 2021-01-23
[teamd]: Increase wait timeout for teamd docker stop to clean Port channels.

judyjoseph:
**- Why I did it**
The Portchannels were not getting cleaned up as the cleanup activity was taking more than 10 secs which is default docker timeout after which a SIGKILL will be send.
Fixes https://github.com/Azure/sonic-buildimage/issues/6199
To check if it works out for this issue in 201911 ? https://github.com/Azure/sonic-buildimage/issues/6503

This issue is significantly seen in master branch compared to 201911 because the Portchannel cleanup takes more time in master. **Test on a DUT with 8 Port Channels.**

**master**
```
    admin@str-s6000-acs-8:~$ time sudo systemctl stop teamd
    real    0m15.599s
    user    0m0.061s
    sys     0m0.038s
```
**Sonic 201911.v58**
```
    admin@str-s6000-acs-8:~$ time sudo systemctl stop teamd
    real    0m5.541s
    user    0m0.020s
    sys     0m0.028s
 ```

**- How I did it**
Increased the timeout, by passing the timeout argument of 60sec now to container stop in case of teamd container.

**- How to verify it**
Checked the following cases and made sure the Portchannels are cleanup properly. Checked that in case of Warm restart of teamd, it doesn't use more time.

Case 1: 
Normal stop of teamd with the fix, used teh script below to try multilple times.
```
admin@str-s6000-acs-8:~$ cat a.sh 
#!/bin/bash

for i in {1..7}
do
  echo "Running $i times"
  sudo systemctl reset-failed swss.service
  sudo systemctl reset-failed syncd.service
  sudo systemctl reset-failed teamd.service
  time sudo systemctl stop teamd.service
  a=`ip link | grep PortChannel | wc -l`
  if [ $a -ne 0 ]
  then
    echo "FAILED"
  fi
  sleep 2m
done

admin@str-s6000-acs-8:~$ ./a.sh 
Running 1 times
real    0m17.692s
user    0m0.033s
sys     0m0.047s

Running 2 times
real    0m19.542s
user    0m0.059s
sys     0m0.039s

Running 3 times
real    0m19.171s
user    0m0.047s
sys     0m0.057s

Running 4 times
real    0m17.320s
user    0m0.061s
sys     0m0.038s

Running 5 times
real    0m16.903s
user    0m0.039s
sys     0m0.058s

Running 6 times
real    0m20.344s
user    0m0.042s
sys     0m0.040s

Running 7 times
real    0m19.468s
[...]

PR #6648, opened 2021-02-02
[teamd] Increase wait timeout for teamd docker stop to clean Port channels

judyjoseph:
**- Why I did it**
Put the fix https://github.com/Azure/sonic-buildimage/pull/6537 in 201911.

**- How I did it**

**- How to verify it**

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012

### DIFF
--- a/files/build_templates/docker_image_ctl.j2
+++ b/files/build_templates/docker_image_ctl.j2
@@ -366,7 +366,12 @@ wait() {
 }
 
 stop() {
+{%- if docker_container_name == "teamd" %}
+    # Longer timeout of 60 sec to wait for Portchannels to be cleaned.
+    docker stop -t 60 {{docker_container_name}}$DEV
+{%- else %}
     docker stop {{docker_container_name}}$DEV
+{%- endif %}
 {%- if docker_container_name == "database" %}
     if [ "$DEV" ]; then 
         ip netns delete "$NET_NS"


files/build_templates/docker_image_ctl.j2
```diff
@@ -366,7 +366,12 @@ wait() {
 }
 
 stop() {
+{%- if docker_container_name == "teamd" %}
+    # Longer timeout of 60 sec to wait for Portchannels to be cleaned.
+    docker stop -t 60 {{docker_container_name}}$DEV
```
nazariig, line 371: @judyjoseph what about default systemd service timeout (90 sec)? Do we need to align also?

judyjoseph: @nazariig will prefer to keep it in sync with master, but we can tune this. 
During my tests I found the cleanup was quick in 201911 and hence even 60sec is a high  value. But if you find that the issue #6537, where you have a large number of POs it takes more time - we can increase it - is that ok ?

nazariig: @judyjoseph ok. Thanks for the clarifications.

nazariig (approved):

judyjoseph:
Adding this PR as reference. https://github.com/Azure/sonic-buildimage/pull/6757

judyjoseph:
retest this please

judyjoseph:
retest this please

PR #6650, opened 2021-02-03
[ci] Archive compiled Debian packages and Python wheels

jleveque:
Archive compiled Debian packages and Python wheels so that the artifacts can be downloaded and used by other pipelines.
Also archive related log files.


qiluo-msft:
Could you also add target.log and target-install.log?

jleveque:
> Could you also add target.log and target-install.log?

Done in latest commit.

qiluo-msft (approved):

lguohan (approved):

PR #6670, opened 2021-02-03
[201911][Arista] Mitigate rare powercycle on SEU scenario

Staphylo:
On the `DCS-7060CX-32S`, a SEU can happen on a CPLD which by default would reboot the platform.
Other SEU scenarios are already handled but this one was missed since it's specific to this platform.
It's a pretty rare case which will now be reported in the syslog the same way others are.


jleveque:
Retest vsimage please

jleveque (approved):

PR #6686, opened 2021-02-05
[sonic-swss-common] Update submodule

qiluo-msft:
Includes below commits
```
9e91e0d 2021-02-04 | Fix: RedisClient.set() should return bool (#454) [Qi Luo]
244a0bf 2021-02-04 | [swig] Add missing functions for ConfigDBConnector (#453) [Qi Luo]
e659698 2021-01-22 | [dbconnector] Make sure json is declared as non null object (#452) [Kamil Cudnik]
```


lguohan:
can you check the build failure?

lguohan:
looks like your change breaks the sonic-swss

qiluo-msft:
Fixed in https://github.com/Azure/sonic-swss-common/pull/454

lguohan (approved):

PR #6756, opened 2021-02-10
Change the order of stopping dependent services of SWSS on 201911

shlomibitton:

**- Why I did it**
When large number of port channels (more than 64) is configured, a config reload command might ends with not all port channel configured and up. Further debug shows that unloading the port channels on the ASIC driver take a lot of time.
With the change, deleting all port channels before the syncd restart will free resources better and the ASIC driver will unload all netdev fast and the operation will execute properly.

**- How I did it**
Change the order of dependent services of SWSS.

**- How to verify it**
Before the fix: 
- Configure more than 80 port channel each with IP address.
- Run 'config reload' command. ASIC driver failure observed and not all port channels are configured properly and up. 

After the fix:
- Configure more than 80 port channel each with IP address.
- Run 'config reload' command. No errors on ASIC driver, all port channel configured and up.

**Which release branch to backport (provide reason below if selected)**
[ ] 201811
[ ] 201911
[ ] 202006
[ ] 202012

**Description for the changelog** 

**A picture of a cute animal (not mandatory but encouraged)**

### DIFF
--- a/files/scripts/swss.sh
+++ b/files/scripts/swss.sh
@@ -101,14 +101,6 @@ start_peer_and_dependent_services() {
 stop_peer_and_dependent_services() {
     # if warm start enabled or peer lock exists, don't stop peer service docker
     if [[ x"$WARM_BOOT" != x"true" ]]; then
-        if [[ ! -z $DEV ]]; then
-            /bin/systemctl stop ${PEER}@$DEV
-        else
-            /bin/systemctl stop ${PEER}
-        fi
-        for dep in ${DEPENDENT}; do
-            /bin/systemctl stop ${dep}
-        done
         for dep in ${MULTI_INST_DEPENDENT}; do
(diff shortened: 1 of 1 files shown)


liat-grozovik (approved):

liat-grozovik:
@abdosi please review. we believe this is needed for 201911 production env.

abdosi:
@yxieca and @judyjoseph please review this.

yxieca (approved):

judyjoseph:
The change looks ok, Could you add the results you observed with and without this change. thanks.

liat-grozovik:
@judyjoseph the comment was updated with more information and how it was verified. please let us know if something is missing, otherwise we strongly suggest to take it to 201911

judyjoseph (approved):

PR #6759, opened 2021-02-10
[swss]: Clear MUX-related state DB tables on start

theasianpianist:

**- Why I did it**
Port configuration changes can leave keys for unused interfaces behind in STATE DB

**- How I did it**
* Add *MUX_CABLE_TABLE* to set of tables to clear on SWSS start, which
will clear HW_MUX_CABLE_TABLE and MUX_CABLE_TABLE
* Order swss to start before pmon to ensure that DBs are cleared before
xcvrd (running inside pmon) starts and re-populates the tables

**- How to verify it**

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012

image


prsunny (commented):
cleanup table lgtm.. can @vdahiya12 , @lguohan to check service dependency.

vdahiya12 (approved):

lguohan:
@jleveque , now pmon is affected by swss.

jleveque:
> @jleveque , now pmon is affected by swss.

Thanks for the heads-up!

lguohan (approved):

theasianpianist:
@jleveque This dependency should not prevent PMON from starting if SWSS fails, since `Before=` does not enforce any dependencies on the health of the SWSS service, it only waits until SWSS has either completely started or failed: https://www.freedesktop.org/software/systemd/man/systemd.unit.html#Before=

jleveque (approved):

lguohan:
in the chassis supervisor card environment, the pmon needs to before swss.

PR #6765, opened 2021-02-11
[multi-asic] Enhanced iptable default rules

abdosi:
What I did:-
- For multi-asic platforms added iptable v4 rule to communicate on docker bridge ip
- For multi-asic platforms extend iptable v4 rule for iptable v6 also
-  For multi-asic program made all internal rules applicable for all protocols (not filter based on tcp/udp). This is done to be consistent same as local host rule
- For multi-asic platforms made nat rule (to forward traffic from namespace to host) generic for all protocols and also use Source IP if present for matching

Why I did:
- We need iptable v4 rule to communicate on docker bridge ip since syslog on multi-asic platforms uses docker bridge ip and for 
  host docker syslog is sent with both source and destination ip as docker bridge ip
- iptable v6 rules are added as  enhancement and to be symmetric as iptable v4 rules.
- Nat rule changes were triggered because even SSH has same use case as SNMP (https://github.com/Azure/sonic-buildimage/pull/5420)
- Need to qualify Source IP with NAT rules is needed since if we load explicit filter rule since we are doing NAT INPUT table will not be used and to make sure we only NAT the valid traffic.

How I verify:
Verified on single and multi-asic platforms

```
Default table  INPUT Rules on host for multi-asic platforms

sudo iptables -L -n
Chain INPUT (policy ACCEPT)
target     prot opt source               destination
ACCEPT     all  --  127.0.0.1            0.0.0.0/0
ACCEPT     all  --  240.127.1.1          240.127.1.1
ACCEPT     all  --  240.127.1.6          240.127.1.1
ACCEPT     all  --  240.127.1.7          240.127.1.1
ACCEPT     all  --  240.127.1.4          240.127.1.1
ACCEPT     all  --  240.127.1.2          240.127.1.1
ACCEPT     all  --  240.127.1.3          240.127.1.1
ACCEPT     all  --  240.127.1.5          240.127.1.1

Default table  INPUT Rules  In Namespace

sudo ip netns exec asic3 iptables -L -n
Chain INPUT (policy ACCEPT)
target     prot opt source               destination
ACCEPT     all  --  127.0.0.1            0.0.0.0/0
ACCEPT     all  --  240.127.1.3          240.127.1.3
ACCEPT     all  --  240.127.1.1          240.127.1.3

Default NAT Table Rules  In Namespace

sudo ip netns exec asic0 iptables -t nat -L -n
Chain PREROUTING (policy ACCEPT)
target     prot opt source               destination
DNAT       tcp  --  0.0.0.0/0              0.0.0.0/0            tcp dpt:161 to:240.127.1.1
DNAT       tcp  --  0.0.0.0/0            0.0.0.0/0            tcp dpt:22 to:240.127.1.1

NAT  Rule after loading below acl.json

sudo ip netns exec asic0 iptables -t nat -L -n
Chain PREROUTING (policy ACCEPT)
target     prot opt source               destination
DNAT       tcp  --  1.1.1.1              0.0.0.0/0            tcp dpt:161 to:240.127.1.1
DNAT       tcp  --  0.0.0.0/0            0.0.0.0/0            tcp dpt:22 to:240.127.1.1

{
    "acl": {
        "acl-sets": {
            "acl-set": {
                "SNMP-ACL": {
                    "acl-entries": {
                        "acl-entry": {
                            "1": {
                                "actions": {
                                    "config": {
                                        "forwarding-action": "ACCEPT"
                                    }
                                },
                                "config": {
                                    "sequence-id": 1
                                },
                                "ip": {
                                    "config": {
                                        "protocol": "IP_UDP",
                                        "source-ip-address": "1.1.1.1/32"
                                    }
                                }
                            }
                        }
                    },
                    "config": {
                        "name": "SNMP-ACL"
                    }
                }
            }
        }
    }
```

### DIFF
--- a/src/sonic-host-services/scripts/caclmgrd
+++ b/src/sonic-host-services/scripts/caclmgrd
@@ -277,19 +277,12 @@ class ControlPlaneAclManager(daemon_base.DaemonBase):
 
         for acl_service in self.ACL_SERVICES:
             if self.ACL_SERVICES[acl_service]["multi_asic_ns_to_host_fwd"]:
+                    # Get the Source IP Set if exists else use default source ip prefix
+                    nat_source_ipv4_set = acl_source_ip_map[acl_service]["ipv4"] if acl_source_ip_map and acl_source_ip_map[acl_service]["ipv4"] else { "0.0.0.0/0" }
+                    nat_source_ipv6_set = acl_source_ip_map[acl_service]["ipv6"] if acl_source_ip_map and acl_source_ip_map[acl_service]["ipv6"] else { "::/0" }
+
                 for ip_protocol in self.ACL_SERVICES[acl_service]["ip_protocols"]:
                     for dst_port in  self.ACL_SERVICES[acl_service]["dst_ports"]: 
-                        # Get the Source IP Set if exists else use default source ip prefix
-                        try:
-                            nat_source_ipv4_set = acl_source_ip_map[acl_service]["ipv4"]
-                        except:
-                            nat_source_ipv4_set = { "0.0.0.0/0" }
-
-                        try:
-                            nat_source_ipv6_set = acl_source_ip_map[acl_service]["ipv6"]
-                        except:
-                            nat_source_ipv6_set = { "::/0" }
-
                         for ipv4_src_ip in nat_source_ipv4_set:
                             # IPv4 rules
                             fwd_traffic_from_namespace_to_host_cmds.append(self.iptables_cmd_ns_prefix[namespace] +
--- a/src/sonic-host-services/scripts/caclmgrd
+++ b/src/sonic-host-services/scripts/caclmgrd
@@ -226,13 +226,31 @@ class ControlPlaneAclManager(daemon_base.DaemonBase):
             allow_internal_docker_ip_cmds.append(self.iptables_cmd_ns_prefix[namespace] + "iptables -A INPUT -s {} -d {} -j ACCEPT".format
                                                 (self.namespace_docker_mgmt_ip[namespace], self.namespace_docker_mgmt_ip[namespace]))
 
+            allow_internal_docker_ip_cmds.append(self.iptables_cmd_ns_prefix[namespace] + "ip6tables -A INPUT -s {} -d {} -j ACCEPT".format
+                                                (self.namespace_docker_mgmt_ipv6[namespace], self.namespace_docker_mgmt_ipv6[namespace]))
+
             # For namespace docker allow all tcp/udp traffic from host docker bridge to its eth0 management ip
             allow_internal_docker_ip_cmds.append(self.iptables_cmd_ns_prefix[namespace] + "iptables -A INPUT -p tcp -s {} -d {} -j ACCEPT".format
                                                  (self.namespace_mgmt_ip, self.namespace_docker_mgmt_ip[namespace]))
 
             allow_internal_docker_ip_cmds.append(self.iptables_cmd_ns_prefix[namespace] + "iptables -A INPUT -p udp -s {} -d {} -j ACCEPT".format
                                                  (self.namespace_mgmt_ip, self.namespace_docker_mgmt_ip[namespace]))
+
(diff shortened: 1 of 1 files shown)


lguohan:
is the test expect to fail?

src/sonic-host-services/scripts/caclmgrd
```diff
@@ -226,20 +226,46 @@ class ControlPlaneAclManager(daemon_base.DaemonBase):
...
+            
             # In host allow all tcp/udp traffic from namespace docker eth0 management ip to host docker bridge
             for docker_mgmt_ip in list(self.namespace_docker_mgmt_ip.values()):
                 allow_internal_docker_ip_cmds.append(self.iptables_cmd_ns_prefix[namespace] + "iptables -A INPUT -p tcp -s {} -d {} -j ACCEPT".format
                                                      (docker_mgmt_ip, self.namespace_mgmt_ip))
 
                 allow_internal_docker_ip_cmds.append(self.iptables_cmd_ns_prefix[namespace] + "iptables -A INPUT -p udp -s {} -d {} -j ACCEPT".format
                                                      (docker_mgmt_ip, self.namespace_mgmt_ip))
+
+            for docker_mgmt_ipv6 in list(self.namespace_docker_mgmt_ipv6.values()):
```
judyjoseph, line 262: LGTM. 
One improvement in these two for loops where we actually create separate iptable rule per namespace from the host ip, we could actually reduce the number of rules into 1 and give the list of ip namespaces addresses separated by commas (-s 240.127.1.6,240.127.1.7..)

abdosi:
@judyjoseph please review again. This has changes to NAT rules also as we discussed offline.

abdosi:
> is the test expect to fail?

@lguohan yes it should not fail. Fixed it.

src/sonic-host-services/scripts/caclmgrd
```diff
@@ -226,54 +230,80 @@ class ControlPlaneAclManager(daemon_base.DaemonBase):
...
+            if self.ACL_SERVICES[acl_service]["multi_asic_ns_to_host_fwd"]:
+                # Get the Source IP Set if exists else use default source ip prefix
+                nat_source_ip_map = acl_source_ip_map[acl_service] if acl_source_ip_map and acl_service in acl_source_ip_map else None
+                nat_source_ipv4_set = nat_source_ip_map["ipv4"] if nat_source_ip_map and "ipv4" in nat_source_ip_map and nat_source_ip_map["ipv4"] else { "0.0.0.0/0" } 
+                nat_source_ipv6_set = nat_source_ip_map["ipv6"] if nat_source_ip_map and "ipv6" in nat_source_ip_map and nat_source_ip_map["ipv6"] else { "::/0" } 
+
+                for ipv4_src_ip in nat_source_ipv4_set:
+                    # IPv4 rules
+                    fwd_traffic_from_namespace_to_host_cmds.append(self.iptables_cmd_ns_prefix[namespace] +
+                                                       "iptables -t nat -A PREROUTING -p {} -s {} --dport {}  -j DNAT --to-destination {}".format
```
judyjoseph, line 288: Here to add an additional loop for protocols, eg: in case of snmp service it supports both tcp, udp.

abdosi: @judyjoseph updated.

abdosi: Thanks @judyjoseph  updated to handle empty ip set condition.

abdosi:
will move below change to separate PR. this change has dependency with sonic-mgmt test case failure. would need to disable the test case/make the change/enable test case again.

- For SNMP remove TCP protocol. We always listen on UDP https://github.com/Azure/sonic-buildimage/blob/d769ef2abd75b2da48bef806b91b9a369b1ecbba/dockers/docker-snmp/snmpd.conf.j2#L23

abdosi:
@judyjoseph can you please review again.

judyjoseph (approved):

PR #6782, opened 2021-02-15
[Mellanox] Add hw-mgmt patch for SimX platform adaptation

shlomibitton:

#### Why I did it
System is stuck on 'starting' state on SimX platform because of infinite loop on 'hw-management-ready.sh' script .
The loop is polling to check if the hw-mgmt sysfs created before proceeding with the flow, for SimX platform the sysfs will never create so the system is not starting properly.

#### How I did it
Add a condition to poll on hw-mgmt sysfs only if the switch is real HW and not SimX platform.

#### How to verify it
Check "systemctl status hw-management.service" output on a SimX switch with this patch, the state will be "active".

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012

### DIFF
--- a/platform/mellanox/hw-management/0003-Make-hw-mgmt-SimX-compatiable.patch
+++ b/platform/mellanox/hw-management/0003-Make-hw-mgmt-SimX-compatiable.patch
@@ -20,7 +20,7 @@ index 3c9f7b6..05d143f 100755
  echo "Start Chassis HW management service."
  logger -t hw-management -p daemon.notice "Start Chassis HW management service."
 diff --git a/usr/usr/bin/hw-management.sh b/usr/usr/bin/hw-management.sh
-index 70f1297..8de1722 100755
+index 70f1297..e427a3d 100755
 --- a/usr/usr/bin/hw-management.sh
 +++ b/usr/usr/bin/hw-management.sh
 @@ -1110,6 +1110,13 @@ do_chip_down()
@@ -29,7 +29,7 @@ index 70f1297..8de1722 100755
  
 +check_simx()
 +{
-+        if [ ! -z "$(lspci -vvv | grep SimX)" ]; then
++        if [ -n "$(lspci -vvv | grep SimX)" ]; then
 +                exit 0
 +        fi
 +}
--- a/platform/mellanox/hw-management/0003-Make-hw-mgmt-SimX-compatiable.patch
+++ b/platform/mellanox/hw-management/0003-Make-hw-mgmt-SimX-compatiable.patch
@@ -1,0 +1,48 @@
+diff --git a/usr/usr/bin/hw-management-ready.sh b/usr/usr/bin/hw-management-ready.sh
+index 3c9f7b6..05d143f 100755
+--- a/usr/usr/bin/hw-management-ready.sh
++++ b/usr/usr/bin/hw-management-ready.sh
+@@ -49,9 +49,12 @@ if [ -d /var/run/hw-management ]; then
+ 	rm -fr /var/run/hw-management
+ fi
+ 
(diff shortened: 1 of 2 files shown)


liat-grozovik (dismissed):

platform/mellanox/hw-management/0003-Make-hw-mgmt-SimX-compatiable.patch
```diff
@@ -0,0 +1,21 @@
...
+ fi
+ 
+-while [ ! -d /sys/devices/platform/mlxplat/mlxreg-hotplug/hwmon ]
+-do
+-	sleep 1
+-done
++if [ "$(cat /sys/devices/virtual/dmi/id/sys_vendor)" != "QEMU" ]; then
++	while [ ! -d /sys/devices/platform/mlxplat/mlxreg-hotplug/hwmon ]
++	do
++		sleep 1
```
nazariig, line 16: @shlomibitton why do we need to sleep? Why not simply return? Does it mean that hw-management service will always be in starting condition?

shlomibitton: The loop is polling to check if the hw-mgmt sysfs created before proceeding with the flow.
AFAIK, we want to wait until it is valid before exiting this script, this is the previous implementation.
The fix is only to skip it if this is a SimX platform.

dprital:
@lguohan - Can you please review ?

platform/mellanox/hw-management/0003-Make-hw-mgmt-SimX-compatiable.patch
nazariig: @shlomibitton `"$(cat /sys/devices/virtual/dmi/id/sys_vendor)" != "QEMU"` this doesn't work anymore. Please use `lspci` instead.

nazariig (changes requested):

platform/mellanox/hw-management/0003-Make-hw-mgmt-SimX-compatiable.patch
```diff
@@ -0,0 +1,48 @@
...
+index 70f1297..bb45ab9 100755
+--- a/usr/usr/bin/hw-management.sh
++++ b/usr/usr/bin/hw-management.sh
+@@ -1110,6 +1110,13 @@ do_chip_down()
+ 	/usr/bin/hw-management-thermal-events.sh change hotplug_asic down %S %p
+ }
+ 
++check_simx()
++{
++        if lspci -vvv | grep SimX &> /dev/null; then
```
nazariig, line 32: @shlomibitton `if [ -n "$(lspci -vvv | grep SimX)" ]; then`

platform/mellanox/hw-management/0003-Make-hw-mgmt-SimX-compatiable.patch
```diff
@@ -0,0 +1,48 @@
...
++++ b/usr/usr/bin/hw-management-ready.sh
+@@ -49,9 +49,12 @@ if [ -d /var/run/hw-management ]; then
+ 	rm -fr /var/run/hw-management
+ fi
+ 
+-while [ ! -d /sys/devices/platform/mlxplat/mlxreg-hotplug/hwmon ]
+-do
+-	sleep 1
+-done
++if [ "$(lspci -vvv | grep SimX)" == "" ]; then
```
nazariig, line 13: @shlomibitton `if [ -z "$(lspci -vvv | grep SimX)" ]; then`

nazariig (approved):

liat-grozovik (approved):

PR #6786, opened 2021-02-16
[sonic-device-data]: add optional port_type as valid identifier for hwsku.json 

carl-nokia:
#### Why I did it
enable automated test suites to selectively run relevant tests ( or not run tests ) based upon a new port_type identifier in hwsku.json
#### How I did it
Modified the valid optional fields in validity check for hwsku.json per recommendation from Joe  in PR 2654

jleveque on Dec 22, 2020 • Contributor
Maybe we could utilize an existing data file like hwsku.json? If we could specify something there that a port is non-SFP, we could skip those ports here.
#### How to verify it
Added a hwsku.json with the new optional port type "port_type" in addition to the required "default_brkout_mode"

backport to 202012 so that the enhanced automation tests can skip copper ports for sfp tests 

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012


src/sonic-device-data/tests/hwsku_json_checker
```diff
@@ -7,7 +7,7 @@ import sys
 
 # Global variable
 PORT_ATTRIBUTES = ["default_brkout_mode"]
-OPTIONAL_PORT_ATTRIBUTES = ["fec", "autoneg"]
+OPTIONAL_PORT_ATTRIBUTES = ["fec", "autoneg","port_type"]
```
jleveque, line 10: Add space after comma.

jleveque (changes requested):

src/sonic-device-data/tests/hwsku_json_checker
carl-nokia: Joe - given that platform.json has been expanded to include "breakout_modes", "index", etc... would it make more sense to add "port_type" there instead of hwsku.json

jleveque: It depends on whether this is something that is fixed at the platform level and will apply to all HwSKUs configured on that platform, or whether it will vary by HwSKU.

carl-nokia:
new PR # 8370 replaces this PR

PR #6824, opened 2021-02-19
[frr] protect template if deployment_id is missing in map when generating dynamic bgp neighbor

qiluo-msft:
#### Why I did it
This is to protect a corner case: if deployment_id is missing in map when generating dynamic bgp neighbor, the template will generate an incomplete line.

```
neighbor dyn_name remote-as
```

#### How I did it
Remove the whole line in the case.

#### How to verify it
Pass the sonic-bgpcfgd unit test. Tested the corner case manually

- [ ] 201811
- [x] 201911
- [x] 202006
- [x] 202012


abdosi (approved):

abdosi (changes requested):
@qiluo-msft we can add unit-test case for this.

abdosi (commented):
can we add unit-test case for this ?

PR #6828, opened 2021-02-19
 .gitmodules: Fix make init

donNewtonAlpha:
* fix url for
* submodule "platform/mellanox/hw-management/hw-mgmt"
* remove trailing / from url

Signed-off-by: Don Newton don@opennetworking.org

#### Why I did it
During make init submodule`` "platform/mellanox/hw-management/hw-mgmt" fails to clone due to trailing / on the url

#### How I did it
Removed the trailing /
#### How to verify it
`make init` and look for errors

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x ] 202012
ONF is using the 202012 branch as the base for our effort

Correct the url line for submodule "platform/mellanox/hw-management/hw-mgmt"

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -59,7 +59,7 @@
 	url = https://github.com/aristanetworks/sonic
 [submodule "platform/mellanox/hw-management/hw-mgmt"]
 	path = platform/mellanox/hw-management/hw-mgmt
-	url = https://github.com/Mellanox/hw-mgmt/
+	url = https://github.com/Mellanox/hw-mgmt
 [submodule "src/redis-dump-load"]
 	path = src/redis-dump-load
 	url = https://github.com/p/redis-dump-load.git


jleveque (approved):
Thanks for the fix. I'm surprised we haven't noticed any issues to date. Must be due to differences in our environments. According to the Git [documentation](https://www.git-scm.com/docs/gitmodules#Documentation/gitmodules.txt-submoduleltnamegtpath), the path must not end with a trailing slash.

donNewtonAlpha:
> Thanks for the fix. I'm surprised we haven't noticed any issues to date. Must be due to differences in our environments. According to the Git [documentation](https://www.git-scm.com/docs/gitmodules#Documentation/gitmodules.txt-submoduleltnamegtpath), the path must not end with a trailing slash.

It doesn't actually break `make init` it just fails on that step.  I imagine you would never really notice unless you were deploying to a mellanox switch.

PR #6835, opened 2021-02-21
Graceful reboot in S6000

rkdevi27:
#### Why I did it
The S6000 devices, the cold reboot is abrupt and it is likely to cause issues which will cause the device to land into EFI shell. Hence the platform reboot will happen after graceful unmount of all the filesystems as in S6100.
#### How I did it
Moved the platform_reboot to platform_reboot_override and hooked it to the systemd shutdown services as in S6100
#### How to verify it
Issue "reboot" command to verify if the reboot is happening gracefully.

- [ ] 201811
- [ ] 201911
- [x] 202006
- [x] 202012

Dell S6000 graceful reboot
[UT_logs.txt](https://github.com/Azure/sonic-buildimage/files/6017256/UT_logs.txt)


jleveque (approved):

lguohan:
how is this tested in other branches? for example 201811,i see it is using python3, but it is available in 201811?

PR #6846, opened 2021-02-22
[minigraph] Support tagged VlanInterface if attached to multiple vlans

qiluo-msft:
Same as https://github.com/Azure/sonic-buildimage/pull/6833
But adapted for 201911 branch

### DIFF
--- a/src/sonic-config-engine/minigraph.py
+++ b/src/sonic-config-engine/minigraph.py
@@ -350,6 +350,13 @@ def parse_dpg(dpg, hname):
         vlan_intfs = []
         vlans = {}
         vlan_members = {}
+        intf_vlan_mbr = defaultdict(list)
+        for vintf in vlanintfs.findall(str(QName(ns, "VlanInterface"))):
+            vlanid = vintf.find(str(QName(ns, "VlanID"))).text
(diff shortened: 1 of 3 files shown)


daall (approved):

src/sonic-config-engine/minigraph.py
```diff
@@ -358,7 +365,10 @@ def parse_dpg(dpg, hname):
             for i, member in enumerate(vmbr_list):
                 vmbr_list[i] = port_alias_map.get(member, member)
                 sonic_vlan_member_name = "Vlan%s" % (vlanid)
-                vlan_members[(sonic_vlan_member_name, vmbr_list[i])] = {'tagging_mode': 'untagged'}
+                if len(intf_vlan_mbr[member]) > 1:
+                    vlan_members[(sonic_vlan_member_name, vmbr_list[i])] = {'tagging_mode': 'tagged'}
```
prsunny, line 369: If there are multiple vlans, does that mean, there are no untagged member? I mean, tagging determines the pvid of port. So all untagged packets on this port would be dropped. Is that the expectation?

qiluo-msft: yes

prsunny (approved):

PR #6865, opened 2021-02-23
[minigraph] set admin_status to down if port not in minigraph

dmytroxshevchuk:

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012


src/sonic-config-engine/minigraph.py
```diff
@@ -1090,6 +1092,7 @@ def parse_xml(filename, platform=None, port_config_file=None, asic_name=None, hw
     port_speeds_default = {}
     port_speed_png = {}
     port_descriptions = {}
+    port_alias_list = []
```
prsunny, line 1095: Do we need to create a new list or can we use existing `port_alias_map`?.

src/sonic-config-engine/minigraph.py
```diff
@@ -1346,6 +1349,11 @@ def parse_xml(filename, platform=None, port_config_file=None, asic_name=None, hw
     for port in ports.values():
         port['pfc_asym'] = 'off'
 
+    # make admin status to down if port not parsed from minigraph
+    for port in ports.values():
+        if port['alias'] not in port_alias_list:
```
prsunny, line 1354: Does this impact multi-asic platforms where there are backplane ports? @arlakshm ?

src/sonic-config-engine/minigraph.py
dmytroxshevchuk: `port_alias_map` contains aliases from config_db, so we cannot use it

src/sonic-config-engine/tests/sample_output/platform_output.json
```diff
@@ -810,7 +810,7 @@
     "index": "7",
     "lanes": "26,27",
     "description": "Eth7/2",
-    "admin_status": "up",
+    "admin_status": "down",
```
prsunny, line 813: Why is the test modified to have all port set to "down"?. I would expect test some port down and some up.

dmytroxshevchuk: Ports described in minigraph.xml as EthernetInterface has up admin_status:
Ethernet8
Ethernet9
Ethernet36
Ethernet98
Ethernet0
Ethernet6
Ethernet4
Ethernet109
Ethernet108

src/sonic-config-engine/minigraph.py
```diff
@@ -1364,6 +1364,11 @@ def parse_xml(filename, platform=None, port_config_file=None, asic_name=None, hw
     for port in ports.values():
         port['pfc_asym'] = 'off'
 
+    # make admin status to down if port not parsed from minigraph
+    for port in ports.keys():
+        if 'admin_status' in ports[port] and port not in port_speeds_default:
```
prsunny, line 1369: This check if `'admin_status' in ports[port]`. What would be the value if its present? Where is that default value coming from?

PR #6885, opened 2021-02-25
[minigraph] Adjust minigraph parsing per Fine Grained ECMP matchmode changes

anish-n:
#### Why I did it
To adjust config db generated via minigraph per matchmode changes in fine grained ecmp. The changes are done so that nexthop IP based filtering can occur to determine routes as requiring Fine Grained ECMP, in the past the only mode was to use the IP prefix of the route for filtering, with this matchmode change we will use nexthop IP based filtering
https://github.com/Azure/SONiC/pull/727
#### How I did it
Change will modify config db entry created for FG_NHG to include 'match_mode': 'nexthop-based' so that nexthop IP based filtering can occur to determine routes as requiring Fine Grained ECMP. Changes also remove FG_NHG_PREFIX entry since its not needed under matchmode nexthop-based.

#### How to verify it
sonic-config-engine unit test 

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012

### DIFF
--- a/src/sonic-config-engine/minigraph.py
+++ b/src/sonic-config-engine/minigraph.py
@@ -146,7 +146,6 @@ def formulate_fine_grained_ecmp(version, dpg_ecmp_content, port_device_map, port
         tag = "fgnhg_v6"
 
     port_nhip_map = dpg_ecmp_content['port_nhip_map']
-    nhgaddr = dpg_ecmp_content['nhgaddr']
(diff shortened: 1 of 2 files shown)


kktheballer (dismissed):
Looks good to me apart from the UT failure. Maybe some ordering issue?

abdosi (approved):

PR #6896, opened 2021-02-26
[openconfig_acl] Add SONiC ACL extension to open config ACL model

daall:
- Add support for VLAN ID match
- Add support for ICMP type/code match

#### Why I did it
To allow users to add ACL rules w/ ICMP and VLAN qualifiers via acl-loader.

#### How I did it
Added `sonic-acl-extension.yang` to augment the existing OpenConfig spec and used pyangbind to generate a new set of python bindings.

#### How to verify it
Re-run the ACL tests w/ the new openconfig_acl.py script and verify that all rules are still loaded properly and dataplane behavior is not impacted.

- [ ] 201811
- [] 201911
- [ ] 202006
- [x] 202012


src/sonic-config-engine/sonic-acl-extension.yang
```diff
@@ -0,0 +1,76 @@
...
+	    }
+    }
+  }
+
+  augment "/oc-acl:acl/oc-acl:acl-sets/oc-acl:acl-set/oc-acl:acl-entries/oc-acl:acl-entry/oc-acl:l2/oc-acl:config" {
+    uses extended-l2-match;
+  }
+
+  augment "/oc-acl:acl/oc-acl:acl-sets/oc-acl:acl-set/oc-acl:acl-entries/oc-acl:acl-entry" {
+	  uses icmp-protocol-fields-top;
```
qiluo-msft, line 74: Indentation not consistent

daall: Fixed, thanks!

qiluo-msft (approved):

lguohan:
openconfig_acl.py seems generated code, can we generate the code during the build?

daall:
> openconfig_acl.py seems generated code, can we generate the code during the build?

I would prefer to merge this as-is to unblock the feature and then refine the process to generate the code at build-time.

PR #6901, opened 2021-02-26
[multi-asic] Changes in FRR templates 

abdosi:
What/Why I did:

- Made the command `next-hop-self force`  only applicable on back-end asic bgp. This is done so that BGPL iBGP session running on backend can send e-BGP learn nexthop. Back end asic FRR  is able to recursively resolve the eBGP nexthop in its routing table since it knows about all the connected routes advertise from front end asic.

- Made all front-end asic bgp use global loopback ip (Loopback0) as router id and back end asic bgp use Loopbacl4096 as ruter-id and originator id for Route-Reflector. This is done so that routes learnt by external peer do not see Loopback4096 as router id in `show ip bgp <route-prerfix>` output.

- To handle above change need to pass Loopback4096 from BGP manager for jinja2 template generation. This was missing and this change/fix is needed for this also https://github.com/Azure/sonic-buildimage/blob/master/dockers/docker-fpm-frr/frr/bgpd/templates/dynamic/instance.conf.j2#L27

- Enhancement to add mult_asic specific bgpd template generation unit test cases.

How I verify:

- Manual Verification of routes for `next-hop-self force` change . Eg on Back-end asic e-BGP route

Before:
Nexthop is changes to internal ip interface (10.1.0.5)

image

After:
e-BGP Peer as nexthop (10.0.0.9) available on backend asic and route nexthop is resolved recursively.

image

image

- Manual Verification for router-id change . Route on external neighbor

Before:
Using Loopback4096 IP (8.0.00) as router-id.
image

After:
Using Global Loopback IP (10.1.0.32) as router id.\
image

- Verify above for v6 also. Ran sonic-mgmt  test_allow_list.py on this changes.

- Added Unit test cases.


abdosi:
@arlakshm can you please review this.

arlakshm (approved):
lgtm

PR #6912, opened 2021-02-26
[swss]: Restore cached FDB and ARP entries after config reload for dual ToR

theasianpianist:
* Only performed after config reload on dual ToR systems

In conjuction with: https://github.com/Azure/sonic-utilities/pull/1465

#### Why I did it
On dual ToR systems, we need to quickly restore the ARP table after config reload so the standby ToR is able to continue normal operation.

#### How I did it
Check for the `/host/config-reload` directory, which is created by the linked PR. If found, copy the ARP and FDB caches to SWSS. Restore FDB entries to APP_DB. Then wait for the VLAN interface to come up, and restore ARP table to the kernel.

#### How to verify it
Ran `config reload -y` on a dual ToR testbed including changes from the linked PR, confirmed that ARP table on standby ToR was restored after.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012

image


dockers/docker-orchagent/kernel_arp_restore.sh
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env bash
+
+function wait_for_intf_up {
+    INTF_NAME=$1
+
+    until [[ `ip link show $INTF_NAME | grep 'state UP'` ]]; do
+        sleep 0.1;
```
prsunny, line 7: This is too minute. I would suggest wait 1 sec. Also, if its never up, we need a way to exit the script after say 5mts or so.

dockers/docker-orchagent/swssconfig.sh
```diff
@@ -29,6 +29,26 @@ function fast_reboot {
...
+    then
+      if [[ -f /fdb.json ]];
+      then
+        swssconfig /fdb.json
+        mv -f /fdb.json /fdb.json.1
+      fi
+
+      if [[ -f /arp.json ]];
+      then
+        swssconfig /arp.json
```
prsunny, line 43: I think, may be we should not restore the `arp.json` entries to APP_DB since we are directly adding to kernel. If we add to APP_DB, arp_refresh script kick in and can result in failed entries temporarily.

dockers/docker-orchagent/kernel_arp_restore.sh
```diff
@@ -0,0 +1,44 @@
...
+
+        # Split the 'NEIGH_TABLE' key with delimiter ':' and take the
+        # second item from the result array which is the device name
+        DEVICE=`echo $KEY | jq -r ". / \":\" | .[1]"`
+
+        # Same as for VLAN, but take the third item which is the IP
+        IP=`echo $KEY | jq -r ". / \":\" | .[2]"`
+
+        wait_for_intf_up $DEVICE
+        ip neigh replace "$IP" dev "$DEVICE" lladdr "$MAC" nud stale
```
prsunny, line 37: if we are restoring IPv6 entries, this should be "ip -6 neigh ..."

dockers/docker-orchagent/kernel_arp_restore.sh
theasianpianist: Ack, will look into a way to timeout

dockers/docker-orchagent/swssconfig.sh
theasianpianist: Ack, will fix

dockers/docker-orchagent/kernel_arp_restore.sh
theasianpianist: I thought `ip neigh` by default could accept both IPv4 and IPv6 entries?

dockers/docker-orchagent/kernel_arp_restore.sh
```diff
@@ -0,0 +1,44 @@
...
+function restore_arp_to_kernel {
+    ARP_FILE='/arp.json.1'
+    NUM_ENTRIES=`jq 'length' $ARP_FILE`
+
+    for i in $( seq 0 $(($NUM_ENTRIES - 1)) ); do
+
+        # For the ith object, get the first key
+        # 'jq' sorts the keys by default so this should always be
+        # the 'NEIGH_TABLE' key, not the 'OP' key
+        KEY=`jq ".[$i] | keys[0]" $ARP_FILE`
```
tahmed-dev, line 20: can we use [`startwith`](https://stedolan.github.io/jq/manual/#startswith(str)) here?

tahmed-dev (approved):
LGTM

dockers/docker-orchagent/kernel_arp_restore.sh
theasianpianist: Do you mean to confirm that the KEY we select is actually the `NEIGH_TABLE` key?

tahmed-dev: Yes, relying on the key would come first sounds like asking for troubles.

theasianpianist: Fixed

dockers/docker-orchagent/supervisord.conf
```diff
@@ -71,6 +71,16 @@ stderr_logfile=syslog
 dependent_startup=true
 dependent_startup_wait_for=orchagent:running
 
+[program:kernel_arp_restore]
```
lguohan, line 74: what is the different between kernel_arp_restore and restore_neighbors below? it seems the similiar things.

prsunny: `restore_neighbors` is executed only during warmboot and restoration is by programming and sending arp requests. `kernel_arp_restore` is only programming to kernel (as dynamic entries) and done for config_reload scenario. Do you prefer we should combine the script?

lguohan: i recall for fast reboot we also restore the arp entries, which script is that? do we restart kernel arp entries for fast reboot?

tahmed-dev (dismissed):

dockers/docker-orchagent/supervisord.conf
prsunny: AFAIK, for fastboot, we only restore to APP_DB so that orchangent programs those ARP entry to ASIC and the [arp_update ](https://github.com/Azure/sonic-buildimage/blob/master/files/scripts/arp_update) script will send out requests to resolve arp.

theasianpianist: It looks like `restore_neighbors` is used primarily during warmboot. It looks like it does accomplish the same thing, but I am hesitant to change anything warmboot related in case it has other side effects.

dockers/docker-orchagent/swssconfig.sh
```diff
@@ -29,6 +29,25 @@ function fast_reboot {
   esac
 }
 
+function config_reload {
```
lguohan, line 32: can you rename this function as restator_neighbors, and then rename the fast_reboot script as restore default_routes. basically, i think we should try to reuse the code here and not to create duplication.

dockers/docker-orchagent/supervisord.conf
lguohan: why fast_reboot cannot use the this kernel_arp_restore script? why only use this script in config reload scenario?

lguohan: it is fine not to combine with restore_neighbors, but why this arp_restore cannot be combined into swssconfig.sh and also use it for fast reboot?

theasianpianist: The issue with placing this in `swssconfig.sh` is that these kernel entries can only be programmed after the VLAN interface comes up, which happens after `swssconfig.sh` has exited

dockers/docker-orchagent/swssconfig.sh
theasianpianist: I can refactor this to re-use more existing code, but not by much since we don't want a lot of overlapping behavior between fast-reboot and config-reload here. E.g. for fast-reboot, we want to restore the ARP table to APP_DB as well as the default routes, but for config reload we do not want to restore the default routes, and we only need the ARP table to be restored to the kernel (if we restored to APP_DB here, there was some concern about having these entries be `FAILED` if the ToR attempted an ARP refresh).

dockers/docker-orchagent/swssconfig.sh
```diff
@@ -22,11 +26,20 @@ function fast_reboot {
...
         mv -f /default_routes.json /default_routes.json.1
       fi
+  fi
+}
 
-      ;;
-    *)
-      ;;
-  esac
+function signal_kernel_restore {
```
lguohan, line 32: why is this only for config reload? can this be used for fast reboot as well?

dockers/docker-orchagent/supervisord.conf
lguohan: it is ok not be combine. but can this be used for fast-reboot as well?

dockers/docker-orchagent/swssconfig.sh
theasianpianist: Fast reboot already restores the neighbor table to APP_DB, and to my knowledge does not need it restored to the kernel.

theasianpianist:
Closing since we are reverting the dependent PR https://github.com/sonic-net/sonic-utilities/pull/2460

PR #6926, opened 2021-03-01
[Mellanox] Change buffer config for new SKU Mellanox-SN2700-D40C8S8

Junchao-Mellanox:
#### Why I did it

Change buffer config for new SKU Mellanox-SN2700-D40C8S8

#### How I did it

Reuse the buffer config of SKU Mellanox-SN2700-D48C8

#### How to verify it

Run sonic-mgmt qos test and all passed

- [ ] 201811
- [x] 201911
- [ ] 202006
- [x] 202012


qiluo-msft (approved):

device/mellanox/x86_64-mlnx_msn2700-r0/Mellanox-SN2700-D40C8S8/buffers_defaults_t0.j2
```diff
@@ -1,102 +0,0 @@
-{% set default_cable = '5m' %}
-{% set ingress_lossless_pool_size =  '7719936' %}
-{% set ingress_lossless_pool_xoff =  '1032192' %}
-{% set egress_lossless_pool_size =  '13945824' %}
-{% set egress_lossy_pool_size =  '7719936' %}
-
-{%- macro generate_port_lists(PORT_ALL) %}
-    {# Generate list of ports #}
-    {%- for port_idx in range(0, 32) %}
```
neethajohn, line 9: @stephenxs, in an earlier PR you mentioned it was okay if this was incorrect since we don't use it. We make use of this for generating the buffer configs in buffers_config.j2. So shouldn't this be aligned according to the port definition for the sku? We are doing it for other vendors

stephenxs: Hi Neetha 
What topology do you use to generate the configuration for this SKU

neethajohn: We use t0-56. If you need a reference, you can take a look at port list for https://github.com/Azure/sonic-buildimage/tree/master/device/arista/x86_64-arista_7260cx3_64/Arista-7260CX3-D108C8. This uses t0-112 topology

stephenxs: Thanks. Will check this.

stephenxs: Hi @neethajohn 
I checked the template [`buffer_config.j2`](https://github.com/Azure/sonic-buildimage/blob/master/files/build_templates/buffers_config.j2#L98). The `generate_port_lists` is called only if `PORT` is not defined.
```
{%- if PORT is not defined %}
    {%- if defs.generate_port_lists(PORT_ALL) %} {% endif %}
{%- else %}
    {%- for port in PORT %}
        {%- if PORT_ALL.append(port) %}{%- endif %}
    {%- endfor %}
{%- endif %}
```
The `PORT` should be defined when the template is rendered by `sonic-cfggen` on Mellanox platform, so I suppose `generate_port_lists` won't be used.
Is that correct?

neethajohn: Using 'load_minigraph', PORT is defined before we render the buffer templates. So we are good. Thank you for checking

neethajohn (approved):

PR #6932, opened 2021-03-01
[openconfig_acl] Allow setting ICMP type/code to 0

daall:

#### Why I did it
There is a bug in how pyangbind translates yang models into python bindings. The model always sets integer values to 0 by default, so there is no way to check if a user has provided a value that is equal to 0. This is problematic for ICMP and VLAN (among others) because 0 is a valid input value.

We also discovered that if the input is an integer and a string value is provided then the field will be silently ignored, which may be problematic in the case of ACL rules.

#### How I did it
I converted ICMP and VLAN fields to `union` types so that `acl-loader` will treat them as null values unless a user explicitly adds an integer value.

#### How to verify it
Run `sonic-mgmt` tests to verify existing acl behavior is intact. See also new tests in https://github.com/Azure/sonic-utilities/pull/1469.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012

### DIFF
--- a/src/sonic-config-engine/openconfig_acl.py
+++ b/src/sonic-config-engine/openconfig_acl.py
@@ -832,7 +832,7 @@ class yc_config_openconfig_acl__acl_acl_sets_acl_set_acl_entries_acl_entry_l2_co
     self.__destination_mac_mask = YANGDynClass(base=RestrictedClassType(base_type=six.text_type, restriction_dict={u'pattern': u'[0-9a-fA-F]{2}(:[0-9a-fA-F]{2}){5}'}), is_leaf=True, yang_name="destination-mac-mask", parent=self, path_helper=self._path_helper, extmethods=self._extmethods, register_paths=True, namespace='http://openconfig.net/yang/acl', defining_module='openconfig-acl', yang_type='yang:mac-address', is_config=True)
     self.__source_mac = YANGDynClass(base=RestrictedClassType(base_type=six.text_type, restriction_dict={u'pattern': u'[0-9a-fA-F]{2}(:[0-9a-fA-F]{2}){5}'}), is_leaf=True, yang_name="source-mac", parent=self, path_helper=self._path_helper, extmethods=self._extmethods, register_paths=True, namespace='http://openconfig.net/yang/acl', defining_module='openconfig-acl', yang_type='yang:mac-address', is_config=True)
     self.__source_mac_mask = YANGDynClass(base=RestrictedClassType(base_type=six.text_type, restriction_dict={u'pattern': u'[0-9a-fA-F]{2}(:[0-9a-fA-F]{2}){5}'}), is_leaf=True, yang_name="source-mac-mask", parent=self, path_helper=self._path_helper, extmethods=self._extmethods, register_paths=True, namespace='http://openconfig.net/yang/acl', defining_module='openconfig-acl', yang_type='yang:mac-address', is_config=True)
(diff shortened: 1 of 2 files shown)


src/sonic-config-engine/sonic-acl-extension.yang
```diff
@@ -7,24 +7,24 @@ module sonic-acl-extension {
   import openconfig-acl { prefix oc-acl; }
 
   typedef vlan-id-type {
-    type uint16 {
```
qiluo-msft, line 10: >uint16 [](start = 9, length = 6)

@praveen-li, we need a nullable integer, ie, missing this field and 0 are different.
Do you know the better way?
Treat it as `string` seems not fit.

praveen-li: @qiluo-msft
We can use YANG unions, Below allows only null as string input, else integer as input from 0-4095.

```
typedef vlan-id-type {
    type uint16 {
        range 0..4095;
     }
     type string {
         pattern "null";
     }
}
```

praveen-li: @qiluo-msft edited above

src/sonic-config-engine/openconfig_acl.py
```diff
@@ -832,7 +832,7 @@ def __init__(self, *args, **kwargs):
...
-    self.__vlan_id = YANGDynClass(base=RestrictedClassType(base_type=RestrictedClassType(base_type=int, restriction_dict={'range': ['0..65535']},int_size=16), restriction_dict={u'range': [u'0..4095']}), is_leaf=True, yang_name="vlan-id", parent=self, path_helper=self._path_helper, extmethods=self._extmethods, register_paths=True, namespace='https://github.com/Azure/sonic-buildimage', defining_module='sonic-acl-extension', yang_type='vlan-id-type', is_config=True)
+    self.__vlan_id = YANGDynClass(base=RestrictedClassType(base_type=six.text_type, restriction_dict={u'length': [u'1..4']}), is_leaf=True, yang_name="vlan-id", parent=self, path_helper=self._path_helper, extmethods=self._extmethods, register_paths=True, namespace='https://github.com/Azure/sonic-buildimage', defining_module='sonic-acl-extension', yang_type='vlan-id-type', is_config=True)
```
praveen-li, line 835: This  is interesting. I remember while writing the translation in sonic-Yang-mgmt, I converted each Yang model in Json format using Libyang and got similar data for each leaf\list\container in yang model. 

FYI: https://github.com/Azure/sonic-buildimage/blob/0f1d41dac2f79f0e29279e497b12af69c0baac32/src/sonic-yang-mgmt/sonic_yang_ext.py#L208

src/sonic-config-engine/sonic-acl-extension.yang
daall: That seems to work, thank you @praveen-li! I will update the PR.

praveen-li: @daall  Cool, if needed we can add:

default "null";

as well, so That if the user doesn't provide it, it will remain null.

daall: For some reason it doesn't seem like `pyangbind` handles that as expected, but if I put the "type string" clause first then it gives us the expected behavior.

qiluo-msft (approved):

PR #6933, opened 2021-03-01
Fix Python 3 'importlib' bug; Add support for Python 2 back in sonic-py-common

jleveque:
#### Why I did it

Fix a strange bug introduced by https://github.com/Azure/sonic-buildimage/pull/6832 which would only occur in environments with both Python 2 and Python 3 installed (e.g., the PMon container). Error messages such as the following would be seen:

```
ERR pmon#ledd[29]: Failed to load ledutil: module 'importlib' has no attribute 'machinery'
```

This is very odd, and it seems like the Python 2 version of importlib, which is basically just a stub, is taking precedence over the Python 3 version. I found that this occurs when calling `import importlib`. However, calling `import importlib.machinery` and `import importlib.util` causes the proper package to be referenced, and the `machinery` and `util` modules are loaded successfully. This is how it is specified in the official documentation, however there is nothing mentioned regarding that it *should* be done this way or that `import importlib` is unreliable.

Also, since sonic-py-common is still used in environments with Python 2 installed we should maintain support for both Python 2 and 3 until we completely deprecate Python 2, so I have added this back in.

#### How I did it
- Call `import importlib.machinery` and `import importlib.util` explicitly
- Re-introduce Python 2 support in `_load_module_from_file()` method. If Python 2, load source via the `imp` module, if Python 3, use the `importlib` module.

#### How to verify it

Ensure `ledd` daemon starts properly with both Python 2 and Python 3

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012

### DIFF
--- a/src/sonic-ctrmgrd/tests/common_test.py
+++ b/src/sonic-ctrmgrd/tests/common_test.py
@@ -1,5 +1,6 @@
 import copy
-import importlib
+import importlib.machinery
+import importlib.util
 import json
 import os
 import subprocess
--- a/src/sonic-host-services/tests/determine-reboot-cause_test.py
+++ b/src/sonic-host-services/tests/determine-reboot-cause_test.py
@@ -1,4 +1,5 @@
-import importlib
+import importlib.machinery
+import importlib.util
 import sys
 import os
 import pytest
--- a/src/sonic-host-services/tests/procdockerstatsd_test.py
+++ b/src/sonic-host-services/tests/procdockerstatsd_test.py
@@ -1,4 +1,5 @@
-import importlib
+import importlib.machinery
+import importlib.util
 import sys
 import os
 import pytest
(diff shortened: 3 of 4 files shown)


yxieca (dismissed):

lguohan:
as my understanding, as part of the pr https://github.com/Azure/sonic-buildimage/pull/6832, you only tested python3, but not python2, therefore python2 is broken?

lguohan:
why is that the unit test determine-reboot-cause_test.py does not catch the issue?

jleveque:
> as my understanding, as part of the pr #6832, you only tested python3, but not python2, therefore python2 is broken?

No. Python 3 was broken in the PMon container, where Python 2 is still installed, because it breaks the import mechanism when calling `import importlib`. I also decided to add Python 2 support back, just in case it is needed on other platforms until we completely deprecate Python 2.

jleveque:
> why is that the unit test determine-reboot-cause_test.py does not catch the issue?

It appears to be environment-specific. I have not been able to determine the exact conditions necessary to produce the issue.

lguohan (approved):

yxieca (approved):

PR #6956, opened 2021-03-03
[CI] Azure pipeline migration. support nightly build on all platforms & branch master

liushilongbuaa:
#### Why I did it
for sonic image nightly build
#### How I did it
pipeline: https://dev.azure.com/mssonic/build/_build?definitionId=72&_a=summary

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012

### DIFF
--- a/.azure-pipelines/official-build.yml
+++ b/.azure-pipelines/official-build.yml
@@ -101,7 +101,7 @@ stages:
       dbg_image: true
       cache_mode: wcache
 
-- stage: Build_marvell-armhf
+- stage: Build_marvell_armhf
   dependsOn: []
   jobs:
   - template: build-template.yml
--- a/.azure-pipelines/official-build.yml
+++ b/.azure-pipelines/official-build.yml
@@ -81,6 +81,7 @@ stages:
       cache_mode: wcache
 
 - stage: Build_arm
+  dependsOn: []
   displayName: 'Build for arm'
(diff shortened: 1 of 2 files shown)


.azure-pipelines/official-build.yml
```diff
@@ -15,8 +15,18 @@ trigger: none
 pr: none
 
 stages:
-- stage: Build
+- stage: Build_barefoot
```
lguohan, line 18: why are we having so many stages?

liushilongbuaa: Just for beter result showing. Here is an example: https://dev.azure.com/mssonic/build/_build?definitionId=72

liushilongbuaa: For [nephos](https://sonic-jenkins.westus2.cloudapp.azure.com/job/nephos/job/buildimage-nephos-all/), the build for master branch breaks for a month. If we keep them in one stage, the result is always fail.

lguohan: stage is designed to have dependency. if we want to make the build independent, then we can create multiple pipeline and each pipeline handle one platform.

liushilongbuaa: we can remove dependency by:   dependsOn: []

Issue #6911, opened 2021-02-26
checkout code failure.


lguohan:
#### Steps to reproduce the issue:
https://dev.azure.com/mssonic/build/_build/results?buildId=5260&view=logs&j=5562265b-363f-5c64-5cff-7b508c285721&t=27753019-8ff4-5eae-cfff-869853f0e297

#### Output of `show version`:

```
(paste your output here)
```

PR #6959, opened 2021-03-04
Fix the azure pipelines checkout code failure issue

xumia:
#### Why I did it
See https://github.com/Azure/sonic-buildimage/issues/6911
The local git repo may have some issue, when doing the merge/checkout multiple times in local cache repo.
The simple solution is to clean everything in the working folder.

#### How I did it
Remove all the files and directories (including hidden).

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012

### DIFF
--- a/.azure-pipelines/build-template.yml
+++ b/.azure-pipelines/build-template.yml
@@ -110,7 +110,7 @@ jobs:
         make USERNAME=admin $CACHE_OPTIONS SONIC_BUILD_JOBS=$(nproc) target/sonic-${{ parameters.platform }}.bin
       fi
     displayName: 'Build sonic image'
-  - template: cleanup.yml
   - publish: $(System.DefaultWorkingDirectory)/
     artifact: sonic-buildimage.${{ parameters.platform }}
     displayName: "Archive sonic image"
+  - template: cleanup.yml
--- a/.azure-pipelines/cleanup.yml
+++ b/.azure-pipelines/cleanup.yml
@@ -10,8 +10,5 @@ steps:
     if sudo [ -f dockerfs/var/run/docker.pid ] ; then
         pid=`sudo cat dockerfs/var/run/docker.pid` ; sudo kill $pid
     fi
-    sudo rm -rf dockerfs
-    sudo rm -rf fsroot
-    username=$(id -un)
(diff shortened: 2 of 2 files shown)


.azure-pipelines/build-template.yml
```diff
@@ -110,7 +110,7 @@ jobs:
         make USERNAME=admin $CACHE_OPTIONS SONIC_BUILD_JOBS=$(nproc) target/sonic-${{ parameters.platform }}.bin
       fi
     displayName: 'Build sonic image'
-  - template: cleanup.yml
   - publish: $(System.DefaultWorkingDirectory)/
     artifact: sonic-buildimage.${{ parameters.platform }}
     displayName: "Archive sonic image"
+  - template: cleanup.yml
```
lguohan, line 116: i do not understand why we move this.

.azure-pipelines/cleanup.yml
```diff
@@ -10,8 +10,5 @@ steps:
     if sudo [ -f dockerfs/var/run/docker.pid ] ; then
         pid=`sudo cat dockerfs/var/run/docker.pid` ; sudo kill $pid
     fi
-    sudo rm -rf dockerfs
-    sudo rm -rf fsroot
-    username=$(id -un)
-    sudo chown -R ${username}.${username} .
+    sudo rm -rf $(ls -A1)
```
lguohan, line 13: can you provide more description on what the issue was?

lguohan:
@xumia, can you address?

.azure-pipelines/build-template.yml
xumia: @lguohan , because the cleanup script is changed to remove the entire working folder including the target folder, if not move the last step, no targets will be uploaded.

.azure-pipelines/cleanup.yml
xumia: @lguohan , for one reason, in PR build, we can not stop user to create files or folders with root permission, it will have issue when trying to checkout. For another reason, the local git repo may have some issue, when doing the merge/checkout multiple time.
The simple solution is to remove the entire folder including the hidden folders/files in the folder.

xumia:
The issue has already fixed.

PR #6985, opened 2021-03-08
Copy dummy flannel.conf to get around absence of CNI Network

renukamanavalan:
#### Why I did it
We skip install of CNI plugin, as we don't need. But this leaves node in "not ready" state, upon joining master. 
To fix, we copy this dummy .conf file in /etc/cni/net.d

#### How I did it
Keep this file in /usr/share/sonic/templates and copy to /etc/cni/net.d upon joining k8s master.

#### How to verify it
Upon configuring master-IP and enable join, watch node join and move to ready state.
You may verify using `kubectl get nodes` command

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012

### DIFF
--- a/src/sonic-ctrmgrd/tests/kube_commands_test.py
+++ b/src/sonic-ctrmgrd/tests/kube_commands_test.py
@@ -113,12 +113,11 @@ None".format(KUBE_ADMIN_CONF),
             "rm -rf {}".format(CNI_DIR),
             "systemctl stop kubelet",
             "modprobe br_netfilter",
-            "systemctl start kubelet",
-            "kubeadm join --discovery-file {} --node-name None".format(
-                KUBE_ADMIN_CONF),
             "mkdir -p {}".format(CNI_DIR),
             "cp {} {}".format(FLANNEL_CONF_FILE, CNI_DIR),
-            "systemctl restart kubelet"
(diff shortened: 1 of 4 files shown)


qiluo-msft:
@isabelmsft Please help review.

src/sonic-ctrmgrd/tests/kube_commands_test.py
```diff
@@ -13,6 +13,8 @@
 
 
 KUBE_ADMIN_CONF = "/tmp/kube_admin.conf"
+FLANNEL_CONF_FILE = "/tmp/flannel.conf"
+CNI_DIR = "/tmp/cni/net.d"
```
isabelmsft, line 17: Why not /etc/cni/net.d?

isabelmsft, line 17: Why not /etc/cni/net.d?

renukamanavalan: This is test code, where there is NO kubernetes installed. All kubernetes calls are mocked and dummy. The test code does not have privilege to operate on /etc, so we set it as /tmp

isabelmsft (approved):

PR #6986, opened 2021-03-08
[CI] Pipeline migration master

liushilongbuaa:

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012


liushilongbuaa:
The former PR is invalid, because of re-forking repo of sonic-buildimage.

liushilongbuaa:
Changed from multi-stages pipeline to single stage pipelines.

lguohan (dismissed):

.azure-pipelines/official-images/official-build-barefoot.yml
```diff
@@ -0,0 +1,24 @@
+# Starter pipeline
```
xumia, line 1: You can use the same yaml file for all build definitions, different definition uses different parameters.

liushilongbuaa: Maybe I can drop another PR for compare. 
This kind of change will remove template parameters limit, is hard to set default values, and is like classic Jenkins job. parameters are in Web config, not in code. I prefer not use the same yaml file.

PR #6990, opened 2021-03-08
dhcp for dual tor: include all vlan intf into downstream intf

trzhang-msft:
Include all vlan interfaces into downstream interface set to allow each dhcp instance to scan over them. Note that with this change, downstream interface set includes not only vlan interfaces but also portchannels. However, this doesn't matter as dhcp uses giaddr or yiaddr to choose out interface.

Tested and verified on a virtual testbed.

### DIFF
--- a/src/isc-dhcp/patch/0009-Support-for-dual-tor-scenario.patch
+++ b/src/isc-dhcp/patch/0009-Support-for-dual-tor-scenario.patch
@@ -238,7 +238,7 @@ index e158efe..055d97f 100644
  		  (flags & INTERFACE_UPSTREAM ? 'Y' : 'N'),
  		  (flags & INTERFACE_DOWNSTREAM ? 'Y' : 'N'));
  
-+	if (flags & INTERFACE_DOWNSTREAM) {
++	if (flags & INTERFACE_DOWNSTREAM || flags & INTERFACE_UPSTREAM) {
 +		ci = ((struct downstream_intf_list *)dmalloc(sizeof *ci, MDL));
 +		if (!ci)
 +			log_fatal("no memory for downstream interface pointer.\n");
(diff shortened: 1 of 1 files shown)


src/isc-dhcp/patch/0009-Support-for-dual-tor-scenario.patch
```diff
@@ -238,7 +238,7 @@ index e158efe..055d97f 100644
  		  (flags & INTERFACE_UPSTREAM ? 'Y' : 'N'),
  		  (flags & INTERFACE_DOWNSTREAM ? 'Y' : 'N'));
  
-+	if (flags & INTERFACE_DOWNSTREAM) {
++	if (flags & INTERFACE_DOWNSTREAM || flags & INTERFACE_UPSTREAM) {
```
tahmed-dev, line 241: We can use the existing `interfaces` which hold list of all interfaces as being done in [isc dhcrelay:849](https://salsa.debian.org/dhcp-team/isc-dhcp/-/blob/master/relay/dhcrelay.c#L849)

trzhang-msft: a) the code above is better: we don't have to include all interfaces, -id will include vlan for this proc, and -iu will include other vlans on the switch; b) the code above is safer: no one knows if some interface will have any setting errors by the code, for example, in patch 0010, i already find out that netmask is not correctly setup. So in short, we should keep the list as short as possible

tahmed-dev: I agree, however on the other hand, this is a legacy code and reducing the amount of code changes helps limit regression or injecting new bugs given that the code was running for quite some time and functioning. I am indifferent though as it is not a complex change.

tahmed-dev (approved):

src/isc-dhcp/patch/0009-Support-for-dual-tor-scenario.patch
jleveque: Without comment, it is a bit confusing why we are including upstream interfaces in the downstream interface list. Maybe consider adding a comment.

trzhang-msft: sure, but I will add a comment on the same line so that we don't have to re-generate the patch

jleveque (approved):

PR #6993, opened 2021-03-09
[build] add sonic-mgmt docker to PR test

yxieca:
#### Why I did it
Found regression on sonic-mgmt docker build.

#### How I did it
Add PR test to catch regression.

#### How to verify it
PR test.

### DIFF
--- a/.azure-pipelines/build-template.yml
+++ b/.azure-pipelines/build-template.yml
@@ -7,6 +7,7 @@ parameters:
   - marvell-armhf
   - mellanox
   - vs
+  - generic
 
 - name: platform_arch
   type: string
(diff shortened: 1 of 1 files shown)


yxieca:
docker-ptf is already built in PR test currently. And the change didn't get sonic-mgmt-docker to build. Investigation is needed.

yxieca:
VS build failure is expected. After #7009 merged, retest will pass.

.azure-pipelines/build-template.yml
```diff
@@ -92,6 +92,7 @@ jobs:
         fi
 
         make USERNAME=admin $CACHE_OPTIONS SONIC_BUILD_JOBS=$(nproc) target/docker-sonic-vs.gz target/sonic-vs.img.gz target/docker-ptf.gz
+        make USERNAME=admin $CACHE_OPTIONS SONIC_BUILD_JOBS=$(nproc) target/docker-sonic-mgmt.gz
```
lguohan, line 95: this increase the vs job time, which is the critical path for running test. suggest to have a new job then.

PR #6999, opened 2021-03-09
Fix get-pip 2.7 url according to upstream announcement

qiluo-msft:
#### Why I did it

ref: https://bootstrap.pypa.io/2.7/get-pip.py

The URL you are using to fetch this script has changed, and this one will no
longer work. Please use get-pip.py from the following URL instead:

    https://bootstrap.pypa.io/pip/2.7/get-pip.py

- [x] 201811
- [x] 201911
- [x] 202006
- [x] 202012


yxieca:
No need to cherry-pick into feature branches. we should use ptf docker built from master branch.

PR #7026, opened 2021-03-11
Update hostcfgd to disable unused services for fabric asics

ngoc-do:
Some services are unnecessary for fabric asics like bgp, lldp, teamd.
We so disable them. But it could be that there are some mixed asics
NPU and fabric in the same system, so we should just disable only
for fabric asics.

This commit is to disable those services only fabric asics. It works together
with https://github.com/Azure/sonic-utilities/pull/1493.

Signed-off-by: ngocdo <ngocdo@arista.com>

#### Why I did it

Fabric asics do not need some services like bgp, lldp, teamd.
They just need database, swss, and syncd running.

#### How I did it

I disable them with `FEATURE` flag configured in `CONFIG_DB`, which is listened and reacted by hostcfgd. 

This PR will update `CONFIG_DB's FEATURE` for fabric asics. The update is executed in load_minigraph, and saved to config_db.json. So the update will run only once.

If a system has both NPU and fabric, only services running on fabric are disabled. NPU's services are not affected.

One field, `always_disabled_on_asics` is added to `FEATURE`'s services that we want to disable for fabric asics. For example, `bgp@4.service` and `bgp@5.service` will be disabled.
```
    "FEATURE": {
        "bgp": {
            "state": "enabled",
            "has_timer": "False",
            "has_global_scope": "False",
            "has_per_asic_scope": "True",
            "auto_restart": "enabled",
            "high_mem_alert": "disabled",
            "always_disabled_on_asics": "4,5"
        },
   }
```

The above setup is at https://github.com/Azure/sonic-utilities/pull/1493.

hostcfg will act on the setup, and disable the services.

#### How to verify it

Tested on systems with VOQ and fabric asics.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012

### DIFF
--- a/src/sonic-host-services/scripts/hostcfgd
+++ b/src/sonic-host-services/scripts/hostcfgd
@@ -437,10 +437,15 @@ class HostConfigDaemon:
         has_global_scope = ast.literal_eval(feature_table[feature_name].get('has_global_scope', 'True'))
         has_per_asic_scope = ast.literal_eval(feature_table[feature_name].get('has_per_asic_scope', 'False'))
 
+        # This feature may be disabled on some asics (but may not be on all)
+        always_disabled_on_asics = ast.literal_eval(feature_table[feature_name].get('always_disabled_on_asics', '()'))
+        disabled_feature_name_suffix_list = ([(feature_name + '@' + str(asic_inst)) for asic_inst in range(device_info.get_num_npus())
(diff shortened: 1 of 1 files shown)


kenneth-arista:
@arlakshm we can discard this PR as this was needed for a hardware platform that we are no longer supporting.

PR #7048, opened 2021-03-13
[201811][sonic-config-engine] Parse AutoNegotiation element from LinkMetadata section of minigraph file

jleveque:
Backport of https://github.com/Azure/sonic-buildimage/pull/7031 to the 201811 branch

#### Why I did it

To enable parsing the `AutoNegotiation` element from the LinkMetadata section of minigraph file

#### How I did it

Parse the value `AutoNegotiation` element from the `LinkMetadata` section of minigraph file. If the element is present, an `autoneg` key will be added to the port in the `PORT` table of Config DB with a value of either `0` or `1`

If an `autoneg` value is present in port_config.ini, the value from the minigraph will take precedence, overriding that value.

Also remove `AutoNegotiation` and `EnableAutoNegotiation` elements from the `DeviceInfo` section, as we will use this data in the `LinkMetadata` section to determine whether to enable auto-negotiation for a port.

#### How to verify it

Added unit tests to validate the behavior described above


lguohan:
can you check the build failure?

jleveque:
All checks have passed except LGTM, which is having trouble locating a commit :/

lguohan (approved):

PR #7062, opened 2021-03-15
[Telemetry docker] add memory and memory swap limits

pra-moh:

- [ ] 201811
- [- ] 201911
- [- ] 202006
- [- ] 202012

### DIFF
--- a/rules/docker-telemetry.mk
+++ b/rules/docker-telemetry.mk
@@ -25,6 +25,7 @@ endif
 
 $(DOCKER_TELEMETRY)_CONTAINER_NAME = telemetry
 $(DOCKER_TELEMETRY)_RUN_OPT += --privileged -t
+$(DOCKER_TELEMETRY)_RUN_OPT += --memory 300m --memory-swap 300m
 $(DOCKER_TELEMETRY)_RUN_OPT += -v /etc/sonic:/etc/sonic:ro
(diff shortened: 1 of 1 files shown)


rlhui:
@pra-moh - thanks, please add testing done.

lguohan:
what is memory-swap?

pra-moh:
> what is memory-swap?



> what is memory-swap?
Adding this argument to avoid container using any extra memory. Using just --memory argument can make docker start memory swap from disk based on some threads. This will make sure it will never use more than 300MB 

Based on docker documentation:
PREVENT A CONTAINER FROM USING SWAP
If --memory and --memory-swap are set to the same value, this prevents containers from using any swap. This is because --memory-swap is the amount of combined memory and swap that can be used, while --memory is only the amount of physical memory that can be used.

https://docs.docker.com/config/containers/resource_constraints/#prevent-a-container-from-using-swap

lguohan:
please describe what kind of testing have you done?

besides, we do not use swap at all on switches. it should be set to zero.

lguohan:
can you update the description?

lguohan (approved):

lguohan:
can you update the description to say how 500m is derived?

rlhui:
Let's revise the memory cap based on latest data.  Thanks.

lguohan:
bad pr, oom triggered system reboot.

hui-ma:
let us revert it.

zhenggen-xu:
We need "--oom-kill-disable" option when we limit the memory so no kernel panic.

PR #7083, opened 2021-03-17
[dockers][supervisor] Increase event buffer size for process exit listener; Set all event buffer sizes to 1024

jleveque:
#### Why I did it

To prevent error [messages](https://dev.azure.com/mssonic/build/_build/results?buildId=2254&view=logs&j=9a13fbcd-e92d-583c-2f89-d81f90cac1fd&t=739db6ba-1b35-5485-5697-de102068d650&l=802) like the following from being logged:

```
Mar 17 02:33:48.523153 vlab-01 INFO swss#supervisord 2021-03-17 02:33:48,518 ERRO pool supervisor-proc-exit-listener event buffer overflowed, discarding event 46
```

This is basically an addendum to https://github.com/Azure/sonic-buildimage/pull/5247, which increased the event buffer size for dependent-startup. While supervisor-proc-exit-listener doesn't subscribe to as many events as dependent-startup, there is still a chance some containers (like swss, as in the example above) have enough processes running to cause an overflow of the default buffer size of 10.

This is especially important for preventing erroneous log_analyzer failures in the sonic-mgmt repo regression tests, which have started occasionally causing PR check builds to fail. Example [here](https://dev.azure.com/mssonic/build/_build/results?buildId=2254&view=logs&j=9a13fbcd-e92d-583c-2f89-d81f90cac1fd&t=739db6ba-1b35-5485-5697-de102068d650&l=802).

I set all supervisor-proc-exit-listener event buffer sizes to 1024, and also updated all dependent-startup event buffer sizes to 1024, as well, to keep things simple, unified, and allow headroom so that we will not need to adjust these values frequently, if at all.

- [ ] 201811
- [x] 201911
- [x] 202006
- [x] 202012

### DIFF
--- a/dockers/docker-database/supervisord.conf.j2
+++ b/dockers/docker-database/supervisord.conf.j2
@@ -8,6 +8,7 @@ command=/usr/bin/supervisor-proc-exit-listener --container-name database
 events=PROCESS_STATE_EXITED,PROCESS_STATE_RUNNING
 autostart=true
 autorestart=unexpected
+buffer_size=1024
 
 [program:rsyslogd]
 command=/usr/sbin/rsyslogd -n -iNONE
--- a/dockers/docker-dhcp-relay/docker-dhcp-relay.supervisord.conf.j2
+++ b/dockers/docker-dhcp-relay/docker-dhcp-relay.supervisord.conf.j2
@@ -17,6 +17,7 @@ command=/usr/bin/supervisor-proc-exit-listener --container-name dhcp_relay
 events=PROCESS_STATE_EXITED,PROCESS_STATE_RUNNING
 autostart=true
 autorestart=unexpected
+buffer_size=50
 
 [program:rsyslogd]
 command=/usr/sbin/rsyslogd -n -iNONE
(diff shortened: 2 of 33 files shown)


lguohan:
not sure why different docker has different buffer size.

jleveque:
> not sure why different docker has different buffer size.

It varies based on the number of processes running in the container. More processes = more events. We could choose a number high enough to work for all containers and apply it everywhere, but this approach is more conservative with memory usage in containers which don't have many processes running.

qiluo-msft:
How large is the one buffer slot in memory? If the buffer overflow is critical error, let's use very large buffer size.

jleveque:
> How large is the one buffer slot in memory? If the buffer overflow is critical error, let's use very large buffer size.

There are different types of events, so they vary, but they appear to be relatively small: https://github.com/Supervisor/supervisor/blob/master/supervisor/events.py. I considered just using a very large size for all. Thoughts?

lguohan:
suggest 1024 for all.

jleveque:
> suggest 1024 for all.

Done in latest commit.

lguohan (approved):

qiluo-msft (approved):

qiluo-msft (approved):

jleveque:
This will not cherry-pick cleanly into 201911, so I opened a separate PR [here](https://github.com/Azure/sonic-buildimage/pull/7106).

daall:
@jleveque can you also open a separate PR for 202012 please? There is a conflict.

jleveque:
> @jleveque can you also open a separate PR for 202012 please? There is a conflict.

202012 PR opened here: https://github.com/Azure/sonic-buildimage/pull/7203

PR #7098, opened 2021-03-18
[mlnx] set temperature threshold values

d-dashkov:
**Why I did it**
Implemented platform API functions for setting / changing temperature threshold.
Associated with PRs: Azure/sonic-platform-common#180, Azure/sonic-platform-daemons#166

**How I did it**
Added implementations for set_high_threshold(self, temperature) and set_high_critical_threshold(self, temperature) from the base planform API.

### DIFF
--- a/platform/mellanox/mlnx-platform-api/sonic_platform/thermal.py
+++ b/platform/mellanox/mlnx-platform-api/sonic_platform/thermal.py
@@ -480,7 +480,6 @@ class Thermal(ThermalBase):
             return None
         return value_float / 1000.0
 
-
     @classmethod
     def _write_generic_file(cls, filename, content):
         """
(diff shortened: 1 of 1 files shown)


keboliu (changes requested):
I don't think these two new API implementations can work, the threshold is read-only, you can not override it.

d-dashkov:
@keboliu
CPU, SFP are RO and as you mentioned I cannot override them. But ASIC, PSU files have RW access mode, so I can set new values and they will work fine. But 
```
lrwxrwxrwx 1 root root /run/hw-management/thermal/mlxsw/temp_trip_crit -> /sys/devices/virtual/thermal/thermal_zone3/trip_point_3_temp
-rw-r--r-- 1 root root /sys/devices/virtual/thermal/thermal_zone3/trip_point_3_temp
```

keboliu:
> @keboliu
> CPU, SFP are RO and as you mentioned I cannot override them. But ASIC, PSU files have RW access mode, so I can set new values and they will work fine.
> 
> ```
> lrwxrwxrwx 1 root root /run/hw-management/thermal/mlxsw/temp_trip_crit -> /sys/devices/virtual/thermal/thermal_zone3/trip_point_3_temp
> -rw-r--r-- 1 root root /sys/devices/virtual/thermal/thermal_zone3/trip_point_3_temp
> ```

what's the motivation to rewrite the  ASIC and PSU-related threshold?  How can you make sure the new threshold is safe for the harware?

PR #7114, opened 2021-03-22
[Mellanox] Fix issue: should not initialize led color in __init__ funcation as platform API will be called by multiple daemons

Junchao-Mellanox:
#### Why I did it

The existing Fan led and Psu led object initialize itself to green color in __init__ method. However, there are multiple daemons calls sonic platform API and there could be a case that:

1. A PSU is removed from system
2. Reboot switch
3. psud detects that 1 PSU is missing and set PSU led to red
4. Other daemon just start up and call sonic platform API, the API set PSU led to green by call PsuLed.__init__

This PR is a partial fix for the issue. As we also need guarantee that the led is initialized with a correct value. I checked existing psud and thermalctld code. psud always initialize the PSU led color on boot up, thermalcltd need some changes to initialize led color on the first run

#### How I did it

Remove the led color initialization code from FanLed.__init__ and PsuLed.__init__

#### How to verify it

Manual test

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012

### DIFF
--- a/platform/mellanox/mlnx-platform-api/sonic_platform/led.py
+++ b/platform/mellanox/mlnx-platform-api/sonic_platform/led.py
@@ -207,8 +207,6 @@ class FanLed(Led):
             self._orange_led_path = os.path.join(Led.LED_PATH, "led_fan_orange")
             self._led_cap_path = os.path.join(Led.LED_PATH, "led_fan_capability")
 
-        self.set_status(Led.STATUS_LED_COLOR_GREEN)
-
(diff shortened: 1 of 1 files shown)


liat-grozovik:
@Junchao-Mellanox  with this partial fix what is the expected behavior and what is still missing? can you think of flows where this will be still not working?

jleveque (approved):

liat-grozovik (approved):

PR #7146, opened 2021-03-24
Update versions_manager.py to make versions map key unique

baxia-lan:
py2/py3 packages names are case insensitive, and the versions map
key should be the same for packages whose name can have different cases.

For example, in files/build/versions/default/versions-py3, package
"click==7.1.2" is pinned; and in
files/build/versions/dockers/docker-sonic-vs/versions-py3, package
"Click==7.0" is pinned.
Without this fix, the aggregated versions-py3 file used for building
docker-sonic-vs looks like below:
...
click==7.1.2
Click==7.0
...
However, we actually want "click==7.0" to overwrite "click==7.1.2" for
docker-sonic-vs build.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012

### DIFF
--- a/scripts/versions_manager.py
+++ b/scripts/versions_manager.py
@@ -22,7 +22,7 @@ class Component:
     ctype -- Component Type, such as deb, py2, etc
     dist  -- Distribution, such as stretch, buster, etc
     arch  -- Architectrue, such as amd64, arm64, etc
-    
+
     '''
     def __init__(self, versions, ctype, dist=ALL_DIST, arch=ALL_ARCH):
         self.versions = versions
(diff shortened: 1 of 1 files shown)


qiluo-msft:
I understand py2/py3 package names are case insensitive. Could you provide details on "deb packages names are case insensitive" ？

baxia-lan:
> I understand py2/py3 package names are case insensitive. Could you provide details on "deb packages names are case insensitive" ？

Luckily the deb package names are all lower cases. Updated the fix to check if filename contains 'py2' or 'py3'.

qiluo-msft (approved):

qiluo-msft:
@xumia Could you help review?

xumia (approved):

PR #7151, opened 2021-03-25
[procdockerstatsd] add missing unit conversion

pra-moh:

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012

### DIFF
--- a/src/sonic-host-services/scripts/procdockerstatsd
+++ b/src/sonic-host-services/scripts/procdockerstatsd
@@ -69,6 +69,7 @@ class ProcDockerStats(daemon_base.DaemonBase):
         UNITS_B = 'B'
         UNITS_KB = 'KB'
         UNITS_MB = 'MB'
+        UNITS_GB = 'GB'
         UNITS_MiB = 'MiB'
         UNITS_GiB = 'GiB'
 
@@ -79,6 +80,8 @@ class ProcDockerStats(daemon_base.DaemonBase):
             value *= 1000
         elif units.lower() == UNITS_MB.lower():
             value *= (1000 * 1000)
+        elif units.lower() == UNITS_GB.lower():
+            value *= (1000 * 1000 * 1000)
         elif units.lower() == UNITS_MiB.lower():
             value *= (1024 * 1024)
         elif units.lower() == UNITS_GiB.lower():


src/sonic-host-services/scripts/procdockerstatsd
```diff
@@ -79,6 +80,8 @@ class ProcDockerStats(daemon_base.DaemonBase):
             value *= 1000
         elif units.lower() == UNITS_MB.lower():
             value *= (1000 * 1000)
+        elif units.lower() == UNITS_GB.lower():
+            value *= (1000 * 1000 * 1000)
```
jleveque, line 84: For brevity/readability, I suggest using an exponent here

```python
value *= (1000 ** 3)
```

Would you mind also making similar changes on lines 82, 86 and 88?

jleveque (changes requested):

src/sonic-host-services/scripts/procdockerstatsd
```diff
@@ -69,6 +69,7 @@ class ProcDockerStats(daemon_base.DaemonBase):
         UNITS_B = 'B'
         UNITS_KB = 'KB'
         UNITS_MB = 'MB'
+        UNITS_GB = 'GB'
         UNITS_MiB = 'MiB'
         UNITS_GiB = 'GiB'
```
jleveque, line 74: For future proofing, I suggest supporting `TB` and `TiB` units.

jleveque (changes requested):

jleveque (approved):

hui-ma (dismissed):

lguohan (changes requested):
please add unit tests

jleveque:
@pra-moh: Please enhance the existing test case to test the new units: https://github.com/Azure/sonic-buildimage/blob/master/src/sonic-host-services/tests/procdockerstatsd_test.py#L27

src/sonic-host-services/tests/procdockerstatsd_test.py
```diff
@@ -36,9 +36,13 @@ def test_convert_to_bytes(self):
             ('500MiB', 524288000),
             ('66.41MiB', 69635932),
             ('333.6MiB', 349804954),
+            ('7GB', 7000000000),
+            ('800GB', 800000000000),
```
jleveque, line 40: Suggest to also add at least one GB test case with a decimal point, following my existing pattern.

src/sonic-host-services/tests/procdockerstatsd_test.py
```diff
@@ -36,9 +36,13 @@ def test_convert_to_bytes(self):
...
             ('1GiB', 1073741824),
             ('500GiB', 536870912000),
-            ('7.751GiB', 8322572878)
+            ('7.751GiB', 8322572878),
+            ('8.1TiB', 8906000000000000),
+            ('12TB', 12000000000000)
```
jleveque, line 45: Suggest to also add at least one TB test case with a decimal point, following my existing pattern.

jleveque (changes requested):

jleveque (approved):

src/sonic-host-services/tests/procdockerstatsd_test.py
```diff
@@ -36,9 +36,14 @@ def test_convert_to_bytes(self):
...
             ('333.6MiB', 349804954),
+            ('7.75GB', 7750000000),
+            ('800GB', 800000000000),
             ('1GiB', 1073741824),
             ('500GiB', 536870912000),
-            ('7.751GiB', 8322572878)
+            ('7.751GiB', 8322572878),
+            ('8.1TiB', 8906044184986),
+            ('12TB', 12000000000000),
+            ('2.25TB', 2250000000000)
```
lguohan, line 46: the test is too simple, we need to have a test that convert from the ps -e output and test the content in the database as end-to-end test.

jleveque: I think this test is appropriate to the `convert_to_bytes()` method. More test cases need to be added in order to ensure full end-to-end functionality.

PR #7195, opened 2021-03-31
libnl3 updates for MPLS

qbdwlr:
What I did
SONiC buildimage support for MPLS:
1) New accessors in libnl3 for MPLS attributes
2) bug fixes in libnl3 for MPLS attribute parsing

Why I did it
SONiC libnl3 support for MPLS

How I verified it
Unit-tests in sonic-swss/tests/test_mpls.py and sonic-utilities/tests
System tests in sonic-mgmt

Details if related
Refer to HLD: Azure/SONiC#706

### DIFF
--- a/src/libnl3/patch/0002-mpls-remove-nl_addr_valid.patch
+++ b/src/libnl3/patch/0002-mpls-remove-nl_addr_valid.patch
@@ -1,0 +1,63 @@
+From c89d1a129f71d3d2f76e6cbadb11ef41d8941a73 Mon Sep 17 00:00:00 2001
+From: Ann Pokora <apokora@juniper.net>
+Date: Tue, 25 May 2021 18:10:04 -0700
+Subject: [PATCH 2/2] mpls remove nl_addr_valid
+
+The removed calls to nl_addr_valid() are passing in a pointer to a binary address
+and the length of the binary address, which does not match expected arguments for
+nl_addr_valid().
+nl_addr_valid() expects a pointer to an ASCII string and the address family of
+the string format.
+The incorrect arguments cause unexpected failures and the expected arguments are
+not available in the context.
+
+Signed-off-by: Ann Pokora <apokora@juniper.net>
+---
+ lib/route/nexthop.c       | 8 --------
+ lib/route/nh_encap_mpls.c | 4 ----
+ 2 files changed, 12 deletions(-)
+
+diff --git a/lib/route/nexthop.c b/lib/route/nexthop.c
+index 7a9904c..ac0095e 100644
+--- a/lib/route/nexthop.c
++++ b/lib/route/nexthop.c
+@@ -351,10 +351,6 @@ int rtnl_route_nh_set_newdst(struct rtnl_nexthop *nh, struct nl_addr *addr)
+ {
+ 	struct nl_addr *old = nh->rtnh_newdst;
+ 
+-	if (!nl_addr_valid(nl_addr_get_binary_addr(addr),
+-			   nl_addr_get_len(addr)))
+-		return -NLE_INVAL;
+-
+ 	if (addr) {
+ 		nh->rtnh_newdst = nl_addr_get(addr);
+ 		nh->ce_mask |= NH_ATTR_NEWDST;
+@@ -378,10 +374,6 @@ int rtnl_route_nh_set_via(struct rtnl_nexthop *nh, struct nl_addr *addr)
+ {
+ 	struct nl_addr *old = nh->rtnh_via;
+ 
+-	if (!nl_addr_valid(nl_addr_get_binary_addr(addr),
+-			   nl_addr_get_len(addr)))
+-		return -NLE_INVAL;
+-
+ 	if (addr) {
+ 		nh->rtnh_via = nl_addr_get(addr);
+ 		nh->ce_mask |= NH_ATTR_VIA;
+diff --git a/lib/route/nh_encap_mpls.c b/lib/route/nh_encap_mpls.c
+index 18336ac..6c5a3c7 100644
+--- a/lib/route/nh_encap_mpls.c
++++ b/lib/route/nh_encap_mpls.c
+@@ -109,10 +109,6 @@ int rtnl_route_nh_encap_mpls(struct rtnl_nexthop *nh,
+ 	if (!addr)
+ 		return -NLE_INVAL;
+ 
+-	if (!nl_addr_valid(nl_addr_get_binary_addr(addr),
+-			   nl_addr_get_len(addr)))
+-		return -NLE_INVAL;
+-
+ 	rtnh_encap = calloc(1, sizeof(*rtnh_encap));
+ 	if (!rtnh_encap)
+ 		return -NLE_NOMEM;
+-- 
+2.7.4
+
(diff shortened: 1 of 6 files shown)


src/libnl3/patch/0001-parse-multipath-infinite-loop-fix.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/lib/route/route_obj.c b/lib/route/route_obj.c
```
lguohan, line 1: are they upstreamed? can you format the patch properly, author, description, bla, bla

qbdwlr: The patch files have been reformatted as requested.  A PR has been created to upstream libnl3 changes, but there have been no PRs merged in that project in 2021, so these patches remain necessary.

src/libnl3/patch/0002-mpls-remove-nl_addr_valid.patch
```diff
@@ -0,0 +1,55 @@
+From c89d1a129f71d3d2f76e6cbadb11ef41d8941a73 Mon Sep 17 00:00:00 2001
+From: Ann Pokora <apokora@juniper.net>
+Date: Tue, 25 May 2021 18:10:04 -0700
+Subject: [PATCH 2/2] mpls remove nl_addr_valid
```
lguohan, line 4: why are we remove the validation?

qbdwlr: @lguohan nl_addr_valid() arguments are supposed to be a pointer to an ASCII string and an address family value (AF_MPLS, AF_INET, AF_INET6)... it then uses these to validate that the string format contains an address for the specified family.
The removed calls to nl_addr_valid() were instead passing in a pointer to a binary address and the length of the binary address... this is completely wrong.  For most cases, the length of the binary address would not match any of the values of supported address families, so nl_addr_valid() would return true.  But if the binary address was an MPLS label stack with 3 labels, then the length would match the value of AF_DECnet (12) and it attempts to parse the binary address as if it were an ASCII string... this fails and nl_addr_valid() returns false.

qbdwlr:
@lguohan @prsunny Please review/approve this sonic-buildimage PR as the associated sonic-swss PR has build dependencies on the libnl3 changes.

src/libnl3/patch/0002-mpls-remove-nl_addr_valid.patch
```diff
@@ -0,0 +1,55 @@
...
+
+diff --git a/lib/route/nexthop.c b/lib/route/nexthop.c
+index 7a9904c..ac0095e 100644
+--- a/lib/route/nexthop.c
++++ b/lib/route/nexthop.c
+@@ -351,10 +351,6 @@ int rtnl_route_nh_set_newdst(struct rtnl_nexthop *nh, struct nl_addr *addr)
+ {
+ 	struct nl_addr *old = nh->rtnh_newdst;
+ 
+-	if (!nl_addr_valid(nl_addr_get_binary_addr(addr),
```
smaheshm, line 20: Instead of removing the validation, can we fix the arguments.

```
replace nl_addr_get_len(addr) with nl_addr_get_family(addr)
```

qbdwlr: @smahesh That still isn't valid.  Look at the code for nl_addr_valid() in libnl3.  It is looking for an ASCII string and validating the string format.  addr does not container a string,  The return from nl_addr_get_binary_addr() is not a string. 
 We would have to first convert addr to string.  It's pointless.  nl_addr_valid() is not used anywhere else in libnl3, so I do not know why someone decided to use it in these MPLS functions.

src/sonic-yang-models/yang-models/sonic-interface.yang
```diff
@@ -57,6 +57,12 @@ module sonic-interface {
 						path "/vrf:sonic-vrf/vrf:VRF/vrf:VRF_LIST/vrf:name";
 					}
 				}
+
+				leaf mpls {
+					type string {
+						pattern "enable|disable";
+					}
```
lguohan, line 64: need description here, what does enable/disable mean here.

src/libnl3/patch/0002-mpls-remove-nl_addr_valid.patch
lguohan: nl_addr_valid() is using the route_nh_set which is not mpls specific. this is the IP functions. how can this function pass now?

src/libnl3/patch/0002-mpls-remove-nl_addr_valid.patch
lguohan: can you put the above explaination into the this patch file?

src/libnl3/patch/0002-mpls-remove-nl_addr_valid.patch
qbdwlr: @lguohan RTA_NEWDST is specific to MPLS.  NEWDST contains the MPLS labelstack associated with the NH.

src/libnl3/patch/0002-mpls-remove-nl_addr_valid.patch
qbdwlr: @lguohan explanation is added to patch file.

src/sonic-yang-models/yang-models/sonic-interface.yang
qbdwlr: @lguohan description is added.

src/libnl3/patch/0002-mpls-remove-nl_addr_valid.patch
```diff
@@ -0,0 +1,57 @@
+From c89d1a129f71d3d2f76e6cbadb11ef41d8941a73 Mon Sep 17 00:00:00 2001
+From: Ann Pokora <apokora@juniper.net>
+Date: Tue, 25 May 2021 18:10:04 -0700
+Subject: [PATCH 2/2] mpls remove nl_addr_valid
+
+The removed calls to nl_addr_valid() are passing in a pointer to a binary address and the length of the binary address, which does not match expected arguments for nl_addr_valid(). nl_addr_valid() expects a pointer to an ASCII string and the address family of the string format.  The incorrect arguments cause unexpected failures and the expected arguments are not available in the context.
```
lguohan, line 6: can you make it more readable, make it multi-line, limit by 80 char per line?

lguohan:
i am only review the libnl part, for the yang model part, it should be reviewed in the sonic yang subgroup meeting. @rlhui,
 is this properly communicated to the team?

rlhui:
> i am only review the libnl part, for the yang model part, it should be reviewed in the sonic yang subgroup meeting. @rlhui,
> is this properly communicated to the team?

I was not aware this PR contains yang changes until last week.   @qbdwlr  could email the yang subgroup sonic-yang-subgroup@googlegroups.com and @anshuv-mfst  to request for review asap.

qbdwlr:
@rlhui @lguohan Changes in sonic-yang-models have been pulled from this PR and placed in a separate PR specifically for sonic-yang-models:  https://github.com/Azure/sonic-buildimage/pull/7881

smaheshm:
Looks good to me.. waiting for builds to pass.

qbdwlr:
@lguohan @smaheshm Please review/approve this PR.  The sonic-swss PR (https://github.com/Azure/sonic-swss/pull/1686) will not build successfully until this PR is merged to master.

smaheshm (approved):
👍

lguohan (approved):

smaheshm:
For posterity can you add the link to libnl3 upstream patch in the description section. Thanks!

PR #7219, opened 2021-04-02
[ci]: get build logs when build fails

lguohan:
#### Why I did it
get build logs when build fails for debugging

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012

### DIFF
--- a/.azure-pipelines/build-template.yml
+++ b/.azure-pipelines/build-template.yml
@@ -114,3 +114,12 @@ jobs:
   - publish: $(System.DefaultWorkingDirectory)/
     artifact: sonic-buildimage.${{ parameters.platform }}
     displayName: "Archive sonic image"
+  - script: |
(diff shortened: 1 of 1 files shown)


.azure-pipelines/build-template.yml
```diff
@@ -108,9 +108,19 @@ jobs:
         fi
 
         make USERNAME=admin $CACHE_OPTIONS SONIC_BUILD_JOBS=$(nproc) target/sonic-${{ parameters.platform }}.bin
+        false
```
liushilongbuaa, line 111: when build succeed, build exit with false. why?

lguohan: this is check failure cause to see the code is working or not.

https://dev.azure.com/mssonic/build/_build/results?buildId=9374&view=results

PR #7228, opened 2021-04-05
[debian] install systemd version 247 from buster-backports

stepanblyschak:

#### Why I did it

Update systemd to v247 in order to pick the fix for "core: coldplug possible nop_job" https://github.com/systemd/systemd/pull/13124

#### How I did it

Install systemd, systemd-sysv from buster-backports. Pass "systemd.unified_cgroup_hierarchy=0" as kernel argument to force systemd to not use unified cgroup hierarchy, otherwise dockerd won't start https://github.com/moby/moby/issues/16238.
Also, chown $FILSYSTEM_ROOT for root, otherwise apt systemd installation complains, see similar https://unix.stackexchange.com/questions/593529/can-not-configure-systemd-inside-a-chrooted-environment

#### How to verify it

Install on the switch, perform basic sanity tests/

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012

Update systemd to v247

### DIFF
--- a/build_debian.sh
+++ b/build_debian.sh
@@ -79,6 +79,8 @@ TARGET_PATH=$TARGET_PATH scripts/build_debian_base_system.sh $CONFIGURED_ARCH $I
 # Prepare buildinfo
 sudo scripts/prepare_debian_image_buildinfo.sh $CONFIGURED_ARCH $IMAGE_DISTRO $FILESYSTEM_ROOT $http_proxy
 
+sudo chown root $FILESYSTEM_ROOT
+
 ## Config hostname and hosts, otherwise 'sudo ...' will complain 'sudo: unable to resolve host ...'
 sudo LANG=C chroot $FILESYSTEM_ROOT /bin/bash -c "echo '$HOSTNAME' > /etc/hostname"
 sudo LANG=C chroot $FILESYSTEM_ROOT /bin/bash -c "echo '127.0.0.1       $HOSTNAME' >> /etc/hosts"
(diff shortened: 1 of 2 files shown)


build_debian.sh
```diff
@@ -79,6 +79,8 @@ TARGET_PATH=$TARGET_PATH scripts/build_debian_base_system.sh $CONFIGURED_ARCH $I
 # Prepare buildinfo
 sudo scripts/prepare_debian_image_buildinfo.sh $CONFIGURED_ARCH $IMAGE_DISTRO $FILESYSTEM_ROOT $http_proxy
 
+sudo chown root $FILESYSTEM_ROOT
```
qiluo-msft, line 82: Why suddenly need to chown? Do you want to change the group at the same time?

lguohan:
is there a behavior change when upgrade to 247 for the unified_cgroup_hierarchy? why is that we do not need pass this option prior to 247?

build_debian.sh
stepanblyschak: Yes, also changed the group to root

stepanblyschak:
@lguohan  yes, there is a behaviour change in systemd v247 comparing to 241. Systemd comes now with cgroup v2 (unified hierarchy) by default. Current docker version does not work with cgroup v2 (comming in docker 20.10). To solve this without updating docker a unified_cgroup_hierarchy is set to 0 in kernel arguments.

qiluo-msft (approved):
LGTM. Please wait other reviewers' approval.

PR #7232, opened 2021-04-05
[multi-asic] optimize api that checks port/port-channel/bgp is internal

abdosi:
What I did:
Optimize the API `is_port_internal` , `is_portchannel_internal` and `is_bgp_internal` to call `get_entry`(Redis hget) instead of `get_table` (Redis keys *)

Why I Did:
Reduces the time taken by show commands.

How I verified:
Verified `show ip bgp summary`  and `show interface status/counters` are fine.


arlakshm (approved):
lgtm

PR #7241, opened 2021-04-06
[multi-asic] Make sure Everflow tables always gets classified as Mirror table

abdosi:
What I did:
Make sure Everflow always gets classified as Mirror table and not as Control Plane on multi-asic platforms.

Why I did:
In Multi-asic platforms we generate Everflow acl table data from minigraph for both host and namespace.
It is possible in multi-asic minigraph if there are no external port-channel (Only Router Port IP Interface) then Everflow table will have no binded interface in host and will gets classified as Control Plane ACL while in namespace gets classified as Mirror Table. 
For ACL Rule generation we read global db as source of truth for acl table information and so for everflow rule generation if tables gets classified as Control plane we can generate rules with invalid action causing orchagent to throw runtime error.

How I did:
If the table is attach to erspan interface in minigraph then it always gets classified as mirror table.

How I verify:

Added unit-test case where minigraph only have External Router Port IP interface and no external port-channels.


arlakshm (approved):
lgtm

PR #7249, opened 2021-04-07
[baseimage]: specify gid for redis group.

renukamanavalan:
#### Why I did it
Default  groupadd for redis, takes 1000 by default. This forces, subsequently created admin group to get 1001.
As all TACACS users are created with 1000 as their gid, they end up in redis group.

#### How I did it
Explicitly provide the gid for redis group.

### How to verify it
Check /etc/group & group of RW users via TACACS

```
   admin@str-s6000-acs-9:~$ cat /etc/group | grep -e "admin\|redis"
    sudo:x:27:admin,user_rw
    docker:x:999:admin,user_rw
    redis:x:1001:admin               --- takes 1001
    admin:x:1000:                       --- retains 1000
    admin@str-s6000-acs-9:~$

    admin@str-s6000-acs-9:~$ ls -l /home
    total 4
    drwxr-xr-x 2 admin   admin   66 Apr  7 14:53 admin
    drwxr-xr-x 2 user_rw admin 4096 Apr  7 15:47 user_rw  -- RW user get admin group
```

- [ ] 201811
- [x ] 201911
- [x ] 202006
- [x ] 202012

### DIFF
--- a/build_debian.sh
+++ b/build_debian.sh
@@ -584,12 +584,10 @@ if [[ $CONFIGURED_ARCH == armhf || $CONFIGURED_ARCH == arm64 ]]; then
     DOCKERFS_PATH=../dockerfs/
 fi
 
-# Ensure admin gid 1000 is available
-gid_1000=$(sudo LANG=C grep -e ":1000:" $FILESYSTEM_ROOT/etc/group)
-if [ -n "${gid_1000}" ]; then
-    if [ "${gid_1000}" != "admin:x:1000:" ]; then
-        die "gid 1000 is in use. Expect: Reserve it for admin group. use:${gid_1000}"
-    fi
+# Ensure admin gid is 1000
+gid_user=$(sudo LANG=C chroot $FILESYSTEM_ROOT id -g $USERNAME) || gid_user="none"
+if [ "${gid_user}" != "1000" ]; then
+    die "expect gid 1000. current:${gid_user}"
 fi
 
 ## Compress docker files
--- a/build_debian.sh
+++ b/build_debian.sh
@@ -244,14 +244,15 @@ sudo cp files/docker/docker.service.conf $_
 ## Fix systemd race between docker and containerd
 sudo sed -i '/After=/s/$/ containerd.service/' $FILESYSTEM_ROOT/lib/systemd/system/docker.service
 
-## Create redis group
-sudo LANG=C chroot $FILESYSTEM_ROOT groupadd -f -g 1001 redis
-
 ## Create default user
 ## Note: user should be in the group with the same name, and also in sudo/docker/redis groups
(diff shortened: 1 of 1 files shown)


lguohan (dismissed):

build_debian.sh
```diff
@@ -245,7 +245,7 @@ sudo cp files/docker/docker.service.conf $_
 sudo sed -i '/After=/s/$/ containerd.service/' $FILESYSTEM_ROOT/lib/systemd/system/docker.service
 
 ## Create redis group
-sudo LANG=C chroot $FILESYSTEM_ROOT groupadd -f redis
+sudo LANG=C chroot $FILESYSTEM_ROOT groupadd -f -g 1001 redis
```
qiluo-msft, line 248: >1001 [](start = 51, length = 4)

I think the motivation is not to fix the gid for `redis`, but fix the gid for `admin` to 1000.
So suggest
1. create a group with gid 1000 and name `$USERNAME`
2. create a group `redis` without specifying gid
3. create user `$USERNAME` with `-g 1000` and also in the groups `sudo,docker,redis`

tahmed-dev (dismissed):

build_debian.sh
renukamanavalan: Agree, we need not  fix gid for redis. But we need to **fix gid for "admin" group**. 
At the same time, it does not limit/restrict anything by explicitly specifying gid for redis.

This fix is nearly a no-op for all purposes as all that is required is a *redis* group and we continue to have it as before.

lguohan: can we pin down the admin group more directly?

renukamanavalan: We could. Just a groupadd. Not sure, about any side effects. Just trying to make the change close to no-op. This change is ~= NO-OP

renukamanavalan: Hold on....There could be something simpler

renukamanavalan: Done. Just shifted order of creation.

build_debian.sh
```diff
@@ -583,6 +584,14 @@ if [[ $CONFIGURED_ARCH == armhf || $CONFIGURED_ARCH == arm64 ]]; then
     DOCKERFS_PATH=../dockerfs/
 fi
 
+# Ensure admin gid 1000 is available
+gid_1000=$(sudo LANG=C grep -e ":1000:" $FILESYSTEM_ROOT/etc/group)
+if [ -n "${gid_1000}" ]; then
+    if [ "${gid_1000}" != "admin:x:1000:" ]; then
```
qiluo-msft, line 590: >admin [](start = 27, length = 5)

`admin` is customizable. Use `$USERNAME` instead.
I guess you can just `id -g $USERNAME`

renukamanavalan: Good catch! Thank you!

qiluo-msft (commented):
Looks great, thanks!

PR #7256, opened 2021-04-07
SONiC QOS Scheduler, WRED, Queue Yangs

AshokDaparthi:
comments

#### Why I did it
Created SONiC Yang model for QOS
Tables:  SCHEDULER, WRED_PROFILE, QUEUE table.

#### How I did it
Defined Yang models for QOS based on Guideline doc:
https://github.com/Azure/SONiC/blob/master/doc/mgmt/SONiC_YANG_Model_Guidelines.md
and
https://github.com/Azure/sonic-utilities/blob/master/doc/Command-Reference.md

#### How to verify it
sonic_yang_models package build


AshokDaparthi:
Will update

PR #7262, opened 2021-04-08
[CI] Fix the wrong environment variable PLATFORM passing into the slave container issue

liushilongbuaa:
#### Why I did it
It is to fix the wrong usage of the build parameter in the azure pipelines' template. The build parameter PLATFORM=XXX can only be set when making configure (by command make configuration). To build platform images, we should avoid to set the value.
make PLATFORM=vs configuration
Bad usage:
make PLATFORM=vs  target/sonic-vs.img.gz
Good usage:
make target/sonic-vs.img.gz

#### How I did it
Change the environment PLATFORM to PLATFORM_AZP to avoid to pass the parameter.

### DIFF
--- a/.azure-pipelines/azure-pipelines-build.yml
+++ b/.azure-pipelines/azure-pipelines-build.yml
@@ -21,7 +21,7 @@ jobs:
     preSteps: ${{ parameters.preSteps }}
     postSteps: ${{ parameters.postSteps }}
     jobVariables:
-      PLATFORM: $(GROUP_NAME)
+      PLATFORM_FULL: $(GROUP_NAME)
       PLATFORM_ARCH: amd64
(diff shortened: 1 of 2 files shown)


xumia:
@qiluo-msft , is it a known issue? We cannot set the parameter "PLATFORM" when building vs images.
make PLATFORM=vs target/sonic-vs.img.gz

SONIC_BUILD_INSTRUCTION :=  make \
                           -f slave.mk \
                           PLATFORM=$(PLATFORM) \

lguohan:
please have proper description?

qiluo-msft:
please have proper title? not a good name 'hot fix'.

qiluo-msft (approved):

Issue #6781, opened 2021-02-14
PortChannel creation with out of range(0-255) value of "min-links", leads to Exception during update_data()


raphaelt-nvidia:
#### Description

Creating a portchannel with an invalid value for min-links results in repeated errors in the log, as well as a stack dump when executing "show interfaces portchannel".

#### Steps to reproduce the issue:
1. config portchannel add PortChannel10 --min-links 256

#### Describe the results you received:

show logging:

Feb 10 11:19:27.156197 r-lionfish-07 ERR snmp#snmp-subagent [ax_interface] ERROR: MIBUpdater.start() caught an unexpected exception during update_data()#012Traceback (most recent call last):#012  File "/usr/local/lib/python3.7/dist-packages/ax_interface/mib.py", line 43, in start#012    self.update_data()#012  File "/usr/local/lib/python3.7/dist-packages/sonic_ax_impl/mibs/ietf/rfc4292.py", line 70, in update_data#012    nexthops = ent["nexthop"]#012  File "/usr/lib/python3/dist-packages/swsscommon/swsscommon.py", line 345, in __getitem__#012    return _swsscommon.FieldValueMap___getitem__(self, key)#012IndexError: key not found

May not be observed in the first couple of minutes, but once it starts, it repeats about every 5 seconds.

Similarly "show interfaces portchannel" may initially produce output like this:

Flags: A - active, I - inactive, Up - up, Dw - Down, N/A - not available,
       S - selected, D - deselected, * - not synced
  No.  Team Dev       Protocol    Ports
-----  -------------  ----------  -------
   10  PortChannel10  N/A

but after that all invocations result in this stack dump:

Traceback (most recent call last):
  File "/usr/local/bin/show", line 8, in <module>
    sys.exit(cli())
  File "/usr/local/lib/python3.7/dist-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.7/dist-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.7/dist-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.7/dist-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.7/dist-packages/click/core.py", line 956, in invoke
[...]

PR #7265, opened 2021-04-08
Increase min-links limit for portchannel from 128 to 1024.

raphaelt-nvidia:
Signed-off-by: Raphael Tryster <raphaelt@nvidia.com>

#### Why I did it

Restrict the min-links parameter in "config portchannel" to the range 1-1024.
Fixes Azure/sonic-buildimage#6781 in conjunction with https://github.com/Azure/sonic-buildimage/pull/1630.
Align YANG model with limits in libteam and sonic-utilties.

#### How I did it

PR 1630 in sonic-utilities prevents CLI user from entering a value outside the allowed range.  This PR does the following:

- Increases the maximum value of min-links from 128 to 1024.
- Provides validation in libteam, incorporating as a patch the code in https://git.kernel.org/pub/scm/linux/kernel/git/jpirko/libteam.git/commit/?id=69a7494bb77dc10bb27076add07b380dbd778592.
- Updates the Yang model upper limit from 128 to 1024 (was inconsistent with libteam value).
- Updates the Yang model lower limit from 1 to 0, since 0 is set as default in sonic-utilities which would fail its new range check otherwise.
- Added Yang tests for valid and invalid value.

#### How to verify it

config portchannel add PortChannel0004 --min-links 1024

Command should be accepted.

show interfaces portchannel

Output should show PortChannel0004, no errors on CLI.

config portchannel add PortChannel0005 --min-links 1025

Command should be rejected

show interfaces portchannel

Output should not show PortChannel0005 , no errors on CLI.

Updates YANG model to allow up to 1024 min_links for portchannel.  Fixes Azure/sonic-buildimage#6781 in conjunction with https://github.com/Azure/sonic-buildimage/pull/1630.

### DIFF
--- a/src/sonic-yang-models/tests/yang_model_tests/tests_config/portchannel.json
+++ b/src/sonic-yang-models/tests/yang_model_tests/tests_config/portchannel.json
@@ -28,6 +28,70 @@
                         "mtu": "9100",
                         "tpid": "0x8100",
                         "lacp_key": "auto",
+			"name": "PortChannel0001"
+                    }
+                ]
+            }
+        }
+    },
+    "PORT_CHANNEL_MAX_VALID_MIN_LINKS": {
+        "sonic-port:sonic-port": {
+            "sonic-port:PORT": {
+                "PORT_LIST": [
+                    {
+                        "admin_status": "up",
+                        "alias": "eth0",
+                        "description": "Ethernet0",
+                        "lanes": "65",
+                        "mtu": 9000,
+                        "name": "Ethernet0",
+                        "speed": 25000
+                    }
+                ]
+            }
+        },
+        "sonic-portchannel:sonic-portchannel": {
+            "sonic-portchannel:PORTCHANNEL": {
+                "PORTCHANNEL_LIST": [
+                    {
+                        "admin_status": "up",
+                        "members": [
+                            "Ethernet0"
+                        ],
+                        "min_links": "1024",
+                        "mtu": "9100",
+                        "name": "PortChannel0001"
+                    }
+                ]
+            }
+        }
+    },
+    "PORT_CHANNEL_OUT_OF_RANGE_MIN_LINKS": {
+        "sonic-port:sonic-port": {
+            "sonic-port:PORT": {
+                "PORT_LIST": [
+                    {
+                        "admin_status": "up",
+                        "alias": "eth0",
+                        "description": "Ethernet0",
+                        "lanes": "65",
+                        "mtu": 9000,
+                        "name": "Ethernet0",
+                        "speed": 25000
+                    }
+                ]
+            }
+        },
+        "sonic-portchannel:sonic-portchannel": {
+            "sonic-portchannel:PORTCHANNEL": {
+                "PORTCHANNEL_LIST": [
+                    {
+                        "admin_status": "up",
+                        "members": [
+                            "Ethernet0"
+                        ],
+                        "min_links": "1025",
+                        "mtu": "9100",
                         "name": "PortChannel0001"
                     }
                 ]
--- a/src/sonic-yang-models/tests/yang_model_tests/tests_config/portchannel.json
+++ b/src/sonic-yang-models/tests/yang_model_tests/tests_config/portchannel.json
@@ -28,7 +28,7 @@
                         "mtu": "9100",
                         "tpid": "0x8100",
                         "lacp_key": "auto",
-			"name": "PortChannel0001"
+                        "name": "PortChannel0001"
                     }
                 ]
             }
--- a/src/sonic-yang-models/yang-models/sonic-portchannel.yang
+++ b/src/sonic-yang-models/yang-models/sonic-portchannel.yang
@@ -26,6 +26,10 @@ module sonic-portchannel {
 
     description "PORTCHANNEL yang Module for SONiC OS";
 
+    revision 2021-05-13 {
+        description "Change min-links valid range from 1-128 to 0-1024";
+    }
+
     revision 2021-03-31 {
         description "Add PortChannel Interface List with VRF attribute";
     }
@@ -75,9 +79,7 @@ module sonic-portchannel {
 
                 leaf min_links {
                     type uint16 {
-                        range 1..1024;
-                    type uint8 {
-                        range 1..128;
+                        range 0..1024;
                     }
                 }
 
--- a/src/sonic-yang-models/yang-models/sonic-portchannel.yang
+++ b/src/sonic-yang-models/yang-models/sonic-portchannel.yang
@@ -26,8 +26,8 @@ module sonic-portchannel {
 
     description "PORTCHANNEL yang Module for SONiC OS";
 
-    revision 2021-05-13 {
-        description "Change min-links valid range from 1-128 to 0-1024";
+    revision 2021-06-13 {
+        description "Change min-links valid range from 1-128 to 1-1024";
     }
 
     revision 2021-03-31 {
@@ -79,7 +79,7 @@ module sonic-portchannel {
 
                 leaf min_links {
                     type uint16 {
-                        range 0..1024;
+                        range 1..1024;
                     }
                 }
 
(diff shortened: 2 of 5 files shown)


lguohan:
do you need to update the libteam as well?

raphaelt-nvidia:
The link I quoted above is the change in libteam.

src/sonic-yang-models/yang-models/sonic-portchannel.yang
```diff
@@ -23,7 +23,7 @@ module sonic-portchannel {
 	description "PORTCHANNEL yang Module for SONiC OS";
 
 	revision 2019-07-01 {
-		description "First Revision";
+		description "Increased range for min_links";
```
anshuv-mfst, line 26: Yang model subgroup meeting 4/8: 
Please confirm if revision needs to be changed. (cc: @praveen-li)

src/sonic-yang-models/yang-models/sonic-portchannel.yang
```diff
@@ -64,8 +64,8 @@ module sonic-portchannel {
 				}
 
 				leaf min_links {
-					type uint8 {
-						range 1..128;
+					type uint16 {
+						range 1..1024;
 					}
```
anshuv-mfst, line 69: Yang model subgroup meeting 4/8:
Do any tests need to be updated with new range limit (128->1024)? (cc: @praveen-li )

src/sonic-yang-models/yang-models/sonic-portchannel.yang
joyas-joseph: For each update to the yang, it is better to maintain a corresponding revision with the description of the change.

joyas-joseph:
> The link I quoted above is the change in libteam.

That change is not available in sonic yet, right? The version being checked out is much older (from Feb 4, 2020).

```
        git clone https://github.com/jpirko/libteam.git
        pushd ./libteam
        git checkout -b teamd -f c7237377dead39ae4a711297203bacf7edb9fa41
```

liat-grozovik:
@lguohan and @judyjoseph can you please clarify what is needed?
do we want to keep this change and work to fix the conflict or should we abandon this as libteam is not the latest one? i dont think libteam should be updated just because of this yang model.

joyas-joseph:
> @lguohan and @judyjoseph can you please clarify what is needed?
> do we want to keep this change and work to fix the conflict or should we abandon this as libteam is not the latest one? i dont think libteam should be updated just because of this yang model.

Let's have it consistent.
We can have the range 1..1024, if we bring in at least the commit (https://github.com/jpirko/libteam/commit/69a7494bb77dc10bb27076add07b380dbd778592) as a patch.
Without that, lets have the range 1..255.

src/sonic-yang-models/yang-models/sonic-portchannel.yang
venkatmahalingam: Looks like a test case needs to be added for the range check.

lguohan:
can you help to resolve conflict?

src/sonic-yang-models/yang-models/sonic-portchannel.yang
```diff
@@ -74,8 +78,8 @@ module sonic-portchannel {
                 }
 
                 leaf min_links {
-                    type uint8 {
-                        range 1..128;
+                    type uint16 {
+                        range 0..1024;
                     }
```
joyas-joseph, line 83: Shouldnt the range be 1..1024?

If we do support value 0, what does it indicate?

src/sonic-yang-models/tests/yang_model_tests/tests_config/portchannel.json
```diff
@@ -28,6 +28,70 @@
                         "mtu": "9100",
                         "tpid": "0x8100",
                         "lacp_key": "auto",
+			"name": "PortChannel0001"
```
qiluo-msft, line 31: >			 [](http://example.com/codeflow?start=0&length=3)

Please use spaces instead of tabs

src/sonic-yang-models/yang-models/sonic-portchannel.yang
qiluo-msft: Agree

qiluo-msft (changes requested):
As comments

src/sonic-yang-models/yang-models/sonic-portchannel.yang
raphaelt-nvidia: I found this necessary to be compatible with related https://github.com/Azure/sonic-utilities/pull/1630.  The original code there sets the default value of min-links to 0, and now that I have introduced range checking there, keeping a lower bound of 1 causes range check to fail.  So this change aligns yang with CLI.

src/sonic-yang-models/tests/yang_model_tests/tests_config/portchannel.json
raphaelt-nvidia: Done

src/sonic-yang-models/yang-models/sonic-portchannel.yang
qiluo-msft: Did you test the 0 in current sonic image? Does teamd support it? ie. you set it as 0, and admin down all the member ports, and the portchannel itself is still up?

raphaelt-nvidia: Results are the same with current master and with the changes in this PR.  Recall that the default value of min-links in sonic-utilities is 0.  In both cases, shutting down all members of the LAG causes the portchannel to go oper down.  I noticed some non-deterministic behaviour, that may be a separate bug: with one of two member ports down, the LAG is sometimes up and sometimes down. 

Since the default is 0, one would expect that in the current master, a portchannel created with default min-links should be up when all members are down.  On the other hand, the work in this PR is supposed to clean up mismatches, and based on Qi's question, I now think that it doesn't do what it should.  What I did was align yang with CLI, but it seems teamd is now the odd one out, as the change in teamd only adjusted the upper limit and kept the lower at 1.  If you think it is reasonable to specify that a portchannel with all members down should be up (i.e. min-links = 0), then we have some more work to do to make teamd comply.  (Any other places?)  If not, then I should do the alignment the other way: put the lower limit in yang back to 1, and change the lower limit and default in sonic-utilities to 1.

qiluo-msft: Thanks for the great insight! I prefer set lower limit of min-link to 1 to make life a little bit easier.

joyas-joseph: I agree. Let's have the lower limit of min-links set to 1.

joyas-joseph (dismissed):

qiluo-msft (approved):

liat-grozovik:
@qiluo-msft and @judyjoseph could you please help to review recent fixes and approve for merge?

joyas-joseph (approved):

qiluo-msft:
I am confused about the numbers in PR description, like 0, 255. So I changed it according to my understanding. Please double check.

qiluo-msft (approved):

raphaelt-nvidia:
> I am confused about the numbers in PR description, like 0, 255. So I changed it according to my understanding. Please double check.

Now I am confused.  What and where did you change?  I think your confusion stems from the mismatches that existed before this work was done, and it appears above in the discussion.  Originally, teamd was 1-255, cli was any integer with default 0, and yang was 1-128.  Now, they should all be aligned on 1-1024.

qiluo-msft:
You can check my changes by clicking the dropdown menu button
image

raphaelt-nvidia:
> You can check my changes by clicking the dropdown menu button
> image

I had a bit of trouble finding it because I see it as Apr 8.  I deduce from this that you are in China (I am in Israel) and github shows the date of a comment according to the viewer's time zone.  Anyway, I saw your changes, and I think they are neither better nor worse than the original, because of the mismatches between the modules.  So, no problem.

PR #7273, opened 2021-04-09
[sonic-py-common] Clear environment variables before running device_info tests

jleveque:
#### Why I did it

To ensure any environment variables which are configured in the build/test environment do not influence the behavior of sonic-py-common during unit tests. For example, variables which might be set by continuous integration pipelines.

#### How I did it

Add class-scoped pytest fixture to `TestDeviceInfo` class which stashes the current environment variables, clears them and yields. Once all the test cases in the class finish, the fixture will restore the original environment variables.

Also remove unnecessary unittest-style setup and teardown functions from interface_test.py

#### How to verify it

Ensure unit tests pass during package build with the following commands:
```
PLATFORM=not_a_valid_platform_string make target/python-wheels/sonic_py_common-1.0-py2-none-any.whl
PLATFORM=not_a_valid_platform_string make target/python-wheels/sonic_py_common-1.0-py3-none-any.whl
```

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012

### DIFF
--- a/src/sonic-py-common/tests/device_info_test.py
+++ b/src/sonic-py-common/tests/device_info_test.py
@@ -9,6 +9,8 @@ else:
     # https://pypi.python.org/pypi/mock
     import mock
 
+import pytest
+
 from sonic_py_common import device_info
 
 from .mock_swsssdk import SonicV2Connector
@@ -51,9 +53,16 @@ EXPECTED_GET_MACHINE_INFO_RESULT = {
 }
 
 class TestDeviceInfo(object):
-    @classmethod
-    def setup_class(cls):
-        print("SETUP")
+    @pytest.fixture(scope="class", autouse=True)
+    def sanitize_environment(self):
+        # Back up environment variables then clear them, in case an environment
+        # variable is set in the test environment (e.g., PLATFORM) which could
+        # modify the behavior of sonic-py-common
+        env_orig = os.environ.copy()
+        os.environ.clear()
+        yield
+        # Restore original environment variables
+        os.environ.update(env_orig)
 
     def test_get_machine_info(self):
         with mock.patch("os.path.isfile") as mock_isfile:
--- a/src/sonic-py-common/tests/device_info_test.py
+++ b/src/sonic-py-common/tests/device_info_test.py
@@ -55,14 +55,10 @@ EXPECTED_GET_MACHINE_INFO_RESULT = {
 class TestDeviceInfo(object):
     @pytest.fixture(scope="class", autouse=True)
     def sanitize_environment(self):
-        # Back up environment variables then clear them, in case an environment
-        # variable is set in the test environment (e.g., PLATFORM) which could
-        # modify the behavior of sonic-py-common
-        env_orig = os.environ.copy()
-        os.environ.clear()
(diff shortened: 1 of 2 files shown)


src/sonic-py-common/tests/device_info_test.py
```diff
@@ -51,9 +53,16 @@
 
 
 class TestDeviceInfo(object):
-    @classmethod
-    def setup_class(cls):
-        print("SETUP")
+    @pytest.fixture(scope="class", autouse=True)
+    def sanitize_environment(self):
```
lguohan, line 57: it seems doing something like tox. https://tox.readthedocs.io/en/latest/index.html#basic-example

tox will take care of environment isolation for you: it will strip away all operating system environment variables not specified via passenv. Furthermore, it will also alter the PATH variable so that your commands resolve first and foremost within the current active tox environment. In general all executables in the path are available in commands, but tox will emit a warning if it was not explicitly allowed via allowlist_externals.

src/sonic-py-common/tests/device_info_test.py
```diff
@@ -51,9 +53,16 @@
...
+        # Back up environment variables then clear them, in case an environment
+        # variable is set in the test environment (e.g., PLATFORM) which could
+        # modify the behavior of sonic-py-common
+        env_orig = os.environ.copy()
+        os.environ.clear()
+        yield
+        # Restore original environment variables
+        os.environ.update(env_orig)
```
lguohan, line 65: are you only modify your own environ, why do you need to restore it? but again, i feel we are creating some virtual env here.

src/sonic-py-common/tests/device_info_test.py
jleveque: True. Do we want to introduce tox to our build process?

src/sonic-py-common/tests/device_info_test.py
jleveque: Correct, there is really no need to restore the old environment, it's just a best practice to leave things as you found them.

src/sonic-py-common/tests/device_info_test.py
```diff
@@ -51,9 +53,16 @@
 
 
 class TestDeviceInfo(object):
-    @classmethod
-    def setup_class(cls):
-        print("SETUP")
+    @pytest.fixture(scope="class", autouse=True)
+    def sanitize_environment(self):
+        # Back up environment variables then clear them, in case an environment
```
qiluo-msft, line 58: >Back up [](http://example.com/codeflow?start=10&length=7)

Mocking is a better way to create an isolated environment.
ref: [https://adamj.eu/tech/2020/10/13/how-to-mock-environment-variables-with-pytest/](https://adamj.eu/tech/2020/10/13/how-to-mock-environment-variables-with-pytest/)

jleveque: Refactored to use `mock` in latest commit.

qiluo-msft (approved):

lguohan (approved):

qiluo-msft (approved):

PR #7173, opened 2021-03-29
Fix CVE-2017-1000487 security alert

xumia:
#### Why I did it
Plexus-utils before 3.0.16 is vulnerable to command injection because it does not correctly process the contents of double quoted strings.

#### How I did it
Upgrade to 3.0.16

#### Which release branch to backport (provide reason below if selected)

- [x] 201811
- [x] 201911
- [x] 202006
- [x] 202012

PR #7278, opened 2021-04-09
Fix CVE-2017-1000487 security alert (#7173)

xumia:
#### Why I did it
Plexus-utils before 3.0.16 is vulnerable to command injection because it does not correctly process the contents of double quoted strings.

Upgrade to 3.0.16

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012


qiluo-msft (approved):
LGTM.

PR #7284, opened 2021-04-10
RADIUS Management User Authentication Feature

a-barboza:
#### Why I did it

HLD: https://github.com/Azure/SONiC/blob/master/doc/aaa/radius_authentication.md
CLI: In a separate PR.

#### How to verify it

UT: src/sonic-host-services/tests/hostcfgd/hostcfgd_radius_test.py

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012


src/sonic-host-services/scripts/hostcfgd
```diff
@@ -56,6 +73,7 @@ def run_cmd(cmd, log_err = True):
         if log_err:
             syslog.syslog(syslog.LOG_ERR, "{} - failed: return code - {}, output:\n{}"
                   .format(err.cmd, err.returncode, err.output))
+    return 0
```
renukamanavalan, line 76: It should be the subprocess return code and err.returncode in case of exception.

src/sonic-host-services/scripts/hostcfgd
```diff
@@ -169,6 +208,29 @@ class AaaCfg(object):
         if modify_conf:
             self.modify_conf_file()
 
+    def tacacs_get_source_intf_ip(self, keys, src_intf):
+        new_ipv4_addr = ""
```
renukamanavalan, line 212: Is the goal to get first IPv4 & IPv6 address ?

Can you please change name to something generic like, "ip-addr-parser" ?

src/sonic-host-services/scripts/hostcfgd
```diff
@@ -185,6 +247,104 @@ class AaaCfg(object):
         if modify_conf:
             self.modify_conf_file()
 
+    def handle_radius_source_intf_ip_chg(self, key, keys):
```
renukamanavalan, line 250: keys is not used.

src/sonic-host-services/scripts/hostcfgd
```diff
@@ -185,6 +247,104 @@ class AaaCfg(object):
...
+        modify_conf=False
+        if 'src_intf' in self.radius_global:
+            if key[0] == self.radius_global['src_intf']:
```
renukamanavalan, line 253: I guess, you modify, if 'src_intf' not in global too,

src/sonic-host-services/scripts/hostcfgd
```diff
@@ -209,29 +369,78 @@ class AaaCfg(object):
...
+
+                if 'src_intf' in server:
+                    # RADIUS: Log a message if src_ip is already defined.
+                    if 'src_ip' in server:
+                        syslog.syslog(syslog.LOG_INFO, \
+            "RADIUS_SERVER|{}: src_intf found. May ignore src_ip".format(addr))
+                    # RADIUS: If server.src_intf, then get the corresponding
+                    # src_ip based on the server.ip, and set it.
+                    src_ip = self.get_interface_ip(server['src_intf'], addr)
+                    if len(src_ip) > 0:
```
renukamanavalan, line 400: You may need to remove the src_ip if len != 0

src/sonic-host-services/scripts/hostcfgd
```diff
@@ -209,29 +369,78 @@ class AaaCfg(object):
...
+        os.rename(PAM_AUTH_CONF + ".tmp", PAM_AUTH_CONF)
 
-        # Modify common-auth include file in /etc/pam.d/login and sshd
+        # Modify common-auth include file in /etc/pam.d/login, sshd, and sudo
         if os.path.isfile(PAM_AUTH_CONF):
-            self.modify_single_file('/etc/pam.d/sshd',  [ "'/^@include/s/common-auth$/common-auth-sonic/'" ])
-            self.modify_single_file('/etc/pam.d/login', [ "'/^@include/s/common-auth$/common-auth-sonic/'" ])
+            self.modify_single_file(ETC_PAMD_SSHD,  [ "'/^@include/s/common-auth$/common-auth-sonic/'" ])
+            self.modify_single_file(ETC_PAMD_LOGIN, [ "'/^@include/s/common-auth$/common-auth-sonic/'" ])
+            self.modify_single_file(ETC_PAMD_SUDO, [ "'/^@include/s/common-auth$/common-auth-sonic/'" ])
```
renukamanavalan, line 425: sudo does not deal with remote auth. 

Why is this change of behavior?

src/sonic-host-services-data/debian/sonic-host-services-data.aaastatsd.service
```diff
@@ -0,0 +1,14 @@
+[Unit]
```
renukamanavalan, line 1: How does service restart works, if it crashes

src/sonic-host-services/scripts/hostcfgd
```diff
@@ -240,6 +449,37 @@ class AaaCfg(object):
...
+                template = env.get_template(template_file)
+                pam_radius_auth_conf = template.render(server=srv)
+
+                open(pam_radius_auth_file, 'a').close()
+                os.chmod(pam_radius_auth_file, 0o600)
+                with open(pam_radius_auth_file, 'w+') as f:
+                    f.write(pam_radius_auth_conf)
+
+        # Start the statistics service
+        cmd = 'service aaastatsd start'
```
renukamanavalan, line 474: Do we stop this, when remote login is *not* enabled ?

Or may be OK to run it always, irrespective of AAA is active or not ?

src/sonic-host-services/scripts/hostcfgd
a-barboza: Does forcing the user to re-authenticate when running the sudo depend on site security policy? I believe this is the reason for  adding it to sudo as well for remote authentication. Otherwise, the re-authentication would occur with the local password (which would not exist for remote authenticated users).

src/sonic-host-services/scripts/hostcfgd
a-barboza: Yes, the fields of the key are not used directly, since we have to get an IP address (v4 or v6) to use. It can be removed.

src/sonic-host-services/scripts/hostcfgd
```diff
@@ -209,29 +369,78 @@ class AaaCfg(object):
...
+                if 'src_intf' in server:
+                    # RADIUS: Log a message if src_ip is already defined.
+                    if 'src_ip' in server:
+                        syslog.syslog(syslog.LOG_INFO, \
+            "RADIUS_SERVER|{}: src_intf found. May ignore src_ip".format(addr))
+                    # RADIUS: If server.src_intf, then get the corresponding
+                    # src_ip based on the server.ip, and set it.
+                    src_ip = self.get_interface_ip(server['src_intf'], addr)
+                    if len(src_ip) > 0:
+                        server['src_ip'] = src_ip
```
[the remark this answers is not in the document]
renukamanavalan: I get that. My comment is, 
server['src_ip'] = src_ip, irrespective of len.
For e.g. len(server['src_ip'] > 0 && len(src_ip) == 0, don't we need to nullify server['src_ip'] entry ?

src/sonic-host-services/scripts/hostcfgd
renukamanavalan: Again, this is change of behavior. 
Today sudo does not do remote user auth. 
Agree, that some security policy might need. In this case, we need to provide a configurable knob and conditionally enable it for sudo,

src/sonic-host-services-data/debian/sonic-host-services-data.aaastatsd.service
renukamanavalan: Good point. So far, hostcfgd happens to be a *great* guy that we never got burnt by it. Thanks to all Try/Catch, I guess.

Said that, if aaaastatsd crashes, it may never come up until some config change related to it, as hostcfgd is less likely to crash.

So let us add auto-restart to it.

src/sonic-host-services/scripts/hostcfgd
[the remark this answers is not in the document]
a-barboza: I think I understand the comment as requesting the following. If the user specified both, a valid 'src_ip' at the global level, and a src_intf for the server, but the src_intf could not resolve to a valid IP address, nullify the 'src_ip' attribute of the server, since the user attempted to over-ride it with a src_intf (which could not resolve to a valid IP address). In other words, if a source interface was specified, ignore src-ip. I will fix with comment

src/sonic-host-services/scripts/hostcfgd
a-barboza: Took out sudo PAM file modification

src/sonic-host-services-data/debian/sonic-host-services-data.aaastatsd.service
a-barboza: ok

src/sonic-host-services/scripts/hostcfgd
renukamanavalan: 1) This is still not fixed. It would still return 0, even in case of exception
2) Noticed there is a duplicate run_cmd at line 40

src/sonic-host-services/scripts/hostcfgd
renukamanavalan: Are you removing "keys" ?

renukamanavalan (approved):

src/sonic-host-services/scripts/aaastatsd
```diff
@@ -0,0 +1,222 @@
...
+
+    def create_file(self, addr):
+        # print( "RadiusStatistics.create_file({})".format(addr))
+        stats_file = RADIUS_PAM_AUTH_CONF_STATS_DIR + addr
+        if self.radius_global['statistics'] == 'False':
+            if os.path.exists(stats_file):
+                os.unlink(stats_file)
+        else:
+            open(stats_file, 'a').close()
+            os.chmod(stats_file, 0o666)
```
maipbui, line 154: Semgrep detects a widely permissive file permission issue.
```
 scripts/aaastatsd
     python.lang.security.audit.insecure-file-permissions.insecure-file-permissions
        These permissions `0o666` are widely permissive and grant access to more people than may be
        necessary. A good default is `0o644` which gives read and write access to yourself and read
        access to everyone else.
        Details: https://sg.run/AXY4

        154┆ os.chmod(stats_file, 0o666)
```
Github code scanning also detects same issue.
Is it possible to limit the rw access to the owner only and give no access to others?

maipbui: +@a-barboza

a-barboza: @maipbui 
I think 0o644 or 0o600 might be ok. There is no sensitive information in this file.

maipbui: Thanks @a-barboza! I’ll raise a PR to address this.

PR #7297, opened 2021-04-12
[DPB][YANG-models] Update portchannel yang model to have lacp_key

DavidZagury:
This PR depends on https://github.com/Azure/sonic-swss/pull/1660

#### Why I did it
To update the yang model to support the new key interduced in https://github.com/Azure/sonic-swss/pull/1660
#### How I did it
Add to the portchanels yang file an definition for the lacp_key

#### How to verify it
Run the command `config interface breakout <interface> <breakout_mode>`

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012

### DIFF
--- a/src/sonic-yang-models/yang-models/sonic-portchannel.yang
+++ b/src/sonic-yang-models/yang-models/sonic-portchannel.yang
@@ -95,7 +95,21 @@ module sonic-portchannel {
                     mandatory true;
                     type stypes:admin_status;
                 }
-            } /* end of list PORTCHANNEL_LIST */
+
+                leaf lacp_key {
+                    /* lacp key should be either auto or a integer in the range of 1 to 65535 */
+                    /* When lacp_key is set to 'auto' the lacp key would be set to be the number */
+                    /* on the end of the PortChannel name with a prefix of 1 */
+                    type union {
+                        type string {
+                            pattern "auto";
+                        }
+                        type uint16 {
+                            range 1..65535;
+                        }
+                    }
+                }
+			} /* end of list PORTCHANNEL_LIST */
 
         } /* end of container PORTCHANNEL */
 
--- a/src/sonic-yang-models/tests/files/sample_config_db.json
+++ b/src/sonic-yang-models/tests/files/sample_config_db.json
@@ -11,7 +11,8 @@
                 "members": [
                     "Ethernet1"
                 ],
-                "mtu": "9100"
+                "mtu": "9100",
+                "lacp_key": "auto"
             },
             "PortChannel0004": {
                 "admin_status": "up",
@@ -19,7 +20,8 @@
                 "members": [
                     "Ethernet2"
                 ],
-                "mtu": "9100"
+                "mtu": "9100",
+                "lacp_key": "auto"
             }
         },
         "PORTCHANNEL_INTERFACE": {
--- a/src/sonic-yang-models/tests/yang_model_tests/tests_config/portchannel.json
+++ b/src/sonic-yang-models/tests/yang_model_tests/tests_config/portchannel.json
@@ -25,6 +25,7 @@
                         ],
                         "min_links": "1",
                         "mtu": "9100",
+                        "lacp_key": "auto",
                         "name": "PortChannel0001"
                     }
                 ]


dprital:
Can someone trigger the tests again ?

anshuv-mfst:
Hi @prsunny - could you please take a look, thanks.

src/sonic-yang-models/yang-models/sonic-portchannel.yang
```diff
@@ -95,7 +95,14 @@ module sonic-portchannel {
                     mandatory true;
                     type stypes:admin_status;
                 }
-            } /* end of list PORTCHANNEL_LIST */
+
+				leaf lacp_key {
+					type string {
+						/* lacp key should be either auto or a integer in the range of 1 to 65535 */
+						pattern '[1-9]{1,4}|[1-5][0-9]{4}|6[0-4][0-9]{3}|65[0-4][0-9]{2}|655[0-2][0-9]|6553[0-5]|auto';
```
judyjoseph, line 102: I find this expression is wrong ?  Shouldn't the starting regex be [1-9][0-9]{1,3}. You could use this link to check out https://regexr.com/

src/sonic-yang-models/yang-models/sonic-portchannel.yang
```diff
@@ -95,7 +95,14 @@ module sonic-portchannel {
                     mandatory true;
                     type stypes:admin_status;
                 }
-            } /* end of list PORTCHANNEL_LIST */
+
+				leaf lacp_key {
```
judyjoseph, line 99: Need to allign the format here.

src/sonic-yang-models/yang-models/sonic-portchannel.yang
```diff
@@ -95,7 +95,14 @@ module sonic-portchannel {
...
+					type string {
+						/* lacp key should be either auto or a integer in the range of 1 to 65535 */
+						pattern '[1-9]{1,4}|[1-5][0-9]{4}|6[0-4][0-9]{3}|65[0-4][0-9]{2}|655[0-2][0-9]|6553[0-5]|auto';
+					}
```
judyjoseph, line 103: I feel we should keep "auto" string with default keyword -- to scenario where  user don't give any input

src/sonic-yang-models/yang-models/sonic-portchannel.yang
qiluo-msft, line 102: >pattern [](http://example.com/codeflow?start=6&length=7)

This regex is too complex. Could you use a `union` of string and uint16 with a range?

DavidZagury: @qiluo-msft  Didn't know that I can do this, fix this. Thank you

src/sonic-yang-models/yang-models/sonic-portchannel.yang
DavidZagury: We can't treat a blank string as "auto", if this is blank we are maintaining the old behavior of keeping lacp key as 0 (although it is not an allowed value by the LACP spec).

This is because we wanted to make sure that if a user is upgrading from an older version and he uses warm-reboot to do so, the connection will remain up and not lose any packets, and to do so we need to keep the lacp key as 0.
So, what we chose to do is to treat the scenrio of lacp_key empty\non existing as lacp key 0 and to not update the DB with lacp key when a warm update has occured.

qiluo-msft (dismissed):

src/sonic-yang-models/yang-models/sonic-portchannel.yang
```diff
@@ -95,7 +95,19 @@ module sonic-portchannel {
                     mandatory true;
                     type stypes:admin_status;
                 }
-            } /* end of list PORTCHANNEL_LIST */
+
+                leaf lacp_key {
+                    /* lacp key should be either auto or a integer in the range of 1 to 65535 */
+                    type union {
+                        type string {
+                            pattern "auto";
```
lguohan, line 103: can you add description for what does auto mean here? it is not clear what it means.

src/sonic-yang-models/yang-models/sonic-portchannel.yang
judyjoseph: Ok I see,  Would be good to add a description on this default case as well then.

src/sonic-yang-models/yang-models/sonic-portchannel.yang
```diff
@@ -95,7 +95,21 @@ module sonic-portchannel {
                     mandatory true;
                     type stypes:admin_status;
                 }
-            } /* end of list PORTCHANNEL_LIST */
+
+                leaf lacp_key {
```
praveen-li, line 99: Kindly add test cases in https://github.com/Azure/sonic-buildimage/blob/master/src/sonic-yang-models/tests/yang_model_tests/tests_config/portchannel.json.

And Sample config in https://github.com/Azure/sonic-buildimage/blob/master/src/sonic-yang-models/tests/files/sample_config_db.json. Thx

liat-grozovik:
@praveen-li can you please review?

qiluo-msft (approved):

PR #7307, opened 2021-04-13
Add split mode to routing config

Pterosaur:
Signed-off-by: Ze Gan <ganze718@gmail.com>

#### Why I did it
For the split mode, the config files, like `bgpd.conf`, `zebra.conf` and so on, were provided by outside. But the `docker_init.sh` will overwrite the outside config files if restart bgp service.

#### How I did it

Add a split mode checking in `docker_init.sh`, if `docker_routing_config_mode` is `split`, don't overwrite the existing routing config files.

#### How to verify it

1. Set split mode in config db
```json
{
    "DEVICE_METADATA": {
        "localhost": {
            "hwsku": "Force10-S6000",
            "platform": "x86_64-kvm_x86_64-r0",
            "docker_routing_config_mode": "split"
            ...
        }
    }
}
```
2. Replace your `bgpd.conf` to `/etc/sonic/frr/bgpd.conf`
3. Restart bgp service by `sudo service bgp restart`

The `/etc/sonic/frr/bgpd.conf` your provided shouldn't be overwritten 

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012

### DIFF
--- a/dockers/docker-fpm-frr/docker_init.sh
+++ b/dockers/docker-fpm-frr/docker_init.sh
@@ -3,20 +3,24 @@
 mkdir -p /etc/frr
 mkdir -p /etc/supervisor/conf.d
 
+ETC_FRR_BACKUP="/etc/frr/backup"
+
+mkdir -p $ETC_FRR_BACKUP
+
 CFGGEN_PARAMS=" \
     -d \
     -y /etc/sonic/constants.yml \
     -t /usr/share/sonic/templates/frr_vars.j2 \
     -t /usr/share/sonic/templates/supervisord/supervisord.conf.j2,/etc/supervisor/conf.d/supervisord.conf \
-    -t /usr/share/sonic/templates/bgpd/gen_bgpd.conf.j2,/etc/frr/bgpd.conf \
+    -t /usr/share/sonic/templates/bgpd/gen_bgpd.conf.j2,$ETC_FRR_BACKUP/bgpd.conf \
     -t /usr/share/sonic/templates/supervisord/critical_processes.j2,/etc/supervisor/critical_processes \
-    -t /usr/share/sonic/templates/zebra/zebra.conf.j2,/etc/frr/zebra.conf \
-    -t /usr/share/sonic/templates/staticd/gen_staticd.conf.j2,/etc/frr/staticd.conf \
-    -t /usr/share/sonic/templates/gen_frr.conf.j2,/etc/frr/frr.conf \
+    -t /usr/share/sonic/templates/zebra/zebra.conf.j2,$ETC_FRR_BACKUP/zebra.conf \
+    -t /usr/share/sonic/templates/staticd/gen_staticd.conf.j2,$ETC_FRR_BACKUP/staticd.conf \
+    -t /usr/share/sonic/templates/gen_frr.conf.j2,$ETC_FRR_BACKUP/frr.conf \
     -t /usr/share/sonic/templates/isolate.j2,/usr/sbin/bgp-isolate \
     -t /usr/share/sonic/templates/unisolate.j2,/usr/sbin/bgp-unisolate \
-    -t /usr/local/sonic/frrcfgd/bfdd.conf.j2,/etc/frr/bfdd.conf \
-    -t /usr/local/sonic/frrcfgd/ospfd.conf.j2,/etc/frr/ospfd.conf \
+    -t /usr/local/sonic/frrcfgd/bfdd.conf.j2,$ETC_FRR_BACKUP/bfdd.conf \
+    -t /usr/local/sonic/frrcfgd/ospfd.conf.j2,$ETC_FRR_BACKUP/ospfd.conf \
 "
 
 FRR_VARS=$(sonic-cfggen $CFGGEN_PARAMS)
@@ -57,6 +61,10 @@ if [[ ! -z "$NAMESPACE_ID" ]]; then
 fi
 
 if [ -z "$CONFIG_TYPE" ] || [ "$CONFIG_TYPE" == "separated" ]; then
+    mv $ETC_FRR_BACKUP/* /etc/frr/
+    echo "no service integrated-vtysh-config" > /etc/frr/vtysh.conf
+    rm -f /etc/frr/frr.conf
+elif [ "$CONFIG_TYPE" == "split" ]; then
     echo "no service integrated-vtysh-config" > /etc/frr/vtysh.conf
     rm -f /etc/frr/frr.conf
 elif [ "$CONFIG_TYPE" == "unified" ]; then


dockers/docker-fpm-frr/docker_init.sh
```diff
@@ -3,20 +3,24 @@
...
-    -t /usr/share/sonic/templates/gen_frr.conf.j2,/etc/frr/frr.conf \
+    -t /usr/share/sonic/templates/zebra/zebra.conf.j2,$ETC_FRR_BACKUP/zebra.conf \
+    -t /usr/share/sonic/templates/staticd/gen_staticd.conf.j2,$ETC_FRR_BACKUP/staticd.conf \
+    -t /usr/share/sonic/templates/gen_frr.conf.j2,$ETC_FRR_BACKUP/frr.conf \
     -t /usr/share/sonic/templates/isolate.j2,/usr/sbin/bgp-isolate \
     -t /usr/share/sonic/templates/unisolate.j2,/usr/sbin/bgp-unisolate \
-    -t /usr/local/sonic/frrcfgd/bfdd.conf.j2,/etc/frr/bfdd.conf \
-    -t /usr/local/sonic/frrcfgd/ospfd.conf.j2,/etc/frr/ospfd.conf \
+    -t /usr/local/sonic/frrcfgd/bfdd.conf.j2,$ETC_FRR_BACKUP/bfdd.conf \
+    -t /usr/local/sonic/frrcfgd/ospfd.conf.j2,$ETC_FRR_BACKUP/ospfd.conf \
```
shi-su, line 23: If config type is not `separated`, do these templates need to be rendered?

Pterosaur: I think they don't need. Should I refactor this script? The current modification try to keep the original code structure without too much refactor.

shi-su: I would recommend doing so since we may generate some files that are never used otherwise. It seems that creating the directory `ETC_FRR_BACKUP` is also unnecessary.

Pterosaur: Done, please review it.

dockers/docker-fpm-frr/docker_init.sh
```diff
@@ -57,9 +47,33 @@ if [[ ! -z "$NAMESPACE_ID" ]]; then
...
+        -t /usr/share/sonic/templates/zebra/zebra.conf.j2,/etc/frr/zebra.conf \
+        -t /usr/share/sonic/templates/staticd/gen_staticd.conf.j2,/etc/frr/staticd.conf \
+    "
+    MGMT_FRAMEWORK_CONFIG=$(echo $FRR_VARS | jq -r '.frr_mgmt_framework_config')
+    if [ -n "$MGMT_FRAMEWORK_CONFIG" ] && [ "$MGMT_FRAMEWORK_CONFIG" != "false" ]; then
+        CFGGEN_PARAMS="$CFGGEN_PARAMS \
+            -t /usr/local/sonic/frrcfgd/bfdd.conf.j2,/etc/frr/bfdd.conf \
+            -t /usr/local/sonic/frrcfgd/ospfd.conf.j2,/etc/frr/ospfd.conf \
+        "
+    fi
```
shi-su, line 63: The original logic is to remove the files otherwise. Wondering if the removal is mandatory? For example, if the files exist previously, is it necessary to remove them here even if it is not generated again from the template?

dockers/docker-fpm-frr/docker_init.sh
shi-su: @tahmed-dev It seems that this is a tradeoff between two options:
1. Generating some files that will never be used in some config type (render all templates anyway in one `sonic-cfggen`);
2. Introducing an additional `sonic-cfggen` call in some config type (get the config type in the first call and only render the necessary templates in the second).

I am not completely sure which one would be better. @tahmed-dev What do you think?

dockers/docker-fpm-frr/docker_init.sh
Pterosaur: Yes, good question. I'm not sure whether any effect will involve if the previous existing files are kept in the `/etc/frr/` folder to the `unified` mode. So I just want to keep the original behavior.

shi-su: I think this is somewhat different from the original behavior though. Let's assume this scenario:
1. `CONFIG_TYPE` is `separated` and `$MGMT_FRAMEWORK_CONFIG` is `false`
2. there are `bfdd.conf` and `ospfd.conf` in `etc/frr` before

The original behavior is to remove `bfdd.conf` and `ospfd.conf`;
The new behavior is to keep `bfdd.conf` and `ospfd.conf`

Pterosaur: Got it, you are right.
I try to remove it in the `else` block. Please review it.

shi-su (approved):

dockers/docker-fpm-frr/docker_init.sh
lguohan: now there are two sonic-cfggen call. it can increase the fast reboot time. @Pterosaur , can you runa sample call on the device to see how much increase the this additiona sonic-cfggen can take?

Pterosaur: Once calls
```
root@ARISTA01T1:/# time sonic-cfggen  -d     -y /etc/sonic/constants.yml     -t /usr/share/sonic/templates/frr_vars.j2     -t /usr/share/sonic/templates/supervisord/supervisord.conf.j2     -t /usr/share/sonic/templates/bgpd/gen_bgpd.conf.j2     -t /usr/share/sonic/templates/supervisord/critical_processes.j2     -t /usr/share/sonic/templates/zebra/zebra.conf.j2     -t /usr/share/sonic/templates/staticd/gen_staticd.conf.j2     -t /usr/share/sonic/templates/gen_frr.conf.j2     -t /usr/share/sonic/templates/isolate.j2     -t /usr/share/sonic/templates/unisolate.j2     -t /usr/local/sonic/frrcfgd/bfdd.conf.j2     -t /usr/local/sonic/frrcfgd/ospfd.conf.j2 > /dev/null

real    0m0.267s
user    0m0.179s
sys     0m0.051s
```

Twice call
```
root@ARISTA01T1:/# time bash -c "sonic-cfggen  -d     -y /etc/sonic/constants.yml     -t /usr/share/sonic/templates/frr_vars.j2     -t /usr/share/sonic/templates/supervisord/supervisord.conf.j2     -t /usr/share/sonic/templates/supervisord/critical_processes.j2     -t /usr/share/sonic/templates/isolate.j2     -t /usr/share/sonic/templates/unisolate.j2 && sonic-cfggen -d         -y /etc/sonic/constants.yml         -t /usr/share/sonic/templates/bgpd/gen_bgpd.conf.j2         -t /usr/share/sonic/templates/zebra/zebra.conf.j2         -t /usr/share/sonic/templates/staticd/gen_staticd.conf.j2         -t /usr/local/sonic/frrcfgd/bfdd.conf.j2         -t /usr/local/sonic/frrcfgd/ospfd.conf.j2 " > /dev/null

real    0m0.405s
user    0m0.285s
sys     0m0.087s
```


It increases about 0.2s.

lguohan: @tahmed-dev, any comments?

lguohan: haven't heard back from tamer. let's do it for now.

lguohan (approved):

assafho:
I am using split mode with 
"service integrated-vtysh-config" > /etc/frr/vtysh.conf
 rm -f /etc/frr/bgpd.conf /etc/frr/zebra.conf /etc/frr/staticd.conf

I see that /etc/sonic/frr/frr.conf is not overwritten after docker restart, however docker init with default configuration and doesn't load frr.conf. Which process is pushing default frr configuration upon startup ?

PR #7319, opened 2021-04-14
Fix the SONiC version starting with HEAD issue

xumia:
#### Why I did it
Fix the SONiC version starting with HEAD issue. For example, sonic-broadcom-HEAD.40620469-1d98d18cf.bin.

#### How I did it
Set the correct branch name after the code checked out.

#### How to verify it
After applied the change "git checkout -b 201911", the version is good.
```
$ git branch
* (HEAD detached at cf54cbc5)
$ sonic_get_version
HEAD.0-cf54cbc5
$ git checkout -b 201911
Switched to a new branch '201911'
$ sonic_get_version     
201911.0-cf54cbc5
```
#### Which release branch to backport (provide reason below if selected)

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012

PR #7320, opened 2021-04-14
[CI] Use the same template in both PR build and official build.

liushilongbuaa:
#### Why I did it
Use the same build template in build process.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012

### DIFF
--- a/.azure-pipelines/azure-pipelines-image-template.yml
+++ b/.azure-pipelines/azure-pipelines-image-template.yml
@@ -45,7 +45,9 @@ jobs:
           ENABLE_DOCKER_BASE_PULL=y make PLATFORM=$(PLATFORM_AZP) PLATFORM_ARCH=$(PLATFORM_ARCH) configure
         displayName: 'Make configure'
     postSteps:
-      - publish: $(System.DefaultWorkingDirectory)/target
+      - script: cp target -r $(Build.ArtifactStagingDirectory)/
+        displayName: Copy Artifacts
+      - publish:  $(Build.ArtifactStagingDirectory)
         artifact: 'sonic-buildimage.$(GROUP_NAME)$(GROUP_EXTNAME)'
         displayName: "Archive sonic image"
       - ${{ parameters.postSteps }}
--- a/.azure-pipelines/cleanup.yml
+++ b/.azure-pipelines/cleanup.yml
@@ -10,8 +10,5 @@ steps:
     if sudo [ -f dockerfs/var/run/docker.pid ] ; then
         pid=`sudo cat dockerfs/var/run/docker.pid` ; sudo kill $pid
     fi
-    sudo rm -rf dockerfs
-    sudo rm -rf fsroot
-    username=$(id -un)
-    sudo chown -R ${username}.${username} .
+    sudo rm -rf $(ls -A1)
   displayName: "Clean Workspace"
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -32,27 +32,18 @@ resources:
 stages:
 - stage: Build
   pool: sonicbld
-
+  variables:
+    CACHE_MODE: rcache
+    ${{ if eq(variables['Build.SourceBranchName'], '202012') }}:
+      VERSION_CONTROL_OPTIONS: 'SONIC_VERSION_CONTROL_COMPONENTS=deb,py2,py3,web'
   jobs:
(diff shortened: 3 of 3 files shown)


azure-pipelines.yml
```diff
@@ -32,27 +32,18 @@ resources:
...
-
-  - template: .azure-pipelines/build-template.yml
-    parameters:
-      platform: mellanox
-      platform_short: mlnx
-      cache_mode: rcache
-      sync_rpc_image: true
-
-  - template: .azure-pipelines/build-template.yml
+  - template: .azure-pipelines/azure-pipelines-build.yml
```
lguohan, line 40: fix this first. https://github.com/Azure/sonic-buildimage/pull/7319/files#r613663606

xumia: @liushilongbuaa @lguohan , the change looks good for both PR build and official build.

liushilongbuaa: we have checked this behavior, there is no impact on PR check.

xumia:
@liushilongbuaa , we should not cache sonic images, it only took 1 second to build vs image.
https://dev.azure.com/mssonic/build/_build/results?buildId=11463&view=logs&jobId=cef3d8a9-152e-5193-620b-567dc18af272&j=cef3d8a9-152e-5193-620b-567dc18af272&t=359769c4-8b5e-5976-a793-85da132e0a6f

.azure-pipelines/cleanup.yml
```diff
@@ -10,8 +10,5 @@ steps:
     if sudo [ -f dockerfs/var/run/docker.pid ] ; then
         pid=`sudo cat dockerfs/var/run/docker.pid` ; sudo kill $pid
     fi
-    sudo rm -rf dockerfs
-    sudo rm -rf fsroot
-    username=$(id -un)
-    sudo chown -R ${username}.${username} .
+    sudo rm -rf $(ls -A1)
```
lguohan, line 13: is this neccessary? if we delete everything, then we will having some difficult time to debug anything.

liushilongbuaa: neccessary. Without this PR build maybe only need 5min. Because target is already exist.

liushilongbuaa: we can remove clean at the end.

.azure-pipelines/azure-pipelines-image-template.yml
```diff
@@ -45,7 +45,9 @@ jobs:
           ENABLE_DOCKER_BASE_PULL=y make PLATFORM=$(PLATFORM_AZP) PLATFORM_ARCH=$(PLATFORM_ARCH) configure
         displayName: 'Make configure'
     postSteps:
-      - publish: $(System.DefaultWorkingDirectory)/target
+      - script: cp target -r $(Build.ArtifactStagingDirectory)/
```
liushilongbuaa, line 48: Keep same file tree with former build template.

.azure-pipelines/cleanup.yml
liushilongbuaa, line 13: PR build maybe only need 5mins. Because target existing.

azure-pipelines.yml
```diff
@@ -32,27 +32,18 @@ resources:
...
     parameters:
-      platform: vs
-      platform_short: vs
-      cache_mode: rcache
+      buildOptions: 'USERNAME=admin SONIC_BUILD_JOBS=$(nproc) ${{ variables.VERSION_CONTROL_OPTIONS }}'
```
qiluo-msft, line 42: >admin [](http://example.com/codeflow?start=30&length=5)

This is the default value, check `DEFAULT_USERNAME`. why you need to specify it in this PR?

.azure-pipelines/azure-pipelines-image-template.yml
qiluo-msft: What is "former build template"? I don't understand the motiviation and why it is required.

liushilongbuaa: Now the published artifacts is short of a directory 'target'. And we don't want to publish source code.

azure-pipelines.yml
liushilongbuaa: It is first used in Guohan's PR pipeline. Maybe there is some corner case, so I kept it.

lguohan (approved):

PR #7321, opened 2021-04-14
[ci] Fix no enough space issue in docker root in multi arch build

xumia:
#### Why I did it
Failed to build the centec-arm64 for no space in docker data root.

#### How I did it
Change to use the data root to the folder under /data.
See detail info about DOCKER_DATA_ROOT_FOR_MULTIARCH in the file Makefile.work.

#### How to verify it
Set the environment variable, then the /data used as docker root.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012

### DIFF
--- a/.azure-pipelines/azure-pipelines-build.yml
+++ b/.azure-pipelines/azure-pipelines-build.yml
@@ -24,6 +24,7 @@ jobs:
       PLATFORM_AZP: $(GROUP_NAME)
       PLATFORM_ARCH: amd64
       BUILD_OPTIONS: ${{ parameters.buildOptions }}
+      DOCKER_DATA_ROOT_FOR_MULTIARCH: /data/multiarch/docker
       dbg_image: false
       swi_image: false
       raw_image: false
--- a/.azure-pipelines/azure-pipelines-build.yml
+++ b/.azure-pipelines/azure-pipelines-build.yml
@@ -24,7 +24,7 @@ jobs:
       PLATFORM_AZP: $(GROUP_NAME)
       PLATFORM_ARCH: amd64
       BUILD_OPTIONS: ${{ parameters.buildOptions }}
-      DOCKER_DATA_ROOT_FOR_MULTIARCH: /data/multiarch/docker
+      DOCKER_DATA_ROOT_FOR_MULTIARCH: /data/march/docker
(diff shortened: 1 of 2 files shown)


.azure-pipelines/azure-pipelines-build.yml
```diff
@@ -24,6 +24,7 @@ jobs:
       PLATFORM_AZP: $(GROUP_NAME)
       PLATFORM_ARCH: amd64
       BUILD_OPTIONS: ${{ parameters.buildOptions }}
+      DOCKER_DATA_ROOT_FOR_MULTIARCH: /data/multiarch/docker
```
lguohan, line 27: where is this variable getting used?

lguohan: azure-pipelines-image-template.yml:          export DOCKER_DATA_ROOT_FOR_MULTIARCH=/data/march/docker

xumia: @lguohan , it is an the variable in Azure pipelines, and will be added as shell environment by azure pipelines' design. And Makefile.work will use it automatically.

lguohan: but there is already export in azure-pipeline-image-template.yml, why that does not work?

xumia: We may need to set the variable for both configuration step and build image step.

lguohan: then, can you remove the export now? also use the original variable defined in the export /data/march/docker

lguohan (approved):

.azure-pipelines/azure-pipelines-build.yml
xumia: I removed the export, and use the old value /data/march/docker as the data root.

xumia:
Login the vm, the dockerd uses the right data root.
root      138380  0.0  0.0  11168  4776 ?        SN   01:13   0:00 sudo dockerd --experimental=true --storage-driver=vfs --data-root=/data/march/docker --exec-root=/var/run/march/docker/ -H unix:///var/run/march/docker.sock -p /var/run/march/docker.pid

PR #7331, opened 2021-04-15
CISCO: Provide support for platform hooks in syncd.sh

VenkatCisco:
#### Why I did it
This is a platform agnostic & vendor agnostic change to files/scripts/syncd.sh to check if 'platform' entry exists in the DEVICE_METADATA table. If it exits it checks if there is a file to source and hence does the needful. 

#### How to verify it
This is a PR followup from the code-review discussion we had with Guhon/Rita and team. This also addresses one of the comment that was raised to convert to sourcing a file instead of executing the same., so the logic within is run in the same shell. 

#### How I did it#### Which release branch to backport (provide reason below if selected)

- [x ] 201811
- [x ] 201911
- [x] 202006
- [x] 202012

### DIFF
--- a/files/scripts/syncd.sh
+++ b/files/scripts/syncd.sh
@@ -28,6 +28,10 @@ function startplatform() {
         debug "Firmware update procedure ended"
     fi
 
+    # This is a platform agnostic approach to check if 'platform'
+    # entry exists in the DEVICE_METADATA table. If it exits then
+    # source the script to perform platform specific checks
+
     PLATFORM_DIR=`sonic-cfggen -H -v DEVICE_METADATA.localhost.platform`
     F="/usr/share/sonic/device/$PLATFORM_DIR/scripts/syncd.sh"
     [ -e $F ] && source $F start $DEV
--- a/files/scripts/syncd.sh
+++ b/files/scripts/syncd.sh
@@ -28,6 +28,10 @@ function startplatform() {
         debug "Firmware update procedure ended"
     fi
 
+    PLATFORM_DIR=`sonic-cfggen -H -v DEVICE_METADATA.localhost.platform`
+    F="/usr/share/sonic/device/$PLATFORM_DIR/scripts/syncd.sh"
+    [ -e $F ] && source $F start $DEV
+
     if [[ x"$WARM_BOOT" != x"true" ]]; then
         if [ x$sonic_asic_platform == x'cavium' ]; then
             /etc/init.d/xpnet.sh start


files/scripts/syncd.sh
```diff
@@ -28,6 +28,10 @@ function startplatform() {
         debug "Firmware update procedure ended"
     fi
 
+    PLATFORM_DIR=`sonic-cfggen -H -v DEVICE_METADATA.localhost.platform`
```
lguohan, line 31: why not use $sonic_asic_platform?

lguohan: if we want to do this one, we should inject the hook for all functions startplatform, waitplatform, stopplatform1, and stopplatform2.

lguohan: and we should also refactor the existing code to use this platform hook.

lguohan (changes requested):
as comments.

files/scripts/syncd.sh
VenkatCisco: > why not use $sonic_asic_platform?

$sonic_asic_platform does not provide platform agnostic approach to otherwise a Platform Independent SONiC code under files/scripts path. Given that this is a run time check, using DEVICE_METADATA seems more as a generic approach

VenkatCisco: > and we should also refactor the existing code to use this platform hook.

I agree with refactoring the existing code. The whole code has vendor hooks and is not maintainable. I can look into that approach. 

Can that be considered as a different exercise given we have dependencies with other vendors using this code.

VenkatCisco:
> as comments.

you mean 'add comments'. Done.

files/scripts/syncd.sh
VenkatCisco: > startplatform, waitplatform, stopplatform1, and stopplatform2.

Presently, we need platform hooks for startplatform() API only. This is an easily extendable hook for any other function. We see an opportunity to restructure files/scripts/syncd.sh to pursue the following instead of making vendor specific calls from each function: 



function startplatform () {
    F="/usr/share/sonic/device/$PLATFORM_DIR/scripts/startplatform_routines.sh"
    [ -x $F ] && $F start $DEV
}

function stopplatform() {
    F="/usr/share/sonic/device/$PLATFORM_DIR/scripts/stopplatform_routines.sh"
    [ -x $F ] && $F start $DEV
}

...
PLATFORM_DIR=`sonic-cfggen -H -v DEVICE_METADATA.localhost.platform`

case "$1" in
    start|wait|stop)
        $1
        ;;
    *)
        echo "Usage: $0 {start|wait|stop}"
        exit 1
        ;;
esac

garigipati31:
Hi Guhon, please let me know if there is anything to be addressed further.

anshuv-mfst:
PR Scrum meeting 5/4: 
Guohan: this should be asic specific and not platform specific. Please replace platform DIR with sonic_asic_platform. 
Customization is asic specific.

anshuv-mfst:
Hi @lguohan - please review and approve, thanks.

VenkatCisco:
Closing ; Tracked via https://github.com/Azure/sonic-buildimage/pull/7635/

PR #7333, opened 2021-04-15
Support no_proxy environment variable in Azure SONiC build files

VenkatCisco:
#### Why I did it
Lack of having the support for no_proxy in SONiC build environment limits enterprise companies to put some internal hacks to ensure the proxy's dont refer to some Intranet site for artifactory downloads etc. Today. Using no_proxy is familiar in proxy settings terminology and excludes traffic destined to certain hosts. 

Most Web clients hence support connection to proxy servers via environment variables:

    http_proxy / HTTP_PROXY
    https_proxy / HTTPS_PROXY
    no_proxy / NO_PROXY

These variables tell the client what URL should be used to access the proxy servers and which exceptions should be made.
 
#### How I did it
Add a new variable called 'no_proxy' where proxy settings are specified. A user may choose not to use the same. 

#### How to verify it
Simply set up the variable in the bash shell at build time. 

export no_proxy=internal.example.com, internal2.example.com

Usage is: 
no_proxy is a comma- or space-separated list of machine or domain names, with optional :port part.  If no :port
part is present, it applies to all ports on that domain.

- [ ] 201811
- [x] 201911
- [x] 202006
- [x] 202012

### DIFF
--- a/Makefile.work
+++ b/Makefile.work
@@ -198,7 +198,7 @@ DOCKER_BASE_BUILD = docker build --no-cache \
 		    -t $(SLAVE_BASE_IMAGE):$(SLAVE_BASE_TAG) \
 		    --build-arg http_proxy=$(http_proxy) \
 		    --build-arg https_proxy=$(https_proxy) \
-            --build-arg no_proxy=$(no_proxy) \
+		    --build-arg no_proxy=$(no_proxy) \
 		    $(SLAVE_DIR)
 
 DOCKER_BASE_PULL = docker pull \
--- a/Makefile.work
+++ b/Makefile.work
@@ -138,6 +138,7 @@ DOCKER_RUN := docker run --rm=true --privileged --init \
     -w $(DOCKER_BUILDER_WORKDIR) \
     -e "http_proxy=$(http_proxy)" \
     -e "https_proxy=$(https_proxy)" \
+    -e "no_proxy=$(no_proxy)" \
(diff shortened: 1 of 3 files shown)


Makefile.work
```diff
@@ -138,6 +138,7 @@ DOCKER_RUN := docker run --rm=true --privileged --init \
     -w $(DOCKER_BUILDER_WORKDIR) \
     -e "http_proxy=$(http_proxy)" \
     -e "https_proxy=$(https_proxy)" \
+    -e "no_proxy=$(no_proxy)" \
```
lguohan, line 141: what is no_proxy? do you have a reference?

VenkatCisco: Please refer to the two links: 
https://www.gnu.org/software/emacs/manual/html_node/url/Proxies.html   

& https://about.gitlab.com/blog/2021/01/27/we-need-to-talk-no-proxy/

lguohan:
can you check build failure and sign cla?

VenkatCisco:
> can you check build failure and sign cla?

Ack, I believe there may be new CLA for a new PR ? I am working with Cisco to get confirmation before i sign the CLA once again.

lguohan (approved):

lguohan:
better to remove CISCO in the pr title. it is not really related to cisco.

VenkatCisco:
> better to remove CISCO in the pr title. it is not really related to cisco.

Sure

PR #7344, opened 2021-04-15
[build] Extend rules/config.user to more Makefiles

bluecmd:
rules/config.user allows overriding default properties without
touching tracked files. This change makes sure all properties
can be set and not just the ones used in slave.mk.

#### Why I did it

I want to set `SONIC_CONFIG_USE_NATIVE_DOCKERD_FOR_BUILD` which is read by `Makefile.work`.

#### How I did it

I extended the existing functionality introduced in https://github.com/Azure/sonic-buildimage/pull/5325.

#### How to verify it

```
echo 'SONIC_CONFIG_USE_NATIVE_DOCKERD_FOR_BUILD = y` > rules/config.user
make [some target]
ps ax | grep 'docker.sock' --color
# Verify that the docker container running has  -v /var/run/docker.sock:/var/run/docker.sock set.
```

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012

Extending functionality of `rules/config.user` to include all build options.

image

### DIFF
--- a/Makefile.work
+++ b/Makefile.work
@@ -91,6 +91,7 @@ SLAVE_DIR = sonic-slave-jessie
 endif
 
 include rules/config
+-include rules/config.user
 
 ifeq ($(ENABLE_DOCKER_BASE_PULL),)
 	override ENABLE_DOCKER_BASE_PULL = n


bluecmd:
@Staphylo: You did the initial addition of the config.user, what do you say about this addition?

bluecmd:
Not sure why `Azure.sonic-buildimage (Test kvmtest-t0) Failing after 53m — Test kvmtest-t0 failed` is happening. Hints welcome!

Staphylo:
That change looks good to me. I actually missed that `include rules/config` reference in `Makefile.work` so thanks for the fix!
It was indeed the intention to include the user config everywhere the base config is included.

Regarding the test runs, you'll have to wait for a project owner to restart the failed jobs.
I'm not sure if the job re-trigger knobs via comments are back with the azure pipelines.

lguohan (approved):

lguohan:
now there is command like "/azpw run"

PR #7352, opened 2021-04-16
Provide support for platform hooks in swss.sh

VenkatCisco:
#### Why I did it
This is a platform agnostic & vendor agnostic change to files/scripts/swss.sh to check if 'platform' entry exists in the DEVICE_METADATA table. If it exits it checks if there is a file to source and hence does the needful.

#### How I did it
This is a PR followup from the code-review discussion we had with Guhon/Rita and team. This also addresses one of the comment that was raised to convert to sourcing a file instead of executing the same., so the logic within is run in the same shell.

#### How to verify it
Verify on the device to ensure any files sourced via the swss.sh hooks are getting executed. 

- [ ] 201811
- [x] 201911
- [x] 202006
- [x] 202012

### DIFF
--- a/files/scripts/swss.sh
+++ b/files/scripts/swss.sh
@@ -149,6 +149,13 @@ start() {
         clean_up_tables STATE_DB "'PORT_TABLE*', 'MGMT_PORT_TABLE*', 'VLAN_TABLE*', 'VLAN_MEMBER_TABLE*', 'LAG_TABLE*', 'LAG_MEMBER_TABLE*', 'INTERFACE_TABLE*', 'MIRROR_SESSION*', 'VRF_TABLE*', 'FDB_TABLE*', 'FG_ROUTE_TABLE*', 'BUFFER_POOL*', 'BUFFER_PROFILE*', '*MUX_CABLE_TABLE*'"
     fi
 
+    # This is a platform agnostic approach to check if 'platform'
+    # entry exists in the DEVICE_METADATA table. If it exits then
+    # source the script to perform platform specific checks
+    PLATFORM_DIR=`sonic-cfggen -H -v DEVICE_METADATA.localhost.platform`
+    F="/usr/share/sonic/device/$PLATFORM_DIR/scripts/swss.sh"
+    [ -e $F ] && source $F start $DEV
+
     # start service docker
     /usr/bin/${SERVICE}.sh start $DEV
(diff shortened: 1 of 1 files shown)


files/scripts/swss.sh
```diff
@@ -149,6 +149,13 @@ start() {
         clean_up_tables STATE_DB "'PORT_TABLE*', 'MGMT_PORT_TABLE*', 'VLAN_TABLE*', 'VLAN_MEMBER_TABLE*', 'LAG_TABLE*', 'LAG_MEMBER_TABLE*', 'INTERFACE_TABLE*', 'MIRROR_SESSION*', 'VRF_TABLE*', 'FDB_TABLE*', 'FG_ROUTE_TABLE*', 'BUFFER_POOL*', 'BUFFER_PROFILE*', '*MUX_CABLE_TABLE*'"
     fi
 
+    # This is a platform agnostic approach to check if 'platform'
+    # entry exists in the DEVICE_METADATA table. If it exits then
+    # source the script to perform platform specific checks
+    PLATFORM_DIR=`sonic-cfggen -H -v DEVICE_METADATA.localhost.platform`
+    F="/usr/share/sonic/device/$PLATFORM_DIR/scripts/swss.sh"
```
judyjoseph, line 156: Does the platform script need to handle stop/wait cases as well ? Another suggestion would be to pass additional parameters like warmboot/fastboot flags, network namespace etc to this script..

judyjoseph:
@VenkatCisco, I have merged a PR https://github.com/Azure/sonic-buildimage/pull/7513 from 201911. This was how we used to parse/extract the PCI-id's from the acis.conf file and populate in DEVICE_METADATA in case of multi-asic platforms.

VenkatCisco:
> @VenkatCisco, I have merged a PR #7513 from 201911. This was how we used to parse/extract the PCI-id's from the acis.conf file and populate in DEVICE_METADATA in case of multi-asic platforms.

HI Judy, as per our earlier review, Guhon & you suggested to use $sonic_asic_platform instead of $PLATFORM_DIR. Accordingly, I have already incorporated those changes in our cisco branch. The diffs are 2 lines as below. 

 F="/usr/share/sonic/device/$sonic_asic_platform/swss.sh"
 [ -e $F ] && source $F start $DEV

VenkatCisco:
tracked via #7621.

PR #7356, opened 2021-04-16
[kubenetes] Kubernetes server configurable using URL

renukamanavalan:
1) Dropped non-required IP update in admin.conf, as all masters use VIP only (#7288)
2) Don't clear VERSION during stop, as it would overwrite new version pending to go.
3) subprocess, get return value from proc and do not imply with presence of data in stderr.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012


isabelmsft (approved):

lguohan:
why not to master and then cherry-pick to 202012?

renukamanavalan:
It is already in master.  PR #7288

renukamanavalan:
PR #7288 already picked into 20212.

PR #7387, opened 2021-04-21
[teamd/patch] do not send LACP PDU if link is down

stepanblyschak:
Otherwise the lacp_port_set_state() returns error when called from
lacp_port_link_update() and lead to not up-to-date cache in
lacp_port->__link_last. This can happen when member flapping happens.
When this issue happens the condition comparing lacp_port->__link_last
to current linkup in lacp_port_link_update() will not pass and member
will never get out of deselected state unless it is flapped again.

This is done as a fix in patch maintained by SONiC but planned to be
upstreamed in libteam as well. Currently we cannot simply propose this
fix in libteam and use libteam master as of important bug fix revert in
jpirko/libteam 61efd6de2fbb8ee077863ee5a355ac3dfd9365b9.

#### Why I did it

Fix an issue that can be reproduced by doing (Ethernet0 is LAG member):
```
config interface shutdown Ethernet0; config interface startup Ethernet0
```

After this operation (probability ~1/3) Ethernet0 will be in Deselected state forever.

#### How I did it
Described in commit message.

#### How to verify it
By trying to reproduce it ~20 times with no repro.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012

### DIFF
--- a/src/libteam/patch/0007-Send-LACP-PDU-immediately-if-our-state-changed.patch
+++ b/src/libteam/patch/0007-Send-LACP-PDU-immediately-if-our-state-changed.patch
@@ -1,27 +1,41 @@
-From 780cc967a5544d3ab6f43cba4076b5e0dbb48395 Mon Sep 17 00:00:00 2001
+From 9a6f63427e843683234a8ea21435b9c69a7a418c Mon Sep 17 00:00:00 2001
 From: Pavel Shirshov <pavelsh@microsoft.com>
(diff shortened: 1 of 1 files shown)


lguohan:
is this upstreamed?

liat-grozovik:
> is this upstreamed?

not yet. Please don't merge yet. There is still work in progress on it.

PR #7399, opened 2021-04-22
[ci] Fix the boolean value case sensitive issue in Azure Pipelines

xumia:
#### Why I did it
Fix the boolean value case sensitive issue in Azure Pipelines

When passing parameters to a template, the "true" or "false" will have case sensitive issue, it should be a type casting issue.
To fix it, we change the true/false to yes/no, to escape the trap.

Support to override the job groups in the template, so PR build has chance to use different build parameters, only build simple targets. For example, for broadcom, we only build target/sonic-broadcom.bin, the other images, such as swi, debug bin, etc, will not be built.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012

### DIFF
--- a/.azure-pipelines/azure-pipelines-build.yml
+++ b/.azure-pipelines/azure-pipelines-build.yml
@@ -25,29 +25,29 @@ jobs:
       PLATFORM_ARCH: amd64
       BUILD_OPTIONS: ${{ parameters.buildOptions }}
       DOCKER_DATA_ROOT_FOR_MULTIARCH: /data/march/docker
-      dbg_image: false
-      swi_image: false
-      raw_image: false
-      sync_rpc_image: false
-      platform_rpc: false
+      dbg_image: no
+      swi_image: no
+      raw_image: no
+      sync_rpc_image: no
+      platform_rpc: no
     jobGroups:
       - name: vs
         variables:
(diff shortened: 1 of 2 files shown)


liushilongbuaa (approved):

lguohan:
how do you differentiate if this is a pr build v.s. nightly build?

xumia:
> how do you differentiate if this is a pr build v.s. nightly build?

For PR build, use the default options, all options, like dbg_image, swi_image, raw_image, etc, are use the default value “no”. It is the old behavior, keep not changed. We can change options if needed.
For Official build, we do not override the setting, so use the options defined in azure-pipelines-build.yml

Take broadcom as an example.
For Official build, the options are as below:
- name: broadcom
  variables:
    dbg_image: yes
    swi_image: yes
    raw_image: yes
    sync_rpc_image: yes
platform_rpc: brcm

For PR build, the we do not set any options:
- name: broadcom
  //variables: -- we do not set the variables here, so use the default options.

So the default options will be used, as below:
Default options:
  dbg_image: no
  swi_image: no
  raw_image: no
  sync_rpc_image: no
  platform_rpc: no

lguohan (approved):
:shipit:

PR #7446, opened 2021-04-27
[Mellanox] Bug fix: exception occurred during chassis object being destroyed

stephenxs:
#### Why I did it
The following error message is observed during chassis object being destroyed
```
Exception ignored in: <function Chassis.__del__ at 0x7fd22165cd08>
  File "/usr/local/lib/python3.7/dist-packages/sonic_platform/chassis.py", line 83, in __del__
ImportError: sys.meta_path is None, Python is likely shutting down
```

The chassis tries to import deinitialize_sdk_handle during being destroyed for the purpose of releasing the sdk_handle.
However, importing another module during shutting down can cause the error because some of the fundamental infrastructures are no longer available.

This error occurs when a chassis object is created and then destroyed in the Python shell.

#### How I did it
To fix it, record the deinitialize_sdk_handle in the chassis object when sdk_handle is being initialized and call the deinitialize handler when the chassis object is being destroyed

#### How to verify it
Manually test.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012

### DIFF
--- a/platform/mellanox/mlnx-platform-api/sonic_platform/chassis.py
+++ b/platform/mellanox/mlnx-platform-api/sonic_platform/chassis.py
@@ -73,6 +73,7 @@ class Chassis(ChassisBase):
         self.sfp_event_initialized = False
         self.reboot_cause_initialized = False
         self.sdk_handle = None
+        self.deinitialize_sdk_handle = None
         logger.log_info("Chassis loaded successfully")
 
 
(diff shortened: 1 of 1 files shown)


liat-grozovik (approved):

keboliu (approved):

liat-grozovik:
@stephenxs not a blocker for this PR but i think it is better to add a test case to be sure it will not raise again (unit test or sonic-mgmt test case).
Once you have it please refer it to the PR.

PR #7461, opened 2021-04-28
[pbh]: Add YANG model

nazariig:
#### Why I did it
* Done as part of PBH development

#### How I did it
* Added PBH YANG model

#### How to verify it
1. Add PBH configuration
2. Run DPB

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012

* N/A

```
      .---.        .-----------
     /     \  __  /    ------
    / /     \(  )/    -----
   //////   ' \/ `   ---
  //// / // :    : ---
 // /   /  /`    '--
//          //..\\
       ====UU====UU====
           '//||\\`
             ''``
```


src/sonic-yang-models/yang-models/sonic-pbh.yang
```diff
@@ -0,0 +1,218 @@
...
+
+				key "hash_name";
+
+				leaf hash_name {
+					description "The name of this hash";
+					type string;
+				}
+
+				leaf hash_field {
+					description "Configures native hash filed for this hash";
```
rathnasabapathyv, line 42: Typo: field

src/sonic-yang-models/yang-models/sonic-pbh.yang
```diff
@@ -0,0 +1,218 @@
...
+				leaf priority {
+					description "Configures priority for this rule";
+					mandatory true;
+					type uint32;
+				}
+
+				leaf gre_key {
+					description "Configures packet match for this rule: GRE key";
+					type string {
+						pattern "(0x){1}[a-fA-F0-9]+/(0x){1}[a-fA-F0-9]+";
```
rathnasabapathyv, line 103: Don't we have to escape the "/" using "\", in this regex pattern to accept an input like this (0xa124/0x1b)?
Asking something like this with "\" escape char in "pattern "(0x){1}[a-fA-F0-9]+\/(0x){1}[a-fA-F0-9]+"";

src/sonic-yang-models/yang-models/sonic-pbh.yang
```diff
@@ -0,0 +1,218 @@
...
+						}
+					}
+					default "";
+				}
+
+				leaf description {
+					description "The description of this table";
+					mandatory true;
+					type string {
+						length 0..255;
```
venkatmahalingam, line 207: What does this field contain when len is 0?

src/sonic-yang-models/yang-models/sonic-pbh.yang
```diff
@@ -0,0 +1,218 @@
...
+
+				key "table_name";
+
+				leaf table_name {
+					description "The name of this table";
+					type string;
+				}
+
+				leaf-list port_list {
+					description "Ports to which this table is applied";
```
venkatmahalingam, line 178: Is it possible to have a single list for both PORT & PORTCHANNEL tables? also, why do we need default "" value? cant we  just avoid this field configuration itself?

venkatmahalingam:
Please add UT test cases.

venkatmahalingam:
Looks like this PR is work-in-progress(Draft), please change the Draft status for review to begin.

src/sonic-yang-models/yang-models/sonic-pbh.yang
nazariig: @venkatmahalingam empty string. It's by design.

src/sonic-yang-models/yang-models/sonic-pbh.yang
nazariig: @rathnasabapathyv No, we don't. This was tested with DPB. No issues observed.

src/sonic-yang-models/yang-models/sonic-pbh.yang
nazariig: @rathnasabapathyv fixed

src/sonic-yang-models/yang-models/sonic-pbh.yang
nazariig: @venkatmahalingam Yes, it is possible. The current implementation is by design. The default "" value means that table can be created without ports/lags. The configuration can be added later via update.

venkatmahalingam: This will confuse the user, we could very well not have the field itself.

src/sonic-yang-models/yang-models/sonic-pbh.yang
venkatmahalingam: not sure how the empty string helps in the description, can we make this field optional?

src/sonic-yang-models/yang-models/sonic-pbh.yang
```diff
@@ -0,0 +1,218 @@
...
+					description "Configures packet match for this rule: GRE key";
+					type string {
+						pattern "(0x){1}[a-fA-F0-9]+/(0x){1}[a-fA-F0-9]+";
+					}
+				}
+
+				leaf ip_protocol {
+					description "Configures packet match for this rule: IP protocol";
+					type string {
+						pattern "(0x){1}[a-fA-F0-9]+/(0x){1}[a-fA-F0-9]+";
```
venkatmahalingam, line 110: For a YANG user, configuring IP protocol like this format looks weird, if you could do the translation that SAI demands internally would be better, this comment is applicable for ipv6_next_header/l4_dst_port/inner_ether_type patterns below.

src/sonic-yang-models/yang-models/sonic-pbh.yang
```diff
@@ -0,0 +1,218 @@
...
+					type string {
+						pattern "SET_ECMP_HASH|SET_LAG_HASH";
+					}
+					default "SET_ECMP_HASH";
+				}
+
+				leaf flow_counter {
+					description "Enables/Disables packet/byte counter for this rule";
+					type string {
+						pattern "DISABLED|ENABLED";
```
venkatmahalingam, line 154: The type definition for this (DISABLED/ENABLED) can go to "sonic-types.yang" file

src/sonic-yang-models/yang-models/sonic-pbh.yang
nazariig: @venkatmahalingam this is by design. We don't want to limit user in configuration flows and expose what SAI provides. Any reason not to do that?

src/sonic-yang-models/yang-models/sonic-pbh.yang
nazariig: @venkatmahalingam we can do it once there will be another schema which requires it

src/sonic-yang-models/yang-models/sonic-pbh.yang
nazariig: @venkatmahalingam ok. We can merge.

src/sonic-yang-models/yang-models/sonic-pbh.yang
nazariig: @venkatmahalingam ok. We can make it optional.

nazariig:
> 
> 
> Please add UT test cases.

@venkatmahalingam done

src/sonic-yang-models/yang-models/sonic-pbh.yang
venkatmahalingam: I suggest to follow "Keyword" names mentioned in the below link for IP protocol names here instead of exposing SAI conventions, we already have pull request (https://github.com/Azure/sonic-buildimage/pull/7051/files) pending for ip_protocol_type in sonic-types.yang file.
https://www.iana.org/assignments/protocol-numbers/protocol-numbers.xhtml

nazariig: @venkatmahalingam let's discuss it on YANG subgroup meeting.

nazariig:
https://github.com/Azure/sonic-buildimage/issues/7733

src/sonic-yang-models/yang-models/sonic-pbh.yang
```diff
@@ -0,0 +1,250 @@
...
+				leaf hash_field {
+					description "Configures native hash field for this hash field";
+					mandatory true;
+					type pbh:hash-field;
+				}
+
+				leaf ip_mask {
+					description "Configures IPv4/IPv6 address mask for this hash field";
+					mandatory true;
+					type union {
```
nazariig, line 79: check ietf for ip addr

src/sonic-yang-models/yang-models/sonic-pbh.yang
```diff
@@ -0,0 +1,250 @@
...
+				}
+
+				leaf gre_key {
+					description "Configures packet match for this rule: GRE key (value/mask)";
+					type string {
+						pattern "(0x){1}[a-fA-F0-9]+/(0x){1}[a-fA-F0-9]+";
+					}
+				}
+
+				leaf ip_protocol {
```
praveen-li, line 156: We need to look at the possibility to remove the mask from next 4 fields....Please check with Anish. Thx.

src/sonic-yang-models/yang-models/sonic-pbh.yang
```diff
@@ -0,0 +1,250 @@
...
+					description "Configures packet match for this rule: IP protocol (value/mask)";
+					type string {
+						pattern "(0x){1}[a-fA-F0-9]+/(0x){1}[a-fA-F0-9]+";
+					}
+				}
+
+				leaf ipv6_next_header {
+					description "Configures packet match for this rule: IPv6 Next header (value/mask)";
```
nazariig, line 164: put reference IANA format

src/sonic-yang-models/yang-models/sonic-pbh.yang
```diff
@@ -0,0 +1,250 @@
...
+				}
+
+				leaf ip_mask {
+					description "Configures IPv4/IPv6 address mask for this hash field";
+					mandatory true;
+					type union {
+						type inet:ipv4-address;
+						type inet:ipv6-address;
+					}
+					when "boolean(../hash_field[.='INNER_DST_IPV4' or .='INNER_SRC_IPV4' or .='INNER_DST_IPV6' or .='INNER_SRC_IPV6'])";
```
praveen-li, line 83: Plz add a test case for this when condition.

src/sonic-yang-models/yang-models/sonic-pbh.yang
```diff
@@ -0,0 +1,250 @@
...
+
+				leaf hash_name {
+					description "The name of this hash";
+					type string;
+				}
+
+				leaf-list hash_field_list {
+					description "The list of hash fields to apply with this hash";
+					min-elements 1;
+					type leafref {
```
praveen-li, line 113: Plz add a Test case to fail it, if no element is defined.

src/sonic-yang-models/yang-models/sonic-pbh.yang
```diff
@@ -0,0 +1,250 @@
...
+
+	import sonic-port {
+		prefix port;
+	}
+
+	import sonic-portchannel {
+		prefix lag;
+	}
+
+	description "PBH YANG Module for SONiC OS";
```
nazariig, line 20: VxLAN & NVGRE with inner frames

praveen-li (commented):
Added comments for few more tests.

src/sonic-yang-models/yang-models/sonic-pbh.yang
nazariig: @praveen-li with the `union` type, there will be no way to have additional validation per `hash_field`, e.g. `ip_mask` can be `IPv4` and `hash_field` is `INNER_DST_IPV6` which will be invalid configuration. Not sure if we can handle it from `xpath` perspective.

nazariig:
@anshuv-mfst the review comments are handled

nazariig:
@anish-n please review & merge

src/sonic-yang-models/yang-models/sonic-pbh.yang
```diff
@@ -0,0 +1,264 @@
...
+
+				leaf hash_field {
+					description "Configures native hash field for this hash field";
+					mandatory true;
+					type pbh:hash-field;
+				}
+
+				leaf ip_mask {
+					description "Configures IPv4/IPv6 address mask for this hash field";
+					mandatory true;
```
anish-n, line 80: Is mandatory "true" correct for this? eg: its not required for inner l4 src/dst ports and ip proto.

nazariig: @anish-n it's perfectly valid. Please refer to UTs. The trick here in using 'when' clause. Mandatory constraint takes place in case 'when' clause is satisfied.

anish-n: I see, thanks for explaining.

anish-n (approved):
High level looks good to me, I am not very familiar with Yang models, so I would recommend merging only after someone with Yang model expertise has also reviewed and signed off.

nazariig:
@praveen-li can you please approve?

src/sonic-yang-models/yang-models/sonic-pbh.yang
nazariig: @anish-n you're welcome.

nazariig:
@praveen-li just a kind reminder

nazariig:
@prsunny please help to merge

prsunny (approved):

PR #7469, opened 2021-04-28
[Kubernetes]: The kube server could be used as http-proxy for docker

renukamanavalan:
#### Why I did it
The SONiC switches get their docker images from local repo, populated during install with container images pre-built into SONiC FW. With the introduction of kubernetes, new docker images available in remote repo could be deployed. This requires dockerd to be able to pull images from remote repo. 

Depending on the Switch network domain & config, it may or may not be able to reach the remote repo. In the case where remote repo is unreachable,  we could potentially make Kubernetes server to also act as http-proxy.

#### How I did it
When admin explicitly enables, the kubernetes-server could be configured as docker-proxy. But any update to docker-proxy has to be via service-conf file environment variable, implying a  "service restart docker" is required. But restart of dockerd is vey expensive, as it would restarts all dockers, including database docker. 

To avoid dockerd restart, pre-configure an http_proxy using an unused IP. When k8s server is enabled to act as http-proxy, an IP table entry would be created to direct all traffic to the configured-unused-proxy-ip to the kubernetes-master IP. This way any update to Kubernetes master config would be just manipulating IPTables, which will be transparent to all modules, until dockerd needs to download from remote repo.

#### How to verify it
Configure a switch such that image repo is unreachable
Pre-configure dockerd with http_proxy.conf using an unused IP (e.g. 172.16.1.1)
Update ctrmgrd.service to invoke ctrmgrd.py with "-p" option.
Configure a k8s server, and deploy an image for feature with set_owner="kube"
Check if switch could successfully download the image or not.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ x] 202012


src/sonic-ctrmgrd/ctrmgr/ctrmgr_iptables.py
```diff
@@ -0,0 +1,136 @@
...
+        with open(DST_FILE, "r") as s:
+            for line in s.readlines():
+                url_match = re.search('^Environment=.HTTP_PROXY=(.+?)"', line)
+                if url_match:
+                    url = urlparse(url_match.group(1))
+                    DST_IP = _get_ip(url.hostname)
+                    DST_PORT = url.port
+                    break
+    else:
+        print("{} not availabl".format(DST_FILE))
```
isabelmsft, line 64: *available

src/sonic-ctrmgrd/ctrmgr/ctrmgr_iptables.py
```diff
@@ -0,0 +1,136 @@
...
+import re
+import socket
+import subprocess
+import syslog
+
+UNIT_TESTING = 0
+
+# NOTE:
+# Unable to use python-iptables as that does not create rules per ip-tables default
+# which is nf_tables. So rules adde via iptc package will not be listed under
```
isabelmsft, line 14: *added

src/sonic-ctrmgrd/ctrmgr/ctrmgr_iptables.py
```diff
@@ -0,0 +1,136 @@
...
+# "sudo iptables -t nat -L -n". But available in kernel. To list, we need to 
+# use legacy mode as "sudo iptables-legacy -t nat -L -n".
+# As we can't use two modes and using non-default could make any debugging effort
+# very tough.
+
+
+from urllib.parse import urlparse
+
+TABLE_NAME = "nat"
```
isabelmsft, line 23: unused

src/sonic-ctrmgrd/tests/ctrmgr_iptables_test.py
```diff
@@ -0,0 +1,179 @@
...
+import re
+import sys
+from unittest.mock import MagicMock, patch
+
+import pytest
+
+from . import common_test
+
+sys.path.append("ctrmgr")
+# import ctrmgr.ctrmgr_iptables
```
isabelmsft, line 11: leftover comment?

src/sonic-ctrmgrd/ctrmgr/ctrmgr_iptables.py
```diff
@@ -0,0 +1,136 @@
...
+    else:
+        return ""
+
+
+def check_proc(proc):
+    if proc.returncode:
+        syslog.syslog(syslog.LOG_ERR, "Failed to run: cmd: {}".format(proc.args))
+        syslog.syslog(syslog.LOG_ERR, "Failed to run: stdout: {}".format(proc.stdout))
+        syslog.syslog(syslog.LOG_ERR, "Failed to run: stderr: {}".format(proc.stderr))
+        if not UNIT_TESTING:
```
isabelmsft, line 86: Doesn't it only enter here if proc.returncode != 0? Do you mean to intent this?

src/sonic-ctrmgrd/ctrmgr/ctrmgr_iptables.py
renukamanavalan: Dropped.

src/sonic-ctrmgrd/ctrmgr/ctrmgr_iptables.py
renukamanavalan: done

src/sonic-ctrmgrd/ctrmgr/ctrmgr_iptables.py
renukamanavalan: I want to assert, but not when in the middle of unit testing

src/sonic-ctrmgrd/tests/ctrmgr_iptables_test.py
renukamanavalan: removed

src/sonic-ctrmgrd/ctrmgr/ctrmgr_iptables.py
```diff
@@ -0,0 +1,135 @@
...
+
+        cmd = "sudo iptables -t nat -n -L OUTPUT {}".format(num)
+        proc = subprocess.run(cmd, shell=True, capture_output=True)
+        check_proc(proc)
+
+        if not proc.stdout:
+            # No more rule
+            break
+
+        rule_dest = _is_rule_match(proc.stdout.decode("utf-8").strip())
```
isabelmsft, line 116: build error:
>           rule_dest = _is_rule_match(proc.stdout.decode("utf-8").strip())
E           AttributeError: 'str' object has no attribute 'decode'

renukamanavalan: Resolved

isabelmsft (approved):

src/sonic-ctrmgrd/ctrmgr/ctrmgr_iptables.py
```diff
@@ -0,0 +1,135 @@
+#!/usr/bin/env python3
```
lguohan, line 1: can you check the hostcfgd, there are nat rules there, do we have conflict there?

lguohan: i suggest to centralize the control plane acl into one place. hostcfgd.

renukamanavalan:
Looking at a different way of meeting this requirement. Hence closing.

src/sonic-ctrmgrd/ctrmgr/ctrmgr_iptables.py
renukamanavalan: Checked this in-depth with @prsunny. 

Currently, hostcfgd touches iptables for “mangle” only, whereas we need to touch nat table. So we can't use hostcfgd.

The cacld manages nat table based on ACL_TABLE & ACL_RULE. But we can't add our rules to ACL_TABLE/RULE, as it is tightly controlled by graph data and any discrepancy will remove the rules that is non-existent in graph.

Per discussion, it is suggested that we keep our IP table update outside cacld too. In other words, let ctrmgrd do that.

NOTE: Each cacld restart, clears iptable "nat" and re-create based on ACL_TABLE & RULE. So to retain our rules, ctrmgrd is made dependent on cacld and start after cacld, and restart for every cacld restart.

isabelmsft (dismissed):

isabelmsft (approved):

PR #7474, opened 2021-04-29
[container_checker] Use Feature table to get running containers 

renukamanavalan:
#### Why I did it
Finding running containers through "docker ps" breaks when kubernetes deploys container, as the names are mangled.

#### How I did it
The data is is available from FEATURE table, which takes care of kubernetes deployment too.

#### How to verify it
Deploy a feature via kubernetes and don't expect error from container_check.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012

### DIFF
--- a/files/image_config/monit/container_checker
+++ b/files/image_config/monit/container_checker
@@ -16,33 +16,12 @@ check program container_checker with path "/usr/bin/container_checker"
     if status != 0 for 5 times within 5 cycles then alert repeat every 1 cycles
 """
 
-import subprocess
+import docker
 import sys
 
 import swsssdk
 from sonic_py_common import multi_asic
-
-
-def get_command_result(command):
-    """
-    @summary: This function will execute the command and return the resulting output.
-    @return: A string which contains the output of command.
-    """
-    command_stdout = ""
-
-    try:
-        proc_instance = subprocess.Popen(command, stdout=subprocess.PIPE, stderr=subprocess.PIPE,
-                                         shell=True, universal_newlines=True)
-        command_stdout, command_stderr = proc_instance.communicate()
-        if proc_instance.returncode != 0:
-            print("Failed to execute the command '{}'. Return code: '{}'".format(
-                command, proc_instance.returncode))
-            sys.exit(1)
-    except (OSError, ValueError) as err:
-        print("Failed to execute the command '{}'. Error: '{}'".format(command, err))
-        sys.exit(2)
-
-    return command_stdout.rstrip().split("\n")
+from swsscommon import swsscommon
 
 
 def get_expected_running_containers():
@@ -60,6 +39,7 @@ def get_expected_running_containers():
     feature_table = config_db.get_table("FEATURE")
 
     expected_running_containers = set()
+    always_running_containers = set()
 
     for container_name in feature_table.keys():
         if feature_table[container_name]["state"] not in ["disabled", "always_disabled"]:
@@ -70,10 +50,12 @@ def get_expected_running_containers():
                     num_asics = multi_asic.get_num_asics()
                     for asic_id in range(num_asics):
                         expected_running_containers.add(container_name + str(asic_id))
+            elif feature_table[container_name]["state"] == 'always_enabled':
+                always_running_containers.add(container_name)
             else:
                 expected_running_containers.add(container_name)
 
-    return expected_running_containers
+    return expected_running_containers, always_running_containers
 
 
 def get_current_running_containers():
@@ -84,12 +66,27 @@ def get_current_running_containers():
     """
     running_containers = set()
 
-    command = "docker ps"
-    command_stdout = get_command_result(command)
-    for line in command_stdout[1:]:
-        running_containers.add(line.split()[-1].strip())
+    state_db = swsscommon.DBConnector("STATE_DB", 0)
+    tbl = swsscommon.Table(state_db, "FEATURE")
+
+    for name in tbl.getKeys():
+        data = dict(tbl.get(name)[1])
+        if data.get('container_id'):
+            running_containers.add(name)
 
     return running_containers
+
+
+def check_dockers(always_running_containers):
+    DOCKER_CLIENT = docker.DockerClient(base_url='unix://var/run/docker.sock')
+    RUNNING = 'running'
+    not_running = set()
+    for name in always_running_containers:
+        container = DOCKER_CLIENT.containers.get(name)
+        container_state = container.attrs['State']
+        if container_state['Status'] != RUNNING:
+            not_running.add(name)
+    return not_running
 
 
 def main():
@@ -98,10 +95,11 @@ def main():
               and the containers which were expected to run. If containers which were exepcted
               to run were not running, then an alerting message will be written into syslog.
     """
-    expected_running_containers = get_expected_running_containers()
+    expected_running_containers, always_running_containers = get_expected_running_containers()
     current_running_containers = get_current_running_containers()
 
     not_running_containers = expected_running_containers.difference(current_running_containers)
+    not_running_containers |= check_dockers(always_running_containers)
     if not_running_containers:
         print("Expected containers not running: " + ", ".join(not_running_containers))
         sys.exit(3)


files/image_config/monit/container_checker
```diff
@@ -84,24 +66,40 @@ def get_current_running_containers():
     """
```
yozhao101, line 66: Since the change will not use `docker ps` to get the running container list, can you reorganize the comments of this function please?

files/image_config/monit/container_checker
```diff
@@ -88,14 +98,12 @@ def check_dockers(always_running_containers):
               in running state.
```
isabelmsft, line 98: Comment says returns containers that are *not* in running state, but code was updated to return running containers

renukamanavalan: Thanks. Corrected.

isabelmsft (approved):

files/image_config/monit/container_checker
```diff
@@ -70,25 +51,60 @@ def get_expected_running_containers():
...
-    command = "docker ps"
-    command_stdout = get_command_result(command)
-    for line in command_stdout[1:]:
-        running_containers.add(line.split()[-1].strip())
+    state_db = swsscommon.DBConnector("STATE_DB", 0)
+    tbl = swsscommon.Table(state_db, "FEATURE")
+    if not tbl.getKeys():
+        return False, None
+
+    for name in tbl.getKeys():
```
yozhao101, line 79: One question: whether the `container_name` in `always_running_containers` is included in `tbl.getKeys()`?

renukamanavalan: rest, except "database".

yozhao101: That's why we still need the code from line 84 to 90, right?

renukamanavalan: yes. 
I could do some optimizations, but realized not worth it.
In fact being generic - helps handle any newer scenarios too, transparently.

files/image_config/monit/container_checker
```diff
@@ -16,50 +16,31 @@ check program container_checker with path "/usr/bin/container_checker"
...
-    @summary: This function will get the expected running containers by following the rule:
+    @summary: This function will get the expected running & always-enabled containers by following the rule:
               The 'state' field of container in 'FEATURE' table should not be 'disabled' or 'always_disabled'.
               If the device has Multi-ASIC, this function will get container list by determining the
               value of field 'has_global_scope', the number of ASICs and the value of field
               'has_per_asic_scope'.
               If the device has single ASIC, the container name was put into the list.
-    @return:  A set which contains the expected running containers.
+    @return:  A set which contains the expected running containers and a set that has 
+              containers marked as "always_enabled"
```
yozhao101, line 36: It would be better to add a punctuation mark `.` at the end of the sentence.

files/image_config/monit/container_checker
```diff
@@ -70,25 +51,60 @@ def get_expected_running_containers():
...
-    @summary: This function will get the current running container list by analyzing the
-              output of command `docker ps`.
-    @return:  A set which contains the current running contianers.
+    @summary: This function will get the current running container list
+              from FEATURE table @ STATE_DB, if this table is available.
+    @return:  a tuple
+              First: Return value indicating if info can be obtained from
+              DB or not.
+              Second: A set which contains the current running containers,
+              if this info is available in DB
```
yozhao101, line 70: It would be better to add a punctuation mark `.` at the end of the sentence.

files/image_config/monit/container_checker
```diff
@@ -98,9 +114,12 @@ def main():
               and the containers which were expected to run. If containers which were exepcted
               to run were not running, then an alerting message will be written into syslog.
     """
-    expected_running_containers = get_expected_running_containers()
-    current_running_containers = get_current_running_containers()
+    expected_running_containers, always_running_containers = get_expected_running_containers()
+    ret, current_running_containers = get_current_running_from_DB(always_running_containers)
```
yozhao101, line 118: One question: would it be feasible to combine the function `get_current_running_from_DB(...)` and `get_current_running_from_dockers()` together? and we can keep the function name as `get_current_running_containers()`.

files/image_config/monit/container_checker
renukamanavalan: done.

files/image_config/monit/container_checker
renukamanavalan: done

files/image_config/monit/container_checker
renukamanavalan: No big opinion in either way. If it would help with code read, which it sounds like, why not.

renukamanavalan: Done.

yozhao101 (approved):

files/image_config/monit/container_checker
```diff
@@ -70,37 +51,87 @@ def get_expected_running_containers():
...
+
+    for name in tbl.getKeys():
+        data = dict(tbl.get(name)[1])
+        if data.get('container_id'):
+            running_containers.add(name)
+
+    DOCKER_CLIENT = docker.DockerClient(base_url='unix://var/run/docker.sock')
+    RUNNING = 'running'
+    for name in always_running_containers:
+        container = DOCKER_CLIENT.containers.get(name)
```
yozhao101, line 87: One question: we probably need handle the exception if the method `get(id_or_name)` did raise one? https://docker-py.readthedocs.io/en/stable/containers.html#docker.models.containers.ContainerCollection.get.

files/image_config/monit/container_checker
```diff
@@ -70,37 +51,87 @@ def get_expected_running_containers():
...
+    """
+    @summary: This function will get all running containers from
+              the list of docker containers in running state.
+    @return:  A set which contains containers that are
+              in running state.
+    """
+    DOCKER_CLIENT = docker.DockerClient(base_url='unix://var/run/docker.sock')
+    running_containers = set()
+    ctrs = DOCKER_CLIENT.containers
+    lst = ctrs.list(filters={"status": "running"})
```
yozhao101, line 105: One question: we probably need handle the exception if the method `list(**kwargs)` did raise one?
https://docker-py.readthedocs.io/en/stable/containers.html#docker.models.containers.ContainerCollection.list.

files/image_config/monit/container_checker
renukamanavalan: Good point. We should.

files/image_config/monit/container_checker
```diff
@@ -70,37 +51,93 @@ def get_expected_running_containers():
...
+    """
+    DOCKER_CLIENT = docker.DockerClient(base_url='unix://var/run/docker.sock')
+    running_containers = set()
+    ctrs = DOCKER_CLIENT.containers
+    try:
+        lst = ctrs.list(filters={"status": "running"})
+        for ctr in lst:
+            running_containers.add(ctr.name)
+    except docker.errors.APIError:
+        pass
```
yozhao101, line 113: Instead of `pass`, can we write an alerting message into syslog to indicate there was an error occurring? like the following message:

    except docker.errors.APIError as err:
        print("Failed to retrieve the running container list. Error: '{}'".format(err))
        sys.exit(2)

renukamanavalan: This implies something is seriously wrong, which would be visible in many different ways. We could stay stick to the purpose of reporting that no dockers are running.

yozhao101 (approved):

files/image_config/monit/container_checker
```diff
@@ -70,37 +51,94 @@ def get_expected_running_containers():
...
+    """
+    DOCKER_CLIENT = docker.DockerClient(base_url='unix://var/run/docker.sock')
+    running_containers = set()
+    ctrs = DOCKER_CLIENT.containers
+    try:
+        lst = ctrs.list(filters={"status": "running"})
+        for ctr in lst:
+            running_containers.add(ctr.name)
+    except docker.errors.APIError:
+        print("Failed to retrieve the running container list. Error: '{}'".format(err))
```
yozhao101, line 113: The variable `err` in the `print(...)` statement need be defined by the exception statement:

        except docker.errors.APIError as err:

files/image_config/monit/container_checker
```diff
@@ -70,37 +51,94 @@ def get_expected_running_containers():
...
+
+    DOCKER_CLIENT = docker.DockerClient(base_url='unix://var/run/docker.sock')
+    RUNNING = 'running'
+    for name in always_running_containers:
+        try:
+            container = DOCKER_CLIENT.containers.get(name)
+            container_state = container.attrs.get('State', {})
+            if container_state.get('Status', "") == RUNNING:
+                running_containers.add(name)
+        except (docker.errors.NotFound, docker.errors.APIError):
```
yozhao101, line 92: We probably also need add an error message at here like the code at line 113.

yozhao101 (approved):

PR #7476, opened 2021-04-29
Remove quagga related code

shi-su:
#### Why I did it
Quagga is no longer being used. Remove quagga related code (e.g., docker-fpm-quagga, sonic-quagga, etc.). 

#### How I did it
Remove quagga related code.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012


platform/p4/docker-sonic-p4.mk
```diff
@@ -1,31 +0,0 @@
-# docker image for p4 sonic docker image
-
-DOCKER_SONIC_P4 = docker-sonic-p4.gz
```
qiluo-msft, line 3: >p4 [](http://example.com/codeflow?start=31&length=2)

Is this related to quagga? Why remove?

shi-su: Removing p4 related part is because it uses quagga regardless of the routing stack selected. Updated the PR description to clarify that.

rules/gobgp.mk
```diff
@@ -1,5 +0,0 @@
-# gobgp package
```
qiluo-msft, line 1: >gobgp [](http://example.com/codeflow?start=2&length=5)

gobgp itself is not related to quagga. Why remove?

shi-su: Undid removal for gobgp.

lguohan: let's keep gobgp and p4. only remove quagga.

qiluo-msft (approved):

qiluo-msft:
@shi-su Could you update the PR title? I think it is not touching p4.

PR #7483, opened 2021-04-29
[makefile] define a do-nothing target for config.user

yxieca:
#### Why I did it
After PR #7344, 'make init' and/or 'make reset' will also build sonic slave dockers.

'-include rules/config.user' is supposed to be fine when the file is missing.  However, when the file is missing, it generates a delayed error which later causes make init and make reset trying to build the sonic slave dockers.
    
#### How I did it
Define a do-nothing target for config.user to catch config.user build therefore preventing other builds to be triggered unexpectedly.

#### How to verify it
did make init and it is now only doing submodule init.

### DIFF
--- a/Makefile.work
+++ b/Makefile.work
@@ -100,7 +100,10 @@ SLAVE_DIR = sonic-slave-jessie
 endif
 
 include rules/config
--include rules/config.user
+
+ifeq (rules/config.user, $(wildcard rules/config.user))
+    include rules/config.user
+endif
 
 ifeq ($(ENABLE_DOCKER_BASE_PULL),)
 	override ENABLE_DOCKER_BASE_PULL = n
--- a/Makefile.work
+++ b/Makefile.work
@@ -99,8 +99,8 @@ else
 SLAVE_DIR = sonic-slave-jessie
 endif
 
-# Define an empty target for rules/config.user so that when
-# the file is missng, make won't try to rebuld everything.
+# Define a do-nothing target for rules/config.user so that when
+# the file is missing, make won't try to rebuld everything.
 rules/config.user:
 	@echo -n ""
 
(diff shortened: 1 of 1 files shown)


qiluo-msft:
It seems missing by design https://github.com/Azure/sonic-buildimage/commit/7e2fa7d9973105752e2e9d4abb3c9031657d0eaa. What issue are you facing?

yxieca:
> It seems missing by design [7e2fa7d](https://github.com/Azure/sonic-buildimage/commit/7e2fa7d9973105752e2e9d4abb3c9031657d0eaa). What issue are you facing?

as in why I did it :-)
After PR #7344, 'make init' and/or 'make reset' will also build sonic slave dockers.

yxieca:
-include is supposed to be fine when file is missing. checking more.

yxieca:
@qiluo-msft you are right the design was that the file can be missing.

@bluecmd  can you take a look at this change. This is less exciting then your change but work-around the issue that sonic slave docker got built with "make init" and "make reset". If you have better way of preventing it, please let me know or propose with another PR.

bluecmd:
Hi!

While I can't say I noticed this behavior when I made the change, I found this passage in the Make docs that might be the reason why the include seems to cause other things to build:

> If an included makefile cannot be found in any of these directories, a warning message is generated, but it is not an immediately fatal error; processing of the makefile containing the include continues. Once it has finished reading makefiles, make will try to remake 
https://www.gnu.org/software/make/manual/html_node/Include.html

Reading that, it seems that this could possibly also be fixed by explicitly defining `rules/config.user` as an empty target that can't be made. Maybe like this (100% untested!):

```make

rules/config.user:

include rules/config
-include rules/config.user
```

Personally, I think the wildcard approach here is pretty smart and nice looking - but I also know we have `-include rules/config.user` in at least one other place, so defining this target once might fix it for everyone.

What do you say about giving this other way a try and see if it works and then we can see which approach seems the right way forward?

qiluo-msft (approved):

PR #7484, opened 2021-04-29
Support for updating tmpfs size according to Image size

lkunjumon:
#### Why I did it
while sonic upgrade, Image will be extracted to tmpfs for installation so tmpfs size should be larger than image size. Image installation will fail if image size is larger than tmpfs size.

we are facing below error while installing debug image with size greater than tmpfs which is 1.5g in marvell armhf platform.

sonic-installer install <url>
New image will be installed, continue? [y/N]: y
Downloading image...
...99%, 1744 MB, 708 KB/s, 0 seconds left...
Installing image SONiC-OS-202012.0-dirty-20210311.224845 and setting it as default...
Command: bash /tmp/sonic_image
tar: installer/fs.zip: Wrote only 7680 of 10240 bytes
tar: installer/onie-image-arm64.conf: Cannot write: No space left on device
tar: Exiting with failure status due to previous errors
Verifying image checksum ... OK.
Preparing image archive ...

#### How I did it
compare downloaded image size with tmpfs size, if size less than image size update the tmpfs size according to image size.

#### How to verify it
Install an Image with size larger than tmpfs. we verified by installing debug image with size 1.9gb which is larger than tmpfs size 1.5gb.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012

### DIFF
--- a/installer/sharch_body.sh
+++ b/installer/sharch_body.sh
@@ -25,7 +25,7 @@ fi
 
 echo " OK."
 
-image_size=$(du "$0" | awk '{print $1}')
+image_size=$(( $(sed -e '1,/^exit_marker$/d' "$0"  | tar --to-stdout -xf - | wc -c) / 1024))
 # Untar and launch install script in a tmpfs
 cur_wd=$(pwd)
 export cur_wd
--- a/installer/sharch_body.sh
+++ b/installer/sharch_body.sh
@@ -34,8 +34,10 @@ tmp_dir=$(mktemp -d)
 if [ "$(id -u)" = "0" ] ; then
     mount -t tmpfs tmpfs-installer $tmp_dir || exit 1
     mount_size=$(df $tmp_dir | tail -1 | tr -s ' ' | cut -d' ' -f4)
-    if [ "$mount_size" -lt "$((image_size*3))" ]; then
-        mount_size=$((((image_size*3)/1024/1024)+1))
+    #checking extra 100KB space in tmp_dir, after image extraction
+    padding=102400
+    if [ "$mount_size" -le "$((image_size + padding))" ]; then
+        mount_size=$(((image_size/1024/1024)+1))
         mount -o remount,size="${mount_size}G" -t tmpfs tmpfs-installer $tmp_dir || exit 1
     fi
 fi
--- a/installer/sharch_body.sh
+++ b/installer/sharch_body.sh
@@ -25,6 +25,7 @@ fi
 
 echo " OK."
 
+image_size=$(du "$0" | awk '{print $1}')
 # Untar and launch install script in a tmpfs
 cur_wd=$(pwd)
 export cur_wd
@@ -32,6 +33,11 @@ archive_path=$(realpath "$0")
 tmp_dir=$(mktemp -d)
 if [ "$(id -u)" = "0" ] ; then
     mount -t tmpfs tmpfs-installer $tmp_dir || exit 1
+    mount_size=$(df $tmp_dir | tail -1 | tr -s ' ' | cut -d' ' -f4)
+    if [ "$mount_size" -lt "$((image_size*3))" ]; then
+        mount_size=$((((image_size*3)/1024/1024)+1))
+        mount -o remount,size="${mount_size}G" -t tmpfs tmpfs-installer $tmp_dir || exit 1
+    fi
 fi
 cd $tmp_dir
 echo -n "Preparing image archive ..."
--- a/installer/sharch_body.sh
+++ b/installer/sharch_body.sh
@@ -25,7 +25,7 @@ fi
 
 echo " OK."
 
-image_size=$(( $(sed -e '1,/^exit_marker$/d' "$0"  | tar --to-stdout -xf - | wc -c) / 1024))
+image_size_in_kb=$((($(sed -e '1,/^exit_marker$/d' "$0"  | tar --to-stdout -xf - | wc -c) + 1023 ) / 1024))
 # Untar and launch install script in a tmpfs
 cur_wd=$(pwd)
 export cur_wd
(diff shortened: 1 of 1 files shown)


installer/sharch_body.sh
```diff
@@ -25,13 +25,19 @@ fi
...
+image_size=$(du "$0" | awk '{print $1}')
 # Untar and launch install script in a tmpfs
 cur_wd=$(pwd)
 export cur_wd
 archive_path=$(realpath "$0")
 tmp_dir=$(mktemp -d)
 if [ "$(id -u)" = "0" ] ; then
     mount -t tmpfs tmpfs-installer $tmp_dir || exit 1
+    mount_size=$(df $tmp_dir | tail -1 | tr -s ' ' | cut -d' ' -f4)
+    if [ "$mount_size" -lt "$((image_size*3))" ]; then
```
qiluo-msft, line 37: >image_size*3 [](http://example.com/codeflow?start=31&length=12)

This seems an estimation. How do you know *3 is enough? Could you get the accurate required size?

lguohan: that must be generated from the build process.

Blueve:
I hit same issue on Celestica platform.
The tmpfs is 991M by default but some of image size are larger than 1G

installer/sharch_body.sh
lguohan: use this to get the original size?

lgh@p330:~$ gzip -l docker-sonic-vs.gz 
         compressed        uncompressed  ratio uncompressed_name
          242618472           684168704  64.5% docker-sonic-vs

antony-rheneus: This script untars sonic-<platform>.bin, which is a ONIE image. We cannot use gzip command. 
Will update the logic by getting image size from the "tar" command, excluding the ONIE installer script

lkunjumon:
Updated logic to get image size  from "tar" command.

installer/sharch_body.sh
```diff
@@ -25,13 +25,19 @@ fi
 
 echo " OK."
 
+image_size=$(( $(sed -e '1,/^exit_marker$/d' "$0"  | tar --to-stdout -xf - | wc -c) / 1024))
```
qiluo-msft, line 28: >/ [](http://example.com/codeflow?start=84&length=1)

Do you want to get the ceiling of division?

installer/sharch_body.sh
```diff
@@ -25,13 +25,19 @@ fi
...
 # Untar and launch install script in a tmpfs
 cur_wd=$(pwd)
 export cur_wd
 archive_path=$(realpath "$0")
 tmp_dir=$(mktemp -d)
 if [ "$(id -u)" = "0" ] ; then
     mount -t tmpfs tmpfs-installer $tmp_dir || exit 1
+    mount_size=$(df $tmp_dir | tail -1 | tr -s ' ' | cut -d' ' -f4)
+    if [ "$mount_size" -lt "$((image_size*3))" ]; then
```
qiluo-msft, line 37: >image_size [](http://example.com/codeflow?start=31&length=10)

Is `image_sizes*1` is good enough?

installer/sharch_body.sh
lkunjumon: okay.will update with ceiling.

installer/sharch_body.sh
lkunjumon: I have addressed the comment. Please review

installer/sharch_body.sh
qiluo-msft: Sorry, how do you get the ceiling? Also there are 2 more division below.

Instead of `x / 1024`,
`(x + 1023) / 1024` will be much safer.

lkunjumon: we can avoid ceiling here as we already adding  ceiling to image_size  by 1G  in this command  " mount_size=$(((image_size/1024/1024)+1)) " .Here if we add it will be ceiled by 1KB only ,so it is not necessary.

lkunjumon:
retest please

installer/sharch_body.sh
qiluo-msft: Okay, you are using padding instead of ceiling. Do you think 1G padding is too large?

lkunjumon: "mount_size=$(((image_size/1024/1024)+1))" Here we are handling ceiling for image_size  and extra free space for the mount , so extra free space will be less than 1GB.

radha-danda:
@lguohan, @qiluo-msft, we have answered the queries on this PR. Kindly let us know if you have any further queries. Can this PR be merged?

installer/sharch_body.sh
qiluo-msft: My only concern is the extreme padding size 1G seem too large. Others look good to me.

radha-danda: @qiluo-msft, 
is it fine to update 32MB extra space? Find below changes. If this is fine, we will push the changes.

mount_size=$((((image_size+1023)/1024) + 32))
mount -o remount,size="${mount_size}M" -t tmpfs tmpfs-installer $tmp_dir || exit 1

qiluo-msft: 32M padding is good. However you divided 3 times, so I think you formula will not work for corner cases.
This is my math
```
bytes = $(sed -e '1,/^exit_marker$/d' "$0"  | tar --to-stdout -xf - | wc -c) 
image_size=$((bytes + 1023) / 1024)
mount_size=$((image_size + 31) / 32) * 32
mount -o remount,size="${mount_size}M"
```

lkunjumon: > 32M padding is good. However you divided 3 times, so I think you formula will not work for corner cases.
> This is my math
> 
> ```
> bytes = $(sed -e '1,/^exit_marker$/d' "$0"  | tar --to-stdout -xf - | wc -c) 
> image_size=$((bytes + 1023) / 1024)
> mount_size=$((image_size + 31) / 32) * 32
> mount -o remount,size="${mount_size}M"
> ```

Updated PR with your comments. Please review.

qiluo-msft (approved):

radha-danda:
> @lguohan, @qiluo-msft, we have answered the queries on this PR. Kindly let us know if you have any further queries. Can this PR be merged?

radha-danda:
@lguohan, can this PR be merged as it is approved by Qi Luo

dflynn-Nokia:
Please cherry-pick this PR to the 202012 and 202106 releases.

PR #7516, opened 2021-05-05
[ci] Fix official builds timeout issue

xumia:
#### Why I did it
Fix official build timeout issue.
For example: https://dev.azure.com/mssonic/build/_build/results?buildId=13575&view=results

#### How I did it
The default value is 600 minutes, it is not enough when building multiple images for a platform, change to 720 minutes.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012


.azure-pipelines/azure-pipelines-job-as-stage.yml
```diff
@@ -0,0 +1,10 @@
+parameters:
```
lguohan, line 1: it is unclear why these changes are related to the pr.

xumia: The official build is using the file, supports to build single platform for platform while using the same yaml file.
The old one build-template.yml is no use any more, can be removed.

xumia: @lguohan , I may not understand your concern, could you please give more detail about it? We can talk about it if needed.

lguohan: according to the description, it increases the timeout from 600m to 720m. then, isn't line 22 on the job-groups.yml doing that? why do we need to add this stage.yml?

xumia: Oh, I got it. I thought only one line change in my PR. The stage.yml was for my test only, it got merged by mistake. I have removed it. Thanks.

lguohan (approved):

Issue #7488, opened 2021-04-29
BGP fails to restore route table before reconciliation in warm reboot


shi-su:
#### Steps to reproduce the issue:
1. Run continuous warm-reboot test on kvm
2. The warm-reboot fails in 1 out of 20 warm-reboots
3. An example failure https://dev.azure.com/mssonic/build/_build/results?buildId=12439&view=logs&j=917c1508-19c3-53d5-10ca-59367f40e278&t=82841ee8-7d01-5dbd-2843-d8193599bb85

#### Describe the results you received:
BGP fails to restore route table before reconciliation in warm reboot. The reconciliation process removes all routes including the default route. The route table gets restores tens of seconds after reconciliation.

Note: this issue is seen on KVM, whether it happens on other platforms has not yet been confirmed.

#### Describe the results you expected:
Route table restores before reconciliation. The reconciliation process does not remove any routes.

#### Output of `show version`:
The latest master image.
```
(paste your output here)
```
#### Output of `show techsupport`:

```
(paste your output here or download and attach the file here )
```

shi-su:
It seems that one of the peers gets updated late (after reconciliation). Here is an example with logs below: 10.0.0.59 gets updated after the reconciliation. It seems that there is some bgp configuration that prevents any routes that do not point to the whole pre-defined peer-group (10.0.0.57, 10.0.0.59, 10.0.0.61, 10.0.0.63). Therefore, without 10.0.0.59 getting updated, all ipv4 routes would not be added and the default route would have packet action as drop.

syslog
```
Apr 24 08:13:19.999392 vlab-01 NOTICE bgp#fpmsyncd: :- main: Warm-Restart timer expired.
Apr 24 08:13:20.001535 vlab-01 NOTICE bgp#fpmsyncd: :- reconcile: Warm-Restart: Initiating reconciliation process for bgp application.
Apr 24 08:13:20.002001 vlab-01 NOTICE bgp#fpmsyncd: :- reconcile: Warm-Restart reconciliation: deleting stale entry 192.184.192.0/25 { nexthop: 10.0.0.57,10.0.0.59,10.0.0.61,10.0.0.63 | ifname: PortChannel0001,PortChannel0002,PortChannel0003,PortChannel0004 } 
[...]

PR #7533, opened 2021-05-06
Reduce route selection deferral timer for bgp graceful restart

shi-su:
#### Why I did it
There are scenarios that End-of-RIB comes from a part of the peers arrives after reconciliation. In such scenarios, if the route selection deferral timer has the default value of 360 seconds, FRR would not set up routes and all routes would be removed after reconciliation. This PR reduces the route selection deferral timer so that at least routes to parts of the peers get restored at the point of reconciliation.

Fix https://github.com/Azure/sonic-buildimage/issues/7488

#### How I did it
Reduce route selection deferral timer for bgp graceful restart to 15 seconds.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012

### DIFF
--- a/dockers/docker-fpm-frr/frr/bgpd/bgpd.main.conf.j2
+++ b/dockers/docker-fpm-frr/frr/bgpd/bgpd.main.conf.j2
@@ -57,7 +57,7 @@ router bgp {{ DEVICE_METADATA['localhost']['bgp_asn'] }}
   bgp graceful-restart restart-time {{ constants.bgp.graceful_restart.restart_time | default(240) }}
   bgp graceful-restart
   bgp graceful-restart preserve-fw-state
-  bgp graceful-restart select-defer-time {{ constants.bgp.graceful_restart.select_defer_time | default(0) }}
+  bgp graceful-restart select-defer-time {{ constants.bgp.graceful_restart.select_defer_time | default(15) }}
 {% endif %}
 !
 {# set router-id #}
--- a/dockers/docker-fpm-frr/frr/bgpd/bgpd.main.conf.j2
+++ b/dockers/docker-fpm-frr/frr/bgpd/bgpd.main.conf.j2
@@ -57,7 +57,7 @@ router bgp {{ DEVICE_METADATA['localhost']['bgp_asn'] }}
   bgp graceful-restart restart-time {{ constants.bgp.graceful_restart.restart_time | default(240) }}
   bgp graceful-restart
   bgp graceful-restart preserve-fw-state
-  bgp graceful-restart select-defer-time {{ constants.bgp.graceful_restart.select_defer_time | default(15) }}
+  bgp graceful-restart select-defer-time {{ constants.bgp.graceful_restart.select_defer_time | default(25) }}
 {% endif %}
 !
 {# set router-id #}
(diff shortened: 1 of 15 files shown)


zhenggen-xu:
If we disable "route selection deferral timer", the restarting speaker could get into the situation where it only gets some peers EOR and then do route selection too early, and then had some unnecessary routes updates at BGP layer.  This could propagate to peers and impact traffic.

Can we just reduce this timer (route selection deferral timer) to some value that fit your environment?  The default seemed to be 360 seconds.  

Also, the original issue (https://github.com/Azure/sonic-buildimage/issues/7488) where we don't get EOR after reconciliation (120 seconds) from one peer is also bit odd as that timer value is very big already.

lguohan:
i agree with @zhenggen-xu, if the deferral time is too small, it could cause negative impact by send the routes to early to the peers. I think this timer can be around 90 seconds so that it will timeout before fpmsync reconcilation timeout, meanwhile not too small so that given peer a chance to send EOR. 

meanwhile, I do think we need to understand when EOR is not received after 120 seconds.

shi-su:
> this timer can be around 90 seconds so that it will timeout before fpmsync reconcilation timeout, meanwhile not too small so that given peer a chance to send EOR.

One tricky thing about this selection deferral timer is that it seems to take effect per peer and starts counting from a peer to get established. That is to say, we are unable to control its timeout in the timeline of reconciliation since we do not know when the first peer gets established.

I am updating the selection deferral timer to 15 seconds, as per my observation, the EOR from all peers is likely to arrive within 15 seconds from the first one to get established. Assuming the reconciliation timer has a 15 seconds margin that the first peer should get established at least 15 seconds before the reconciliation happens, such a deferral timer should address the issue while not impose a negative effect.

In the meantime, we can put some effort into understanding why EOR comes after 120 seconds.

lguohan:
> as per my observation, the EOR from all peers is likely to arrive within 15 seconds from the first one to get established
is this valid? should we give more tolerance here?

shi-su:
> > as per my observation, the EOR from all peers is likely to arrive within 15 seconds from the first one to get established
> 
> is this valid? should we give more tolerance here?

I think we need to have an elaborate analysis on the timing to find a good value for the timer here. Let me collect more data and will circle back with my findings.

shi-su:
I collected some data on the KVM platform about the timing of the first peer gets established after the reconciliation timer starts, and the interval between the first peer gets established and the last EOR is received. The distribution is presented in the figure attached. The results show that the typical time for the first peer to be established after the reconciliation timer starts is about 70 seconds, whereas the interval between the first peer gets established and the last EOR received ranges between 5 to 20 seconds. So I think 25 seconds would be a reasonable choice for the selection deferral timer to ensure some routes can be selected while avoiding unnecessary route updates.

I think another method might be extending the reconciliation timer to the same value as the selection deferral timer. This could better prevent running into the empty route table issue. The price is that the control plane might come back a bit later.

 
illustration

yxieca (approved):

judyjoseph:
Please raise a new PR for 202111 branch, as this patch cannot be cleanly cherry-picked.

shi-su:
> Please raise a new PR for 202111 branch, as this patch cannot be cleanly cherry-picked.

This PR is already included in 202111 branch, no need for cherry-picking.

PR #7545, opened 2021-05-06
[201911] [Innovium] Add new platforms and config updates

tonytitus:
#### Why I did it
Update Innovium configs + Add new platforms supporting Innovium chips
#### How I did it
NA
#### How to verify it
Sonic PTF
Will create separate PRs for other branches

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012

Update Innovium configs + Add new platforms supporting Innovium chips


jleveque (changes requested):
Please fix LGTM alerts

device/cameo/x86_64-cameo_esc600_128q-r0/plugins/eeprom.py
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env python
```
jleveque, line 1: These old platform plugins (eeprom.py, psuutil.py, sfputil.py) are deprecated in favor of the new sonic-platform package. Support for these will be removed soon in the master branch. I will allow these in the 201911 branch, but for master you should generate a Python 3 sonic-platform wheel package instead.

jleveque:
"Wrong number of arguments in a class instantiation" alerts are false alarms due to limitation/bug in LGTM. These alerts are acceptable.

jleveque (approved):

PR #7551, opened 2021-05-06
RADIUS Management User Authentication

a-barboza:
Fixed the RADIUS client debian packages build
  for arm64, armhf PLATFORM_ARCH, by using the latest
  config.sub/config.guess

#### Why I did it
Fixed the RADIUS client debian packages build for arm64, armhf PLATFORM_ARCH
#### How I did it
Used the latest config.sub/config.guess
#### How to verify it
Build the RADIUS client debian packages for arm64,armhf

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012

  Fixed the RADIUS client debian packages build
  for arm64, armhf PLATFORM_ARCH, by using the latest
  config.sub/config.guess


lguohan:
config.sub and config.guess are generated file, we should not check-in those files. 

can you take a look at debian packaging to see how they can enable the build of arm64?

https://packages.debian.org/buster/freeradius

a-barboza:
> config.sub and config.guess are generated file, we should not check-in those files.
> 
> can you take a look at debian packaging to see how they can enable the build of arm64?
> 
> https://packages.debian.org/buster/freeradius

I used that similar technique as freeradius maintainers. Please see this commit id.

https://github.com/FreeRADIUS/freeradius-server/commit/6d47caadada66812854aaad332c1564cd927f64d

lguohan:
i did not realize your version so old? 1.1.1? can you use later version. buster is using 3.0.17 now.

a-barboza:
> i did not realize your version so old? 1.1.1? can you use later version. buster is using 3.0.17 now.

That is the server version. We are using the pam_radius client which does not have an official debian version, so it is being built from github.com/pam_radius. Some of the PEAP modules are used from freeradius-server for PEAP authentication

lguohan:
for client, debian packaging also do not have guess. can you check how debian package and build on arm64? https://packages.debian.org/buster/libpam-radius-auth

why do we need build server here?

a-barboza:
> for client, debian packaging also do not have guess. can you check how debian package and build on arm64? https://packages.debian.org/buster/libpam-radius-auth
> 
> why do we need build server here?

Cc: @adyeung

Got it. I will check the debian source package for libpam-radius-auth for buster for arm64/armhf compile. 

The github sources for libpam-radius-auth (https://github.com/FreeRADIUS/pam_radius) did not have support for the PEAP/MSCHAPv2 authentication, therefore the library from the freeradius-server was used to implement this. Another option would be to look for this library if it appears separately, or write one from scratch. The latter would require considerable more effort, therefore the option of using freeradius-server (from that older version) was chosen.

lguohan:
if you just want to use the library, why not use this one? https://packages.debian.org/buster/libfreeradius3

a-barboza:
> if you just want to use the library, why not use this one? https://packages.debian.org/buster/libfreeradius3

libfreeradius3 is being built from the freeradius (previously freeradius-server) source tar ball. At the time of original writing the freeradius 1.1.x tree was being used. Now, it is 3.0.x. So, I understand the ask to mean: upgrade to 3.0.x for using the latest library from freeradius (server) source package. The re-write to use the 3.0.x interfaces would require more time.

lguohan:
already, please fix the client side first? and then plan for upgrade to use the freeradius library accordingly.

a-barboza:
> understood. Will update PR for the pam_radius to use the buster debian arm64.

Updated on 05/14/21

lguohan (approved):

PR #7382, opened 2021-04-20
[platform] Add serial number and model number to Mellanox PSU platform implementation

robocoder99:
#### Why I did it

We want to add the ability for the command `show platform psustatus` to show the serial number and part number of the PSU devices on Mellanox platforms. This will be useful for data-center management of field replaceable units (FRUs) on switches.

#### How I did it

I implemented the platform 2.0 functions `get_model()` and `get_serial()` for the PSU in the mellanox platform API by referencing the sysfs nodes provided by the [hw-management](https://github.com/Azure/sonic-buildimage/tree/master/platform/mellanox/hw-management) module.

#### How to verify it

I ran `show platform psustatus` and verified that the serial number and model number were present on platforms where the information was accessible through sysfs. (Currently functioning on all mellanox platforms other than MSN2100 and MSN2010)

#### Which release branch to backport (provide reason below if selected)

None

#### Description for the changelog

[platform] Add serial number and model number to Mellanox PSU platform implementation

#### A picture of a cute animal (not mandatory but encouraged)

97f64614927328325cf385c8be8aa80b

PR #7552, opened 2021-05-06
[Mellanox] Implement Hardware Revision Platform API Call for Mellanox Chassis and PSU

robocoder99:
#### Why I did it

This pull request allows calls to be made through the platform 2.0 API that retrieve the PSU and Chassis hardware revision on Mellanox platforms. Access to these values will aid customers in determining their hardware revisions for debugging and technical support. These values are intended to be eventually exposed through the CLI. 

#### How I did it

For the PSU hardware revision I used the existing VPD function calls implemented in https://github.com/Azure/sonic-buildimage/pull/7382

For the Chassis hardware revision I parsed the SMBIOS / DMI type 2 information to retrieve the information. 

#### How to verify it

To verify on a switch open a python 3 interpreter with root access and run the following script....

```
import sonic_platform
platform_chassis = sonic_platform.platform.Platform().get_chassis()

chassis_rev = platform_chassis.get_revision()
psu_rev = platform_chassis.get_all_psus()[0].get_revision()

print("Chassis Rev:", chassis_rev)
print("PSU Rev:", psu_rev)
```

The hardware revision of the chassis and the psu should print successfully. (Generally for mellanox this will look like "A[1-3]") 

Verified on all mellanox platforms aside from MSN2010.

[platform] Implement Hardware Revision Platform API Call for Mellanox Chassis and PSU

goats

### DIFF
--- a/platform/mellanox/mlnx-platform-api/sonic_platform/chassis.py
+++ b/platform/mellanox/mlnx-platform-api/sonic_platform/chassis.py
@@ -26,6 +26,23 @@ MAX_SELECT_DELAY = 3600
 
 MLNX_NUM_PSU = 2
 
+DMI_FILE = '/sys/firmware/dmi/entries/2-0/raw'
+DMI_HEADER_LEN = 15
+DMI_PRODUCT_NAME = "Product Name"
+DMI_MANUFACTURER = "Manufacturer"
+DMI_VERSION = "Version"
+DMI_SERIAL = "Serial Number"
+DMI_ASSET_TAG = "Asset Tag"
+DMI_LOC = "Location In Chassis"
+DMI_TABLE_MAP = {
+                    DMI_PRODUCT_NAME: 0,
+                    DMI_MANUFACTURER: 1,
+                    DMI_VERSION: 2,
+                    DMI_SERIAL: 3,
+                    DMI_ASSET_TAG: 4,
+                    DMI_LOC: 5
+                }
+
 EEPROM_CACHE_ROOT = '/var/cache/sonic/decode-syseeprom'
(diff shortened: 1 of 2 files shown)


platform/mellanox/mlnx-platform-api/sonic_platform/chassis.py
```diff
@@ -67,6 +84,9 @@ def __init__(self):
         # Initialize Platform name
         self.platform_name = device_info.get_platform()
 
+        # Initialize DMI data
```
liat-grozovik, line 87: Do we have such init on different APIs? i think eeprom or maybe transceiver has it as well.
Is it ok to have the same init on different APIs? if so, no problem. lets just make sure it is.
Also, recently I saw other PRs to improve the init as some APIs can be from pmon and some from the host, so need to see if such improvement is relevant here as well. Maybe not

robocoder99: Do you mind linking in the PRs that you have seen that may be relevant? At least using the method that I am there will be no problem performing this on multiple APIs as far as linux is concerned it is reading a file.

robocoder99: I added lazy loading, I believe this should cover this concern. The DMI will now only be read when the hardware revision is specifically requested AND the dmi has not already been read.

Junchao-Mellanox (approved):

keboliu (approved):

liat-grozovik (approved):

jleveque (approved):

PR #7564, opened 2021-05-08
[Mellanox] Use a service to clear EEPROM cache

Junchao-Mellanox:
#### Why I did it

Use a service to clear eeprom cache during boot up so that CLI does not have to remove and update cache everytime.

#### How I did it

Use a service to clear eeprom. It is a one shot which runs before PMON service.

#### How to verify it

Manual test.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012

### DIFF
--- a/files/build_templates/docker_image_ctl.j2
+++ b/files/build_templates/docker_image_ctl.j2
@@ -393,6 +393,7 @@ start() {
         -v mlnx_sdk_ready:/tmp \
         -e SX_API_SOCKET_FILE=/var/run/sx_sdk/sx_api.sock \
         -v /dev/shm:/dev/shm:rw \
+        -v /var/cache/sonic/decode-syseeprom:/var/cache/sonic/decode-syseeprom:rw \
 {%- else %}
 {%- if mount_default_tmpfs|default("n") == "y" %}
         --tmpfs /tmp \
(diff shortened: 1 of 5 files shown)


jleveque:
Now that we are using syseepromd to obtain EEPROM info from the hardware and store to State DB, I feel like this mechanism of caching EEPROM data to a file is now redundant. I think we might want to plan on removing it altogether.

@Junchao-Mellanox: What are your thoughts?

Junchao-Mellanox:
Hi @jleveque, if we are going to remove the eeprom cache and save the data in DB, this PR is not necessary. In that case, 
1. If platform API need get data from eeprom internally, it has to access State DB. For example, chassis name and part number are usually got from eeprom. I am OK with that, but I don't see any platform API directly access DB for now, is that a platform API design rule?
2. I guess vendor's platform API need to be updated if the cache mechanism is removed 

Any comment on this?

jleveque:
> Hi @jleveque, if we are going to remove the eeprom cache and save the data in DB, this PR is not necessary. In that case,
> 
> 1. If platform API need get data from eeprom internally, it has to access State DB. For example, chassis name and part number are usually got from eeprom. I am OK with that, but I don't see any platform API directly access DB for now, is that a platform API design rule?

Yes, the platform API is designed to interface with the platform hardware, it is not intended to access the DB. However, syseepromd uses the platform API to pull the data from the EEPROM and stores it in State DB. All applications in SONiC should reference that data stored in State DB rather than calling into the platform API directly. With this approach, there should be no need to cache the EEPROM data to a file. The platform API should only be used to get the data by syseepromd, and once it's stored to the DB, it is essentially cached.

> 2. I guess vendor's platform API implementation need to be updated if the cache mechanism is removed

Yes, this would most likely require modifying existing implementations.

We could move forward with this PR for the time being, scope out the work to remove the caching, then create a HLD.

PR #7574, opened 2021-05-10
Update the netifaces pip3 package version

qiluo-msft:
#### Why I did it
202012 branch is using reproducible build. The versions-py3 file must contains correct pip3 package version. Otherwise we will get a build error

```
Step 31/74 : RUN pip3 install          scapy==2.4.4          pyroute2==0.5.14          netifaces==0.10.9
 ---> Running in d7a2401dd21d
Collecting scapy==2.4.4
  Downloading scapy-2.4.4.tar.gz (1.0 MB)
Collecting pyroute2==0.5.14
  Downloading pyroute2-0.5.14.tar.gz (873 kB)
ERROR: Cannot install netifaces==0.10.9 because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested netifaces==0.10.9
    The user requested (constraint) netifaces==0.10.7

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict

ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/user_guide/#fixing-conflicting-dependencies
The command '/bin/sh -c pip3 install          scapy==2.4.4          pyroute2==0.5.14          netifaces==0.10.9' returned a non-zero code: 1
[  FAIL LOG END  ] [ target/docker-sonic-vs.gz ]
```

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012

### DIFF
--- a/files/build/versions/dockers/docker-sonic-vs/versions-py3
+++ b/files/build/versions/dockers/docker-sonic-vs/versions-py3
@@ -5,7 +5,7 @@ deepdiff==5.2.3
 ijson==2.6.1
 jsondiff==1.3.0
 M2Crypto==0.37.1
-netifaces==0.10.7
+netifaces==0.10.9
 ordered-set==4.0.2
 parameterized==0.8.1
 pexpect==4.8.0


xumia (approved):

xumia:
@qiluo-msft , It will change back to old value if version upgraded?

xumia:
@qiluo-msft , Looks like the version has already in https://github.com/Azure/sonic-buildimage/blob/202012/files/build/versions/default/versions-py3

qiluo-msft:
It's different file files/build/versions/dockers/docker-sonic-vs/versions-py3

PR #7586, opened 2021-05-12
[ntp] Fix ntp.conf template to allow setting of source port in CONFIG_DB

robocoder99:
#### Why I did it
Currently, there is a bug in the ntp.conf jinja2 template where it will ignore the `src_intf` directive in CONFIG_DB if there are multiple IP addresses associated with an interface. This code change fixes that bug and allows the template to select the correct source interface for NTP. 

#### How I did it
I did this by modifying the macro in `ntp.conf.j2` which determines if there is an ip address associated with an interface to set a state variable when it detects a valid interface entry in CONFIG_DB instead of outputting "true" directly (which could result in multiple "trues" outputted for interfaces with multiple valid IP addresses). 

#### How to verify it

1. Add two ipv4 addresses to an interface in SONiC

2. Add the following configuration to `config_db.json` 

```
{
"NTP": {
    "global": {
        "src_intf": "Ethernet1"
        }
    }
}
```
Replace `Ethernet1` with the interface name of the one you assigned the IP addresses to.

3. Run `sudo config reload -y`

4. Open `/etc/ntp.conf` and verify that the following line exists

```
...
interface listen Ethernet1
...
```
The interface specified should be the one set in the previous steps. 

[ntp] Fix ntp.conf template to allow setting of source port in CONFIG_DB

Light-Gray-German-Angora-Rabbit_Hidden-Springs-Farm_Shutterstock-e1612565280404

### DIFF
--- a/files/image_config/ntp/ntp.conf.j2
+++ b/files/image_config/ntp/ntp.conf.j2
@@ -41,13 +41,15 @@ interface ignore wildcard
 # if the source interface is configured but no ip on that interface, then listen on another
 # interface based on existing logic
 {%- macro check_ip_on_interface(interface_name, table_name) %}
+    {%- set ns = namespace(valid_intf = 'false') %}
     {%- if table_name %}
         {%- for (name, source_prefix) in table_name|pfx_filter %}
             {%- if source_prefix and name == interface_name %}
-true
+                {%- set ns.valid_intf = 'true' %}
             {%- endif %}
         {%- endfor %}
     {%- endif %}
+{{ ns.valid_intf }}
 {%- endmacro %}
 
 {% set ns = namespace(source_intf = "") %}
--- a/src/sonic-config-engine/tests/data/ntp/ntp_interfaces.json
+++ b/src/sonic-config-engine/tests/data/ntp/ntp_interfaces.json
@@ -1,0 +1,12 @@
+{
+    "NTP": {
+        "global": {
+            "src_intf": "Ethernet0"
+        }
+    },
+    "INTERFACE": {
+        "Ethernet0": {},
+        "Ethernet0|10.0.0.0/31": {},
+        "Ethernet0|192.168.0.122/24": {}
+    }
+}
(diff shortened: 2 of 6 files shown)


lguohan:
can you add test in src/sonic-config-engine/tests to test the template?

files/image_config/ntp/ntp.conf.j2
```diff
@@ -41,13 +41,15 @@ interface ignore wildcard
...
+    {%- set ns = namespace(valid_intf = 'false') %}
     {%- if table_name %}
         {%- for (name, source_prefix) in table_name|pfx_filter %}
             {%- if source_prefix and name == interface_name %}
-true
+                {%- set ns.valid_intf = 'true' %}
             {%- endif %}
         {%- endfor %}
     {%- endif %}
+{{ ns.valid_intf }}
```
yxieca, line 52: Do you have to set ns.valid_intf to false as default value? What if the condition is not met?

robocoder99: It is set to a default value of false here https://github.com/Azure/sonic-buildimage/pull/7586/files#diff-0a4c1f55aab272c1613693aea10e89dbab5b46a6dd65182f9139717693be421dR44. 

If the condition is not met (i.e. it never finds a valid ip on the interface) it simply returns false.

yxieca (dismissed):

robocoder99:
@lguohan To verify that we are on the same page I assume you are referring to a new test here (https://github.com/Azure/sonic-buildimage/blob/master/src/sonic-config-engine/tests/test_j2files.py) that would validate the output of `ntp.conf` correct?

robocoder99:
I would use https://github.com/Azure/sonic-buildimage/blob/master/src/sonic-config-engine/tests/test_j2files.py#L291 as a template, it seems to be a similar test to the one I would want to perform (i.e. provide a minimal config_db file and check that the desired template output matches a ground truth) 

One question would be that there are many branches within this template file. Is our desire to test all of these branches? That may be out of the scope of this pull request. I would propose just adding a test for the case of multiple valid IP addresses for a given Ethernet interface.

lguohan:
i am ok with your proposal.

robocoder99:
@lguohan @yxieca  Added requested unit tests, please take a look when you can.

liat-grozovik (approved):

yxieca (approved):

PR #7588, opened 2021-05-12
Fix the /var wrong permission issue (#7571)

xumia:
#### Why I did it
Build broken when reproducible build enabled, relative to backport package systemd.
The root cause is the bad permission of /var.

See https://github.com/Azure/sonic-buildimage/pull/7571

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012


lguohan:
cherry-picked #7571

PR #7599, opened 2021-05-13
add config-setup.service as dependency for pcie-check.service

sujinmkang:
#### Why I did it
start pcie-check.service after config-setup.service since pcie_util depends on device_info which is available with config db metadata.
#### How I did it
Add config-setup.service as a dependency of pcie-check.service
#### How to verify it
Upon reboot, check if the pcie-check.sh throws the platform api error which is dependent on DEVICE_METADATA

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012

### DIFF
--- a/files/image_config/pcie-check/pcie-check.service
+++ b/files/image_config/pcie-check/pcie-check.service
@@ -1,6 +1,6 @@
 [Unit]
 Description=Check the PCIe device presence and status
-After=rc.local.service database.service
+After=rc.local.service database.service config-setup.service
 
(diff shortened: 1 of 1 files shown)


jleveque:
@sujinmkang: In the "Why I did it" section, can you please state the reason why the dependency is needed?

jleveque (approved):

shlomibitton:
@jleveque can you please cherry-pick this change to 202012 also?

PR #7617, opened 2021-05-14
[muti-asic] Updated BGP community for Internal routes

abdosi:
What I did:

Following changes are done:

- Internal routes are tagged with `no-export` instead pf `local-AS`
- Option to add User Define BGP community on top of `no-export`

How I verify:

- Unit test for BGP Template verification
- Manual Verification on BGPMON.
- https://github.com/Azure/sonic-mgmt/pull/3484


arlakshm (approved):
lgtm

PR #7630, opened 2021-05-17
[TPID CONFIG]TPID attribute Yang model and default TPID for Minigraph to configDB Changes

gechiang:
#### Why I did it
This PR is part of the TPID CONFIG SONiC feature set to support port/LAG TPID configuration.  
TPID HLD PR: (https://github.com/Azure/SONiC/pull/681)

There are total of 3 PRs across different submodules to achieve this feature support.

- SONiC-Utilities (https://github.com/Azure/sonic-utilities/pull/1618) - To handle TPID CLI related changes.
- SONiC-Swss (https://github.com/Azure/sonic-swss/pull/1747) - To handle Orchagent changes triggered from TPID configuration changes to interact with SYNCd SAI.
- and this PR which handles the new port/LAG TPID attribute on Yang model and Minigraph to configDB changes.

Please refer to the TPID HLD listed above for more details on the TPID configuration feature support.
Here is a quick summary for this PR changes:
By default TPID for port/LAG is 0x8100.  With HW SKU that has capability to handle TPID configuration user is allowed to configure the port/LAG to non default TPID values such as (0x9100, 0x9200, 0x88A8). 
This PR is to define this new TPID attribute for Port and LAG in the Yang model as well as to ensure all the ports/LAGs in the config DB generated from Minigraph should have a default TPID value of 0x8100 .

#### How to verify it
Test code were added for the Yang model to test the new TPID attribute in PORT and Portchannel.
Test code were also updated in test config gen, minigraph as well as multi-asic config gen to ensure the expected default TPID is in place.
Complete feature testing of TPID functionality were also performed on DUTs that has SAI support for TPID attribute to ensure the feature is working properly.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012

TPID port/LAG attribute Yang model and default TPID for Minigraph to configDB Changes


src/sonic-yang-models/tests/files/sample_config_db.json
```diff
@@ -252,258 +254,295 @@
...
                 "speed": "11100",
+                "tpid": "0x8100",
                 "admin_status": "up"
             },
             "Ethernet14": {
                 "alias": "Eth4/3",
                 "lanes": "79",
                 "description": "",
                 "speed": "11100",
+                "tpid": "0x8100",
```
praveen-li, line 369: We do not need to add it in all ports, just add it in 1 or 2 ports.

src/sonic-yang-models/yang-models/sonic-portchannel.yang
```diff
@@ -109,7 +109,15 @@ module sonic-portchannel {
                         }
                     }
                 }
-			} /* end of list PORTCHANNEL_LIST */
+
+                leaf tpid {
+                    type string {
+                        length 6;
```
praveen-li, line 115: See, if we can move type to stypes.yang, and use that typedef here.

src/sonic-yang-models/yang-models/sonic-port.yang
```diff
@@ -96,6 +96,14 @@ module sonic-port{
 						pattern "on|off";
 					}
 				}
+
+				leaf tpid {
```
gechiang, line 100: 1. Add description field to describe what this attribute is about
2. move to common so both port and LAG can share to avoid duplication

gechiang: we will not follow the open config model for this... for simplicity handling in config DB.

src/sonic-yang-models/tests/yang_model_tests/tests_config/portchannel.json
```diff
@@ -10,6 +10,7 @@
                         "lanes": "65",
                         "mtu": 9000,
                         "name": "Ethernet0",
+                        "tpid": "0x8100",
```
praveen-li, line 13: We can avoid updating old tests, since it is not mandatory field. Just add it in new tests.

src/sonic-yang-models/tests/yang_model_tests/tests_config/port.json
```diff
@@ -78,10 +82,28 @@
...
+                        "tpid": "0x8100",
                         "autoneg": 0
                     }
                 ]
             }
         }
-    }
+    },
+
+    "PORT_INVALID_TPID_TEST": {
```
praveen-li, line 93: We do not need extra test, once type is moved in stypes.

src/sonic-yang-models/tests/yang_model_tests/tests_config/port.json
```diff
@@ -12,6 +12,7 @@
                         "mtu": 9000,
                         "pfc_asym": "off",
                         "name": "Ethernet8",
+                        "tpid": "0x8100",
```
praveen-li, line 15: No need to update old tests.

src/sonic-yang-models/yang-models/sonic-port.yang
gechiang: In the future when stateDB validation method is agreed, we can come back to this and enhance it to check capability as part of the yang model validation for TPID attribute

src/sonic-config-engine/tests/test_multinpu_cfggen.py
```diff
@@ -149,16 +149,16 @@ def test_frontend_asic_portchannels(self):
         argument = "-m {} -p {} -n asic0 --var-json \"PORTCHANNEL\"".format(self.sample_graph, self.port_config[0])
         output = json.loads(self.run_script(argument))
         self.assertDictEqual(output, \
-            {'PortChannel0002': {'admin_status': 'up', 'min_links': '2', 'members': ['Ethernet0', 'Ethernet4'], 'mtu': '9100'},
-            'PortChannel4001': {'admin_status': 'up', 'min_links': '2', 'members': ['Ethernet-BP0', 'Ethernet-BP4'], 'mtu': '9100'},
-            'PortChannel4002': {'admin_status': 'up', 'min_links': '2', 'members': ['Ethernet-BP8', 'Ethernet-BP12'], 'mtu': '9100'}})
+                {'PortChannel0002': {'admin_status': 'up', 'min_links': '2', 'members': ['Ethernet0', 'Ethernet4'], 'mtu': '9100', 'tpid': '0x8100'},
```
dgsudharsan, line 152: Can you please correct the alignment in this file?

src/sonic-yang-models/tests/files/sample_config_db.json
gechiang: Going forward after this PR is approved and merged, the config DB when generated from the Minigraph will always have TPID 0x8100.  It is similar to the MTU handling where there is a default value set in config DB.
I like to keep this change to set TPID attribute for all the interfaces (port and LAG). this way it is more in sync when user inspect config DB.

src/sonic-yang-models/yang-models/sonic-portchannel.yang
gechiang: Agreed. Will have this changed in next commit.

src/sonic-yang-models/tests/yang_model_tests/tests_config/port.json
gechiang: Not sure I understood your comment.  moving the tpid type to a common typedef allows sharing of the same definition  but testing is still required isn't it?  LThis was added so that we can see that the Yang model correctly catches any attempt of using a wrong TPID value that is not specified in the model.  Please clarify ...
Thanks!

src/sonic-yang-models/tests/yang_model_tests/tests_config/portchannel.json
gechiang: I am actually leveraging this same test config file for the new attribute TPID testing.  Are you saying that I should generate a new test file just for TPID purpose?  since TPID is a new attribute for both port and portchannel, I thought of just expanding on existing test case  instead of generating a new file for just one TPID test...
There are no other tests needed other than ensure that the model can catch invalid TPID values.

src/sonic-config-engine/tests/test_multinpu_cfggen.py
gechiang: fixed.

src/sonic-yang-models/tests/yang_model_tests/tests_config/port.json
gechiang: TPID will be present in all config DB after thei feature is checked in so making sure it is consistently present in all interface configs.

lguohan (approved):

PR #7636, opened 2021-05-18
[202012] [sonic-snmpagent] submodule update

liorghub:
Includes below commits:
748b545 2021-04-05  [SNMP] Update description of entPhysicalDescr mib in case interface is not configured.

Signed-off-by: liora liora@nvidia.com


liorghub:
@lguohan @dprital @keboliu 
test_bgp_update_timer failed with the following error:
no bgp update withdraw route {'prefix': '10.10.100.128/27', 'nexthop': '192.168.0.2'} from 192.168.0.1 to 192.168.0.3

My changes are not related to BGP, hence there is a good chance that the test is not stable.
Can you rerun please?

qiluo-msft (approved):

PR #7652, opened 2021-05-19
[sonic-py-common] Add platform and chassis info methods to device_info

robocoder99:
#### Why I did it
These methods were added to make some convenient platform and chassis information methods accessible through sonic-py-common. These methods were refactored from sonic-utilities and are used in the `show platform summary` and `show version` commands. 

#### How I did it
There are two methods, one is `get_platform_info()` which simply calls local methods to collect useful platform information into a dictionary format, this came directly from sonic-utilities.

The other method is `get_chassis_info()` which retrieves the chassis model, serial number, and hardware revision from the STATE_DB directly. 

#### How to verify it
Open a python3 interpreter and enter the following lines
```
Python 3.7.3 (default, Jan 22 2021, 20:04:44)
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from sonic_py_common import device_info
>>> device_info.get_platform_info()
{'platform': 'x86_64-mlnx_msn2700-r0', 'hwsku': 'ACS-MSN2700', 'asic_type': 'mellanox', 'asic_count': 1}
>>> device_info.get_chassis_info()
{'model':'MSN2700-CS2FO', 'serial':'MT1623X09522', 'revision': 'A1'}
```

For different platforms these values will be different, simply check if they are successfully populated (Not empty or N/A). If it says "Not Provided" this means that the requested information is not encoded into the DMI of the switch.

[sonic-py-common] Add platform and chassis info methods to device_info

Highland-Cow-Lavena-and-calf-Star

### DIFF
--- a/src/sonic-py-common/sonic_py_common/device_info.py
+++ b/src/sonic-py-common/sonic_py_common/device_info.py
@@ -346,7 +346,7 @@ def get_chassis_info():
         chassis_info_dict['serial'] = db.get(db.STATE_DB, table, CHASSIS_INFO_SERIAL_FIELD)
         chassis_info_dict['model'] = db.get(db.STATE_DB, table, CHASSIS_INFO_MODEL_FIELD)
         chassis_info_dict['revision'] = db.get(db.STATE_DB, table, CHASSIS_INFO_REV_FIELD)
-    except:
+    except Exception:
         pass
 
     return chassis_info_dict
--- a/src/sonic-py-common/sonic_py_common/device_info.py
+++ b/src/sonic-py-common/sonic_py_common/device_info.py
@@ -7,6 +7,8 @@ import subprocess
 import yaml
 from natsort import natsorted
 
+from . import multi_asic
+
 # TODO: Replace with swsscommon
 from swsssdk import ConfigDBConnector, SonicDBConfig, SonicV2Connector
 
@@ -31,6 +33,52 @@ NAMESPACE_PATH_GLOB = "/run/netns/*"
 ASIC_CONF_FILENAME = "asic.conf"
 FRONTEND_ASIC_SUB_ROLE = "FrontEnd"
 BACKEND_ASIC_SUB_ROLE = "BackEnd"
+
+# Chassis STATE_DB keys
+CHASSIS_INFO_TABLE = 'CHASSIS_INFO|chassis {}'
+CHASSIS_INFO_CARD_NUM_FIELD = 'module_num'
+CHASSIS_INFO_SERIAL_FIELD = 'serial'
+CHASSIS_INFO_MODEL_FIELD = 'model'
+CHASSIS_INFO_REV_FIELD = 'revision'
+
+# Get hardware information
+def get_platform_info():
+    """
+    This function is used to get the HW info helper function
+    """
+
+    hw_info_dict = {}
+
+    version_info = device_info.get_sonic_version_info()
+
+    hw_info_dict['platform'] = device_info.get_platform()
+    hw_info_dict['hwsku'] = device_info.get_hwsku()
+    hw_info_dict['asic_type'] = version_info['asic_type']
+    hw_info_dict['asic_count'] = multi_asic.get_num_asics()
+
+    return hw_info_dict
+
+
+def get_chassis_info():
+    """
+    This function is used to get the Chassis serial / model / rev number
+    """
+
+    chassis_info_dict = {}
+
+    try:
+        # Init statedb connection
+        db = swsscommon.SonicV2Connector()
+        db.connect(db.STATE_DB)
+        table = CHASSIS_INFO_TABLE.format(1)
+
+        chassis_info_dict['serial'] = db.get(db.STATE_DB, table, CHASSIS_INFO_SERIAL_FIELD)
+        chassis_info_dict['model'] = db.get(db.STATE_DB, table, CHASSIS_INFO_MODEL_FIELD)
+        chassis_info_dict['revision'] = db.get(db.STATE_DB, table, CHASSIS_INFO_REV_FIELD)
+    except:
+        pass
+
+    return chassis_info_dict
 
 
 def get_localhost_info(field):
--- a/src/sonic-py-common/sonic_py_common/device_info.py
+++ b/src/sonic-py-common/sonic_py_common/device_info.py
@@ -39,46 +39,6 @@ CHASSIS_INFO_SERIAL_FIELD = 'serial'
 CHASSIS_INFO_MODEL_FIELD = 'model'
 CHASSIS_INFO_REV_FIELD = 'revision'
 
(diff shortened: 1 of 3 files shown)


robocoder99:
Not sure what is causing the build failures here (insert classic "all tests are passing on my machine")

Relevant line from CI build....
https://dev.azure.com/mssonic/build/_build/results?buildId=15822&view=logs&j=88ce9a53-729c-5fa9-7b6e-3d98f2488e3f&t=8d99be27-49d0-54d0-99b1-cfc0d47f0318&l=2096

```
_____________________ ERROR collecting tests/vlan_test.py ______________________
ImportError while importing test module '/sonic/src/sonic-utilities/tests/vlan_test.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/sonic/src/sonic-utilities/.eggs/pyroute2-0.6.1-py3.7.egg/pyroute2/__init__.py:44: in <module>
    from importlib import metadata
E   ImportError: cannot import name 'metadata' from 'importlib' (/usr/lib/python3.7/importlib/__init__.py)

During handling of the above exception, another exception occurred:
/usr/lib/python3/dist-packages/_pytest/python.py:450: in _importtestmodule
    mod = self.fspath.pyimport(ensuresyspath=importmode)
/usr/lib/python3/dist-packages/py/_path/local.py:668: in pyimport
    __import__(modname)
/usr/lib/python3/dist-packages/_pytest/assertion/rewrite.py:294: in load_module
    six.exec_(co, mod.__dict__)
tests/vlan_test.py:7: in <module>
    import config.main as config
config/main.py:24: in <module>
    from utilities_common.db import Db
utilities_common/db.py:4: in <module>
    from utilities_common.multi_asic import multi_asic_ns_choices
utilities_common/multi_asic.py:6: in <module>
    import pyroute2
/sonic/src/sonic-utilities/.eggs/pyroute2-0.6.1-py3.7.egg/pyroute2/__init__.py:46: in <module>
    import importlib_metadata as metadata
E   ModuleNotFoundError: No module named 'importlib_metadata'
```

Nothing I did should have touched that. Perhaps an infra issue?

robocoder99:
> 
> 
> This pull request **introduces 1 alert** when merging [a305336](https://github.com/Azure/sonic-buildimage/commit/a3053363082bf878ccde46271ca0e4ff85fa22e6) into [c646257](https://github.com/Azure/sonic-buildimage/commit/c6462577a9957fd7a6f41b681c8139e59bd30061) - [view on LGTM.com](https://lgtm.com/projects/g/Azure/sonic-buildimage/rev/pr-9151511baec0165f05cd22bc9e147059b690ff52)
> 
> **new alerts:**
> 
>     * 1 for Except block handles 'BaseException'

This is expected. The block of code calling this makes a redis db call which from my experience right now can return a plethora of errors (if the whole service is down, timeout etc.), I want this to nicely handle that situation and simply return the dictionary with as many values as it was able to successfully retrieve. Handling of not having this information is done by the caller.

src/sonic-py-common/sonic_py_common/device_info.py
```diff
@@ -32,6 +32,53 @@
 FRONTEND_ASIC_SUB_ROLE = "FrontEnd"
 BACKEND_ASIC_SUB_ROLE = "BackEnd"
 
+# Chassis STATE_DB keys
+CHASSIS_INFO_TABLE = 'CHASSIS_INFO|chassis {}'
+CHASSIS_INFO_CARD_NUM_FIELD = 'module_num'
+CHASSIS_INFO_SERIAL_FIELD = 'serial'
+CHASSIS_INFO_MODEL_FIELD = 'model'
+CHASSIS_INFO_REV_FIELD = 'revision'
+
```
jleveque, line 41: Per PEP8, there should be two blank lines before function definitions. Also suggest moving the new functions to the end of the section, mainly because `get_platform_info` relies on other functions. Even though Python isn't picky about definition order, I still prefer it. Suggest moving these functions directly above

```python
#
# Multi-NPU functionality
#
```

jleveque (changes requested):

src/sonic-py-common/sonic_py_common/device_info.py
robocoder99: Are there any linting tools built into the SONiC codebase as of now?

If not, I would highly suggest adding some linting tools to the push gate checks or adding a auto-formatter / checker to the repo if style guidelines are going to be enforced. PEP8 is a massive standard and knowing precisely what we want to standardize on will help developers push code that they know matches our guidelines.

jleveque:
Please fix LGTM alert.

robocoder99:
Did you see my earlier comment about the LGTM alert?

jleveque:
> Did you see my earlier comment about the LGTM alert?

If you need to catch all exceptions, please use `except Exception` rather than simply `except`. This is the preferred method, and it will resolve the LGTM alert.

jleveque:
Here's a little bit more about the difference: https://stackoverflow.com/a/18982754

robocoder99:
Thats very interesting, I was not aware of that distinction. Thanks!

jleveque:
> Thats very interesting, I was not aware of that distinction. Thanks!

I wasn't either until we began using the LGTM plugin. :)

src/sonic-py-common/sonic_py_common/device_info.py
jleveque: Currently we don't have linting tools built in. We're looking into it as a low priority. In the meantime, I suggest using the `autopep8` tool. Fee free to adjust the max line length to ~100-120 (e.g., `autopep8 --max-line-length 120 <file>`)

src/sonic-py-common/sonic_py_common/device_info.py
```diff
@@ -303,6 +310,47 @@ def get_sonic_version_file():
 
     return SONIC_VERSION_YAML_PATH
 
+
+# Get hardware information
+def get_platform_info():
```
jleveque, line 315: Just thinking that we have `get_platform()` and `get_platform_info()`. Hard to discern the difference by name. If we can't think of a more descriptive name, we can merge this.

robocoder99: I'm not sure there is a more descriptive method name possible given the purpose of the function is to provide various platform information that we have deemed "good to know" to the user. 

Alternates could be `get_platform_summary()` or `get_hardware_facts()` 

But unless you *love* any of these suggestions compared to what we have I suggest we keep with the current one as I don't see a large difference between these suggestions other than personal taste. 

Open to suggestions as well.

jleveque: I agree it's hard to think of much better. We can move ahead with this.

jleveque (approved):

PR #7663, opened 2021-05-21
[CI] Enable 201911 branch PR check

liushilongbuaa:
#### Why I did it
Enable azure pipeline PR check for 201911 branch.
#### How I did it
Add 201911 branch to trigger. But Test stage is failing now, so I commented the Test stage leaving a TODO. When Test stage passed, uncomment it.
#### How to verify it
In this PR, check is already enabled.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012


azure-pipelines.yml
```diff
@@ -17,6 +17,7 @@ pr:
     include:
       - master
       - 202012
+      - 201911
```
xumia, line 20: @lguohan , we need to add 201811 as well, right?

xumia:
@lguohan , do we need to add 201811 as well? Need to add a new commit in 201811 branch.

liat-grozovik (approved):

liat-grozovik:
@liushilongbuaa can you please check why it is not working?

azure-pipelines.yml
```diff
@@ -45,84 +46,83 @@ stages:
       - name: vs
       - name: broadcom
       - name: mellanox
-
-- stage: Test
```
xumia, line 49: If you want to disable the Test stage, please use.
condition: false

@qiluo-msft , do you think we can only enable the build without Test? It should be better than without any check in 2001911.

xumia (approved):

qiluo-msft (approved):

azure-pipelines.yml
qiluo-msft: Agree. At lease we have some build checker, better than nothing.

PR #7666, opened 2021-05-21
[build] Refactor device-data Makefile and fix bug

joe-tricklebank-owens:
#### Why I did it

There was a missing ; which meant the pai.profile processing wasn't happening

#### How I did it

Refactored the processing slightly to make extensions easier and fixed the missing ; in the process

### DIFF
--- a/src/sonic-device-data/Makefile
+++ b/src/sonic-device-data/Makefile
@@ -17,11 +17,13 @@ $(addprefix $(DEST)/, $(MAIN_TARGET)): $(DEST)/% :
 	# Create hwsku for virtual switch
 	for d in `find -L ../../../device -maxdepth 3 -mindepth 3 -type d | grep -vE "(plugins|led-code)"`; do \
 		cp -Lr $$d device/x86_64-kvm_x86_64-r0/ ; \
-		cp ./sai.vs_profile device/x86_64-kvm_x86_64-r0/$$(basename $$d)/sai.profile; \
-		grep -v ^# device/x86_64-kvm_x86_64-r0/$$(basename $$d)/port_config.ini | awk '{i=i+1;print "eth"i":"$$2}' > device/x86_64-kvm_x86_64-r0/$$(basename $$d)/lanemap.ini
-		cp ./pai.vs_profile device/x86_64-kvm_x86_64-r0/$$(basename $$d)/pai.profile; \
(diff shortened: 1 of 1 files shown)


src/sonic-device-data/Makefile
```diff
@@ -17,13 +17,15 @@ $(addprefix $(DEST)/, $(MAIN_TARGET)): $(DEST)/% :
...
 	for d in `find -L ../../../device -maxdepth 3 -mindepth 3 -type d | grep -vE "(plugins|led-code)"`; do \
 		cp -Lr $$d device/x86_64-kvm_x86_64-r0/ ; \
-		cp ./sai.vs_profile device/x86_64-kvm_x86_64-r0/$$(basename $$d)/sai.profile; \
-		grep -v ^# device/x86_64-kvm_x86_64-r0/$$(basename $$d)/port_config.ini | awk '{i=i+1;print "eth"i":"$$2}' > device/x86_64-kvm_x86_64-r0/$$(basename $$d)/lanemap.ini
-		cp ./pai.vs_profile device/x86_64-kvm_x86_64-r0/$$(basename $$d)/pai.profile; \
 		grep -v ^# device/x86_64-kvm_x86_64-r0/$$(basename $$d)/port_config.ini | awk '{i=i+1;print "eth"i":"$$2}' > device/x86_64-kvm_x86_64-r0/$$(basename $$d)/lanemap.ini
 
 	done;
 
+	for d in `find -L ../../../device -maxdepth 3 -mindepth 3 -type d | grep -vE "(plugins|led-code)"`; do \
```
jleveque, line 24: Seems redundant to perform the same loop twice. Why not refactor similar to the following?

```
	# Create hwsku for virtual switch
	for d in `find -L ../../../device -maxdepth 3 -mindepth 3 -type d | grep -vE "(plugins|led-code)"`; do \
		cp -Lr $$d device/x86_64-kvm_x86_64-r0/; \
		cp ./sai.vs_profile device/x86_64-kvm_x86_64-r0/$$(basename $$d)/sai.profile; \
		cp ./pai.vs_profile device/x86_64-kvm_x86_64-r0/$$(basename $$d)/pai.profile; \
		grep -v ^# device/x86_64-kvm_x86_64-r0/$$(basename $$d)/port_config.ini | awk '{i=i+1;print "eth"i":"$$2}' > device/x86_64-kvm_x86_64-r0/$$(basename $$d)/lanemap.ini; \
	done

```

joe-tricklebank-owens: We have an internal-only enhancement to make to the second loop that's not shown here (we don't want to copy over the VS sai/pai profile files for our specific device folder)

PR #7637, opened 2021-05-18
Support L1 & L3 Config generation in SONiC 

VenkatCisco:
#### Why I did it
This provides support for: PR #7074.

#### How I did it
Extend sonic-config-engine/config_samples.py to provide support for l1 & l3

#### How to verify it
Verify system bring up with L1 & L3 configuration

#### Which release branch to backport (provide reason below if selected)

- [ ] 201811
- [X] 201911
- [ ] 202006
- [X] 202012

PR #7691, opened 2021-05-24
[sonic-config-engine] Add support for generating a minimal configuration preset

joe-tricklebank-owens:
#### Why I did it

To allow a default_sku to be built using minimal configuration and remove superfluous, templated BGP configuration.

#### How I did it

Added a new function, _generate_minimal_config()_, that can be referenced by the default_sku file.

### DIFF
--- a/src/sonic-config-engine/config_samples.py
+++ b/src/sonic-config-engine/config_samples.py
@@ -40,6 +40,18 @@ def generate_t1_sample_config(data):
         port_count += 1
     return data
 
+def generate_minimal_config(data):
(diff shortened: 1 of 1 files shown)


anamehra:
There is a similar PR https://github.com/Azure/sonic-buildimage/pull/7637

joe-tricklebank-owens:
> There is a similar PR #7637

Ah, indeed! Thanks for pointing this out, I think your L3 config gives us what we need, so I'll close down this PR.  Thanks!

PR #7692, opened 2021-05-24
[build] For SONIC_MAKE_DEBS targets, call autogen.sh if it exists in the project's root directory

joe-tricklebank-owens:
#### Why I did it

Some projects require auto-generation of files to be run before they can be compiled.  This enhancement adds the ability to do that, if required.

#### How I did it

If it exists for the project, added a call to _autogen.sh_ before invoking the make command for the project.


jleveque (approved):

qiluo-msft:
Could you give an example of which project is requiring this functionality? Is it possible to implement it in the Makefile in the source folder?

joe-tricklebank-owens:
> Could you give an example of which project is requiring this functionality? Is it possible to implement it in the Makefile in the source folder?

@qiluo-msft unfortunately the project we're working on isn't open sourced, so I can't really reference it.  Suffice it to say, due to the complexity of the project, it can't be built into the Makefile system easily.  It's also worth saying that there's a similar call in the SONIC_DPKG_DEBS rule, so we're just replicating that functionality

qiluo-msft:
I feel weird that "it can't be built into the Makefile system easily". `autogen.sh` is not a well-known component for a Makefile controlled project, so I recommend not to add new feature at this general place.

PR #7658, opened 2021-05-20
[docker-database] Fix python version in shebang

msosyak:
#### Why I did it
To avoid the following logs 
```
Mar 15 15:52:04.599302 igk-dut-04 INFO database#/supervisord: flushdb /bin/bash: /usr/local/bin/flush_unused_database: /usr/bin/python: bad interpreter: No such file or directory
Mar 15 15:52:04.599947 igk-dut-04 INFO database#supervisord 2021-03-15 15:52:04,599 INFO exited: flushdb (exit status 126; not expected)
```

#### How I did it
Fix  shebang
#### How to verify it
Check the logs
#### Which release branch to backport (provide reason below if selected)

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012

PR #7700, opened 2021-05-25
[docker-database] Fix Python3 issue

msosyak:
#### Why I did it
To avoid the following error
```
  File "/usr/local/bin/flush_unused_database", line 10, in <module>
    if 'PONG' in output:
TypeError: a bytes-like object is required, not 'str'
```
`communicate` method returns the strings if streams were opened in text mode; otherwise, bytes.
In our case text arg  in Popen is not true and that means that `communicate` return the bytes
#### How I did it
Set `text=True` to get strings instead of bytes
#### How to verify it
run `/usr/local/bin/flush_unused_database` inside database container

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012

### DIFF
--- a/dockers/docker-database/flush_unused_database
+++ b/dockers/docker-database/flush_unused_database
@@ -6,7 +6,7 @@ import time
 
 while(True):
     output = subprocess.Popen(['sonic-db-cli', 'PING'], stdout=subprocess.PIPE).communicate()[0]
-    if 'PONG' in output:
+    if b'PONG' in output:
         break
     time.sleep(1)
 
--- a/dockers/docker-database/flush_unused_database
+++ b/dockers/docker-database/flush_unused_database
@@ -5,8 +5,8 @@ import subprocess
 import time
 
 while(True):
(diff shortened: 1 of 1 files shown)


msosyak:
@qiluo-msft I have missed one more fix in the previous PR https://github.com/Azure/sonic-buildimage/pull/7658. Check it, please.

dockers/docker-database/flush_unused_database
```diff
@@ -6,7 +6,7 @@ import time
 
 while(True):
     output = subprocess.Popen(['sonic-db-cli', 'PING'], stdout=subprocess.PIPE).communicate()[0]
```
qiluo-msft, line 8: >Popen [](http://example.com/codeflow?start=24&length=5)

Thanks for the fix! I see the `text=True` could also fix this, and it is more readable. Could you instead use that optional parameter?

msosyak: Done. Description updated

akokhan (approved):

qiluo-msft (approved):

PR #7702, opened 2021-05-25
[CI] Adjust azp build artifacts and platform to keep same with Jenkins

liushilongbuaa:
#### Why I did it
Adjust azp official build artifacts and platform to keep same with Jenkins.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012


abdosi (approved):

xumia:
@liushilongbuaa ,
Do we need to change the trigger as well?
We remove some of the platforms, but the trigger not changed, it will be failed with no job to run, although it is not a big issue.

xumia (approved):

Issue #7683, opened 2021-05-24
The "debian" folder of a newly added platform driver is ignored to be uploaded.


Komil-yc:
#### Description

    There is a pattern "platform/**/debian/*" described in .gitignore.
That causes GIT to ignore all files in debian/ used to build Debian package. 
A newly added platform will miss debian/ folder to finish their building. 
The change is introduced by below PR:
      https://github.com/Azure/sonic-buildimage/pull/4803
    
Is that an expected behavior to ignore debian/ folder automatically and
should we upload them by another way like "git add -f" ?
#### Steps to reproduce the issue:
1.
2.
3.

#### Output of `show version`:

```
(paste your output here)
```
#### Output of `show techsupport`:

```
(paste your output here or download and attach the file here )
```

anshuv-mfst:
Hi @jleveque - could you please take a look, thanks.

jleveque:
I agree that this ignore is a bit severe, and could cause vendors to miss adding their debian directory. I will open a PR to remove this line. However, I think each vendor should be responsible for providing one or more `.gitignore` files to properly ignore generated files in their platform directories.

bluecmd:
I created a PR to mitigate this issue as I lost a few hours today because of files being hidden from `git status`; https://github.com/sonic-net/sonic-buildimage/pull/14565

PR #7720, opened 2021-05-26
[.gitignore] Do not blanket-ignore all  `debian` directories

jleveque:
#### Why I did it

Having a blanket statement to ignore all files under `platform/**/debian/` can pose issues when a vendor adds a new platform, as any new files under a `debian/` directory will not show up in a Git changelist, therefore they can be missed in the commit.

Resolves https://github.com/Azure/sonic-buildimage/issues/7683

#### How I did it

Remove the blanket ignore statement from .gitignore. Platform vendors should be responsible for providing one or more .gitignore files to properly ignore generated files in their platform directories.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,7 +59,6 @@ src/**/*.o
 platform/**/*.egg-info
 platform/**/*-none-any.whl
 platform/**/.pybuild
-platform/**/debian/*
 platform/**/build
 platform/**/*.ko
 platform/**/*.mod.c


lguohan:
how can we enforce the vendor will be adding the debian in the .gitignore?

jleveque:
> how can we enforce the vendor will be adding the debian in the .gitignore?

I don't have a good idea to enforce this. Any suggestions are welcome.

PR #7721, opened 2021-05-26
[Mellanox] Implement auto_firmware_update platform API for to support fwutil auto-update

robocoder99:
#### Why I did it
The Mellanox platform is required to support the `fwutil auto-update` feature defined [here ](https://github.com/sujinmkang/SONiC/blob/357485991d768a9fa78873bb083e3979fbc5cf71/doc/fwutil/fwutil.md#2224-update-all-commands)

This is to allow switches, when performing SONiC upgrades to choose whether to perform firmware upgrades that may interrupt the data plane through a cold boot. 

#### How I did it
Two methods were added to the component implementations for mellanox. 

In the base `Component` class we add a default function that chooses to skip the installation of any firmware unless the `cold` boot option is provided. This is because the Mellanox platform, by default, does not support installing firmware on ONIE, the CPLD, or the BIOS "on-the-fly". 

In the `ComponentSSD` class we add a function that behaves similarly but uses the Mellanox specific SSD firmware upgrade tool to check if the current SSD supports being upgraded on the fly in order to decide whether to skip or perform the installation.

#### How to verify it
Unit tests are included with this PR. These test will run on build of target `sonic-mellanox.bin` 

You may also perform `fwutil auto-update ...` commands after https://github.com/Azure/sonic-utilities/pull/1242 is merged in.  

[platform][mellanox] Implement auto_firmware_update for mellanox 

Great_Pyrenees_Sheep_Dog_Guarding_the_Flock_5113678413

### DIFF
--- a/platform/mellanox/mlnx-platform-api/sonic_platform/component.py
+++ b/platform/mellanox/mlnx-platform-api/sonic_platform/component.py
@@ -20,10 +20,10 @@ try:
     else:
         import ConfigParser as configparser
 
-    from sonic_platform_base.component_base import ComponentBase,
-                                                    FW_AUTO_UPDATED,
-                                                    FW_AUTO_ERR_BOOT_TYPE,
-                                                    FW_AUTO_ERR_IMAGE,
+    from sonic_platform_base.component_base import ComponentBase,           \
+                                                    FW_AUTO_UPDATED,        \
+                                                    FW_AUTO_ERR_BOOT_TYPE,  \
+                                                    FW_AUTO_ERR_IMAGE,      \
                                                     FW_AUTO_ERR_UKNOWN
 except ImportError as e:
     raise ImportError(str(e) + "- required module not found")
--- a/platform/mellanox/mlnx-platform-api/sonic_platform/component.py
+++ b/platform/mellanox/mlnx-platform-api/sonic_platform/component.py
@@ -20,7 +20,11 @@ try:
     else:
         import ConfigParser as configparser
 
-    from sonic_platform_base.component_base import ComponentBase
+    from sonic_platform_base.component_base import ComponentBase,
+                                                    FW_AUTO_UPDATED,
+                                                    FW_AUTO_ERR_BOOT_TYPE,
+                                                    FW_AUTO_ERR_IMAGE,
+                                                    FW_AUTO_ERR_UKNOWN
 except ImportError as e:
     raise ImportError(str(e) + "- required module not found")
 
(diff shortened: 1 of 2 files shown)


liat-grozovik:
@robocoder99 can you confirm submodule is updated? then we can rerun the tests of this PR.

platform/mellanox/mlnx-platform-api/sonic_platform/component.py
```diff
@@ -469,6 +497,40 @@ def __install_firmware(self, image_path):
...
+                # No power cycle required
+                supported_boot += ['warm', 'fast', 'none', 'any']
+        except RuntimeError:
+            # Unknown error from firmware probe
+            return FW_AUTO_ERR_UKNOWN
+
+        if boot_action in supported_boot:
+            if self.install_firmware(image_path):
+                # Successful update
+                return FW_AUTO_UPDATED
```
sujinmkang, line 527: Could it be more like below? If it doesn't need any complete action after installation, then it could be updated otherwise, it should return the installed status as we defined in hld.
```
 if self.get_firmware_update_notification(image_path) is None: 
    return FW_ATUO_UPDATED
else:
    return FW_AUTO_INSTALLED 
```

robocoder99: No mellanox components need action at or after boot. We always do a full install and then some type of boot (warm / cold / power cycle) in all cases. 

Due to this we either install completely just prior to the reboot or we don't install at all, greatly simplifying this logic.

sujinmkang: I think if the reboot needs to be followed then the return should be "installed" not "updated".
"updated" means no additional action needed for the completion of the firmware update.

robocoder99: Can you please point out to me where this is documented? I cannot seem to find it anywhere. I will make an update.

robocoder99: @sujinmkang Is the updated vs installed flag used to do any logic? I'm a little confused at what the purpose is. For example, should there ever be an issue if I always pass installed as the return code so that the intended reboot is always followed?

robocoder99:
@sujinmkang Can you please review?

platform/mellanox/mlnx-platform-api/sonic_platform/component.py
sujinmkang: Please 2.2.2.4 update all commands at https://github.com/Azure/SONiC/blob/master/doc/fwutil/fwutil.md

sujinmkang (approved):

PR #7731, opened 2021-05-27
Support to build armhf/arm64 platforms on arm based system

xumia:
#### Why I did it
Support to build armhf/arm64 platforms on arm based system without qemu simulator.
When building the armhf/arm64 on arm based system, it is not necessary to use qemu simulator.

#### How I did it
Build armhf on armhf system, or build arm64 on arm64 system, by default, qemu simulator will not be used.
When building armhf on arm64, and you have enabled armhf docker, then it will build images without simulator automatically. It is based how the docker service is run.

Docker base image change:
For amd64, change from debian:<os-release>to amd64/debian:<os-release>
For arm64, change from multiarch/debian-debootstrap:arm64-<release> to arm64v8/debian:<os-release>
For armhf, change from multiarch/debian-debootstrap:armhf-<release> to arm32v7/debian:<os-release>
See https://github.com/docker-library/official-images#architectures-other-than-amd64
The mapping relations:
arm32v6  --- armel
arm32v7 --- armhf
arm64v8 --- arm64

Docker image armhf deprecated info: https://hub.docker.com/r/armhf/debian, using arm32v7 instead.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012

### DIFF
--- a/Makefile.work
+++ b/Makefile.work
@@ -47,12 +47,13 @@ SHELL = /bin/bash
 USER := $(shell id -un)
 PWD := $(shell pwd)
 USER_LC := $(shell echo $(USER) | tr A-Z a-z)
+DOCKER_MACHINE := $(shell docker run --rm debian:buster uname -m)
 
 comma := ,
 
-ifeq ($(shell uname -m), aarch64)
+ifeq ($(DOCKER_MACHINE), aarch64)
 COMPILE_HOST_ARCH := arm64
-else ifeq ($(shell uname -m), armv7l)
+else ifeq ($(shell echo $(DOCKER_MACHINE) | grep -qE "armv7l|armv8l" && echo y),y)
 COMPILE_HOST_ARCH := armhf
 else
 COMPILE_HOST_ARCH := amd64
@@ -286,7 +287,7 @@ SONIC_BUILD_INSTRUCTION :=  make \
 .DEFAULT_GOAL :=  all
 
 %::
-ifneq (,$(filter $(CONFIGURED_ARCH), armhf arm64))
+ifeq ($(MULTIARCH_QEMU_ENVIRON), y)
 	@$(DOCKER_MULTIARCH_CHECK)
 ifneq ($(BLDENV), )
 	@$(DOCKER_SERVICE_MULTIARCH_CHECK)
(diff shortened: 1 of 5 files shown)


Makefile.work
```diff
@@ -51,12 +51,14 @@ comma := ,
...
 ifeq ($(shell uname -m), aarch64)
 COMPILE_HOST_ARCH := arm64
-else ifeq ($(shell uname -m), armv7l)
+else ifneq (,$(filter $(shell uname -m),armv7l armv8l))
 COMPILE_HOST_ARCH := armhf
 else
 COMPILE_HOST_ARCH := amd64
 endif
 
+COMPILE_HOST_ARCHS := $(COMPILE_HOST_ARCH) $(HOST_ARCHS)
```
qiluo-msft, line 60: >HOST_ARCHS [](http://example.com/codeflow?start=45&length=10)

not defined?

Makefile.work
```diff
@@ -51,12 +51,14 @@ comma := ,
 
 ifeq ($(shell uname -m), aarch64)
 COMPILE_HOST_ARCH := arm64
-else ifeq ($(shell uname -m), armv7l)
+else ifneq (,$(filter $(shell uname -m),armv7l armv8l))
```
qiluo-msft, line 54: >ifneq [](http://example.com/codeflow?start=5&length=5)

Is it possible to test postively instead of negatively?

Makefile.work
```diff
@@ -111,18 +113,13 @@ ifeq ($(ENABLE_DOCKER_BASE_PULL),)
 	override ENABLE_DOCKER_BASE_PULL = n
 endif
 
-ifeq ($(CONFIGURED_ARCH),amd64)
+ifneq (,$(filter $(CONFIGURED_ARCH),amd64 $(COMPILE_HOST_ARCHS)))
```
qiluo-msft, line 116: Could you add some comment for complex condition check?

dockers/docker-base/Dockerfile.j2
```diff
@@ -1,6 +1,6 @@
-{% if CONFIGURED_ARCH == "armhf" and MULTIARCH_QEMU_ENVIRON == "y" %}
```
qiluo-msft, line 1: >MULTIARCH_QEMU_ENVIRON [](http://example.com/codeflow?start=37&length=22)

If removed, do we remove support on any platform?

Makefile.work
xumia: Changed to use ifeq, thanks.

dockers/docker-base/Dockerfile.j2
xumia: For armhf, we can always use multiarch/debian-debootstrap:armhf-jessie, it should not be relative to MULTIARCH_QEMU_ENVIRON yes or not.

Makefile.work
qiluo-msft: I search the repo and could not found `HOST_ARCHS`

xumia: Yes, it is not defined, normally, we are not necessary to define it, the default value is empty.
If we want to support build armhf on arm64, we can set HOST_ARCHS=armhf, it means that the architecture armhf is supported on arm64, although it does not match with "uname -m".

qiluo-msft: OK, this paragraph is good code comment, could you add it? Otherwise user could not easily figure out this hidden feature.

lguohan:
what about building arm64 on armhf? that is probably not supported, correct?

lguohan (dismissed):

xumia:
> what about building arm64 on armhf? that is probably not supported, correct?

It is not supported, it needs to use only qemu vm to support.

xumia:
@qiluo-msft , I do an improvement to detect the docker environment in the build time to decide whether the qemu simulator needed. When the docker arch of the build environment is the same as the build target, then no simulator required.

lguohan (approved):

qiluo-msft (approved):

PR #7772, opened 2021-06-01
[dhcp_relay] DHCP relay support for IPv6

shlomibitton:
#### Why I did it
Currently SONiC use the 'isc-dhcp-relay' package to allow DHCP relay functionality on IPv4 networks only.
This will allow the IPv6 functionality along the IPv4 type.

#### How I did it
Edit supervisord template to start DHCPv6 instances when configured to do so on Config DB.
Align cfg unit test to the new change.
Add DHCPv6 relay minigraph parsing support and a suitable t0 topology xml file for UT.

#### How to verify it
Configure DHCPv6 agents as described on the feature HLD: https://github.com/Azure/SONiC/pull/765
Test it with real client/server with IPv6 or use the dedicated automatic test: https://github.com/Azure/sonic-mgmt/pull/3565

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [x] 202106


santoshdoke (approved):

src/dhcpmon/src/dhcp_device.c
```diff
@@ -152,6 +196,45 @@ static void handle_dhcp_option_53(dhcp_device_context_t *context,
...
+    case DHCPv6_MESSAGE_TYPE_CONFIRM:
+    case DHCPv6_MESSAGE_TYPE_RENEW:
+    case DHCPv6_MESSAGE_TYPE_REBIND:
+    case DHCPv6_MESSAGE_TYPE_RELEASE:
+    case DHCPv6_MESSAGE_TYPE_DECLINE:
+    case DHCPv6_MESSAGE_TYPE_ADVERTISE:
+    case DHCPv6_MESSAGE_TYPE_REPLY:
+    case DHCPv6_MESSAGE_TYPE_RECONFIGURE:
+    case DHCPv6_MESSAGE_TYPE_INFORMATION_REQUEST:
+    case DHCPv6_MESSAGE_TYPE_RELAY_FORWARD:
```
tahmed-dev, line 227: can we instead of counting relay forward and relay reply to count those packet against the underlying type with dir set as :
relay-forward: TX dir of client packets
relay-reply: RX dir of server packets.

src/dhcpmon/src/dhcp_device.c
```diff
@@ -254,34 +386,90 @@ static bool dhcp_device_is_dhcp_inactive(uint64_t counters[][DHCP_DIR_COUNT][DHC
...
+ */
+static bool dhcp_device_is_dhcpv6_msg_unhealthy(dhcpv6_message_type_t type,
+                                                uint64_t v6counters[][DHCP_DIR_COUNT][DHCPv6_MESSAGE_TYPE_COUNT])
+{
+    // check if DHCP message 'type' is being relayed
+    switch (type) {
+        case DHCPv6_MESSAGE_TYPE_SOLICIT:
+        case DHCPv6_MESSAGE_TYPE_REQUEST:
+            return ((v6counters[DHCP_COUNTERS_CURRENT][DHCP_RX][type] > v6counters[DHCP_COUNTERS_SNAPSHOT][DHCP_RX][type]) &&
+                    (v6counters[DHCP_COUNTERS_CURRENT][DHCP_TX][DHCPv6_MESSAGE_TYPE_RELAY_FORWARD] <= v6counters[DHCP_COUNTERS_SNAPSHOT][DHCP_TX][DHCPv6_MESSAGE_TYPE_RELAY_FORWARD]));
```
tahmed-dev, line 416: once we have the relay forward counted as TX, this condition should be simplified.

src/dhcpmon/src/dhcp_device.c
```diff
@@ -254,34 +386,90 @@ static bool dhcp_device_is_dhcp_inactive(uint64_t counters[][DHCP_DIR_COUNT][DHC
...
+
+        case DHCPv6_MESSAGE_TYPE_RELAY_REPLY:
```
tahmed-dev, line 418: I believe the whole `switch` case could be eliminate.

tahmed-dev (changes requested):
Nice work, thanks!

Can we split this PR into 3 as follows:

1. Changes related to graph parsing
2. Changes related to ISC DHCPv6 params + required cfggen change to pass the test
3. Changes related to dhcpmon + ditto for cfggen
3.

src/dhcpmon/src/dhcp_device.c
shlomibitton: The current implementation is counting with the direction.
On the condition to check if health status is ok, we check if solicit/request packets received and then check if relay-forward packet transmitted, can you elaborate how is it possible to simplify this check?

src/dhcpmon/src/dhcp_device.c
shlomibitton: can you explain why we can eliminate the switch case here?
The check is dependent on the packet type, for solicit/request packets we need to check relay-forward packet.
For relay-reply packet we need to check advertise/reply packet.
Are you agree or have a better idea?

shlomibitton:
> Nice work, thanks!
> 
> Can we split this PR into 3 as follows:
> 
> 1. Changes related to graph parsing
> 2. Changes related to ISC DHCPv6 params + required cfggen change to pass the test
> 3. Changes related to dhcpmon + ditto for cfggen

@tahmed-dev I divided this PR into 3:
https://github.com/Azure/sonic-buildimage/pull/7772
https://github.com/Azure/sonic-buildimage/pull/8060
https://github.com/Azure/sonic-buildimage/pull/8059

src/sonic-config-engine/minigraph.py
```diff
@@ -555,6 +555,12 @@ def parse_dpg(dpg, hname):
                 vdhcpserver_list = vintfdhcpservers.split(';')
                 vlan_attributes['dhcp_servers'] = vdhcpserver_list
 
+            vintf_node = vintf.find(str(QName(ns, "Dhcpv6Relays")))
+            if vintf_node is not None and vintf_node.text is not None:
```
tahmed-dev, line 559: I'd still put code chunk into #8059 . It is better to have each PR to minimal independent functional code changes.

shlomibitton: OK, will re open the other PR, but this PR can't pass all checks if the other is nor merged.

dockers/docker-dhcp-relay/docker-dhcp-relay.supervisord.conf.j2
```diff
@@ -102,33 +115,88 @@ stderr_logfile=syslog
 dependent_startup=true
 dependent_startup_wait_for=start:exited
 
+{% endif %}
```
tahmed-dev, line 118: Have tried Jinja2 include [directive](https://jinja.palletsprojects.com/en/3.0.x/templates/#include)? I guess it would make sense to split the file into three components; 1) DHCPv4 Relay, 2) DHCPv6 Relay, and 3) DHCP MONitor. However this may not be related, we can at least have the DHCPv6 Relay in a separate file.

src/dhcpmon/src/dhcp_device.c
shlomibitton: Addressed on a different PR.

src/dhcpmon/src/dhcp_device.c
shlomibitton: Addressed on a different PR.

src/dhcpmon/src/dhcp_device.c
shlomibitton: Addressed on a different PR.

dockers/docker-dhcp-relay/docker-dhcp-relay.supervisord.conf.j2
shlomibitton: @tahmed-dev if you are referring to the test failure, the DHCP MONitor entries are common to both DHCP types, the added flags and 'dependent_startup_wait_for' values are dependent on the DHCP servers IP type configured on the xml file and parsed by the minigraph parser.
Even if we separate it into 3 files the test will still fail without the correct xml and parser support for this type.
Anyway, I think separating it into 3 files will make it more complex, the current implementation is simple and clear.
What about merging this PR https://github.com/Azure/sonic-buildimage/pull/8125 and then re-trigger this one?

tahmed-dev: @shlomibitton my point is related to code maintenance and have changes in a self-contained file and is not related to functionality. It just the file grows really big with repeated code chunks. I was hoping to find a way to parameterize the file so it would boil down to single file with either IPv4/6 as aparameter, however I am not expert in Jinja2.

shlomibitton: @tahmed-dev can you check it now? I divided the code to several files to make it easier to maintenance.

shlomibitton:
Test failure is not related to this change:
stderr_lines: WARNING: No route found for IPv6 destination

dockers/docker-dhcp-relay/dhcpv4-relay.agents.j2
```diff
@@ -0,0 +1,40 @@
...
+priority=3
+autostart=false
+autorestart=false
+stdout_logfile=syslog
+stderr_logfile=syslog
+dependent_startup=true
+dependent_startup_wait_for=start:exited
+
+{% endif %}
+{% endif %}
```
tahmed-dev, line 40: nit: missing eol here and in other files.

tahmed-dev (approved):
LGTM, just nit related to eol at the end of each new file.

dockers/docker-dhcp-relay/dhcpv4-relay.agents.j2
shlomibitton: Fixed

shlomibitton:
> LGTM, just nit related to eol at the end of each new file.

@tahmed-dev I fixed all EOL for all files, please check

tahmed-dev (approved):

qiluo-msft:
This PR could not be cleanly cherry-pick to 202012. Please submit another PR.

PR #7789, opened 2021-06-03
[dhcp_relay] Disable dhcp_relay for ToRRouter switches type by the feature manager

shlomibitton:

#### Why I did it
Currently dhcp packets are disabled by the COPP manager for non ToRRouter type switches.
Even if the feature is enabled, DHCP packets wont hook to the CPU since the COPP manager will not trap this packets.
This change is to disable dhcp_relay by default for non ToRRouter switches from init_cfg.json.
With this approach, if the user want to enable the feature for non ToRRouter switches, manual enablement is required by the 'feature' configuration.
This is to keep the current approach for MSFT production issue with dhcp relay for non ToRRouter switched and allow the user to decide if to use it or not.

#### How I did it
Configure dhcp_relay 'disabled' by default on init_cfg.json for non ToRRouter switches.
Remove the exclusion of dhcp packets on copp_cfg.json

#### How to verify it
- Enable dhcp_relay feature on a non ToRRouter switch.
- Unit-tests modified so the default values on mocked CONFIG DB in 'test_vectors.py' for dhcp_relay will be 'disabled'.
This is by the change for 'init_cfg.json.j2'.
For ToRRouter the state will change from 'disabled' to 'enabled'.
Another test case added for a 'ToR' switch type, this is to test the state is 'enabled' if the user configured it to be so.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012

### DIFF
--- a/files/image_config/copp/copp_cfg.j2
+++ b/files/image_config/copp/copp_cfg.j2
@@ -79,12 +79,10 @@
 		    "trap_ids": "lldp",
 		    "trap_group": "queue4_group3"
 	    },
-{% if not (DEVICE_METADATA is defined and DEVICE_METADATA['localhost'] is defined and DEVICE_METADATA['localhost']['type'] is defined and DEVICE_METADATA['localhost']['type'] != "ToRRouter") %}
-	    "dhcp": {
+	    "dhcp_relay": {
 		    "trap_ids": "dhcp,dhcpv6",
 		    "trap_group": "queue4_group3"
 	    },
-{% endif %}
 	    "udld": {
 		    "trap_ids": "udld",
 		    "trap_group": "queue4_group3"
--- a/src/sonic-host-services/tests/hostcfgd/test_vectors.py
+++ b/src/sonic-host-services/tests/hostcfgd/test_vectors.py
@@ -29,7 +29,7 @@ HOSTCFGD_TEST_VECTOR = [
                         "has_timer": "False",
                         "high_mem_alert": "disabled",
                         "set_owner": "kube",
-                        "state": "enabled"
+                        "state": "{% if not (DEVICE_METADATA is defined and DEVICE_METADATA['localhost'] is defined and DEVICE_METADATA['localhost']['type'] is defined and DEVICE_METADATA['localhost']['type'] != 'ToRRouter') %}enabled{% else %}disabled{% endif %}"
                     },
                     "mux": {
                         "auto_restart": "enabled",
@@ -125,7 +125,7 @@ HOSTCFGD_TEST_VECTOR = [
                         "has_timer": "False",
                         "high_mem_alert": "disabled",
                         "set_owner": "kube",
-                        "state": "enabled"
+                        "state": "{% if not (DEVICE_METADATA is defined and DEVICE_METADATA['localhost'] is defined and DEVICE_METADATA['localhost']['type'] is defined and DEVICE_METADATA['localhost']['type'] != 'ToRRouter') %}enabled{% else %}disabled{% endif %}"
                     },
                     "mux": {
                         "auto_restart": "enabled",
@@ -157,7 +157,7 @@ HOSTCFGD_TEST_VECTOR = [
                         "has_timer": "False",
                         "high_mem_alert": "disabled",
                         "set_owner": "kube",
-                        "state": "enabled"
+                        "state": "disabled"
                     },
                     "mux": {
                         "auto_restart": "enabled",
@@ -181,9 +181,6 @@ HOSTCFGD_TEST_VECTOR = [
                 },
             },
             "expected_subprocess_calls": [
-                call("sudo systemctl unmask dhcp_relay.service", shell=True),
-                call("sudo systemctl enable dhcp_relay.service", shell=True),
-                call("sudo systemctl start dhcp_relay.service", shell=True),
                 call("sudo systemctl stop mux.service", shell=True),
                 call("sudo systemctl disable mux.service", shell=True),
                 call("sudo systemctl mask mux.service", shell=True),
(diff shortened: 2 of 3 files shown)


liat-grozovik:
@dgsudharsan i am unable to assign you as a reviewer but appreciate if you can review and add comments.

shlomibitton:
@liat-grozovik he did the internal review and approve it :)

src/sonic-host-services/scripts/hostcfgd
```diff
@@ -793,6 +793,12 @@ class HostConfigDaemon:
                 syslog.syslog(syslog.LOG_WARNING, "Enable state of feature '{}' is None".format(feature_name))
                 continue
 
+            # dhcp_relay WA for production issue
```
liat-grozovik, line 796: what do you mean by WA? to ensure upgrade and backwards compatibility?

liat-grozovik (commented):
LGTM beside of the small comment which is mostly documentation of a so called WA

src/sonic-host-services/scripts/hostcfgd
shlomibitton: MSFT have a production issue, they want only ToRRouter switches to enable dhcp relay service.
The previous approach was to disable the feature by not hooking dhcp packets, this approach allow the user to enable it if it is configured to do so but still address the MSFT requirement to enable it by default only for ToRRouter.
If it is not a ToRRouter the user will have to override the configuration to enable it.

prsunny:
lgtm, did you test if rule gets added if the feature is enabled at a later stage? @dgsudharsan to review

shlomibitton:
> lgtm, did you test if rule gets added if the feature is enabled at a later stage? @dgsudharsan to review

Yes, for ToRRouter it will be enabled by default, for other types it will be disabled by default.
But, if the user enabled the feature with a non ToRRouter switch it will stay enabled.
I added a UT cases to verify it as well.

liat-grozovik (dismissed):

dgsudharsan (approved):

liat-grozovik:
@prsunny is it ok to merge?

files/image_config/copp/copp_cfg.j2
```diff
@@ -79,12 +79,10 @@
 		    "trap_ids": "lldp",
 		    "trap_group": "queue4_group3"
 	    },
-{% if not (DEVICE_METADATA is defined and DEVICE_METADATA['localhost'] is defined and DEVICE_METADATA['localhost']['type'] is defined and DEVICE_METADATA['localhost']['type'] != "ToRRouter") %}
-	    "dhcp": {
+	    "dhcp_relay": {
```
tahmed-dev, line 82: There is a requirement to not trap those DHCP packet on non-ToR deployment. This is a change is requirement. I'd suggest to have copp_mgr adapt and support his requirement.

src/sonic-host-services/scripts/hostcfgd
```diff
@@ -793,6 +793,12 @@ class HostConfigDaemon:
...
+            if feature_name == 'dhcp_relay':
+                device_metadata = self.config_db.get_table('DEVICE_METADATA')
+                if not (device_metadata and device_metadata['localhost'] and device_metadata['localhost']['type'] and device_metadata['localhost']['type'] != "ToRRouter"):
```
tahmed-dev, line 799: Please move this to state field in the init_cfg.json similar to `mux` feature in the test cases.

src/sonic-host-services/tests/hostcfgd/test_vectors.py
```diff
@@ -207,6 +204,96 @@
...
+                        "state": "disabled"
+                    },
+                    "mux": {
+                        "auto_restart": "enabled",
+                        "has_global_scope": "True",
+                        "has_per_asic_scope": "False",
+                        "has_timer": "False",
+                        "high_mem_alert": "disabled",
+                        "set_owner": "local",
+                        "state": "{% if 'subtype' in DEVICE_METADATA['localhost'] and DEVICE_METADATA['localhost']['subtype'] == 'DualToR' %}enabled{% else %}always_disabled{% endif %}"
```
tahmed-dev, line 224: Here is example of `mux` feature.

tahmed-dev (changes requested):

files/image_config/copp/copp_cfg.j2
dgsudharsan: This will be handled in hostcfgd where dhcp_relay feature will be disabled on non ToR deployment

tahmed-dev: CPU trap will be on T1/T2 deployment, right? It is now up to ctrl plane to forward those packet. This is a change in behavior.

dgsudharsan: CPU trap will be dynamically enabled based on feature. If a feature is disabled, then the corresponding trap will not be installed. COPP manager handles this based on the fact that feature table and copp table have same keys. That's the reason we changed dhcp keyword in copp file to dhcp_relay to match the feature, thereby making it to listen to feature notifications and installing trap only if enabled

tahmed-dev: Can you please point me to this code? I was not aware of that.

dgsudharsan: https://github.com/Azure/sonic-swss/blob/a3bc89bf3cff4bdf2a6ba80a0f4dd9fedb8f65ba/cfgmgr/coppmgr.cpp#L56

tahmed-dev: Thanks @dgsudharsan!

liat-grozovik:
@shlomibitton please help to resolve conflicts due to recent changes in hostcfgd.

src/sonic-host-services/scripts/hostcfgd
shlomibitton: I tried to do so on init_cfg.json but this is not possible since this file is used on build stage and the value of "device_metadata['localhost']['type']" is not exist, so the approach here is to disable dhcp_relay by default.
hostcfgd will check on runtime if this is a ToRRouter switch and enable it accordingly.
If not, it will be disabled as desired, but if the user explicitly request to enable it, this will be accepted as well.

src/sonic-host-services/tests/hostcfgd/test_vectors.py
shlomibitton: Since the dhcp_relay is disabled by default from init_cfg.json there is no reason to add this condition, it will be always disabled.
The UT should check if ToRRouter switched to 'enabled' state by hostcfgd, and if a non ToRRouter switch, which a user manually enabled, will stay enabled.

shlomibitton:
> @shlomibitton please help to resolve conflicts due to recent changes in hostcfgd.

@liat-grozovik resolved the conflicts

src/sonic-host-services/scripts/hostcfgd
tahmed-dev: The content of the state will be evaluated when loading the init_cfg.j2 for the first time. Also, the build J2 template will not render contents inside `append` call. Something like the following:

```
{% do features.append(("dhcp_relay", "{% if 'type' in DEVICE_METADATA['localhost'] and DEVICE_METADATA['localhost']['type'] == 'ToRRouter' %}enabled{% else %}disabled{% endif %}", false, "enabled")) %}
```

Then we can remove the change in `hostcfgd` all together.

src/sonic-host-services/tests/hostcfgd/test_vectors.py
tahmed-dev: The point is to move the code from `hostcfg` to `init_cfg.json.j2`

src/sonic-host-services/scripts/hostcfgd
shlomibitton: @tahmed-dev can you please check the changes now?

src/sonic-host-services/tests/hostcfgd/test_vectors.py
```diff
@@ -213,6 +210,99 @@
...
                 },
+                "FEATURE": {
+                    "dhcp_relay": {
+                        "auto_restart": "enabled",
+                        "has_global_scope": "True",
+                        "has_per_asic_scope": "False",
+                        "has_timer": "False",
+                        "high_mem_alert": "disabled",
+                        "set_owner": "kube",
+                        "state": "{% if not (DEVICE_METADATA is defined and DEVICE_METADATA['localhost'] is defined and DEVICE_METADATA['localhost']['type'] is defined and DEVICE_METADATA['localhost']['type'] != 'ToRRouter') %}enabled{% else %}disabled{% endif %}"
```
tahmed-dev, line 221: why use -ve logic? curious.

tahmed-dev (approved):

PR #7794, opened 2021-06-04
[barefoot][platform] Chassis.get_reboot_cause

vboykox:

#### Why I did it
To fix determine-reboot-cause service which was failing due to non-implemented thrown from get_reboot_case, if the reboot was done with `sudo reboot` (cold reboot)

#### How I did it
Added get_reboot_case implementation that will always return Non-Hardware reason.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012

barefoot implementation of sonic_platform.chassis.Chassis.get_reboot_cause

### DIFF
--- a/platform/barefoot/sonic-platform-modules-bfn-montara/sonic_platform/chassis.py
+++ b/platform/barefoot/sonic-platform-modules-bfn-montara/sonic_platform/chassis.py
@@ -138,3 +138,16 @@ class Chassis(ChassisBase):
     def get_change_event(self, timeout=0):
         ready, event_sfp = Sfp.get_transceiver_change_event(timeout)
         return ready, { 'sfp': event_sfp } if ready else {}
+
+    def get_reboot_cause(self):
+        """
+        Retrieves the cause of the previous reboot
+
+        Returns:
+            A tuple (string, string) where the first element is a string
+            containing the cause of the previous reboot. This string must be
+            one of the predefined strings in this class. If the first string
+            is "REBOOT_CAUSE_HARDWARE_OTHER", the second string can be used
+            to pass a description of the reboot cause.
+        """
+        return self.REBOOT_CAUSE_NON_HARDWARE, ''


vboykox:
@akokhan

vboykox:
@jleveque to me it looks confusing that if platform does not support determining any of REBOOT_CAUSE_ it should return Non-Hardware to make determine-reboot-cause work.
imho REBOOT_CAUSE_NON_HARDWARE should only mean none of other conditions, which could be tracked by platform (power loss, etc), happened.
The alternative PR: https://github.com/Azure/sonic-buildimage/pull/7795

jleveque:
> @jleveque to me it looks confusing that if platform does not support determining any of REBOOT_CAUSE_ it should return Non-Hardware to make determine-reboot-cause work.
> imho REBOOT_CAUSE_NON_HARDWARE should only mean none of other conditions, which could be tracked by platform (power loss, etc), happened.
> The alternative PR: #7796

`REBOOT_CAUSE_NON_HARDWARE` implies that a hardware-initiated reboot cause (e.g., power loss, thermal overload, etc.) did not occur. If the platform doesn't support detecting any hardware-initiated reboot causes, then `REBOOT_CAUSE_NON_HARDWARE` makes sense, because we were unable to determine that a hardware-initiated reboot occurred.

vboykox:
@jleveque For now we will not support it in SONiC.
So we have to state that reason is not hardware even if device lost power. There is no "I don't know" option.

jleveque:
> @jleveque For now we will not support it in SONiC.
> So we have to state that reason is not hardware even if device lost power. There is no "I don't know" option.

If the platform API returns `REBOOT_CAUSE_NON_HARDWARE`, the reboot cause determination logic will check to see if a software reboot was initiated. If there is no record of a software reboot, the reboot cause will be "Unknown". So in your case, if a hardware-initiated reboot occurs but you return `REBOOT_CAUSE_NON_HARDWARE`, the reboot cause should be "Unknown".

vboykox:
@jleveque Yes, specifically for determine-reboot-cause, it makes sense to return NON_HARDWARE in combination with sw cause check.
Thank you for explanation!

jleveque (approved):

vboykox:
@lguohan

vboykox:
@lguohan

jleveque (approved):

PR #7822, opened 2021-06-08
[Arista] Remove 7050-QX32S unnecessary configs for cl72 and external phy

zzhiyuan:
#### Why I did it
Ask from Microsoft regarding these configs.

port_init_cl72_hg - Related to cl72 on fabric ( This sku doesn't use )
port_init_cl72 - disabling cl72 ( default is disabled )
port_phy_addr - related to external phy addressing ( This sku doesn't have any external phy )

These are not thought to have impact whether the configs are here or they are not. Since they are unnecessary this PR is to remove them.

#### How I did it
Remove the unnecessary configs from the HWSKUs of 7050-QX32S

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012


yxieca (approved):

lguohan:
does this have any implication on warm reboot?

zzhiyuan:
> does this have any implication on warm reboot?

I don't believe there is any impact to warm reboot.

PR #1754, opened 2018-05-31
[sonic-platform-daemons] Update submodule

jleveque:
Includes:
  - https://github.com/Azure/sonic-platform-daemons/pull/16: [ledd] Use select() with timeout for AppDB notifications

PR #7838, opened 2021-06-10
Adding mmu buffer yang files

AmitKaushik7:
#### Why I did it
Added Buffer yang files derived from sonic/mgmt-cvl/testdata/schema from sonic-mgmt-common,  used by mgmt-framework.

Updated BUFFER_PG|({ifname},)*|{pg_num} to BUFFER_PG|{ifname}|{pg_num} in sonic-buffer-pg.yang.
This change is required for configuration migration for dynamic port breakout operation.

Added sonic-buffer-queue.yang for BUFFER_QUEUE

Tables: BUFFER_POOL, BUFFER_PROFILE, BUFFER_PG, BUFFER_QUEUE.

#### How I did it
Defined Yang models for BUFFER tables based on Guideline doc:
https://github.com/Azure/SONiC/blob/master/doc/mgmt/SONiC_YANG_Model_Guidelines.md
and
https://github.com/Azure/sonic-utilities/blob/master/doc/Command-Reference.md

#### How to verify it
sonic_yang_models package build.

MMU Buffer yang files (BUFFER_POOL/BUFFER_PROFILE/BUFFER_PG/BUFFER_QUEUE)


AmitKaushik7:
Iguohan/Neethajohn   Could you review the PR and share you feedback

src/sonic-yang-models/yang-models/sonic-buffer-profile.yang
```diff
@@ -0,0 +1,67 @@
...
+
+			list BUFFER_PROFILE_LIST {
+				key "name";
+
+				leaf name {
+					type string;
+				}
+
+				leaf static_th {
+					type uint64;
```
lguohan, line 36: need description for all features

src/sonic-yang-models/yang-models/sonic-buffer-pg.yang
```diff
@@ -0,0 +1,66 @@
...
+	organization
+		"SONiC";
+
+	contact
+		"SONiC";
+
+	description
+		"SONIC BUFFER PG";
+
+	revision 2019-05-15 {
```
praveen-li, line 26: We want to keep this revision date or latest ?

src/sonic-yang-models/yang-models/sonic-buffer-pg.yang
```diff
@@ -0,0 +1,66 @@
...
+	}
+
+
+	container sonic-buffer-pg {
+
+		container BUFFER_PG {
+
+			list BUFFER_PG_LIST {
+				key "ifname pg_num";
+				//sonic-ext:key-pattern "BUFFER_PG|{ifname}|{pg_num}"; //special pattern used for extracting keys from
```
praveen-li, line 38: Plz remove the comments...

src/sonic-yang-models/yang-models/sonic-buffer-pg.yang
```diff
@@ -0,0 +1,66 @@
...
+					type leafref {
+						path "/prt:sonic-port/prt:PORT/prt:PORT_LIST/prt:name";
+					}
+				}
+
+				leaf pg_num {
+					type string {
+						pattern "[0-7]((-)[0-7])?" {
+							error-message "Invalid Buffer PG number";
+							error-app-tag pg-num-invalid;
```
praveen-li, line 52: If the file has tabs, plz convert to spaces.

src/sonic-yang-models/yang-models/sonic-buffer-pg.yang
```diff
@@ -0,0 +1,66 @@
...
+
+		container BUFFER_PG {
+
+			list BUFFER_PG_LIST {
+				key "ifname pg_num";
+				//sonic-ext:key-pattern "BUFFER_PG|{ifname}|{pg_num}"; //special pattern used for extracting keys from
+				//redis-key and fill populate the yang instance
+				// Total list instance = number(key1) * number(key2) * number(key3)
+
+				leaf ifname {
```
praveen-li, line 42: if logical interfaces are not allowed and this is always physical port, then better to use port_name\port. Agree?

src/sonic-yang-models/yang-models/sonic-buffer-pg.yang
```diff
@@ -0,0 +1,66 @@
...
+					type leafref {
+						path "/prt:sonic-port/prt:PORT/prt:PORT_LIST/prt:name";
+					}
+				}
+
+				leaf pg_num {
+					type string {
+						pattern "[0-7]((-)[0-7])?" {
```
lguohan, line 50: 0-7

src/sonic-yang-models/yang-models/sonic-buffer-pg.yang
```diff
@@ -0,0 +1,66 @@
...
+
+	container sonic-buffer-pg {
+
+		container BUFFER_PG {
+
+			list BUFFER_PG_LIST {
+				key "ifname pg_num";
+				//sonic-ext:key-pattern "BUFFER_PG|{ifname}|{pg_num}"; //special pattern used for extracting keys from
+				//redis-key and fill populate the yang instance
+				// Total list instance = number(key1) * number(key2) * number(key3)
```
lguohan, line 40: remove the commented line

src/sonic-yang-models/yang-models/sonic-buffer-queue.yang
```diff
@@ -0,0 +1,65 @@
...
+
+				leaf ifname {
+					type leafref {
+						path "/prt:sonic-port/prt:PORT/prt:PORT_LIST/prt:name";
+					}
+				}
+
+				leaf qindex {
+					type string {
+						pattern "[0-7]((-)[0-7])?"{
```
praveen-li, line 49: pattern must change 0-15,

src/sonic-yang-models/yang-models/sonic-buffer-pool.yang
```diff
@@ -0,0 +1,51 @@
...
+
+				leaf mode {
+					type enumeration {
+						enum static;
+						enum dynamic;
+					}
+				}
+
+				leaf size {
+					type uint64;
```
lguohan, line 45: what is the unit, is it bytes, can you clarify?

src/sonic-yang-models/yang-models/sonic-buffer-queue.yang
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+
+	container sonic-buffer-queue {
+
+		container BUFFER_QUEUE {
+
+			list BUFFER_QUEUE_LIST {
+				key "ifname qindex";
+				//sonic-ext:key-pattern "QUEUE|{ifname}|{qindex}"; //special pattern used for extracting keys from redis-key and populate the yang instance
```
praveen-li, line 38: same for this comments, ...remove

src/sonic-yang-models/yang-models/sonic-buffer-queue.yang
```diff
@@ -0,0 +1,65 @@
...
+										   // Total list instances = number(key1) * number(key2) * number(key3)
+
+				leaf ifname {
+					type leafref {
+						path "/prt:sonic-port/prt:PORT/prt:PORT_LIST/prt:name";
+					}
+				}
+
+				leaf qindex {
```
praveen-li, line 47: plz add descriptions for all leafs.

src/sonic-yang-models/yang-models/sonic-buffer-profile.yang
```diff
@@ -0,0 +1,67 @@
...
+					type int64;
+				}
+
+				leaf size {
+					type uint64;
+				}
+
+				leaf pool {
+					type leafref {
+						path "/bpl:sonic-buffer-pool/bpl:BUFFER_POOL/bpl:BUFFER_POOL_LIST/bpl:name";
```
dgsudharsan, line 49: Please mandate the buffer pool

src/sonic-yang-models/yang-models/sonic-buffer-profile.yang
```diff
@@ -0,0 +1,67 @@
...
+
+				leaf dynamic_th {
+					type int64;
+				}
+
+				leaf size {
+					type uint64;
+				}
+
+				leaf pool {
```
praveen-li, line 47: should be mandatory ?

src/sonic-yang-models/yang-models/sonic-buffer-profile.yang
```diff
@@ -0,0 +1,67 @@
...
+					type leafref {
+						path "/bpl:sonic-buffer-pool/bpl:BUFFER_POOL/bpl:BUFFER_POOL_LIST/bpl:name";
+					}
+				}
+
```
dgsudharsan, line 52: Can you please confirm if we need to add buffer profile threshold mode?

src/sonic-yang-models/yang-models/sonic-buffer-profile.yang
```diff
@@ -0,0 +1,67 @@
...
+					type string;
+				}
+
+				leaf static_th {
+					type uint64;
+				}
+
+				leaf dynamic_th {
+					type int64;
+				}
```
lguohan, line 41: is there way to express the constraint that static_th and dynamic_th as mutually exclusive?

dgsudharsan (changes requested):

src/sonic-yang-models/yang-models/sonic-buffer-profile.yang
```diff
@@ -0,0 +1,67 @@
...
+		container BUFFER_PROFILE {
+
+			list BUFFER_PROFILE_LIST {
+				key "name";
+
+				leaf name {
+					type string;
+				}
+
+				leaf static_th {
```
praveen-li, line 35: We can have either static or dynamic, we may need choice statement here

praveen-li (commented):
Updated comments from YANG WG.

src/sonic-yang-models/yang-models/sonic-buffer-queue.yang
```diff
@@ -0,0 +1,65 @@
...
+						pattern "[0-7]((-)[0-7])?"{
+							error-message "Invalid Q-index";
+							error-app-tag qindex-invalid;
+						}
+					}
+				}
+
+				leaf profile { //Hash reference key
+					type leafref {
+						path "/bpf:sonic-buffer-profile/bpf:BUFFER_PROFILE/bpf:BUFFER_PROFILE_LIST/bpf:name";
```
venkatmahalingam, line 58: This PR has a dependency on https://github.com/Azure/sonic-swss/pull/1754, please merge this PR once the dependent PR is merged.

src/sonic-yang-models/yang-models/sonic-buffer-pool.yang
```diff
@@ -0,0 +1,51 @@
...
+				}
```
venkatmahalingam, line 46: xoff field is missing in the buffer pool table.

src/sonic-yang-models/yang-models/sonic-buffer-profile.yang
```diff
@@ -0,0 +1,67 @@
...
+
+				leaf pool {
+					type leafref {
+						path "/bpl:sonic-buffer-pool/bpl:BUFFER_POOL/bpl:BUFFER_POOL_LIST/bpl:name";
+					}
+				}
+
+				leaf xon_offset {
+					type uint64;
+				}
```
venkatmahalingam, line 55: Please add descriptions for fields.

src/sonic-yang-models/yang-models/sonic-buffer-profile.yang
```diff
@@ -0,0 +1,67 @@
...
+				leaf name {
+					type string;
+				}
+
+				leaf static_th {
+					type uint64;
+				}
+
+				leaf dynamic_th {
+					type int64;
```
venkatmahalingam, line 40: I think, the valid range is -6 to 3, please check and update.

src/sonic-yang-models/yang-models/sonic-buffer-profile.yang
```diff
@@ -0,0 +1,67 @@
...
+				}
+
+				leaf size {
+					type uint64;
```
venkatmahalingam, line 44: Please refer the below SAI spec and update the default values for fields accordingly.

https://github.com/opencomputeproject/SAI/blob/master/inc/saibuffer.h

src/sonic-yang-models/yang-models/sonic-buffer-profile.yang
venkatmahalingam: We can try with count(field-name) = 0 or 1 condition in a must statement.

src/sonic-yang-models/yang-models/sonic-buffer-profile.yang
```diff
@@ -0,0 +1,67 @@
...
+	organization
+		"SONiC";
+
+	contact
+		"SONiC";
+
+	description
+		"SONIC BUFFER PROFILE";
+
+	revision 2019-05-15 {
```
venkatmahalingam, line 18: Update the revision.

src/sonic-yang-models/yang-models/sonic-buffer-profile.yang
```diff
@@ -0,0 +1,67 @@
...
+
+		container BUFFER_PROFILE {
+
+			list BUFFER_PROFILE_LIST {
+				key "name";
+
+				leaf name {
+					type string;
+				}
+
```
venkatmahalingam, line 34: As discussed in the meeting, can we add optional field mode (static/dynamic) as well?

src/sonic-yang-models/yang-models/sonic-buffer-pg.yang
venkatmahalingam: What if there is an overlap PG entries i.e same PG number for more than one entry, is that a valid config? if no, how are we planning to validate that? thru custom validation?

src/sonic-yang-models/yang-models/sonic-buffer-queue.yang
AmitKaushik7: Sure

src/sonic-yang-models/yang-models/sonic-buffer-profile.yang
AmitKaushik7: Venkat, there should not be a need to add an optional field "mode". Only one of static_th/dynamic_th must be specified and mode can be derived from threshold defined.  static_th/dynmic_th can be made mandatory and using choice  we can select one of then,

src/sonic-yang-models/yang-models/sonic-buffer-profile.yang
AmitKaushik7: Yes, will add a choice statement here.

src/sonic-yang-models/yang-models/sonic-buffer-profile.yang
AmitKaushik7: I will try with choice statement.

src/sonic-yang-models/yang-models/sonic-buffer-profile.yang
venkatmahalingam: This mode field is already present in the backend code and present in SAI spec as well under profile (SAI_BUFFER_PROFILE_ATTR_THRESHOLD_MODE), so, in order to match the existing implementation, we have to add it.

AmitKaushik7: Please check bufferorch.cpp, the mode value is derived based on static_th/dynamic_th value and 2 SAI Attributes are set accordingly (SAI_BUFFER_PROFILE_ATTR_THRESHOLD_MODE 
 and SAI_BUFFER_PROFILE_ATTR_SHARED_DYNAMIC_TH/SAI_BUFFER_PROFILE_ATTR_SHARED_STATIC_TH).

venkatmahalingam: I wanted to have the mode under buffer-profile table because some platforms use it, but since the backend is assuming the mode from static_th/dynamic_th, we can avoid the mode field in the profile table and update platform-specific buffer defaults, thanks!

https://github.com/Azure/sonic-buildimage/blob/master/device/dell/x86_64-dellemc_s5232f_c3538-r0/DellEMC-S5232f-C32/buffers_defaults_t0.j2#L26

src/sonic-yang-models/yang-models/sonic-buffer-pg.yang
AmitKaushik7: changed to port.

src/sonic-yang-models/yang-models/sonic-buffer-queue.yang
AmitKaushik7: changed to 0-15.

src/sonic-yang-models/yang-models/sonic-buffer-pg.yang
AmitKaushik7: Removed Tabs

src/sonic-yang-models/yang-models/sonic-buffer-pg.yang
AmitKaushik7: Removed comments.

src/sonic-yang-models/yang-models/sonic-buffer-profile.yang
AmitKaushik7: Note required. Backend can  derive it based on static_th/dynamic_th defined mode.

src/sonic-yang-models/yang-models/sonic-buffer-profile.yang
AmitKaushik7: Done.

src/sonic-yang-models/yang-models/sonic-buffer-pool.yang
AmitKaushik7: Bytes. Updated the description for the same.

src/sonic-yang-models/yang-models/sonic-port-ingress-profile-list.yang
```diff
@@ -0,0 +1,55 @@
...
+
+    revision 2021-07-01 {
+        description
+            "Initial revision.";
+    }
+
+
+    container sonic-port-ingress-profile-list {
+
+        container PORT_INGRESS_PROFILE_LIST {
```
venkatmahalingam, line 34: Where do we process this table in the backend?

src/sonic-yang-models/yang-models/sonic-buffer-profile.yang
AmitKaushik7: Praveen, 
   I tried following choice/case statement for the threshold, but got the error mentioned below. 
Statement
=======
                choice threshold_type {
                    mandatory true;
                    case static {
                        leaf static_th {
                            type uint64;
                            description "The maximum size of the buffer pool the port can occupy";
                        }
                    }
                    case dynamic {
                        leaf dynamic_th {
                            type int32 {
                                range "-7..3";
                            }
                            description "The maximum proportion of the free size of the buffer pool the port can occupy (Alpha value)";
                        }
                    }
                }

Error:
====
           self.root = None
           self.sysLog(msg="Data Loading Failed:{}".format(str(e)), \
            debug=syslog.LOG_ERR, doPrint=True)
>          raise SonicYangException("Data Loading Failed\n{}".format(str(e)))
E          sonic_yang_ext.SonicYangException: Data Loading Failed
E          string indices must be integers

src/sonic-yang-models/yang-models/sonic-port-ingress-profile-list.yang
AmitKaushik7: Renamed it to BUFFER_PORT_INGRESS_PROFILE_LIST, it is processed in bufferorch.cpp in the backend.

src/sonic-yang-models/yang-models/sonic-buffer-profile.yang
AmitKaushik7: Tried with choice but hitting error. Please refer comments from Praveen below.

src/sonic-yang-models/yang-models/sonic-buffer-pg.yang
AmitKaushik7: It should depend on the App code whether it needs to be overwritten or not. If we have 2 entries like "Ethernet4:3-4" and "Ethernet4:4-5", then both these configs are valid and should be applied in the order. Application in this case will be overriding  PG 4 with "Ethernet4:4-5" config. I am not aware if we can validate such overlap in yang.  The only solution I can think of is to split the PG(4 and 5) separately instead of range(say 4-5). This should be true not only for PG, but for other interfaces where we are specifying the range(Eg. Ethernet4-Ethernet10).

src/sonic-yang-models/yang-models/sonic-buffer-profile.yang
AmitKaushik7: Praveen, Do you see any issue with the above yang block  for choice.

src/sonic-yang-models/yang-models/sonic-buffer-pool.yang
```diff
@@ -0,0 +1,63 @@
...
+    container sonic-buffer-pool {
+
+        container BUFFER_POOL {
+
+            list BUFFER_POOL_LIST {
+                key "name";
+
+                leaf name {
+                    type string;
+                    description "Buffer Pool name";
```
venkatmahalingam, line 28: I'm not sure we allow any name here, I see the backend code written with a specific name e.g egress_lossless_pool, ingress_lossless_pool..etc, please check and restrict the name usage.

AmitKaushik7: Hi Venkat
   egress_lossless_pool/ingress_lossy_pool etc are defined in buffer_defaults j2 files only. There is no restriction for the name in the cpp files and it is just a string and it just has to be unique only. I dont think there is a need of name restriction here. 
Thanks
Amit

venkatmahalingam: Please see the code below, there is an assumption to use a particular pool name.
https://github.com/Azure/sonic-swss/blob/master/cfgmgr/buffermgr.cpp#L173

AmitKaushik7: This is added only for ingress_lossless_pool in sonic-swss/cfgmgr/buffermgr.h.
#define INGRESS_LOSSLESS_PG_POOL_NAME "ingress_lossless_pool"
What about other buffer pools- egress/ingress lossy/lossless. (egress_lossy_pool/ingress_lossy_pool) etc?

src/sonic-yang-models/yang-models/sonic-buffer-queue.yang
AmitKaushik7: Venkat, Azure/sonic-swss#1754 is still not merged. Does the current PR has dependency on #1754.

zhangyanzhao:
@venkatmahalingam will take a further look, then we can merge.

src/sonic-yang-models/yang-models/sonic-buffer-queue.yang
venkatmahalingam: Yes, the current backend expectation is to have a table name (BUFFER_PROFILE) also while referencing to profile table from BUFFER_QUEUE and other tables.

127.0.0.1:6379[4]> hgetall BUFFER_QUEUE|Ethernet38|5-6
1) "profile"
2) "[BUFFER_PROFILE|egress_lossy_profile]"

src/sonic-yang-models/yang-models/sonic-buffer-pool.yang
venkatmahalingam: ok, let's have a string type for now to accept any pool name.

venkatmahalingam:
> @venkatmahalingam will take a further look, then we can merge.

This PR has a dependency on Azure/sonic-swss#1754

src/sonic-yang-models/yang-models/sonic-buffer-pool.yang
```diff
@@ -0,0 +1,63 @@
...
+                    mandatory true;
+                    type enumeration {
+                        enum ingress;
+                        enum egress;
+                    }
+                    description "Buffer Pool Type";
+                }
+
+                leaf mode {
+                    default dynamic;
```
stephenxs, line 41: Shouldn't it be mandatory as well?

src/sonic-yang-models/yang-models/sonic-buffer-port-egress-profile-list.yang
```diff
@@ -0,0 +1,55 @@
...
+
+                leaf port {
+                    type leafref {
+                        path "/prt:sonic-port/prt:PORT/prt:PORT_LIST/prt:name";
+                    }
+                }
+
+                leaf-list profile_list {
+                    type leafref {
+                        path "/bpf:sonic-buffer-profile/bpf:BUFFER_PROFILE/bpf:BUFFER_PROFILE_LIST/bpf:name";
```
stephenxs, line 47: Is it possible to check whether an egress profile is referenced here?
An egress profile is a buffer profile that references a buffer pool with `egress` as type

src/sonic-yang-models/yang-models/sonic-buffer-port-ingress-profile-list.yang
```diff
@@ -0,0 +1,55 @@
...
+
+                leaf port {
+                    type leafref {
+                        path "/prt:sonic-port/prt:PORT/prt:PORT_LIST/prt:name";
+                    }
+                }
+
+                leaf-list profile_list {
+                    type leafref {
+                        path "/bpf:sonic-buffer-profile/bpf:BUFFER_PROFILE/bpf:BUFFER_PROFILE_LIST/bpf:name";
```
stephenxs, line 47: The same as egress profile list.

src/sonic-yang-models/yang-models/sonic-buffer-profile.yang
```diff
@@ -0,0 +1,82 @@
...
+                    type int32 {
+                        range "-7..3";
+                    }
+                    description "The maximum proportion of the free size of the buffer pool the port can occupy (Alpha value)";
+                }
+
+                leaf size {
+                    mandatory true;
+                    type uint64;
+                    description "Headroom Size";
```
stephenxs, line 51: Per my understanding, headroom is on the ingress side only. But the buffer profiles are configured on both sides.
Better to change it to `reserved size`.

src/sonic-yang-models/yang-models/sonic-buffer-profile.yang
```diff
@@ -0,0 +1,82 @@
...
+                }
+
+                leaf static_th {
+                    type uint64;
+                    description "The maximum size of the buffer pool the port can occupy";
+                }
+
+                leaf dynamic_th {
+                    type int32 {
+                        range "-7..3";
```
stephenxs, line 43: I think the range should be the union of the legal range of all vendors 
`-8..7` should be better.
BTW the bounders are inclusive, right?

stephenxs (changes requested):

src/sonic-yang-models/yang-models/sonic-buffer-profile.yang
```diff
@@ -0,0 +1,82 @@
...
+                    default 0;
+                    description "Xon Threshold for ingress PG";
+                }
+
+                leaf xoff {
+                    type uint64;
+                    default 0;
+                    description "Xoff Threshold for ingress PG";
+                }
+            }
```
stephenxs, line 79: There is another leaf `headroom_type` if the buffer model is dynamic. It is handled by dynamic buffer manager.

Check [schema](https://github.com/Azure/SONiC/blob/master/doc/qos/dynamically-headroom-calculation.md#schema-7) and [example](https://github.com/Azure/SONiC/blob/master/doc/qos/dynamically-headroom-calculation.md#dynamic-calculation-with-non-default-alpha) for details

zhangyanzhao:
BRCM team will follow-up this.

src/sonic-yang-models/yang-models/sonic-buffer-port-egress-profile-list.yang
AmitKaushik7: No I dont think it can be checked for leafref.

src/sonic-yang-models/yang-models/sonic-buffer-port-ingress-profile-list.yang
AmitKaushik7: No I dont think it can be checked for leafref.

AmitKaushik7:
@neethajohn  Can you see if this PR can be merged. Stephen had approved the changes. Builds & tests failure were not related to the changes in this PR. After rebase all builds are successful and tests are passing.

neethajohn (approved):

PR #7846, opened 2021-06-10
Grouping delayed services under a target for config reload checks

dgsudharsan:
#### Why I did it
Create a target for delayed service timers. Few services in sonic have delayed to speed up the bring up of the system and essential services. However there is no way to track when they start. This will be a problem when executing config reload as config reload expects all services to be up. Hence grouped all the timers that trigger the delayed services under one target so that they could be tracked in 'config reload' command

#### How I did it
Created delay.target service and add created dependency on the delayed targets.

#### How to verify it
This is verified using config reload command. The command should not execute after boot, until all the delayed services are up. Introduced sonic-mgmt test cases to cover the logic.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012


dgsudharsan:
@jleveque Can you please review this as part of the config reload changes?

files/build_templates/mgmt-framework.timer
```diff
@@ -8,4 +8,4 @@ OnBootSec=3min 30 sec
 Unit=mgmt-framework.service
 
 [Install]
-WantedBy=timers.target sonic.target
+WantedBy=timers.target sonic.target delay.target
```
jleveque, line 11: Maybe the target should have a more descriptive name? Maybe `sonic-delayed.target`?

dgsudharsan: Updated.

jleveque (approved):

jleveque:
Looks good to me. Please wait for other reviewers.

qiluo-msft:
Where is "sonic-mgmt test cases to cover the logic"?

dgsudharsan:
> Where is "sonic-mgmt test cases to cover the logic"?

https://github.com/Azure/sonic-mgmt/pull/3657

qiluo-msft (approved):

PR #7863, opened 2021-06-14
Bgpcli

devadityaojha:
#### Why I did it
Found a bug, that invokes a nonetype error when running "show ip bgp network" on a single ASIC vs
#### How I did it
Added an if-condition to raise an exception, specifically an attribute error, when there is a single asic vs i.e. when the asic_name is None or does not have a "startswith" attribute. 
#### How to verify it
Run "show ip bgp network" on the VS CLI

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012


smaheshm:
can you add before and after fix output in the 'verify' section.

src/sonic-py-common/sonic_py_common/multi_asic.py
```diff
@@ -132,6 +132,12 @@ def get_asic_id_from_name(asic_name):
     Returns:
         asic id.
     """
+    #Exception addition
+    if asic_name is None:
+        raise AttributeError('Single ASIC platform cannot have this attribute.')
+    elif not hasattr(asic_name, 'startswith'):
```
smaheshm, line 138: is the objective here to check if 'asic_name' is an instance of 'str'?

src/sonic-py-common/sonic_py_common/multi_asic.py
```diff
@@ -132,6 +132,12 @@ def get_asic_id_from_name(asic_name):
     Returns:
         asic id.
     """
+    #Exception addition
+    if asic_name is None:
```
smaheshm, line 136: The 'check' and the exception error message is bit confusing. Instead of 'AttributeError', raise 'ValueError'

rlhui:
@smaheshm - should this be continued and what's needed?

smaheshm:
> @smaheshm - should this be continued and what's needed?

not needed. Will close it.

PR #7872, opened 2021-06-15
[build][multi-asic][vs]: Add a new NUM_ASIC build parameter

SuvarnaMeenakshi:
#### Why I did it
Virtual switch image is required in single and multi-asic form to run VS tests.
To build a multi-asic VS image, <platform>/asic.conf has to be updated with right number of asics and device_id.
This is diff is to:
1. Help generate multi-asic VS image by passing the NUM_ASIC in build parameter.
2. Help build single asic and multi-asic VS using the same packages with just asic.conf updated with the right number of asics. 
#### How I did it
1. Add a new build parameter NUM_ASIC which will have a default value of 1.
2. For VS platform, Export this variable which will be used by asic.conf.j2 to generate asic.conf.
#### How to verify it
Build a single asic VS image as being today:
make target/sonic-vs.img.gz - Ensure that /usr/share/sonic/device/<platform>/asic.conf has "NUM_ASIC=1".
In the same workspace, build multi-asic VS image:
make NUM_ASIC_VS=2 target/sonic-vs.img.gz - /usr/share/sonic/device/<platform>/asic.conf has 
NUM_ASIC=2

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012


slave.mk
```diff
@@ -198,6 +198,14 @@ ifeq ($(BUILD_LOG_TIMESTAMP),)
 override BUILD_LOG_TIMESTAMP := $(DEFAULT_BUILD_LOG_TIMESTAMP)
 endif
 
+ifeq ($(NUM_ASIC),)
+override NUM_ASIC := $(DEFAULT_NUM_ASIC)
```
qiluo-msft, line 202: >NUM_ASIC [](http://example.com/codeflow?start=9&length=8)

NUM_ASIC is a runtime parameter. The same sonic image could run either 6-ASIC switch or 66-ASIC switch (just example). Why the build process need to know the number?

SuvarnaMeenakshi: That is right, this is mainly done to make the building easier in Azure Pipeline script. If we go with the current build setup,
We will need to something like this:
```
          set -ex
          if [ $(GROUP_NAME) == vs ]; then
            sudo echo NUM_ASIC=$(num_asic) > ./device/virtual/x86_64-kvm_x86_64-r0/asic.conf
            asic_id=0
            while [ $asic_id -le $(num_asic) ]; do
               sudo echo DEV_ID_ASIC_$asic_id=$asic_id >> ./device/virtual/x86_64-kvm_x86_64-r0/asic.conf
               asic_id=$(( $asic_id + 1 ))
            done
            make $BUILD_OPTIONS target/sonic-vs.img.gz && mv target/sonic-vs.img.gz target/sonic-masic-vs.img.gz
            sudo echo NUM_ASIC=1 > ./device/virtual/x86_64-kvm_x86_64-r0/asic.conf
            rm ./target/debs/buster/sonic-device-data_1.0-1_all.deb
            make $BUILD_OPTIONS target/docker-sonic-vs.gz target/sonic-vs.img.gz target/docker-ptf.gz
```
If we have a build parameter, then we could directly build multi-asic VS image as in the PR:
https://github.com/Azure/sonic-buildimage/pull/7873/

qiluo-msft: Now I understand you are using the parameter only for KVM image. Could you change parameter name to sth like `NUM_VIRTUAL_ASIC` or `NUM_ASIC_VS` to make it explicit?

SuvarnaMeenakshi: Updated as suggested.

lguohan:
i do not fully agree on the approach. we use single brcm image for single asic and multi-asic platform, why do we need to build two images for single asic and multi-asic platform? we should only build one image, to solve the issue, we should build two onie image that one present as single asic, and another present multi-asic.

SuvarnaMeenakshi:
Follow comment above (to solve the issue, we should build two onie image that one present as single asic, and another present multi-asic.) , created 2 new PRs:
https://github.com/Azure/sonic-buildimage/pull/8269
https://github.com/Azure/sonic-buildimage/pull/8270

PR #6148, opened 2020-12-07
[docker-mgmt-framework]: update mgmt framework docker to support sonic-cli cmd

kwangsuk:
**- Why I did it**
- migrate to python3 support
- add dependent packages for Klish 
- allow login as non-root user

**- How I did it**
update sonic-cli script to start Klish with user name, system name and timeout
update the Dockerfile.j2 to resolve dependent packages
add python3-dev for Klish use 

**- How to verify it**
Incremental buster build with https://github.com/Azure/sonic-mgmt-framework/pull/76 and verify the sonic-cli

**- Which release branch to backport (provide reason below if selected)**
- [ ] 201811
- [ ] 201911
- [ ] 202006

**- Description for the changelog**
Migrate to python3.7 support, update sonic-cli script and resolve package dependencies 

**- A picture of a cute animal (not mandatory but encouraged)**

PR #7880, opened 2021-06-15
Revert "[docker-mgmt-framework]: update mgmt framework docker to support sonic-cli cmd"

renukamanavalan:
Reverts Azure/sonic-buildimage#6148.

Causes debug build failure.

When building docker-sonic-mgmt-framework-dbg, openssl-client will be installed, it will add a group as below. During the docker build, the /host_etc does not exist.

```
johnar@0c7bd7e98baf:/sonic/dockers/docker-sonic-mgmt-framework$ grep openssh Dockerfile-dbg 
RUN apt-get update && apt-get install -f -y gdb gdbserver vim openssh-client sshpass strace

I have no name!@fd2a27db6a3c:/# /usr/sbin/groupadd -g 100 ssh
groupadd: cannot open /etc/group
I have no name!@fd2a27db6a3c:/#

I have no name!@fd2a27db6a3c:/# ls /etc/group -l
lrwxrwxrwx 1 0 0 15 Jun 14 10:08 /etc/group -> /host_etc/group


```

### DIFF
--- a/dockers/docker-sonic-mgmt-framework/Dockerfile.j2
+++ b/dockers/docker-sonic-mgmt-framework/Dockerfile.j2
@@ -7,7 +7,21 @@ RUN [ -f /etc/rsyslog.conf ] && sed -ri "s/%syslogtag%/$docker_container_name#%s
 ENV DEBIAN_FRONTEND=noninteractive
 
 RUN apt-get update && \
-    apt-get install -y g++ python3-dev libxml2 libcurl3-gnutls libcjson-dev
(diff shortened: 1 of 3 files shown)


jeff-yin:
I don't understand how reverting #6148 fixes the issue described in the problem description. 
If the problem is due to the fact that openssh-client is being installed, that installation was introduced by https://github.com/Azure/sonic-buildimage/commit/fc55329b220e21a25dd145197002a356ef271641 which was merged last year. So what else changed recently?

PR #7881, opened 2021-06-15
sonic-yang-models updates for MPLS

qbdwlr:
What I did
SONiC YANG model support in buildimage for MPLS:
1) sonic-yang-model support for MPLS enable/disable
2) sonic-yang-model support for MPLS CRM thresholds

Why I did it
SONiC yang-models support for MPLS

How I verified it
Unit-tests in sonic-swss/tests/test_mpls.py and sonic-utilities/tests
System tests in sonic-mgmt

Details if related
Refer to MPLS HLD https://github.com/Azure/SONiC/pull/706

### DIFF
--- a/src/sonic-yang-models/yang-models/sonic-crm.yang
+++ b/src/sonic-yang-models/yang-models/sonic-crm.yang
@@ -353,6 +353,58 @@ module sonic-crm {
                     type threshold;
                 }
 
+                leaf mpls_inseg_threshold_type {
+                    description "CRM threshold support for MPLS in-segment entries";
+
+                    must "(((current()='PERCENTAGE' or current()='percentage') and
+                        ../mpls_inseg_high_threshold<100 and
+                        ../mpls_inseg_low_threshold<100) or
+                        (current()!='PERCENTAGE' and current()!='percentage'))";
+                    type stypes:crm_threshold_type;
+                }
+
+                leaf mpls_inseg_high_threshold {
+                    description "CRM threshold support for MPLS in-segment entries";
+
+                    must "(current() > ../mpls_inseg_low_threshold)"
+                    {
+                        error-message "high_threshold should be more than low_threshold";
+                    }
+                    type threshold;
+                }
+
+                leaf mpls_inseg_low_threshold {
+                    description "CRM threshold support for MPLS in-segment entries";
+
+                    type threshold;
+                }
+
+                leaf mpls_nexthop_threshold_type {
+                    description "CRM threshold support for MPLS next-hops";
+
+                    must "(((current()='PERCENTAGE' or current()='percentage') and
+                        ../mpls_nexthop_high_threshold<100 and
+                        ../mpls_nexthop_low_threshold<100) or
+                        (current()!='PERCENTAGE' and current()!='percentage'))";
+                    type stypes:crm_threshold_type;
+                }
+
+                leaf mpls_nexthop_high_threshold {
+                    description "CRM threshold support for MPLS next-hops";
+
+                    must "(current() > ../mpls_nexthop_low_threshold)"
+                    {
+                        error-message "high_threshold should be more than low_threshold";
+                    }
+                    type threshold;
+                }
+
+                leaf mpls_nexthop_low_threshold {
+                    description "CRM threshold support for MPLS next-hops";
+
+                    type threshold;
+                }
+
             }
             /* end of Config */
         }
--- a/src/sonic-yang-models/yang-models/sonic-interface.yang
+++ b/src/sonic-yang-models/yang-models/sonic-interface.yang
@@ -68,6 +68,13 @@ module sonic-interface {
 					}
                                         default "0";
 				}
+				leaf mpls {
+					description "Enable/disable MPLS routing for the interface";
+					type enumeration {
+						enum enable;
+						enum disable;
+					}
+				}
 			}
 			/* end of INTERFACE_LIST */
 
--- a/src/sonic-yang-models/yang-models/sonic-vlan.yang
+++ b/src/sonic-yang-models/yang-models/sonic-vlan.yang
@@ -80,6 +80,13 @@ module sonic-vlan {
 					}
                                         default "0";
 				}
+				leaf mpls {
+					description "Enable/disable MPLS routing for the vlan interface";
+					type enumeration {
+						enum enable;
+						enum disable;
+					}
+				}
 			}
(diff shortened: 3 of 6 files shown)


src/sonic-yang-models/yang-models/sonic-interface.yang
```diff
@@ -68,6 +68,12 @@ module sonic-interface {
 					}
                                         default "0";
 				}
```
kperumalbfn, line 70: Can you add Yang model tests for mpls enable/disable on RIFs? Most of tests are in sonic-yang-models/tests/yang_model_tests/tests

src/sonic-yang-models/yang-models/sonic-crm.yang
```diff
@@ -353,6 +353,46 @@ module sonic-crm {
                     type threshold;
                 }
 
+                leaf mpls_inseg_threshold_type {
```
kperumalbfn, line 356: Please add tests for MPLS CRM.

src/sonic-yang-models/yang-models/sonic-interface.yang
```diff
@@ -68,6 +68,12 @@ module sonic-interface {
...
+				leaf mpls {
+					description "Enable/disable MPLS routing for the interface";
+					type string {
+						pattern "enable|disable";
```
kperumalbfn, line 74: Default value for all the interface types?

qbdwlr: @kperumalbfn I looked at the types of tests currently under sonic-yang-models.  It looked like it was mostly parameter validation.  Since this mpls knob has two fixed values, I wasn't sure what more needed to be checked for that.  Do you have something specific in mind?  Can you provide a pointer to an example?

src/sonic-yang-models/yang-models/sonic-crm.yang
qbdwlr: @kperumalbfn I looked at the existing CRM tests in sonic-yang-models.  It appears that there exist currently generic parameter checking that is applicable for *all* CRM types.  I don't think there needs to be anything specific for MPLS, since it can already make use of the generic CRM tests.

src/sonic-yang-models/yang-models/sonic-interface.yang
qbdwlr: @kperumalbfn I looked at the types of tests currently under sonic-yang-models.  It looked like it was mostly parameter validation.  Since this mpls knob has two fixed values, I wasn't sure what more needed to be checked for that.  Do you have something specific in mind?  Can you provide a pointer to an example?

src/sonic-yang-models/yang-models/sonic-interface.yang
qbdwlr: @kperumalbfn I don't think we want to add a default value here, do we?  What happens if the platform does not support this attribute?  I think we do not want to explicitly set the attribute in that case.

src/sonic-yang-models/yang-models/sonic-interface.yang
kperumalbfn: I think it will be good to have a test for one type of interface. SAI default value for SAI_ROUTER_INTERFACE_ATTR_ADMIN_MPLS_STATE is disabled, we can use that. There are examples for NAT_ZONE in sonic-yang-models/tests/yang_model_tests/tests/interface.json

src/sonic-yang-models/yang-models/sonic-vlan.yang
```diff
@@ -80,6 +80,13 @@ module sonic-vlan {
 					}
                                         default "0";
 				}
+				leaf mpls {
+					description "Enable/disable MPLS routing for the vlan interface";
+					type enumeration {
+						enum enable;
+						enum disable;
```
venkatmahalingam, line 87: Please define typedef for enable/disable and use it here.

liat-grozovik:
@qbdwlr can you pls share some update on the review comments handling? it is blocking dpb from being executed on master.

src/sonic-yang-models/yang-models/sonic-crm.yang
qbdwlr: @kperumalbfn I have added CRM tests for the 2 MPLS CRM types as requested.

src/sonic-yang-models/yang-models/sonic-interface.yang
qbdwlr: @kperumalbfn  Type for mpls attribute has been changed from string to enum, which I believe should eliminate the need for valid/invalid attribute value tests.  No explicit default should be defined in yang models due to lack of support for this atttribute in many existing platforms.

src/sonic-yang-models/yang-models/sonic-vlan.yang
qbdwlr: @venkatmahalingam There seems to be some inconsistency with regards to typedefs in sonic-yang-models.  I see instances of enable/disable enum in sonic-device_metadata.yang and sonic-flex_counter.yang without a generic typedef being referenced.  Also, I see that sonic-types.yang now requires a revision.  I would suggest that a new revision of sonic-types.yang that incorporates a generic enable/disable typedef should be outside of the scope of this PR and done as a general cleanup of the yang_models to avoid churn on yang revisions.

qbdwlr:
@zhangyanzhao @liat-grozovik There is nothing preventing this PR from being merged in its current state.  It may be merged now to resolve DPB issue.

src/sonic-yang-models/yang-models/sonic-vlan.yang
venkatmahalingam: why can't we add typedef for enable/disable in the sonic-types.yang and use it in this YANG? we can correct typedef in the other modules later?

qbdwlr: @venkatmahalingam Presumably any addition to sonic-types.yang requires a new revision for that file.  It makes more sense to bundle multiple changes to sonic-types.yang together to avoid revision churn.  This should be done as general cleanup prior to new release and should not be included in the scope of this PR.

zhangyanzhao:
We will create a new PR to track the "why can't we add typedef for enable/disable in the sonic-types.yang and use it in this YANG? we can correct typedef in the other modules later?" issue. @venkatmahalingam please approve this PR and Guohan will merge.

venkatmahalingam (approved):

PR #7890, opened 2021-06-16
[CI] Fix azp built docker-sonic-mgmt image

liushilongbuaa:
#### Why I did it
UserID is different with the image built by Jenkins if we build docker-sonic-mgmt in sonicbld pool.
So we build this image in sonictest pool. There is a sonictmp user with UserID 1001.
This will make docker image same as the image built in Jenkins

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012


xumia (approved):

lguohan:
the idea way is to use sonicbld, this is only a temp fix. we need to remove the user from the docker.

PR #7911, opened 2021-06-18
[voqsystemlagid] Fix for timing issue in setting system lag id boundary

vganesan-nokia:
#### Why I did it

In VOQ chassis systems, system lag id boundary is set in redis-chassis server immediately after starting the database-chassis container in the supervisor card. This system lad id boundary was not set correctly all the time. Intermittently there was failure in setting these values in CHASSIS_APP_DB in redis-chassis server. This PR has fix for this.

#### How I did it

The voq system lag id boundary is set in redis-chassis. Changes include
setting this from database-chassis container. This fixes a timing issue
in finding datbase_config.json file from redis directory which is
created from database container. Since database container usually
starts after database-chassis container the existence of this file is
unreliable while running the command. Running the command under
database-chassis container makes sure that the database_config.json form
redis-chassis directory is guaranteed to be available and hence fixes the
timing issue.

#### How to verify it

In VOQ chassis system, set the system lag id start and system lag id end in chassisdb.conf. Reboot the chassis and observe that SYSTEM_LAG_ID_START and SYSTEM_LAG_ID_END are set correctly as mentioned in chassisdb.conf and there are no failures reported in syslog.

N/A

The voq system lag id boundary is set in redis-chassis. Changes include setting this from database-chassis container. This fixes a timing issue in finding datbase_config.json file from redis directory which is created from database container. Since database container usually starts after database-chassis container the existence of this file is unreliable while running the command. Running the command under database-chassis container makes sure that the database_config.json from redis-chassis directory is guaranteed to be available and hence fixes the timing issue.

### DIFF
--- a/files/build_templates/docker_image_ctl.j2
+++ b/files/build_templates/docker_image_ctl.j2
@@ -108,8 +108,8 @@ function setPlatformLagIdBoundaries()
     CHASSIS_CONF=/usr/share/sonic/device/$PLATFORM/chassisdb.conf
     if [ -f "$CHASSIS_CONF" ]; then
         source $CHASSIS_CONF
-        $SONIC_DB_CLI CHASSIS_APP_DB SET "SYSTEM_LAG_ID_START" "$lag_id_start"
-        $SONIC_DB_CLI CHASSIS_APP_DB SET "SYSTEM_LAG_ID_END" "$lag_id_end"
+        docker exec -i ${DOCKERNAME} $SONIC_DB_CLI CHASSIS_APP_DB SET "SYSTEM_LAG_ID_START" "$lag_id_start"
+        docker exec -i ${DOCKERNAME} $SONIC_DB_CLI CHASSIS_APP_DB SET "SYSTEM_LAG_ID_END" "$lag_id_end"
     fi
 }
 {%- endif %}


vganesan-nokia:
This is fix for issue https://github.com/Azure/sonic-buildimage/issues/7912

ysmanman (approved):

files/build_templates/docker_image_ctl.j2
```diff
@@ -108,8 +108,8 @@ function setPlatformLagIdBoundaries()
     CHASSIS_CONF=/usr/share/sonic/device/$PLATFORM/chassisdb.conf
     if [ -f "$CHASSIS_CONF" ]; then
         source $CHASSIS_CONF
-        $SONIC_DB_CLI CHASSIS_APP_DB SET "SYSTEM_LAG_ID_START" "$lag_id_start"
-        $SONIC_DB_CLI CHASSIS_APP_DB SET "SYSTEM_LAG_ID_END" "$lag_id_end"
+        docker exec -i ${DOCKERNAME} $SONIC_DB_CLI CHASSIS_APP_DB SET "SYSTEM_LAG_ID_START" "$lag_id_start"
```
judyjoseph, line 111: Could you share any errors seen in syslog when it doesn't work ? I am trying to understand why it would fail .. and how this fix is helping to slove it.

As we actually wait for supervisor in docker to be up before invoking setPlatformLagIdBoundaries() here
```
        until [[ ($(docker exec -i ${DOCKERNAME} pgrep -x -c supervisord) -gt 0) &&
                 ($(docker exec -i ${DOCKERNAME} $SONIC_DB_CLI CHASSIS_APP_DB PING | grep -c True) -gt 0) ]]; do
           sleep 1
        done
        setPlatformLagIdBoundaries
```

vganesan-nokia: Without this fix we would get the error : **Invalid database name input : 'CHASSIS_APP_DB`**
We are trying to set the lag id boundaries (SYSTEM_LAG_ID_START and SYSTEM_LAG_ID_END) in chassis app db **redis_chassis** server (not in regular **redis** server. Please note that in supervisor card we run two redis servers viz. **redis** in *database* container and **redis_chassis** server in *database-chassis* container). The check that you mentioned makes sure that this **redis_chassis** server is up and running before we set these variables. But the command not invoked under the *database-chassis* container actually accesses the /var/run/redis/sonic-db/database_config.jon to resolve CHASSIS_APP_DB. This is incorrect. This file is for regular **redis** server running in *database* container which may or may not exist at the time of invoking this boundary setting command (since these two containers are started asynchronously or *database* container is started after *database-chassis* container). Since there is no namespaces in the supervisor card to run the command for correct redis server (which is redis_chassis), we run the command under the correct container context.

judyjoseph: Thanks @vganesan-nokia for the details, then it is right to run the sonic-db-cli in the correct docker database-chassis. 

Ideally I would have assumed that the redis-chassis server also is running in the database docker itself, instead of creating a separate database-chassis docker with the same database_config.json file.

judyjoseph (approved):
LGTM

PR #7917, opened 2021-06-18
SONiC MAC ACL Yang model update to add support for Source MAC, Destination MAC, Ethertype pattern update, VLAN_ID, PCP, DEI fields.

abhishekd0:
Signed-off-by: Abhishek Dharwadkar <48468617+abhishekd0@users.noreply.github.com>

#### Why I did it
Currently only IP ACL and related model is defined. Support for MAC ACL is missing. Added support for it.

#### How I did it
ACL_RULE table is added with new MAC ACL related fields namely Source MAC, Destination MAC, Ethertype (Pattern updated to match any valid Ethertypes), VLAN, PCP, DEI

#### How to verify it
Yang model tests are attached.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012

Add support for MAC ACL yang model.

### DIFF
--- a/src/sonic-yang-models/tests/yang_model_tests/tests_config/acl.json
+++ b/src/sonic-yang-models/tests/yang_model_tests/tests_config/acl.json
@@ -657,5 +657,123 @@
                 ]
             }
         }
+    },
+    "ACL_TABLE_L2_ACL_FIELDS": {
+        "sonic-acl:sonic-acl": {
+            "sonic-acl:ACL_RULE": {
+                "ACL_RULE_LIST": [
+                    {
+                        "ACL_TABLE_NAME": "L2ACL",
+                        "SRC_MAC": "00:00:AB:CD:EF:00/FF:FF:FF:00:00:00",
+                        "DST_MAC": "00:00:AB:CD:EF:FF/FF:FF:FF:FF:FF:FF",
+                        "ETHER_TYPE": "0x0800",
+                        "PCP": "5/5",
+                        "DEI": "0",
+                        "PACKET_ACTION": "FORWARD",
+                        "PRIORITY": 999980,
+                        "RULE_NAME": "Rule_20"
+                    }
+                ]
+            },
+            "sonic-acl:ACL_TABLE": {
+                "ACL_TABLE_LIST": [
+                    {
+                        "ACL_TABLE_NAME": "L2ACL",
+                        "policy_desc": "L2ACL Test",
+                        "ports": [ "" ],
+                        "stage": "INGRESS",
+                        "type": "L2"
+                    }
+                ]
+            }
+        }
+    },
+    "ACL_TABLE_L3_RULE_WITH_L2_FIELDS": {
+        "sonic-acl:sonic-acl": {
+            "sonic-acl:ACL_RULE": {
+                "ACL_RULE_LIST": [
+                    {
+                        "ACL_TABLE_NAME": "L3ACL-MAC-FIELDS",
+                        "SRC_MAC": "00:00:AB:CD:EF:00/FF:FF:FF:00:00:00",
+                        "DST_MAC": "00:00:AB:CD:EF:FF/FF:FF:FF:FF:FF:FF",
+                        "ETHER_TYPE": "0x0800",
+                        "PCP": "5/5",
+                        "DEI": "0",
+                        "PACKET_ACTION": "FORWARD",
+                        "PRIORITY": 999980,
+                        "RULE_NAME": "Rule_20"
+                    }
+                ]
+            },
+            "sonic-acl:ACL_TABLE": {
+                "ACL_TABLE_LIST": [
+                    {
+                        "ACL_TABLE_NAME": "L3ACL-MAC-FIELDS",
+                        "policy_desc": "L2ACL Test",
+                        "ports": [ "" ],
+                        "stage": "INGRESS",
+                        "type": "L3"
+                    }
+                ]
+            }
+        }
+    },
+    "ACL_RULE_L2_INVALID_MAC": {
+        "sonic-acl:sonic-acl": {
+            "sonic-acl:ACL_RULE": {
+                "ACL_RULE_LIST": [
+                    {
+                        "ACL_TABLE_NAME": "L2ACL_INVALID_MAC",
+                        "SRC_MAC": "00.00.AB.CD.EF.00/FF.FF.FF.00.00.00",
+                        "DST_MAC": "00.00.AB.CD.EF.FF/FF.FF.FF.FF.FF.FF",
+                        "ETHER_TYPE": "0x0800",
+                        "PCP": "5/5",
+                        "DEI": "0",
+                        "PACKET_ACTION": "FORWARD",
+                        "PRIORITY": 999980,
+                        "RULE_NAME": "Rule_20"
+                    }
+                ]
+            },
+            "sonic-acl:ACL_TABLE": {
+                "ACL_TABLE_LIST": [
+                    {
+                        "ACL_TABLE_NAME": "L2ACL_INVALID_MAC",
+                        "policy_desc": "L2ACL Test",
+                        "ports": [ "" ],
+                        "stage": "INGRESS",
+                        "type": "L2"
+                    }
+                ]
+            }
+        }
+    },
+    "ACL_RULE_L2_INVALID_ETHER": {
+        "sonic-acl:sonic-acl": {
+            "sonic-acl:ACL_RULE": {
+                "ACL_RULE_LIST": [
+                    {
+                        "ACL_TABLE_NAME": "L2ACL_INVALID_ETHER",
+                        "SRC_MAC": "00.00.AB.CD.EF.00/FF.FF.FF.00.00.00",
+                        "DST_MAC": "00.00.AB.CD.EF.FF/FF.FF.FF.FF.FF.FF",
+                        "ETHER_TYPE": "64",
+                        "PACKET_ACTION": "FORWARD",
+                        "PRIORITY": 999980,
+                        "RULE_NAME": "Rule_20"
+                    }
+                ]
+            },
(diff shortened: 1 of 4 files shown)


src/sonic-yang-models/yang-models/sonic-acl.yang
```diff
@@ -139,7 +151,7 @@ module sonic-acl {
 
 				leaf ETHER_TYPE {
 					type string {
-						pattern "(0x88CC|0x8100|0x8915|0x0806|0x0800|0x86DD|0x8847)";
+					    pattern "0x[6-9a-fA-F][0-9a-fA-F]{2,3}|0x[1-9a-fA-F][0-9a-fA-F]{3,4}";
```
bingwang-ms, line 154: Why the ```ETHER_TYPE``` filed is updated?

abhishekd0: For L2ACL user can specify any valid ethertype for the L2 traffic filtering.

bingwang-ms: I see. Thanks

bingwang-ms (approved):

zhangyanzhao:
branch conflicts need be resolved before merge.

qiluo-msft (dismissed):

src/sonic-yang-models/yang-templates/sonic-acl.yang.j2
```diff
@@ -75,8 +75,20 @@ module sonic-acl {
...
 				}
 
-				choice ip_prefix {
-
+				choice src_dst_address {
+					case l2_src_dst_address {
+						when "(/acl:sonic-acl/acl:ACL_TABLE/acl:ACL_TABLE_LIST[ACL_TABLE_NAME=current()/acl:ACL_TABLE_NAME]/acl:type = 'L2')";
+						leaf SRC_MAC {
+							type string {
+								pattern "[0-9a-fA-F]{2}(:[0-9a-fA-F]{2}){5}|[0-9a-fA-F]{2}(:[0-9a-fA-F]{2}){5}/[0-9a-fA-F]{2}(:[0-9a-fA-F]{2}){5}";
```
qiluo-msft, line 83: >pattern [](http://example.com/codeflow?start=8&length=7)

Should we define a shared type, and name it like stypes:mac_and_mask?
I see the pattern repeated below.

qiluo-msft:
@bingwang-ms Could you also help review?

src/sonic-yang-models/yang-templates/sonic-acl.yang.j2
abhishekd0: Added common type and used it in acl.yang

src/sonic-yang-models/yang-templates/sonic-acl.yang.j2
```diff
@@ -144,7 +152,7 @@ module sonic-acl {
 
 				leaf ETHER_TYPE {
 					type string {
-						pattern "(0x88CC|0x8100|0x8915|0x0806|0x0800|0x86DD|0x8847)";
+					    pattern "0x[6-9a-fA-F][0-9a-fA-F]{2,3}|0x[1-9a-fA-F][0-9a-fA-F]{3,4}";
```
qiluo-msft, line 155: >pattern [](http://example.com/codeflow?start=9&length=7)

This change is not related to your PR title or description. And seems the old code is better.
Could you focus on MAC part as original commit?

src/sonic-yang-models/yang-templates/sonic-acl.yang.j2
```diff
@@ -220,6 +228,26 @@ module sonic-acl {
 				leaf INNER_L4_DST_PORT {
 					type uint16;
 				}
+
+				leaf VLAN_ID {
```
qiluo-msft, line 232: >VLAN_ID [](http://example.com/codeflow?start=9&length=7)

This change is not related to your PR title or description. 
Could you focus on MAC part as original commit?

src/sonic-yang-models/yang-templates/sonic-acl.yang.j2
abhishekd0: This was discussed earlier also. Please see https://github.com/Azure/sonic-buildimage/pull/7917#discussion_r656718927. User should be able to specify any valid Ethertype for L2 ACL.

src/sonic-yang-models/yang-templates/sonic-acl.yang.j2
abhishekd0: VLAN ID is part of the MAC header. This PR adds various other MAC header keys like PCP DEI etc. I will update the description to explicitly list all added fields

src/sonic-yang-models/yang-templates/sonic-acl.yang.j2
```diff
@@ -220,6 +228,26 @@ module sonic-acl {
...
+					type uint16 {
+						range 1..4093;
```
qiluo-msft, line 234: >4093 [](http://example.com/codeflow?start=15&length=4)

4094 should be good?
How about `0..4095`? the borders are revered but still valid.

src/sonic-yang-models/yang-templates/sonic-acl.yang.j2
qiluo-msft: Also update the PR title?

src/sonic-yang-models/yang-templates/sonic-acl.yang.j2
abhishekd0: Updated to 1-4094 which matches the "config vlan" CLI from [sonic-utilities](https://github.com/Azure/sonic-utilities/blob/1ea88e21f72ccbc73e18d9fa5d1223b378000490/utilities_common/cli.py#L246) .  0 is reserved for priority tag and will be assigned PVID which will be in range 1-4094. 4095 is reserved as well .

src/sonic-yang-models/yang-templates/sonic-acl.yang.j2
abhishekd0: Done

src/sonic-yang-models/yang-templates/sonic-acl.yang.j2
ganglyu: Is 0x10000 acceptable?
Should we define a shared type?

abhishekd0: Fixed regex.  This is used only once. Can be moved in future to a shared type if its used/repeated multiple times.

ganglyu (dismissed):

src/sonic-yang-models/yang-templates/sonic-acl.yang.j2
ganglyu: 0x0800 was acceptable, but not now.

src/sonic-yang-models/yang-templates/sonic-acl.yang.j2
```diff
@@ -144,7 +152,7 @@ module sonic-acl {
 
 				leaf ETHER_TYPE {
 					type string {
-						pattern "(0x88CC|0x8100|0x8915|0x0806|0x0800|0x86DD|0x8847)";
+					    pattern "0x[6-9a-fA-F][0-9a-fA-F]{2}|0x[1-9a-fA-F][0-9a-fA-F]{3}";
```
ganglyu, line 155: 0x0800 was acceptable, but not now.

abhishekd0: Added only those 2 to pattern to maintain backwards compatibility

src/sonic-yang-models/yang-templates/sonic-acl.yang.j2
```diff
@@ -144,7 +152,7 @@ module sonic-acl {
 
 				leaf ETHER_TYPE {
 					type string {
-						pattern "(0x88CC|0x8100|0x8915|0x0806|0x0800|0x86DD|0x8847)";
+					    pattern "0x[6-9a-fA-F][0-9a-fA-F]{2}|0x[1-9a-fA-F][0-9a-fA-F]{3}|0x0806|0x0800";
```
ganglyu, line 155: 0x0[6-9a-fA-F][0-9a-fA-F]{2}

src/sonic-yang-models/tests/yang_model_tests/tests_config/acl.json
```diff
@@ -657,5 +657,123 @@
...
+    },
+    "ACL_TABLE_L2_ACL_FIELDS": {
+        "sonic-acl:sonic-acl": {
+            "sonic-acl:ACL_RULE": {
+                "ACL_RULE_LIST": [
+                    {
+                        "ACL_TABLE_NAME": "L2ACL",
+                        "SRC_MAC": "00:00:AB:CD:EF:00/FF:FF:FF:00:00:00",
+                        "DST_MAC": "00:00:AB:CD:EF:FF/FF:FF:FF:FF:FF:FF",
+                        "ETHER_TYPE": "0x800",
```
ganglyu, line 669: Can this field pass the UT?

src/sonic-yang-models/yang-templates/sonic-acl.yang.j2
abhishekd0: Leading zeros as no significance and this is represented as hex number. Requester has to take the extra step to add a 0 which is not useful

src/sonic-yang-models/tests/yang_model_tests/tests_config/acl.json
abhishekd0: updated

ganglyu (approved):

qiluo-msft (approved):

PR #7920, opened 2021-06-19
Check platform reboot cause to see if any reset happened during fast/warm-reboot

sujinmkang:
#### Why I did it
Check platform reboot cause to see if any reset happened during fast/warm-reboot 
While starting dockers with boot_type, it's only checking the /proc/cmdline.
If any unexpected Platform reset happens during fast/warm-reboot, it should be detected as platform reboot 
and "FAST_REBOOT|system" should be set in that case.
#### How I did it
Add the platform reboot cause parsing when determining the boot_type
#### How to verify it
Check the logic with reboot
Convert the device with fast-reload and see if the boot_type is detected correctly

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012

### DIFF
--- a/files/build_templates/docker_image_ctl.j2
+++ b/files/build_templates/docker_image_ctl.j2
@@ -60,6 +60,17 @@ function getMountPoint()
     echo $1 | python -c "import sys, json, os; mnts = [x for x in json.load(sys.stdin)[0]['Mounts'] if x['Destination'] == '/usr/share/sonic/hwsku']; print '' if len(mnts) == 0 else os.path.abspath(mnts[0]['Source'])" 2>/dev/null
 }
 
+function isPlatformReset()
+{
+    output=$(python3 -c "import sonic_platform.platform; p = sonic_platform.platform.Platform(); c = p.get_chassis(); hw_rc_major, hw_rc_minor = c.get_reboot_cause(); print(hw_rc_major)" 2>/dev/null)
+    echo "${output}"
+    if [[ "${output}" != "Non-Hardware" ]]; then
+        return 0
+    else
+        return 1
+    fi
+}
+
 function getBootType()
 {
     # same code snippet in files/scripts/syncd.sh
@@ -76,6 +87,9 @@ function getBootType()
     *)
         TYPE='cold'
     esac
+    if isPlatformReset; then
+        TYPE='cold'
+    fi
     echo "${TYPE}"
 }
 
--- a/files/build_templates/docker_image_ctl.j2
+++ b/files/build_templates/docker_image_ctl.j2
@@ -64,9 +64,12 @@ function getPlatformRebootCause()
 {
     IMPORT_ERR="*ImportError*"
     err=$(python -c "import sonic_platform.platform" 2>&1 > /dev/null)
+    err2=$(python3 -c "import sonic_platform.platform" 2>&1 > /dev/null)
+    output="Non-Hardware"
     if [[ "$err" =~ "$IMPORT_ERR"]]; then
          output=$(python3 -c "import sonic_platform.platform; p = sonic_platform.platform.Platform(); c = p.get_chassis(); hw_rc_major, hw_rc_minor = c.get_reboot_cause(); print(hw_rc_major)" 2>/dev/null)
-    else
+    fi
+    if [[ "$err2" =~ "$IMPORT_ERR"]]; then
          output=$(python -c "import sonic_platform.platform; p = sonic_platform.platform.Platform(); c = p.get_chassis(); hw_rc_major, hw_rc_minor = c.get_reboot_cause(); print(hw_rc_major)" 2>/dev/null)
     fi
     if [[ "${output}" == "Non-Hardware" ]]; then
--- a/files/build_templates/docker_image_ctl.j2
+++ b/files/build_templates/docker_image_ctl.j2
@@ -112,6 +112,14 @@ function preStartAction()
     fi
 {%- elif docker_container_name == "snmp" %}
     $SONIC_DB_CLI STATE_DB HSET 'DEVICE_METADATA|localhost' chassis_serial_number $(decode-syseeprom -s)
+{%- elif docker_container_name == "swss" %}
+    if [[ "$BOOT_TYPE" == "fast" ]] && [[ -d /host/fast-reboot ]]; then
+        # Clear the FAST_REBOOT|system db setting if PLATFORM_REBOOT_CAUSE is not "software" before starting swss
+        if  [[ "$PLATFORM_REBOOT_CAUSE" != "software" ]]; then
+            # Delete the FAST_REBOOT|system db setting 
+            $SONIC_DB_CLI STATE_DB DEL "FAST_REBOOT|system" &>/dev/null
+        fi
+    fi
 {%- else %}
     : # nothing
 {%- endif %}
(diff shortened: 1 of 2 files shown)


files/build_templates/docker_image_ctl.j2
```diff
@@ -60,6 +60,17 @@ function getMountPoint()
     echo $1 | python -c "import sys, json, os; mnts = [x for x in json.load(sys.stdin)[0]['Mounts'] if x['Destination'] == '/usr/share/sonic/hwsku']; print '' if len(mnts) == 0 else os.path.abspath(mnts[0]['Source'])" 2>/dev/null
 }
 
+function isPlatformReset()
+{
+    output=$(python3 -c "import sonic_platform.platform; p = sonic_platform.platform.Platform(); c = p.get_chassis(); hw_rc_major, hw_rc_minor = c.get_reboot_cause(); print(hw_rc_major)" 2>/dev/null)
+    echo "${output}"
```
jleveque, line 66: Maybe this should echo more context, rather than just the reboot cause?

files/build_templates/docker_image_ctl.j2
```diff
@@ -60,6 +60,17 @@ function getMountPoint()
...
+    if [[ "${output}" != "Non-Hardware" ]]; then
+        return 0
+    else
+        return 1
```
jleveque, line 70: Maybe add a comment here to explain this logic?

files/build_templates/docker_image_ctl.j2
santhosh-kt: I commented about it yesterday but it said pending status.

santhosh-kt:
Uploaded UT test logs in https://github.com/Azure/sonic-buildimage/pull/8024 for your reference.
(OS9 to SONIC - cold reset.txt)
(OS9 to SONIC - fastboot.txt)

files/build_templates/docker_image_ctl.j2
```diff
@@ -60,6 +60,16 @@ function getMountPoint()
     echo $1 | python -c "import sys, json, os; mnts = [x for x in json.load(sys.stdin)[0]['Mounts'] if x['Destination'] == '/usr/share/sonic/hwsku']; print '' if len(mnts) == 0 else os.path.abspath(mnts[0]['Source'])" 2>/dev/null
 }
 
+function isPlatformReset()
+{
+    output=$(python3 -c "import sonic_platform.platform; p = sonic_platform.platform.Platform(); c = p.get_chassis(); hw_rc_major, hw_rc_minor = c.get_reboot_cause(); print(hw_rc_major)" 2>/dev/null)
```
santhosh-kt, line 65: @sujinmkang : On checking the logs double time, found out one of the process(database.sh) is mentioning it as cold reset despite fast-boot happened. I suspect this is because either determine-reboot-cause or process-reboot-cause function is not up until that time?
Will there be any impact if database is still considering the BOOT_TYPE as cold?
```
sonic database.sh[680]: BOOT_TYPE  we received cold
sonic lldp.sh[2314]: BOOT_TYPE  we received fast
sonic pmon.sh[2308]: BOOT_TYPE  we received fast
sonic-s6100-01 bgp.sh[2886]: BOOT_TYPE  we received fast
sonic-s6100-01 swss.sh[2887]: BOOT_TYPE  we received fast
sonic-s6100-01 teamd.sh[3655]: BOOT_TYPE  we received fast
sonic-s6100-01 snmp.sh[3658]: BOOT_TYPE  we received fast
sonic-s6100-01 radv.sh[3651]: BOOT_TYPE  we received fast
sonic-s6100-01 syncd.sh[3656]: BOOT_TYPE  we received fast
sonic-s6100-01 dhcp_relay.sh[4441]: BOOT_TYPE  we received fast
```

sujinmkang: @santhosh-kt we're not using determine-reboot-cause or process-reboot-cause for this determination but only using platform api. Probably does the database always consider the boot_type as cold even without this change? Since we're sending the same boot_type to all container starting script. Right?

santhosh-kt: Sorry! I misunderstood. Except database.sh script all other daemon scripts are called lately if I understood. 

So all those scripts(except database.sh) could have easily interpreted the boot type since get_reboot_cause() is showing correct value as track_reboot_reason.sh script got executed before these scripts.

But for database.sh, this get_reboot_cause() is showing wrong value as it runs even before s6100_platform.sh itself.
My question is, will wrong BOOT_TYPE has any impact in database.sh script?

sujinmkang: @santhosh-kt can we detect the platform driver got initialized or not? Or should we check the exception from platform api call to see the platform driver initialization status? Any other suggestion?

sujinmkang: @santhosh-kt I checked the database container start code, I don't think the boot_type is be used for database container starts.

sujinmkang: @yxieca I don't think the database container is starting differently based on the boot-type. do you have any comments or concern?

yxieca: Database container start code has a special handling during warm reboot: https://github.com/Azure/sonic-buildimage/blob/master/files/build_templates/docker_image_ctl.j2#L87

sujinmkang:
@yxieca Since the platform driver is not ready by database container starts and we are sure the platform driver is ready before swss starting, I changed the code to Delete "FAST_REBOOT|system" key when swss starts instead of checking the platform reset in database starts. What do you think this change?

yxieca:
> @yxieca Since the platform driver is not ready by database container starts and we are sure the platform driver is ready before swss starting, I changed the code to Delete "FAST_REBOOT|system" key when swss starts instead of checking the platform reset in database starts. What do you think this change?

Is this question related to this PR? I didn't find the change matches this question?

sujinmkang:
@santhosh-kt Can you please verify this new PR with the scenario? 
@yxieca  Can you please review the logic to detect the platform reset/reboot cause during fast-reload/reboot case?

santhosh-kt:
Hi @sujinmkang : Following changes can be tried out in master image but as I mentioned earlier, latest master raw image will expand upto 3GB and it will cause disk space issue while conversion.
I see to try the same in 201811 but we don't have determine-reboot-cause changes as well as no support on previous-reboot-cause.json.

sujinmkang:
@santhosh-kt I will create a new PR for 201911

sujinmkang:
@santhosh-kt Can you please give me some more information about the 3G image issue? Is the issue still existing without my PR or does it happen with my PR?

files/build_templates/docker_image_ctl.j2
```diff
@@ -95,6 +95,23 @@ function preStartAction()
...
+            REG_BOOT_TYPE="fast*"
+	    CAUSE_NO_AVAIL="\"N/A\""
+            REBOOT_CAUSE="$(cat /host/reboot-cause/previous-reboot-cause.json | jq '.cause')"
+            EXTRA_CAUSE="$(cat /host/reboot-cause/previous-reboot-cause.json | jq '.comment')"
+
+            # Clear the FAST_REBOOT|system db setting if EXTRA_REBOOT_CAUSE is not "N/A" before starting swss
+            if [[ $REBOOT_CAUSE =~ $REG_BOOT_TYPE ]]; then
+                if [[ "${EXTRA_CAUSE}" != "${CAUSE_NO_AVAIL}" ]]; then
+                    # Delete the FAST_REBOOT|system db setting
+                    $SONIC_DB_CLI STATE_DB DEL "FAST_REBOOT|system" &>/dev/null
```
yxieca, line 110: How can you guarantee that syncd didn't take fast-reboot approach? syncd could have read this value before swss deletes it.

files/build_templates/docker_image_ctl.j2
```diff
@@ -95,6 +95,23 @@ function preStartAction()
     fi
 {%- elif docker_container_name == "snmp" %}
     $SONIC_DB_CLI STATE_DB HSET 'DEVICE_METADATA|localhost' chassis_serial_number $(decode-syseeprom -s)
+{%- elif docker_container_name == "swss" %}
+    if [[ "$BOOT_TYPE" == "fast" ]] && [[ -d /host/fast-reboot ]]; then
+        if [[ -f /host/reboot-cause/previous-reboot-cause.json ]]; then
+            REG_BOOT_TYPE="fast*"
+	    CAUSE_NO_AVAIL="\"N/A\""
```
yxieca, line 102: can you use white space instead of tab?

sujinmkang:
Closing this PR since we don't need this PR for master branch

PR #7927, opened 2021-06-21
[Mellanox] Return N/A for PSU's model, serial and revision on platforms with fixed PSU

stephenxs:

#### Why I did it
The methods `get_model`, `get_serial`, and `get_revision` have been implemented by reading relevant information from VPD and then recording the information into relevant fields.
However, there is no VPD data on platforms with fixed PSUs and relevant fields haven't been initialized, which causes the methods to throw exceptions. which in turn prevents `psud` from inserting fields into PSU table.
Eventually, this causes `show platform psustatus` doesn't output correct info.
```
admin@sonic:~$ show platform psustatus 
PSU    Model    Serial    HW Rev    Voltage (V)    Current (A)    Power (W)    Status    LED
-----  -------  --------  --------  -------------  -------------  -----------  --------  -----
PSU 1                                                                          OK        green
PSU 2                                                                          OK        green
```

#### How I did it
Initliaze those fields as `N/A` on systems with fixed PSUs.

#### How to verify it
Manually test.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012

### DIFF
--- a/platform/mellanox/mlnx-platform-api/sonic_platform/psu.py
+++ b/platform/mellanox/mlnx-platform-api/sonic_platform/psu.py
@@ -88,6 +88,10 @@ class Psu(PsuBase):
         self.psu_data = DEVICE_DATA[platform]['psus']
         psu_vpd = filemap[PSU_VPD]
 
+        self.model = "N/A"
+        self.serial = "N/A"
+        self.rev = "N/A"
+
         if psu_vpd is not None:
             self.psu_vpd = os.path.join(self.psu_path, psu_vpd.format(self.index))
             self.vpd_data = self._read_vpd_file(self.psu_vpd)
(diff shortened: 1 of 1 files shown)


keboliu (dismissed):

stephenxs:
Vs test failed due to failure when setting up one of the VMs.
Need to trigger rerun

jleveque (approved):

platform/mellanox/mlnx-platform-api/sonic_platform/psu.py
```diff
@@ -112,7 +112,9 @@ def __init__(self, psu_index, platform):
 
         else: 
             logger.log_info("Not reading PSU{} VPD data: Platform is fixed".format(self.index))
-
+            self.model = "N/A"
```
robocoder99, line 115: This would be better put at around Line 94 as general best practice. This will ensure that no matter how the code changes below if it ever does these values will always be initialized. 

Thanks for the fix!

robocoder99 (changes requested):

platform/mellanox/mlnx-platform-api/sonic_platform/psu.py
robocoder99: You can also then remove my defaults at Lines 98 / 104 / 110 as well.

stephenxs: Fixed

jleveque (approved):

jleveque:
@robocoder99: Please review again.

robocoder99 (approved):

PR #7939, opened 2021-06-22
[sonic-yang] Backlinks are none when there is 0 backlinks

ghooo:
#### Why I did it
Backlinks are none when there is 0 backlinks. without this change `backlinks.number()` throws an exception when backlinks==None.

#### How I did it
Allow backlinks to be None, so we get an empty list as a return from find_data_dependencies


lguohan:
@ghooo , can you push to your own fork, not azure repo.

PR #7708, opened 2021-05-25
[Mellanox] Update SKUs to enable SDK dumps

DavidZagury:
This PR depends on https://github.com/Azure/sonic-sairedis/pull/833
#### Why I did it
To create SDK dump on Mellanox devices when SDK event has occurred.

#### How I did it
Set the SKUs keys needed to initialize the feature in SAI.

#### How to verify it
Simulate SDK event and check that dump is created in the expected path.

#### Which release branch to backport (provide reason below if selected)

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012

PR #7978, opened 2021-06-25
[Mellanox] Update SKUs to enable SDK dumps (#7708)

DavidZagury:
#### Why I did it
To create SDK dump on Mellanox devices when SDK event has occurred.

#### How I did it
Set the SKUs keys needed to initialize the feature in SAI.

#### How to verify it
Simulate SDK event and check that dump is created in the expected path.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012


liat-grozovik:
Create PR for 202012 as https://github.com/Azure/sonic-buildimage/pull/7708 cannot be cleanly cherry picked.

PR #7997, opened 2021-06-28
Added retry to sonic-mgmt docker container

slutati1536:
Signed-off-by: Sharon Lutati <slutati@nvidia.com>

#### Why I did it

the motivation for this PR is to add retry_call to several test cases in the community, for example, the following cases:

- test_show_platform_fanstatus_mocked
- test_show_platform_temperature_mocked

are executing a command once and comparing the output to the expected mock data,
sometimes differences between the mock and the actual are causing the tests to fail. 

retry will make these tests more stable.
retry will also be more efficient than sleep which will cause the tests to run longer because sometimes it is not necessary to sleep all that time, retry will only run a function only until it passed.

#### How I did it

added retry to the docker file

#### How to verify it
I run the tests with retry on the docker after installing the retry package 

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012

adding retry to dockerfile so that retry module will be supported on sonic-mgmt

### DIFF
--- a/dockers/docker-sonic-mgmt/Dockerfile.j2
+++ b/dockers/docker-sonic-mgmt/Dockerfile.j2
@@ -69,6 +69,7 @@ RUN pip install cffi==1.10.0 \
                 tabulate \
                 textfsm \
                 virtualenv \
+		retry \
     && git clone https://github.com/p4lang/scapy-vxlan.git \
     && cd scapy-vxlan \
     && python setup.py install \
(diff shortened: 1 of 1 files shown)


dockers/docker-sonic-mgmt/Dockerfile.j2
```diff
@@ -69,6 +69,7 @@ RUN pip install cffi==1.10.0 \
                 tabulate \
                 textfsm \
                 virtualenv \
+		retry \
```
keboliu, line 72: please make the ident aligned.

yxieca: Please use white spaces instead of tabs. If this was the problem.

slutati1536: @keboliu @yxieca I fixed the indentation in file

yxieca (approved):

PR #8005, opened 2021-06-29
[Arista] Update phy-credo service for config load

byu343:
Signed-off-by: Boyang Yu <byu@arista.com>

#### Why I did it
phy-credo.service was only run once in booting. With this change, it will be restarted when running 'config reload'

#### How I did it
Change the service from WantedBy=multi-user.target to wantedBy=sonic.target

#### How to verify it
Checked that service is restarted after running 'config reload -y', and the ports are still brought up.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012


lguohan (approved):

arlakshm:
@byu343, is this change need for chassis Linecard. Do we need to cherry-pick this change in `202106`?

byu343:
@arlakshm, yes, it is. And we need it in 202106.

PR #7857, opened 2021-06-12
[vs] add information about features to config db

stepanblyschak:
vs has components from swss, bgp, teamd and nat. This table is needed by this change https://github.com/Azure/sonic-utilities/pull/1554.

Signed-off-by: Stepan Blyschak <stepanb@nvidia.com>

#### Why I did it
Because https://github.com/Azure/sonic-utilities/pull/1554 requires "config warm_restart enable FEATURE" and the FEATURE has to be in feature table.

#### How I did it
Added FEATURE table to VS.

#### How to verify it
Run VS tests.

#### Which release branch to backport (provide reason below if selected)

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012

PR #8008, opened 2021-06-29
[vstes] Virtual chssis config rendering from default_config.json

vganesan-nokia:
#### Why I did it

To fix VS virtual_chassis tests failure issue https://github.com/Azure/sonic-swss/issues/1809

#### How I did it

There was problem in rendering virtual chassis default_config.json into config_db.json. This problem was introduced by changes in PR https://github.com/Azure/sonic-buildimage/pull/7857. The problem was that introduction of using init_cfg.json.j2 template did not have instruction in init_config.json.j2 to merge the virtual chassis default_config.json. Ths fix is to have a separate invokiation of sonic-cfggen to combine init_config.json and virtual chassis default_config.json
 
#### How to verify it

VS test for test_virtual_chassis.py passes all tests consistently.

### DIFF
--- a/platform/vs/docker-sonic-vs/start.sh
+++ b/platform/vs/docker-sonic-vs/start.sh
@@ -27,17 +27,18 @@ popd
 
 [ -d /etc/sonic ] || mkdir -p /etc/sonic
 
-if [[ -f /usr/share/sonic/virtual_chassis/default_config.json ]]; then
-    CHASS_CFG="-j /usr/share/sonic/virtual_chassis/default_config.json"
-fi
-
 # Note: libswsscommon requires a dabase_config file in /var/run/redis/sonic-db/
 # Prepare this file before any dependent application, such as sonic-cfggen
 mkdir -p /var/run/redis/sonic-db
 cp /etc/default/sonic-db/database_config.json /var/run/redis/sonic-db/
 
 SYSTEM_MAC_ADDRESS=$(ip link show eth0 | grep ether | awk '{print $2}')
-sonic-cfggen -t /usr/share/sonic/templates/init_cfg.json.j2 -a "{\"system_mac\": \"$SYSTEM_MAC_ADDRESS\"}" $CHASS_CFG > /etc/sonic/init_cfg.json
+sonic-cfggen -t /usr/share/sonic/templates/init_cfg.json.j2 -a "{\"system_mac\": \"$SYSTEM_MAC_ADDRESS\"}" > /etc/sonic/init_cfg.json
+
+if [[ -f /usr/share/sonic/virtual_chassis/default_config.json ]]; then
+    sonic-cfggen -j /etc/sonic/init_cfg.json -j /usr/share/sonic/virtual_chassis/default_config.json --print-data > /tmp/init_cfg.json
+    mv /tmp/init_cfg.json /etc/sonic/init_cfg.json
+fi
 
 if [ -f /etc/sonic/config_db.json ]; then
     sonic-cfggen -j /etc/sonic/init_cfg.json -j /etc/sonic/config_db.json --print-data > /tmp/config_db.json


platform/vs/docker-sonic-vs/start.sh
```diff
@@ -27,17 +27,18 @@ popd
...
 # Prepare this file before any dependent application, such as sonic-cfggen
 mkdir -p /var/run/redis/sonic-db
 cp /etc/default/sonic-db/database_config.json /var/run/redis/sonic-db/
 
 SYSTEM_MAC_ADDRESS=$(ip link show eth0 | grep ether | awk '{print $2}')
-sonic-cfggen -t /usr/share/sonic/templates/init_cfg.json.j2 -a "{\"system_mac\": \"$SYSTEM_MAC_ADDRESS\"}" $CHASS_CFG > /etc/sonic/init_cfg.json
+sonic-cfggen -t /usr/share/sonic/templates/init_cfg.json.j2 -a "{\"system_mac\": \"$SYSTEM_MAC_ADDRESS\"}" > /etc/sonic/init_cfg.json
+
+if [[ -f /usr/share/sonic/virtual_chassis/default_config.json ]]; then
+    sonic-cfggen -j /etc/sonic/init_cfg.json -j /usr/share/sonic/virtual_chassis/default_config.json --print-data > /tmp/init_cfg.json
```
prsunny, line 39: why dont we write directly to /etc/sonic/init_cfg.json

prsunny:
@stepanblyschak , can you please review?

platform/vs/docker-sonic-vs/start.sh
vganesan-nokia: The template /usr/share/sonic/templates/init_cfg.json.j2 does not have j2 code to take the /usr/share/sonic/virtual_chassis/default_config.json contents as data and generate the result with the default_config,json contents in /etc/sonic/init_cfg.json

prsunny: ok, but do you need to write to /tmp/init_cfg.json or can you update init_cfg.json?

vganesan-nokia: I think it is possible to update the init_cfg.json directly since sonic-cfggen loads all the contents first as data and then produces the output). But for here, I just followed the model used for generating config_db.json (given just below these lines from 43-45) to appear consistent

prsunny (approved):

platform/vs/docker-sonic-vs/start.sh
stepanblyschak: @vganesan-nokia Does this mean that "-t" & "-j" does not work togather in sonic-cfggen?

```
-j JSON, --json JSON  json file that contains additional variables
```

```
-t TEMPLATE, --template TEMPLATE
                        render the data with the template file
```

vganesan-nokia: Yes. When -t option is used the template file, the contents of the json file given with option -j are loaded as data to the template. If the template does not have explicit j2 code to take these data, the rendering will not have the json files contents in the output.

stepanblyschak: Thanks for details, now I understand.

stepanblyschak (approved):

vganesan-nokia:
@prsunny, would you please merge this PR so that we can merge the dependent PR Azure/sonic-swss#1811? Once both of these PRs are merged, we can add back the skipped virtual chassis tests.

vganesan-nokia:
@prsunny, @anshuv-mfst, would you please make arrangements to get this PR merged? Merging this will enable merging Azure/sonic-swss#1811 and adding back the skipped virtual chassis tests in test_virtual_chassis.py. I am also preparing a UT update PR as requested in PR Azure/sonic-swss#1823, which requires enabling back the skipped tests in test_viratual_chassis.py.
Thanks

PR #8012, opened 2021-06-29
[Barefoot] Enable 'TELEMETRY_WRITABLE' for telemetry container

YuliiaBashko:
#### Why I did it
For testing `gNMI/gNOI` we need to enable r/w mode for telemetry container

#### How I did it
Added `TELEMETRY_WRITABLE=y` option to `make` command
#### How to verify it
Build SONiC with command `make TELEMETRY_WRITABLE=y target/sonic-aboot-barefoot.swi`

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012

### DIFF
--- a/.azure-pipelines/azure-pipelines-build.yml
+++ b/.azure-pipelines/azure-pipelines-build.yml
@@ -48,7 +48,7 @@ jobs:
             docker_syncd_rpc_image: yes
             platform_rpc: bfn
             swi_image: yes
-            BUILD_OPTIONS: $(BUILD_OPTIONS) TELEMETRY_WRITABLE=y
+            BUILD_OPTIONS: ${{ parameters.buildOptions }} TELEMETRY_WRITABLE=y
 
         - name: broadcom
           variables:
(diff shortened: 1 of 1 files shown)


YuliiaBashko:
@xumia Could you please take a look?

.azure-pipelines/azure-pipelines-build.yml
```diff
@@ -48,6 +48,7 @@ jobs:
             docker_syncd_rpc_image: yes
             platform_rpc: bfn
             swi_image: yes
+            BUILD_OPTIONS: $(BUILD_OPTIONS) TELEMETRY_WRITABLE=y
```
xumia, line 51: Do you mean as below?
BUILD_OPTIONS: ${{ parameters.buildOptions }} TELEMETRY_WRITABLE=y

YuliiaBashko: I see, thanks, I\`ll fix this. So it means that my change wouldn\`t  work at all?

akokhan (approved):

YuliiaBashko:
@xumia Could you please check my new change?

lguohan:
telemtry_writable should be consistent across different platform. we cannot enable write only on a particular platform.

lguohan:
that is not the current plan. the viable solution is to make the telemetry writable as a runtime option, then user can enable it on the fly and no longer a build option.

PR #8015, opened 2021-06-29
[ci]: Disable nephos docker syncd rpc temporarily

xumia:
#### Why I did it
Disable the nephos docker syncd rpc image temporarily

The package libboost-atomic1.71.0 does not exist in the debian:stretch, it has no chance to build a successful rpc image.
Revert the change until the rpc image upgraded to buster.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012

### DIFF
--- a/.azure-pipelines/azure-pipelines-build.yml
+++ b/.azure-pipelines/azure-pipelines-build.yml
@@ -93,7 +93,7 @@ jobs:
         - name: nephos
           variables:
             dbg_image: yes
-            docker_syncd_rpc_image: yes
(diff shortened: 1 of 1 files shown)


xumia:
Another approach is to revert the features that have dependency on libboost-atomic1.71.0, and remove the package in the Dockerfile.

xumia:
@lguohan , can we disable the nephos docker syncd rpc temporarily? We need to make sure the build work, it blocks for few weeks.

qiluo-msft (approved):

PR #8034, opened 2021-07-01
Replace swsssdk with swsscommon in sonic-host-services

qiluo-msft:
#### Why I did it
swsssdk will be deprecated. Use swsscommon instead.

#### How to verify it
Unit test

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [x] 202106


src/sonic-host-services/scripts/hostcfgd
```diff
@@ -9,7 +9,7 @@ import syslog
 
 import jinja2
 from sonic_py_common import device_info
-from swsssdk import ConfigDBConnector
+from swsscommon import swsscommon
```
sujinmkang, line 12: Can we just import only ConfigDBConnector if we don't need whole swsscommon module then we don't need any other changes in this file?
from swsscommon.swsscommon import ConfigDBConnector

qiluo-msft: Resolved.

sujinmkang (approved):

PR #8049, opened 2021-07-02
[multiasic][voq][bgpconf] Fix for the issue of same BGP router id in all asics

vganesan-nokia:
#### Why I did it

For multiasic, the back end asics use ip addresss of Loopback4096 for BGP router id. In VOQ multi-asic chassis there are no back end asics. All the asics are front end and the iBGP connections are established via Ethernet-IB of asics. Since these asics are not designated as BackEnd, the ip address of interface Loopback0 is used as BGP router id. Since the ip address of Loopback0 is same for all the asics in the line card, same router id is used for voq iBGP configurations and hence the iBGP connections are not established. Changes are done to fix this

#### How I did it

To fix this (i.e., to have different router ids for differen iBGP peers), we need to use the ip address of Loopback4096 for
the front end asics also in VOQ chassis systems. The bgp main configuration template is adjusted to take ip address of Loopback4096 as router id for asics (both in single asic line card and multi-asic line card) in chassis voq systems irrespective of whether the asic is front end or not.

#### How to verify it

- In  multi-asic voq chassis system have the following in the config_db.json of multiple asics
- Configure different IP addresses for Loopback4096 loop back interface in different asics
- Configure BGP_VOQ_CHASSIS_NEIGHBORS for the Ethernet-IB interfaces.
- Reboot the chassis
- Observe that all iBGP sessions are established state.

### DIFF
--- a/dockers/docker-fpm-frr/frr/bgpd/bgpd.main.conf.j2
+++ b/dockers/docker-fpm-frr/frr/bgpd/bgpd.main.conf.j2
@@ -60,7 +60,7 @@ router bgp {{ DEVICE_METADATA['localhost']['bgp_asn'] }}
 {% endif %}
 !
 {# set router-id #}
-{% if DEVICE_METADATA['localhost']['sub_role'] == 'BackEnd' %}
+{% if DEVICE_METADATA['localhost']['sub_role'] == 'BackEnd' or  (multi_asic is defined and DEVICE_METADATA['localhost']['switch_type'] == 'voq') %}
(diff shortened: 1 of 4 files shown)


abdosi (changes requested):
please add unit-test case to address template changes

vganesan-nokia:
> please add unit-test case to address template changes

Added unit tests.

abdosi (approved):

arlakshm (approved):

PR #8063, opened 2021-07-05
[Mellanox][pcied] Ignore bus on pcie.yaml for Mellanox switches

DavidZagury:
#### Why I did it
BIOS upgrade on rare cases cannot guarantee bus value remain the same on every BIOS release. Ignoring this field in order for pcied not to fail but still verify device id in a different way. The solution is future proof and will not require changes in code when new BIOS version is available

#### How I did it
Since bus is not a fixed value (it is determined by the bios version) we are ignoring this field, and instead checking if there is a device that match on all other fields that and in addition has a matching device id.

#### How to verify it
Verify no errors or failures in pcied on different BIOS version with the same code base.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [x] 202106

### DIFF
--- a/platform/mellanox/mlnx-platform-api/sonic_platform/pcie.py
+++ b/platform/mellanox/mlnx-platform-api/sonic_platform/pcie.py
@@ -1,0 +1,69 @@
+########################################################################
+#
+# Copyright (C) 2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# Module contains a platform specific implementation of SONiC Platform
+# Base PCIe class
+#
+########################################################################
+import os
+import re
+
+try:
+    from sonic_platform_base.sonic_pcie.pcie_common import PcieUtil
+except ImportError as e:
+    raise ImportError(str(e) + "- required module not found")
+
+SYSFS_PCI_DEVICE_PATH = '/sys/bus/pci/devices/'
+
+
+class Pcie(PcieUtil):
+    # check the current PCIe device with config file and return the result
+    # use bus from _device_id_to_bus_map instead of from yaml file
+    def get_pcie_check(self):
+        self.load_config_file()
+        for item_conf in self.confInfo:
+            id_conf = item_conf["id"]
+            dev_conf = item_conf["dev"]
+            fn_conf = item_conf["fn"]
+            bus_conf = self._device_id_to_bus_map.get(id_conf)
+            if bus_conf and self.check_pcie_sysfs(bus=int(bus_conf, base=16), device=int(dev_conf, base=16),
+                                                  func=int(fn_conf, base=16)):
+                item_conf["result"] = "Passed"
+            else:
+                item_conf["result"] = "Failed"
+        return self.confInfo
+
+    # Create
+    def _create_device_id_to_bus_map(self):
+        self._device_id_to_bus_map = {}
+        self.load_config_file()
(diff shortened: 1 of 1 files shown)


BasimShalata:
@sujinmkang can you please review.
its based on your changes and as we discussed in our previous meeting

keboliu (approved):

platform/mellanox/mlnx-platform-api/sonic_platform/pcie.py
```diff
@@ -0,0 +1,54 @@
...
+
+    # Create
+    def _create_device_id_to_bus_map(self):
+        self._device_id_to_bus_map = {}
+        self.load_config_file()
+        device_folders = os.listdir(SYSFS_PCI_DEVICE_PATH)
+        for folder in device_folders:
+            # For each folder in the sysfs tree we check if it matches the normal PCIe device folder pattern,
+            # If match we add the device id from the device file and the bus from the folder name to the map
+            pattern_for_device_folder = re.search('....:(..):..\..', folder)
```
qiluo-msft, line 45: >....:(..):..\.. [](http://example.com/codeflow?start=51&length=15)

This regex is hard to read. Could you give some examples and explanations?

platform/mellanox/mlnx-platform-api/sonic_platform/pcie.py
```diff
@@ -0,0 +1,54 @@
...
+            if pattern_for_device_folder:
+                bus = pattern_for_device_folder.group(1)
+                with open(os.path.join('/sys/bus/pci/devices', folder, 'device'), 'r') as device_file:
+                    device_id = device_file.read().replace('\n', '').replace('0x', '')
```
qiluo-msft, line 49: >device_id = device_file.read().replace('\n', '').replace('0x', '') [](http://example.com/codeflow?start=20&length=66)

Could you give a sample file content? If future change happens, we know what is broken and what is expected output.

qiluo-msft (approved):

platform/mellanox/mlnx-platform-api/sonic_platform/pcie.py
DavidZagury: An example of the contact of a file:
> 0x6fe2

I am looking for the '6fe2' part only. I will add a comment regarding this, and maybe will make it simpler by changing the first replace to strip

platform/mellanox/mlnx-platform-api/sonic_platform/pcie.py
DavidZagury: An example for a dir name is: 0000:ff:0b.1
The name of the folder is created from: {4 digit of domain - always 0}:{2 hex digit of bus}:{2 hex digit - dev}.{fn}

You can see this pattern of the folder name also here:
https://github.com/Azure/sonic-platform-common/blob/87c81de7193ee68568543d982fb8ba38f5e4ad07/sonic_platform_base/sonic_pcie/pcie_common.py#L106

I use this regex to look for the bus (in the above example 'ff').
The dots match any character, so I am checking if the folder that starts with 4 characters (domain), colon, 2 characters of bus (they are in parentheses to create the group I am looking for), a colon, 2 characters of dev, a dot, and the fn.

I tried to create the simplest regex, but I would be glad to hear if you have a different suggestion

sujinmkang (approved):

PR #8065, opened 2021-07-05
[multi-asic][cli][chassis-db] Avoid connecting to chassis db when cli commands are executed from linecards

vganesan-nokia:
#### Why I did it

Currently, for all the cli commands, we connect to all databases mentioned in the database_config.json. The database_config.json also includes the databases from chassis redis server from supervisor card. It is unnecessary to connect to databases from chassis redis server when cli commands are executed form linecard. But we need to allow connection to chassis databases when the cli commands are executed from supervisor card. This PR fixes this problem

#### How I did it

This PR requires that platform_env.conf in supervisor card includes a variable **"supervisor"** with value 1 to identify the supervisor card. The connect_to_all_dbs_for_ns() is changed to skip chassis databases form the list of collected databases if the card is not supervisor card. This PR requires sonic-utilities PR https://github.com/Azure/sonic-utilities/pull/1707 to have complete fix

#### How to verify it

In voq chassis, execute cli commands (show, config etc.) and make sure that commands are successful irrespective whether there is supervisor card or not.

N/A

multi-asic cli to skip connecting to chassis databases when commands are executed from line card.

### DIFF
--- a/src/sonic-py-common/sonic_py_common/device_info.py
+++ b/src/sonic-py-common/sonic_py_common/device_info.py
@@ -374,6 +374,22 @@ def is_multi_npu():
     return (num_npus > 1)
 
 
+def is_voq_supervisor():
+    asic_conf_file_path = get_asic_conf_file_path()
+    if asic_conf_file_path is None:
+        return False
+    with open(asic_conf_file_path) as asic_conf_file:
+        for line in asic_conf_file:
+            tokens = line.split('=')
+            if len(tokens) < 2:
+               continue
+            if tokens[0].lower() == 'voq_supervisor':
+                val = tokens[1].strip()
+                if val == '1':
+                    return True
+        return False
+ 
+
 def get_npu_id_from_name(npu_name):
     if npu_name.startswith(NPU_NAME_PREFIX):
         return npu_name[len(NPU_NAME_PREFIX):]
--- a/src/sonic-py-common/sonic_py_common/device_info.py
+++ b/src/sonic-py-common/sonic_py_common/device_info.py
@@ -29,6 +29,7 @@ HWSKU_JSON_FILE = 'hwsku.json'
 NPU_NAME_PREFIX = "asic"
 NAMESPACE_PATH_GLOB = "/run/netns/*"
 ASIC_CONF_FILENAME = "asic.conf"
+PLATFORM_ENV_CONF_FILENAME = "platform_env.conf"
 FRONTEND_ASIC_SUB_ROLE = "FrontEnd"
 BACKEND_ASIC_SUB_ROLE = "BackEnd"
 
@@ -160,6 +161,29 @@ def get_asic_conf_file_path():
     for asic_conf_file_path in asic_conf_path_candidates:
         if os.path.isfile(asic_conf_file_path):
             return asic_conf_file_path
+
+    return None
+
+
+def get_platform_env_conf_file_path():
+    """
+    Retrieves the path to the PLATFORM ENV conguration file on the device
+
+    Returns:
+        A string containing the path to the PLATFORM ENV conguration file on success,
+        None on failure
+    """
+    platform_env_conf_path_candidates = []
+
+    platform_env_conf_path_candidates.append(os.path.join(CONTAINER_PLATFORM_PATH, PLATFORM_ENV_CONF_FILENAME))
+
+    platform = get_platform()
+    if platform:
+        platform_env_conf_path_candidates.append(os.path.join(HOST_DEVICE_PATH, platform, PLATFORM_ENV_CONF_FILENAME))
+
+    for platform_env_conf_file_path in platform_env_conf_path_candidates:
+        if os.path.isfile(platform_env_conf_file_path):
+            return platform_env_conf_file_path
 
     return None
 
@@ -374,16 +398,16 @@ def is_multi_npu():
     return (num_npus > 1)
 
 
-def is_voq_supervisor():
-    asic_conf_file_path = get_asic_conf_file_path()
-    if asic_conf_file_path is None:
+def is_supervisor():
+    platform_env_conf_file_path = get_platform_env_conf_file_path()
+    if platform_env_conf_file_path is None:
         return False
-    with open(asic_conf_file_path) as asic_conf_file:
-        for line in asic_conf_file:
+    with open(platform_env_conf_file_path) as platform_env_conf_file:
+        for line in platform_env_conf_file:
             tokens = line.split('=')
             if len(tokens) < 2:
                continue
-            if tokens[0].lower() == 'voq_supervisor':
+            if tokens[0].lower() == 'supervisor':
                 val = tokens[1].strip()
                 if val == '1':
                     return True
--- a/src/sonic-py-common/sonic_py_common/multi_asic.py
+++ b/src/sonic-py-common/sonic_py_common/multi_asic.py
@@ -8,6 +8,7 @@ from swsscommon import swsscommon
 from .device_info import CONTAINER_PLATFORM_PATH
 from .device_info import HOST_DEVICE_PATH
 from .device_info import get_platform
+from .device_info import is_voq_supervisor
 
 ASIC_NAME_PREFIX = 'asic'
 NAMESPACE_PATH_GLOB = '/run/netns/*'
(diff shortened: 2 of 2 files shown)


rlhui:
@lguohan - would you like to opt out of reviewing this PR?

anshuv-mfst:
Hi @mlorrillere - could you please review, thanks.

mlorrillere:
Hi @vganesan-nokia ,

It seems that on a multi-asic pizza box this would require to configure VOQ_SUPERVISOR=1 as well.

Would it make more sense to rely on start_chassis_db=1 set in chassisdb.conf to connect to chassis db from CLI?

vganesan-nokia:
> Hi @vganesan-nokia ,
> 
> It seems that on a multi-asic pizza box this would require to configure VOQ_SUPERVISOR=1 as well.
> 
> Would it make more sense to rely on start_chassis_db=1 set in chassisdb.conf to connect to chassis db from CLI?

Hi @mlorrillere, In multi-asic pizza box systems which are not distributed VOQ systems, we always want to avoid connecting to chassis databases since they do not exist in pizza box systems.  Simply not having this VOQ_SUPERVISOR variable in the asic.conf  (as it is today in multi-asic pizza box systems) will be enough. If we do not have this variable defined, we'll skip connecting to chassis databases. Also chassisdb.conf may not exist in pizza box systems

mlorrillere:
> > Hi @vganesan-nokia ,
> > It seems that on a multi-asic pizza box this would require to configure VOQ_SUPERVISOR=1 as well.
> > Would it make more sense to rely on start_chassis_db=1 set in chassisdb.conf to connect to chassis db from CLI?
> 
> Hi @mlorrillere, In multi-asic pizza box systems which are not distributed VOQ systems, we always want to avoid connecting to chassis databases since they do not exist in pizza box systems. Simply not having this VOQ_SUPERVISOR variable in the asic.conf (as it is today in multi-asic pizza box systems) will be enough. If we do not have this variable defined, we'll skip connecting to chassis databases. Also chassisdb.conf may not exist in pizza box systems

Right, but on a multi-asic pizza box that is a VOQ system, we would need to set VOQ_SUPERVISOR=1. Am I correct?

vganesan-nokia:
> > > Hi @vganesan-nokia ,
> > > It seems that on a multi-asic pizza box this would require to configure VOQ_SUPERVISOR=1 as well.
> > > Would it make more sense to rely on start_chassis_db=1 set in chassisdb.conf to connect to chassis db from CLI?
> > 
> > 
> > Hi @mlorrillere, In multi-asic pizza box systems which are not distributed VOQ systems, we always want to avoid connecting to chassis databases since they do not exist in pizza box systems. Simply not having this VOQ_SUPERVISOR variable in the asic.conf (as it is today in multi-asic pizza box systems) will be enough. If we do not have this variable defined, we'll skip connecting to chassis databases. Also chassisdb.conf may not exist in pizza box systems
> 
> Right, but on a multi-asic pizza box that is a VOQ system, we would need to set VOQ_SUPERVISOR=1. Am I correct?

I see. In that case, I imagine, we'll be running the redis-chassis server (where the chassis databases are hosted) along with other asic instance specific redis servers in the same box. The VOQ pizza box behaves like a supervisor card in addition to being a line card or vice versa. I think setting VOQ_SUPERVISOR=1 in this case is o.k. and connecting to the chassis databases should be fine.

**Using chassisdb.conf vs asic,conf for VOQ_SUPERVISOR:** The existing CLI implementation has tested infra apis for setting path for asic.conf for file access. I just made use of it. If we have to use chassisdb.conf, we need to add additional code similar to asic.conf path setting apis for file access. I see using chassisdb.conf avoids using an additional variable in asic.conf. Otherwise start_chassis_db in chassisdb.conf and VOQ_SUPERVISOR in asic.conf give same information (that the host is a supervisor card). I do not see any significant advantage in choosing chassisdb.conf over asic.conf. If avoiding using additional variable in asic.conf is preferred over adding additional code for chassisdb.conf, I can do the modifications.

mlorrillere (dismissed):

src/sonic-py-common/sonic_py_common/device_info.py
```diff
@@ -374,6 +374,22 @@ def is_multi_npu():
     return (num_npus > 1)
 
 
+def is_voq_supervisor():
```
judyjoseph, line 377: Do we need this check to be specific to voq chassis ? Can we generalize to all chassis flavors

src/sonic-py-common/sonic_py_common/device_info.py
```diff
@@ -374,6 +374,22 @@ def is_multi_npu():
...
+    asic_conf_file_path = get_asic_conf_file_path()
```
judyjoseph, line 378: I feel it is better to use the platform_env.conf file (eg: https://github.com/Azure/sonic-buildimage/blob/master/device/arista/x86_64-arista_7800_sup/platform_env.conf) for defining voq_supervisor instead of using asic.conf file. Need to add API's for getting this file similar to get_asic_conf_file_path().

src/sonic-py-common/sonic_py_common/multi_asic.py
```diff
@@ -40,15 +41,22 @@ def connect_config_db_for_ns(namespace=DEFAULT_NAMESPACE):
       handle to the config_db for a namespace
     """
     SonicDBConfig.load_sonic_global_db_config()
-    config_db = ConfigDBConnector(namespace=namespace)
+    if namespace is not None and namespace != DEFAULT_NAMESPACE:
+        config_db = ConfigDBConnector(use_unix_socket_path=True, namespace=namespace)
+    else:
+        config_db = ConfigDBConnector(namespace=namespace)
```
arlakshm, line 47: why do we need this change ?

src/sonic-py-common/sonic_py_common/device_info.py
arlakshm: I agree, can we make it `is_supervisor()` instead ?

arlakshm (commented):
as comments

src/sonic-py-common/sonic_py_common/device_info.py
vganesan-nokia: Yes. I'll change this to is_supervisor()

src/sonic-py-common/sonic_py_common/multi_asic.py
vganesan-nokia: These changes are not required for this fix. I'll remove.

src/sonic-py-common/sonic_py_common/device_info.py
vganesan-nokia: Fixed.

src/sonic-py-common/sonic_py_common/device_info.py
vganesan-nokia: Fixed. The supervisor card type determination is done using a variable "supervisor" form platform_env.conf instead of using asic.conf

src/sonic-py-common/sonic_py_common/multi_asic.py
vganesan-nokia: Fixed.

arlakshm:
@vganesan-nokia, can you resolve the conflicts ?

vganesan-nokia:
@arlakshm, I force pushed to resolve the conflict. But the build seems to be stuck. Would you please re-start the azure pipeline? Thanks

src/sonic-py-common/sonic_py_common/multi_asic.py
```diff
@@ -56,7 +61,15 @@ def connect_to_all_dbs_for_ns(namespace=DEFAULT_NAMESPACE):
         handle to all the dbs for a namespaces
     """
     db = swsscommon.SonicV2Connector(namespace=namespace)
-    for db_id in db.get_db_list():
+    db_list = list(db.get_db_list())
+    if not is_supervisor():
+        try:
+            db_list.remove('CHASSIS_APP_DB')
+            db_list.remove('CHASSIS_STATE_DB')
```
judyjoseph, line 68: Good to have an function returning back a list of chassis db instances in supervisor. 
But it is ok to do it later as well.

judyjoseph (approved):

arlakshm (approved):

PR #7191, opened 2021-03-30
SONiC Yang model support for LLDP

sandeep-kulambi:
#### What I did 
Created SONiC Yang model for LLDP

#### How I did it
Added LLDP and LLDP_PORT tables for global and port level configurations

#### How to verify it
Compile sonic_yang_mgmt-1.0-py3-none-any.whl
tests/libyang-python-tests/test_sonic_yang.py - test_xlate_rev_xlate fails as the output generated in revXlateConfig.json has one of the field values "True" instead of "true" which causes failure in comparison.

PR #8120, opened 2021-07-07
SONiC Yang model support for LLDP

ph408077:
What I did
Created SONiC Yang model for LLDP

How I did it
Added LLDP and LLDP_PORT tables for global and port level configurations

How to verify it
Compile sonic_yang_mgmt-1.0-py3-none-any.whl

### DIFF
--- a/src/sonic-yang-models/tests/files/sample_config_db.json
+++ b/src/sonic-yang-models/tests/files/sample_config_db.json
@@ -924,6 +924,24 @@
                 "trap_ids": "ip2me",
                 "trap_group": "queue1_group1"
             }
+        },
+        "LLDP": {
+            "GLOBAL": {
+                "mode": "TRANSMIT",
+                "enabled": "true",
+                "hello_time": "12",
+                "multiplier": "5",
+                "supp_mgmt_address_tlv": "true",
+                "supp_system_capabilities_tlv": "false",
+                "system_name": "sonic",
+                "system_description": "sonic-system"
+            }
+        },
+        "LLDP_PORT": {
+            "Ethernet0": {
+                "mode": "TRANSMIT",
+                "enabled": "true"
+            }
         }
     },
 
--- a/src/sonic-yang-models/yang-models/sonic-lldp.yang
+++ b/src/sonic-yang-models/yang-models/sonic-lldp.yang
@@ -1,0 +1,154 @@
+module sonic-lldp {
+    namespace "http://github.com/Azure/sonic-lldp";
+    prefix slldp;
+    yang-version 1.1;
+
+    import sonic-port {
+        prefix prt;
+    }
+
+    import sonic-extension {
+        prefix sonic-ext;
+    }
+
+    organization
+        "SONiC";
+
+    contact
+        "SONiC";
+
+    description
+        "SONiC LLDP yang model";
+
+    revision 2020-04-16 {
+        description
+            "Initial revision.";
+    }
+
+    grouping lldp_mode_config {
+        leaf enabled {
+            type boolean;
+            description
+                "Enable/Disable LLDP";
+        }
+
+        leaf mode {
+            type enumeration {
+                enum RECEIVE;
+                enum TRANSMIT;
+            }
+
+            description
+                "RX/TX mode for LLDP frames";
+        }
+    }
+
+    container sonic-lldp {
+        container LLDP {
+            container GLOBAL {
+
+                leaf hello_time {
+                    type uint8 {
+                        range "5..254" {
+                            error-message "Invalid hello timer value.";
+                        }
+                    }
+                    default 30;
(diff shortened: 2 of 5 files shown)


anshuv-mfst:
Hi @venkatmahalingam , @praveen-li - could you please review and approve. This is dup of #7191

src/sonic-yang-models/yang-models/sonic-lldp.yang
```diff
@@ -0,0 +1,154 @@
...
+    organization
+        "SONiC";
+
+    contact
+        "SONiC";
+
+    description
+        "SONiC LLDP yang model";
+
+    revision 2020-04-16 {
```
venkatmahalingam, line 23: Please update the date.

src/sonic-yang-models/yang-models/sonic-lldp.yang
```diff
@@ -0,0 +1,154 @@
...
+                    type enumeration {
+                        enum RECEIVE;
+                        enum TRANSMIT;
+                    }
+
+                    description
+                        "RX/TX mode for LLDP frames";
+                }
+
+                //uses lldp_mode_config;
```
venkatmahalingam, line 118: We should use the grouping, Praveen mentioned in the last meeting that only few lines are required in the infra, @praveen-li please provide the infra code for testing.

src/sonic-yang-models/yang-models/sonic-lldp.yang
```diff
@@ -0,0 +1,154 @@
...
+                }
+
+                leaf system_description {
+                    type string;
+                    description
+                        "System description";
+                }
+
+                leaf supp_mgmt_address_tlv {
+                    type boolean;
```
venkatmahalingam, line 89: no default value for this field?

src/sonic-yang-models/yang-models/sonic-lldp.yang
```diff
@@ -0,0 +1,154 @@
...
+                    default true;
+                    description
+                        "Enable/Disable LLDP";
+                }
+
+                leaf mode {
+                    type enumeration {
+                        enum RECEIVE;
+                        enum TRANSMIT;
+                    }
```
venkatmahalingam, line 112: no default value?

src/sonic-yang-models/yang-models/sonic-lldp.yang
ph408077: Done

src/sonic-yang-models/yang-models/sonic-lldp.yang
ph408077: Added default value

src/sonic-yang-models/yang-models/sonic-lldp.yang
ph408077: This is optional field, so user can set the mode to either tx-only or rx-only and 'both tx and rx' is controlled by enabled field

ph408077 (commented):
/azpw run

src/sonic-yang-models/tests/files/sample_config_db.json
```diff
@@ -924,6 +924,24 @@
                 "trap_ids": "ip2me",
                 "trap_group": "queue1_group1"
             }
+        },
+        "LLDP": {
```
praveen-li, line 928: BTW what is the source of this config?, this config is not listed on https://github.com/Azure/SONiC/wiki/Configuration.

src/sonic-yang-models/yang-models/sonic-lldp.yang
```diff
@@ -0,0 +1,155 @@
...
+
+    description
+        "SONiC LLDP yang model";
+
+    revision 2021-07-08 {
+        description
+            "Initial revision.";
+    }
+
+    grouping lldp_mode_config {
```
praveen-li, line 28: Any reason by grouping is used?

src/sonic-yang-models/tests/files/sample_config_db.json
ph408077: LLDP parameters were directly configured via lldpcli. To support north bound interfaces,these configs were added by Broadcom to configure LLDP global and per port parameters and store the same in Config DB.

src/sonic-yang-models/yang-models/sonic-lldp.yang
ph408077: Enabled and mode leaf is used in Global and per Port config. So it was grouped and used in these nodes. However usage of grouping is leading to compilation error, its commented for now. But ideally since its used in two places, grouping is defined.

ph408077:
Hi @venkatmahalingam , @praveen-li - Do you have there any further comments or questions

PR #8144, opened 2021-07-09
[gearbox] Add gbsyncd container for Credo gearbox chips

byu343:
#### Why I did it
This change is to add a gbsyncd container to accommodate the syncd process and the SAI libraries for the Credo gearbox chips.

#### How I did it
This container works similar to the existing Broadcom syncd container. Its main difference is that the SAI-related dynamic libraries are replaced by the ones for Credo gearbox chips, and the container only reacts to SAI events for the gearbox chips. The SAI libraries will be provided by the package libsai-credo_1.0_amd64.deb.

For the image build, the added container will be built and included in the Broadcom platform image, after $(LIBSAI_CREDO)_URL = is replaced to the correct value. For now, as $(LIBSAI_CREDO)_URL is empty, the container build is skipped in the image build.   

After the container is included in the image, in the runtime, the container will begin with checking the existence of /usr/share/sonic/hwsku/gearbox_config.json; if that file is not provided, the container will exit by itself. Therefore, for platforms unrelated to the Credo chips, as long as they are not providing the file, they will not be affected by this change.  

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [x] 202106


platform/broadcom/rules.dep
```diff
@@ -17,6 +17,7 @@ include $(PLATFORM_PATH)/platform-modules-brcm-xlr-gts.dep
 include $(PLATFORM_PATH)/docker-syncd-brcm.dep
 include $(PLATFORM_PATH)/docker-syncd-brcm-rpc.dep
 include $(PLATFORM_PATH)/docker-saiserver-brcm.dep
+include $(PLATFORM_PATH)/docker-gbsyncd-credo.dep
```
lguohan, line 20: credo is not specific to broadcom, i feel this docker should be a generic docker except on vs platform or we can say which platform will include this docker initially it would be on broadcom only, but this docker can be available on other platforms as well.

platform/components/docker-gbsyncd-credo/supervisord.conf.j2
```diff
@@ -0,0 +1,42 @@
...
+autostart=false
+autorestart=false
+startsecs=0
+stdout_logfile=syslog
+stderr_logfile=syslog
+dependent_startup=true
+dependent_startup_wait_for=rsyslogd:running
+
+[program:syncd]
+command=/usr/bin/syncd -s -p /etc/sai.d/pai.profile -x /usr/share/sonic/hwsku/context_config.json -g 1
```
lguohan, line 35: let's use /etc/sai.d/psai.profile as a name convention. instead of calling pai, let's call it psai

platform/components/docker-gbsyncd-credo/critical_processes.j2
```diff
@@ -0,0 +1 @@
+program:syncd
```
lguohan, line 1: I noticed that docker-gbsyncd-vs is using gbsyncdmgrd script from sonic-sairedis repo which checks if gearbox_config.json exists or not, if not exist then it will not start the syncd process.

lguohan: one question here is that how do we know the gearbox is available on this platform? for example, many broadcom platformd does not have gearbox, and therefore, there is no need to start gbsyncd docker. gbsyncdmgrd is one way to handle such issue, but a better solution might be not start the gbsyncd docker at all if gearbox is not available on this platform.

byu343: Currently, it is handled by platform/components/docker-gbsyncd-credo/docker-init.sh, for which the container exits by itself if no gearbox_config.json.

I'll check whether it can be changed to gbsyncdmgrd or not starting the container at all.

In platform/components/docker-gbsyncd-credo/docker-init.sh,
GB_CONFIG=/usr/share/sonic/hwsku/gearbox_config.json

if [ ! -f $GB_CONFIG ]; then
   exit 0
fi

byu343: To address the problem of not starting gbsyncd containers unrelated to a platform, the change "Separate gbsyncd-credo service" is committed. 

After the change, it is assumed that each gbsyncd variant has its own service and docker image. For example, for gbsyncd-credo, there are gbsyncd-credo.service and gbsyncd-credo docker; for gbsyncd-vs, there are gbsyncd-vs.service and gbsyncd-vs docker. Potentially, a new service file and docker image should be created if there is a new type of gbsyncd added in the future.

In the booting, all kinds of gbsyncd services installed will all be started, and each of them use ExecCondition=/usr/bin/gbsyncd-platform.sh to determine whether to start the corresponding docker or just skip it, by checking the file gbsyncd.ini in the device platform folder. If the result is skipping, it will leave some messages in the log, e.g.,

Jul 22 21:49:06.264558 gd270 INFO systemd[1]: Starting gbsyncd-credo service...
Jul 22 21:49:07.117152 gd270 NOTICE systemd[1]: gbsyncd-credo.service: Control process exited, code=exited, status=1/FAILURE
Jul 22 21:49:07.117328 gd270 INFO systemd[1]: gbsyncd-credo.service: Skipped due to 'exec-condition'.
Jul 22 21:49:07.118970 gd270 INFO systemd[1]: Condition check resulted in gbsyncd-credo service being skipped.

platform/components/docker-gbsyncd-credo/Dockerfile.j2
```diff
@@ -0,0 +1,40 @@
...
+## Make apt-get non-interactive
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN apt-get update
+
+RUN apt-get install -f -y iproute2 libcap2-bin libprotobuf-dev
+
+COPY \
+{% for deb in docker_gbsyncd_credo_debs.split(' ') -%}
+{% if 'libsaibcm' not in deb -%}
```
lguohan, line 15: not sure why we have saibcm here.

byu343:
The build failure is because the base version of sonic-buildimage we tested against is different. I am looking into it and will fix it ASAP.

lguohan (approved):

PR #8146, opened 2021-07-09
[arista] Add gearbox configs for Arista 7280cr3mk

byu343:
#### Why I did it
This change adds the gearbox config files to Arista 7280cr3mk

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106


device/arista/x86_64-arista_7280cr3mk_32p4/Arista-7280CR3-C40/context_config.json
```diff
@@ -0,0 +1,157 @@
...
+                {
+                    "index" : 0,
+                    "hwinfo" : ""
+                }
+            ]
+        },
+        {
+            "guid" : 1,
+            "name" : "phy1",
+            "dbAsic" : "GB_ASIC_DB",
```
kcudnik, line 23: duplicated databases, please read here: https://github.com/Azure/sonic-sairedis/wiki/Context-configuration

kcudnik: i also think what you intended is to have 2 syncd , first with ASIC_DB and second with GB_ASIC_DB, first will have 1 switch, and 2nd will have 8 phys

device/arista/x86_64-arista_7280cr3mk_32p4/Arista-7280CR3-C40/pai.profile
```diff
@@ -0,0 +1 @@
+SAI_KEY_INIT_CONFIG_FILE=/usr/share/sonic/hwsku/gearbox_100G_PAM4.xml
```
lguohan, line 1: can you rename this file to psai.profile?

device/arista/x86_64-arista_7280cr3mk_32p4/Arista-7280CR3-C40/gearbox_config.json
```diff
@@ -0,0 +1,326 @@
...
+      "sai_init_config_file": "",
+      "phy_access": "mdio",
+      "bus_id": 0,
+      "hwinfo": "mdio4_0_0/0"
+    },
+    {
+      "phy_id": 6,
+      "name": "phy6",
+      "address": "6",
+      "lib_name": "/usr/lib/libsai_credo.so",
```
lguohan, line 67: is this really used? i thought the name is libsai.so.1?

device/arista/x86_64-arista_7280cr3mk_32p4/Arista-7280CR3-C40/phy1_config_1.json
```diff
@@ -0,0 +1,101 @@
+{
+  "lanes": [
```
lguohan, line 2: what does this lanes use for?

device/arista/x86_64-arista_7280cr3mk_32p4/Arista-7280CR3-C40/gearbox_100G_PAM4.xml
```diff
@@ -0,0 +1,37 @@
...
+    <mode>gearbox</mode>
+    <topology>2</topology>
+    <tx-taps>
+        <PAM4>2,-8,17,0,0</PAM4>
+        <NRZ>0,-8,17,0,0</NRZ>
+    </tx-taps>
+    <tx-taps-scale>0,0,1,0,0</tx-taps-scale>
+
+    <lane id="0" system-side="true" />
+    <lane id="1" system-side="true" />
```
lguohan, line 14: why do we need this file? we have phy_config_1.json, then we should not need this xml? it seems redundant information here.

device/arista/x86_64-arista_7280cr3mk_32p4/Arista-7280CR3-C40/context_config.json
shyam77git: @kcudnik : in this context_config.json file, trying to see/understand what's the purpose of having both ASIC_DB and GB_ASIC_DB blocks. Is it like this:
If a ASIC complex on the board has both ASIC and PHY on the same chip/SOC, then specify both of them.
If a PHY chip/device is Ext to ASIC (i.e. independent chip on the board), then just specifying GB_ASIC_DB block is sufficient?

kcudnik: from my point of view, how hardware is organized is not relevant, depends how you want to present it in database and configuration file. you **CANT** have multiple syncd pointing to the same databases ex. ASIC_DB, each context entry with separate GUID is a separate syncd instance. please read context config wiki page

byu343: @shyam77git , for this platform, guid 0 is for ASIC, guid 1 and thereafter are for external phys now. Later, we may have a further change to cover all external phys by guid 1

shyam77git: @byu343, per this platform (this PR):
>> guid 1 and thereafter are for external phys now. 
Can you clarify now vs later? 
From what I gather from context config wiki page & from Kamil's comment here, multiple guid pointing to same database (e.g GB_ASIC_DB ) is not supported so how come this model works (i.e. different guid for each PHY)?

byu343: @kcudnik I tried to set all phys to share the same context guid 1, and it has two new issues:

In https://github.com/Azure/sonic-swss/blob/master/orchagent/saihelper.cpp#L421, SAI_REDIS_SWITCH_ATTR_CONTEXT is set to phy_id, which is different for each phy. How about add a new property called context to gearbox_config.json, set it to 1 for all phys and then change line 421 to attr.value.u64 = phy->context.

In https://github.com/opencomputeproject/SAI/blob/master/inc/saiport.h#L748, SAI_PORT_ATTR_HW_LANE_LIST is defined as key. For ports with the same in-chip id but from different phy chips, they have the same SAI_PORT_ATTR_HW_LANE_LIST. After putting those ports in the same context, we encountered an error related to key conflict, which is from https://github.com/Azure/sonic-sairedis/blob/master/meta/Meta.cpp#L5060. I tried to keep SAI_PORT_ATTR_HW_LANE_LIST from being key in saiport.h#L748,  and it seems good. As that involves the change to SAI headers, is there any other better solution?

kcudnik: this model works if you each syncd will have its own database

kcudnik: we will not add any special properties like you proposed, context_config.json describes each syncd instance and that's it. its common for all vendors and architectures

byu343: @shyam77git, between now vs later, maybe you can just skip the "now" model, as it does not follow the design in the wiki page and will very likely be deprecated. If you are still interested in why multiple guid pointing to same database could work, the tricky part is we did only create the syncd process for guid 1. It was done in that way to bypass some other issues.

byu343: @kcudnik ,
The new property proposed is for gearbox_config.json, not for context_config.json.  Let me just create the new PR for it, to let it be clear in discussion.

device/arista/x86_64-arista_7280cr3mk_32p4/Arista-7280CR3-C40/gearbox_config.json
byu343: No, not really used

device/arista/x86_64-arista_7280cr3mk_32p4/Arista-7280CR3-C40/gearbox_100G_PAM4.xml
byu343: This file is used by credo driver directly.  I'll check whether some items in it can be removed. Still, there are some have to be kept for the current driver design, like <name>CSDK-B52</name> for determining the chip model to initialize.

device/arista/x86_64-arista_7280cr3mk_32p4/Arista-7280CR3-C40/pai.profile
byu343: Yes

device/arista/x86_64-arista_7280cr3mk_32p4/Arista-7280CR3-C40/gearbox_config.json
lguohan: if not used, can we remove it?

device/arista/x86_64-arista_7280cr3mk_32p4/Arista-7280CR3-C40/gearbox_config.json
```diff
@@ -0,0 +1,334 @@
...
+      "bus_id": 0,
+      "context_id": 1,
+      "hwinfo": "mdio1_0_0/0"
+    },
+    {
+      "phy_id": 3,
+      "name": "phy3",
+      "address": "3",
+      "lib_name": "",
+      "firmware_path": "/usr/share/phy_fw.bin",
```
lguohan, line 34: is this used? if not can we remove?

byu343: Yes, firmware_path is used.

device/arista/x86_64-arista_7280cr3mk_32p4/Arista-7280CR3-C40/gearbox_config.json
byu343: It has been removed.

device/arista/x86_64-arista_7280cr3mk_32p4/Arista-7280CR3-C40/gearbox_100G_PAM4.xml
byu343: It turns out that for now the items in this file cannot be removed, as they are still used by the driver.

device/arista/x86_64-arista_7280cr3mk_32p4/Arista-7280CR3-C40/context_config.json
byu343: @kcudnik The code has been updated. 
For context_config.json, all the phys are defined in the same context. 
For gearbox_config.json, a property of context_id is added to each phy. It is to be used by gearsyncd in sonic-swss, as gearsyncd needs to specify the context for each phy when create_switch(): https://github.com/Azure/sonic-swss/pull/1826/files#diff-be825ae02c0d05e1a802a265c12721e5747035acab211f3bc7e6a219cf500e03L421.  phy_id in gearbox_config.json was used to specify the context which seems not always true. To support context_id in gearsyncd, a change to swss is created: https://github.com/Azure/sonic-swss/pull/1826.

shyam77git: >> phy_id in gearbox_config.json was used to specify the context which seems not always true.

@byu343 
To elaborate more on this, share the understanding and ensure we all are on same page:
[1] Based on existing codebase in swss/saihelper, looks like each phy_id (in gearbox_config.json) corresponds to unique/separate guid (of context_config.json). whereas, with the change in context_config.json all PHYs (Switches) are under same guid (global context).
[2] At swss/sai code/workflow level, this change was done because phy_id was/is being passed as SAI_REDIS_SWITCH_ATTR_CONTEXT. 
[3] phy_id is unique for each PHY. However, context_config.json puts all PHYs (switches) under one (common) guid. 
[4] globalContext gets populated with phy_id for every create_switch() call which make globalContext different from every create_switch() call , whereas the expectation is to have a common (same) id for all PHYs/Switches under the guid 1 of the revised context_config.json. For this purpose, context_id field was introduced for all PHYs under the same roof (i.e. guid 1 GB_ASIC_DB of context_config.json) and populate with same value for all these PHYs.

@kcudnik - can you please share you thoughts/comments on this and related PR (1826)

<Excerpt from current codebase>
saihelper.cpp 
initSaiPhyApi() 
{
...
    /* Must be last Attribute */
    attr.id = SAI_REDIS_SWITCH_ATTR_CONTEXT;
    attr.value.u64 = phy->phy_id;
    attrs.push_back(attr);

    status = sai_switch_api->create_switch(&phyOid, (uint32_t)attrs.size(), attrs.data());
....
}

Sai.cpp 
sai_status_t Sai::create(
{
...
      if (attr_list[attr_count - 1].id == SAI_REDIS_SWITCH_ATTR_CONTEXT)
        {
            globalContext = attr_list[--attr_count].value.u32;
        }
...
}

Thanks,
Shyam

device/arista/x86_64-arista_7280cr3mk_32p4/Arista-7280CR3-C40/context_config.json
```diff
@@ -0,0 +1,66 @@
...
+                },
+                {
+                    "index" : 6,
+                    "hwinfo" : "mdio6_0_0/0"
+                },
+                {
+                    "index" : 7,
+                    "hwinfo" : "mdio7_0_0/0"
+                }
+            ]
```
shyam77git, line 63: Are all 8 PHYs (0 - 7) belonging to same NPU/ASIC? 

Is/Are there PHY(s) which needs to be on another NPU? 
In that case, would it be following SONiC multi-NPU/multi-ASIC architecture ? (i.e. have these gearbox_config, phy<x>_config_1 , context_config json files under /usr/share/sonic/device/swss@<NPU> syncd@<NPU>  and gbsyncd@<NPU> )?

byu343: For this specific platform, the setting is 8 external phy chips connected 1 ASIC.
So far for the platforms we planned to support, we haven't considered about the multi-ASIC case.

device/arista/x86_64-arista_7280cr3mk_32p4/Arista-7280CR3-C40/gearbox_100G_PAM4.xml
lguohan: we will keep it for now, can you check with credo to remove the support for this file?

device/arista/x86_64-arista_7280cr3mk_32p4/Arista-7280CR3-C40/phy2_config_1.json
```diff
@@ -0,0 +1,101 @@
+{
+  "lanes": [
```
lguohan, line 2: it is unclear how the lanes are mapped to the ports, where do we get those information?

device/arista/x86_64-arista_7280cr3mk_32p4/Arista-7280CR3-C40/gearbox_config.json
```diff
@@ -0,0 +1,334 @@
...
+    {
+      "phy_id": 1,
+      "name": "phy1",
+      "address": "1",
+      "lib_name": "",
+      "firmware_path": "/usr/share/phy_fw.bin",
+      "config_file": "/usr/share/sonic/hwsku/phy1_config_1.json",
+      "sai_init_config_file": "",
+      "phy_access": "mdio",
+      "bus_id": 0,
```
lguohan, line 12: is this really used? if not, we should have remove it.

device/arista/x86_64-arista_7280cr3mk_32p4/Arista-7280CR3-C40/gearbox_config.json
```diff
@@ -0,0 +1,334 @@
...
+      "context_id": 1,
+      "hwinfo": "mdio2_0_0/0"
+    },
+    {
+      "phy_id": 4,
+      "name": "phy4",
+      "address": "4",
+      "lib_name": "",
+      "firmware_path": "/usr/share/phy_fw.bin",
+      "config_file": "/usr/share/sonic/hwsku/phy4_config_1.json",
```
lguohan, line 48: >_1 [](http://example.com/codeflow?start=56&length=2)

is there a reason to add _1 here? seems not needed.

device/arista/x86_64-arista_7280cr3mk_32p4/Arista-7280CR3-C40/gearbox_config.json
```diff
@@ -0,0 +1,334 @@
...
+      "address": "1",
+      "lib_name": "",
+      "firmware_path": "/usr/share/phy_fw.bin",
+      "config_file": "/usr/share/sonic/hwsku/phy1_config_1.json",
+      "sai_init_config_file": "",
+      "phy_access": "mdio",
+      "bus_id": 0,
+      "context_id": 1,
+      "hwinfo": "mdio0_0_0/0"
+    },
```
shyam77git, line 15: @byu343  : Can you please share/update as to which azure branch you referring to to put add and validate these fields? Is it azure/master? which git commit/ reference point?

device/arista/x86_64-arista_7280cr3mk_32p4/Arista-7280CR3-C40/context_config.json
kcudnik: > whereas the expectation is to have a common (same) id for all PHYs/Switches under the guid 1 of the revised context_config.json.

what do you mean to its expected to have common id ? which id are you talking about here?

device/arista/x86_64-arista_7280cr3mk_32p4/Arista-7280CR3-C40/context_config.json
```diff
@@ -0,0 +1,66 @@
...
+        },
+        {
+            "guid" : 1,
+            "name" : "phys",
+            "dbAsic" : "GB_ASIC_DB",
+            "dbCounters" : "GB_COUNTERS_DB",
+            "dbFlex": "GB_FLEX_COUNTER_DB",
+            "dbState" : "STATE_DB",
+            "zmq_enable" : false,
+            "zmq_endpoint": "tcp://127.0.0.1:5565",
```
kcudnik, line 28: you still have the same zmq addresses for both contexts guid 0 and 1

device/arista/x86_64-arista_7280cr3mk_32p4/Arista-7280CR3-C40/phy2_config_1.json
byu343: The section of lanes is not used, and the data put here is to pass the parser checking.  The related info is obtained from gearbox_100G_PAM4.xml, the file only applied to credo driver.

device/arista/x86_64-arista_7280cr3mk_32p4/Arista-7280CR3-C40/gearbox_config.json
byu343: We followed "src/sonic-swss/gearsyncd/tests/configs/positive/phy1_config_1.json".  Other than that, it is not needed.

device/arista/x86_64-arista_7280cr3mk_32p4/Arista-7280CR3-C40/gearbox_config.json
byu343: Bus_id is not really used. The value put here is to pass the parser checking: https://github.com/Azure/sonic-swss/blob/ec104c1ba079901a4af1fb72981d34e0069ceeac/gearsyncd/gearboxparser.cpp#L140

lguohan, line 12: this is not used in the driver, but parser requires that.

device/arista/x86_64-arista_7280cr3mk_32p4/Arista-7280CR3-C40/gearbox_100G_PAM4.xml
byu343: sure

device/arista/x86_64-arista_7280cr3mk_32p4/Arista-7280CR3-C40/gearbox_config.json
byu343: The base sonic-buildimage image version we are using for testing is e43967645.  We plan to bring the support to master and 202106, so if any issue is from on them later, we will submit some fix.

device/arista/x86_64-arista_7280cr3mk_32p4/Arista-7280CR3-C40/context_config.json
byu343: This has been fixed.

jimmyzhai (approved):

PR #8157, opened 2021-07-12
[hostcfgd] [202012] Enhance hostcfgd to check feature state and run less system calls

shlomibitton:

#### Why I did it
Currently hostcfgd is implemented in a way each feature which is enabled/disabled triggering execution of systemctl enable/unmask commands which eventually trigger 'systemctl daemon-reload' command.
Each call like this cost 0.6s and overall add a overhead of ~12 seconds of CPU time.
This change will verify the desired state of a feature and the current state of this feature on systemd and trigger a system call only when must.

#### How I did it
Check each feature status on systemd before executing a system call to enable and reload the systemctl daemon.

#### How to verify it
Build an image with this change and observe less system calls are executed.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012
- [ ] 202106

### DIFF
--- a/src/sonic-host-services/scripts/hostcfgd
+++ b/src/sonic-host-services/scripts/hostcfgd
@@ -457,6 +457,13 @@ class HostConfigDaemon:
     def enable_feature(self, feature_names, feature_suffixes):
         start_cmds = []
         for feature_name in feature_names:
+            # Check if it is already enabled, if yes skip the system call
+            cmd = "sudo systemctl status {}.{} | grep Loaded".format(feature_name, feature_suffixes[-1])
+            proc = subprocess.Popen(cmd, shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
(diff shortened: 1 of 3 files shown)


shlomibitton:
@yxieca can you please check the warm_boot test failure? It does not seems to be related to this change.
I see in the test failure: "WARNING: No route found for IPv6 destination"

vaibhavhd:
> @yxieca can you please check the warm_boot test failure? It does not seems to be related to this change.
> I see in the test failure: "WARNING: No route found for IPv6 destination"

The real failure here is the exceeded downtime:
`downtime"FAILED:dut:Longest downtime period must be less then 200 seconds. It was 0:03:24.487656"`

The ` "WARNING: No route found for IPv6 destination"` is benign and appears on every run of this test.

I have triggered a re-run and this should likely pass.

lguohan (approved):

PR #8187, opened 2021-07-15
[sonic-yang] Backlinks are none when there is 0 backlinks

ghooo:
#### Why I did it
Backlinks are none when there is 0 backlinks. without this change `backlinks.number()` throws an exception when backlinks==None.

#### How I did it
Allow backlinks to be None, so we get an empty list as a return from find_data_dependencies

#### How to verify it
```python
    def test_find_leaf_dependencies_no_schema_dependencies(self):
        sy = sonic_yang.SonicYang(YANG_DIR)
        sy.loadYangModel()
        sy.loadData(json.loads("""
{
    "PORT": {
        "Ethernet0": {
            "alias":"eth0",
            "lanes": "65",
            "speed": "100000"
        }
    }
}
            """))

        # the PORT alias has no leafrefs pointing to it
        xpath = "/sonic-port:sonic-port/PORT/PORT_LIST[name='Ethernet0']/alias"
        deps = sy.find_data_dependencies(xpath)

        self.assertEqual(0, len(deps))
```
(added UT for the same idea)

### DIFF
--- a/src/sonic-yang-mgmt/sonic_yang.py
+++ b/src/sonic-yang-mgmt/sonic_yang.py
@@ -526,7 +526,7 @@ class SonicYang(SonicYangExtMixin):
 
             schema_node = ly.Schema_Node_Leaf(data_node.schema())
             backlinks = schema_node.backlinks()
-            if backlinks.number() > 0:
+            if backlinks is not None and backlinks.number() > 0:
                 for link in backlinks.schema():
                      node_set = node.find_path(link.path())
                      for data_set in node_set.data():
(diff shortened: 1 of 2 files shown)


qiluo-msft:
@pchaudhary Could you help review

qiluo-msft:
@praveen-li Could you help review

qiluo-msft (approved):

qiluo-msft (approved):

praveen-li (approved):
LGTM, fix the alignment if wrong, before merge. Thx.

PR #8191, opened 2021-07-15
Upgrade SONiC to Debian Bullseye

saiarcot895:
#### Why I did it

In preparation for Debian Bullseye, upgrade SONiC's base system to be based on Bullseye, which was released in August 2021. Major changes from Buster are:

1. Kernel is now based on 5.10.x (currently, official Debian Bullseye is publishing the 5.10.70 kernel)
2. Most Python 2 packages (as well as `pip2.7`) have been removed from Bullseye. The Python 2 interpreter is still available.

#### How I did it

The kernel has been upgraded to 5.10.46, and the base system is now based on Bullseye. Containers are still based on Buster.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106


lguohan:
can you resolve the conflict so that the build can proceed?

saiarcot895:
Conflicts have been resolved. The expectation is that the kvm image should build (none of the containers will get built), broadcom will fail to build (vendor platform modules need changes), mellanox likely will fail to build.

lguohan:
can you check the kvm build, seems failed.

lguohan:
kvm build still failing...

lguohan:
can you resolve merge conflict, i think the last commit on swss-common is no longer needed.

saiarcot895:
The `libswsscommon`, `python-swsscommon`, and `python3-swsscommon` packages are all installed on the base image itself, not just in the containers, so I do think they need the Bullseye changes.

Currently, though, the CI run is reporting a test failure in the `target/python-wheels/sonic_config_engine-1.0-py3-none-any.whl` target:

```
test_render_template (tests.test_minigraph_case.TestCfgGenCaseInsensitive) ... Traceback (most recent call last):
  File "/sonic/src/sonic-config-engine/tests/../sonic-cfggen", line 443, in <module>
    main()
  File "/sonic/src/sonic-config-engine/tests/../sonic-cfggen", line 405, in main
    env = _get_jinja2_env(paths)
  File "/sonic/src/sonic-config-engine/tests/../sonic-cfggen", line 237, in _get_jinja2_env
    redis_bcc = RedisBytecodeCache(SonicV2Connector(host='127.0.0.1'))
  File "/usr/lib/python3/dist-packages/swsscommon/swsscommon.py", line 1697, in __init__
    for db_name in self.get_db_list():
  File "/usr/lib/python3/dist-packages/swsscommon/swsscommon.py", line 1633, in get_db_list
    return _swsscommon.SonicV2Connector_Native_get_db_list(self)
RuntimeError: Sonic database config file doesn't exist at /var/run/redis/sonic-db/database_config.json
```

lguohan:
swss-common submodule has already been updated in master branch, you do not need to update it again.

rules/swsssdk-py2.mk
```diff
@@ -1,7 +1,10 @@
-# swsssdk python2 wheel
+# Python 2 packages shouldn't be built for Bullseye
+ifneq ($(BLDENV), bullseye)
```
qiluo-msft, line 2: I believe this will be applicable to Bullseye and all future versions. It is not convenient
1. to include every future version names
2. to repeat in every py2 rule file

Suggest centralize the check and share a variable like ENABLE_PY2_WHEEL

files/build_templates/sonic_debian_extension.j2
```diff
@@ -56,9 +56,7 @@ FILESYSTEM_ROOT_ETC_SONIC="$FILESYSTEM_ROOT_ETC/sonic"
 GENERATED_SERVICE_FILE="$FILESYSTEM_ROOT/etc/sonic/generated_services.conf"
 
 clean_sys() {
-    sudo chroot $FILESYSTEM_ROOT umount /sys/fs/cgroup/*            \
-                                        /sys/fs/cgroup              \
-                                        /sys || true
+    sudo chroot $FILESYSTEM_ROOT umount -R /sys || true
```
qiluo-msft, line 59: >sudo chroot $FILESYSTEM_ROOT umount -R /sys || true [](http://example.com/codeflow?start=4&length=51)

Why you need to change this part? Is it an optimization?

saiarcot895: `/sys/fs/cgroup/*` and `/sys/fs/cgroup` didn't appear to be mount points, and so `umount` would error out. However, just in case they  did get mounted through some code path, I added the `-R` flag to `umount`, so that they would still get umounted if they were indeed mounted.

rules/swsssdk-py2.mk
saiarcot895: My thinking is that once the containers are upgraded to Bullseye, the code to build the Python 2 packages would be entirely removed. I can still define a central variable if you want.

files/build_templates/sonic_debian_extension.j2
saiarcot895: I'm going to revert this part of the change, looks like `/sys/fs/cgroup` does get mounted later in the script.

lguohan:
looks like we still have the sonic-cfggen unit test failure.

rules/swsssdk-py2.mk
qiluo-msft: We still have jessie in the codebase. So your think is a long roadmap. It does not harm to have a central variable

saiarcot895:
Build failures are caused by the `sonic-telemetry` build/test overwriting `/var/run/redis/sonic-db/database_config.json`, and then removing  the `/var/run/redis/sonic-db/` directory. Because the same slave container is used for all of the builds, these changes persist across builds.

The commits responsible for this change are Azure/sonic-telemetry@4bb02f5acbba49f2cd7be7f34583968a5ce30f8a (overwriting `/var/run/redis/sonic-db/database_config.json`) and Azure/sonic-telemetry@6df988cce714ab7be3a914069ad24d30d153da13 (removing the `/var/run/redis/sonic-db/` directory).

```
 $ grep database_config.json target/debs/buster/*.log
target/debs/buster/libswsscommon_1.0.0_amd64.deb-install.log:Configuration file '/var/run/redis/sonic-db/database_config.json', does not exist on system.
target/debs/buster/libswsscommon_1.0.0_amd64.deb.log: /usr/bin/install -c -m 644 database_config.json '/sonic/src/sonic-swss-common/debian/tmp/var/run/redis/sonic-db'
target/debs/buster/sonic-telemetry_0.1_amd64.deb.log:sudo cp ./testdata/database_config.json /var/run/redis/sonic-db/
```

saiarcot895:
Waiting for #8282 to be merged.

platform/barefoot/docker-syncd-bfn-rpc.mk
```diff
@@ -19,7 +19,7 @@ endif
 $(DOCKER_SYNCD_BFN_RPC)_CONTAINER_NAME = syncd
 $(DOCKER_SYNCD_BFN_RPC)_VERSION = 1.0.0-rpc
 $(DOCKER_SYNCD_BFN_RPC)_PACKAGE_NAME = syncd
-$(DOCKER_SYNCD_BFN_RPC)_RUN_OPT += --net=host --privileged -t
```
qiluo-msft, line 22: Could you share more insight why "--net=host is specified twice" ? Maybe we should fix the other place.

saiarcot895: Commit c948379 made some changes in [files/build_templates/docker_image_ctl.j2](https://github.com/Azure/sonic-buildimage/commit/c9483796dc7dc6ad68b19e3970d851ed43ffbad3#diff-130e5ab75471398db06b73a4c6e4b56517f42e3c1f17870b5309620fc0e2fafb). Specifically, it added some code here to explicitly specify which network namespace/topology the container should run in (with the `--net=` flag): either the host namespace (so that it has all of the host network interfaces accessible within the container), connected to the Docker bridge with veth interfaces (this is the default), or the namespace of another container.

If the container isn't a per-ASIC container (meaning there's only one for the system), then it'll put it in the host namespace. Otherwise, if it's in a multi-ASIC environment and the database container is being started, then that container will be connected via the default Docker bridge. Otherwise, it'll use the namespace of the database container for that ASIC.

This change also removed all of the explicit `--net=host` flags specified in the run options for other containers (see the files changed in `rules/` directory in that commit), but didn't handle the ones that were defined in the `platform/` directory. So I'm replicating the change made in that commit for these other containers.

lguohan:
i can see the src/sonic-swss-common as a conflict and it blocks the build, can you consider merge your changes in src/sonic-swss-common into the repo so that we won't have such conflict in the future.

src/ifupdown2/Makefile
```diff
@@ -13,6 +13,14 @@ $(addprefix $(DEST)/, $(MAIN_TARGET)): $(DEST)/% :
 	tar -z -f $(IFUPDOWN2_VERSION).tar.gz -x
 	pushd ./ifupdown2-$(IFUPDOWN2_VERSION)
 
+	git init
+	git add -f *
+	git commit -m "unmodified lldp source"
```
lguohan, line 18: copy-paste error?

alexrallen:
Hi @saiarcot895 if possible could you please avoid force pushing to this branch? 

I know its generally not a problem on feature branches but I am currently working on aligning Mellanox code to this branch in parallel and it would be extremely helpful for me to have a consistent history so that I can better merge in changes from your end as they come in. 

Please let me know if there are any problems or anything I can do to help.

saiarcot895:
Hi @alexrallen ,

I can reduce the number of force pushes that I do, and stick with merge commits for now. Note that I will be doing one force-push likely next week.

paulmenzel:
Thank you for this work. Could you please update the merge/pull request description with the biggest changes? Something like:

1.  Linux kernel upgrade from 4.19 to 5.10
2.  Python upgrade …
3.  … (other SONiC affecting components)

Did you perform boot time and bandwidth/performance tests? Any difference? What about memory usage and so on?

saiarcot895:
> Thank you for this work. Could you please update the merge/pull request description with the biggest changes? Something like:
> 
> 1. Linux kernel upgrade from 4.19 to 5.10
> 2. Python upgrade …
> 3. … (other SONiC affecting components)
> 
> Did you perform boot time and bandwidth/performance tests? Any difference? What about memory usage and so on?

Currently, functionality and bootup is being worked on; performance tests have not been done.

sonic-slave-bullseye/Dockerfile.j2
```diff
@@ -3,7 +3,9 @@ FROM multiarch/qemu-user-static:x86_64-arm-5.0.0-2 as qemu
 FROM multiarch/debian-debootstrap:armhf-bullseye
 COPY --from=qemu /usr/bin/qemu-arm-static /usr/bin
 {%- elif CONFIGURED_ARCH == "arm64" and MULTIARCH_QEMU_ENVIRON == "y" %}
+FROM multiarch/qemu-user-static:x86_64-aarch64-5.2.0-2 as qemu
```
lguohan, line 6: i do not know why this is needed?

sonic-slave-bullseye/Dockerfile.j2
```diff
@@ -349,19 +353,24 @@ RUN apt-get update && apt-get install -y \
 
 RUN apt-get -y build-dep openssh
 
+# Build fix for ARM64 and ARMHF /etc/debian_version
+{%- if CONFIGURED_ARCH == "armhf" or CONFIGURED_ARCH == "arm64" %}
+RUN apt upgrade -y base-files
+{%- endif %}
```
lguohan, line 359: is this really an issue?

saiarcot895: Yes. This Bullseye arm64 container was last published well before the release of Bullseye. Because of that, `/etc/debian_version` contains `bullseye/sid` instead of 10.11, and after `sources.list` is changed to use our repo list, `lsb_release -cs` fails because that file doesn't contain numbers (details on that are [here](https://github.com/saiarcot895/sonic-buildimage/pull/7#discussion_r732025357)). Additionally, the container update for bullseye arm64 seems to have been [disabled in the multiarch repo](https://github.com/multiarch/debian-debootstrap/blob/master/.github/workflows/actions.yml#L27), although I'm not entirely sure why.

Updating the `base-files` package updates `/etc/debian_version` to the current version of the file, and then `lsb_release -cs` works normally.

sonic-slave-bullseye/Dockerfile.j2
saiarcot895: I believe it's so that if the `qemu-*-static` binaries are not loaded and persisted in memory (i.e. by using the `--persistent` flag), then there will be a version inside the container that can be used. Looks like arm has similar code as well.

lguohan:
can you fix the lgtm?

lguohan (approved):

PR #8223, opened 2021-07-20
hostcfgd: Handle missed tacacs updates between load & listen

renukamanavalan:
#### Why I did it
The time gap between last  config load & db-listen seem to have increased.
Any config updates that occurred in this gap gets missed by db-listen.
This could miss updating /etc/pam.d/common-auth-sonic

#### How I did it
Add a one shot timer, just before db-listen. The timer will fire after the subscribe is done
When the timer fires, reload tacacs & aaa

#### How to verify it
A tough repro. So you may explicitly add  some sleep just before listen.
Make sure the sleep is at least 10 seconds less than timer.
Write a log just before sleep.
restart hostcfgd and watch syslog for the added log line
Upon this log, make some tacacs updates that would have some change in
/etc/pam.d/common-auth-sonic.
Watch for changes in that file

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106

### DIFF
--- a/files/image_config/hostcfgd/hostcfgd
+++ b/files/image_config/hostcfgd/hostcfgd
@@ -4,6 +4,7 @@
 import ast
 import os
 import time
+import threading
 import sys
 import subprocess
 import syslog
@@ -24,6 +25,20 @@ NSS_CONF = "/etc/nsswitch.conf"
 TACPLUS_SERVER_PASSKEY_DEFAULT = ""
 TACPLUS_SERVER_TIMEOUT_DEFAULT = "5"
 TACPLUS_SERVER_AUTH_TYPE_DEFAULT = "pap"
+
+global_lock = None
+
+class lock_mgr:
+    def __init__(self):
+        self.lock = global_lock
+
+    def __enter__( self ):
+        if self.lock:
+            self.lock.acquire()
+
+    def __exit__( self, exc_type, exc_value, traceback ):
+        if self.lock:
+            self.lock.release()
 
 
 def is_true(val):
@@ -123,7 +138,7 @@ class Iptables(object):
                               .format(err.cmd, err.returncode, err.output))
 
 class AaaCfg(object):
-    def __init__(self):
+    def __init__(self, config_db):
         self.auth_default = {
             'login': 'local',
         }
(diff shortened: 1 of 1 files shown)


files/image_config/hostcfgd/hostcfgd
```diff
@@ -25,6 +26,20 @@ TACPLUS_SERVER_PASSKEY_DEFAULT = ""
 TACPLUS_SERVER_TIMEOUT_DEFAULT = "5"
 TACPLUS_SERVER_AUTH_TYPE_DEFAULT = "pap"
 
+global_lock = None
```
arlakshm, line 29: iso of `global_lock`, can we move this lock to the` AaaCfg()` ? To prevent another classes in this file from using this lock

arlakshm (commented):
Do we need this change in master as well ?

files/image_config/hostcfgd/hostcfgd
renukamanavalan: If anyone else (_another subscriber_) gets a need, they could use this.

abdosi:
@renukamanavalan and @yxieca  : when the issue is seen the steps are:-

a) load minigraph
b) config tacacs * commands

Basically delay bewteen restart hostcfgd and config command is 10-15 seconds. I think if we can randomize delay we might be able to repro this

Restart of hostcfgd service as part of load minigraph
image
Config save after load minigrpah

image

Tacacs configuration
image

files/image_config/hostcfgd/hostcfgd
```diff
@@ -136,49 +151,42 @@ class AaaCfg(object):
...
+
     def modify_single_file(self, filename, operations=None):
         if operations:
             cmd = "sed -e {0} {1} > {1}.new; mv -f {1} {1}.old; mv -f {1}.new {1}".format(' -e '.join(operations), filename)
             os.system(cmd)
 
     def modify_conf_file(self):
+        with lock_mgr():
+            self.auth = self.config_db.get_table('AAA').get("authentication", {})
+            if 'failthrough' in self.auth:
```
abdosi, line 179: can we add some syslog in the AAA load() so that we can see going forward timestamp when these gets executed.

files/image_config/hostcfgd/hostcfgd
```diff
@@ -537,6 +554,10 @@ class HostConfigDaemon:
         # Defer load until subscribe
         self.load()
```
abdosi, line 555: syslog here also will be useful

renukamanavalan:
> @renukamanavalan and @yxieca : when the issue is seen the steps are:-
> 
> a) load minigraph
> b) config tacacs * commands
> 
> Basically delay bewteen restart hostcfgd and config command is 10-15 seconds. I think if we can randomize delay we might be able to repro this
> 
> Restart of hostcfgd service as part of load minigraph
> Config save after load minigrpah
> 
> 
> Tacacs configuration

Managed a repro with following script. I could repro reliably in two systems. In one case, the pause time of 5 seconds worked, but in another it took 11 seconds. Again this is pretty dependent on many other factors. With help of load_minigraph, we could put control plane on high load, which helped get a distinct pause between load & listen. 

```
for ((;;))
do
    tsleep=1
    tinterval=1

    for ((i=0; i<16; i++))
    do
        for ((j=0; j<5; j++))
        do
            config load_minigraph -y
            sleep 30s
            systemctl restart hostcfgd

            # variable sleep since hostcfgd restart
            sleep ${tsleep}

            # Update config
            config tacacs passkey xxxx
            config tacacs  authtype login
            config aaa authentication failthrough enable
            config aaa authentication login tacacs+
            echo "tacacs configured; sleep 30s before testing pam file"

            # Pause to allow hostcfgd to react
            sleep 30s

            # Check pam.d conf file
            n=$(grep testing123 /etc/pam.d/common-auth-sonic | wc -l)
            if [ $n -ne 1 ]; then
                echo "$(date): sleep little longer"
                sleep 30s
                n=$(grep testing123 /etc/pam.d/common-auth-sonic | wc -l)
                if [ $n -ne 1 ]; then
                    echo "$(date): Failure encountered"
                    exit -2
                fi
            fi
            echo "$(date): Run i=$i j=$j tsleep=${tsleep} complete"
        done
        tsleep=$((tsleep + tinterval))
    done
done

```

files/image_config/hostcfgd/hostcfgd
```diff
@@ -136,49 +151,42 @@ class AaaCfg(object):
...
+    def tacacs_global_update(self, key):
         if key == 'global':
-            self.tacplus_global = data
-            if modify_conf:
-                self.modify_conf_file()
-
-    def tacacs_server_update(self, key, data, modify_conf=True):
-        if data == {}:
-            if key in self.tacplus_servers:
-                del self.tacplus_servers[key]
```
abdosi, line 169: do we don't need this code anymore ? all these changes looks unrelated to this fix.
Can we have separate PR for reason of these changes or at least  can we add to PR description

renukamanavalan: We don't need this code anymore. The whole purpose of this code is to keep a local cache of tacacs & aaa. Maintain the cache on updates and use the cache. We don't do caching anymore.

We get it when we need it. This handles the missed-updates in simple manner and btw, we don't gain any by keeping duplicate cache in mem.

abdosi (approved):

PR #8074, opened 2021-07-06
[Juniper] Migrating to SONiC Platform API package

ciju-juniper:
This patch set introduce platform API package for QFX5210 & QFX5200 platforms

Signed-off-by: Ciju Rajan K <crajank@juniper.net>

PR #8257, opened 2021-07-26
[Juniper] Migrating to SONiC Platform API package

ciju-juniper:
This patch set introduce platform API package for  QFX5210 & QFX5200 platforms

Signed-off-by: Ciju Rajan K <crajank@juniper.net>

This is the follow-up pull request from https://github.com/Azure/sonic-buildimage/pull/8074


ciju-juniper:
@jleveque  Could you help me to understand what is wrong with the constructor calls? Lgtm issues the warning "17 for Wrong number of arguments in a class instantiation". I tried to suppress it, but still the warnings are mentioned in the summary. There are not any warnings shown in the lgtm code view.

PR #8274, opened 2021-07-28
[docker-orchagent][201911] Pass ASIC vendor information to swss docker as docker level environment variable

stephenxs:
#### Why I did it
Recently, the reserved buffer of admin-down ports is going to be reclaimed.
However, the way to do this differs among vendors.
We need to find a way to pass vendor information to swss docker.

#### How I did it
Fetch the ASIC vendor information when the docker is created and pass it to the docker as environment variable `ASIC_VENDOR`.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106

### DIFF
--- a/files/build_templates/docker_image_ctl.j2
+++ b/files/build_templates/docker_image_ctl.j2
@@ -245,11 +245,6 @@ start() {
     echo "Creating new {{docker_container_name}}$DEV container with HWSKU $HWSKU"
     {%- endif %}
 
-    {%- if docker_container_name == "swss" %}
-    # Obtain the vendor name
-    ASIC_VENDOR=`$SONIC_CFGGEN -y /etc/sonic/sonic_version.yml -v asic_type`
-    {%- endif %}
-
     # In Multi ASIC platforms the global database config file database_global.json will exist.
     # Parse the file and get the include path for the database_config.json files used in
     # various namesapces. The database_config paths are relative to the DIR of SONIC_DB_GLOBAL_JSON.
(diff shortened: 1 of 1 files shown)


files/build_templates/docker_image_ctl.j2
```diff
@@ -245,6 +245,11 @@ start() {
     echo "Creating new {{docker_container_name}}$DEV container with HWSKU $HWSKU"
     {%- endif %}
 
+    {%- if docker_container_name == "swss" %}
+    # Obtain the vendor name
+    ASIC_VENDOR=`$SONIC_CFGGEN -y /etc/sonic/sonic_version.yml -v asic_type`
```
neethajohn, line 250: Use the 'sonic_asic_platform' var instead of defining  "ASIC_VENDOR"

stephenxs: fixed.

qiluo-msft:
You can get the info from inside swss container
```
sonic-cfggen -y /etc/sonic/sonic_version.yml -v asic_type
```

If you still want to add a PR, could you add the feature on master first and request a cherry-pick?

stephenxs:
> You can get the info from inside swss container
> 
> ```
> sonic-cfggen -y /etc/sonic/sonic_version.yml -v asic_type
> ```
> 
> If you still want to add a PR, could you add the feature on master first and request a cherry-pick?

Thanks, Qi.
I think fetching it once when the container is created is better than fetching it every time the daemon starts in swss container.
In 202012 and above, we have similiar logic.

files/build_templates/docker_image_ctl.j2
qiluo-msft: I agree with @neethajohn 's suggestion and it should go to master, 202012 and later braches.

lguohan:
what is the motivation to move expose asic vendor to swss docker?

neethajohn:
> what is the motivation to move expose asic vendor to swss docker?

Related to the comments in https://github.com/Azure/sonic-swss/pull/1837

files/build_templates/docker_image_ctl.j2
stephenxs: @qiluo-msft PR to master created #8361

keboliu (approved):

qiluo-msft (approved):

PR #8289, opened 2021-07-30
[sonic-frr/Makefile]: Checkout gracefully if branch exists already.

praveen-li:
This is useful to avoid "branch exists" error while rebuild FRR.

Signed-off-by: Praveen Chaudhary<pchaudhary@linkedin.com>

#### Why I did it

If branch exist already, today FRR build fails. Branch may exist on rebuild.
With this PR,  we will Checkout branch gracefully if branch exists already.

#### How I did it

Use git checkout $(FRR_BRANCH) || git checkout -b $(FRR_BRANCH) origin/$(FRR_BRANCH) in src/sonic-frr/Makefile.

#### How to verify it

Rebuild works fine. Fresh build works fine as well.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106

### DIFF
--- a/src/sonic-frr/Makefile
+++ b/src/sonic-frr/Makefile
@@ -10,7 +10,7 @@ STG_BRANCH = stg_temp.$(SUFFIX)
 $(addprefix $(DEST)/, $(MAIN_TARGET)): $(DEST)/% :
 	# Build the package
 	pushd ./frr
-	git checkout -b $(FRR_BRANCH) origin/$(FRR_BRANCH)
+	git checkout $(FRR_BRANCH) || git checkout -b $(FRR_BRANCH) origin/$(FRR_BRANCH)
 	stg branch --create $(STG_BRANCH) $(FRR_TAG)
 	stg import -s ../patch/series
 	tools/tarsource.sh -V -e '-sonic'
--- a/src/sonic-frr/Makefile
+++ b/src/sonic-frr/Makefile
@@ -10,7 +10,7 @@ STG_BRANCH = stg_temp.$(SUFFIX)
 $(addprefix $(DEST)/, $(MAIN_TARGET)): $(DEST)/% :
 	# Build the package
 	pushd ./frr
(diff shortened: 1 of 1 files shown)


src/sonic-frr/Makefile
```diff
@@ -10,7 +10,7 @@ STG_BRANCH = stg_temp.$(SUFFIX)
 $(addprefix $(DEST)/, $(MAIN_TARGET)): $(DEST)/% :
 	# Build the package
 	pushd ./frr
-	git checkout -b $(FRR_BRANCH) origin/$(FRR_BRANCH)
+	git checkout $(FRR_BRANCH) || git checkout -b $(FRR_BRANCH) origin/$(FRR_BRANCH)
```
shi-su, line 13: In what scenarios does this issue happen? I thought a normal rebuild should not trigger this since the branch should get removed after the build.

praveen-li: Yeah, I notice that we clear branch at end. I think, on fresh Azure code, I notice FRR build failure first, and then i tried  to rebuild immediately, I notice 'branch exists' failure. So trying to addressed it. Seems, I should have tried to debug\resolve the first failure.  
Anyways, right now, if you do not see any side effect then we can put this fix, it helps ,while working on FRR patch in /patch/series folder. @shi-su

shi-su: Would this command `git checkout -b $(FRR_BRANCH) origin/$(FRR_BRANCH) || git checkout $(FRR_BRANCH)` work for this case? Though I currently could not think about any cases that cause side effects with your fix, I am a little unassured. I felt swapping the two commands may better align with the original behavior.

praveen-li: Agree, let me change it and we can keep both commands.

shi-su (approved):

liushilongbuaa:
Approve for 202106 branch to avoid build failure.

PR #8307, opened 2021-08-02
Add thrift 0.13.0 

msosyak:
#### Why I did it
To bump thrift version to 0.13.0, to fix some dependencies issues.

#### How I did it
As there are dependencies between thrift and saithrinft server  (https://github.com/opencomputeproject/SAI/tree/bf3630316cbef484e1015478b5746067bac4e2b8/test/saithrift) which is used by syncd-rpc to update thrift version, I also need to make changes in saithrinft server, and then SAI ref point should be updated in sairedis, and then sairedis ref point should be updated too. It is too many change, so I decided to add thrift 0.13.0 as separeate target to be able to work and test father changes in saithrinft and one when appropriate changes will be merged to SAI and ref points will be updated I will squash this and the old thrift target.  I was not able to build thrift deb pkg by original rules, so I copied `debian `folder from the old version and tune it for newer one.

#### How to verify it
```
make init
make configure PLATFORM=vs
make target/debs/buster/libthrift_0.13.0_amd64.deb
```

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106

```
 /\__/\
(=^.^=)
 (")(")_/
```


msosyak:
@qiluo-msft @xumia Please review

qiluo-msft:
Do you get the files from upstream? If mostly yes, do not copy them, instead, download it from a upstream repo, and apply your patch, and then build.

ref: https://github.com/Azure/sonic-buildimage/tree/master/src/initramfs-tools

msosyak:
I was not able to build thrift deb pkg by original rules https://github.com/apache/thrift/tree/0.13.0/debian , so I copied debian folder from the old version https://github.com/Azure/sonic-buildimage/blob/91432fff65347464d5e67cb80a39493acb91e570/src/thrift/Makefile#L19  and tune it for a newer one.  I have tried to add patches to not copy but download and patch. But there are many changes including files renaming and deleting, so the copy of all debian folders looks more clear.

I have also tried to search for the newer version of thrift packages on https://sonicstorage.blob.core.windows.net/packages/debian, but looks like I do not have permission to browse the storage or do not found the proper link to do it. Will be very appreciate if you share the info on how to browse https://sonicstorage.blob.core.windows.net/packages/debian and push/update files there if it is possible.

qiluo-msft:
Did you try https://packages.debian.org/bullseye/libthrift-0.13.0?
Download Source Package thrift:
[thrift_0.13.0-6.dsc]
[thrift_0.13.0.orig.tar.gz]
[thrift_0.13.0-6.debian.tar.xz]

msosyak:
Yes. It is a bullseye package and it cant be installed due to unmet dependencies till SONiC is buster based. Also, we need to install the thrift package not only in SONiC host but in syncd-rpc and docker-ptf containers too, which are not always the same Debian based. And it is more flexible to build thrift from sources for appropriate env then use prebuilt package.

msosyak:
No, I haven't tried. I missed that there is an archive with the source code. Just tried and it works. Thank you for the tip. I have updated PR to use source archive from debian.org

src/thrift_0_13_0/.gitignore
```diff
@@ -0,0 +1,3 @@
+!0001-Disable-unneeded-packages.patch
+!Makefile
+!series
```
qiluo-msft, line 3: Seems this file is not necessary. They are not ignored by default.

src/thrift_0_13_0/Makefile
```diff
@@ -0,0 +1,30 @@
...
+DERIVED_TARGETS = libthrift-dev_$(THRIFT_VERSION_FULL)_$(CONFIGURED_ARCH).deb \
+		  python3-thrift_$(THRIFT_VERSION_FULL)_$(CONFIGURED_ARCH).deb \
+		  thrift-compiler_$(THRIFT_VERSION_FULL)_$(CONFIGURED_ARCH).deb
+
+THRIFT_LINK_PRE = http://deb.debian.org/debian/pool/main/t/thrift/
+
+$(addprefix $(DEST)/, $(MAIN_TARGET)): $(DEST)/% :
+	rm -rf thrift-$(THRIFT_VERSION)
+
+	wget "$(THRIFT_LINK_PRE)/thrift_$(THRIFT_VERSION_FULL).debian.tar.xz"
```
qiluo-msft, line 18: >wget [](http://example.com/codeflow?start=1&length=4)

If you run `make` twice, 2nd time wget will rename because name conflicts with local.

rules/thrift_0_13_0.dep
```diff
@@ -0,0 +1,10 @@
+
+SPATH       := $($(LIBTHRIFT_0_13_0)_SRC_PATH)
+DEP_FILES   := $(SONIC_COMMON_FILES_LIST) rules/thrift_0_13_0.mk rules/thrift_0_13_0.dep
+DEP_FILES   += $(SONIC_COMMON_BASE_FILES_LIST)
+DEP_FILES   += $(shell git ls-files $(SPATH))
+
+$(LIBTHRIFT_0_13_0)_CACHE_MODE  := GIT_CONTENT_SHA
+$(LIBTHRIFT_0_13_0)_DEP_FLAGS   := $(SONIC_COMMON_FLAGS_LIST)
+$(LIBTHRIFT_0_13_0)_DEP_FILES   := $(DEP_FILES)
+
```
qiluo-msft, line 10: Remove extra empty lines at top and bottom

src/thrift_0_13_0/Makefile
```diff
@@ -0,0 +1,30 @@
...
+	wget "$(THRIFT_LINK_PRE)/thrift_$(THRIFT_VERSION).orig.tar.gz"
+	wget "$(THRIFT_LINK_PRE)/thrift_$(THRIFT_VERSION_FULL).dsc"
+
+	dpkg-source -x thrift_$(THRIFT_VERSION_FULL).dsc
+	pushd thrift-$(THRIFT_VERSION)
+
+	DEB_BUILD_OPTIONS=nocheck dpkg-buildpackage -d -rfakeroot -b -us -uc -j$(SONIC_CONFIG_MAKE_JOBS) --admindir $(SONIC_DPKG_ADMINDIR)
```
qiluo-msft, line 25: >buildpackage [](http://example.com/codeflow?start=32&length=12)

Did you actually apply your patch before the build?

src/thrift_0_13_0/Makefile
```diff
@@ -0,0 +1,30 @@
+SHELL = /bin/bash
+.ONESHELL:
+.SHELLFLAGS += -e
+
+THRIFT_VERSION = 0.13.0
+THRIFT_VERSION_FULL = $(THRIFT_VERSION)-6
+
+MAIN_TARGET = libthrift-$(THRIFT_VERSION)_$(THRIFT_VERSION_FULL)_$(CONFIGURED_ARCH).deb
```
qiluo-msft, line 8: >libthrift [](http://example.com/codeflow?start=14&length=9)

Is it true that you only need to build it for Debian versions before bulleyes?

rules/thrift_0_13_0.dep
msosyak: done

src/thrift_0_13_0/Makefile
msosyak: Yep, missed few lines

src/thrift_0_13_0/Makefile
msosyak: It is better to build from sources for bulleyes too

src/thrift_0_13_0/Makefile
qiluo-msft: `wget -O` will be better than `rm` a wildcard.

src/thrift_0_13_0/.gitignore
qiluo-msft: Still have this concern.

src/thrift_0_13_0/patch/0001-Disable-unneeded-packages.patch
```diff
@@ -0,0 +1,112 @@
...
+--- a/debian/control
++++ b/debian/control
+@@ -13,8 +13,8 @@ Build-Depends: debhelper-compat (= 11), dh-python,
+  python3-all-dev,
+  python3-six,
+  golang-go, golang-github-golang-mock-dev,
+- pkg-php-tools (>= 1.14~), php-dev, phpunit, php-codesniffer,
+- perl (>= 5.22), libbit-vector-perl, libclass-accessor-perl,
++#  pkg-php-tools (>= 1.14~), php-dev, phpunit, php-codesniffer,
++#  perl (>= 5.22), libbit-vector-perl, libclass-accessor-perl,
```
qiluo-msft, line 22: Look like you are commenting many code. Could you just delete instread of commenting. It will make patch more readable.

src/thrift_0_13_0/.gitignore
msosyak: Just follow the same approach as in src/thrift, but agree it is not the best.
Added here just one rule thrift*   to ignore downloaded archives and extracted files

src/thrift_0_13_0/Makefile
msosyak: Ok

src/thrift_0_13_0/patch/0001-Disable-unneeded-packages.patch
msosyak: Done

qiluo-msft (approved):

msosyak:
@qiluo-msft Your approval was dismissed when I force pushed the latest commit to retrigger the CI as it was failed.
Not the CI passed, so could we merge this?

qiluo-msft (approved):

msosyak:
@qiluo-msft Do you have some concerns or could we merge this?

PR #8309, opened 2021-08-02
[Mellanox][Pcie] Fix issue on pcied with an id that contains only decimal digits was treated as a decimal number

DavidZagury:
#### Why I did it
A device that contains only decimal digits was mistreated as a decimal integer resulting in failure to find it in the id to bus map.

#### How I did it
Made sure id from the yaml file was treated as a string

#### How to verify it
Open log on 4600, make sure no error appear.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [X] 202012
- [X] 202106


BasimShalata:
@sujinmkang can you please review this PR.

sujinmkang (approved):

PR #8373, opened 2021-08-08
Fix make issue when the degree of parallelism is more than 4

liushilongbuaa:
#### Why I did it
dpkg lock can't avoid all parallel installing.
#### How I did it
Add retry for installing packages.
#### How to verify it
trigger one build on an 8 cores agent.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012
- [x] 202106


slave.mk
```diff
@@ -549,8 +549,13 @@ $(SONIC_INSTALL_DEBS) : $(DEBS_PATH)/%-install : .platform $$(addsuffix -install
 		$(foreach deb, $($*_CONFLICT_DEBS), \
 			{ while dpkg -s $(firstword $(subst _, ,$(basename $(deb)))) | grep "^Version: $(word 2, $(subst _, ,$(basename $(deb))))" &> /dev/null; do echo "waiting for $(deb) to be uninstalled" $(LOG); sleep 1; done } )
 		# put a lock here because dpkg does not allow installing packages in parallel
+		# this lock didn't avoid all parallel installing. Add some retry.
```
lguohan, line 552: why this lock did not avoid all parallel installing? what missed?

liushilongbuaa: From Guohan's PR, we can see it is the following package holding the lock:
src/openssh/Makefile:22:        sudo http_proxy=$(http_proxy) apt-get -y build-dep openssh

Maybe adding retry is still the best way to resolve this issue?

PR #8381, opened 2021-08-09
[openssh]: move build dep installation to sonic-slave-buster

lguohan:
install build dep causes dpkg lock issue in parallel build

#### Why I did it
install build dep causes dpkg lock issue in parallel build

#### How I did it
move build-dep into sonic-slave docker

#### How to verify it
did parallel build and check

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012
- [x] 202106

### DIFF
--- a/sonic-slave-buster/Dockerfile.j2
+++ b/sonic-slave-buster/Dockerfile.j2
@@ -348,6 +348,8 @@ RUN apt-get update && apt-get install -y \
         libsystemd-dev \
         pkg-config
 
+RUN apt-get -y build-dep openssh
+
 # Build fix for ARMHF buster libsairedis
 {%- if CONFIGURED_ARCH == "armhf" %}
        # Install doxygen build dependency packages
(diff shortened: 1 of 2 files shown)


sonic-slave-buster/Dockerfile.j2
```diff
@@ -348,6 +348,8 @@ RUN apt-get update && apt-get install -y \
         libsystemd-dev \
         pkg-config
 
+RUN apt-get -y build-dep openssh
```
qiluo-msft, line 351: >build-dep [](http://example.com/codeflow?start=15&length=9)

Not a blocking issue: I see `build-dep` installed twice in this file, so you can skip it.

qiluo-msft (approved):

sonic-slave-buster/Dockerfile.j2
lguohan: build-dep is a keyword, not a package.

qiluo-msft: Got it. My bad.

Issue #8389, opened 2021-08-09
[rsyslogd][multiasic] rsyslogd process exited with FATAL status in mutliasic platform database container


mlok-nokia:
#### Description

In multiasic platform, rsyslogd process exited with FATAL status in the global instance database container.  In the database container, /etc/rsyslog.conf is empty file.

#### Steps to reproduce the issue:
1.  go to the the database container: docker exec -it database bash
2. execute command :  supervisorctl status

#### Describe the results you received:
```
admin@sonic:~$ docker exec -it database bash
root@sonic:/# supervisorctl status
flushdb                          EXITED    Aug 08 05:33 PM
redis                            RUNNING   pid 38, uptime 21:32:12
rsyslogd                         FATAL     Exited too quickly (process log may have details)
supervisor-proc-exit-listener    RUNNING   pid 36, uptime 21:32:12
root@sonic:/# ls -al /etc/rsyslog.conf 
-rw-r--r-- 1 root root 0 Aug  8 17:28 /etc/rsyslog.conf
root@sonic:/# 
```
#### Describe the results you expected:
```
admin@sonic:~$ docker exec -it database bash
root@sonic:/# supervisorctl status
flushdb                          EXITED    Aug 08 03:20 AM
redis                            RUNNING   pid 38, uptime 1 day, 11:49:24
rsyslogd                         RUNNING   pid 37, uptime 1 day, 11:49:24
supervisor-proc-exit-listener    RUNNING   pid 36, uptime 1 day, 11:49:24
root@sonic:/# ls /etc/rsyslog.conf 
/etc/rsyslog.conf
root@sonic:/# ls -al /etc/rsyslog.conf 
-rw-r--r-- 1 root root 1924 Aug  7 21:39 /etc/rsyslog.conf 
```
#### Output of `show version`:
SONiC buildimage from commit 'https://github.com/Azure/sonic-buildimage/commit/3b3012791311ea7724975c8485236a0ee88d7510', 

However, looking at the latest code, the issue still exists.

mlok-nokia:
Issue is fixed

PR #8390, opened 2021-08-09
[rsyslogd][multiasic] fixed rsyslogd FATAL issue in the database container in multi-asic box

mlok-nokia:
#### Why I did it

Fix for issue https://github.com/Azure/sonic-buildimage/issues/8389

#### How I did it
The /etc/rsyslog.conf is empty file which cause the FATAL of the process rsyslogd in the global instance database container.  The function updateSyslogConf() should only generate the rsyslog.conf for containers in the namespace. it should not do it for the containers in the global instance. Instead, default rsyslog.conf should be used.  Especially for database container, updateSyslogConf() is called before the database container is created.  The result cause the sonic-cfggen failed to generate the rsyslog.conf.  

#### How to verify it
-- Go to the database container and execute the supervisorctl status command:
```
admin@sonic:~$ docker exec -it database bash
root@sonic:/# supervisorctl status
flushdb                          EXITED    Aug 08 03:20 AM
redis                            RUNNING   pid 38, uptime 1 day, 11:49:24
rsyslogd                         RUNNING   pid 37, uptime 1 day, 11:49:24
supervisor-proc-exit-listener    RUNNING   pid 36, uptime 1 day, 11:49:24
root@sonic:/# ls /etc/rsyslog.conf 
/etc/rsyslog.conf
root@sonic:/# ls -al /etc/rsyslog.conf 
-rw-r--r-- 1 root root 1924 Aug  7 21:39 /etc/rsyslog.conf
root@sonic2:/# 
```
-- /etc/rsyslog.conf should not be a empty file

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106

### DIFF
--- a/files/build_templates/docker_image_ctl.j2
+++ b/files/build_templates/docker_image_ctl.j2
@@ -36,8 +36,11 @@ function updateSyslogConf()
         TARGET_IP=$(docker network inspect bridge --format={{ "'{{(index .IPAM.Config 0).Gateway}}'" }})
         CONTAINER_NAME="$DOCKERNAME"
         TMP_FILE="/tmp/rsyslog.$CONTAINER_NAME.conf"
-
+        {%- if docker_container_name == "database" %} 
+        python -c "import jinja2, os; paths=['/usr/share/sonic/templates']; loader = jinja2.FileSystemLoader(paths); env = jinja2.Environment(loader=loader, trim_blocks=True); template_file='/usr/share/sonic/templates/rsyslog-container.conf.j2'; template = env.get_template(os.path.basename(template_file)); data=template.render({\"target_ip\":\"$TARGET_IP\",\"container_name\":\"$CONTAINER_NAME\"}); print(data)" > $TMP_FILE
+        {%- else %}
         sonic-cfggen -t /usr/share/sonic/templates/rsyslog-container.conf.j2 -a "{\"target_ip\": \"$TARGET_IP\", \"container_name\": \"$CONTAINER_NAME\" }"  > $TMP_FILE
+        {%- endif %}
         docker cp $TMP_FILE ${DOCKERNAME}:/etc/rsyslog.conf
         rm -rf $TMP_FILE
     fi


zhangyanzhao:
@arlakshm  please help to review. Thanks.

files/build_templates/docker_image_ctl.j2
```diff
@@ -36,8 +36,11 @@ function updateSyslogConf()
         TARGET_IP=$(docker network inspect bridge --format={{ "'{{(index .IPAM.Config 0).Gateway}}'" }})
         CONTAINER_NAME="$DOCKERNAME"
         TMP_FILE="/tmp/rsyslog.$CONTAINER_NAME.conf"
-
+        {%- if docker_container_name == "database" %} 
+        python -c "import jinja2, os; paths=['/usr/share/sonic/templates']; loader = jinja2.FileSystemLoader(paths); env = jinja2.Environment(loader=loader, trim_blocks=True); template_file='/usr/share/sonic/templates/rsyslog-container.conf.j2'; template = env.get_template(os.path.basename(template_file)); data=template.render({\"target_ip\":\"$TARGET_IP\",\"container_name\":\"$CONTAINER_NAME\"}); print(data)" > $TMP_FILE
+        {%- else %}
```
arlakshm, line 41: This code is needed only for containers which are attached to namespace. Can we add check to execute this  block only  if the `$DEV` is set?

something like this?

```
 # Also update the container name
    if [[ ($NUM_ASIC -gt 1)  && "$DEV" ]]; then
        TARGET_IP=$(docker network inspect bridge --format='{{(index .IPAM.Config 0).Gateway}}')
        CONTAINER_NAME="$DOCKERNAME"
        TMP_FILE="/tmp/rsyslog.$CONTAINER_NAME.conf"

        sonic-cfggen -t /usr/share/sonic/templates/rsyslog-container.conf.j2 -a "{\"target_ip\": \"$TARGET_IP\", \"container_name\": \"$CONTAINER_NAME\" }"  > $TMP_FILE
        docker cp $TMP_FILE ${DOCKERNAME}:/etc/rsyslog.conf
        rm -rf $TMP_FILE
    fi

```

arlakshm (changes requested):
as comments

files/build_templates/docker_image_ctl.j2
arlakshm: In multi asic platform the container attached to namespace are in a docker bridge network. 
we use `240.127.1.1`(docker bridge ip on the host) rsyslog server address to make sure syslogs generated in containers attached to namespace reaches the host.

For containers running on host, they can use the `127.0.0.1` as the syslog server because they are not in the docker bridge network

arlakshm (approved):
lgtm

PR #8416, opened 2021-08-10
[macsec] Add eapol to copp config

byu343:
#### Why I did it
This change enables the control packets of MACsec to be processed by CPU.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106

### DIFF
--- a/files/image_config/copp/copp_cfg.j2
+++ b/files/image_config/copp/copp_cfg.j2
@@ -91,7 +91,7 @@
 		    "trap_ids": "ip2me",
 		    "trap_group": "queue1_group1"
 	    },
-	    "eapol": {
+	    "macsec": {
 		    "trap_ids": "eapol",
 		    "trap_group": "queue4_group3"
 	    },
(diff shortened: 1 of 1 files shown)


lguohan:
@prsunny, will this break on platforms that does not support this trap id?

prsunny:
> @prsunny, will this break on platforms that does not support this trap id?

Yes, it will break. We should have a corresponding feature name and only enable for those platforms that support it. I see that we've a [macsec ](https://github.com/Azure/sonic-buildimage/blob/master/files/build_templates/init_cfg.json.j2#L38)feature. Can the trap be named as "macsec" instead of "eapol"?

lguohan:
@prsunny, how do you suggest to move ahead? how can we enable this trap when macsec feature is enabled?

files/image_config/copp/copp_cfg.j2
```diff
@@ -91,6 +91,10 @@
 		    "trap_ids": "ip2me",
 		    "trap_group": "queue1_group1"
 	    },
+	    "eapol": {
```
lguohan, line 94: change name to eapol -> macsec, then accordingly to prince, this trap will not be installed until the feature is enabled.

lguohan: @byu343, can you do that? then we can merge.

lguohan (approved):

PR #8417, opened 2021-08-10
[docker-ptf]PTF adds xmlrunner for xunit report

richardyu-ms:
#### Why I did it
add xmlrunner to generate the test report in xunit format

#### How I did it
add xmlrunner to generate the test report in xunit format

#### How to verify it
built the ptf docker and check the function with test cases
- vxlan/test_vxlan_decap.py
- local test case for sai testing

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012
- [x] 202106

### DIFF
--- a/dockers/docker-ptf/Dockerfile.j2
+++ b/dockers/docker-ptf/Dockerfile.j2
@@ -120,6 +120,7 @@ RUN rm -rf /debs \
     && pip install exabgp==3.4.17\
     && pip install pyaml   \
     && pip install pybrctl pyro4 rpyc yabgp \
+    && pip install unittest-xml-reporting \
     && mkdir -p /opt       \
     && cd /opt             \
     && wget https://raw.githubusercontent.com/p4lang/ptf/master/ptf_nn/ptf_nn_agent.py


wangxin (commented):
The xmlrunner package is pretty old. Not sure if it supports python3 well. We need to consider python3 support while adding packages. Is the unittest-xml-reporting package a better option?

Reference: https://github.com/xmlrunner/unittest-xml-reporting

bingwang-ms (approved):

wangxin (approved):

lguohan:
@wangxin , why not using pytest report? can we converge on using one report same as sonic-mgmt?

wangxin:
@guohan, talked with @richardyu-ms, each SAI PTF script has many test cases. However, for a pytest case running the PTF script, it will only report single result based on pass or fail of the PTF script. It would be better for the PTF script to generate a junit xml result with more fine grained results. The junit xml result can be copied back to sonic-mgmt. Then we can take advantage of the existing tools to process the PTF script junit XML result files to generate better test reports.

PR #8455, opened 2021-08-13
[voq][bgpcfg] BGP_VOQ_CHASSIS_NEIGHBORS timers config

vganesan-nokia:
#### Why I did it

Currently, voq chassis bgp neighbor timer configuration is done similar to general neighbors. i.e., keepalive and holdtime timers are set to configured values, if configured,, otherwise set to default of 60 and 180. This is causing a problem of configuring timer values to 0 & 0 since minigraph configures 0 & 0 for voq chassis neighbors. Voq chassis neighbors are equivalent to multi-asic internal neighbors but we need to configure timers to values other than 0 & 0 since the voq chassis bgp neighbors may go down due to line card failures, for example. This PR fixes this problem of having 0 and 0 timer confiuration for voq chassis neighbors.

#### How I did it

Since the voq chassis bgp neighbors are similar to multi-asic internal neighbors, the voq chassis bgp timers are configured similar to multi-asic internal bgp neighbors. i.e, 3 & 10 are configured (hardcoded) for keepalive and holdtime respectively and the configurations given via config db are ignored. Also, similar to bgp internal neighbors, connection retry timer is also configured for voq chassis bgp neighbors.

#### How to verify it

- Configure BGP_VOQ_CHASSIS_NEIGHBOR using minigraph. The minigraph will generate these neighbors with keepalive and holdtime timer values of 0 & 0 respectively.
- Reboot the chassis with this configuration
- After chassis is up and operational, check the bgp summary and observe voq chassis bgp sessions are up.
- Check the bgp confiuration for the internal neighbors in FRR (using show running-configuration, for example) and observe 3 & 10 are configured for keepalive and holdtime timers for voq chassis bgp neighbors. Also observe that connection retry timer is configured to 10.

N/A

Keepalive, Holddown and Connection retry timers configuration for BGP_VOQ_CHASSIS_NEIGHBORs

### DIFF
--- a/dockers/docker-fpm-frr/frr/bgpd/templates/voq_chassis/instance.conf.j2
+++ b/dockers/docker-fpm-frr/frr/bgpd/templates/voq_chassis/instance.conf.j2
@@ -7,11 +7,8 @@
   neighbor {{ neighbor_addr }} peer-group VOQ_CHASSIS_PEER
   neighbor {{ neighbor_addr }} remote-as {{ bgp_session['asn'] }}
   neighbor {{ neighbor_addr }} description {{ bgp_session['name'] }}
-{# set the bgp neighbor timers if they have not default values #}
-{% if     (bgp_session['keepalive'] is defined and bgp_session['keepalive'] | int != 60)
-      or  (bgp_session['holdtime'] is defined  and bgp_session['holdtime']  | int != 180) %}
-  neighbor {{ neighbor_addr }} timers {{ bgp_session['keepalive'] | default("60") }} {{ bgp_session['holdtime'] | default("180") }}
-{% endif %}
+  neighbor {{ neighbor_addr }} timers 3 10
+  neighbor {{ neighbor_addr }} timers connect 10
 !
 {% if 'admin_status' in bgp_session and bgp_session['admin_status'] == 'down' or 'admin_status' not in bgp_session and 'default_bgp_status' in CONFIG_DB__DEVICE_METADATA['localhost'] and CONFIG_DB__DEVICE_METADATA['localhost']['default_bgp_status'] == 'down' %}
   neighbor {{ neighbor_addr }} shutdown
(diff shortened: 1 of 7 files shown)


anshuv-mfst:
Hi @ysmanman - could you please assign Arista reviewers to the PR, thanks.

ysmanman:
Add @jmmikkel who was reviewing the change.

jmmikkel:
The code change is trivial but I don't understand the motivation for it. If the timers should't be set to 0, why are they 0 in the minigraph? Shouldn't whatever's generating the minigraph be what's aware of the differences between peers in the different tables?

Because by ignoring the config, you also remove the ability for someone to set the timers themselves. Maybe that's not considered important?

vganesan-nokia:
> The code change is trivial but I don't understand the motivation for it. If the timers should't be set to 0, why are they 0 in the minigraph? Shouldn't whatever's generating the minigraph be what's aware of the differences between peers in the different tables?
> 
> Because by ignoring the config, you also remove the ability for someone to set the timers themselves. Maybe that's not considered important?

It seems yes. When this issue was brought up in one of the sonic chassis subroup meetings, all agreed that voq chassis neighbors are similar to multi-asic internal neighbors and these timers should be hard coded similar to bgp_internal neighbors.

jmmikkel:
> It seems yes. When this issue was brought up in one of the sonic chassis subroup meetings, all agreed that voq chassis neighbors are similar to multi-asic internal neighbors and these timers should be hard coded similar to bgp_internal neighbors.

VOQ neighbors are indeed *similar* to multi-asic, but they are not the *same*. The fact you need this change demonstrates that treating them the *same* elsewhere is wrong, so why not change the assumption elsewhere?

However, if it has already been determined that this explicit ignoring of config is "right" I am not sure why my review is required.

arlakshm (approved):

abdosi (approved):

PR #8467, opened 2021-08-13
[ci]: Change to build marvell-armhf on armhf pool

xumia:
#### Why I did it
Change the agent pool to build marvell-armhf to reduce the build time.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106


qiluo-msft:
Should this PR go master branch first, and then cherry-picked to 202012?

lguohan:
arm pool is too small do we want to enable it now?

xumia:
> arm pool is too small do we want to enable it now?

Is it ok to only enable for 202012 official build?

Issue #8404, opened 2021-08-10
[Broadcom platform] CLI change MTU causes portmgrd crashed


vincentchiang-ec:
#### Description

Use CLI to change MTU, portmgrd is crashed.

#### Steps to reproduce the issue:
1. Use CLI "sudo config interface mtu Ethernet0 1501" to change MTU value

#### Describe the results you received:
Syslog:
```
Aug 10 00:51:19.879027 sonic NOTICE swss#orchagent: :- doPortTask: Set port Ethernet0 MTU to 1501
Aug 10 00:51:19.880180 sonic NOTICE swss#orchagent: :- setRouterIntfsMtu: Set router interface Ethernet0 MTU to 1501
Aug 10 00:51:19.881378 sonic INFO swss#/supervisord: portmgrd Error: mtu greater than device maximum.
Aug 10 00:51:19.881936 sonic ERR swss#portmgrd: :- main: Runtime error: /sbin/ip link set dev "Ethernet0" mtu "1501" :
Aug 10 00:51:20.577196 sonic INFO swss#supervisord 2021-08-10 00:51:20,576 INFO exited: portmgrd (exit status 255; not expected)
Aug 10 00:51:21.590176 sonic INFO swss#/supervisor-proc-exit-listener: Process 'portmgrd' exited unexpectedly. Terminating supervisor 'swss'
```

#### Output of `show version`:

```
root@sonic:~# show ver

SONiC Software Version: SONiC.master.25538-5e435e05a
Distribution: Debian 10.10
Kernel: 4.19.0-12-2-amd64
Build commit: 5e435e05a
Build date: Mon Jul 26 15:32:45 UTC 2021
Built by: AzDevOps@sonic-build-workers-000J8D

Platform: x86_64-accton_as7816_64x-r0
HwSKU: Accton-AS7816-64X
ASIC: broadcom
ASIC Count: 1
Serial Number: None
Model Number: None
Hardware Revision: None
Uptime: 01:00:15 up 28 min,  2 users,  load average: 0.00, 0.28, 0.50
```

#### Additional information you deem important (e.g. issue happens only occasionally):
This issue is fixed by https://github.com/Azure/sonic-buildimage/pull/3895 before, but the modification is removed after https://github.com/Azure/sonic-buildimage/pull/6526 merged.

gechiang:
I am able to reproduce this.
[...]

PR #6526, opened 2021-01-21
[BCMSAI] Update BCMSAI debian to 4.3.0.10 with 6.5.21 SDK, and opennsl module to 6.5.21

smaheshm:
**- Why I did it**
  * Update BCMSAI debian to 4.3.0.10 with 6.5.21 SDK
  * Update opennsl module to 4.3.0.10

**- How I did it**
  * Merge BRCM 6.5.21 GPL modules
  * Update BCMSAI debian package pointer.

**- How to verify it**
  * Manual verification on switch.
```
admin@str-s6000-acs-8:~$ show version                                                                                                                                                 

SONiC Software Version: SONiC.bcmsai4.3.0-dirty-20210121.183527
Distribution: Debian 10.7
Kernel: 4.19.0-9-2-amd64
Build commit: 7d1622b6
Build date: Thu Jan 21 22:16:04 UTC 2021
Built by: samaddik@samaddik-vm-01

Platform: x86_64-dell_s6000_s1220-r0
HwSKU: Force10-S6000
ASIC: broadcom
ASIC Count: 1
Serial Number: 1QBRX42
Uptime: 00:25:47 up 11 min,  1 user,  load average: 12.36, 11.06, 5.82

Docker images:
REPOSITORY                    TAG                                 IMAGE ID            SIZE
docker-syncd-brcm             bcmsai4.3.0-dirty-20210121.183527   cd4348b7c5a8        640MB
docker-syncd-brcm             latest                              cd4348b7c5a8        640MB
docker-database               bcmsai4.3.0-dirty-20210121.183527   6bb6d360e390        394MB
docker-database               latest                              6bb6d360e390        394MB
docker-sflow                  bcmsai4.3.0-dirty-20210121.183527   ff127b4007df        405MB
docker-sflow                  latest                              ff127b4007df        405MB
docker-teamd                  bcmsai4.3.0-dirty-20210121.183527   2c002ebfe189        404MB
docker-teamd                  latest                              2c002ebfe189        404MB
docker-router-advertiser      bcmsai4.3.0-dirty-20210121.183527   f09b66265f1d        394MB
docker-router-advertiser      latest                              f09b66265f1d        394MB
docker-platform-monitor       bcmsai4.3.0-dirty-20210121.183527   695db9e59ac7        601MB
docker-platform-monitor       latest                              695db9e59ac7        601MB
docker-lldp                   bcmsai4.3.0-dirty-20210121.183527   2ad83b82f29f        434MB
[...]

PR #3895, opened 2019-12-13
[broadcom]: fix KNET MTU setting with Linux Kernel > 4.10.0

ishidawataru:
We need to set min_mtu/max_mtu properly to support MTU setting

https://github.com/torvalds/linux/commit/61e84623

Signed-off-by: Wataru Ishida <ishida@nel-america.com>

PR #8472, opened 2021-08-13
Reapply the fix to address set MTU > 1500 causing portmgrd crash on BRCM platforms

gechiang:
#### Why I did it
This is to address the issue (https://github.com/Azure/sonic-buildimage/issues/8404)
The original fix which was introduced by (https://github.com/Azure/sonic-buildimage/pull/3895) was removed when onboarding BRCM SAI 4.3  (https://github.com/Azure/sonic-buildimage/pull/6526).  
At that time this change was removed because there was another way to set the port default MTU to 9100 which caused this change being removed.  It was most likely not anticipated that user may need to change the port MTU via CLI at which time this fix when not present will cause portmrgd to crash.
I have create a Case with BRCM (CS00012203989) to ask them to look into this issue to either take this change in and include it to all their future SAI releases and patch their current SAI releases or provide a better alternative to address this issue at which time BRCM team should revert this PR when raising their official fix to address the set interface MTU > 1500 causing portmgrd crash issue.

#### How I did it
Add back the same change done by (https://github.com/Azure/sonic-buildimage/pull/3895)

#### How to verify it
Loaded the image and tested with the "sudo config interface mtu Ethernet0 1501" and observe MTU setting go through fine without causing any crash.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [X] 202012
- [X] 202106

Fix set MTU > 1500 on interface causing portmgrd crash on BRCM platforms

### DIFF
--- a/platform/broadcom/saibcm-modules/systems/linux/kernel/modules/bcm-knet/bcm-knet.c
+++ b/platform/broadcom/saibcm-modules/systems/linux/kernel/modules/bcm-knet/bcm-knet.c
@@ -6341,6 +6341,11 @@ bkn_init_ndev(u8 *mac, char *name)
         dev->mtu = rx_buffer_size;
     }
 
+#if (LINUX_VERSION_CODE >= KERNEL_VERSION(4,10,0))
+    dev->min_mtu = 68;
+    dev->max_mtu = rx_buffer_size;
+#endif
+
     /* Device vectors */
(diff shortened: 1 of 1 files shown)


smaheshm (approved):
Looks good. wait for the builds to pass

lguohan (approved):

PR #8502, opened 2021-08-17
Enable sysctl  fib_multipath_use_neigh

abdosi:
What I did:
Enable   fib_multipath_use_neigh for v4  
https://www.kernel.org/doc/Documentation/networking/ip-sysctl.txt

Why I did:
This is helpful if the neighbor are not directly connected then Kernel forward to unreachable neighbor option. With this option forwarding using neighbor state to be valid. More details: https://patchwork.ozlabs.org/project/netdev/patch/1459463081-20206-1-git-send-email-dsa@cumulusnetworks.com/

How I verify:
This is useful in chassis scenario where there is ECMP route on LC to reach other LC's. If the neighbor of LC's is not reachable without this fix kernel can use the path with that neighbor and cause traffic drop/BGP down. After this fix working as expected.


files/image_config/sysctl/sysctl-net.conf
```diff
@@ -37,3 +37,4 @@ net.ipv4.udp_l3mdev_accept=1
 net.core.rmem_max=3145728
 net.core.wmem_max=3145728
 net.core.somaxconn=512
+net.ipv4.fib_multipath_use_neigh=1
```
prsunny, line 40: any equivalent for ipv6?

prsunny (approved):

files/image_config/sysctl/sysctl-net.conf
abdosi: @prsunny looks like for ipv6 it is handle differently as the default behavior https://github.com/torvalds/linux/blob/v4.19/net/ipv6/route.c#L611.

IPv6 does not even use CONFIG_IP_MULTIPATH option also.

PR #8542, opened 2021-08-20
Qiluo/devicefolder

qiluo-msft:
#### Why I did it
Reorganize the folder structure under /device. Move
```
/device/PLATFORM_VENDOR/PLATFORM_STRING
```
to
```
/device/ASIC_VENDOR/PLATFORM_VENDOR/PLATFORM_STRING
```

And build sonic-device-data_1.0-1_all.deb with only platforms using specific ASIC_VENDOR.

#### Which release branch to backport (provide reason below if selected)

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106

PR #8552, opened 2021-08-23
sonic_installer verification

wen587:
#### Why I did it
sonic_installer should verify the platform matches the binary image before executing install operation
#### How I did it
Build up a dictionary that keeps the mapping of the platform to ASIC, then check if the expected ASIC matches the ASIC of image build.
#### How to verify it
Install mellanox image on broadcom will fail and vice versa.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106


qiluo-msft:
You mentioned 
> Install mellanox image on broadcom will fail and vice versa.
Did you test both onie install and sonic-installer?

dockers/docker-platform-monitor/docker-pmon.supervisord.conf.j2
```diff
@@ -150,7 +150,7 @@ dependent_startup_wait_for=rsyslogd:running
 
 {% if not skip_pcied %}
 [program:pcied]
-command=/usr/local/bin/pcied
```
qiluo-msft, line 153: >command [](http://example.com/codeflow?start=0&length=7)

This change seems not related to your goal.

installer/platform2asic.conf
```diff
@@ -0,0 +1,69 @@
+# List of two main platfrom: broadcom and mellanox
```
qiluo-msft, line 1: ># List of two main platfrom [](http://example.com/codeflow?start=0&length=27)

I am proposing a general solution in PR [https://github.com/Azure/sonic-buildimage/pull/8542](https://github.com/Azure/sonic-buildimage/pull/8542).

onie-mk-demo.sh
```diff
@@ -93,6 +94,7 @@ sed -i -e "s/%%DEMO_TYPE%%/$demo_type/g" \
        -e "s/%%ONIE_IMAGE_PART_SIZE%%/$onie_image_part_size/" \
        -e "s/%%EXTRA_CMDLINE_LINUX%%/$EXTRA_CMDLINE_LINUX/" \
        -e "s@%%OUTPUT_RAW_IMAGE%%@$output_raw_image@" \
+       -e "s/%%ONIE_MACHINE%%/$machine/g" \
```
qiluo-msft, line 97: >ONIE_MACHINE [](http://example.com/codeflow?start=15&length=12)

Naming is confusing.

`onie_machine` is an old concept in the world of ONIE bootloader, it means what is the running hardware.

What you are trying to name, is the target ASIC vendor related to SONiC image.

qiluo-msft, line 97: >$machine [](http://example.com/codeflow?start=30&length=8)

You are building the image ground truth into `installer.sh`. 

Alternatively, I suggest that `/etc/sonic/sonic_version.yml` asic_type field is the ground truth. Then you don't need to change `installer.sh`.

This is the same to other bootloader such as aboot. In future, we may have similar logic in bootloader other than ONIE.

wen587:
Please check out another approach.
https://github.com/Azure/sonic-buildimage/pull/8648
&
https://github.com/Azure/sonic-utilities/pull/1791

PR #523, opened 2017-04-20
[docker-ptf]replace ping utility from inetutils-ping to iputils-ping

maggiemsft:
the inetutils-ping does not have source interface '-I' option

Issue #521, opened 2017-04-20
teamd: Error: either "dev" is duplicate, or "boot" is a garbage.


stcheng:
This is because the folder /etc/teamd is empty.
Need to check if /etc/teamd folder is empty or not.
https://github.com/Azure/sonic-buildimage/pull/441/files could resolve this issue.
But line https://github.com/marian-pritsak/sonic-buildimage/blob/97a59f15d5c57b83e604efd50cb674a818f6fec9/dockers/docker-teamd/start.sh#L18 will still return /etc/teamd/* when the folder is empty.

stcheng:
no loner using /etc/teamd to store the static configurations
close this sissue

PR #512, opened 2017-04-17
Update SAI submodule

pavel-shirshov:

PR #510, opened 2017-04-16
[build]: Adding support for Free-Range-Routing stack.

rodnymolina:
- Extending SONiC building infrastructure to provide users
           with greater flexibility, by allowing them to elect a
           routing-stack different than the default one (quagga). The desired
           routing-stack will be defined in rules/config file.

         - As part of these changes I'm adding support for
           Free-Range-Routing (FRR) stack. Quagga will continue to be
           the default routing-stack.

Signed-off-by: Rodny Molina <rodny@linkedin.com>

PR #8566, opened 2021-08-24
Updating sonic-swss-common submodule

TACappleman:
> Add a new field for FLEX_COUNTER_TABLE to indicate delay for flex counters (#523)
> [logger] Make map access thread safe and proper terminate thread (#510)
> Test with coverage output, publish gcov on Azure pipelines (#517)
> Update schema.h to include config_db DHCP and state_db counter table (#521)
> [schema] Add next hop group table to schema (#475)
> Fix: DBInterface::get() return nullable strings (#516)
> [netlink]refill netlink cache when failing to get the link object by name. (#506)
> Fix test_ConfigDBSubscribe timing risk (#512)
> added missing headers for building with gcc-10.3 (#494)
> Modify the hardcode separator ":" to variable. (#473)

#### Why I did it
Required dependency for proposed sonic-swss PR: https://github.com/Azure/sonic-swss/pull/1702

#### How I did it
Standard submodule update

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106


smaheshm:
@TACappleman needs conflict resolution.

PR #8583, opened 2021-08-25
Configure midplane interface in multi-asic namespaces on a chassis

mlorrillere:
This change is to create an `eth1` interface in multi-asic namespaces on a linecard so that containers can connect to CHASSIS_APP_DB

#### Why I did it
Multi-asic instances running on a linecard can't connect to CHASSIS_APP_DB (running on the supervisor) because they are bridged using default docker bridge and can only access local linecard resources.

#### How I did it
- `platform_env.conf` on a linecard will define `MIDPLANE_SUBNET` which is the subnet used for the internal network
- `MIDPLANE_SUBNET` is chosen by the vendor. The only restriction is that should be a /16 subnet.
- `sonic-platform-modules-*` is to create an `eth1-midplane` interface on the linecard that can connect to the supervisor. The IP assigned to `eth1-midplane` should be part of `MIDPLANE_SUBNET` and in the form `X.Y.<slot id>.1`
- `interfaces-config.sh` will create for each namespace of a multi-asic linecard (defined as `NUM_ASIC > 1` and `MIDPLANE_SUBNET` present) an ipvlan `eth1` interface that is attached to `eth1-midplane`. This interface is assigned an IP in the range defined by `MIDPLANE_SUBNET` (i.e. `X.Y.<slot id>.(10 + namespace id)`)

To allow using a `MIDPLANE_SUBNET` in the `127.X.Y.Z/16` range, we change the loopback address of the namespace `lo` interface to `127.0.0.1/16` (default is `/8`). This step is already done on the host as per [PR5353](https://github.com/Azure/sonic-buildimage/pull/5353).

When `MIDPLANE_SUBNET` is in the `127.X.Y.Z/16` range, we enable routing on local network so that `127/16` can be used.

#### How to verify it

Loopback address of a namespace changed to 127.0.0.1/16:
```console
root@sonic:~# sonic-netns-exec asic0 ip addr show dev lo
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/16 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host
       valid_lft forever preferred_lft forever
```

New `eth1` interface of asic0 namespace on a linecard in slot 6:
```console
root@sonic:~# sonic-netns-exec asic0 ip addr show dev eth1
11: eth1@Bridge: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UNKNOWN group default qlen 1000
    link/ether d4:af:f7:2f:fe:39 brd ff:ff:ff:ff:ff:ff
    inet 127.100.6.10/16 scope host eth1
       valid_lft forever preferred_lft forever
    inet6 fe80::d4af:f700:12f:fe39/64 scope link
       valid_lft forever preferred_lft forever
```

New `eth1` interface of asic1 namespace on a linecard in slot 6:
```console
root@sonic:~# sonic-netns-exec asic1 ip addr show dev eth1
12: eth1@Loopback0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UNKNOWN group default qlen 1000
    link/ether d4:af:f7:2f:fe:39 brd ff:ff:ff:ff:ff:ff
    inet 127.100.6.11/16 scope host eth1
       valid_lft forever preferred_lft forever
    inet6 fe80::d4af:f700:22f:fe39/64 scope link
       valid_lft forever preferred_lft forever
```

Check connectivity from namespace asic0 to CHASSIS_APP_DB:
```console
root@sonic:~# sonic-db-cli -n asic0 CHASSIS_APP_DB PING
True
```

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106

Configure midplane interface in multi-asic namespaces on a chassis


files/image_config/interfaces/interfaces-config.sh
```diff
@@ -34,6 +34,48 @@ for intf_pid in $(ls -1 /var/run/dhclient*.Ethernet*.pid 2> /dev/null); do
...
+       if [ "$?" -ne "0" ]; then # namespace doesn't exist
+          continue
+       fi
+
+       # Use /16 for loopback interface
+       ip netns exec $NET_NS ip addr add 127.0.0.1/16 dev lo
+       ip netns exec $NET_NS ip addr del 127.0.0.1/8 dev lo
+
+       # Create eth1 in database instance
+       ip link add name ns-eth1 link eth1-midplane type ipvlan mode l2
```
judyjoseph, line 60: Ok so here we assume the eth1-midplane interface already exists by the time this script is run. Also this interface is created by a platform specific script/service,

files/image_config/interfaces/interfaces-config.sh
```diff
@@ -34,6 +34,48 @@ for intf_pid in $(ls -1 /var/run/dhclient*.Ethernet*.pid 2> /dev/null); do
...
+
+# Setup eth1 if we connect to a remote chassis DB.
+PLATFORM=${PLATFORM:-`sonic-cfggen -H -v DEVICE_METADATA.localhost.platform`}
+PLATFORM_ENV="/usr/share/sonic/device/$PLATFORM/platform_env.conf"
+[ -f $PLATFORM_ENV ] && source $PLATFORM_ENV
+
+ASIC_CONF="/usr/share/sonic/device/$PLATFORM/asic.conf"
+[ -f $ASIC_CONF ] && source $ASIC_CONF
+
+if [[ -n "$MIDPLANE_SUBNET" && ($NUM_ASIC -gt 1) ]]; then
```
judyjoseph, line 46: The MIDPLANE_SUBNET should match with the "chassisdb_address" in the platform file eg: device/arista/x86_64-arista_7800_sup/chassisdb.conf ?

judyjoseph:
Thanks for the PR. Could you update the PR description with more details like ip route o/p in namespace/host , arp entries learned etc on a linecard as well as supervisor card.

judyjoseph:
@arlakshm @abdosi @minionatwork could you take a look as well

files/image_config/interfaces/interfaces-config.sh
mlorrillere: That's correct. Assumption here is that eth1-midplane was created by a sonic-platform-* service.

files/image_config/interfaces/interfaces-config.sh
mlorrillere: Done.

mlorrillere:
@judyjoseph @minionatwork 

> Thanks for the PR. Could you update the PR description with more details like ip route o/p in namespace/host , arp entries learned etc on a linecard as well as supervisor card.

Description updated with requested output.

sudurais:
Looks good to me.

files/image_config/interfaces/interfaces-config.sh
mlok-nokia: Since platform_env.conf and chassisdb.conf  are in the same format, also the subnet and db_address are related,  I will suggest to consolidate them into one file.  If we don't like the same of the file, we can rename it sometimes in the future.

anamehra: > Since platform_env.conf and chassisdb.conf are in the same format, also the subnet and db_address are related, I will suggest to consolidate them into one file. If we don't like the same of the file, we can rename it sometimes in the future.

I agree. A single generic file should be used for all chassis-specific configurations. There would be more configs in future, like http-config, and I suggest using a more scalable solution like a yaml config file.

mlorrillere: I moved this definition to `chassisdb.conf` and renamed it `midplane_subnet` (small caps) to be consistent with `chassisdb_address`.

anamehra:
Is there a particular reason the midplane interface name has to be 'eth1-midplane'? Could this be platform defined in chassisd.conf and sourced in interfaces-config.sh?

PR #8622, opened 2021-08-28
[Arista] Fix flash size computation for Lodoga

Staphylo:
#### Why I did it

The Lodoga platform also matched crow which was hardcoding the flash
size to 3700. 

#### How I did it

By removing the 3700 hardcoding on crow, this change enables autodetect on Clearlake
which in turns allows autodetect for Lodoga.

The threshold was also bumped from 3700 to 4000 because size computation can
differ slightly and report slightly above 3700.

#### How to verify it

Verify that `docker_inram` is no longer set on `/proc/cmdline`

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012
- [x] 202106

Fix flash size computation in boot0 for Lodoga

### DIFF
--- a/files/Aboot/boot0.j2
+++ b/files/Aboot/boot0.j2
@@ -478,7 +478,6 @@ write_platform_specific_cmdline() {
     if [ "$platform" = "crow" ]; then
         # Assuming sid=Clearlake
         aboot_machine=arista_7050_qx32s
-        flash_size=3700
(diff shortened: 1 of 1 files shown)


yxieca:
I think this file can use some comments? It would be confusing to see a change related to 7050qx addresses issues on 7050cx platform.

yxieca (approved):

PR #1761, opened 2018-06-01
Fix the bcmsh issues.

zhenggen-xu:
-- bcmsh in docker is not executable.
-- bcmsh is not copied to /usr/bin/ at host side

**- What I did**
Fix the issues with bcmsh:
--bcmsh is not executable today in docker, so if you type bcmsh, it won't work.
--bcmsh is not at /usr/bin so you can not type that command directly.

**- How I did it**
Fix above two issues.

**- How to verify it**
bcmsh is working at host and docker.

admin@lnos-x1-a-asw03:~$ bcmsh 
Press Enter to show prompt.
Press Ctrl+C to exit.

drivshell>

**- Description for the changelog**

Fix bcmsh issues (#1761)

**- A picture of a cute animal (not mandatory but encouraged)**

PR #8632, opened 2021-08-30
[yang-models] Removing sonic-acl import from sonic-nat.yang

ghooo:
#### Why I did it
Fixing issue [[sonic-utilities] Unit test failed when building sonic-utilities #1761](https://github.com/Azure/sonic-utilities/issues/1761)

Importing `sonic-acl` caused getting references by `backlinks()` to break, 

#### How I did it
solution is to comment out the importing statement as it is not used anyway.

#### How to verify it
Ran sonic-utilities unit-tests locally after the fix, and all passed.

### DIFF
--- a/src/sonic-yang-models/yang-models/sonic-nat.yang
+++ b/src/sonic-yang-models/yang-models/sonic-nat.yang
@@ -7,9 +7,10 @@ module sonic-nat {
         prefix inet;
     }
 
-    import sonic-acl {
-        prefix sacl;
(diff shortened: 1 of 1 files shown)


qiluo-msft (dismissed):

qiluo-msft:
@venkatmahalingam, Could you help review?

qiluo-msft:
@AkhileshSamineni , Could you help review?

PR #8648, opened 2021-09-01
IMAGE_ASIC info needed from sonic_installer force install

wen587:
#### Why I did it
Prerequisite for sonic_installer force install
#### How I did it
Add IMAGE_ASIC in install.sh
#### How to verify it
Just add a MACRO

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106

### DIFF
--- a/onie-mk-demo.sh
+++ b/onie-mk-demo.sh
@@ -93,7 +93,6 @@ sed -i -e "s/%%DEMO_TYPE%%/$demo_type/g" \
        -e "s/%%ONIE_IMAGE_PART_SIZE%%/$onie_image_part_size/" \
        -e "s/%%EXTRA_CMDLINE_LINUX%%/$EXTRA_CMDLINE_LINUX/" \
        -e "s@%%OUTPUT_RAW_IMAGE%%@$output_raw_image@" \
-       -e "s/%%IMAGE_ASIC%%/$machine/g" \
     $tmp_installdir/install.sh || clean_up 1
 echo -n "."
 cp -r $* $tmp_installdir || clean_up 1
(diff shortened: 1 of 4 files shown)


onie-mk-demo.sh
```diff
@@ -93,6 +93,7 @@ sed -i -e "s/%%DEMO_TYPE%%/$demo_type/g" \
        -e "s/%%ONIE_IMAGE_PART_SIZE%%/$onie_image_part_size/" \
        -e "s/%%EXTRA_CMDLINE_LINUX%%/$EXTRA_CMDLINE_LINUX/" \
        -e "s@%%OUTPUT_RAW_IMAGE%%@$output_raw_image@" \
+       -e "s/%%IMAGE_ASIC%%/$machine/g" \
     $tmp_installdir/install.sh || clean_up 1
```
qiluo-msft, line 97: >install.sh [](http://example.com/codeflow?start=20&length=10)

You don't need to embed the info into this file. $machine is already in the file `machine.conf`, which is in the same directory of `install.sh` during installation.

qiluo-msft:
>     . ./machine.conf

$machine is available after this command line.

---
Refers to: installer/arm64/install.sh:33 in 8f128c1. [](commit_id = 8f128c1829009e4c83c9d16e06bf029d57c28c5d, deletion_comment = False)

onie-mk-demo.sh
wen587: I saw the statement is 
` if [ -r machine.conf] . ./machine.conf `
so we cannot guarantee the machine.conf is always available,right? 
otherwise we can skip the if statement.

qiluo-msft: The Line 104~105 will guarantee the file exists in onie image. The code you mention should be fixed.

PR #8689, opened 2021-09-06
[CBF]  Added configuration templates to generate configs for CBF

Cosmin-Jinga-MS:
[build_templates]: Added default configuration file for CBF
[rules]: Added loading rule for CBF config

#### Why I did it
The CBF default config is required to load default start-up config on CBF capable platforms 
#### How I did it
Added the default config files among the existing ones, mirroring the QOS implementation
Unit tests are in:
https://github.com/Azure/sonic-utilities/pull/1799
#### How to verify it
Deploy SBI on a CBF capable platform(e.g. SODA)

Default CBF config files

Signed-off-by: v-cjinga@microsoft.com


files/build_templates/cbf_config.j2
```diff
@@ -0,0 +1,82 @@
+{
+    "DSCP_TO_TC_MAP": {
```
smaheshm, line 2: DSCP_TO_TC_MAP is already used by qos_config.j2.

If the intention is to use the same mapping, no need to add here.  QoS config generation will create the config for DSCP_TO_TC_MAP. Having in two places will result in race condition where the last config generation will overwrite the DSCP_TO_TC_MAP.

abdosi:
can you please provide what does FC and CBF stands for ?

PR #8707, opened 2021-09-08
[Nokia ixs7215] Miscellaneous platform API fixes

dflynn-Nokia:
This commit delivers the following fixes for the Nokia ixs7215 platform

- Fix bug in a fan API error path
- Add support for setting the fan drawer led
- Add support for getting/setting the front panel PSU status led
- Add support for getting the min/max observed temperature value

#### Why I did it
Fix bugs in various platform APIs
These fixes improve the sonic-mgmt platform API test pass rate

#### How to verify it
Run the following sonic-mgmt test cases
- test-fan.py
- test-fan-drawer.py
- test-thermal.py
- test-psu.py

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012
- [x] 202106

### DIFF
--- a/platform/marvell-armhf/sonic-platform-nokia/7215/sonic_platform/thermal.py
+++ b/platform/marvell-armhf/sonic-platform-nokia/7215/sonic_platform/thermal.py
@@ -33,9 +33,12 @@ class Thermal(ThermalBase):
                     "CPU Core")
 
     def __init__(self, thermal_index):
+        ThermalBase.__init__(self)
         self.index = thermal_index + 1
         self.is_psu_thermal = False
         self.dependency = None
+        self._minimum = None
+        self._maximum = None
         self.thermal_high_threshold_file = None
         # PCB temperature sensors
         if self.index < 3:
@@ -163,6 +166,10 @@ class Thermal(ThermalBase):
             self.thermal_temperature_file)
         if (thermal_temperature != 'ERR'):
             thermal_temperature = float(thermal_temperature) / 1000
+            if self._minimum is None or self._minimum > thermal_temperature:
+                self._minimum = thermal_temperature
+            if self._maximum is None or self._maximum < thermal_temperature:
+                self._maximum = thermal_temperature
         else:
             thermal_temperature = 0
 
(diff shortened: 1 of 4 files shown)


platform/marvell-armhf/sonic-platform-nokia/7215/sonic_platform/thermal.py
```diff
@@ -226,6 +233,16 @@ def get_high_critical_threshold(self):
 
         return float("{:.3f}".format(thermal_high_crit_threshold))
 
+    def get_minimum_recorded(self):
+        if self._minimum is None:
+           self.get_temperature()
```
Blueve, line 238: Do we have any other chance to update the min/max temperature?
Maybe we can always invoke get_temperature and then return the min/max temperature

Blueve (changes requested):

platform/marvell-armhf/sonic-platform-nokia/7215/sonic_platform/thermal.py
dflynn-Nokia: I'm not sure what you are getting at here.
The min and max temp values are set/adjusted within each call to get_temperature()
The logic here is simply to account for the case where get_minimum_recorded() is called before the first call to get_temperature(). By calling get_temperature() here we ensure that the self._minimum variable is set to a temperature value.

Blueve: If no one call `get_temperature` then the min/max value will never have a chance to update.

I checked other implementation and they will call get_temperature and try refresh the min/max during get min/max value. e.g. https://github.com/Azure/sonic-buildimage/blob/e5c2937faec615367a04075ab1a4dc7aeeee1be2/platform/broadcom/sonic-platform-modules-ruijie/b6510-48vs8cq/sonic_platform/thermal.py

dflynn-Nokia: I see your point. By my thinking get_temperature() will be called regularly by the traditional thermalctld daemon used on our platform but I will make the change as you suggest. FWIW, I modeled my approach after an earlier version of Arista platform code. I can see that in their latest version they now use the approach that you suggest.

Blueve (approved):

PR #8723, opened 2021-09-09
[202012] [fstrim] use absolute seconds in time granularity

yxieca:
#### Why I did it
At the initial boot, vs images seems to have trouble with the time delay granularity of 10m, 1h, etc. But seems to be good with absolute second values. With this change, vs image will behave more like physical DUTs.

#### How to verify it
This PR test. Also downloaded the vs image built here and tried manually.


yxieca:
Not the real issue/fix.

PR #8736, opened 2021-09-13
[Dell] S6000 I2C not responding to certain optics

santhosh-kt:
#### Why I did it
- "sfputil show eeprom" and "sfpshow eeprom" commands displays inconsistency results for certain optics.
- For every execution of the commands, various results are displayed.
- Only certain optics EEPROM contents were not retrieved.
#### How I did it
On branch Master
Your branch is up to date with 'origin/Master'.

Changes to be committed:
  (use "git restore --staged <file>..." to unstage)
        modified:   device/dell/x86_64-dell_s6000_s1220-r0/plugins/sfputil.py
        modified:   platform/broadcom/sonic-platform-modules-dell/s6000/modules/dell_s6000_platform.c
        modified:   platform/broadcom/sonic-platform-modules-dell/s6000/scripts/s6000_platform.sh
        modified:   platform/broadcom/sonic-platform-modules-dell/s6000/sonic_platform/sfp.py

#### How to verify it
- Just execute "sfputil show eeprom" and "sfpshow eeprom" commands.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106

### DIFF
--- a/platform/broadcom/sonic-platform-modules-dell/s6000/sonic_platform/sfp.py
+++ b/platform/broadcom/sonic-platform-modules-dell/s6000/sonic_platform/sfp.py
@@ -155,7 +155,7 @@ class Sfp(SfpBase):
         if (self.sfpInfo is None):
             return None
 
-        SFP_LOCK_FILE="/var/run/platform_cache/sfp_lock"
+        SFP_LOCK_FILE="/etc/sonic/sfp_lock"
         try:
             fd = open(SFP_LOCK_FILE, "r")
         except IOError as e:
@@ -434,7 +434,7 @@ class Sfp(SfpBase):
         Retrieves the presence of the sfp
         """
         presence_ctrl = self.sfp_control + 'qsfp_modprs'
-        SFP_LOCK_FILE="/var/run/platform_cache/sfp_lock"
+        SFP_LOCK_FILE="/etc/sonic/sfp_lock"
 
         try:
             fd = open(SFP_LOCK_FILE, "r")
--- a/platform/broadcom/sonic-platform-modules-dell/s6000/sonic_platform/sfp.py
+++ b/platform/broadcom/sonic-platform-modules-dell/s6000/sonic_platform/sfp.py
@@ -12,6 +12,7 @@ try:
     import re
     import struct
     import time
+    import fcntl
     from sonic_platform_base.sfp_base import SfpBase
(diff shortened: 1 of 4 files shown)


platform/broadcom/sonic-platform-modules-dell/s6000/sonic_platform/sfp.py
```diff
@@ -154,6 +155,15 @@ def _get_eeprom_data(self, eeprom_key):
         if (self.sfpInfo is None):
             return None
 
+        SFP_LOCK_FILE="/etc/sonic/sfp_lock"
```
prgeor, line 158: Please consider using python native Lock [https://docs.python.org/3/library/multiprocessing.html#synchronization-between-processes](https://docs.python.org/3/library/multiprocessing.html#synchronization-between-processes). This is more exception friendly and easier to debug deadlock scenarios. Moreover the file lock is originally designed for mutual exclusion for file access.

platform/broadcom/sonic-platform-modules-dell/s6000/sonic_platform/sfp.py
```diff
@@ -437,12 +458,69 @@ def get_presence(self):
...
+        mask = (1 << index)
+
+        if ((reg_value & mask) == 1):
+            modsel_state = False
+        else:
+            modsel_state = True
+
+        return modsel_state
+
+    def set_modsel(self):
```
prgeor, line 494: If you bring the lock inside set_modsel(), then you don't have to lock/unlock every place where set_modsel() is called, keeps code changes to minimum.

prgeor (changes requested):
@santhosh-kt please address comment

platform/broadcom/sonic-platform-modules-dell/s6000/sonic_platform/sfp.py
santhosh-kt: The lock is based on set_modsel() but we cant introduce it here.
A scenario:
1) When sfputil tries to read a port, it checks for modprs register and to read that we are setting modsel register.
2) Before we actually reading modprs register, if pmon docker or any other process trying to set modsel register then it will have the race condition here.
3) To avoid this we are locking before modsel register is being set and releasing the lock only after the requirements are over(In our case after retriving modprs register).

PR #8746, opened 2021-09-14
Remove the rw folder from the image after installing in KVM

saiarcot895:
#### Why I did it

When the image is installed from within KVM and then loaded, some files
(such as timer stamp files) are created as part of that bootup that then
get into the final image. This can cause some side effects, such as
systemd thinking that some persistent timers need to run because the
last trigger time got missed.

Therefore, at the end of the check_install.py script, remove the rw
folder so that it doesn't exist in the image, and that when this image
is started up in a KVM setup for the first time, it starts with a truly
clean slate.

Without this change, the issue seen was that for fstrim.timer, a stamp
file would be present in /var/lib/systemd/timers (and for other timers
that are marked as persistent). This would then cause fstrim.service to
get started immediately when starting a QEMU setup if the timer for that
service missed a trigger, and not wait 10 minutes after bootup. In the
case of fstrim.timer, that means if the image was started in QEMU after
next Monday, since that timer is scheduled to be triggered weekly.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106

### DIFF
--- a/check_install.py
+++ b/check_install.py
@@ -37,22 +37,19 @@ def main():
                 raise
             time.sleep(1)
 
-    # select ONIE embed
+    # select default SONiC Image
     p.expect(grub_selection)
-    p.sendline(KEY_DOWN)
+    p.sendline()
 
-    # install sonic image
+    # bootup sonic image
     while True:
-        i = p.expect([login_prompt, passwd_prompt, grub_selection, cmd_prompt])
+        i = p.expect([login_prompt, passwd_prompt, cmd_prompt])
         if i == 0:
             # send user name
             p.sendline(args.u)
         elif i == 1:
             # send password
             p.sendline(args.P)
-        elif i == 2:
-            # select onie install
-            p.sendline()
         else:
             break
 
--- a/check_install.py
+++ b/check_install.py
@@ -15,11 +15,6 @@ def main():
 
     args = parser.parse_args()
 
-    KEY_UP = '\x1b[A'
-    KEY_DOWN = '\x1b[B'
-    KEY_RIGHT = '\x1b[C'
-    KEY_LEFT = '\x1b[D'
-
     login_prompt = 'sonic login:'
     passwd_prompt = 'Password:'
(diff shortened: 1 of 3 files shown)


yxieca:
test is failing, did this remove being too aggresive?

saiarcot895:
Yes, the `firstboot` file in the `/host/` directory was no longer present, and so none of the first boot steps ran. One of these steps is to create the `/var/platform` directory, which is then bind-mounted into the mgmt-framework container.

I'm working on splitting up the `check_install.py` script into two separate scripts. The first script will install the image, and the second script will boot into the image and verify that things are working. The KVM instance for the second script will use the `-snapshot` flag, so that any changes made by that KVM instance aren't actually written to disk.

check_install.py
```diff
@@ -64,6 +61,8 @@ def main():
     p.expect([cmd_prompt])
     p.sendline('show ip bgp sum')
     p.expect([cmd_prompt])
+    p.sendline('sudo rm -r /host/image-*/rw/')
```
qiluo-msft, line 64: If you think there is garbage in this folder, could you delete during building the image. Is it a issue on non-KVM hardware?

saiarcot895: Actually, that command isn't needed anymore with the changes in my second commit. Let me remove that.

The files in that folder comes when kvm boots up the SONiC image to verify basic functionality is working, not when the image is built. One type of file that's written into that folder under `/var/lib/systemd/timers` is timer stamp files, saying when was the last time a timer was started or was fired. The presence of these files affects the fstrim and logrotate timers when the image is actually deployed into a KVM/QEMU environment. It'll cause systemd to immediately fire the timers when they are started instead of waiting till whenever the next "event" is supposed to be, because systemd will think that based on the stamp files, these timers should have already ran.

This is not an issue on non-KVM images, because they are not booted up during image build; their first bootup would be on the hardware.

yxieca (approved):

yxieca (approved):

PR #8761, opened 2021-09-15
Routed subinterface enhancements

preetham-singh:
#### Why I did it
Routed subinterfae enhancements HLD [#833](https://github.com/Azure/SONiC/pull/833)
Adding python API support to get routed subinterface long name to get correct parent interface for the routed subinterface.

#### How I did it
Adding python API.

#### How to verify it
Configuring routed subinterface on Physical & port channel subinterfaces.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106


zhangyanzhao:
HLD is not merged yet, we should merge the HLD before merging this code PR.

src/sonic-py-common/sonic_py_common/interface.py
```diff
@@ -88,5 +105,30 @@ def get_port_table_name(interface_name):
...
+        if interface_name.startswith(physical_subinterface_prefix()) or interface_name.startswith(portchannel_subinterface_prefix()):
+            return "VLAN_SUB_INTERFACE"
     else:
         return ""
+
+def intf_get_longname(intf):
+    if intf is None:
+       return None
+
+    if VLAN_SUB_INTERFACE_SEPARATOR in intf:
```
venkatmahalingam, line 118: Do we need to check Eth or Po prefix as well instead of blindly checking for "." in the intf string and returning long interface name?

preetham-singh: In certain case we need to fetch full name for the given subinterface parent interface which could be in short name format. Just checking for "." will not suffice here.

venkatmahalingam (dismissed):

prsunny:
Please add some tests here - https://github.com/Azure/sonic-buildimage/blob/master/src/sonic-py-common/tests/interface_test.py

src/sonic-py-common/sonic_py_common/interface.py
```diff
@@ -88,5 +105,47 @@ def get_port_table_name(interface_name):
...
+    if intf.startswith("Eth"):
+        intf_index=intf[len("Eth"):sub_intf_sep_idx]
+        return "Ethernet"+intf_index+VLAN_SUB_INTERFACE_SEPARATOR+sub_intf_idx
+    elif intf.startswith("Po"):
+        intf_index=intf[len("Po"):sub_intf_sep_idx]
+        return "PortChannel"+intf_index+VLAN_SUB_INTERFACE_SEPARATOR+sub_intf_idx
+    else:
+        return str(intf)
+
+def intf_get_longname(intf):
```
prsunny, line 131: please follow a similar naming convention in the file.. suggest rename to `get_subintf_name`

prsunny (dismissed):

src/sonic-py-common/sonic_py_common/interface.py
preetham-singh: Since this is being used to get longname for both subinterface and parent of subinterface, I have updated to use get_intf_longname as part of next commit.

prsunny:
@preetham-singh , could you please rebase?

prsunny:
looks like incorrect rebase. Could you please fix?

prsunny (approved):

PR #8768, opened 2021-09-15
[Reclaim buffer] Reclaim unused buffers by applying zero buffer profiles

stephenxs:
#### Why I did it
Support zero buffer profiles
1. Add buffer profiles and pool definition for zero buffer profiles
2. Support applying zero profiles on `INACTIVE PORTS`
3. Enable `dynamic buffer manager` to load zero pools and profiles from a JSON file

Dependency: It depends on Azure/sonic-swss#1910 and submodule advancing PR once the former merged.

#### How I did it
1. Add buffer profiles and pool definition for zero buffer profiles
   - If the buffer model is static:
      - Apply normal buffer profiles to admin-up ports
      - Apply zero buffer profiles to admin-down ports
   - If the buffer model is dynamic:
      - Apply normal buffer profiles to all ports
      - buffer manager will take care when a port is shut down
2. Update `buffers_config.j2` to support `INACTIVE PORTS` by extending the existing macros to generate the various buffer objects, including PGs, queues, ingress/egress profile lists
   - Originally, all the macros to generate the above buffer objects took active ports only as an argument
   - Now that buffer items need to be generated on inactive ports as well, an extra argument representing the inactive ports need to be added
   - To be backward compatible, a new series of macros are introduced to take both active and inactive ports as arguments
   - The original version (with active ports only) will be checked first. If it is not defined, then the extended version will be called
   - Only vendors who support zero profiles need to change their buffer templates
3. Enable `buffer manager` to load zero pools and profiles from a JSON file:
   - The JSON file is provided on a per-platform basis
   - It is copied from `platform/<vendor>` folder to `/usr/share/sonic/temlates` folder in compiling time and rendered when the swss container is being created.
4. To make code clean and reduce redundant code, extract common macros from buffer_defaults_t{0,1}.j2 of all SKUs to two common files:
   - One in Mellanox-SN2700-D48C8 for single ingress pool mode
   - The other in ACS-MSN2700 for double ingress pool mode

   Those files of all other SKUs will be symbol link to the above files
5. Update `sonic-cfggen` test accordingly:
   - Adjust example output file of JSON template for unit test
   - Add unit test in for Mellanox's new buffer templates.

#### How to verify it
1. Regression test.
2. Unit test in sonic-cfggen

Run regression test and manually test.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012
- [x] 202106


files/build_templates/buffers_config.j2
```diff
@@ -21,6 +21,30 @@ def
 {%-     set switch_role      = '' %}
 {%- endif -%}
 
+{%- set PORT_ALL  = [] %}
+
+{%- if PORT is not defined %}
```
stephenxs, line 26: check whether "zero port" will affect this logic

files/build_templates/buffers_config.j2
```diff
@@ -21,6 +21,30 @@ def
...
+    {%- if defs.generate_port_lists(PORT_ALL) %} {% endif %}
```
neethajohn, line 27: we no longer have this macro. will this cause issues if device is brought up with empty ports?

stephenxs: Fixed by testing whether it is defined ahead of calling it.

files/build_templates/buffers_config.j2
stephenxs: Updated code to tolerance zero port scenario.

files/build_templates/buffers_config.j2
neethajohn: Can we check if a testcase can be added for rendering template in zero port case?

stephenxs: I'm afraid we can't for now.
The j2 rendering test is built on top of minigraph loading logic which assumes there always be at least one port existing in the minigraph.

keboliu (approved):

src/sonic-config-engine/tests/sample_output/py3/buffers-mellanox2410.json
```diff
@@ -0,0 +1,1163 @@
+
+{
+    "CABLE_LENGTH": {
```
neethajohn, line 3: What is the diff between this one and buffers-mellanox2700.json?

device/mellanox/x86_64-mlnx_msn3800-r0/Mellanox-SN3800-D100C12S2/buffers_defaults_t0.j2
```diff
@@ -4,97 +4,20 @@
 {% set egress_lossless_pool_size =  '34287552' %}
 {% set egress_lossy_pool_size =  '20664320' %}
 
-{%- macro generate_port_lists(PORT_ALL) %}
-    {# Generate list of ports #}
-    {%- for port_idx in range(0, 32) %}
-        {%- if PORT_ALL.append("Ethernet%d" % (port_idx)) %}{%- endif %}
-    {%- endfor %}
-{%- endmacro %}
+{% import 'buffers_defaults_objects.j2' as defs with context %}
```
neethajohn, line 7: Missing buffers_defaults_objects.j2 for this sku?

neethajohn (commented):
@stephenxs, this PR still contains the code from #9133

stephenxs:
> @stephenxs, this PR still contains the code from #9133

Yes for now. Will remove them after that one is merged otherwise there is dependencies

src/sonic-config-engine/tests/sample_output/py3/buffers-mellanox2410.json
stephenxs: 2700 should be single ingress pool. This one should be double ingress pools

device/mellanox/x86_64-mlnx_msn3800-r0/Mellanox-SN3800-D100C12S2/buffers_defaults_t0.j2
stephenxs: Fixed.

neethajohn (dismissed):

neethajohn (approved):

stephenxs:
It depends on Azure/sonic-swss#1910 and submodule advancing PR once the former merged.

PR #8781, opened 2021-09-16
[docker-sonic-vs] Use database_config.json from package libswsscommon

jimmyzhai:
#### Why I did it
The file platform/vs/docker-sonic-vs/database_config.json can be easily out of sync against standard schema file database_config.json from package libswsscommon. This can cause tricky issues while accessing redis DB.
#### How I did it
1/ remove  file platform/vs/docker-sonic-vs/database_config.json
2/ copy database_config.json of package libswsscommon to /etc/default/sonic-db/

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106


platform/vs/docker-sonic-vs/Dockerfile.j2
```diff
@@ -180,6 +179,9 @@ COPY ["hwsku.json", "/usr/share/sonic/device/x86_64-kvm_x86_64-r0/brcm_gearbox_v
 # Workaround the tcpdump issue
 RUN mv /usr/sbin/tcpdump /usr/bin/tcpdump
 
+# Backup database_config.json from package libswsscommon, restore it in start.sh
+RUN cp /var/run/redis/sonic-db/database_config.json /etc/default/sonic-db/
```
lguohan, line 183: let's not copy host version in docker-sonic-vs, it could be different

lguohan: let's modify the database_config.json

jimmyzhai: Why could it be different?

lguohan: they are different systems, it could be different.

PR #8793, opened 2021-09-18
Added HTTP server support on chassis system

anamehra:
Signed-off-by: Anand Mehra <anamehra@cisco.com>

#### Why I did it

This commit introduces HTTP server for Supervisor Card. HTTP will be used by Line Cards to download Sonic image
file from Supervisor card. This may also be used to download configuration files or any other data file from Supervisor.

#### How I did it

HTTP server is enabled by a systemd service, http-server.service and runs on Supervisor only when
/etc/sonic/http_config.yaml file is present.

On Supervisor Card, the platform shall provide HTTP configuration via devices/chassis.yaml file.
Following HTTP configuration may be provided by chassis.yaml:

http-config:
    http_server_dir: <http server home dir path>
    http_server_ip: <ip>
    http_server_port: <port>
    start_http_server: 'yes'

On Line Card, chassis.yaml shall only contain:
http-config:
    http_server_ip: 127.1.1.1

http_server_dir:         Optional; HTTP home directory, default: /var/www
http_server_ip:        Required; IP to map http-server hostname used by HTTP Server.
                                  This should be the internal IP address used by the internal chassis network.
http_server_port:    Optional; Port to bind, default: 8000.
start_http_server:   Optional; If 'yes' or '1', start HTTP server on the card.

chassis.yaml file is parsed by a new service, config-chassis.service. This service parses the http-config
from chassis.yaml to set up pre-requisites for http-server.service. config-chassis.service is made generic
to be used for other chassis-based configurations in the future.

config-chassis.service does the following two operations for HTTP server based on http-config block:

If start_http_server is 'yes' or '1', it creates /etc/sonic/http_config.yaml file with the HTTP configuration
provided.
Create the http_server_ip to http-server hostname mapping in /etc/hosts. Line Cards may also provide
http-config with this entry only to create hostname mapping during boot.

#### How to verify it
create devices/<platform>/chassis.yaml file with following details:
http-config:
    http_server_dir: <http server home dir path>
    http_server_ip: <ip>
    http_server_port: <port>
    start_http_server: 'yes'

During boot config-chassis.servcei should parse this file and create /etc/hosts entry for 'http-server' hostname
and /etc/sonic/http_config.yaml

http-server.service should start and create http server.

Place a file at the http_server_dir path and use curl to download:
curl http://http-server:<http_server_port>/<file> -o file

validate the file.

[UT.txt](https://github.com/Azure/sonic-buildimage/files/7188900/UT.txt)

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106


files/build_templates/config-chassis.service.j2
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Config chassis service
+After=rc-local.service
+Requires=rc-local.service
+Before=config-chassisd.service
```
anamehra, line 5: Fix: config-chassisdb.service

rlhui:
@mlok-nokia - thanks for your review on this.

files/image_config/config-chassis/config-chassis.py
```diff
@@ -0,0 +1,114 @@
...
+        proc = subprocess.Popen([SONIC_CFGGEN_PATH, '-H', '-v', PLATFORM_KEY],
+                            stdout=subprocess.PIPE,
+                            shell=False,
+                            stderr=subprocess.STDOUT)
+        outp = proc.communicate()[0]
+        proc.wait()
+        self.platform = outp.decode().rstrip('\n')
+
+    def config_http(self):
+        ''' Setup HTTP server configiguration.
```
anamehra, line 72: configiguration -> configuration

rlhui:
@ysmanman, @staphylo - would you please help review this? Thanks.

files/image_config/config-chassis/config-chassis.py
```diff
@@ -0,0 +1,114 @@
...
+class ChassisConfig():
+    def __init__(self, args, **kw):
+        self.chassis_conf = None
+        self.http_config = None
+        if args.config_file:
+            chassis_yaml = args.config_file
+        else:
+            # get chassis.yaml path based on platform type
+            self.get_platform()
+            chassis_yaml=os.path.join('/usr/share/sonic/device', self.platform, 'chassis.yaml')
```
arlakshm, line 47: can you commit a sample `chassis.yaml` for the VS platform for reference?

files/image_config/config-chassis/config-chassis.py
```diff
@@ -0,0 +1,114 @@
...
+
+            # Extract http config
+            self.http_config = self.chassis_conf.get('http_config')
+        else:
+            logger.info('chassis config file {} not present - skip config'.format(chassis_yaml))
+
+    def get_platform(self):
+        ''' Get the platform type from DEVICE_METADATA '''
+        SONIC_CFGGEN_PATH = '/usr/local/bin/sonic-cfggen'
+        PLATFORM_KEY = 'DEVICE_METADATA.localhost.platform'
```
arlakshm, line 61: On a system before the config\minigraph is loaded, `'DEVICE_METADATA.localhost.platform'` may not be available

files/image_config/http-server/http-server.py
```diff
@@ -0,0 +1,85 @@
...
+            # Get Server port from config
+            self.http_server_port = self.http_config.get('http_server_port', HTTP_DEFAULT_BIND_PORT)
+
+            # Create HTTP home dir path if not present
+            path = Path(http_dir)
+            path.mkdir(parents=True, exist_ok=True)
+
+    def run(self):
+        if self.http_config:
+            with socketserver.TCPServer(('http-server', self.http_server_port), HTTPRequestHandler) as httpd:
```
arlakshm, line 70: what will be the ip address on which the http server is running? localhost?

arlakshm (commented):
Can you add some sonic-mgmt tests to test this ?

lguohan:
can we put such service into container? I do not think we want to have such services in the base image unless it is absolutely necessary.

anamehra:
> can we put such service into container? I do not think we want to have such services in the base image unless it is absolutely necessary.

Hi @lguohan , we had an initial proposal to introduce docker for http service but in the chassis community group meeting it was decided to keep it in host with a lightweight python3 http server. Also, this PR introduces a configuration that could be used to enable or disable the service from a platform config file in case the service is not required.

files/image_config/config-chassis/config-chassis.py
anamehra: sonic-cfggen call get_platform which fetch the platform from machine.conf. -H flag helps to construct DEVICE_METADATA info. So we should be good. database.sh use same command to get platform info which is before config is loaded.

files/image_config/http-server/http-server.py
anamehra: The IP address comes from the http_config provided in chassis.conf

anamehra:
> Can you add some sonic-mgmt tests to test this ?
Hi @arlakshm ,
May I add this as a separate PR later? I updated this PR for your other comments.

files/image_config/config-chassis/config-chassis.py
```diff
@@ -0,0 +1,114 @@
+#!/usr/bin/env python3
+# 
+###########################################################################
```
judyjoseph, line 3: Revisiting this PR again, have a question - if we really need this additional service config-chassis.service to configure the chassis? Seems like we are binding starting of http service with chassis.  

Can we make it as a generic http service and have control variables in for eg: in platform_env.conf to start stop service?

judyjoseph:
Adding comments here from the discussion earlier. There were a few changes suggested. 

1.	Add http_server to FEATURE table, with default state as disabled and auto_restart enabled by default. (Another suggestion was to keep this FEATURE enabled by default, but then to make sure in LC and other pizza boxes it should be a noop) 
2.	http_server to be added as part of host_services package : sonic-buildimage/tree/master/src/sonic-host-services/scripts
so that it will be part of the sonic_host_services-1.0-py3-none-any.whl
3.	The http configuration initialization done config-chassis.service will be moved to http-server.service itself.
4.	The startup http configuration can be kept in platform_env.conf file. Another proposal was to replace the existing chassisdb.conf  yaml formal (eg:  chassis.yaml ). This could be taken up later.

anamehra:
Hi @judyjoseph , @abdosi , I have created new PR with the changes discussed
https://github.com/Azure/sonic-buildimage/pull/10120

I am closing this PR to track changes via new PR.

PR #8812, opened 2021-09-22
Load global config in caclmgrd only in multi asic NPU

dgsudharsan:
#### Why I did it
The database_global.json file is generated only in multi asic scenario. However, caclmgrd calls initialization based on this file without multi npu check.
This leads to following error in single asic scenario
Sep 21 22:35:20.996798 r-leopard-32 ERR caclmgrd[11819]: :- initializeGlobalConfig: Sonic database config global file doesn't exist at /var/run/redis/sonic-db/database_global.json

#### How I did it
Added if multi npu check before invoking the load global config.

#### How to verify it
Restart caclmgrd after this change and check if no error log is thrown.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012
- [x] 202106

### DIFF
--- a/src/sonic-host-services/scripts/caclmgrd
+++ b/src/sonic-host-services/scripts/caclmgrd
@@ -101,7 +101,9 @@ class ControlPlaneAclManager(daemon_base.DaemonBase):
         self.lock[DEFAULT_NAMESPACE] = threading.Lock()
         self.num_changes[DEFAULT_NAMESPACE] = 0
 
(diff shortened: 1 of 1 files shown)


dgsudharsan:
@arlakshm @judyjoseph  Please review

arlakshm (approved):
Lgtm

dgsudharsan:
@arlakshm @judyjoseph @lguohan  Can we merge this if changes are good?

PR #8814, opened 2021-09-22
add onie image install's platform verification

wen587:
#### Why I did it
Provide onie image install verification
#### How I did it
Add devices folder based on platform_asic and check if current platform inside the list
#### How to verify it
In onie env, install a bin file which differs the running platform's ASIC. For example;
install sonic-broadcom.bin to mellanox ASIC platform
install sonic-vs.bin to broadcom ASIC platform

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106

### DIFF
--- a/build_image.sh
+++ b/build_image.sh
@@ -87,11 +87,24 @@ generate_onie_installer_image()
           $ONIE_INSTALLER_PAYLOAD
 }
 
+generate_onie_device_list()
+{
+    # Generate asic-specific ONIE device list
+    rm -rf ./installer/x86_64/devices/
+    mkdir -p ./installer/x86_64/devices/
+    for d in `find -L ./device  -maxdepth 2 -mindepth 2 -type d`; do
+            if [ -f $d/platform_asic ] && grep -Fxq "$CONFIGURED_PLATFORM" $d/platform_asic; then
+                touch ./installer/x86_64/devices/${d##*/};
+            fi;
+    done
+}
+
 if [ "$IMAGE_TYPE" = "onie" ]; then
     echo "Build ONIE installer"
     mkdir -p `dirname $OUTPUT_ONIE_IMAGE`
     sudo rm -f $OUTPUT_ONIE_IMAGE
 
+    generate_onie_device_list
     generate_onie_installer_image
 
 ## Build a raw partition dump image using the ONIE installer that can be
--- a/installer/x86_64/install.sh
+++ b/installer/x86_64/install.sh
@@ -92,6 +92,39 @@ VAR_LOG_SIZE=4096
 
 [ -r platforms/$onie_platform ] && . platforms/$onie_platform
 
+# Verify image platform is inside devices list
+# Temporarily set TIMEOUT as 10 seconds
+if [ "$install_env" = "onie" ] && [ -d devices ]; then
+    if [ ! -f devices/$onie_platform ]; then
+        TIMEOUT=10
+        echo "The image you're trying to install is of a different platform as the running plaform"
+        while true; do
+            read -t $TIMEOUT -r -p "Do you still wish to install this image?[Y/n]" input
+            STATUS=$?
+            if test $STATUS -eq 0; then
+                case $input in
+                    [Yy])
+                        echo "Force installing..."
+                        break
+                        ;;
+                    [Nn])
+                        echo "Exited installation!"
+                        exit 1
+                        ;;
+                    *)
+                        echo "Error: Invalid input"
+                        ;;
+                esac
+            else
+                echo
+                echo "Timeout! Installation cancelled."
+                exit 1
+            fi
+        done
+
+    fi
+fi
+
 # Pick up console port and speed from install enviroment if not defined yet.
 # Console port and speed setting in cmdline is like "console=ttyS0,9600n",
 # so we can use pattern 'console=ttyS[0-9]+,[0-9]+' to match it.


wen587:
Based on https://github.com/Azure/sonic-buildimage/pull/8542.
Wait for above PR merge first.

build_image.sh
```diff
@@ -87,11 +87,24 @@ generate_onie_installer_image()
           $ONIE_INSTALLER_PAYLOAD
 }
 
+generate_onie_device_list()
+{
+    # Generate asic-specific ONIE device list
+    rm -rf ./installer/x86_64/devices/
+    mkdir -p ./installer/x86_64/devices/
+    for d in `find -L ./device  -maxdepth 2 -mindepth 2 -type d`; do
+            if [ -f $d/platform_asic ] && grep -Fxq "$CONFIGURED_PLATFORM" $d/platform_asic; then
```
qiluo-msft, line 96: >Fxq [](http://example.com/codeflow?start=48&length=3)

Will the pattern `broadcom` match the file content `broadcom-dnx`? I guess no, just double check.

qiluo-msft, line 96: >$d [](http://example.com/codeflow?start=20&length=2)

Do you need `$$d`?

wen587: It will not match broadcom-dnx.

wen587: No. It is shell script. Seems we only use $$d in makefile.

build_image.sh
```diff
@@ -87,11 +87,24 @@ generate_onie_installer_image()
...
+                touch ./installer/x86_64/devices/${d##*/};
```
qiluo-msft, line 97: >touch ./installer/x86_64/devices/${d##*/}; [](http://example.com/codeflow?start=16&length=42)

Creat one file, which contains the list of onie platform string?

build_image.sh
qiluo-msft, line 96: >             [](http://example.com/codeflow?start=0&length=12)

Too much indentation?

installer/x86_64/install.sh
```diff
@@ -92,6 +92,39 @@ VAR_LOG_SIZE=4096
 
 [ -r platforms/$onie_platform ] && . platforms/$onie_platform
 
+# Verify image platform is inside devices list
+# Temporarily set TIMEOUT as 10 seconds
+if [ "$install_env" = "onie" ] && [ -d devices ]; then
```
qiluo-msft, line 97: >onie [](http://example.com/codeflow?start=23&length=4)

Is it possible to run on all the envrionment besides ONIE?

build_image.sh
```diff
@@ -87,11 +87,24 @@ generate_onie_installer_image()
...
+            fi;
+    done
+}
+
 if [ "$IMAGE_TYPE" = "onie" ]; then
```
qiluo-msft, line 102: >onie [](http://example.com/codeflow?start=22&length=4)

How about other `IMAGE_TYPE`? such as Aboot?

installer/x86_64/install.sh
qiluo-msft, line 97: >-d devices [](http://example.com/codeflow?start=36&length=10)

This should not happen according to your code. If you want to handle this case, it should be an immediate failure.

installer/x86_64/install.sh
```diff
@@ -92,6 +92,39 @@ VAR_LOG_SIZE=4096
...
+    if [ ! -f devices/$onie_platform ]; then
+        TIMEOUT=10
+        echo "The image you're trying to install is of a different platform as the running plaform"
+        while true; do
+            read -t $TIMEOUT -r -p "Do you still wish to install this image?[Y/n]" input
```
qiluo-msft, line 102: >[ [](http://example.com/codeflow?start=76&length=1)

Add a blank char in between?

qiluo-msft, line 102: >Y [](http://example.com/codeflow?start=77&length=1)

If you use uppercase `Y`, it normally means the default when user input 'Enter' directly.

installer/x86_64/install.sh
```diff
@@ -92,6 +92,39 @@ VAR_LOG_SIZE=4096
...
+                        ;;
+                esac
+            else
+                echo
+                echo "Timeout! Installation cancelled."
+                exit 1
+            fi
+        done
+
+    fi
```
qiluo-msft, line 125: Remove extra empty line

build_image.sh
wen587: clean up soon

installer/x86_64/install.sh
wen587: That's true. I was thinking the wrong way.

installer/x86_64/install.sh
wen587: will change to [y/n] later

installer/x86_64/install.sh
wen587: done

build_image.sh
wen587: Also add Aboot support

build_image.sh
```diff
@@ -87,11 +87,24 @@ generate_onie_installer_image()
           $ONIE_INSTALLER_PAYLOAD
 }
 
+generate_onie_device_list()
+{
+    # Generate asic-specific ONIE device list
+    rm -rf ./installer/x86_64/devices/
+    mkdir -p ./installer/x86_64/devices/
```
qiluo-msft, line 94: You just need to create an empty file in this folder. No need to remove directory.

installer/x86_64/install.sh
qiluo-msft: It could be an option among sonic/onie/build

qiluo-msft:
> #!/bin/bash

I notice there are many CPU architectures now besides x86_64. Check `installer/*/install.sh`.

---
Refers to: build_image.sh:1 in e53fc70. [](commit_id = e53fc70400ff0f5f777629480508fa93ec10d974, deletion_comment = False)

qiluo-msft:
> #!/bin/sh

I notice there are many CPU architectures now besides x86_64. Check `installer/*/install.sh`.

---
Refers to: installer/x86_64/install.sh:1 in e53fc70. [](commit_id = e53fc70400ff0f5f777629480508fa93ec10d974, deletion_comment = False)

installer/x86_64/install.sh
wen587: We have sonic_installer to block wrong platform already. It might be tedious if user input --skip-platform-check and be prompted here the second time to confirm install.

build_image.sh
wen587: [generate_onie_installer_image](https://github.com/Azure/sonic-buildimage/blob/master/build_image.sh#L72)
I am following the above code to refresh the folder every time. And the folder needs to be created the first time.

installer/x86_64/install.sh
qiluo-msft: Reasonable. Could you change sonic-utilities code with same logic, checking platform_asic instead of previous image asic.

qiluo-msft: This comment could be closed.

installer/x86_64/install.sh
```diff
@@ -92,6 +92,38 @@ VAR_LOG_SIZE=4096
 
 [ -r platforms/$onie_platform ] && . platforms/$onie_platform
 
+# Verify image platform is inside devices list
+# Temporarily set INPUT_TIMEOUT as 10 seconds
+if [ "$install_env" = "onie" ] && [ "$machine" != "generic" ]; then
+    if ! grep -Fxq "$onie_platform" devices/platforms_asic; then
+        INPUT_TIMEOUT=10
+        echo "The image you're trying to install is of a different platform as the running platform"
```
qiluo-msft, line 100: >platform [](http://example.com/codeflow?start=67&length=8)

asic type

qiluo-msft, line 100: >platform [](http://example.com/codeflow?start=91&length=8)

asic

installer/x86_64/install.sh
```diff
@@ -92,6 +92,38 @@ VAR_LOG_SIZE=4096
...
+        while true; do
+            read -t $INPUT_TIMEOUT -r -p "Do you still wish to install this image? [y/n]: " input
```
qiluo-msft, line 102: >$INPUT_TIMEOUT [](http://example.com/codeflow?start=20&length=14)

Actually there is no requirement to timeout. If the script is used by automation, the prompt will be enough to break the normal path (The automation will have its own timeout).

If the script is used by user, it is too rush for them to response in 10 seconds, and without any hint.

build_image.sh
```diff
@@ -170,6 +183,13 @@ elif [ "$IMAGE_TYPE" = "aboot" ]; then
     zip -g $ABOOT_BOOT_IMAGE version
     rm version
 
+    generate_onie_device_list
+    if [ -f ./installer/x86_64/devices/platforms_asic ]; then
+        cp ./installer/x86_64/devices/platforms_asic .platforms_asic
+        zip -g $OUTPUT_ABOOT_IMAGE .platforms_asic
+        rm .platforms_asic
+    fi
```
qiluo-msft, line 191: Implement a new function `generate_aboot_device_list` ? If some logic is common, you can extract into another function like  `generate_device_list`.

installer/x86_64/install.sh
wen587: will change

wen587: will change

installer/x86_64/install.sh
wen587: will remove it

build_image.sh
wen587: I think it should be fine. 
The extra code here is just for the case that if aboot has a generic build. We should simply bypass the cp process.

build_image.sh
```diff
@@ -87,11 +87,24 @@ generate_onie_installer_image()
...
 }
 
+generate_device_list()
+{
+    # Generate asic-specific ONIE device list
+    rm -rf ./installer/x86_64/devices/
+    mkdir -p ./installer/x86_64/devices/
+    for d in `find -L ./device  -maxdepth 2 -mindepth 2 -type d`; do
+        if [ -f $d/platform_asic ] && grep -Fxq "$CONFIGURED_PLATFORM" $d/platform_asic; then
+            echo "${d##*/}" >> ./installer/x86_64/devices/platforms_asic;
```
qiluo-msft, line 97: >./installer/x86_64/devices/platforms_asic [](http://example.com/codeflow?start=31&length=41)

This filename could be replace by a function parameter, and onie/aboot will call the function with different argument.

Then you don't need the extra step `cp ./installer/x86_64/devices/platforms_asic .platforms_asic`

wen587:
Ready to be merged.

build_image.sh
```diff
@@ -170,6 +184,13 @@ elif [ "$IMAGE_TYPE" = "aboot" ]; then
     zip -g $ABOOT_BOOT_IMAGE version
     rm version
 
+    generate_device_list ".platforms_asic"
+    # Check file existence in case of general build
```
qiluo-msft, line 188: >Check file existence in case of general build [](http://example.com/codeflow?start=6&length=45)

What do you mean by "general build"? Is this file always existing after previous generating function?

build_image.sh
```diff
@@ -87,11 +87,25 @@ generate_onie_installer_image()
           $ONIE_INSTALLER_PAYLOAD
 }
 
+generate_device_list()
+{
+    # Generate asic-specific device list
+    local platforms_asic=$1
+    for d in `find -L ./device  -maxdepth 2 -mindepth 2 -type d`; do
+        if [ -f $d/platform_asic ] && grep -Fxq "$CONFIGURED_PLATFORM" $d/platform_asic; then
+            echo "${d##*/}" >> "$platforms_asic";
```
qiluo-msft, line 96: >echo "${d##*/}" >> "$platforms_asic"; [](http://example.com/codeflow?start=12&length=37)

You need to clean the file at the begining in case the file is not empty. No need to `rm` before this function call.
If there is no platforms using CONFIGURED_PLATFORM asic, this function should generated an empty file.

wen587: In case of general build, sonic installer verification would bypass because $platforms_asic file is not there. But if it is a empty file, it will fail because no platform is in the file. I guess general build means we could install it to any asic.
CONFIGURED_PLATFORM is general by default.
Check out here. [general](https://github.com/Azure/sonic-buildimage/blob/master/onie-image.conf#L11)

build_image.sh
wen587: I think the file doesn't exist in some case. For example, if the TARGET_MACHINE was set to 'general', it won't match any platform_asic. I will bypass the check here.

TACappleman:
@wen587 is there a way to bypass this check (or respond to it) in automation, e.g. if it's hit in a Jenkins run?

wen587:
> @wen587 is there a way to bypass this check (or respond to it) in automation, e.g. if it's hit in a Jenkins run?

If you are using `sonic-installer install`, you can use `sonic-installer install -f` to bypass the check and force install.
If you are using `onie install`, a char of `Y` or `y` need to be responded if the platform does not match.

PR #8857, opened 2021-09-28
[device]: Add SAI checksum verify to TD3 config

theasianpianist:

#### Why I did it
A new config option was added to control the value of IPV4_INCR_CHECKSUM_ORIGINAL_VALUE_VERIFY in the EGR_FLEX_CONFIG control register (this prevents checksums of 0xffff from being propagated to other devices)

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106


gechiang (approved):
Looks good to me!

gechiang:
@theasianpianist Please note that the necessary changes in BRCM SAI is only available in 202012 BRCM SAI 4.3.5.1-3. This means that this PR ideally should wait until the same fix is applied to the master branch BRCM SAI 5.0 before you merge it...
I would suggest that you raise a separate PR for the 202012 branch with this same change so that the 202012 branch which is ready with the BRCM SAI support can start getting this complete fix.

gechiang:
@theasianpianist you should also press BRCM to make sure they push this same change into their REL_5.0 SUG repo and any future release that they are also developing such as 6.0.  They miss this type of things quite often and needs to be reminded unfortunately or else there will be a surprised gap...

prsunny (approved):

gechiang:
@theasianpianist , Should this SOC property be applied to all TD3 based HW SKUs or just the two you updated?

theasianpianist:
> @theasianpianist , Should this SOC property be applied to all TD3 based HW SKUs or just the two you updated?

Checked offline with @lguohan, we only need to update the 7050CX3 SKUs

PR #8880, opened 2021-09-30
[sonic-yang]: introduce yang template for cvl yang

lguohan:
…models

python3 setup.py build
python3 setup.py sdist

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106

### DIFF
--- a/src/sonic-yang-models/setup.py
+++ b/src/sonic-yang-models/setup.py
@@ -1,8 +1,43 @@
+import shutil
+import os
+import glob
+import jinja2
 from setuptools import setup, find_packages
+from setuptools.command.build_py import build_py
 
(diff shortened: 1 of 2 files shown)


lguohan:
related #6861

anand-kumar-subramanian (changes requested):
Yang model type should be CVL and PV(python validator) instead of int and ext.

zhangyanzhao:
> Yang model type should be CVL and PV(python validator) instead of int and ext.

@lguohan  would you please double check this and get the build issue fixed? Thanks

anand-kumar-subramanian (approved):

PR #8889, opened 2021-10-01
sysready

sg893052:
#### Why I did it
System Ready Feature Implementation as per https://github.com/Azure/SONiC/pull/875/files
At present, there is no mechanism to know that the system is up with all the essential sonic services and also, all the docker apps are ready along with port ready status to start the network traffic. With the asynchronous architecture of SONiC, we will not be able to verify if the config has been applied all the way down to the HW. But we can get the closest up status of each app and arrive at the system readiness.

#### How I did it
A new python based System monitor framework is introduced to monitor all the essential system host services including docker wrapper services on an event based model and declare the system is ready. This framework gives provision for docker apps to notify its closest up status. CLIs are provided to fetch the current system status and also service running status and its app ready status along with failure reason if any.

#### How to verify it
"show system status core" click CLI
"show system status all" click CLI
Syslogs for system ready

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106

### DIFF
--- a/files/build_templates/init_cfg.json.j2
+++ b/files/build_templates/init_cfg.json.j2
@@ -51,7 +51,27 @@
 {%- if feature in ["lldp", "pmon", "radv", "snmp", "telemetry"] %}
             "set_owner": "kube", {% else %}
             "set_owner": "local", {% endif %} {% endif %}
-            "high_mem_alert": "disabled"
+            "high_mem_alert": "disabled",
+{%- if feature in ["bgp", "swss", "pmon", "nat", "teamd", "dhcp_relay", "sflow", "l2mcd", "udld", "stp", "snmp", "lldp", "radv", "iccpd", "syncd", "vrrp", "mgmt-framework", "tam"] %}
+            "check_up_status" : "false"
+{%- else %}
+            "check_up_status" : "false"
+{%- endif %}
+        }{% if not loop.last %},{% endif -%}
+{% endfor %}
+    },
+{%- set host_features = [("caclmgrd", false),
+                        ("cron", false),
+                        ("docker", false),
+                        ("hostcfgd", false)] %}
+    "HOST_FEATURE": {
+{%- for host_feature, check_up_status in host_features %}
+        "{{host_feature}}": {
+{%- if host_feature in ["caclmgrd", "hostcfgd"] %}
+            "check_up_status" : "false"
+{%- else %}
+            "check_up_status" : "false"
+{%- endif %}
         }{% if not loop.last %},{% endif -%}
 {% endfor %}
     }
--- a/files/image_config/sysmonitor/sysmonitor.py
+++ b/files/image_config/sysmonitor/sysmonitor.py
@@ -1,0 +1,691 @@
+#!/usr/bin/python3
+
+from datetime import datetime
+import time
+import datetime
+import os
+import subprocess
+import sys
+import threading
+import syslog
+import argparse
+import multiprocessing as mp
+from swsssdk import ConfigDBConnector
+from swsssdk import SonicV2Connector
+import socket
+import json
+import fcntl
+import stat
+
+SYSLOG_IDENTIFIER="system#monitor"
+STATE_FEATURE_TABLE_NAME = "FEATURE"
+REDIS_TIMEOUT_MS = 0
+SYSTEM_STATE="DOWN"
+logger = None
+SYSTEM_CORESRV_STATE="DOWN"
+SYSTEM_ALLSRV_STATE="DOWN"
+SYSREADY_LOCKFILE="/var/run/sysready.lock"
+core_dnsrvs_name_list=[]
+dnsrvs_name_list=[]
+allsrvs_dict={}
+coresrvs_dict={}
+allsrvs_status="DOWN"
+coresrvs_status="DOWN"
+spl_srv_list= ['database-chassis', 'gbsyncd']
+core_srv_list = [
+        'swss.service',
+        'bgp.service',
+        'teamd.service',
+        'pmon.service',
+        'syncd.service',
+        'database.service',
+        'mgmt-framework.service',
+]
+
+class FileLock:
+   def __init__(self, lock_file):
+      self.f = open(lock_file, 'w')
+
+   def lock(self):
+      fcntl.flock(self.f, fcntl.LOCK_EX)
+
+   def unlock(self):
+      fcntl.flock(self.f, fcntl.LOCK_UN)
+
+sysready_lock = FileLock(SYSREADY_LOCKFILE)
+
+
+class Logger(object):
+    def __init__(self, syslog_identifier):
+        syslog.openlog(ident=syslog_identifier, logoption=syslog.LOG_NDELAY, facility=syslog.LOG_DAEMON)
+
+    #def __del__(self):
+        #syslog.closelog()
+
+    def log_emerg(self, msg, also_print_to_console=False):
+        syslog.syslog(syslog.LOG_EMERG, msg)
+
+        if also_print_to_console:
+            print(msg)
+
+    def log_crit(self, msg, also_print_to_console=False):
+        syslog.syslog(syslog.LOG_CRIT, msg)
+
+        if also_print_to_console:
+            print(msg)
+
+    def log_alert(self, msg, also_print_to_console=False):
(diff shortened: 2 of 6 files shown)


sg893052:
> This pull request **introduces 2 alerts** when merging [5c28c6d](https://github.com/Azure/sonic-buildimage/commit/5c28c6d4c91ce8c4a76ce3bd5b8d6f8fd9a453ba) into [df6361f](https://github.com/Azure/sonic-buildimage/commit/df6361f50c1e8fb1917788e3a51d40cb1c72069d) - [view on LGTM.com](https://lgtm.com/projects/g/Azure/sonic-buildimage/rev/pr-17f62f3695cd48967e1c4499eb95409fb677af28)
> 
> **new alerts:**
> 
> * 2 for Unreachable code

src/sonic-yang-models/yang-models/sonic-feature.yang
```diff
@@ -66,7 +66,37 @@ module sonic-feature{
...
+        container HOST_FEATURE {
+
+            description "host feature table in config_db.json";
+
+            list HOST_FEATURE_LIST {
+
+                key "name";
+
+                leaf name {
+                    description "host feature name in host Feature table";
```
venkatmahalingam, line 87: Can you give some examples of host features like hostcfgd, caclmgrd...etc?

files/build_templates/init_cfg.json.j2
```diff
@@ -51,7 +51,27 @@
 {%- if feature in ["lldp", "pmon", "radv", "snmp", "telemetry"] %}
             "set_owner": "kube", {% else %}
             "set_owner": "local", {% endif %} {% endif %}
-            "high_mem_alert": "disabled"
+            "high_mem_alert": "disabled",
+{%- if feature in ["bgp", "swss", "pmon", "nat", "teamd", "dhcp_relay", "sflow", "l2mcd", "udld", "stp", "snmp", "lldp", "radv", "iccpd", "syncd", "vrrp", "mgmt-framework", "tam"] %}
+            "check_up_status" : "false"
```
venkatmahalingam, line 56: Why do we need this initialization, cant we assume if check_up_status field is not present,  it's false?

files/image_config/sysmonitor/sysmonitor.py
```diff
@@ -0,0 +1,691 @@
...
+    while True:
+        event = QUEUE.get()
+        #logger.log_info( "System event [ "+event+" ] is received")
+        try:
+            sysready_lock.lock()
+            check_unit_status(event, st_db, ap_db, config_db)
+            sysready_lock.unlock()
+        except Exception as e:
+            logger.log_error( str(e))
+            time.sleep(2)
```
venkatmahalingam, line 671: Do we need to periodically check the status every 2 secs? what's the CPU usage for this service?

sg893052: sleep(2) is invoked only in case of exception. 
There is no periodic check. 
As the design is on event based framework, the logic is to check the status of a service unit that was received in the queue. It will not hog the cpu.

files/build_templates/init_cfg.json.j2
sg893052: This is part of framework provision for dockers to set "true" against "check_up_status" flag.
For now, all these dockers check_up_status is set to 'false'. Later once these dockers implement the logic of marking the up_status flag in STATE_DB, can set it to 'true' eventually.

src/sonic-yang-models/yang-models/sonic-feature.yang
sg893052: Sure, will add.

sg893052:
[system_ready_ut.pdf](https://github.com/Azure/sonic-buildimage/files/7486396/system_ready_ut.pdf)

venkatmahalingam (approved):

zhangyanzhao:
Venkat approved and all build check passed. Merge it

lguohan:
to note, this pr has been reverted. per our discussion, i think this pr need mellanox folks to review and approve.

liat-grozovik:
@zhangyanzhao as we have comments on the HLD itself i suggest to revert this PR as well.
it should not be merged without agreeing on the HLD.

zhangyanzhao:
Add yang label since it has yang related change

files/image_config/sysmonitor/sysmonitor.py
```diff
@@ -0,0 +1,691 @@
+#!/usr/bin/python3
+
+from datetime import datetime
```
lguohan, line 3: need to package it into a proper python package, need unit test to validate the service.

files/image_config/sysmonitor/sysmonitor.py
```diff
@@ -0,0 +1,691 @@
...
+core_dnsrvs_name_list=[]
+dnsrvs_name_list=[]
+allsrvs_dict={}
+coresrvs_dict={}
+allsrvs_status="DOWN"
+coresrvs_status="DOWN"
+spl_srv_list= ['database-chassis', 'gbsyncd']
+core_srv_list = [
+        'swss.service',
+        'bgp.service',
```
lguohan, line 37: i am not sure bgp.service is a core service, we need to check how this interack with application extension framework.

files/image_config/sysmonitor/sysmonitor.py
```diff
@@ -0,0 +1,691 @@
...
+import datetime
+import os
+import subprocess
+import sys
+import threading
+import syslog
+import argparse
+import multiprocessing as mp
+from swsssdk import ConfigDBConnector
+from swsssdk import SonicV2Connector
```
lguohan, line 14: swsssdk is on the deprecation path, it cannot be used any more.

files/image_config/sysmonitor/sysmonitor.py
```diff
@@ -0,0 +1,691 @@
...
+            connection.sendall(json.dumps(fail_res).encode('utf-8'))
+
+        connection.close()
+
+    #sock.close() #lgtm [py/unreachable-statement]
+
+
+def db_connect():
+    try:
+        st_db = SonicV2Connector()
```
lguohan, line 626: we should all use the swsscommon library.

src/sonic-yang-models/yang-models/sonic-feature.yang
```diff
@@ -66,7 +66,38 @@ module sonic-feature{
...
+                
+                leaf check_up_status {
+                    description "This configuration controls the system ready tool to check
+                                the app ready/up status";
+                    type boolean;
+                    default false;
+                }
+            }
+        }
+        container HOST_FEATURE {
```
lguohan, line 78: we should rename sonic-host-feature

lguohan, line 78: i am not sure we should introduce such yang model, why not in the feature table, we add a flag to indicate if it is host or container feature.

files/build_templates/init_cfg.json.j2
```diff
@@ -51,7 +51,27 @@
 {%- if feature in ["lldp", "pmon", "radv", "snmp", "telemetry"] %}
             "set_owner": "kube", {% else %}
             "set_owner": "local", {% endif %} {% endif %}
-            "high_mem_alert": "disabled"
+            "high_mem_alert": "disabled",
+{%- if feature in ["bgp", "swss", "pmon", "nat", "teamd", "dhcp_relay", "sflow", "l2mcd", "udld", "stp", "snmp", "lldp", "radv", "iccpd", "syncd", "vrrp", "mgmt-framework", "tam"] %}
```
stepanblyschak, line 55: dhcp_relay is now an application extension package. We should not have it hardcoded as we may have images with dhcp_relay and without it.
We need to add this new field here - https://github.com/Azure/sonic-utilities/blob/master/sonic_package_manager/service_creator/feature.py#L34.
And we need to add a new manifest variable to control whether "check_up_status" should be up true or false which will also be an indication whether docker implements marking the up_status flag in STATE_DB.

files/image_config/sysmonitor/sysmonitor.py
stepanblyschak: We will probably need to define whether a docker is a core or not in the docker's manifest.

PR #8917, opened 2021-10-07
Fix uninitialized variable in minigraph parsing

TACappleman:
#### Why I did it
Local variable may not be created in minigraph parsing, causing crash

#### How I did it
Moved variable creation earlier


TACappleman:
@arlakshm could you approve and merge this PR please?

arlakshm (approved):

PR #8950, opened 2021-10-11
[sonic-slave]: Add gmock for sonic-swss-common tests 

donNewtonAlpha:
Signed-off-by: Don Newton <don@opennetworking.org>

#### Why I did it
Sonic-swss-common requires gmock for staged unit tests
#### How I did it
Installed dependency in sonic-buster-build docker file
#### How to verify it
rebuild the build container

### DIFF
--- a/sonic-slave-bullseye/Dockerfile.j2
+++ b/sonic-slave-bullseye/Dockerfile.j2
@@ -235,6 +235,8 @@ RUN apt-get update && apt-get install -y \
 # For gtest
         libgtest-dev \
         cmake \
+# For gmock
+        libgmock-dev \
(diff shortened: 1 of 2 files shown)


bocon13 (approved):

bocon13:
@lguohan @yxieca Please help us backport this to 202106 and 202012.

Currently all active branches share the same Docker tag:
`sonicdev-microsoft.azurecr.io:443/sonic-slave-buster:latest`

Longer term, we should fix this, but for now, this is causing trouble for PINS. Specifically, if the 202012 container is the most recently pushed, we are missing gmock.

Thanks!

PR #8951, opened 2021-10-11
[201811] Invoke disk check periodically

renukamanavalan:
#### Why I did it
When disk become read-only (kernel bug), it blocks new remote users access.
This fix would enable remote user access even in read-only state

#### How I did it
Keep monitoring disk state.
If it becomes read-only, mount /etc & home dir in tmpfs to become RW and periodically write error logs
With /etc & /home set in RW state, this allows remote user access

#### How to verify it
1) Make sure TACACS is configured
2) Make disk read-only (`sudo bash -c "echo u > /proc/sysrq-trigger"`)
3) Try logging in as remote user, who has not logged in before.

- [x] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106

### DIFF
--- a/files/image_config/monit/conf.d/sonic-host
+++ b/files/image_config/monit/conf.d/sonic-host
@@ -25,6 +25,5 @@ check process rsyslog with pidfile /var/run/rsyslogd.pid
 # Raise syslog error message, in case of underlying issues
 #
 check program diskCheck with path "/usr/local/bin/disk_check.py"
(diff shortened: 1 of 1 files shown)


yxieca:
@renukamanavalan please complete the PR template, especially please elaborate who did you test the change? I think this change needs to have a matching submodule update.

yxieca (approved):

PR #8952, opened 2021-10-11
Advance the broadcom SAI to 5.0.0.11

judyjoseph:
#### Why I did it
Advance the broadcom SAI to 5.0.0.11

#### How I did it

Includes the following commits/CSPs
```
9a1765cf CS00012203989: Max mtu not set at init knet interfaces
9b59e33f Fixes for CS00012195261, CS00012195512, CS00012194313, CS00012178692, CS00012192505, SDK SIDs
```

#### How to verify it
<To be updated>

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106

### DIFF
--- a/platform/broadcom/saibcm-modules/debian/changelog
+++ b/platform/broadcom/saibcm-modules/debian/changelog
@@ -1,3 +1,9 @@
+opennsl (5.0.0.11) unstable; urgency=medium
+
+  * Update to Broadcom SAI 5.0.0.11
+
+ -- Judy Joseph <jujoseph@microsoft.com>  Fri, 8 Oct 2021 18:36:38 +0000
+
 opennsl (5.0.0.4) unstable; urgency=medium
 
   * Update to Broadcom SAI 5.0.0.4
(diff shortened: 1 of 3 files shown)


platform/broadcom/saibcm-modules/debian/changelog
```diff
@@ -1,3 +1,9 @@
+opennsl (5.0.0.11) unstable; urgency=medium
```
gechiang, line 1: Wondering if this file was changed and included when you built the libsaibcm debians?  If not, this version will not be reflected when one issue bcmcmd "bsv" correct?

judyjoseph: libsaibcm is itself a debian and it has the version defined there in saibcm 
```
networking-acs-saibcm$ git grep 5.0.0.11   
RELEASE:5.0.0.11
debian/changelog:saibcm (5.0.0.11) unstable; urgency=medium
include/brcm_sai_version.h:#define BRCM_SAI_VER "5.0.0.11"
```
We update this modules file -- so that we can track which saibcm release caused this change.

gechiang: I guess there are no minor involved yet this time so it is probably ok for now... it there are minor involved such as something like 5.0.0.11-1 then the "-1" will be missing if the debian/changelog is not included as part of the debian build...

PR #6971, opened 2021-03-05
[Quanta] Add and update platform and device files

roberthong-qct:
#### Why I did it
Add device and platform code for ix7-bwde, ix8a-bwde.
Support platform API 2.0 for all quanta platforms except for ix1b
#### How I did it
Implement code
#### How to verify it
for all quanta platforms except for ix1b:
sfputil show eeprom
show platform fan
show platform psu
show platform syseeprom
show platform temp
#### Which release branch to backport (provide reason below if selected)

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012

PR #8967, opened 2021-10-13
[202012] [Innovium] Add new platforms

Hanly-Cameo:
#### Why I did it
Add Cameo boxes support 
#### How to verify it
verify with sonic cli(platform driver)

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012
- [ ] 202106


Komil-yc:
**"40 for Wrong number of arguments in a class instantiation"** this should be limitation or bug in LGTM, same alerts are present in previous [PR#7545](https://github.com/Azure/sonic-buildimage/pull/7545)

**"5 for Module imports itself"** and **"5 for 'import *' may pollute namespace"** happen on other PR(e.g. [#6971](https://github.com/Azure/sonic-buildimage/pull/6971) ), they are ignored.

Komil-yc:
**"4 for Modification of parameter with default"** 
I thank they are misjudgment by LGTM. 
The variable "lanes" presents a new list and it's element are copied from A_lane.
The following assignment on "lanes" should not modify the original referenced list.
圖片

PR #8983, opened 2021-10-15
[Mellanox] Support PSU voltage high/low thresholds and power max threshold

stephenxs:
#### Why I did it
Support PSU voltage high/low thresholds and power max threshold

- Add thresholds support for voltage and power.
  As thresholds are not supported on all platforms, we need to check the capability first and fetch thresholds only if it is supported.

#### How to verify it
Run regression test and manual test.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106


keboliu (approved):

keboliu (approved):

sujinmkang:
@stephenxs can the capability be added to platform.json so that the sonic-mgmt test can skip or run the tests based on the field?

sujinmkang (approved):

PR #8989, opened 2021-10-15
[chassis][mgmtif][defaultroute] Fix for issue of mgmt if default route

vganesan-nokia:
#### Why I did it

If management interface is configured in config db, a default route is added with the management interface (eth0) as the gateway with same admin distance as iBGP routes. This is done as a backup reachability for management interface. In chassis using multi-asic image, this route was interfering with the default routes received from the iBGP sessions since static route has higher preference (i.e., the route on eth0 is selected as best route). However since routesync does not send any route on eth0 to orchagent, we do not have a default route in the datapath. This PR fixes this issue

#### How I did it

The backup default route management interface as gateway is added as part of staticd configuration in FRR. This is changed to set the administrative distance greater than the administrative distance of iBGP routes.

#### How to verify it

- In multi-asic imaged chassis (especially with line card with single asics), configure eth0 management interface in MGMT_INTERFACE table.
- Configure a default route with gateway on an eBGP interface. This default route from eBGP sessions will be leaned internally via iBGP in all other asics (via backend asics' iBGP sessions)
- Bring up the chassis and make sure the bgp is up and running and all iBGP and eBGP sessions are up.
- Check the route table for 0.0.0.0/0 route. 
- Observe that the there is a route for 0.0.0.0/0 with gateway equal to the management interface address and this route is not the best route.

Fix for issue of static default route on management interface interfering with default routes learned via iBGP.

### DIFF
--- a/dockers/docker-fpm-frr/frr/staticd/staticd.default_route.conf.j2
+++ b/dockers/docker-fpm-frr/frr/staticd/staticd.default_route.conf.j2
@@ -1,9 +1,17 @@
 !
 {% block default_route %}
+{% if DEVICE_METADATA['localhost']['sub_role'] == 'FrontEnd' or DEVICE_METADATA['localhost']['sub_role'] == 'BackEnd' or DEVICE_METADATA['localhost']['switch_type'] == 'voq' %}
+{% set multi_asic_chassis = True %}
+{% endif %}
 ! set static default route to mgmt gateway as a backup to learned default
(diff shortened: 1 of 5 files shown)


abdosi:
@arlakshm please check

vganesan-nokia:
With PR https://github.com/Azure/sonic-buildimage/pull/9182, since the static default route addition on mgmt interface is removed, this PR is no longer needed. Closing this PR.

PR #9036, opened 2021-10-21
[dhcp_relay] fix data type in dhcp6relay, add protection in packet data parsing

kellyyeh:
#### Why I did it
Error in socket receiver, causing dhcp6relay to be stuck in infinite while loop when parsing through packet data.

#### How I did it
Fix recv datatype, add a protection check to check if the next pointer of packet is the same as current pointer.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012
- [ ] 202106

Signed off by Kelly Yeh kellyyeh@microsoft.com

### DIFF
--- a/src/dhcp6relay/src/relay.cpp
+++ b/src/dhcp6relay/src/relay.cpp
@@ -538,7 +538,7 @@ void callback(evutil_socket_t fd, short event, void *arg) {
         do {
             ext_header = (const struct ip6_ext *)current_position;
             current_position += ext_header->ip6e_len;
-            if((current_position == prev) || (current_position <= (uint8_t *)ptr + sizeof(message_buffer))) {
+            if((current_position == prev) || (current_position >= (uint8_t *)ptr + sizeof(message_buffer))) {
                 return;
             }
             prev = current_position;


src/dhcp6relay/src/relay.cpp
```diff
@@ -515,26 +515,32 @@ void relay_client(int sock, const uint8_t *msg, int32_t len, const ip6_hdr *ip_h
 void callback(evutil_socket_t fd, short event, void *arg) {
     struct relay_config *config = (struct relay_config *)arg;
     static uint8_t message_buffer[4096];
-    uint32_t len = recv(config->filter, message_buffer, 4096, 0);
+    int32_t len = recv(config->filter, message_buffer, 4096, 0);
     if (len <= 0) {
         syslog(LOG_WARNING, "recv: Failed to receive data at filter socket\n");
```
saiarcot895, line 520: Can you print `errno` as well here, and add a `return`?

src/dhcp6relay/src/relay.cpp
```diff
@@ -515,26 +515,32 @@ void relay_client(int sock, const uint8_t *msg, int32_t len, const ip6_hdr *ip_h
...
 
     auto ip_header = parse_ip6_hdr(current_position, &tmp);
     current_position = tmp;
 
+    prev = current_position;
     if (ip_header->ip6_ctlun.ip6_un1.ip6_un1_nxt != IPPROTO_UDP) {
         const struct ip6_ext *ext_header;
         do {
             ext_header = (const struct ip6_ext *)current_position;
             current_position += ext_header->ip6e_len;
```
saiarcot895, line 539: Is it worth it to add a bounds check here? Basically to make sure that the new `current_position` is within the `message_buffer`? This would prevent a malicious crafted packet from crashing dhcp6relay.

src/dhcp6relay/src/relay.cpp
```diff
@@ -515,29 +515,36 @@ void relay_client(int sock, const uint8_t *msg, int32_t len, const ip6_hdr *ip_h
...
+            if((current_position == prev) || (current_position <= (uint8_t *)ptr + sizeof(message_buffer))) {
```
yxieca, line 541: should be '>=', right?

kellyyeh: yes, thanks for catching that!

PR #9083, opened 2021-10-27
[PINS] Build P4RT container for PINS

bocon13:
- Add INCLUDE_PINS to config to enable/disable container
- Add Docker files and supporting resources
- Add sonic-pins submodule and associated make files

Submission containing materials of a third party:
    Copyright Google LLC; Licensed under Apache 2.0

Co-authored-by: Runming Wu <runmingwu@google.com>
Co-authored-by: Robert J. Halstead <rhalstea@google.com>
Co-authored-by: Srikishen Pondicherry Shanmugam <kishanps@google.com>
Co-authored-by: Brian O'Connor <bocon@opennetworking.org>

#### Why I did it

Adds P4RT container to SONiC for PINS

The P4RT app is covered by this HLD:
https://github.com/pins/SONiC/blob/master/doc/pins/p4rt_app_hld.md

#### How I did it

Followed the pattern and templates used for other SONiC applications

#### How to verify it

Build SONiC with INCLUDE_P4RT set to "y".
Verify that the resulting build has a container called "p4rt" running.
You can verify that the service is up by running the following command on the SONiC switch:
```bash
sudo netstat -lpnt | grep p4rt
```
You should see the service listening on TCP port 9559.

#### Which release branch to backport (provide reason below if selected)

None

#### Description for the changelog

Build P4RT container for PINS

PR #9084, opened 2021-10-27
[PINS] Add submit to ingress port to copp_cfg file

bocon13:
Submission containing materials of a third party:
    Copyright Google LLC; Licensed under Apache 2.0

Co-authored-by: Glenn Connery <gconnery@google.com>

Depends on:
- [ ] https://github.com/Azure/sonic-buildimage/pull/9083
- [ ] https://github.com/Azure/sonic-swss/pull/1952

#### Why I did it

Adds support for submit_to_ingress port required for PINS packet out path

More details in this HLD:
https://github.com/pins/SONiC/blob/master/doc/pins/Packet_io.md

#### How to verify it

Build the SONiC image and verify that the submit_to_ingress netdev is created

None

Add submit to ingress port to copp_cfg file

### DIFF
--- a/files/image_config/copp/copp_cfg.j2
+++ b/files/image_config/copp/copp_cfg.j2
@@ -13,7 +13,7 @@
 		    "trap_priority":"4",
 		    "queue": "4"
 	    },
-	    "queue4_group2": {            
+	    "queue4_group2": {
 		    "trap_action":"copy",
 		    "trap_priority":"4",
 		    "queue": "4",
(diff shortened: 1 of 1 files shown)


files/image_config/copp/copp_cfg.j2
```diff
@@ -60,7 +60,10 @@
 		    "trap_action": "trap",
 		    "trap_priority": "1"
 
-	    }
+	    }{%- if include_p4rt == "y" %},
+	    "trap.group.send_to_ingress" : {
+		    "submit_to_ingress_name" : "send_to_ingress"
```
bocon13, line 65: Can we move this to the CONFIG_DB port table?
Issue: we can't specify that this is a host interface for CPU there today.

Can we move this to the P4Orch constructor?

bocon13: CoPP group has policers and other things. It doesn't really fit here.

bocon13: Proposal for MVP: put it in P4Orch
Longer term: make this part of the PORT_TABLE

PR #9098, opened 2021-10-28
[Yang model] Allow user to set none value for interface type

Junchao-Mellanox:
**What I did**

Allow user to set none value for interface type

**Why I did it**

Say user has speed=50G, interface_type=CR2, and user wants to change it to 10G and CR. Currently, there is no CLI can support it, because:

1. If you change the interface type to CR first, SAI will get 50G and CR, SAI will report error because not supported
2. If you change the speed to 10G first, SAI will get 10G and CR2, also not supported

The only way for now is to manually change the CONFIG_DB and do a config reload, this is not user friendly.

This PR allow user to set none value to interface type. So there is a way to achieve the goal via CLI:

1. config interface type XXX none
2. config interface speed XXX 10000
3. config interface type XXX CR

**How I verified it**

Manual test

**Details if related**

### DIFF
--- a/src/sonic-yang-models/yang-models/sonic-types.yang
+++ b/src/sonic-yang-models/yang-models/sonic-types.yang
@@ -129,7 +129,7 @@ module sonic-types {
             enum UDP;
         }
     }
-    
+
     typedef interface_type {
         type enumeration {
             enum CR;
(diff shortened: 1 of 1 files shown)


keboliu (approved):

liat-grozovik:
@lguohan could you please help to merge?

judyjoseph (approved):

PR #9123, opened 2021-10-29
TACACS: Don't send sshd's bad password to AAA

renukamanavalan:
#### Why I did it
When sshd realizes that this login can't succeed due to internal device state
or configuration, instead of failing right there, it proceeds to prompt for
password, so as the user does not get any clue on where is the failure point.

Yet to ensure that this login does not proceed, sshd replaces user provided password
with a specific pattern of characters matching length of user provided password.
This pattern is "<BS><LF><CR><DEL>INCORRECT", which is bound to fail.

If user provided length is smaller/equal, the substring of pattern is overwritten.
If user provided length is greater, the pattern is repeated until length is exhausted.

But if the PAM-tacacs plugin would send this password to AAA, the user could get
locked out by AAA, for providing incorrect value.

#### How I did it
Hence this fix, matches obtained password against the pattern. If match, fail just before
reaching AAA server.

#### How to verify it

1. Make sure tacacs is properly configured.
2. Try logging in as, say "user-A"; ensure it succeeds
3. Pick another user, say user-B and ensure this user has not logged into this device before (_look into /etc/passed & folders under /home_)
4. Disable monit service (_as that could fix the issue using disk_check.py_)
5. Start TCP dump for all TACACS servers.
6. Simulate Read-only disk
7. Try logging in using  user-B.
8. Verify it fails, after 3 attempts
9. Stop tcp dump.
10. TCP dump should show "authentication" for user-A only

- [x] 201811
- [x] 201911
- [x] 202006
- [x] 202012
- [x] 202106

### DIFF
--- a/src/tacacs/pam/0010-handle-bad-password-set-by-sshd.patch
+++ b/src/tacacs/pam/0010-handle-bad-password-set-by-sshd.patch
@@ -1,0 +1,99 @@
+From eb7f56e4a75050091916841d2d9daec646a4210a Mon Sep 17 00:00:00 2001
+From: Renuka Manavalan <remanava@microsoft.com>
+Date: Fri, 29 Oct 2021 21:40:55 +0000
+Subject: [PATCH] handle bad password set by sshd
+
+---
+ pam_tacplus.c |  7 +++++++
+ support.c     | 39 ++++++++++++++++++++++++++++++++++++++-
+ support.h     |  1 +
+ 3 files changed, 46 insertions(+), 1 deletion(-)
+
+diff --git a/pam_tacplus.c b/pam_tacplus.c
+index d57657a..2f1ed46 100644
+--- a/pam_tacplus.c
++++ b/pam_tacplus.c
+@@ -248,6 +248,13 @@ int pam_sm_authenticate (pam_handle_t * pamh, int flags,
+         return PAM_CRED_INSUFFICIENT;
+     }
+ 
++    if (validate_not_sshd_bad_pass(pass) != PAM_SUCCESS) {
++        syslog(LOG_ERR, "auth fail: Password incorrect");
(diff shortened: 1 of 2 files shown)


liuh-80 (approved):
I checked this bug still not fix by the upstream project: there are some discussion about this in upstream project 3 years ago, but they never fix it.

src/tacacs/pam/0010-handle-bad-password-set-by-sshd.patch
```diff
@@ -0,0 +1,99 @@
...
+index d57657a..2f1ed46 100644
+--- a/pam_tacplus.c
++++ b/pam_tacplus.c
+@@ -248,6 +248,13 @@ int pam_sm_authenticate (pam_handle_t * pamh, int flags,
+         return PAM_CRED_INSUFFICIENT;
+     }
+ 
++    if (validate_not_sshd_bad_pass(pass) != PAM_SUCCESS) {
++        syslog(LOG_ERR, "auth fail: Password incorrect");
++        bzero(pass, strlen (pass));
```
liuh-80, line 22: The bzero function been deprecated because the compiler may remove the code: https://man7.org/linux/man-pages/man3/bzero.3.html

However this is not a bug, because when code run to here, the 'pass' only contains SSHD_BAD_PASS, so no risk to leak user secret.

renukamanavalan: The code is currently using bzero in other places too. That is how, I put it.
Does it mean, we patch other occurrences too ?

renukamanavalan: Searched across and fixed all.

liuh-80 (approved):

qiluo-msft:

PR #9138, opened 2021-11-01
[201811][dhcp_relay] Fix sendto: invalid destination address error

kellyyeh:
#### Why I did it
Not all relay forward packets are sent to desired dhcpv6_servers.
Sendto error: invalid destination address

#### How I did it
Memory gets deallocated at end of loop_relay, need to allocate object on the heap.

#### How to verify it
Build dhcp6relay debian package and run dhcpv6 relay tests

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106

Signed off by: Kelly Yeh kellyyeh@microsoft.com

### DIFF
--- a/src/dhcp6relay/src/relay.cpp
+++ b/src/dhcp6relay/src/relay.cpp
@@ -20,7 +20,7 @@ struct event *ev_sigint;
 struct event *ev_sigterm;
 static std::string counter_table = "DHCPv6_COUNTER_TABLE|";
 swss::DBConnector state_db(6, "localhost", 6379, 0);
-swss::RedisClient m_applDbRedisClient(&state_db);
+swss::RedisClient m_stateDbRedisClient(&state_db);
 
 /* DHCPv6 filter */
 /* sudo tcpdump -dd "ip6 dst ff02::1:2 && udp dst port 547" */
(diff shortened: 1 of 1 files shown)


src/dhcp6relay/src/relay.cpp
```diff
@@ -681,24 +680,26 @@ void dhcp6relay_stop()
  */
 void loop_relay(std::vector<relay_config> *vlans) {
     std::vector<int> sockets;
-    
-    for(std::size_t i = 0; i<vlans->size(); i++) {
-        struct relay_config config = vlans->at(i);
+    std::vector<relay_config *> configs;
+
+    for(relay_config vlan: *vlans) {
+        relay_config *config = new relay_config(vlan);
```
saiarcot895, line 686: So this memory allocation can be avoided by resuing the object that's already in `vlans` (this list will exist for the duration of the application, since this is actually allocated in the `main` method). To do this, use `relay_config *config = &vlans->at(i)`. That way, you don't need to free the memory at the end of this function.

saiarcot895 (changes requested):

src/dhcp6relay/src/relay.cpp
kellyyeh: Thanks for pointing it out! Updated

saiarcot895 (approved):

yxieca (approved):

PR #9143, opened 2021-11-02
[dhcp_relay] Fix sendto: invalid destination address error

kellyyeh:
#### Why I did it
Not all relay forward packets are sent to desired dhcpv6_servers.
Sendto error: invalid destination address

#### How I did it
Memory gets deallocated at end of loop_relay, need to allocate object on the heap.

#### How to verify it
Build dhcp6relay debian package and run dhcpv6 relay tests

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012
- [ ] 202106

### DIFF
--- a/src/dhcp6relay/src/relay.cpp
+++ b/src/dhcp6relay/src/relay.cpp
@@ -682,16 +682,16 @@ void dhcp6relay_stop()
 void loop_relay(std::vector<relay_config> *vlans, swss::DBConnector *db) {
     std::vector<int> sockets;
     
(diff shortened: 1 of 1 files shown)


saiarcot895 (approved):

lguohan:
why we could not catch this issue in the sonic-mgmt test?

kellyyeh:
@lguohan There were 4 server addresses, and only one was affected, so packets were still successfully relayed to the other three destination addresses. Sonic-mgmt only checks if >=1 expected packet is received.

PR #9151, opened 2021-11-02
Simplify the broadcom sai packages reference

xumia:
#### Why I did it
For broadcom sai, we only need to upgrade the version, not necessary the token part in the url.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106

### DIFF
--- a/platform/broadcom/sai.mk
+++ b/platform/broadcom/sai.mk
@@ -1,12 +1,16 @@
-BRCM_SAI = libsaibcm_6.0.0.10-1_amd64.deb
-$(BRCM_SAI)_URL = "https://sonicstorage.blob.core.windows.net/packages/bcmsai/6.0/master/XGS/libsaibcm_6.0.0.10-1_amd64.deb?sv=2015-04-05&sr=b&sig=5fW6otWPCPrsSIXeomEcllS8sEDXMo0YOD8UhB370U0%3D&se=2035-07-22T20%3A50%3A51Z&sp=r"
-BRCM_SAI_DEV = libsaibcm-dev_6.0.0.10-1_amd64.deb
+LIBSAIBCM_VERSION = 6.0.0.10-1
+LIBSAIBCM_BRANCH_NAME = REL_$(shell echo $(LIBSAIBCM_VERSION) | cut -d. -f1,2)
+LIBSAIBCM_URL_PREFIX = "https://sonicstorage.blob.core.windows.net/public/sai/bcmsai/$(LIBSAIBCM_BRANCH_NAME)/$(LIBSAIBCM_VERSION)"
+
+BRCM_SAI = libsaibcm_$(LIBSAIBCM_VERSION)_amd64.deb
+$(BRCM_SAI)_URL = "$(LIBSAIBCM_URL_PREFIX)/$(BRCM_SAI)"
+BRCM_SAI_DEV = libsaibcm-dev_$(LIBSAIBCM_VERSION)_amd64.deb
 $(eval $(call add_derived_package,$(BRCM_SAI),$(BRCM_SAI_DEV)))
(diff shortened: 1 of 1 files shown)


platform/broadcom/sai.mk
```diff
@@ -1,12 +1,16 @@
-BRCM_SAI = libsaibcm_6.0.0.10_amd64.deb
-$(BRCM_SAI)_URL = "https://sonicstorage.blob.core.windows.net/packages/bcmsai/6.0/master/libsaibcm_6.0.0.10_amd64.deb?sv=2020-08-04&st=2021-10-20T15%3A46%3A03Z&se=2030-10-21T15%3A46%3A00Z&sr=b&sp=r&sig=zhosphas9cBuH%2B7mLC%2ByarQPQSIe2LY%2FXOATIW96cVY%3D"
-BRCM_SAI_DEV = libsaibcm-dev_6.0.0.10_amd64.deb
+LIBSAIBCM_VERSION = 4.3.5.1
```
xumia, line 1: The version is not correct, it is only an example

xumia:
@gechiang, it is only a sample PR, please skip the version 6.0 -> 4.3 change.

rules/config
```diff
@@ -202,3 +202,5 @@ BUILD_MULTIASIC_KVM = n
 
 # INCLUDE_MUX - build docker-mux for dual ToR (Gemini)
 INCLUDE_MUX = y
+
+STORAGE_PUBLIC_TOKEN = ?st=2021-11-02T13%3A34%3A45Z&se=2030-11-03T13%3A34%3A00Z&sp=r&sv=2018-03-28&sr=c&sig=7MmsVW2EUGYSUr94IPJGBxwI1qYI0VQx%2Fa7z0koJT60%3D
```
lguohan, line 206: is there plan to remove the this token? it does not seem that the token is useful.

platform/broadcom/sai.mk
```diff
@@ -1,12 +1,16 @@
...
+LIBSAIBCM_MAIN_VERSION = 4.3
```
xumia, line 2: Can be calculated from the LIBSAIBCM_VERSION .

platform/broadcom/sai.mk
```diff
@@ -1,12 +1,16 @@
...
+LIBSAIBCM_BUILD = 48112
```
xumia, line 3: We can remove it, for one package version it should always to the unique commit.

rules/config
xumia: @lguohan , do you mean to publish a storage account public accessible? so we do not need to token at all?

platform/broadcom/sai.mk
```diff
@@ -1,12 +1,16 @@
-BRCM_SAI = libsaibcm_6.0.0.10-1_amd64.deb
-$(BRCM_SAI)_URL = "https://sonicstorage.blob.core.windows.net/packages/bcmsai/6.0/master/XGS/libsaibcm_6.0.0.10-1_amd64.deb?sv=2015-04-05&sr=b&sig=5fW6otWPCPrsSIXeomEcllS8sEDXMo0YOD8UhB370U0%3D&se=2035-07-22T20%3A50%3A51Z&sp=r"
-BRCM_SAI_DEV = libsaibcm-dev_6.0.0.10-1_amd64.deb
+LIBSAIBCM_VERSION = 6.0.0.10-1
+LIBSAIBCM_BRANCH_NAME = REL_$(shell echo $(LIBSAIBCM_VERSION) | cut -d. -f1,2)
```
gechiang, line 2: Currently BRCM does not have a standard way to name their releases.  We have been trying to follow their naming where we bring in their code to libsaibcm side.  Here are the names so far:
For very old releases such as 20181130:
       3.5-201811
For 20191130:
      /rita/3_7
For 20201231:
     REL_4.3.5.1
For 20210630:
     REL_5.2
For master:
     REL_6.0

So not sure what you proposed here may cause future release from BRCM naming issue?
REL_4.3.5
REL_

xumia: @gechiang , can we simply use REL_6.0? If the branch change, we need to update it, otherwise, we only update the version.

gechiang (approved):

lguohan (approved):

PR #9152, opened 2021-11-02
[7060] put docker partition in the ramdisk for 7060

yxieca:
#### Why I did it
7060 has 4GB hard drive and they cannot fit full image from 202012 branch.

#### How I did it
Put docker partition in ram so that we could install full image on 7060
Put /var/log partition back on harddrive with space saved from docker partition.

#### How to verify it
1. One image can be installed.
2. Can upgrade from first image to second image.
3. warm reboot test passes (lag recovered increased from 60 to 76 seconds, still within budget)
4. fast reboot test passes
5. this change only affects 7060, 7060_qx still use harddrive for docker partition due to memory limitation.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012
- [x] 202106

### DIFF
--- a/files/Aboot/boot0.j2
+++ b/files/Aboot/boot0.j2
@@ -468,6 +468,8 @@ write_platform_specific_cmdline() {
     # detect the size of the flash partition from name in Aboot/EOS/SONiC
     local flash_size=$(($(df "$target_path" | tail -1 | tr -s ' ' | cut -f2 -d' ') / 1000))
 
+    local force_log_on_disk="no"
+
     if [ "$platform" = "raven" ]; then
         # Assuming sid=Cloverdale
         aboot_machine=arista_7050_qx32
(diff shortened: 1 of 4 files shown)


lguohan:
since this mainly affects fast reboot and warm reboot control plane down time, how much increase do we see without using ram v.s. using ram? what is the tolerance we have for fast reboot/warm reboot now on this platform?

lguohan (approved):

lguohan:
We discussed this subject. There is a PKRAM approach, however it was not merged in the kernel and we cannot really use it.
Meanwhile, there is an existing mechanism use the DRAM to simulate persistent memory to sustain kexec reboot, and that
has been in the kernel for a long time since Linux kernel 4.2.

   Is it possible for someone to further investigate if this approach can be used.

[RFC 00/43] PKRAM: Preserved-over-Kexec RAM - Anthony Yznaga (kernel.org)

How to Emulate Persistent Memory Using Dynamic Random-access Memory... (intel.com)

PR #9179, opened 2021-11-05
dhcp6relay: remove line overwriting docker-dhcp-relay variable

saiarcot895:
The dhcp6relay rules file had a line overwriting a variable for
docker-dhcp-relay. Remove that line.

This line caused a limited impact where if some (many?) of the docker
containers were already built, except for dhcp-relay, and the build
failed or was interrupted, then dhcp-relay container would fail to build
because this variable was overwritten and the python3-swsscommon
wouldn't get installed into the slave container. Most builds would be
fine, though.

#### How to verify it

To see the failure:
1. Start a regular build (for any platform).
2. When the containers are getting built, before the dhcp-relay container is built, interrupt the build with Ctrl-C.
3. If it's timed right, after restarting the build, the dhcp-relay container fails to build.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106


kellyyeh (approved):

lguohan:
do we need this for 202012?

PR #8242, opened 2021-07-21
[Mellanox] [SKU] Fix the shared headroom for 4600C-C64 SKU

vivekreddynv:
#### How I did it

1) Removed ingress_lossy_pool from the BUFFER_POOL list
2) Fx the the egress_lossless_pool_size value 

#### How to verify it

1) Reboot with the updated SKU.
2) Perform a qos reload avd verify if the buffer profiles are applied properly without any error

#### Which release branch to backport (provide reason below if selected)

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012
- [ ] 202106

PR #9215, opened 2021-11-10
[Mellanox] Adjust buffer parameters with 2km cable supported for 4600C non-generic SKUs

stephenxs:
#### Why I did it

Also recalculated all parameters with the latest algorithm with per-speed peer response time taken into account

#### How I did it

Detailed information of each SKU:
- C64:
  - t0: 32 100G downlinks and 32 100G uplinks
  - t1: 56 100G downlinks and 8 100G uplinks with 2km-cable supported
- D112C8: 112 50G downlinks and 8 100G uplinks.
- D48C40: 48 50G downlinks, 32 100G downlinks, and 8 100G uplinks
- D100C12S2: 4 100G downlinks, 2 10G downlinks, 100 50G downlinks, and 8 100G uplinks

2km cable is supported for C64 on t1 only

#### How to verify it

Run regression test (QoS)

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012
- [x] 202106


stephenxs:
It can be cherry-picked to 202012 cleanly.
For 202106, PR #8242 need to be cherry-picked first and then this one can be cherry-picked cleanly.

liat-grozovik:
@stephenxs if this was taken into consideration new logic resides only on master/202111 I don't think we should set the label as it cannot be cleanly cherry picked
Please correct me if I am wrong.

stephenxs:
> @stephenxs if this was taken into consideration new logic resides only on master/202111 I don't think we should set the label as it cannot be cleanly cherry picked Please correct me if I am wrong.

As offline discussed, it isn't related to the per speed peer response time for dynamic buffer calculation. So let's request it for 202012 and 202106 (after PR #8242 being cherry-picked).

baiwei0427 (approved):

device/mellanox/x86_64-mlnx_msn4600c-r0/Mellanox-SN4600C-D112C8/pg_profile_lookup.ini
```diff
@@ -17,22 +17,22 @@
 #  PG lossless profiles.
 # speed cable  size   xon    xoff   threshold
   10000    5m 19456  19456  24576          0
-  25000    5m 19456  19456  24576          0
```
neethajohn, line 20: @stephenxs, the 2km support is only for SN4600C-C64? If that is the case, let's create a separate pg_profile_lookup.ini for all other 4600skus without the 2km and share them. right now this file is shared by all 4600C skus

stephenxs: Yes. We had an internal discussion with Arch and agreed on using the updated algorithm to calculate the headroom size for all 4600c SKUs. So this common pg profile lookup.ini is updated
Will that do?

neethajohn (approved):

stephenxs:
> > @stephenxs if this was taken into consideration new logic resides only on master/202111 I don't think we should set the label as it cannot be cleanly cherry picked Please correct me if I am wrong.
> 
> As offline discussed, it isn't related to the per speed peer response time for dynamic buffer calculation. So let's request it for 202012 and 202106 (after PR #8242 being cherry-picked).

Add "Request for 202106" as the dependent PR has been cherry-picked.

PR #9224, opened 2021-11-10
[sonic-yang-mgmt] Adding in_place flag to sonic_yang.loadData

ghooo:
#### Why I did it
sonic_yang.loadData was by default removing tables without YANG from the input configdbJson i.e. modifying it in place

#### How I did it
Added an `in_place` flag so caller of `loadData` can decide whether to modify input in place or not.

#### How to verify it
unit-test added

### DIFF
--- a/src/sonic-yang-mgmt/sonic_yang_ext.py
+++ b/src/sonic-yang-mgmt/sonic_yang_ext.py
@@ -2,6 +2,7 @@
 # class sonic_yang. A separate file is used to avoid a single large file.
 
 from __future__ import print_function
+import copy
 import yang as ly
 import syslog
 from json import dump, dumps, loads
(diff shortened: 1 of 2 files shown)


qiluo-msft:
@praveen-li Could you help review?

src/sonic-yang-mgmt/sonic_yang_ext.py
```diff
@@ -1038,16 +1039,21 @@ def _findXpathList(self, xpath, list, keys):
     load_data: load Config DB, crop, xlate and create data tree from it. (Public)
     input:    data
               debug Flag
+              remove_non_yang_in_place Flag that indicates whether to remove tables without YANG
+                                       from configdbJson directly or create a copy and modify it
     returns:  True - success   False - failed
     """
-    def loadData(self, configdbJson, debug=False):
+    def loadData(self, configdbJson, debug=False, remove_non_yang_in_place=True):
```
qiluo-msft, line 1046: >remove_non_yang_in_place [](http://example.com/codeflow?start=50&length=24)

About the parameter name, how about `copyJson`?

src/sonic-yang-mgmt/sonic_yang_ext.py
```diff
@@ -1038,16 +1039,21 @@ def _findXpathList(self, xpath, list, keys):
     load_data: load Config DB, crop, xlate and create data tree from it. (Public)
     input:    data
               debug Flag
+              remove_non_yang_in_place Flag that indicates whether to remove tables without YANG
+                                       from configdbJson directly or create a copy and modify it
```
qiluo-msft, line 1043: Modify configdbJson directly or ...

qiluo-msft:
>            raise SonicYangException("Data Loading Failed\n{}".format(str(e)))

Should we return False according to the function comment?

---
Refers to: src/sonic-yang-mgmt/sonic_yang_ext.py:1073 in 699c9d4. [](commit_id = 699c9d4aad40b5bdd02d0b852d7fb5badea0e438, deletion_comment = False)

PR #9235, opened 2021-11-11
[mellanox]: Disable caching for the issu-version file

saiarcot895:

#### Why I did it

This is to work around the current build issue where Mellanox builds are
failing. This is so that issu-version is always "built", so that copy is
made into the bullseye directory.

#### How I did it

The issu-version file for Mellanox is generated from the Mellanox SDK
libraries. The SDK is installed into a Buster docker container, but the
issu-version file goes onto the base OS, which is Bullseye. To work
around this, the issu-version build rules explicitly copies the
issu-version file to target/files/bullseye/ during the Buster build.

Because of our build infra, if caching is enabled and a cache is being
used, then for issu-version, since it is technically built as part of
Buster, then only target/files/buster/issu-version is saved into the
cache, and target/files/bullseye/issu-version isn't cached. If this
cache gets used, then target/files/bullseye/issu-version is missing, and
the final image build fails.

#### How to verify it

The CI build for the PR should be sufficient to verify it. If the mellanox build there succeeds, then the regular master branch build should succeed as well.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106


yxieca (approved):

lguohan:
mellanox effort: https://github.com/Azure/sonic-buildimage/pull/9234

lguohan:
@alexrallen , i merged this pr to unblock.

Issue #9121, opened 2021-10-29
[kvm][configdb] After building a KVM, a deprecated field still exist /VLAN/Vlan1000/members


ghooo:
#### Description
After building a KVM, config-db has a field that should deprecated/revmoved:
/VLAN/Vlan1000/members

#### Steps to reproduce the issue:
1. Build up a KVM env
2. `show running-config all`
3. look for `members` under `/VLAN/Vlan1000`

#### Output of `show version`:
```sh
admin@vlab-01:~$ show version

SONiC Software Version: SONiC.master.37690-1a2e85248
Distribution: Debian 10.10
Kernel: 4.19.0-12-2-amd64
Build commit: 1a2e85248
Build date: Fri Sep 17 14:17:43 UTC 2021
Built by: AzDevOps@sonic-build-workers-000PC0

Platform: x86_64-kvm_x86_64-r0
HwSKU: Force10-S6000
ASIC: vs
ASIC Count: 1
Serial Number: N/A
Model Number: N/A
Hardware Revision: N/A
Uptime: 20:13:46 up 13 days, 23:47,  1 user,  load average: 1.11, 0.61, 0.41

Docker images:
REPOSITORY                    TAG                      IMAGE ID            SIZE
docker-platform-monitor       latest                   521ca112ce1b        630MB
docker-platform-monitor       master.37690-1a2e85248   521ca112ce1b        630MB
docker-sonic-mgmt-framework   latest                   007c6afa0bef        570MB
docker-sonic-mgmt-framework   master.37690-1a2e85248   007c6afa0bef        570MB
docker-snmp                   latest                   d919e712bdd7        455MB
docker-snmp                   master.37690-1a2e85248   d919e712bdd7        455MB
docker-sonic-telemetry        latest                   88af4994a44d        502MB
docker-sonic-telemetry        master.37690-1a2e85248   88af4994a44d        502MB
docker-dhcp-relay             latest                   a22ce5135236        420MB
docker-lldp                   latest                   942d89861049        453MB
docker-lldp                   master.37690-1a2e85248   942d89861049        453MB
docker-fpm-frr                latest                   6f30a087f8f0        443MB
docker-fpm-frr                master.37690-1a2e85248   6f30a087f8f0        443MB
docker-database               latest                   d7fc3d3a7350        413MB
docker-database               master.37690-1a2e85248   d7fc3d3a7350        413MB
docker-macsec                 latest                   2bcd6845512d        428MB
[...]

Issue #8306, opened 2021-08-02
"show vlan config" command display wrong member information


tomer-israel:
#### Description

"show vlan config" may display wrong information about vlan members after generating vlan members configuration from minigraph and removing vlan members.

the problem is related to a change in the vlan schema that was changed on this PR:
https://github.com/Azure/sonic-utilities/pull/1038
 
In the new schema, we no longer use the “members” field on the VLAN table:
```
    "VLAN": {
        "Vlan1000": {
           ….. 
           "members": [
                "Ethernet8",
                "Ethernet16",
                "Ethernet24",
                "Ethernet32"
            ],
           "vlanid": "1000"
       }
```

And the member info is taken only from vlan members tables:
   ```
 "VLAN_MEMBER": {
        "Vlan1000|Ethernet8": {
            "tagging_mode": "untagged"
        },
        "Vlan1000|Ethernet16": {
            "tagging_mode": "untagged"
        },
        ……
```

We can see that on the new implementation we don’t add the new members to the “members” field on the VLAN table.

In the load_minigraph script we still generate this “members” field on the VLAN table and when we remove member from a vlan it will remove the VLAN_MEMBER table only, the “members” field on the VLAN table is not changing and the member that was removed is still there.

On this scenario when we run show command we see members exist on this vlan although it was removed.. 
We see this problematic situation because the show command is still using this “members” field when it displays the vlan member information. The “members” field info from the VLAN table is added to the VLAN_MEMBER tables info and together it displayed.

#### Steps to reproduce the issue:
1. load minigraph file with vlan members configuration
2. remove the vlan member
3. "show vlan config" will still display the removed vlan member

Example to vlan member configuration on the minigraph xml file:
```
...
      <VlanInterfaces>
        <VlanInterface>
          <Name>Vlan1000</Name>
      <AttachTo>etp2;etp3;etp4;etp5;etp6;etp7;etp8;etp9;etp10;etp11;etp12;etp13;etp14;etp15;etp16;etp17;etp18;etp19;etp20;etp21;etp22;etp23;etp24;etp25</AttachTo>
          <NoDhcpRelay>False</NoDhcpRelay>
          <StaticDHCPRelay>0.0.0.0/0</StaticDHCPRelay>
          <Type i:nil="true"/>
          <DhcpRelays>192.0.0.1;192.0.0.2;192.0.0.3;192.0.0.4</DhcpRelays>
          <VlanID>1000</VlanID>
[...]

PR #9237, opened 2021-11-11
[minigraph] No longer generate the “members” field on the VLAN table in ConfigDB

qiluo-msft:
#### Why I did it
This field is deprecated.
Fixes https://github.com/Azure/sonic-buildimage/issues/8306
Fixes https://github.com/Azure/sonic-buildimage/issues/9121

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106

### DIFF
--- a/src/sonic-config-engine/minigraph.py
+++ b/src/sonic-config-engine/minigraph.py
@@ -531,6 +531,7 @@ def parse_dpg(dpg, hname):
         vlanintfs = child.find(str(QName(ns, "VlanInterfaces")))
         vlans = {}
         vlan_members = {}
+        vlan_member_list = {}
         dhcp_relay_table = {}
         vlantype_name = ""
         intf_vlan_mbr = defaultdict(list)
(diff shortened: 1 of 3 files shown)


qiluo-msft:
@tomer-israel Could you also help review?

theasianpianist:
There may still be some sonic-mgmt tests or utilities that use this field, are we concerned with those?

qiluo-msft:
Then it should be a bug and we need to fix if any. Are you aware of any usage?

ganglyu:
> Then it should be a bug and we need to fix if any. Are you aware of any usage?

test_vlan is still using this field.

qiluo-msft:
@praveen-li Please help review

praveen-li:
> Then it should be a bug and we need to fix if any. Are you aware of any usage?

We would have come to know about sonic-utilities tests at build time itself. As I know there are none, which failed. Looks good to me, we need these changes. Thx for fixing it.

praveen-li (approved):
LGTM

ganglyu (approved):
I have updated sonic-mgmt to use VLAN_MEMBER table.

PR #9248, opened 2021-11-12
[Arista] Increase switch PCIe timeout for 7060

zzhiyuan:
The platform-init, similar to hwsku-init, would be scripts that need to
be called for a specific platform.

#### Why I did it
Arista 7060 platform has a rare and unreproduceable PCIe timeout that could possibly be solved with increasing the switch PCIe timeout value. To do this we'll call a script for this platform to increase the PCIe timeout on boot-up.

No issues would be expected from the setpci command. From the PCIe spec:

"Software is permitted to change the value in this field at any
time. For Requests already pending when the Completion
Timeout Value is changed, hardware is permitted to use either
the new or the old value for the outstanding Requests, and is
permitted to base the start time for each Request either on when
this value was changed or on when each request was issued. "

#### How I did it
Add "platform-init" support in swss docker similar to how "hwsku-init" is called, only this would be for any device belonging to a platform. Then the script would reside in device data folder.

Additionally, add pciutils dependency to docker-orchagent so it can run the setpci commands.

#### How to verify it
On bootup of an Arista 7060, can execute:
lspci -vv -s 01:00.0 | grep -i "devctl2"
In order to check that the timeout has changed.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106

### DIFF
--- a/device/arista/x86_64-arista_7060_cx32s/platform-init
+++ b/device/arista/x86_64-arista_7060_cx32s/platform-init
@@ -1,0 +1,3 @@
+# Increase PCIe timeout value to 210ms
+setpci -s01:00.0 CAP_EXP+0x28.B=6
+setpci -s01:00.1 CAP_EXP+0x28.B=6
--- a/dockers/docker-orchagent/Dockerfile.j2
+++ b/dockers/docker-orchagent/Dockerfile.j2
@@ -19,6 +19,7 @@ RUN apt-get update &&       \
         bridge-utils        \
         conntrack           \
         ndppd               \
+        pciutils            \
         # Needed for installing netifaces Python package
         build-essential     \
         python3-dev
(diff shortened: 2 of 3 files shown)


zzhiyuan:
Testing done:

1. On 7060 platform, changed counter poll interval to 10ms.
2. With 50ms PCIe timeout, saw no errors.
3. With 210ms PCIe timeout, saw no errors.
4. Run script to alternate PCIe timeout from 50ms to 210ms and back again repeatedly with no sleep - in an hour of run saw 18 occurences of "getResAvailableCounters: Failed to get availability for object_type 63"
5. Run script to alternate PCIe timeout from 50ms to 210ms and back again repeatedly and echo the loop count - in several hours of run with 500,000+ loops, did not see any errors.

From these findings I believe there is no danger to changing the timeout on a production device.

sujinmkang (approved):

sujinmkang:
@zzhiyuan is x86_64-arista_7060_cx32s the only platform sku applicable for this change?

sujinmkang (approved):

gechiang:
looks like this fix is already included and picked up in 20191130 branch but missing in 202012 and 20181130 branches.
202205 was later branch after this fix already exists so it has the changes by default.

PR #9273, opened 2021-11-16
[YANG] add YANG model support for CABLE_LENGTH

lolyu:
#### Why I did it
Add YANG model support for table `CABLE_LENGTH`

#### How I did it
1. Add the YANG model file
2. Add the test description file and config file
3. add list `CABLE_LENGTH_LIST` to the `qos_maps_model ` list in `sonic-yang-ext`, as it has an inner list.

#### How to verify it
Build `sonic-yang-model` and `sonic-yang-mgmt` 

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106

### DIFF
--- a/src/sonic-yang-mgmt/sonic_yang_ext.py
+++ b/src/sonic-yang-mgmt/sonic_yang_ext.py
@@ -8,14 +8,17 @@ from json import dump, dumps, loads
 from xmltodict import parse
 from glob import glob
 
-Type_1_list_maps_model = ['DSCP_TO_TC_MAP_LIST',
+Type_1_list_maps_model = [
+    'DSCP_TO_TC_MAP_LIST',
     'DOT1P_TO_TC_MAP_LIST',
     'TC_TO_PRIORITY_GROUP_MAP_LIST',
     'TC_TO_QUEUE_MAP_LIST',
     'MAP_PFC_PRIORITY_TO_QUEUE_LIST',
     'PFC_PRIORITY_TO_PRIORITY_GROUP_MAP_LIST',
     'DSCP_TO_FC_MAP_LIST',
-    'EXP_TO_FC_MAP_LIST']
+    'EXP_TO_FC_MAP_LIST',
+    'CABLE_LENGTH_LIST'
+]
 
 """
 This is the Exception thrown out of all public function of this class.
--- a/src/sonic-yang-models/yang-models/sonic-cable-length.yang
+++ b/src/sonic-yang-models/yang-models/sonic-cable-length.yang
@@ -27,12 +27,12 @@ module sonic-cable-length {
                 leaf name {
                     type string {
                         pattern '[a-zA-Z0-9]{1}([-a-zA-Z0-9_]{0,31})' {
-                            error-message "Invalid cable length table name.";
-                            error-app-tag cable-length-invalid-table-name;    
+                            error-message "Invalid cable length list name.";
+                            error-app-tag cable-length-invalid-list-name;
                         }
                         length 1..32 {
-                            error-message "Invalid length for the cable length table name.";
-                            error-app-tag cable-length-invalid-table-name-length;
+                            error-message "Invalid length for the cable length list name.";
+                            error-app-tag cable-length-invalid-list-name-length;
                         }
                     }
                 }
--- a/src/sonic-yang-models/yang-models/sonic-cable-length.yang
+++ b/src/sonic-yang-models/yang-models/sonic-cable-length.yang
@@ -1,0 +1,61 @@
+module sonic-cable-length {
+
+    yang-version 1.1;
+
+    namespace  "http://github.com/Azure/sonic-cable-length";
+
+    prefix cable-length;
+
+    import sonic-port {
+        prefix port;
+    }
+
+    description "CABLE_LENGTH YANG module for SONiC OS";
+
+    revision 2021-11-11 {
+        description "Initial version";
+    }
+
+    container sonic-cable-length {
+        container CABLE_LENGTH {
+
+            description "CABLE_LENGTH part of config_db.json";
+
+            list CABLE_LENGTH_LIST {
+                key "name";
+
+                leaf name {
+                    type string {
+                        pattern '[a-zA-Z0-9]{1}([-a-zA-Z0-9_]{0,31})' {
+                            error-message "Invalid cable length table name.";
+                            error-app-tag cable-length-invalid-table-name;    
+                        }
+                        length 1..32 {
+                            error-message "Invalid length for the cable length table name.";
+                            error-app-tag cable-length-invalid-table-name-length;
+                        }
+                    }
+                }
+
+                list CABLE_LENGTH {
+                    key "port";
+
+                    leaf port {
+                        type leafref {
+                            path /port:sonic-port/port:PORT/port:PORT_LIST/port:name;
+                        }
+                    }
+
+                    leaf length {
+                        type string {
+                            pattern '[0-9]+m' {
+                                error-message "Invalid cable length.";
+                                error-app-tag cable-length-invalid-length;
+                            }
+                        }
+                    }
+                }
+            }
+        }
+    }
+}
--- a/src/sonic-yang-models/setup.py
+++ b/src/sonic-yang-models/setup.py
@@ -90,6 +90,7 @@ setup(
                          './yang-models/sonic-buffer-port-egress-profile-list.yang',
                          './yang-models/sonic-buffer-profile.yang',
                          './yang-models/sonic-buffer-queue.yang',
+                         './yang-models/sonic-cable-length.yang',
                          './yang-models/sonic-copp.yang',
                          './yang-models/sonic-crm.yang',
                          './yang-models/sonic-device_metadata.yang',
--- a/src/sonic-yang-models/tests/files/sample_config_db.json
+++ b/src/sonic-yang-models/tests/files/sample_config_db.json
@@ -375,6 +375,12 @@
             "20.1.1.1": {},
             "2001:aa:aa::aa": {}
         },
+        "CABLE_LENGTH": {
+            "AZURE": {
+                "Ethernet0": "5m",
+                "Ethernet1": "5m"
+            }
+        },
         "PORT": {
             "Ethernet0": {
                 "alias": "Eth1/1",
--- a/src/sonic-yang-models/tests/yang_model_tests/tests/cable_length.json
+++ b/src/sonic-yang-models/tests/yang_model_tests/tests/cable_length.json
@@ -1,0 +1,13 @@
+{
+    "CABLE_LENGTH_TEST": {
+        "desc": "Configure CABLE_LENGTH table."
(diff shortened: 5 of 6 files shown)


src/sonic-yang-models/yang-models/sonic-cable-length.yang
```diff
@@ -0,0 +1,61 @@
...
+
+                    leaf port {
+                        type leafref {
+                            path /port:sonic-port/port:PORT/port:PORT_LIST/port:name;
+                        }
+                    }
+
+                    leaf length {
+                        type string {
+                            pattern '[0-9]+m' {
```
stephenxs, line 51: Can we put an upper bound on it?
I think the length should be `2..5`, which means the maximum length it represents is `9999m` which should suffice for all scenario

lolyu: updated, thanks!

stephenxs (dismissed):

src/sonic-yang-models/yang-models/sonic-cable-length.yang
```diff
@@ -0,0 +1,65 @@
...
+
+            description "CABLE_LENGTH part of config_db.json";
+
+            list CABLE_LENGTH_LIST {
+                key "name";
+
+                leaf name {
+                    type string {
+                        pattern '[a-zA-Z0-9]{1}([-a-zA-Z0-9_]{0,31})' {
+                            error-message "Invalid cable length table name.";
```
qiluo-msft, line 30: >table [](http://example.com/codeflow?start=64&length=5)

Is it a table or a list?

src/sonic-yang-models/yang-models/sonic-cable-length.yang
qiluo-msft: Where does the maximum come from? Do we have physical limitation somewhere?

src/sonic-yang-models/yang-models/sonic-cable-length.yang
lolyu: it is in a YANG list, updated.

src/sonic-yang-models/yang-models/sonic-cable-length.yang
lolyu: @stephenxs, could you help to clarify? thanks!

stephenxs: Not really.
CABLE_LENGTH is used for lossless traffic for calculating headroom sizes.
Some platforms can have limitations on the headroom sizes of a port but not cable length itself.
In the meanwhile, we need to put an upper bound on the cable length. I think ~10km (9999m) should be large enough for all scenarios.
How do you think @qiluo-msft @neethajohn

src/sonic-yang-models/yang-models/sonic-cable-length.yang
```diff
@@ -0,0 +1,65 @@
...
+    container sonic-cable-length {
+        container CABLE_LENGTH {
+
+            description "CABLE_LENGTH part of config_db.json";
+
+            list CABLE_LENGTH_LIST {
+                key "name";
+
+                leaf name {
+                    type string {
```
venkatmahalingam, line 28: I think, this name is expected to be AZURE now, correct? if yes, let's just put what we're supporting now.

stephenxs: I don’t tend to agree on that. IMO any name should be supported.

neethajohn: don't see "AZURE" used anywhere in the backend except in the unit tests. will that be a problem if we allow any string?

lolyu: I see there isn't any restriction of the hash name in `buffermgr`:
https://github.com/Azure/sonic-swss/blob/d95823d057035e518c38cdea40e3a330e4bb9255/cfgmgr/buffermgr.cpp#L384-L388

stephenxs: I think the logic in the code should be to load the only item in `CABLE_LENGTH` table regardless of its name.
In the vs test, it is assumed that the vs image is deployed via using default configuration rendered from templates in sonic-buildimage. In the templates, the names of the items are `AZURE`.
But a user can configure it to other names in his/her production environment.

src/sonic-yang-mgmt/sonic_yang_ext.py
```diff
@@ -8,12 +8,15 @@
...
 from glob import glob
 
-qos_maps_model = ['DSCP_TO_TC_MAP_LIST',
-            'DOT1P_TO_TC_MAP_LIST',
-            'TC_TO_PRIORITY_GROUP_MAP_LIST',
-            'TC_TO_QUEUE_MAP_LIST',
-            'MAP_PFC_PRIORITY_TO_QUEUE_LIST',
-            'PFC_PRIORITY_TO_PRIORITY_GROUP_MAP_LIST']
+qos_maps_model = [
+    'DSCP_TO_TC_MAP_LIST',
```
venkatmahalingam, line 12: Any reason for adding DSCP_TO_TC_MAP_LIST here?

src/sonic-yang-models/yang-models/sonic-cable-length.yang
venkatmahalingam: Ok, I've not seen any other name usage other than AZURE in the name, also can the CABLE_LENGTH_LIST list have more than one entry? please add the test cases with names other than AZURE.

src/sonic-yang-mgmt/sonic_yang_ext.py
lolyu: I just reorg the indent here, and add `CABLE_LENGTH_LIST` to the list `qos_maps_model`.

src/sonic-yang-models/yang-models/sonic-cable-length.yang
stephenxs: Just FYI, AFAIK it is assumed only one entry in the list and only the first one is handled.

lolyu: @venkatmahalingam, added unit test, please review, thanks!

src/sonic-yang-mgmt/sonic_yang_ext.py
venkatmahalingam: ok

src/sonic-yang-models/yang-models/sonic-cable-length.yang
venkatmahalingam: I've reviewed the new changes, I'm okay to approved as soon as the below comment is addressed.

lolyu: Could you please be more specific about which comment should be addressed? @venkatmahalingam

venkatmahalingam: I'm referring to the comment on the length field.

                    leaf length {
                        type string {
                            pattern '[0-9]+m' {
In the meanwhile, we need to put an upper bound on the cable length. I think ~10km (9999m) should be large enough for all scenarios.

src/sonic-yang-models/yang-models/sonic-cable-length.yang
neethajohn: I hear there can be cable lengths of 120km as well. what should be the upper bound then? @stephenxs, any suggestions

stephenxs: @neethajohn 
I'm not sure whether 120km cable can be supported for lossless traffic because the headroom required, per my calculation, can be as large as 14M...
But if we know for sure that some vendor can support it, or this table will be used for other purposes than lossless traffic in the future, I'm ok to enlarge or even remove the upper bound.

venkatmahalingam: @stephenxs I think, we can keep the upper bound open for now, even if we put any upper value it may not work unless the platform itself supports it.

stephenxs: I’m OK to this approach.

qiluo-msft: I vote for no upper limit.

src/sonic-yang-models/yang-models/sonic-cable-length.yang
```diff
@@ -0,0 +1,65 @@
...
+                        }
+                    }
+
+                    leaf length {
+                        type string {
+                            pattern '[0-9]+m' {
+                                error-message "Invalid cable length.";
+                                error-app-tag cable-length-invalid-length;
+                            }
+                            length 2..5 {
```
qiluo-msft, line 55: >2 [](http://example.com/codeflow?start=35&length=1)

I am confused about the lower limit.
Does `2` means at least 10m cable? Why 9m is not valid?

stephenxs: 2 means a digit plus ‘m’.

src/sonic-yang-mgmt/sonic_yang_ext.py
```diff
@@ -8,12 +8,15 @@
 from xmltodict import parse
 from glob import glob
 
-qos_maps_model = ['DSCP_TO_TC_MAP_LIST',
-            'DOT1P_TO_TC_MAP_LIST',
-            'TC_TO_PRIORITY_GROUP_MAP_LIST',
-            'TC_TO_QUEUE_MAP_LIST',
-            'MAP_PFC_PRIORITY_TO_QUEUE_LIST',
-            'PFC_PRIORITY_TO_PRIORITY_GROUP_MAP_LIST']
+qos_maps_model = [
```
praveen-li, line 11: Approve change in this file.

praveen-li (approved):
Approve Py Lib part. Thx

src/sonic-yang-models/yang-models/sonic-cable-length.yang
neethajohn: @lolyu, since all have agreed to have no upper  bound, please remove this restriction

src/sonic-yang-models/yang-models/sonic-cable-length.yang
lolyu: Updated to remove the upper limit, thank you!

neethajohn:
@lolyu , please resolve merge conflicts

src/sonic-yang-models/yang-models/sonic-cable-length.yang
neethajohn: agreed that we don't need an upper bound now

lolyu:
Conflict resolved with rebase, thanks!

neethajohn (approved):

PR #9274, opened 2021-11-16
Arm64 fixes  and optimizations

saiarcot895:
#### Why I did it

Currently, Bullseye arm64 slave container fails to build in our CI build environment, possibly due to an older version of qemu being used. Try to force registration of the qemu interpreters always to get around this issue.

#### How to verify it

Check to see if the Bullseye arm64 slave container is built.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106

Always register the qemu interpreters in the kernel. The current code doesn't properly check to see if they are registered, so just do it always for now.

Also reduce the number of calls made when referencing variables that should never be changing in a build.


sonic-slave-bullseye/Dockerfile.j2
```diff
@@ -1,5 +1,5 @@
 {%- if CONFIGURED_ARCH == "armhf" and MULTIARCH_QEMU_ENVIRON == "y" %}
-FROM multiarch/qemu-user-static:x86_64-arm-5.2.0-2 as qemu
+FROM multiarch/qemu-user-static:x86_64-arm-6.1.0-8 as qemu
```
qiluo-msft, line 2: >6.1.0-8 [](http://example.com/codeflow?start=43&length=7)

This version is same as below aarch64. Will it always be the same in future? If yes, use a constant?

slave.mk
```diff
@@ -27,12 +27,12 @@ SRC_PATH = src
 RULES_PATH = rules
 TARGET_PATH = target
 DOCKERS_PATH = dockers
-BLDENV = $(shell lsb_release -cs)
+BLDENV := $(shell lsb_release -cs)
```
qiluo-msft, line 30: >:= [](http://example.com/codeflow?start=7&length=2)

Is it fixing a bug? Is it really needed?

saiarcot895: This is just an optimization, so that `lsb_release -cs` is evaluated only once instead of every single time that `BLDENV` is referenced. For native arch builds, the change isn't really noticeable (maybe a second or two), but for arm64 builds on amd64, it appears to reduce the time it takes to run `make configure` by 2 minutes for each distro.

sonic-slave-bullseye/Dockerfile.j2
saiarcot895: I think it makes sense to have it be the same for both armhf and aarch64. It may or may not need changing in the future.

I'll make a variable for this.

Makefile.work
```diff
@@ -185,8 +185,10 @@ ifneq ($(SONIC_DPKG_CACHE_SOURCE),)
 endif
 
 ifeq ($(SONIC_CONFIG_USE_NATIVE_DOCKERD_FOR_BUILD), y)
+ifneq ($(MULTIARCH_QEMU_ENVIRON), y)
```
xumia, line 188: Does MULTIARCH_QEMU_ENVIRON support native docker? If not, can we simply print a wrong usage if MULTIARCH_QEMU_ENVIRON set to y.

xumia: For MULTIARCH_QEMU_ENVIRON=y, native docker will always be used.

DOCKER_RUN += -v /var/run/march/docker.sock:/var/run/docker.sock

xumia (approved):

Makefile.work
saiarcot895: I don't entirely understand the current multiarch code, but from what I can tell, there are actually two native dockerd instances being started up. The first one is for iptables rules to get applied, and the second one is to load the `docker-*.gz` container files. I don't know why there's two separate dockerd instances here and why the one that is already running natively can't be used.

qiluo-msft (dismissed):
Please also check with other reviewers.

sonic-slave-bullseye/Dockerfile.j2
saiarcot895: Docker doesn't like setting an `ARG` before any `FROM` commands, which leaves two alternatives:

* Restructure the armhf and arm64 conditionals to bring in the Debian image first, then set the `ARG`, then bring in the qemu image.
* Make it a Jinja2 variable, and specify the qemu version in `Makefile.work` when parsing the template.

The second might be nicer, since that'll allow setting a single version to be used across all slave images (Stretch, Buster, and Bullseye). If that's fine, I'll change it to that.

saiarcot895:
@qiluo-msft @xumia could you re-review this?

xumia (approved):

judyjoseph:
Please raise a new PR for 202111 branch, as this patch cannot be cleanly cherry-picked.

PR #9075, opened 2021-10-26
[yang] add set_owner to feature yang

arlakshm:
Signed-off-by: Arvindsrinivasan Lakshmi Narasimhan <arlakshm@microsoft.com>

#### Why I did it
Add the configuration for the set_owner in the `feature` yang model

#### How I did it
Add new leaf `set_pwner` to the `feature` yang model

#### How to verify it
compile `sonic_yang_mgmt-1.0-py3-none-any.whl`

#### Which release branch to backport (provide reason below if selected)

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106

PR #9276, opened 2021-11-16
[DPB][YANG] Add POLL_INTERVAL in flex_counter yang model

mykolaxgerasymenko:
#### Why I did it
DPB falls due to missing POLL_INTERVAL in sonic-flex_counter yang model.

#### How I did it
Added POLL_INTERVAL leaf to ACL container in sonic-flex_counter yang model.

#### How to verify it
Run the command config interface breakout <interface> <breakout_mode>

**NOTE:**
To verify this fix, a PR ([add set_owner to feature yang](https://github.com/Azure/sonic-buildimage/pull/9075)) that fix another bug in SONiC should be merged to master.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106


mykolaxgerasymenko:
@stepanblyschak I found that you added this ACL container to sonic-flex_counter yang model.
Please review my changes.

akokhan (approved):

src/sonic-yang-models/yang-models/sonic-flex_counter.yang
```diff
@@ -159,6 +159,11 @@ module sonic-flex_counter {
                 leaf FLEX_COUNTER_DELAY_STATUS {
                     type flex_delay_status;
                 }
+                leaf POLL_INTERVAL {
```
stepanblyschak, line 162: How about adding POLL_INTERVAL leaf to every container on the same level here? Why is this fix specific to ACL?

src/sonic-yang-models/yang-models/sonic-flex_counter.yang
```diff
@@ -159,6 +159,11 @@ module sonic-flex_counter {
...
+                    type uint32 {
+                        range 1..4294967295;
```
stepanblyschak, line 164: I think POLL_INTERVAL range is limited by the CLI. For example - https://github.com/Azure/sonic-utilities/blob/master/counterpoll/main.py#L94.
Do you think we can have the same range here for consistency or you want to relax the constraint of the CLI command?
I am fine with not limiting the upper bound of the polling interval however the lower bound is useful. Setting the counters polling interval to < 100ms will produce a huge load on the CPU.

src/sonic-yang-models/yang-models/sonic-flex_counter.yang
```diff
@@ -159,6 +159,11 @@ module sonic-flex_counter {
...
+                    }
+                }
```
stepanblyschak, line 166: Could you please add UT?

src/sonic-yang-models/yang-models/sonic-flex_counter.yang
mykolaxgerasymenko: I added the POLL_INTERVAL leaf to ACL container, because only the ACL requires this for DPB feature.
But I will try to add this leaf to other containers and test it.

src/sonic-yang-models/yang-models/sonic-flex_counter.yang
mykolaxgerasymenko: I don't want to relax the constraint of the CLI command, I just couldn't find info about the correct range, so I added a range for uint32 type.
If I understand correctly, are there different ranges for different containers?
For example, for ACL we have range: 1000..30000 - https://github.com/Azure/sonic-utilities/blob/master/counterpoll/main.py#L254
Is this correctly?

stepanblyschak: Yes, it is correct

src/sonic-yang-models/yang-models/sonic-flex_counter.yang
stepanblyschak: Do you have other counters enabled in your configuration? I believe if you do, then it will also affect DPB. Don't see anything special about ACL. I think it is better to add it to this PR.

src/sonic-yang-models/yang-models/sonic-flex_counter.yang
mykolaxgerasymenko: UT were added, please review.

src/sonic-yang-models/yang-models/sonic-flex_counter.yang
mykolaxgerasymenko: I added POLL_INTERVAL for other containers. But I didn't find poll intervals for some container here: https://github.com/Azure/sonic-utilities/blob/master/counterpoll/main.py
For example, there is no polling interval range for RIF: https://github.com/Azure/sonic-utilities/blob/master/counterpoll/main.py#L254
Therefore, I did not add the poll interval for containers which I did not find in this file.
Maybe you know where I can find these polling intervals?

stepanblyschak (dismissed):

src/sonic-yang-models/yang-models/sonic-flex_counter.yang
stepanblyschak: @noaOrMlnx Please, comment on the RIF poll interval question

mykolaxgerasymenko:
@praveen-li @lguohan @neethajohn 
Please review.

src/sonic-yang-models/yang-models/sonic-flex_counter.yang
mykolaxgerasymenko: @noaOrMlnx thank you.
But here only the default value of poll_interval for RIF.
Do we need a range of polling interval here for the RIF as well as for other containers? 
Or we can use only default value without range?

src/sonic-yang-models/yang-models/sonic-flex_counter.yang
```diff
@@ -119,6 +149,12 @@ module sonic-flex_counter {
                 leaf FLEX_COUNTER_DELAY_STATUS {
                     type flex_delay_status;
                 }
+                leaf POLL_INTERVAL {
```
praveen-li, line 152: If it stayed in same range and with same default value, then we can have typedef similar as Example:flex_delay_status.

praveen-li (commented):
Looks good overall to me, Kindly address the comments from other reviewers. Thx

src/sonic-yang-models/yang-models/sonic-flex_counter.yang
mykolaxgerasymenko: But we have three different ranges and three different default values for POLL_INTERVAL for different containers.

mykolaxgerasymenko: @praveen-li 
I added a typedef for poll_interval range 1000..30000 and default 10000.
Comments from other reviewers have been resolved.
Please review.

src/sonic-yang-models/yang-models/sonic-flex_counter.yang
```diff
@@ -89,6 +105,12 @@ module sonic-flex_counter {
...
+                    type uint32 {
+                        range 100..30000;
+                    }
+                    default 1000;
```
dgsudharsan, line 112: By setting default this goes into CONFIG_DB and we don't want it which makes it tough to differentiate from user configuration. There was a consensus in YANG subgroup not to have default in yang. Please remove the default.

dgsudharsan (changes requested):

src/sonic-yang-models/yang-models/sonic-flex_counter.yang
mykolaxgerasymenko: @dgsudharsan Do you mean to remove the default for the PORT container or for all containers?

dgsudharsan: The default value for interval is not required. It is handled by application in the backend. Please remove the 'default' that is introduced in this PR

mykolaxgerasymenko: The 'default' values have been removed.

src/sonic-yang-models/yang-models/sonic-flex_counter.yang
mykolaxgerasymenko: @praveen-li I added a typedef for all poll_interval ranges. Please review.

dgsudharsan (dismissed):

src/sonic-yang-models/yang-models/sonic-flex_counter.yang
```diff
@@ -39,6 +57,9 @@ module sonic-flex_counter {
                 leaf FLEX_COUNTER_DELAY_STATUS {
                     type flex_delay_status;
                 }
+                leaf POLL_INTERVAL {
+                    type poll_interval_range_1s_30s;
```
qiluo-msft, line 61: >poll_interval_range_1s_30s [](http://example.com/codeflow?start=25&length=26)

Is there a technical reason that the maximal is 30s? How about just using non-zero uint32?

src/sonic-yang-models/tests/yang_model_tests/tests_config/flex_counter.json
```diff
@@ -3,7 +3,8 @@
         "sonic-flex_counter:sonic-flex_counter": {
             "sonic-flex_counter:FLEX_COUNTER_TABLE": {
                 "BUFFER_POOL_WATERMARK": {
-                    "FLEX_COUNTER_STATUS": "enable"
+                    "FLEX_COUNTER_STATUS": "enable",
```
praveen-li, line 6: Where are we using this test config?

praveen-li: Add test which violates the range please. Thx

src/sonic-yang-models/yang-models/sonic-flex_counter.yang
mykolaxgerasymenko: @qiluo-msft 
It was suggested in the comments to use ranges from this file: https://github.com/Azure/sonic-utilities/blob/master/counterpoll/main.py#L26
When the PR was opened, this range was from 1 to 30 seconds. But two weeks ago, the maximum was increased to the 60 seconds.
About using non-zero uint32: Do you mean using the whole range for uint32: range 1..4294967295?
Is this correct?

qiluo-msft: Yes that's my question. How do you think?

mykolaxgerasymenko: As far as I understand, the POLL_INTERVAL range is limited by the CLI. For example - https://github.com/Azure/sonic-utilities/blob/master/counterpoll/main.py#L94.
I think we can have the same range here for consistency. But we should limit the lower bound.
As for me we can use not limiting the upper bound of the polling interval, however the lower bound is useful. Setting the counters polling interval to < 100ms will produce a huge load on the CPU.
The same was written in the comments: https://github.com/Azure/sonic-buildimage/pull/9276#discussion_r750784662
So, we can use the same following range for all containers: 100..4294967295.
What do you think?

src/sonic-yang-models/yang-models/sonic-flex_counter.yang
mykolaxgerasymenko: Added RIF poll interval

src/sonic-yang-models/tests/yang_model_tests/tests_config/flex_counter.json
mykolaxgerasymenko: @praveen-li 
Added polling intervals to test config.
Added test cases with a valid polling interval and a test which violates the range.
Please review.

src/sonic-yang-models/yang-models/sonic-flex_counter.yang
mykolaxgerasymenko: @qiluo-msft 
Extended poll interval range to max uint32 with limiting lower bound: range 100..4294967295.
Please review.

mykolaxgerasymenko:
@qiluo-msft @praveen-li 
Addresses comments were resolved. Could you please review?

qiluo-msft (dismissed):

qiluo-msft:
Please resolve conflict.

qiluo-msft (approved):

qiluo-msft:
@akokhan, @stepanblyschak, @dgsudharsan Your review conclusions are dismissed due to new commits. Would you like to check again?

dgsudharsan (approved):

stepanblyschak (approved):

mykolaxgerasymenko:
> Please resolve conflict.

Conflict has been resolved.

qiluo-msft:
You mentioned 

> To verify this fix, a PR ([add set_owner to feature yang](https://github.com/Azure/sonic-buildimage/pull/9075)) that fix another bug in SONiC should be merged to master.

Can we merge this PR before that one?

mykolaxgerasymenko:
> You mentioned
> 
> > To verify this fix, a PR ([add set_owner to feature yang](https://github.com/Azure/sonic-buildimage/pull/9075)) that fix another bug in SONiC should be merged to master.
> 
> Can we merge this PR before that one?

@qiluo-msft 
I think we can merge this PR before #9075. These are two independent issues.
But merging can lead to conflicts in 9075 that need to be resolved.
Also we can merge #9418.

arlakshm:
cherry-pick failing in 202106. Please create a new PR

PR #9298, opened 2021-11-18
Added option to enable/disable Static Route Advertisement

abdosi:
What I did:

Added option 'advertise':'true'/'false' to STATIC_TALE to control whether given route need to be advertise into BGP.

How I did:

- Add a route in FRR with tag (1 for advertise and 2 for not advertise)
- Create Route Map to match on tag as 1 and permit
- Attach route-map to `redistribute static`.
- Minigraph option to parse `advertise` option of Static Route.

How I verify:

- Unit Test cases
- Manual Verification Device


abdosi:
@anamehra

src/sonic-bgpcfgd/bgpcfgd/managers_static_rt.py
```diff
@@ -99,44 +102,55 @@ def split_key(key):
...
-            else:
-                op = self.OP_ADD
+        if route_tag != cur_route_tag:
+            for ip_nh in cur_nh_set:
+                op_cmds = op_cmd_list.setdefault(self.OP_DELETE, [])
+                op_cmds.append(self.generate_command(self.OP_DELETE, ip_nh, ip_prefix, vrf, cur_route_tag))
+            for ip_nh in ip_nh_set:
+                op_cmds = op_cmd_list.setdefault(self.OP_ADD, [])
+                op_cmds.append(self.generate_command(self.OP_ADD, ip_nh, ip_prefix, vrf, route_tag))
+        else:
```
shi-su, line 114: I was wondering if there are cases where both `route_tag` and `ip_nh_set` change at the same time? If that is the case, would this if-else check skip nh configuration?

abdosi: @shi-su we are deleting all the previous route/nexthop and creating based on new route/nexthop update. This should make sure we use latest nh configuration.

anamehra (approved):

shi-su (approved):

PR #9301, opened 2021-11-18
td2/td3 change cpu cos num to 10

prsunny:
#### Why I did it
bcm_num_cos=10 forces SDK default to map internal priority 0-7 to COS 0. It is now consistent with other SKUs. 

#### How I did it
Modified config.bcm

#### How to verify it
Verified the Cosq for trapped packets

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012
- [x] 202106


lguohan (approved):

gechiang (commented):
Should we consider S6000 HW SKUs as well?  They are also TD2 based I think...

prsunny:
> Should we consider S6000 HW SKUs as well? They are also TD2 based I think...

@gechiang, its already set as 10 for S6000.

gechiang:
> > Should we consider S6000 HW SKUs as well? They are also TD2 based I think...
> 
> @gechiang, its already set as 10 for S6000.

That's great!  Thanks!

qiluo-msft:

PR #9304, opened 2021-11-18
ACL_RULE can set ACCEPT in packet action but yang model doesn't support that

wen587:
#### Why I did it
ACL_RULE can set ACCEPT in packet action but yang model doesn't support that
#### How I did it
Add ACCEPT to packet_action
#### How to verify it
`sudo config apply-patch acl.json` will not report failure

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106

### DIFF
--- a/src/sonic-yang-models/yang-models/sonic-types.yang
+++ b/src/sonic-yang-models/yang-models/sonic-types.yang
@@ -60,6 +60,7 @@ module sonic-types {
 
     typedef packet_action{
         type enumeration {
+            enum ACCEPT;
             enum DROP;
             enum FORWARD;
             enum REDIRECT;


src/sonic-yang-models/yang-models/sonic-types.yang
```diff
@@ -60,6 +60,7 @@ module sonic-types {
 
     typedef packet_action{
         type enumeration {
+            enum ACCEPT;
```
qiluo-msft, line 63: Is it only used in control plane ACL? If yes, should we limit the enum used by data plane ACL?

wen587: Didn't find any doc talking more about PACKET_ACTION.
[ACL Design](https://github.com/Azure/SONiC/blob/master/doc/acl/ACL-High-Level-Design.mdl) mentioned "FORWARD and DROP" for "L3" tables

qiluo-msft:
Will you close this PR and in favor of https://github.com/Azure/sonic-buildimage/pull/9784?

wen587:
tracked here: https://github.com/Azure/sonic-buildimage/pull/9784

PR #9310, opened 2021-11-18
TACACS: Don't send sshd's bad password to AAA 

renukamanavalan:
[cherry-pick PR #9123 ]

#### Why I did it
When sshd realizes that this login can't succeed due to internal device state
or configuration, instead of failing right there, it proceeds to prompt for
password, so as the user does not get any clue on where is the failure point.

Yet to ensure that this login does not proceed, sshd replaces user provided password
with a specific pattern of characters matching length of user provided password.
This pattern is `"<BS><LF><CR><DEL>INCORRECT`", which is bound to fail.

If user provided length is smaller/equal, the substring of pattern is overwritten.
If user provided length is greater, the pattern is repeated until length is exhausted.

But if the PAM-tacacs plugin would send this password to AAA, the user could get
locked out by AAA, for providing incorrect value.

#### How I did it
Hence this fix, matches obtained password against the pattern. If match, fail just before
reaching AAA server.

#### How to verify it

1. Make sure tacacs is properly configured.
2. Try logging in as, say "user-A"; ensure it succeeds
3. Pick another user, say user-B and ensure this user has not logged into this device before (_look into /etc/passed & folders under /home_)
4. Disable monit service (_as that could fix the issue using disk_check.py_)
5. Start TCP dump for all TACACS servers.
6. Simulate Read-only disk
7. Try logging in using  user-B.
8. Verify it fails, after 3 attempts
9. Stop tcp dump.
10. TCP dump should show "authentication" for user-A only

- [ ] 201811
- [x] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106

### DIFF
--- a/src/tacacs/pam/Makefile
+++ b/src/tacacs/pam/Makefile
@@ -20,6 +20,7 @@ $(addprefix $(DEST)/, $(MAIN_TARGET)): $(DEST)/% :
 	git apply ../0004-management-vrf-support.patch
 	git apply ../0005-pam-Modify-parsing-of-IP-address-and-port-number-to-.patch
 	git apply ../0006-Add-support-for-source-ip-address.patch
+	git apply ../0007-handle-bad-password-set-by-sshd.patch
 
 	dpkg-buildpackage -rfakeroot -b -us -uc -j$(SONIC_CONFIG_MAKE_JOBS) --admindir $(SONIC_DPKG_ADMINDIR)
 	popd
--- a/src/tacacs/pam/0007-handle-bad-password-set-by-sshd.patch
+++ b/src/tacacs/pam/0007-handle-bad-password-set-by-sshd.patch
@@ -1,0 +1,133 @@
+From f2687e7a442c83e19190695021fb9a60fe07ba60 Mon Sep 17 00:00:00 2001
+From: Renuka Manavalan <remanava@microsoft.com>
+Date: Wed, 17 Nov 2021 02:31:45 +0000
+Subject: [PATCH] handle bad password set by sshd
+
+---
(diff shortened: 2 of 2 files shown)


lguohan:
better to keep the same description as the original pr

src/tacacs/pam/Makefile
```diff
@@ -20,6 +20,7 @@ $(addprefix $(DEST)/, $(MAIN_TARGET)): $(DEST)/% :
 	git apply ../0004-management-vrf-support.patch
 	git apply ../0005-pam-Modify-parsing-of-IP-address-and-port-number-to-.patch
 	git apply ../0006-Add-support-for-source-ip-address.patch
+	git apply ../0007-handle-bad-password-set-by-sshd.patch
```
qiluo-msft, line 23: Keep it as 0010 as the master commit? Then later if we decide to backport more patches, others will be easier.

renukamanavalan: Not really. As master has diverged well with 3 patches, this patch had to be manually re-created. The contents of 0010 & 0007 are different but for the same purpose. Similarly, in future, if we would need any patch from master it has to be re-created manually. The title of the patch would indicate the patch purpose, which is same in both master & here.

qiluo-msft: Sorry for misleading. Is it possible to keep original file name `../0010-handle-bad-password-set-by-sshd.patch`. I am okay if the file content changes.

renukamanavalan: I got your point. 
The number is just meant for sequence of application. So here, this patch has to be applied after 0006, hence 0007.

qiluo-msft (approved):

PR #9331, opened 2021-11-20
[Build]: Fix the version not found issue

xumia:
#### Why I did it
When we update the a sai package downing from a remote server, we need to update the version file as well currently, but the reproducible build feature is not enabled in master, it can only be detected when merging the code into the release branches, such as 202106, 202012, etc.
The reproducible feature is to reduce the build failure, not need to break the build when the version not specified. If version not specified, the best choice is to accept the version from remote server.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012
- [x] 202106


qiluo-msft (approved):

liushilongbuaa (approved):

yxieca:
This change is in 202205 since day 1.

PR #9362, opened 2021-11-24
[Build]: Enable arm pr check

xumia:
#### Why I did it
Support marvell-armhf dpkg cache and the azp check.
Waiting for merging PR https://github.com/Azure/sonic-buildimage/pull/9381 to 202012 branch, so only azp template change in this PR.

Move the VS build to a new stage BuildVS, change the Test stage only depending on BuildVS, running the BuildVS and the other platform's build in parallel. The Test stage do not has dependency on the marvel-armhf build, reduce the overall build time caused by longer build time of marvel-armhf build.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -31,7 +31,7 @@ resources:
     endpoint: build
 
 stages:
-- stage: Build
+- stage: BuildVS
   pool: sonicbld
   variables:
     CACHE_MODE: rcache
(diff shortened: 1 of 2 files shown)


azure-pipelines.yml
```diff
@@ -43,10 +43,29 @@ stages:
       buildOptions: 'USERNAME=admin SONIC_BUILD_JOBS=$(nproc) ${{ variables.VERSION_CONTROL_OPTIONS }}'
       jobGroups:
       - name: vs
+
+- stage: Build
+  pool: sonicbld
+  dependsOn: []
```
qiluo-msft, line 49: Is it fixing something? do we need it?

azure-pipelines.yml
```diff
@@ -31,7 +31,7 @@ resources:
     endpoint: build
 
 stages:
-- stage: Build
+- stage: BuildVS
```
qiluo-msft, line 34: >BuildVS [](http://example.com/codeflow?start=9&length=7)

I think you extracted vs from below section, what is the motivation? I see previous code simpler.

azure-pipelines.yml
```diff
@@ -43,10 +43,29 @@ stages:
       buildOptions: 'USERNAME=admin SONIC_BUILD_JOBS=$(nproc) ${{ variables.VERSION_CONTROL_OPTIONS }}'
       jobGroups:
```
qiluo-msft, line 44: I see these lines repeated in below stage. Can we reuse?

azure-pipelines.yml
xumia: @qiluo-msft , it is to declare the build stage does not depend on any other stage.
A new stage BuildVS is added, the Test stage only depends on the BuildVS stage.
The Build stage, including broadcom, mellanox, marvell-armhf, can be run in parallel with BuildVS stage.
The Build stage will take more time than before, since marvel-armhf will take longer time to build.

azure-pipelines.yml
xumia: I can check if we can reuse it, it depends on if the variables can be used another variable in compiling time.

qiluo-msft (approved):
With minor comment on reuse, others lgtm.

lguohan:
do you know why marvell_armhf build still take 5 hours for the pr build?

xumia:
> do you know why marvell_armhf build still take 5 hours for the pr build?

No dpkg cache now, We need to merge the PR, and do a official build, then we have dpkg cache.

lguohan (approved):

PR #9379, opened 2021-11-25
[sonic-utilities] update submodule

stepanblyschak:
```
c05845d Add trap flow counter support (#1868)
ef82f00 [load_minigraph] Delay pfcwd start until the buffer templates are rendered (#1937)
f5e5a56 [sonic-package-manager] support sonic-cli-gen and packages with YANG model (#1650)
64777a4 generic_config_updater: Filename changed & VLAN validator added (#1919)
1f8f6ab [config reload] Update command reference (#1941)
```

#### Why I did it

To pick new commits from utilities.

#### How I did it
Updated submodule.

#### How to verify it
Build.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106


stepanblyschak:
@lguohan @liat-grozovik  This submodule update brings in the YANG and CLI autogeneration support to sonic-package-manager. Since we are executing some sonic-package-manager operations at build time (to install dhcp-relay and generate shutdown order files) and the sonic-package-manager instantiates ConfigMgmt class it requires that at least the /etc/sonic/init_cfg.json is aligned with YANG models, otherwise it is going to fail. This is a good check as discussed with @praveen-li and will block PRs that add new data into init_cfg.json without the schema for that data. But right now init_cfg.json is not aligned. This PR requires this fix https://github.com/Azure/sonic-buildimage/pull/9276 to be merged in order for checkers to pass.

PR #9486, opened 2021-12-09
[bgp] Enable BGP Graceful Restart based on device role

abdosi:
What I did:
Updated Jinja Template to enable BGP Graceful Restart based on device role. By default it will be enable only if the device role type is `TorRouter`. 

Why I did:-
By default FRR is configured in Graceful Helper mode. Graceful Restart is needed on T0/TorRouter only since the device can go for warm-reboot. For T1/LeafRouter it need to be in Helper mode only.  For T2  and above we can disable Graceful Restart feature completely support by FRR 7.4 onwards (https://github.com/FRRouting/frr/pull/5210). However As per FRR current code it send BGP EOR packet only if Peer has advertised Graceful Restart or Helper capability. And as mention in BGP RFC https://datatracker.ietf.org/doc/html/rfc4724 sending of BGP EOR is better for overall network convergance so we are having BGP Helper mode on for all role other than `T0/TorRouter`

Trigger for this change was eg:
a)  if we cold-reboot the T1 router T0 will not withdraw T1 routes till Graceful Timer is not expired and this causes traffic blackhole for that duration.
b) For Chassis iBGP session to detect BGP peer down and withdraw routes immediately

How I verify:
1.  Updated Unit Test case.
2. Manual Verification.
3. We will create test for this in `sonic-mgmt` 

```
On Instance0 GR commands configured with Restart Timer as 240 sec and Instance1 has no GR configuration 

1.	We can see Instance1 by default goes into helper mode

•	Instance 0 : show ip bgp neighbor O/P
                                                                                                                                                                               
  Graceful restart information:
    End-of-RIB send: IPv4 Unicast
    End-of-RIB received: IPv4 Unicast
    Local GR Mode: Restart*
    Remote GR Mode: Helper
    R bit: False
    Timers:
      Configured Restart Time(sec): 240
      Received Restart Time(sec): 120
    IPv4 Unicast:
      F bit: False
      End-of-RIB sent: Yes
      End-of-RIB sent after update: No
      End-of-RIB received: Yes
      Timers:
        Configured Stale Path Time(sec): 360
        Configured Selection Deferral Time(sec): 45

•	Instance 1 : show ip bgp neighbor O/P

  Graceful restart information:
    End-of-RIB send: IPv4 Unicast
    End-of-RIB received: IPv4 Unicast
    Local GR Mode: Helper*
    Remote GR Mode: Restart
    R bit: True
    Timers:
      Configured Restart Time(sec): 120
      Received Restart Time(sec): 240
    IPv4 Unicast:
      F bit: True
      End-of-RIB sent: Yes
      End-of-RIB sent after update: Yes
      End-of-RIB received: Yes
      Timers:
        Configured Stale Path Time(sec): 360

2.	When the Helper BGP instance goes down (Instance 1) Restart BGP Instance (Instance0) remove the route immediately.

3.	When the Restart BGP instance goes down (Instance 0) Helper BGP Instance (Instance1) remove the route after 4 min (240 sec as configured).

```


abdosi:
> For T2 role, should we add `bgp graceful-restart-disable ` config to disable helper mode as well?

@arlakshm PR description has reason of not doing that as per current understanding. In future we can reference this change again.

arlakshm (approved):
lgtm

yxieca (approved):

abdosi:
@yxieca @qiluo-msft  can we take this for 202012.

abdosi:
@qiluo-msft created 202012 PR: https://github.com/sonic-net/sonic-buildimage/pull/12990

PR #9493, opened 2021-12-10
Sync SONiC host's timezone to all containers

gord1306:
Sync the content of /etc/timezone and /etc/localtime from the SONiC host to all containers
during their own systemd services be started.

#### Why I did it
The containers are not awared the timezone change in the SONiC host, it may cause some info 
are inconsistent e.g. time gap between syslog and swss.rec/sairedis.rec

#### How I did it
In the systemd scripts of services, then timezone of containers will be sync in the postStartAction()
and therefore the containers will have same timezone after their own services are restarted

#### How to verify it
Change the timezone in the SONiC host and then exec the sudo config reload -y
To check the containers timezone are changed

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012
- [x] 202106


bsun-sudo:
If there is a runtime change of timezone in the host, would all running containers pick up the new timezone automatically?

gord1306:
> If there is a runtime change of timezone in the host, would all running containers pick up the new timezone automatically?

No, the timezone would not be applied automatically. It is only applied during service starting. 
Users can use the command - e.g. `systemctl restart systemctl restart sonic.target` to make the timezone be sync to the containers

bsun-sudo:
> `systemctl restart systemctl restart sonic.target`

Do you mean a docker has to be restarted to sync its timezone with the host's timezone?

gord1306:
> > `systemctl restart systemctl restart sonic.target`
> 
> Do you mean each docker need to be restarted to sync its timezone with the host's timezone? Or only restart the services in each docker use the timezone?

Yes, the script would be executed while the service is restarted and the restarted operation includes restarting the container. It is triggered by using the `systemctl restart` command. If you are asking for the `docker restart` command only, the answer would be no. The script is not triggered to execute by the `docker restart` command

gord1306:
Closed PR due to there is similar implementation https://github.com/sonic-net/sonic-buildimage/pull/14000

PR #9429, opened 2021-12-02
sonic-bgp-cmn/admin_status fixed

MaratGubaiev:
#### Why I did it
fixing https://github.com/Azure/sonic-buildimage/issues/9350 "[yang-models] BGP neighbor admin_status should be up/down"

#### Which release branch to backport (provide reason below if selected)

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106

PR #9496, opened 2021-12-10
Generic updater

MaratGubaiev:
#### Why I did it
fixing https://github.com/Azure/sonic-utilities/issues/1935 "Running "sudo config apply-patch ..." in latest (as of 11/15) master image applied in SONiC switches fails with inability to parse config for given yang models."

#### How I did it
If fact I just checked and applied [the updates](https://github.com/Azure/sonic-utilities/files/7557303/yang_updated.tar.gz) provided by @renukamanavalan with some changes.

#### How to verify it
Perform `sudo config apply-patch <some_patch>.json-patch`
Patch should be applied and there should be no errors*.
<details><summary>Example of a dummy patch</summary>

```
[
    { "op": "replace", "path": "/PORT/Ethernet124/admin_status", "value": "up" }
]

```
</details>
*NOTE that errors might be caused by improper config too.<br><br><br><br>

@renukamanavalan @praveen-li @prsunny @zhangyanzhao 
I think that despite the changes fix the issue, they are not complete.
Could you provide any documentation to make YANG models correspond to actual config?

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106

### DIFF
--- a/src/sonic-yang-models/yang-models/sonic-bgp-neighbor.yang
+++ b/src/sonic-yang-models/yang-models/sonic-bgp-neighbor.yang
@@ -3,6 +3,10 @@ module sonic-bgp-neighbor {
     prefix bgpnbr;
     yang-version 1.1;
 
+    import sonic-types {
+        prefix stypes;
+    }
+
     import ietf-inet-types {
         prefix inet;
     }
@@ -58,7 +62,7 @@ module sonic-bgp-neighbor {
                 }
 
                 leaf admin_status {
-                    type string;
+                    type stypes:admin_status;
                     description "Admin status";
                 }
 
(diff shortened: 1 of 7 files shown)


src/sonic-yang-models/yang-models/sonic-bgp-neighbor.yang
```diff
@@ -61,6 +61,11 @@ module sonic-bgp-neighbor {
                     description "BGP Neighbor address";
                 }
 
+                leaf admin_status {
```
qiluo-msft, line 64: >admin_status [](http://example.com/codeflow?start=21&length=12)

You just change the leaves' order. Is it really needed?

qiluo-msft:
We fixed many yang issues, for example https://github.com/Azure/sonic-buildimage/pull/9429.
Could you check the latest code and reduce the scope for truly needed modification?

MaratGubaiev:
> Could you check the latest code and reduce the scope for truly needed modification?

Yes, sure: I have seen that many fixes are already done and merging them with my changes have made a mess. I will re-check it and commit again.

MaratGubaiev:
@qiluo-msft 
I created new pull request https://github.com/Azure/sonic-buildimage/pull/9517

PR #9504, opened 2021-12-10
[rsyslog]Setting log file size to 16Mb

dgsudharsan:
Signed-off-by: Sudharsan Dhamal Gopalarathnam <sudharsand@nvidia.com>

#### Why I did it
The existing log file size in sonic is 1 Mb. Over a period of time this leads to huge number of log files which becomes difficult for monitoring applications to handle.
Instead of large number of small files, the size of the log file is not set to 16 Mb which reduces the number of files over a period of time.

#### How I did it
Changed the size parameter and related macros in logrotate config for rsyslog

#### How to verify it
Execute logrotate manually and verify the limit when the file gets rotated.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [x] 202111


liat-grozovik (dismissed):

liat-grozovik:
@lguohan could you please approve this fix?

lguohan:
i think we will have issue when disk size is small, can this be a template? keep 1M for disk size /var/log partition less than 200MB.

yxieca:
@dgsudharsan Do we need any 'default' configuration before the logrotate config was rendered? Have you confirmed that the next logrotate will pick up the new config after rendering? Can you also add a sonic-mgmt test for this?

dgsudharsan:
> @dgsudharsan Do we need any 'default' configuration before the logrotate config was rendered? Have you confirmed that the next logrotate will pick up the new config after rendering? Can you also add a sonic-mgmt test for this?

Yes I tested this scenario  logrotate picks up after rendered. Its only few seconds difference between cron and rendering. 
There are multiple scenarios
1) Onie install - there will be no rsyslog config in logrotate and so it will not be considered for logrotate. However the next time logrotate runs after 10 minutes, there will be rsyslog logrotate config which will be used. 
2) Upgrade - I believe the conf files under /etc would be copied during upgrade and old conf file would be used before rendering and after will pick up the new one
3) reboot - The conf file will already be rendered before reboot and hence that will be used. 

Regarding sonic-mgmt, I don't think there are any test cases although the suite covers logrotate as a utility for all existing tests. 
Do you recommend to write a test case for covering logrotate functionality? If yes I believe that needs to be considered as a separate effort.

yxieca:
> Regarding sonic-mgmt, I don't think there are any test cases although the suite covers logrotate as a utility for all existing tests. Do you recommend to write a test case for covering logrotate functionality? If yes I believe that needs to be considered as a separate effort.

@dgsudharsan thanks for the details. Yes please track a different effort and add a sonic-mgmt test.

yxieca (approved):

yxieca (approved):

PR #9510, opened 2021-12-12
[Mellanox] Exclude CPU board and switch board sensors on all systems beside SN2201

liorghub:
#### Why I did it
2 new sensors that were added for SN2201 needed to be excluded for all other systems but it was missed.

#### How I did it
I excluded the relevant sensors.

#### How to verify it
Run show platform temperature

- [x] 202111

This PR is fixing issue in "show platform temperature", now the right sensors are being showed.


liat-grozovik (approved):

liorghub:
@Junchao-Mellanox Is it possible to add true only for Alligator and have false as default for all other systems?

PR #9527, opened 2021-12-14
Reduce disk space for image

saiarcot895:
#### Why I did it

Some platforms have low amounts of disk storage (4GB of storage, for example), and with the current image, only one image 
 (maybe two if there's no large files being stored) can be effectively stored on disk. Make some changes in the squashfs compression used for the base filesystem and in the docker containers to save some space.

#### How I did it

1. For base filesystem, change the squashfs compression format from gzip to zstd with 1MB block size. zstd achieves much higher compression ratios for a small increase in decompression time. (When the zstd code in the kernel gets updated to current code, it should have higher decompression speeds compared to gzip.)
2. Create a standalone docker layer that just has swss. This is because there are 5 containers that are each installing swss, and so this takes up 5x the amount of space needed.
3. For docker-platform-monitor, remove Python 2 support entirely. With Bullseye, the platform scripts must be in Python 3, since Python 2 support is gone. Because of how we're bringing in the platform scripts from the base filesystem into the container, this change will also affect the container.

#### How to verify it

A Broadcom image should take about 250 MB less disk space with these changes.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106


build_debian.sh
```diff
@@ -319,8 +319,6 @@ sudo LANG=C DEBIAN_FRONTEND=noninteractive chroot $FILESYSTEM_ROOT apt-get -y in
     python3-pip             \
     python-is-python3       \
     cron                    \
-    libprotobuf23           \
-    libgrpc++               \
```
yxieca, line 323: @vdahiya12 do we need to keep libgrpc?

yxieca (dismissed):
LGTM except I don't know if we can remove the grpc lib. Please wait for Vaibhav's comments.

build_debian.sh
vdahiya12: @yxieca we dont need this for the new driver, But sonic-telemetry might need this. @hui-ma to comment.

saiarcot895: For reference, this was added as part of a commit adding support for Nokia IXR7250E and to support the Nokia NDK about 2 months ago. From what I can tell, the grpcio python package is needed (and installed in the pmon docker container), but the C/C++ libraries aren't needed there.

vdahiya12: I agree @mlok-nokia , could you please clarify why you install the mentioned grpc packages in host?

mlok-nokia: The Nokia platform device manager is running the host which requires grpc and protobuf libs.

saiarcot895: @mlok-nokia is this just python scripts or is there a C/C++ library? If just python scripts, I don't see anything in the documentation or the quick start guide that says the C/C++ libraries also need to be installed.

mlok-nokia: Both C/C++ and python.

saiarcot895: Got it, I'll revert this change.

One things is that the ++ is being treated as regex, so the development headers are being installed as well. I'm assuming the development headers aren't needed at runtime, so I'll make sure the runtime libraries are installed.

mlok-nokia: Thanks, The headers are not needed.

yxieca (approved):

lguohan:
i suggest to combine last two commits, so that we can keep your original commits and not use squash merge.

saiarcot895:
> i suggest to combine last two commits, so that we can keep your original commits and not use squash merge.

Done

lguohan (approved):

prgeor (approved):

build_debian.sh
```diff
@@ -581,7 +582,7 @@ sudo rm -f $ONIE_INSTALLER_PAYLOAD $FILESYSTEM_SQUASHFS
 sudo du -hsx $FILESYSTEM_ROOT
 sudo mkdir -p $FILESYSTEM_ROOT/var/lib/docker
 scripts/collect_host_image_version_files.sh $TARGET_PATH $FILESYSTEM_ROOT
-sudo mksquashfs $FILESYSTEM_ROOT $FILESYSTEM_SQUASHFS -e boot -e var/lib/docker -e $PLATFORM_DIR
+sudo mksquashfs $FILESYSTEM_ROOT $FILESYSTEM_SQUASHFS -comp zstd -b 1M -e boot -e var/lib/docker -e $PLATFORM_DIR
```
jusherma, line 585: What was the purpose of changing block size from default of 128K to 1M? It seems like it takes more memory at build time with this change

Issue #9534, opened 2021-12-14
[yang-models] find_data_dependencies by 'must' statement not only 'leafref'


ghooo:
#### Description

The problem is current implementation of [sonic-yang-mgmt::find_data_dependencies](https://github.com/Azure/sonic-buildimage/blob/f2774b635dee2d5c0e30489632b47e0210a712dd/src/sonic-yang-mgmt/sonic_yang.py#L518) does not get referrers if they are using `must` statement, it has to use `leafref`.

Investigate getting referrers by `must` statements as well.

PR #9535, opened 2021-12-14
[yang-model] Using 'leafref' instead of 'must' for loopback

ghooo:
#### Why I did it
Fix issue https://github.com/Azure/sonic-utilities/issues/1962

The problem is current implementation of [sonic-yang-mgmt::find_data_dependencies](https://github.com/Azure/sonic-buildimage/blob/f2774b635dee2d5c0e30489632b47e0210a712dd/src/sonic-yang-mgmt/sonic_yang.py#L518) does not get referrers if they are using `must` statement, it has to use `leafref`.

For now we can convert `must` to `leafref` if possible. In the future we will investigate get referrers by `must` statements as well https://github.com/Azure/sonic-buildimage/issues/9534

#### How I did it
Instead of `must` use `leafref`

#### How to verify it
unit-test

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106

### DIFF
--- a/src/sonic-yang-models/tests/yang_model_tests/tests/loopback.json
+++ b/src/sonic-yang-models/tests/yang_model_tests/tests/loopback.json
@@ -1,7 +1,7 @@
 {
-    "LOOPBACK_IPPREFIX_PORT_MUST_CONDITION_FALSE": {
-	"desc": "Loopback Ip-prefix port-name must condition failure.",
-	"eStrKey" : "Must"
+    "LOOPBACK_IPPREFIX_PORT_FOR_NON_EXIST_INTERFACE": {
+	"desc": "Configure Loopback Ip-prefix for non-existing Loopback interface.",
+	"eStrKey" : "LeafRef"
     },
     "LOOPBACK_INTERFACE_WRONG_NAT_ZONE_RANGE": {
         "desc": "Configure wrong value for nat zone.",
--- a/src/sonic-yang-models/tests/yang_model_tests/tests/loopback.json
+++ b/src/sonic-yang-models/tests/yang_model_tests/tests/loopback.json
@@ -1,7 +1,7 @@
 {
     "LOOPBACK_IPPREFIX_PORT_FOR_NON_EXIST_INTERFACE": {
-	"desc": "Configure Loopback Ip-prefix for non-existing Loopback interface.",
-	"eStrKey" : "LeafRef"
+        "desc": "Configure Loopback Ip-prefix for non-existing Loopback interface.",
+        "eStrKey" : "LeafRef"
     },
     "LOOPBACK_INTERFACE_WRONG_NAT_ZONE_RANGE": {
         "desc": "Configure wrong value for nat zone.",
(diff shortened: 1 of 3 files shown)


src/sonic-yang-models/tests/yang_model_tests/tests/loopback.json
```diff
@@ -1,7 +1,7 @@
 {
-    "LOOPBACK_IPPREFIX_PORT_MUST_CONDITION_FALSE": {
-	"desc": "Loopback Ip-prefix port-name must condition failure.",
-	"eStrKey" : "Must"
+    "LOOPBACK_IPPREFIX_PORT_FOR_NON_EXIST_INTERFACE": {
+	"desc": "Configure Loopback Ip-prefix for non-existing Loopback interface.",
```
qiluo-msft, line 3: >	 [](http://example.com/codeflow?start=0&length=1)

Replace tab with 4 spaces. Please format the whole file's indentation.

ghooo: updated

qiluo-msft (approved):
LGTM. Please wait a little longer for more eyes.

src/sonic-yang-models/yang-models/sonic-loopback-interface.yang
```diff
@@ -63,12 +63,11 @@ module sonic-loopback-interface {
...
 
                 leaf name{
-                    /* This node must be present in LOOPBACK_INTERFACE_LIST */
-                    must "(current() = ../../LOOPBACK_INTERFACE_LIST[name=current()]/name)"
-                    {
-                        error-message "Must condition not satisfied, Try adding lo<>: {}, Example: 'lo1': {}";
+                    description "Loopback interface name";
+
+                    type leafref {
+                        path "../../LOOPBACK_INTERFACE_LIST/name";
```
praveen-li, line 69: Lets keep error-msg, something as below: 

error-message "Loopback VRF interface must exist, Try adding lo<>: {}, Example: 'lo1': {}";
description "Loopback interface name";

praveen-li: plz see if we must do similar changes in VLAN_INTERFACE and in INTERFACE. Thx

ganglyu (approved):

src/sonic-yang-models/yang-models/sonic-loopback-interface.yang
ghooo: I haven't seen `error-message` used with leafref before in any sonic yang models. I think it is not supported:
```
libyang[0]: Invalid keyword "error-message". (path: /sonic-loopback-interface:sonic-loopback-interface/LOOPBACK_INTERFACE/LOOPBACK_INTERFACE_IPPREFIX_LIST/name)
```

Tried like this:
```
                leaf name{
                    error-message "Loopback VRF interface must exist, Try adding lo<>: {}, Example: 'lo1': {}";

                    description "Loopback interface name";

                    type leafref {
                        path "../../LOOPBACK_INTERFACE_LIST/name";
                    }
                }
```

and this:
```
                leaf name{
                    description "Loopback interface name";

                    type leafref {
                        error-message "Loopback VRF interface must exist, Try adding lo<>: {}, Example: 'lo1': {}";
                        path "../../LOOPBACK_INTERFACE_LIST/name";
                    }
                }
```

ghooo: > plz see if we must do similar changes in VLAN_INTERFACE and in INTERFACE. Thx

Yes, we might need to fix these as well. I will start first by checking if we can can the referrers from must statements #9534

PR #9566, opened 2021-12-16
[Mellanox] [202012] fail the build when hw-mgmt patches do not apply

stepanblyschak:
Taken from https://github.com/Azure/sonic-buildimage/pull/9539

####  Why I did it
To fix an issue that hw-mgmt patches were not applied. One patch was already in upstream hw-mgmt package thus applying it again caused an error and no other patches were applied. Also, I did it to improve the Makefile, so that the make will fail in case patches fail to apply.

####  How I did it
Removed obsolete patch, made applying patches a hard failure in the build.

####  How to verify it
Run the make and verify patches are applied.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106

### DIFF
--- a/platform/mellanox/hw-management/Makefile
+++ b/platform/mellanox/hw-management/Makefile
@@ -6,7 +6,8 @@ MAIN_TARGET = hw-management_1.mlnx.$(MLNX_HW_MANAGEMENT_VERSION)_amd64.deb
 $(addprefix $(DEST)/, $(MAIN_TARGET)): $(DEST)/% :
 
 	pushd hw-mgmt
-	git am ../*.patch
+	git stash
+	git apply -3 ../*.patch || exit 1
 	chmod +x ./debian/rules
 	KVERSION=$(KVERSION) dpkg-buildpackage -us -uc -b -rfakeroot -j$(SONIC_CONFIG_MAKE_JOBS) --admindir $(SONIC_DPKG_ADMINDIR)
 	popd


liat-grozovik (approved):

platform/mellanox/hw-management/Makefile
```diff
@@ -6,7 +6,8 @@ MAIN_TARGET = hw-management_1.mlnx.$(MLNX_HW_MANAGEMENT_VERSION)_amd64.deb
 $(addprefix $(DEST)/, $(MAIN_TARGET)): $(DEST)/% :
 
 	pushd hw-mgmt
-	git am ../*.patch
+	git stash
```
qiluo-msft, line 9: What is the purpose of `git stash`? Do you intend to `git reset --hard`?

stepanblyschak: The idea was that if I develop a patch for hw-mgmt and have some changes that are not commited the stash will save them.

liat-grozovik:
@qiluo-msft could you please refer to the answer to  your question if all good, approve and merge?

platform/mellanox/hw-management/Makefile
qiluo-msft: This will impact the whole local repo. Assume other developers modify some file in this repo locally, and build locally, if hw-managment is also built in the process, the local changes will be stashed, and it is surprising.

qiluo-msft: I notice the same issue on master branch. So if you agree, please fix on master first.

stepanblyschak: @qiluo-msft  If we look at other 3rd party projects built with patches in SONiC, e.g libteam the whole libteam is deleted every time you build which might also be a surprise if modifying libteam locally. ```git reset --hard``` would be even more destructive because local changes are not saved.

qiluo-msft: Sorry I just notice `git stash` in submodule, not in the sonic-buildimage repo. Then it will mostly not impact other developers.

qiluo-msft (approved):

PR #9573, opened 2021-12-17
Support SONiC OpenSSL FIPS 140-3 based on SymCrypt engine

xumia:
#### Why I did it
Support OpenSSL FIPS 140-3, see design doc: https://github.com/Azure/SONiC/blob/master/doc/fips/SONiC-OpenSSL-FIPS-140-3.md.

#### How I did it
Install the fips packages.
To build the fips packages, see https://github.com/Azure/sonic-fips
Azure pipelines: https://dev.azure.com/mssonic/build/_build?definitionId=412

#### How to verify it

Validate the SymCrypt engine:
```
admin@sonic:~$ dpkg-query -W | grep openssl
openssl 1.1.1k-1+deb11u1+fips
symcrypt-openssl        0.1

admin@sonic:~$ openssl engine -v | grep -i symcrypt
(symcrypt) SCOSSL (SymCrypt engine for OpenSSL)
admin@sonic:~$ 
```

Install debug version of symcrypt and validate:
```
$ gdb --args openssl rand 10 
(gdb) b scossl_rand_bytes
(gdb) r
Starting program: /usr/bin/openssl rand 10

Breakpoint 1, scossl_rand_bytes (buf=0x7ffd81ce49f0 "200\024\t\300\004X\020\211\263!", num=10)
    at /src/SymCrypt-OpenSSL/SymCryptEngine/src/scossl_rand.c:23
23          SymCryptRandom(buf, num);
```

Test golang: when the fips is enabled, the test.go will call go boring api, call openssl api, and then call symcrypt api.
```
test.go:
import (
    "crypto/rand"
    "crypto/rsa"
)
func main() {
    rsa.GenerateKey(rand.Reader, 2048)
}
```
```
$ gdb test
(gdb) b scossl_rsa_keygen
(gdb) r
Thread 1 "test" hit Breakpoint 1, scossl_rsa_keygen (rsa=0x12299b0, bits=2048, e=0x122e490, cb=0x0) at /src/SymCrypt-OpenSSL/SymCryptEngine/src/scossl_rsa.c:638
638         PBYTE   ppbPrimes[2] = { 0 };

(gdb) bt
#0  scossl_rsa_keygen (rsa=0x12299b0, bits=2048, e=0x122e490, cb=0x0) at /src/SymCrypt-OpenSSL/SymCryptEngine/src/scossl_rsa.c:638
#1  0x00007f4e966a8446 in RSA_generate_key_ex (rsa=0x12299b0, bits=2048, e_value=0x122e490, cb=0x0) at ../crypto/rsa/rsa_gen.c:35
#2  0x00000000004025aa in crypto/internal/boring(.text) ()
...
#8  0x0000000000401c0e in _cgo_b67541c0c44c_Cfunc__goboringcrypto_RSA_generate_key_fips ()
```

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106


files/build_templates/sonic_debian_extension.j2
```diff
@@ -590,6 +590,8 @@ sudo dpkg --root=$FILESYSTEM_ROOT -i {{deb}} || sudo LANG=C DEBIAN_FRONTEND=noni
 {% endfor %}
 {% endif %}
 
+sudo LANG=C DEBIAN_FRONTEND=noninteractive chroot $FILESYSTEM_ROOT apt-get --fix-broken install -y
```
qiluo-msft, line 593: >--fix-broken [](http://example.com/codeflow?start=75&length=12)

Why need this line?

rules/sonic-fips.mk
```diff
@@ -0,0 +1,41 @@
+# fips packages
+
+FIPS_VERSION = 0.1
+FIPS_OPENSSH_VERSION = 8.4p1-5+fips
+FIPS_PYTHON_MAIN_VERSION = 3.9
+FIPS_PYTHON_VERSION = 3.9.2-1+fips
+FIPS_GOLANG_MAIN_VERSION = 1.15
+FIPS_GOLANG_VERSION = 1.15.15-1~deb11u2+fips
+BLDENV ?= $(shell lsb_release -cs)
```
qiluo-msft, line 9: >BLDENV [](http://example.com/codeflow?start=0&length=6)

This is already used by slave.mk. Do you reassign?

rules/sonic-fips.mk
```diff
@@ -0,0 +1,41 @@
...
+FIPS_URL_PREFIX = https://sonicstorage.blob.core.windows.net/public/fips/$(BLDENV)/$(FIPS_VERSION)
+
+SYMCRYPT_OPENSSL_NAME = symcrypt-openssl
+SYMCRYPT_OPENSSL = fips/$(SYMCRYPT_OPENSSL_NAME)_$(FIPS_VERSION)_$(CONFIGURED_ARCH).deb
+$(SYMCRYPT_OPENSSL)_SRC_PATH = $(SRC_PATH)/sonic-fips
```
qiluo-msft, line 14: >sonic-fips [](http://example.com/codeflow?start=43&length=10)

Do you want to add a new submodule in this PR?

Makefile.work
```diff
@@ -133,13 +133,17 @@ endif
 endif
 SLAVE_IMAGE = $(SLAVE_BASE_IMAGE)-$(USER_LC)
 
+ifneq ($(PLATFORM_ARCH),amd64)
+    override INCLUDE_FIPS := n
```
qiluo-msft, line 137: >INCLUDE_FIPS [](http://example.com/codeflow?start=13&length=12)

How about `ENABLE_FIPS`?

slave.mk
```diff
@@ -93,7 +93,7 @@ ifneq ($(CONFIGURED_PLATFORM),generic)
 endif
 
 configure :
-	@mkdir -p $(DEBS_PATH)
+	@mkdir -p $(DEBS_PATH)/fips
```
qiluo-msft, line 96: >fips [](http://example.com/codeflow?start=24&length=4)

Why use a subfolder?

rules/sonic-fips.mk
xumia: The variable is not exported in slave.mk, we use the same way with the slave.mk. May be better to export it in slave.mk.

rules/sonic-fips.mk
xumia: No, we will use binaries only, it will take long time to build it.

files/build_templates/sonic_debian_extension.j2
xumia: I double check the build, it can be removed, thanks.
+ sudo LANG=C DEBIAN_FRONTEND=noninteractive chroot ./fsroot-vs apt-get --fix-broken install -y
Reading package lists...
Building dependency tree...
Reading state information...
The following packages were automatically installed and are no longer required:
  dirmngr gir1.2-packagekitglib-1.0 gnupg gnupg-l10n gnupg-utils gpg-agent
  gpg-wks-client gpg-wks-server gpgsm iso-codes libcurl3-gnutls libksba8
  libnpth0 libpackagekit-glib2-18 lsb-release pinentry-curses python-pip-whl
  python3-dbus python3-pycurl python3-setuptools python3-software-properties
  python3-wheel
Use 'sudo apt autoremove' to remove them.
**0 upgraded, 0 newly installed, 0 to remove and 6 not upgraded.**

Makefile.work
xumia: Sure, I will change to use ENABLE_FIPS.

saiarcot895: Is there a reason to use override here? If it's because FIPS modules aren't supported on anything besides amd64, then it's better to error out rather than silently disabling the FIPS option.

rules/sonic-fips.mk
saiarcot895: If these packages need to be updated (for fixing security issues, or features/functionality), is there an easy way to rebuild the FIPS packages?

sonic-slave-bullseye/Dockerfile.j2
```diff
@@ -397,6 +397,13 @@ RUN sudo augtool --autosave "set /files/etc/dpkg/dpkg.cfg/force-confold"
 RUN apt-get -y build-dep linux
 
 # For gobgp and telemetry build
+{%- if INCLUDE_FIPS == "y" %}
+RUN wget -O golang-go.deb 'https://sonicstorage.blob.core.windows.net/public/fips/bullseye/0.1/golang-1.15-go_1.15.15-1~deb11u2%2Bfips_{{ CONFIGURED_ARCH }}.deb' \
```
qiluo-msft, line 401: >1.15 [](http://example.com/codeflow?start=102&length=4)

Can we use the same version as in `else` branch? Ideally, we should share the version constant, and build any version specified with FIPS.

rules/sonic-fips.mk
qiluo-msft: If they are the same thing, better to export in slave.mk

saiarcot895: The `BLDENV` variable should be available in this context. It's being used in `rules/snmpd.mk` for example, so the export shouldn't be needed.

rules/sonic-fips.mk
xumia: Yes, see https://github.com/Azure/sonic-fips.
I will add the artifacts publish automation later, currently, copy the artifacts to storage container manually.

rules/sonic-fips.mk
xumia: Fixed, thank.

sonic-slave-bullseye/Dockerfile.j2
xumia: Change all to use 1.15 for bullseye, align to the default one in bullseye.
And change to use the debian packages, not to download from google web site.

Makefile.work
xumia: Fixed, if not supported, print error message.

slave.mk
xumia: It is to save to the fips packages in folder, some of the packages, such as openssh-server, has two different versions, fips&none-fips one, better to place in different folders.
It will break the existing installation script (dpkg -i openssh-server*.dep), if different version of packages in the same folder.

sonic-slave-bullseye/Dockerfile.j2
```diff
@@ -397,20 +397,13 @@ RUN sudo augtool --autosave "set /files/etc/dpkg/dpkg.cfg/force-confold"
...
-{%- if CONFIGURED_ARCH == "armhf" %}
- && wget https://storage.googleapis.com/golang/go$VERSION.linux-armv6l.tar.gz \
- && tar -C /usr/local -xzf go$VERSION.linux-armv6l.tar.gz \
-{%- elif CONFIGURED_ARCH == "arm64" %}
- && wget https://storage.googleapis.com/golang/go$VERSION.linux-arm64.tar.gz \
- && tar -C /usr/local -xzf go$VERSION.linux-arm64.tar.gz \
-{%- else %}
- && wget https://storage.googleapis.com/golang/go$VERSION.linux-amd64.tar.gz \
- && tar -C /usr/local -xzf go$VERSION.linux-amd64.tar.gz \
+RUN apt-get install -y golang-1.15
```
qiluo-msft, line 400: >golang-1.15 [](http://example.com/codeflow?start=23&length=11)

In bulleyes package server, golang (metapackage) is roughly same as golang-1.15. Should we just install golang?

rules/sonic-fips.mk
qiluo-msft: which storage? Do you mean FIPS_URL_PREFIX? I notice the name difference: sonic-fips vs fips.

xumia: Yes, the same some of the other artifacts, sonicstorage.

xumia:
@saiarcot895 , could you please help review again?

slave.mk
```diff
@@ -327,6 +329,7 @@ $(info "INCLUDE_P4RT"                    : "$(INCLUDE_P4RT)")
 $(info "INCLUDE_KUBERNETES"              : "$(INCLUDE_KUBERNETES)")
 $(info "INCLUDE_MACSEC"                  : "$(INCLUDE_MACSEC)")
 $(info "INCLUDE_MUX"                     : "$(INCLUDE_MUX)")
+$(info "ENABLE_FIPS_FEATUR               : "$(ENABLE_FIPS_FEATURE)")
```
saiarcot895, line 332: Typo and missing closing quote here

rules/config
```diff
@@ -221,3 +221,12 @@ ENABLE_ASAN ?= n
...
 DEFAULT_CONTAINER_REGISTRY ?=
+
+# ENABLE_FIPS_FEATURE - support FIPS feature, only for amd64 or arm64, armhf not supported yet
+# ENABLE_FIPS - support FIPS flag, if enabled, no additional config requred for the image to support FIPS
+ifeq ($(CONFIGURED_ARCH), armhf)
+ENABLE_FIPS_FEATURE ?= n
+else
+ENABLE_FIPS_FEATURE ?= y
+endif
+ENABLE_FIPS ?= n
```
saiarcot895, line 232: So `ENABLE_FIPS_FEATURE` affects build-time, and causes the FIPS binaries to get installed into the image. `ENABLE_FIPS` affects runtime, and tells the FIPS binaries to actually use FIPS. Is this correct?

If so, maybe renaming `ENABLE_FIPS` to `ENABLE_FIPS_AT_RUNTIME` may explain it better? Also, this flag would depend on `ENABLE_FIPS_FEATURE` to be `y`, so adding a check for that would be good.

files/build_templates/sonic_debian_extension.j2
```diff
@@ -625,6 +620,11 @@ sudo dpkg --root=$FILESYSTEM_ROOT -P {{ debname }}
 {% endfor %}
 {% endif %}
 
+# Remove sshd host keys, and will regenerate on first sshd start. This needs to be
+# done again here because our custom version of sshd is being installed, which
+# will regenerate the sshd host keys.
+sudo rm -f $FILESYSTEM_ROOT/etc/ssh/ssh_host_*_key*
```
qiluo-msft, line 626: Why move this statement?

xumia: When installing openssl/openssh for fips, the ssh key will be generated again, so moving the removal step after all installing steps are complete.

qiluo-msft: As my understanding, in original code, removal step is after all  openssl/openssh installing steps.

If you have a good reason, is it a general bug fix?

installer/x86_64/install.sh
```diff
@@ -666,6 +666,11 @@ else # install_env = "onie"
     fi
 fi
 
+# Add extra linux command line
+extra_cmdline_linux=%%EXTRA_CMDLINE_LINUX%%
+echo "EXTRA_CMDLINE_LINUX=$extra_cmdline_linux"
+ONIE_PLATFORM_EXTRA_CMDLINE_LINUX="$ONIE_PLATFORM_EXTRA_CMDLINE_LINUX $extra_cmdline_linux"
```
qiluo-msft, line 672: >ONIE_PLATFORM_EXTRA_CMDLINE_LINUX [](http://example.com/codeflow?start=0&length=33)

Do not combine %%EXTRA_CMDLINE_LINUX%% with ONIE_PLATFORM_EXTRA_CMDLINE_LINUX. Just add %%EXTRA_CMDLINE_LINUX%% to below menuentry template.

qiluo-msft: Sorry, by semantics, GRUB_CMDLINE_LINUX should include %%EXTRA_CMDLINE_LINUX%%

xumia: Thanks, fixed, better to use GRUB_CMDLINE_LINUX .

rules/config
xumia: @saiarcot895 , thanks for your comment. Yes, the ENABLE_FIPS affects the runtime, it will enable the fips by default without any config. In the last community review, it is requested to add the feature. If the value is not set, we have to set change the Linux parameter in the sonic image installation or upgrade time.
Looks like ENABLE_FIPS_AT_RUNTIME is more clear, but a little strange, not necessary to emphasize it, how about using ENABLE_FIPS without change, but if you still think better to change to ENABLE_FIPS_AT_RUNTIME, I can change it.

For the ENABLE_FIPS check, I have added a check in Makefile.work, if the ENABLE_FIPS is set, but ENABLE_FIPS_FEATURE is not set, the build will be failed.

slave.mk
xumia: Fixed, thanks.

rules/config
```diff
@@ -224,3 +224,12 @@ ENABLE_ASAN ?= n
 
 # reset default container registry from dockerhub to other
 DEFAULT_CONTAINER_REGISTRY ?=
+
+# ENABLE_FIPS_FEATURE - support FIPS feature, only for amd64 or arm64, armhf not supported yet
+# ENABLE_FIPS - support FIPS flag, if enabled, no additional config requred for the image to support FIPS
+ifeq ($(CONFIGURED_ARCH), armhf)
```
qiluo-msft, line 230: >CONFIGURED_ARCH [](http://example.com/codeflow?start=8&length=15)

`CONFIGURED_ARCH` is defined in Makefile.work. Ideally this file should only contain simple configurable variable, not depend on others. If you want to override, you may do in Makefile.work

files/build_templates/sonic_debian_extension.j2
xumia: It is not a bug, there is an additional installation step to install debian packages, the FIPS packages are installed in the step, see https://github.com/Azure/sonic-buildimage/blob/53e5fe6a93b051c8319d51e109d317beb5166e77/files/build_templates/sonic_debian_extension.j2#L598

We move the removal step, because the FIPS packages will generate the ssh keys again.

rules/config
xumia: Qi, thanks for comment. Agree, move the logic to the Makefile.work, and keep rule/config simple and clean.

qiluo-msft: Will you remove these line in this file?

rules/config
saiarcot895: Ok, I'm fine with leaving the names as-is.

rules/config
xumia: Removed, thanks.

qiluo-msft (approved):

xumia:
@saiarcot895 , could you please help review it again? Thanks.

saiarcot895 (approved):

PR #9596, opened 2021-12-20
[build] Fix docker-sonic-mgmt pylint dependency lazy-object-proxy version

liushilongbuaa:
#### Why I did it
https://pypi.org/project/lazy-object-proxy/#description
lazy-object-proxy's latest version does not support python2, while pylint denpends on lazy-object-proxy.

1.7.1 (2021-12-15)
Removed most of the Python 2 support code and fixed python_requires to require at least Python 3.6.

Note that 1.7.0 has been yanked because it could not install on Python 2.7. Installing lazy-object-proxy on Python 2.7 should automatically fall back to the 1.6.0 release now.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106

### DIFF
--- a/dockers/docker-sonic-mgmt/Dockerfile.j2
+++ b/dockers/docker-sonic-mgmt/Dockerfile.j2
@@ -56,6 +56,7 @@ RUN pip install cffi==1.10.0 \
                 psutil \
                 pyasn1==0.1.9 \
                 pyfiglet \
+                lazy-object-proxy==1.6.0 \
                 pylint==1.8.1 \
                 pyro4 \
                 pysnmp==4.2.5 \


dockers/docker-sonic-mgmt/Dockerfile.j2
```diff
@@ -56,6 +56,7 @@ RUN pip install cffi==1.10.0 \
                 psutil \
                 pyasn1==0.1.9 \
                 pyfiglet \
+                lazy-object-proxy==1.6.0 \
```
xumia, line 59: Can we upgrade pylint to fix it? So no need to install lazy-object-proxy explicitly.

liushilongbuaa: I have tried pylint 1.81~1.9.5, but failed.
seems pylint not specific lazy-object-proxy version.

xumia (approved):

PR #9600, opened 2021-12-21
[telemetry] Roll over streaming telemetry secrets by ACMS.

yozhao101:

#### Why I did it
Initially the secrets of streaming telemetry container is managed by dSMS and deployed by Hardware Proxy. Since the secrets were not always be able to be deployed successfully on production device by Hardware Proxy, we decide to leverage the `acms` agent running in ACMS container to manage and deploy the secrets.

#### How I did it
I mainly did the following two changes: **First** I added a script in streaming telemtry container and this script will check in background to see whether the certificate and key were rolled over and updated by `acms` agent or not. The streaming telemetry container will be restarted if and only if the certificate and key were actually rolled over. **Second**, In the script `telemetry.sh`, the telemetry server process will only be started if and only if the certificate and key files did exist on the device. Otherwise, it will check the certificate and key files periodically.

#### How to verify it
I did this implementation and tested them on virtual switch.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106

### DIFF
--- a/dockers/docker-sonic-telemetry/certificate_rollover_checker
+++ b/dockers/docker-sonic-telemetry/certificate_rollover_checker
@@ -1,0 +1,139 @@
+#!/usr/bin/env python3
+
+"""
+certificate_rollover_checker
+
+This script will be leveraged to periodically check whether the certificate and private key
+files of streaming telemetry were rolled over by dSMS service or not. The streaming telemetry
+container will be restarted if the certificate and private key are rolled over by dSMS service
+and then updated by ACMS agent running in ACMS container.
+"""
+
+import os
+import signal
+import sys
+import syslog
+import time
+
+from swsscommon import swsscommon
+
+CERTIFICATE_CHECKING_INTERVAL_SECS = 3600
+
+
+def get_file_last_mod_time(file_path):
+    """Gets the last modification time of a specific file.
+
+    Args:
+        file_path: A string represents the file path.
+
+    Returns:
+        last_mod_time: A float number in seconds represents the last moditification time of file
+                       since epoch.
+    """
+    last_mod_time = 0.0
+
+    try:
+        last_mod_time = os.path.getmtime(file_path)
+    except OSError as error:
+        syslog.syslog(syslog.LOG_ERR,
+                      "Could not get last modification time of the file and error message is '{}'.".format(error))
+        sys.exit(1)
+
+    return last_mod_time
+
+
+def restart_streaming_telemetry():
+    """Restarts the streaming telemetry container by terminating the root process.
+
+    Args:
+        None
+
+    Returns:
+        None
+    """
+    root_process_pid = os.getppid()
+    syslog.syslog(syslog.LOG_INFO,
+                  "Restarting streaming telemetry service by terminating the process with pid: '{}'".format(root_process_pid))
+    os.kill(root_process_pid, signal.SIGTERM)
+    sys.exit(0)
+
+
+def certificate_rollover_check():
+    """Checks certificate and key files and restart streaming telemetry contianer if necessary.
+
+    Checks the last modification time of certificate and private key files of streaming telemetry
+    to see whether they were already rolled over by dSMS service and updated by ACMS agent running
+    in ACMS container. The streaming telemetry container will be restarted if they were rolled over.
+
+    Args:
+        None
+
+    Returns:
+        None
+    """
+    certificate_path = ""
+    private_key_path = ""
+    certificate_last_mod_time = 0
+    private_key_last_mod_time = 0
+
+    config_db = swsscommon.DBConnector("CONFIG_DB", 0)
+    telemetry_table = swsscommon.Table(config_db, "TELEMETRY")
+    telemetry_table_keys = telemetry_table.getKeys()
+    if "certs" in telemetry_table_keys:
+        certs_info = dict(telemetry_table.get("certs")[1])
+        if "server_crt_acms" in certs_info and "server_key_acms" in certs_info:
+            certificate_path = certs_info["server_crt_acms"]
+            private_key_path = certs_info["server_key_acms"]
+            syslog.syslog(syslog.LOG_INFO, "Path of certificate file is '{}'".format(certificate_path))
+            syslog.syslog(syslog.LOG_INFO, "Path of key file is '{}'".format(private_key_path))
+        else:
+            syslog.syslog(syslog.LOG_ERR,
+                          "Failed to retrieve the path of certificate and key file from 'TELEMETRY' table!")
+            sys.exit(2)
+    else:
+        syslog.syslog(syslog.LOG_ERR,
+                      "Failed to retrieve the certificate information from 'TELEMETRY' table!")
+        sys.exit(3)
+
+    while True:
+        if not os.path.exists(certificate_path) or not os.path.exists(private_key_path):
+            syslog.syslog(syslog.LOG_ERR,
+                          "Certificate or key file did not exist on device and sleep '{}' seconds to check again {} ...".format(CERTIFICATE_CHECKING_INTERVAL_SECS))
+            time.sleep(CERTIFICATE_CHECKING_INTERVAL_SECS)
+        else:
+            break
+
+    certificate_last_mod_time = get_file_last_mod_time(certificate_path)
+    private_key_last_mod_time = get_file_last_mod_time(private_key_path)
+
+    while True:
+        certificate_mod_time = get_file_last_mod_time(certificate_path)
+        private_key_mod_time = get_file_last_mod_time(private_key_path)
+        syslog.syslog(syslog.LOG_INFO,
+                      "Last modification time of certificate file is: '{}'".format(time.ctime(certificate_last_mod_time)))
+        syslog.syslog(syslog.LOG_INFO,
+                      "Last modification time of key file is: '{}'".format(time.ctime(private_key_last_mod_time)))
+
+        if (certificate_mod_time > certificate_last_mod_time
+           or private_key_mod_time > private_key_last_mod_time):
+            syslog.syslog(syslog.LOG_INFO,
+                          "Last modification time of certificate file is changed to '{}': ".format(time.ctime(certificate_mod_time)))
+            syslog.syslog(syslog.LOG_INFO,
+                          "Last modification time of key file is changed to '{}': ".format(time.ctime(private_key_mod_time)))
+            syslog.syslog(syslog.LOG_INFO,
+                          "Secrets were rolled over and restarting streaming telemetry service ...")
+            restart_streaming_telemetry()
+
+        # Wait for specified seconds and then do the next round checking
+        syslog.syslog(syslog.LOG_INFO,
+                      "Sleeping '{}' seconds before doing the next round rollover checking ...".format(CERTIFICATE_CHECKING_INTERVAL_SECS))
+        time.sleep(CERTIFICATE_CHECKING_INTERVAL_SECS)
+
+
+def main():
+    certificate_rollover_check()
+
+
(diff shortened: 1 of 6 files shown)


dockers/docker-sonic-telemetry/certificate_rollover_checker
```diff
@@ -0,0 +1,136 @@
...
+    Args:
+        None
+
+    Returns:
+        None
+    """
+    root_process_pid = os.getppid()
+    syslog.syslog(syslog.LOG_INFO,
+                  "Restarting streaming telemetry service by terminating the process with pid: '{}'".format(root_process_pid))
+    os.kill(root_process_pid, signal.SIGTERM)
```
qiluo-msft, line 55: >SIGTERM [](http://example.com/codeflow?start=37&length=7)

The modern convention is to use sighup (ref: [https://stackoverflow.com/a/28327659/2514803](https://stackoverflow.com/a/28327659/2514803) ).

The benefit is not to explicitly terminate the other process and trigger critical process monitor alerts.

You may need to add the sighup handler in sonic-telemetry if not supported.

files/build_templates/sonic_debian_extension.j2
```diff
@@ -127,6 +127,10 @@ sudo rm -rf $FILESYSTEM_ROOT/$REDIS_DUMP_LOAD_PY3_WHEEL_NAME
 # Install Python module for psutil
 sudo https_proxy=$https_proxy LANG=C chroot $FILESYSTEM_ROOT pip3 install psutil
 
+# Install Python module for inotify
+sudo https_proxy=$https_proxy LANG=C chroot $FILESYSTEM_ROOT pip3 install inotify
+
+
```
qiluo-msft, line 133: Remove extra empty line.

dockers/docker-sonic-telemetry/certificate_rollover_checker
```diff
@@ -0,0 +1,136 @@
...
+            restart_streaming_telemetry()
+
+        # Wait for specified seconds and then do the next round checking
+        syslog.syslog(syslog.LOG_INFO,
+                      "Sleeping '{}' seconds before doing the next round rollover checking ...".format(CERTIFICATE_CHECKING_INTERVAL_SECS))
+        time.sleep(CERTIFICATE_CHECKING_INTERVAL_SECS)
+
+
+def main():
+    certificate_rollover_check()
```
qiluo-msft, line 131: >certificate_rollover_check [](http://example.com/codeflow?start=4&length=26)

Could you try inotify ([https://www.linuxjournal.com/content/linux-filesystem-events-inotify](https://www.linuxjournal.com/content/linux-filesystem-events-inotify) ) instead of reinvent the wheel?

qiluo-msft (changes requested):
As comments

files/build_templates/sonic_debian_extension.j2
yozhao101: Updated.

dockers/docker-sonic-telemetry/certificate_rollover_checker
yozhao101: Updated.

dockers/docker-sonic-telemetry/certificate_rollover_checker
yozhao101: I think we can probably use the `supervisorctl restart telemetry` command to only restart the streaming telemetry server process once the secrets were rotated. This can avoid triggerring the critical process alerts.

qiluo-msft: Restarting is using `SIGTERM` and `SIGKILL` internally. One big concern is graceful shutdown. Considering the client will often fetch large amount of data, graceful shutdown will make client easier.

yozhao101: I got the point. We can do this by either sending the signal `SIGKILL` or the command `supervisorctl restart telemetry`. However, our main focus is how we can do some cleanup before gracefully stopping the telemetry server process and disconnecting with gNMI client side.

dockers/docker-sonic-telemetry/certificate_rotation_checker
```diff
@@ -0,0 +1,225 @@
...
+            private_key_file_rotated = True
+
+        if certificate_file_rotated and private_key_file_rotated:
+            certificate_file_rotated = False
+            private_key_file_rotated = False
+            syslog.syslog(syslog.LOG_INFO,
+                          "Certificate and private key were rotated and restarting telemetry server process ...")
+            restart_telemetry_server()
+
+            # Wait for specified seconds and then do the next round checking
```
qiluo-msft, line 155: >Wait for specified second [](http://example.com/codeflow?start=14&length=25)

You don't need sleep in an event loop.

dockers/docker-sonic-telemetry/certificate_rotation_checker
```diff
@@ -0,0 +1,225 @@
...
+        None.
+    """
+    certificate_file_rotated = False
+    private_key_file_rotated = False
+
+    inotify_instance = inotify.adapters.Inotify()
+    inotify_instance.add_watch(CREDENTIALS_DIR_PATH)
+    for event in inotify_instance.event_gen(yield_nones=False):
+        header, event_type, monitoring_path, file_name = event
+        if (file_name == certificate_file_name
```
qiluo-msft, line 141: >if [](http://example.com/codeflow?start=8&length=2)

The FSM logic is complex and may be messed up by some input sequence. Could you use one file as the main indicator, and always rotate if that file changed.

ref: [https://github.com/Azure/sonic-restapi/blob/94805a39ac0712219f7dc08faa2cfdbf371dd177/go-server-server/main.go#L103](https://github.com/Azure/sonic-restapi/blob/94805a39ac0712219f7dc08faa2cfdbf371dd177/go-server-server/main.go#L103)

dockers/docker-sonic-telemetry/certificate_rotation_checker
yozhao101: Initially what I am thinking is since the directory `/etc/sonic/credentials/` will store all the certificates and keys for different applications in the future and if one or several certificates under this directory are rotated too frequently (by accident?), this event loop will be executed continuously and then burn too much CPU cycles potentially.  Another thinking the polling frequency of our internal tool to get the update from certificate storage is 8 ~ 24 hours. That's why I put a sleep function at here.

I will let `inotify` to monitor the certificate file directly instead of monitoring the `/etc/sonic/credentials/` and then we can remove this sleep function in event loop.

dockers/docker-sonic-telemetry/certificate_rotation_checker
yozhao101: Updated and use the rotation of certificate file as the main indicator.

dockers/docker-sonic-telemetry/certificate_rotation_checker
qiluo-msft: Got your motivation, and the considering may be helpful. However, event you sleep, the event will be queued and processed after each sleep, you don't save too much?

dockers/docker-sonic-telemetry/certificate_rotation_checker
qiluo-msft: Please make sure describe the main file in document? This is very critical design assumption and the cert rotator should treat it as a contract.

dockers/docker-sonic-telemetry/certificate_rotation_checker
```diff
@@ -0,0 +1,205 @@
...
+        certificate_file_name: A string indicates the telemetry certificate file name.
+
+    Returns:
+        None.
+    """
+    certificate_file_rotated = False
+
+    inotify_instance = inotify.adapters.Inotify()
+    inotify_instance.add_watch(CREDENTIALS_DIR_PATH)
+    for event in inotify_instance.event_gen(yield_nones=False):
```
qiluo-msft, line 136: >inotify_instance [](http://example.com/codeflow?start=17&length=16)

In extreme case, the file is deleted by a malicious user, will the inotify_instance still working? I think its link to inode, and deleting file will destroy the inode.

If this is true, a crash is better than a dead loop.

dockers/docker-sonic-telemetry/certificate_rotation_checker
```diff
@@ -0,0 +1,205 @@
...
+        else:
+            syslog.syslog(syslog.LOG_ERR,
+                          "Failed to retrieve the path of certificate file from 'TELEMETRY' table!")
+            sys.exit(3)
+    else:
+        syslog.syslog(syslog.LOG_ERR,
+                      "Failed to retrieve the certificate information from 'TELEMETRY' table!")
+        sys.exit(4)
+
+    while True:
```
qiluo-msft, line 181: >while True: [](http://example.com/codeflow?start=4&length=11)

The intention of this loop is to wait until two files exist. why only check one file?

yozhao101: Updated and checks the existence of both two files.

dockers/docker-sonic-telemetry/certificate_rotation_checker
yozhao101: Updated the design document.

dockers/docker-sonic-telemetry/certificate_rotation_checker
yozhao101: The `inotify` is `inode` based and it will monitor the credentials directory `/etc/sonic/credentials/` to see whether the telemetry certificate file was rotated or not. If certificate file was deleted by accidentally, the `inotify_instance` will not be impacted.

I updated the PR to log an error message if the certificate was deleted. What I am thinking is if the certificate was restored later, then it can be treated as a kind of `rotation` operation and the telemetry server will be restarted by this script.

dockers/docker-sonic-telemetry/certificate_rotation_checker
yozhao101: Agree and this will not save too much!

liat-grozovik:
@yozhao101 could you please make sure your PR is rebased to latest and rerun checkers?
this fix is needed for 202111 and we appreciate if it can be ready for signoff.

liat-grozovik:
@yozhao101 could you please rebase your work on latest and rerun checkers?
@qiluo-msft could you review?

yozhao101:
@qiluo-msft Can you help me review this PR please?

dockers/docker-sonic-telemetry/certificate_rotation_checker
```diff
@@ -0,0 +1,214 @@
...
+                          "Failed to retrieve the path of certificate or private key file from 'TELEMETRY' table!")
+            sys.exit(1)
+    else:
+        syslog.syslog(syslog.LOG_ERR,
+                      "Failed to retrieve the certificate information from 'TELEMETRY' table!")
+        sys.exit(2)
+
+    while True:
+        if not os.path.exists(certificate_file_path) or not os.path.exists(private_key_file_path):
+            syslog.syslog(syslog.LOG_ERR,
```
qiluo-msft, line 192: >LOG_ERR [](http://example.com/codeflow?start=33&length=7)

Instead of error message and keep loop, it is better to just simply kill telemetry daemon.

liat-grozovik:
@yozhao101 any update on this PR? should it go to 202205? if so, can you add the proper label?

zhangyanzhao:
@yozhao101 can you please help to response the comments and provide an update on this PR? Thanks.

jeflo:
Hi @yozhao101 this looks like a very useful change. Would like to know what is the status of this PR and what's the plan to get it to master? Thanks.

PR #9604, opened 2021-12-21
[Build]: Improve the linux kernel build cache hit rate

xumia:
#### Why I did it
Improve the linux kernel build cache hit rate.
Current the the hit rate is around 85.8% (based on the last 3 month, 3479 PR builds totally, 494 PR build not hit).
We can improve the hit rate up to 95% or better.
The Linux kernel build will take really long time, most of the PRs are nothing to do with the kernel change. The remaining cache options should be enough to detect the Linux kernel cache status (dirty or not).

#### How I did it
Some of the file changes are not relative to the Linux kernel build, such as sonic-slave-stretch/Dockerfile.j2, we can exclude it.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106

### DIFF
--- a/rules/linux-kernel.dep
+++ b/rules/linux-kernel.dep
@@ -1,7 +1,6 @@
 
 SPATH       := $($(LINUX_HEADERS_COMMON)_SRC_PATH)
-DEP_FILES   := $(SONIC_COMMON_FILES_LIST) rules/linux-kernel.mk rules/linux-kernel.dep
-DEP_FILES   += $(SONIC_COMMON_BASE_FILES_LIST)
+DEP_FILES   := rules/linux-kernel.mk rules/linux-kernel.dep
 SMDEP_FILES := $(addprefix $(SPATH)/,$(shell cd $(SPATH) && git ls-files))
 
 DEP_FLAGS := $(SONIC_COMMON_FLAGS_LIST) \
--- a/Makefile.cache
+++ b/Makefile.cache
@@ -68,6 +68,7 @@
 # Run the 'touch cache.skip.common' command in the base directory to exclude the common files from caching
 SONIC_COMMON_FILES_LIST :=  $(if $(wildcard cache.skip.common),, .platform slave.mk rules/functions Makefile.cache)
 SONIC_COMMON_FLAGS_LIST :=  $(CONFIGURED_PLATFORM) \
+                            $(CONFIGURED_ARCH) \
                             $(BLDENV) \
                             $(SONIC_DEBUGGING_ON) \
                             $(SONIC_PROFILING_ON) $(SONIC_ENABLE_SYNCD_RPC)


qiluo-msft:
> Some of the file changes are relative to the Linux kernel build, such as sonic-slave-stretch/Dockerfile.j2, we can exclude it.

Do you mean "not relative"?

rules/linux-kernel.dep
```diff
@@ -1,7 +1,6 @@
 
 SPATH       := $($(LINUX_HEADERS_COMMON)_SRC_PATH)
-DEP_FILES   := $(SONIC_COMMON_FILES_LIST) rules/linux-kernel.mk rules/linux-kernel.dep
-DEP_FILES   += $(SONIC_COMMON_BASE_FILES_LIST)
+DEP_FILES   := rules/linux-kernel.mk rules/linux-kernel.dep
```
qiluo-msft, line 3: >DEP_FILES [](http://example.com/codeflow?start=0&length=9)

Do you exclude `.arch` file? Should we include it?

xumia:
> > Some of the file changes are relative to the Linux kernel build, such as sonic-slave-stretch/Dockerfile.j2, we can exclude it.
> 
> Do you mean "not relative"?

Yes, fixed, thanks.

rules/linux-kernel.dep
xumia: @qiluo-msft , the file .arch not in the SONIC_COMMON_FILES_LIST or SONIC_COMMON_BASE_FILES_LIST, see https://github.com/Azure/sonic-buildimage/blob/3e4658231421e9f23996ab7699acbc0fa3b9da2b/Makefile.cache#L69 and https://github.com/Azure/sonic-buildimage/blob/3e4658231421e9f23996ab7699acbc0fa3b9da2b/Makefile.cache#L76

We have another option SONIC_COMMON_FLAGS_LIST contains the arch info, not changed.

qiluo-msft: I am wondering if there will be a problem if one CONFIGURED_PLATFORM has two arch. The kernel build depends arch [https://github.com/Azure/sonic-linux-kernel/blob/10b1d7bec7c8336aa0ddbe4c755fd46e8b64b3e0/Makefile#L110](https://github.com/Azure/sonic-linux-kernel/blob/10b1d7bec7c8336aa0ddbe4c755fd46e8b64b3e0/Makefile#L110)

xumia: Currently, the CONFIGURED_PLATFORM contains arch, such as marvel_armhf, centec-arm64, do we have any other scenarios that have none unique platform name?
To solve the issue, we should add the variable CONFIGURED_ARCH in SONIC_COMMON_FLAGS_LIST, no SONIC_COMMON_FILES_LIST.

qiluo-msft (approved):

PR #9607, opened 2021-12-21
[pmon] Move smartctl from pmon to host

alexrallen:
#### Why I did it

Need to be able to run `smartctl` when pmon docker is not running. 

#### How I did it

Removed the pmon dependency for pmon as well as the command wrapper and added it to the debian-extension. 

#### How to verify it

* Stop pmon
* Run `smartctl` from the host and verify it runs without error

None.

[pmon] Move smartctl from pmon to host


qiluo-msft:
What is the reason that pmon docker is not running?

alexrallen:
I need smartctl to be accessible during reboot in order to perform firmware installations. I also believe that it is important for a diagnostic tool such as smartctl to not be dependent on high level functionality like docker.

sujinmkang (dismissed):

alexrallen:
@sujinmkang I fixed the conflict, please re-review.

sujinmkang (approved):

judyjoseph:
@alexrallen There are conflicts merging into 202111, please raise a new PR on branch 20211.

PR #9647, opened 2021-12-27
Fix docker-sonic-mgmt reproducible related issue.

liushilongbuaa:
#### Why I did it
when using 'sed -i' the directory must also be writable as sed will create a temporary file in that directory first.
This issue breaks building sonic-mgmt docker image.
we should use sudo to resolve this issue, but sonic-base image haven't installed sudo tool.
So we should use both 'sed -i' and 'sudo sed -i'
#### How I did it
add retry with sudo.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106

### DIFF
--- a/src/sonic-build-hooks/scripts/buildinfo_base.sh
+++ b/src/sonic-build-hooks/scripts/buildinfo_base.sh
@@ -17,6 +17,12 @@ REPR_MIRROR_URL_PATTERN='http:\/\/packages.trafficmanager.net\/debian'
 
 URL_PREFIX=$(echo "${PACKAGE_URL_PREFIX}" | sed -E "s#(//[^/]*/).*#\1#")
 
+if [[ $USER == 'root' ]];then
+    SUDO=''
+else
+    SUDO=sudo
+fi
+
 log_err()
 {
     echo "$1" >> $LOG_PATH/error.log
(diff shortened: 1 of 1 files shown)


src/sonic-build-hooks/scripts/buildinfo_base.sh
```diff
@@ -72,7 +72,7 @@ set_reproducible_mirrors()
 
     local mirrors="/etc/apt/sources.list $(find /etc/apt/sources.list.d/ -type f)"
     for mirror in $mirrors; do
-        sed -i "$expression" "$mirror"
+        sed -i "$expression" "$mirror" || sudo sed -i "$expression" "$mirror"
```
xumia, line 75: Can we change to run the command as root in docker-sonic-mgmt?

liushilongbuaa: how about now?

xumia (approved):

src/sonic-build-hooks/scripts/buildinfo_base.sh
```diff
@@ -17,6 +17,12 @@ REPR_MIRROR_URL_PATTERN='http:\/\/packages.trafficmanager.net\/debian'
 
 URL_PREFIX=$(echo "${PACKAGE_URL_PREFIX}" | sed -E "s#(//[^/]*/).*#\1#")
 
+if [ $USER != 'root' ] && [ -n $(which sudo) ];then
+    SUDO=sudo
+else
+    SUDO=''
+fi
```
xumia, line 24: Can be simplified to 
[ $USER != 'root' ] && [ -n $(which sudo) ] && SUDO=sudo

xumia:
@liushilongbuaa , we should not need to merge to release branches, right?

PR #9651, opened 2021-12-28
Enable/Disable Order ECMP feature.

abdosi:
What I did:
Updated Jinja2 Template in `switch.json.j2`  for enabling/disabling Order ECMP feature based on device role.
Changes as per design: https://github.com/Azure/SONiC/pull/896

How I verify:
Manual Verification for template getting generated correctly.


dockers/docker-orchagent/switch.json.j2
```diff
@@ -20,7 +20,12 @@
         "SWITCH_TABLE:switch": {
             "ecmp_hash_seed": "{{ hash_seed_value }}",
             "lag_hash_seed": "{{ hash_seed_value }}",
-            "fdb_aging_time": "600"
+            "fdb_aging_time": "600",
+{% if DEVICE_METADATA.localhost.type and "LeafRouter" in DEVICE_METADATA.localhost.type %}
```
shi-su, line 24: Just wanted to confirm. In the design, whether ordered ECMP is enabled is only dependent on the type. There will be no additional switch for it?

abdosi: yes for now device type only. There was thinking to use asic_type also but now we are using sai query capability and enable/disable based on it.

shi-su (approved):

PR #9655, opened 2021-12-28
[201811] Fix DHCP Relay Monitor Jinaja2 Template file.

abdosi:
Why I did:

DHCP Relay docker does not start if we don't have v6 DHCP Relay address as this caused jinja2 template generation failure.

What I did:

Fixed the template . The fix is to make template generation same as present in master code.

How I verify:

Manual Verification and make sure dhcp relay docker is running fine.


kellyyeh (approved):

abdosi:
@yxieca Fix is merged in 201811.

PR #9659, opened 2021-12-29
[sonic-yang-mgmt] Adding flag to disable/enable log printing

ghooo:
#### Why I did it
Fixes https://github.com/Azure/sonic-utilities/issues/1871

From [generic-config-updater](https://github.com/Azure/sonic-utilities/tree/master/generic_config_updater) we call `sonic-yang-mgmt` multiple times in order to check a certain change to ConfigDb is valid or not. It is expected for some changes to be invalid, so always printing errors from `sonic-yang-mgmt` makes the output hard to read.

In this PR, we are adding a way to control if logs should be printed or not.

#### How I did it
- Added `print_log_enabled` flag to sonic_yang ctor
- Converted all `print` statements to `sysLog(..., doPrint=True)`

#### How to verify it
unit-test passing means the change did not break logs.

#### Info about libyang logging
libyang provides an extensive logging logic which can support a lot of scenarios:
- ly_log_level: setting logging level
  - LY_LLERR
  - LY_LLWRN
  - ... 
- ly_set_log_clb: setting log callback to customize the default behavior which is printing the msgs
- ly_log_options: setting logging options 
  - LY_LOLOG: If callback is set use it, otherwise just print. If flag is not set, do nothing.
  - ...

For more info refer to:
- https://netopeer.liberouter.org/doc/libyang/devel/html/group__logopts.html#gaff80501597ed76344a679be2b90a1d0a
- https://netopeer.liberouter.org/doc/libyang/devel/html/group__log.html#gac88b78694dfe9efe0450a69603f7eceb

#### What's next?
Consume the new flag `print_log_enabled` in [generic-config-updater](https://github.com/Azure/sonic-utilities/tree/master/generic_config_updater) to reduce the logging clutter. 

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106

### DIFF
--- a/src/sonic-yang-mgmt/sonic_yang.py
+++ b/src/sonic-yang-mgmt/sonic_yang.py
@@ -13,6 +13,7 @@ i.e. it is mixin not parent class.
 class SonicYang(SonicYangExtMixin):
 
     def __init__(self, yang_dir, debug=False):
+        self.enable_logging()
         self.yang_dir = yang_dir
         self.ctx = None
         self.module = None
@@ -50,7 +51,18 @@ class SonicYang(SonicYangExtMixin):
     def __del__(self):
         pass
 
+    def disable_logging(self):
+        ly.set_log_options(0)
+        self.syslogging_disabled = True
+
+    def enable_logging(self):
+        # ly.LY_LOLOG: If callback is set use it, otherwise just print. If flag is not set, do nothing.
+        ly.set_log_options(ly.LY_LOLOG)
+        self.syslogging_disabled = False
+
     def sysLog(self, debug=syslog.LOG_INFO, msg=None, doPrint=False):
+        if self.syslogging_disabled:
+            return
 
         # log debug only if enabled
         if self.DEBUG == False and debug == syslog.LOG_DEBUG:
@@ -64,7 +76,7 @@ class SonicYang(SonicYangExtMixin):
         return
 
     def fail(self, e):
-        print(e)
+        self.sysLog(msg=e, debug=syslog.LOG_ERR, doPrint=True)
         raise e
 
     """
@@ -76,7 +88,7 @@ class SonicYang(SonicYangExtMixin):
         try:
             return self.ctx.parse_module_path(yang_file, ly.LYS_IN_YANG)
         except Exception as e:
-            print("Failed to load yang module file: " + yang_file)
+            self.sysLog(msg="Failed to load yang module file: " + yang_file, debug=syslog.LOG_ERR, doPrint=True)
             self.fail(e)
 
     """
@@ -120,7 +132,7 @@ class SonicYang(SonicYangExtMixin):
             try:
                 ctx.parse_module_path(str(file), ly.LYS_IN_YANG)
             except Exception as e:
-                print("Failed to parse yang module file: " + file)
+                self.sysLog(msg="Failed to parse yang module file: " + file, debug=syslog.LOG_ERR, doPrint=True)
                 self.fail(e)
 
         return ctx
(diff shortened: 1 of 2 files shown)


src/sonic-yang-mgmt/sonic_yang.py
```diff
@@ -50,7 +51,18 @@ def __init__(self, yang_dir, debug=False):
     def __del__(self):
         pass
 
+    def disable_logging(self):
+        ly.set_log_options(0)
+        self.syslogging_disabled = True
+
+    def enable_logging(self):
+        # ly.LY_LOLOG: If callback is set use it, otherwise just print. If flag is not set, do nothing.
+        ly.set_log_options(ly.LY_LOLOG)
```
ghooo, line 60: An alternative solution to always disable libyang logging. Instead we can use the following:
```
        ly.set_log_options(ly.LY_LOSTORE_LAST)
```
or
```
        ly.set_log_options(ly.LY_LOSTORE)
```
Then in all lines where we call any functionality in libyang we print the errors ourselves using:
```
get_ly_errors(ctx)
```

This gives us more fine tuned control over how logs show up. For my case, I only needed a way to disable/enable logging, so do not need to make it more complex.

ghooo: The option suggested in this comment also provide consistency across instances of `sonic_yang`, assume there are multiple instances created, each instance should have control of how it logs. But relying `ly.LY_LOLOG` can cause inconsistencies between instance. If we rely instead on `ly.LY_LOSTORE` or `ly.LY_LOSTORE_LAST` and `get_ly_errors(ctx)` this can only be associated with a single `context` and there is a single `context` per `sonic_yang` object.

For `generic-config-updater` use case we don't manipulate multiple instances of `sonic_yang` at once, so we don't care about this problem. Also manipulating `yang` global state using `ly.set_log_options` can be done anyway by different files, so there is no easy to fix the logging flags consistency issues.

ghooo, line 60: The default flags were:
`LY_LOLOG | LY_LOSTORE_LAST`

LY_LOSTORE_LAST: means only the last error is stored and can be accessed using `ly.get_ly_errors(ctx)`. In `sonic_yang` we don't use `get_ly_errors(ctx)` so there is no need to store the last errors.

ghooo: [reference](https://netopeer.liberouter.org/doc/libyang/devel/html/group__log.html#ga9a039b617d81acfe361f641e5a4f8618)

src/sonic-yang-mgmt/sonic_yang.py
```diff
@@ -50,7 +51,18 @@ def __init__(self, yang_dir, debug=False):
...
+        self.syslogging_disabled = False
+
     def sysLog(self, debug=syslog.LOG_INFO, msg=None, doPrint=False):
+        if self.syslogging_disabled:
```
qiluo-msft, line 64: >syslogging_disabled [](http://example.com/codeflow?start=16&length=19)

The name is confusing, it actually disabled syslog and print. Your intention is to only disable print?

ghooo: my intention is to disable logging

ghooo: I think it will be hard to read the logs as well if we keep printing errors during a change validation

ghooo: I see why it might be confusing, the name of the property `syslogging_disabled` but it is disabling it for both `syslog` and `print`. I got the name of the property from the name of the method `sysLog`.

ghooo: renamed to `logging_disabled`

praveen-li (approved):
LGTM, BTW I cannot recall why the name of argument is 'debug' in sysLog function. We can change it to 'level' or 'logLevel', if needed.  
def sysLog(self, debug=syslog.LOG_INFO, msg=None, doPrint=False):

Approved, even if above is not addressed. Thx

qiluo-msft (approved):

ghooo:
@praveen-li let me rename the flag in a separate PR, the build takes a very long time to finish and I have an ask I need to finish soon.

PR #9674, opened 2022-01-05
Updated Internal BGP Templates for chassis packet

abdosi:
What/Why I did:

```
   Some time jinja2 template renderation of Loopbcack4096 was not happening
    correctly because of timing issue where DB might not be poupulated.
    So Instead of using IP Address using Interface name.

    Fixes: https://github.com/Azure/sonic-buildimage/issues/9610
```

How I verify:

Unit Test.


abdosi:
@anamehra Can you please check if this fixes the issue.

anamehra:
> @anamehra Can you please check if this fixes the issue.

@abdosi , I validated this change before the holidays and it worked fine. You may go ahead with the merge.

anamehra (approved):

abdosi:
@arlakshm can you please help review/approve this.

arlakshm:
@abdosi, as discussed offline can you verify the race condition where the Loopback4096 ip address is configuration after the bgp configuration is applied in FRR

abdosi:
> @abdosi, as discussed offline can you verify the race condition where the Loopback4096 ip address is configuration after the bgp configuration is applied in FRR

@arlakshm I have done some more regression testing with this change and it looks fine. Can you please approve this.

PR #9688, opened 2022-01-06
Correct the pkg name for thrift.0.13.0

richardyu-ms:
Correct thrift.0.13.0 dependent package name.
In previous code, the buildout target was named as PYTHON3_THRIFT_0_13_0
But when add the prackage to LIBTHRIFT_0_13_0, it typo as PYTHON_THRIFT_0_13_0

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106


azure-pipelines.yml
```diff
@@ -37,7 +37,7 @@ stages:
   variables:
   - template: .azure-pipelines/azure-pipelines-repd-build-variables.yml
   - name: CACHE_MODE
-    value: rcache
+    value: none
```
lguohan, line 40: why change this?

qiluo-msft (commented):
Share the same concern. Other looks good.

azure-pipelines.yml
richardyu-ms: Use this to identify a potential issue. which addressed at https://github.com/Azure/sonic-buildimage/pull/9705#pullrequestreview-847106456
this will be removed.
Thanks for the question and sorry for the confusion :)

qiluo-msft (approved):

PR #9707, opened 2022-01-08
[minigraph] Enhanced parser to parse interface name for static route nexthop

abdosi:
What I did:-

Enhanced minigraph parser to parse interface name associated with static route nexthop

Why I did:-

One of the use case to support interface name is Chassis Packet. For Chassis Packet we have Static Routes configured to route traffic across line-card. If the FRR programs static route without the interface name then in case if the ip interface that is associated with the nexthop goes down FRR resolves static route nexthop over the default route as we have FRR config `ip nht-resolve-via-default` which causes undesired behavior. Having interface name with Static Route prevents recursive lookup on default route.

How I verify:

- Updated  unit-test cases
- Manual verification


abdosi:
@anamehra please help to validate this change.

abdosi:
/ azp run Azure.sonic-buildimage

arlakshm (approved):

PR #9740, opened 2022-01-12
[dhcp_relay] Check payload size to prevent buffer overflow in dhcpv6 option

kellyyeh:
#### Why I did it
When parsing through relay reply packets, parse_dhcpv6_option introduces vulnerability in buffer overflow, allowing up to 2 byte/65535 copied into a fixed buffer of size 4096.

#### How I did it
Add bound checks to before and after option parsing

#### How to verify it
Send a dummy packet with option length greater than 4096. dhcp6relay should not perform memcpy on invalid option length.

- [x] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012
- [ ] 202106


src/dhcp6relay/src/relay.cpp
```diff
@@ -492,14 +492,17 @@ void relay_client(int sock, const uint8_t *msg, int32_t len, const ip6_hdr *ip_h
     auto position = current_position + sizeof(struct dhcpv6_option);
     auto dhcpv6msg = parse_dhcpv6_hdr(position);
 
-    while ((current_position - msg) != len) {
+    while ((current_position - msg) != len && (current_position - msg) < len)  {
```
yxieca, line 495: I think second condition already covered the first one?

src/dhcp6relay/src/relay.cpp
```diff
@@ -492,14 +492,17 @@ void relay_client(int sock, const uint8_t *msg, int32_t len, const ip6_hdr *ip_h
...
         auto option = parse_dhcpv6_opt(current_position, &tmp);
         current_position = tmp;
+        if(option->option_length > (msg-current_buffer_position)) {
```
yxieca, line 498: what is msg-current_buffer_position? Isn't this negative?

yxieca (approved):

yxieca (dismissed):

yxieca (approved):

PR #9748, opened 2022-01-12
[dhcp_relay] Add isc-dhcp-server

kellyyeh:
#### Why I did it
Add isc-dhcp-server to dhcp relay docker

#### How I did it
Add isc-dhcp-server package

#### How to verify it
Run dhcpd in dhcp relay docker

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012
- [ ] 202106


rules/isc-dhcp-server.mk
```diff
@@ -0,0 +1,15 @@
+# isc-dhcp-server
+
+ISC_DHCP_VERSION = 4.3.3
+ISC_DHCP_VERSION_FULL = ${ISC_DHCP_VERSION}-6
+
+export ISC_DHCP_VERSION ISC_DHCP_VERSION_FULL
+
+ISC_DHCP_SERVER = isc-dhcp-server_$(ISC_DHCP_VERSION_FULL)_$(CONFIGURED_ARCH).deb
+$(ISC_DHCP_SERVER)_SRC_PATH = $(SRC_PATH)/isc-dhcp
```
saiarcot895, line 9: Since both `isc-dhcp-server` and `isc-dhcp-relay` have the same source package, I think it's best if they're both specified in the same rules file.

sonic-slave-bullseye/Dockerfile.j2
```diff
@@ -468,6 +468,9 @@ RUN apt-get install -y xsltproc
 # Install dependencies for isc-dhcp-relay build
 RUN apt-get -y build-dep isc-dhcp
 
+# Install dependencies for isc-dhcp-server build
+RUN apt-get -y build-dep isc-dhcp-server
```
saiarcot895, line 472: This isn't needed, since `isc-dhcp` is the source package name, and that's specified above.

rules/isc-dhcp-server.mk
kellyyeh: Updated

PR #9757, opened 2022-01-13
[redis] Upgrade redis version

xjasonlyu:
#### Why I did it

The current redis version of SONiC is `6.0.6`, which contains many high-risky security issues like CVEs that are fixed in the latest version. The Redis release notes also highly recommend to upgrade with SECURITY urgency.

```
================================================================================
Redis 6.0.16 Released Mon Oct 4 12:00:00 IDT 2021
================================================================================

Upgrade urgency: SECURITY, contains fixes to security issues.

Security Fixes:
* (CVE-2021-41099) Integer to heap buffer overflow handling certain string
  commands and network payloads, when proto-max-bulk-len is manually configured
  to a non-default, very large value [reported by yiyuaner].
* (CVE-2021-32762) Integer to heap buffer overflow issue in redis-cli and
  redis-sentinel parsing large multi-bulk replies on some older and less common
  platforms [reported by Microsoft Vulnerability Research].
* (CVE-2021-32687) Integer to heap buffer overflow with intsets, when
  set-max-intset-entries is manually configured to a non-default, very large
  value [reported by Pawel Wieczorkiewicz, AWS].
* (CVE-2021-32675) Denial Of Service when processing RESP request payloads with
  a large number of elements on many connections.
* (CVE-2021-32672) Random heap reading issue with Lua Debugger [reported by
  Meir Shpilraien].
* (CVE-2021-32628) Integer to heap buffer overflow handling ziplist-encoded
  data types, when configuring a large, non-default value for
  hash-max-ziplist-entries, hash-max-ziplist-value, zset-max-ziplist-entries
  or zset-max-ziplist-value [reported by sundb].
* (CVE-2021-32627) Integer to heap buffer overflow issue with streams, when
  configuring a non-default, large value for proto-max-bulk-len and
  client-query-buffer-limit [reported by sundb].
* (CVE-2021-32626) Specially crafted Lua scripts may result with Heap buffer
  overflow [reported by Meir Shpilraien].

Other bug fixes:
* Fix appendfsync to always guarantee fsync before reply, on MacOS and FreeBSD (kqueue) (#9416)
* Fix the wrong mis-detection of sync_file_range system call, affecting performance (#9371)
* Fix replication issues when repl-diskless-load is used (#9280)
```

#### How I did it

Edit `Dockerfile.j2` file

#### How to verify it

Check redis version

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106

This PR will upgrade redis-server version to `6.0.16`.


qiluo-msft:
Thanks for the contribution!
I checked all the CVE are fixed in bullseye (security). Can you just `apt-get install -t bullseye-security redis-tools`?
We previously backup deb packages on sonicstorage in case of Debian package server unavailable. However it is not needed any more, and we have reproducible build enable on release branches. (@xumia for vis)

ref: https://security-tracker.debian.org/tracker/CVE-2021-32626

xjasonlyu:
> Thanks for the contribution! I checked all the CVE are fixed in bullseye (security). Can you just `apt-get install -t bullseye-security redis-tools`? We previously backup deb packages on sonicstorage in case of Debian package server unavailable. However it is not needed any more, and we have reproducible build enable on release branches. (@xumia for vis)
> 
> ref: https://security-tracker.debian.org/tracker/CVE-2021-32626

The Debian release used by SONiC's database docker image is `buster`, so running `apt-get install -t bullseye-security redis-tools` will cause errors since it uses `bullseye` source. We can add bullseye's source to buster's apt source, but I believe it's not recommended. In this case, the manual installation might be better?

qiluo-msft:
> FROM docker-config-engine-buster

Could you try `docker-config-engine-bullseye` as the image FROM?
One of our tasks is to upgrade all image to bullseye version.

---
Refers to: dockers/docker-database/Dockerfile.j2:2 in d052cb7. [](commit_id = d052cb748987a7930cc1e97b600081b053592102, deletion_comment = False)

dockers/docker-database/Dockerfile.j2
```diff
@@ -11,22 +11,7 @@ ENV DEBIAN_FRONTEND=noninteractive
...
-RUN dpkg -i redis-tools_6.0.6-1~bpo10+1_arm64.deb redis-server_6.0.6-1~bpo10+1_arm64.deb || apt-get install -f
-RUN rm redis-tools_6.0.6-1~bpo10+1_arm64.deb redis-server_6.0.6-1~bpo10+1_arm64.deb
-{% else %}
-RUN curl -o redis-tools_6.0.6-1~bpo10+1_amd64.deb "https://sonicstorage.blob.core.windows.net/packages/redis/redis-tools_6.0.6-1~bpo10+1_amd64.deb?sv=2015-04-05&sr=b&sig=73zbmjkf3pi%2Bn0R8Hy7CWT2EUvOAyzM5aLYJWCLySGM%3D&se=2030-09-06T19%3A44%3A59Z&sp=r"
-RUN curl -o redis-server_6.0.6-1~bpo10+1_amd64.deb "https://sonicstorage.blob.core.windows.net/packages/redis/redis-server_6.0.6-1~bpo10+1_amd64.deb?sv=2015-04-05&sr=b&sig=2Ketg7BmkZEaTxR%2FgvAFVmhjn7ywdmkc7l2T2rsL57o%3D&se=2030-09-06T19%3A45%3A20Z&sp=r"
-RUN dpkg -i redis-tools_6.0.6-1~bpo10+1_amd64.deb redis-server_6.0.6-1~bpo10+1_amd64.deb || apt-get install -f
-RUN rm redis-tools_6.0.6-1~bpo10+1_amd64.deb redis-server_6.0.6-1~bpo10+1_amd64.deb
-{% endif %}
+RUN apt-get install -y -t bullseye-security redis-tools redis-server
```
xumia, line 14: How about using buster-security, the issue has fixed in buster-security, see https://security-tracker.debian.org/tracker/CVE-2021-32626
To upgrade bullseye-security, it is another task, we can focus on security only in this PR if it works in buster.

qiluo-msft: Agree with focus on security first.

xjasonlyu: But the Redis version in buster is 5.0.x, which means a major downgrade from current Redis version 6.0.x then.

qiluo-msft: @xjasonlyu Please go ahead with manual installation of the version used in bullseye-security. Also add a TODO comment that we should directly install bullseye-security from package server after the OS release version upgraded.

xjasonlyu: @qiluo-msft The Redis 6.0.16 requires libc6 >= 2.29 to install, however, the highest libc6 version supported by buster is 2.28. The debian community says that install libc6 higher than 2.28 may corrupt the system, so I wonder can we simply use static compiled binary (redis-server/redis-cli, etc.) to avoid the problem? Or do you have any other suggestions?

qiluo-msft: Would you like to upgrade database container to bullseye first? If yes, we are okay with this direction.

dockers/docker-database/Dockerfile.j2
```diff
@@ -1,5 +1,5 @@
 {% from "dockers/dockerfile-macros.j2" import install_debian_packages, install_python_wheels, copy_files %}
-FROM docker-config-engine-buster
+FROM docker-config-engine-bullseye
```
qiluo-msft, line 2: Don't forget the change the dependencies in the file rules/docker-database.mk

dockers/docker-database/Dockerfile.j2
```diff
@@ -11,21 +11,22 @@ ENV DEBIAN_FRONTEND=noninteractive
 RUN apt-get update
 
 # Install redis-server
+# TODO: Require direct bullseye-security installation after the OS release version upgraded
```
qiluo-msft, line 14: I suggest just "direct bullseye-security installation after the OS release version upgraded"

dockers/docker-database/Dockerfile.j2
```diff
@@ -11,21 +11,22 @@ ENV DEBIAN_FRONTEND=noninteractive
...
 {% if CONFIGURED_ARCH == "armhf" %}
-RUN curl -k -o redis-tools_6.0.6-1~bpo10+1_armhf.deb "https://sonicstorage.blob.core.windows.net/packages/redis/redis-tools_6.0.6-1_bpo10+1_armhf.deb?sv=2015-04-05&sr=b&sig=67vHAMxsl%2BS3X1KsqhdYhakJkGdg5FKSPgU8kUiw4as%3D&se=2030-10-24T04%3A22%3A40Z&sp=r"
-RUN curl -k -o redis-server_6.0.6-1~bpo10+1_armhf.deb "https://sonicstorage.blob.core.windows.net/packages/redis/redis-server_6.0.6-1_bpo10+1_armhf.deb?sv=2015-04-05&sr=b&sig=xTdayvm0RBguxi9suyv855jKRjU%2FmKQ8nHuct4WSX%2FA%3D&se=2030-10-24T04%3A22%3A05Z&sp=r"
-RUN dpkg -i redis-tools_6.0.6-1~bpo10+1_armhf.deb redis-server_6.0.6-1~bpo10+1_armhf.deb || apt-get install -f
-RUN rm redis-tools_6.0.6-1~bpo10+1_armhf.deb redis-server_6.0.6-1~bpo10+1_armhf.deb
+RUN curl -sS -o redis-tools_6.0.15-1_armhf.deb "http://ftp.debian.org/debian/pool/main/r/redis/redis-tools_6.0.15-1_armhf.deb"    && \
```
saiarcot895, line 16: I think this can be changed to use apt now, since we're getting from the Debian repos?

dockers/docker-database/Dockerfile.j2
saiarcot895: Bullseye-security is part of the `sources.list` file, so just saying `apt install redis-server` should install the latest version (whether that be from bullseye-security or bullseye-updates). Given that, does bullseye-security need to be explicitly mentioned?

saiarcot895 (changes requested):

dockers/docker-database/Dockerfile.j2
xjasonlyu: OK

dockers/docker-database/Dockerfile.j2
xjasonlyu: Yes, apt-get can handle all the jobs now.

dockers/docker-database/Dockerfile.j2
xjasonlyu: @qiluo-msft I agree with @saiarcot895, since we have upgraded to bullseye, so we can use apt installation directly.

rules/docker-database.mk
```diff
@@ -29,5 +29,5 @@ $(DOCKER_DATABASE)_BASE_IMAGE_FILES += redis-cli:/usr/bin/redis-cli
 $(DOCKER_DATABASE)_FILES += $(SYSCTL_NET_CONFIG) $(SUPERVISOR_PROC_EXIT_LISTENER_SCRIPT)
 $(DOCKER_DATABASE)_FILES += $(UPDATE_CHASSISDB_CONFIG_SCRIPT)
 
-SONIC_BUSTER_DOCKERS += $(DOCKER_DATABASE)
-SONIC_BUSTER_DBG_DOCKERS += $(DOCKER_DATABASE_DBG)
+SONIC_BULLSEYE_DOCKERS += $(DOCKER_DATABASE)
```
saiarcot895, line 32: These variables aren't actually defined; only the containers that are based on a Debian version that doesn't match the base image (i.e. Bullseye) need to be added into the `SONIC_*_DOCKERS` variable. Since the base image is Bullseye, there is no `SONIC_BULLSEYE_DOCKERS` or `SONIC_BULLSEYE_DBG_DOCKERS` variables.

xjasonlyu:
@qiluo-msft what's your opinion? stick to manual installation or just 'apt-get'?

rules/docker-database.mk
xjasonlyu: Sorry, I don't quite understand those definitions in the mk file. Did you suggest that I should delete this line since `SONIC_BULLSEYE_DOCKERS` or `SONIC_BULLSEYE_DBG_DOCKERS` doesn't exist?

saiarcot895: Yes. Neither of those variables exist.

Those version-specific variables (stretch, Jessie, buster) are defined in our build scripts for all older versions of Debian, since the containers may be based on an older version, and dependencies need to be correctly handled. Therefore, since bullseye is the current version, there's no `SONIC_BULLSEYE_DOCKERS` or `SONIC_BULLSEYE_DBG_DOCKERS`.

qiluo-msft (approved):

qiluo-msft:
Could you update PR description? See you are not using Redis 6.2.6.

xjasonlyu:
> Could you update PR description? See you are not using Redis 6.2.6.

Updated.

Issue #8558, opened 2021-08-23
[TH3][202012] WARMReboot test in test_advanced_reboot.py failed due to MAC events reported by SAI to SONIC during warmboot


gechiang:
#### Description
While testing TH3 Advanced reboot, part of the warmreboot test we observed Orchagent crashed.  Upon closer examination on the syslog we found that the crash was due to BRCM SAI returning a set of bridge ports that belong to VLAN 1000 instead of the default VLAN (1).  This caused the meta checker to catch that the reference count being > 0 and failed the attempt to remove these non default bridge ports.
This issue is easily reproduced by executing the "test_advanced_reboot.py" on a TH3 T0 DUT where Orchagent Core is observed.

Here is the portion of the syslog that shows the issue:
```
...
Aug 23 18:28:05.723987 str2-z9332f-05 NOTICE swss#portmgrd: :- doTask: Configure Ethernet24 MTU to 9100
Aug 23 18:28:05.724304 str2-z9332f-05 NOTICE swss#portsyncd: :- onMsg: nlmsg type:16 key:Ethernet2 admin:1 oper:1 addr:c8:f7:50:ed:15:41 ifindex:37 master:17
Aug 23 18:28:05.727381 str2-z9332f-05 NOTICE swss#portsyncd: message repeated 4 times: [ :- onMsg: nlmsg type:16 key:Ethernet2 admin:1 oper:1 addr:c8:f7:50:ed:15:41 ifindex:37 master:17]
Aug 23 18:28:05.727381 str2-z9332f-05 NOTICE swss#orchagent: :- removeDefaultVlanMembers: Remove 0 VLAN members from default VLAN
Aug 23 18:28:05.728018 str2-z9332f-05 INFO kernel: [   29.844864] Bridge: port 24(Ethernet2) entered blocking state
Aug 23 18:28:05.728030 str2-z9332f-05 INFO kernel: [   29.844866] Bridge: port 24(Ethernet2) entered disabled state
Aug 23 18:28:05.728032 str2-z9332f-05 INFO kernel: [   29.845297] device Ethernet2 entered promiscuous mode
Aug 23 18:28:05.728034 str2-z9332f-05 INFO kernel: [   29.845575] Bridge: port 24(Ethernet2) entered blocking state
Aug 23 18:28:05.728037 str2-z9332f-05 INFO kernel: [   29.845578] Bridge: port 24(Ethernet2) entered forwarding state
[...]

PR #9760, opened 2022-01-13
[202012][BRCM TH3] Add SOC properties to prevent FDB events during warmboot

gechiang:
#### Why I did it
This is to fix issue reported by (https://github.com/Azure/sonic-buildimage/issues/8558)
The fix requires 2 changes:

1.  in BRCM SAI 4.3.5.2 which can be found in this PR (https://github.com/Azure/sonic-buildimage/pull/9754)
2.  This PR itself where it needs to set 2 SOC properties in order to activate the code changed in BRCM SAI 4.3.5.2

#### How to verify it

1. Repeat the steps documented in issue (https://github.com/Azure/sonic-buildimage/issues/8558) without these 2 SOC properties on a TH3 based HWSKU (Z9332F) and with any BRCM SAI (including 4.3.5.2) and see the issue.
2. Enable these two SOC properties and repeat the same steps again with Z9332F running with BRCM SAI 4.3.5.2 and observe that the problem no longer can be reproduced.

Once verified that the fix that went into SAI 4.3.5.2 is also present in the SAI 6.0 branch, we should then port this to master branch (running BRCM SAI 6.0.x).
There is no plan to port this to 

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106

[202012][BRCM TH3] Add SOC properties to prevent FDB events during warmboot


gechiang:
Please ignore this PR.  Will start a new one shortly...

PR #9762, opened 2022-01-14
[swss]: Reduce tunnel_packet_handler memory usage

theasianpianist:

#### Why I did it
The tunnel packet handler gradually leaks memory

#### How I did it
- Remove call to the `link_lookup` method from the Pyroute2 library
- Use the `store=0` argument for the Scapy sniffer

#### How to verify it
- Run the packet handler and observe SWSS memory usage using `docker stat`

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012
- [ ] 202106

### DIFF
--- a/dockers/docker-orchagent/tunnel_packet_handler.py
+++ b/dockers/docker-orchagent/tunnel_packet_handler.py
@@ -9,15 +9,17 @@ destination IP to trigger the process of obtaining neighbor information
 """
 import subprocess
 import time
-
 from datetime import datetime
 from ipaddress import ip_interface
+
+from swsssdk import ConfigDBConnector, SonicV2Connector
+from sonic_py_common import logger as log
+
 from pyroute2 import IPRoute
 from scapy.layers.inet import IP
 from scapy.layers.inet6 import IPv6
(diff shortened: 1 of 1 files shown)


prsunny:
let's plan to add a sonic-mgmt test for this. Mainly since we are manipulating the netlink attribute and in case future kernel versions has any change, it would be easy to catch.

dockers/docker-orchagent/tunnel_packet_handler.py
```diff
@@ -68,19 +74,32 @@ def portchannel_intfs(self):
...
+    def get_intf_name(self, msg):
+        """
+        Gets the interface name for a netlink msg
+
+        Returns:
+            (str) The interface name, or the empty string if no interface
+                  name was found
         """
-        Gets a mapping of interface kernel indices to portchannel interfaces
+        attr_list = msg['attrs']
```
prsunny, line 85: will there be a case if attr_list is null?

prsunny (approved):

prsunny (approved):

PR #4412, opened 2020-04-11
[caclmgrd] Add some default ACCEPT rules and lastly drop all incoming packets

jleveque:
**- What I did**

Modified behavior to enhance device security as follows:

- Upon starting or receiving notification of ACL table/rule changes in Config DB:
    1. Add iptables/ip6tables commands to allow all incoming packets from established TCP sessions or new TCP sessions which are related to established TCP sessions
    2. Add iptables/ip6tables commands to allow bidirectional ICMPv4 ping and traceroute
    3. Add iptables/ip6tables commands to allow bidirectional ICMPv6 ping and traceroute
    4. Add iptables/ip6tables commands to allow all incoming Neighbor Discovery Protocol (NDP) NS/NA/RS/RA messages
    5. Add iptables/ip6tables commands to allow all incoming IPv4 DHCP packets
    6. Add iptables/ip6tables commands to allow all incoming IPv6 DHCP packets
    7. Add iptables/ip6tables commands to allow all incoming BGP traffic
    8. Add iptables/ip6tables commands for all ACL rules for recognized services (currently SSH, SNMP, NTP)
    9. For all services which we did not find configured ACL rules, add iptables/ip6tables commands to allow all incoming packets for those services (allows the device to accept SSH connections before the device is configured)
    10. Add iptables rules to drop all packets destined for loopback interface IP addresses
    11. Add iptables rules to drop all packets destined for management interface IP addresses
    12. Add iptables rules to drop all packets destined for point-to-point interface IP addresses
    13. Add iptables rules to drop all packets destined for our VLAN interface gateway IP addresses
    14. Add iptables/ip6tables commands to allow all incoming packets with TTL of 0 or 1 (This allows the device to respond to tools like tcptraceroute)
    15. If we found control plane ACLs in the configuration and applied them, we lastly add iptables/ip6tables commands to drop all other incoming packets

Issue #4797, opened 2020-06-18
iptables blocking access to /32 Loopback address


scottlaird:
**Description**

The default iptables rules on SONiC block access to the .0 address of each interface netblock.  When Loopback is configured with a /32, this produces a rule that blocks all incoming traffic to loopback *except* for BGP and DHCP.  This is likely also an issue with /31 links, but not with /30 or larger.

**Steps to reproduce the issue:**
1.  Set loopback to a /32.  Example:
```
    "LOOPBACK_INTERFACE": {
        "Loopback0": {}, 
        "Loopback0|10.0.0.4/32": {}, 
        "Loopback0|2001:470:e959::3/128": {}
    }, 
```
2. Attempt to ssh, snmpwalk, etc the loopback interface and fail.
3. Run iptables -L -v -n and see this:
```
# iptables -L -v -n
Chain INPUT (policy ACCEPT 3199 packets, 357K bytes)
 pkts bytes target     prot opt in     out     source               destination         
  12M 1345M ACCEPT     all  --  lo     *       127.0.0.1            0.0.0.0/0           
57309   18M ACCEPT     tcp  --  *      *       0.0.0.0/0            0.0.0.0/0            ctstate RELATED,ESTABLISHED
 372K 2757M ACCEPT     icmp --  *      *       0.0.0.0/0            0.0.0.0/0            icmptype 8
  408 34272 ACCEPT     icmp --  *      *       0.0.0.0/0            0.0.0.0/0            icmptype 0
  104  9620 ACCEPT     icmp --  *      *       0.0.0.0/0            0.0.0.0/0            icmptype 3
    0     0 ACCEPT     icmp --  *      *       0.0.0.0/0            0.0.0.0/0            icmptype 11
14284 6250K ACCEPT     udp  --  *      *       0.0.0.0/0            0.0.0.0/0            udp spts:67:68 dpts:67:68
    0     0 ACCEPT     udp  --  *      *       0.0.0.0/0            0.0.0.0/0            udp spts:546:547 dpts:546:547
  768 46740 ACCEPT     tcp  --  *      *       0.0.0.0/0            0.0.0.0/0            tcp dpt:179
   42  1680 ACCEPT     tcp  --  *      *       0.0.0.0/0            0.0.0.0/0            tcp spt:179
10504  701K DROP       all  --  *      *       0.0.0.0/0            10.0.0.4            
    0     0 DROP       all  --  *      *       0.0.0.0/0            10.0.1.8           
[...]

PR #9826, opened 2022-01-21
caclmgrd: correct IP2ME address logic

### LINKED ISSUE #7008: Incorrect iptable rules from caclmgrd
#### Description

There are a couple of issues that I want to bring to your notice.

a. caclmgrd does not handle mgmt IP address to form the iptable rules if the mgmt interface is using DHCP.
b. caclmgrd seem to form incorrect IP address while configuring the IP table rules. If the configured IP address has 24-bit or lower netmask then the iptable rules are not formed correctly. Consider the following two cases:

i. IP address with 24-bit netmask and the corresponding iptable rule:
MGMT_INTERFACE|eth0|192.168.122.94/24
MGMT_PORT|eth0

/var/log/syslog.1:Mar 10 20:08:06.648400 sonic INFO caclmgrd[1171]:   iptables -A INPUT -d 192.168.122.0/32 -j DROP

The above rule looks incorrect, it should have been "iptables -A INPUT -d 192.168.122.0/24 -j DROP".

ii. IP address with 32-bit netmask and the corresponding iptable rule:
MGMT_INTERFACE|eth0|192.168.122.94/32
MGMT_PORT|eth0

/var/log/syslog:Mar 10 20:26:17.493740 sonic INFO caclmgrd[1095]:   iptables -A INPUT -d 192.168.122.94/32 -j DROP

The above is correct.

1. Configure IP address with 24-bit netmask
2. Check the iptable rule

The control traffic is not dropped on the interface that has 24-bit netmask.

The expectation was that by default Sonic blocks all control traffic except BGP, few ICMP packets (0,3,8,11), DHCP (v4 and v6) from what is seen in the code.
However, I am not sure what the expectation of Sonic is. Is it suppose to just allow only the traffic mentioned above? It seem to be intended to drop control traffic such as SSH and SNMP by default. Please clarify what the expectations are.

root@sonic:/home/cisco# show ver

SONiC Software Version: SONiC.202012.26-dirty-20210226.212041
Distribution: Debian 10.8
Kernel: 4.19.0-9-amd64
Build commit: 91f30291
Build date: Sat Feb 27 04:48:54 UTC 2021
Built by: jenkins@mb12-004

#### Additional information you deem important (e.g. issue happens only occasionally):

The issue seems to be in this function generate_block_ip2me_traffic_iptables_commands:

                     ip_ntwrk = ipaddress.ip_network(iface_cidr, strict=False)

                    # For VLAN interfaces, the IP address we want to block is the default gateway (i.e.,
                    # the first available host IP address of the VLAN subnet)
                    ip_addr = next(ip_ntwrk.hosts()) if iface_table_name == "VLAN_INTERFACE" else ip_ntwrk.network_address

                    if isinstance(ip_ntwrk, ipaddress.IPv4Network):
                        block_ip2me_cmds.append(self.iptables_cmd_ns_prefix[namespace] + "iptables -A INPUT -d {}/{} -j DROP".format(ip_addr, ip_ntwrk.max_prefixlen))

 
ip_ntwrk always gets netmask bits masked out, so for 192.168.122.94/24  ip_ntwrk becomes 192.168.122.0/24. And command is formed with ip_ntwrk.max_prefixlen as the netmask which is 32-bit.

bluecmd:
#### Why I did it

Fixes #7008.

Ensure that the destination IP that is permitted is the interface IP.

Currently the logic does not handle IP interfaces correctly and only produces correct results for `/32` or `/128` subnets.

```json
{
    "LOOPBACK_INTERFACE": {
        "Loopback0": {},
        "Loopback0|10.10.10.10/32": {}
    },
    "MGMT_INTERFACE": {
        "eth0|172.18.0.249/24": {
            "gwaddr": "172.18.0.1"
        }
    },
    "VLAN_INTERFACE": {
        "Vlan110": {},
        "Vlan110|172.18.0.250/24": {},
    }
}
```

See `iptables -vnL`
```
    0     0 DROP       all  --  *      *       0.0.0.0/0            10.10.10.10         
    0     0 DROP       all  --  *      *       0.0.0.0/0            172.18.0.0          
    0     0 DROP       all  --  *      *       0.0.0.0/0            172.18.0.1         
```

#### How I did it

Corrected the logic and the iptables commands executed.

It is likely that the initial implementation was only ever tested on L3 interfaces with `/32` and VLAN-interfaces where the switch had the first IP in the network (e.g. `192.168.0.1/24`).

#### How to verify it

 1. Apply patch
 2. Configure some interfaces and VLAN interfaces
 3. Execute `iptables -vnL`

You will see something like this:
```
    0     0 DROP       all  --  *      *       0.0.0.0/0            10.10.10.10          /* Block IP2ME on interface Loopback0 */
    0     0 DROP       all  --  *      *       0.0.0.0/0            172.18.0.250         /* Block IP2ME on interface Vlan110 */
```

Note that MGMT_INTERFACEs are not blocked by default.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012
- [x] 202106
- [x] 202111

caclmgrd: handle IP2ME subnets

image

Note: The original code was added in https://github.com/Azure/sonic-buildimage/pull/4412

### DIFF
--- a/src/sonic-host-services/scripts/caclmgrd
+++ b/src/sonic-host-services/scripts/caclmgrd
@@ -215,18 +215,12 @@ class ControlPlaneAclManager(daemon_base.DaemonBase):
                 for key, _ in iface_table.items():
                     if not _ip_prefix_in_key(key):
                         continue
-
                     iface_name, iface_cidr = key
-                    ip_ntwrk = ipaddress.ip_network(iface_cidr, strict=False)
-
-                    # For VLAN interfaces, the IP address we want to block is the default gateway (i.e.,
-                    # the first available host IP address of the VLAN subnet)
-                    ip_addr = next(ip_ntwrk.hosts()) if iface_table_name == "VLAN_INTERFACE" else ip_ntwrk.network_address
-
-                    if isinstance(ip_ntwrk, ipaddress.IPv4Network):
-                        block_ip2me_cmds.append(self.iptables_cmd_ns_prefix[namespace] + "iptables -A INPUT -d {}/{} -j DROP".format(ip_addr, ip_ntwrk.max_prefixlen))
-                    elif isinstance(ip_ntwrk, ipaddress.IPv6Network):
-                        block_ip2me_cmds.append(self.iptables_cmd_ns_prefix[namespace] + "ip6tables -A INPUT -d {}/{} -j DROP".format(ip_addr, ip_ntwrk.max_prefixlen))
+                    ip_iface = ipaddress.ip_interface(iface_cidr)
+                    if isinstance(ip_iface, ipaddress.IPv4Interface):
+                        block_ip2me_cmds.append(self.iptables_cmd_ns_prefix[namespace] + "iptables -A INPUT -d {} -j DROP -m comment --comment 'Block IP2ME on interface {}'".format(ip_iface.ip, iface_name))
+                    elif isinstance(ip_iface, ipaddress.IPv6Interface):
+                        block_ip2me_cmds.append(self.iptables_cmd_ns_prefix[namespace] + "ip6tables -A INPUT -d {} -j DROP -m comment --comment 'Block IP2ME on interface {}'".format(ip_iface.ip, iface_name))
                     else:
                         self.log_warning("Unrecognized IP address type on interface '{}': {}".format(iface_name, ip_ntwrk))
 
--- a/src/sonic-host-services/scripts/caclmgrd
+++ b/src/sonic-host-services/scripts/caclmgrd
@@ -218,12 +218,28 @@ class ControlPlaneAclManager(daemon_base.DaemonBase):
                         continue
                     iface_name, iface_cidr = key
                     ip_iface = ipaddress.ip_interface(iface_cidr)
+                    ip_ntwrk = ipaddress.ip_network(iface_cidr, strict=False)
+                    ip_first_addr = next(iter(ip_ntwrk.hosts()), None)
+
                     if isinstance(ip_iface, ipaddress.IPv4Interface):
                         block_ip2me_cmds.append(self.iptables_cmd_ns_prefix[namespace] + "iptables -A INPUT -d {} -j DROP -m comment --comment 'Block IP2ME on interface {}'".format(ip_iface.ip, iface_name))
                     elif isinstance(ip_iface, ipaddress.IPv6Interface):
                         block_ip2me_cmds.append(self.iptables_cmd_ns_prefix[namespace] + "ip6tables -A INPUT -d {} -j DROP -m comment --comment 'Block IP2ME on interface {}'".format(ip_iface.ip, iface_name))
                     else:
-                        self.log_warning("Unrecognized IP address type on interface '{}': {}".format(iface_name, ip_ntwrk))
+                        self.log_warning("Unrecognized IP address type on interface '{}': {}".format(iface_name, ip_iface))
+
+                    # For more information about this part, see
+                    # https://github.com/Azure/sonic-buildimage/pull/9826
+                    # For VLAN interfaces, we additionally block the first IP
+                    # in the subnet.
+                    if (iface_table_name == "VLAN_INTERFACE" and
+                        ip_first_addr is not None and ip_first_addr != ip_iface.ip):
+                        if isinstance(ip_ntwrk, ipaddress.IPv4Network):
+                            block_ip2me_cmds.append(self.iptables_cmd_ns_prefix[namespace] + "iptables -A INPUT -d {} -j DROP -m comment --comment 'Block IP2ME (first IP) on interface {}'".format(ip_first_addr, iface_name))
+                        elif isinstance(ip_ntwrk, ipaddress.IPv6Network):
+                            block_ip2me_cmds.append(self.iptables_cmd_ns_prefix[namespace] + "ip6tables -A INPUT -d {} -j DROP -m comment --comment 'Block IP2ME (first IP) on interface {}'".format(ip_first_addr, iface_name))
+                        else:
+                            self.log_warning("Unrecognized IP address type on interface '{}': {}".format(iface_name, ip_ntwrk))
 
         return block_ip2me_cmds
 
--- a/src/sonic-host-services/scripts/caclmgrd
+++ b/src/sonic-host-services/scripts/caclmgrd
@@ -198,9 +198,10 @@ class ControlPlaneAclManager(daemon_base.DaemonBase):
         return tcp_flags_str
 
     def generate_block_ip2me_traffic_iptables_commands(self, namespace):
+        # We do not block MGMT_INTERFACE IPs by default to allow users
+        # to manage the box with default configuration.
         INTERFACE_TABLE_NAME_LIST = [
             "LOOPBACK_INTERFACE",
-            "MGMT_INTERFACE",
             "VLAN_INTERFACE",
             "PORTCHANNEL_INTERFACE",
             "INTERFACE"
--- a/src/sonic-host-services/tests/caclmgrd/caclmgrd_dhcp_test.py
+++ b/src/sonic-host-services/tests/caclmgrd/caclmgrd_dhcp_test.py
@@ -14,19 +14,20 @@ from tests.common.mock_configdb import MockConfigDb
 DBCONFIG_PATH = '/var/run/redis/sonic-db/database_config.json'
 
 
-swsscommon.swsscommon.ConfigDBConnector = MockConfigDb
-test_path = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
-modules_path = os.path.dirname(test_path)
-scripts_path = os.path.join(modules_path, "scripts")
-sys.path.insert(0, modules_path)
-caclmgrd_path = os.path.join(scripts_path, 'caclmgrd')
-caclmgrd = load_module_from_source('caclmgrd', caclmgrd_path)
-
-
 class TestCaclmgrdDhcp(TestCase):
     """
         Test caclmgrd dhcp
     """
+    def setUp(self):
+
+        swsscommon.swsscommon.ConfigDBConnector = MockConfigDb
+        test_path = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+        modules_path = os.path.dirname(test_path)
+        scripts_path = os.path.join(modules_path, "scripts")
+        sys.path.insert(0, modules_path)
+        caclmgrd_path = os.path.join(scripts_path, 'caclmgrd')
+        self.caclmgrd = load_module_from_source('caclmgrd', caclmgrd_path)
+
     @parameterized.expand(CACLMGRD_DHCP_TEST_VECTOR)
     @patchfs
     def test_caclmgrd_dhcp(self, test_name, test_data, fs):
@@ -46,7 +47,7 @@ class TestCaclmgrdDhcp(TestCase):
 
             mark = test_data["mark"]
 
-            caclmgrd_daemon = caclmgrd.ControlPlaneAclManager("caclmgrd")
+            caclmgrd_daemon = self.caclmgrd.ControlPlaneAclManager("caclmgrd")
             mux_update = test_data["mux_update"]
 
             for key,data in mux_update:
--- a/src/sonic-host-services/tests/caclmgrd/caclmgrd_ip2me_test.py
+++ b/src/sonic-host-services/tests/caclmgrd/caclmgrd_ip2me_test.py
@@ -26,6 +26,7 @@ class TestCaclmgrdIP2Me(TestCase):
         sys.path.insert(0, modules_path)
         caclmgrd_path = os.path.join(scripts_path, 'caclmgrd')
         self.caclmgrd = load_module_from_source('caclmgrd', caclmgrd_path)
+        self.maxDiff = None
 
     @parameterized.expand(CACLMGRD_IP2ME_TEST_VECTOR)
     @patchfs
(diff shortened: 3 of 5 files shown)


bluecmd:
I see. This change fixed so all IP2ME interfaces are now blocked, so the tests fail because even management interfaces are now blocked by default. I will fix.

src/sonic-host-services/scripts/caclmgrd
```diff
@@ -215,18 +216,12 @@ class ControlPlaneAclManager(daemon_base.DaemonBase):
                 for key, _ in iface_table.items():
                     if not _ip_prefix_in_key(key):
                         continue
-
                     iface_name, iface_cidr = key
-                    ip_ntwrk = ipaddress.ip_network(iface_cidr, strict=False)
-
-                    # For VLAN interfaces, the IP address we want to block is the default gateway (i.e.,
```
prsunny, line 222: Looks like the comment is to block default g/w and the code change here is intended.

bluecmd: Blocking the default GW in the INPUT table doesn't make sense -- unless you are the default GW in which case my code does the exact same thing.

prsunny: ok

prsunny:
PR is missing the unit tests

bluecmd:
> PR is missing the unit tests

I can create new ones if you want, but there are no tests for this code right now. Let me know if this is something you want me to research how to do.

bluecmd:
I have to find where `test_show_dhcpv6_helper.py::TestDhcpRelayHelper::test_show_dhcpv6_helper_0_DHCPv6_Helpers` lives so I can fix that as well, or adopt my fix to not make it fail. Hints are appreciated, can't seem to find it in a 5 min search.

bluecmd:
@prsunny @yxieca Ready for review, this time with tests! :-)

dockers/docker-dhcp-relay/cli-plugin-tests/test_show_dhcpv6_helper.py
```diff
@@ -34,7 +34,8 @@ def test_show_dhcpv6_helper(self, test_name, test_data, fs):
             fs.create_file(DBCONFIG_PATH) 
         MockConfigDb.set_config_db(test_data["config_db"])
         runner = CliRunner()
-        table = MockConfigDb.get_table(self, "DHCP_RELAY")
+        config_db = MockConfigDb()
+        table = config_db.get_table("DHCP_RELAY")
```
yxieca, line 38: Is this change related to IP2ME?

bluecmd: Yes, the short summary is that it fixes a bug in how MockConfigDb is called from `test_show_dhcpv6_helper`. Without this fix  `MockConfigDb.get_table` will crash when referencing it's `self`. The previous author seems to have assumed that get_table is static, which it is not, and accidentally passed `self` as an argument which causes all kinds of weirdness.

yxieca: Thanks for addressing this issue! just figure that this change should be in a separate PR.

bluecmd:
> @prsunny the change looks good to me. The only thing is that I am not sure if relaxing mgmt cacl aligns with our design?

To be clear, this PR does not relax the cacls for MGMT in **_practice_** - unless I am missing something. My assessment is that the previous MGMT cacls never worked as intended and wouldn't block any traffic unless these very specific scenarios were met:
  - You used a /32 management address.
  - You used the network identity as the MGMT address, e.g. `192.168.0.0/24` as opposed to e.g. `192.168.0.1/24`.

I would be very surprised if anyone used SONiC in these scenarios - and if they did, they would have no way to SSH into the switch as the default factory configuration would block SSH on the MGMT interface. It stands to reason they would have reported an issue.

yxieca (approved):

yxieca:
Let's merge #9903 before this PR.

prsunny:
> > > @prsunny the change looks good to me. The only thing is that I am not sure if relaxing mgmt cacl aligns with our design?
> > 
> > 
> > To be clear, this PR does not relax the cacls for MGMT in **_practice_** - unless I am missing something. My assessment is that the previous MGMT cacls never worked as intended and wouldn't block any traffic unless these very specific scenarios were met:
> > 
> > * You used a /32 management address.
> > * You used the network identity as the MGMT address, e.g. `192.168.0.0/24` as opposed to e.g. `192.168.0.1/24`.
> > 
> > I would be very surprised if anyone used SONiC in these scenarios - and if they did, they would have no way to SSH into the switch as the default factory configuration would block SSH on the MGMT interface. It stands to reason they would have reported an issue.
> 
> Thanks for the clarifications!



> > > @prsunny the change looks good to me. The only thing is that I am not sure if relaxing mgmt cacl aligns with our design?
> > 
> > 
> > To be clear, this PR does not relax the cacls for MGMT in **_practice_** - unless I am missing something. My assessment is that the previous MGMT cacls never worked as intended and wouldn't block any traffic unless these very specific scenarios were met:
> > 
> > * You used a /32 management address.
> > * You used the network identity as the MGMT address, e.g. `192.168.0.0/24` as opposed to e.g. `192.168.0.1/24`.
> > 
> > I would be very surprised if anyone used SONiC in these scenarios - and if they did, they would have no way to SSH into the switch as the default factory configuration would block SSH on the MGMT interface. It stands to reason they would have reported an issue.
> 
> Thanks for the clarifications!

I'm not sure I understand the case here. Currently we install drop rules as below:

1. For Vlan interface, the drop shall be added to the default gw or the first IP address in the subnet
2. For Mgmt interface, it shall be to the first ip (gw) with /32. This will not block anyone from ssh'ing to the device.

Our requirement is to block kernel accepting any packet to the gw IPs.

prsunny (changes requested):
please check comment

bluecmd:
Sure,

> I'm not sure I understand the case here. Currently we install drop rules as below:
> 
>     1. For Vlan interface, the drop shall be added to the default gw or the first IP address in the subnet
> 
>     2. For Mgmt interface, it shall be to the first ip (gw) with /32. This will not block anyone from ssh'ing to the device.
> 
> 
> Our requirement is to block kernel accepting any packet to the gw IPs.

Can you explain why these cases are formulated this way? I suspect these are not the true requirements, but rather a misstaken way to block the IPs configured on the switch. As I explained above, the current behavior does not make any sense as it would be almost impossible for it to block packets in real life scenarios. It does not protect the control plane currently in practice, so the scenario does not make sense as currently formulated in the code.

To avoid talking in circles, we can break it down if we look at the original PR (#4412) these are the relevant scenarios:

- Add iptables rules to drop all packets destined for loopback interface IP addresses
- Add iptables rules to drop all packets destined for management interface IP addresses
- Add iptables rules to drop all packets destined for point-to-point interface IP addresses

These were broken before this PR and only supported `/32` - with this PR it fixes to work with any subnet mask. Do we agree that this PR fixes these issues? 

- Add iptables rules to drop all packets destined for our VLAN interface gateway IP addresses

This one is a bit more difficult. The only interpretation I can make of this sentence that make sense is that the assumption that the SONiC switch **is** the gateway. Thus we can simplify the sentence to be like the other interfaces above; "Add iptables rules to drop all packets destined for VLAN interface IP addresses". In the case where we **are not** the gateway, the packets would never enter the `INPUT` table anyway in the controlplane, so any other interpretation does in my honest opinion seems moot.

Is that answer to your question @prsunny ?

dockers/docker-dhcp-relay/cli-plugin-tests/test_show_dhcpv6_helper.py
bluecmd: Rebased upon new master to remove this part from this PR. Please re-review @yxieca :-).

prsunny:
> Sure,
> 
> > I'm not sure I understand the case here. Currently we install drop rules as below:
> > ```
> > 1. For Vlan interface, the drop shall be added to the default gw or the first IP address in the subnet
> > 
> > 2. For Mgmt interface, it shall be to the first ip (gw) with /32. This will not block anyone from ssh'ing to the device.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Our requirement is to block kernel accepting any packet to the gw IPs.
> 
> Can you explain why these cases are formulated this way? I suspect these are not the true requirements, but rather a misstaken way to block the IPs configured on the switch. As I explained above, the current behavior does not make any sense as it would be almost impossible for it to block packets in real life scenarios. It does not protect the control plane currently in practice, so the scenario does not make sense as currently formulated in the code.
> 
> To avoid talking in circles, we can break it down if we look at the original PR (#4412) these are the relevant scenarios:
> 
> * Add iptables rules to drop all packets destined for loopback interface IP addresses
> * Add iptables rules to drop all packets destined for management interface IP addresses
> * Add iptables rules to drop all packets destined for point-to-point interface IP addresses
> 
> These were broken before this PR and only supported `/32` - with this PR it fixes to work with any subnet mask. Do we agree that this PR fixes these issues?
> 
> * Add iptables rules to drop all packets destined for our VLAN interface gateway IP addresses
> 
> This one is a bit more difficult. The only interpretation I can make of this sentence that make sense is that the assumption that the SONiC switch **is** the gateway. Thus we can simplify the sentence to be like the other interfaces above; "Add iptables rules to drop all packets destined for VLAN interface IP addresses". In the case where we **are not** the gateway, the packets would never enter the `INPUT` table anyway in the controlplane, so any other interpretation does in my honest opinion seems moot.
> 
> Is that answer to your question @prsunny ?

I think you are missing one point from #4412. 

i.e Add iptables rules to drop all packets destined for our VLAN interface gateway IP addresses

Are you saying this was an incorrect requirement? I'm lost here. What is that we are trying to solve and what is the issue you are facing?

prsunny:
Can we discuss this in the next Sonic issue triage meeting?

bluecmd:
> Can we discuss this in the next Sonic issue triage meeting?

Yes, I'd be happy to - where/when can I join? I agree this might be the best way to walk through this.

> I think you are missing one point from #4412.
> 
> i.e Add iptables rules to drop all packets destined for our VLAN interface gateway IP addresses

I am sorry if that wasn't clear, but I tried addressing that in my previous comment in the last paragraph.

> Are you saying this was an incorrect requirement? I'm lost here. What is that we are trying to solve and what is the issue you are facing?

Well, I admit I do not know the high-level requirement - but yes, I believe the requirement as literally stated in the old PR does not make sense.

The problem I am trying to solve: I want the control plane ACLs to block access to my switch's L3 interfaces when not using a `/32` subnet mask.

prsunny:
It Wednesdays 8:00am PST. Please join the sonicproject@googlegroups.com

bluecmd:
Very good. I will join the next one (February 16 it seems). If you want to schedule something before that let me know.

bluecmd:
Related: I found this old comment by @jleveque https://github.com/Azure/sonic-buildimage/issues/4797#issuecomment-646425329 - he seems to have thought the behavior described by @scottlaird indeed was indicative of a bug. I agree with their assessment and this PR indeed fixes the oddity are discussing at that particular point in the conversation.

bluecmd:
From discussions on triage meeting today:

1. We arrived that the `MGMT_INTERFACE` exception is needed in order to allow a default way to SSH to the switch on provision.
2. We agree that we should the existing `VLAN_INTERFACE` behavior as a safety net to preserve current DROPs, even if they are not in use. We can clean these up further along the line if we want.

I will make the adoptions needed for (2).

src/sonic-host-services/scripts/caclmgrd
```diff
@@ -215,20 +216,33 @@ class ControlPlaneAclManager(daemon_base.DaemonBase):
...
+                    ip_iface = ipaddress.ip_interface(iface_cidr)
+                    if isinstance(ip_iface, ipaddress.IPv4Interface):
+                        block_ip2me_cmds.append(self.iptables_cmd_ns_prefix[namespace] + "iptables -A INPUT -d {} -j DROP -m comment --comment 'Block IP2ME on interface {}'".format(ip_iface.ip, iface_name))
+                    elif isinstance(ip_iface, ipaddress.IPv6Interface):
+                        block_ip2me_cmds.append(self.iptables_cmd_ns_prefix[namespace] + "ip6tables -A INPUT -d {} -j DROP -m comment --comment 'Block IP2ME on interface {}'".format(ip_iface.ip, iface_name))
                     else:
-                        self.log_warning("Unrecognized IP address type on interface '{}': {}".format(iface_name, ip_ntwrk))
+                        self.log_warning("Unrecognized IP address type on interface '{}': {}".format(iface_name, ip_iface))
+
+                    # In versions before early 2022 the logic was to also drop packets
```
prsunny, line 228: Please retain the previous logic for VLAN_INTERFACE. and add the fix for IP2ME. Please try to keep the changes minimal and have the comment in PR description. We don't want any version info in the code.

bluecmd: Sounds good, I have tried my best to retain as much of the logic now. It is hard for two reasons:

- To calculate the correct address one should use `ipaddress.ip_interface`, while the previous logic needs `ipaddress.ip_network`.
- This means we need to have two groups of `isinstance` checks unless we want to add another layer of looping. To me it seems more readable to have these two groups.
- Finally, this code was not tested before so there were some valid edge cases that would crash; like having a `/32` on a VLAN interface, and the fact that python3/python2 has changed how iterators work (thus the extra `iter()` call).

bluecmd:
@prsunny What are the next steps here? The tests seems to be failing in the current master branch already, so I cannot provide a PR with all tests working. Do you want to wait until somebody fixes the master branch before merging these, or are you fine accepting this patch with the tests broken in master?

prsunny:
@bluecmd , would you please rebase to latest?

bluecmd:
> @bluecmd , would you please rebase to latest?

Done!

bluecmd:
I will move this PR to https://github.com/sonic-net/sonic-host-services now when this code has moved.

bluecmd:
See https://github.com/sonic-net/sonic-host-services/pull/4 for the first part of this "version 2" attempt :-)

PR #9856, opened 2022-01-26
[Arista] rename management interface in initrd

Staphylo:
#### Why I did it

On some products the pci enumeration adds randomness into which nic gets
initialized first.
Because SONiC doesn't use deterministic interface naming but instead old
style interface naming, this leads to eth0 not always being the
management port.
To make sure eth0 is always the management port (SONiC expectation)
rename the interfaces in the initramfs for Arista products.

#### How I did it

Rename interfaces in the initramfs based on kernel cmdline parameters

#### How to verify it

eth0 is the management port

Rename management interface to always be eth0 on Arista devices


Staphylo:
@prgeor

prgeor (approved):

PR #9857, opened 2022-01-26
[asan] Add address sanitizer support.

oleksandrivantsiv:
Implement infrastructure that allows enabling address sanitizer
for docker containers. Enable address sanitizer for SWSS container.

#### Why I did it
To add a possibility to compile SONiC applications with address sanitizer (ASAN). 
ASAN is a memory error detector for C/C++. It finds:
- Use after free (dangling pointer dereference)
- Heap buffer overflow
- Stack buffer overflow
- Global buffer overflow
- Use after return
- Use after the scope
- Initialization order bugs
- Memory leaks

#### How I did it
By adding new **ENABLE_ASAN** configuration option.
#### How to verify it
By default **ASAN** is disabled and the SONiC image is not affected.
When **ASAN** is enabled it inspects all allocation, deallocation, and memory usage that the application does in run time. To verify whether the application has memory errors tests that trigger memory usage of the application should be run. Ideally, the whole regression tests should be run. Memory leaks reports will be placed in **/var/log/asan/** directory of SONiC host OS.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106


dgsudharsan (dismissed):

prsunny (approved):
lgtm, request Qi/Ying to review as well

dockers/docker-orchagent/Dockerfile.j2
```diff
@@ -70,10 +70,14 @@ COPY ["files/arp_update", "/usr/bin"]
...
 COPY ["enable_counters.py", "tunnel_packet_handler.py", "/usr/bin/"]
-COPY ["docker-init.sh", "orchagent.sh", "swssconfig.sh", "buffermgrd.sh", "/usr/bin/"]
+COPY ["orchagent.sh", "swssconfig.sh", "buffermgrd.sh", "/usr/bin/"]
 COPY ["files/supervisor-proc-exit-listener", "/usr/bin"]
 
 # Copy all Jinja2 template files into the templates folder
 COPY ["*.j2", "/usr/share/sonic/templates/"]
 
+RUN sonic-cfggen -a "{\"ENABLE_ASAN\":\"{{ENABLE_ASAN}}\"}" -t /usr/share/sonic/templates/docker-init.j2 > /usr/bin/docker-init.sh
+RUN rm -f /usr/share/sonic/templates/docker-init.j2
```
yxieca, line 80: Why removing this template?

oleksandrivantsiv: It is needed only during compilation to generate **docker-init.sh** file. It won't be used in run-time

yxieca (approved):

dockers/docker-orchagent/Dockerfile.j2
akokhan: in such case probably it's better to remove it in the same `RUN` as compilation to avoid additional layer to be created

liat-grozovik:
@oleksandrivantsiv please resolve conflicts so we can merge following the approval by Ying

PR #9858, opened 2022-01-26
[pmon] Fix chassis_db_init exit not being expected

alexrallen:
#### Why I did it

Error log was shown on switches during boot
```
 Dec 22 04:27:16.710999 r-bulldog-03 INFO pmon#supervisord 2021-12-22 04:27:16,709 INFO exited: chassis_db_init (exit status 0; not expected)
```

#### How I did it

Add exit code zero as an expected exit code and also disable autorestart. 

#### How to verify it

Boot the switch and ensure the above log line does not appear. 

None. Not a functional bug. 

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [X] 202106
- [X] 202111

[pmon] Fix chassis_db_init exit not being expected

shutterstock_1189520911


dgsudharsan (approved):

sujinmkang:
@alexrallen which platform do you see this error log message and auto restart?
If the platform doesn't need the chassis_db_init then why don't you add skip_chassis_db_init=True to  pmon_daemon_control.json?

alexrallen:
All platforms that I tested on show this (three separate mellanox platforms).

sujinmkang:
@alexrallen Can you add " and IS_MODULAR_CHASSIS == 1" for chassis_db_init similar to chassisd instead of changing the exit code?  chassisd and chassis_db_init are for modular chassis. So it shouldn
t start for non-modular chassis.

alexrallen:
@sujinmkang I am the original author of chassis_db_init. 

This module is intended to populate redis with metadata about the chassis such as model number, serial number, and hardware revision. This information is pulled from the Platform API chassis object. As such, this is able to run on all platforms.

sujinmkang:
@alexrallen The default exit code is 0 so we don't need to set the exit code. But my question is why it crashes with "not expected". Is there anything we can improve the exit / event handling for chassis_db_init?

alexrallen:
The program is correctly exiting with code 0 according to the log. Do you know why else it would say "unexpected"? The program is not crashing it is working fine.

liat-grozovik (approved):

PR #9867, opened 2022-01-27
[Test - Do Not Review] Limit the memory usage of telemetry container.

yozhao101:
Signed-off-by: Ubuntu <yozhao@yozhao-dev.lhaujduh2rsebddzdvd5lg4tuc.phxx.internal.cloudapp.net>

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106

### DIFF
--- a/rules/docker-telemetry.mk
+++ b/rules/docker-telemetry.mk
@@ -29,6 +29,8 @@ endif
 
 $(DOCKER_TELEMETRY)_CONTAINER_NAME = telemetry
 $(DOCKER_TELEMETRY)_RUN_OPT += --privileged -t
+$(DOCKER_TELEMETRY)_RUN_OPT += --memory 400m
+$(DOCKER_TELEMETRY)_RUN_OPT += --oom-kill-disable
 $(DOCKER_TELEMETRY)_RUN_OPT += -v /etc/sonic:/etc/sonic:ro
 $(DOCKER_TELEMETRY)_RUN_OPT += -v /usr/share/sonic/scripts:/usr/share/sonic/scripts:ro
 $(DOCKER_TELEMETRY)_RUN_OPT += -v /var/run/dbus:/var/run/dbus:rw


qiluo-msft (approved):

qiluo-msft:
Could you merge latest master? The build should be fixed.

rules/docker-telemetry.mk
```diff
@@ -29,6 +29,8 @@ endif
 
 $(DOCKER_TELEMETRY)_CONTAINER_NAME = telemetry
 $(DOCKER_TELEMETRY)_RUN_OPT += --privileged -t
+$(DOCKER_TELEMETRY)_RUN_OPT += --memory 400m
```
qiluo-msft, line 32: Should we also limit the CPU, like
```
--cpus="0.3"
```

liuh-80: I think docker will report error when you only set memory limit, in my test, I need set both memory and memory-swap:       --memory="400m" --memory-swap="400m"

liuh-80: Also suggest also update dockers/docker-sonic-telemetry/base_image_files/monit_telemetry, there are following config:
check program container_memory_telemetry with path "/usr/bin/memory_checker telemetry **419430400**"
    if status == 3 for 10 times within 20 cycles then exec "/usr/bin/restart_service telemetry"

Currently this monitor will restart container when memory reach to 400MB, there maybe a risk the container OOM and hang there, but memory usage not reach exactly 400MB. suggest update the container_memory_telemetry parameter to little bit smaller than 400 MB.

yozhao101: > I think docker will report error when you only set memory limit, in my test, I need set both memory and memory-swap: --memory="400m" --memory-swap="400m"

Can you please specify what is error message docker will report  if we only set `--memory 400m`? 

What I observed on lab device is if we set `--memory 400m` and `--oom-kill-disable` and use the `stress` utility to consumes memory up to 1GB, then the commands `docker stats` and `docker stats --no-stream` will be struck without any response.

If we set both `--memory 400m` and `--memory-swap 400m`, this will prevent telemetry container swapping memory pages to disk: https://docs.docker.com/config/containers/resource_constraints/#--memory-swap-details 

I prefer setting both flags to disable the swapping.

yozhao101: > Should we also limit the CPU, like
> 
> ```
> --cpus="0.3"
> ```

I am verifying whether this setting is necessary or not and will update later.

rules/docker-telemetry.mk
```diff
@@ -29,6 +29,10 @@ endif
...
+$(DOCKER_TELEMETRY)_RUN_OPT += --memory-swap 400m
+$(DOCKER_TELEMETRY)_RUN_OPT += --cpus 0.3
+$(DOCKER_TELEMETRY)_RUN_OPT += --oom-kill-disable
```
qiluo-msft, line 35: >oom [](http://example.com/codeflow?start=33&length=3)

Offline discussed. will update this PR and will also update the Monit script memory_checkerto read the memory usage of telemetry from file instead of docker stats.

rules/docker-telemetry.mk
liuh-80: After I check again, I think it's not necessary to set the --memory-swap option:

1. SONiC not enable swap, there are only 2 lines of config in fstab file
    https://github.com/Azure/sonic-buildimage/blob/bdfe89c4ad22cf396f4ea87a2ccdcb918a8eb564/build_debian.sh

         ## Config basic fstab
         sudo LANG=C chroot $FILESYSTEM_ROOT /bin/bash -c 'echo "proc /proc proc defaults 0 0" >> /etc/fstab'
         sudo LANG=C chroot $FILESYSTEM_ROOT /bin/bash -c 'echo "sysfs /sys sysfs defaults 0 0" >> /etc/fstab'


2. For the --memory options, there will be a warning message on some system:
    WARNING: Your kernel does not support swap limit capabilities or the cgroup is not mounted. Memory limited without swap.

PR #9880, opened 2022-01-28
[yang] Fixing groupings when grouping is in the same module file

ghooo:
#### Why I did it
If the grouping is in the same file in the module, fetching the grouping fails

#### How I did it
fixing the `uses_module_name` when the grouping is under the same file

#### How to verify it
Enabled a grouping under the same file which is `lldp`, there is a test in sonic-yang-mgmt that translates `sample-config-db` into yang format. This test passes with grouping in `lldp` used.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106


praveen-li (approved):
LGTM

qiluo-msft (approved):

PR #9890, opened 2022-01-29
Close console session due to user inactivity

prgeor:
Signed-off-by: Prince George <prgeor@microsoft.com>

#### Why I did it
The default sudo [password prompt timeout](https://www.sudo.ws/docs/man/1.7.10/sudoers.man/#:~:text=disable%20word%20wrap\).-,passwd_timeout,-Number%20of%20minutes) is NOT working. This is causing the console session to be held up even though the user in active for a long period of time. Due to this, TMOUT or auto-logout is not taking effect because user cannot login to bash, leaving the console unusable.

#### How I did it
Explicitly set the password prompt timeout in sudoers

#### How to verify it
Verified 
1. The sudo password prompt timeout after 5 mins of inactivity
2. After user has logged in to bash via console, the session timeout if user has not pressed key for more than 15 mins (TMOUT)

- [x] 201811
- [x] 201911
- [x] 202006
- [x] 202012
- [x ] 202106

### DIFF
--- a/files/image_config/bash/bash.bashrc
+++ b/files/image_config/bash/bash.bashrc
@@ -55,7 +55,7 @@ if [ -x /usr/lib/command-not-found -o -x /usr/share/command-not-found/command-no
 fi
 
 # Automatically log out console ttyS* sessions after 15 minutes of inactivity
-tty | grep ttyS >/dev/null && TMOUT=900
+tty | egrep -q '^/dev/ttyS[[:digit:]]+$' && TMOUT=900
 
 # if SSH_TARGET_CONSOLE_LINE was set, attach to console line interactive cli directly
 if [ -n "$SSH_TARGET_CONSOLE_LINE" ]; then
--- a/files/image_config/sudoers/sudoers
+++ b/files/image_config/sudoers/sudoers
@@ -60,6 +60,9 @@ ALL     ALL=NOPASSWD: READ_ONLY_CMDS
 # Prevent password related command into syslog
 Defaults!PASSWD_CMDS !syslog
 
+# Make sure sudo password prompt times out after 5 mins
+Defaults passwd_timeout=5
+
 # See sudoers(5) for more information on "#include" directives:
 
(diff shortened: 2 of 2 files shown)


files/image_config/sudoers/sudoers
```diff
@@ -60,6 +60,9 @@ ALL     ALL=NOPASSWD: READ_ONLY_CMDS
 # Prevent password related command into syslog
 Defaults!PASSWD_CMDS !syslog
 
+# Make sure sudo password prompt times out after 5 mins
+Defaults passwd_timeout=5
```
qiluo-msft, line 64: >passwd_timeout [](http://example.com/codeflow?start=9&length=14)

Is it a well known issue in upstream? Do you have a tracking link for this bug?

files/image_config/bash/bash.bashrc
```diff
@@ -55,7 +55,7 @@ if [ -x /usr/lib/command-not-found -o -x /usr/share/command-not-found/command-no
 fi
 
 # Automatically log out console ttyS* sessions after 15 minutes of inactivity
-tty | grep ttyS >/dev/null && TMOUT=900
+tty | egrep -q '^/dev/ttyS[[:digit:]]+$' && TMOUT=900
```
qiluo-msft, line 58: >egrep [](http://example.com/codeflow?start=6&length=5)

Is it a bug fix or refactor?

qiluo-msft (approved):

files/image_config/bash/bash.bashrc
prgeor: refactor

files/image_config/sudoers/sudoers
prgeor: sonic bug [https://msazure.visualstudio.com/One/_workitems/edit/13121024/](https://msazure.visualstudio.com/One/_workitems/edit/13121024/)

prgeor:
@rlhui please help merge this PR to 201911

PR #9934, opened 2022-02-08
[radv] Support multiple ipv6 prefixes per vlan interface

kellyyeh:
#### Why I did it
Radvd.conf.j2 template creates two copies of the vlan interface when there are more than one ipv6 address assigned to a single vlan interface. Changed the format to add prefixes under the same vlan interface block.

#### How I did it
Modifies radvd.conf.j2 and added unit tests

#### How to verify it
Configure multiple ipv6 address to the same vlan, start radvd
Unit test will check if radvd.conf with multiple ipv6 addresses is formed correctly

- [ ] 201811
- [x] 201911
- [ ] 202006
- [x] 202012
- [ ] 202106


dockers/docker-router-advertiser/radvd.conf.j2
```diff
@@ -23,15 +34,15 @@ interface {{ name }}
     AdvOtherConfigFlag off;
     AdvLinkMTU 9100;
     AdvHomeAgentFlag off;
+{% for prefix in prefixes %}
```
prsunny, line 37: do we have some unit tests for this?

kellyyeh: not yet. Is there existing unit test for radv? Should it be added to https://github.com/Azure/sonic-buildimage/tree/master/src/sonic-config-engine/tests/sample_output ?

prsunny: yes, lets try add it there

dockers/docker-router-advertiser/radvd.conf.j2
```diff
@@ -10,9 +10,19 @@
 
 {# If our configuration has VLAN interfaces... #}
 {% if VLAN_INTERFACE %}
+{% set vlan_list = dict() %}
+{% for vlan_name in VLAN %}
```
prsunny, line 14: why we need to loop on VLAN table and check if `name==vlan_name`? I think to keep it same logic as before, can't we just need to loop VLAN_INTERFACE?

kellyyeh: VLAN table gives the vlan names, and VLAN_INTERFACE gives tuples like: "Vlan1000, fc02:1000::1" , "Vlan1000, fc02:2000::1". My intent is to create a prefix_list for every VLAN, and loop through VLAN_INTERFACE to add all ipv6 to corresponding VLAN's prefix_list.

prsunny (dismissed):
looks good now, thanks

prsunny (approved):

abdosi:
@kellyyeh please create direct PR for 201911 as there is cherry-pick conflict.

cc @prsunny

qiluo-msft:

Issue #9742, opened 2022-01-12
[yang-models] POLICER table missing


wen587:
#### Description
POLICER yang is needed. It is used to set rate limit for mirror-session

#### Steps to reproduce the issue:
1. Sample configDB table about policer.
```
admin@vlab-01:~/everflow$ cat policer.json
{
  "POLICER": {
    "everflow_static_policer": {
    "meter_type": "bytes",
    "mode": "sr_tcm",
    "cir": "12500000",
    "cbs": "12500000",
    "red_packet_action": "drop"
    }
  }
}
```
2. `apply-patch` to add policer
```
admin@vlab-01:~/everflow$ sudo config apply-patch -i /FEATURE -i /QUEUE -i /SCHEDULER add_policer.json
Patch Applier: Patch application starting.
Patch Applier: Patch: [{"op": "add", "path": "/POLICER", "value": {"everflow_static_policer": {"meter_type": "bytes", "mode": "sr_tcm", "cir": "12500000", "cbs": "12500000", "red_packet_action": "drop"}}}]
Patch Applier: Getting current config db.
Patch Applier: Simulating the target full config after applying the patch.
Patch Applier: Validating target config does not have empty tables, since they do not show up in ConfigDb.
Patch Applier: Sorting patch updates.
Note: Below table(s) have no YANG models:
BGP_PEER_RANGE, CONSOLE_SWITCH, DEVICE_NEIGHBOR_METADATA, DHCP_SERVER, KDUMP, RESTAPI, SNMP, SNMP_COMMUNITY, TELEMETRY, POLICER,
Note: Below table(s) have no YANG models:
POLICER,
Failed to apply patch
Usage: config apply-patch [OPTIONS] PATCH_FILE_PATH
Try "config apply-patch -h" for help.

Error: Given YANG covered config patch is not valid because it has changes to tables without YANG models

```
3.

#### Describe the results you received:
Action failed.

#### Describe the results you expected:
Yang model should provide verification.

#### Output of `show version`:

```
admin@vlab-01:~/everflow$ show ver

SONiC Software Version: SONiC.master.58770-6402a0226
Distribution: Debian 11.1
Kernel: 5.10.0-8-2-amd64
Build commit: 6402a0226
Build date: Mon Dec 13 09:02:14 UTC 2021
Built by: AzDevOps@sonic-build-workers-000Z9S

Platform: x86_64-kvm_x86_64-r0
HwSKU: Force10-S6000
ASIC: vs
[...]

Issue #9743, opened 2022-01-12
[yang-models] sonic-mirror-session.yang miss policer leaf


wen587:
#### Description
sonic-mirror-session.yang doesn't support policer where it is needed for rate limit

#### Steps to reproduce the issue:
1. Sample mirror session config contains policer.
```
admin@vlab-01:~/everflow$ cat session.json
{
  "MIRROR_SESSION": {
    "mirror_session_dscp": {
    "dscp": "5",
    "dst_ip": "2.2.2.2",
    "policer": "everflow_static_policer",
    "src_ip": "1.1.1.1",
    "ttl": "64",
    "type": "ERSPAN"
    }
  }
}

```
2.
3.

#### Describe the results you expected:
We need to add Yang model for policer first then add the leaf for sonic-mirror-session.yang.

#### Output of `show version`:

```
admin@vlab-01:~/everflow$ show ver

SONiC Software Version: SONiC.master.58770-6402a0226
Distribution: Debian 11.1
Kernel: 5.10.0-8-2-amd64
Build commit: 6402a0226
Build date: Mon Dec 13 09:02:14 UTC 2021
Built by: AzDevOps@sonic-build-workers-000Z9S

Platform: x86_64-kvm_x86_64-r0
HwSKU: Force10-S6000
ASIC: vs
ASIC Count: 1
Serial Number: N/A
Model Number: N/A
Hardware Revision: N/A
Uptime: 08:30:18 up  2:47,  2 users,  load average: 0.21, 0.35, 0.27

```
#### Output of `show techsupport`:

```
(paste your output here or download and attach the file here )
```

wen587:
POLICER yang missing should be solved first to fix above issue.
https://github.com/Azure/sonic-buildimage/issues/9742

zhangyanzhao:
@rupesh-k can you please help to take a look? Thanks.

PR #9948, opened 2022-02-09
[yang]YANG model for policer table

dgsudharsan:
#### Why I did it
Added yang model for policer table
Fixes https://github.com/Azure/sonic-buildimage/issues/9742 and https://github.com/Azure/sonic-buildimage/issues/9743
#### How I did it
Creating yang model for policer

#### How to verify it
Added UT to verify the yang model

The configuration schema for policer is added in the pull request https://github.com/Azure/sonic-swss/pull/2144

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106


src/sonic-yang-models/tests/yang_model_tests/tests/policer.json
```diff
@@ -0,0 +1,61 @@
+{
+    "POLICER_TABLE": {
+    "desc": "Configure policer with all fields."
```
qiluo-msft, line 3: >"desc" [](http://example.com/codeflow?start=4&length=6)

Could you reformat this json file?

src/sonic-yang-models/yang-templates/sonic-policer.yang.j2
```diff
@@ -0,0 +1,126 @@
...
+					description "Policer color Source";
+				} 
+
+				leaf cir {
+					type uint64;
+					default 0;
+					description
+						"Committed information rate for the dual-rate token
+						bucket policer.  This value represents the rate at which
+						tokens are added to the primary bucket.";
```
ganglyu, line 57: what's the unit? kbps?

src/sonic-yang-models/yang-templates/sonic-policer.yang.j2
```diff
@@ -0,0 +1,126 @@
...
+					}
+					must "(current() >= current()/../cir)" {
+						error-message "cbs must be greater than or equal to cir";
+					}
+					type uint64;
+					default 0;
+					description
+						"Committed burst size for the dual-rate token bucket
+						policer.  This value represents the depth of the token
+						bucket.";
```
ganglyu, line 72: what's the unit? bytes?

src/sonic-yang-models/yang-templates/sonic-policer.yang.j2
```diff
@@ -0,0 +1,126 @@
...
+						"Committed information rate for the dual-rate token
+						bucket policer.  This value represents the rate at which
+						tokens are added to the primary bucket.";
+				}
+
+				leaf cbs {
+					must "((current()/../cir) and (current()/../cir > 0))" {
+						error-message "cbs can't be configured without cir.";
+					}
+					must "(current() >= current()/../cir)" {
```
ganglyu, line 64: Please refer to https://datatracker.ietf.org/doc/html/rfc4115.
The CBS and EBS are measured in bytes and must configure to be greater than the expected maximum length of the incoming PDU.  The CIR and EIR are both measured in bits/s.
I think we need to convert unit for comparing.

dgsudharsan:
@ganglyu  Please refer to copp yang table. I have reused the same definitions used there. Basically copp internally uses policer and both should be aligned in yang definitions and so I defined as in https://github.com/Azure/sonic-buildimage/blob/master/src/sonic-yang-models/yang-models/sonic-copp.yang

src/sonic-yang-models/yang-templates/sonic-policer.yang.j2
dgsudharsan: It depends on the meter type defined. The meter type can be bytes or packets. This field by itself has no units. Please refer to https://github.com/opencomputeproject/SAI/blob/16c11ee7062e630f1eaf4bb3e44d31ab58eafa5b/inc/saipolicer.h#L123

src/sonic-yang-models/yang-templates/sonic-policer.yang.j2
dgsudharsan: It depends on the meter type defined. The meter type can be bytes or packets. This field by itself has no units. Please refer to https://github.com/opencomputeproject/SAI/blob/16c11ee7062e630f1eaf4bb3e44d31ab58eafa5b/inc/saipolicer.h#L123

src/sonic-yang-models/yang-templates/sonic-policer.yang.j2
dgsudharsan: Please refer to my earlier comments regarding the units.

ganglyu: Can you update the descriptions for CIR, CBS, PIR and PBS to add unit?

src/sonic-yang-models/yang-templates/sonic-policer.yang.j2
ganglyu: > @ganglyu Please refer to copp yang table. I have reused the same definitions used there. Basically copp internally uses policer and both should be aligned in yang definitions and so I defined as in https://github.com/Azure/sonic-buildimage/blob/master/src/sonic-yang-models/yang-models/sonic-copp.yang

Can you update the descriptions for CIR, PIR, CBS and PBS to add unit?

ganglyu:
> @ganglyu Please refer to copp yang table. I have reused the same definitions used there. Basically copp internally uses policer and both should be aligned in yang definitions and so I defined as in https://github.com/Azure/sonic-buildimage/blob/master/src/sonic-yang-models/yang-models/sonic-copp.yang

Can you update the descriptions for CIR, PIR, CBS and PBS to add unit?

src/sonic-yang-models/tests/yang_model_tests/tests/policer.json
dgsudharsan: Done

dgsudharsan:
> > @ganglyu Please refer to copp yang table. I have reused the same definitions used there. Basically copp internally uses policer and both should be aligned in yang definitions and so I defined as in https://github.com/Azure/sonic-buildimage/blob/master/src/sonic-yang-models/yang-models/sonic-copp.yang
> 
> Can you update the descriptions for CIR, PIR, CBS and PBS to add unit?

Done

ganglyu (approved):

liat-grozovik:
@lguohan could you please signoff?

qiluo-msft (approved):

PR #9969, opened 2022-02-11
[AS9716-32D] Modify thermal code to python3

jostar-yang:
Signed-off-by: Jostar Yang <jostar_yang@accton.com.tw>

#### Why I did it
1. Modify thermal code from python2 to python3
2. Fix thermal code bug
#### How I did it
1. Remove py2 code and use py3 api or format
2. Fix thermal bug
#### How to verify it
Test accton_as9716_32d_monitor.py and there is no error. And it can read fan and thermal sysfs.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012
- [x] 202106


jostar-yang:
Add test information.
Add state=MAX when fan is plugged out.
Plug out one fan , other fan speed will change to 100%.
But plug in the fan back , all fan speed still are 100%.
- How to verify it
Run thermal plan service, plug out fan-6.
Check fan speed=100%. Insert back fan-6, and check fan speed should according to thermal plan.

PR #9986, opened 2022-02-15
[containerd]Fixing container commands when mode is local and state is disabled

dgsudharsan:
#### Why I did it
During warm-reboot and fast-reboot the below error logs appear
Feb 3 22:05:15.187408 r-lionfish-13 ERR container: docker cmd: kill for nat failed with 404 Client Error for http+docker://localhost/v1.41/containers/nat/json: Not Found ("No such container: nat")

The container command when called for local mode doesn't check if it is enabled before calling docker kill which throws the above errors.
https://github.com/Azure/sonic-utilities/blob/b6ca76b4821453171d9607383b808385968eeeeb/scripts/fast-reboot#L699

#### How I did it
Checking feature state if local mode and returning error exit code along with valid debug message.

#### How to verify it
Manually tested with warm-reboot and fast-reboot
Added UT to verify it.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [x] 202111


src/sonic-ctrmgrd/ctrmgr/container
```diff
@@ -389,24 +403,26 @@ def main():
     args = parser.parse_args()
     kwargs = {}
 
+    ret = 0
```
renukamanavalan, line 406: Currently unused
Do we need?

dgsudharsan: Hi Renuka. Do you mean the variable ret? It is used below where the return value is returned at the end of function so when the process exits it will have appropriate return code.

src/sonic-ctrmgrd/ctrmgr/container
```diff
@@ -301,19 +306,25 @@ def container_kill(feature, **kwargs):
...
+    debug_msg("{}: set_owner:{} current_owner:{} remote_state:{} docker_id:{} state:{}".format(
+        feature, set_owner, current_owner, remote_state, docker_id, state))
 
     if remove_label:
         set_label(feature, False)
 
+    if set_owner == "local":
+        if state not in ["enabled", "always_enabled"]:
+            debug_msg("{} is not enabled".format(feature))
+            return FAILURE
```
liat-grozovik, line 324: @dgsudharsan should we return failure in this case? if just means nothing to kill and the docker command will not executed

src/sonic-ctrmgrd/tests/container_test.py
```diff
@@ -498,6 +523,24 @@ def test_kill(self, mock_docker, mock_table, mock_conn):
...
+    @patch("container.docker.from_env")
+    def test_invalid_kill(self, mock_docker, mock_table, mock_conn):
+        self.init()
+        common_test.set_mock(mock_table, mock_conn, mock_docker)
+
+        for (i, ct_data) in invalid_kill_test_data.items():
+            common_test.do_start_test("container_test:container_kill", i, ct_data)
+
+            ret = container.container_kill("sflow")
+
```
liat-grozovik, line 537: extra new line, i believe should be removed.

src/sonic-ctrmgrd/tests/container_test.py
```diff
@@ -498,6 +523,24 @@ def test_kill(self, mock_docker, mock_table, mock_conn):
...
+            assert ret != 0
+            ret = common_test.check_tables_returned()
```
liat-grozovik, line 539: add new line?

src/sonic-ctrmgrd/ctrmgr/container
dgsudharsan: The failure return code is to signify the fact that the user input is wrong (docker is not running) and cannot be killed. This error code is returned when the container command is executed which is inline with many other linux commands.

renukamanavalan: This PR is to avoid the failure.
It was "_No such container: nat_"
Would the caller (warm/fast reboot) would still fail, because of this return code?

dgsudharsan: No. The error code will not fail the warm-reboot and its validated.
https://github.com/Azure/sonic-utilities/blob/b6ca76b4821453171d9607383b808385968eeeeb/scripts/fast-reboot#L699

renukamanavalan (approved):

PR #9992, opened 2022-02-15
Enable pulling the base slave docker image by default

saiarcot895:
The base slave image is built roughly every night. Instead of defaulting
to building it locally every time, try to pull this image from the
registry. This saves a bit of time, and only the user-specific container
has to be built.

#### Why I did it

Building the base slave image can take 30 minutes at least. However, there's already a nightly task to build the slave image. Depending on when the user tries to build an image, the base slave image may already be available, and that can be used instead of rebuilding locally.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111


saiarcot895:
@lguohan @xumia Could you review this? Also, are there any pipelines that would need `ENABLE_DOCKER_BASE_PULL=n` to be explicitly specified, so that they rebuild the slave containers?

PR #10008, opened 2022-02-17
[memory_monitoring] Enhance the feature of monitoring the memory usage of containers

yozhao101:

#### Why I did it
Currently we have a PR (https://github.com/Azure/sonic-buildimage/pull/9867/files) which will put a hard limitation on memory usage of telemetry container.

During testing the above PR, we found that the commands `docker stats`, `docker start`, `docker stop` and `docker exec` will be no response if memory usage of telemetry container reaches or exceeds the configured hard limitation. This behavior will fail the background script `memory_checker` and `restart_service` since they depend on the command output of `docker stats`, operation of `docker start` and `docker stop`,

#### How I did it

1. Instead of depending on the output of `docker stats`, the background script `memory_checker` will calculate the memory usage of a container based on values reading from the cgroup subsystem files `memory.usage_in_bytes` and `memory.stats`. 

      I refer to this Docker official document (https://docs.docker.com/engine/reference/commandline/stats/#extended-description) to make sure the memory usage of a specific container reading from command output of `docker stats` is equal to the value subtracting cache usage from the total memory usage.

2. Initially the `restart_service` script will leverage the command `sudo systemctl restart <container_name>.service` to restart the specified container. However, this command will fail since the `docker stop` and `docker start` will be no response if memory usage exceeds the limitation.

    

#### How to verify it
I manually tested this implementation on DuT `str-msn2700-20`.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x ] 202012
- [ ] 202106
- [ ] 202111

### DIFF
--- a/files/image_config/monit/memory_checker
+++ b/files/image_config/monit/memory_checker
@@ -25,15 +25,17 @@ import sys
 import syslog
 import re
 
+CGROUP_DOCKER_MEMORY_DIR = "/sys/fs/cgroup/memory/docker/"
+
 
 def get_command_result(command):
-    """Executes the command and return the resulting output.
+    """Executes command and returns command's stdout.
 
     Args:
         command: A string contains the command to be executed.
 
     Returns:
-        A string which contains the output of command.
+        A string contains command's stdout.
     """
     command_stdout = ""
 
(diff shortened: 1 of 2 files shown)


FengPan-Frank (commented):
There's conflicts exist on the PR, I update it into https://github.com/sonic-net/sonic-buildimage/pull/19179 instead, will reply and address comments there as well.

PR #10036, opened 2022-02-19
[swss]: Wait for vlan intf to start ndppd

theasianpianist:

#### Why I did it
If the VLAN interface is not up when ndppd starts, it will fail to enable `allmulti` mode on the interface and be unable to process received NDP packets

The following logs are seen:
```
/var/log/syslog.33.gz:Feb 18 10:33:12.825406 sonic INFO swss#/supervisord: ndppd (error) Failed to set allmulti: No such device
```
#### How I did it
Use the `wait_for_link` script currently used by `radv` to delay ndppd startup until the vlan interface is ready

#### How to verify it
Apply the changes to a device. `config reload` the device and confirm that the above error logs are not observed when ndppd starts. Run the `arp/test_arp_dualtor.py::test_proxy_arp` test case and verify it passes. 

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

### DIFF
--- a/dockers/docker-orchagent/wait_for_link.sh.j2
+++ b/dockers/docker-orchagent/wait_for_link.sh.j2
@@ -1,0 +1,30 @@
+#!/usr/bin/env bash
+
+VLAN_TABLE_PREFIX="VLAN_TABLE"
+
+function wait_until_iface_ready
+{
+    TABLE_PREFIX=$1
+    IFACE=$2
+
+    echo "Waiting until interface $IFACE is ready..."
+
+    # Wait for the interface to come up
+    # (i.e., interface is present in STATE_DB and state is "ok")
+    while true; do
+        RESULT=$(sonic-db-cli STATE_DB HGET "${TABLE_PREFIX}|${IFACE}" "state" 2> /dev/null)
+        if [ x"$RESULT" == x"ok" ]; then
+            break
+        fi
+
+        sleep 1
+    done
+
+    echo "Interface ${IFACE} is ready!"
+}
+
+
+# Wait for all interfaces to be up and ready
+{% for (name, prefix) in VLAN_INTERFACE|pfx_filter %}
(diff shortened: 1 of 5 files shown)


dockers/docker-orchagent/wait_for_link.sh.j2
```diff
@@ -0,0 +1,30 @@
...
+{
+    TABLE_PREFIX=$1
+    IFACE=$2
+
+    echo "Waiting until interface $IFACE is ready..."
+
+    # Wait for the interface to come up
+    # (i.e., interface is present in STATE_DB and state is "ok")
+    while true; do
+        RESULT=$(sonic-db-cli STATE_DB HGET "${TABLE_PREFIX}|${IFACE}" "state" 2> /dev/null)
```
prsunny, line 15: During SWSS init, there is a cleanup of STATE_DB tables. Can you please confirm, this script is always executed after that cleanup?

theasianpianist: This script waits for `swssconfig.sh` to exit, and `swssconfig.sh` [waits for the `/ready` flag](https://github.com/Azure/sonic-buildimage/blob/a7f613038338093964a1d5299eb3e4f4fcbd0f0b/dockers/docker-orchagent/swssconfig.sh#L39) which is only created after the [DBs are cleared](https://github.com/Azure/sonic-buildimage/blob/a7f613038338093964a1d5299eb3e4f4fcbd0f0b/files/scripts/swss.sh#L175)

prsunny (approved):

dockers/docker-orchagent/wait_for_link.sh.j2
prsunny: So just to confirm, ndppd won't get started on a T1 topology with this change, right?

PR #10043, opened 2022-02-20
sonic-slave-buster pins the versions of Jinja2 and MarkupSafe in py3

qiluo-msft:
#### Why I did it
Upstream breaking change, ref discussion https://github.com/pallets/markupsafe/issues/282

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

### DIFF
--- a/sonic-slave-buster/Dockerfile.j2
+++ b/sonic-slave-buster/Dockerfile.j2
@@ -465,6 +465,11 @@ RUN pip3 uninstall -y enum34
 
 # For templating
 RUN pip2 install j2cli==0.3.10
+# Note: Jinja2 depends on MarkupSafe, however markupsafe 2.1.0 breaks Jinja2 2.10
+# Debian buster dist-packages include python3-markupsafe (1.1.0-1) and python3-jinja2 (2.10-2)
+# If not pinning the versions, any requirement like `MarkupSafe>=2.0` will pull latest into site-packages and mess up.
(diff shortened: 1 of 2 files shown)


sonic-slave-buster/Dockerfile.j2
```diff
@@ -465,6 +465,11 @@ RUN pip3 uninstall -y enum34
 
 # For templating
 RUN pip2 install j2cli==0.3.10
+# Note: Jinja2 depends on MarkupSafe, however markupsafe 2.1.0 breaks Jinja2 2.10
+# Debian buster dist-packages include python3-markupsafe (1.1.0-1) and python3-jinja2 (2.10-2)
+# If not pinning the versions, any requirement like `MarkupSafe>=2.0` will pull latest into site-packages and mess up.
+RUN pip3 install MarkupSafe==2.0.1
+RUN pip3 install Jinja2==3.0.3
```
davidism, line 472: Since you're pinning to Jinja2==3.0.3, you can pin to MarkupSafe==2.1.0. It's only if you're using an old version of Jinja that you'd need to pin an old version of MarkupSafe as well.

xumia (approved):

sonic-slave-buster/Dockerfile.j2
qiluo-msft: We will consider upgrade MarkupSafe version later.

PR #10064, opened 2022-02-22
[bbr] Add ability to disable bbr for storage backend

neethajohn:
#### Why I did it
BBR is enabled by default and needs to be disabled for storage backend

#### How I did it
Introduce a new var 'storage_backend_state' in constants.yml which will be used to set/unset bbr configurations for storage backend devices

#### How to verify it
Unit test and tested the change on storage backend device

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012
- [x] 202106
- [x] 202111

### DIFF
--- a/files/image_config/constants/constants.yml
+++ b/files/image_config/constants/constants.yml
@@ -32,6 +32,7 @@ constants:
     bbr:
       enabled: true
       default_state: "disabled"
+      storage_backend_state: "disabled"
     peers:
       general: # peer_type
         db_table: "BGP_NEIGHBOR"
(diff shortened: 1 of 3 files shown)


files/image_config/constants/constants.yml
```diff
@@ -32,6 +32,7 @@ constants:
     bbr:
       enabled: true
```
prsunny, line 33: Can we set this to 'disabled' in case of storage backend?

neethajohn: This file is a common file and not a j2. So setting this to disabled will disable bbr for all roles.

prsunny: agree, i was thinking more like changing this to j2

neethajohn:
Based on internal discussions, closing this PR

PR #10075, opened 2022-02-24
Fix Headroom value for 7260C64 SKU

abdosi:
What I did:
Updated the Headroom value for (100G,5m) in 7260C64 SKU.

Why I did:
Previous value were not correct.


abdosi:
cc @yxieca

prgeor (approved):

Issue #9929, opened 2022-02-08
[yang] missing MIRROR_INGRESS_ACTION and MIRROR_EGRESS_ACTION in sonic-acl.yang


wen587:
#### Description
missing MIRROR_INGRESS_ACTION and MIRROR_EGRESS_ACTION in sonic-acl.yang
which is used for EverflowAlwaysOn in Generic Config Updater E2E test.

#### Steps to reproduce the issue:
1. 
```
admin@vlab-01:~/everflow$ sudo config acl add table EVERFLOW_DSCP MIRROR_DSCP

admin@vlab-01:~/everflow$ sudo config apply-patch -i /FEATURE -i /QUEUE -i /SCHEDULER add_rule.json
Patch Applier: Patch application starting.
Patch Applier: Patch: [{"op": "add", "path": "/ACL_RULE", "value": {"EVERFLOW_DSCP|RULE_1": {"DSCP": "5", "MIRROR_INGRESS_ACTION": "mirror_session_dscp", "PRIORITY": "9999"}}}]
Patch Applier: Getting current config db.
Patch Applier: Simulating the target full config after applying the patch.
Patch Applier: Validating target config does not have empty tables, since they do not show up in ConfigDb.
Patch Applier: Sorting patch updates.
Note: Below table(s) have no YANG models:
BGP_PEER_RANGE, CONSOLE_SWITCH, DEVICE_NEIGHBOR_METADATA, DHCP_SERVER, KDUMP, RESTAPI, SNMP, SNMP_COMMUNITY, TELEMETRY,
sonic_yang(3):All Keys are not parsed in ACL_RULE
dict_keys(['EVERFLOW_DSCP|RULE_1'])
sonic_yang(3):exceptionList:["'MIRROR_INGRESS_ACTION'"]   <=====================
sonic_yang(3):Data Loading Failed:All Keys are not parsed in ACL_RULE
dict_keys(['EVERFLOW_DSCP|RULE_1'])
Failed to apply patch
Usage: config apply-patch [OPTIONS] PATCH_FILE_PATH
Try "config apply-patch -h" for help.

Error: Given patch is not valid because it will result in an invalid config
```
2.
3.

#### Describe the results you received:
See above

#### Describe the results you expected:
no error expected

#### Output of `show version`:

```
admin@vlab-01:~/everflow$ show ver

SONiC Software Version: SONiC.master.59050-dirty-20211214.171137
Distribution: Debian 11.1
[...]

PR #10100, opened 2022-02-28
[yang] support acl MIRROR_ACTION

wen587:
Fixes https://github.com/Azure/sonic-buildimage/issues/9929
#### Why I did it
ACL doesn't have mirror related action
#### How I did it
Add 'MIRROR_INGRESS_ACTION' and 'MIRROR_EGRESS_ACTION' to sonic-acl.yang.j2
#### How to verify it
Run the YANG model unit tests

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111


dgsudharsan (changes requested):
Please add sample configuration here https://github.com/Azure/sonic-buildimage/blob/master/src/sonic-yang-models/tests/files/sample_config_db.json

wen587:
> Please add sample configuration here https://github.com/Azure/sonic-buildimage/blob/master/src/sonic-yang-models/tests/files/sample_config_db.json

Added.

dgsudharsan (approved):

wen587:
@ganglyu @qiluo-msft @bingwang-ms  could you help review. Thanks

src/sonic-yang-models/yang-templates/sonic-acl.yang.j2
```diff
@@ -65,6 +69,18 @@ module sonic-acl {
...
+					type leafref {
+						path "/sms:sonic-mirror-session/sms:MIRROR_SESSION/sms:MIRROR_SESSION_LIST/sms:name";
+					}
+				}
+
+				leaf MIRROR_EGRESS_ACTION {
+					type leafref {
+						path "/sms:sonic-mirror-session/sms:MIRROR_SESSION/sms:MIRROR_SESSION_LIST/sms:name";
+					}
+				}
```
ganglyu, line 82: There are INGRESS_ACTION and EGRESS_ACTION, I think we can check acl rule. For example, acl stage is ingress, MIRROR_EGRESS_ACTION is invalid.

wen587: https://github.com/Azure/sonic-mgmt/blob/master/tests/everflow/everflow_test_utilities.py#L147-L150
Looks valid action.

bingwang-ms (approved):

PR #10104, opened 2022-02-28
Image build time improvements

saiarcot895:
#### Why I did it

The last step of the SONiC build (building `sonic-vs.bin` or `sonic-broadcom.bin` or `sonic-aboot-broadcom.bin`) seemed to take a long time. I started looking at what it was doing and what could be done to reduce the time spent there.

#### How I did it

There are two major changes in this PR (see the individual commits for more details on what was done there):

1. The first one is an issue being caused by debootstrap that would cause the `/proc` filesystem within the slave container to get umounted. This meant that `/proc` is now an empty folder. There were two visible side effects to this: each invocation of sudo added 0.8 seconds to the runtime (sudo itself would take 0.8 seconds more, _then_ execute the actual command), and from the host machine, trying to `docker exec` into the container while the Debian image build was happening would fail, because it expects `/proc/self/fd` to be available. This is fixed in debootstrap 1.0.124, but since there's no backported version available for Bullseye, I am rebuilding debootstrap with the patch added in.
2. The second one is to try to reduce the `fsync` calls done during deb package installations by dpkg. dpkg calls `fsync` many times, to make sure all new files are written to disk and that the filesystem is in a consistent state, so that if some system crash were to happen, it would be somewhat possible to recover the system. However, because of the nature of the build that we're doing, this is not really important for us. The `eatmydata` package can suppress these `fsync` calls and potentially improve application performance, at the risk of losing data or having the filesystem in an inconsistent state in the event of a system crash.

#### How to verify it

On my dev VM, before any of these changes, building the virtual switch image took 28 minutes (this is only counting the part of the build where `build_debian.sh` is run to build the base Debian image, load in all of the containers, and make image files). After the debootstrap patch, this decreased to 22 minutes. With the use of `eatmydata`, this decreased to 19-20 minutes. Actual numbers will vary depending on the system.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111


lguohan (approved):

build_debian.sh
```diff
@@ -113,6 +113,12 @@ sudo cp files/apt/apt.conf.d/{81norecommends,apt-{clean,gzip-indexes,no-language
 ## Note: set lang to prevent locale warnings in your chroot
 sudo LANG=C chroot $FILESYSTEM_ROOT apt-get -y update
 sudo LANG=C chroot $FILESYSTEM_ROOT apt-get -y upgrade
+
+echo '[INFO] Install and setup eatmydata'
+sudo LANG=C chroot $FILESYSTEM_ROOT apt-get -y install eatmydata
+sudo ln -s /usr/bin/eatmydata $FILESYSTEM_ROOT/usr/local/bin/dpkg
```
qiluo-msft, line 119: >eatmydata [](http://example.com/codeflow?start=20&length=9)

Why create this symlink? If needed, install it in slave?

build_debian.sh
```diff
@@ -568,6 +574,10 @@ scripts/collect_host_image_version_files.sh $TARGET_PATH $FILESYSTEM_ROOT
 # Remove GCC
 sudo LANG=C DEBIAN_FRONTEND=noninteractive chroot $FILESYSTEM_ROOT apt-get -y remove gcc
 
+# Remove eatmydata
+sudo rm $FILESYSTEM_ROOT/etc/apt/apt.conf.d/00image-install-eatmydata $FILESYSTEM_ROOT/usr/local/bin/dpkg
+sudo LANG=C DEBIAN_FRONTEND=noninteractive chroot $FILESYSTEM_ROOT apt-get -y remove eatmydata
```
qiluo-msft, line 579: >eatmydata [](http://example.com/codeflow?start=85&length=9)

After removing in chroot, you have a dead symlink outside.

build_debian.sh
saiarcot895: The symlink is being created within the chroot for use within the chroot (the link is being created at `$FILESYSTEM_ROOT/usr/local/bin/dpkg` and pointing to `/usr/bin/eatmydata`). When chrooted, this allows all calls to `dpkg` to actually call `eatmydata`, which will then look for the actual binary in `PATH`.

build_debian.sh
saiarcot895: The symlink is being created inside the chroot, and is being deleted on line 578, so there shouldn't be any dead symlink in the slave container.

build_debian.sh
qiluo-msft: You mean something like below?
```
sudo  LANG=C chroot $FILESYSTEM_ROOT ln -s /usr/bin/eatmydata /usr/local/bin/dpkg
```

If yes, one minor comment: it is not obvious that the intention is `ln` inside chroot.

saiarcot895: Yes. I was trying to avoid the chroot here, but I'll change this to use chroot to make it clear.

saiarcot895: This line has now been changed to use `chroot` to make it clear that the symlink is meant for use inside the chroot.

saiarcot895:
@qiluo-msft can you re-review this?

@xumia looks like the change you did for making sure `/proc` is mounted would've fixed the first change in this PR. I can remove this part of the change if you want.

xumia:
> @qiluo-msft can you re-review this?
> 
> @xumia looks like the change you did for making sure `/proc` is mounted would've fixed the first change in this PR. I can remove this part of the change if you want.

Yes, I have a PR to make sure the /proc is mounted in the slave container. It should not be a correct usage to umount /proc, has impact on multiple programs, including the debootstrap (there is a warning message, complain /proc not mounted).
If mount/umount the /proc from the fsroot filesystem, then it is good. It is my understanding, the command "umount_on_exit /proc" is running in the fsroot, right?

saiarcot895:
Correct, `umount_on_exit` works on filesystems within the fsroot.

yxieca:
@qiluo-msft Are we ok to merge this PR?

qiluo-msft (approved):

PR #10112, opened 2022-03-01
[Inspur] Modify MGMT code for IPv4/IPv6

inspurSDN:
#### Why I did it
     We can set IPv4 and IPv6 separately.
#### How I did it
     We always send the dhcp request and get the dhcp ip, when we set the static ip, we will check the ip in redis DB.
     When the static IP is found in redis DB, and then the dhcp ip will be released. 
#### How to verify it
     Set the static IP for IPv6 and DHCP ip for IPv4. And the we can check the IP.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111


lguohan:
is this in master image already? if not, please submit patch to master branch.

PR #10072, opened 2022-02-23
[Mellanox] Fix DPB supported breakout modes

alexrallen:
#### Why I did it

Removing DPB breakout modes that require adjacent ports to be disabled as that is not supported by the current DPB infrastructure.

Correspondingly had to remove the `hwsku.json` file from any SKUs which utilized these removed modes such that the system will fall back to `ports_config.ini` and DPB will not be supported for those SKUs.

#### How I did it

Modified the `platform.json` files of Mellanox devices.

#### How to verify it

Execute `show int break [Ethernet]` on the affected platforms and ensure there are no modes present that would require an adjacent port to be disabled to function. 

#### Which release branch to backport (provide reason below if selected)

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [x] 202111

#### Description for the changelog
[mellanox] Fix DPB supported breakout modes

#### A picture of a cute animal (not mandatory but encouraged)
borzoi_body_image_2

PR #10128, opened 2022-03-02
[202111] [Mellanox] Fix DPB supported breakout modes

alexrallen:
**Cherry pick of https://github.com/Azure/sonic-buildimage/pull/10072**

#### Why I did it

Removing DPB breakout modes that require adjacent ports to be disabled as that is not supported by the current DPB infrastructure.

Correspondingly had to remove the `hwsku.json` file from any SKUs which utilized these removed modes such that the system will fall back to `ports_config.ini` and DPB will not be supported for those SKUs.

#### How I did it

Modified the `platform.json` files of Mellanox devices.

#### How to verify it

Execute `show int break [Ethernet]` on the affected platforms and ensure there are no modes present that would require an adjacent port to be disabled to function. 

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [x] 202111

[mellanox] Fix DPB supported breakout modes

borzoi_body_image_2


alexrallen:
@judyjoseph Please see cherry-pick here.

liat-grozovik (approved):

Issue #10096, opened 2022-02-27
[ERSPAN] Hexadecimal values are not accepted for GRE parameter


raphaelt-nvidia:
#### Description

In June 2021, I opened https://github.com/Azure/sonic-buildimage/issues/7990 [mirroring] Missing type validations in config mirror-session command.  This was addressed by https://github.com/Azure/sonic-utilities/pull/1825 Validate input of config mirror_session add.  In particular, that PR introduced

GRE_TYPE_RANGE = click.IntRange(min=0, max=65535)

The result of this change is that whereas previously, any value was accepted for the GRE parameter, including garbage, now only a DECIMAL integer between 0 and 65535 is accepted.

The command reference https://github.com/Azure/sonic-utilities/blob/master/doc/Command-Reference.md#mirroring-config-commands doesn't define it explicitly, but the following text suggests that hex format should at least be allowed, if not preferred:

optional - GRE Type in case if user wants to send the packet via GRE tunnel. GRE type could be anything; it could also be left as empty; by default, it is 0x8949 for Mellanox; and 0x88be for the rest of the chips.

@bingwang-ms What do you think the correct validation should be?  Accept decimal, hexadecimal, both?

#### Steps to reproduce the issue:

sudo config mirror_session erspan add ERSPAN 100.0.0.1 100.0.0.2 0 11 0x8949

#### Describe the results you received:

Usage: config mirror_session erspan add [OPTIONS] <session_name> <src_ip>
                                        <dst_ip> <dscp> <ttl> [gre_type]
                                        [queue] [src_port] [direction]
Try "config mirror_session erspan add -h" for help.

Error: Invalid value for "[gre_type]": 0x8949 is not a valid integer

#### Describe the results you expected:

No error, command accepted.

#### Output of `show version`:

SONiC Software Version: SONiC.master.289-44028723e_Internal
Distribution: Debian 11.2
Kernel: 5.10.0-8-2-amd64
Build commit: 44028723e
Build date: Thu Feb 24 09:14:41 UTC 2022
Built by: sw-r2d2-bot@r-build-sonic-ci02-243

Platform: x86_64-nvidia_sn5600_simx-r0
HwSKU: ACS-SN5600
ASIC: mellanox
ASIC Count: 1
Serial Number: MT2022X08597
Model Number: MSN4700-WS2FO
Hardware Revision: N/A
Uptime: 13:55:31 up 11:41,  1 user,  load average: 0.00, 0.00, 0.00

[...]

PR #10140, opened 2022-03-03
Update YANG model for mirror session to support decimal value for GRE type

bingwang-ms:
Signed-off-by: bingwang <bingwang@microsoft.com>

#### Why I did it
PR  https://github.com/Azure/sonic-utilities/pull/1825 added validation for the input of `config mirror session add`, and only decimal value is accepted.
An issue https://github.com/Azure/sonic-buildimage/issues/10096 was raised to suggest accepting HEX value as well, and the suggestion makes sense to me.

To accept HEX value for GRE type, and keep backward compatibility as well, I updated the YANG model to support both decimal and hexadecimal input for GRE type.

#### How I did it
Update the regex for GRE type.

#### How to verify it
Verified by UT
```
platform linux -- Python 3.9.2, pytest-6.0.2, py-1.10.0, pluggy-0.13.0
rootdir: /sonic/src/sonic-yang-models
plugins: pyfakefs-4.5.4, cov-2.10.1
collected 3 items                                                                                                                                                                                     

tests/test_sonic_yang_models.py ..                                                                                                                                                              [ 66%]
tests/yang_model_tests/test_yang_model.py .                                                                                                                                                     [100%]

========================================================================================== 3 passed in 2.53s ==========================================================================================
```

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

Update YANG model for mirror session to support decimal value for GRE type.

### DIFF
--- a/src/sonic-yang-models/tests/yang_model_tests/tests_config/mirror_session.json
+++ b/src/sonic-yang-models/tests/yang_model_tests/tests_config/mirror_session.json
@@ -31,6 +31,22 @@
             }
         }
     },
+    "MIRROR_ERSPAN_ENTRY_WITH_VALID_DEC_VALUES_1": {
+        "sonic-mirror-session:sonic-mirror-session": {
+            "MIRROR_SESSION": {
+                "MIRROR_SESSION_LIST": [
+                    {
+                        "name": "erspan",
+                        "type": "ERSPAN",
+                        "dst_ip": "11.1.1.1",
+                        "src_ip": "10.1.1.1",
+                        "gre_type": "65535",
+                        "dscp": "10"
+                    }
+                ]
+            }
+        }
+    },
     "MIRROR_ERSPAN_ENTRY_WRONG_TYPE": {
         "sonic-mirror-session:sonic-mirror-session": {
             "MIRROR_SESSION": {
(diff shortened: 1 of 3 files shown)


src/sonic-yang-models/yang-models/sonic-mirror-session.yang
```diff
@@ -99,7 +99,7 @@ module sonic-mirror-session {
                 leaf gre_type {
                     when "current()/../type = 'ERSPAN'";
                     type string {
-                        pattern "0[xX][0-9a-fA-F]*";
+                        pattern "0[xX][0-9a-fA-F]*|[1-9][0-9]*";
```
dgsudharsan, line 102: I don't think this should be the pattern. This would allow values out of range. The pattern should restrict values from 0-65535.

dgsudharsan (changes requested):

src/sonic-yang-models/yang-models/sonic-mirror-session.yang
bingwang-ms: Good catch. I updated the pattern for matching decimal, and some test cases as well. Not we only accept values in range [0, 65535].
Please be noticed that values as `0x0` and `0` are also accepted.

bingwang-ms:
@dgsudharsan Please help take a look one more time. Thanks

src/sonic-yang-models/tests/yang_model_tests/tests_config/mirror_session.json
```diff
@@ -15,6 +15,70 @@
...
+                        "dst_ip": "11.1.1.1",
+                        "src_ip": "10.1.1.1",
+                        "gre_type": "0",
+                        "dscp": "10"
+                    }
+                ]
+            }
+        }
+    },
+    "MIRROR_ERSPAN_ENTRY_WITH_VALID_DEC_VALUES_3": {
```
dgsudharsan, line 66: Isn't this hex value example for GRE type?

bingwang-ms: Good catch. I fixed the typo. Thanks

dgsudharsan (approved):

dgsudharsan:
@qiluo-msft  Can you please review and sign off?

qiluo-msft (approved):

PR #10160, opened 2022-03-07
[YANG ] Support vlan sub intf short naming format

lolyu:
#### Why I did it
To support vlan sub intf short naming format, like subport `Eth8.100` for parent port`Ethernet8` with vlan id `100`.

#### How I did it
Add checks to `must` condition to verify there is a port or portchannel with the index same as the vlan sub intf.

#### How to verify it
Run the unttests.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [x] 202111

### DIFF
--- a/src/sonic-yang-models/yang-models/sonic-vlan-sub-interface.yang
+++ b/src/sonic-yang-models/yang-models/sonic-vlan-sub-interface.yang
@@ -14,6 +14,10 @@ module sonic-vlan-sub-interface {
         prefix port;
     }
 
+    import sonic-portchannel {
+        prefix lag;
+    }
+
     import sonic-vrf {
         prefix vrf;
     }
--- a/src/sonic-yang-models/yang-models/sonic-vlan-sub-interface.yang
+++ b/src/sonic-yang-models/yang-models/sonic-vlan-sub-interface.yang
@@ -36,9 +36,11 @@ module sonic-vlan-sub-interface {
                 key "name";
 
                 leaf name {
-                    must "substring-before(current(), '.') = /port:sonic-port/port:PORT/port:PORT_LIST/port:name"
+                    must "(substring-before(current(), '.') = /port:sonic-port/port:PORT/port:PORT_LIST/port:name) or " +
+                         "(starts-with(substring-before(current(), '.'), 'Eth') and " +
+                         "concat('Ethernet', substring-after(substring-before(current(), '.'), 'Eth')) = /port:sonic-port/port:PORT/port:PORT_LIST/port:name)"
                     {
-                        error-message "Must condition not satisfied, Try adding PORT: {}, Example: 'Ethernet0': {}";
+                        error-message "Must condition not satisfied, please follow vlan sub interface naming convention";
                     }
 
                     // check if the vlan sub interface have the form as <portname>.<vlan_id>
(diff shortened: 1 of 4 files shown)


src/sonic-yang-models/yang-models/sonic-vlan-sub-interface.yang
```diff
@@ -36,9 +36,11 @@ module sonic-vlan-sub-interface {
                 key "name";
 
                 leaf name {
-                    must "substring-before(current(), '.') = /port:sonic-port/port:PORT/port:PORT_LIST/port:name"
+                    must "(substring-before(current(), '.') = /port:sonic-port/port:PORT/port:PORT_LIST/port:name) or " +
+                         "(starts-with(substring-before(current(), '.'), 'Eth') and " +
+                         "concat('Ethernet', substring-after(substring-before(current(), '.'), 'Eth')) = /port:sonic-port/port:PORT/port:PORT_LIST/port:name)"
```
neethajohn, line 41: shouldn't the left side be concat('Ethernet', substring-after(current(), '.'))

src/sonic-yang-models/yang-models/sonic-vlan-sub-interface.yang
```diff
@@ -36,9 +36,11 @@ module sonic-vlan-sub-interface {
                 key "name";
 
                 leaf name {
-                    must "substring-before(current(), '.') = /port:sonic-port/port:PORT/port:PORT_LIST/port:name"
+                    must "(substring-before(current(), '.') = /port:sonic-port/port:PORT/port:PORT_LIST/port:name) or " +
+                         "(starts-with(substring-before(current(), '.'), 'Eth') and " +
```
neethajohn, line 40: should we not check for an exact match for 'Eth' instead of starts-with?

lolyu: This is implicitly checked in LINE41.
If the sub intf name is `Eth40.100`, LINE41 will ensure the port `Ethernet40` with port index as `40` substring between `Eth` and `.` is existed in the port list. If the sub intf name is something like `Ether40.100`, it will fail the check.

src/sonic-yang-models/yang-models/sonic-vlan-sub-interface.yang
lolyu: ` substring-after(current(), '.')` will return the vlan id, what we need is the substring between `Eth` and `.`, which contains the port index, right?

ganglyu (approved):

src/sonic-yang-models/yang-models/sonic-vlan-sub-interface.yang
neethajohn: My bad. you are right

ganglyu (dismissed):

dgsudharsan (changes requested):
Can you please add an example of short naming format here 
https://github.com/Azure/sonic-buildimage/blob/d112e7cca9087d02892ada9d70bddcc5d5616c24/src/sonic-yang-models/tests/files/sample_config_db.json#L900

dgsudharsan (approved):

neethajohn (approved):

qiluo-msft (dismissed):

ganglyu (approved):

lolyu:
I've rebased this PR to the newest master branch, all the approvals are dismissed. Could you please help review again, thanks! @neethajohn @ganglyu @qiluo-msft

ganglyu (approved):

qiluo-msft (approved):

PR #10166, opened 2022-03-07
Add a config variable to override default container registry instead of dockerhub.

liushilongbuaa:
#### Why I did it
Add a config variable DEFAULT_CONTAINER_REGISTRY to override default container registry.
#### How I did it
It can be set in config file and can be overridden by command line.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111


xumia:
@liushilongbuaa , thanks for the contribution to make the container registry endpoint configurable.
Please help add some tests to use the new ACR, and not break the docker reproducible build feature, thanks.

Makefile.work
```diff
@@ -112,9 +112,20 @@ endif
...
 include rules/config
 -include rules/config.user
 
+ifeq ($(SONIC_DEFAULT_CONTAINER_REGISTRY),)
+SONIC_DEFAULT_CONTAINER_REGISTRY := $(DEFAULT_CONTAINER_REGISTRY)
+endif
+ifneq ($(SONIC_DEFAULT_CONTAINER_REGISTRY),)
+SONIC_DEFAULT_CONTAINER_REGISTRY := $(SONIC_DEFAULT_CONTAINER_REGISTRY)/
+endif
+
```
xumia, line 128: Looks like the logic can be simplified, reduce one if block.

DEFAULT_CONTAINER_REGISTRY is from rule/config, SONIC_DEFAULT_CONTAINER_REGISTRY is the user input.
We can simply override the DEFAULT_CONTAINER_REGISTRY, when SONIC_DEFAULT_CONTAINER_REGISTRY is not empty.

ifneq ($(SONIC_DEFAULT_CONTAINER_REGISTRY),)
DEFAULT_CONTAINER_REGISTRY := $(SONIC_DEFAULT_CONTAINER_REGISTRY)
endif

When passing the value to j2, we can define a new parameter, for example:

ifneq ($(DEFAULT_CONTAINER_REGISTRY),)
DEFAULT_CONTAINER_REGISTRY_PREFIX:= $(DEFAULT_CONTAINER_REGISTRY)/
endif

rules/config
```diff
@@ -211,3 +211,6 @@ INCLUDE_MUX = y
 
 # ENABLE_ASAN - enable address sanitizer
 ENABLE_ASAN ?= n
+
+# reset default container registry from dockerhub to other
+DEFAULT_CONTAINER_REGISTRY=
```
xumia, line 216: Change to:
DEFAULT_CONTAINER_REGISTRY ?=

liushilongbuaa:
> @liushilongbuaa , thanks for the contribution to make the container registry endpoint configurable. Please help add some tests to use the new ACR, and not break the docker reproducible build feature, thanks.

test finished.

Makefile
```diff
@@ -66,6 +66,7 @@ init:
 #
 # Function to invoke target $@ in Makefile.work with proper BLDENV
 #
+export DEFAULT_CONTAINER_REGISTRY
```
xumia, line 69: Remove it?

xumia (approved):

PR #10172, opened 2022-03-07
Image disk space reduction

saiarcot895:
#### Why I did it

Reduce the disk space taken up during bootup and runtime.

#### How I did it

1. Remove python package cache from the base image and from the containers.
2. During bootup, if logs are to be stored in memory, then don't create the `var-log.ext4` file just to delete it later during bootup.
3. For the partition containing `/host`, don't reserve any blocks for just the root user. This just makes sure all disk space is available for all users, if needed during upgrades (for example).

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012
- [x] 202106
- [x] 202111

### DIFF
--- a/files/initramfs-tools/varlog
+++ b/files/initramfs-tools/varlog
@@ -11,16 +11,26 @@ case $1 in
     ;;
 esac
 
+logs_inram=false
+
 # Extract kernel parameters
 set -- $(cat /proc/cmdline)
 for x in "$@"; do
     case "$x" in
         varlog_size=*)
             varlog_size="${x#varlog_size=}"
+            ;;
+        logs_inram=on)
+            logs_inram=true
+            ;;
     esac
 done
 
 [ -z "$varlog_size" ] && exit 0
+
+# If logs are being stored in memory, then don't bother
+# creating the log file just to have it deleted afterwards.
+$logs_inram && exit 0
 
 # exit when the var_log.ext4 exists and the size matches
 if [ -e "${rootmnt}/host/disk-img/var-log.ext4" ]; then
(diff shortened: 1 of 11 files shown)


installer/x86_64/platforms_asic
```diff
@@ -0,0 +1,3 @@
+x86_64-kvm_x86_64-r0
+x86_64-kvm_x86_64_4_asic-r0
+x86_64-kvm_x86_64_6_asic-r0
```
xumia, line 3: Do we need to merge it? Or it will be auto-generated?

saiarcot895: No, this is not expected. Looks like an auto-generated file got added in.

xumia (approved):

files/initramfs-tools/varlog
```diff
@@ -11,17 +11,27 @@ case $1 in
...
+            logs_inram=true
+            ;;
     esac
 done
 
 [ -z "$varlog_size" ] && exit 0
 
+# If logs are being stored in memory, then don't bother
+# creating the log file just to have it deleted afterwards.
+$logs_inram && exit 0
```
yxieca, line 33: I've seen this syntax causing trouble when -x was used. Something related to newer version of sh treat unmatched condition as failure. It would be safer to use if statement.
I can see that this syntax was used at line 29 as well.

Not making this issue blocking.

yxieca (approved):

files/initramfs-tools/varlog
saiarcot895: Yeah true, I was a bit surprised that it was working here, but for now, I matched the syntax on line 29.

qiluo-msft:

judyjoseph:
@saiarcot895, commit could not be cleanly cherry-picked to 202111. Please submit another PR.

PR #10175, opened 2022-03-08
[Build]: Fix installing dpkg packages in parallel issue

xumia:
#### Why I did it
Fix the debian packages installing in parallel issue.
Add apt hook command to support apt to print no version control info.

#### How I did it
Add a file lock, it can block the apt/apt-get installation for 10 min in max if another apt/apt-get installation command is running. It only has effect on the build time.

Clarify the feature to the lock in slave.mk:
Currently, we already has a lock in the slave.mk, it has no effect when installing the package inside of the makefile. Comparing the method, the method in the pr is more general, take effect when using the command apt/apt-get.
And it is not as a replacement for the debian installation lock in the slave.mk, another feature of the lock in slave.mk is to block some of the conflict packages installation, so make the conflict targets not building in the same time.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012
- [x] 202106
- [x] 202111

### DIFF
--- a/src/sonic-build-hooks/hooks/apt
+++ b/src/sonic-build-hooks/hooks/apt
@@ -1,4 +1,5 @@
 #!/bin/bash
 
 . /usr/local/share/buildinfo/scripts/buildinfo_base.sh
-APT_REAL_COMMAND=$(get_command apt) ./apt-get "$@"
+
+APT_REAL_COMMAND=$(get_command apt) $(dirname "$0")/apt-get "$@"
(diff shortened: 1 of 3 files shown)


liushilongbuaa (approved):

qiluo-msft:
Could you explain the original problem? What is the error message?
I am confused at `dpkg` in title, seem you are only fixing `apt` or `apt-get`.

Issue #9885, opened 2022-01-28
VS image build failed when P4RT container is enabled


zhangineer:
#### Steps to reproduce the issue:
1. Proceed with the normal building process to choose `PLATFORM=vs`
2. Execute `make SONIC_BUILD_JOBS=4 INSTALL_DEBUG_TOOLS=y target/sonic-vs.bin`

#### Describe the results you received:

ERROR: invalid type '' for SAI_ACL_COUNTER_ATTR_TABLE_ID
ERROR: flags are not defined for SAI_ACL_COUNTER_ATTR_TABLE_ID
ERROR: FATAL ERROR === MUST FIX === : Can't use an undefined value as an ARRAY reference at parse.pl line 1676.

make[4]: *** [Makefile:107: saimetadata.c] Error 1
make[4]: Leaving directory '/sonic/src/sonic-sairedis/SAI/meta'
make[3]: *** [Makefile:1144: saimeta-gen] Error 2
make[3]: Leaving directory '/sonic/src/sonic-sairedis/meta'
make[2]: *** [Makefile:444: install-recursive] Error 1
make[2]: Leaving directory '/sonic/src/sonic-sairedis'
dh_auto_install: error: make -j1 install DESTDIR=/sonic/src/sonic-sairedis/debian/tmp AM_UPDATE_INFO_DIR=no returned exit code 2
make[1]: *** [debian/rules:59: binary-syncd-vs] Error 25
make[1]: Leaving directory '/sonic/src/sonic-sairedis'
dpkg-buildpackage: error: fakeroot debian/rules binary-syncd-vs subprocess returned exit status 2
[  FAIL LOG END  ] [ target/debs/bullseye/libsairedis_1.0.0_amd64.deb ]
make: *** [slave.mk:511: target/debs/bullseye/libsairedis_1.0.0_amd64.deb] Error 1
make: *** Waiting for unfinished jobs....
[ 01 ] [ target/debs/bullseye/linux-headers-5.10.0-8-2-common_5.10.46-4_all.deb ]
make[1]: *** [Makefile.work:311: target/sonic-vs.bin] Error 2
make[1]: Leaving directory '/home/sonic/sonic-buildimage'
make: *** [Makefile:33: target/sonic-vs.bin] Error 2

#### Describe the results you expected:
Expected no errors with successful image build. When I disabled the P4RT container, the image build process went through successfully.

zhangyanzhao:
[...]

PR #10182, opened 2022-03-08
[PINS] update sonic-p4rt docker to bullseye

rhalstea:
#### Why I did it
SONiC is migrating to bullseye. This will update the sonic-pins container to bullseye.

#### How I did it
The [sonic-pins code](https://github.com/Azure/sonic-buildimage/blob/master/rules/p4rt.mk) isn't dependent on any architecture so it will already build successfully for bullseye. This PR updates the docker to use bullseye.

#### How to verify it
Today we cannot build the docker-sonic-p4rt.gz target (e.g. Issue #9885). With this change the docker will build successfully. The P4RT executable will not run, because of a missing runtime library, libgmpxx, which I'll address in a followup PR.

None.

Update docker-sonic-p4rt.gz target to build with Bullseye instead of Buster.

N/A


bocon13:
Thanks @rhalstea for working on this!

Here's the fix on the packaging side: https://github.com/Azure/sonic-pins/pull/4
After that's merged, we'll need to just update the submodule.

pettershao-ragilenetworks:
@rhalstea I faced the same build error when I build libsairedis locally, how can I fix this, thanks!:
```
ERROR: invalid type '' for SAI_ACL_COUNTER_ATTR_TABLE_ID
ERROR: flags are not defined for SAI_ACL_COUNTER_ATTR_TABLE_ID
ERROR: FATAL ERROR === MUST FIX === : Can't use an undefined value as an ARRAY reference at parse.pl line 1676.
```

rhalstea:
@pettershao-ragilenetworks I haven't looked into the siaredis issue. This fix is addressing the incompatibilities in the p4rt docker. Have you tried the suggestions in this comment: https://github.com/Azure/sonic-buildimage/issues/9885#issuecomment-1035239310

lguohan (approved):

rhalstea:
Thanks @lguohan, Is there anything else needed before we merge this?

qiluo-msft (approved):

PR #10210, opened 2022-03-11
Add failure details in sonic-config-engine unit test.

liushilongbuaa:
#### Why I did it
When tests in sonic-config-engine failed, there is no detail massage.
#### How I did it
Using existing function to show failure details when assert failed.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

### DIFF
--- a/src/sonic-config-engine/tests/test_j2files.py
+++ b/src/sonic-config-engine/tests/test_j2files.py
@@ -1,4 +1,3 @@
-import filecmp
 import json
 import os
 import shutil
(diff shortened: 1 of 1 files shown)


src/sonic-config-engine/tests/common_utils.py
```diff
@@ -40,7 +40,11 @@ def cmp(file1, file2):
             obj1 = json.load(f)
         with open(file2, 'r') as f:
             obj2 = json.load(f)
-        return obj1 == obj2
+        if obj1 == obj2:
+            return true
+        else:
+            os.system("git diff --no-index " + file1 +" "+ file2)
```
xumia, line 46: I would like to make the feature more general, without git dependency.
User can copy the code in a none-git environment to do test. For the feature, it has no dependency on git.

https://stackoverflow.com/questions/28838170/multilevel-json-diff-in-python

xumia (approved):

Issue #9643, opened 2021-12-23
config-setup factory failed to parse platform.json


mykolaxgerasymenko:
#### Description
If the platform.json file includes some fields with a boolean type value, an error occurred while parsing platform.json during `config-setup factory`.

#### Steps to reproduce the issue:
1) Deploy SONiC image 
2) Added to platform.json file to node psus one more field with boolean type "temperature": false. 
For example: 

```
        "psus": [ 
            { 
                "name": "psu-1", 
                "temperature": false 
            }, 
            { 
                "name": "psu-2", 
                "temperature": false 
            } 
        ], 
```

3) Remove config_db.json file 

`sudo rm -f /etc/sonic/config_db.json `

4) Configure new config_db.json file 

`sudo config-setup factory `

**Note:** 
platform.json file must include an "interfaces" key with real interfaces inside, which describes all info about them. 
For example: 

```
    "interfaces": { 
        "Ethernet0": { 
            "index": "1,1,1,1,1,1,1,1", 
            "lanes": "0,1,2,3,4,5,6,7", 
            "breakout_modes": { 
                "1x400G[200G]": ["Ethernet0"], 
                "2x200G[100G,40G]": ["Ethernet0, Ethernet4"], 
                "4x100G[50G]": ["Ethernet0", "Ethernet2", "Ethernet4", "Ethernet6"], 
                "8x50G[25G,10G]": ["Ethernet0", "Ethernet1", "Ethernet2", "Ethernet3", "Ethernet4", "Ethernet5", "Ethernet6", "Ethernet7"], 
                "1x200G[100G,40G](4)+4x50G[25G,10G](4)": ["Ethernet0", "Ethernet4", "Ethernet5", "Ethernet6", "Ethernet7"], 
                "4x50G[25G,10G](4)+1x200G[100G,40G](4)": ["Ethernet0", "Ethernet1", "Ethernet2", "Ethernet3", "Ethernet4"] 
            } 
        }, 
    } 
```

If the "interfaces" key is empty, then the portconfig.py script will be use port_config.ini instead of platform.json to configure config_db.json file. 

#### Describe the results you received:
Error logs: 

```
admin@sonic:~$ sudo config-setup factory 
Traceback (most recent call last): 
  File "/usr/local/bin/sonic-cfggen", line 445, in <module> 
    main() 
  File "/usr/local/bin/sonic-cfggen", line 306, in main 
[...]

PR #10224, opened 2022-03-11
[sonic-config-engine][portconfig] Do not parse JSON as Python AST

vboykox:

#### Why I did it
To fix https://github.com/Azure/sonic-buildimage/issues/9643

#### How I did it
Instead of ast.literal_eval added python2 compat code for json strings unicode -> str convertion.

We need python2 compatibility since py2 sonic config engine (buster/sonic_config_engine-1.0-py2-none-any.whl target) is still included into the build (ENABLE_PY2_MODULES flag is set for buster). Once we abandon buster and python2, this compat and ast.literal_eval could be cleaned up all through the code base.

#### How to verify it
run steps from the linked issue

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [x] 202106
- [x] 202111

### DIFF
--- a/src/sonic-config-engine/portconfig.py
+++ b/src/sonic-config-engine/portconfig.py
@@ -45,16 +45,25 @@ BRKOUT_PATTERN_GROUPS = 6
 #
 # Helper Functions
 #
+
+# For python2 compatibility
+def py2JsonStrHook(j):
+    if isinstance(j, unicode):
+        return j.encode('utf-8', 'backslashreplace')
+    if isinstance(j, list):
+        return [py2JsonStrHook(item) for item in j]
+    if isinstance(j, dict):
+        return {py2JsonStrHook(key): py2JsonStrHook(value)
+            for key, value in j.iteritems()}
+    return j
+
 def readJson(filename):
     # Read 'platform.json' or 'hwsku.json' file
     try:
         with open(filename) as fp:
-            try:
-                data = json.load(fp)
-            except json.JSONDecodeError:
-                print("Json file does not exist")
-        data_dict = ast.literal_eval(json.dumps(data))
-        return data_dict
+            if sys.version_info.major == 2:
+                return json.load(fp, object_hook=py2JsonStrHook)
+            return json.load(fp)
     except Exception as e:
         print("error occurred while parsing json: {}".format(sys.exc_info()[1]))
         return None
--- a/src/sonic-config-engine/tests/sample_platform.json
+++ b/src/sonic-config-engine/tests/sample_platform.json
@@ -1,4 +1,16 @@
 {
+    "chassis": {
+        "psus": [
+            {
+                "name": "PSU 1",
+                "temperature": false
+            },
+            {
+                "name": "PSU 2",
+                "temperature": false
+            }
+        ]
+    },
     "interfaces": {
         "Ethernet0": {
             "index": "1,1,1,1",


vboykox:
@[qiluo-msft](https://github.com/qiluo-msft)

vboykox:
[samaity](https://github.com/samaity) 
[zhenggen-xu](https://github.com/zhenggen-xu)
[renukamanavalan](https://github.com/renukamanavalan)
Could you please review?

vboykox:
@prsunny

vboykox:
@akokhan

akokhan (dismissed):

vboykox:
@qiluo-msft @lguohan please review & merge

src/sonic-config-engine/portconfig.py
```diff
@@ -45,16 +46,23 @@
 #
 # Helper Functions
 #
+
+# For python2 compatibility
+def jsonStrHook(j):
```
qiluo-msft, line 51: >jsonStrHook [](http://example.com/codeflow?start=4&length=11)

Why we need this hook for python2? Can we assume platform.json and hwsku.json do not contain chars not in ascii range?

If we really need this hook for python2, can we add it only for python2? Python3 just has no hook at all.

vboykox: @qiluo-msft 
The only reason I added py2 compatibility is failing unit tests for sonic-config-engine py2 package. These are built because ENABLE_PY2_MODULES is yes for buster and jessie. Any reason to keep it as of now?

Regarding non-ascii chars Idk. It's so happens we have an implicit requirement to not have bools which we need for sure. Plus I have not found any platform.json in master having non-ascii. So currently I cannot think of a non-compromise solution. Ideally we just drop py2.

qiluo-msft: If we really need this hook for python2, can we add it only for python2? Python3 just has no hook at all.

vboykox: Alternatively in json we get from json.load we can replace bool values to some string makers and then replace the markers back with False or True before passing to ast.literal_eval. This also would keep compatibility with non-ascii. 

@qiluo-msft But do we really need py2 compatibility? What are the plans for buster?

qiluo-msft: @saiarcot895 Could you help answer the py2 plan on buster?

saiarcot895: Buster will continue to have Python 2.

However, from what I can tell, `rules/docker-config-engine-buster.mk` is using the python 3 version of sonic-config-engine, so this might not be needed on Buster containers. (I verified on a couple Buster containers that `sonic-cfggen` is calling the Python 3 interpreter.)

Stretch is using the python 2 version of sonic-config-engine. Looks like the syncd container for innovium and nephos are on Stretch.

src/sonic-config-engine/portconfig.py
```diff
@@ -45,16 +46,23 @@
 #
 # Helper Functions
 #
+
+# For python2 compatibility
```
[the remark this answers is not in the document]
vboykox: I made changes addressing your comments
1. on non-ascii chars - I kept the behavior the same as previous - instead of throwing UnicodeEcodeError, encode with hex escapes.
2. moved the compatibility code under sys.version_info.major == 2 check
3. added bool values to sample_platform.json. The changes are already covered by test_cfggen_platformJson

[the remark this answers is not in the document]
vboykox: @qiluo-msft please review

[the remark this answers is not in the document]
vboykox: @qiluo-msft please review

qiluo-msft (approved):

src/sonic-config-engine/portconfig.py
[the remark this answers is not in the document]
vboykox: @qiluo-msft @saiarcot895 thanks

PR #10242, opened 2022-03-15
[fast-reboot] Optimize PMON and LLDP services start up CPU consumption

shlomibitton:

#### Why I did it

Following this finding, some adaptations to init of PMON and LLDP ae required to off load the CPU consumption during create_switch execution.

#### How I did it

- Change the way PMON init bash script validate the existence of python2 and python3 platform packages to a more efficient way.
- On LLDP, move the sonic-cfggen call by "start.sh" script triggered by supervisord to the init bash script of this docker.

By doing this 2 adaptations PMON is not consuming a lot of CPU during init and LLDP consume the CPU before create_switch is executed and not interfering it.

#### How to verify it

Run fast-reboot on MLNX platform and observe ~2 seconds less in create_switch execution time.
Attached to this PR, bootchart outputs with the different utilizations

**With no optimizations:**

bootchart_no_optimizations

**After PMON optimizations:**

bootchart_202111_pmon_optimized_sonic_db_cli_optimized

**After PMON and LLDP optimizations:**

bootchart_202111_pmon_optimized_sonic_db_cli_optimized_lldp_optimized

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [x] 202111

### DIFF
--- a/dockers/docker-lldp/docker-lldp-init.sh
+++ b/dockers/docker-lldp/docker-lldp-init.sh
@@ -1,4 +1,7 @@
 #!/usr/bin/env bash
+
+. /usr/bin/start.sh
+
 #Generate supervisord.conf based on device metadata
 mkdir -p /etc/supervisor/conf.d/
 sonic-cfggen -d -t /usr/share/sonic/templates/supervisord.conf.j2 > /etc/supervisor/conf.d/supervisord.conf
--- a/dockers/docker-platform-monitor/docker_init.j2
+++ b/dockers/docker-platform-monitor/docker_init.j2
@@ -42,8 +42,8 @@ if [ -e /usr/share/sonic/platform/platform_wait ]; then
 fi
 
 # If the Python 2 sonic-platform package is not installed, try to install it
-python2 -c "import sonic_platform" > /dev/null 2>&1 || pip2 show sonic-platform > /dev/null 2>&1
-if [ $? -ne 0 ]; then
+PYTHON2_PLATFORM_PACKAGE=/usr/local/lib/python2.7/dist-packages/sonic_platform
+if [ ! -d "$PYTHON2_PLATFORM_PACKAGE" ]; then
     SONIC_PLATFORM_WHEEL="/usr/share/sonic/platform/sonic_platform-1.0-py2-none-any.whl"
     echo "sonic-platform package not installed, attempting to install..."
     if [ -e ${SONIC_PLATFORM_WHEEL} ]; then
@@ -59,8 +59,8 @@ if [ $? -ne 0 ]; then
 fi
 
 # If the Python 3 sonic-platform package is not installed, try to install it
-python3 -c "import sonic_platform" > /dev/null 2>&1 || pip3 show sonic-platform > /dev/null 2>&1
-if [ $? -ne 0 ]; then
+PYTHON3_PLATFORM_PACKAGE=/usr/local/lib/python3.7/dist-packages/sonic_platform
+if [ ! -d "$PYTHON3_PLATFORM_PACKAGE" ]; then
     SONIC_PLATFORM_WHEEL="/usr/share/sonic/platform/sonic_platform-1.0-py3-none-any.whl"
     echo "sonic-platform package not installed, attempting to install..."
     if [ -e ${SONIC_PLATFORM_WHEEL} ]; then
--- a/dockers/docker-lldp/supervisord.conf.j2
+++ b/dockers/docker-lldp/supervisord.conf.j2
@@ -28,17 +28,6 @@ stdout_logfile=syslog
 stderr_logfile=syslog
 dependent_startup=true
 
-[program:start]
-command=/usr/bin/start.sh
-priority=2
-autostart=false
-autorestart=false
-startsecs=0
-stdout_logfile=syslog
-stderr_logfile=syslog
-dependent_startup=true
-dependent_startup_wait_for=rsyslogd:running
-
 [program:lldpd]
 # https://github.com/vincentbernat/lldpd/commit/9856f2792c301116cc4a3fcfba91b9672ee5db1f
 # - `-d` means to stay in foreground, log to syslog


liat-grozovik:
@yxieca , @stepanblyschak  FYI, please review

liat-grozovik:
@vaibhavhd could you please help to review?

liat-grozovik:
@yxieca , @stepanblyschak kindly reminder, please review

dockers/docker-platform-monitor/docker_init.j2
```diff
@@ -42,8 +42,8 @@ if [ -e /usr/share/sonic/platform/platform_wait ]; then
 fi
 
 # If the Python 2 sonic-platform package is not installed, try to install it
-python2 -c "import sonic_platform" > /dev/null 2>&1 || pip2 show sonic-platform > /dev/null 2>&1
-if [ $? -ne 0 ]; then
+PYTHON2_PLATFORM_PACKAGE=/usr/local/lib/python2.7/dist-packages/sonic_platform
```
stepanblyschak, line 45: The package can be anywhere in PYTHONPATH. Suggest to use pkgutil.

dockers/docker-platform-monitor/docker_init.j2
```diff
@@ -59,8 +59,8 @@ if [ $? -ne 0 ]; then
 fi
 
 # If the Python 3 sonic-platform package is not installed, try to install it
-python3 -c "import sonic_platform" > /dev/null 2>&1 || pip3 show sonic-platform > /dev/null 2>&1
-if [ $? -ne 0 ]; then
+PYTHON3_PLATFORM_PACKAGE=/usr/local/lib/python3.7/dist-packages/sonic_platform
```
stepanblyschak, line 62: The python minor version might change. Suggest to use pkgutil.

dockers/docker-lldp/docker-lldp-init.sh
```diff
@@ -1,4 +1,7 @@
 #!/usr/bin/env bash
+
+. /usr/bin/start.sh
```
stepanblyschak, line 3: How to make sure this won't interffer with create_switch in the future?

shlomibitton: I can't think of a way to make sure this won't interfere in the future.
I think the best approach to improve the performance and ensure the stability of the solution is to delay these services in 90 seconds (fastboot criteria for control plane), this improve the performance even more.
The suggested services to delay is: dhcp_relay, radv, pmon and lldp.
In the future, as part of the HLD I am currently working, we will have a better way to indicate the end of fast-reboot flow and we can start the services according to it.
@liat-grozovik  @stepanblyschak @vaibhavhd @yxieca what do you think about this approach?

liat-grozovik:
@shlomibitton request change indication was removed. Still need to handle, right?

dockers/docker-lldp/docker-lldp-init.sh
shlomibitton: > @shlomibitton request change indication was removed. Still need to handle, right?

Still waiting for  @vaibhavhd and @yxieca reply for the new suggestion following Stepan comment.
To make it stable we can use the suggested approach, need to get the approval first and I will update the PR accordingly.

dockers/docker-platform-monitor/docker_init.j2
vaibhavhd: Yes, the paths can change, and pkgutil should be better tool here.

Also, will the new check of path existence validate both `sonic_platform` and `sonic-platform` availability? The previous check was doing that, and I see only `sonic_platform` in new check.

dockers/docker-platform-monitor/docker_init.j2
vaibhavhd: Same comment as for Python2.

dockers/docker-lldp/docker-lldp-init.sh
vaibhavhd: I am confused about what are we proposing. This PR is preempting lldp execution  (`LLDP consume the CPU before create_switch is executed and not interfering it.`), but we are also saying that in future PR we are going to defer the execution post fast-reboot completion.
Do we need this change right now, if we are soon going to depend upon fast-reboot end-indicator to start the deferred set of services?
I am not sure what is the impact of delaying LLDP.

vaibhavhd: Ok, I understand the new-design in future-branch part.
Still some open questions. I agree that lldp startup sequence can be adjusted to prevent overlap with create-switch. But my doubt was - should it be done before create switch (this PR), or after create-switch (deferred start in 202205)?

Also I believe that this change will affect warm-boot process too. Do you agree? If yes, then title and description should be updated.

shlomibitton: I suggest to change this PR with the after create-switch approach - deferred start.
Following Stepan comment, this approach is lack of a sync mechanism to ensure it will not overlap again in the future.
The other suggestion is to delay the entire service startup until init flow has finished with a 90 seconds timer and this is much more reliable and a bit similar to the new design of 202205.
I mean, with the new implementation these services will be delayed as well, the only difference is to start them with a different mechanism and not a timer.
I agree it will affect warm-reboot and the description will be modified as well.
If this approach is acceptable I will update all as needed.

shlomibitton: @vaibhavhd what do you think? Can I proceed and modify the PR to this suggestion?

vaibhavhd: I replied back on the email regarding this. Sorry for stalling this. The new suggested approach LGTM.
However, I am not confident about impact of delaying lldp with a 90s timer. May be @prsunny  or @yxieca can vet this from lldp point of view.

At this stage, I also would suggest to put two different PRs for pmon and lldp.

shlomibitton: thanks @vaibhavhd, I will create a PR and test it.

shlomibitton: @vaibhavhd @liat-grozovik @stepanblyschak @yxieca I created 2 new PR's with the suggested approach, can you please review? Thanks

PMON: https://github.com/Azure/sonic-buildimage/pull/10567
LLDP: https://github.com/Azure/sonic-buildimage/pull/10568

shlomibitton:
@vaibhavhd @liat-grozovik @stepanblyschak @yxieca I created 2 new PR's with the suggested approach, can you please review? Thanks

PMON: https://github.com/Azure/sonic-buildimage/pull/10567
LLDP: https://github.com/Azure/sonic-buildimage/pull/10568

Closing this PR.

PR #10274, opened 2022-03-18
[Build]: Use one debian mirror config

xumia:
#### Why I did it
Use one debian mirror config.
The empty config in https://github.com/Azure/sonic-buildimage/blob/master/files/image_config/apt/sources.list overrides the file https://github.com/Azure/sonic-buildimage/blob/master/files/apt/sources.list.amd64 (armhf/arm64), it does not make sense.
All the content in files/image_config/apt is no use, any one wants to add mirror config, please add in files/apt.

#### How I did it
Remove files/image_config/apt and the reference.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012
- [x] 202106
- [x] 202111


xumia:
@antony-rheneus , could you please help review it? It does not break armhf build, can we simplify it? See the checks in https://github.com/Azure/sonic-buildimage/pull/10281
https://dev.azure.com/mssonic/be1b070f-be15-4154-aade-b1d3bfb17054/_build/results?buildId=82352

xumia:
@saiarcot895, could you please help review it?

files/apt/sources.list.amd64
```diff
@@ -3,13 +3,13 @@
 
 deb [arch=amd64] http://deb.debian.org/debian/ bullseye main contrib non-free
```
saiarcot895, line 4: Since we have `debian-archive.trafficmanager.net` providing the bullseye packages, I don't think we need `deb.debian.org` as well, right?

xumia: @saiarcot895 , thanks for your comment, I think it should be good to remove deb.debian.org, updated.

saiarcot895: Also, what's the difference between debian-archive.trafficmanager.net and packages.trafficmanager.net? Are both needed/recommended?

xumia: @saiarcot895 , not much different, but packages.trafficmanager.net is for reproducible build, which contains version frozen packages, debian-archive.trafficmanager.net may not contain them.

saiarcot895 (approved):

PR #10285, opened 2022-03-18
Don't use python/python2 for restapi start scripts

saiarcot895:
Python 2 isn't installed by default in Buster and Bullseye containers,
and the scripts/modules can be used with Python 3, so make sure Python 3
is used.

#### Why I did it

After the Buster and Bullseye upgrade for the restapi container, processes will no longer start because supervisord is trying to call `python` and `python2`, both of which are unavailable.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012
- [ ] 202106
- [ ] 202111


lguohan:
test failure to due to this issue. https://github.com/Azure/sonic-mgmt/pull/5374

PR #10230, opened 2022-03-14
Update docker-sonic-vs infrastructure in order to run CoPP UT

noaOrMlnx:
#### Why I did it
In order to be able to run CoPP UT in docker-sonic-vs
#### How I did it
N/A
#### How to verify it
Run Copp UT
#### Which release branch to backport (provide reason below if selected)

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106

PR #10290, opened 2022-03-20
[202012][CoPP] Fix the issue that trap entry can not be installed for DHCP and DHCPv6

stephenxs:
Fix the issue that TRAP entry can not be installed via aligning the entry names of DHCP in copp_cfg.j2 with the feature name
There is a logic in CoPP manager which loads CoPP items only if the feature is enabled.
It requires the feature name to align with CoPP item name otherwise the CoPP item can not be loaded.
Currently, DHCP CoPP item is used by feature dhcp_relay but the name doesn't align, which prevents DHCP CoPP entries from being installed.
This is to fix it.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111


dgsudharsan:
@stephenxs  Can we close this PR? We already have one for 202012 raised here https://github.com/Azure/sonic-buildimage/pull/10230

Issue #9801, opened 2022-01-19
[yang] DPB fails when Dynamic Buffer is configured


alexrallen:
When we configure the following buffer profiles....

````
    "BUFFER_PORT_EGRESS_PROFILE_LIST": {
        "Ethernet0": {
            "profile_list": "egress_lossless_profile,egress_lossy_profile"
        }
    }
```

Dynamic port breakout fails with...

```
admin@r-lionfish-14:~$ sudo config interface breakout Ethernet232 4x25G[10G,1G] -y 

Running Breakout Mode : 1x100G[50G,40G,25G,10G,1G] 
Target Breakout Mode : 4x25G[10G,1G]
Ports to be deleted : 
 {
    "Ethernet232": "100000"
}
Ports to be added : 
 {
    "Ethernet232": "25000",
    "Ethernet233": "25000",
    "Ethernet234": "25000",
    "Ethernet235": "25000"
}

After running Logic to limit the impact
Final list of ports to be deleted : 
 {
    "Ethernet232": "100000"
} 

Final list of ports to be added :  
 {
    "Ethernet232": "25000",
    "Ethernet233": "25000",
    "Ethernet234": "25000",
    "Ethernet235": "25000"
}

Note: Below table(s) have no YANG models:

DEFAULT_LOSSLESS_BUFFER_PARAMETER, KDUMP, LOSSLESS_TRAFFIC_PATTERN, SNMP, SNMP_COMMUNITY, WJH, WJH_CHANNEL, XCVRD_LOG, 

libyang[0]: Duplicated instance of "profile_list" leaf-list ("e"). (path: /sonic-buffer-port-egress-profile-list:sonic-buffer-port-egress-profile-list/BUFFER_PORT_EGRESS_PROFILE_LIST/BUFFER_PORT_EGRESS_PROFILE_LIST_LIST[port='Ethernet0']/profile_list[.='e'])

sonic_yang(3):Data Loading Failed:Duplicated instance of "profile_list" leaf-list ("e").
Data Loading Failed
Duplicated instance of "profile_list" leaf-list ("e").

ConfigMgmt Class creation failed

Failed to break out Port. Error: Failed to load the config. Error: ConfigMgmtDPB Class creation failed
```


This is likely due to the fact that the Yang model is configured with the following expectation for profile_list (a JSON list) which does not match what is the appropriate configuration (a comma separated string).

```
[...]

PR #10291, opened 2022-03-21
[YANG] Fix issue: Non compliant leaf list in config_db schema

Junchao-Mellanox:
#### Why I did it

Fix issue: Non compliant leaf list in config_db schema: https://github.com/Azure/sonic-buildimage/issues/9801

#### How I did it

The basic flow of DPB is like:
1.	Transfer config db json value to YANG json value, name it “yangIn”
2.	Validate “yangIn” by libyang
3.	Generate a YANG json value to represent the target configuration, name it “yangTarget”
4.	Do diff between “yangIn” and “yangTarget”
5.	Apply the diff to CONFIG DB json and save it back to DB
 
The fix:
•	For step #1, If value of a leaf-list field string type, transfer it to a list by splitting it with “,” the purpose here is to make step#2 happy. We also need to save <table_name>.<key>.<field_name> to a set named “leaf_list_with_string_value_set”.
•	For step#5, loop “leaf_list_with_string_value_set” and change those fields back to a string.

#### How to verify it

1.Manual test
2. Changed sample config DB and unit test passed

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [x] 202111


dgsudharsan (approved):

Junchao-Mellanox:
Hi @qiluo-msft  and @praveen-li , could you please review?

zhangyanzhao:
@praveen-li would you please help to review this PR?

liat-grozovik:
Hi @qiluo-msft and @praveen-li kindly reminder to review. As this is also needed for 202111 we need to urge you on that. sorry.

zhangyanzhao:
@praveen-li would you please help to review and approve this PR? Thanks.

liat-grozovik:
@praveen-li kindly reminder. this fix is must for 202111

praveen-li:
Hi Liat

I will surely review it today,

Thanks,

Regards
Praveen

src/sonic-yang-mgmt/sonic_yang_ext.py
```diff
@@ -566,6 +574,8 @@ def _xlateList(self, model, yang, config, table, exceptionList):
                 exceptionList.append(str(e))
                 # with multilist, we continue matching other keys.
                 continue
+            finally:
+                self.elementPath.pop()
```
praveen-li, line 578: okie, seems you want to record it in a set, if it is a string leaf-list

src/sonic-yang-mgmt/sonic_yang_ext.py
```diff
@@ -734,9 +750,12 @@ def _revYangConvert(val):
 
         # if it is a leaf-list do it for each element
         if leafDict[key]['__isleafList']:
-            vValue = list()
-            for v in value:
-                vValue.append(_revYangConvert(v))
+            if tuple(self.elementPath) in self.leaf_list_with_string_value_set:
```
praveen-li, line 753: if table1 and entry 1 is in set. And table2 and entry2 is in set too. Then (table1, entry2) may create false true, right ? i.e. same leaf name in 2 tables.

praveen-li, line 753: We use this lib for reverse translation as well, i.e. from yang to configdb, in that case self.leaf_list_with_string_value_set will be empty right. [i.e. only reverse translation cases.]

src/sonic-yang-mgmt/sonic_yang_ext.py
```diff
@@ -874,8 +897,10 @@ def _revXlateContainerInContainer(self, model, yang, config, table):
         if yang.get(modelContainer['@name']):
             config[modelContainer['@name']] = dict()
             self.sysLog(msg="revXlateContainerInContainer {}".format(modelContainer['@name']))
+            self.elementPath.append(modelContainer['@name'])
             self._revXlateContainer(modelContainer, yang[modelContainer['@name']], \
                 config[modelContainer['@name']], table)
+            self.elementPath.pop()
```
praveen-li, line 903: We need to change these fields in backend eventually. IMHO we should list xpath of these fields in a list\map. And do the special processing when xpath matches. Let me know you view?

praveen-li (changes requested):

src/sonic-yang-mgmt/sonic_yang_ext.py
Junchao-Mellanox: No. An element in self.leaf_list_with_string_value_set would be like ['PORT', 'Ethernet0', 'speed'], another table with the same leaf name could be ['INTERFACE', 'loopback0', 'speed'], it will not generate false true.

Junchao-Mellanox: What could be the case of that [only reverse translation cases]?

src/sonic-yang-mgmt/sonic_yang_ext.py
Junchao-Mellanox: I am not sure that how xpath could work in such case. IMO, if the current solution is a valid fix, I would prefer not to introduce another solution as this is just a workaround which will be replaced by a final solution(change these fields in backend) in future.

For the xpath you mentioned, could you elaborate how it will work in following functions:
 ```python

    def _findYangTypedValue(self, key, value, leafDict):

        # convert config DB string to yang Type
        def _yangConvert(val):
            # Convert everything to string
            val = str(val)
            # find type of this key from yang leaf
            type = leafDict[key]['type']['@name']

            if 'uint' in type:
                vValue = int(val, 10)
            # TODO: find type of leafref from schema node
            elif 'leafref' in type:
                vValue = val
            #TODO: find type in sonic-head, as of now, all are enumeration
            elif 'stypes:' in type:
                vValue = val
            else:
                vValue = val
            return vValue

        # if it is a leaf-list do it for each element
        if leafDict[key]['__isleafList']:
            vValue = list()
            for v in value:
                vValue.append(_yangConvert(v))
        else:
            vValue = _yangConvert(value)

        return vValue

    def _revFindYangTypedValue(self, key, value, leafDict):

        # convert yang Type to config DB string
        def _revYangConvert(val):
            # config DB has only strings, thank god for that :), wait not yet!!!
            return str(val)

        # if it is a leaf-list do it for each element
        if leafDict[key]['__isleafList']:
            vValue = list()
            for v in value:
                vValue.append(_revYangConvert(v))
        elif leafDict[key]['type']['@name'] == 'boolean':
            vValue = 'true' if value else 'false'
        else:
            vValue = _revYangConvert(value)

        return vValue
 ```

Junchao-Mellanox: Had a call with Praveen, will use a hardcodes list to define those fields which are defined as leaf-list but have string value.

src/sonic-yang-mgmt/sonic_yang_ext.py
Junchao-Mellanox: Praveen provided a case that "config reload from YANG". So this is a valid concern. I will fix.

wen587:
Hi I saw testcase fail on GCU generic_config_updater/test_bgp_prefix.py
It seems the changes doesn't pass yang validation with below configuration.

```
admin@vlab-01:~/hijack/test$ cat add1.load
{
    "BGP_ALLOWED_PREFIXES": {
        "DEPLOYMENT_ID|0": {
            "prefixes_v4": [
                "10.20.0.0/16"
            ],
            "prefixes_v6": [
                "fc02:20::/64"
            ]
        }
    }
}

admin@vlab-01:~/hijack/test$ sudo config load -y add1.load
Running command: /usr/local/bin/sonic-cfggen -j add1.load --write-to-db

admin@vlab-01:~/hijack/test$ cd ~
admin@vlab-01:~$ cat verify_configDB_yang.py
import subprocess
import sonic_yang
import json
#import pytest
def get_config_db_json():
    cmd = "show runningconfiguration all"
    result = subprocess.Popen(cmd, shell=True, text=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
    text, err = result.communicate()
    return json.loads(text)
configDbJson = get_config_db_json()
sy = sonic_yang.SonicYang("/usr/local/yang-models")
sy.loadYangModel()
sy.loadData(configDbJson )
sy.validate_data_tree()

admin@vlab-01:~$ python verify_configDB_yang.py
sonic_yang(6):Note: Below table(s) have no YANG models: CONSOLE_SWITCH, DEVICE_NEIGHBOR_METADATA, DHCP_SERVER, KDUMP, RESTAPI, SNMP, SNMP_COMMUNITY, TELEMETRY
sonic_yang(3):All Keys are not parsed in BGP_ALLOWED_PREFIXES
dict_keys(['DEPLOYMENT_ID|0'])
sonic_yang(3):exceptionList:['Value not found for deployment id community in DEPLOYMENT_ID|0', "'list' object has no attribute 'split'"]
sonic_yang(3):Data Loading Failed:All Keys are not parsed in BGP_ALLOWED_PREFIXES
dict_keys(['DEPLOYMENT_ID|0'])
  File "/usr/local/lib/python3.9/dist-packages/sonic_yang_ext.py", line 1107, in loadData
    self._xlateConfigDB(xlateFile=xlateFile)
  File "/usr/local/lib/python3.9/dist-packages/sonic_yang_ext.py", line 727, in _xlateConfigDB
    self._xlateConfigDBtoYang(jIn, yangJ)
  File "/usr/local/lib/python3.9/dist-packages/sonic_yang_ext.py", line 713, in _xlateConfigDBtoYang
    self._xlateContainer(cmap['container'], yangJ[key][subkey], \
  File "/usr/local/lib/python3.9/dist-packages/sonic_yang_ext.py", line 692, in _xlateContainer
    raise(Exception("All Keys are not parsed in {}\n{}".format(table, \
Exception: All Keys are not parsed in BGP_ALLOWED_PREFIXES
dict_keys(['DEPLOYMENT_ID|0'])

During handling of the above exception, another exception occurred:

  File "/home/admin/verify_configDB_yang.py", line 13, in <module>
    sy.loadData(configDbJson )
  File "/usr/local/lib/python3.9/dist-packages/sonic_yang_ext.py", line 1117, in loadData
    raise SonicYangException("Data Loading Failed\n{}".format(str(e)))
sonic_yang_ext.SonicYangException: Data Loading Failed
All Keys are not parsed in BGP_ALLOWED_PREFIXES
dict_keys(['DEPLOYMENT_ID|0'])
```

liat-grozovik:
@praveen-li could you please help to review following comments provided?

liat-grozovik:
@praveen-li kindly reminder. this must get in 2021111 so appreciate if we can close the loop. thanks.

zhangyanzhao:
Once Praveen's comment is addressed, he will approve and merge.

Junchao-Mellanox:
Hi @praveen-li , I have fixed the comments, could you please review and sign-off?

praveen-li (approved):
LGTM, thanks for fixing it

dgsudharsan (approved):

zhangyanzhao:
Qi will merge

qiluo-msft (approved):

Junchao-Mellanox:
no clean cherry-pick, i will create separate PR for 202111.

PR #10324, opened 2022-03-22
[201911] Backport FRR patch tp FRR-7.2 to handle pthread race in peer notify message handling

abdosi:
What I did:

Backport FRR patch https://github.com/FRRouting/frr/pull/8220 on FRR 7.2. Fixes the Issue https://github.com/FRRouting/frr/issues/8213

Why I did:-

Because of this race-condition we saw GR getting triggered even though BGP shut is given on peer device. 

How I verify:

After patching this fix GR is not triggered on doing BGP shut on peer.


lguohan (approved):

prsunny (approved):

lguohan:
better to change your pr title to be more meaningful

PR #10325, opened 2022-03-22
Check to see that the py2 and py3 version files exist before trying to sort them

saiarcot895:
For Bullseye, Python 2 isn't present at all. This means that in certain
build cases (such as building something only for Bullseye), the version
file may not exist, and so the sort command would fail.

For most normal build commands, this probably won't be an issue, because
the SONiC build will start with Buster (which has both Python 2 and
Python 3 wheels built), and so the py2 and py3 files will be present
even during the Bullseye builds.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111


xumia (approved):

xumia:
@saiarcot895 , thanks for fixing the issue.

PR #10352, opened 2022-03-25
Parallel building of sonic dockers using native dockerd(dood).

Kalimuthu-Velappan:
Currently, the build dockers are created as user dockers(docker-base-stretch-<user>, etc) that are
specific to each user. But the sonic dockers (docker-database, docker-swss, etc) are
created with a fixed docker name and are common to all the users.

    docker-database:latest
    docker-swss:latest

When multiple builds are triggered on the same build server that creates parallel building issues because
all the build jobs are trying to create the same docker with the latest tag.
This happens only when sonic dockers are built using native host dockerd for sonic docker image creation.

This patch creates all sonic dockers as user sonic dockers and then, while
saving and loading the user sonic dockers, it renames the user sonic
dockers into correct sonic dockers with the tag as latest.

            docker-database:latest <== SAVE/LOAD ==> docker-database-<user>:tag

The user sonic docker names are derived from 'DOCKER_USERNAME and DOCKER_USERTAG' make env variable and using Jinja template, it replaces the FROM docker name with correct user sonic docker name for loading and saving the docker image.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

### DIFF
--- a/dockers/docker-config-engine-buster/Dockerfile.j2
+++ b/dockers/docker-config-engine-buster/Dockerfile.j2
@@ -1,5 +1,5 @@
 {% from "dockers/dockerfile-macros.j2" import install_debian_packages, install_python_wheels, copy_files %}
-FROM docker-base-buster
+FROM docker-base-buster-{{DOCKER_USERNAME}}:{{DOCKER_USERTAG}}
 
 ## Make apt-get non-interactive
 ENV DEBIAN_FRONTEND=noninteractive
--- a/Makefile
+++ b/Makefile
@@ -40,6 +40,7 @@ endif
 ifeq ($(NOBULLSEYE), 0)
 	BLDENV=bullseye make -f Makefile.work $@
 endif
+	BLDENV=bullseye make -f Makefile.work docker-cleanup
 
 jessie:
 	@echo "+++ Making $@ +++"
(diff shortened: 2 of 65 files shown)


dockers/docker-config-engine-buster/Dockerfile.j2
```diff
@@ -1,5 +1,5 @@
 {% from "dockers/dockerfile-macros.j2" import install_debian_packages, install_python_wheels, copy_files %}
-FROM docker-base-buster
+FROM {{ docker_config_engine_buster_load_image }}
```
xumia, line 2: Is it possible to define a variable contains the info that we can easily know which image it bases on?
For instance, {{ docker_base_buster_image }}
All the images can be exposed as variables before building dockers.
docker_base_buster_image=docker-base-buster-$(user)

Kalimuthu-Velappan: Done

xumia (approved):

xumia:
@liushilongbuaa , could you please help review it? Is it any impact on the reproducible build of docker image?

liushilongbuaa:
> @liushilongbuaa , could you please help review it? Is it any impact on the reproducible build of docker image?

It looks good for me

liushilongbuaa (approved):

PR #10358, opened 2022-03-27
[Build]: Fix host image debian package version issue

xumia:
#### Why I did it
Fix host image debian package version issue.
The package dependencies may have issue, when some of debian packages of the base image are upgraded. For example, libc is installed in base image, but if the mirror has new version, when running "apt-get upgrade", the package will be upgraded unexpected. To avoid such issue, need to add the versions when building the host image.

#### How I did it
The package versions of host-image should contain host-base-image.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012
- [x] 202106
- [x] 202111


liushilongbuaa (approved):

xumia:
**By pass the test stage for this PR, it blocks the 202012 branch official build, and there are several submodule builds that have the dependencies.**
Justification: it only changes the build options.

liushilongbuaa (approved):

PR #10369, opened 2022-03-28
[Security]: Enable hardening build options

xumia:
#### Why I did it
See https://wiki.debian.org/Hardening

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [x] 202111


slave.mk
```diff
@@ -268,6 +268,11 @@ export SONIC_ROUTING_STACK
 export FRR_USER_UID
 export FRR_USER_GID
 
+###############################################################################
+## Build Options
+###############################################################################
+export DEB_BUILD_MAINT_OPTIONS = hardening=+all
```
saiarcot895, line 274: The manpage suggests that `DEB_BUILD_MAINT_OPTIONS` should be used only within `debian/rules`, and `DEB_BUILD_OPTIONS` should be used outside `debian/rules`, so that hardening options can be easily overridden in `debian/rules` if necessary. Given this, should this be changed to use `DEB_BUILD_OPTIONS` instead?

xumia: @saiarcot895 , thanks for your comment. Changed to use DEB_BUILD_OPTIONS instead.

saiarcot895 (approved):

saiarcot895:
Looks like this setting won't have much of an impact, because `DEB_BUILD_OPTIONS` gets overridden for both [SONIC_MAKE_DEBS](https://github.com/Azure/sonic-buildimage/blob/cd330f0e70e646c2ff6fc88c6c8fb9f75e8b40e7/slave.mk#L508) and [SONIC_DPKG_DEBS](https://github.com/Azure/sonic-buildimage/blob/cd330f0e70e646c2ff6fc88c6c8fb9f75e8b40e7/slave.mk#L551), so this setting might end up not doing anything to add hardening build options.

PR #10421, opened 2022-03-31
Removed python2 dependency for sonic-pcied in sonic-platform-daemons

assrinivasan:
Removed python2 dependency for sonic-pcied in sonic-platform-daemons that was causing unit test errors in sonic_pcied.

#### Why I did it
The current unit test for the PCIE daemon contains test code that is only compatible with Python3. This was causing build issues on account of the python3 wheel being dependent on the Python2 wheel.

#### How I did it
Removed all rules relating to building the Python2 wheel and the PCIE daemon's Python 3 wheel's dependencies on it.

#### How to verify it
Clone and build sonic-buildimage. This should, among other things, build target/python-wheels/buster/sonic_pcied-1.0-**py3**-none-any.whl without building target/python-wheels/buster/sonic_pcied-1.0-**py2**-none-any.whl

This is my dog Disco. She's pretty cute.

IMG_2177


vdahiya12:
Please also remove the corresponding configuration from Docker supervisord jinja templates.

dockers/docker-platform-monitor/docker-pmon.supervisord.conf.j2
```diff
@@ -171,7 +171,7 @@ dependent_startup_wait_for=rsyslogd:running
 
 {% if not skip_pcied %}
 [program:pcied]
-command={% if API_VERSION == 3 and 'pcied' not in python2_daemons %}python3 {% else %} python2 {% endif %}/usr/local/bin/pcied
+command= python3 /usr/local/bin/pcied
```
qiluo-msft, line 174: >python3 [](http://example.com/codeflow?start=9&length=7)

You can just remove `python3` because there is correct shebang line inside the script.

And remove the blank before after `=`.

sujinmkang (approved):

assrinivasan:
Can I merge this?

qiluo-msft (approved):

Issue #9746, opened 2022-01-12
[yang] validating TACPLUS with empty “global” fails, while non-empty “global” does not fail


wen587:
#### Description
yang validation fail when `global` is empty, but it succeeds when nonempty.

_Background of tacacs table._ 
In t0 topo, tacacs has default value while configDB doesn't contain those table.
After sudo config tacacs value, its table will appear in configDB.
Then the table will contains empty value after set back to default.

#### Steps to reproduce the issue:
1. config tacacs and let TACPLUS table appear in configDB
```
admin@vlab-01:~$ show run all | grep -w TACPLUS
admin@vlab-01:~$ sudo config tacacs timeout 3
admin@vlab-01:~$ show run all | grep -w TACPLUS -A3
    "TACPLUS": {
        "global": {
            "timeout": "3"
        }

```
2. Verify by python script. Pass. (Some yang is outdated so we ignore them)
```
admin@vlab-01:~$ cd /usr/local/yang-models/
admin@vlab-01:/usr/local/yang-models$ sudo mkdir ignore
admin@vlab-01:/usr/local/yang-models$ sudo mv sonic-queue.yang sonic-scheduler.yang sonic-feature.yang ignore/
admin@vlab-01:/usr/local/yang-models$ cd -
/home/admin

admin@vlab-01:~$ cat verify_configDB_yang.py
import subprocess
import sonic_yang
import json
def get_config_db_json():
    cmd = "show runningconfiguration all"
    result = subprocess.Popen(cmd, shell=True, text=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
    text, err = result.communicate()
    return json.loads(text)
configDbJson = get_config_db_json()
sy = sonic_yang.SonicYang("/usr/local/yang-models")
sy.loadYangModel()
sy.loadData(configDbJson )
sy.validate_data_tree()
admin@vlab-01:~$

admin@vlab-01:~$ python verify_configDB_yang.py
Note: Below table(s) have no YANG models:
BGP_PEER_RANGE, CONSOLE_SWITCH, DEVICE_NEIGHBOR_METADATA, DHCP_SERVER, FEATURE, KDUMP, QUEUE, RESTAPI, SCHEDULER, SNMP, SNMP_COMMUNITY, TELEMETRY,
admin@vlab-01:~$
```
3. set tacacs back to default and run python script again. It fails.
```
admin@vlab-01:~$ sudo config tacacs default timeout
[...]

PR #10431, opened 2022-04-01
Fix yang validation failure when table contains empty value

wen587:
#### Why I did it
Fix https://github.com/Azure/sonic-buildimage/issues/9746
#### How I did it
Split the check condition based on non-exist and zero length.
#### How to verify it
Run verification script when table contains empty value

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

### DIFF
--- a/src/sonic-yang-mgmt/sonic_yang_ext.py
+++ b/src/sonic-yang-mgmt/sonic_yang_ext.py
@@ -591,22 +591,22 @@ class SonicYangExtMixin:
     """
     def _xlateContainerInContainer(self, model, yang, configC, table):
         ccontainer = model
-        #print(ccontainer['@name'])
-        yang[ccontainer['@name']] = dict()
-        if configC.get(ccontainer['@name']) is None:
+        ccName = ccontainer['@name']
+        yang[ccName] = dict()
+        if ccName not in configC:
             return
-        if len(ccontainer['@name']) == 0:
+        if len(ccName) == 0:
             # Empty container, clean config and return
-            del configC[ccontainer['@name']]
+            del configC[ccName]
             return
-        self.sysLog(msg="xlateProcessListOfContainer: {}".format(ccontainer['@name']))
-        self._xlateContainer(ccontainer, yang[ccontainer['@name']], \
-        configC[ccontainer['@name']], table)
+        self.sysLog(msg="xlateProcessListOfContainer: {}".format(ccName))
+        self._xlateContainer(ccontainer, yang[ccName], \
+        configC[ccName], table)
         # clean empty container
-        if len(yang[ccontainer['@name']]) == 0:
-            del yang[ccontainer['@name']]
+        if len(yang[ccName]) == 0:
+            del yang[ccName]
         # remove copy after processing
-        del configC[ccontainer['@name']]
+        del configC[ccName]
 
         return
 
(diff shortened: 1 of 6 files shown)


src/sonic-yang-mgmt/sonic_yang_ext.py
```diff
@@ -593,7 +593,11 @@ def _xlateContainerInContainer(self, model, yang, configC, table):
         ccontainer = model
         #print(ccontainer['@name'])
         yang[ccontainer['@name']] = dict()
-        if not configC.get(ccontainer['@name']):
+        if configC.get(ccontainer['@name']) is None:
```
qiluo-msft, line 596: >configC.get(ccontainer['@name']) is None [](http://example.com/codeflow?start=11&length=40)

How about
```
if ccontainer['@name'] not in configC:
```

BTW, you can extract `ccontainer['@name']` as a variable, no need to calculate many times.

And remove `#print` above.

qiluo-msft (commented):
LGTM

src/sonic-yang-mgmt/sonic_yang_ext.py
```diff
@@ -591,18 +591,22 @@ def _xlateListInContainer(self, model, yang, configC, table, exceptionList):
...
-        yang[ccontainer['@name']] = dict()
-        if not configC.get(ccontainer['@name']):
+        ccName = ccontainer['@name']
+        yang[ccName] = dict()
+        if ccName not in configC:
             return
-        self.sysLog(msg="xlateProcessListOfContainer: {}".format(ccontainer['@name']))
-        self._xlateContainer(ccontainer, yang[ccontainer['@name']], \
-        configC[ccontainer['@name']], table)
+        if len(ccName) == 0:
```
praveen-li, line 598: if len(configC[ccName]) ==0: 

i.e if value is zero, delete the key. Here, delete the key means processing is done,

praveen-li (commented):
One Dev Changes is needed.

Will it be possible for you to add a test case for this please, may be in same reference under which you found the issue. Or simply by adding a config in https://github.com/Azure/sonic-buildimage/blob/master/src/sonic-yang-models/tests/files/sample_config_db.json

src/sonic-yang-mgmt/sonic_yang_ext.py
wen587: Hi, I saw TACPLUS table already has configuration in sample_config_db.json.
I added a UT for field `global` with empty value.
From where I saw the issue, it failed at GCU test after other test configured TACPLUS table.
I also added a empty json apply GCU test https://github.com/Azure/sonic-mgmt/pull/5459  for simply checking the configDB.

src/sonic-yang-models/tests/files/sample_config_db.json
```diff
@@ -304,7 +304,7 @@
                 "switch_id": "2",
                 "switch_type": "voq",
                 "max_cores": "8",
-                "sub_role": "FrondEnd",
```
praveen-li, line 307: [10:18 PM] Praveen Chaudhary
in this file: src/sonic-yang-models/tests/files/sample_config_db.json

[10:18 PM] Praveen Chaudhary
we need global {} config for test to exercise this code 

[10:18 PM] Praveen Chaudhary
plz add that and we are good

praveen-li (approved):
LGTM

qiluo-msft (approved):

PR #10432, opened 2022-04-01
[procdockerstatsd] Use psutil library to update process stats instead of ps command

MuLinForest:
#### Why I did it
The value of /oc-sys:system/oc-sys:processes in RESTCONF is retrieved from stateDB. And it is written by `procdockerstatsd`.
However, some statistics is missing, like start-time, uptime, etc.
In order to get these statistics, it is easier to use `psutil` to obtain process information.

#### How I did it
Use `psutil` library to obtain the process information

Signed-off-by: MuLin <mulin_huang@edge-core.com>


gord1306:
@jleveque could you help to review this PR, thanks.

PR #10445, opened 2022-04-02
[Ci]: check if there is a sonic dirty version issue

xumia:
#### Why I did it
[Ci]: check if there is a sonic dirty version issue
If there is a dirty version issue in PR build, the build will be failed.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012
- [x] 202106
- [x] 202111

### DIFF
--- a/.azure-pipelines/azure-pipelines-build.yml
+++ b/.azure-pipelines/azure-pipelines-build.yml
@@ -131,3 +131,4 @@ jobs:
             make $BUILD_OPTIONS target/sonic-$(GROUP_NAME).bin
           fi
         displayName: "Build sonic image"
+      - template: check-dirty-version.yml
--- a/.azure-pipelines/check-dirty-version.yml
+++ b/.azure-pipelines/check-dirty-version.yml
@@ -3,7 +3,7 @@ steps:
     . functions.sh
     SONIC_VERSION=$(sonic_get_version)
     echo "SONIC_VERSION=$SONIC_VERSION"
-    if [ "$SONIC_VERSION" == *dirty* ]; then
+    if [[ "$SONIC_VERSION" == *dirty* ]]; then
       # Print the detail dirty info
       git status --untracked-files=no -s --ignore-submodules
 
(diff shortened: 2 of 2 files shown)


.azure-pipelines/check-dirty-version.yml
```diff
@@ -0,0 +1,16 @@
+steps:
+- script: |
+    . functions.sh
+    SONIC_VERSION=$(sonic_get_version)
+    echo "SONIC_VERSION=$SONIC_VERSION"
+    if [ "$SONIC_VERSION" == *dirty* ]; then
+      # Print the detail dirty info
+      git status --untracked-files=no -s --ignore-submodules
+
+      # Exit with error, if it is a PR build
```
xumia, line 10: We can block official build if the issue has fixed in all release branches.

liushilongbuaa: It seems that this step will change in future.
Can we use a template in master branch?

liushilongbuaa (dismissed):

.azure-pipelines/check-dirty-version.yml
xumia: The check-dirty-version.yml is a template file.

xumia:
@liushilongbuaa , please help review again, fix a bash string any string comparing issue, and verify it works as expected. See the check in https://github.com/Azure/sonic-buildimage/pull/10495

.azure-pipelines/azure-pipelines-build.yml
```diff
@@ -131,3 +131,4 @@ jobs:
             make $BUILD_OPTIONS target/sonic-$(GROUP_NAME).bin
           fi
         displayName: "Build sonic image"
+      - template: check-dirty-version.yml
```
qiluo-msft, line 134: >template: check-dirty-version.yml [](http://example.com/codeflow?start=8&length=33)

Is it too late. If we detect dirty, we don't need to build at all.

xumia: @qiluo-msft , the dirty is detected after build, during the build, it may not have the dirty issue. In most case 99%, no one changes the source code, so I think it is not late to check it after build.

qiluo-msft (approved):

PR #10470, opened 2022-04-05
[asan] add address sanitizer support to docker-sonic-vs

Yakiv-Huryk:
Add address sanitizer support to docker-sonic-vs

#### Why I did it
To support docker-sonic-vs image with ASAN.
#### How I did it
* made the supervisord.conf a template
* added the 'log_path' environment variable for ASAN-enabled daemons
* added supervisord.conf.j2 generation and ASAN lib to the docker-sonic-vs/Dockerfile.j2
#### How to verify it
* made a build with ENABLE_ASAN=y
* run the tests, checked ASAN reports

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111


liat-grozovik:
@prsunny could you please help to review or suggest reviewer?

saiarcot895:
It may be useful to also have it dump a core file (adding `disable_coredump=0:unmap_shadow_on_exit=1:abort_on_error=1`), so that it'd be easier to debug the cause of the issue after it exists.

liat-grozovik:
> > It may be useful to also have it dump a core file (adding `disable_coredump=0:unmap_shadow_on_exit=1:abort_on_error=1`), so that it'd be easier to debug the cause of the issue after it exists.
> 
> Currently, it's out of scope for this stage. The log with stack trace seems enough for the bugs that were found. We might want to consider adding it in the future if we see that it's useful.

i tend to agree. We have another feature of auto techsupport which is now extended to memory leaks and later can also be extended to asan. for now lets focus on enabling this functionality and later on enhance it

platform/vs/docker-sonic-vs/supervisord.conf.j2
```diff
@@ -42,6 +42,9 @@ autostart=false
 autorestart=false
 stdout_logfile=syslog
 stderr_logfile=syslog
+{% if ENABLE_ASAN == "y" %}
+environment=ASAN_OPTIONS="log_path=/var/log/asan/syncd-asan.log"
```
saiarcot895, line 46: Does `/var/log/asan/` need to be created first? When testing ASAN for another application, with this `log_path`, ASAN didn't write the log file because `/var/log/asan` didn't exist.

Yakiv-Huryk: For the vs, it is not convenient to have the logs inside of a container. The idea is to map this dir to the host (e.g. -v /asan_reports:/var/log/asan). It makes the automation/analysis of the reports much easier.

saiarcot895 (approved):

PR #10471, opened 2022-04-05
Added support for front panel port prefix regex

itamar-talmon:
- What I did
Removed the dependency on the "Ethernet" string in the SONiC code base and added support
for extending the front panel port name pattern.

- How I did it
1. Introduced FRONT_PANEL_PORT_PREFIX_REGEX that extends the old FRONT_PANEL_PORT_PREFIX ("Ethernet")
2. Updated all the relevant usage of the "Ethernet" throughout the code base to use the new regex pattern

- How to verify it
Pass all UT and CI testing.

- Why I did it
In order to support distinguishing between different types of front panel ports in a maintainable fashion.
Specifically, we are planning to bring up a system with 'service' ports (in addition to the regular ethernet data ports) - these
are lower speed ports that used for connection to accelerators, internal loopbacks and more.

**- Related Commits and Merge Strategy**
_This is part of a group of related commits and should be merged **after https://github.com/Azure/sonic-swss-common/pull/598**._

The full merge order is:

1. swss-common - https://github.com/Azure/sonic-swss-common/pull/598
2. sonic-buildimage - https://github.com/Azure/sonic-buildimage/pull/10471
3. swsssdk - https://github.com/Azure/sonic-py-swsssdk/pull/121
4. all the rest
     [ https://github.com/Azure/sonic-utilities/pull/2127](https://github.com/Azure/sonic-utilities/pull/2127)
     [ https://github.com/Azure/sonic-snmpagent/pull/251](https://github.com/Azure/sonic-snmpagent/pull/251)
     [ https://github.com/Azure/sonic-swss/pull/2223](https://github.com/Azure/sonic-swss/pull/2223)
     [ https://github.com/Azure/sonic-platform-daemons/pull/252](https://github.com/Azure/sonic-platform-daemons/pull/252)
     [ https://github.com/Azure/sonic-platform-common/pull/274](https://github.com/Azure/sonic-platform-common/pull/274)

### DIFF
--- a/src/sonic-config-engine/portconfig.py
+++ b/src/sonic-config-engine/portconfig.py
@@ -238,7 +238,8 @@ class BreakoutCfg(object):
             return hash((self.num_ports, tuple(self.supported_speed), self.num_assigned_lanes))
 
     def __init__(self, name, bmode, properties):
-        self._interface_base_id = int(name.replace(PORT_STR, ''))
+        self._name = name
+        self._front_panel_prefix, self._interface_base_id = self._parse_name_to_prefix_and_id()
         self._properties = properties
         self._lanes = properties ['lanes'].split(',')
         self._indexes = properties ['index'].split(',')
@@ -253,6 +254,12 @@ class BreakoutCfg(object):
 
         if not self._breakout_capabilities:
             raise RuntimeError("Unsupported breakout mode {}!".format(bmode))
+
+    def _parse_name_to_prefix_and_id(self):
+        match = re.match(swsscommon.FRONT_PANEL_PORT_REGEX, self._name)
+        if not match:
+            raise RuntimeError("Failed to parse front panel prefix of {}!", self._name)
+        return match.group(1), int(match.group(2))
 
     def _re_group_to_entry(self, group):
         if len(group) != BRKOUT_PATTERN_GROUPS:
@@ -300,7 +307,7 @@ class BreakoutCfg(object):
             lanes_per_port = entry.num_assigned_lanes // entry.num_ports
 
             for port in range(entry.num_ports):
-                interface_name = PORT_STR + str(self._interface_base_id + lane_id)
+                interface_name = self._front_panel_prefix + str(self._interface_base_id + lane_id)
 
                 lanes = self._lanes[lane_id:lane_id + lanes_per_port]
 
--- a/src/sonic-device-data/tests/hwsku_json_checker
+++ b/src/sonic-device-data/tests/hwsku_json_checker
@@ -8,7 +8,7 @@ import sys
 # Global variable
 PORT_ATTRIBUTES = ["default_brkout_mode"]
 OPTIONAL_PORT_ATTRIBUTES = ["fec", "autoneg", "port_type"]
-PORT_REG = "Ethernet(\d+)"
+PORT_REG = "(Ethernet)(\d+)"
 HWSKU_JSON = '*hwsku.json'
 INTF_KEY = "interfaces"
 
--- a/src/sonic-device-data/tests/platform_json_checker
+++ b/src/sonic-device-data/tests/platform_json_checker
@@ -8,7 +8,8 @@ import sys
 # Global variable
 PORT_ATTRIBUTES = ["index", "lanes", "breakout_modes"]
 ATTR_LEN = len(PORT_ATTRIBUTES)
-PORT_REG = "Ethernet(\d+)"
+# PORT_REG should be taken from scheme.h FRONT_PANEL_PORT_PREFIX_REGEX
+PORT_REG = "(Ethernet)(\d+)"
 PLATFORM_JSON = '*platform.json'
 INTF_KEY = "interfaces"
 CHASSIS_KEY = "chassis"
--- a/src/sonic-py-common/sonic_py_common/interface.py
+++ b/src/sonic-py-common/sonic_py_common/interface.py
@@ -7,8 +7,12 @@ SONiC interface types and access functions.
  "Human readable interface string":"Sonic interface prefix"
  Currently this is a static mapping, but in future could be stored as metadata in DB.
 """
+import re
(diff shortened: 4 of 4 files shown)


itamar-talmon:
This PR depends on https://github.com/Azure/sonic-swss-common/pull/598 so it will not pass until it merged

rules/sonic-py-common.mk
```diff
@@ -4,7 +4,8 @@ ifeq ($(ENABLE_PY2_MODULES), y)
     SONIC_PY_COMMON_PY2 = sonic_py_common-1.0-py2-none-any.whl
     $(SONIC_PY_COMMON_PY2)_SRC_PATH = $(SRC_PATH)/sonic-py-common
     $(SONIC_PY_COMMON_PY2)_DEPENDS += $(SWSSSDK_PY2)
-    $(SONIC_PY_COMMON_PY2)_PYTHON_VERSION = 2
+    $(SONIC_PY_COMMON_PY2)_DEBS_DEPENDS += $(PYTHON_SWSSCOMMON)
```
qiluo-msft, line 7: >PYTHON_SWSSCOMMON [](http://example.com/codeflow?start=45&length=17)

Thanks for fixing the dependency! It is not related to front panel port prefix. Could you raise a separate PR for the fix? We may need to backport the fix.

src/sonic-config-engine/portconfig.py
```diff
@@ -258,6 +259,14 @@ def __init__(self, name, bmode, properties):
 
         if not self._breakout_capabilities:
             raise RuntimeError("Unsupported breakout mode {}!".format(bmode))
+    def _parse_name_to_prefix_and_id(self):
```
qiluo-msft, line 262: >_parse_name_to_prefix_and_id [](http://example.com/codeflow?start=8&length=28)

Leave 1~2 lines before and after this function. Not use 0 or too many empty lines.

src/sonic-config-engine/portconfig.py
```diff
@@ -258,6 +259,14 @@ def __init__(self, name, bmode, properties):
...
+        match = re.search(swsscommon.FRONT_PANEL_PORT_PREFIX_REGEX + "(\d+)", self._name)
```
qiluo-msft, line 263: >FRONT_PANEL_PORT_PREFIX_REGEX [](http://example.com/codeflow?start=37&length=29)

There is an implicit assumption: there must be one and only one group in FRONT_PANEL_PORT_PREFIX_REGEX. This is not flexible and not easy to change in future.

Better to capture the prefix and id into one regex, no concatenating.

rules/sonic-py-common.mk
itamar-talmon: The reason I have to add the dependency here is because I needed the new port panel regex from swsscomon in sonic_py_common, if you think it's worth to make it a separate PR I can do it.

src/sonic-config-engine/portconfig.py
itamar-talmon: done

src/sonic-config-engine/portconfig.py
itamar-talmon: done

itamar-talmon:
@kcudnik @qiluo-msft - could you please check it out after my update?
Checks will fail until related commits will be merged (in the description).
Thank you!

rules/sonic-py-common.mk
qiluo-msft: The dependency fix itself is a good bug fix. Please help.

src/sonic-config-engine/portconfig.py
```diff
@@ -259,6 +260,13 @@ def __init__(self, name, bmode, properties):
         if not self._breakout_capabilities:
             raise RuntimeError("Unsupported breakout mode {}!".format(bmode))
 
+    def _parse_name_to_prefix_and_id(self):
+        match = re.match(swsscommon.FRONT_PANEL_PORT_REGEX, self._name)
+        try:
+            return match.group(1), int(match.group(2))
+        except Exception:
```
qiluo-msft, line 267: >Exception [](http://example.com/codeflow?start=15&length=9)

Please use more specific exception type.

src/sonic-config-engine/portconfig.py
```diff
@@ -259,6 +260,13 @@ def __init__(self, name, bmode, properties):
         if not self._breakout_capabilities:
             raise RuntimeError("Unsupported breakout mode {}!".format(bmode))
 
+    def _parse_name_to_prefix_and_id(self):
+        match = re.match(swsscommon.FRONT_PANEL_PORT_REGEX, self._name)
+        try:
```
qiluo-msft, line 265: >try [](http://example.com/codeflow?start=8&length=3)

Just check `match` instead of `try/except`?

rules/sonic-py-common.mk
itamar-talmon: Sure, no problem, i'll create a new PR with only the dependency fix and send it to you.

src/sonic-config-engine/portconfig.py
itamar-talmon: Yes, good idea, i'll do that

src/sonic-config-engine/portconfig.py
itamar-talmon: I removed the exception due to the above comment

rules/sonic-py-common.mk
itamar-talmon: Actually this was fixed in master few weeks ago by - https://github.com/Azure/sonic-buildimage/commit/e4ed475fcde0d68abb3b2814a71b76296642a7ee.
I just took the latest master and removed my change.

itamar-talmon:
@kcudnik @qiluo-msft - I updated the code with the new comments and rebased since the dependency fix was already merged to master by a different commit (you can check my comment there). Please check, thank you!

qiluo-msft (approved):

PR #10478, opened 2022-04-06
[orchagent] Set hash_seed for EPMS and MgmtTsToR

kellyyeh:
#### Why I did it
Add new device type EPMS and MgmtTsToR for setting hash_seed in orchagent docker.
hash_seed is automatically set to 0, this is mainly for reference of where "ToRRouter" is used in orchagent

#### How I did it
Update switch.json.j2

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

### DIFF
--- a/dockers/docker-orchagent/switch.json.j2
+++ b/dockers/docker-orchagent/switch.json.j2
@@ -3,7 +3,7 @@
 {% set hash_seed = 0 %}
 {% set hash_seed_offset = 0 %}
 {% if DEVICE_METADATA.localhost.type %}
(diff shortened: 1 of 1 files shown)


yxieca (approved):

Blueve:
Do we need back port this fix to 202012?

Blueve:
The default value is 0, so this change will not change configuration we generated for MgmtTsToR.

kellyyeh:
> The default value is 0, so this change will not change configuration we generated for MgmtTsToR.

Yes, it is just for reference if new devices get introduced in the future, therefore no need to backport

PR #10472, opened 2022-04-05
[sonic-swss]: Update submodule

saiarcot895:
This submodule update brings in the following changes:

```
50d5be2 Make changes to support compiling on Bullseye with GCC 10 (#2216)
0870cf5 [mirrororch]: Implement HW resources availability validation for SPAN/ERSPAN (#2187)
f4ec565 [vlanmgrd] fix use-after-free memory issue (#2211)
c2de7fc [QosOrch] The notifications cannot be drained in QosOrch in case the first one needs to retry (#2206)
5575935 [neighsyncd] increase neighsyncd timeout (#2209)
0f06910 [PBH] Implement Edit Flows (#2169)
6241bbf Remove redundant and problematic code to skip "pool" field in buffer profile handling (#2197)
a55343c [azp]: Set diff coverage threshhold to 80% (#2188)
390cae1 [portsorch]: Prevent LAG member configuration when port has active ACL binding (#2165)
c1d47e6 [VNET]Fixing nexthop group delete during route change (#2198)
8941cc0 [BFD]Registering BFD state change callback during session creation (#2202)
680c539 [vxlan] Remove tunnel map objects on VNET tunnel removal (#2150)
20dde0c Fix for handling broadcom DNX ASIC to have ipv4 and ipv6 ACL rules in separate tables. (#2178)
5b7c949 [FdbOrch] SAI_FDB_EVENT_MOVE generates update with empty update.entry.port_name (#2200)
7350d49 [Vxlanmgr] vnet netdev cleanup during config reload fix (#2191)
2bef62b Validate LAG has members before mirror session create (#2130)
1e4d4ce [VS test] Increase VS test time, skip dpb flaky test (#2195)
6eda965 [vstest]Migrating vs tests from using click commands to direct DB access (#2179)
```

Signed-off-by: Saikrishna Arcot <sarcot@microsoft.com>

#### Which release branch to backport (provide reason below if selected)

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

PR #10404, opened 2022-03-30
[submodule]: Advance sonic-swss submodule

nazariig:
Signed-off-by: Nazarii Hnydyn <nazariig@nvidia.com>

#### Why I did it
In order to include the following commit:
0f06910 [PBH] Implement Edit Flows (Azure/sonic-swss#2169)

[sonic-swss](https://github.com/Azure/sonic-swss)
```
50d5be2 Make changes to support compiling on Bullseye with GCC 10 (#2216)
0870cf5 [mirrororch]: Implement HW resources availability validation for SPAN/ERSPAN (#2187)
f4ec565 [vlanmgrd] fix use-after-free memory issue (#2211)
c2de7fc [QosOrch] The notifications cannot be drained in QosOrch in case the first one needs to retry (#2206)
5575935 [neighsyncd] increase neighsyncd timeout (#2209)
0f06910 [PBH] Implement Edit Flows (#2169)
6241bbf Remove redundant and problematic code to skip "pool" field in buffer profile handling (#2197)
a55343c [azp]: Set diff coverage threshhold to 80% (#2188)
390cae1 [portsorch]: Prevent LAG member configuration when port has active ACL binding (#2165)
c1d47e6 [VNET]Fixing nexthop group delete during route change (#2198)
8941cc0 [BFD]Registering BFD state change callback during session creation (#2202)
680c539 [vxlan] Remove tunnel map objects on VNET tunnel removal (#2150)
20dde0c Fix for handling broadcom DNX ASIC to have ipv4 and ipv6 ACL rules in separate tables. (#2178)
5b7c949 [FdbOrch] SAI_FDB_EVENT_MOVE generates update with empty update.entry.port_name (#2200)
7350d49 [Vxlanmgr] vnet netdev cleanup during config reload fix (#2191)
2bef62b Validate LAG has members before mirror session create (#2130)
1e4d4ce [VS test] Increase VS test time, skip dpb flaky test (#2195)
6eda965 [vstest]Migrating vs tests from using click commands to direct DB access (#2179)
```

#### How I did it
* Advanced submodule

#### How to verify it
1. make configure PLATFORM=mellanox
2. make target/sonic-mellanox.bin

#### Which release branch to backport (provide reason below if selected)

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

#### Link to config_db schema for YANG module changes

[...]

PR #10484, opened 2022-04-06
[swss]: Convert swss docker to bullseye

theasianpianist:

#### Why I did it
Update SWSS docker to bullseye

#### How I did it
Create new `docker-swss-layer-bullseye` docker and use it as the base for `docker-orchagent`

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111


lguohan (approved):

rules/docker-orchagent.mk
```diff
@@ -39,5 +39,5 @@ $(DOCKER_ORCHAGENT)_RUN_OPT += -v /var/log/swss:/var/log/swss:rw
 $(DOCKER_ORCHAGENT)_BASE_IMAGE_FILES += swssloglevel:/usr/bin/swssloglevel
 $(DOCKER_ORCHAGENT)_FILES += $(ARP_UPDATE_SCRIPT) $(ARP_UPDATE_VARS_TEMPLATE) $(SUPERVISOR_PROC_EXIT_LISTENER_SCRIPT)
 
-SONIC_BUSTER_DOCKERS += $(DOCKER_ORCHAGENT)
-SONIC_BUSTER_DBG_DOCKERS += $(DOCKER_ORCHAGENT_DBG)
+SONIC_BULLSEYE_DOCKERS += $(DOCKER_ORCHAGENT)
```
saiarcot895, line 42: `SONIC_BULLSEYE_DOCKERS` and `SONIC_BULLSEYE_DBG_DOCKERS` don't exist and aren't used, so lines adding to these variables can be removed.

saiarcot895:
Also, unless I'm missing something, #10472 will need to be merged in to get the swss fixes for compiling on Bullseye.

dockers/docker-orchagent/Dockerfile.j2
```diff
@@ -28,14 +28,14 @@ RUN apt-get update &&       \
 # Fix for gcc/python/iputils-ping not found in arm docker
 RUN apt-get install -f -y python2.7 python2.7-dev
 RUN apt-get install -y   \
-        gcc-8            \
+        gcc-10           \
```
saiarcot895, line 31: Does it need to be a specific version of GCC? Can it just be `gcc` here?

dockers/docker-orchagent/Dockerfile.j2
```diff
@@ -28,14 +28,14 @@ RUN apt-get update &&       \
...
         iputils-ping
 {% endif %}
 {% if CONFIGURED_ARCH == "armhf" %}
-RUN ln -s -f /usr/bin/gcc-8 /usr/bin/arm-linux-gnueabihf-gcc
+RUN ln -s -f /usr/bin/gcc-10 /usr/bin/arm-linux-gnueabihf-gcc
```
saiarcot895, line 35: I think if the above is changed to `gcc` instead of `gcc-10`, then these symlinks won't be needed.

dockers/docker-orchagent/Dockerfile.j2
saiarcot895: Actually, I realized that Python 2.7 is no logner available in Bullseye, so this needs to be changed anyways.

dockers/docker-swss-layer-bullseye/Dockerfile.j2
```diff
@@ -0,0 +1,22 @@
+{% from "dockers/dockerfile-macros.j2" import install_debian_packages, install_python_wheels, copy_files %}
+FROM docker-config-engine-bullseye
+
+## Make apt-get non-interactive
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN apt-get update
+
+{% if docker_swss_layer_bullseye_debs.strip() -%}
```
qiluo-msft, line 9: >docker_swss_layer_bullseye_debs [](http://example.com/codeflow?start=6&length=31)

The "-layer-" name is special. Could you just call it "docker-swss-buster"?

qiluo-msft:
You have a very big submodule change inside. Better to raise it in another PR?

prsunny:
submodule PR is merged - https://github.com/Azure/sonic-buildimage/pull/10404. Please rebase

dockers/docker-swss-layer-bullseye/Dockerfile.j2
theasianpianist: I am keeping convention with the existing `docker_swss_layer_buster`. As I understand it, multiple dockers require the same set of packages which is why we have the `docker_swss_layer_buster`. Once all the dockers are upgraded to bullseye, would the same be true, and we would require this `docker_swss_layer_bullseye` for common packages?

qiluo-msft (dismissed):

dockers/docker-orchagent/Dockerfile.j2
```diff
@@ -28,15 +28,9 @@ RUN apt-get update &&       \
 # Fix for gcc/python/iputils-ping not found in arm docker
 RUN apt-get install -f -y python2.7 python2.7-dev
```
saiarcot895, line 29: Does python 2.7 really need to be installed for armhf/arm64?

dockers/docker-swss-layer-bullseye/Dockerfile.j2
```diff
@@ -0,0 +1,22 @@
+{% from "dockers/dockerfile-macros.j2" import install_debian_packages, install_python_wheels, copy_files %}
+FROM docker-config-engine-bullseye
```
saiarcot895, line 2: This needs to be `docker-config-engine-bullseye-{{DOCKER_USERNAME}}:{{DOCKER_USERTAG}}`.

saiarcot895 (approved):

PR #10496, opened 2022-04-07
Add two extra lossless queues for bounced back traffic

bingwang-ms:
#### Why I did it
This PR is to add two extra lossless queues for bounced back traffic.
HLD https://github.com/Azure/SONiC/pull/950

SKUs include
```
Arista-7050CX3-32S-C32
Arista-7050CX3-32S-D48C8
Arista-7260CX3-D108C8
Arista-7260CX3-C64
Arista-7260CX3-Q64
```
#### How I did it
Update the `buffers.json.j2` template and `buffers_config.j2` template to generate new `BUFFER_QUEUE` table.
- For T1 devices, queue `2` and queue `6` are set as lossless queues on T0 facing ports.
- For T0 devices, queue `2` and queue `6` are set as lossless queues on T1 facing ports.
- Queue 7 is added as a new lossy queue as DSCP 48 is mapped to TC 7, and then mapped into Queue 7

#### How to verify it
1. Verified by UT
2. Verified by coping the new template and generate buffer config with `sonic-cfggen`

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012
- [ ] 202106
- [ ] 202111

This PR is to add two extra lossless queues for bounced back traffic.


lguohan:
can we have an enable/disable flag for this feature?

bingwang-ms:
> can we have an enable/disable flag for this feature?

Do you mean a flag for these extra queues, or the whole QoS remapping feature?

bingwang-ms:
> > can we have an enable/disable flag for this feature?
> 
> Do you mean a flag for these extra queues, or the whole QoS remapping feature? I think we can add a global flag in `DEVICE_METADATA` table.
> 
> ```
> "DEVICE_METADATA": {
>         "localhost": {
>             "mux_tunnel_qos_remapping" : "enable"
>         }
>     },
> ```
> 
> If the flag `mux_tunnel_qos_remapping` is set to `enable`, then we are going to generate the new `qos.config` and `buffer.config`, and then `orchanget` will consume it. Otherwise nothing will be changed. How do you think?

I drafted a HLD for adding a `FLAGS` table. Please see also https://github.com/Azure/SONiC/pull/982

files/build_templates/buffers_config.j2
```diff
@@ -182,6 +190,8 @@ def
 {{ defs.generate_queue_buffers(port_names_active) }}
 {% elif defs.generate_queue_buffers_with_inactive_ports is defined %}
 {{ defs.generate_queue_buffers_with_inactive_ports(port_names_active, port_names_inactive) }}
+{% elif generate_extra_lossless_buffers is defined %}
```
stephenxs, line 193: 1. Looks like the name `generate_extra_lossless_buffers` is a bit misleading.
   The macro is introduced to generate queue buffers for both lossless and lossy queues with additional lossless queues on ports between leaf routers and dual ToR routers. Maybe `generate_queue_buffers_with_extra_lossless_queues` or `generate_queue_buffers_dual_tor` is better.
2. Some vendors, like Nvidia, provide a `_with_inactive_ports` version queue/PG buffer generating macros for the purpose of reclaiming unused buffer. In this version, an extra argument `port_names_inactive` is passed.
So you may want to provide a `_with_inactive_ports` version for generating queue buffers for ports connecting to dual ToR.
3. The macro is defined in per-SKU `buffer_config_tx.j2` which was imported as `defs`. So we need `defs.` to reference it, like other macros

bingwang-ms: Thanks for the grest suggetsions.
For 1, I renamed the macro to `generate_queue_buffers_with_extra_lossless_queues`. It's more accurate. Thanks
For 2, I added the call of macro `generate_queue_buffers_with_extra_lossless_queues_with_inactive_ports` in `buffers_config.j2` . But the macro is not defined yet. We may implement the macro when we are updating template for NVIDIA devices
For 3, I didn't do the change as I think it makes more sense to define the macro `generate_queue_buffers_with_extra_lossless_queues` in `buffers.json.j2` as the macro is exactly the same for different SKUs and roles (T0 and T1). If we move the macro into `buffers_defaults_t0.j2` or `buffers_defaults_t1.j2`, that means lots of copy-paste. 
BTW, I also update the macro to accept port names joint with `,` like the existing macros.

Thanks

files/build_templates/buffers_config.j2
```diff
@@ -128,6 +128,16 @@ def
 {%- endfor %}
 {%- set port_names_active  = port_names_list_active  | join(',') %}
 
+{%- set port_names_list_extra_queues = [] %}
+{%- for port in PORT_ACTIVE %}
+    {% if (generate_queue_buffers_with_extra_lossless_queues is defined or generate_queue_buffers_with_extra_lossless_queues_with_inactive_ports is defined) and 
+            ('type' in DEVICE_METADATA['localhost'] and DEVICE_METADATA['localhost']['type'] == 'LeafRouter' and DEVICE_NEIGHBOR_METADATA is defined and DEVICE_NEIGHBOR[port].name in DEVICE_NEIGHBOR_METADATA and DEVICE_NEIGHBOR_METADATA[DEVICE_NEIGHBOR[port].name].type == 'ToRRouter') or
+            ('subtype' in DEVICE_METADATA['localhost'] and DEVICE_METADATA['localhost']['subtype'] == 'DualToR' and DEVICE_NEIGHBOR_METADATA is defined and DEVICE_NEIGHBOR[port].name in DEVICE_NEIGHBOR_METADATA and DEVICE_NEIGHBOR_METADATA[DEVICE_NEIGHBOR[port].name].type == 'LeafRouter') %}
```
neethajohn, line 135: can you put these in brackets since they should be evaluated first before the 'and' 
macro present and (type is leaf or dualtor)

files/build_templates/buffers_config.j2
```diff
@@ -128,6 +128,16 @@ def
...
+        {%- if port_names_list_extra_queues.append(port) %}{%- endif %}
+    {% endif %}
+{%- endfor %}
+{%- set port_names_extra_queues = port_names_list_extra_queues | join(',') %}
+
 {%- set port_names_list_inactive  = [] %}
 {%- for port in PORT_INACTIVE %}
```
neethajohn, line 142: you will need to have the same logic as 133 here as well when generate_queue_buffers_with_extra_lossless_queues_with_inactive_ports is defined, to find out the ports in the inactive list that need the extra queues

files/build_templates/buffers_config.j2
```diff
@@ -128,6 +128,16 @@ def
 {%- endfor %}
 {%- set port_names_active  = port_names_list_active  | join(',') %}
 
+{%- set port_names_list_extra_queues = [] %}
+{%- for port in PORT_ACTIVE %}
+    {% if (generate_queue_buffers_with_extra_lossless_queues is defined or generate_queue_buffers_with_extra_lossless_queues_with_inactive_ports is defined) and
```
neethajohn, line 133: are you able to access the macro as such? don't you need it to be defs.macro

neethajohn (changes requested):

files/build_templates/buffers_config.j2
bingwang-ms: I don't think it's a valid path.
Because the `PORT_INACTIVE` is the ports that are not in `DEVICE_NEIGHBOR`. So, we have no info to determine the type of neighbor. Hence, we don't know if extra queues are required because we need the neighbor's info.

files/build_templates/buffers_config.j2
bingwang-ms: Thanks. I added a `refs.` in front the call of `generate_queue_buffers_with_extra_lossless_queues_and_inactive_ports` so that vendors can define this macros as other marcos.

device/arista/x86_64-arista_7260cx3_64/Arista-7260CX3-D108C8/buffers.json.j2
```diff
@@ -1,3 +0,0 @@
-{%- set default_topo = 't0' %}
```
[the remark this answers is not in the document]
bingwang-ms: Good catch.
I moved the macro definition into a new template file `buffers_extra_queues.j2`, and the macro is imported to `buffers_defaults_t0.j2` and `buffers_defaults_t1.j2` for the invloved SKUs. So `buffers.json.j2` is not changed now.

files/build_templates/buffers_config.j2
bingwang-ms: Fixed.

files/build_templates/buffers_config.j2
bingwang-ms: Yes, I'm able to access the macro because it's defined in `buffers.json.j2`. The macro is defined before `{%- include 'buffers_config.j2' %}`, so it's not in any namespace. We can access the macro in `buffers_config.j2` directly.

But now we have to add namespace `defs` to visit the macro `generate_queue_buffers_with_extra_lossless_queues ` as it's now defined in `buffers_default_tX.j2` now.

stephenxs (dismissed):

files/build_templates/buffers_config.j2
neethajohn: Makes sense. My bad

neethajohn (dismissed):

stephenxs (approved):

stephenxs (dismissed):

neethajohn (dismissed):

bingwang-ms:
@stephenxs @neethajohn  The last commit introduced the `SYSTEM_DEFAULT` table into the template. Please find more details at https://github.com/Azure/sonic-buildimage/pull/10877.
If `TunnelQosRemap` is not enabled, the template will not be effective, then nothing is changed.
I also added two new PGs 2 and 6 in `BUFFER_PG` table
https://github.com/Azure/sonic-buildimage/blob/1aeb85b0135ba2d9b0af17842d24ac1b0d8d7abd/device/arista/x86_64-arista_7050cx3_32s/Arista-7050CX3-32S-D48C8/buffers_extra_queues.j2#L37-L53
It's because the port level `TC_TO_PG` map is changed as required by Broadcom. We are going to map TC 2 -> PG2 and TC 6 -> PG 6 at the port level.
Please feel free to discuss and comment. Thanks

stephenxs:
> @stephenxs @neethajohn The last commit introduced the `SYSTEM_DEFAULT` table into the template. Please find more details at #10877. If `TunnelQosRemap` is not enabled, the template will not be effective, then nothing is changed. I also added two new PGs 2 and 6 in `BUFFER_PG` table
> 
> https://github.com/Azure/sonic-buildimage/blob/1aeb85b0135ba2d9b0af17842d24ac1b0d8d7abd/device/arista/x86_64-arista_7050cx3_32s/Arista-7050CX3-32S-D48C8/buffers_extra_queues.j2#L37-L53
> 
> 
> It's because the port level `TC_TO_PG` map is changed as required by Broadcom. We are going to map TC 2 -> PG2 and TC 6 -> PG 6 at the port level.
> Please feel free to discuss and comment. Thanks

@bingwang-ms @neethajohn 
So this means for all ports we will have 2 additional PGs? 
- ports between T1 and dual ToR, there are 2 additional lossless PGs
- other ports, which had only 1 lossy and 2 lossless PGs, have 3 lossy and 2 lossless PGs
right?
Can we define the map on a per-vendor basis? We would still have TC 2/6 -> PG 0 because it doesn't make sense to allocate two additional lossy PGs and this wastes shared buffer.

bingwang-ms:
> > @stephenxs @neethajohn The last commit introduced the `SYSTEM_DEFAULT` table into the template. Please find more details at #10877. If `TunnelQosRemap` is not enabled, the template will not be effective, then nothing is changed. I also added two new PGs 2 and 6 in `BUFFER_PG` table
> > https://github.com/Azure/sonic-buildimage/blob/1aeb85b0135ba2d9b0af17842d24ac1b0d8d7abd/device/arista/x86_64-arista_7050cx3_32s/Arista-7050CX3-32S-D48C8/buffers_extra_queues.j2#L37-L53
> > 
> > It's because the port level `TC_TO_PG` map is changed as required by Broadcom. We are going to map TC 2 -> PG2 and TC 6 -> PG 6 at the port level.
> > Please feel free to discuss and comment. Thanks
> 
> @bingwang-ms @neethajohn So this means for all ports we will have 2 additional PGs?
> 
> * ports between T1 and dual ToR, there are 2 additional lossless PGs
> * other ports, which had only 1 lossy and 2 lossless PGs, have 3 lossy and 2 lossless PGs
>   right?
> 
> Taking T1 uplinks as an example, we do not have traffic with DSCP 2/6 (mapping to TC 2/6) received on uplink ports. Even if there are such TC => PG mappings, we do not need to allocate buffer for those PGs, right?

Make sense. There will be no traffic going lossy PG 2 and 6 of T1 uplink ports. I have removed the change. Thanks

device/arista/x86_64-arista_7050cx3_32s/Arista-7050CX3-32S-C32/buffers_defaults_t0.j2
```diff
@@ -44,3 +44,11 @@
         }
     },
 {%- endmacro %}
+
+{% if (SYSTEM_DEFAULTS is defined) and ('tunnel_qos_remap' in SYSTEM_DEFAULTS) and (SYSTEM_DEFAULTS['tunnel_qos_remap']['status'] == 'enabled') %}
```
stephenxs, line 48: We can also check `SYSTEM_DEFAULTS` in `buffers_config.j2` instead of in each SKU's buffer template. By doing so, a lot of redundant code can be avoided.
The idea is to generate an empty `port_names_list_extra_queues` list when remapping is disabled in `SYSTEM_DEFAULTS`
How do you think? @bingwang-ms

bingwang-ms: Good suggestion. I've updated the `buffers_config.j2` to add the check, and remove the check from `buffer_defaults.j2`. Please help review. Thanks

stephenxs (approved):
LGTM.

files/build_templates/buffers_config.j2
```diff
@@ -128,6 +128,16 @@ def
 {%- endfor %}
 {%- set port_names_active  = port_names_list_active  | join(',') %}
 
+{%- set port_names_list_extra_queues = [] %}
+{%- for port in PORT_ACTIVE %}
+    {%- if ((SYSTEM_DEFAULTS is defined) and ('tunnel_qos_remap' in SYSTEM_DEFAULTS) and (SYSTEM_DEFAULTS['tunnel_qos_remap']['status'] == 'enabled')) and (defs.generate_queue_buffers_with_extra_lossless_queues is defined or defs.generate_queue_buffers_with_extra_lossless_queues_with_inactive_ports is defined) and
```
neethajohn, line 133: We need not check for the existence of the macros (generate_queue) here. We are anyways checking it in L195 and L197

bingwang-ms: Make sense. Updated. Thanks

neethajohn (approved):

PR #10499, opened 2022-04-07
Enable P4RT at build time and disable at startup

ntoorchi:
#### Why I did it
Currently at the Azure build system, the P4RT container is disabled by default at the build time. Here the goal is to include the P4RT container at the build time while disabling it at the runtime. The user can enable/disable the p4rt app through the config based on the preference. 

#### How I did it
Changed the config in rules/config and init-cfg.json.j2

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111


bocon13:
@ntoorchi can you add some time and space comparisons?

rhalstea:
@ntoorchi can we rebuild this, or do we know why the build failed?

ntoorchi:
> @ntoorchi can we rebuild this, or do we know why the build failed?

@rhalstea marvell_armhf fails. It's due to this block:
https://github.com/sonic-net/sonic-buildimage/blob/485edbfd67530744f9ce39c2e6878731a33ef1ae/slave.mk#L178
The error is `recipe commences before first target.  Stop.` so it might be due to `@echo`.

rhalstea:
Thanks @ntoorchi,
 
Was this re-based to be at the most recent sonic-buildimage? I'm not sure how to tell if it was, but marvell_armhf doesn't seem to have an issue on the mainline.

ntoorchi:
@rhalstea no this is not rebased recently, but I tried on a locally rebased sonic-buildimage too. This looks like a syntax error in the `slave.mk`, happens when `INCLUDE_P4RT=y` and `CONFIGURED_ARCH=armhf` (running on arm processor). I guess the reason is the tab before `@echo` when it is not under any target.  I tried using  `$(info ...)` instead of echo and it worked. Should I add the fix to this PR or open a separate PR or issue?

ntoorchi:
@prsunny please review this PR, aiming to include the P4RT at the build time while disabling it at the runtime. The PR also includes a fix to the `slave.mk` in order to pass `marvell_armhf` build. The other pipeline builds were passing previously, but currently failing in master due to an upstream issue. I will re-run the pipeline as soon as it is fixed.

ntoorchi:
@rhalstea all check passed!

rhalstea:
Thanks @ntoorchi!

@prsunny would you be able to review this? Asking since you seem to be who we mainly send PINS PRs to :)

CC'ing: @bhagatyj and @bocon13

bhagatyj:
LGTM from PINS

ntoorchi:
@qiluo-msft, @zhangyanzhao can you please review this PR or assign someone to review? The goal is to include the P4RT at the build time. However it is disabled by default at the runtime. The PR also includes a fix to the slave.mk in order to pass marvell_armhf build.

xumia (approved):

qiluo-msft:
Share the same concern: can you add some time and space comparisons?

ntoorchi:
> Share the same concern: can you add some time and space comparisons?

Thanks @qiluo-msft for reviewing. I am trying to use Azure Pipeline artifacts as references but build time and size usually varies by time. In general, I have noticed build time is quite comparable, and there is some increase in size due to adding P4RT.  

Pipeline it is taking a long time (or failing) at this moment. But the `sonic-buildimage.broadcom` was successful and it took around 6h while Master build has taken around 8h. But at the normal condition, the PR build time was quite comparable with other builds, e.g. around 2h 30m.

Comparing size of downloaded artifact with the one from the commit [9750cb4](https://github.com/sonic-net/sonic-buildimage/commit/9750cb48c642281157ddd405674bdca9710e1edc), there is slightly increase in size due to adding P4RT. We have disabled some of the debugging tools to minimize the overhead.

| commit | size sonic-buildimage.broadcom | size sonic-aboot-broadcom.swi | 
| --- | --- | --- | 
| PR 10499 | 10.27G | 1.13G | 
| commit 9750cd4 | 9.88G | 1.12 G | 

Please let me know if you have any specific comparison in mind.

qiluo-msft:
sonic-buildimage.broadcom increases by 390M. That's significant increasing.
Since most users do not use P4RT, could you maintain private branches/patches on top of the public branches, so there is no impact to unrelated users. We have to support various platforms and some have super small disks.

bocon13:
I'm trying to understand the requirement here.

From my perspective, it looks like enabling PINS adds ~10MB to the switch image that we deploy. This seems like a very small impact to users, and it will be great to have PINS in the image to enable more users to try it out without rebuilding.

The ~400MB is the build time artifacts on a machine that likely has 1TB+ available. Personally, I don't see a problem with this, but if there is concern, we can just add a step to remove intermediate artifacts from the PINS build before they are stored in the CI system.

What do you think?

qiluo-msft:
I see your point and refine my last comment. The size of build time artifacts is not a concern. ~10M of image size increasing is still a valid concern. @lguohan do you have any comment?

ntoorchi:
> I see your point and refine my last comment. The size of build time artifacts is not a concern. ~10M of image size increasing is still a valid concern. @lguohan do you have any comment?

Thanks @qiluo-msft for your comment. As @bocon13 mentioned, enabling PINS would be valuable as currently many vendors are trying it out. It will be more convenient if it's present in the image, encouraging more users to trying it. 

cc @bhagatyj @rhalstea

bhagatyj:
Since PINS introduction last year, multiple organizations have been working on different use cases. We are also in the path of upstreaming tests. As a percentage of image size, the 10M is quite small. For this size, this container brings in a critical functionality of SDN and remote controllability.

qiluo-msft (approved):

qiluo-msft:
Please solve the conflicts.

ntoorchi:
> Please solve the conflicts.

Thanks Qi, I rebased and resolved the conflict but the pipeline builds are failing at this moment.

rhalstea:
Thanks @ntoorchi @qiluo-msft and @xumia,

The build and test look good. Can we merge this?

PR #10511, opened 2022-04-08
[202012][services] kill container on stop in warm/fast mode

stepanblyschak:

#### Why I did it
To optimize stop on warm boot.

#### How I did it
Added kill for containers

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012
- [ ] 202106
- [ ] 202111


vaibhavhd:
These changes LGTM, but there is an error in PR tests on master branch for the corresponding change. I would suggest to RCA those issues to ensure there isn't a risk introduced.

files/scripts/radv.sh
```diff
@@ -1,84 +0,0 @@
-#!/bin/bash
```
vaibhavhd, line 1: Do you know why GIT shows that original `radv.sh` was deleted? Is it because the new script has new permissions (chmod permissions)? 

In the same regard, does other `sh` scripts introduced here have `executable` permissions? I think that's necessary.

stepanblyschak: All have executable permissions. radv.sh and service_mgmt.sh are not the same.

vaibhavhd: I understand that they are different files. My minor concern was why GIT shows that original `radv.sh` was deleted. This usually happens when contents are replaced and permissions are changed.

image

vaibhavhd (approved):

vaibhavhd:
I just realized that this PR hasn't run any tests, and the one on master has gone through round of testing and failed.
I suggest we sort out and understand the cause of failure/regression on master before we merge this.

files/scripts/radv.sh
stepanblyschak: @vaibhavhd The content was changed, please check the line SERVICE="radv" and SERVICE="${SCRIPT_NAME%.*}" in service_mgmt.sh

vaibhavhd:
@stepanblyschak I am not able to merge this as the branch is `out-of-date`. Can you update the branch with latest of 202012 and retry?

Issue #10326, opened 2022-03-22
[202012] DHCPmon core generated in fast-reboot recovery path


vaibhavhd:
DHCMPmon core generated after fast-reboot - `dhcpmon-Vlan1000 (terminated by SIGSEGV (core dumped); not expected)`

#### Steps to reproduce the issue:
1. Install latest 202012 image. (57 in the case of this repro).
2. Do fast-reboot. 
3. Check logs for dhcmpmon failure and core generated. This issue is NOT consistently reproducible.

#### Describe the results you received:

```
Mar 21 05:45:36.675483 str2-7050cx3-acs-02 INFO dhcp_relay#/supervisord: start container_up: BEGIN
Mar 21 05:45:36.675483 str2-7050cx3-acs-02 INFO dhcp_relay#/supervisord: start container_up: args: feature=dhcp_relay, owner=local, version=20201231.57 DB: set_owner=kube state_data={
Mar 21 05:45:36.675483 str2-7050cx3-acs-02 INFO dhcp_relay#/supervisord: start container_up: args: feature=dhcp_relay, owner=local, version=20201231.57 DB: set_owner=kube state_data={
Mar 21 05:45:36.675483 str2-7050cx3-acs-02 INFO dhcp_relay#/supervisord: start     "current_owner": "local",
Mar 21 05:45:36.675483 str2-7050cx3-acs-02 INFO dhcp_relay#/supervisord: start     "update_time": "2022-03-21 05:45:29",
Mar 21 05:45:36.675483 str2-7050cx3-acs-02 INFO dhcp_relay#/supervisord: start     "container_id": "dhcp_relay",
Mar 21 05:45:36.675483 str2-7050cx3-acs-02 INFO dhcp_relay#/supervisord: start     "container_id": "dhcp_relay",
Mar 21 05:45:36.675521 str2-7050cx3-acs-02 INFO dhcp_relay#/supervisord: start     "remote_state": "none",
Mar 21 05:45:36.675521 str2-7050cx3-acs-02 INFO dhcp_relay#/supervisord: start     "container_version": "0.0.0",
Mar 21 05:45:36.675544 str2-7050cx3-acs-02 INFO dhcp_relay#/supervisord: start     "system_state": "up"
Mar 21 05:45:36.675544 str2-7050cx3-acs-02 INFO dhcp_relay#/supervisord: start }
[...]

PR #10553, opened 2022-04-12
[DHCPv6 relay monitor] Add an offset bound check during DHCPv6 header inspection to prevent segmentation fault

shlomibitton:
#### Why I did it
Fixes issue https://github.com/Azure/sonic-buildimage/issues/10326
Adding an offset bound check during DHCPv6 header inspection to prevent segmentation fault issue.

#### How I did it
Each time offset is incremented to fetch a different field in the DHCPv6 header, check if the buffer contains this offset and only then try to fetch it.
If the offset is out of the DHCP header bounds, print a message to syslog and return.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012
- [ ] 202106
- [x] 202111

### DIFF
--- a/src/dhcpmon/src/dhcp_device.c
+++ b/src/dhcpmon/src/dhcp_device.c
@@ -75,6 +75,15 @@ static bool dhcpv6_enabled;
 #define OP_JSET     (BPF_JMP | BPF_JSET | BPF_K)    /** bpf jset */
 #define OP_LDXB     (BPF_LDX | BPF_B    | BPF_MSH)  /** bpf ldxb */
 
+#define BOUND_CHECK(offset)                                            \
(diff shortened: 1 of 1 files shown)


shlomibitton:
@yxieca @kellyyeh Can you please review?
Thanks

liat-grozovik (approved):

kellyyeh:
@shlomibitton I think this fix should only be added to 202106 since it does not use dhcp6relay and relies on dhcpmon for counter.
We are reverting dhcpv6mon on all other branches.

shlomibitton:
@yxieca @kellyyeh so I will create a new PR against 202106 only, thanks

PR #10555, opened 2022-04-12
[CG-Fix-CVE-2021-44906] Patching on thrift.0.14.1 for package minimist

richardyu-ms:
Signed-off-by: richardyu-ms <richard.yu@microsoft.com>

#### Why I did it
resolve alert CVE-2021-44906
#### How I did it
Patching on thrift.0.14.1 for package minimist
#### How to verify it
build thrift 0.14.1 in build docker
build barefoot, broadcom and mellanox images

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [x] 202111


msosyak:
Please select 202111 branch to backport

msosyak (dismissed):

xumia (dismissed):

src/thrift_0_14_1/Makefile
```diff
@@ -0,0 +1,31 @@
...
+.SHELLFLAGS += -e -x
+
+THRIFT_VERSION = 0.14.1
+
+MAIN_TARGET = libthrift0_$(THRIFT_VERSION)_$(CONFIGURED_ARCH).deb
+DERIVED_TARGETS = libthrift-dev_$(THRIFT_VERSION)_$(CONFIGURED_ARCH).deb \
+		  python3-thrift_$(THRIFT_VERSION)_$(CONFIGURED_ARCH).deb \
+		  thrift-compiler_$(THRIFT_VERSION)_$(CONFIGURED_ARCH).deb
+
+THRIFT_LINK_PRE = https://archive.apache.org/dist/thrift
```
qiluo-msft, line 12: >https://archive.apache.org/dist/thrift [](http://example.com/codeflow?start=18&length=38)

@xumia, will reproducible build help if this download link is down?

src/thrift_0_14_1/Makefile
```diff
@@ -0,0 +1,31 @@
...
+$(addprefix $(DEST)/, $(MAIN_TARGET)): $(DEST)/% :
+	rm -rf thrift-$(THRIFT_VERSION)
+
+	wget -O "thrift_$(THRIFT_VERSION).tar.gz" "$(THRIFT_LINK_PRE)/$(THRIFT_VERSION)/thrift-$(THRIFT_VERSION).tar.gz"
+
+	tar -xvzf ./thrift_$(THRIFT_VERSION).tar.gz
+	pushd thrift-$(THRIFT_VERSION)
+
+	# Disable php perl and few other packages as they need additional packages to be installed
+	patch -p1 < ../patch/0001-Remove-unneeded-packages.patch
```
qiluo-msft, line 23: >patch [](http://example.com/codeflow?start=1&length=5)

keep using a series file as before?

qiluo-msft: slave.mk support *.patch folder, you may use it. ref: [https://github.com/Azure/sonic-buildimage/blob/0191300b96465e3124733a1ab91b9510c9e0393c/slave.mk#L461](https://github.com/Azure/sonic-buildimage/blob/0191300b96465e3124733a1ab91b9510c9e0393c/slave.mk#L461)

src/thrift_0_14_1/Makefile
xumia: Yes, wget will work, see line 17, it is only a prefix.

src/thrift_0_14_1/Makefile
richardyu-ms: there we used this method, build from the download src package
1. code in submodule will trigger the security alert, even we do not need those package
2. Patch will be restored after submodule build (use the SONIC_DPKG_DEB), in this way, it will use the series to apply patches
3. Build workflow change, like conditional restore after building, will impact other build targets, and need more effort

then we get the solution
1. introduce the source only when build needs them, then, we use make to download and unpackage the source code
2. make file to generate the debs, and everything will be under control by 
3. just one package changed, and we used this way in thrift 0.13.0 and thrift 0.11.0

qiluo-msft (approved):

PR #10560, opened 2022-04-13
[DHCPv6 relay monitor] [202106] Add an offset bound check during DHCPv6 header inspection to prevent segmentation fault

shlomibitton:
#### Why I did it
Fixes issue https://github.com/Azure/sonic-buildimage/issues/10326
Adding an offset bound check during DHCPv6 header inspection to prevent segmentation fault issue.

#### How I did it
Each time offset is incremented to fetch a different field in the DHCPv6 header, check if the buffer contains this offset and only then try to fetch it.
If the offset is out of the DHCP header bounds, print a message to syslog and return.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [x] 202106
- [ ] 202111


shlomibitton:
@yxieca @kellyyeh can you please review this? this is for 202106

shlomibitton:
@kellyyeh @yxieca all checkers passed and the PR approved can we proceed to merge?
Thanks

PR #3505, opened 2019-09-24
[Mellanox] Stop pmon ahead of syncd

stephenxs:
**- What I did**
***Issue Overview***
****shutdown flow****
For any shutdown flow, which means all dockers are stopped in order, pmon docker stops after syncd docker has stopped, causing pmon docker fail to release sx_core resources and leaving sx_core in a bad state. The related logs are like the following:
```
INFO syncd.sh[23597]: modprobe: FATAL: Module sx_core is in use.
INFO syncd.sh[23597]: Unloading sx_core[FAILED]
INFO syncd.sh[23597]: rmmod: ERROR: Module sx_core is in use
```
****config reload & service swss.restart****
In the flows like "config reload" and "service swss restart", the failure cause further consequences:
1. sx_core initialization error with error message like "sx_core: create EMAD sdq 0 failed. err: -16"
2. syncd fails to execute the create switch api with error message "syncd_main: Runtime error: :- processEvent: failed to execute api: create, key: SAI_OBJECT_TYPE_SWITCH:oid:0x21000000000000, status: SAI_STATUS_FAILURE"
3. swss fails to call SAI API "SAI_SWITCH_ATTR_INIT_SWITCH", which causes orchagent to restart. This will introduce an extra 1 or 2 minutes for the system to be available, failing related test cases. 

****reboot, warm-reboot & fast-reboot****
In the reboot flows including "reboot", "fast-reboot" and "warm-reboot" this failure doesn't have further negative effects since the system has already rebooted. In addition, "warm-reboot" requires the system to be shutdown as soon as possible to meet the GR time restriction of both BGP and LACP. "fast-reboot" also requires to meet the GR time restriction of BGP which is longer than LACP. In this sense, any unnecessary steps should be avoided. It's better to keep those flows untouched.

****summary****
To summarize, we have to come up with a way to ensure:

1. shutdown pmon docker ahead of syncd for "config reload" or "service swss restart" flow;
2. don't shutdown pmon docker ahead of syncd for "fast-reboot" or "warm-reboot" flow in order to save time.
3. for "reboot" flow, either order is acceptable.

***Solution***
To solve the issue, pmon shoud be stopped ahead of syncd stopped for all flows except for the warm-reboot.

**- How I did it**
1. To stop pmon ahead of syncd stopped. This is done in /usr/local/bin/syncd.sh::stop() and for all shutdown sequence.
[...]

PR #10567, opened 2022-04-13
[Fastboot] Delay PMON service for better fastboot performance

shlomibitton:

#### Why I did it

Following this finding, and the motivation to ensure these services will not interfere in the future, PMON is delayed in 90 seconds until the system finish the init flow after fastboot.

bootchart_no_optimizations

#### How I did it

- Add a timer for PMON service.
- Exclude for MLNX platform the start trigger of PMON when SYNCD starts in case of fastboot.
- Copy the timer file to the host bin image.

#### How to verify it

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012
- [ ] 202106
- [x] 202111


files/scripts/syncd.sh
```diff
@@ -46,6 +46,10 @@ function startplatform() {
 function waitplatform() {
     
     if [[ x"$sonic_asic_platform" == x"mellanox" ]]; then
+        BOOT_TYPE=`getBootType`
+        if [[ x"$BOOT_TYPE" == x"fast" ]]; then
+            return
+        fi
         debug "Starting pmon service..."
         /bin/systemctl start pmon
         debug "Started pmon service"
```
stepanblyschak, line 55: This was added by https://github.com/Azure/sonic-buildimage/pull/3505 to fix some issues related to pmon/SDK. Won't that issue happen again with your change?

vaibhavhd: This change can be made generic to all platforms. CC @yxieca.

Also, if this is not made generic, then the comments on `files/build_templates/pmon.timer` should be modified to say:
`# This delay is for fast-reboot performance on MLNX platform`

vaibhavhd: Also, don't you want to do this change for warm-reboot as well? I am curious why only for fast-reboot?

files/scripts/syncd.sh
```diff
@@ -46,6 +46,10 @@ function startplatform() {
 function waitplatform() {
     
     if [[ x"$sonic_asic_platform" == x"mellanox" ]]; then
+        BOOT_TYPE=`getBootType`
+        if [[ x"$BOOT_TYPE" == x"fast" ]]; then
+            return
```
vaibhavhd, line 51: It would be meaningful to have a log line here mentioning that for fastboot (or warm), `pmon` start will be delayed as governed by `files/build_templates/pmon.timer`

shlomibitton: Log added.

files/scripts/syncd.sh
shlomibitton: It is a generic change, this exception is only for MLNX platform, so I added a condition for it in order to keep the delay and not start PMON with SYNCD start flow.
Regarding warm-boot, I totally agree we can do it for both types, fixed it.

files/build_templates/pmon.timer
```diff
@@ -0,0 +1,12 @@
+[Unit]
+# This delay is for fast-reboot performance
```
vaibhavhd, line 2: Correction - fast-reboot and warm-reboot performance.

files/scripts/syncd.sh
```diff
@@ -44,11 +44,16 @@ function startplatform() {
...
 
 function waitplatform() {
-    
+
+    BOOT_TYPE=`getBootType`
     if [[ x"$sonic_asic_platform" == x"mellanox" ]]; then
-        debug "Starting pmon service..."
-        /bin/systemctl start pmon
-        debug "Started pmon service"
+        if [[ x"$BOOT_TYPE" != x"cold" ]]; then
```
vaibhavhd, line 50: Same comment as on lldp PR:
Minor: This is correct, but it may be better to match against whitelist BOOT_TYPEs (fast, warm, fastfast).
This is so that this condition does not incorrectly evaluate to true for any other boottype that gets added in future.

vaibhavhd (approved):
Minor comments. LGTM otherwise.

shlomibitton:
@vaibhavhd @stepanblyschak I fixed your comments, can you please review and approve?
Thanks

stepanblyschak (dismissed):

liat-grozovik:
@vaibhavhd kindly reminder to review

vaibhavhd (approved):

liat-grozovik (approved):

liat-grozovik:
As prev approved by @vaibhavhd and merge only done to avoid checkers failures, PR is now merged.

qiluo-msft:

shlomibitton:
@qiluo-msft I created a separate PR for 202012:
https://github.com/Azure/sonic-buildimage/pull/10745

PR #10568, opened 2022-04-13
[Fastboot] Delay LLDP service for better fastboot performance

shlomibitton:

#### Why I did it

Following this finding, and the motivation to ensure these services will not interfere in the future, LLDP is delayed in 90 seconds until the system finish the init flow after fastboot.

bootchart_no_optimizations

#### How I did it

- Add a timer for LLDP service.
- Copy the timer file to the host bin image.

#### How to verify it

This PR is dependent on PR: https://github.com/Azure/sonic-buildimage/pull/10567

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012
- [ ] 202106
- [x] 202111

### DIFF
--- a/files/build_templates/per_namespace/lldp.timer
+++ b/files/build_templates/per_namespace/lldp.timer
@@ -1,0 +1,12 @@
+[Unit]
+# This delay is for fast-reboot performance
+Description=Delays LLDP docker until SONiC has started
+PartOf=lldp.service
+
+[Timer]
+OnUnitActiveSec=0 sec
+OnBootSec=1min 30 sec
+Unit=lldp.service
+
+[Install]
+WantedBy=timers.target sonic.target sonic-delayed.target
--- a/files/scripts/syncd.sh
+++ b/files/scripts/syncd.sh
@@ -50,7 +50,7 @@ function waitplatform() {
         /bin/systemctl start pmon
         debug "Started pmon service"
     fi
-    if [[ x"$BOOT_TYPE" != x"cold" ]]; then
+    if [[ x"$BOOT_TYPE" = @(x"fast"|x"warm"|x"fastfast") ]]; then
         debug "LLDP service is delayed by a timer for better fast/warm boot performance"
     else
         debug "Starting lldp service..."
--- a/files/scripts/syncd.sh
+++ b/files/scripts/syncd.sh
@@ -44,11 +44,18 @@ function startplatform() {
 }
 
 function waitplatform() {
-    
+
     if [[ x"$sonic_asic_platform" == x"mellanox" ]]; then
         debug "Starting pmon service..."
         /bin/systemctl start pmon
         debug "Started pmon service"
+    fi
+    if [[ x"$BOOT_TYPE" != x"cold" ]]; then
+        debug "LLDP service is delayed by a timer for better fast/warm boot performance"
+    else
+        debug "Starting lldp service..."
+        /bin/systemctl start lldp
+        debug "Started lldp service"
     fi
 }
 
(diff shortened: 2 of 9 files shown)


yxieca (approved):

files/build_templates/per_namespace/lldp.timer
```diff
@@ -0,0 +1,12 @@
+[Unit]
+# This delay is for fast-reboot performance
+Description=Delays LLDP docker until SONiC has started
+PartOf=lldp.service
+
+[Timer]
+OnUnitActiveSec=0 sec
+OnBootSec=1min 30 sec
+Unit=lldp.service
```
stepanblyschak, line 9: lldp.service is a per-namespace container. Will that work for multi-asic? Should we have several timers?

vaibhavhd:
I think this change will impact all kinds of reboot - warm/fast/cold/etc. Do you agree?

If yes, is that what we want to do? Can we do this only for fast and warmboot case?

shlomibitton:
> I think this change will impact all kinds of reboot - warm/fast/cold/etc. Do you agree?
> 
> If yes, is that what we want to do? Can we do this only for fast and warmboot case?

@vaibhavhd I agree, I added a condition for it on syncd.sh like we do for pmon today.

files/scripts/syncd.sh
```diff
@@ -44,12 +44,19 @@ function startplatform() {
...
 
 function waitplatform() {
-    
+
     if [[ x"$sonic_asic_platform" == x"mellanox" ]]; then
         debug "Starting pmon service..."
         /bin/systemctl start pmon
         debug "Started pmon service"
     fi
+    if [[ x"$BOOT_TYPE" != x"cold" ]]; then
```
vaibhavhd, line 53: Minor: This is correct, but it may be better to match against whitelist BOOT_TYPEs (fast, warm, fastfast).
This is so that this condition does not evaluate to true for any other boottype that gets added in future.

files/build_templates/per_namespace/lldp.timer
```diff
@@ -0,0 +1,12 @@
+[Unit]
+# This delay is for fast-reboot performance
```
vaibhavhd, line 2: Correction - fast-reboot and warm-reboot performance.

vaibhavhd (dismissed):
Minor comments pending, LGTM otherwise. Please address Stepan's concerns before merging.

files/build_templates/per_namespace/lldp.timer
stepanblyschak: Add  has_timer to this file https://github.com/Azure/sonic-buildimage/blob/master/files/build_templates/init_cfg.json.j2#L32

shlomibitton:
@vaibhavhd @stepanblyschak I fixed your comments, can you please review and approve?
Thanks

files/build_templates/per_namespace/lldp.timer
shlomibitton: Done

files/scripts/syncd.sh
shlomibitton: Done

files/build_templates/per_namespace/lldp.timer
shlomibitton: Done

liat-grozovik:
@vaibhavhd @stepanblyschak kindly reminder to review

vaibhavhd (approved):

stepanblyschak (approved):

qiluo-msft:

shlomibitton:
@qiluo-msft I created a separate PR for 202012:
https://github.com/Azure/sonic-buildimage/pull/10744

PR #10573, opened 2022-04-14
Fix race condition between networking service and interface-config service

Junchao-Mellanox:
#### Why I did it

The PR is aimed to fix a bug that mgmt port eth0 may loss IP even if user configured static IP of eth0. This is not a always reproduceable issue, the reproducing flow is like:

1.	Systemd starts networking service, which runs a dhcp based configuration and assigned an ip from dhcp.
2.	Systemd starts interface-config service who depends on networking service
3.	Interface-config service runs command  “ifdown –force eth0”, check [line](https://github.com/Azure/sonic-buildimage/blob/16717d2dc51f74fa711ed7b4392ce5e4f7e71c29/files/image_config/interfaces/interfaces-config.sh#L4). but networking service is still running so that this [line](https://github.com/CumulusNetworks/ifupdown2/blob/ac32bec0e24d64c583778f387050a7b6f4269db0/ifupdown2/ifupdown/main.py#L74) failed with error: “error: Another instance of this program is already running.”. This error is printed by ifupdown2 lib who is the main process of networking service. So, ifdown actually does not work here, the ip of eth0 is not down.
4.	Interface-config service updates /etc/networking/interface to static configuration.
5.	Interface-config service runs command “systemctl restart networking”. This command kills the previous networking related processes (log: networking.service: Main process exited, code=killed, status=15/TERM), and try to reconfigure the ip address with static configuration. But it detects that the configured IP and the existing IP are the same, and it does not really configure the ip to kernel. Hence, the ip is still getting from dhcp. (this could be a bug of ifupdown2: previous ip is from dhcp, new ip is a static ip, it treats them as same instead of re-configuring the IP)
6.	When the lease of the ip expires, the ip of eth0 is removed by kernel and the issue reproduces.

The issue is not always reproduceable because networking service usually runs fast so that it won't hit step#3.

#### How I did it

Check networking service state before running "ifdown –force eth0", wait for it done if it is activating.

#### How to verify it

Manual test.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

### DIFF
--- a/files/image_config/interfaces/interfaces-config.sh
+++ b/files/image_config/interfaces/interfaces-config.sh
@@ -9,7 +9,7 @@ function wait_networking_service_done() {
     while [[ "${_WDOG_CNT}" -le "${_WDOG_MAX}" ]]; do
         networking_status="$(systemctl is-active networking 2>&1)"
 
-        if [ "${networking_status}" == active ] || [ "${networking_status}" == inactive ] || [ "${networking_status}" == failed ] ; then
+        if [[ "${networking_status}" == active || "${networking_status}" == inactive || "${networking_status}" == failed ]] ; then
             return
         fi
 
@@ -29,11 +29,11 @@ if [[ $(ifquery --running eth0) ]]; then
 fi
 
 # Check if ZTP DHCP policy has been installed
-if [ -e /etc/network/ifupdown2/policy.d/ztp_dhcp.json ]; then
+if [[ -e /etc/network/ifupdown2/policy.d/ztp_dhcp.json ]]; then
     # Obtain port operational state information
     redis-dump -d 0 -k "PORT_TABLE:Ethernet*"  -y > /tmp/ztp_port_data.json
 
-    if [ $? -ne 0 ] || [ ! -e /tmp/ztp_port_data.json ] || [ "$(cat /tmp/ztp_port_data.json)" = "" ]; then
+    if [[ $? -ne 0 || ! -e /tmp/ztp_port_data.json || "$(cat /tmp/ztp_port_data.json)" = "" ]]; then
         echo "{}" > /tmp/ztp_port_data.json
     fi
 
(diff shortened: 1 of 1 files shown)


keboliu (dismissed):

liat-grozovik:
@qiluo-msft  could you please help to review or suggest someone who can?

Junchao-Mellanox:
Hi @yxieca , could you please help review and sign-off?

files/image_config/interfaces/interfaces-config.sh
```diff
@@ -1,6 +1,30 @@
...
+    local -r _TIMEOUT="1s"
+
+    while [[ "${_WDOG_CNT}" -le "${_WDOG_MAX}" ]]; do
+        networking_status="$(systemctl is-active networking 2>&1)"
+
+        if [[ "${networking_status}" == active ]]; then
+            return
+        fi
+
+        echo "interfaces-config: networking service is running, wait for it done"
```
yxieca, line 16: What if the networking service is stopped?

Junchao-Mellanox: Fixed

liat-grozovik:
@yxieca kindly reminder to review fixes and signoff

files/image_config/interfaces/interfaces-config.sh
```diff
@@ -1,6 +1,30 @@
...
+function wait_networking_service_done() {
+    local -i _WDOG_CNT="1"
+    local -ir _WDOG_MAX="30"
+
+    local -r _TIMEOUT="1s"
+
+    while [[ "${_WDOG_CNT}" -le "${_WDOG_MAX}" ]]; do
+        networking_status="$(systemctl is-active networking 2>&1)"
+
+        if [ "${networking_status}" == active ] || [ "${networking_status}" == inactive ] || [ "${networking_status}" == failed ] ; then
```
yxieca, line 12: Please be consistent with '[[' and ']]' as rest of the script.

Junchao-Mellanox: Fixed.

Junchao-Mellanox:
@yxieca kindly reminder.

yxieca (approved):

Junchao-Mellanox:
Hi @judyjoseph , could you please help cherry-pick to 202111?

For 202012, there is no clean cherry-pick, I will add separate PR.

PR #8927, opened 2021-10-08
Fix tagged VlanInterface if attached to multiple vlan as untagged member

qiluo-msft:
#### Why I did it
Fix several bugs:
1. If one vlan member belongs to multiple vlans, and if any of the vlans is "Tagged" type, we respect the tagged type
2. If one vlan member belongs to multiple vlans, and all of the vlans have no "Tagged" type, we override it to be a tagged member
3. make sure `vlantype_name` is assigned correctly in each iteration

#### How to verify it
1. Test the command line to parse a minigraph and make sure the output does not change.
```
./sonic-cfggen -m minigraph.mlnx20.xml
```
The minigraph is for HwSKU Mellanox-SN2700-D40C8S8.

2. Test on a DUT with HwSKU Mellanox-SN2700-D40C8S8
```
sudo config load_minigraph
show vlan brief
```
Checked the "Port Tagging" column in the output.

#### Which release branch to backport (provide reason below if selected)

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012
- [ ] 202106

PR #10589, opened 2022-04-16
[202012] Fix tagged VlanInterface if attached to multiple vlan as untagged member

qiluo-msft:
Backport https://github.com/Azure/sonic-buildimage/pull/8927 to 202012 branch


kktheballer (approved):
Validated AzD scenario using 202012 minigraph.py change

neethajohn (approved):

PR #10228, opened 2022-03-14
[sonic-cfggen] make minigraph parser fail when speed and lanes are not in PORT table

wen587:
Fix https://github.com/Azure/sonic-buildimage/issues/9676
[Done] Lanes is fixed by https://github.com/Azure/sonic-buildimage/pull/10362
[Done]Wait for UT speed update
#### Why I did it
Config db schema generated by minigraph can’t pass yang validation, PORT table does not have 'lanes' field.
#### How I did it
cfggen command fail when 'lanes' and 'speed' are not provided
#### How to verify it
Run 'sonic-cfggen -m xxx.xml --print-data' to make sure command fail when 'lanes' not in PORT table
#### Which release branch to backport (provide reason below if selected)

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

PR #10617, opened 2022-04-20
[multi-asic]: Skip must field check in config_db for default namespace in multi-asic platform

SuvarnaMeenakshi:

#### Why I did it
For multi-asic VS testbed bring up, minigraph is copied to DUT and topology service is started.
topology service uses hwsku from minigraph to run the topology script for the right hwsku, using the command below:
```
  HWSKU=`sonic-cfggen -m /etc/sonic/minigraph.xml -v "DEVICE_METADATA['localhost']['hwsku']" 2>&1`
```
The above command fails after PR 10228.
After https://github.com/Azure/sonic-buildimage/pull/10228;   https://github.com/Azure/sonic-buildimage/blob/master/files/image_config/topology/topology.sh#L14 fails to get hwsku as the must-have fields are not present in default namespace for multi-asic platform.

#### How I did it
Check must-have fields only for single asic or only on namespaces in multi-asic platform.

#### How to verify it
Able to generate config_db in multi-asic platform without warning message.
Able to start topology service in multi-asic vs platform.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

### DIFF
--- a/src/sonic-config-engine/sonic-cfggen
+++ b/src/sonic-config-engine/sonic-cfggen
@@ -32,7 +32,7 @@ from functools import partial
 from minigraph import minigraph_encoder, parse_xml, parse_device_desc_xml, parse_asic_sub_role, parse_asic_switch_type
 from portconfig import get_port_config, get_breakout_mode
 from redis_bcc import RedisBytecodeCache
-from sonic_py_common.multi_asic import get_asic_id_from_name, get_asic_device_id
+from sonic_py_common.multi_asic import get_asic_id_from_name, get_asic_device_id, is_multi_asic
 from sonic_py_common import device_info
 from swsscommon.swsscommon import SonicV2Connector, ConfigDBConnector, SonicDBConfig, ConfigDBPipeConnector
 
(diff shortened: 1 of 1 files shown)


arlakshm (approved):
lgtm

qiluo-msft:
The `_must_field_by_yang` is reverted, so this PR is not needed any more. You may check conflicting for details.

ganglyu:
If multi-asic platform has a port without lanes and speed, maybe we should modify sonic-port.yang?
@SuvarnaMeenakshi Can you explain the scenario?

SuvarnaMeenakshi:
> platform has a port without lanes and speed, maybe we should modify sonic-port.yang

In multi-asic, currently the PORT table is generated for default namespace, though the default namespace does not have any asic or PORT associated. PORT table is generated because minigraph.xml contains PORTS for the device. But as there port_config.ini for default namespace, there is no lanes generated. 
@arlakshm Currently lanes and speed are mandatory fields in sonic-port.yang, can this be made non-mandatory fields?

SuvarnaMeenakshi:
closing this PR as https://github.com/Azure/sonic-buildimage/pull/10683 reverts the PR #10228.

ganglyu:
> > platform has a port without lanes and speed, maybe we should modify sonic-port.yang
> 
> In multi-asic, currently the PORT table is generated for default namespace, though the default namespace does not have any asic or PORT associated. PORT table is generated because minigraph.xml contains PORTS for the device. But as there is no port_config.ini for default namespace, there is no lanes generated. @arlakshm Currently lanes and speed are mandatory fields in sonic-port.yang, can this be made non-mandatory fields?

Maybe we can delete PORT table in this case?

PR #10619, opened 2022-04-20
[Mellanox] Adjust PSU voltage WA

Junchao-Mellanox:
#### Why I did it

InvalidPsuVolWA.run might raise exception if user power off PSU when it is running. This exception is not caught and will be raised to psud which causes psud failed to update PSU data to DB.

#### How I did it

1. Change the log level when WA does not work. This could happen when user power off PSU, hence changing the log level from error to warning is better
2. Change the wait time from 5 to 1 to avoid introduce too much delay in psud. 1 second is usually enough per my test
3. Give a default return value for function get_voltage_low_threshold and get_voltage_high_threshold to avoid exception reach to psud

#### How to verify it

Manual test.
Run sonic-mgmt regression

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [x] 202111

### DIFF
--- a/platform/mellanox/mlnx-platform-api/sonic_platform/psu.py
+++ b/platform/mellanox/mlnx-platform-api/sonic_platform/psu.py
@@ -397,6 +397,7 @@ class Psu(FixedPsu):
 
         return None
 
+    @utils.default_return(None)
     def get_voltage_high_threshold(self):
         """
         Retrieves the high threshold PSU voltage output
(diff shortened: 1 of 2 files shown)


keboliu (approved):

liat-grozovik (approved):

Junchao-Mellanox:
@judyjoseph could you please help cherry-pick to 202111?

PR #10624, opened 2022-04-20
[Arista] Fix arista-net initramfs hook

Staphylo:
#### Why I did it

The `arista-net` initramfs hook exits early when a mgmt NIC is missing.
This leads to mgmt interfaces not properly renamed and possibly `eth0` missing.

#### How I did it

Fixed the line in `arista-net` that fails when one NIC is missing.
The error is tied to the syntax which is incompatible with `set -e`

#### How to verify it

Tested on products where one of the management NIC is missing.

Backport will be made manually to 202111 and 202012 as `arista-net`
has changed since.

Fix `arista-net` initramfs hook when a mgmt NIC is missing

### DIFF
--- a/files/initramfs-tools/arista-net
+++ b/files/initramfs-tools/arista-net
@@ -47,7 +47,9 @@ arista_net_rename() {
     local new_name="$2"
     local from_name="$3"
     devname=$(arista_net_devname "$device_path" "$from_name")
-    [ -n "$devname" ] && ip link set "$devname" name "$new_name"
+    if [ -n "$devname" ]; then
+        ip link set "$devname" name "$new_name"
+    fi
 }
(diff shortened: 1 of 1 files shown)


zyboy2000:
rw_dir=${rootmnt}/host/$image_dir/rw
work_dir=${rootmnt}/host/$image_dir/work
mkdir -p "$rw_dir"
mkdir -p "$work_dir"
mount -n -o lowerdir=${rootmnt},upperdir=${rw_dir},workdir=${work_dir} -t overlay root-overlay ${rootmnt}

after sonic-install,  $(rw_dir) has  /etc ,/home  ...............directory  Where did these files come from？（which script ?） Thanks

PR #10689, opened 2022-04-26
[DHCP relay] Enlarge alias size to be equal to the definition of linu…

irene-pan1202:
#### Why I did it
The maximum alias size defined by isc-dhcp is 16, if the alias length exceeds 16, which will cause the Circuit-ID of option82 error.

#### How I did it
Enlarge alias size to be equal to the definition of linux source.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111


src/isc-dhcp/patch/0013-Enlarge-alias-size-to-be-equal-to-the-def-of-linux.patch
jasper-hou-micas, line 1: It shall be better if the  following log be kept.
log_debug("Failed to retrieve alias for interface name '%s'. Defaulting to interface name.", ifname)

gord1306: Given the planned migration of isc-dhcp-relay to kea, I'm wondering whether this patch is still necessary

PR #10698, opened 2022-04-27
[201811] Minigraph parser changes to select mmu profiles based on SonicQosProfile attribute

neethajohn:
#### Why I did it
There is a need to select different mmu profiles based on deployment type

#### How I did it
There will be separate subfolders (RDMA-CENTRIC, TCP-CENTRIC, BALANCED) in each hwsku folder which contains deployment specific mmu and qos settings. SonicQosProfile attribute in the minigraph will be used to determine which settings to use. If that attribute is not present, the default settings that exist in the hwsku folder will be used

#### How to verify it
Manually verified by creating subfolders under DX010 folder and updated minigraph on the device with SonicQosProfile pointing to one of the subfolders. Verified by loading minigraph and confirming the mmu and ecn settings are as expected.

Unit tests have been updated to include the SonicQosProfile attribute in the minigraphs used for buffer and qos rendering. Since the dir structure does not exist now, code flow will take no action. Once the dir structure exists, the profile selection path will get executed but since all 3 subfolders have identical values, existing test cases should hold good.

Build sonic-config-engine python wheel with the changes and build was successful.

### DIFF
--- a/src/sonic-config-engine/minigraph.py
+++ b/src/sonic-config-engine/minigraph.py
@@ -543,8 +543,7 @@ def select_mmu_profiles(profile, platform, hwsku):
             file_in_dir = os.path.join(dir_path, file_item)
             if os.path.isfile(file_in_dir):
                 base_file = os.path.join(path, file_item)
-                exec_cmd("sudo cp {} {}".format(file_in_dir, base_file)
-
+                exec_cmd("sudo cp {} {}".format(file_in_dir, base_file))
 
 ###############################################################################
 #
--- a/src/sonic-config-engine/minigraph.py
+++ b/src/sonic-config-engine/minigraph.py
@@ -410,6 +410,7 @@ def parse_meta(meta, hname):
     mgmt_routes = []
     erspan_dst = []
     deployment_id = None
+    qos_profile = []
     device_metas = meta.find(str(QName(ns, "Devices")))
     for device in device_metas.findall(str(QName(ns1, "DeviceMetadata"))):
         if device.find(str(QName(ns1, "Name"))).text.lower() == hname.lower():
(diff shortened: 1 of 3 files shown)


src/sonic-config-engine/minigraph.py
```diff
@@ -410,6 +410,7 @@ def parse_meta(meta, hname):
     mgmt_routes = []
     erspan_dst = []
     deployment_id = None
+    qos_profile = []
```
yxieca, line 413: why is this a list?

src/sonic-config-engine/tests/sample-dell-6100-t0-minigraph.xml
```diff
@@ -716,6 +716,11 @@
             <a:Reference i:nil="true"/>
             <a:Value>10.0.0.16</a:Value>
           </a:DeviceProperty>
+          <a:DeviceProperty>
+          <a:Name>SonicQosProfile</a:Name>
+          <a:Reference i:nil="true"/>
+          <a:Value>Rdma-centric</a:Value>
```
yxieca, line 722: RDMA-CENTRIC?

neethajohn: not needed. L540 in parser takes care of converting it to uppercase

src/sonic-config-engine/minigraph.py
neethajohn: Fixed

src/sonic-config-engine/minigraph.py
```diff
@@ -517,6 +520,31 @@ def filter_acl_mirror_table_bindings(acls, neighbors, port_channels):
...
+              the copy list in the dir path and copy it over to the base path.
+    """
+    if not profile:
+        return
+
+    files_to_copy = ['pg_profile_lookup.ini', 'qos.json.j2', 'buffers_defaults_t0.j2', 'buffers_defaults_t1.j2']
+
+    path = os.path.join('/usr/share/sonic/device', platform, hwsku)
+
+    dir_path = os.path.join(path, profile.upper())
```
yxieca, line 540: Not sure we want to do profile.upper() here, this will limit our options to up case only.

neethajohn: I added that based on the naming for the subfolders where everything is in upper case. Should we rely on the exact name in the minigraph?

yxieca: I think we should rely on exact match.

yxieca (approved):

lguohan (approved):

PR #10702, opened 2022-04-28
Add gbsyncd-credo to barefoot platform

byu343:
#### Why I did it
This change will install the gbsyncd-credo image to barefoot platform, which has already been installed to broadcom platform. The gbsyncd-credo is needed by arista_7170b_64c. 
#### How to verify it
We verified that the build can pass and the docker image is in the barefoot image.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111


platform/components/docker-gbsyncd-credo/Dockerfile.j2
```diff
@@ -20,7 +20,7 @@ debs/
 
 RUN dpkg -i \
 {% for deb in docker_gbsyncd_credo_debs.split(' ') -%}
-{% if 'libsaibcm' not in deb -%}
+{% if 'libsaibcm' not in deb and 'bfnsdk' not in deb and 'bfnplatform' not in deb -%}
```
jimmyzhai, line 23: Why are bfnsdk and bfnplatform packages in DEPENDS list of docker gbsyncd-credo? Even I do not remember for libsaibcm. If not dependent, let's remove the condition checks; otherwise to do them in previous `COPY` command is enough.

byu343: The reason is that gbsyncd docker depends on credo SAI and SYNCD, and SYNCD depends on the ASIC SAI, but we cannot install both ASIC SAI and credo SAI in the same docker.
The idea here is to NOT install the ASIC SAI packages to the gbsyncd docker, and it seems whether to copy them doesn't matter.

jimmyzhai: But ASIC SAI is only dependent while compiling syncd, or building SYNCD package. ASIC SAI package should not be in the list of `docker_gbsyncd_credo_debs`. Please double check.

byu343: I checked it again and noticed that $(SYNCD)_DEPENDS are not in docker_gbsyncd_credo_debs , while $(SYNCD)_RDEPENDS are in the docker_gbsyncd_credo_debs. For broadcom platform, it works fine without filtering libsaibcm from docker_gbsyncd_credo_debs now (likely some change after the initial submission of docker-gbsyncd-credo made it possible). For the barefoot platform, it is still have $(SYNCD)_RDEPENDS, so I make the change "[Refactor dependency to bfn_sai](https://github.com/Azure/sonic-buildimage/pull/10702/commits/1bf4cec287902519417d9b108c2efb75166db8da)" in this PR to fix it.

jimmyzhai (approved):
LGTM

PR #10714, opened 2022-04-29
[buffer] RDMA buffer config profiles

kevinskwang:
#### Why I did it
Different traffic pattern requires different RDMA tuning parameters. But currently we just support one for each hwsku.
#### How I did it
Separate into three profiles: 
RDMA-CENTRIC: (RDMA centric)  
TCP-CENTRIC: (TCP Friendly)  
BALANCED: Default is BALANCED. When missing, acting as BALANCED.  

And move the profiles to a common directory, all the hwsku which shared the profile would link to the common profile from the device specific diretory.
#### How to verify it
Build image and the buffer/qos function should be work fine with different profile groups.

- [x] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111


device/common/profiles/th2/7260/BALANCED/pg_profile_lookup.ini
```diff
@@ -0,0 +1,8 @@
+# PG lossless profiles.
```
neethajohn, line 1: please include the 50g speed and 5m,40m,300m lines as well

yxieca: Updated with copying 100G data (the original 50G data was higher than 100G config.

device/common/profiles/th/gen/BALANCED/buffers.json.j2
```diff
@@ -0,0 +1,2 @@
+{%- set default_topo = 't1' %}
```
neethajohn, line 1: this is different for each hwsku. for eg, for 7060-D48C8 it is 't0' whereas for 7060 C32 it is 't1'

neethajohn (approved):

PR #10720, opened 2022-05-01
[BFN] Canceling PSU platform API calls on SIGTERM

akokhan:
Signed-off-by: Andriy Kokhan <andriyx.kokhan@intel.com>
Co-authored-by: Taras Keryk <tarasx.keryk@intel.com>

#### Why I did it
Sometime, SIGTERM processing by psud takes more then default 10sec (please see stopwaitsecs in http://supervisord.org/configuration.html).

Due to this, the following two testcases may fail:
```
test_pmon_psud_stop_and_start_status
test_pmon_psud_term_and_start_status
```

Also, limited `get_bmc_version()` execution time to 1 sec.

#### How I did it
Introduced two levels of SIGTERM handlers:
1. The PSU class level `signal_handler()` that executes a default handler (`psud` SIGTERM handler) and avoids subsequent `psu_info_get()` executions.
2. The `cancel_on_sigterm()` decorator that executes class level `signal_handler()` and raises an exception to cancel current `psu_info_get()` execution.

#### How to verify it
Run SONiC CTs:
```
test_pmon_psud_stop_and_start_status
test_pmon_psud_term_and_start_status
```

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [x] 202111

### DIFF
--- a/platform/barefoot/sonic-platform-modules-bfn-montara/sonic_platform/platform_utils.py
+++ b/platform/barefoot/sonic-platform-modules-bfn-montara/sonic_platform/platform_utils.py
@@ -3,6 +3,8 @@
 try:
     import os
     import subprocess
+    import signal
+    from functools import wraps
 
 except ImportError as e:
     raise ImportError(str(e) + "- required module not found")
@@ -20,3 +22,24 @@ def file_create(path, mode=None):
         run_cmd(['touch', path])
     if (mode is not None):    
         run_cmd(['chmod', mode, path])
+
+def cancel_on_sigterm(func):
+    """
+    Wrapper for a function which has to be cancel on SIGTERM
+    """
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        def handler(sig, frame):
+            if sigterm_handler:
+                sigterm_handler(sig, frame)
+            raise Exception("Canceling {}() execution...".format(func.__name__))
+
+        sigterm_handler = signal.getsignal(signal.SIGTERM)
+        signal.signal(signal.SIGTERM, handler)
+        result = None
+        try:
+            result = func(*args, **kwargs)
+        finally:
+            signal.signal(signal.SIGTERM, sigterm_handler)
+        return result
+    return wrapper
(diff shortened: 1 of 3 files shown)


akokhan:
@prgeor , please review. Thank you

akokhan:
@lguohan , please approve and merge. Thanks

akokhan:
@prgeor , @lguohan , could you please take a look. Thank you!

akokhan:
@prgeor , could you please take a look?

akokhan:
@prgeor , could you please approve and merge? Thanks

akokhan:
@lguohan , please approve and merge. Thanks

akokhan:
@qiluo-msft , please approve and merge. Thank you

qiluo-msft:
Sorry for late response, we will soon review it.

liuh-80 (commented):
I checked this PR and it's not conflict with the recent change in libswsscommon.
Because this project is a pure python project and not using pubsub from swsscommon, so the signal handler in this PR will not be blocked by the infinite loop in pubsub.

platform/barefoot/sonic-platform-modules-bfn-montara/sonic_platform/platform_utils.py
```diff
@@ -18,5 +26,55 @@ def run_cmd(cmd):
...
         run_cmd(['touch', path])
-    if (mode is not None):    
+    if (mode is not None):
         run_cmd(['chmod', mode, path])
+
+def cancel_on_sigterm(func):
+    """
+    Wrapper for a function which has to be cancel on SIGTERM
+    """
+    @wraps(func)
```
liuh-80, line 36: I have a concern about the 'signal handler backup/restore' this decorator:
1. Currently this code will work well because it's only 1 decorator. and seems there is no another code in this project to similar backup/restore.

2. However, because this is a decorator, so it's maybe use to another different method in future, then if 2 methods using this decorator running simultaneously, the signal handler will messed up by this decorator. here is a example:

step1: App using default signal handler handeler_D.

step1: Method1 start, signaler handler switch to handler_N, bacup handler is signal_D

step2: Method2 start, signal handler switch to handler_M, and backup handler will be handler_N

step3: Method1 end, signal handler restore to handler_D

step4: Method2 end, signal handler restore to handler_N, which is wrong because signal handler should be handler_D

I suggest define a global signal handler for this app, and also backup/restore the global signal handler.

akokhan: I agree that this scenario is possible in multi-threaded environment. But this is not a case with psud as of now. So, not sure whether it worth the efforts. Probably we can address this separately. What do you think?

prgeor (approved):

PR #10725, opened 2022-05-02
[PINS] Allow P4RT App to read configs from CONFIG_DB.

rhalstea:
#### Why I did it
The initial commit for the P4RT docker hard coded all the flags which makes it difficult to configure at runtime. Reading them from the CONFIG_DB allows for more flexibility.

#### How I did it
Updated the p4rt.sh startup script to first read from CONFIG_DB then apply any configs it finds. If a config is not set then a default (specified by the executable) is used. 

#### How to verify it
Update data in the CONFIG_DB and restart the P4RT docker.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

The P4RT app can be configured through the CONFIG_DB.

### DIFF
--- a/dockers/docker-sonic-p4rt/Dockerfile.j2
+++ b/dockers/docker-sonic-p4rt/Dockerfile.j2
@@ -24,6 +24,7 @@ RUN apt-get clean -y      && \
     rm -rf /debs
 
 COPY ["start.sh", "p4rt.sh", "/usr/bin/"]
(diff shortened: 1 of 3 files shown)


rhalstea:
Hi @lguohan @zhangyanzhao,

Would someone have time to review this change? All it does is take sonic-pins flags that were previously hard coded and let users configure them at startup time.

zhangyanzhao:
@prsunny

prsunny (approved):

PR #10731, opened 2022-05-03
Added Support for BGP allow list feature to have route-map action of setting tag

abdosi:
What I did:
Added support to create route-map action `set tag <user define value>`
when the the allow prefix list  matches. The tag can ben define by user in
constants.yml.

Why I did:
Since for Allow List feature we `call` from base route-map allow-list route-map having `set tag` option provides way for base route-map to do ` match tag` and take any further action if needed. Adding tag provide metadata that can used by base route-map

Verify:

Unit test updated
Manual Verification.


prsunny (commented):
lgtm, can you also explain in description as to what would be a use-case/scenario where we need to use tag?

abdosi:
> lgtm, can you also explain in description as to what would be a use-case/scenario where we need to use tag?

@prsunny use case being we want to take further action if we are matching prefix list route-map.

prsunny (approved):
lgtm

StormLiangMS (approved):
LGTM

PR #10737, opened 2022-05-03
Revert "Fix buffer template for Arista SKU."

qiluo-msft:
Reverts Azure/sonic-buildimage#10663

To fix the unit test of sonic-config-engine.


device/arista/x86_64-arista_7800r3_48cq2_lc/Arista-7800R3-48CQ2-C48/buffers_defaults_t2.j2
```diff
@@ -18,26 +19,26 @@
     },
     "BUFFER_PROFILE": {
         "ingress_lossless_profile": {
-            "pool":"ingress_lossless_pool",
+            "pool":"[BUFFER_POOL|ingress_lossless_pool]",
```
neethajohn, line 22: this format is incorrect. we don't use referencing in master branch

PR #10744, opened 2022-05-04
[202012] [Fastboot] Delay LLDP service for better fastboot performance

shlomibitton:
This PR is to backport a fix https://github.com/Azure/sonic-buildimage/pull/10568
This PR is dependent on PR: https://github.com/Azure/sonic-buildimage/pull/10745
#### Why I did it

Following this finding, and the motivation to ensure these services will not interfere in the future, LLDP is delayed in 90 seconds until the system finish the init flow after fastboot.

#### How I did it

Add a timer for LLDP service.
Copy the timer file to the host bin image.

#### How to verify it

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012
- [ ] 202106
- [ ] 202111


liat-grozovik:
there is a failure related to LinkManagerStateMachineTest which is not the scope of this fix.
retriggering. if still failing will need the owner to take further look

liat-grozovik (approved):

PR #10745, opened 2022-05-04
[202012] [Fastboot] Delay PMON service for better fastboot performance

shlomibitton:
#### Why I did it

Following this finding, and the motivation to ensure these services will not interfere in the future, PMON is delayed in 90 seconds until the system finish the init flow after fastboot.

#### How I did it

Add a timer for PMON service.
Exclude for MLNX platform the start trigger of PMON when SYNCD starts in case of fastboot.
Copy the timer file to the host bin image.

#### How to verify it

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012
- [ ] 202106
- [ ] 202111


files/build_templates/init_cfg.json.j2
```diff
@@ -21,8 +21,8 @@
 {%- set features = [("bgp", "enabled", false, "enabled"),
                    ("database", "always_enabled", false, "always_enabled"),
                    ("dhcp_relay", "enabled", false, "enabled"),
-                   ("lldp", "enabled", false, "enabled"),
-                   ("pmon", "enabled", false, "enabled"),
+                   ("lldp", "enabled", true, "enabled"),
```
vivekreddynv, line 24: why was timer for lldp set to true here? Isn't there a separate PR for that. https://github.com/Azure/sonic-buildimage/pull/10744

shlomibitton: there is, it was a mistake, fixed now

qiluo-msft (approved):

PR #10750, opened 2022-05-05
[build][asan] make dpkg cache asan-aware

Yakiv-Huryk:
Currently, the build with ASAN_ENABLE=y reuses the packages built with
ASAN_ENABLE=n (and vice versa). To address this issue, ASAN_ENABLE is added to the
SONIC_COMMON_FLAGS_LIST.

Signed-off-by: Yakiv Huryk <yhuryk@nvidia.com>

#### Why I did it
To make dpkg cache use/rebuild the packages for ASAN_ENABLE=y/n.
#### How I did it
Added ASAN_ENABLE to the SONIC_COMMON_FLAGS_LIST which is a part of DEP_FLAGS for all the packages.
#### How to verify it
Built with ASAN_ENABLE=y/n and checked the .flags .log files.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

### DIFF
--- a/Makefile.cache
+++ b/Makefile.cache
@@ -71,7 +71,9 @@ SONIC_COMMON_FLAGS_LIST :=  $(CONFIGURED_PLATFORM) \
                             $(CONFIGURED_ARCH) \
                             $(BLDENV) \
                             $(SONIC_DEBUGGING_ON) \
-                            $(SONIC_PROFILING_ON) $(SONIC_ENABLE_SYNCD_RPC)
+                            $(SONIC_PROFILING_ON) \
+                            $(SONIC_ENABLE_SYNCD_RPC) \
+                            $(ENABLE_ASAN)
 SONIC_COMMON_DPKG_LIST  :=  debian/control debian/changelog debian/rules \
                             debian/compat debian/install debian/copyright
 SONIC_COMMON_BASE_FILES_LIST  := sonic-slave-jessie/Dockerfile.j2 sonic-slave-jessie/Dockerfile.user.j2 \


Makefile.cache
```diff
@@ -71,7 +71,9 @@ SONIC_COMMON_FLAGS_LIST :=  $(CONFIGURED_PLATFORM) \
                             $(CONFIGURED_ARCH) \
                             $(BLDENV) \
                             $(SONIC_DEBUGGING_ON) \
-                            $(SONIC_PROFILING_ON) $(SONIC_ENABLE_SYNCD_RPC)
+                            $(SONIC_PROFILING_ON) \
+                            $(SONIC_ENABLE_SYNCD_RPC) \
+                            $(ENABLE_ASAN)
```
xumia, line 76: Where is the flag used? It is only a flag? why not use cache=none/wcache, if we want to disable/rebuild cache?

Yakiv-Huryk: It's used for building a sonic image with ASAN-enabled containers (currently, swss and mellanox-syncd are supported).
More details: https://github.com/Azure/sonic-buildimage/pull/9857
The idea is to have the same cache working (in rw mode) for both regular and ASAN-enabled builds.
Currently, it's not possible since there is no distinction between packages built with/without ASAN_ENABLE=y.

liat-grozovik:
@xumia could you please review?

Makefile.cache
xumia: Can we only enable it for special targets, such as only for docker-orchagent? It will make the all the targets with  ENABLE_ASAN flag not able to share with the ones without the flag. Is it expected?

Yakiv-Huryk: Ok, this will work as well. I've updated the PR.

xumia (approved):

PR #10779, opened 2022-05-07
[Mellanox] Credo Y-cable | add more log info, checks, fix exception message

ayurkiv-nvda:
#### Why I did it
Script fails when there is an exception while reading

#### How I did it
Add more logs and checks. Fix wrong variable naming and messages.

#### How to verify it
Provoke exception while **read_eeprom()** and check that it is handled properly

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111


liat-grozovik:
@ayurkiv-nvda good to have more logs but does it fix the read exception? why do we get to that flow?

ayurkiv-nvda:
> @ayurkiv-nvda good to have more logs but does it fix the read exception? why do we get to that flow?

@liat-grozovik the problem here is when exception happen code cant handle it properly because there are typo "device_address" instead of "offset"

nazariig (approved):

stephenxs (approved):

liat-grozovik (approved):

PR #10800, opened 2022-05-10
[Arista] Update platform submodules

Staphylo:
- Refactor pci topology logic for chassis (fixes some chassis commands and chassisd on linecard)
 - Introduce new cooling algorithm
 - Fix linecard poweroff logic when supervisor is going down
 - Fix linecard status led leading to `system-health` crashing
 - Misc fixes


kenneth-arista:
Tagging @arlakshm for awareness

arlakshm (approved):

PR #10751, opened 2022-05-05
[reboiot-cause] Fix a broken symlink of previous-reboot-cause file removal issue

mlok-nokia:
#### Why I did it
When the previous-reboot-cause has a broken symlink, the following back trace is shown in the syslog and system-running state show degraded. And rebooting the system will not be able to generated a new symlink of the new previous-reboot-cause. 
fixes #10746
#### How I did it
Somehow, when the  symlink file /host/reboot-cause/previous-reboot-cause is broken (which its destination files doesn't exist in this case),  the current condition check "if os.path,exists(PREVIOUS_REBOOT_CAUSE_FILE)" will return False in determine-reboot-cause script.  Hence, the current previous-reboot-cause is not been removed and the recreation of the new previous-reboot-cause failed.  In case of previous-reboot-cause is a broken synlink file, add condition os.path.islink(PREVIOUS_REBOOT_CAUSE) to check and allow the remove operation happens.

#### How to verify it
1) Manually make the /host/reboot-cause/previous-reboot-cause to be a broken symlink file by removing its destination file
2) reboot the system
3) verify the syslog, the backtrace should not be shown and "systemctl is-system-running" should show "running" instead of "degraded"

#### Which release branch to backport (provide reason below if selected)

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

PR #10813, opened 2022-05-11
Revert "[reboiot-cause] Fix a broken symlink of previous-reboot-cause file removal issue"

lizhuhuams:
Reverts Azure/sonic-buildimage#10751


lizhuhuams (commented):
need double check the changes.

PR #10871, opened 2022-05-19
[201911] Advance sonic-swss pointer

neethajohn:
Signed-off-by: Neetha John <nejo@microsoft.com>

Submodule update contains the following commit
923db73 [201911][pfcwd] Avoid ingress drop by not attaching zero profiles when pfc storm is detected (Azure/sonic-swss#2279)


abdosi:
i can manually update this.

Issue #10141, opened 2022-03-03
[GCU or bgpcfgd]bgp prefix updating only fail for prefixes_v4 removal when GCU apply-patch


wen587:
#### Description
Test based on BGP_ALLOWED_PREFIX yang https://github.com/Azure/sonic-buildimage/pull/10142
bgp prefix removal only fails on prefixes_v4 when GCU apply-patch to remove.
Not sure it is a bug from `bgpcfgd` or `GCU`.

#### Steps to reproduce the issue:
1. Load bgp prefix config and check it takes into effect
```
admin@vlab-01:~/hijack$ cat add.load
{
    "BGP_ALLOWED_PREFIXES": {
        "DEPLOYMENT_ID|0": {
            "prefixes_v4": [
                "10.20.0.0/16",
                "10.30.0.0/16"
            ],
            "prefixes_v6": [
                "fc02:20::/64",
                "fc02:30::/64"
            ]
        }
    }
}

admin@vlab-01:~/hijack$ sudo config load -y add.load
Running command: /usr/local/bin/sonic-cfggen -j add.load --write-to-db

admin@vlab-01:~/hijack$ show run all | grep BGP_ALL -A11
    "BGP_ALLOWED_PREFIXES": {
        "DEPLOYMENT_ID|0": {
            "prefixes_v4": [
                "10.20.0.0/16",
                "20.20.0.0/16"
            ],
            "prefixes_v6": [
                "fc02:20::/64",
                "fc02:30::/64"
            ]
        }
    },

admin@vlab-01:~/hijack$ show run bgp | grep prefix
ip prefix-list PL_ALLOW_LIST_DEPLOYMENT_ID_0_COMMUNITY_empty_V4 seq 10 deny 0.0.0.0/0 le 17
ip prefix-list PL_ALLOW_LIST_DEPLOYMENT_ID_0_COMMUNITY_empty_V4 seq 20 permit 127.0.0.1/32
ip prefix-list PL_ALLOW_LIST_DEPLOYMENT_ID_0_COMMUNITY_empty_V4 seq 30 permit 10.20.0.0/16 le 32  <======
ip prefix-list PL_ALLOW_LIST_DEPLOYMENT_ID_0_COMMUNITY_empty_V4 seq 40 permit 20.20.0.0/16 le 32  <======
ip prefix-list PL_LoopbackV4 seq 5 permit 10.1.0.32/32
[...]

PR #10876, opened 2022-05-19
cherry pick [bgpcfgd] to support removal part of configuration of bgp allowed pre…

StormLiangMS:
cherry pick from master to 202012

Why I did it
Failed to remove part of configuration of bgp allowed prefix list. The details in #10141

How I did it
There are two issues:

In FRR, ipv6 default route is ::/0, but in the configuration, it is 0::/0, string comparison would be false, but why ipv4 failed to remove the allowed prefix list, ipv6 works? Looks into next one for the answer.

The current managers_allow_list doesn’t support removal part of the prefix list. But why IPv6 works in 1? It is because the bug for the IPv6 default route comparison, it would do the update no matter what is the operation (the code will compare the prefix list in the FRR and configuration db, if all configurations in db are presented in FRR, it do nothing, otherwise it will update the prefix list based on the configuration from db).

How to verify it
Follow the step in #10141

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111


wen587 (approved):

StormLiangMS:
already cherry-picked into 202012, this is a redundant PR, close.

PR #10877, opened 2022-05-19
Define `SYSTEM_DEFAULTS` table to control tunnel_qos_remap

bingwang-ms:
Signed-off-by: bingwang <wang.bing@microsoft.com>

#### Why I did it

This PR is to define a new table `SYSTEM_DEFAULTS`  to turn on/off new features in SONiC.
Currently, only the flag to turn on/off `tunnel_qos_remap` is in the new table.
HLD https://github.com/sonic-net/SONiC/pull/982

#### How I did it
1. Define a new section in `minigraph.xml`

```
<SystemDefaultsDeclaration>
    <a:SystemDefaults xmlns:a="http://schemas.datacontract.org/2004/07/Microsoft.Search.Autopilot.Evolution">
		 <a:SystemDefault>
            <a:Name>TunnelQosRemapEnabled</a:Name>
            <a:Value>True</a:Value>
         </a:SystemDefault>
    </a:SystemDefaults>
</SystemDefaultsDeclaration>
```
2. Update `minigrapg.py` to parse the new section, and write into `config_db`.
3. The Tunnel Qos remapping feature will be turn off if `TunnelQosRemapEnabled=False`.  Otherwise `minigraph.py` will try to parse new fields in `minigraph.xml`, and write the parsed value into `config_db`.
The extra values for tunnel qos remapping are as below.
```
 "src_ip": "25.1.1.10",
 "decap_dscp_to_tc_map": "AZURE_TUNNEL",
 "decap_tc_to_pg_map": "AZURE_TUNNEL",
 "encap_tc_to_dscp_map": "AZURE_TUNNEL",
 "encap_tc_to_queue_map": "AZURE_TUNNEL"
```
Changes for the templates will be raised in another PR.

#### How to verify it
Verified by adding new UT cases.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

Define `SYSTEM_DEFAULTS` table to control tunnel_qos_remap.


lguohan (dismissed):

src/sonic-config-engine/minigraph.py
```diff
@@ -1714,7 +1765,14 @@ def get_tunnel_entries(tunnel_intfs, lo_intfs, hostname):
     for type, tunnel_dict in tunnel_intfs.items():
         for tunnel_key, tunnel_attr in tunnel_dict.items():
             tunnel_attr['dst_ip'] = lo_addr
+
+            if (tunnel_qos_remap.get('status') == 'enabled') and (mux_tunnel_name == tunnel_key) and (peer_switch_ip is not None):
+                tunnel_attr['src_ip'] = peer_switch_ip
```
prsunny, line 1770: Since the minigraph is anyways changing, can we have the `src_ip` to be coming from minigraph `TunnelInterface `instead of having the logic to determine from  `peer_switch_ip` ?

bingwang-ms: I was also thinking about that. The problem is, I'm afraid it's not easy to implement that logic in Metadata.
For example, for upper_tor, the peer_ip would be `10.1.0.33`, while for `lower_tor`, it would be `10.1.0.32`. That means we have to generate different minigraph for upper_tor/lower_tor. I don't konw if it's doable.

src/sonic-config-engine/minigraph.py
```diff
@@ -899,6 +920,26 @@ def parse_meta(meta, hname):
     return syslog_servers, dhcp_servers, dhcpv6_servers, ntp_servers, tacacs_servers, mgmt_routes, erspan_dst, deployment_id, region, cloudtype, resource_type, downstream_subrole, switch_id, switch_type, max_cores, kube_data
 
 
+def parse_system_defaults(meta):
+    system_default_values = {}
+
+    system_defaults = meta.find(str(QName(ns1, "SystemDefaults")))
+    
+    for system_default in system_defaults.findall(str(QName(ns1, "SystemDefault"))):
```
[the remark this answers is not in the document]
bingwang-ms: Thanks. Fixed.

lolyu (approved):
LGTM, thanks!

src/sonic-config-engine/minigraph.py
prsunny: IMO, this would be pretty simple to implement. And btw, the minigraph is indeed different for upper and lower tor. Approving this!

prsunny (approved):

neethajohn (approved):

PR #10901, opened 2022-05-23
[Mellanox] [pmon] Fix for PMON service not starting when restarting SWSS service after fast/warm reboot

shlomibitton:
#### Why I did it

Recent change to delay PMON service in case of fast/warm reboot introduce an issue when restarting only SWSS service after fast/warm reboot for Nvidia platform.
Since the timer is triggered only when the system boot, in a scenario when the system is after a fast/warm reboot and the user restart SWSS service, as part of syncd.sh script, PMON service will stop but the timer will not start again.

#### How I did it

On syncd.sh script, in case of fast/warm indication, check if pmon.timer is running.
If it is running it means we are at the first boot and continue normally.
If it is not running, meaning the service was restarted, start the timer to keep the system behavior consistent.

#### How to verify it

1. Run fast/warm reboot.
2. service swss restart.
3. Observe PMON service starting.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [x] 202111
- [x] 202205


liat-grozovik (dismissed):

liat-grozovik:
@shlomibitton could you please checkout the failures?

shlomibitton:
> @shlomibitton could you please checkout the failures?

@liat-grozovik looks like environment issue, I retriggered.

shlomibitton:
@liat-grozovik all checkers are now passing

liat-grozovik (approved):

PR #10902, opened 2022-05-23
[202012] [Mellanox] [pmon] Fix for PMON service not starting when restarting SWSS service after fast/warm reboot

shlomibitton:
#### Why I did it

Recent change to delay PMON service in case of fast/warm reboot introduce an issue when restarting only SWSS service after fast/warm reboot for Nvidia platform.
Since the timer is triggered only when the system boot, in a scenario when the system is after a fast/warm reboot and the user restart SWSS service, as part of syncd.sh script, PMON service will stop but the timer will not start again.

#### How I did it

On syncd.sh script, in case of fast/warm indication, check if pmon.timer is running.
If it is running it means we are at the first boot and continue normally.
If it is not running, meaning the service was restarted, start the timer to keep the system behavior consistent.

#### How to verify it

1. Run fast/warm reboot.
2. service swss restart.
3. Observe PMON service starting.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111


liat-grozovik:
@stepanblyschak could you please review?

files/scripts/syncd.sh
```diff
@@ -40,7 +40,12 @@ function waitplatform() {
     BOOT_TYPE=`getBootType`
     if [[ x"$sonic_asic_platform" == x"mellanox" ]]; then
         if [[ x"$BOOT_TYPE" = @(x"fast"|x"warm"|x"fastfast") ]]; then
-            debug "PMON service is delayed by a timer for better fast/warm boot performance"
+            PMON_TIMER_STATUS=$(systemctl is-active pmon.timer)
+            if [[ "$PMON_TIMER_STATUS" = "inactive" ]]; then
```
lguohan, line 44: should be consistent style as line 41.

shlomibitton: Done.

liat-grozovik (approved):

liat-grozovik:
@qiluo-msft FYI, this need to get into 202012 with high prio.

qiluo-msft (approved):

PR #10910, opened 2022-05-24
[202012][openssh] openssh: Upgrade from 7.9 to 8.4, to match version in buster-backports

Blueve:
Signed-off-by: Jing Kan <jika@microsoft.com>

#### Why I did it

Upgrade 202012 sshd to 1:8.4p1-2 for security concern.

#### How I did it

Bullseye has already upgrade to use 1:8.4p1-5 and well tested.

This PR is for buster base image, we want to upgrade our openssh to buster-backports stable version 1:8.4p1-2

#### How to verify it

Build the openssh-server deb package and manually install it to a DUT
- [x] Test ssh
- [x] Test console reverse ssh

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111


Blueve:
Hello @saiarcot895 Do you think we can upgrade the openssh server version in 202012?

Blueve:
The build failed due to incorrect dependencies version

saiarcot895:
It might be fine to upgrade to 8.4 on buster, but one thing that'll need to be verified is that anything using SSH is still able to connect to this new version. OpenSSH may disable/remove support for less secure ciphers/encryption/signature methods, and while most clients will be fine, anything that is using very old versions of OpenSSH or using other implementations may have issues, because they may be trying to use those less secure ciphers/encryption/signature methods.

Blueve:
> It might be fine to upgrade to 8.4 on buster, but one thing that'll need to be verified is that anything using SSH is still able to connect to this new version. OpenSSH may disable/remove support for less secure ciphers/encryption/signature methods, and while most clients will be fine, anything that is using very old versions of OpenSSH or using other implementations may have issues, because they may be trying to use those less secure ciphers/encryption/signature methods.

Yes, that is a valid concern...

On the other hand, seems like some dependencies are not exists in buster-backports.
I downgrade to https://sources.debian.org/patches/openssh/1:8.4p1-2~bpo10+1/

xumia (approved):

PR #10913, opened 2022-05-24
[Arista] Update supervisor configurations

Staphylo:
The `use_pci_id_chassis` file is missing for the supervisor card, thus preventing proper init.
Some other configurations are no longer necessary and can safely be removed.


device/arista/x86_64-arista_7800_sup/Arista-7808R3A-FM/asic.conf
```diff
@@ -1,2 +0,0 @@
-NUM_ASIC=12
```
arlakshm, line 1: We need asic.conf  to see if the device is single or multi asic?
This PR https://github.com/Azure/sonic-buildimage/pull/9681 is just adding the pcie address to config_db. We still need `asic.conf`

Staphylo: The `asic.conf` in use by sonic is left untouched, see https://github.com/Azure/sonic-buildimage/blob/master/device/arista/x86_64-arista_7800_sup/asic.conf

The 3 `asic.conf` removed by this PR are hwsku specific ones that have never been handled by the master code.
We used these internally to enable testing on different chassis sizes and kind of fabric cards.

arlakshm (approved):

yxieca:
@arlakshm @Staphylo this cherry-pick yields an empty commit. Is that expected?

PR #10917, opened 2022-05-25
[macsec] Parse masec_enabled and macsec_profile from minigraph

judyjoseph:
#### Why I did it
In the Minigraph the following details are present for macsec configuration
```
  <LinkMetadataDeclaration>
    <Link xmlns:a="http://schemas.datacontract.org/2004/07/Microsoft.Search.Autopilot.Evolution">
      <a:LinkMetadata>
        <a:Name i:nil="true"/>
        <a:Properties>
          <a:DeviceProperty>
            <a:Name>MacSecEnabled</a:Name>
            <a:Value>True</a:Value>
          </a:DeviceProperty>
        </a:Properties>
        <a:Key>linecard-1:Ethernet1/1;ARISTA01-RH:Ethernet1/1</a:Key>
      </a:LinkMetadata>
    </Link>
    <Properties xmlns:a="http://schemas.datacontract.org/2004/07/Microsoft.Search.Autopilot.Evolution"/>
  </LinkMetadataDeclaration>

<a:DeviceProperty>
  <a:Name>MacSecProfile</a:Name>
  <a:Value>PrimaryKey="macsec-profile" FallbackKey="macsec-profile2" MacsecPolicy=""</a:Value>
```
Added the following
1. Support to parse the macsec enabled flag per interface
2. extract the primary macsec profile name in minigraph and attach it to interfaces where macsecEnabled flag is set.
3. The PrimaryKey in the MacSecProfile DeviceProperty, is the macsec profile used by default.
4. The FallbackKey in the MacSecProfile DeviceProperty, is the the macsec profile that will be used during key rotation. This is a manual process for now. 

The macsec_profile eg: as below will be pushed into the device as a json file and loaded into the config DB's using "config load command"

```
1. cat macsec_profile.json

      {
          "MACSEC_PROFILE": {
              "macsec-profile": {
                  "cipher_suite": "GCM-AES-XPN-256",
                  "primary_cak": "cak",
                  "primary_ckn": "ckn",
                  "priority": "0",
                  "rekey_period": "60"
              },
              "macsec-profile2": {
                  "cipher_suite": "GCM-AES-XPN-256",
                  "primary_cak": "cak",  
                  "primary_ckn": "ckn",
                  "priority": "0",
                  "rekey_period": "60"
              }
          }
      }

2. config load macsec_profile.json

```

#### How I did it
Added parsing logic in minigraph.py

#### How to verify it
Verified by generating config_db.json in a multi-asic device.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111


src/sonic-config-engine/minigraph.py
```diff
@@ -1824,6 +1837,30 @@ def parse_asic_meta_get_devices(root):
 
     return local_devices
 
+def parse_meta_get_macsec_profile(root, hname):
```
arlakshm, line 1840: Can we add parsing of `macsec-profile` in this function  https://github.com/Azure/sonic-buildimage/blob/ab0930313b34aa37819b4a5a325ed8b2f633e028/src/sonic-config-engine/minigraph.py#L873?

arlakshm (commented):
Please add unit tests for these ?

judyjoseph:
> Please add unit tests for these ?

Added unit tests, but there are some validation failures which needs to be further discussed. 

Do we need macsec-profile yang checks enforced here, as we are not populating macsec_profile table contents from minigraph.

```
    Validating yang schema
sonic_yang(6):Note: Below table(s) have no YANG models: PEER_SWITCH, VOQ_INBAND_INTERFACE, SYSTEM_PORT, CONSOLE_PORT, TUNNEL, MUX_CABLE, DEVICE_NEIGHBOR_METADATA, DHCP_SERVER, TELEMETRY, RESTAPI, CONSOLE_SWITCH
sonic_yang(3):Data Loading Failed:Leafref "/sonic-macsec:sonic-macsec/sonic-macsec:MACSEC_PROFILE/sonic-macsec:MACSEC_PROFILE_LIST/sonic-macsec:name" of value "macsec-profile" points to a non-existing leaf.
yang data generated from /sonic/src/sonic-config-engine/tests/sample-voq-graph.xml is not valid: Data Loading Failed
Leafref "/sonic-macsec:sonic-macsec/sonic-macsec:MACSEC_PROFILE/sonic-macsec:MACSEC_PROFILE_LIST/sonic-macsec:name" of value "macsec-profile" points to a non-existing leaf.
```

judyjoseph:
> > Please add unit tests for these ?
> 
> Added unit tests, but there are some validation failures which needs to be further discussed.
> 
> Do we need macsec-profile yang checks enforced here, as we are not populating macsec_profile table contents from minigraph.
> 
> ```
>     Validating yang schema
> sonic_yang(6):Note: Below table(s) have no YANG models: PEER_SWITCH, VOQ_INBAND_INTERFACE, SYSTEM_PORT, CONSOLE_PORT, TUNNEL, MUX_CABLE, DEVICE_NEIGHBOR_METADATA, DHCP_SERVER, TELEMETRY, RESTAPI, CONSOLE_SWITCH
> sonic_yang(3):Data Loading Failed:Leafref "/sonic-macsec:sonic-macsec/sonic-macsec:MACSEC_PROFILE/sonic-macsec:MACSEC_PROFILE_LIST/sonic-macsec:name" of value "macsec-profile" points to a non-existing leaf.
> yang data generated from /sonic/src/sonic-config-engine/tests/sample-voq-graph.xml is not valid: Data Loading Failed
> Leafref "/sonic-macsec:sonic-macsec/sonic-macsec:MACSEC_PROFILE/sonic-macsec:MACSEC_PROFILE_LIST/sonic-macsec:name" of value "macsec-profile" points to a non-existing leaf.
> ```

This issue is fixed by updating the testcases. It is passing now.

arlakshm (approved):

src/sonic-config-engine/minigraph.py
```diff
@@ -991,6 +995,8 @@ def parse_linkmeta(meta, hname):
                 lower_tor_hostname = value
             elif name == "AutoNegotiation":
                 auto_negotiation = value
+            elif name == "MacSecEnabled":
+                macsec_enabled = value
```
qiluo-msft, line 999: >macsec_enabled [](http://example.com/codeflow?start=16&length=14)

An explicit `False` value will override if `macsec_enabled = True` in previous iteration. I think this is not expected.

src/sonic-config-engine/minigraph.py
```diff
@@ -1002,14 +1008,28 @@ def parse_linkmeta(meta, hname):
                 linkmetas[port]["PeerSwitch"] = upper_tor_hostname
         if auto_negotiation:
             linkmetas[port]["AutoNegotiation"] = auto_negotiation
+        if macsec_enabled:
+            linkmetas[port]["MacSecEnabled"] = macsec_enabled
     return linkmetas
 
+def parse_macsec_profile(val_string):
+    macsec_profile = {}
+    values = val_string.strip().split(' ')
```
qiluo-msft, line 1017: >val_string [](http://example.com/codeflow?start=13&length=10)

The parsing logic is too strict and not easy to read/extend. Could you explore libraries like [https://docs.python.org/3/library/configparser.html](https://docs.python.org/3/library/configparser.html)?

src/sonic-config-engine/tests/common_utils.py
```diff
@@ -73,6 +73,9 @@ def validate(self, argument):
                 cmd += ' -p ' + args.port_config
             if args.namespace is not None:
                 cmd += ' -n ' + args.namespace
+            if "-j " in argument:
```
qiluo-msft, line 76: > if "-j " in argument: [](http://example.com/codeflow?start=11&length=22)

Could you check `args`?

src/sonic-config-engine/tests/common_utils.py
```diff
@@ -73,6 +73,9 @@ def validate(self, argument):
...
+                strs = argument.split(' ')
+                cmd += " -j {}".format(next(strs[i+1] for i in range(len(strs)) if strs[i]=='-j'))
```
qiluo-msft, line 78: >{} [](http://example.com/codeflow?start=28&length=2)

Could you use `args.json`?

src/sonic-config-engine/minigraph.py
judyjoseph: My understanding is as below for each interface there is a LinkMeta . It is initialized False and looped per property for that port. 

```
            <a:LinkMetadata>
                <a:Name i:nil="true"/>
                <a:Properties>
                    <a:DeviceProperty>
                        <a:Name>MacSecEnabled</a:Name>
                        <a:Reference i:nil="true"/>
                        <a:Value>True</a:Value>
                    </a:DeviceProperty>
                </a:Properties>
                <a:Key>STG01-0101-0400-01T2-lc3:Eth48-ASIC0;STG30-0100-0100-02RHE:Ethernet4/7/1</a:Key>
            </a:LinkMetadata>
            <a:LinkMetadata>
                <a:Name i:nil="true"/>
                <a:Properties>
                    <a:DeviceProperty>
                        <a:Name>MacSecEnabled</a:Name>
                        <a:Reference i:nil="true"/>
                        <a:Value>True</a:Value>
                    </a:DeviceProperty>
                </a:Properties>
                <a:Key>STG01-0101-0400-01T2-lc3:Eth56-ASIC0;STG30-0100-0100-02RHE:Ethernet4/8/1</a:Key>
            </a:LinkMetadata>
```

src/sonic-config-engine/tests/common_utils.py
judyjoseph: Updated to use args itself.

src/sonic-config-engine/tests/common_utils.py
judyjoseph: Updated to use args itself.

src/sonic-config-engine/minigraph.py
judyjoseph: @Qi I am finding this config parser library used when we have to read from an input file and while parsing it has restrictions like need to have [SECTION] header etc - which I feel is not applicable here.

I checked the following, and the current parsing should take care of them.
> the macsec profile name cannot be with spaces
> the profile will have PrimaryKey and FallbackKey.

judyjoseph:
> @judyjoseph, LGTM. We will add tests for `packet-chassis` in another PR ?

yes, I will raise a separate PR for packet-chassis as the sample xml/portconfig.ini needs to be modified a bit which I will confirm, thanks.

ganglyu (approved):

PR #10581, opened 2022-04-14
[portconfig]: Remove try block for db config initialization

SuvarnaMeenakshi:
#### Why I did it
1. Provide fix for comment: https://github.com/Azure/sonic-buildimage/pull/10475/files#r847753187; 
Follow up issue to be done in a different PR:
 https://github.com/Azure/sonic-buildimage/pull/10362/ - modified 2 test cases where the test case is required to get PORT table from config_db, the PR modified to use port_config.ini instead. 
#### How I did it
1. Try exception is not required in this scenario, so remove and modify to initial db config according to single or multi-asic platforms.

#### How to verify it
Verified on multi-asic device.
#### Which release branch to backport (provide reason below if selected)

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

PR #10475, opened 2022-04-06
[202012][multi-asic][sonic-config-engine]: Get PORT table from namespace config db

SuvarnaMeenakshi:
Why I did it
Cherry-pick of: https://github.com/Azure/sonic-buildimage/pull/7632
portconfig.py gets PORT table from config_db if it is present. If not, port_config.ini files are parsed.
For multi-asic platform, if namespace is passed to get_port_config(), config_db connection was done to host namespace always and not to asic specific namespace.
Provides fix for: https://github.com/Azure/sonic-buildimage/issues/7161

How I did it
Modify db connection function to connect to namespace config_db.

#### How to verify it
Unit-test passed.
Verified on multi-asic VS platform.
#### Which release branch to backport (provide reason below if selected)

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

PR #10960, opened 2022-05-27
[portconfig]: Remove try exception during config_db initialization.

SuvarnaMeenakshi:
loading database config already handles exception so this is not
required in portconfig. Move laoding database config to application
code instead of portconfig as portconfig is used as a library.

#### Why I did it
Provide fix for comment: https://github.com/Azure/sonic-buildimage/pull/10475/files#r847753187;
Move laoding database config to application code instead of portconfig as portconfig is used as a library.
https://github.com/Azure/sonic-buildimage/pull/10581 was raised for this fix, but had to be reverted due to issue with multi-asic platform. 

#### How I did it
Remove try exception handing from portconfig.py during config_db intialization.
Move loading of database config to application that uses portconfig.py.

#### How to verify it
unit-test passes.
Verified that it does not cause issue during boot up of multi-asic VS image.
Verified that config_db generation was successful in multi-asic VS.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111


qiluo-msft (approved):

src/sonic-config-engine/sonic-cfggen
```diff
@@ -138,6 +138,13 @@ def ip_network(value):
         return "Invalid ip address %s" % value
     return r_v.network
 
+def load_namespace_config(asic_name):
+    if not SonicDBConfig.isInit():
+        if is_multi_asic():
+            SonicDBConfig.load_sonic_global_db_config(namespace=asic_name)
```
arlakshm, line 144: when we are parsing the host or device configuration on multi asic platform `asic_name` will be None.  In that case will `SonicDBConfig.load_sonic_global_db_config(namespace=None)` work ?

arlakshm (approved):

SuvarnaMeenakshi:
> initializeGlobalConfig

This error message is seen on an image without this change as well. Checked on 4-asic VS testbed "master.101959-392899682" image. Still shows the same error:
Jun  1 20:58:36.426814 vlab-08 ERR python3: :- initializeGlobalConfig: SonicDBConfig Global config is already initialized

SuvarnaMeenakshi:
> > initializeGlobalConfig
> 
> This error message is seen on an image without this change as well. Checked on 4-asic VS testbed "master.101959-392899682" image. Still shows the same error: Jun 1 20:58:36.426814 vlab-08 ERR python3: :- initializeGlobalConfig: SonicDBConfig Global config is already initialized

Confirmed that this error message is seen on image without this change on multi-asic VS.
ERR python3: :- initializeGlobalConfig: SonicDBConfig Global config is already initialized and - not sure which script is throwing this error message.
ERR caclmgrd[19224]: :- initializeGlobalConfig: SonicDBConfig Global config is already initialized

qiluo-msft:
Need backport to other branches?

gechiang:
Please note that "https://github.com/sonic-net/sonic-buildimage/pull/18609" was created for 202205 for this fix which included two additional changes not included in this PR.  I have requested @deepak-singhal0408 to raise another PR to inlcude those two additional changes in the master branch as well to ensure that the fix is also complete in master.

PR #10986, opened 2022-05-31
[sonic-feature.yang] fix check_up_status field type

stepanblyschak:

#### Why I did it

An issue is encountered when a value "False" is written for a feature in "check_up_status" field, which does not pass YANG validation.

#### How I did it

 We usually use stypes::boolean_type for such fields, even in this YANG model. This custom type, supports "False" value.

#### How to verify it

Write "False" in "check_up_status" field and see if YANG validation passes.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111


src/sonic-yang-models/yang-models/sonic-feature.yang
```diff
@@ -90,8 +90,8 @@ module sonic-feature{
                 leaf check_up_status {
                     description "This configuration controls the system ready tool to check
                         the app ready/up status";
-                    type boolean;
-                    default false;
+                    type stypes:boolean_type;
+                    default "false";
```
qiluo-msft, line 94: Add unit test?

stepanblyschak: @qiluo-msft Without this fix sonic-utilities submodule update is blocked. There were no UT when this field was added. I suggest to take this fix to unblock submodule update and the owner of new feature adds UT.

qiluo-msft: @wen587 for awareness of blocking issue of sonic-utilities submodule update

qiluo-msft (approved):

liat-grozovik:
@stepanblyschak please chck if this is needed for 202205 and if so update PR desciption/label

PR #10992, opened 2022-05-31
[ssip]: Add YANG model

nazariig:
**HLD:** https://github.com/sonic-net/SONiC/pull/1002

#### Why I did it
* To implement Syslog Source IP feature

#### How I did it
* Added the relevant yang model 

#### How to verify it
* Added unit test

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

* N/A

* N/A

```
      .---.        .-----------
     /     \  __  /    ------
    / /     \(  )/    -----
   //////   ' \/ `   ---
  //// / // :    : ---
 // /   /  /`    '--
//          //..\\
       ====UU====UU====
           '//||\\`
             ''``
```

### DIFF
--- a/src/sonic-yang-models/tests/yang_model_tests/tests_config/syslog.json
+++ b/src/sonic-yang-models/tests/yang_model_tests/tests_config/syslog.json
@@ -24,6 +24,13 @@
                 ]
             }
         },
+        "sonic-mgmt_vrf:sonic-mgmt_vrf": {
+            "sonic-mgmt_vrf:MGMT_VRF_CONFIG": {
+                "sonic-mgmt_vrf:vrf_global": {
+                    "mgmtVrfEnabled": "true"
+                }
+            }
+        },
         "sonic-vrf:sonic-vrf": {
             "sonic-vrf:VRF": {
                 "VRF_LIST": [
@@ -41,6 +48,18 @@
                     {
                         "server_address": "100.0.0.1",
                         "source": "invalid_value"
+                    }
+                ]
+            }
+        }
+    },
+    "SYSLOG_SERVER_SOURCE_IP_FAMILY_MISMATCH": {
+        "sonic-syslog:sonic-syslog": {
+            "sonic-syslog:SYSLOG_SERVER": {
+                "SYSLOG_SERVER_LIST": [
+                    {
+                        "server_address": "100.0.0.1",
+                        "source": "1111::1111"
                     }
                 ]
             }
@@ -67,6 +86,25 @@
                         "vrf": "invalid_value"
                     }
                 ]
+            }
+        }
+    },
+    "SYSLOG_SERVER_DISABLED_MGMT_VRF": {
+        "sonic-syslog:sonic-syslog": {
+            "sonic-syslog:SYSLOG_SERVER": {
+                "SYSLOG_SERVER_LIST": [
+                    {
+                        "server_address": "100.0.0.1",
+                        "vrf": "mgmt"
+                    }
+                ]
+            }
+        },
+        "sonic-mgmt_vrf:sonic-mgmt_vrf": {
+            "sonic-mgmt_vrf:MGMT_VRF_CONFIG": {
+                "sonic-mgmt_vrf:vrf_global": {
+                    "mgmtVrfEnabled": "false"
+                }
             }
         }
     },
--- a/src/sonic-yang-models/tests/yang_model_tests/tests/syslog.json
+++ b/src/sonic-yang-models/tests/yang_model_tests/tests/syslog.json
@@ -6,6 +6,10 @@
         "desc": "Configure invalid SOURCE in SYSLOG_SERVER.",
         "eStrKey": "InvalidValue"
     },
+    "SYSLOG_SERVER_SOURCE_IP_FAMILY_MISMATCH": {
+        "desc": "Configure SOURCE IP family mismatch in SYSLOG_SERVER.",
+        "eStrKey": "Must"
+    },
     "SYSLOG_SERVER_INVALID_PORT": {
         "desc": "Configure invalid PORT in SYSLOG_SERVER.",
         "eStrKey": "InvalidValue"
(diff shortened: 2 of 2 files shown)


src/sonic-yang-models/yang-models/sonic-syslog.yang
```diff
@@ -9,22 +9,64 @@ module sonic-syslog {
...
                 leaf server_address {
-                    description "syslog server ip address";
+                    description "Syslog server IP address";
                     type inet:ip-address;
                 }
-            } /* end of list SYSLOG_SERVER_LIST */ 
 
-        } /* end of container SYSLOG_SERVER */
+                leaf source {
+                    description "Syslog source IP address";
```
venkatmahalingam, line 46: What happens when the source IP is not configured within a VRF (say. mgmt or data VRF)? rsyslog takes available IP from the kernel and send the log messages?

src/sonic-yang-models/tests/yang_model_tests/tests_config/syslog.json
```diff
@@ -1,4 +1,75 @@
...
+                        "server_address": "100.0.0.1",
+                        "source": "1.1.1.1",
+                        "port": "514",
+                        "vrf": "default"
+                    },
+                    {
+                        "server_address": "200.0.0.1",
+                        "source": "1.1.1.1",
+                        "port": "514",
+                        "vrf": "mgmt"
```
venkatmahalingam, line 16: Suggest to have a validation for mgmt VRF ' (/mvrf:sonic-mgmt_vrf/mvrf:MGMT_VRF_CONFIG/mvrf:vrf_global/mvrf:mgmtVrfEnabled = 'true')' as well.

src/sonic-yang-models/yang-models/sonic-syslog.yang
```diff
@@ -9,22 +9,64 @@ module sonic-syslog {
...
 
                 leaf server_address {
-                    description "syslog server ip address";
+                    description "Syslog server IP address";
                     type inet:ip-address;
                 }
-            } /* end of list SYSLOG_SERVER_LIST */ 
 
-        } /* end of container SYSLOG_SERVER */
+                leaf source {
```
dgsudharsan, line 45: Should the source ip family type and server_address ip family type match?

dgsudharsan (changes requested):
Can you please update https://github.com/Azure/sonic-buildimage/blob/master/src/sonic-yang-models/doc/Configuration.md as well?

ganglyu:
would you please add the link for HLD?

src/sonic-yang-models/tests/yang_model_tests/tests_config/syslog.json
nazariig: @venkatmahalingam not possible since it goes against the feature design: with the approach suggested won't be able to use a definition for config generator (template renderer)

src/sonic-yang-models/yang-models/sonic-syslog.yang
nazariig: @dgsudharsan for sure. The CLI has this kind of check

dgsudharsan: Then we need to add this in yang as well

src/sonic-yang-models/yang-models/sonic-syslog.yang
venkatmahalingam: SONIC YANG model will be used for validating the configs present in json file for config load/reload, so YANG validation is important.

src/sonic-yang-models/tests/yang_model_tests/tests_config/syslog.json
venkatmahalingam: YANG should do all possible validations (including custom validations if required), delegating any validation to CLI is not a good idea.

src/sonic-yang-models/yang-models/sonic-syslog.yang
nazariig: @venkatmahalingam handled

src/sonic-yang-models/tests/yang_model_tests/tests_config/syslog.json
nazariig: @venkatmahalingam done

src/sonic-yang-models/yang-models/sonic-syslog.yang
nazariig: @dgsudharsan done

src/sonic-yang-models/yang-models/sonic-syslog.yang
nazariig: @venkatmahalingam such kind of validation is not possible with the current DB schema and YANG approach since we don't reference a configured IPs. Also not sure if it's even possible to add such a constraint in current YANG model to validate whether interface IP is configured across the specific VRF. This brings a lot of complexity. Any suggestions are appreciated.

nazariig:
> Can you please update https://github.com/Azure/sonic-buildimage/blob/master/src/sonic-yang-models/doc/Configuration.md as well?

@dgsudharsan will do

dgsudharsan (approved):

src/sonic-yang-models/yang-models/sonic-syslog.yang
```diff
@@ -9,22 +9,72 @@ module sonic-syslog {
...
                 leaf server_address {
-                    description "syslog server ip address";
+                    description "Syslog server IP address";
+                    type inet:ip-address;
+                }
+
+                leaf source {
+                    description "Syslog source IP address";
                     type inet:ip-address;
+                    must "(contains(current(), '.') and contains(../server_address, '.'))
```
qiluo-msft, line 52: >. [](http://example.com/codeflow?start=48&length=1)

ipv6 address could also container `.`. Please check [https://en.wikipedia.org/w/index.php?title=IPv6_address&section=19#Special_addresses](https://en.wikipedia.org/w/index.php?title=IPv6_address&section=19#Special_addresses)

nazariig: @qiluo-msft do we have such a use case in SONiC?

qiluo-msft (approved):

nazariig:
> Can you please update https://github.com/Azure/sonic-buildimage/blob/master/src/sonic-yang-models/doc/Configuration.md as well?

@dgsudharsan done: https://github.com/Azure/sonic-buildimage/pull/11363

PR #11000, opened 2022-06-01
[swss] Adding bgp container as dependent of swss

Ndancejic:
What I did:
Added bgp as a dependent of swss

Why I did it:
bgp container was not restarting on swss crash. When swss crashes, linkmgrd
doesn't initate a switchover because it cannot access the default route from
orchagent. Bringing down bgp with swss will isolate the ToR, causing linkmgrd
to initiate a switchover to the peer ToR avoiding significant packet loss.

How I did it:
Added bgp to DEPENDENT

Signed-off-by: Nikola Dancejic <ndancejic@microsoft.com>

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

### DIFF
--- a/files/scripts/swss.sh
+++ b/files/scripts/swss.sh
@@ -9,7 +9,7 @@ LOCKFILE="/tmp/swss-syncd-lock$DEV"
 NAMESPACE_PREFIX="asic"
 ETC_SONIC_PATH="/etc/sonic/"
 
-DEPENDENT="radv"
+DEPENDENT="radv bgp"
 MULTI_INST_DEPENDENT="teamd"
 
 . /usr/local/bin/asic_status.sh


lguohan:
can you have more description, why we want to bgp restart when swss crash?

prsunny (approved):

Ndancejic:
rebased changes onto latest master

SuvarnaMeenakshi:
@Ndancejic - bgp is a per-asic service, so it will run for each namespace on multi-asic platform, similar to swss/syncd/teamd/lldp.
So, bgp should be added in multi-asic dependent service list along with teamd.
MULTI_INST_DEPENDENT="teamd"
More details on the issue: https://github.com/sonic-net/sonic-buildimage/issues/11653

prsunny:
> @Ndancejic - bgp is a per-asic service, so it will run for each namespace on multi-asic platform, similar to swss/syncd/teamd/lldp. So, bgp should be added in multi-asic dependent service list along with teamd. MULTI_INST_DEPENDENT="teamd" More details on the issue: #11653

@SuvarnaMeenakshi , the concern is by having 'bgp' in MULTI_INST, restarting bgp is also restarting swss whereas ideally we ONLY want the 'vice-versa' behavior. This will be a behavior change from 201911. We don't want swss to be restarted and impacting dataplane when bgp restarts. Infact this is the case today if we restart teamd on a single ASIC devices. @Ndancejic shall raise a github issue for the same. We have to get that reverse dependency fixed before having anything in MULTI_INST.

SuvarnaMeenakshi:
> > @Ndancejic - bgp is a per-asic service, so it will run for each namespace on multi-asic platform, similar to swss/syncd/teamd/lldp. So, bgp should be added in multi-asic dependent service list along with teamd. MULTI_INST_DEPENDENT="teamd" More details on the issue: #11653
> 
> @SuvarnaMeenakshi , the concern is by having 'bgp' in MULTI_INST, restarting bgp is also restarting swss whereas ideally we ONLY want the 'vice-versa' behavior. This will be a behavior change from 201911. We don't want swss to be restarted and impacting dataplane when bgp restarts. Infact this is the case today if we restart teamd on a single ASIC devices. @Ndancejic shall raise a github issue for the same. We have to get that reverse dependency fixed before having anything in MULTI_INST.

Sure, but that will be the case for single asic as well right? restarting bgp in single asic should cause restart of swss with this change? Can we revert this PR until a clean fix is available for single and multi-asic? This is stopping us from enabling sanity test in multi-asic vs testing which is to be enabled for PR checker, also this is in 202205 which will impact Chassis as well.

prsunny:
With this change restarting bgp is *not* causing swss restart. But if we put this in multi_inst, it is causing swss restart. Ok, lets discuss offline on the next steps and understand a bit more.

wenyiz2021:
> > > @Ndancejic - bgp is a per-asic service, so it will run for each namespace on multi-asic platform, similar to swss/syncd/teamd/lldp. So, bgp should be added in multi-asic dependent service list along with teamd. MULTI_INST_DEPENDENT="teamd" More details on the issue: #11653
> > 
> > 
> > @SuvarnaMeenakshi , the concern is by having 'bgp' in MULTI_INST, restarting bgp is also restarting swss whereas ideally we ONLY want the 'vice-versa' behavior. This will be a behavior change from 201911. We don't want swss to be restarted and impacting dataplane when bgp restarts. Infact this is the case today if we restart teamd on a single ASIC devices. @Ndancejic shall raise a github issue for the same. We have to get that reverse dependency fixed before having anything in MULTI_INST.
> 
> Sure, but that will be the case for single asic as well right? restarting bgp in single asic should cause restart of swss with this change? Can we revert this PR until a clean fix is available for single and multi-asic? This is stopping us from enabling sanity test in multi-asic vs testing which is to be enabled for PR checker, also this is in 202205 which will impact Chassis as well.

just to point out PR   https://github.com/sonic-net/sonic-mgmt/pull/6083 is dependent on this.. if any revert or change to be made for this fix

wenyiz2021:
> > @Ndancejic - bgp is a per-asic service, so it will run for each namespace on multi-asic platform, similar to swss/syncd/teamd/lldp. So, bgp should be added in multi-asic dependent service list along with teamd. MULTI_INST_DEPENDENT="teamd" More details on the issue: #11653
> 
> @SuvarnaMeenakshi , the concern is by having 'bgp' in MULTI_INST, restarting bgp is also restarting swss whereas ideally we ONLY want the 'vice-versa' behavior. This will be a behavior change from 201911. We don't want swss to be restarted and impacting dataplane when bgp restarts. Infact this is the case today if we restart teamd on a single ASIC devices. @Ndancejic shall raise a github issue for the same. We have to get that reverse dependency fixed before having anything in MULTI_INST.

my understanding is current fix still should be reverted as bgp on duthost is starting which doesn't work for multi-asic, right? but before having the change in multi_inst, need to reverse the dependency of swss restart if bgp restart.. from 201911, 202012,... master branch?

Ndancejic:
Tested out single ASIC with the following change seems to work for us: bgp goes down without also taking down swss. can you confirm that this also works on multi-ASIC. I'll open a PR in the meantime to run through PR tests:

```
diff --git a/files/scripts/swss.sh b/files/scripts/swss.sh
index 93a3c3dcf..e67faf75c 100755
--- a/files/scripts/swss.sh
+++ b/files/scripts/swss.sh
@@ -309,9 +309,11 @@ function check_peer_gbsyncd()
 if [ "$DEV" ]; then
     NET_NS="$NAMESPACE_PREFIX$DEV" #name of the network namespace
     SONIC_DB_CLI="sonic-db-cli -n $NET_NS"
+    DEPENDENT+=" bgp@${DEV}"
 else
     NET_NS=""
     SONIC_DB_CLI="sonic-db-cli"
+    DEPENDENT+=" bgp"
 fi
```

Ndancejic:
https://github.com/sonic-net/sonic-buildimage/pull/11691

PR #11012, opened 2022-06-02
Add IP interface loopback action related content to YANG models.

liorghub:
#### Why I did it
Add IP interface loopback action related content to the required YANG models.

#### How I did it
Update the relevant YANG models.

#### How to verify it
New unittests were added.

- [x] 202205

### DIFF
--- a/src/sonic-yang-models/yang-models/sonic-interface.yang
+++ b/src/sonic-yang-models/yang-models/sonic-interface.yang
@@ -77,9 +77,7 @@ module sonic-interface {
 
                 leaf loopback_action {
                     description "Packet action when a packet ingress and gets routed on the same IP interface";
-                    type string{
-                        pattern "drop|forward";
-                    }
+                    type stypes:loopback_action;
                 }
 			}
 			/* end of INTERFACE_LIST */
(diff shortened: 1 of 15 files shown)


liorghub:
The following test is failing:
ERROR dhcp_relay/test_dhcp_relay.py::test_dhcp_relay_after_link_flap

dgsudharsan (dismissed):

src/sonic-yang-models/yang-models/sonic-interface.yang
```diff
@@ -66,13 +66,19 @@ module sonic-interface {
...
 					description "Enable/disable MPLS routing for the interface";
 					type enumeration {
 						enum enable;
 						enum disable;
 					}
 				}
+
+                leaf loopback_action {
+                    description "Packet action when a packet ingress and gets routed on the same IP interface";
+                    type stypes:loopback_action;
```
prsunny, line 80: shouldn't we specify the default action?

dgsudharsan: @prsunny In general no. If we specify default action, this will be written to config_db. In case if the action is not implemented by other vendors there would be a SAI failure followed by orchagent crash.
So previously in yang subgroup we decided not to have default action for attributes that would be programmed into ASIC

prsunny (approved):

dgsudharsan (approved):

ganglyu (approved):

PR #11036, opened 2022-06-06
Fix ndppd occupy high CPU usage problem

kuanyu99:
#### Why I did it

In t1-topology, we frequently see the ndppd use a lot of CPU usage.

#### How I did it

Make the route-ttl setting correctly generate inside the /etc/ndppd.conf of swss container.

#### How to verify it

Run the test_cpu_memory_usage.py for multiple times and make sure the ndppd will not fail the test.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012
- [x] 202106
- [x] 202111

Based on the PR#7456, the route-ttl setting should not put inside the vlan interface if-else block

### DIFF
--- a/dockers/docker-orchagent/ndppd.conf.j2
+++ b/dockers/docker-orchagent/ndppd.conf.j2
@@ -6,6 +6,7 @@
 {% endblock banner %}
 # Config file for ndppd, the NDP Proxy Daemon
 # See man page for ndppd.conf.5 for descriptions of all available options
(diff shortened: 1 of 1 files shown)


theasianpianist:
Can we change this to not run on T1 devices altogether?

kuanyu99:
> Can we change this to not run on T1 devices altogether?

Do you mean if the ndppd can be an optional program during the device in a T1 topology?

PR #11063, opened 2022-06-07
[ci] Build docker-sonic-vs.gz in common lib pipeline to increase component pipeline success rate.

liushilongbuaa:
#### Why I did it
Build docker-sonic-vs.gz in common lib pipeline for other pipelines download.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

### DIFF
--- a/.azure-pipelines/template-commonlib.yml
+++ b/.azure-pipelines/template-commonlib.yml
@@ -9,7 +9,7 @@ parameters:
 jobs:
 - ${{ each arch in parameters.archs }}:
   - job: Build_${{ arch }}
-    timeoutInMinutes: 120
+    timeoutInMinutes: 240
     ${{ if eq(arch,'amd64') }}:
       pool: sonicbld
     ${{ else }}:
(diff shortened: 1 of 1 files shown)


.azure-pipelines/template-commonlib.yml
```diff
@@ -39,9 +39,10 @@ jobs:
       displayName: Make configure
     - script: |
         set -ex
+        grep amd64 .arch && extarget=target/docker-sonic-vs.gz
```
xumia, line 42: docker-sonic-vs.gz should be only in one of the BLDENV.

.azure-pipelines/template-commonlib.yml
```diff
@@ -39,9 +39,10 @@ jobs:
...
         for bldenv in $(bldenvs)
         do
-          SONIC_BUILD_JOBS=$(nproc) BLDENV=$bldenv make -f Makefile.work lib-packages ENABLE_DOCKER_BASE_PULL=y
```
xumia, line 44: Why remove the option ENABLE_DOCKER_BASE_PULL=y

liushilongbuaa: make configure step has already pulled docker base image.

.azure-pipelines/template-commonlib.yml
liushilongbuaa: If split docker-sonic-vs.gz into another pipeline, it is complex to refer it. We need to add another step to download.
If they are in the same pipeline, we only need to add patterns in old download step.

liushilongbuaa: We can see this test job: https://dev.azure.com/mssonic/build/_build/results?buildId=107333&view=results
Although it is in buster, it is ok to run in bullseye. It will do nothing.

PR #10519, opened 2022-04-09
Increase LAG Ids to 1024

ysmanman:
#### Why I did it
Today at most 128 LAGs are supported. This is not sufficient if there are many LAGs with just few ports.

#### How I did it
Increase LAG Ids to 1024 for DNX device.

#### Which release branch to backport (provide reason below if selected)

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

PR #11070, opened 2022-06-08
[Arista] Enable larger number of LAGs on 7800 LCs

kenneth-arista:
For 7800 LCs, set LAG mode to support 1024 number of 16-member system
LAGs.

#### Why I did it
The SOC property changes are necessary to match https://github.com/Azure/sonic-buildimage/pull/10519 which increases the number of system LAG IDs to 1024.

For 7800 LCs, set LAG mode to support 1024 number of 16-member system
LAGs.


kenneth-arista:
@arlakshm please review

arlakshm (approved):

PR #11075, opened 2022-06-08
[QoS] Support dynamic headroom calculation for Barefoot platforms

mmysle:
Signed-off-by: Michał Mysłek <michal.myslek@intel.com>

**What I did**
Adding the dynamic headroom calculation support for Barefoot platforms.

**Why I did it**
Enabling dynamic mode for barefoot case.

**How I verified it**
The community tests are adjusted and pass.

**Details if related**

### DIFF
--- a/device/barefoot/x86_64-accton_as9516_32d-r0/newport/buffers_defaults_t1.j2
+++ b/device/barefoot/x86_64-accton_as9516_32d-r0/newport/buffers_defaults_t1.j2
@@ -1,8 +1,8 @@
 {% set default_cable = '5m' %}
-{% set ingress_lossless_pool_size =  '23850816' %}
-{% set ingress_lossy_pool_size =  '36222208' %}
-{% set egress_lossless_pool_size =  '29482816' %}
-{% set egress_lossy_pool_size =  '26400000' %}
+{% set ingress_lossless_pool_size = '43067728' %}
+{% set ingress_lossy_pool_size = '3520000' %}
+{% set egress_lossless_pool_size = '46749824' %}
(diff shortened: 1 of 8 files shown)


device/barefoot/x86_64-accton_as9516_32d-r0/newport/buffers_defaults_t1.j2
```diff
@@ -14,13 +14,17 @@
 {%- macro generate_buffer_pool_and_profiles() %}
     "BUFFER_POOL": {
         "ingress_lossless_pool": {
+            {%- if dynamic_mode is not defined %}
             "size": "{{ ingress_lossless_pool_size }}",
+            {%- endif %}
             "type": "ingress",
             "mode": "dynamic",
-            "xoff": "36222208"
+            "xoff": "3153920"
```
stephenxs, line 22: Hi
From this PR I see the buffer pool sizes (`ingress_lossless_pool_size`, `ingress_lossy_pool_size`, and `egress_lossy_pool`) are dynamically calculated and the shared headroom pool size (`ingress_lossless_pool.xoff`) is a fixed number.
But in PR Azure/sonic-swss#2317 and Azure/sonic-swss#2323, I see the buffer pool sizes are fixed and the shared headroom pool size is dynamically calculated.
These result in both the buffer pool sizes and the shared headroom pool size are static. Is this a mistake or by intention?
If you are going to have the buffer pool sizes static and the shared headroom pool size dynamic, I would like to suggest
- always defining `size`s in all pools regardless of whether `dynamic_mode` is defined
- defining `xoff` in `ingress_lossless_pool` only if `dynamic_mode` is defined

PR #11082, opened 2022-06-09
[202012] [sonic-config-engine] Generate expected output with different cable len

neethajohn:
#### Why I did it
To address internal build failures where the cable len for some of the skus is set to 300m for all tiers. 

#### How I did it
For the buffers test, generate a new output file based off the original expected output with CABLE_LENGTH table updated to use 300m. In the comparison logic, compare against each of the expected output files and if any matches, the testcase is set to pass

#### How to verify it
Build sonic-config-engine wheel with the changes

### DIFF
--- a/src/sonic-config-engine/tests/test_j2files.py
+++ b/src/sonic-config-engine/tests/test_j2files.py
@@ -69,6 +69,20 @@ class TestJ2Files(TestCase):
         if not dir_exist:
             os.system('sudo rmdir /host')
 
+    def modify_cable_len(self, base_file, file_dir):
+        input_file = os.path.join(file_dir, base_file)
+        with open(input_file, 'r') as ifd:
+            object = json.load(ifd)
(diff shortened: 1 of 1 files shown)


src/sonic-config-engine/tests/test_j2files.py
```diff
@@ -355,8 +369,24 @@ def _test_buffers_render_template(self, vendor, platform, sku, minigraph, buffer
...
+            sample_output_file = os.path.join(out_file_dir, out_file)
+            if utils.cmp(sample_output_file, self.output_file):
+                match = True
+                break
+            else:
+                diff = diff + str(self.run_diff(sample_output_file, self.output_file))
+
+        os.remove(os.path.join(out_file_dir, expected_files[1]))
+
+        if match:
```
qiluo-msft, line 386: you can simplify if/else by `assert match, diff`

qiluo-msft:
Change target branch to master, and cherry-pick to 202012?

bingwang-ms (approved):

PR #11095, opened 2022-06-10
[minigraph] Support parse IPv6 in device_desc.xml

Blueve:
#### Why I did it

minigraph parser can parse initial configuration from a fragment of minigraph (device_desc.xml), however it can only get IPv4 address from it and due to some reason, if the device only has IPv6 management IP address, then the IPv4 address will be `0.0.0.0/0` which will cause `config load_mgmt_config` command fail

#### How I did it
1. Avoid put `0.0.0.0/0` to mgmt config
2. Put IPv6 mgmt IP information to mgmt config if possiable

#### How to verify it

- [x] UT

- [ ] 201811
- [x] 201911
- [x] 202006
- [x] 202012
- [x] 202106
- [x] 202111
- [x] 202205


src/sonic-config-engine/minigraph.py
```diff
@@ -1848,10 +1851,16 @@ def parse_device_desc_xml(filename):
...
-    mgmtipn = ipaddress.ip_network(UNICODE_TYPE(mgmt_prefix), False)
-    gwaddr = ipaddress.ip_address((next(mgmtipn.hosts())))
-    results['MGMT_INTERFACE'] = {('eth0', mgmt_prefix): {'gwaddr': gwaddr}}
+    results['MGMT_INTERFACE'] = {}
+    if mgmt_prefix and mgmt_prefix != '0.0.0.0/0':
+        mgmtipn = ipaddress.ip_network(UNICODE_TYPE(mgmt_prefix), False)
+        gwaddr = ipaddress.ip_address((next(mgmtipn.hosts())))
+        results['MGMT_INTERFACE'].update({('eth0', mgmt_prefix): {'gwaddr': gwaddr}})
+
+    if mgmt_prefix_v6 and mgmt_prefix_v6 != '::/0':
```
qiluo-msft, line 1860: >mgmt_prefix_v6 != '::/0' [](http://example.com/codeflow?start=26&length=24)

Do we really need to check with all zero IPv4/IPv6? If yes, compare it by ipaddress class, not by string.

Blueve: Yes, minigraph will provide default value (all zero addresses). I will update with ipn comparison

ganglyu (approved):

src/sonic-config-engine/tests/simple-sample-device-desc-ipv6-only.xml
```diff
@@ -0,0 +1,11 @@
+<Device i:type="ToRRouter" xmlns="Microsoft.Search.Autopilot.Evolution" xmlns:i="http://www.w3.org/2001/XMLSchema-instance">
+  <Hostname>switch-t0</Hostname>
+  <HwSku>Force10-S6000</HwSku>
+  <ClusterName>AAA00PrdStr00</ClusterName>
+  <ManagementAddress xmlns:a="Microsoft.Search.Autopilot.NetMux">
+    <a:IPPrefix>0.0.0.0/0</a:IPPrefix>
+    </ManagementAddress>
+    <ManagementAddressV6 xmlns:a="Microsoft.Search.Autopilot.NetMux">
```
qiluo-msft, line 8: >ManagementAddressV6 [](http://example.com/codeflow?start=5&length=19)

Could you reformat this XML file?

src/sonic-config-engine/tests/simple-sample-device-desc-ipv6-only.xml
```diff
@@ -0,0 +1,11 @@
+<Device i:type="ToRRouter" xmlns="Microsoft.Search.Autopilot.Evolution" xmlns:i="http://www.w3.org/2001/XMLSchema-instance">
+  <Hostname>switch-t0</Hostname>
+  <HwSku>Force10-S6000</HwSku>
+  <ClusterName>AAA00PrdStr00</ClusterName>
+  <ManagementAddress xmlns:a="Microsoft.Search.Autopilot.NetMux">
```
qiluo-msft, line 5: >ManagementAddress [](http://example.com/codeflow?start=3&length=17)

Could you create another test data with no `ManagementAddress` node at all?

PR #11027, opened 2022-06-04
[Tunnel PFC][Fix bug] Fix bug and Tests for adding property 'sai_remap_prio_on_tnl_egress'

richardyu-ms:
#### Why I did it
    Fix a bug in the template
    Add tests for adding property 'sai_remap_prio_on_tnl_egress', this
    property should only be added in dual tor environment.
    

#### How I did it
Add tests for adding property 'sai_remap_prio_on_tnl_egress'
#### How to verify it
    Test done:
    Run test test_j2files.py
```
python3 -m pytest src/sonic-config-engine/tests/test_j2files.py
=================================================================== test session starts ===================================================================
platform linux -- Python 3.7.3, pytest-3.10.1, py-1.7.0, pluggy-0.8.0
rootdir: /data/forks/sonic-buildimage/src/sonic-config-engine, inifile:
plugins: cov-2.6.0
collected 27 items                                                                                                                                        

src/sonic-config-engine/tests/test_j2files.py ...........................                                                                           [100%]

=============================================================== 27 passed in 18.84 seconds ================================================================

```
#### Which release branch to backport (provide reason below if selected)

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

Issue #11042, opened 2022-06-06
check the config.bcm template in the build


lguohan:
#### Steps to reproduce the issue:

If the new config.bcm property is hardcoded, we need to add it to permitted_list. Otherwise it will break the build.
If the new config.bcm property is added by template, then config_checker does not check it.
We need to make sure the config_check also checks the template.

#### Output of `show version`:

```
(paste your output here)
```
#### Output of `show techsupport`:

```
(paste your output here or download and attach the file here )
```

PR #11099, opened 2022-06-10
[202012][Tunnel PFC][Fix bug] Fix bug and Tests for adding property 'sai_remap_prio_on_tnl_egress'

richardyu-ms:
#### Why I did it
manual cherry-pick from https://github.com/Azure/sonic-buildimage/pull/11027
Fix a bug in the template
Add tests for adding property 'sai_remap_prio_on_tnl_egress', this
property should only be added in dual tor environment.

#### How I did it
Add tests for adding property 'sai_remap_prio_on_tnl_egress'

#### How to verify it
Test done:
Run test test_j2files.py
```

src/sonic-config-engine/tests/test_j2files.py ........................                                                                              [100%]

=============================================================== 24 passed in 13.31 seconds ================================================================
richardyu@3209c1b66df2:/data/forks/sonic-buildimage$  python3 -m pytest src/sonic-config-engine/tests/test_j2files.py -k 'test_config_brcm_render_template'
=================================================================== test session starts ===================================================================
platform linux -- Python 3.7.3, pytest-3.10.1, py-1.7.0, pluggy-0.8.0
rootdir: /data/forks/sonic-buildimage/src/sonic-config-engine, inifile:
plugins: cov-2.6.0
collected 24 items / 23 deselected 
```

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111


device/arista/x86_64-arista_7260cx3_64/Arista-7260CX3-D108C8/config.bcm.j2
```diff
@@ -952,5 +952,3 @@ serdes_preemphasis_131=0x580c
 
 mmu_init_config="MSFT-TH2-Tier0"
 {{ IPinIP_sock }}
-
-phy_an_lt_msft=1
```
lguohan, line 956: why do you remove this attribute?

src/sonic-config-engine/tests/test_j2files.py
```diff
@@ -336,6 +336,37 @@ def test_buffers_mellanox2410_render_template(self):
...
+        ]
+        for i, config in enumerate(config_bcm_sample_outputs):
+            device_template_path = os.path.join(self.test_dir, './data/j2_template')
+            config_sample_output = config_bcm_sample_outputs[i]
+            sample_minigraph_file = os.path.join(self.test_dir,sample_minigraph_files[i])
+            port_config_ini_file = os.path.join(device_template_path, 'port_config.ini')
+            config_bcm_file = os.path.join(device_template_path, 'config.bcm.j2')
+            config_test_output = os.path.join(self.test_dir, 'config_output.bcm')
+
+            argument = '-m ' + sample_minigraph_file + ' -p ' + port_config_ini_file + ' -t ' + config_bcm_file + ' > ' + config_test_output
```
neethajohn, line 362: I just realized that we are creating our own config_bcm_file template here and using that. Then how do we intend to catch issues when the original .bcm in each of the hwsku folders is modified?

device/arista/x86_64-arista_7260cx3_64/Arista-7260CX3-D108C8/config.bcm.j2
richardyu-ms: fixed, mistakenly changed it when making the manual cherry-pick

src/sonic-config-engine/tests/test_j2files.py
lguohan: we do not. i think the purpose is to mainly to test the template.

i think it is a problem for the unit test to test some files outside of this sonic-config-engine folder, making the sonic-config-engine not self-contained.

richardyu-ms: I think the checks for j2 config, might be divided into two parts, j2 template, and the actual output.

For the j2 file, I chose not to check from the original j2, apart from the self-contained problem, those config files might differ from different platforms, if we want to check them all and make sure the output synced up with the origin file, then we need to copy all those expected output files to the test folder and modify them every time when the origin file changed. Then, in this PR, I only checked the template part.

For checking the content of the generated bcm config, it should be like other .bcm checks. 
It can be checked in the config_checker. Like, illegal/unexpected properties, syntax errors (unexpected token), etc.

Actually, there is an issue with checking the generated bcm config. 
https://github.com/Azure/sonic-buildimage/issues/11042

neethajohn (approved):

PR #2016, opened 2018-09-05
[build]: use vfs storage driver to build dockers

lguohan:
seen issues to build dockers using aufs in ubuntu 18.04

Immedidate dockers are exported to docker file and then
imported into sonic image. Thus, whether using vfs or aufs
as the storage driver does not matter for the immediate build.

Signed-off-by: Guohan Lu <gulv@microsoft.com>

**- What I did**

**- How I did it**

**- How to verify it**
build the image and load to DUT to test.

**- Description for the changelog**

**- A picture of a cute animal (not mandatory but encouraged)**

PR #11111, opened 2022-06-11
[Build] Improve docker build performance

xumia:
#### Why I did it
The docker storage driver vfs is not a good option for build, it uses the “deep copy” when building a new layer, leads to lower performance and more space used on disk than other storage drivers.
A better docker storage driver is the default one overlay2, it is a modern union filesystem.

1. Improve build performance, reduce the build time
    Reduce around 3 hours in Broadcom official build, from 12.5 hours to 9.5 hours.
2. Reduce the disk write bytes
    Reduce from 1600G to 360G in Broadcom official build.
3. Reduce the required disk size during the build.

#### How I did it
1. Change the storage driver setting of the docker in slave container.
2. Mount the local file system to /var/lib/docker, since the overlay fs as the backing filesystem is not supported.

#### How to verify it

Verify the build option:
For default option:
BLDENV=buster make -f Makefile.work docker-start KEEP_SLAVE_ON=yes
```
xumia@dccae0cd559c:/sonic$ docker info | grep "Storage Driver"
Storage Driver: vfs
```
Enable overlay2:
SONIC_SLAVE_DOCKER_DRIVER=overlay2 BLDENV=buster make -f Makefile.work docker-start KEEP_SLAVE_ON=yes
```
xumia@ba4b89040dbb:/sonic$ docker info | grep "Storage Driver"
Storage Driver: overlay2
```
Compare the disk write bytes metrics:
image

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

### DIFF
--- a/Makefile.work
+++ b/Makefile.work
@@ -137,6 +137,7 @@ MULTIARCH_QEMU_ENVIRON = y
 endif
 endif
 SLAVE_IMAGE = $(SLAVE_BASE_IMAGE)-$(USER_LC)
+DOCKER_ROOT = $(PWD)/fsroot.docker.$(BLDENV)
 
 # Support FIPS feature, armhf not supported yet
 ifeq ($(PLATFORM_ARCH),armhf)
@@ -181,6 +182,7 @@ endif
 DOCKER_LOCKFILE_SAVE := $(DOCKER_LOCKDIR)/docker_save.lock
 $(shell mkdir -m 0777 -p $(DOCKER_LOCKDIR))
 $(shell [ -f $(DOCKER_LOCKFILE_SAVE) ] || (touch $(DOCKER_LOCKFILE_SAVE) && chmod 0777 $(DOCKER_LOCKFILE_SAVE)))
+$(shell sudo rm -rf $(DOCKER_ROOT) && mkdir -p $(DOCKER_ROOT))
 
 ifeq ($(DOCKER_BUILDER_MOUNT),)
 override DOCKER_BUILDER_MOUNT := "$(PWD):/sonic"
@@ -193,6 +195,7 @@ endif
 DOCKER_RUN := docker run --rm=true --privileged --init \
     -v $(DOCKER_BUILDER_MOUNT) \
     -v "$(DOCKER_LOCKDIR):$(DOCKER_LOCKDIR)" \
+    -v $(DOCKER_ROOT):/var/lib/docker \
     -w $(DOCKER_BUILDER_WORKDIR) \
     -e "http_proxy=$(http_proxy)" \
     -e "https_proxy=$(https_proxy)" \
--- a/Makefile.work
+++ b/Makefile.work
@@ -195,13 +195,17 @@ endif
 DOCKER_RUN := docker run --rm=true --privileged --init \
     -v $(DOCKER_BUILDER_MOUNT) \
     -v "$(DOCKER_LOCKDIR):$(DOCKER_LOCKDIR)" \
-    -v $(DOCKER_ROOT):/var/lib/docker \
     -w $(DOCKER_BUILDER_WORKDIR) \
     -e "http_proxy=$(http_proxy)" \
     -e "https_proxy=$(https_proxy)" \
     -e "no_proxy=$(no_proxy)" \
     -i$(shell { if [ -t 0 ]; then echo t; fi }) \
     $(SONIC_BUILDER_EXTRA_CMDLINE)
+
+# Mount the $(DOCKER_ROOT) to /var/lib/docker in the slave container, the overlay fs is not supported as dockerd root folder.
+ifneq ($(filter $(SONIC_SLAVE_DOCKER_DRIVER),overlay overlay2),)
+	DOCKER_RUN += -v $(DOCKER_ROOT):/var/lib/docker
+endif
 
 ifneq ($(DOCKER_BUILDER_USER_MOUNT),)
 	DOCKER_RUN += $(foreach mount,$(subst $(comma), ,$(DOCKER_BUILDER_USER_MOUNT)), $(addprefix -v , $(mount)))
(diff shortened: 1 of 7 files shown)


lguohan:
what is the difference between this one with the SONIC_CONFIG_USE_NATIVE_DOCKERD_FOR_BUILD?

xumia:
> 2\. /var/lib/docker,

It uses different docker instances, the native docker uses the same docker instance started on the host.
In the native docker mode, in the salve container, you can see everything the same as in the host.
```
$ BLDENV=buster make -f Makefile.work sonic-slave-bash SONIC_CONFIG_USE_NATIVE_DOCKERD_FOR_BUILD=y

xumia@d48d9949fb53:/sonic$ sudo docker ps | grep d48d9949fb53
d48d9949fb53        sonic-slave-buster-xumia:c177586b087   "bash"                   About a minute ago   Up About a minute   22/tcp              strange_kilby
xumia@d48d9949fb53:/sonic$ 
```

xumia:
> what is the difference between this one with the SONIC_CONFIG_USE_NATIVE_DOCKERD_FOR_BUILD?



> what is the difference between this one with the SONIC_CONFIG_USE_NATIVE_DOCKERD_FOR_BUILD?



> > what is the difference between this one with the SONIC_CONFIG_USE_NATIVE_DOCKERD_FOR_BUILD?
> 
> It uses different docker instances, the native docker uses the same docker instance started on the host. In the native docker mode, in the salve container, you can see everything the same as in the host. If multiple users use the same build system, they are sharing the same docker instance. We do not use the native docker in azp.
The PR is for native docker, it still starts a docker instance inside of the docker slave container, the behavior is not changed.
> 
> ```
> $ BLDENV=buster make -f Makefile.work sonic-slave-bash SONIC_CONFIG_USE_NATIVE_DOCKERD_FOR_BUILD=y
> 
> xumia@d48d9949fb53:/sonic$ sudo docker ps | grep d48d9949fb53
> d48d9949fb53        sonic-slave-buster-xumia:c177586b087   "bash"                   About a minute ago   Up About a minute   22/tcp              strange_kilby
> xumia@d48d9949fb53:/sonic$ 
> ```

lguohan:
Impacts:
The default docker storage driver vfs is not configured in sonic slave images. If someone uses the images for the other scenarios, not sonic build, and want to use the docker daemon inside of the salve docker, needs to config the docker options himself or herself.

I am wondering if you can keep vfs as the default option and add new option for this? i think the slave image is used in some other places.

Makefile.work
```diff
@@ -193,6 +195,7 @@ endif
 DOCKER_RUN := docker run --rm=true --privileged --init \
     -v $(DOCKER_BUILDER_MOUNT) \
     -v "$(DOCKER_LOCKDIR):$(DOCKER_LOCKDIR)" \
+    -v $(DOCKER_ROOT):/var/lib/docker \
```
lguohan, line 198: is there way to cleanup this folder? I guess it would be better if we do not reuse the host docker folder for this inside docker builds.

Makefile.work
```diff
@@ -181,6 +182,7 @@ endif
 DOCKER_LOCKFILE_SAVE := $(DOCKER_LOCKDIR)/docker_save.lock
 $(shell mkdir -m 0777 -p $(DOCKER_LOCKDIR))
 $(shell [ -f $(DOCKER_LOCKFILE_SAVE) ] || (touch $(DOCKER_LOCKFILE_SAVE) && chmod 0777 $(DOCKER_LOCKFILE_SAVE)))
+$(shell sudo rm -rf $(DOCKER_ROOT) && mkdir -p $(DOCKER_ROOT))
```
lguohan, line 185: is there way to cleanup this folder?

xumia: Yes, the folder can be cleaned up when calling "make reset".
https://github.com/Azure/sonic-buildimage/blob/356b51f4d6c903e5a40c9a3ab01feae5182e0724/Makefile.work#L466

For azp, we will do cleanup for the whole workspace.

xumia:
> > The default docker storage driver vfs is not configured in sonic slave images. If someone uses the images for the other
> > scenarios, not sonic build, and want to use the docker daemon inside of the salve docker, needs to config the docker options
> > himself or herself.
> 
> I am wondering if you can keep vfs as the default option and add new option for this? i think the slave image is used in some other places.

Yes, fixed. We can config the overlay2 option inside of the slave containers, not necessary to change in the slave image.

Makefile.work
xumia: We have already added cleanup for "fsroot*", it can be cleaned up by "make reset".

rules/config
```diff
@@ -229,3 +229,6 @@ DEFAULT_CONTAINER_REGISTRY ?=
 # ENABLE_FIPS - support FIPS flag, if enabled, no additional config requred for the image to support FIPS
 ENABLE_FIPS_FEATURE ?= y
 ENABLE_FIPS ?= n
+
+# SONIC_SLAVE_DOCKER_DRIVER - set the sonic slave docker storage driver
+SONIC_SLAVE_DOCKER_DRIVER ?= overlay2
```
lguohan, line 234: suggest to keep the default option as vfs, and we can experiment this overlay2 in our azure pipeline more before we change this option?

lguohan:
for reference, why we changed to vfs in the first place. https://github.com/Azure/sonic-buildimage/pull/2016

rules/config
xumia: Yes, it makes sense to only enable it in azp.
Fixed, thanks.

xumia:
> for reference, why we changed to vfs in the first place. #2016

Thanks for the info. It is true, the overlay fs cannot be used as backing filesystem of docker. If the source folder is an overlay fs, then it will be an issue, need to mount a non-overlay fs for the slave docker. In the azp agents, it should work fine.

lguohan (approved):

jusherma:

xumia:
> This change assumes that all users building SONIC have password-less sudo access, which is not the case in our build environment. This is causing build problems for us.

@jusherma , it was caused by the following line?
It is to clean up the data, maybe we can move the cleanup action inside of the slave container in slave.mk.
```
$(shell sudo rm -rf $(DOCKER_ROOT) && mkdir -p $(DOCKER_ROOT))
```

PR #11117, opened 2022-06-13
Add Yang model for `SYSTEM_DEFAULTS` table

bingwang-ms:
Signed-off-by: bingwang <wang.bing@microsoft.com>

#### Why I did it
This PR is to defind Yang model for `SYSTEM_DEFAULTS` table.
The table was introduced in PR https://github.com/sonic-net/SONiC/pull/982
The table will be like
```
"SYSTEM_DEFAULTS": {
        "tunnel_qos_remap": {
            "status": "enabled"
        }
}
```

#### How I did it
Add a new yang file `sonic-system-defaults.yang`.

#### How to verify it
Verified by UT
```
collected 3 items                                                                                                                                                                                     

tests/test_sonic_yang_models.py ..                                                                                                                                                              [ 66%]
tests/yang_model_tests/test_yang_model.py .                                                                                                                                                     [100%]

====================================================================================== 3 passed in 2.27 seconds =======================================================================================
```

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

Add Yang model for `SYSTEM_DEFAULTS` table

TODO


src/sonic-yang-models/setup.py
```diff
@@ -104,6 +104,7 @@ def run(self):
                          './yang-models/sonic-extension.yang',
                          './yang-models/sonic-flex_counter.yang',
                          './yang-models/sonic-feature.yang',
+                         './yang-models/sonic-system-defaults.yang',
```
qiluo-msft, line 107: Also need updating the file src/sonic-yang-models/doc/Configuration.md

src/sonic-yang-models/yang-models/sonic-system-defaults.yang
```diff
@@ -0,0 +1,41 @@
...
+                    type string {
+                        length 1..32;
+                    }
+                }
+
+                leaf status {
+                    description "default status of the feature";
+                    type string {
+                        pattern "enabled|disabled";
+                    }
```
ganglyu, line 35: Can you use stypes:admin_status?

bingwang-ms: I think the value of `stypes:admin_status` is `up` or `down`. I think it would be better to use `enabled` or `disabled` here.

ganglyu: https://github.com/Azure/sonic-buildimage/blob/master/src/sonic-yang-models/yang-templates/sonic-types.yang.j2#L126-L131
How about admin_mode?

ganglyu (approved):

src/sonic-yang-models/yang-models/sonic-system-defaults.yang
bingwang-ms: Yes, `admin_mode` would be good. Updated.

qiluo-msft (approved):

wen587 (approved):

qiluo-msft (approved):

bingwang-ms:
PR for `202205` branch is filed https://github.com/sonic-net/sonic-buildimage/pull/15129
The change is already included in `202211` branch. So we don't need to do the cherry-pick to `202211`. I have removed the labels.

PR #11141, opened 2022-06-14
Update include_macsec flag if type is SpineRouter

judyjoseph:
#### Why I did it
 Update the feature table in STATE_DB in database running in namespace for multi-asic platforms.
This will be used by features running in namespaces.

This is needed by features like https://github.com/Azure/sonic-swss/pull/2286, which need to know the state of feature before adding feature based condition

Also add the support to enable macsec when type == "SpineRouter"

#### How I did it
Updated the init_cfg.json.j2 file and hostcfgd.

#### How to verify it
Verified with feature running in multiple namespaces 

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205

### DIFF
--- a/src/sonic-host-services/scripts/hostcfgd
+++ b/src/sonic-host-services/scripts/hostcfgd
@@ -11,7 +11,7 @@ import signal
 
 import jinja2
 from sonic_py_common import device_info
-from swsscommon.swsscommon import ConfigDBConnector, DBConnector, Table, SonicDBConfig
+from swsscommon.swsscommon import ConfigDBConnector, DBConnector, Table
 
 # FILE
 PAM_AUTH_CONF = "/etc/pam.d/common-auth-sonic"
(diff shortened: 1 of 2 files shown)


src/sonic-host-services/scripts/hostcfgd
```diff
@@ -399,6 +399,12 @@ class FeatureHandler(object):
     def set_feature_state(self, feature, state):
         self._feature_state_table.set(feature.name, [('state', state)])
 
+        # Update the feature state in STATE_DB in namespaces in multi-asic platform
+        namespaces = device_info.get_namespaces()
+        for namespace in namespaces:
+            db_conn = DBConnector(STATE_DB, 0, False, namespace);
+            feature_state_tbl = Table(db_conn, 'FEATURE')
+            feature_state_tbl.set(feature.name, [('state', state)])
```
judyjoseph, line 408: Check if the hostcfgd starts before orchagent, to make sure there is no timeing issue.

lguohan: wonder if we really want feature table to be in each namespace.

judyjoseph: I have updated this.

arlakshm (approved):

PR #11146, opened 2022-06-15
Remove minigraph loading in script

wen587:
#### Why I did it
Minigraph will be deprecated in the future. So minigraph related reload should be deleted.
#### How I did it
Remove unused load_minigraph

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205

### DIFF
--- a/files/image_config/config-setup/config-setup
+++ b/files/image_config/config-setup/config-setup
@@ -31,7 +31,6 @@ CONFIG_DB_JSON=/etc/sonic/config_db.json
 CONFIG_DB_PATH=/etc/sonic/
 CONFIG_DB_PREFIX=config_db
 CONFIG_DB_SUFFIX=.json
-MINGRAPH_FILE=/etc/sonic/minigraph.xml
 TMP_ZTP_CONFIG_DB_JSON=/tmp/ztp_config_db.json
 FACTORY_DEFAULT_HOOKS=/etc/config-setup/factory-default-hooks.d
 CONFIG_PRE_MIGRATION_HOOKS=/etc/config-setup/config-migration-pre-hooks.d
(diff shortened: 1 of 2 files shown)


wen587:
In sonic-ztp repo, do we need to remove minigraph here?
https://github.com/Azure/sonic-ztp/blob/master/src/usr/lib/ztp/plugins/graphservice#L90

files/image_config/updategraph/updategraph
```diff
@@ -17,13 +10,6 @@ fi
 
 . /etc/sonic/updategraph.conf
 
-if [ "$enabled" = "reload_only" ]; then
```
qiluo-msft, line 20: We can still keep the if statement, and echo an error message to stderr, and exit with non zero.

qiluo-msft:
@abdosi Please review on multi-ASIC use case.

files/image_config/config-setup/config-setup
```diff
@@ -329,13 +320,8 @@ do_config_migration()
...
         # Disable updategraph
         disable_updategraph
-    elif [ -r ${MINGRAPH_FILE} ]; then
-        echo "Use minigraph.xml from old system..."
-        reload_minigraph
-        # Disable updategraph
-        disable_updategraph
     else
-        echo "Didn't found neither config_db.json nor minigraph.xml ..."
+        echo "Didn't found neither config_db.json..."
```
qiluo-msft, line 324: >neither [](http://example.com/codeflow?start=27&length=7)

remove "neither"

yxieca (changes requested):
By keeping config_db.json before rebooting into new image. This code path is already avoided. There is NO point of removing this code.

Risk to nightly tests needs to be assesses.

files/image_config/updategraph/updategraph
```diff
@@ -18,10 +11,8 @@ fi
 . /etc/sonic/updategraph.conf
 
 if [ "$enabled" = "reload_only" ]; then
```
qiluo-msft, line 13: We can focus on this file in this PR. Which seems safe to merge.

wen587:
@yxieca, coud you review again?

yxieca (approved):

qiluo-msft (approved):

PR #11152, opened 2022-06-15
[bgpcfgd] improve the log when delete a loopback interface

StormLiangMS:
#### Why I did it
The bgpcfgd doesn't support deletion of 'zebra set src', if an interface is deleted, the bgpcfgd will drop a warning message. In current implementation, we only care about the loopback0 interface but not others. 
To improve the log print to have the key info, which will give the name of the deleted interface. We can ignore it if it is not the loopback0 interface. The application layer should be aware of that update and deletion is not supported, delete or update with a new address of loopback0 could cause issue, this log can give enough info to root cause the issue.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205

### DIFF
--- a/src/sonic-bgpcfgd/bgpcfgd/managers_setsrc.py
+++ b/src/sonic-bgpcfgd/bgpcfgd/managers_setsrc.py
@@ -64,4 +64,4 @@ class ZebraSetSrc(Manager):
     def del_handler(self, key):
         """ Implementation of 'DEL' command for this class """
         self.directory.remove(self.db_name, self.table_name, key)
-        log_warn("Delete command is not supported for 'zebra set src' templates")
+        log_warn("Delete key '%s' is not supported for 'zebra set src' templates" % str(key))
--- a/src/sonic-bgpcfgd/tests/test_setsrc.py
+++ b/src/sonic-bgpcfgd/tests/test_setsrc.py
@@ -58,5 +58,6 @@ def test_set_handler_invalid_ip(mocked_log_err):
 @patch('bgpcfgd.managers_setsrc.log_warn')
 def test_del_handler(mocked_log_warn):
     m = constructor()
-    m.del_handler("Loopback0|10.1.0.32/32")
-    mocked_log_warn.assert_called_with("Delete command is not supported for 'zebra set src' templates")
+    del_key = "Loopback0|10.1.0.32/32"
+    m.del_handler(del_key)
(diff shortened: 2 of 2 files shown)


prsunny (approved):
Do we need to have a test for this?

wen587:
Hi @StormLiangMS, is the bgpcfgd using zebra? Is it an issue when deleting the loopback interface?
Please see the below scenario, it will warn us when we remove loopback by CLI.

Remove Loopback interface through Config CLI:
```
admin@vlab-01:~$ show ip int | grep Loop
Loopback0                 10.1.0.32/32         up/up         N/A             N/A
admin@vlab-01:~$ sudo config interface ip remove Loopback0 10.1.0.32/32
admin@vlab-01:~$
```
Check Syslog:
```
admin@vlab-01:~$ show logging -f
...
Jun  1 06:13:45.658305 vlab-01 DEBUG bgp#bgpcfgd: Received message : '('Loopback0|10.1.0.32/32', 'DEL', ())'
Jun  1 06:13:45.658305 vlab-01 WARNING bgp#bgpcfgd: Delete command is not supported for 'zebra set src' templates
```

StormLiangMS:
> Hi @StormLiangMS, is the bgpcfgd using zebra? Is it an issue when deleting the loopback interface? Please see the below scenario, it will warn us when we remove loopback by CLI.
> 
> Remove Loopback interface through Config CLI:
> 
> ```
> admin@vlab-01:~$ show ip int | grep Loop
> Loopback0                 10.1.0.32/32         up/up         N/A             N/A
> admin@vlab-01:~$ sudo config interface ip remove Loopback0 10.1.0.32/32
> admin@vlab-01:~$
> ```
> 
> Check Syslog:
> 
> ```
> admin@vlab-01:~$ show logging -f
> ...
> Jun  1 06:13:45.658305 vlab-01 DEBUG bgp#bgpcfgd: Received message : '('Loopback0|10.1.0.32/32', 'DEL', ())'
> Jun  1 06:13:45.658305 vlab-01 WARNING bgp#bgpcfgd: Delete command is not supported for 'zebra set src' templates
> ```

@wen587 
1. bgpcfgd will pass this configuration to zebra to have the source IP for the route. 2. deletion and update is not supported, if the loopback0 interface is removed, currently bgpcfgd will do nothing. And the routes which distributed to kernel would be deleted.

StormLiangMS:
> Do we need to have a test for this?

@prsunny Yes, changed the unit test for this new log changes.

wen587 (approved):

PR #11159, opened 2022-06-16
[202205][bcm sai] upgrade Broadcom SAI to 7.1.0.0-5

yxieca:
#### Why I did it

- Enable Microsoft AN/LT patch


platform/broadcom/sai.mk
```diff
@@ -1,4 +1,4 @@
-LIBSAIBCM_VERSION = 7.1.0.0-1
+LIBSAIBCM_VERSION = 7.1.0.0-5
```
lguohan, line 1: update xgs only thanks.

yxieca: Done.

yxieca: Fully separated the platforms.

lguohan (approved):

PR #11188, opened 2022-06-19
Change severity of log messages for cases where docker container was stopped during service checker operation

liorghub:
#### Why I did it
There might be a case where service checker periodic operation determined that specific container is running but when it tries to perform an operation on it, it was already closed by the user. This is a valid flow and we should not log an error message, informative warning is enough. 

#### How I did it
I reduce log severity.

#### How to verify it
I verified it manually. 

- [x] 202111
- [x] 202205


dprital:
@prsunny - Can you please approve ?

prsunny:
> @prsunny - Can you please approve ?

I'm not familiar with this code. Adding @qiluo-msft to check.

qiluo-msft (approved):

PR #11204, opened 2022-06-21
Add support for generating interface configuration in /etc/network/interfaces for multiple management interfaces

geogchen:
Signed-off-by: George Chen gechen@microsoft.com

#### Why I did it
The interfaces template, interfaces.j2, currently hardcodes to eth0 and therefore cannot support multiple management interfaces.  This pull request introduces change to support multiple management interfaces.

#### How I did it
1) Update interfaces.j2 to iterate over MGMT_INTERFACE where it is currently not, and refer to interface name rather than hardcoded eth0.
2) Update minigraph.py to process mgmt_intf in sorted order.  This is needed because in Python 3.x, the dictionary keys are presented in such a way that eth1 in [t0-sample-graph-two-mgmt.xml](https://github.com/Azure/sonic-buildimage/compare/master...geogchen:dev/gechen/multi_mgmt_interface?expand=1#diff-0a80119cc01db12d4686c311fc7be9281147ff487247d611610f4784e526f454) appears first, and it gets rewritten to be eth0 into MGMT_INTERFACE.  To prevent the interface-reordering, process the interfaces in sorted order.

#### How to verify it
Unit test.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205

Add support to generating multiple management interfaces for /etc/network/interfaces 

### DIFF
--- a/src/sonic-config-engine/minigraph.py
+++ b/src/sonic-config-engine/minigraph.py
@@ -10,6 +10,7 @@ from collections import defaultdict
 from lxml import etree as ET
 from lxml.etree import QName
 
+from natsort import natsorted
 
 from portconfig import get_port_config
 from sonic_py_common.interface import backplane_prefix
@@ -1432,7 +1433,7 @@ def parse_xml(filename, platform=None, port_config_file=None, asic_name=None, hw
     results['MGMT_INTERFACE'] = {}
     mgmt_intf_count = 0
     mgmt_alias_reverse_mapping = {}
-    sorted_keys = sorted(mgmt_intf.keys(), key=lambda x : x[0].lower())
+    sorted_keys = natsorted(mgmt_intf.keys(), key=lambda x : x[0].lower())
     for key in sorted_keys:
         alias = key[0]
         if alias in mgmt_alias_reverse_mapping:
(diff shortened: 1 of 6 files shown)


src/sonic-config-engine/minigraph.py
```diff
@@ -1432,7 +1432,8 @@ def parse_xml(filename, platform=None, port_config_file=None, asic_name=None, hw
     results['MGMT_INTERFACE'] = {}
     mgmt_intf_count = 0
     mgmt_alias_reverse_mapping = {}
-    for key in mgmt_intf:
+    sorted_keys = sorted(mgmt_intf.keys(), key=lambda x : x[0].lower())
```
qiluo-msft, line 1435: >sorted [](http://example.com/codeflow?start=18&length=6)

Is `sorted` really needed?

geogchen: without sorting, in Py 3 eth1 is returned first, and will be re-written as eth0 in [this line](https://github.com/Azure/sonic-buildimage/pull/11204/files/6351b9a9a5b5bb0daa508842128ce131cd3b4ec5#diff-466ae5de99730b4f3dde9c932e9086536e3a0226858c735fbf776453e8d86fafR1441).

qiluo-msft: Strictly speaking, better to use natural sort, like Management1~Management11

qiluo-msft (approved):

src/sonic-config-engine/minigraph.py
geogchen: changed to use natsort

PR #11221, opened 2022-06-22
Minigraph parser changes for storage backend acl

neethajohn:
Signed-off-by: Neetha John <nejo@microsoft.com>

#### Why I did it
For storage backend, certain rules will be applied to the DATAACL table to allow only vlan tagged packets and drop untagged packets. 

#### How I did it
Create DATAACL table if the device is a storage backend device
To avoid ACL resource issues, remove EVERFLOW related tables if the device is a storage backend device

#### How to verify it
Added the following unit tests
 - verify that EVERFLOW acl tables is removed and DATAACL table is added for storage backend tor
 - verify that no DATAACL tables are created and EVERFLOW tables exist for storage backend leaf

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [x] 202111
- [x] 202205

### DIFF
--- a/src/sonic-config-engine/minigraph.py
+++ b/src/sonic-config-engine/minigraph.py
@@ -1169,18 +1169,18 @@ def parse_spine_chassis_fe(results, vni, lo_intfs, phyport_intfs, pc_intfs, pc_m
 
 def filter_acl_table_for_backend(acls, vlan_members):
     filter_acls = {}
-    for acl_name in acls:
+    for acl_name, value in acls.items():
         if 'everflow' not in acl_name.lower():
-            filter_acls[acl_name] = acls[acl_name]
+            filter_acls[acl_name] = value
 
+    ports = set()
+    for vlan, member in vlan_members:
+        ports.add(member)
     filter_acls['DATAACL'] = { 'policy_desc': 'DATAACL',
                                'stage': 'ingress',
-                               'type': 'L3'
+                               'type': 'L3',
+                               'ports': list(ports)
                              }
-    ports = set()
-    for vlan, member in vlan_members.keys():
-        ports.add(member)
-    filter_acls['DATAACL']['ports'] = list(ports)
     return filter_acls
 
 def filter_acl_table_bindings(acls, neighbors, port_channels, sub_role, device_type, is_storage_device, vlan_members):
--- a/src/sonic-config-engine/minigraph.py
+++ b/src/sonic-config-engine/minigraph.py
@@ -1177,10 +1177,10 @@ def filter_acl_table_for_backend(acls, vlan_members):
                                'stage': 'ingress',
                                'type': 'L3'
                              }
-    ports = list()
+    ports = set()
     for vlan, member in vlan_members.keys():
-        ports.append(member)
-    filter_acls['DATAACL']['ports'] = ports
+        ports.add(member)
+    filter_acls['DATAACL']['ports'] = list(ports)
     return filter_acls
 
 def filter_acl_table_bindings(acls, neighbors, port_channels, sub_role, device_type, is_storage_device, vlan_members):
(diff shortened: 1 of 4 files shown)


src/sonic-config-engine/minigraph.py
```diff
@@ -1142,7 +1142,26 @@ def parse_spine_chassis_fe(results, vni, lo_intfs, phyport_intfs, pc_intfs, pc_m
 #
 ###############################################################################
 
-def filter_acl_table_bindings(acls, neighbors, port_channels, sub_role):
+def filter_acl_table_for_backend(acls, vlan_members):
+    filter_acls = {}
+    for acl_name in acls:
```
qiluo-msft, line 1147: >acls [](http://example.com/codeflow?start=20&length=4)

you can get key and value at the same time `in acls.items()`.

src/sonic-config-engine/minigraph.py
```diff
@@ -1142,7 +1142,26 @@ def parse_spine_chassis_fe(results, vni, lo_intfs, phyport_intfs, pc_intfs, pc_m
...
+        if 'everflow' not in acl_name.lower():
+            filter_acls[acl_name] = acls[acl_name]
+
+    filter_acls['DATAACL'] = { 'policy_desc': 'DATAACL',
+                               'stage': 'ingress',
+                               'type': 'L3'
+                             }
+    ports = set()
+    for vlan, member in vlan_members.keys():
```
qiluo-msft, line 1156: >.keys() [](http://example.com/codeflow?start=36&length=7)

You can remove `.keys()`.

src/sonic-config-engine/minigraph.py
```diff
@@ -1142,7 +1142,26 @@ def parse_spine_chassis_fe(results, vni, lo_intfs, phyport_intfs, pc_intfs, pc_m
...
 ###############################################################################
 
-def filter_acl_table_bindings(acls, neighbors, port_channels, sub_role):
+def filter_acl_table_for_backend(acls, vlan_members):
+    filter_acls = {}
+    for acl_name in acls:
+        if 'everflow' not in acl_name.lower():
+            filter_acls[acl_name] = acls[acl_name]
+
+    filter_acls['DATAACL'] = { 'policy_desc': 'DATAACL',
```
qiluo-msft, line 1151: >filter_acls [](http://example.com/codeflow?start=4&length=11)

You can assign `ports` in the same line, just move this line to end of function.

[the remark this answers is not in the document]
neethajohn: this function is called only for storage backend ToR (L1163) which does not have any DATAACLS. so overriding should be okay.

src/sonic-config-engine/minigraph.py
neethajohn: Done

src/sonic-config-engine/minigraph.py
neethajohn: Done

src/sonic-config-engine/minigraph.py
neethajohn: Done

qiluo-msft (approved):

src/sonic-config-engine/minigraph.py
```diff
@@ -1167,7 +1167,26 @@ def parse_spine_chassis_fe(results, vni, lo_intfs, phyport_intfs, pc_intfs, pc_m
...
+                               'stage': 'ingress',
+                               'type': 'L3',
+                               'ports': list(ports)
+                             }
+    return filter_acls
+
+def filter_acl_table_bindings(acls, neighbors, port_channels, sub_role, device_type, is_storage_device, vlan_members):
+    if device_type == 'BackEndToRRouter' and is_storage_device:
+        return filter_acl_table_for_backend(acls, vlan_members)
```
prsunny, line 1188: Prior to this change, the rest of the code was executed for BackendToR.. like portchannel, mirror etc. Is that not required anymore?

yxieca (approved):

PR #11291, opened 2022-06-29
[yang]Adding a tool for generating documentation based on yang model

dgsudharsan:
#### Why I did it
Added a tool to generate configuration based on yang model

#### How I did it
Parse the yang model and create documentation based on description field.

#### How to verify it
Added UT to verify it.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205


zhangyanzhao:
This tool is very helpful to improve the user experience and dev productivity. Love this idea and thanks for the contribution!

src/sonic-yang-mgmt/tests/test_cfghelp.py
```diff
@@ -0,0 +1,170 @@
...
+
+    def test_dummy_run(self):
+        argument = ''
+        output = self.run_script(argument)
+        self.assertEqual(output, output1)
+
+    def test_single_table(self):
+        argument = '-t AUTO_TECHSUPPORT'
+        output = self.run_script(argument)
+        self.assertEqual(output, techsupport_table_output)
```
ganglyu, line 149: If we modify techsupport yang model, this unit test will fail?

ganglyu: Maybe we can create some dummy yang models for unit test.
And then we can modify Yang model, and don't need to update this unit test.

dgsudharsan: Hi @ganglyu ,
I prefer to use any active yang model rather than introducing a dummy for the test. We do have instances of such tests in other repositories where if schema is changed, the UT needs to be updated.
I don't think the yang model I chosen would change quite often. Even if someone changes it and accidentally misses to update UT, it will be caught during build time and build will fail, forcing the user to modify the test.

ganglyu (approved):

PR #11298, opened 2022-06-30
[Marvell-armhf] Fix kernel hang due to kernel upgrade to 5.10.103

pavannaregundi:
Change-Id: I015856d173d50d94e30d8c555590efb70eb712ae
Signed-off-by: Pavan Naregundi <pnaregundi@marvell.com>

#### Why I did it
Kernel hang was cause due to increase in size of kernel during the upgrade to 5.10.103 and uboot env setting for fdt and initrd address was not good enough.

#### How I did it
Have given sufficient room for kernel, by moving fdt_addr and initrd_addr.

#### How to verify it
Verified by loading image on Marvell Armada 380/385.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [x ] 202205

### DIFF
--- a/platform/marvell-armhf/platform.conf
+++ b/platform/marvell-armhf/platform.conf
@@ -6,9 +6,9 @@ echo "Preparing for installation ... "
 
 # global defines
 kernel_addr=0x1100000
-fdt_addr=0x1000000
-fdt_high=0x10fffff
-initrd_addr=0x2000000
+fdt_addr=0x2800000
+fdt_high=0x28fffff
+initrd_addr=0x2900000
 
 kernel_fname="/boot/vmlinuz-5.10.0-12-2-armmp"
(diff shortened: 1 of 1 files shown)


saiarcot895:
With this change, does it mean that 28MB is reserved for the kernel?

pavannaregundi:
> With this change, does it mean that 28MB is reserved for the kernel?

kernel_addr is a location of zImage loaded from disk and uncompressed kernel will be reloaded, So have chosen a safe range. 24M is diff of kernel_addr to fdt_addr.

radha-danda:
> With this change, does it mean that 28MB is reserved for the kernel?

@saiarcot895, Pavan has responded to your query. do you have any further questions? If not can you please review and merge the PR?

PR #11341, opened 2022-07-05
[arm] Refactor installer and build to allow arm builds targeted at grub platforms

alexrallen:
Refactors the SONiC Installer to support greater flexibility in building for a given architecture and bootloader. 

#### Why I did it

Currently the SONiC installer assumes that if a platform is ARM based that it uses the `uboot` bootloader and uses the `grub` bootloader otherwise. This is not a correct assumption to make as ARM is not strictly tied to uboot and x86 is not strictly tied to grub. 

#### How I did it

To implement this I introduce the following changes:
* Remove the different arch folders from the `installer/` directory
* Merge the generic components of the ARM and x86 installer into `installer/installer.sh`
* Refactor x86 + grub specific functions into `installer/default_platform.conf` 
* Modify installer to call `default_platform.conf` file and also call `platform/[platform]/patform.conf` file as well to override as needed
* Update references to the installer in the `build_image.sh` script
* Add `TARGET_BOOTLOADER` variable that is by default `uboot` for ARM devices and `grub` for x86 unless overridden in `platform/[platform]/rules.mk`
* Update bootloader logic in `build_debian.sh` to be based on `TARGET_BOOTLOADER` instead of `TARGET_ARCH` and to reference the grub package in a generic manner

#### How to verify it

This has been tested on a ARM test platform as well as on Mellanox amd64 switches as well to ensure there was no impact. 

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205

[arm] Refactor installer and build to allow arm builds targeted at grub platforms

N/A

possum

### DIFF
--- a/build_image.sh
+++ b/build_image.sh
@@ -84,7 +84,7 @@ generate_onie_installer_image()
 
     ## Generate an ONIE installer image
     ## Note: Don't leave blank between lines. It is single line command.
-    ./onie-mk-demo.sh $TARGET_PLATFORM $TARGET_MACHINE $TARGET_PLATFORM-$TARGET_MACHINE-$ONIEIMAGE_VERSION \
+    ./onie-mk-demo.sh $CONFIGURED_ARCH $TARGET_MACHINE $TARGET_PLATFORM-$TARGET_MACHINE-$ONIEIMAGE_VERSION \
           installer platform/$TARGET_MACHINE/platform.conf $output_file OS $IMAGE_VERSION $ONIE_IMAGE_PART_SIZE \
           $ONIE_INSTALLER_PAYLOAD
 }
--- a/installer/install.sh
+++ b/installer/install.sh
@@ -62,8 +62,14 @@ if [ -r ./machine.conf ]; then
     read_conf_file "./machine.conf"
 fi
 
-if [ -r ./onie-image*.conf ]; then
-. ./onie-image*.conf
+# Load generic onie-image.conf
+if [ -r ./onie-image.conf ]; then
+. ./onie-image.conf
+fi
+
+# Load arch-specific onie-image-[arch].conf if exists
+if [ -r ./onie-image-*.conf ]; then
+. ./onie-image-*.conf
 fi
 
 echo "ONIE Installer: platform: $platform"
--- a/onie-mk-demo.sh
+++ b/onie-mk-demo.sh
@@ -77,7 +77,7 @@ mkdir $tmp_installdir || clean_up 1
 
 cp -r $installer_dir/* $tmp_installdir || clean_up 1
 cp onie-image.conf $tmp_installdir
-cp onie-image-*.conf $tmp_installdir
+cp onie-image-$arch.conf $tmp_installdir
 
 # Set sonic fips config for the installer script
 if [ "$ENABLE_FIPS" = "y" ]; then


dgsudharsan (approved):

installer/install.sh
```diff
@@ -30,7 +59,7 @@ fi
 
 cd $(dirname $0)
 if [ -r ./machine.conf ]; then
-    . ./machine.conf
+    read_conf_file "./machine.conf"
 fi
 
 if [ -r ./onie-image-arm64.conf ]; then
```
qiluo-msft, line 65: >onie-image-arm64.conf [](http://example.com/codeflow?start=10&length=21)

Will you add another block for `onie-image-armhf.conf` ?

alexrallen: Yes I will need to. Good catch thanks.

alexrallen: Ultimately with these changes I am just looking to make sure the build succeeds and SONiC successfully installs via both ONIE and SONiC-SONiC upgrade / downgrade.

installer/default_platform.conf
```diff
@@ -1,119 +1,5 @@
...
-# Get platform specific linux kernel command line arguments
-ONIE_PLATFORM_EXTRA_CMDLINE_LINUX=""
-
-# Default var/log device size in MB
-VAR_LOG_SIZE=4096
-
-[ -r platforms/$onie_platform ] && . platforms/$onie_platform
-
-# Verify image platform is inside devices list
-if [ "$install_env" = "onie" ]; then
```
wen587, line 96: Hi, after removing this part, how do you handle the below cases:
1. install sonic-mellnoax.bin in broadcom ASIC. (That could be caused by human error)
2. install sonic-broadcom.bin in mellonox ASIC.

build_image.sh
```diff
@@ -111,7 +111,7 @@ if [ "$IMAGE_TYPE" = "onie" ]; then
     mkdir -p `dirname $OUTPUT_ONIE_IMAGE`
     sudo rm -f $OUTPUT_ONIE_IMAGE
 
-    generate_device_list "./installer/$TARGET_PLATFORM/platforms_asic"
```
wen587, line 114: sonic-utilities also need to change if the path change.
https://github.com/Azure/sonic-utilities/blob/master/sonic_installer/bootloader/grub.py#L21

wen587 (changes requested):

installer/default_platform.conf
wen587: Hi @liat-grozovik, I think we are on the same page. I meant we need to prevent the above cases.
Those cases were prevented by the removed code. I am asking how to prevent these cases without the below handling.
```
if ! grep -Fxq "$onie_platform" platforms_asic; then
        echo "The image you're trying to install is of a different ASIC type as the running platform's ASIC"
```

alexrallen: The code you show me as removing was relocated here
https://github.com/alexrallen/sonic-buildimage/blob/sonic_arm_grub/installer/install.sh#L99

git did not handle this refactor well, apologies for that. Let me know any other questions.

installer/install.sh
```diff
@@ -30,11 +59,11 @@ fi
...
 cd $(dirname $0)
 if [ -r ./machine.conf ]; then
-    . ./machine.conf
+    read_conf_file "./machine.conf"
 fi
 
-if [ -r ./onie-image-arm64.conf ]; then
-    . ./onie-image-arm64.conf
+if [ -r ./onie-image*.conf ]; then
+. ./onie-image*.conf
```
qiluo-msft, line 66: If there are multiple files matching the wildcard, only one is executed.

```
$ . a.sh
a
$ . b.sh
b
$ . *.sh
a
```

qiluo-msft: Seems there is a case of `onie-image.conf` and `onie-image-*.conf` both exist for one platform.

qiluo-msft: Still applicable.

xumia:
@alexrallen looks like one of the template variable (%%EXTRA_CMDLINE_LINUX%%) is not replaced. Is it relative to the change?

log:
```
EXTRA_CMDLINE_LINUX=
ONIE: Unable to find 'Part Number' TLV in EEPROM data.
Success: Support tarball created: /tmp/onie-support-kvm_x86_64.tar.bz2
Installing for i386-pc platform.
Installation finished. No error reported.
Switch CPU vendor is: GenuineIntel
Switch CPU cstates are: disabled
EXTRA_CMDLINE_LINUX=%%EXTRA_CMDLINE_LINUX%%
Installed SONiC base image SONiC-OS successfully
ONIE: NOS install successful: file://dev/vdb/onie-installer.bin
```

And Grub settings:
```
menuentry 'SONiC-OS-master-10946.127441-ab961b5e3' {
        search --no-floppy --label --set=root SONiC-OS
        echo    'Loading SONiC-OS OS kernel ...'
        insmod gzio
        if [ x = xxen ]; then insmod xzio; insmod lzopio; fi
        insmod part_msdos
        insmod ext2
        linux   /image-master-10946.127441-ab961b5e3/boot/vmlinuz-5.10.0-12-2-amd64 root=UUID=0f1b4b07-9742-426d-bd3a-a1e789401536 rw console=tty0 console=ttyS0,115200n8 quiet intel_idle.max_cstate=0 %%EXTRA_CMDLINE_LINUX%%                  net.ifnames=0 biosdevname=0                 loop=image-master-10946.127441-ab961b5e3/fs.squashfs loopfstype=squashfs                                       systemd.unified_cgroup_hierarchy=0                 apparmor=1 security=apparmor varlog_size=4096 usbcore.autosuspend=-1 
        echo    'Loading SONiC-OS OS initial ramdisk ...'
        initrd  /image-master-10946.127441-ab961b5e3/boot/initrd.img-5.10.0-12-2-amd64
}
```

installer/default_platform.conf
```diff
@@ -1,119 +1,5 @@
 #!/bin/sh
```
qiluo-msft, line 1: It is not ideal that we have .conf file with so complex bash code. And there are many variables need replacement, like `%%SONIC_ROOT%%`.

PR #11367, opened 2022-07-06
[caclmgrd]: Add infrastructure to support adding feature specific acls

SuvarnaMeenakshi:

#### Why I did it
Add infrastructure to support adding feature specific acls.
If feature specific ACLs has to be added:
```
if feature_name in self.feature_present and self.feature_present.get('feature_name'):
    add_feature_specific_acls()
```

#### How I did it
Add function to get features present in feature table.

#### How to verify it
unit-test passes.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [x] 202205

### DIFF
--- a/src/sonic-host-services/scripts/caclmgrd
+++ b/src/sonic-host-services/scripts/caclmgrd
@@ -29,6 +29,7 @@ SYSLOG_IDENTIFIER = "caclmgrd"
 
 DEFAULT_NAMESPACE = ''
 
+
 # ========================== Helper Functions =========================
 
 
--- a/src/sonic-host-services/scripts/caclmgrd
+++ b/src/sonic-host-services/scripts/caclmgrd
@@ -54,6 +54,7 @@ class ControlPlaneAclManager(daemon_base.DaemonBase):
     Attributes:
         config_db: Handle to Config Redis database via SwSS SDK
     """
+    FEATURE_TABLE = "FEATURE"
     ACL_TABLE = "ACL_TABLE"
     ACL_RULE = "ACL_RULE"
     DEVICE_METADATA_TABLE = "DEVICE_METADATA"
--- a/src/sonic-host-services/scripts/caclmgrd
+++ b/src/sonic-host-services/scripts/caclmgrd
@@ -29,9 +29,6 @@ SYSLOG_IDENTIFIER = "caclmgrd"
 
 DEFAULT_NAMESPACE = ''
 
-FEATURES_PRESENT = [""]
-
-
 # ========================== Helper Functions =========================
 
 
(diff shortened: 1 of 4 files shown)


src/sonic-host-services/scripts/caclmgrd
```diff
@@ -29,6 +29,8 @@ SYSLOG_IDENTIFIER = "caclmgrd"
 
 DEFAULT_NAMESPACE = ''
 
+FEATURES_PRESENT = [""]
```
SuvarnaMeenakshi, line 32: Remove this and udpate for all features from FEATURE_TABLE in config_db

src/sonic-host-services/scripts/caclmgrd
```diff
@@ -201,6 +206,12 @@ class ControlPlaneAclManager(daemon_base.DaemonBase):
         tcp_flags_str = tcp_flags_str[:-1]
         return tcp_flags_str
 
+    def update_feature_present(self):
```
qiluo-msft, line 209: >update_feature_present [](http://example.com/codeflow?start=8&length=22)

Could you add unit tests to cover new code?

src/sonic-host-services/scripts/caclmgrd
```diff
@@ -201,6 +206,12 @@ class ControlPlaneAclManager(daemon_base.DaemonBase):
...
+        feature_tb_info = self.config_db_map[DEFAULT_NAMESPACE].get_table(self.FEATURE_TABLE)
+        if feature_tb_info:
+            for k, v in feature_tb_info.items():
+                self.feature_present[k] = True
```
qiluo-msft, line 213: >feature_present [](http://example.com/codeflow?start=21&length=15)

Seems this is a generic oneshot cache. Suggest leverage common library like `cached_property`. ref: [https://www.geeksforgeeks.org/python-functools-cached_property/](https://www.geeksforgeeks.org/python-functools-cached_property/)

SuvarnaMeenakshi: thank you for the pointer. Currently, we are updating this feature_present map with the features that are present only once during initialization so it is not repeatedlyinvoked.
Also, if we use cached_property, we will invoke the function once, which could be in per-thread context, which might cause issues. So not using cached_property in this case.

src/sonic-host-services/scripts/caclmgrd
SuvarnaMeenakshi: Added unit-test and mock data for the unit-test.

qiluo-msft (approved):

abdosi (approved):
LGTM

PR #11368, opened 2022-07-06
Add support to generate /e/n/i when there are multiple MGMT_INTERFACE

geogchen:
#### Why I did it
Currently interfaces.j2 hardcodes to eth0 even when there are multiple interfaces in MGMT_INTERFACE.  This change adds support to generate /e/n/i when there are multiple interfaces in MGMT_INTERFACE.

#### How I did it
By removing hardcoded eth0 when looping through MGMT_INTERFACE.

#### How to verify it
Verified through unit test.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205

### DIFF
--- a/files/image_config/interfaces/interfaces.j2
+++ b/files/image_config/interfaces/interfaces.j2
@@ -33,6 +33,7 @@ iface lo inet loopback
 {% if (ZTP_DHCP_DISABLED is not defined) and (ZTP is defined) and (ZTP['mode'] is defined and ZTP['mode']['profile'] == 'active') %}
 auto eth0
 
+
 # ZTP out-of-band interface
 allow-hotplug eth0
 {% if ZTP['mode']['ipv4'] == 'true' %}
--- a/files/image_config/interfaces/interfaces.j2
+++ b/files/image_config/interfaces/interfaces.j2
@@ -30,9 +30,8 @@ iface lo inet loopback
 {% block mgmt_interface %}
 
 # The management network interface
-auto eth0
 {% if (ZTP_DHCP_DISABLED is not defined) and (ZTP is defined) and (ZTP['mode'] is defined and ZTP['mode']['profile'] == 'active') %}
-
+auto eth0
 
 # ZTP out-of-band interface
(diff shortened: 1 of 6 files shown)


ganglyu (approved):

files/image_config/interfaces/interfaces.j2
```diff
@@ -30,8 +30,8 @@ iface lo inet loopback
 {% block mgmt_interface %}
 
 # The management network interface
-auto eth0
 {% if (ZTP_DHCP_DISABLED is not defined) and (ZTP is defined) and (ZTP['mode'] is defined and ZTP['mode']['profile'] == 'active') %}
+auto eth0
```
qiluo-msft, line 34: >eth0 [](http://example.com/codeflow?start=5&length=4)

Just to confirm, you will not change this for multiple mgmt ports?

files/image_config/interfaces/interfaces.j2
```diff
@@ -64,8 +64,11 @@ iface {{ port }} inet6 dhcp
 
 {% else %}
 {% if MGMT_INTERFACE %}
+{% for (name, prefix) in MGMT_INTERFACE|pfx_filter|unique(attribute=0) %}
```
qiluo-msft, line 67: >unique [](http://example.com/codeflow?start=51&length=6)

Why one `for` has `unique`, but below `for` do not have? Just use one `for` block?

geogchen: if unique is not applied here, and each interface has an ipv4 and ipv6 address, then the interface will be repeated.
E.g., if we have
eth0|1.2.3.4
eth0|abcd::
eth1|2.3.4.5
eth1|cdef::

without unique, it would generate the following:
auto eth0
auto eth0
auto eth1
auto eth1

The loop below this does not need unique because ipv4 and ipv6 need to be iterated for the interface.

qiluo-msft (approved):

PR #11386, opened 2022-07-08
[ci] Fix azure pipeline test stage dependency issue.

liushilongbuaa:
#### Why I did it
When any of the test job failed in the test stage, the rerun will not work, the test stage will be skipped automaticall, so we do not have chance to rerun the test stage again, and the checks of the test will be always in failed status, block the PR to merge forever.

It should be caused by the condition in the Test stage, we should specify the status of the BuildVS stage.

#### How I did it
Fix stage dependency logic.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -77,7 +77,7 @@ stages:
 
 - stage: Test
   dependsOn: BuildVS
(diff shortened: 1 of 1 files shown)


xumia:
Merging the PR, bypass all the checks.
Justification: only azp change, it blocks PR to merge.

PR #10154, opened 2022-03-04
[sonic-cfggen]: Fix generated deployment_id

ganglyu:
Signed-off-by: Gang Lv ganglv@microsoft.com

#### Why I did it
Config db schema generated by minigraph can’t pass yang validation, deployment_id can’t be none for yang validation.

#### How I did it
Update minigraph.py, skip deployment_id with None value

#### How to verify it
Run UT for sonic-config-enginue.
Run command 'sonic-cfggen -m tests/multi_npu_data/sample-minigraph-noportchannel.xml -p tests/multi_npu_data/sample_port_config-3.ini -n asic3 --print-data'.

#### Which release branch to backport (provide reason below if selected)

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

#### Description for the changelog

Fix #9592

PR #11407, opened 2022-07-10
Added Support for deployment_id parsing for Device Asic metadata

abdosi:
What I did:
Added Support for deployment_id parsing for Device Asic metadata. 

Why I did:-
Deployment Id is used in BGP docker for FRR template generation. For multi-asic platforms running in namespace without deployment id  as key in DEVICE_METADATA FRR template generation fails. This change is needed after this https://github.com/Azure/sonic-buildimage/pull/10154 where if deployment_id is none we don't update DEVICE_METADA dictionary. 

How I verify:-
Added unit-test.

### DIFF
--- a/src/sonic-config-engine/tests/test_multinpu_cfggen.py
+++ b/src/sonic-config-engine/tests/test_multinpu_cfggen.py
@@ -304,6 +304,7 @@ class TestMultiNpuCfgGen(TestCase):
                 self.assertEqual(output['localhost']['sub_role'], 'FrontEnd')
             else:
                 self.assertEqual(output['localhost']['sub_role'], 'BackEnd')
+            self.assertEqual(output['localhost']['deployment_id'], "1")
 
     def test_global_asic_acl(self):
         argument = "-m {} -p {}  --var-json \"ACL_TABLE\"".format(self.sample_graph, self.sample_port_config)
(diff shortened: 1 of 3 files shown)


src/sonic-config-engine/tests/test_multinpu_cfggen.py
```diff
@@ -304,6 +304,7 @@ def test_device_asic_metadata(self):
                 self.assertEqual(output['localhost']['sub_role'], 'FrontEnd')
             else:
                 self.assertEqual(output['localhost']['sub_role'], 'BackEnd')
+            self.assertEqual(output['localhost']['deployment_id'], "1")
```
ganglyu, line 307: If there's no DeploymentId, what's the expected output?

abdosi: the behavior will be same as your PR: https://github.com/Azure/sonic-buildimage/pull/10154. It will be none and not get populated in result dictionary

ganglyu (approved):

abdosi:
@ganglyu  Can you reapprove this one: https://github.com/Azure/sonic-buildimage/pull/11454/

I am closing this PR since my source branch is not correct.

PR #11417, opened 2022-07-12
[build] don't require passwordless sudo (Closes #11292)

### LINKED ISSUE #11292: Builds now require password-less sudo access

_Originally posted by @jusherma in https://github.com/Azure/sonic-buildimage/issues/11111#issuecomment-1170621783_

jusherma:
#### Why I did it

Not all build environments have passwordless sudo enabled for all users

#### How I did it

Instead of using `sudo` to delete fsroot directories, mount them in a small, temporary docker container and delete them from there

#### How to verify it

Build in an environment where the build user does not have passwordless sudo enabled and confirm that no sudo password prompts are seen

Needs to be included in any branch that contains #11111 

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [x] 202106
- [x] 202111
- [x] 202205

Re-allow builds without passwordless sudo authentication

### DIFF
--- a/Makefile.work
+++ b/Makefile.work
@@ -188,7 +188,8 @@ endif
 DOCKER_LOCKFILE_SAVE := $(DOCKER_LOCKDIR)/docker_save.lock
 $(shell mkdir -m 0777 -p $(DOCKER_LOCKDIR))
 $(shell [ -f $(DOCKER_LOCKFILE_SAVE) ] || (touch $(DOCKER_LOCKFILE_SAVE) && chmod 0777 $(DOCKER_LOCKFILE_SAVE)))
-$(shell sudo rm -rf $(DOCKER_ROOT) && mkdir -p $(DOCKER_ROOT))
+$(docker run --rm -v $(DOCKER_ROOT)\:/mount alpine sh -c 'rm -rf /mount/')
+$(mkdir -p $(DOCKER_ROOT))
 
 ifeq ($(DOCKER_BUILDER_MOUNT),)
 override DOCKER_BUILDER_MOUNT := "$(PWD):/sonic"


jusherma:
@xumia can you please review?

jusherma:
@qiluo-msft @xumia @lguohan Please review. This PR is very small and has been open for 10 day nows

xumia (approved):

kv-y:
@jusherma @xumia 
This fix seems doesn't work correctly (I'm testing in Ubuntu 20.04 (docker 20.10.20)).
Yes, we don't need sudo anymore but it also doesn't remove fsroot directory.
You can try to execute this docker command manually and you get `Resource busy` error if you try to remove /mount in container.
Could you please review it again?

I think it should be something like this:
`$(shell docker run --rm -v $(PWD)\:$(PWD) alpine sh -c 'rm -rf $(DOCKER_ROOT)')`
`$(shell mkdir -p $(DOCKER_ROOT))`

Makefile.work
```diff
@@ -188,7 +188,8 @@ endif
 DOCKER_LOCKFILE_SAVE := $(DOCKER_LOCKDIR)/docker_save.lock
 $(shell mkdir -m 0777 -p $(DOCKER_LOCKDIR))
 $(shell [ -f $(DOCKER_LOCKFILE_SAVE) ] || (touch $(DOCKER_LOCKFILE_SAVE) && chmod 0777 $(DOCKER_LOCKFILE_SAVE)))
-$(shell sudo rm -rf $(DOCKER_ROOT) && mkdir -p $(DOCKER_ROOT))
+$(docker run --rm -v $(DOCKER_ROOT)\:/mount alpine sh -c 'rm -rf /mount/')
```
qiluo-msft, line 191: >alpine [](http://example.com/codeflow?start=44&length=6)

This docker image is unnecessary dependency.

PR #11458, opened 2022-07-15
[caclmgrd][dualtor] add iptables rule for dualtor gRPC to allow packets

vdahiya12:
getting forwarded from loopback IP

Signed-off-by: vaibhav-dahiya <vdahiya@microsoft.com>

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205

### DIFF
--- a/src/sonic-host-services/scripts/caclmgrd
+++ b/src/sonic-host-services/scripts/caclmgrd
@@ -56,6 +56,7 @@ class ControlPlaneAclManager(daemon_base.DaemonBase):
     ACL_RULE = "ACL_RULE"
     DEVICE_METADATA_TABLE = "DEVICE_METADATA"
     MUX_CABLE_TABLE = "MUX_CABLE_TABLE"
+    CONFIG_MUX_CABLE = "MUX_CABLE"
 
     ACL_TABLE_TYPE_CTRLPLANE = "CTRLPLANE"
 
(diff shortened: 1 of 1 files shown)


src/sonic-host-services/scripts/caclmgrd
```diff
@@ -273,6 +274,41 @@ class ControlPlaneAclManager(daemon_base.DaemonBase):
 
         return allow_internal_docker_ip_cmds
 
+    def get_loopback_ip_from_loopback_table(self, loopback_table):
```
prsunny, line 277: please provide "loopback3" as an argument so that this function can be re-used later

prsunny:
I see the master PR is closed. It should ideally go to master and cherry-pick to 202205.

src/sonic-host-services/scripts/caclmgrd
```diff
@@ -273,6 +274,42 @@ class ControlPlaneAclManager(daemon_base.DaemonBase):
 
         return allow_internal_docker_ip_cmds
 
+    def get_loopback_ip_from_loopback_table(self, loopback_table, loopback_name):
```
prsunny, line 277: i think the name is not aligned. I suggest define this as 'get_ip_from_loopback_table' and just pass loopback_name, if function is for loopback table. If not, rename to get_ip_from_interface_table' and pass table and name. Suggest to make it generic.

src/sonic-host-services/scripts/caclmgrd
```diff
@@ -273,6 +274,42 @@ class ControlPlaneAclManager(daemon_base.DaemonBase):
...
+            loopback_name = 'Loopback3'
+            loopback_address = self.get_loopback_ip_from_loopback_table(loopback_table, loopback_name)
+            if loopback_address is not None:
+                mux_table = self.config_db_map[DEFAULT_NAMESPACE].get_table(self.CONFIG_MUX_CABLE)
+                mux_table_keys = mux_table.keys()
+                for key in mux_table_keys:
+                    kvp = mux_table.get(key)
+                    if 'cable_type' in kvp and kvp['cable_type'] == 'active-active':
+                        fwd_dualtor_grpc_traffic_from_host_to_soc_cmds.append(self.iptables_cmd_ns_prefix[namespace] + 
+                                "iptables -t nat -D POSTROUTING --destination {} -j SNAT --to-source {}".format(kvp['soc_ipv4'], loopback_address))
```
prsunny, line 307: I wouldn't recommend deleting and reinstalling the entries as this may be just an overhead with so many rule.  Can we just do one flush of this?

prsunny (commented):
add unit tests.

src/sonic-host-services/scripts/caclmgrd
vdahiya12: fixed

src/sonic-host-services/scripts/caclmgrd
vdahiya12: Not sure if this is ok ?
```fwd_dualtor_grpc_traffic_from_host_to_soc_cmds.append(self.iptables_cmd_ns_prefix[namespace] + 
                    "iptables -t nat -D POSTROUTING --flush")```
I think the flush option only takes in a chain but does not take the source, destination etc options.

vdahiya12:
> add unit tests.

added

prsunny (approved):

PR #11465, opened 2022-07-18
Enable ARP Update Script for Packet based chassis.

abdosi:
What I did:
```
    Following chnages done for packet based chassis:-
    1> Run arp_update on LC's to resolve static route nexthops over backend
    port-channel interfaces.
    2> On Supervsior make sure arp_update exit gracefully
```

Why I did:
In Packet based chassis it's possible Static Route toward other  ASIC's/{LC's,ASIC's} via backed port-channel  nexthop might not be resolved (eg: if the peers is not ready) when route is added and OA starts Arp resolver. This causes the arp entry to go into failed state. Because of this Static Route and corresponding any e-BGP routes resolved over static route gets added by OA as TEMP ROUTE  to point to any one randomly picked resolved nexthop and not  to ECMP/Nexthop group. OA will then keep on trying to add route over the nexthop group/ecmp if possible.

Above retry causes high CPU Utilization by OA. Also since only one nexthop is programmed we are not utilizing ECMP over backend ASIC.

Adding arp_update logic make sure all nexthops goes in resolved/stale state (when all peers's are ready and in good state)  and once they are resolved Static Route and corresponding any e-BGP routes resolved over static route will get updated by OA to point to ECMP/Next-hop group.

Please note this change does not handle Backed-End Link and ASIC Failure Scenario. Those will be handle as Future Enhancements using BFD. 

How I tested:
Manual Verification. After adding the logic OA high cpu utilization  is gone and also all routes point to ECMP.


abdosi:
cc @anamehra

dockers/docker-orchagent/docker-init.j2
```diff
@@ -34,9 +34,12 @@ if [ -x /usr/share/sonic/hwsku/hwsku-init ]; then
     /usr/share/sonic/hwsku/hwsku-init
 fi
 
-# Start arp_update and NDP proxy daemon when VLAN exists
-if [ "$VLAN" != "" ]; then
+# Start arp update when VLAN exists or switch type as chassis packet for backend port channel interfaces
+if [[ "$VLAN" != "" ]] || [[ "$SWITCH_TYPE" == "chassis-packet" ]]; then
```
prsunny, line 38: may be we should have a separate check for "chassis-packet and copy the arp.conf. Let Vlan check be separate.

prsunny (approved):
lgtm, minor comment

abdosi:
> lgtm, minor comment

thanks @prsunny I will address it as part of another PR that I am working on.

PR #11469, opened 2022-07-18
Remove swsssdk from rules and image.

liuh-80:
#### Why I did it
To deprecate swsssdk, remove all dependency to it. 

#### How I did it
Remove swsssdk from rules and build image scripts.

#### How to verify it
Pass all UT and E2E test case

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205

Remove swsssdk from rules and build image scripts.


liuh-80:
Remove swsssdk from rules first, if UT passed, will remove from image later.

qiluo-msft (approved):

rules/docker-config-engine.mk
```diff
@@ -4,7 +4,6 @@ DOCKER_CONFIG_ENGINE = docker-config-engine.gz
 $(DOCKER_CONFIG_ENGINE)_PATH = $(DOCKERS_PATH)/docker-config-engine
 $(DOCKER_CONFIG_ENGINE)_DEPENDS += $(LIBSWSSCOMMON) \
                                    $(SONIC_DB_CLI)
-$(DOCKER_CONFIG_ENGINE)_PYTHON_WHEELS += $(SWSSSDK_PY2)
```
liuh-80, line 7: Database container depends on hredis
And after remove SWSSSDK_PY2, the indirect to hredis been removed, so some test case failed.

liuh-80:
This PR need wait for 3 vendor side change complete first.
Because this change will remove swsssdk from dockers, and if any vendor's code running in docker still suing swsssdk, their code will break.

liuh-80:
Will check if can move some change which does not depend on vender code to another PR.

liuh-80:
WIll fix merge conflict and update soon.

rules/docker-config-engine.mk
baxia-lan: Does this removal also deprecate the redis-dump-load functionality? 
What if users still want to use redis-dump-load?

liuh-80: > redis-dump-load
It's been porting to sonic_py_common:
https://github.com/sonic-net/sonic-buildimage/blob/b1102ad91344669072f7769baef693bd18365984/src/sonic-py-common/sonic_py_common/sonic_db_dump_load.py

mint570: Our container does not have redis-dump. We have include this change: https://github.com/sonic-net/sonic-buildimage/pull/12185

Is there other changes that we are missing?

mint570: Looks like it is missing the entry script:

cat /usr/local/bin/redis-dump
#!/usr/bin/python3
# -*- coding: utf-8 -*-
import re
import sys
from redisdl import main
if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw|\.exe)?$', '', sys.argv[0])
    sys.exit(main())

mint570: Looks like we are using sonic-db-dump instead of redis-dump now.
Are they basically the same thing?
Are we removing redis-dump?

mint570: I understand now. sonic-db-dump-load is built on top of redis-dump-load. So it needs redis-dump-load as dependency. Currently, for sonic_py_common python3 package, it does not list redis-dump-load as dependency. And hence sonic-db-dump-load doesn't work. This needs to be fixed.

liuh-80: @mint570 , here is the fix PR: https://github.com/sonic-net/sonic-buildimage/pull/14347

liuh-80: @mint570, may I know how you found this issue, some command failed or errro in syslog? I will create a sonic-mgmt test case to prevent this happen again, so need understand the break scenario.

mint570: I just called sonic-db-dump in the container, it failed with the redisdl import.
(Initially I was looking for redis-dump, but there is no such tool in the container.)

Issue #11472, opened 2022-07-18
monit errors in the logs  "ERR /memory_checker: Failed to retrieve the running container list from docker daemon"


liorghub:
#### Description
The following error appear in log during deinit flow:
ERR /memory_checker: Failed to retrieve the running container list from docker daemon! Error message is: 'Error while fetching server API version: ('Connection aborted.', FileNotFoundError(2, 'No such file or directory'))'

#### Steps to reproduce the issue:
This is issue is not deterministic. It happens when Monit plugin memory_checker that run periodically is running after reboot was perfomed. The plugin is trying to fetch the running containers but Docker engine was already stopped by systemd. 

#### Describe the results you received:
The following error appear in log:
ERR /memory_checker: Failed to retrieve the running container list from docker daemon! Error message is: 'Error while fetching server API version: ('Connection aborted.', FileNotFoundError(2, 'No such file or directory'))'

#### Describe the results you expected:
If Docker engine is not running, do not try to fetch running containers and return empty list.

#### Output of `show version`:
```
root@r-lionfish-14:/home/admin# show version

SONiC Software Version: SONiC.202012.121748-81f200fde
Distribution: Debian 10.12
Kernel: 4.19.0-12-2-amd64
Build commit: 81f200fde
Build date: Wed Jul 13 13:56:52 UTC 2022
Built by: AzDevOps@sonic-build-workers-001RQ0

Platform: x86_64-mlnx_msn3420-r0
HwSKU: ACS-MSN3420
ASIC: mellanox
ASIC Count: 1
Serial Number: MT2019X13879
Uptime: 09:01:14 up 0 min,  1 user,  load average: 2.14, 0.55, 0.19

Docker images:
REPOSITORY                    TAG                       IMAGE ID            SIZE
docker-syncd-mlnx             202012.121748-81f200fde   4ff0a131809c        859MB
docker-syncd-mlnx             latest                    4ff0a131809c        859MB
docker-sonic-mgmt-framework   202012.121748-81f200fde   ce220f832981        687MB
docker-sonic-mgmt-framework   latest                    ce220f832981        687MB
docker-sonic-telemetry        202012.121748-81f200fde   4fa7aae2af85        452MB
docker-sonic-telemetry        latest                    4fa7aae2af85        452MB
docker-teamd                  202012.121748-81f200fde   84db114155f5        373MB
docker-teamd                  latest                    84db114155f5        373MB
[...]

PR #11129, opened 2022-06-14
[memory_checker] Do not check memory usage of containers which are not created

yozhao101:
Signed-off-by: Yong Zhao <yozhao@microsoft.com>

#### Why I did it
This PR aims to fix an issue (https://github.com/Azure/sonic-buildimage/issues/10088) by enhancing the script `memory_checker`. 

Specifically, if container is not created successfully during device is booted/rebooted, then `memory_checker` do not need check its memory usage.

#### How I did it
In the script `memory_checker`, a function is added to get names of running containers. If the specified container name is not in current running container list, then this script will exit without checking its memory usage.

#### How to verify it
I tested on a lab device by following the steps:

1. Stops `telemetry` container with command `sudo systemctl stop telemetry.service`
2. Removes `telemetry` container with command `docker rm telemetry`
3. Checks whether the script `memory_checker` ran by Monit will generate the syslog message saying it will exit without checking memory usage of `telemetry`:
        
        Jun 14 15:42:08.580612 str-s6000-on-2 INFO /memory_checker: [memory_checker] Exits without checking memory usage since container 'telemetry' is not running!

#### Which release branch to backport (provide reason below if selected)

- [ ] 201811
- [x ] 201911
- [x ] 202006
- [ x] 202012
- [x ] 202106
- [x ] 202111

PR #11476, opened 2022-07-18
[memory_checker] Do not check memory usage of containers if docker daemon is not running

liorghub:
Fix in Monit memory_checker plugin. Skip fetching running containers if docker engine is down (can happen in deinit).
This PR fixes issue https://github.com/Azure/sonic-buildimage/issues/11472.

Signed-off-by: liora <liora@nvidia.com>

#### Why I did it
In the case where Monit runs during deinit flow, memory_checker plugin is fetching the running containers without checking if Docker service is still running. I added this check.

#### How I did it
Use systemctl is-active to check if Docker engine is still running.

#### How to verify it
Use systemctl to stop docker engine and reload Monit, no errors in log and relevant print appears in log.

The fix is required in 202205 and 202012 since the PR that introduced the issue was cherry picked to those branches (https://github.com/Azure/sonic-buildimage/pull/11129).

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012
- [ ] 202106
- [ ] 202111
- [x] 202205

### DIFF
--- a/files/image_config/monit/memory_checker
+++ b/files/image_config/monit/memory_checker
@@ -96,16 +96,16 @@ def check_memory_usage(container_name, threshold_value):
         sys.exit(4)
 
 
-def is_service_active(service):
+def is_service_active(service_name):
     """Test if service is running.
 
     Args:
-        command: A string contains the service name
+        service_name: A string contains the service name
 
     Returns:
         True if service is running, False otherwise
     """
-    status = subprocess.run("systemctl is-active --quiet {}".format(service), shell=True, check=False)
+    status = subprocess.run("systemctl is-active --quiet {}".format(service_name), shell=True, check=False)
     return status.returncode == 0
 
 
--- a/files/image_config/monit/memory_checker
+++ b/files/image_config/monit/memory_checker
@@ -24,7 +24,6 @@ import subprocess
 import sys
 import syslog
 import re
-import os
 
 import docker
 
@@ -97,9 +96,9 @@ def check_memory_usage(container_name, threshold_value):
         sys.exit(4)
 
 
-def is_service_active(feat):
-    status = os.system('systemctl is-active --quiet {}'.format(feat))
-    return status == 0
+def is_service_active(service):
+    status = subprocess.run("systemctl is-active --quiet {}".format(service), shell=True, check=False)
+    return status.returncode == 0
 
 
 def get_running_container_names():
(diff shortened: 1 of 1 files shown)


files/image_config/monit/memory_checker
```diff
@@ -96,6 +96,11 @@ def check_memory_usage(container_name, threshold_value):
         sys.exit(4)
 
 
+def is_service_active(service):
+    status = subprocess.run("systemctl is-active --quiet {}".format(service), shell=True, check=False)
```
yozhao101, line 100: Can we use the command `systemctl is-active docker` to check whether the `docker daemon` is active or not? Then we can ues the existing function at line 31 `get_command_result(...)`.

yozhao101: Can we also add comments for this function to explain why we need this function?

files/image_config/monit/memory_checker
```diff
@@ -128,6 +133,10 @@ def main():
     parser.add_argument("threshold_value", type=int, help="threshold value in bytes")
     args = parser.parse_args()
 
+    if not is_service_active("docker"):
+        syslog.syslog(syslog.LOG_INFO,"Skipping memory check: docker daemon is down")
```
yozhao101, line 137: Can we rephrase the syslog message as following?

`syslog.syslog(syslog.LOG_INFO, "[memory_checker] Exits without checking memory usage of container '{}' since docker daemon is not running!".format(args.container_name))`

files/image_config/monit/memory_checker
liorghub: I wanted to use get_command_result() but couldnt from the following reasons:
1. It returns only the command output. I do not need the command output, I need to test the command return code.
2. If command return code is not 0, it fails with error to log and sys.exit(1). In my case return code not 0 means that docker daemon is not running which is a valid flow.

I added function documentation.

files/image_config/monit/memory_checker
liorghub: Done

files/image_config/monit/memory_checker
```diff
@@ -96,6 +96,19 @@ def check_memory_usage(container_name, threshold_value):
         sys.exit(4)
 
 
+def is_service_active(service):
+    """Test if service is running.
+
+    Args:
+        command: A string contains the service name
```
yozhao101, line 103: `command` should be `service_name`, right?

files/image_config/monit/memory_checker
```diff
@@ -96,6 +96,19 @@ def check_memory_usage(container_name, threshold_value):
         sys.exit(4)
 
 
+def is_service_active(service):
```
yozhao101, line 99: `service_name` would be better?

files/image_config/monit/memory_checker
liorghub: Done.

files/image_config/monit/memory_checker
liorghub: Done.

yozhao101:
@qiluo-msft Can you please help review this PR? Thanks.

dprital:
@yozhao101  , @qiluo-msft  - now that all the comments were addressed, can you please review and approve ?
Thanks in advance.

yozhao101 (approved):

PR #11483, opened 2022-07-19
[YANG] Add constraint for loopback interface name formatting

isabelmsft:
#### Why I did it
Loopback interface YANG model was missing the constraint for Loobback interface name formatting (should be LoopbackXXXX)

#### How I did it
Add constraint to YANG model

#### How to verify it
Try configuring Loopback interface name to valid and invalid names

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205

### DIFF
--- a/src/sonic-yang-models/yang-models/sonic-loopback-interface.yang
+++ b/src/sonic-yang-models/yang-models/sonic-loopback-interface.yang
@@ -35,8 +35,11 @@ module sonic-loopback-interface {
             list LOOPBACK_INTERFACE_LIST {
                 key "name";
 
-                leaf name{
-                    type string;
+                leaf name {
+                    type string {
+                        length 1..128;
+                        pattern 'Loopback[0-9]{1,3}';
+                    };
                 }
 
                 leaf vrf_name {


src/sonic-yang-models/yang-models/sonic-loopback-interface.yang
```diff
@@ -35,8 +35,11 @@ module sonic-loopback-interface {
             list LOOPBACK_INTERFACE_LIST {
                 key "name";
 
-                leaf name{
-                    type string;
+                leaf name {
+                    type string {
+                        length 1..128;
+                        pattern 'Loopback[0-9]{1,3}';
```
qiluo-msft, line 41: Please add unit test cases for valid/invalid usage?

src/sonic-yang-models/yang-models/sonic-loopback-interface.yang
```diff
@@ -35,8 +35,11 @@ module sonic-loopback-interface {
             list LOOPBACK_INTERFACE_LIST {
                 key "name";
 
-                leaf name{
-                    type string;
+                leaf name {
+                    type string {
+                        length 1..128;
```
ganglyu, line 40: you have defined pattern, is length still necessary?

src/sonic-yang-models/yang-models/sonic-loopback-interface.yang
qiluo-msft: Please add the related checking code into PR description.

ganglyu:
buildimage pipeline failed, please build target or run unit test before commit.

PR #11506, opened 2022-07-21
Replace string token function in saibcm-modules

maipbui:
#### Why I did it
Remove insecure function `strtok()`
- `strtok` instantly alternates the first argument buffer, permanently deleting the delimiter character.
- Thread insecurity when there are two or more threads use this function.
#### How I did it
Replace `strtok()` by `strtok_r()` function - reentrant version strtok(), thread safe.
Use `strdup()` - to make a copy and avoid the function changes the first argument buffer.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205


platform/broadcom/saibcm-modules/systems/bde/linux/kernel/linux-kernel-bde.c
```diff
@@ -3199,12 +3199,10 @@ _init(void)
         unsigned int eb_ba = 0x0;
 
         gprintk("EB bus info: %s\n", eb_bus);
-        tok = strtok(eb_bus,",");
-        while (tok) {
+        while ((tok = strtok_r(eb_bus, ",", &eb_bus))) {
```
qiluo-msft, line 3203: >eb_bus [](http://example.com/codeflow?start=45&length=6)

You need provide a temporary pointer to store the output parameter, and it could not be `eb_bus`.
ref: [https://stackoverflow.com/a/15961298/2514803](https://stackoverflow.com/a/15961298/2514803)

qiluo-msft:
> #include <gmodule.h>

As with all bounds-checked functions, strtok_s is only guaranteed to be available if __STDC_LIB_EXT1__ is defined by the implementation and if the user defines __STDC_WANT_LIB_EXT1__ to the integer constant 1 before including <string.h>.

ref: [https://en.cppreference.com/w/c/string/byte/strtok](https://en.cppreference.com/w/c/string/byte/strtok)

---
Refers to: platform/broadcom/saibcm-modules/systems/bde/linux/kernel/linux-kernel-bde.c:21 in 7bb265b. [](commit_id = 7bb265b3f0f9e44e76eb5acead00b1f240e3f919, deletion_comment = False)

platform/broadcom/saibcm-modules/systems/bde/linux/kernel/linux-kernel-bde.c
maipbui: Fixed.

maipbui:
> > #include <gmodule.h>
> 
> As with all bounds-checked functions, strtok_s is only guaranteed to be available if **STDC_LIB_EXT1** is defined by the implementation and if the user defines **STDC_WANT_LIB_EXT1** to the integer constant 1 before including <string.h>.
> 
> ref: https://en.cppreference.com/w/c/string/byte/strtok
> 
> Refers to: platform/broadcom/saibcm-modules/systems/bde/linux/kernel/linux-kernel-bde.c:21 in [7bb265b](https://github.com/Azure/sonic-buildimage/commit/7bb265b3f0f9e44e76eb5acead00b1f240e3f919). [](commit_id = [7bb265b](https://github.com/Azure/sonic-buildimage/commit/7bb265b3f0f9e44e76eb5acead00b1f240e3f919), deletion_comment = False)

I'm using `strtok_r()`, (ref: [https://linux.die.net/man/3/strtok_r](https://linux.die.net/man/3/strtok_r)), which is recommended by the static code analysis tool. 

I deleted the declaration of `strtok()` function and added the declaration of `strtok_r()`. Refs: I got the `strtok_r()` declaration from [https://code.woboq.org/userspace/glibc/string/strtok_r.c.html](https://code.woboq.org/userspace/glibc/string/strtok_r.c.html) and [https://zhuanlan.zhihu.com/p/399398742](https://zhuanlan.zhihu.com/p/399398742)

PR #11571, opened 2022-07-28
[Mellanox] Configure buffer for queue 7 on T1 uplink ports in dual ToR scenario

stephenxs:
#### Why I did it

Configure buffer for queue 7 on T1 uplink ports in dual ToR scenario
On uplink ports `DSCP 48` => `TC 7` => `queue 7`

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [x] 202205

### DIFF
--- a/device/mellanox/x86_64-mlnx_msn2700-r0/Mellanox-SN2700-D48C8/buffers_defaults_objects.j2
+++ b/device/mellanox/x86_64-mlnx_msn2700-r0/Mellanox-SN2700-D48C8/buffers_defaults_objects.j2
@@ -143,7 +143,11 @@
 {% endfor %}
 {% for port in port_names_active.split(',') %}
 {% if port not in port_names_extra_queues.split(',') %}
+{% if port_names_extra_queues|length > 0 %}
+        "{{ port }}|5-7": {
+{% else %}
         "{{ port }}|5-6": {
+{% endif %}
             "profile" : "q_lossy_profile"
         }{% if not loop.last %},{% endif %}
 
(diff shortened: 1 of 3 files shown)


device/mellanox/x86_64-mlnx_msn2700-r0/Mellanox-SN2700-D48C8/buffers_defaults_objects.j2
```diff
@@ -143,7 +143,11 @@
 {% endfor %}
 {% for port in port_names_active.split(',') %}
 {% if port not in port_names_extra_queues.split(',') %}
+{% if port_names_extra_queues|length > 0 %}
+        "{{ port }}|5-7": {
+{% else %}
         "{{ port }}|5-6": {
+{% endif %}
```
neethajohn, line 150: Why do we need this check? for both uplinks and downlinks, dscp 48->tc 7-> q7. so we need q7 in both cases

stephenxs: @neethajohn 
This is not to check whether it’s uplink or downlinks. It’s to check whether it’s the dual ToR scenario.
- For dual tor uplink, we need 5-7.
- For non dual tor scenario, we still use 5-6
- For dual tor down link, we have 5, 6, and 7 in which 6 is lossless and 5, 7 are lossy. It is not defined here

You can check the update in example output as well to have a better understanding.

neethajohn (approved):

bingwang-ms (approved):

PR #11572, opened 2022-07-28
[Bug][Build] Fix the variable patterns not replaced issue

xumia:
#### Why I did it

The %%EXTRA_CMDLINE_LINUX%% is not replaced to the real value, it has impact on the kernel parameter settings.
See the log sonic-vs.img.gz.log in the latest master build. In the grub.cfg, the %%EXTRA_CMDLINE_LINUX%% is set in the linux command line.
```
Installing for i386-pc platform.
Installation finished. No error reported.
Switch CPU vendor is: GenuineIntel
Switch CPU cstates are: disabled
EXTRA_CMDLINE_LINUX=%%EXTRA_CMDLINE_LINUX%%
Installed SONiC base image SONiC-OS successfully
ONIE: NOS install successful: file://dev/vdb/onie-installer.bin
```

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205

### DIFF
--- a/installer/default_platform.conf
+++ b/installer/default_platform.conf
@@ -471,7 +471,6 @@ EOF
     fi
 
 # Add extra linux command line
-extra_cmdline_linux=%%EXTRA_CMDLINE_LINUX%%
 echo "EXTRA_CMDLINE_LINUX=$extra_cmdline_linux"
 GRUB_CMDLINE_LINUX="$GRUB_CMDLINE_LINUX $extra_cmdline_linux"
 
--- a/installer/default_platform.conf
+++ b/installer/default_platform.conf
@@ -64,7 +64,6 @@ create_partition()
         # determine ONIE partition type
         onie_partition_type=$(${onie_bin} onie-sysinfo -t)
         # demo partition size in MB
-        demo_part_size="%%ONIE_IMAGE_PART_SIZE%%"
         if [ "$firmware" = "uefi" ] ; then
             create_demo_uefi_partition $blk_dev
         elif [ "$onie_partition_type" = "gpt" ] ; then
--- a/installer/install.sh
+++ b/installer/install.sh
@@ -236,6 +236,9 @@ if [ "$install_env" = "onie" ]; then
     fi
 fi
 
+demo_part_size="%%ONIE_IMAGE_PART_SIZE%%"
+echo "ONIE_IMAGE_PART_SIZE=$demo_part_size"
+
 extra_cmdline_linux=%%EXTRA_CMDLINE_LINUX%%
 echo "EXTRA_CMDLINE_LINUX=$extra_cmdline_linux"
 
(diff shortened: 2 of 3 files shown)


xumia:
@alexrallen, could you please take a look? Thanks.

alexrallen (changes requested):
It looks like when I refactored some of this code in `install.sh` outside of that file I failed to consider this find/replace operation we were performing. Thanks for catching this. 

I would prefer from an cleanliness perspective to reduce the amount of `sed` processing that we do here. 

I suggest we setup the setup the variables in `install.sh` and then `default_installer.conf` and any scripts called by the installer will inherit variables without having to add this regex code to onie-mk each time (and prevent bugs like this in the future)

An example of this would be removing the following line from `default_platform.conf`:
https://github.com/Azure/sonic-buildimage/blob/master/installer/default_platform.conf#L474
So that this line in `install.sh` is able to pass the value down to `default_platform.conf`
https://github.com/Azure/sonic-buildimage/blob/master/installer/install.sh#L239

In addition it seems there are three instances of having these templates in `default_platform.conf` 
https://github.com/Azure/sonic-buildimage/blob/master/installer/default_platform.conf#L474
https://github.com/Azure/sonic-buildimage/blob/master/installer/default_platform.conf#L463
https://github.com/Azure/sonic-buildimage/blob/master/installer/default_platform.conf#L67

We need to make sure all three of these instances are handled (I don't think the `%%SONIC_ROOT%%` is handled in this patch.

Let me know if there are any questions.

xumia:
> SONIC_ROOT

@alexrallen, thanks for your comment, I will change it. Do you know what value should be set for SONIC_ROOT?

xumia:
For %%SONIC_ROOT%%, looks like it is not used at all, I do not fix SONIC_ROOT in this PR.
We support 4 image types: onie, raw, kvm, aboot. The first 3 image types are use onie (calling generate_onie_installer_image), aboot use files/Aboot/boot0.j2, not relative to install.sh.
@alexrallen, do you think it is good or not?

xumia:
@alexrallen, could you please help review it again? Thanks.

xumia:
@dgsudharsan,  could you please take a look the PR?

alexrallen (approved):
This looks fine to me. Thanks @xumia

installer/default_platform.conf
```diff
@@ -471,7 +470,6 @@ EOF
     fi
 
 # Add extra linux command line
-extra_cmdline_linux=%%EXTRA_CMDLINE_LINUX%%
```
qiluo-msft, line 474: >EXTRA_CMDLINE_LINUX [](http://example.com/codeflow?start=22&length=19)

The 3 lines were added in one commit, it makes sense to remove all 3 lines together.

qiluo-msft (approved):

Issue #11381, opened 2022-07-07
[VS] ARP packets are being dropped on the VS Image When used as a L2 device


vivekrnv:
#### Steps to reproduce the issue:
1. Topology: HA <-> Eth0, HB <-> Eth4
2. config vlan add 10
3. config vlan member add 10 Ethernet0 -u
4. config vlan member add 10 Ethernet4 -u
5. Assign IP addresses in the same network to HA & HB
6. Ping HB from HA

#### Describe the results you received:
- ARP request is being dropped by the VS
- This is because of the ebtable rules configured.

These rules might be required on the real HW but these shouldn't be there for VS
```
-A FORWARD -p ARP -j DROP
-A FORWARD -p 802_1Q --vlan-encap ARP -j DROP
```

#### Describe the results you expected:
ARP shouldn't be dropped

#### Work-Around:

To allow ARP frames forwarded via the switch, please, remove corresponding ebtables entries:
```
sudo ebtables -D FORWARD -p 802_1Q --vlan-encap ARP -j DROP
sudo ebtables -D FORWARD -p ARP -j DROP
```

prsunny:
This can be fixed by bypassing for the KVM image. Can be checked for platform in [build_debian.sh](https://github.com/Azure/sonic-buildimage/blob/master/build_debian.sh#L561) or [docker_image_ctl.j2](https://github.com/Azure/sonic-buildimage/blob/master/files/build_templates/docker_image_ctl.j2#L193). Request help from community.

PR #11585, opened 2022-07-29
[KVM]Preventing ebtables cfg to be applied on KVM

dgsudharsan:
#### Why I did it
To fix https://github.com/Azure/sonic-buildimage/issues/11381

Preventing ebtables rules to be applied on KVM image. The ebtables rules in  SONiC are added to prevent ARP as well as L2 forwarding to be blocked in linux kernel since the hardware will take care of the actual L2 forward. However this is not the case with KVM where linux needs to forward even L2 packets

#### How I did it
Added check in docker_image_ctl.j2 to apply the ebtables rules only when platform is not vs which is set for KVM

#### How to verify it
Loaded KVM image built with the changes and verified ebtables are not present

Before
```
root@sonic:~# ebtables -L
Bridge table: filter

Bridge chain: INPUT, entries: 0, policy: ACCEPT

Bridge chain: FORWARD, entries: 3, policy: ACCEPT
-d BGA -j DROP
-p ARP -j DROP
-p 802_1Q --vlan-encap ARP -j DROP

Bridge chain: OUTPUT, entries: 0, policy: ACCEPT
root@sonic:~#
```

After
```
ebtables -L
Bridge table: filter

Bridge chain: INPUT, entries: 0, policy: ACCEPT

Bridge chain: FORWARD, entries: 0, policy: ACCEPT

Bridge chain: OUTPUT, entries: 0, policy: ACCEPT
```

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205

### DIFF
--- a/files/build_templates/docker_image_ctl.j2
+++ b/files/build_templates/docker_image_ctl.j2
@@ -190,8 +190,9 @@ function postStartAction()
        fi
     fi
     # Setup ebtables configuration
+{%- if sonic_asic_platform != "vs" %}
     ebtables_config
-
+{%- endif %}
     # chassisdb starts before database starts, bypass the PING check since other
     # databases are not availbale until database container is ready.
     # also chassisdb doesn't support warm/fast reboot, its dump.rdb is deleted


volodymyrsamotiy (approved):

prsunny (approved):

dgsudharsan:
@prsunny To resolve the sonic-mgmt failure we require this PR https://github.com/sonic-net/sonic-mgmt/pull/6073

PR #11590, opened 2022-08-01
[swss.sh/syncd.sh] Trap only on EXIT

stepanblyschak:
When using trap on SIGTERM the script will not react to the SIGTERM signal sent while a child is executing.
I.e, the following script does not react on SIGTERM sent to it if it is
waiting for sleep to finish:

```
#!/bin/bash
trap "echo Handled SIGTERM" 0 2 3 15

echo "Before sleep"
sleep inf
echo "After sleep"
```

On SIGTERM sent to the followig script:

```
+ trap 'echo Handled SIGTERM' 15
+ echo Before sleep
Before sleep
+ sleep inf

Terminated
++ echo Handled SIGTERM
Handled SIGTERM
+ echo After sleep
After sleep
```

"echo After sleep" shouldn't be printed, the script has to exit.

Instead, trap only on EXIT which covers also a scenario with exit on
SIGINT, SIGTERM.

#### Why I did it

When NVIDIA switch is first booting FW upgrade process starts, if user does ```ztp disable -y``` during that process, which calls ```config reload -y -f``` all services will be terminated, including syncd and running FW upgrade process, however, it is observed that syncd.sh does not terminate but continues loading kernel drivers. This will cause an issue on next syncd start leading to switch init failure.

#### How I did it

Changed the signal for trap.

#### How to verify it

Boot NVIDIA switch first time, do "ztp disable -y", verify after some time all services are started and running.

- [ ] 201811
- [x] 201911
- [x] 202006
- [x] 202012
- [x] 202106
- [x] 202111
- [x] 202205

### DIFF
--- a/files/scripts/swss.sh
+++ b/files/scripts/swss.sh
@@ -38,7 +38,7 @@ function lock_service_state_change()
 
     exec {LOCKFD}>${LOCKFILE}
     /usr/bin/flock -x ${LOCKFD}
-    trap "/usr/bin/flock -u ${LOCKFD}" 0 2 3 15
+    trap "/usr/bin/flock -u ${LOCKFD}" EXIT
 
     debug "Locked ${LOCKFILE} (${LOCKFD}) from ${SERVICE}$DEV service"
 }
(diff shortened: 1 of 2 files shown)


liat-grozovik:
@prsunny  could you please help to review or assing someone?

liat-grozovik:
@yxieca , @prsunny please help to review.
i added label only to the main supported branches. this is the prio. it can be taken to other branches as well as needed.

yxieca (approved):

PR #11278, opened 2022-06-28
[Mellanox] Add arch folder to SDK binary location

alexrallen:
#### Why I did it

This is for the eventual support of multiple architectures for the mellanox platform. 

#### How I did it

Change the location of the binaries in `Switch-SDK-drivers` so that the path specifies the target architecture in addition to the target distribution that the debians are built for. 

This is the most straightforward way to separate binaries built against different architectures and selectively target them for installation in the mellanox SONiC image. 

#### How to verify it

Build SONiC for mellanox and verify it compiles successfully. 

#### Which release branch to backport (provide reason below if selected)

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012
- [ ] 202106
- [x] 202111
- [x] 202205

#### Description for the changelog
[mellanox] Add arch folder to SDK binary location

PR #11598, opened 2022-08-01
[Mellanox] [202111] Add arch folder to SDK binary location (#11278)

vivekrnv:
- Why I did it
Backport https://github.com/Azure/sonic-buildimage/pull/11278 to 202111 .
This is for the eventual support of multiple architectures for the mellanox platform.

- How I did it
Change the location of the binaries in Switch-SDK-drivers so that the path specifies the target architecture in addition to the target distribution that the debians are built for.

This is the most straightforward way to separate binaries built against different architectures and selectively target them for installation in the mellanox SONiC image.

- How to verify it
Build SONiC for mellanox and verify it compiles successfully.


dgsudharsan (approved):

dgsudharsan:
@liat-grozovik  Can we merge this?

liat-grozovik (approved):

PR #11616, opened 2022-08-04
parse 400g zr port config from minigraph

arlakshm:
Signed-off-by: Arvindsrinivasan Lakshmi Narasimhan <arlakshm@microsoft.com>

#### Why I did it
Generate the port configuration required 400G ZR port from minigraph.

#### How I did it
Add parse logic to get `tx_power` and `laser_freq` from `LinkMetadata` section of the minigraph.
Add UT for packet-chassis and voq chassis

#### How to verify it
UT

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205

### DIFF
--- a/src/sonic-config-engine/minigraph.py
+++ b/src/sonic-config-engine/minigraph.py
@@ -973,7 +973,8 @@ def parse_linkmeta(meta, hname):
         lower_tor_hostname = ''
         auto_negotiation = None
         macsec_enabled = False
-
+        tx_power = None
+        laser_freq = None
         properties = linkmeta.find(str(QName(ns1, "Properties")))
         for device_property in properties.findall(str(QName(ns1, "DeviceProperty"))):
             name = device_property.find(str(QName(ns1, "Name"))).text
@@ -990,6 +991,10 @@ def parse_linkmeta(meta, hname):
                 auto_negotiation = value
             elif name == "MacSecEnabled":
                 macsec_enabled = value
+            elif name == "TxPower":
+                tx_power = value
+            elif name == "Frequency":
+                laser_freq = value
 
         linkmetas[port] = {}
         if fec_disabled:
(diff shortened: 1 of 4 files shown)


src/sonic-config-engine/minigraph.py
```diff
@@ -1003,6 +1008,11 @@ def parse_linkmeta(meta, hname):
             linkmetas[port]["AutoNegotiation"] = auto_negotiation
         if macsec_enabled:
             linkmetas[port]["MacSecEnabled"] = macsec_enabled
+        if tx_power:
+            linkmetas[port]["tx_power"] = tx_power
```
prgeor, line 1012: tx_power can be float, so is that taken care?

src/sonic-config-engine/tests/sample-voq-graph.xml
```diff
@@ -79,6 +79,20 @@
...
+      <a:LinkMetadata>
+        <a:Name i:nil="true"/>
+        <a:Properties>
+          <a:DeviceProperty>
+            <a:Name>Frequency</a:Name>
+            <a:Value>195.875</a:Value>
+          </a:DeviceProperty>
+          <a:DeviceProperty>
+            <a:Name>TxPower</a:Name>
+            <a:Value>-10</a:Value>
```
prgeor, line 91: can we have one test case for float value? like -10.3? see https://github.com/sonic-net/sonic-buildimage/pull/11053

src/sonic-config-engine/minigraph.py
arlakshm: Yes, this is taken care

src/sonic-config-engine/tests/sample-voq-graph.xml
arlakshm: updated the test to test this

prgeor (approved):

PR #11630, opened 2022-08-05
Support TC value above 7 in yang model

bingwang-ms:
#### Why I did it
This PR is to update YANG model to support TC value above `7`.
Since PR https://github.com/sonic-net/sonic-buildimage/pull/10936 , we will have `TC 8` in some platform. Hence the YANG models are required to be updated to support TC value larger than `8`. 

#### How I did it
Update YANG model to support more values for TC. The new range is `[0,15]`

#### How to verify it
Verified by UT.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205

Support TC value above 7 in yang model


wen587 (approved):
LGTM

qiluo-msft (approved):

ganglyu (approved):

PR #11650, opened 2022-08-08
[write_standby] update write_standby.py script

yxieca:
#### Why I did it
The initial value has to be present for the state machines to work. In active-standby dual-tor scenario, or any hardware mux scenario, the value will be updtaed eventually with a delay.

However, in active-active dual-tor scenario, there is no other mechanism to initialize the value and get state machines started.
So this script will have to write something at start up time.

For active-active dualtor, 'active' is a more preferred initial value, the state machine will switch the state to standby soon if
link prober found link not in good state.

#### How I did it
Update the script to always provide initial values.

#### How to verify it
Tested on active-active dual-tor testbed.

### DIFF
--- a/files/scripts/write_standby.py
+++ b/files/scripts/write_standby.py
@@ -96,14 +96,21 @@ class MuxStateWriter(object):
 
         return status and value == 'true' 
 
-    def get_auto_mux_intfs(self):
+    def get_all_mux_intfs_modes(self):
         """
-        Returns a list of all mux cable interfaces that are configured to auto-switch
+        Returns a list of all mux cable interfaces, with suggested modes
+        Setting mux initial modes is crucial to kick off the statemachines,
+        have to set the modes for all mux/gRPC ports.
         """
+        intf_modes = {}
         all_intfs = self.config_db.get_table('MUX_CABLE')
-        auto_intfs = [intf for intf, status in all_intfs.items()
-                                if status['state'].lower() == 'auto']
-        return auto_intfs
+        for intf, status in all_intfs.items():
+            state = status['state'].lower()
+            if state == 'standby':
+                intf_modes[intf] = 'standby'
+            elif state in ['active', 'auto', 'manual']:
+                intf_modes[intf] = 'active'
+        return intf_modes
 
     def tunnel_exists(self):
         """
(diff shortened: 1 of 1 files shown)


zjswhhh (approved):
LGTM.

files/scripts/write_standby.py
```diff
@@ -96,14 +96,21 @@ def is_warmrestart(self):
...
         all_intfs = self.config_db.get_table('MUX_CABLE')
-        auto_intfs = [intf for intf, status in all_intfs.items()
-                                if status['state'].lower() == 'auto']
-        return auto_intfs
+        for intf, status in all_intfs.items():
+            state = status['state'].lower()
+            if state == 'standby':
+                intf_modes[intf] = 'standby'
+            elif state in ['active', 'auto', 'manual']:
+                intf_modes[intf] = 'active'
```
prsunny, line 112: how does this behave for y-cable dual-tor?

yxieca: updated. thanks.

files/scripts/write_standby.py
```diff
@@ -144,19 +157,23 @@ def apply_mux_config(self):
...
             logger.log_error("Timed out waiting for tunnel {}, mux state will not be written".format(self.tunnel_name))
 
 
 if __name__ == '__main__':
-    mux_writer = MuxStateWriter()
+    parser = argparse.ArgumentParser(description='Write initial mux state')
+    parser.add_argument('-s', '--state',
+                        help='state: intial state for "auto" and/or "manual" config, default "standby"',
+                        type=str, required=False, default='standby')
+    args = parser.parse_args()
```
prsunny, line 177: i think we've few places where write_standby script is used/planned to use.  So just want to confirm that we tested the case if no param given, it is default to 'standby'.

prsunny (approved):
lgtm, @theasianpianist please review/sign-off.

files/scripts/write_standby.py
yxieca: yes. I tested that.

theasianpianist (approved):
lgtm! Wondering if with these changes it would be appropriate to rename this script to something like `write_mux_state`?

PR #11651, opened 2022-08-08
[Bgpcfgd] Enhance add_peer/add_peer_ipv6 unit tests

geogchen:
#### Why I did it
The current input to add_peer/add_peer_ipv6 is admin status change, update the UT to supply new peer information.
Current UT does not check for case when check_neig_meta is true, update UT to check for this case

#### How I did it
By changing the input to add_peer/add_peer_ipv6
By modifying load_constants/constructor to take constants path as an input, and add two UT that uses a version of constants.yml that sets check_neig_meta to true.

#### How to verify it
UT failing before the change, and passing after the change.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205


files/image_config/constants/constants_use_neig.yml
```diff
@@ -0,0 +1,60 @@
+constants:
```
StormLiangMS, line 1: @geogchen This UT looks good to me, but adding one new file under this folder maybe not suitable, we use one unified constants.yml for all platforms, could we mock the data in the testcases other than add a new file in this folder just for different parameter changes?
And there is an internal repo https://msazure.visualstudio.com/One/_git/Networking-acs-buildimage, I checked the constants.yml, which has use_neighbors_meta set to true, maybe you can just keep the UT part and when it synced into internal, you can run the UT with internal repo with the value "ture" for use_neighbors_meta.

StormLiangMS: @geogchen Suggest to create a folder under src/sonic-bgpcfgd/tests/data, to hold variant constants.yaml files for test purpose.

src/sonic-bgpcfgd/tests/test_bgp.py
```diff
@@ -38,75 +46,89 @@ def constructor():
...
+        assert res, "Expect True return value for peer update"
+        mocked_log_err.assert_called_with("Peer 'default|10.10.10.1': Can't update the peer. It has wrong attribute value attr['admin_status'] = 'invalid'")
 
 def test_add_peer():
-    m = constructor()
-    res = m.set_handler("30.30.30.1", {"local_addr": "30.30.30.30", "admin_status": "up"})
-    assert res, "Expect True return value"
+    for constant in load_constant_files():
+        m = constructor(constant)
+        res = m.set_handler("30.30.30.1", {'asn': '65200', 'holdtime': '180', 'keepalive': '60', 'local_addr': '30.30.30.30', 'name': 'TOR', 'nhopself': '0', 'rrclient': '0'})
```
StormLiangMS, line 97: @geogchen this one "admin_status": "up" is missing, what's the reason we don't have this in the add peer test?

src/sonic-bgpcfgd/tests/test_bgp.py
```diff
@@ -4,15 +4,23 @@
...
+from .util import load_constants, CONSTANTS_PATH
 from swsscommon import swsscommon
 import bgpcfgd.managers_bgp
 
 TEMPLATE_PATH = os.path.abspath('../../dockers/docker-fpm-frr/frr')
 
-def constructor():
+def load_constant_files():
+    path = "tests/data/constants"
+    constant_files = [os.path.abspath(os.path.join(path, name)) for name in os.listdir(path)
```
StormLiangMS, line 15: @geogchen 
For the constants file which is included in the image, I think we should also include it in the testing for the verification, otherwise, the illegal change in the files/image_config/constants/constants.yml won't be catched up by the UT.

src/sonic-bgpcfgd/tests/test_bgp.py
geogchen: when I did trace over the add_peer call, I did not see admin status included.

src/sonic-bgpcfgd/tests/test_bgp.py
geogchen: Changed to include constants.yml under image_config/constants

files/image_config/constants/constants_use_neig.yml
geogchen: added a constants folder under tests/data, and moved the variant constants.yml to this new folder.

StormLiangMS (approved):
LGTM

PR #11725, opened 2022-08-12
[chassis] Configure midplane network on supervisor

mlorrillere:
#### Why I did it
Multi-asic Docker instances are created behind Docker's default bridge
which doesn't allow talking to other Docker instances that are in the
host network (like database-chassis).

On linecards, we configure midplane interfaces to let per-asic docker
containers talk to CHASSIS_DB on the supervisor through internal chassis
network.

On the supervisor we don't need to use chassis internal network, but we
still need a similar setup in order to allow fabric containers to talk
to database-chassis

#### How I did it
Reusing existing code used to setup the midplane network on a linecard,
except that we only need a pair of veth interfaces on the supervisor to allow
per-asic docker instances access to the midplane subnet.

### DIFF
--- a/files/build_templates/docker_image_ctl.j2
+++ b/files/build_templates/docker_image_ctl.j2
@@ -170,16 +170,26 @@ function postStartAction()
            ip netns exec "$NET_NS" ip addr add 127.0.0.1/16 dev lo
            ip netns exec "$NET_NS" ip addr del 127.0.0.1/8 dev lo
 
+           slot_id=$(python3 -c 'import sonic_platform.platform; platform_chassis = sonic_platform.platform.Platform().get_chassis(); print(platform_chassis.get_my_slot())' 2>/dev/null)
+           supervisor_slot_id=$(python3 -c 'import sonic_platform.platform; platform_chassis = sonic_platform.platform.Platform().get_chassis(); print(platform_chassis.get_supervisor_slot())' 2>/dev/null)
+
            # Create eth1 in database instance
-           ip link add name ns-eth1"$NET_NS" link eth1-midplane type macvlan mode bridge
+           if [[ "${slot_id}" == "${supervisor_slot_id}" ]]; then
+                   ip link add name ns-eth1"$NET_NS" type veth peer name eth1@"$NET_NS"
+                   ip link set dev eth1@"$NET_NS" master br1
+                   ip link set dev eth1@"$NET_NS" up
+           else
+                   ip link add name ns-eth1"$NET_NS" link eth1-midplane type macvlan mode bridge
+           fi
+
+           # Create eth1 in database instance
            ip link set dev ns-eth1"$NET_NS" netns "$NET_NS"
            ip netns exec "$NET_NS" ip link set ns-eth1"$NET_NS" name eth1
 
            # Configure IP address and enable eth1
-           lc_slot_id=$(python3 -c 'import sonic_platform.platform; platform_chassis = sonic_platform.platform.Platform().get_chassis(); print(platform_chassis.get_my_slot())' 2>/dev/null)
-           lc_ip_address=`echo $midplane_subnet | awk -F. '{print $1 "." $2}'`.$lc_slot_id.$(($DEV + 10))
-           lc_subnet_mask=${midplane_subnet#*/}
-           ip netns exec "$NET_NS" ip addr add $lc_ip_address/$lc_subnet_mask dev eth1
+           slot_ip_address=`echo $midplane_subnet | awk -F. '{print $1 "." $2}'`.$slot_id.$(($DEV + 10))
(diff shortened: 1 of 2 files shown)


mlorrillere:
CC @arlakshm @abdosi @judyjoseph @minionatwork

files/build_templates/docker_image_ctl.j2
```diff
@@ -170,16 +170,27 @@ function postStartAction()
...
+	   if [[ "${slot_id}" == "${supervisor_slot_id}" ]]; then
+		   ip link add name ns-eth1"$NET_NS" type veth peer name eth1@"$NET_NS"
+		   ip link set dev eth1@"$NET_NS" master br1
+		   ip link set dev eth1@"$NET_NS" up
+	   else
+		   ip link add name ns-eth1"$NET_NS" link eth1-midplane type macvlan mode bridge
+	   fi
+
            # Create eth1 in database instance
            ip link add name ns-eth1"$NET_NS" link eth1-midplane type macvlan mode bridge
```
arlakshm, line 186: Do we need this line, seems same as line 182 ?

abdosi:
wondering this change can be be handle in platform code ?  Not all vendors/platform have br1 define on supervisor. 

Similar to creation of eth1-midplane is platform/vendor specific.

files/build_templates/docker_image_ctl.j2
mlorrillere: Right, we don't need this line.

mlorrillere:
> wondering this change can be be handle in platform code ? Not all vendors/platform have br1 define on supervisor.
> 
> Similar to creation of eth1-midplane is platform/vendor specific.

I don't think we can easily move that change in platform code because all namespaces are created and destroyed here, that's the reason the entire logic was moved from interface-config to database.sh in #10500 

Here we assume `br1` exists (same as `eth1-midplane`). If there is no consensus on the name of the bridge we can add that in `chassisdb.conf` like we do for `midplane_subnet`.

files/build_templates/docker_image_ctl.j2
mlorrillere: Should be fixed now.

arlakshm:
> > wondering this change can be be handle in platform code ? Not all vendors/platform have br1 define on supervisor.
> > Similar to creation of eth1-midplane is platform/vendor specific.
> 
> I don't think we can easily move that change in platform code because all namespaces are created and destroyed here, that's the reason the entire logic was moved from interface-config to database.sh in #10500
> 
> Here we assume `br1` exists (same as `eth1-midplane`). If there is no consensus on the name of the bridge we can add that in `chassisdb.conf` like we do for `midplane_subnet`.

Hi @mlorrillere, not all platform need a software bridge  to communicate between LC and Sup. So they dont need to create `br1`.  On platforms which don't create the bridge we create an `eth1-midplane` interface. Wondering on the platfrom which need a software bridge we care create a bridge with name `eth1-midplane` ?

mlorrillere:
> > > wondering this change can be be handle in platform code ? Not all vendors/platform have br1 define on supervisor.
> > > Similar to creation of eth1-midplane is platform/vendor specific.
> > 
> > 
> > I don't think we can easily move that change in platform code because all namespaces are created and destroyed here, that's the reason the entire logic was moved from interface-config to database.sh in #10500
> > Here we assume `br1` exists (same as `eth1-midplane`). If there is no consensus on the name of the bridge we can add that in `chassisdb.conf` like we do for `midplane_subnet`.
> 
> Hi @mlorrillere, not all platform need a software bridge to communicate between LC and Sup. So they dont need to create `br1`. On platforms which don't create the bridge we create an `eth1-midplane` interface. Wondering on the platfrom which need a software bridge we care create a bridge with name `eth1-midplane` ?

`eth1-midplane` is the name of the interface on the linecard side. On the sup side the endpoint of the `eth1-midplane` of all linecards is added to `br1`.

In practice this PR is already a noop on platforms that don't have `br1` because this code is conditional on `midplane_subnet` being defined in `chassisdb.conf` of the sup (which is usually not defined, hence why it is added to `device/arista/x86_64-arista_7800_sup/chassisdb.conf`). Since this is not obvious, an alternative could be to add a `midplane_bridge=br1` to `chassisdb.conf`, that way we could clearly see which part of the code is run on the sup/linecard.

mlorrillere:
As discussed offline, we will keep `br1` in this PR as changing the name of the bridge requires updating Arista's platform modules. We will update Arista platform modules to create the bridge as `eth1-midplane` and update docker_image_ctl.j2 when we bump-up sonic-platform-modules-arista.

mlorrillere:
Hi @arlakshm ,
I believe we will need to port this PR to 202205 branch.

arlakshm (approved):

PR #11740, opened 2022-08-15
[device/celestica][e1031] Mitigation for command injection vulnerability

maipbui:

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205

### DIFF
--- a/device/celestica/x86_64-cel_e1031-r0/sonic_platform/common.py
+++ b/device/celestica/x86_64-cel_e1031-r0/sonic_platform/common.py
@@ -1,6 +1,7 @@
 import os
 import imp
 import yaml
+import shlex
 import subprocess
 
 from sonic_py_common import device_info
@@ -47,7 +48,7 @@ class Common:
         output = ""
         try:
             p = subprocess.Popen(
-                command, shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+                shlex.split(command), stdout=subprocess.PIPE, stderr=subprocess.PIPE)
             raw_data, err = p.communicate()
             if p.returncode == 0:
                 status, output = True, raw_data.strip()
@@ -67,7 +68,7 @@ class Common:
             cleaned_input = input_translator.get(input)
 
         elif type(input_translator) is str:
-            cleaned_input = eval(input_translator.format(input))
+            cleaned_input = eval(input_translator.format(input), {"__builtins__": None}, {})
 
         return cleaned_input
 
@@ -77,9 +78,9 @@ class Common:
         if type(output_translator) is dict:
             output = output_translator.get(output)
         elif type(output_translator) is str:
-            output = eval(output_translator.format(output))
+            output = eval(output_translator.format(output), {"__builtins__": None}, {})
         elif type(output_translator) is list:
-            output = eval(output_translator[index].format(output))
+            output = eval(output_translator[index].format(output), {"__builtins__": None}, {})
 
         return output
 
@@ -195,7 +196,8 @@ class Common:
         return True
 
     def is_host(self):
-        return os.system(self.HOST_CHK_CMD) == 0
+        command = shlex.split(self.HOST_CHK_CMD)
+        return subprocess.run(command).returncode == 0
 
     def load_json_file(self, path):
         """
--- a/device/celestica/x86_64-cel_e1031-r0/sonic_platform/component.py
+++ b/device/celestica/x86_64-cel_e1031-r0/sonic_platform/component.py
@@ -65,7 +65,7 @@ class Component(ComponentBase):
         # Retrieves the cpld register value
         cmd = "echo {1} > {0}; cat {0}".format(GETREG_PATH, register)
         p = subprocess.Popen(
-            cmd, shell=True, universal_newlines=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+            shlex.split(cmd), universal_newlines=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
         raw_data, err = p.communicate()
         if err is not '':
             return None
@@ -76,11 +76,11 @@ class Component(ComponentBase):
         cpld_version = dict()
         with open(SMC_CPLD_PATH, 'r') as fd:
             smc_cpld_version = fd.read()
-        smc_cpld_version = 'None' if smc_cpld_version is 'None' else "{}.{}".format(
+        smc_cpld_version = 'None' if smc_cpld_version == 'None' else "{}.{}".format(
             int(smc_cpld_version[2], 16), int(smc_cpld_version[3], 16))
 
         mmc_cpld_version = self.get_register_value(MMC_CPLD_ADDR)
-        mmc_cpld_version = 'None' if mmc_cpld_version is 'None' else "{}.{}".format(
+        mmc_cpld_version = 'None' if mmc_cpld_version == 'None' else "{}.{}".format(
             int(mmc_cpld_version[2], 16), int(mmc_cpld_version[3], 16))
 
         cpld_version["SMC_CPLD"] = smc_cpld_version


device/celestica/x86_64-cel_e1031-r0/sonic_platform/common.py
```diff
@@ -47,7 +48,7 @@ def run_command(self, command):
         output = ""
         try:
             p = subprocess.Popen(
-                command, shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+                split(quote(command)), stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
qiluo-msft, line 51: >split [](http://example.com/codeflow?start=16&length=5)

This will break many command, like "echo hello".

device/celestica/x86_64-cel_e1031-r0/sonic_platform/common.py
```diff
@@ -195,7 +196,8 @@ def write_txt_file(self, file_path, value):
         return True
 
     def is_host(self):
-        return os.system(self.HOST_CHK_CMD) == 0
+        command = split(quote(self.HOST_CHK_CMD))
```
qiluo-msft, line 199: >split [](http://example.com/codeflow?start=18&length=5)

This command could not run.

device/celestica/x86_64-cel_e1031-r0/sonic_platform/component.py
```diff
@@ -40,7 +40,7 @@ def __run_command(self, command):
         # Run bash command and print output to stdout
         try:
             process = subprocess.Popen(
-                shlex.split(command), universal_newlines=True, stdout=subprocess.PIPE)
+                split(quote(command)), universal_newlines=True, stdout=subprocess.PIPE)
```
qiluo-msft, line 43: >quote [](http://example.com/codeflow?start=22&length=5)

It will break commands, like `echo Hello`.

device/celestica/x86_64-cel_e1031-r0/sonic_platform/component.py
```diff
@@ -65,7 +65,7 @@ def get_register_value(self, register):
         # Retrieves the cpld register value
         cmd = "echo {1} > {0}; cat {0}".format(GETREG_PATH, register)
         p = subprocess.Popen(
-            cmd, shell=True, universal_newlines=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+            split(quote(cmd)), universal_newlines=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
qiluo-msft, line 68: >quote [](http://example.com/codeflow?start=18&length=5)

broken

device/celestica/x86_64-cel_silverstone-r0/sonic_platform/psu.py
```diff
@@ -33,20 +33,15 @@
...
-PSU2_VOUT_SS_ID = "0x40"
-PSU2_COUT_SS_ID = "0x41"
-PSU2_POUT_SS_ID = "0x42"
-PSU2_STATUS_REG = "0x2f"
-
 PSU1_FRU_ID = 3
 
 SS_READ_OFFSET = 0
 
+PSU_VOUT_SS_ID = {"PSU1_VOUT_SS_ID": "0x36", "PSU2_VOUT_SS_ID": "0x40"}
```
qiluo-msft, line 40: >PSU_VOUT_SS_ID [](http://example.com/codeflow?start=0&length=14)

Is an array better than a dict?

maipbui: I believe a dict is better than array since it is easier to get value of the key and faster to lookup.

device/celestica/x86_64-cel_e1031-r0/sonic_platform/common.py
maipbui: Fixed to use `split` only.

device/celestica/x86_64-cel_e1031-r0/sonic_platform/common.py
```diff
@@ -195,7 +196,8 @@ def write_txt_file(self, file_path, value):
         return True
 
     def is_host(self):
-        return os.system(self.HOST_CHK_CMD) == 0
+        command = split(self.HOST_CHK_CMD)
+        return subprocess.run(command).returncode == 0
```
qiluo-msft, line 200: >subprocess.run [](http://example.com/codeflow?start=15&length=14)

For this special case, the safe way is
`subprocess.run(['docker'], stdout=None, stderr=None).returncode`

qiluo-msft:
>     def get_output(self, index, config, default):

Seems this function is not used. Please remove.
@mudsut4ke Could you help review?

---
Refers to: device/celestica/x86_64-cel_e1031-r0/sonic_platform/common.py:226 in ca56944. [](commit_id = ca56944b57aec4178b91ddc2084fd779b9f7fc42, deletion_comment = False)

qiluo-msft:
>     def set_output(self, index, input, config):

Seems this function is not used. Please remove.
@mudsut4ke Could you help review?

---
Refers to: device/celestica/x86_64-cel_e1031-r0/sonic_platform/common.py:281 in ca56944. [](commit_id = ca56944b57aec4178b91ddc2084fd779b9f7fc42, deletion_comment = False)

qiluo-msft:
>         #     install_command = "afulnx_64 %s /p /b /n /x /r" % image_path

`image_path` is an input parameter. A malicious user could inject some string not as simple as one token, and split into multiple parts.

---
Refers to: device/celestica/x86_64-cel_e1031-r0/sonic_platform/component.py:164 in ca56944. [](commit_id = ca56944b57aec4178b91ddc2084fd779b9f7fc42, deletion_comment = False)

device/celestica/x86_64-cel_e1031-r0/sonic_platform/common.py
qiluo-msft: The stdout/stderr redirect is not needed inside `HOST_CHK_CMD` string.

device/celestica/x86_64-cel_silverstone-r0/sonic_platform/psu.py
maipbui: Fixed.

device/celestica/x86_64-cel_e1031-r0/sonic_platform/common.py
maipbui: Fixed.

qiluo-msft: Suggest this way.
```
HOST_CHK_CMD = ['docker']
```
Benefit:
1. minimal code change
2. string array will be super safe and one less command line parsing.

device/celestica/x86_64-cel_silverstone-r0/sonic_platform/helper.py
```diff
@@ -28,23 +28,25 @@ def pci_get_value(self, resource, offset):
...
-                cmd, shell=True, universal_newlines=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
-            raw_data, err = p.communicate()
-            if err == '':
-                result = raw_data.strip()
-        except:
+            cmd = ' '.join(cmd1_args) + ' | ' + ' '.join(cmd2_args)
+            with subprocess.Popen(cmd1_args, universal_newlines=True, stdout=subprocess.PIPE) as p1:
+                with subprocess.Popen(cmd2_args, universal_newlines=True, stdin=p1.stdout, stdout=subprocess.PIPE) as p2:
+                    raw_data = p2.communicate()[0]
+                    if p2.returncode != 0:
```
qiluo-msft, line 39: >p2.returncode [](http://example.com/codeflow?start=23&length=13)

Also check `p1.returncode != 0` ?

qiluo-msft (approved):

device/celestica/x86_64-cel_seastone_2-r0/plugins/psuutil.py
```diff
@@ -13,16 +12,19 @@ class PsuUtil(PsuBase):
...
-
-        if proc.returncode != 0:
-            sys.exit(proc.returncode)
-
+    def run_command(self, cmd1_args, cmd2_args):
+        with subprocess.Popen(cmd1_args, universal_newlines=True, stdout=subprocess.PIPE) as p1:
+            with subprocess.Popen(cmd2_args, universal_newlines=True, stdin=p1.stdout, stdout=subprocess.PIPE) as p2:
+                out = p2.communicate()[0]
+                p1.wait()
+                if p1.returncode != 0 or p2.returncode != 0:
```
qiluo-msft, line 23: >if p1.returncode != 0 or p2.returncode != 0: [](http://example.com/codeflow?start=16&length=44)

you can simplify even further:
remove the `or` check
```
                    if p1.returncode != 0:
                        sys.exit(p1.returncode)
                    else if p2.returncode != 0:
                        sys.exit(p2.returncode)
```

device/celestica/x86_64-cel_seastone_2-r0/plugins/psuutil.py
```diff
@@ -13,16 +12,19 @@ class PsuUtil(PsuBase):
...
 
-    def run_command(self, command):
-        proc = subprocess.Popen(command, shell=True, universal_newlines=True, stdout=subprocess.PIPE)
-        (out, err) = proc.communicate()
-
-        if proc.returncode != 0:
-            sys.exit(proc.returncode)
-
+    def run_command(self, cmd1_args, cmd2_args):
+        with subprocess.Popen(cmd1_args, universal_newlines=True, stdout=subprocess.PIPE) as p1:
```
qiluo-msft, line 19: >with subprocess.Popen(cmd1_args, universal_newlines=True, stdout=subprocess.PIPE) as p1: [](http://example.com/codeflow?start=8&length=88)

Possible to use the function in common library?

qiluo-msft: Use a loop to check exitcode?

qiluo-msft (approved):

qiluo-msft:
@Blueve is connecting with Celestica reviewers.

Blueve:
Celestica team acked to review

device/celestica/x86_64-cel_e1031-r0/sonic_platform/common.py
```diff
@@ -195,7 +185,7 @@ def write_txt_file(self, file_path, value):
         return True
 
     def is_host(self):
-        return os.system(self.HOST_CHK_CMD) == 0
+        return subprocess.call(self.HOST_CHK_CMD) == 0
```
qnos, line 188: Should pass DEVNULL to stdout and stderr parameters to suppress the output message.
Replace with return subprocess.call(HOST_CHK_CMD, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL) == 0

qnos, line 188: If use subprocess.call, need to suppress stdout and stderr and handle docker command not found exception in docker space.
The following example codes work in both host and docker mode.
``` python3
def is_host():
    try:
        subprocess.call(HOST_CHK_CMD, stdout=subprocess.DEVNULL,
            stderr=subprocess.DEVNULL)
    except FileNotFoundError:
        return False
    return True
```

device/celestica/x86_64-cel_seastone-r0/sonic_platform/helper.py
```diff
@@ -15,7 +15,7 @@ def __init__(self):
         (self.platform, self.hwsku) = device_info.get_platform_and_hwsku()
 
     def is_host(self):
-        return os.system(HOST_CHK_CMD) == 0
+        return subprocess.call(HOST_CHK_CMD) == 0
```
qnos, line 18: Please consider changing the is_host function implementation as suggested above.

device/celestica/x86_64-cel_silverstone-r0/sonic_platform/helper.py
```diff
@@ -13,7 +14,7 @@ def __init__(self):
         pass
 
     def is_host(self):
-        return os.system(HOST_CHK_CMD) == 0
+        return subprocess.call(HOST_CHK_CMD) == 0
```
qnos, line 17: Please consider changing the is_host function implementation as suggested above.

device/celestica/x86_64-cel_silverstone-r0/sonic_platform/sfp.py
```diff
@@ -270,7 +269,7 @@ def __convert_string_to_num(self, value_str):
             return 'N/A'
 
     def __is_host(self):
-        return os.system(self.HOST_CHK_CMD) == 0
+        return subprocess.call(self.HOST_CHK_CMD) == 0
```
qnos, line 272: Please consider changing the is_host function implementation as suggested above.

platform/broadcom/sonic-platform-modules-cel/services/platform_api/sonic_platform/common.py
```diff
@@ -166,7 +167,7 @@ def write_txt_file(self, file_path, value):
         return True
 
     def is_host(self):
-        return os.system(self.HOST_CHK_CMD) == 0
+        return subprocess.call(self.HOST_CHK_CMD) == 0
```
qnos, line 170: Please consider changing the is_host function implementation as suggested above.

device/celestica/x86_64-cel_e1031-r0/sonic_platform/common.py
maipbui: I think os.system is equivalent with subprocess.call without specifying stdour and stderr to DEVNULL. Ref: https://docs.python.org/3/library/subprocess.html#replacing-os-system

qnos (changes requested):
Except the is_host() function implementation, agree with all other changes.

device/celestica/x86_64-cel_e1031-r0/sonic_platform/common.py
maipbui: I got what you meant, I will change per your suggestion

qnos: > I think os.system is equivalent with subprocess.call without specifying stdout and stderr to DEVNULL. Ref: https://docs.python.org/3/library/subprocess.html#replacing-os-system

Yes, no essential differences between the two functions, but HOST_CHK_CMD had been changed from "docker > /dev/null 2>&1" to ["docker"], so there is no way to suppress the 'docker' command output any more in host mode. And handling the FileNotFoundError exception is also necessary for docker mode.

maipbui: Fixed. Could you review and approve? Thank you!

PR #11750, opened 2022-08-16
202205: Update bcm config file system_ref_core_clock_khz param for j2cplus linecards

skbarista:
Update the bcm config file system_ref_core_clock_khz param to handlesystems with J2cplus linecards.

We need system_ref_core_clock_khz to be set to 1600000 for supporting j2 and j2cplus linecards on the same chassis.


skbarista:
@lguohan please discard this pr. Created it by mistake

PR #11754, opened 2022-08-17
Added support to add gbsyncd in Feature Table of Host Config DB

abdosi:
Why I did:

In case of multi-asic platforms `gbsyncd` is not getting added to Feature Table of Host Config DB. Without this container_checker complains of not needed gbsyncd container's are running.

How I did:
Update Both Host and Namespace config db when `gbsyncd` docker is starting.

How I verify:
Verified on Multi-asic platfroms.


abdosi:
cc anamehra

anamehra:
LGTM

arlakshm (approved):

anamehra (approved):

PR #11763, opened 2022-08-17
[snmpd]: Update to 5.9+dfsg-4+deb11u1 to match Debian version

saiarcot895:
This brings in some security fixes.

#### Why I did it

Debian has updated snmpd to 5.9+dfsg-4+deb11u1. This is necessary so that a SONiC image can be built.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [x] 202205


src/snmpd/Makefile
```diff
@@ -31,27 +31,13 @@ $(addprefix $(DEST)/, $(MAIN_TARGET)): $(DEST)/% :
 	rm -rf net-snmp-$(SNMPD_VERSION)
 
 	# download debian net-snmp
-ifneq (,$(findstring 5.9,$(SNMPD_VERSION)))
```
qiluo-msft, line 34: If we upgrade again, do we need to check another version? Or we can just check the old versions?

saiarcot895: This if-check was initially added because the version in Bullseye was `5.9+dfsg-3+b1`, but the source files were the same as `5.9+dfsg-3`. The `+b1` indicates a binary-only upload, for rebuilding against a newer version of perl. Therefore, to get around that, the sources for `5.9+dfsg-3` were downloaded, and the version was manually corrected.

Since this isn't a binary-only upload, this check isn't needed anymore. If there is a binary-only upload in the future, though, (for any package) then some custom handling might be needed.

qiluo-msft (approved):

PR #11770, opened 2022-08-18
[docker-macsec]: Add dependencies of MACsec

Pterosaur:
Signed-off-by: Ze Gan <ganze718@gmail.com>

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205

### DIFF
--- a/rules/docker-macsec.mk
+++ b/rules/docker-macsec.mk
@@ -42,6 +42,10 @@ $(DOCKER_MACSEC)_RUN_OPT += --privileged -t
 $(DOCKER_MACSEC)_RUN_OPT += -v /etc/sonic:/etc/sonic:ro
 $(DOCKER_MACSEC)_RUN_OPT += -v /host/warmboot:/var/warmboot
 
+$(DOCKER_MACSEC)_SERVICE_REQUIRES = updategraph
+$(DOCKER_MACSEC)_SERVICE_AFTER = swss syncd
+$(DOCKER_MACSEC)_SERVICE_DEPENDENT_OF = swss
+
 $(DOCKER_MACSEC)_CLI_CONFIG_PLUGIN = /cli/config/plugins/macsec.py
 $(DOCKER_MACSEC)_CLI_SHOW_PLUGIN = /cli/show/plugins/show_macsec.py
 


files/scripts/swss.sh
```diff
@@ -28,7 +28,7 @@ function read_dependent_services()
     fi
 
     if [[ -f ${ETC_SONIC_PATH}/${SERVICE}_multi_inst_dependent ]]; then
-        MULTI_INST_DEPENDENT="${MULTI_INST_DEPENDENT} cat ${ETC_SONIC_PATH}/${SERVICE}_multi_inst_dependent"
+        MULTI_INST_DEPENDENT="${MULTI_INST_DEPENDENT} $(cat ${ETC_SONIC_PATH}/${SERVICE}_multi_inst_dependent)"
```
judyjoseph, line 31: I am thinking we should use DEPENDENT, instead of MULTI_INST_DEPENDENT for macsec.

with DEPENDENT - if swss goes down dependent services go down. eg: macsec service here will go down

with MULTI_INST_DEPENDENT - if swss goes down dependent services go down. In addition if dependent service like macsec go down -- we bring down evering like swss/syncd etc. We would not need that for macsec.

Pterosaur: https://github.com/sonic-net/sonic-utilities/blob/7704f6378724468417212f9f05c7a371311670cf/sonic_package_manager/service_creator/creator.py#L358-L362

https://github.com/sonic-net/sonic-buildimage/blob/56a679dc09081b41f55e3671a914eb9c7e4b0d8f/rules/functions#L184-L185

I think, A service as a MULTI_INST_DEPENDENT or DEPENDENT depends on the location of xxx_.service.j2.
But right now, the macsec.service.j2 is in the folder of `per_namespace`. I think it's make sense. So, I think macsec as MULTI_INST_DEPENDENT is following the current convention.

rules/docker-macsec.mk
```diff
@@ -42,6 +42,10 @@ $(DOCKER_MACSEC)_RUN_OPT += --privileged -t
 $(DOCKER_MACSEC)_RUN_OPT += -v /etc/sonic:/etc/sonic:ro
 $(DOCKER_MACSEC)_RUN_OPT += -v /host/warmboot:/var/warmboot
 
+$(DOCKER_MACSEC)_SERVICE_REQUIRES = updategraph
+$(DOCKER_MACSEC)_SERVICE_AFTER = swss syncd
```
shyam77git, line 46: shouldn't gbsyncd too part of this?
gbsyncd is a vertical parallel to swss and syncd docker containers. Together the all three (swss, gbsyncd, syncd) complete interfaces related work-flow for SKU's having PHY (Gearbox)

files/scripts/swss.sh
judyjoseph: The variable DEPENDENT and MULTI_ASIC_DEPENDENT -- the functionality wise it is not the same. I think we should change that name to may be MULTI_ASIC_PEER or so.

https://github.com/sonic-net/sonic-buildimage/blob/master/files/scripts/swss.sh#L251, So we it is in MULTI_ASIC_DEPENDENT we add that to the list of containers /usr/bin/docker-wait-any check for exit, and if any of them exit will restart swss service.

judyjoseph (approved):

rules/docker-macsec.mk
Pterosaur: @jimmyzhai said gbsyncd is as a runtime service which cannot be bounded at make time. And gbsyncd has been bound to syncd, so we don't need to explicitly bind gbsyncd at here.

files/scripts/swss.sh
```diff
@@ -306,6 +306,15 @@ function check_peer_gbsyncd()
     fi
 }
 
+function check_macsec()
+{
+    MACSEC_STATE=`show feature status | grep macsec | awk '{print $2}'`
+
+    if [[ ${MACSEC_STATE} == 'enabled' ]]; then
+        DEPENDENT="${DEPENDENT} macsec"
```
abdosi, line 314: does not correct it should be like this https://github.com/sonic-net/sonic-buildimage/commit/23dcfdf9b672a1f637f618741ca26c31490fe623

Otherwise i think multi-asic might now work

abdosi (changes requested):

files/scripts/swss.sh
Pterosaur: Make sense, Thanks for your suggestion.
@judyjoseph Could you help to check the latest commit agian.

judyjoseph: Yes it is fine now, thanks to @abdosi for pointing it out.
Btw @Pterosaur may be this branch needs a rebase the script don't have the bgp changes.

Pterosaur: Yes, Done.

Pterosaur:
@abdosi Could you please help to check this PR again? Thanks~

shyam77git:
@Pterosaur 
Don't see any UT steps, logs to this PR.
could you please add them? also, please add, update which UT done on simulator or physical (HW)

abdosi (approved):

PR #11771, opened 2022-08-18
Add gRPC default port 50051 to reserved ports

gord1306:
#### Why I did it
The port 50051 may be occupied by the other network connections than gRPC

#### How I did it
Add port 50051 into the reserved ports to prevent from being chosen as the
ephemeral port when establishing a network connection.

#### How to verify it
Generate connection make sure there no 50051 port be used

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205

Signed-off-by: gord_chen <gord_chen@edge-core.com>


gord1306:
Fixed by https://github.com/sonic-net/sonic-buildimage/pull/17907

PR #11781, opened 2022-08-19
Add new Redis database PROFILE_DB

liuh-80:
Add new Redis database PROFILE_DB

#### Why I did it
Add new PROFILE_DB so profile in J2 template will render to this DB.
Also, swss-common will change to read profile from this new DB.
This will simplify reboot handling logic.

#### How I did it
Add PROFILE_DB config in database_config.j2 and database_config.json.

#### How to verify it
Pass all E2E test case.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205

Add new Redis database PROFILE_DB

### DIFF
--- a/dockers/docker-database/database_config.json.j2
+++ b/dockers/docker-database/database_config.json.j2
@@ -93,6 +93,11 @@
             "id" : 14,
             "separator": ":",
             "instance" : "redis"
+        },
+        "STATIC_CONFIG_DB" : {
+            "id" : 15,
+            "separator": ":",
+            "instance" : "redis"
         }
     },
     "VERSION" : "1.0"
--- a/dockers/docker-database/database_config.json.j2
+++ b/dockers/docker-database/database_config.json.j2
@@ -94,7 +94,7 @@
             "separator": ":",
             "instance" : "redis"
         },
-        "STATIC_CONFIG_DB" : {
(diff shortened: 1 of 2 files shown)


dockers/docker-database/database_config.json.j2
```diff
@@ -93,6 +93,11 @@
             "id" : 14,
             "separator": ":",
             "instance" : "redis"
+        },
+        "PROFILE_DB" : {
```
qiluo-msft, line 97: PROFILE_DB is similar to CONFIG_DB. As a general practice, we prefer `|` as separator for newly added DB index.

liuh-80: Fixed, also will fix config files in swss-common PR later.

liuh-80:
Close because this issue: 
Commenter does not have sufficient privileges for PR 11781 in repo sonic-net/sonic-buildimage

dockers/docker-database/database_config.json.j2
```diff
@@ -93,6 +93,11 @@
...
+            "id" : 15,
```
qiluo-msft, line 98: This id will conflict with TEST_DB.

PR #11786, opened 2022-08-19
Adding support for get/set low power mode for QSFPs in PDDF common APIs

FuzailBrcm:
#### Why I did it

Issue#11785

Unable to get/set low power mode for QSFP28 and QSFP+

Platforms which use PDDF have common PDDF platform APIs. Since sfp_refactor framework takes care of only CMIS transceivers and platform APIs are supposed to take care of the low power mode for QSFPs.

#### How I did it
Modified pddf_sfp.py to include get/set APIs for lpmode in case of QSFP

#### How to verify it

> sfputil show lpmode
> sfputil lpmode on/off <>

LOGS:

```
>>> platform_chassis.get_sfp(55).set_lpmode(True)

True

>>> 

>>> platform_chassis.get_sfp(55).get_lpmode()

True

>>> platform_chassis.get_sfp(55).set_lpmode(False)

True

>>> 

>>> platform_chassis.get_sfp(55).get_lpmode()

False
```

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205

### DIFF
--- a/platform/pddf/platform-api-pddf-base/sonic_platform_pddf_base/pddf_sfp.py
+++ b/platform/pddf/platform-api-pddf-base/sonic_platform_pddf_base/pddf_sfp.py
@@ -204,10 +204,10 @@ class PddfSfp(SfpOptoeBase):
                     lpmode = super().get_lpmode()
                 elif xcvr_id == 0x11 or xcvr_id == 0x0d or xcvr_id == 0x0c:
                     # QSFP28, QSFP+, QSFP
-                    val = self.read_eeprom(QSFP_PWR_CTRL_ADDR, 1)
-                    if val is not None:
-                        if (ord(val) & 0x3) == 0x3:
-                            lpmode = True
+                    power_set = self.get_power_set()
+                    power_override = self.get_power_override()
+                    # By default the lpmode pin is pulled high as mentioned in the sff community
+                    return power_set if power_override else True
 
         return lpmode
 
@@ -341,12 +341,10 @@ class PddfSfp(SfpOptoeBase):
                     status = super().set_lpmode(lpmode)
                 elif xcvr_id == 0x11 or xcvr_id == 0x0d or xcvr_id == 0x0c:
                     # QSFP28, QSFP+, QSFP
-                    val = self.read_eeprom(QSFP_PWR_CTRL_ADDR, 1)
-                    if val is not None:
-                        val = ord(val) & 0xf0
-                        val |= 0x3 if lpmode else 0xd
-                        buf = bytearray([val])
-                        status = self.write_eeprom(QSFP_PWR_CTRL_ADDR, 1, buf)
+                    if lpmode is True:
+                        self.set_power_override(True, True)
+                    else:
+                        self.set_power_override(True, False)
 
         return status
 
--- a/platform/pddf/platform-api-pddf-base/sonic_platform_pddf_base/pddf_sfp.py
+++ b/platform/pddf/platform-api-pddf-base/sonic_platform_pddf_base/pddf_sfp.py
@@ -10,6 +10,8 @@ try:
 except ImportError as e:
     raise ImportError(str(e) + "- required module not found")
 
+QSFP_PWR_CTRL_ADDR = 93
+
 
 class PddfSfp(SfpOptoeBase):
     """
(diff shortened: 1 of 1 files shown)


FuzailBrcm:
@prgeor 
Can you review this?

jostar-yang (approved):
It is fine to me

zhangyanzhao:
@prgeor  can you please help to take a look and merge? Thanks.

platform/pddf/platform-api-pddf-base/sonic_platform_pddf_base/pddf_sfp.py
```diff
@@ -321,8 +333,20 @@ def set_lpmode(self, lpmode):
...
+                    # Use common SfpOptoeBase implementation for set_lpmode
+                    status = super().set_lpmode(lpmode)
+                elif xcvr_id == 0x11 or xcvr_id == 0x0d or xcvr_id == 0x0c:
+                    # QSFP28, QSFP+, QSFP
+                    val = self.read_eeprom(QSFP_PWR_CTRL_ADDR, 1)
+                    if val is not None:
+                        val = ord(val) & 0xf0
+                        val |= 0x3 if lpmode else 0xd
+                        buf = bytearray([val])
+                        status = self.write_eeprom(QSFP_PWR_CTRL_ADDR, 1, buf)
```
prgeor, line 349: use [set_power_override](https://github.com/sonic-net/sonic-platform-common/blob/master/sonic_platform_base/sonic_xcvr/api/public/sff8436.py#L253) for QSFP28/QSFP+. For eg see [here](https://github.com/sonic-net/sonic-buildimage/blob/master/device/accton/x86_64-accton_as4630_54pe-r0/sonic_platform/sfp.py#L372)

platform/pddf/platform-api-pddf-base/sonic_platform_pddf_base/pddf_sfp.py
```diff
@@ -194,8 +196,18 @@ def get_lpmode(self):
...
+                if xcvr_id == 0x18 or xcvr_id == 0x19 or xcvr_id == 0x1e:
+                    # QSFP-DD or OSFP
+                    # Use common SfpOptoeBase implementation for get_lpmode
+                    lpmode = super().get_lpmode()
+                elif xcvr_id == 0x11 or xcvr_id == 0x0d or xcvr_id == 0x0c:
+                    # QSFP28, QSFP+, QSFP
+                    power_set = self.get_power_set()
+                    power_override = self.get_power_override()
+                    # By default the lpmode pin is pulled high as mentioned in the sff community
+                    return power_set if power_override else True
```
FuzailBrcm, line 210: Please check this line. We are returning True in case the power_override is not set. 

**SFF_8436**

4.1.1.3 LPMode
The LPMode pin shall be pulled up to Vcc in the QSFP+ module. This function is affected
by the LPMode pin and the combination of the Power_over-ride and Power_set software
control bits (Address A0h, byte 93 bits 0,1).

**SFF_8636**

QSFP+ and QSFP28 modules have the LPMode input signal (see SFF-8679) that can be used by the host system to
force the module into Low Power Mode. Low Power Mode for those modules is defined as maximum power
consumption of 1.5W. If the LPMode input signal is pulled low by the host system, the module is then capable of
entering High Power Mode. The maximum power consumption in High Power Mode depends on the module Power
Class as advertised in the Extended Identifier (see 6.3.2), Page 00h, byte 129, bits 1-0 and 7-6.
The operation of the LPMode input signal can be overridden by the host writing a ‘1’ to byte 93, bit 0. In that case,
the function of the LPMode input signal is replaced by byte 93, bit 1.
SFF-8436 has 4 power classes from 1.5 to 3.5 W. Only bits 7-6 are used to define those power classes. At revision
1.9 of this specification, 3 new higher power classes, 4.0 W, 4.5 W, and 5.0 W were added. These power classes,
designated power classes 5, 6 and 7, are designated using bits 1-0 of the Extended Identifier byte, page 00h byte
129. In order to protect legacy host systems designed to support only the original 4 power classes, the High Power
Class Enable control was defined at byte 93, bit 2. Modules in power classes 5, 6, 7 or 8 are required to limit power
consumption to no more than a power class 4 module if the High Power Class Enable, byte 93 bit 2 control is not
set. They are not required to be functional in this situation.

FuzailBrcm:
@jostar-yang , @prgeor , @ds952811 

Can you review it again?

prgeor (approved):

PR #11795, opened 2022-08-19
New API to support runtime metadata needed for Feature Table field jinja rendering.

abdosi:
What I did:
Added new API to return runtime metdata dictionary as needed during Feature Table field rendering by hostcfgd.

Why I did:
Details: https://github.com/sonic-net/sonic-buildimage/pull/11796 and https://github.com/sonic-net/sonic-host-services/pull/14

How I Verify:
Manual Verification


abdosi:
cc @anamehra @SuvarnaMeenakshi

src/sonic-py-common/sonic_py_common/device_info.py
```diff
@@ -441,7 +441,7 @@ def is_packet_chassis():
 
 
 def is_chassis():
-    return is_voq_chassis() or is_packet_chassis()
+    return is_voq_chassis() or is_packet_chassis() or is_supervisor()
```
arlakshm, line 444: remove is_supervisor() check here ?

abdosi: @arlakshm Done

arlakshm (approved):

PR #11796, opened 2022-08-19
Added Support to runtime render bgp and teamd feature state and lldp has_asic_scope flag

abdosi:
What I did:
Added Support to runtime render bgp and teamd feature `state` and lldp `has_asic_scope` flag

How I did:

Define new runtime metadata `dict` as define in https://github.com/sonic-net/sonic-buildimage/pull/11795 which is used for rendering by hostcfgd.

- bgp: Enable Feature if Ethernet ports are present on device and Device is not Chassis supervisor
- teamd: Enable Feature if Ethernet ports are present on device
- lldp: Enable Multi-asic scope if Ethernet ports are present on device and Device is not Chassis supervisor

Also in `swss.sh` only add teamd as Dependent service if Ethernet ports are present on the device.
Same approach can be used for other docker/services as needed.

Hostcfgd modication: https://github.com/sonic-net/sonic-host-services/pull/14

Order of PR merge will be:
https://github.com/sonic-net/sonic-buildimage/pull/11795
https://github.com/sonic-net/sonic-host-services/pull/14 and Submodule Update of same.
Finally this PR will be merged.

Why I did:
Not to create dockers which are not needed.

How I Verify:
Manual Verification.


abdosi:
cc @anamehra @SuvarnaMeenakshi

files/build_templates/init_cfg.json.j2
```diff
@@ -32,7 +32,7 @@
             "tsa_enabled": "false"
         }
     },
-{%- set features = [("bgp", "enabled", false, "enabled"),
+{%- set features = [("bgp", "{% if not DEVICE_RUNTIME_METADATA['ETHERNET_PORTS_PRESENT'] or ('CHASSIS_METADATA' in DEVICE_RUNTIME_METADATA and DEVICE_RUNTIME_METADATA['CHASSIS_METADATA']['module_type'] in ['supervisor']) %}disabled{% else %}enabled{% endif %}", false, "enabled"),
```
mlok-nokia, line 35: In order to allow the data value to be generated with variable in the init_cfg.json after build,  need to have {% raw  %}  .... {% endraw %}.  Such as below
{% raw %} "{% if not DEVICE_RUNTIME_METADATA['ETHERNET_PORTS_PRESENT'] or ('CHASSIS_METADATA' in DEVICE_RUNTIME_METADATA and DEVICE_RUNTIME_METADATA['CHASSIS_METADATA']['module_type'] in ['supervisor']) %}disabled{% else %}enabled{% endif %}", false, "enabled")",{% endraw %}

files/build_templates/init_cfg.json.j2
```diff
@@ -41,7 +41,7 @@
                    ("snmp", "enabled", true, "enabled"),
                    ("swss", "enabled", false, "enabled"),
                    ("syncd", "enabled", false, "enabled"),
-                   ("teamd", "enabled", false, "enabled")] %}
+                   ("teamd", "{% if not DEVICE_RUNTIME_METADATA['ETHERNET_PORTS_PRESENT'] %}disabled{% else %}enabled{% endif %}", false, "enabled")] %}
```
mlok-nokia, line 44: it looks like this line also require  {%raw %} ..... {% endraw %}

files/build_templates/init_cfg.json.j2
```diff
@@ -60,7 +60,11 @@
             "state": "{{state}}",
             "has_timer" : {{has_timer | lower()}},
             "has_global_scope": {% if feature + '.service' in installer_services.split(' ') %}true{% else %}false{% endif %},
+{%- if feature in ["lldp"] %}
+            "has_per_asic_scope": "{% if not DEVICE_RUNTIME_METADATA['ETHERNET_PORTS_PRESENT'] or ('CHASSIS_METADATA' in DEVICE_RUNTIME_METADATA and DEVICE_RUNTIME_METADATA['CHASSIS_METADATA']['module_type'] in ['supervisor']) %}False{% else %}True{% endif %}",
```
mlok-nokia, line 64: this also requires  {% raw %} .... {% endraw %}

abdosi:
@mlok-nokia : Can you please approve this.

abdosi:
@arlakshm can you please review this.

files/build_templates/init_cfg.json.j2
abdosi: Updated by @mlok-nokia

files/scripts/swss.sh
```diff
@@ -321,6 +315,22 @@ function check_macsec()
     fi
 }
 
+function check_ports_present()
+{
+    PORT_CONFIG_INI=/usr/share/sonic/device/$PLATFORM/$HWSKU/$DEV/port_config.ini
+    HWSKU_JSON=/usr/share/sonic/device/$PLATFORM/$HWSKU/$DEV/hwsku.json
+
+    if [[ -f $PORT_CONFIG_INI ]] || [[ -f $HWSKU_JSON ]]; then
```
judyjoseph, line 323: Should we check for presence of "interfaces" in platform.json file? Does all platforms need to define hwsku.json file

files/build_templates/init_cfg.json.j2
```diff
@@ -61,7 +61,11 @@
             "state": "{{state}}",
             "has_timer" : {{has_timer | lower()}},
             "has_global_scope": {% if feature + '.service' in installer_services.split(' ') %}true{% else %}false{% endif %},
+{%- if feature in ["lldp"] %}
+            "has_per_asic_scope": {% raw %}"{% if not DEVICE_RUNTIME_METADATA['ETHERNET_PORTS_PRESENT'] or ('CHASSIS_METADATA' in DEVICE_RUNTIME_METADATA and DEVICE_RUNTIME_METADATA['CHASSIS_METADATA']['module_type'] in ['supervisor']) %}False{% else %}True{% endif %}"{% endraw %},
+{%- else %}
             "has_per_asic_scope": {% if feature + '@.service' in installer_services.split(' ') %}true{% else %}false{% endif %},
```
judyjoseph, line 67: has_per_asic_scope is changed from boolean_type to string, should we change here too ?

abdosi: @judyjoseph : thanks for the comment and catching this. Updated.

files/scripts/swss.sh
abdosi: @judyjoseph no all platforms need not define hwsku.json file. Check is if either one of the file (port_config.ini or platform.json ) present then we have have front panel ports on the system.  Current behaviour assume this files are define only when there are Ethernet ports on the system

judyjoseph (approved):

yxieca:
@abdosi this PR cannot be cherry-picked cleanly. Can you raise separate PR for 202205?

PR #11821, opened 2022-08-23
[mux] Exit to write standby state to `active-active` ports

lolyu:
Signed-off-by: Longxiang Lyu <lolv@microsoft.com>

#### Why I did it
After stopping the `mux` container, `show mux status` returns `active` for `active-active` ports, which should be in `standby` state.

#### How I did it
Let `mux` exit by calling `write_standby.py` with args to write `standby` for `active-active` ports.

#### How to verify it
Stop `mux` container and verify `active-active` ports are in `standby`.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [x] 202205

### DIFF
--- a/files/build_templates/mux.service.j2
+++ b/files/build_templates/mux.service.j2
@@ -14,7 +14,7 @@ ExecStartPre=/usr/local/bin/mark_dhcp_packet.py
 ExecStartPre=/usr/bin/{{docker_container_name}}.sh start
 ExecStart=/usr/bin/{{docker_container_name}}.sh wait
 ExecStop=/usr/bin/{{docker_container_name}}.sh stop
-ExecStopPost=/usr/local/bin/write_standby.py
+ExecStopPost=/usr/local/bin/write_standby.py -a standby
 Restart=always
 RestartSec=30
 


yxieca (approved):

zjswhhh (approved):

zjswhhh:
Maybe need to do the same thing for bgp service: 
https://github.com/sonic-net/sonic-buildimage/blob/234b4973cddb0d3db550341b42c20785e9a0108c/files/build_templates/per_namespace/bgp.service.j2#L18

yxieca:
We should also pass in the module name, in this case 'mux', to the write class, so that it could check if we are warm restarting mux docker.

lolyu:
> We should also pass in the module name, in this case 'mux', to the write class, so that it could check if we are warm restarting mux docker.
> 
> You might want to change the parameter to --shutdown <mux|bgp> so it ready for BGP for the next PR>

`write_standby.py` checks if it is warm restart before writing the configs:
https://github.com/sonic-net/sonic-buildimage/blob/214e394ac01c411b5aac228c14a30fdbe3f73ad5/files/scripts/write_standby.py#L156

lolyu:
Hi, @yxieca, updated, please take a review, thanks!

yxieca (approved):

PR #11827, opened 2022-08-23
[bgp][service] Start bgp service after interfaces-config service

vboykox:

#### Why I did it
interfaces-config service restarts networking service, during the restart loopback interface address is being removed and reassigned back, leaving loopback without an ipv4 address for a while.
On SONiC startup and config reload interfaces-config and bgp services start in parallel and sometimes
fpmsyncd in bgp attempts bind to loopback while it does not have an address, fails with the log
`Exception "Cannot assign requested address" had been thrown in daemon`
and exits with rc 0.
```
root@sonic:/# supervisorctl status
fpmsyncd                         EXITED    Jul 20 05:04 AM
zebra                            RUNNING   pid 35, uptime 6:15:05
zsocket                          EXITED    Jul 20 05:04 AM
```
```
docker logs bgp
INFO exited: fpmsyncd (exit status 0; expected)
```
With fpmsyncd dead, configured routes do not appear in the database.

#### How I did it
Added ordering dependency on interfaces-config service into bgp.config
#### How to verify it
Itself the issue reproduces quite rarely, but one can gain the time interval between networking down and networking up in interfaces-config.sh like this:
```
diff --git a/files/image_config/interfaces/interfaces-config.sh b/files/image_config/interfaces/interfaces-config.sh
index f6aa4147a..87caceeff 100755
--- a/files/image_config/interfaces/interfaces-config.sh
+++ b/files/image_config/interfaces/interfaces-config.sh
@@ -63,7 +63,11 @@ done
 # Read sysctl conf files again
 sysctl -p /etc/sysctl.d/90-dhcp6-systcl.conf

-systemctl restart networking
+# systemctl restart networking
+
+systemctl start networking
+sleep 10
+systemctl stop networking

 # Clean-up created files
 rm -f /tmp/ztp_input.json /tmp/ztp_port_data.json
```
with this change the issue reproduces on every config reload.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [x] 202106
- [x] 202111
- [x] 202205

### DIFF
--- a/files/build_templates/per_namespace/bgp.service.j2
+++ b/files/build_templates/per_namespace/bgp.service.j2
@@ -7,6 +7,7 @@ After=updategraph.service
 BindsTo=sonic.target
 After=sonic.target
 Before=ntp-config.service
+After=interfaces-config.service
 StartLimitIntervalSec=1200
 StartLimitBurst=3
 


vboykox:
@lguohan Another thing we can do is to make the interfaces-config service run Before sonic target instead of After.
This will simplify dependencies for (future) services which bind to loopback.
From my preliminary runs this approach won't affect startup time.

vboykox:
@lguohan please review

yxieca (approved):

vboykox:
@yxieca @prsunny @lguohan please merge, I do not have merge permissions.

msosyak (approved):

dgsudharsan (approved):

dgsudharsan:
@vboykox Do we know why we started seeing this issue only recently? I believe some recent commits might have exposed this issue. Do you suspect any commit?

yxieca:
@vboykox  can you check the test failure?

vboykox:
> @vboykox Do we know why we started seeing this issue only recently? I believe some recent commits might have exposed this issue. Do you suspect any commit?

@dgsudharsan  I have not done such analysis, but I would look for changes in startup
times/order of other services and possibly on bootchart integration.
For any change that could put interface-config and bgp closer at startup - like database startup times improvement.

I know about the issue since the mid of June. It reproduces approx once-twice in two weeks considering everyday runs. Can not say if there was any recent change in the frequency.

Are you also observing the issue?

dgsudharsan:
> > @vboykox Do we know why we started seeing this issue only recently? I believe some recent commits might have exposed this issue. Do you suspect any commit?
> 
> @dgsudharsan I have not done such analysis, but I would look for changes in startup times/order of other services and possibly on bootchart integration. For any change that could put interface-config and bgp closer at startup - like database startup times improvement.
> 
> I know about the issue since the mid of June. It reproduces approx once-twice in two weeks considering everyday runs. Can not say if there was any recent change in the frequency.
> 
> Are you also observing the issue?

Yes we are also observing the same issue only recently but it doesn't reproduce that frequently as you mentioned.

PR #10087, opened 2022-02-25
[smartmontools] Install smartmontools with apt-get and upgrade it to 7.2-1

keboliu:
#### Why I did it
Smartmontools 6.6 has issue with read SMART info of nvme SSD
Smartmontools can be installed with apt-get, no need to build and instal

```
root@sonic:/home/admin# smartctl -a /dev/nvme0n1
smartctl 6.6 2017-11-05 r4594 [x86_64-linux-5.10.0-8-2-amd64] (local build)
Copyright (C) 2002-17, Bruce Allen, Christian Franke, [www.smartmontools.org](https://nam11.safelinks.protection.outlook.com/?url=http%3A%2F%2Fwww.smartmontools.org%2F&data=04%7C01%7Ckebol%40nvidia.com%7C7bd4d46f236441d936d608d9f695d1ad%7C43083d15727340c1b7db39efd9ccc17a%7C0%7C0%7C637811945587622712%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000&sdata=I6Q%2Fsbo1sucai1mMCygQrN5mEGS1ZrHTQYIbfIRqTw0%3D&reserved=0)

=== START OF INFORMATION SECTION ===
Model Number:                       SFPC020GM1EC2TO-I-5E-11P-STD
Serial Number:                      A0221030722410000027
Firmware Version:                   COT6OQ
PCI Vendor/Subsystem ID:            0x1dd4
IEEE OUI Identifier:                0x8c6078

……

0 +     512       0         0

=== START OF SMART DATA SECTION ===
Read NVMe SMART/Health Information failed: NVMe Status 0x02
```

#### How I did it

Use apt-get to install smartmontools 7.2-1
Remove previous make files for smartmontools 6.6

#### How to verify it

verify with "smartctl" can read out correct SMART info on NVME ssd.
verify "show platform ssdhealth" can still work

#### Which release branch to backport (provide reason below if selected)

- [ ] 201811
- [x] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

PR #11837, opened 2022-08-24
[pmon] Add smartmontools to pmon docker

cytsao1:
#### Why I did it
Some APIs in the pmon docker rely on smartmontools to execute correctly (such as reading from some sensor types), but smartmontools was moved outside of the docker earlier on.
example: in PMON, thermalctld updates the database with sensor temperature using get_temperature(), and the ssd sensor get_temperature is dependent on smartctl.

#### How I did it
Add smartmontools to platform monitor Dockerfile

#### How to verify it
Run smartctl inside pmon docker and verify any APIs dependent on it

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205

[pmon] Add smartmontools to pmon docker


cytsao1:
@abdosi Could this get reviewed?

bmridul:
@abdosi @rlhui , Pls follow up on this review.

abdosi:
@mlok-nokia can you please help review this ?

gechiang:
@cytsao1 you mentioned in the description:
"...  but smartmontools was moved outside of the docker earlier on."

Can you share which PR shows this was moved out of the PMON as I would like to find out what was the original reason for this moving outside of the docker.  Don't want to fix this and break whatever was the reason that caused it to move out in the first place if it was done to fix some issue...

cytsao1:
> @cytsao1 you mentioned in the description: "... but smartmontools was moved outside of the docker earlier on."
> 
> Can you share which PR shows this was moved out of the PMON as I would like to find out what was the original reason for this moving outside of the docker. Don't want to fix this and break whatever was the reason that caused it to move out in the first place if it was done to fix some issue...

Sure, here is the PR https://github.com/sonic-net/sonic-buildimage/pull/9607.

gechiang:
@alexrallen @liat-grozovik ,
Can you help review this PR as it seems that your previous PR (https://github.com/sonic-net/sonic-buildimage/pull/9607) is causing this issue that @cytsao1 is trying to fix by placing smartmontools back to pmon docker.  Do you see any issue for this change and if you do, please suggest a way so both are accomplished where it does not break the change introduced by your PR (https://github.com/sonic-net/sonic-buildimage/pull/9607) at the same time the need of this PR trying to address is also accomplished.
Thanks!

alexrallen:
As far as I am concerned... smartmontools can be in pmon without causing issues as long as the system does not have a problem with *both* smartmontools in pmon *and* on the host. However, it absolutely *must* be on the host. 

I am not sure if two instances in docker and on the host will "play nice" with each other but this 100% needs to be confirmed before merging this PR.

We need smartmontools on the host because smartmontools gives us important access into our SSD and it is poor architecture to require docker to be running in order to access these because we might need to either as the system starts up or when the system is going down (my original PR was made because the Mellanox firmware update flow was broken because docker had to stop before we performed some SSD operations).

dockers/docker-platform-monitor/Dockerfile.j2
```diff
@@ -26,7 +26,8 @@ RUN apt-get update &&   \
         psmisc          \
         python3-jsonschema \
         libpci3         \
-        iputils-ping
+        iputils-ping    \
+        smartmontools
```
alexrallen, line 30: Please confirm that this will not conflict with the smartmontools installed on the host. 

Its also worth noting that `rules/smartmontools.mk` should have probably been cleaned up with https://github.com/sonic-net/sonic-buildimage/pull/10087 but it was not, now might be a good time to do so now that smartmontools is getting some attention.

alexrallen (changes requested):

dockers/docker-platform-monitor/Dockerfile.j2
alexrallen: It might also be good practice to follow #10087 in hard-coding the version number to `smartmontools=7.2-1` so that we don't end up with mismatched versions. Again, not sure how particular its going to be about two instances of it running at the same time but it seems to me a version mismatch could exacerbate that.

cytsao1: I've tested, and installing smartmontools in the docker does not conflict with the host.
As suggested, I've added the same version number to the docker and removed the $(SMARTMONTOOLS) files that remained.

cytsao1: Both would have to be updated in the future. Unless there is a way for both to pick up the version from the same place (unsure if this is possible to be done easily), I can add a comment in both locations.

gechiang: @cytsao1 If this is not possible to only have one update and both are taken cared of, let's go with the "add comment in both locations" for now.
Thanks!

alexrallen: This is relatively easy to do using Makefile variables in SONiC. In fact this is a great suggestion that I would like to see implemented. 

The rest of the changes LGTM. I'll approve once the above is resolved.

gechiang: @alexrallen if You can share a sample coding in the Makefile variable and @cytsao1 can give it a try?
Thanks so much!

dgsudharsan (approved):

dockers/docker-platform-monitor/Dockerfile.j2
gechiang: > Both would have to be updated in the future. Unless there is a way for both to pick up the version from the same place (unsure if this is possible to be done easily), I can add a comment in both locations.

In the interest of time, can you please go ahead with adding the comments on both locations so that at least in the future whoever need to upgrade it to future releases, both locations will be updated to avoid inconsistencies.

If you can figure out how to do this in the Make file to cover it with a variable that would be even better.

cytsao1: I've added the comments for now.

gechiang:
@cytsao1 Can you try pull in the latest from sonic-buildimage to your view so that the fixes that was causing the PR build issue can also be in your view to pass the PR test?

cytsao1:
> @cytsao1 Can you try pull in the latest from sonic-buildimage to your view so that the fixes that was causing the PR build issue can also be in your view to pass the PR test?

@gechiang Ok, done.

cytsao1:
@gechiang Could you take a look at this again?

gechiang:
@lguohan, @yxieca, We have seen this PR test keeps failing after many attempts to rerun and even rebase the latest from sonic-buildimage...  The failures seems not related to this PR.  Is there something else we can try to allow this PR to pass the necessary test and be picked up?
Thanks!

gechiang (approved):

cytsao1:
@abdosi Could this get merged into 202205 branch as well, or should new PR be opened?

PR #11921, opened 2022-09-01
Define whether a test is required by code

yejianquan:
#### Why I did it
Define whether a test job is required before merging by code.
Let the failure of multi-asic and t0-sonic don't block pr merge.
The 'required' configuration can be modified by the owner in the future.
#### How I did it
Required:
t1-lag, t0
Not required:
multi-asic, t0-sonic
#### How to verify it
AZP itself verifies it.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -81,6 +81,10 @@ stages:
   - name: testbed_file
     value: vtestbed.csv
 
+# For every test job:
+# continueOnError: false means it's a required test job and will block merge if it fails
+# continueOnError: true means it's an optional test job and will not block merge even though it fails(unless a required test job depends on its result)
+
   jobs:
   - job:
     pool: sonictest
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -133,7 +133,7 @@ stages:
     pool: sonictest
     displayName: "kvmtest-t0-part1"
     timeoutInMinutes: 360
-
+    continueOnError: true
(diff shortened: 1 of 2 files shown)


azure-pipelines.yml
```diff
@@ -133,7 +133,7 @@ stages:
     pool: sonictest
     displayName: "kvmtest-t0-part1"
     timeoutInMinutes: 360
-
+    continueOnError: true
```
SuvarnaMeenakshi, line 136: is t0 not mandatory?

SuvarnaMeenakshi:
https://github.com/sonic-net/sonic-buildimage/blob/master/.azure-pipelines/run-test-template.yml#L55 - can this be removed/

azure-pipelines.yml
yejianquan: t0-part 1 and t0-part2 are set as not mandatory,
kvmtest-t0 will check the results of part1 and part2, and it's mandatory

yejianquan:
> https://github.com/sonic-net/sonic-buildimage/blob/master/.azure-pipelines/run-test-template.yml#L55 - can this be removed/

@SuvarnaMeenakshi  Yes, removed

SuvarnaMeenakshi (approved):

PR #11943, opened 2022-09-02
Restructuring of Makefile.work

Kalimuthu-Velappan:
- The Makefile.work becomes complex and it is very difficult to manages
the changes across branches.
- Restructured the Makefile.work and it becomes more readable.
- Added $(QUIET) option to turn on command echo mode through command line option.
- Exported the SONIC_OVERRIDE_BUILD_VARS variable, through which make
options can be set dynamically.
	Eg: make SONIC_OVERRIDE_BUILD_VARS='INCLUDE_NAT=y'

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205


Makefile
```diff
@@ -5,6 +5,13 @@ NOSTRETCH ?= 0
 NOBUSTER ?= 0
 NOBULLSEYE ?= 0
 
+export Q ?= @
+ifneq ($(QUIET),)
+override Q:=
+endif
+SONIC_OVERRIDE_BUILD_VARS += Q='$(Q)'
+export SONIC_OVERRIDE_BUILD_VARS
```
saiarcot895, line 13: What's the purpose of `SONIC_OVERRIDE_BUILD_VARS`? What do you mean by setting make options dynamically?

Makefile
```diff
@@ -5,6 +5,13 @@ NOSTRETCH ?= 0
 NOBUSTER ?= 0
 NOBULLSEYE ?= 0
 
+export Q ?= @
+ifneq ($(QUIET),)
```
saiarcot895, line 9: Can this be changed to `ifeq ($(QUIET),1)`? Setting `QUIET=0` shouldn't make it quiet.

Makefile
Kalimuthu-Velappan: In Makefile.work, SONIC_OVERRIDE_BUILD_VARS flag is already passed as a dynamic argument to the slave build container. 
      ```
SONIC_BUILD_INSTRUCTION :=  make \
                           -f slave.mk \
                              .....
                            $(SONIC_SLAVE_DOCKER_DRIVER=$(SONIC_SLAVE_DOCKER_DRIVER) \
                            **$(SONIC_OVERRIDE_BUILD_VARS)**
```

So setting any value to this variable will become an argument to the slave build container. As the name suggests, any make variable can be set dynamically. 

Already so many variables are being already passed as individual arguments. So the SONIC_OVERRIDE_BUILD_VARS variable will eliminate the future addition to it.

Makefile
Kalimuthu-Velappan: Sure, I will update the change

PR #11969, opened 2022-09-05
[POC] Test sonic-db-cli change

liuh-80:
This just a POC PR, need use this to test if sonic-db-cli change will break any existing code


liuh-80:
close because validate passed

PR #11985, opened 2022-09-06
[device/accton] Replace os.system and remove subprocess with shell=True

maipbui:

#### Why I did it
`subprocess.Popen()` and `subprocess.run()` is used with `shell=True`, which is very dangerous for shell injection.
#### How I did it
Replace `os` by `subprocess`
Remove unused functions

- [ ] 201811
- [X] 201911
- [X] 202006
- [X] 202012
- [X] 202106
- [X] 202111
- [X] 202205

### DIFF
--- a/device/accton/x86_64-accton_as4630_54pe-r0/sonic_platform/sfp.py
+++ b/device/accton/x86_64-accton_as4630_54pe-r0/sonic_platform/sfp.py
@@ -6,10 +6,9 @@
 #
 #############################################################################
 
-import os
 import time
 import sys
-
+import subprocess
 from ctypes import create_string_buffer
 
 try:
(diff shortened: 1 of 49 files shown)


device/accton/x86_64-accton_as4630_54pe-r0/sonic_platform/sfp.py
```diff
@@ -60,7 +59,7 @@ def get_eeprom_path(self):
         return self.port_to_eeprom_mapping[self.port_num]
 
     def __is_host(self):
-        return os.system(self.HOST_CHK_CMD) == 0
+        return subprocess.run(self.HOST_CHK_CMD, stdout=None, stderr=None).returncode == 0
```
qiluo-msft, line 62: >, stdout=None, stderr=None [](http://example.com/codeflow?start=47&length=26)

You are using default values for both parameters. So you can remove them.

qiluo-msft:
>         return self.__run_command(install_command)

It is not defined. @simonJi2018 Could you check whether it is a syntax error in original code?

---
Refers to: device/accton/x86_64-accton_as7116_54x-r0/sonic_platform/component.py:78 in 69b7f29. [](commit_id = 69b7f29915bf44592432643331d9288c6695406e, deletion_comment = False)

qiluo-msft (approved):

qiluo-msft (approved):

maipbui:
@ec-michael-shih @jeffhtgt @seanwu-ec @jostar-yang Could you help review and verify?

PR #11996, opened 2022-09-07
Remove jinja2_cache 

dprital:
#### Why I did it

As part of [Persistent log level HLD](https://github.com/sonic-net/SONiC/pull/1041) ,  LOGLEVEL_DB content is moved to CONFIG_DB. 
In addition, it was decided to remove jinja2_cache which currently appear on LOGLEVEL_DB 

This cache was added to speed up template rendering in start scripts. There were a lot of them rendered during system start. This caused a delay in warm boot LAG restore time. It was tested and verified that with and without the cache we don't see any difference in this timing now. It is probably due to a lot of other optimizations done to sonic-cfggen. Since there is no noticeable improvement made by j2 cache now it is safe to remove it.

#### How I did it

Remove redis_bcc.py file and and remove the bytcode_cache from sonic-sfggen

#### How to verify it

Warm boot was tested with \ without this jinja2_cache and it there is no difference in performance

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205


stepanblyschak (approved):

qiluo-msft (approved):

yxieca:
@dprital  this change cannot be cherry-picked cleanly to 202205. Please raise separate PR.

dprital:
> Warm boot was tested with \ without this jinja2_cache and it there is no difference in performance

Done --> https://github.com/sonic-net/sonic-buildimage/pull/12155

Thanks,
Dror.

PR #12000, opened 2022-09-08
01.Version-cache - restructuring of Makefile.work

Kalimuthu-Velappan:
- The Makefile.work becomes complex and it is very difficult to manage the changes across branches.
- Restructured the Makefile.work and it becomes more readable.
- Added $(QUIET) option to turn on command echo mode through command line option.
- Exported the SONIC_BUILD_VARS variable, through which make options can be set dynamically.
	Eg: make SONIC_BUILD_VARS='INCLUDE_NAT=y'

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205

### DIFF
--- a/Makefile.work
+++ b/Makefile.work
@@ -124,7 +124,7 @@ endif
 # Define a do-nothing target for rules/config.user so that when
 # the file is missing, make won't try to rebuld everything.
 rules/config.user:
-	@echo -n ""
+	$(Q)echo -n ""
 
 include rules/config
 -include rules/config.user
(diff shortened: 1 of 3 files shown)


adyeung:
@xumia @liushilongbuaa submitter has taken the time to split the original code PR to smaller submissions for review, pls help review and merge

xumia (approved):

Makefile.work
```diff
@@ -415,87 +498,75 @@ SONIC_BUILD_INSTRUCTION :=  make \
...
+endif
+
+ifeq "$(KEEP_SLAVE_ON)" "yes"
+SLAVE_SHELL={ /bin/bash; }
+endif
+
 .DEFAULT_GOAL :=  all
 
-%::
+%:: | sonic-build-hooks
```
liushilongbuaa, line 516: Why?

Kalimuthu-Velappan: There is no dependency check for the sonic-build-hooks target, so we are checking this target always.

Kalimuthu-Velappan:
I tried the below commands, but it doesn't generate the expected out file in the host.

#cat Dockerfile
  FROM debian:buster AS build-stage
  RUN mkdir -p /test
  RUN echo "MY CNT" > /test/myfile.txt

   FROM scratch AS export-stage
  COPY --from=build-stage /test /


#docker build --no-cache -o out . 
#ls
Dockerfile

Makefile.work
```diff
@@ -415,87 +498,75 @@ SONIC_BUILD_INSTRUCTION :=  make \
 
 .PHONY: sonic-slave-build sonic-slave-bash init reset
 
+
+ifeq ($(filter clean,$(MAKECMDGOALS)),)
+COLLECT_BUILD_VERSION = { DBGOPT='$(DBGOPT)' scripts/collect_build_version_files.sh $$?; }
+endif
```
kv-y, line 504: @Kalimuthu-Velappan 

`make clean` is broken now because of these changes:
```
bash: -c: line 1: syntax error near unexpected token `;'
bash: -c: line 1: `make -f slave.mk PLATFORM= PLATFORM_ARCH=amd64 
....
MIRROR_URLS= MIRROR_SECURITY_URLS=  Q=@ clean; ; '
make[1]: *** [Makefile.work:530: clean] Error 2
```
If COLLECT_BUILD_VERSION is not set we get double ;; in 
`bash -c "$(FIX_DOCKER_GROUP); $(SONIC_BUILD_INSTRUCTION) $@; $(COLLECT_BUILD_VERSION); $(SLAVE_SHELL)"`

kv-y: Also I'm not fully understand why we need this specifc handling for `clean` command.
I mean why we filter only `clean` here, but not `distclean` e.g.

PR #12009, opened 2022-09-08
[Build] Fix dbus-run-session command not found issue when install dbus-python

xumia:
#### Why I did it
Fix the dbus-pyhon installation failure when building docker-sonic-vs, caused by the command dbus-run-session not found.
The command "dbus-run-session" should be the new dependency introduced in dbus-python 1.3.2, the old version 1.2.18 does not have the issue.

#### How I did it
Install the dbus debian package which contains the command dbus-run-session.
It is not a blocking issue on release branches. The release branches with reproducible build feature can avoid such issue in official builds and PR builds, it only block the version upgrade (trying to upgrade from 1.2.18 to 1.3.2).

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012
- [x] 202106
- [x] 202111
- [x] 202205

### DIFF
--- a/platform/vs/docker-sonic-vs/Dockerfile.j2
+++ b/platform/vs/docker-sonic-vs/Dockerfile.j2
@@ -72,7 +72,8 @@ RUN apt-get install -y net-tools \
                        {%- if ENABLE_ASAN == "y" %}
                        libasan5 \
                        {%- endif %}
-                       libsystemd0
+                       libsystemd0 \
+                       dbus
 
 # Install redis-server
 {% if CONFIGURED_ARCH == "armhf" %}


platform/vs/docker-sonic-vs/Dockerfile.j2
```diff
@@ -72,7 +72,8 @@ RUN apt-get install -y net-tools \
                        {%- if ENABLE_ASAN == "y" %}
                        libasan5 \
                        {%- endif %}
-                       libsystemd0
+                       libsystemd0 \
+                       dbus
```
StormLiangMS, line 76: @xumia just one quick question, where to specify the version of dbus for this branch?

StormLiangMS (approved):
LGTM

platform/vs/docker-sonic-vs/Dockerfile.j2
xumia: @StormLiangMS , you mean for reproducible build right?
We can specify version in release branches, such as 202012, 202205, not in master branch.

prsunny:
hi @xumia , can you also raise PRs to 202012/202205?

xumia:
> hi @xumia , can you also raise PRs to 202012/202205?

Has already merged in 202205 branch, no affect 202012.
git diff origin/202012..origin/202106 src/sonic-host-services/setup.py
image

Issue #11995, opened 2022-09-07
'eth0' oper down causing BGP default route missing


prsunny:
With a stable system with BGP learnt default route, and oper down of mgmt port (eth0) is causing default route deleted from APP_DB.  This issue is seen on 202205 as well. 

Logs:

/var/log/syslog.444.gz:Aug 31 17:52:41.788610  INFO kernel: [2430739.815597] tg3 0000:04:00.0 eth0: Link is down

swss rec
2022-08-31.17:52:41.801386|ROUTE_TABLE:2603:10e2:11:154f::/64|DEL
2022-08-31.17:52:41.801423|ROUTE_TABLE:25.192.243.64/26|DEL
2022-08-31.17:52:41.801458|ROUTE_TABLE:0.0.0.0/0|DEL
2022-08-31.17:52:41.801492|ROUTE_TABLE:::/0|DEL

#### Steps to reproduce the issue:
1. Reboot the system with 202205 image
2. Learn BGP default route (0.0.0.0/0) from peers
3. Ensure default route is also present in default table for eth0 ('ip route show table default')
4. Bring down eth0 from management switch (simulate oper-down on DUT)
5. Check for default route in APP_DB.

#### Describe the results you received:
APP_DB default route is removed

#### Describe the results you expected:
APP_DB default route must not be changed during eth0 link-up/down events

#### Output of `show version`:

```
(paste your output here)
```
#### Output of `show techsupport`:

```
(paste your output here or download and attach the file here )
```

arlakshm:
Logs from debugging so far. On bootup we see the  kernel routes in `zebra` for the `default` table
```
admin@sonic:~$ ip link show dev eth0
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast state UP mode DEFAULT group default qlen 1000
    link/ether ec:f4:bb:fe:80:84 brd ff:ff:ff:ff:ff:ff
admin@sonic:~$ vtysh -c "show ip route table all" | more
Codes: K - kernel route, C - connected, S - static, R - RIP,
       O - OSPF, I - IS-IS, B - BGP, E - EIGRP, N - NHRP,
       T - Table, v - VNC, V - VNC-Direct, A - Babel, D - SHARP,
       F - PBR, f - OpenFabric,
       > - selected route, * - FIB route, q - queued, r - rejected, b - backup

VRF default table 253:
[...]

PR #12018, opened 2022-09-08
[zebra] ignore route from default table

yxieca:
#### Why I did it
Fix issue #11995 : when management port goes down, zebra will remove the default route and causing ASIC to blackhole IO.

#### How I did it
Ignore route changes in default table.

#### How to verify it
Cause management port to go down and verify that default route is not removed.


prsunny:
@hasan-brcm, please review

prsunny (approved):

Stephenxf:
The issue https://github.com/sonic-net/sonic-buildimage/issues/12753 is related to this change. PR https://github.com/sonic-net/sonic-buildimage/pull/12754 is raised to address it.

PR #12033, opened 2022-09-09
Fix dhcp option buffer issue

ganglyu:
Signed-off-by: Gang Lv ganglv@microsoft.com

#### Why I did it
Current isc-dhcp uses below code to remove DHCP option:
`memmove(sp, op, op[1] + 2);`
`sp += op[1] + 2;`

strip1

sp points to the option to be stripped, we can call it as option S.
op points to the option after options S, we can call it as option O.
DHCP option is a typical type-length-value structure, the first byte is type, the second byte is length, and remain parts are value.
In this case, option O length is bigger than option S, and more than 2 bytes, after the memmove, we will get this result:

strip2

Now Option S and Option O are overwritten, op[1] was the length of Option O, and it's modified after memmove.
But current implementation is still using op[1] as length to update sp (sp+=op[1]+2), so we get the wrong sp.

#### How I did it
Create patch from https://github.com/isc-projects/dhcp
The new impelementation use mlen to store the length of Option O before memmove, that's how it fixed the bug.
`size_t mlen = op[1] + 2;`
`memmove(sp, op, mlen);`
`sp += mlen;`

#### How to verify it
I have a PR for sonic-mgmt to cover this issue:
https://github.com/sonic-net/sonic-mgmt/pull/6330

- [ ] 201811
- [x] 201911
- [x] 202006
- [x] 202012
- [x] 202106
- [x] 202111
- [x] 202205

### DIFF
--- a/src/isc-dhcp/patch/0013-Fix-dhcrelay-agent-option-buffer-pointer-logic.patch
+++ b/src/isc-dhcp/patch/0013-Fix-dhcrelay-agent-option-buffer-pointer-logic.patch
@@ -1,0 +1,53 @@
+From 0a2f9a62bceb90b0d30461add2e25c4ce7a24547 Mon Sep 17 00:00:00 2001
+From: Thomas Markwalder <tmark@isc.org>
+Date: Fri, 20 Dec 2019 10:11:54 -0500
+Subject: [PATCH] [#71] Fix dhcrelay agent option buffer pointer logic
+
+relay/dhcrelay.c
+    strip_relay_agent_options()
+    strip_relay_agent_options()
+    - corrected buffer pointer logic
+
+---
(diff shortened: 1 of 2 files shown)


qiluo-msft:
Possible to test in KVM?

ganglyu:
> Possible to test in KVM?

I have used kvm and sonic-mgmt to verify this issue.

qiluo-msft:
Please merge it together with sonic-mgmt PR.

qiluo-msft (approved):
LGTM

ganglyu:
Add end2end test:
https://github.com/sonic-net/sonic-mgmt/pull/6330

kellyyeh (approved):

PR #12064, opened 2022-09-13
Fix to improve hostname handling

devpatha:
#### Why I did it
If config_db.json is missing hostname entry, hostname-config.sh ends up deleting existing entry too and hostname changes to default 'localhost'

#### How I did it
If hostname is not available from config_db.json, do not edit /etc/hostname and exit

#### How to verify it
with missing hostname entry in config_db.json, /etc/hostname entry is not cleared

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012 for Cisco 8102 
- [ ] 202106
- [ ] 202111
- [ ] 202205

### DIFF
--- a/files/image_config/hostname/hostname-config.sh
+++ b/files/image_config/hostname/hostname-config.sh
@@ -4,8 +4,8 @@ CURRENT_HOSTNAME=`hostname`
 HOSTNAME=`sonic-cfggen -d -v DEVICE_METADATA[\'localhost\'][\'hostname\']`
 
 if [ -z "$HOSTNAME" ] ; then
-       echo "Missing hostname in the config file!"
-       exit 0
+       echo "Missing hostname in the config file, setting to default 'sonic'"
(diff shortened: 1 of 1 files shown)


files/image_config/hostname/hostname-config.sh
```diff
@@ -3,6 +3,11 @@
 CURRENT_HOSTNAME=`hostname`
 HOSTNAME=`sonic-cfggen -d -v DEVICE_METADATA[\'localhost\'][\'hostname\']`
 
+if [ -z "$HOSTNAME" ] ; then
+       echo "Missing hostname in the config file!"
+       exit 0
+fi
```
yxieca, line 9: Why not set a good default hostname instead?

devpatha: shall we default to "sonic"?

yxieca (approved):

PR #12066, opened 2022-09-13
Add support for platform topology configuration service

anamehra:
This service invokes the platform plugin for platform specific topology
    configuration.
    The path for platform plugin script is:
    /usr/share/sonic/device/$PLATFORM/plugins/config-topology.sh
    If the platform plugin is not available, this service does nothing.

#### Why I did it
For the chassis, the platform requires to generate the chassis topology from the available inventory. Based on the inventory config files for the platform and ASICs are generated. Some configs are global as well as ASIC specific. This needs to be done on both RP and LC and requires the availability of database-chassis and database services.
 Adding this systemd dependency in platform side service unit files causes first boot issues as the database service is not known when rc.local starts the platform service.
Secondly, sonic-systemd-generator generates the dependency on multi-instance services during boot based on num_asics and for that Sonic systemd generator needs to know about the platform services. For example, if a service defines:
After=database.service
sonic systemd generator creates the dependency for the global database and all namespace database instances
After=database.service
After=database@0.service
After=database@1.service

This is not possible if the service is added as a platform-side service.

This is also helpful in use cases where a platform service may require to be restarted when database-chassis or database service restarts. Without a proper dependency defined on Sonic side, the platform service won't restart.

Systemd-sonic-generator issue:
The check for multi instance service, checks if the srvice name string is present in the multi_instance_service[].
This results in database-chassis marked as multi instance service when added as dependency.

#### How I did it
Added a common service that defines the dependency on Sonic build side to handle the above limitations. The service calls a platform plugin, if present. If the platform plugin is not available, the service exits as successful and will do nothing for the platform.

#### How to verify it
1. Load the image with changes.
2. check systemctl status config-topology.service logs
3. Validated that the services starts after database services and before config-setup service.

Fix for systemd-sonic-generator:
The service name may contain @.service or .service. Remove these
postfixes and extract service name. Compare service name for the absolute
match in multi_instance_services[].
This is to prevent services like database-chassis and systemd-timesyncd marked
as multi instance services as they contain strings 'database' and 'syncd' respectively
which are multi instance services in multi_instance_services[].

UT logs:
[UT.txt](https://github.com/sonic-net/sonic-buildimage/files/10396524/UT.txt)

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [x] 202205

### DIFF
--- a/files/image_config/config-topology/config-topology.service
+++ b/files/image_config/config-topology/config-topology.service
@@ -4,7 +4,6 @@ After=database.service
 After=database-chassis.service
 Requires=database.service
 Requires=database-chassis.service
-PartOf=database.service
 Before=config-setup.service
 
 [Service]
--- a/files/image_config/config-topology/config-topology.service
+++ b/files/image_config/config-topology/config-topology.service
@@ -1,7 +1,9 @@
 [Unit]
 Description=Platform topology configuration service
-Requires=database.service
 After=database.service
+After=database-chassis.service
+Requires=database.service
+Requires=database-chassis.service
 PartOf=database.service
 Before=config-setup.service
 
--- a/files/image_config/config-topology/config-topology.service
+++ b/files/image_config/config-topology/config-topology.service
@@ -1,0 +1,17 @@
+[Unit]
+Description=Platform topology configuration service
+Requires=database.service
+After=database.service
+PartOf=database.service
+Before=config-setup.service
+
+[Service]
+Type=oneshot
+User=root
+RemainAfterExit=yes
+ExecStart=/usr/bin/config-topology.sh start
+ExecStop=/usr/bin/config-topology.sh stop
+
+[Install]
+WantedBy=multi-user.target
+
--- a/src/systemd-sonic-generator/systemd-sonic-generator.c
+++ b/src/systemd-sonic-generator/systemd-sonic-generator.c
@@ -104,7 +104,22 @@ static int get_target_lines(char* unit_file, char* target_lines[]) {
 static bool is_multi_instance_service(char *service_name){
     int i;
     for(i=0; i < num_multi_inst; i++){
-        if (strstr(service_name, multi_instance_services[i]) != NULL) {
+        /*
+         * The service name may contain @.service or .service. Remove these
+         * postfixes and extract service name. Compare service name for absolute
+         * match in multi_instance_services[].
+         * This is to prevent services like database-chassis and systemd-timesyncd marked
+         * as multi instance services as they contain strings 'database' and 'syncd' respectively
+         * which are multi instance services in multi_instance_services[].
+         */
+        char *token = strtok(service_name, "@");
+        if (token) {
+            if (strstr(token, ".service") != NULL) {
+                /* If we are here, service_name did not have '@' delimiter but contains '.service' */
+                token = strtok(service_name, ".");
+            }
+        }
+        if (strncmp(service_name, multi_instance_services[i], strlen(service_name)) == 0) {
             return true;
         }
     }
--- a/src/systemd-sonic-generator/systemd-sonic-generator.c
+++ b/src/systemd-sonic-generator/systemd-sonic-generator.c
@@ -112,11 +112,12 @@ static bool is_multi_instance_service(char *service_name){
          * as multi instance services as they contain strings 'database' and 'syncd' respectively
          * which are multi instance services in multi_instance_services[].
          */
-        char *token = strtok(service_name, "@");
+        char *saveptr;
+        char *token = strtok_r(service_name, "@", &saveptr);
         if (token) {
             if (strstr(token, ".service") != NULL) {
                 /* If we are here, service_name did not have '@' delimiter but contains '.service' */
-                token = strtok(service_name, ".");
+                token = strtok_r(service_name, ".", &saveptr);
             }
         }
         if (strncmp(service_name, multi_instance_services[i], strlen(service_name)) == 0) {
--- a/files/build_templates/config-setup.service.j2
+++ b/files/build_templates/config-setup.service.j2
@@ -2,6 +2,8 @@
 Description=Config initialization and migration service
 After=rc-local.service
 After=database.service
+After=config-topology.service
+Requires=config-topology.service
 Requires=database.service
 {% if sonic_asic_platform == 'mellanox' -%}
 Requires=hw-management.service
--- a/files/build_templates/sonic_debian_extension.j2
+++ b/files/build_templates/sonic_debian_extension.j2
@@ -554,6 +554,11 @@ echo "topology.service" | sudo tee -a $GENERATED_SERVICE_FILE
 sudo cp $IMAGE_CONFIGS/topology/topology.sh $FILESYSTEM_ROOT/usr/bin
 {%- endif %}
 
+# Copy platform topology configuration scripts
+sudo cp $IMAGE_CONFIGS/config-topology/config-topology.service $FILESYSTEM_ROOT_USR_LIB_SYSTEMD_SYSTEM
+echo "config-topology.service" | sudo tee -a $GENERATED_SERVICE_FILE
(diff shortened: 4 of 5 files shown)


anamehra:
Hi @abdosi , FYI-

rlhui:
Can we elaborate the advantage/reason of adding this? Thanks.

files/image_config/config-topology/config-topology.service
```diff
@@ -0,0 +1,17 @@
...
+[Service]
+Type=oneshot
+User=root
+RemainAfterExit=yes
+ExecStart=/usr/bin/config-topology.sh start
+ExecStop=/usr/bin/config-topology.sh stop
+
+[Install]
+WantedBy=multi-user.target
+
```
SuvarnaMeenakshi, line 17: Should the swss/syncd service files also be modified to start after config-topology.service?

rlhui:
> Can we elaborate the advantage/reason of adding this? Thanks.

@anamehra , reminder on this, thanks

rlhui:
@anamehra  - all tests are failing

anamehra:
> > Can we elaborate the advantage/reason of adding this? Thanks.
> 
> @anamehra , reminder on this, thanks

Hi @rlhui , I added more details after the chassis group discussion.

anamehra:
@abdosi , the logs from the build are lost, could you please rerun the checks?

anamehra:
Hi @SuvarnaMeenakshi , I have added explicit dependency in config-setup.service to prevent application service s from starting when config-topology fails. Please review.

abdosi:
@SuvarnaMeenakshi  can you please review and sign-off for this PR ? There is some fix in systemd-generator also.

src/systemd-sonic-generator/systemd-sonic-generator.c
```diff
@@ -104,7 +104,23 @@ static int get_target_lines(char* unit_file, char* target_lines[]) {
...
+         * match in multi_instance_services[].
+         * This is to prevent services like database-chassis and systemd-timesyncd marked
+         * as multi instance services as they contain strings 'database' and 'syncd' respectively
+         * which are multi instance services in multi_instance_services[].
+         */
+        char *saveptr;
+        char *token = strtok_r(service_name, "@", &saveptr);
+        if (token) {
+            if (strstr(token, ".service") != NULL) {
+                /* If we are here, service_name did not have '@' delimiter but contains '.service' */
```
SuvarnaMeenakshi, line 119: Do we need this if condition here? or is this added for additional check?

Ideally, we will not enter this condition as "service_name" that is being passed to this function should not include '.service.
https://github.com/sonic-net/sonic-buildimage/blob/master/src/systemd-sonic-generator/systemd-sonic-generator.c#L215
Test code :
#include<string.h>
#include<stdio.h>
int main(){
        char service_name[] = "swss.service";
        char *save_ptr2;
        char *type, *service;
        service = strtok_r(service_name, ".", &save_ptr2);
        type = strtok_r(NULL, "\n", &save_ptr2);
        printf("\nservice=%s, type=%s", service, type);
}
output:
service=swss, type=service
The value is service is passed to is_multi_instance_service()

files/image_config/config-topology/config-topology.service
```diff
@@ -0,0 +1,19 @@
+[Unit]
+Description=Platform topology configuration service
+After=database.service
+After=database-chassis.service
+Requires=database.service
+Requires=database-chassis.service
```
SuvarnaMeenakshi, line 6: Do we need After and Require dependency to database-chassis?
database service will start only after database-chassis service, so the ordering should automatically be taken care of without this explicit dependency, right?

files/image_config/config-topology/config-topology.service
```diff
@@ -0,0 +1,19 @@
...
+PartOf=database.service
```
SuvarnaMeenakshi, line 7: Do we need "PartOf" when "Requires" is added?

files/image_config/config-topology/config-topology.service
anamehra: Requires is required for both as the service needs to restart on database or database-chassis restart. It depends on both.
Regarding "After=", defining explicit dependency on order should be fine.

SuvarnaMeenakshi: without explicit dependency to database-chassis, if database-chassis restarts, database will restart and in turn config-topology should restart. correct?

anamehra: database service unit file does not define "Requires" depenedency on database-chassis. Database service does not restart when database-chassis restarts.

Here are the logs where I restart database-chassis couple of mins ago, but database service did not restart:
root@sonic:/home/cisco# systemctl status database-chassis.service 
● database-chassis.service - database-chassis container
     Loaded: loaded (/lib/systemd/system/database-chassis.service; enabled-runtime; vendor preset: enabled)
     Active: active (running) since Thu 2023-01-19 23:42:30 UTC; 1min 44s ago
   Main PID: 2766315 (database.sh)
      Tasks: 12 (limit: 38080)
     Memory: 23.5M
     CGroup: /system.slice/database-chassis.service
             ├─2766315 /bin/bash /usr/bin/database.sh wait chassisdb
             └─2766324 docker wait database-chassis

Jan 19 23:42:26 sonic systemd[1]: Starting database-chassis container...
Jan 19 23:42:27 sonic database.sh[2766041]: Starting existing database-chassis container
Jan 19 23:42:28 sonic database.sh[2766093]: database-chassis
Jan 19 23:42:29 sonic database.sh[2766215]: Invalid database name input : 'CHASSIS_APP_DB'
Jan 19 23:42:29 sonic database.sh[2766215]: Unable to connect to redis: Cannot assign requested address
Jan 19 23:42:30 sonic systemd[1]: Started database-chassis container.
root@sonic:/home/cisco# systemctl status database
● database.service - Database container
     Loaded: loaded (/lib/systemd/system/database.service; enabled; vendor preset: enabled)
    Drop-In: /etc/systemd/system/database.service.d
             └─auto_restart.conf
     Active: active (running) since Sat 2023-01-14 15:51:46 UTC; 5 days ago
   Main PID: 3127 (database.sh)
      Tasks: 13 (limit: 38080)
     Memory: 32.1M
     CGroup: /system.slice/database.service
             ├─3127 /bin/bash /usr/bin/database.sh wait
             └─3201 docker wait database

SuvarnaMeenakshi: Thank you for the details!

src/systemd-sonic-generator/systemd-sonic-generator.c
SuvarnaMeenakshi: @anamehra  - this comment is pending, the other changes look ok to me.

anamehra: I saw <>.service being passed in some instances in my UT and that is why I added this additional check.
before L215, it is doing a strdup on word and using service_name to run strtok_r but passing word.

There is another instance https://github.com/sonic-net/sonic-buildimage/blob/master/src/systemd-sonic-generator/systemd-sonic-generator.c#L267

anamehra: Hi @SuvarnaMeenakshi , please let me know if you have any more queries. Thanks

abdosi:
@SuvarnaMeenakshi  can you please review again

src/systemd-sonic-generator/systemd-sonic-generator.c
SuvarnaMeenakshi: Got it, thank you for the pointer. 
Seems like this  change will handle both situation on service_name.type or service_name being passed.

SuvarnaMeenakshi (approved):
lgtm

abdosi (approved):

yxieca:
@anamehra @rlhui why is this change not needed for 202211 branch?

PR #12108, opened 2022-09-19
[inventec] Replace os.system and remove subprocess with shell=True

maipbui:

Dependency: [https://github.com/sonic-net/sonic-buildimage/pull/12065](https://github.com/sonic-net/sonic-buildimage/pull/12065)
#### Why I did it
1. `getstatusoutput` is used without a static string and it uses `shell=True`
2. `subprocess()` - when using with `shell=True` is dangerous. Using subprocess function without a static string can lead to command injection.
3. `os` - not secure against maliciously constructed input and dangerous if used to evaluate dynamic content.
#### How I did it
1. use `getstatusoutput` without shell=True
2. `subprocess()` - use `shell=False` instead. use an array string. Ref: [https://semgrep.dev/docs/cheat-sheets/python-command-injection/#mitigation](https://semgrep.dev/docs/cheat-sheets/python-command-injection/#mitigation)
3. `os` - use with `subprocess`

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205

### DIFF
--- a/platform/broadcom/sonic-platform-modules-inventec/common/utils/asic_monitor.py
+++ b/platform/broadcom/sonic-platform-modules-inventec/common/utils/asic_monitor.py
@@ -17,13 +17,11 @@ options:
 
 try:
     import os
-    import commands
-    import sys, getopt
-    import logging
     import re
     import time
     import syslog
     from sonic_sfp.bcmshell import bcmshell
+    from sonic_py_common.general import getstatusoutput_noshell
     
 except ImportError as e:
     raise ImportError("%s - required module not found" % str(e))
@@ -53,10 +51,10 @@ def initialLoop():
             bcm_obj = BCMUtil()
             bcm_obj.execute_command("echo")
             initialNotOK = False
-            print bcm_obj
+            print(bcm_obj)
             log_message( syslog.LOG_INFO, "BCMUtil Object initialed successfully" )  
-        except Exception, e:   
-            print "Exception. The warning is {0}".format(str(e)) 
+        except Exception as e:   
+            print("Exception. The warning is {0}".format(str(e)))
             time.sleep(10)
             
 class BCMUtil(bcmshell):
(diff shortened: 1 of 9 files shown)


platform/broadcom/sonic-platform-modules-inventec/common/utils/asic_monitor.py
```diff
@@ -66,7 +64,7 @@ class BCMUtil(bcmshell):
     
     def get_platform(self):
         if self.platform is None:
-            self.platform = os.popen("uname -n").read().strip()
```
qiluo-msft, line 69: >.strip() [](http://example.com/codeflow?start=55&length=8)

Do you want to keep `.strip()`? It also applies to blank char.

platform/broadcom/sonic-platform-modules-inventec/d6332/utils/inventec_d6332_util.py
```diff
@@ -236,8 +237,9 @@ def system_install(boot_option):
 	      return status
     for addr_offset in range (0,FAN_NUM):
         addr=FAN_VPD_ADDR_BASE+addr_offset
-        cmd = "i2cdetect -y "+str(FAN_VPD_CHANNEL)+" "+str(addr)+" "+str(addr)+" | grep "+str(hex(addr)).replace('0x','')
-        result=os.system(cmd)
+        cmd1 = ["i2cdetect", "-y", str(FAN_VPD_CHANNEL), str(addr), str(addr)]
+        cmd2 = ["grep", str(hex(addr)).replace('0x','')]
```
qiluo-msft, line 241: >replace [](http://example.com/codeflow?start=39&length=7)

better way: [https://stackoverflow.com/a/16414603/2514803](https://stackoverflow.com/a/16414603/2514803)

maipbui: Fixed, thanks!

qiluo-msft (approved):

maipbui:
@klhaung @david-xk @CynthiaINV Could you help review and verify?

linalvin (approved):

PR #12127, opened 2022-09-20
[chassis-packet] fix the issue of internal ip arp not getting resolved.

abdosi:
Why I did it:
 Fix the issue where arp_update will not ping some of the ip's even
 though they are in failed state since `grep` of that ip on `ip neigh show`
command does not do exact word match and can return multiple match.

How I fixed it:
Using `-w` option of grep.

```
Before for example search of 10.0.0.5 will result in multiple entries and that causes failure
admin@str2-xxxxx-lc1-3:~$ sudo ip -4 -n asic0 neigh show | grep  "10.0.0.5"
10.0.0.5 dev PortChannel104 lladdr 3a:8d:6a:dd:b6:0d REACHABLE
10.0.0.55 dev Ethernet27 lladdr 06:5c:07:85:41:e2 REACHABLE

After fix (using -w option) we get exact match
admin@str2-xxxx-lc1-3:~$ sudo ip -4 -n asic0 neigh show | grep -w "10.0.0.5"
10.0.0.5 dev PortChannel104 lladdr 3a:8d:6a:dd:b6:0d REACHABLE
```

How I verify:
Manually after fix all the ip's arp are resolved.


abdosi:
cc @anamehra

files/scripts/arp_update
```diff
@@ -22,10 +22,10 @@ while /bin/true; do
       fi
       for nexthop in $STATIC_ROUTE_NEXTHOPS; do
           if [[ $nexthop == *"."* ]]; then
-              neigh_state=( $(ip -4 neigh show | grep $nexthop | tr -s ' ' | cut -d ' ' -f 3,4) )
+              neigh_state=( $(ip -4 neigh show | grep -w $nexthop | tr -s ' ' | cut -d ' ' -f 3,4) )
```
prsunny, line 25: hi @abdosi , can you please provide in description with an example of what was before and what is return with this options. Will be good to refer in future for such changes

abdosi: @prsunny updated the PR description with the example

prsunny (approved):

prsunny:
@theasianpianist for viz

PR #12164, opened 2022-09-22
[Build] Use apt-get to predictably support dependency ordered configuration of lazy packages

lornelong:
#### Why I did it

- The current lazy installer relies on a filename sort for both unpack and configuration steps. When systemd services are configured [started] by multiple packages the order is by filename not by the declared package dependencies.   This can cause the start order of services to differ between first-boot and subsequent boots.   Declared systemd service dependencies further exacerbate the issue (e.g. blocking the first-boot script).

- The current installer leaves packages un-configured if the package dependency order does not match the filename order.

- This also fixes a trivial bug in #10923 where externally downloaded dependencies are duplicated across lazy package device directories.

#### How I did it
Changed the staging and first-boot scripts to use apt-get:

`dpkg -i /host/image-$SONIC_VERSION/platform/$platform/*.deb`

becomes

`apt-get -y install /host/image-$SONIC_VERSION/platform/$platform/*.deb`

when dependencies are detected during image staging.

#### How to verify it
Apt-get  [critical rules](https://github.com/Debian/apt/blob/main/apt-pkg/orderlist.cc)

Add a Depends= to the control information of a package.  Grep the syslog for rc.local between images and observe the configuration order of packages change.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ X] 202012
- [ ] 202106
- [ ] 202111
- [ X] 202205

Use apt-get to predictably support dependency ordered configuration of lazy packages


lornelong:
@xumia

files/build_templates/sonic_debian_extension.j2
```diff
@@ -646,12 +646,19 @@ sudo mkdir -p $FILESYSTEM_ROOT/$PLATFORM_DIR/common
...
 
 sudo dpkg --root=$FILESYSTEM_ROOT -P {{ debname }}
 
 {% endfor %}
+# create a trivial apt repo if any of the debs have dependencies, including between lazy debs
+if [ $(for f in $FILESYSTEM_ROOT/$PLATFORM_DIR/common/*.deb; do \
+           sudo dpkg -I $f | grep "Depends:\|Pre-Depends:"; done | wc -l) -gt 0 ]; then
+    sudo dpkg-scanpackages $FILESYSTEM_ROOT/$PLATFORM_DIR/common | \
+        sudo gzip | sudo tee $FILESYSTEM_ROOT/$PLATFORM_DIR/common/Packages.gz > /dev/null
+fi
```
xumia, line 661: Can we list out all the packages and the order that are required for a specific platform? Can We simple installed the in sequence?
For instance, we have 3 packages, A.deb, B.deb, and C.deb, A-depends->B-depends->C.
Packages.list
```
C.deb B.deb A.deb
```
We can reduce the complex logic to install the packages, line 287 to 300, make thing simple.
It is only one of the suggestions, not sure if it works for you.

lornelong: @xumia Just looking for predictable packaging from the vendor side. I will try refactoring as you suggest.  apt-get has different orders for the unpack and configure steps.  Unpack remains a lexicographical sort except when PreDepends= is used.  Configure is a topological sort by Depends= .  Both ordered lists are in the stdout of apt install -s (simulate) but screen scraping is sketchy business....maybe there is another way to capture.   But in the end, apt is there and its logic is maintained and tested by several distros.

lornelong: @xumia Since apt wraps dpkg I wrote a wrapper for dpkg to see if we could get the list reliably.   This methodology seems to work, but I would like your input before I invest much more.

/usr/bin/dpkg wrapper:

```
#!/bin/bash
A="$@"
#echo "dpkg $A" >> /var/log/apt.trace
if [[ $A == *"unpack"* ]]; then
    if [[ $A == *"recursive"* ]]; then 
        # apt decide to rename the packages in temp space
        ls ${A##* } | cut -d '-' -f 2-  >> /var/log/dpkg.order
    else
        # apt listed the debs on the command line
        for word in $A; do
            if [[ $word == *".deb" ]]; then
                echo ${word##*/} >> /var/log/dpkg.order
            fi
        done
    fi	
fi
/usr/bin/dpkg.bin $A
```

The installer becomes (an appropriate filename when done):

`dpkg -I $(cat /var/log/dpkg.order)`

Of course there is all the other setup moved to the build template and a bunch of testing.  Thoughts?

xumia: How about changing the installation script, unpack the packages, and then configure them.
```
dpkg --force-all --unpack *
dpkg --configure -a
```

Test 1: Failed
```
docker run --rm -it debian:bullseye bash
Inside of the docker container:
cd ~
apt-get update
apt-get download $(apt-cache depends --recurse --no-recommends --no-suggests --no-conflicts --no-breaks --no-replaces --no-enhances python3-dev | grep "^\w")
dpkg -i *

```
Error messages as below:
```
Processing triggers for libc-bin (2.31-13+deb11u4) ...
 python3-minimal_3.9.2-3_amd64.deb
 python3_3.9.2-3_amd64.deb
 python3-dev
 python3-distutils
 python3-lib2to3
```

Test 2: succeeded
```
docker run --rm -it debian:bullseye bash
Inside of the docker container:
cd ~
apt-get update
apt-get download $(apt-cache depends --recurse --no-recommends --no-suggests --no-conflicts --no-breaks --no-replaces --no-enhances python3-dev | grep "^\w")
dpkg -i *
```

Only some warning messages:
```
dpkg: regarding python3-minimal_3.9.2-3_amd64.deb containing python3-minimal, pre-dependency problem:
 python3-minimal pre-depends on python3.9-minimal (>= 3.9.2-0~)
  python3.9-minimal is not installed.

dpkg: warning: ignoring pre-dependency problem!


```

lornelong: Thats pretty much what apt does under the covers, except apt sorts the dependent packages in order first. Apt either renames the debs with numeric prefixes in temp space or supplies them on the command line in order.  The order sets the configure order.  The problem we have is that different debs deliver different systemd services and we need the postint scripts [which start them] to be run in order.

lornelong: After reviewing with @xumia The late binding in the installer appears to carry the lowest risk and liability. Syncing the fork to rerun the checks.

xumia (approved):

sachinv-msft:
@wangxin could you please take a look at this?

PR #12177, opened 2022-09-23
Replace os.system and remove subprocess with shell=True

maipbui:

#### Why I did it
#### How I did it
Replace `os` by `subprocess`

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205

### DIFF
--- a/dockers/docker-lldp/lldpmgrd
+++ b/dockers/docker-lldp/lldpmgrd
@@ -80,10 +80,10 @@ class LldpManager(daemon_base.DaemonBase):
         self.port_init_done = False
 
     def update_hostname(self, hostname):
-        cmd = "lldpcli configure system hostname {0}".format(hostname)
+        cmd = ["lldpcli", "configure", "system", "hostname", hostname]
         self.log_debug("Running command: '{}'".format(cmd))
 
-        proc = subprocess.Popen(cmd, shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        proc = subprocess.Popen(cmd,stdout=subprocess.PIPE, stderr=subprocess.PIPE)
         (stdout, stderr) = proc.communicate()
 
         if proc.returncode != 0:
@@ -93,15 +93,15 @@ class LldpManager(daemon_base.DaemonBase):
 
     def update_mgmt_addr(self, ip):
         if ip == "None":
-            cmd = "lldpcli unconfigure system ip management pattern"
+            cmd = ["lldpcli", "unconfigure", "system", "ip", "management", "pattern"]
             self.log_info("Mgmt IP {0} deleted".format(self.mgmt_ip))
         else:
-            cmd = "lldpcli configure system ip management pattern {0}".format(ip)
+            cmd = ["lldpcli", "configure", "system", "ip", "management", "pattern", ip]
             self.log_info("Mgmt IP changed old ip {0}, new ip {1}".format(self.mgmt_ip, ip))
 
         self.log_debug("Running command: '{}'".format(cmd))
 
-        proc = subprocess.Popen(cmd, shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        proc = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
         (stdout, stderr) = proc.communicate()
 
         if proc.returncode != 0:
@@ -150,11 +150,11 @@ class LldpManager(daemon_base.DaemonBase):
         # Get the port description. If None or empty string, we'll skip this configuration
         port_desc = port_table_dict.get("description")
         
-        lldpcli_cmd = "lldpcli configure ports {0} lldp portidsubtype local {1}".format(port_name, port_alias)
+        lldpcli_cmd = ["lldpcli", "configure", "ports", port_name, "lldp", "portidsubtype", "local", port_alias]
 
         # if there is a description available, also configure that
         if port_desc:
-            lldpcli_cmd += " description '{}'".format(port_desc)
+            lldpcli_cmd += ["description", port_desc]
         else:
             self.log_info("Unable to retrieve description for port '{}'. Not adding port description".format(port_name))
 
@@ -330,7 +330,7 @@ class LldpManager(daemon_base.DaemonBase):
                     self.port_init_done = self.port_config_done = True
             if self.port_init_done and self.port_config_done:
                 self.port_init_done = self.port_config_done = False
-                rc, stderr = run_cmd(self, "lldpcli resume")
+                rc, stderr = run_cmd(self, ["lldpcli", "resume"])
                 if rc != 0:
                     self.log_error("Failed to resume lldpd with command: 'lldpcli resume': {}".format(stderr))
                     sys.exit(1)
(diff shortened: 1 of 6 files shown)


dockers/docker-lldp/lldpmgrd
```diff
@@ -80,10 +80,10 @@ class LldpManager(daemon_base.DaemonBase):
         self.port_init_done = False
 
     def update_hostname(self, hostname):
-        cmd = "lldpcli configure system hostname {0}".format(hostname)
+        cmd = ["lldpcli", "configure", "system", "hostname" hostname]
```
qiluo-msft, line 83: >  [](http://example.com/codeflow?start=59&length=1)

Do you miss a separator?

dockers/docker-lldp/lldpmgrd
```diff
@@ -150,11 +150,11 @@ class LldpManager(daemon_base.DaemonBase):
         # Get the port description. If None or empty string, we'll skip this configuration
         port_desc = port_table_dict.get("description")
         
-        lldpcli_cmd = "lldpcli configure ports {0} lldp portidsubtype local {1}".format(port_name, port_alias)
+        lldpcli_cmd = ["lldpcli", "configure", "ports", port_name, "lldp", "portidsubtype", "local", port_alias]
 
         # if there is a description available, also configure that
         if port_desc:
-            lldpcli_cmd += " description '{}'".format(port_desc)
+            lldpcli_cmd = ["lldpcli", "configure", "ports", port_name, "lldp", "portidsubtype", "local", port_alias, "description", port_desc]
```
qiluo-msft, line 157: >= [](http://example.com/codeflow?start=24&length=1)

You can still append to an array?

src/sonic-ctrmgrd/tests/ctrmgr_iptables_test.py
```diff
@@ -103,30 +103,29 @@ def __init__(self, ret, stdout, stderr):
...
-    cmd_prefix = "sudo iptables -t nat "
-    list_cmd = "{}-n -L OUTPUT ".format(cmd_prefix)
-    del_cmd = "{}-D OUTPUT ".format(cmd_prefix)
-    ins_cmd = "{}-A OUTPUT -p tcp -d ".format(cmd_prefix)
+    list_cmd = ["sudo", "iptables", "-t", "nat", "-n", "-L", "OUTPUT"]
+    del_cmd = ["sudo", "iptables", "-t", "nat", "-D", "OUTPUT"]
+    ins_cmd = ["sudo", "iptables", "-t", "nat", "-A", "OUTPUT", "-p", "tcp", "-d"]
 
-    assert shell
+    assert shell == False
```
qiluo-msft, line 110: >shell == False [](http://example.com/codeflow?start=11&length=14)

Why change it?

src/sonic-ctrmgrd/tests/ctrmgr_iptables_test.py
```diff
@@ -103,30 +103,29 @@ def __init__(self, ret, stdout, stderr):
...
     
     print("cmd={}".format(cmd))
-    if cmd.startswith(list_cmd):
-        num = int(cmd[len(list_cmd):])
+    if list_cmd == cmd[:len(list_cmd)]:
+        num = int(cmd[-1])
```
qiluo-msft, line 114: >-1 [](http://example.com/codeflow?start=22&length=2)

The logic is different from original logic, even I consider string vs. array of strings.

src/sonic-ctrmgrd/tests/ctrmgr_iptables_test.py
```diff
@@ -103,30 +103,29 @@ def __init__(self, ret, stdout, stderr):
...
         out = current_rules[num] if len(current_rules) > num else ""
         return proc(0, out, "")
 
-    if cmd.startswith(del_cmd):
-        num = int(cmd[len(del_cmd):])
+    if del_cmd == cmd[:len(del_cmd)]:
+        num = int(cmd[-1])
```
qiluo-msft, line 119: >cmd [](http://example.com/codeflow?start=18&length=3)

the same.

src/sonic-eventd/tools/events_volume_test.py
```diff
@@ -33,16 +33,13 @@ def run_test(process, file, count, duplicate):
     time.sleep(2) # buffer for events_tool to startup 
     logging.info("Generating logger messages\n")
     for i in range(count):
-        line = ""
-        state = "up"
         if duplicate:
-            line = "{} test message testmessage state up".format(process)
+            command = ["logger", "-p", "local0.notice", "-t", process, "test", "message", "testmessage", "state", "up"]
```
qiluo-msft, line 37: >logger", "-p", "local0.notice", "-t" [](http://example.com/codeflow?start=24&length=36)

this sub array repeated several times, better to extract as a variable.

src/sonic-config-engine/tests/test_cfggen.py
```diff
@@ -52,13 +52,12 @@ def tearDown(self):
             pass
 
     def run_script(self, argument, check_stderr=False, verbose=False):
-        print('\n    Running sonic-cfggen ' + argument)
+        print('\n    Running sonic-cfggen ', argument)
         self.assertTrue(self.yang.validate(argument))
-
         if check_stderr:
-            output = subprocess.check_output(self.script_file + ' ' + argument, stderr=subprocess.STDOUT, shell=True)
+            output = subprocess.check_output(self.script_file + argument, stderr=subprocess.STDOUT)
```
qiluo-msft, line 58: >self.script_file + argument [](http://example.com/codeflow?start=45&length=27)

Will you change it from a string to an array of strings?

src/sonic-ctrmgrd/tests/ctrmgr_iptables_test.py
maipbui: Because subprocess.run should be used with shell=False instead of True in original code. https://github.com/sonic-net/sonic-buildimage/blob/8d7f66ab175aaca05395314a26b2b1dab85eb576/src/sonic-ctrmgrd/ctrmgr/ctrmgr_iptables.py#L109

src/sonic-ctrmgrd/tests/ctrmgr_iptables_test.py
maipbui: Could you explain how so? 
I think the purpose is to get the `num` variable from this line, and UT tests in this repo pass.
https://github.com/sonic-net/sonic-buildimage/blob/8d7f66ab175aaca05395314a26b2b1dab85eb576/src/sonic-ctrmgrd/ctrmgr/ctrmgr_iptables.py#L108

maipbui: Maybe change to `int(cmd[len(list_cmd):][0])`?

qiluo-msft: Suggest `int(cmd[len(list_cmd)])`, however need check index first.

qiluo-msft (approved):

maipbui:
Splitting this PR to small PRs for easier control

qiluo-msft (approved):

PR #12227, opened 2022-09-29
Revert "Support for serdes platform library debian installation for Innovium SONiC image"

prgeor:
Reverts sonic-net/sonic-buildimage#11920. serdes platform should be open sourced and cannot be a deb package.


antony-rheneus:
@prgeor, For serdes SI settings each vendor has different configuration. And cannot be followed as similar to Brcm https://github.com/sonic-net/SONiC/blob/master/doc/media-settings/Media-based-Port-settings.md

Also the source code cannot be open sourced due to licensing agreement. 
Do you have any suggestions or any other approach which other vendors follow for it?

PR #12287, opened 2022-10-05
[Yang model] add Restapi yang file

prsunny:
#### Why I did it
Add yang model for Restapi

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205

### DIFF
--- a/src/sonic-yang-models/tests/yang_model_tests/tests/restapi.json
+++ b/src/sonic-yang-models/tests/yang_model_tests/tests/restapi.json
@@ -1,9 +1,9 @@
 {
     "RESTAPI_TABLE_WITH_INCORRECT_CERT": {
-        "desc": "TABLE_WITH_INCORRECT_CERT failure.",
-        "eStrKey": "Pattern"
+        "desc": "RESTAPI TABLE_WITH_INCORRECT_CERT failure.",
+        "eStr": ["server_crt"]
     },
     "RESTAPI_TABLE_WITH_VALID_CONFIG": {
-        "desc": "TABLE WITH VALID CONFIG."
+        "desc": "RESTAPI TABLE WITH VALID CONFIG."
     }
 }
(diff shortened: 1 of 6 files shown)


src/sonic-yang-models/yang-models/sonic-restapi.yang
```diff
@@ -0,0 +1,63 @@
...
+                    type string {
+                        pattern '(/[a-zA-Z0-9_-]+)*/([a-zA-Z0-9_-]+).pem';
+                    }
+                    description "Local path for ca_crt.";
+                }
+
+                leaf server_crt {
+                    type string {
+                        pattern '(/[a-zA-Z0-9_-]+)*/([a-zA-Z0-9_-]+).crt';
+                    }
```
ganglyu, line 42: Do we really need to limit the suffix as crt? Can we use xxx.cer?

prsunny: i think for restapi, it is currently fixed to crt. We may not need to provide an extension support

src/sonic-yang-models/tests/yang_model_tests/tests/restapi.json
```diff
@@ -0,0 +1,9 @@
+{
+    "RESTAPI_TABLE_WITH_INCORRECT_CERT": {
+        "desc": "TABLE_WITH_INCORRECT_CERT failure.",
+        "eStrKey": "Pattern"
```
praveen-li, line 4: "eStr": ["server_crt"]

This way, Libyang error will be matched against only server_crt field.

src/sonic-yang-models/tests/yang_model_tests/tests/restapi.json
```diff
@@ -0,0 +1,9 @@
...
+    },
+    "RESTAPI_TABLE_WITH_VALID_CONFIG": {
+        "desc": "TABLE WITH VALID CONFIG."
```
praveen-li, line 7: "RESTAPI TABLE WITH VALID CONFIG."

Desc is helpful in case of test failures. Thx

praveen-li (approved):
LGTM, barring minor issues in Tests. Plz fix them.

ganglyu (approved):

PR #12327, opened 2022-10-10
[openssh] Update openssh make file, add missing dependency to libnl.

liuh-80:
Update openssh make file, add missing dependency to libnl.

#### Why I did it
Openssh indirectly depends on libnl.
Another PR need add new patch to openssh, after adding new patch to openssh, PR build failed with libnl missing error.

#### How I did it
Update openssh make file, add missing dependency to libnl.

#### How to verify it
Pass all test case

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205

Update openssh make file, add missing dependency to libnl.

### DIFF
--- a/rules/openssh.mk
+++ b/rules/openssh.mk
@@ -6,6 +6,7 @@ export OPENSSH_VERSION
 
 OPENSSH_SERVER = openssh-server_$(OPENSSH_VERSION)_$(CONFIGURED_ARCH).deb
 $(OPENSSH_SERVER)_SRC_PATH = $(SRC_PATH)/openssh
+$(OPENSSH_SERVER)_DEPENDS += $(LIBNL_ROUTE3_DEV)
 SONIC_MAKE_DEBS += $(OPENSSH_SERVER)
 
 # The .c, .cpp, .h & .hpp files under src/{$DBG_SRC_ARCHIVE list}


liuh-80:
Close, now sonic using fips version.

liuh-80:
Build failed with this error, seems a agent pool issue:
E: Could not get lock /var/lib/dpkg/lock-frontend. It is held by process 58398 (dpkg)
E: Unable to acquire the dpkg frontend lock (/var/lib/dpkg/lock-frontend), is another process using it?
make[1]: *** [Makefile:9: /sonic/target/debs/bullseye/openssh-server_8.4p1-5+deb11u1_amd64.deb] Error 100

rules/openssh.mk
```diff
@@ -6,6 +6,7 @@ export OPENSSH_VERSION
 
 OPENSSH_SERVER = openssh-server_$(OPENSSH_VERSION)_$(CONFIGURED_ARCH).deb
 $(OPENSSH_SERVER)_SRC_PATH = $(SRC_PATH)/openssh
+$(OPENSSH_SERVER)_DEPENDS +=  $(LIBNL3_DEV) $(LIBNL_ROUTE3_DEV)
```
qiluo-msft, line 9: >LIBNL_ROUTE3_DEV [](http://example.com/codeflow?start=46&length=16)

Seem this fix is a valid fix, and has no relationship with the other PR 12447 .

I am worried about why this bug is not exposed before? Could you check previous build log and possible find a build bug?

qiluo-msft: @xumia please be aware of this issue.

liuh-80: > > LIBNL_ROUTE3_DEV
> 
> Seem this fix is a valid fix, and has no relationship with the other PR 12447 .
> 
> I am worried about why this bug is not exposed before? Could you check previous build log and possible find a build bug?

I checked the build log and not found build issue caused by this.
When build on local devbox or run a full build, in most case we build everything from source code, so the libnl will be build and install before openssh build.

Here is how this issue happen:
    When build on azure pipeline, we are using cache, so when change openssh code, the libnl will not build because there is no code change, also libnl will not be installed from cache. then openssh will build break because missing package.

xumia (approved):

liuh-80:
Request this for 202205 branch, later there will be another PR depends on this also will request for 202205 branch:

https://github.com/sonic-net/sonic-buildimage/pull/12447

PR #12331, opened 2022-10-10
Add k8s master table yang model

lixiaoyuner:
Signed-off-by: Yun Li <yunli1@microsoft.com>

#### Why I did it
Add k8s master table yang model

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205


src/sonic-yang-models/yang-models/sonic-kubernetes_master.yang
```diff
@@ -0,0 +1,60 @@
...
+
+        container KUBERNETES_MASTER {
+
+            description "KUBERNETES_MASTER part of config_db.json";
+
+            container SERVER {
+
+                leaf ip {
+                    description "Kubernetes api server endpoint";
+                    type string;
```
renukamanavalan, line 32: This should use IPAddress type

src/sonic-yang-models/yang-models/sonic-kubernetes_master.yang
```diff
@@ -0,0 +1,60 @@
...
+                    description "This configuration identicates whether disable kubernetes";
+                    type stypes:boolean_type;
+                    default "false";
+                }
+
+                leaf insecure {
+                    description "This configuration identicates it will download kubernetes
+                                CA by http other than https";
+                    type stypes:boolean_type;
+                    default "true";
```
renukamanavalan, line 50: This should be false too.

renukamanavalan:
@praveen-li, Please review this PR.

ganglyu:
Would you please update doc/Configuration.md?

ganglyu (approved):

src/sonic-yang-models/yang-models/sonic-kubernetes_master.yang
lixiaoyuner: It should be IPAddress type and DNS domain type, currently use inet:host to cover both of them.

src/sonic-yang-models/yang-models/sonic-kubernetes_master.yang
lixiaoyuner: Currently sonic join logic is to download k8a default ca.crt and combine it with local ame certificate, we get the kubeconfig credential. The default ca.crt is stored in k8s configmap, we download it from apiserver endpoint, and sonic impossibly have the default ca.crt, but if sonic want to establish secure connection with apiserver, sonic needs k8s default ca.crt. So, the downloading k8s default ca request can only be insecure currently. So, I set the default value "true".

lixiaoyuner:
> Would you please update doc/Configuration.md?

Updated

renukamanavalan (approved):

PR #12369, opened 2022-10-12
Fixes the exit status of slave build.

Kalimuthu-Velappan:
This PR fixes the issue reported in PR#12367
https://github.com/sonic-net/sonic-buildimage/pull/12367

The issue is that exit code always being 0 for the builds that are failed.

Fix is added in the Makefile.work to return the error code when the slave build is failed with an error.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205

### DIFF
--- a/Makefile.work
+++ b/Makefile.work
@@ -500,7 +500,7 @@ SONIC_BUILD_INSTRUCTION :=  $(MAKE) \
 
 
 ifeq ($(filter clean,$(MAKECMDGOALS)),)
-COLLECT_BUILD_VERSION = { DBGOPT='$(DBGOPT)' scripts/collect_build_version_files.sh $$?; }
+COLLECT_BUILD_VERSION = { DBGOPT='$(DBGOPT)' scripts/collect_build_version_files.sh \$$?; }
 endif
 
 ifdef SOURCE_FOLDER
@@ -508,10 +508,11 @@ ifdef SOURCE_FOLDER
 endif
 
 ifeq "$(KEEP_SLAVE_ON)" "yes"
-SLAVE_SHELL={ /bin/bash; }
+SLAVE_SHELL={ ret=\$$?; /bin/bash; exit \$$ret; }
 endif
 
 .DEFAULT_GOAL :=  all
+.SHELLFLAGS += -e
 
 %:: | sonic-build-hooks
 ifneq ($(filter y, $(MULTIARCH_QEMU_ENVIRON) $(CROSS_BUILD_ENVIRON)),)
@@ -527,7 +528,7 @@ endif
 
 	$(Q)$(DOCKER_RUN) \
 		$(SLAVE_IMAGE):$(SLAVE_TAG) \
-		bash -c "$(SONIC_BUILD_INSTRUCTION) $@;$(COLLECT_BUILD_VERSION); $(SLAVE_SHELL)"
+		bash -c "$(SONIC_BUILD_INSTRUCTION) $@; $(COLLECT_BUILD_VERSION); $(SLAVE_SHELL)" 
 	$(Q)$(docker-image-cleanup)
 
 docker-cleanup:


Makefile.work
```diff
@@ -527,7 +530,8 @@ endif
 
 	$(Q)$(DOCKER_RUN) \
 		$(SLAVE_IMAGE):$(SLAVE_TAG) \
-		bash -c "$(SONIC_BUILD_INSTRUCTION) $@;$(COLLECT_BUILD_VERSION); $(SLAVE_SHELL)"
+		bash -c "$(SONIC_BUILD_INSTRUCTION) $@ && $(COLLECT_BUILD_VERSION) && $(SLAVE_SHELL)" \
```
saiarcot895, line 533: This could be problematic. Currently, `$(COLLECT_BUILD_VERSION)` is meant to be run always (whether the actual build passed or not). That script then exits with the exit code that's passed in.

Makefile.work
```diff
@@ -527,7 +530,8 @@ endif
...
+		|| $(ERROR_EXIT)
```
saiarcot895, line 534: I just realized, running the slave shell here (in `$(ERROR_EXIT)`) doesn't make sense, because the `docker run` command has already finished, and this will just launch the shell outside of the docker environment.

Makefile.work
Kalimuthu-Velappan: done

Makefile.work
Kalimuthu-Velappan: done

saiarcot895 (approved):

Makefile.work
```diff
@@ -527,7 +530,8 @@ endif
 
 	$(Q)$(DOCKER_RUN) \
 		$(SLAVE_IMAGE):$(SLAVE_TAG) \
-		bash -c "$(SONIC_BUILD_INSTRUCTION) $@;$(COLLECT_BUILD_VERSION); $(SLAVE_SHELL)"
+		bash -c "$(SONIC_BUILD_INSTRUCTION) $@; $(COLLECT_BUILD_VERSION) && $(SLAVE_SHELL) || $(ERROR_EXIT)" \
```
xumia, line 533: How about using the SHELLFLAGS?
```
.SHELLFLAGS += -e
```

Kalimuthu-Velappan: done

xumia (approved):

Makefile.work
```diff
@@ -509,9 +509,13 @@ endif
 
 ifeq "$(KEEP_SLAVE_ON)" "yes"
 SLAVE_SHELL={ /bin/bash; }
+else
+SLAVE_SHELL={ true; }
 endif
+ERROR_EXIT={ $(SLAVE_SHELL); exit -1; }
```
xumia, line 515: Looks like it can be simplified as below:
```
ifeq "$(KEEP_SLAVE_ON)" "yes"
SLAVE_SHELL={ ret=\$$?; /bin/bash; exit $ret; }
endif
```

Makefile.work
```diff
@@ -527,7 +531,8 @@ endif
 
 	$(Q)$(DOCKER_RUN) \
 		$(SLAVE_IMAGE):$(SLAVE_TAG) \
-		bash -c "$(SONIC_BUILD_INSTRUCTION) $@;$(COLLECT_BUILD_VERSION); $(SLAVE_SHELL)"
+		bash -c "$(SONIC_BUILD_INSTRUCTION) $@; $(COLLECT_BUILD_VERSION) && $(SLAVE_SHELL) || $(ERROR_EXIT)" \
+		|| exit -1
```
xumia, line 535: "exit -1" can be removed when ".SHELLFLAGS += -e" specified.

Makefile.work
Kalimuthu-Velappan: Done, thanks. it is much simpler now

xumia:
Merging the PR by force, reason: all the build returning succeeded state, it impacts all PR checks.

PR #12379, opened 2022-10-12
add Restapi Yang model

prsunny:
#### Why I did it

Reverting PR https://github.com/sonic-net/sonic-buildimage/pull/12374

Fix build issues

This PR fixed https://github.com/sonic-net/sonic-buildimage/issues/10548

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205

### DIFF
--- a/src/sonic-yang-models/yang-models/sonic-restapi.yang
+++ b/src/sonic-yang-models/yang-models/sonic-restapi.yang
@@ -31,7 +31,7 @@ module sonic-restapi {
 
                 leaf ca_crt {
                     type string {
-                        pattern '(/[a-zA-Z0-9_-]+)*/([a-zA-Z0-9_-]+).pem';
+                        pattern '(/[a-zA-Z0-9_-]+)*/([a-zA-Z0-9_-]+).([a-z]+)';
                     }
                     description "Local path for ca_crt.";
                 }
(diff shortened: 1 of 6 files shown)


src/sonic-yang-models/yang-models/sonic-restapi.yang
```diff
@@ -0,0 +1,63 @@
...
+                leaf server_crt {
+                    type string {
+                        pattern '(/[a-zA-Z0-9_-]+)*/([a-zA-Z0-9_-]+).crt';
+                    }
+                    description "Local path for server_crt.";
+                }
+
+                leaf client_crt_cname {
+                    type string {
+                        pattern '(/[a-zA-Z0-9_-.]+)*/([a-zA-Z0-9_-.]+)./([a-z]{3})';
```
qiluo-msft, line 48: >([a-z]{3}) [](http://example.com/codeflow?start=72&length=10)

What is the issue in original PR? Do you add a new test case for the issue?

ganglyu (approved):

src/sonic-yang-models/yang-models/sonic-restapi.yang
prsunny: it was a syntax error in regex. i've added few more tests to cover.

qiluo-msft (approved):

PR #12381, opened 2022-10-12
[BGP] starting BGP service after swss

yxieca:
#### Why I did it
BGP service has always been starting after interface-config. However, recently we discovered an issue where some BGP sessions are unable to establish due to BGP daemon not able to read the interface IP.

This issue was clearly observed after upgrading to FRR 8.2.2. See more details in #12380.

#### How I did it
Delaying starting BGP seems to be a workaround for this issue.

However, caution is that this delay might impact warm reboot timing and other timing sequences.

This workaround is reducing the probability of hitting the issue by close to 100X. However, this workaround is not bulletproof as test shows. It is still preferrable to have a proper FRR fix and revert this change in the future.

#### How to verify it
Continuously issuing config reload and check BGP session status afterwards.

### DIFF
--- a/files/build_templates/per_namespace/bgp.service.j2
+++ b/files/build_templates/per_namespace/bgp.service.j2
@@ -7,6 +7,7 @@ After=updategraph.service
 BindsTo=sonic.target
 After=sonic.target
 Before=ntp-config.service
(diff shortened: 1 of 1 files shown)


yxieca:
Cherry-pick was done by PR https://github.com/sonic-net/sonic-buildimage/pull/15312

PR #12426, opened 2022-10-18
[VXLAN] Reply arp for loopback ip address.

iMasaruOki:
Signed-off-by: Masaru OKI <masaru.oki@gmail.com>

#### Why I did it
VXLAN VTEP IP address is assigned into loopback interface.
So far, arp_ignore = 2 - reply only if the target IP address is 
local address configured on the incoming interface.
VXLAN tunneling doesn't work with this configuration.

#### How I did it
arp_ignore changes to 0 - reply for any local target IP address,
 configured on any interface.

#### How to verify it
1. assign loopback ip address
2. add routing information (e.g. bgp)
3. ping peer's loopback ip address

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205


files/image_config/sysctl/sysctl-net.conf
```diff
@@ -14,7 +14,7 @@ net.ipv4.conf.all.arp_accept=0
 net.ipv4.conf.all.arp_announce=1
 net.ipv4.conf.all.arp_filter=0
 net.ipv4.conf.all.arp_notify=1
-net.ipv4.conf.all.arp_ignore=2
+net.ipv4.conf.all.arp_ignore=0
```
theasianpianist, line 17: I'm trying to investigate and find out why this was originally set to 2, will come back once I have an answer

theasianpianist: This would make the Loopback IP reachable from any connected device, including servers/blades configured in a Vlan. I@prsunny @lguohan is this acceptable? I can't think of any other impact this might have. Since this change is making `arp_ignore` more permissive, existing ARP functionality should hopefully be unaffected.

theasianpianist:
@iMasaruOki If there is a BGP session established between the two devices you are using, the Loopback IP should already be pingable between them without changing this ARP setting since the route to the Loopback IP should be advertised via BGP. If that's not working then I think there may be other issues that need to be addressed.

dulitz:
I think there's another use case for the underlying change. If you're using VLANs for port isolation (since SONiC doesn't explicitly support port isolation), you may want to assign a /32 network to the VLAN. But you still want the switch to answer `who-has` for its own addresses, which are outside that /32.

Yes, it is odd for the next-hop to be outside the network, but there are hosts happy to do this and if you need to conserve IP addresses it's one of the tools available -- or it is if your switch answers `who-has` for its own addresses outside the /32.

PR #12489, opened 2022-10-25
[containercfgd] Add containercfgd and syslog rate limit configuration support

Junchao-Mellanox:
HLD: https://github.com/sonic-net/SONiC/pull/1049

#### Why I did it

Add a containercfgd to handler container side CONFIG DB changes

#### How I did it

1. add a daemon containercfgd to run for each containers
2. add syslog handler to handle syslog rate limit config change

#### How to verify it

unit test

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205


Junchao-Mellanox:
Build issue pending on PR https://github.com/sonic-net/sonic-buildimage/pull/12510

files/build_templates/docker_image_ctl.j2
```diff
@@ -34,16 +34,24 @@ function updateSyslogConf()
...
+        if [ "$CONTAINER_EXISTS" = "yes" ]; then
+            # database configuration has been synced to /etc/rsyslog.conf
+            # no need generate it to save boot time
+            return
+        fi
+        TARGET_IP="127.0.0.1"
     fi
+    CONTAINER_NAME="$DOCKERNAME"
+    TMP_FILE="/tmp/rsyslog.$CONTAINER_NAME.conf"
+    rm -rf $TMP_FILE
```
wen587, line 47: Is the remove necessary?

files/image_config/rsyslog/rsyslog-container.conf.j2
```diff
@@ -14,8 +14,24 @@ $ModLoad imuxsock # provides support for local system logging
...
+{% endif %}
+{% endif %}
+{% endif %}
+
+{% if rate_limit_interval is defined %}
+$SystemLogRateLimitInterval {{ rate_limit_interval }}
+{% endif %}
+{% if rate_limit_burst is defined %}
+$SystemLogRateLimitBurst {{ rate_limit_burst }}
+{% endif %}
```
mkeda, line 34: This change means you have unlimited messages unless you turn on the rate-limiting feature?  Is it possible to have rate limiting but not make it configurable?

src/sonic-containercfgd/tests/conftest.py
```diff
@@ -0,0 +1,5 @@
+import os
+
+
+def init_env():
+    os.environ["CONTAINER_NAME"] = "2"
```
mkeda, line 5: Is there any significance to 2?

src/sonic-containercfgd/tests/test_config_daemon.py
```diff
@@ -0,0 +1,54 @@
...
+def test_init_data_handler():
+    mock_handler_cls = mock.MagicMock()
+    mock_handler_instance = mock.MagicMock()
+    mock_handler_instance.handle_init_data = mock.MagicMock()
+    mock_handler_cls.return_value = mock_handler_instance
+
+    containercfgd.ContainerConfigDaemon.register_handler('LoadMock', mock_handler_cls)
+    daemon = containercfgd.ContainerConfigDaemon()
+    daemon.init_data_handler({})
+    mock_handler_instance.handle_init_data.assert_called_once
```
mkeda, line 35: missing () ?

src/sonic-containercfgd/tests/test_syslog_config.py
```diff
@@ -0,0 +1,82 @@
...
+    handler.handle_init_data(init_data)
+    handler.update_syslog_config.assert_called_once()
+
+
+@mock.patch('containercfgd.containercfgd.run_command')
+@mock.patch('containercfgd.containercfgd.SyslogHandler.parse_syslog_conf', mock.MagicMock(return_value=('100', '200', '127.0.0.1')))
+def test_update_syslog_config(mock_run_cmd):
+    mock_run_cmd.return_value = ""
+    handler = containercfgd.SyslogHandler()
+    #handler.parse_syslog_conf = mock.MagicMock(return_value=('100', '200', '127.0.0.1'))
```
mkeda, line 56: still needed?

src/sonic-containercfgd/containercfgd/containercfgd.py
```diff
@@ -0,0 +1,198 @@
...
+container_name = None
+
+
+def run_command(command):
+    """
+    Utility function to run an shell command and return the output.
+    :param command: Shell command string.
+    :return: Output of the shell command.
+    """
+    return subprocess.check_output(shlex.split(command), text=True, stderr=subprocess.PIPE)
```
mkeda, line 32: No need to catch exceptions here?

src/sonic-containercfgd/containercfgd/containercfgd.py
```diff
@@ -0,0 +1,198 @@
...
+        table_name (str): CONFIG DB table name
+    """
+    def wrapper(object_type):
+        ContainerConfigDaemon.register_handler(table_name, object_type)
+        return object_type
+    return wrapper
+
+
+@config_handler(SYSLOG_CONFIG_FEATURE_TABLE)
+class SyslogHandler:
```
mkeda, line 96: This seems very similar to the code in the sonic-host-services pull request.  Is there any way to share that?

src/sonic-containercfgd/containercfgd/containercfgd.py
```diff
@@ -0,0 +1,198 @@
...
+
+        logger.log_notice(f'Configure syslog rate limit interval={new_interval}, burst={new_burst}')
+
+        if os.path.exists(self.TMP_SYSLOG_CONF_PATH):
+            os.remove(self.TMP_SYSLOG_CONF_PATH)
+        with open(self.TMP_SYSLOG_CONF_PATH, 'w+') as f:
+            output = run_command('sonic-cfggen -d -t /usr/share/sonic/templates/rsyslog-container.conf.j2 -a "{{\\"target_ip\\": \\"{}\\", \\"container_name\\": \\"{}\\" }}"'.format(self.target_ip, container_name))
+            f.write(output)
+        run_command('cp {} {}'.format(self.TMP_SYSLOG_CONF_PATH, self.SYSLOG_CONF_PATH))
+        run_command("supervisorctl restart rsyslogd")
```
mkeda, line 156: In the sonic-host-services PR, you also had 'systemctl reset-failed rsyslog-config rsyslog'.  Is that needed here?

mkeda (changes requested):
Thank you for your work to add this feature!

src/sonic-containercfgd/containercfgd/containercfgd.py
Junchao-Mellanox: No, according to supervisor doc, supervisorctl has no such function.

files/build_templates/docker_image_ctl.j2
Junchao-Mellanox: will remove this line

files/image_config/rsyslog/rsyslog-container.conf.j2
Junchao-Mellanox: `This change means you have unlimited messages unless you turn on the rate-limiting feature?`
Yes.
`Is it possible to have rate limiting but not make it configurable?`
Default rate-limit configuration is located at files/build_templates/init_cfg.json.j2, check this PR: https://github.com/sonic-net/sonic-buildimage/pull/12490

src/sonic-containercfgd/tests/conftest.py
Junchao-Mellanox: No, here is just for making sure there is an env variable named "CONTAINER_NAME". I will change it to something different and add comment.

src/sonic-containercfgd/containercfgd/containercfgd.py
Junchao-Mellanox: Currently, I would like to let caller to handler the exception, but I am not sure if it is best practice. Any suggestion?

src/sonic-containercfgd/containercfgd/containercfgd.py
Junchao-Mellanox: This is good suggestion. I also thought of sharing the logic between sonic-host-services and containercfgd, however, I finnaly decided to leave it as is because:

1. cannot find a good place to do so. The logic for SyslogHandler is not so general to put it into sonic_py_common. 
2. The logic in sonic-host-services and containercfgd are similar but not exactly the same.

Junchao-Mellanox:
Hi @mkeda, I have fixed/replied all comments. Would you please review the new changes?

src/sonic-containercfgd/containercfgd/containercfgd.py
mkeda: I think it's ok to let the caller to handle the exception in this case.

mkeda (approved):

keboliu (approved):

Junchao-Mellanox:
Hi @saiarcot895, could you please kindly help review this?

src/sonic-containercfgd/containercfgd/containercfgd.py
saiarcot895, line 156: Can these be split into individual entries (such as `["cp", self.TMP_SYSLOG_CONF_PATH, self.SYSLOG_CONF_PATH]` and passed directly into `subprocess.check_output` instead of using `shlex.split`? That will prevent code injections from outside sources doing something that shouldn't be done.

saiarcot895: That's true, because it's split into an array, the first token will always be the command that gets executed. However, if, hypothetically, `self.TMP_SYSLOG_CONF_PATH` was `/tmp/file name`, then `shlex.split` will split `cp /tmp/file name /tmp/file_copy` into `["cp", "/tmp/file", "name", "/tmp/file_copy"]`, which is incorrect.

I think considering we know what these arguments should be, just doing the splitting ourselves would be better. (It would also make the JSON argument a couple lines above this a bit easier to read).

Junchao-Mellanox: fixed

saiarcot895 (approved):

PR #12497, opened 2022-10-25
Add yang_config_validation to DEVICE_METADATA yang model

isabelmsft:
#### Why I did it
New field `yang_config_validation` needs to be represented in DEVICE_METADATA yang model, configured in minigraph.py

#### How I did it
Add field to yang model, set value in minigraph.py

#### How to verify it
Use GCU to apply patch configuring `yang_config_validation` field

### DIFF
--- a/src/sonic-config-engine/minigraph.py
+++ b/src/sonic-config-engine/minigraph.py
@@ -1450,7 +1450,8 @@ def parse_xml(filename, platform=None, port_config_file=None, asic_name=None, hw
         'hostname': hostname,
         'hwsku': hwsku,
         'type': device_type,
-        'synchronous_mode': 'enable'
+        'synchronous_mode': 'enable',
+        'yang_config_validation': 'disable'
         }
     }
 
--- a/src/sonic-config-engine/minigraph.py
+++ b/src/sonic-config-engine/minigraph.py
@@ -1450,8 +1450,7 @@ def parse_xml(filename, platform=None, port_config_file=None, asic_name=None, hw
         'hostname': hostname,
         'hwsku': hwsku,
         'type': device_type,
-        'synchronous_mode': 'enable',
-        'yang_config_validation': 'disable'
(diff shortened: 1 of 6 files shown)


qiluo-msft:
@ganglyu could you check?

src/sonic-config-engine/minigraph.py
```diff
@@ -1450,7 +1450,8 @@ def parse_xml(filename, platform=None, port_config_file=None, asic_name=None, hw
         'hostname': hostname,
         'hwsku': hwsku,
         'type': device_type,
-        'synchronous_mode': 'enable'
+        'synchronous_mode': 'enable',
+        'yang_config_validation': 'disable'
```
ganglyu, line 1454: Please create another PR for sonic-config-engine, and update unit test for sonic-config-engine.

src/sonic-yang-models/yang-models/sonic-device_metadata.yang
```diff
@@ -119,6 +119,14 @@ module sonic-device_metadata {
                     default enable;
                 }
 
+                leaf yang_config_validation {
+                    type enumeration {
```
ganglyu, line 123: Can you use mode-status instead?
https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-yang-models/yang-templates/sonic-types.yang.j2#L178

ganglyu:
Please update Configuration.md and sample_config_db.json.

src/sonic-yang-models/yang-models/sonic-device_metadata.yang
isabelmsft: Changed to use stypes:mode-status

ganglyu (approved):

src/sonic-config-engine/minigraph.py
isabelmsft: Created separate PR

PR #12500, opened 2022-10-26
[202012] [Mellanox] [SFP Platform API] Fix DOM support capability issues on QSFP and CMIS cables

keboliu:
#### Why I did it
1. For QSFP cables(SFF 8636), if the cable is only compliant with some old revision(e.g. lower than 0x08), in these revisions some of the DOM support status bits are not defined yet, so the DOM support status of the cable is also 'unknown', currently will treat this case as DOM support status 'supported', but for copper cables, it never can support DOM, copper cables should be ruled out in this case.
2. For CMIS cable, there is a bug in determining the DOM support status, the temperature and voltage support statuses are always set to true, this is not correct, they shall be determined in the same way as other DOM capability support statuses.

#### How I did it
1. Add a function to judge whether a QSFP cable is a copper cable or not, it judges from the specification compliance bytes and ext specification compliance byte.
2. CMIS cable temperature and voltage DOM support status shall be judged in the same way as other capabilities.
#### How to verify it
Verify the platform API and 'show interface transceiver eeprom -d' command on various types of CMIS and QSFP cables to make sure the output is as expected.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205

### DIFF
--- a/platform/mellanox/mlnx-platform-api/sonic_platform/sfp.py
+++ b/platform/mellanox/mlnx-platform-api/sonic_platform/sfp.py
@@ -411,6 +411,26 @@ class SFP(SfpBase):
             # in this case we treat it as the default type according to the SKU
             self.sfp_type = sfp_type
 
+    def _is_qsfp_copper(self):
+        # This function read the specification compliance byte and
+        # ext specification compliance byte to judge whether the cable is copper or not.
+        # It is only designed for SFF 8636 cables, SFF 8472 and CMIS cables don't apply.
+
+        # copper cables defined in SFF8024 "Extended Specification Compliance Codes"
+        copper_ext_specification_list = ['08', '0b', '0c', '0d', '19', '30', '32', '40']
+
+        offset = PAGE00_HIGH_MEMORY_OFFSET
+        qsfp_specification_raw = self._read_eeprom_specific_bytes(offset + XCVR_COMPLIANCE_CODE_OFFSET, 1)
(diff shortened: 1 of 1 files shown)


keboliu:
This PR has a dependency on https://github.com/sonic-net/sonic-buildimage/pull/12399

keboliu:
@prgeor would you please help to review?

platform/mellanox/mlnx-platform-api/sonic_platform/sfp.py
```diff
@@ -411,6 +411,26 @@ def _detect_sfp_type(self, sfp_type):
             # in this case we treat it as the default type according to the SKU
             self.sfp_type = sfp_type
 
+    def _is_qsfp_copper(self):
```
prgeor, line 414: PR title says CMIS, so where is CMIS copper cable check?

keboliu: The CMIS cable issue is not about copper or not, it is arbitrarily setting the temperature and voltage support statuses to true w/o reading the EEPROM, please check the PR description: "For CMIS cable, there is a bug in determining the DOM support status, the temperature and voltage support statuses are always set to true, this is not correct, they shall be determined in the same way as other DOM capability support statuses."

prgeor (approved):

PR #12517, opened 2022-10-26
Revert "[build] don't require passwordless sudo (Closes #11292)"

### LINKED ISSUE #11292: Builds now require password-less sudo access

qiluo-msft:
Reverts sonic-net/sonic-buildimage#11417

This docker image `alpine` is unnecessary dependency.

### DIFF
--- a/Makefile.work
+++ b/Makefile.work
@@ -241,8 +241,7 @@ endif
 DOCKER_LOCKFILE_SAVE := $(DOCKER_LOCKDIR)/docker_save.lock
 $(shell mkdir -m 0777 -p $(DOCKER_LOCKDIR))
 $(shell [ -f $(DOCKER_LOCKFILE_SAVE) ] || (touch $(DOCKER_LOCKFILE_SAVE) && chmod 0777 $(DOCKER_LOCKFILE_SAVE)))
-$(docker run --rm -v $(DOCKER_ROOT)\:/mount alpine sh -c 'rm -rf /mount/')
-$(mkdir -p $(DOCKER_ROOT))
+$(shell sudo rm -rf $(DOCKER_ROOT) && mkdir -p $(DOCKER_ROOT))
 
 ifeq ($(DOCKER_BUILDER_MOUNT),)
 override DOCKER_BUILDER_MOUNT := "$(PWD):/sonic"


qiluo-msft:
@jusherma Could you help review?

jusherma (changes requested):
This is reverting a change which fixed a build issue (#11292). This should not be reverted

kv-y:
> This is reverting a change which fixed a build issue (#11292). This should not be reverted

Current code also doesn't work. It does nothing.
1. Word `shell` is missed
2. You can't remove mounted directory inside a container. Try and you get `Resoure busy` error.
Possible solution is mount parent directory of $(DOCKER_ROOT).
3. Instead of `alpine` probably better to use `debian:$(BLDENV)` container.

Makefile.work
```diff
@@ -241,8 +241,7 @@ endif
 DOCKER_LOCKFILE_SAVE := $(DOCKER_LOCKDIR)/docker_save.lock
 $(shell mkdir -m 0777 -p $(DOCKER_LOCKDIR))
 $(shell [ -f $(DOCKER_LOCKFILE_SAVE) ] || (touch $(DOCKER_LOCKFILE_SAVE) && chmod 0777 $(DOCKER_LOCKFILE_SAVE)))
-$(docker run --rm -v $(DOCKER_ROOT)\:/mount alpine sh -c 'rm -rf /mount/')
-$(mkdir -p $(DOCKER_ROOT))
+$(shell sudo rm -rf $(DOCKER_ROOT) && mkdir -p $(DOCKER_ROOT))
```
kv-y, line 244:
```suggestion
+ifneq ($(filter $(SONIC_SLAVE_DOCKER_DRIVER),overlay overlay2),)
+$(shell docker run --rm -v $$(dirname "$(DOCKER_ROOT)")\:/mount debian:buster sh -c 'rm -rf /mount/$$(basename "$(DOCKER_ROOT)")')
+$(shell mkdir -p $(DOCKER_ROOT))
+endif
```

1. debian:buster because we already use it  for DOCKER_MACHINE in Makefile.work, but maybe need a variable for default release name.
2. dirname and basename because we can't remove bind mounted directory itself
3. We need to execute this command only if we set SONIC_SLAVE_DOCKER_DRIVER to overlay/overlay2
4. I'm not sure about mkdir.  If directory doesn't exist docker creates this directory anyway, but owner is root in this case.

jusherma: Thanks Konstantin. I not sure if `$(shell )` is needed. My understanding is that it tells `make` to treat the command's output as if it were part of the makefile (https://www.gnu.org/software/make/manual/html_node/Shell-Function.html) 

But if the above works, I'm not too particular about how it works. My only requirement is that `sudo` isn't used as that breaks the builds for us

kv-y: @xumia
Could you please review this.
You added this feature in https://github.com/sonic-net/sonic-buildimage/pull/11111
But side effect of your commit was requiring sudo for each make.

There was a fix for this in https://github.com/sonic-net/sonic-buildimage/pull/11417.
But fix doesn't work correctly because:
1. word `shell` was missed
2. It's not possible to remove bind-mounted directory itself
So at this moment fsroot directories are not removed when you do make.

Also this fix has some drawbacks:
1. Docker image `alpine` is unnecessary dependency. We can use debian:buster instead (we already use it for DOCKER_MACHINE detection)
2. We don't have to remove fsroot directory and run docker container for this every time (e.g. when we use a native docker)
3. We don't have to recreate this directory because docker does it anyway (if we use -v option).

I can do a pull request instead of suggestion. But in this case it will take some time because I haven't signed CCLA yet.

qiluo-msft:
Close in favor of https://github.com/sonic-net/sonic-buildimage/pull/15132

PR #12532, opened 2022-10-27
[Aboot] Support boot from device

WeirdMachine:
#### Why I did it
I have a Arista Switch which has a slot for an m2 ssd drive that is by default not used. I added the m2 SSD. 
However the boot script and some other tooling, assumes that everything is installed on the internal flash and not the drive (SSD). The boot-config file is never updated on the flash, which leads to a failing boot as aboot always reads the boot-config from flash.

#### How I did it

Ensure boot-config is mirrored to the flash when updated.

#### How to verify it
Install SONiC on Arista device with an additional m2 SSD, referenced as drive.

Add support for drive device for aboot


WeirdMachine:
Support for the installer: https://github.com/sonic-net/sonic-utilities/pull/2469

PR #12534, opened 2022-10-27
[sonic-ctrmgrd] Replace os.system and remove subprocess with shell=True

maipbui:

#### Why I did it
`subprocess.Popen()` and `subprocess.run()` is used with `shell=True`, which is very dangerous for shell injection.
#### How I did it
Replace `os` by `subprocess`, remove `shell=True`
#### How to verify it
Passed UT
Tested in DUT

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205

### DIFF
--- a/src/sonic-ctrmgrd/ctrmgr/container_startup.py
+++ b/src/sonic-ctrmgrd/ctrmgr/container_startup.py
@@ -120,7 +120,6 @@ def get_docker_id():
     cmd3 = ['cut', '-f1', '-d', '/']
     cmd4 = ['rev']
     _, output = getstatusoutput_noshell_pipe(cmd0, cmd1, cmd2, cmd3, cmd4)
-    # output = proc.communicate()[0].decode("utf-8")
     return output.strip()[:12]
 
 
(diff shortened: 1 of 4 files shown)


src/sonic-ctrmgrd/ctrmgr/container_startup.py
```diff
@@ -114,9 +114,13 @@ def get_docker_id():
...
-    cmd = 'cat /proc/self/cgroup | grep -e ":memory:" | rev | cut -f1 -d\'/\' | rev'
-    proc = subprocess.Popen(cmd, shell=True, stdout=subprocess.PIPE)
-    output = proc.communicate()[0].decode("utf-8")
+    cmd0 = ['cat', '/proc/self/cgroup']
+    cmd1 = ['grep', '-e', ":memory:"]
+    cmd2 = ['rev']
+    cmd3 = ['cut', '-f1', '-d', '/']
+    cmd4 = ['rev']
+    _, output = getstatusoutput_noshell_pipe(cmd0, cmd1, cmd2, cmd3, cmd4)
+    # output = proc.communicate()[0].decode("utf-8")
```
qiluo-msft, line 123: ># output = proc.communicate()[0].decode("utf-8") [](http://example.com/codeflow?start=4&length=48)

Do you need this line of code? if not, delete it.

renukamanavalan: @lixiaoyuner Can you please review ?

PR #12537, opened 2022-10-27
[sonic-yang-mgmt] Replace subprocess using with shell=True

maipbui:

#### Why I did it
#### How I did it
#### How to verify it
Pass UT

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205

### DIFF
--- a/src/sonic-yang-mgmt/tests/test_cfghelp.py
+++ b/src/sonic-yang-mgmt/tests/test_cfghelp.py
@@ -1,7 +1,7 @@
 import json
 import subprocess
 import os
-
+from shlex import join
 from unittest import TestCase
 
 output1="""\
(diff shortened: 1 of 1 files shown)


src/sonic-yang-mgmt/tests/test_cfghelp.py
```diff
@@ -126,8 +126,8 @@ def setUp(self):
         self.script_file = os.path.join(self.test_dir, '..', 'sonic-cfg-help')
 
     def run_script(self, argument):
-        print('\n    Running sonic-cfg-help ' + argument)
-        output = subprocess.check_output(self.script_file + ' ' + argument, shell=True)
+        print('\n    Running sonic-cfg-help ', argument)
```
qiluo-msft, line 129: >, [](http://example.com/codeflow?start=45&length=1)

Semgrep alerts on this line?

maipbui: No, but `argument` variable is a list, cannot be concatenated with string in the print statement.

qiluo-msft:
```suggestion
        print('\n    Running sonic-cfg-help ' + shlex.join(argument))
```

maipbui: Updated.

qiluo-msft (approved):

PR #12549, opened 2022-10-29
[openssh]: Restore behavior of ClientAliveCountMax=0

saiarcot895:
OpenSSH 8.2 changed the behavior of ClientAliveCountMax=0 such that setting it to 0 disables connection-killing entirely when the connection is idle. Revert that change.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [x] 202205


src/openssh/Makefile
```diff
@@ -23,7 +23,6 @@ ifeq ($(CROSS_BUILD_ENVIRON), y)
 	patch -p1 < ../patch/cross-compile-changes.patch
 	dpkg-buildpackage -rfakeroot -b -us -uc -a$(CONFIGURED_ARCH) -Pcross,nocheck -j$(SONIC_CONFIG_MAKE_JOBS) --admindir $(SONIC_DPKG_ADMINDIR)
 else
-	sudo http_proxy=$(http_proxy) apt-get -y build-dep openssh
```
Blueve, line 26: Is safe to remove this?

saiarcot895: Yes, in fact, it is necessary. Prior to this commit, the broadcom build failed due to simultaneous package installations happening. There shouldn't be any package installations happening during builds (except those that `slave.mk` itself is doing).

As per `git blame`, this line got added in the PR that added armhf crosscompile support. I'm not sure why it was added, or how I missed it in the PR review, or how the build has been working so well so far, but this shouldn't be here.

yxieca (approved):

PR #12562, opened 2022-10-31
[sonic-py-common] Remove subprocess with shell=True

maipbui:

#### Why I did it
#### How I did it
#### How to verify it
Manual test
Pass UT

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205

### DIFF
--- a/src/sonic-py-common/sonic_py_common/device_info.py
+++ b/src/sonic-py-common/sonic_py_common/device_info.py
@@ -533,17 +533,17 @@ def _valid_mac_address(mac):
 
 def run_command(cmd):
     proc = subprocess.Popen(cmd, universal_newlines=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
-    (mac, err) = proc.communicate()
-    return (mac, err)
+    (out, err) = proc.communicate()
+    return (out, err)
 
 
 def run_command_pipe(cmd0, cmd1, cmd2):
-    exitcodes, mac = getstatusoutput_noshell_pipe(cmd0, cmd1, cmd2)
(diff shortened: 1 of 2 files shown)


src/sonic-py-common/sonic_py_common/device_info.py
```diff
@@ -531,7 +531,27 @@ def _valid_mac_address(mac):
     return bool(re.match("^([0-9A-Fa-f]{2}[:-]){5}([0-9A-Fa-f]{2})$", mac))
 
 
+def run_command(cmd):
+    proc = subprocess.Popen(cmd, universal_newlines=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    (mac, err) = proc.communicate()
```
qiluo-msft, line 536: >mac [](http://example.com/codeflow?start=5&length=3)

`mac` is hard to understand. Do you mean `out`?

src/sonic-py-common/sonic_py_common/device_info.py
```diff
@@ -531,7 +531,27 @@ def _valid_mac_address(mac):
...
+    return (mac, err)
+
+
+def run_command_pipe(cmd0, cmd1, cmd2):
+    exitcodes, mac = getstatusoutput_noshell_pipe(cmd0, cmd1, cmd2)
```
qiluo-msft, line 541: >mac [](http://example.com/codeflow?start=15&length=3)

The same.

qiluo-msft (approved):

PR #12597, opened 2022-11-03
Fix removing of old docker container

kv-y:
#### Why I did it

scripts/collect_docker_version_files.sh doesn't remove existing docker container.
It's possible to get "Error response from daemon: Conflict. The container name ... is already in use by container ...".

#### How I did it

Use $DOCKER_CONTAINER instead of $DOCKER_IMAGE to remove existing docker container.

#### How to verify it

Manually create docker container and run scripts/collect_docker_version_files.sh
Existing container will be successfully removed in this case.

### DIFF
--- a/scripts/collect_docker_version_files.sh
+++ b/scripts/collect_docker_version_files.sh
@@ -15,8 +15,8 @@ mkdir -p $TARGET_VERSIONS_PATH
 export DOCKER_CLI_EXPERIMENTAL=enabled
 
 # Remove the old docker container if existing
-if docker container inspect $DOCKER_IMAGE > /dev/null 2>&1; then
-    docker container rm $DOCKER_IMAGE > /dev/null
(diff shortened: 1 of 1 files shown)


kv-y:
@xumia 
Could you please review this PR?

You added scripts/collect_docker_version_files.sh for Repr build in 2020 and this bug exists since then.
So if we really need to remove existing container before creating new one then this PR fixes this bug.
If we want to get an error message when container already exists then we can remove this code.

But now in this code we try to inspect and remove docker image with docker container command like this;
`docker container inspect sonic-slave-bullseye-user:20d23eb25b2 > /dev/null 2>&1`
It's not correct anyway. This error was not detected cause you redirect all stderr/stdout to /dev/null.

kv-y:
@xumia or @lguohan 

Could you please review this PR?
There were many fixes in scripts/collect_docker_version_files.sh last week for vcache, but this incorrect code for removing old docker container is still here.

lguohan (approved):

PR #12628, opened 2022-11-07
Make determine-reboot-cause into oneshot service

cytsao1:
#### Why I did it
"show reboot-cause history" does not report the latest reboot-cause because process-reboot-cause does not wait for determine-reboot-cause to complete running before it starts.
Currently the system just assumes that process-reboot-cause will run after determine-reboot-cause based on timing. This fix is fixing the race condition that might happen if the determine-reboot-cause does not finish before process-reboot-cause.

#### How I did it
This PR makes determine-reboot-cause.service into a oneshot service to ensure process-reboot-cause only runs once determine-reboot-cause completes.
This fix depends on a PR for process-reboot-cause.service dependency, which needs to be committed to 202205 branch. 
https://github.com/sonic-net/sonic-buildimage/pull/11213

#### How to verify it
Run "show reboot-cause" and "show reboot-cause history"
The latest entry in reboot-cause history should match the output of show reboot-cause.

Can check the journal logs of determine-reboot-cause.service and process-reboot-cause.service after a reboot to confirm that process-reboot-cause only runs after determine-reboot cause.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [x] 202205

Update determine-reboot-cause.service to be oneshot service to ensure it completes before process-reboot-cause.service runs


cytsao1:
@gechiang Could you check this?

prgeor (approved):

bmridul:
@abdosi , @gechiang Pls include this in 202205 branch.

PR #12653, opened 2022-11-09
Add platform files for Innovium platform

rck-innovium:
Signed-off-by: rck-innovium rck@innovium.com

#### Why I did it
Add platform files for critical processes and default qos config for Innovium platforms

#### How I did it
Added default files for critical processes and qos config

#### How to verify it

Tested with autorestart/test_container_autorestart.py::test_containers_autorestart

```
PASSED autorestart/test_container_autorestart.py::test_containers_autorestart[sonic-xxx-dut-sonic-xxx-dut|syncd]
```

Add platform files for critical processes and default qos config for Innovium platforms

### DIFF
--- a/platform/innovium/docker-syncd-invm/ivm_start.sh
+++ b/platform/innovium/docker-syncd-invm/ivm_start.sh
@@ -1,0 +1,39 @@
+#!/usr/bin/env bash
+
+TOPO_CONF=/usr/share/sonic/platform/topo.conf
+if [ -f ${TOPO_CONF} ]; then
+ topo=`cat ${TOPO_CONF}`
+else
+ topo="none"
+fi
+
+if [ ${topo} != "none" ]; then
+    BUFFER_CONFIG=/usr/share/sonic/hwsku/buffers_defaults_${topo}.j2
+    QOS_CONFIG=/usr/share/sonic/hwsku/qos_defaults_${topo}.j2
+
+    if [ "$(/usr/local/bin/sonic-cfggen -d -v TC_TO_PRIORITY_GROUP_MAP)" ] || \
+       [ "$(/usr/local/bin/sonic-cfggen -d -v MAP_PFC_PRIORITY_TO_QUEUE)" ] || \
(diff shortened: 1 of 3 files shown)


rlhui (approved):

platform/innovium/docker-syncd-invm/ivm_start.sh
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env bash
+
+TOPO_CONF=/usr/share/sonic/platform/topo.conf
+if [ -f ${TOPO_CONF} ]; then
+ topo=`cat ${TOPO_CONF}`
+else
+ topo="none"
+fi
+
+if [ ${topo} != "none" ]; then
```
lguohan, line 10: need to explain the motivation why this skip logic is added? this should not be platform specific.

rck-innovium: @lguohan 

This logic is added to apply the default buffers and QOS config based on the device's SKU and role.

These defaults are applied only if the user has not applied any qos/buffer config in the config\_db during boot. In out platform, either the operator's qos config or the defaults must be applied in config\_db for traffic to flow in the switch. 

This check is made platform specific since in other platforms, the SAI adapter may have internally set up the buffer and qos defaults and may not need the defaults in the config\_db. However, in our platforms all the configuration are set up via the SAI APIs.

Please refer the below for sample default files.

[https://github.com/sonic-net/sonic-buildimage/blob/master/device/wistron/x86_64-wistron_6512_32r-r0/wistron_6512_32r/buffers_defaults_t1.j2](https://github.com/sonic-net/sonic-buildimage/blob/master/device/wistron/x86_64-wistron_6512_32r-r0/wistron_6512_32r/buffers_defaults_t1.j2)

https://github.com/sonic-net/sonic-buildimage/blob/master/device/wistron/x86_64-wistron_6512_32r-r0/wistron_6512_32r/qos_defaults_t1.j2

rck-innovium:
@lguohan  Please review and merge.

prgeor (approved):

PR #12678, opened 2022-11-11
Add California-SB237 feature. Requires to change default user password

andriydnvd:
Signed-off-by: Andriy Dobush <andriyd@nvidia.com>

#### Why I did it
Add support of California-SB237 conformance.
https://github.com/sonic-net/SONiC/tree/master/doc/California-SB237

#### How I did it
Expire user passwords during build

#### How to verify it
Enable build flag and check if default user is prompted for a new password

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205


liuh-80 (commented):
Is there a PR to validate this change work?

And as we know if we enable this feature following will break:
1. https://github.com/sonic-net/sonic-buildimage/blob/master/check_install.py run in build image and will try to login to device.
2. UTs in sonic-mgmt will break because they will login to device.

I think we need fix 1, and what's the plan for fix 2?

Also, we need UT in sonic-mgmt to cover this feature.

Yarden-Z:
> Is there a PR to validate this change work?
> 
> And as we know if we enable this feature following will break:
> 
>     1. https://github.com/sonic-net/sonic-buildimage/blob/master/check_install.py run in build image and will try to login to device.
> 
>     2. UTs in sonic-mgmt will break because they will login to device.
> 
> 
> I think we need fix 1, and what's the plan for fix 2?
> 
> Also, we need UT in sonic-mgmt to cover this feature.

Why do we need UT in sonic-mgmt for this feature?
This only works on initial boot, so this will require adjustments...
Initial discussions in regards to testing talked about handling the infra required changes and that will suffice as the test suite.
Are there additional areas where you think we should test this?

andriydnvd:
> Is there a PR to validate this change work?
> 
> And as we know if we enable this feature following will break:
> 
> 1. https://github.com/sonic-net/sonic-buildimage/blob/master/check_install.py run in build image and will try to login to device.

Updated check_install.py in build image to be ready for prompting new password. Original password is reverted back by script

liuh-80:
I create following PR to validate this change:
https://github.com/sonic-net/sonic-buildimage/pull/12983

As you can see, after enable this feature, every UT failed in 'prepare testbed' stage:
https://dev.azure.com/mssonic/build/_build/results?buildId=185495&view=logs&j=a35c5435-5100-57cf-f655-60b256043d84&t=c9f737d3-1e43-5e60-26d7-f79fbfea8dbb

I'm not check detail, but I think that step failed because current code trying to login but not reset password.

I suggest fix UT break issue in sonic-mgmt before this PR merge. the fix can reuse the logic in check_install.py.

Current UT PR only add new UT to cover this feature, but existed UTs will break when this feature enabled.

liuh-80 (approved):
According to discussion in https://github.com/sonic-net/sonic-mgmt/pull/6863
There will be another PR to change sonic-mgmt to support this feature enabled.

check_install.py
```diff
@@ -11,6 +11,7 @@ def main():
     parser = argparse.ArgumentParser(description='test_login cmdline parser')
     parser.add_argument('-u', default="admin", help='login user name')
     parser.add_argument('-P', default="YourPaSsWoRd", help='login password')
+    parser.add_argument('-P2', default="Test@2022", help='new password')
```
qiluo-msft, line 14: This will conflict with `-P`. For multiple char option, suggest `--P2`.
`-P2` == `-P` + `-2` in many command option conventions.
ref: https://stackoverflow.com/a/21286343

qiluo-msft: This is still applicable.

andriydnvd: Agree, can it be some literal line '-N', default="Test@2022", help='new password'  ?

qiluo-msft (approved):

PR #12692, opened 2022-11-13
Add Secure Boot Support

davidpil2002:
#### Why I did it
Add Secure Boot support to SONiC OS.
Secure Boot (SB) is a verification mechanism for ensuring that code launched by a computer's UEFI firmware is trusted. It is designed to protect a system against malicious code being loaded and executed early in the boot process before the operating system has been loaded.
#### How I did it
Added a signing process to sign the following components:
shim, grub, Linux kernel, and kernel modules when doing the build, and when feature is enabled in build time according to the HLD explanations (the feature is disabled by default).
#### How to verify it
There are self-verifications of each boot component when building the image, in addition, there is an existing end-to-end test in sonic-mgmt repo that checks that the boot succeeds when loading a secure system (details below).

How to build a sonic image with secure boot feature: (more description in HLD)
1. Required to use the following build flags from rules/config:
- SECURE_UPGRADE_MODE="dev"
- SECURE_UPGRADE_DEV_SIGNING_KEY="/path/to/private/key.pem"
- SECURE_UPGRADE_DEV_SIGNING_CERT="/path/to/cert/key.pem"
After setting those flags should build the sonic-buildimage.
2. Before installing the image, should prepared the setup (switch device) with the follow:
- check that the device support UEFI
- stored pub keys in UEFI DB
- enabled Secure Boot flag in UEFI
3. How to run a test that verify the Secure Boot flow:
  The existing test "test_upgrade_path" under "sonic-mgmt/tests/upgrade_path/test_upgrade_path", is enough to validate proper boot
  You need to specify the following arguments:
  Base_image_list your_secure_image
  Taget_image_list your_second_secure_image
  Upgrade_type cold
  And run the test, basically the test will install the base image given in the parameter and then upgrade to target image by doing cold reboot and validates all the services are up and working correctly

note: the path to the existing sonic-mgmt test can be found in the link in the HLD description.

- [ ] master
- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [X] 202211

No REDIS DB used in this feature

HLD link: https://github.com/sonic-net/SONiC/pull/1028


davidpil2002:
hi @lguohan, 

Apparently, there is some test issue when connecting to the host, the attached error is below.
For now, I didn't find something related to this specific commit, can you help with this issue?

details from log:
link: https://www.testbed-tools.org/scheduler/testplan/7215
error msg:
E           AnsibleConnectionFailure: Host unreachable

@liat-grozovik ++

rules/config
```diff
@@ -208,6 +208,14 @@ SONIC_ENABLE_IMAGE_SIGNATURE ?= n
 # The absolute path should be provided.
 SONIC_ENABLE_SECUREBOOT_SIGNATURE ?= n
 
+# Full Secure Boot feature flags.
+# SECURE_UPGRADE_DEV_SIGNING_KEY - path to development signing key, used for image signing during build
+# SECURE_UPGRADE_DEV_SIGNING_CERT - path to development signing certificate, used for image signing during build
+# SECURE_UPGRADE_MODE - enum value for secure upgrade mode, valid options are "dev", "prod" and "no_sign"
+SECURE_UPGRADE_DEV_SIGNING_KEY = /sonic/files/nv_onyx_key.pem
+SECURE_UPGRADE_DEV_SIGNING_CERT =  /sonic/files/nv_onyx_key_certificate.pem
```
lguohan, line 216: what does this nv_onyx mean here? is that nvidia specific?

lguohan:
how can we test this SB feature? is there a virtual environment that we can test and validate?

liat-grozovik:
If I am not mistaken, we are working on a functional test as it requires specific image built with it


scripts/signing_kernel_modules.sh
```diff
@@ -0,0 +1,120 @@
+#!/bin/bash
+# This script is signing kernel modules by using sign-file tool 
+
+usage() {
+    cat <<EOF
+$0: # Display Help
+$0 <LINUX_KERNEL_VERSION> <PEM_CERT> <PEM_PRIVATE_KEY> <LOCAL_SIGN_FILE> <LOCAL_EXTRACT_CERT> <KERNEL_MODULES_DIR>
+Sign kernel modules in <KERNEL_MODULES_DIR> using private & public keys.
```
lguohan, line 8: i suggest to use getopt to define those option clearly instead of just providing all these raw arugments

scripts/signing_secure_boot_dev.sh
```diff
@@ -0,0 +1,152 @@
...
+
+# rename signed vmlinuz with the name vmlinuz without signed suffix
+sudo cp ${CURR_VMLINUZ}-signed ${CURR_VMLINUZ}
+
+sudo bash scripts/secure_boot_signature_verification.sh -c $PEM_CERT -e ${CURR_VMLINUZ}
+
+#########################
+# Kernel Modules signing
+#########################
+sudo bash scripts/signing_kernel_modules.sh $LINUX_KERNEL_VERSION ${PEM_CERT} ${PEM_PRIV_KEY}
```
lguohan, line 150: i do not see module dir provided as required by the script, why?

installer/default_platform.conf
```diff
@@ -405,7 +475,8 @@ bootloader_menu_config()
     fi
 
     DEFAULT_GRUB_SERIAL_COMMAND="serial --port=${CONSOLE_PORT} --speed=${CONSOLE_SPEED} --word=8 --parity=no --stop=1"
-    DEFAULT_GRUB_CMDLINE_LINUX="console=tty0 console=ttyS${CONSOLE_DEV},${CONSOLE_SPEED}n8 quiet $CSTATES"
+    # DEFAULT_GRUB_CMDLINE_LINUX="console=tty0 console=ttyS${CONSOLE_DEV},${CONSOLE_SPEED}n8 quiet $CSTATES"
+    DEFAULT_GRUB_CMDLINE_LINUX="console=tty0 console=ttyS${CONSOLE_DEV},${CONSOLE_SPEED}n8 loglevel=7 systemd.log_level=debug $CSTATES"
```
lguohan, line 479: why are we adding loglevel=7 here? is this related to sb?

lguohan:
the design doc says only signing mellanox specific kernel module, wonder why?

lguohan:
the pr needs to provide steps how to build sb-enabled image for other to try it out. otherwise, it would be difficult for community members to check.

rules/config
davidpil2002: just an example of a private and public key location and names, 
modified to something more generic:
```
key_private_dev.pem
key_certificate_dev.pem
```

installer/default_platform.conf
davidpil2002: I used it to debug the kernel, 
removed - thanks

scripts/signing_kernel_modules.sh
davidpil2002: done

davidpil2002:
> how can we test this SB feature? is there a virtual environment that we can test and validate?

with a physical device:
1. enable the Secure Boot flag in UEFI.
2. store the public keys in UEFI.
3. build the sonic-buildimage using the Secure Boot flags according to the HLD explanations.
4. manually, check that the boot and init flow works.
5. automation - run the corresponding sonic-mgmt tests. (pointer to sonic-mgmt test should be attached in the HLD description)

In general - a successful secure boot flow is one where you load all kernel modules and nothing fails.
---------------------------------
with a virtual machine:
1. should verify that the VM supports UEFI and supports storing pub keys in the DB and Secure Boot UEFI flag.
In addition - currently the Linux is pulling the public key from the UEFI DB in order to verify the sbsign output for each secure boot component. The VM would have to support this basic Linux flow.
2. Do steps like a physical device.

scripts/signing_secure_boot_dev.sh
davidpil2002: I added the parameter, in general, if you don't, it will sign all the kernel modules from the CWD

davidpil2002:
> the design doc says only signing mellanox specific kernel module, wonder why?

fix the doc to generic vendor:
`fsroot-vendor_name instead fsroot-mellanox`

davidpil2002:
> the pr needs to provide steps how to build sb-enabled image for other to try it out. otherwise, it would be difficult for community members to check.

The HLD shows all the steps in general, but I added as per your requested more description in this PR.

davidpil2002:
> the pr needs to provide steps how to build sb-enabled image for other to try it out. otherwise, it would be difficult for community members to check.

I added the steps in the PR description

davidpil2002:
> the pr needs to provide steps how to build sb-enabled image for other to try it out. otherwise, it would be difficult for community members to check.

done

davidpil2002:
> the design doc says only signing mellanox specific kernel module, wonder why?

fixed

davidpil2002 (commented):
Ack all the comments/concerns & add one commit

build_debian.sh
```diff
@@ -618,6 +619,56 @@ then
 
 fi
 
+# #################
+#   secure boot 
+# #################
+if [[ $SECURE_UPGRADE_MODE == 'dev' || $SECURE_UPGRADE_MODE == "prod" && $SONIC_ENABLE_SECUREBOOT_SIGNATURE != 'y' ]]; then
+    # note: SONIC_ENABLE_SECUREBOOT_SIGNATURE is a feature that signing just kernel,
```
sacnaik, line 626: I think we should club this now and single solution for UEFI secure boot, We added this earlier to sign the Linux kernel and copy the signed boot loader component (Ex:  SHIM and Monolithic GRUB) from HW vendor ONIE ESP partition.

build_debian.sh
```diff
@@ -618,6 +619,56 @@ then
...
+#   secure boot 
+# #################
+if [[ $SECURE_UPGRADE_MODE == 'dev' || $SECURE_UPGRADE_MODE == "prod" && $SONIC_ENABLE_SECUREBOOT_SIGNATURE != 'y' ]]; then
+    # note: SONIC_ENABLE_SECUREBOOT_SIGNATURE is a feature that signing just kernel, 
+    # SECURE_UPGRADE_MODE is signing all the boot component including kernel.
+    # its required to do not enable both features together to avoid conflicts.
+    echo "Secure Boot support build stage: Starting .."
+
+    # debian secure boot dependecies
+    sudo LANG=C DEBIAN_FRONTEND=noninteractive chroot $FILESYSTEM_ROOT apt-get -y install      \
```
sacnaik, line 632: Are we downloading prebuilt binaries of SHIM and GRUB?

davidpil2002: yes, we are downloading the shim and grub from Debian sources unsigned, them the build sign them and copy to the corresponding boot location.

davidpil2002:
Hi @abhishekd0,
Hope all is well, 
we answered all the comments and concerns, can you approve this PR?
Thanks
David

build_debian.sh
sacnaik: Isn't this security gap? Who does auditing on these prebuilt images ? If SONiC vendor is signing these images then the SONiC vendor should able to audit them right or why not take Debian distribution signed images directly.

lguohan:
this pr seems too large, and contains multiple features, can we break this pr into smaller pieces so that it becomes more clear.

for example, the kernel module signing, shim signing, we can break into separate pr.

lguohan:
the instruction to repro seems not clear enough, we need more detailed instructions on how to build/how to verify especially for in a virtual environment.

scripts/signing_kernel_modules.sh
```diff
@@ -0,0 +1,124 @@
...
+LINUX_KERNEL_VERSION
+PEM_CERT                             public key (pem format)
+PEM_PRIVATE_KEY                      private key (pem format)
+LOCAL_SIGN_FILE                      path of the sign-file tool for signing Kernel Modules, if the value is empty it will used the sign-file installed in /usr/lib/linux-kbuild-<version>/scripts
+LOCAL_EXTRACT_CERT                   path of the extract-cert tool for Extract X.509 certificate, if the value is empty it will used the extract-cert installed in /usr/lib/linux-kbuild-<version>/scripts
+KERNEL_MODULES_DIR                   root directory of all the kernel modules to be sign by the script, if the value empty it will use the call script location as root.
+
+Runs examples:
+1. sudo bash scripts/signing_kernel_modules.sh 5.10.0-8-2 cert.pem priv-key.pem
+2. sudo bash scripts/signing_kernel_modules.sh 5.10.0-8-2 cert.pem priv-key.pem fsroot-mellanox /usr/lib/linux-kbuild-5.10/scripts/extract-cert /usr/lib/linux-kbuild-5.10/scripts/sign-file
```
lguohan, line 20: please update the example here since you updated the getopts?

scripts/signing_secure_boot_prod.sh
```diff
@@ -0,0 +1 @@
+#In this script Vendor should code  the logic to build a secure boot image by using vendor flows
```
lguohan, line 1: this looks like placeholder, i do not know how to deal with it. why is this actually a placeholder.

scripts/signing_kernel_modules.sh
```diff
@@ -0,0 +1,124 @@
...
+
+if [ ! -d "$KERNEL_MODULES_DIR" ]; then
+  # If the user do not provide a KERNEL_MODULES_DIR, the script is going to search in the script call path for Kernel modules.
+  KERNEL_MODULES_DIR="./"
+  echo "KERNEL_MODULES_DIR set to default path: $KERNEL_MODULES_DIR"
+
+fi
+
+# find all the kernel modules.
+modules_list=$(sudo find ${KERNEL_MODULES_DIR} -name "*.ko")
```
lguohan, line 86: what is the kernel_module_dir scattered in multiple places? i also remember bash has some limitation on the length of maximum arguments, i doubt this will work.

scripts/signing_kernel_modules.sh
```diff
@@ -0,0 +1,124 @@
...
+${LOCAL_EXTRACT_CERT} ${local_sign_key} ${local_sign_cert}
+
+# Do sign for each found module
+kernel_modules_cnt=0
+for mod in $modules_list
+do
+    echo "signing module named: ${mod} .."
+    echo "${LOCAL_SIGN_FILE} sha512 ${local_sign_key} ${local_sign_cert} ${mod}"
+    kernel_modules_cnt=$((kernel_modules_cnt+1))
+    sudo ${LOCAL_SIGN_FILE} sha512 ${local_sign_key} ${local_sign_cert} ${mod}
```
lguohan, line 114: i susgest remove sudo in the script, if the script requires root user, check it in the script to prevent people to run it from normal user. using sudo inside the script to elavate the privlege in general is not a good practice.

build_debian.sh
```diff
@@ -618,6 +619,56 @@ then
...
+
+    if [[ $SECURE_UPGRADE_MODE == 'dev' ]]; then
+        # development signing & verification 
+
+        if [ ! -f $SECURE_UPGRADE_DEV_SIGNING_KEY ]; then
+            echo "Error: SONiC SECURE_UPGRADE_DEV_SIGNING_KEY=$SECURE_UPGRADE_DEV_SIGNING_KEY key missing"
+            exit 1
+        fi
+
+        sudo bash scripts/signing_secure_boot_dev.sh -a $CONFIGURED_ARCH \
```
lguohan, line 649: remove sudo

davidpil2002:
> this pr seems too large, and contains multiple features, can we break this pr into smaller pieces so that it becomes more clear.
> 
> for example, the kernel module signing, shim signing, we can break into separate pr.

the signing flow has dependencies between the different boot loaders, so breaking it into small PRs will create build issues and problems to test the commits separately.
As a result of that, if better to code in one PR.

scripts/signing_secure_boot_prod.sh
davidpil2002: This script is a placeholder to give modularity to the vendor for the signing flow. 
The vendor can save in this script his own signing flow, signing in the same build server or external sign server, and can decide which sign tools to use.

Details explained in the HLD: 
https://github.com/sonic-net/SONiC/blob/6ec54974b9814b49e5acef1dc544b51d91198424/doc/secure_boot/hld_secure_boot.md#ArchitectureDesign

scripts/signing_kernel_modules.sh
davidpil2002: this code is working for sure, it was already tested on different devices.
The reason that the code is looking for all the kernel object by using "find", is to make the signing flow more robust, that because each vendor have its own kernel modules and it is not feasible to save all the kernel module's path. So by finding all the kernel modules, if any vendor add some new kernel module the code will find it and sign it as well.

scripts/signing_kernel_modules.sh
davidpil2002: done

scripts/signing_kernel_modules.sh
davidpil2002: this script exists and is called just when building the image and cannot be accessed in a runtime setup, anyway I removed the sudo inside the script and saved the sudo only in the build_debian.sh script like was usage in this way on different occasions in the script.

build_debian.sh
davidpil2002: explained in the comment above.

build_debian.sh
```diff
@@ -618,6 +619,56 @@ then
...
+                                                     -l $LINUX_KERNEL_VERSION \
+                                                     -c $SECURE_UPGRADE_DEV_SIGNING_CERT \
+                                                     -p $SECURE_UPGRADE_DEV_SIGNING_KEY
+    elif [[ $SECURE_UPGRADE_MODE == "prod" ]]; then
+        #  Here Vendor signing should be implemented
+        OUTPUT_SEC_BOOT_DIR=$FILESYSTEM_ROOT/boot
+        sudo bash scripts/signing_secure_boot_prod.sh $CONFIGURED_ARCH $FILESYSTEM_ROOT $LINUX_KERNEL_VERSION $OUTPUT_SEC_BOOT_DIR
+        
+        # verifying all EFI files and kernel modules in $OUTPUT_SEC_BOOT_DIR
+        bash scripts/secure_boot_signature_verification.sh -e $OUTPUT_SEC_BOOT_DIR \
```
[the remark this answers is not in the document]
davidpil2002: remove the bash and code ./ instead

davidpil2002:
> the instruction to repro seems not clear enough, we need more detailed instructions on how to build/how to verify especially for in a virtual environment.

There are instructions on how to build the sonic image with secure boot feature in this PR description and in the HLD, 
About the verification, there is also descriptions in the PR. 
pls let me know which steps are not clear.
about virtualization, you can test the feature in the same way as a physical device, you need to have a VM that supports UEFI with a secure boot feature and support to add certificates to the UEFI DB.

davidpil2002 (commented):
remove sudo redundancy calls and update code comments

build_debian.sh
davidpil2002: well, SONiC downloads many Debian packages from Debian sources, so probably should be required to do a robust solution, because even if you audit the download of secure boot loaders components still SONiC it's downloading some kernel modules or some tools from some Debian sources without audit.
In general, the solution should be to have some server as a repository to contain all the deb resources and authenticate them...
About the other question, in our HLD review, we are giving the Vendor the ability to sign with his own keys so they are not limitate to using the Microsoft pub key only.

scripts/signing_kernel_modules.sh
lguohan: in the build image, it should have multiple kernel module dir, can KERNEL_MODULES_DIR be a list?

installer/default_platform.conf
```diff
@@ -469,9 +539,34 @@ EOF
...
+mkdir -p /boot/efi/EFI/debian/
+echo "cp $tmp_config /boot/efi/EFI/debian/grub.cfg"
+cp $tmp_config /boot/efi/EFI/debian/grub.cfg
+
 # Add extra linux command line
 echo "EXTRA_CMDLINE_LINUX=$extra_cmdline_linux"
 GRUB_CMDLINE_LINUX="$GRUB_CMDLINE_LINUX $extra_cmdline_linux"
+GRUB_CFG_LINUX_CMD=""
+GRUB_CFG_INITRD_CMD=""
+if [ "$firmware" = "uefi" ] ; then
```
lguohan, line 561: what about the bios is uefi, but do not support secure boot?

scripts/signing_kernel_modules.sh
lguohan: i can see sudo still here.

scripts/signing_secure_boot_dev.sh
```diff
@@ -0,0 +1,121 @@
...
+# clean old files
+clean_file ${CURR_VMLINUZ}-signed
+
+echo "signing ${CURR_VMLINUZ} .."
+${EFI_SIGNING} -p $PEM_PRIV_KEY -c $PEM_CERT -e ${CURR_VMLINUZ} -s ${CURR_VMLINUZ}-signed
+
+# rename signed vmlinuz with the name vmlinuz without signed suffix
+mv ${CURR_VMLINUZ}-signed ${CURR_VMLINUZ}
+
+./scripts/secure_boot_signature_verification.sh -c $PEM_CERT -e ${CURR_VMLINUZ}
```
lguohan, line 114: i do not understand this, why the verification is performed before the kernel module signing part?

build_debian.sh
lguohan: i am not sure sonic is downloading any kernel binary modules, all kernel and modules are build from sources.

davidpil2002: in the build_debian.sh many pkgs are downloading and not build from source if you search for the word "apt-get" you will find, i.e like this one:
`sudo LANG=C chroot $FILESYSTEM_ROOT apt-get -y install apt-transport-https \
                                                       ca-certificates \
                                                       curl \
                                                       gnupg2 \
                                                       software-properties-common

`
But it's not a big change for me, so if its better I can download it in a Makefile, similar like we did with src/iptables.

scripts/signing_secure_boot_dev.sh
davidpil2002: this step it's verifying the Linux Kernel (vmlinuz) and not the kernel modules.

rules/config
```diff
@@ -208,6 +208,14 @@ SONIC_ENABLE_IMAGE_SIGNATURE ?= n
 # The absolute path should be provided.
 SONIC_ENABLE_SECUREBOOT_SIGNATURE ?= n
 
+# Full Secure Boot feature flags.
+# SECURE_UPGRADE_DEV_SIGNING_KEY - path to development signing key, used for image signing during build
+# SECURE_UPGRADE_DEV_SIGNING_CERT - path to development signing certificate, used for image signing during build
+# SECURE_UPGRADE_MODE - enum value for secure upgrade mode, valid options are "dev", "prod" and "no_sign"
+SECURE_UPGRADE_DEV_SIGNING_KEY = /sonic/your/private/key/path/private_key.pem
+SECURE_UPGRADE_DEV_SIGNING_CERT =  /sonic/your/certificate/path/cert.pem
```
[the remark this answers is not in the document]
davidpil2002: if the `SECURE_UPGRADE_MODE` is `no_sign` the other parameter will be not use.
By default, according the HLD the feature mode is `no_sign` so it's not matter what it's the content of the other parameters

installer/default_platform.conf
davidpil2002: Good comment, 
first of all, it will work, `linuxefi` command it's a command similar than `linux` to load the kernel from grub.
The different it's that `linuxefi` is supported by UEFI system (when secure boot enabled or disabled).
The `linux` command is not supporting some new feature that exist in UEFI like Secure Boot.
So, in general it should be better to always use this command in UEFI systems, and this also give the possibility to move from secure boot enable and disable with flexibility.

scripts/signing_secure_boot_dev.sh
```diff
@@ -0,0 +1,121 @@
+#!/bin/bash
```
sacnaik, line 1: DEV vs PROD are actually key types right? Why to mandate the signing method as DEV vs PROD ?
Why not DEV keys are external signing server or Production signing is based on local public key. Not every vendor would have external signing server.
By default SONiC should support sbsign method and signing private keys to be available within build system for both production and development purpose and additional hook to provide for vendor specific signing where vendor can implement their choice of signing or external signing service. The vendor specific signing should be kept out side of build hook and linked via build configuration.

build_debian.sh
davidpil2002: The source code unfortunately has also a vulnerability...
Well, I found that many packages are getting the sources code from http debian sources instead https. I already raise this issue to the community.
This issue can affect many features not just the SHIM and GRUB.

But, like I mentioned before, from my side it's not a big modification and I will do this small modification to GRUB & SHIM.
I probably will do a similar solution creating a Makefile like was used with iptables sources, but I will use a https source instead http source.
My plan is to use this source: 
https://deb.debian.org/debian/pool/main/g/grub2/

pls let me know if you have still any concern, thanks

sacnaik: Yes, if you could bring both SHIM and GRUB like iptable src would be good. Cisco can also add this as separate PR. We have raised PR. https://github.com/sonic-net/sonic-buildimage/pull/12901, OR you can also bring it that.

sacnaik: We can modify that PR so that it can go on top of this PR.

sacnaik: Let us know.

Yarden-Z: I'd suggest that we commit this PR with our current changes and then as an improvement to the PR - you'll commit your changes with the adjusted shim & grub src in a follow up PR

build_debian.sh
davidpil2002: could be OK, but we need to check that there are no conflicts,
are you planning to use this flag in other places also?

scripts/signing_kernel_modules.sh
davidpil2002: the KERNEL_MODULES_DIR is assigned to be the file system root, and we are searching all the kernel objects inside this directory.

Because the search is recursive, it's like searching in multiples list dir.
It's literally searching in all the file system.

scripts/signing_secure_boot_dev.sh
Yarden-Z: This is exactly why we implemented this feature the way we showcased here.
In our dev signing script, we are using a local script since we are going to assume a dev environment functions as an environment that has everything at hand - a build server, dev keys, dev certs all accessible. Also the reason why we are using it as a folder/location access to the keys (assuming they reside in some FS location accessible by the build server)
For production environment - we have no idea what each environment requests, it might be the same as dev (in this case we can just copy the script) or it could be that there is a signing server that resides in some secure location where you have to call specific API calls with a certain api token to send requests to it. 
Therefor - production value signing is open for the customer and dev signing is given here in full.
We provided a placeholder for the build hook to be called.

DavidZagury:
@davidpil2002 I have added to this branch the key and certificate mounts we discussed about

scripts/signing_kernel_modules.sh
davidpil2002: fixed

scripts/efi-sign.sh
```diff
@@ -0,0 +1,63 @@
...
+# - shim
+# - grub
+# - vmlinuz
+#
+print_usage() {
+    cat <<EOF
+
+$0: Usage
+$0 -p <PRIVATE_KEY_PEM> -c <CERT_PEM> -e <EFI_FILE> -s <EFI_FILE_SIGNED>
+Usage example: efi-sign.sh priv-key.pem pub-key.pem shimx64.efi shimx64-signed.efi
```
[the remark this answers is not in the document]
davidpil2002: fixed description

davidpil2002 (commented):
fix/answer all comments and upstream 1 commit with fixes

build_debian.sh
rlhui: @Yarden-Z - the current PR has this gap/comment that needs to be addressed.  Could this PR be updated/separated so that it does not download pre-built binaries?

liat-grozovik:
@lguohan , @rlhui can you please go over the changes following the review comments or let us know who is expected to review?
we would like to close the loop of the feature and backport it to 202211 asap.

build_debian.sh
sacnaik: It may be better to call/restate Prod and Dev as signing types instead of stating/giving the perspective of Dev signing expects keys from local and production signing expects remote server. Not all vendor has remote signing server that does not mean their images are not production image. HLD can provide guidelines on what is the preferable method in case of the release signing key to keep securely. I see this infra supports both local and remote key signing methods but the definition of the workflow may require to be restated.

build_debian.sh
```diff
@@ -628,6 +629,56 @@ then
...
+
+        sudo ./scripts/signing_secure_boot_dev.sh -a $CONFIGURED_ARCH \
+                                                  -r $FILESYSTEM_ROOT \
+                                                  -l $LINUX_KERNEL_VERSION \
+                                                  -c $SECURE_UPGRADE_DEV_SIGNING_CERT \
+                                                  -p $SECURE_UPGRADE_DEV_SIGNING_KEY
+    elif [[ $SECURE_UPGRADE_MODE == "prod" ]]; then
+        #  Here Vendor signing should be implemented
+        OUTPUT_SEC_BOOT_DIR=$FILESYSTEM_ROOT/boot
+        sudo ./scripts/signing_secure_boot_prod.sh $CONFIGURED_ARCH $FILESYSTEM_ROOT $LINUX_KERNEL_VERSION $OUTPUT_SEC_BOOT_DIR
```
sacnaik, line 667: The empty signing_secure_boot_prod.sh require the vendor to implement their signing method after the git pull. At build, time adding vendor-specific code after git pull, will make a dirty flag set for the image built indicating source code polluted after pull. It would be better to provide VENDOR_SPECIFIC_SIGNING_TOOL build environmental variable in rules/config to point vendor-specific signing tool.

davidpil2002: In continue of the yesterday meeting, we agree to this comment, and add a one more build environmental variable like you mentioned and mount this one in slave docker, commit will be push soon.

build_debian.sh
davidpil2002: This is a naming conflict as I see (not a functional issue).
I will do some additional review of the naming conventions and see if from the security POV it's OK to named it local and remote instead dev and prod.
@Yarden-Z 
@eddyk-nvidia

scripts/signing_secure_boot_prod.sh
```diff
@@ -0,0 +1,2 @@
+# In this script Vendor should code  the logic to build a secure boot image by using production vendor flows.
```
sacnaik, line 1: Do we still need this file now?? The Vendor scripts will be anyway outside the git repo.

slave.mk
```diff
@@ -376,6 +376,10 @@ $(info "USE_NATIVE_DOCKERD_FOR_BUILD"    : "$(SONIC_CONFIG_USE_NATIVE_DOCKERD_FO
 $(info "SONIC_USE_DOCKER_BUILDKIT"       : "$(SONIC_USE_DOCKER_BUILDKIT)")
 $(info "USERNAME"                        : "$(USERNAME)")
 $(info "PASSWORD"                        : "$(PASSWORD)")
+$(info "SECURE_UPGRADE_MODE"             : "$(SECURE_UPGRADE_MODE)")
+$(info "SECURE_UPGRADE_DEV_SIGNING_KEY"  : "$(SECURE_UPGRADE_DEV_SIGNING_KEY)")
+$(info "SECURE_UPGRADE_DEV_SIGNING_CERT" : "$(SECURE_UPGRADE_DEV_SIGNING_CERT)")
+$(info "SECURE_UPGRADE_PROD_SIGNING_TOOL": "$(SECURE_UPGRADE_PROD_SIGNING_TOOL)")
```
sacnaik, line 382: Update on HLD is needed for SECURE_UPGRADE_PROD_SIGNING_TOOL

sacnaik: And also PR description should be updated for SECURE_UPGRADE_PROD_SIGNING_TOOL

Makefile.work
```diff
@@ -312,6 +312,13 @@ endif
 ifneq ($(SECURE_UPGRADE_DEV_SIGNING_CERT),)
     DOCKER_RUN += -v $(SECURE_UPGRADE_DEV_SIGNING_CERT):$(SECURE_UPGRADE_DEV_SIGNING_CERT):ro
 endif
+# Mount the Signing prod tool in the slave container
+$(info "SECURE_UPGRADE_PROD_SIGNING_TOOL": "$(SECURE_UPGRADE_PROD_SIGNING_TOOL)")
+ifneq ($(SECURE_UPGRADE_PROD_SIGNING_TOOL),)
+	SECURE_UPGRADE_PROD_SIGNING_TOOL_DST = /sonic/scripts/$(shell basename -- $(SECURE_UPGRADE_PROD_SIGNING_TOOL))
+	DOCKER_RUN += -v $(SECURE_UPGRADE_PROD_SIGNING_TOOL):$(SECURE_UPGRADE_PROD_SIGNING_TOOL_DST):ro
+	SECURE_UPGRADE_PROD_SIGNING_TOOL := $(SECURE_UPGRADE_PROD_SIGNING_TOOL_SOURCE)
```
sacnaik, line 320: This is not correct I think. from where SECURE_UPGRADE_PROD_SIGNING_TOOL_SOURCE is coming from?

The tool path is pointed by SECURE_UPGRADE_PROD_SIGNING_TOOL_DST.
Anyway in my opinion you don't need to alter the path  from vendor-specific to /sonic/script
Mount them as it is
Ex:
DOCKER_RUN += -v $(SECURE_UPGRADE_PROD_SIGNING_TOOL):$(SECURE_UPGRADE_PROD_SIGNING_TOOL):ro

build_debian.sh
```diff
@@ -629,6 +630,56 @@ then
...
+
+        sudo ./scripts/signing_secure_boot_dev.sh -a $CONFIGURED_ARCH \
+                                                  -r $FILESYSTEM_ROOT \
+                                                  -l $LINUX_KERNEL_VERSION \
+                                                  -c $SECURE_UPGRADE_DEV_SIGNING_CERT \
+                                                  -p $SECURE_UPGRADE_DEV_SIGNING_KEY
+    elif [[ $SECURE_UPGRADE_MODE == "prod" ]]; then
+        #  Here Vendor signing should be implemented
+        OUTPUT_SEC_BOOT_DIR=$FILESYSTEM_ROOT/boot
+        sudo $SECURE_UPGRADE_PROD_SIGNING_TOOL $CONFIGURED_ARCH $FILESYSTEM_ROOT $LINUX_KERNEL_VERSION $OUTPUT_SEC_BOOT_DIR
```
sacnaik, line 668: Why don't we create a script argument the same as the dev, If the vendor doesn't want then the vendor can some inputs.
If the argument options are not supplied as -a -r -l -c and -p then the order of incoming argument has to be in a fixed order to map in vendors script.

IMO keeping the argument layout the same as DEV and PROD, it keeps great flexibility for vendors too. 
Vendors may ignore some argument inside their script.

slave.mk
sacnaik: SECURE_UPGRADE_DEV_SIGNING_KEY is a secure-boot key, Kernel module signing may use the same key or a different key. If the build system exposes SECURE_UPGRADE_KO_SIGNING_KEY then the vendor can have the flexibility to choose a separate key or pass both variables with the same keys.

liat-grozovik:
But this was not part of the HLD
So if this is now a new req I suggest it should be handled as a separate work on top of the existing one

files/build_templates/sonic_debian_extension.j2
```diff
@@ -78,6 +78,10 @@ fi
 # Update apt's snapshot of its repos
 sudo LANG=C DEBIAN_FRONTEND=noninteractive chroot $FILESYSTEM_ROOT apt-get update
 
+# Install efitools to support secure upgrade
+sudo LANG=C DEBIAN_FRONTEND=noninteractive chroot $FILESYSTEM_ROOT apt-get -y install efitools
```
qiluo-msft, line 82: >efitools [](http://example.com/codeflow?start=86&length=8)

We should only install these 2 packages in UEFI platforms?

installer/default_platform.conf
```diff
@@ -359,6 +359,68 @@ demo_install_uefi_grub()
...
+demo_install_uefi_shim()
+{
+    echo "demo_install_uefi_shim(): Installing Secure Boot components"
+
+    local demo_mnt="$1"
+    local blk_dev="$2"
+
+    # make sure /boot/efi is mounted
+    if ! mount | grep -q "/boot/efi"; then
+        mount /boot/efi
```
qiluo-msft, line 372: >mount [](http://example.com/codeflow?start=8&length=5)

Need to check the return code.

installer/default_platform.conf
```diff
@@ -359,6 +359,68 @@ demo_install_uefi_grub()
...
+    fi
+
+    # Look for the EFI system partition UUID on the same block device as
+    # the ONIE-BOOT partition.
+    local uefi_part=0
+    for p in $(seq 8) ; do
+        if sgdisk -i $p $blk_dev | grep -q C12A7328-F81F-11D2-BA4B-00A0C93EC93B ; then
```
qiluo-msft, line 379: >C12A7328-F81F-11D2-BA4B-00A0C93EC93B [](http://example.com/codeflow?start=43&length=36)

Magic string?

installer/default_platform.conf
```diff
@@ -359,6 +359,68 @@ demo_install_uefi_grub()
...
+
+    # make sure /boot/efi is mounted
+    if ! mount | grep -q "/boot/efi"; then
+        mount /boot/efi
+    fi
+
+    # Look for the EFI system partition UUID on the same block device as
+    # the ONIE-BOOT partition.
+    local uefi_part=0
+    for p in $(seq 8) ; do
```
qiluo-msft, line 378: >8 [](http://example.com/codeflow?start=19&length=1)

Magic number?

scripts/signing_kernel_modules.sh
```diff
@@ -0,0 +1,124 @@
...
+
+if [ ! -f ${LOCAL_EXTRACT_CERT} ]; then
+    echo "ERROR: LOCAL_EXTRACT_CERT=${LOCAL_EXTRACT_CERT} file does not exist"
+    usage
+    exit 1
+fi
+
+if [ ! -d "$KERNEL_MODULES_DIR" ]; then
+  # If the user do not provide a KERNEL_MODULES_DIR, the script is going to search in the script call path for Kernel modules.
+  KERNEL_MODULES_DIR="./"
```
qiluo-msft, line 80: >   [](http://example.com/codeflow?start=0&length=2)

always indent with 4xN spaces.

rules/config
```diff
@@ -217,6 +217,15 @@ SONIC_ENABLE_IMAGE_SIGNATURE ?= n
...
 
+# Full Secure Boot feature flags.
+# SECURE_UPGRADE_DEV_SIGNING_KEY - path to development signing key, used for image signing during build
+# SECURE_UPGRADE_DEV_SIGNING_CERT - path to development signing certificate, used for image signing during build
+# SECURE_UPGRADE_MODE - enum value for secure upgrade mode, valid options are "dev", "prod" and "no_sign"
+# SECURE_UPGRADE_PROD_SIGNING_TOOL - path to a vendor signing tool for production flow.
+SECURE_UPGRADE_DEV_SIGNING_KEY = /sonic/your/private/key/path/private_key.pem
+SECURE_UPGRADE_DEV_SIGNING_CERT =  /sonic/your/certificate/path/cert.pem
+SECURE_UPGRADE_MODE = "no_sign"
+SECURE_UPGRADE_PROD_SIGNING_TOOL ?= none
```
qiluo-msft, line 228: >none [](http://example.com/codeflow?start=36&length=4)

`none` is ambiguous. It could be a real path.

build_debian.sh
qiluo-msft, line 668: >SECURE_UPGRADE_PROD_SIGNING_TOOL [](http://example.com/codeflow?start=14&length=32)

you need to check if the var has valid path.

build_debian.sh
```diff
@@ -629,6 +630,56 @@ then
...
+            echo "Error: SONiC SECURE_UPGRADE_DEV_SIGNING_KEY=$SECURE_UPGRADE_DEV_SIGNING_KEY key missing"
+            exit 1
+        fi
+
+        sudo ./scripts/signing_secure_boot_dev.sh -a $CONFIGURED_ARCH \
+                                                  -r $FILESYSTEM_ROOT \
+                                                  -l $LINUX_KERNEL_VERSION \
+                                                  -c $SECURE_UPGRADE_DEV_SIGNING_CERT \
+                                                  -p $SECURE_UPGRADE_DEV_SIGNING_KEY
+    elif [[ $SECURE_UPGRADE_MODE == "prod" ]]; then
```
qiluo-msft, line 665: >elif [](http://example.com/codeflow?start=4&length=4)

We should minimize the difference between two modes.
1. share OUTPUT_SEC_BOOT_DIR
2. verify EFI files and kernel modules
3. verify vmlinuz file

Makefile.work
sacnaik: The SECURE_UPGRADE_PROD_SIGNING_TOOL when not used should be empty. But in rules/config the default to none. The "none" is itself some value and hence make 
ifneq ($(SECURE_UPGRADE_PROD_SIGNING_TOOL),) instruction wont work

I think this supposed to be
ifneq ($(SECURE_UPGRADE_PROD_SIGNING_TOOL),none)

slave.mk
davidpil2002: In our HLD we are using the same private key for signing boot loaders components and kernel modules, this is in order to make it more easily from the user's POV.
If you propose to enhance it and have multiple keys, I suggest doing a future meeting about it.
For now, in this PR, let's save the HLD conventions about it.

Makefile.work
davidpil2002: agree, I will add this condition

scripts/signing_secure_boot_prod.sh
davidpil2002: fixed

build_debian.sh
davidpil2002: done

rules/config
davidpil2002: done

build_debian.sh
davidpil2002: well, the verification is done by the same script name `./scripts/secure_boot_signature_verification.sh`, but it's not exactly the same flow.

in development, the script `scripts/signing_secure_boot_dev.sh` is doing the verification after each local signing, so if there is some failure the build will stop immediately and not continue the signing of another component.

in prod, the sign happened in a remote server so all the signing logic coded in the script `SECURE_UPGRADE_PROD_SIGNING_TOOL` will finish, and then the verification step will start doing the verification of all the components.

scripts/signing_kernel_modules.sh
davidpil2002: done

files/build_templates/sonic_debian_extension.j2
davidpil2002: we are installing the `efitools` always, and the following just when the feature is enabled:
shim-unsigned
grub-efi
code can be found in `build_debian.sh`.

installer/default_platform.conf
davidpil2002: actually, I copied those few lines from a SONiC code in the function `demo_install_uefi_grub` that was coded 7 years ago.
I will add your comment and create a function to avoid code duplication.

build_debian.sh
sacnaik: Ack

scripts/signing_secure_boot_prod.sh
sacnaik: Ack

slave.mk
sacnaik: Okay. Ack

Makefile.work
sacnaik: Ack

sacnaik (approved):

zhangyanzhao:
@davidpil2002 can you please sign the EasyCLA? That will block the PR merge.

installer/default_platform.conf
Yarden-Z: @davidpil2002 ?

installer/default_platform.conf
Yarden-Z: @davidpil2002

files/build_templates/sonic_debian_extension.j2
qiluo-msft: you comment is not relevant to my comment.

davidpil2002: pls can you elaborate your comment?

davidpil2002: I think now Im understand your question, efitools is only for UEFI platform, so we can add a condition that checking if the platform support UEFI and if yes install, other not.
I will add this condition, thanks

davidpil2002 (commented):
fix all comments, pls approve/review @qiluo-msft @Yarden-Z

abdosi:
/ azp run Azure.sonic-buildimage

abdosi:
@sacnaik can you create PR for 202205 for this change ?

davidpil2002:
> @sacnaik can you create PR for 202205 for this change ?

This PR is pending the @qiluo-msft approval, after the approval, we can cherry-pick. is it that OK?

build_debian.sh
qiluo-msft: It seems that the verification method could be unified. So the if-elif block will be smaller. If you really want "verification after each local signing, so if there is some failure the build will stop immediately", you may have duplicated verification, and it does not harm much.

rules/config
```diff
@@ -217,6 +217,15 @@ SONIC_ENABLE_IMAGE_SIGNATURE ?= n
 # The absolute path should be provided.
 SONIC_ENABLE_SECUREBOOT_SIGNATURE ?= n
 
+# Full Secure Boot feature flags.
+# SECURE_UPGRADE_DEV_SIGNING_KEY - path to development signing key, used for image signing during build
```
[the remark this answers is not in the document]
qiluo-msft: As an audience, I am confused by the name and curious why it is only limited to image upgrade, not normal bootup.

build_debian.sh
davidpil2002: I not sure 100% its possible, probably the best it's to schedule a meeting to understand exactly the details.
Or maybe if it possible, can you approve this PR as is, and I will schedule a meeting and create an additional PR to review your suggestion?
I preferred to do not postpone this merge only in result of a verification step, if its ok from your side as well.

qiluo-msft (approved):

ITJamie:
Please note the following issue with a build of master: https://github.com/sonic-net/sonic-buildimage/issues/14316. install failed due to missing mokutil

build_debian.sh
```diff
@@ -631,6 +632,62 @@ then
...
+        if [ ! -f $SECURE_UPGRADE_PROD_SIGNING_TOOL ]; then
+            echo "Error: SONiC SECURE_UPGRADE_PROD_SIGNING_TOOL=$SECURE_UPGRADE_PROD_SIGNING_TOOL script missing"
+            exit 1
+        fi
+
+        sudo $SECURE_UPGRADE_PROD_SIGNING_TOOL $CONFIGURED_ARCH $FILESYSTEM_ROOT $LINUX_KERNEL_VERSION $OUTPUT_SEC_BOOT_DIR
+        
+        # verifying all EFI files and kernel modules in $OUTPUT_SEC_BOOT_DIR
+        sudo ./scripts/secure_boot_signature_verification.sh -e $OUTPUT_SEC_BOOT_DIR \
+                                                             -c $SECURE_UPGRADE_DEV_SIGNING_CERT \
```
gordon-nexthop, line 680: Why is this `DEV_SIGNING_CERT` and not something like `PROD_SIGNING_CERT` are these the same thing?

davidpil2002: In few words, when using DEV the flow is use an script inside the repo for the SECURE BOOT,
other than PROD, which is using an additional parameter that support to inject an external script for using in this SECURE BOOT flow, the content of this external script is ukown, is not in the repo, so, vendor can use their logic without exposing their methodology.
For example, the signing can be done in an external server, and not in the same build server.

gordon-nexthop: Thank you for responding! Yes I understand the difference between the two flows. To my understanding this is still in the PROD flow but we are still using `DEV_SIGNING_CERT`. Are the certificates just the same then, irrespective of which flow we are using?

PR #12557, opened 2022-10-31
[Build] Support j2 template for debian sources

xumia:
#### Why I did it
1. Unify the Debian mirror sources, not code change required for Debian mirror sources when upgrading to the new Debian distribution.
2. Support to customize the Debian mirror sources during the build

#### Which release branch to backport (provide reason below if selected)

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205

PR #12702, opened 2022-11-14
[Build] Update SLAVE_BASE_TAG and DPKG cache if Debian mirrors were changed

kv-y:
#### Why I did it

After #12557 sources.lists are generated.
So we need to recalculate SLAVE_BASE_TAG if mirrors were changed.
Also we need to rebuild DPKG cache in this case.

#### How I did it

1. Use generated sources.list for SLAVE_BASE_TAG
2. Add MIRROR_URLS and MIRROR_SECURITY_URLS to SONIC_COMMON_FLAGS_LIST

Signed-off-by: Konstantin Vasin k.vasin@yadro.com

### DIFF
--- a/Makefile.cache
+++ b/Makefile.cache
@@ -70,6 +70,7 @@ SONIC_COMMON_FILES_LIST :=  $(if $(wildcard cache.skip.common),, .platform slave
 SONIC_COMMON_FLAGS_LIST :=  $(CONFIGURED_PLATFORM) \
                             $(CONFIGURED_ARCH) \
                             $(BLDENV) \
+                            $(MIRROR_URLS) $(MIRROR_SECURITY_URLS) \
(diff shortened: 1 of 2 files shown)


kv-y:
@xumia FYI
Before #12557 debian mirrors were in sonic-slave-xxx/Dockerfile.j2.
Also sources.list in dockers/docker-base-xxx were in deps list for docker-base-xxx.

Now if I change MIRROR_URLS and/or MIRROR_SECURITY_URLS then sonic-slave-xxx images and docker-base (from DPKG cache) doesn't update.
I'm not sure about solution based on SONIC_COMMON_FLAGS_LIST, but I can't use files (sources.list) from docker-base-xxx directory because they are generated after dependency check.

xumia (approved):

xumia:
@kv-y , thanks for fixing it.

PR #12005, opened 2022-09-08
05.Version cache - docker dpkg caching support

Kalimuthu-Velappan:
This feature caches all the deb files during docker build and stores them
into version cache.

It loads the cache file if already exists in the version cache and copies the extracted
deb file from cache file into Debian cache path( /var/cache/apt/archives).

The apt-install always installs the deb file from the cache if exists, this
avoid unnecessary package download from the repo and speeds up the overall build.

The cache file is selected based on the SHA value of version dependency
files.

#### Which release branch to backport (provide reason below if selected)

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205

PR #12713, opened 2022-11-15
[Build] fix make clean

kv-y:
#### Why I did it

`make clean` is broken after #12000:
```
bash: -c: line 1: syntax error near unexpected token `;'
bash: -c: line 1: `make -f slave.mk PLATFORM= PLATFORM_ARCH=amd64 MULTIARCH_QEMU_ENVIRON=n 
...
MIRROR_URLS= MIRROR_SECURITY_URLS=  Q=@ clean; ; '
make[1]: *** [Makefile.work:531: clean] Error 2
```
#### How I did it
Remove a conditional for `clean` command.


kv-y:
@Kalimuthu-Velappan 

You added this conditional: ifeq ($(filter clean,$(MAKECMDGOALS)),)  in #12000.
If it's not for some special purpose then I think we can just remove it to fix `make clean` command.

Kalimuthu-Velappan:
> @Kalimuthu-Velappan
> 
> You added this conditional: ifeq ($(filter clean,$(MAKECMDGOALS)),) in #12000. If it's not for some special purpose then I think we can just remove it to fix `make clean` command.
> 
> Another possible solution is:
> 
>     1. Use filter or filter-out not only for clean but for other targets like distclean and list too.
> 
>     2. Fix double semicolon error.

You can go ahead and commit the current fix. I will address this in a separate PR.

xumia (approved):

lguohan:
@kv-y , i resolve the conflict, can you check if the pr still good?

kv-y:
> You can go ahead and commit the current fix. I will address this in a separate PR.

@Kalimuthu-Velappan 
Are you still going to address this bug in a separate PR?
I mean this code `filter clean,$(MAKECMDGOALS)` was added as part of you PR for vcache.
PR for vcache was merged in #12000 and #12005.

If you don't need a special handling for `clean` command for vcache framework and it's just for refactoring then we can merge this PR.

PR #12779, opened 2022-11-21
[build]: force to use xz for deb control file compression

lguohan:

#### Why I did it
in ubuntu 22.04, dpkg-deb start to use zstd compression for build control file. such format is not supported within stretch docker build environment.

therefore add -Zxz option for force using xz for control file which is the previous behavior.

#### How I did it
use -Zxz to force using xz compression

#### How to verify it
build on ubuntu 22.04

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205


lguohan:
dup of https://github.com/sonic-net/sonic-buildimage/pull/12823

PR #12803, opened 2022-11-22
[Build] tune wget settings for debootstrap

kv-y:
#### Why I did it

wget is used by debootstrap to download deb packages for base system.
Sometime it can fail to download one or more packages and we get the  error:

E: Couldn't download packages: libffi7
chroot: failed to run command '/debootstrap/debootstrap': No such file or directory

#### How I did it

1. Set number of tries to 3 (default is 20)
If 3 is not enough then we need to fix deb server or our network connection.

2. Set the connect timeout to 10 seconds
(default value for Linux is 127 sec if tcp_syn_retries = 6)
There is no reason to wait so long.

3. Set the read (and write) timeout to 15 seconds.
 (default value for wget is 900 sec)
Again in this case there is no reason to wait so long.
Optimal solution is to do next try.

4. Consider "connection refused" a transient error and try again.
This is the most important option here.
Sometimes wget fails to connect to servers like deb.debian.org and doesn't retry.
Most of the time even 1 additional try is enough to reconnect successfully.


kv-y:
@xumia 
Could you please review?

Debootstrap fails was very annoying to me. Because it's a final part of build.
These errors were not rare when we download packages for debootstrap from deb.debian.org or another public debian mirrors. trafficmanager.net is more stable but it's not used for arm (see scripts/build_debian_base_system.sh)

src/debootstrap/tune-wget-settings.patch
```diff
@@ -0,0 +1,33 @@
+From a056a2d8cec783002b13c3592846e83ea1e9b5a2 Mon Sep 17 00:00:00 2001
+From: Konstantin Vasin <k.vasin@yadro.com>
+Date: Sun, 20 Nov 2022 19:22:59 +0300
+Subject: [PATCH] tune wget settings
+
+1. Set number of tries to 3 (default is 20).
+2. Set the connect timeout to 10 seconds
+(default value for Linux is 127 sec if tcp_syn_retries == 6).
+3. Set the read (and write) timeout to 15 seconds.
```
xumia, line 9: Does it have impact on the users who have very bad network, for the first try? For example, very long latency, it can download the file with long time, but retry not help. It will change the default options only on the retry attempts (the second retry and the third retry), or all steps including the first attempt?

kv-y: It changes 1st attempt too.
When I tested with deb.debian.org most fails were like this:

`Resolving deb.debian.org (deb.debian.org)... 199.232.138.132, 2a04:4e42:3a::644
Connecting to deb.debian.org (deb.debian.org)|199.232.138.132|:80...`
After that it's waiting 127 sec and fail or try again if --retry-connrefused is set.

Also one time I got read error:
`Resolving deb.debian.org (deb.debian.org)... 199.232.138.132, 2a04:4e42:3a::644
Connecting to deb.debian.org (deb.debian.org)|199.232.138.132|:80... connected.
HTTP request sent, awaiting response...`
After that it's waiting 900 sec and try again.

So I reduced this timeouts to spedup build.
But ofc I can do additional tests using  [netem](https://wiki.linuxfoundation.org/networking/netem)
It allows to simulate bad network connection and customize latency, rate and packet loss.

kv-y: Results with bad network connection were more stable than high bandwidth network connection.
Maybe this error is some sort of "rate limit" of public debian servers.

Also I have tested another public debian mirror.
And unlike deb.debian.org it sometimes responses me after 60-100 sec connection timeout.
So I decided to remove both these options: --connect-timeout and --read-timeout and use default values.
I'm still sure  that short values are better for more faster build if we have a stable network connection.
But as a generic solution default values of timeouts are acceptable.

kv-y:
`wget -t 3 --retry-connrefused` passed a test successfully.
Run 200 times `debootstrap --download-only --variant=minbase --arch arm64 bullseye $1 http://deb.debian.org/debian`
There were 6 retries due to connection timeout. Without fix it would lead to 6 failed builds.
With this fix deb packages were successfully downloaded after 1 retry:

```
Resolving deb.debian.org (deb.debian.org)... 199.232.138.132, 2a04:4e42:3a::644
Connecting to deb.debian.org (deb.debian.org)|199.232.138.132|:80... failed: Connection timed out.
Connecting to deb.debian.org (deb.debian.org)|2a04:4e42:3a::644|:80... failed: Network is unreachable.
Retrying.

(try: 2)  http://deb.debian.org/debian/pool/main/u/util-linux/bsdutils_2.36.1-8+deb11u1_arm64.deb
Connecting to deb.debian.org (deb.debian.org)|199.232.138.132|:80... connected.
HTTP request sent, awaiting response... 200 OK 
Length: 145188 (142K) [application/vnd.debian.binary-package]
```

Yes, we have 127 sec timeout here between 1st and 2nd try, but it's not critical due to low rate of this errors.

kv-y:
@xumia 
Could you please review updated version?

PR #12819, opened 2022-11-24
supporting redfish

pettershao-ragilenetworks:

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205


Blueve:
Do we have any design specification regarding this feature? Why is it not containerized? 
How the redfish server enabled and configured? Is there a feature flag or build parameter to disable it?

lguohan:
adding redfish support is a new feature, please go through standard sonic community design review process.

pettershao-ragilenetworks:
I did not record specific data. But this is the reason, if there is no request, you need to use curl to get data from bmc, and each query needs to start a curl process. pmon is a circular query, which will pull up multiple processes in a loop, which consumes a lot. Using request does not need to pull up the process. @lguohan

PR #12824, opened 2022-11-24
better solution for STATIC_ROUTE_EXPIRY_TIME check

jcaiMR:
#### Why I did it
1. swss-common behavior change between 202012 and master
   https://github.com/sonic-net/sonic-swss-common/commit/e95a4666ad90f13c3675efb9238f273df5232e91
2. This behavior change will cause different return value:
    master:  key not exist, swss-common api will return None.
    202012:  key not exist, swss-common api will return empty string.
3. Based on 1 and 2, modify the original static route expiry code to make the code safe for different kinds of key/value pair in both master and 202012.

#### How I did it
Add protection code when swss-common api return None/"". 

#### How to verify it
apply in bgpcfgd, verify different key/value pair. 

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012
- [ ] 202106
- [ ] 202111
- [x] 202205

### DIFF
--- a/src/sonic-bgpcfgd/bgpcfgd/static_rt_timer.py
+++ b/src/sonic-bgpcfgd/bgpcfgd/static_rt_timer.py
@@ -15,17 +15,15 @@ class StaticRouteTimer(object):
     MAX_TIMER     = 1800
 
     def set_timer(self):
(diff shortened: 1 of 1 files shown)


StormLiangMS (approved):
LGTM

Issue #12362, opened 2022-10-11
[Build] Incorrect info in Build Configuration for SONIC_VERSION_CONTROL_COMPONENTS


kv-y:
#### Description
Build configuration provides incorect info about SONIC_VERSION_CONTROL_COMPONENTS.
It always show value from rules/config.
https://github.com/sonic-net/sonic-buildimage/blob/e1765121b2fa3a5481e80c7db15da597964ab30f/slave.mk#L398

#### Steps to reproduce the issue:
1. make configure PLATFORM=broadcom SONIC_VERSION_CONTROL_COMPONENTS=all
2. check value of SONIC_VERSION_CONTROL_COMPONENTS in Build configuration.

#### Describe the results you received:
"SONIC_VERSION_CONTROL_COMPONENTS": "none"

#### Describe the results you expected:
"SONIC_VERSION_CONTROL_COMPONENTS": "all"

prgeor:
@liushilongbuaa can you help check this introduced by PR https://github.com/sonic-net/sonic-buildimage/pull/8116

liushilongbuaa:
It is a print error.

PR #12867, opened 2022-11-29
Convert all make rules to be overridable, clean up typos

andrewimeson:
Various guides on the internet suggest editing rules/config to enable features. This slightly complicates maintaining a build, because you either have to `git stash` (and hope that when you re-apply a change to the file hasn't caused a conflict) before pulling, or commit your changes and merge or rebase upstream changes. Prior to this these were the only configurable values in the file:

 - `SONIC_DPKG_CACHE_METHOD`
 - `SONIC_DPKG_CACHE_SOURCE`
 - `INCLUDE_KUBERNETES_MASTER`
 - `SONIC_ENABLE_IMAGE_SIGNATURE`
 - `SONIC_ENABLE_SECUREBOOT_SIGNATURE`
 - `PACKAGE_URL_PREFIX`
 - `SONIC_VERSION_CONTROL_COMPONENTS`
 - `REGISTRY_PORT`
 - `REGISTRY_SERVER`
 - `ENABLE_ASAN`
 - `DEFAULT_CONTAINER_REGISTRY`
 - `ENABLE_FIPS_FEATURE`
 - `ENABLE_FIPS`
 - `SONIC_SLAVE_DOCKER_DRIVER`

#### Why I did it

Since these are build flags, it shouldn't require modifying source-controlled files to build.

#### How I did it

Switching from variable assignment to optional variable assignment if unset.

#### How to verify it

```sh-session
$ make init
$ make configure PLATFORM=broadcom
$ make ENABLE_SFLOW_DROPMON=y SONIC_BUILD_JOBS=4 target/sonic-broadcom.bin
…
SONiC Build System

Build Configuration
"CONFIGURED_PLATFORM"             : "broadcom"
SONiC Build System

Build Configuration

…
"ENABLE_SFLOW_DROPMON"            : "y"
```

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205

Allow overriding rules/config values with make

Bartalina the house sheep @ 1-2 days old
image

### DIFF
--- a/rules/config
+++ b/rules/config
@@ -4,21 +4,20 @@
 
 # SONIC_CONFIG_PRINT_DEPENDENCIES - show dependencies for each invoked target.
 # Before executing rule for each target its dependencies are printed to console.
-# Uncomment next line to enable:
-# SONIC_CONFIG_PRINT_DEPENDENCIES = y
+SONIC_CONFIG_PRINT_DEPENDENCIES ?= n
 
 # SONIC_CONFIG_BUILD_JOBS - set number of jobs for parallel build.
 # Corresponding -j argument will be passed to make command inside docker
 # container.
-SONIC_CONFIG_BUILD_JOBS = 1
+SONIC_CONFIG_BUILD_JOBS ?= 1
 
 # SONIC_CONFIG_MAKE_JOBS - set number of parallel make jobs per package.
 # Corresponding -j argument will be passed to make/dpkg commands that build separate packages
-SONIC_CONFIG_MAKE_JOBS = $(shell nproc)
+SONIC_CONFIG_MAKE_JOBS ?= $(shell nproc)
 
 # DEFAULT_BUILD_LOG_TIMESTAMP - add timestamp in build log
 # Supported format: simple, none
-DEFAULT_BUILD_LOG_TIMESTAMP = none
+DEFAULT_BUILD_LOG_TIMESTAMP ?= none
 
 # SONIC_USE_DOCKER_BUILDKIT - use docker buildkit for build.
 # If set to y SONiC build system will set environment variable DOCKER_BUILDKIT=1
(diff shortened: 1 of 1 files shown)


andrewimeson:
I learned on the mailing list that I can create `rules/config.user` which will get read in after `rules/config`.

I think it may make sense to support both? It's a little confusing to have only some flags configureable via `make FLAG_NAME=y …`.

andrewimeson:
example of an issue due to this confusion
https://github.com/sonic-net/sonic-buildimage/issues/12362

rules/config
```diff
@@ -4,107 +4,102 @@
...
 # Special handling of the docker image file names is needed to avoid conflicts with
 # other SONiC build jobs on the same server. This requires changes to the Dockerfile.j2 FROM statement
 # in the dockers/ and platform/ subdirs to use a variable reference instead of an explicit image name.
-# SONIC_CONFIG_USE_NATIVE_DOCKERD_FOR_BUILD = y
+SONIC_CONFIG_USE_NATIVE_DOCKERD_FOR_BUILD ?= n
 
 # SONIC_CONFIG_ENABLE_COLORS - enable colored output in build system.
-# Comment next line to disable:
-# SONIC_CONFIG_ENABLE_COLORS = y
+SONIC_CONFIG_ENABLE_COLORS ?= y
```
saiarcot895, line 38: Contrary to the original description, this might actually be disabled by default. It's also not used anywhere (besides defining some `log_*` color functions), as far as I can tell, but it's probably best to default this to `n` instead.

saiarcot895:
> example of an issue due to this confusion #12362

This is actually due to a different reason, I think related to `Makefile.work` not propagating this variable into the slave docker container. However, I think it turns out that the intended value for that variable does get applied due to it getting stored in `src/sonic-build-hooks/buildinfo/config/buildinfo.config`. Regardless, this should probably get fixed to at least show the correct value.

PR #12915, opened 2022-12-02
[TestbedV2][master] Remove timeout in each step.

yutongzhang-microsoft:
Previously, we set timeout in each step such as Lock testbed, Prepare testbed, Run test and KVM dump. When some issue suck like retry happens in one step, it will cause timeout error, but actually, it only needs more time to success. In this pr, we remove the timeout limit in each step and control the timeout outside in each job. When the job runs more than four hours, it will be cancelled.

#### Why I did it
Previously, we set timeout in each step such as Lock testbed, Prepare testbed, Run test and KVM dump. When some issue suck like retry happens in one step, it will cause timeout error, but actually, it only needs more time to success. In this pr, we remove the timeout limit in each step and control the timeout outside in each job. When the job runs more than four hours, it will be cancelled.

#### How I did it
Remove the timeout parameter in each step, and control the timeout outside in each job.

#### How to verify it
Set the timeout of one job to 4 hours, and when timeout happens, azure pipeline will cancel this job.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205


yejianquan (approved):
LGTM

PR #12973, opened 2022-12-06
[MASIC] [azp] remove official-build-multi-asic.yml

wenyiz2021:
this official-build-multi-asic.yml file is not needed now, original purpose was to test stability of build & test of multi-asic topos, before multi-asic PR check were brought up.
testbedv2 has substituted classic multi-asic PR check.
resources could be saved in sonic-ma to have none standby agent and mostly no job will be running in sonic-ma.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211


SuvarnaMeenakshi:
Thank you, agreed that we can remove, we could remove the pipeline yml as well?

SuvarnaMeenakshi:
The multi-asic VS pipeline is disabled as we have the PR checker added for multi-asic VS.
https://dev.azure.com/mssonic/build/_build?definitionId=342&_a=summary

PR #13030, opened 2022-12-12
[Build] mount cgroup2 in chroot to fix build on ubuntu 22.04

kv-y:
#### Why I did it

Ubuntu 22.04 uses cgroup2 by default, but docker.sh doesn't  mount it.
As a result we get an error when trying to run `docker info` in chroot env:

ERROR: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?

#### How I did it

mount cgroup2 in chroot if all enabled kernel cgroup controllers are currently not in use by cgroup1

#### How to verify it

With this fix the build finished successfully on ubuntu 22.04

### DIFF
--- a/files/docker/docker
+++ b/files/docker/docker
@@ -63,7 +63,8 @@ cgroupfs_mount() {
 	# see also https://github.com/tianon/cgroupfs-mount/blob/master/cgroupfs-mount
 	if grep -v '^#' /etc/fstab | grep -q cgroup \
 		|| [ ! -e /proc/cgroups ] \
-		|| [ ! -d /sys/fs/cgroup ]; then
+		|| [ ! -d /sys/fs/cgroup ] \
+		|| [ "$(stat -fc '%T' /sys/fs/cgroup)" = "cgroup2fs" ]; then
 		return
 	fi
 	if ! mountpoint -q /sys/fs/cgroup; then
--- a/files/docker/docker
+++ b/files/docker/docker
@@ -67,6 +67,13 @@ cgroupfs_mount() {
 		return
 	fi
 	if ! mountpoint -q /sys/fs/cgroup; then
+		# cgroup2
+		# https://github.com/systemd/systemd/blob/5c6c587ce24096d36826418b5390599d1e5ad55c/src/shared/cgroup-setup.c#L35-L74
+		if awk '!/^#/ && $4 == 1 && $2 != 0 { count++ } END { exit count != 0 }' /proc/cgroups; then
+			mount -t cgroup2 cgroup2 /sys/fs/cgroup
+			return
+		fi
+		# cgroup1
 		mount -t tmpfs -o uid=0,gid=0,mode=0755 cgroup /sys/fs/cgroup
 	fi
 	(


files/build_templates/sonic_debian_extension.j2
```diff
@@ -64,6 +64,9 @@ clean_sys() {
 trap_push clean_sys
 sudo LANG=C chroot $FILESYSTEM_ROOT mount sysfs /sys -t sysfs
 
+[ -d "/sys/fs/cgroup" ] && [ "$(stat -fc '%T' /sys/fs/cgroup)" = "cgroup2fs" ] && \
+    sudo LANG=C chroot $FILESYSTEM_ROOT mount cgroup2 /sys/fs/cgroup -t cgroup2
```
lguohan, line 68: it would be better if you can modify the files/docker/docker incorporate cgroup2 support there.

kv-y: docker.sh is running in chroot, so I can't check fs on /sys/fs/cgroup then.
I think I need additional time to investigate this.

kv-y: @lguohan 
I think new version also will work, but I'm not sure about builds with enabled CROSS_BUILD_ENVIRON. I haven't test them.

kv-y:
So we need to mount cgroup in chroot environment on /sys/fs/cgroup.
Because inside chroot we don't know which cgroup version is used by the host we have two possible solutions:
1. cgroup tree for chroot is mounted by the host  (it was my 1st version of this fix)
2. cgroup tree is mounted inside chroot based on info from /proc/cgroups (it's current version of this fix)

My 2nd version based on this code from systemd: https://github.com/systemd/systemd/blob/5c6c587ce24096d36826418b5390599d1e5ad55c/src/shared/cgroup-setup.c#L35-L74

1. We parse info from /proc/cgroups
2. Skip header line started from #
3. Skip controller if it's disabled (4th column = 0)
4. Count number of controllers with non-zero of hierarchy_id (2nd column)

If this number is not zero then we assume some of controllers are used by host system and the host system uses hybrid or legacy cgroup tree. In this case we can't use unified cgroup tree inside chroot and mount old cgroup tree (v1).
If this number is zero then we assume host system uses unified cgroup tree and we need to mount cgroup2 inside chroot.

lguohan (approved):

kv-y:
@lguohan 

I think I can add one more fix in this code.
Now if we run script docker.sh 2nd time (e.g. `docker.sh restart`) when cgroup2 is already mounted on /sys/fs/cgroup this code is still executed:

```
	(
		cd /sys/fs/cgroup
		for sys in $(awk '!/^#/ { if ($4 == 1) print $1 }' /proc/cgroups); do
			mkdir -p $sys
			if ! mountpoint -q $sys; then
				if ! mount -n -t cgroup -o $sys cgroup $sys; then
					rmdir $sys || true
				fi
			fi
		done
	)
```
Nothing critical and these controllers will not be mounted.
But because this part is for cgroupv1 only I think need to slighly change conditions to not execute this code if cgroup2fs  already mounted on /sys/fs/cgroup.

Issue #12954, opened 2022-12-06
[CLI] [dhcp6relay] config vlan dhcp_relay is accepting ipv6 addresses and restarting the service 


vivekrnv:
#### Description

dhcpv6 relay needs config only in DHCP_RELAY table and not in VLAN table and thus the "config vlan" cli shouldn't accept an ipv6 address.

#### Steps to reproduce the issue:
```
1.  config vlan add 100
2.  config vlan dhcp_relay add 100 fc00::3:2
```

#### Describe the results you received:

```
Added DHCP relay destination addresses ['fc00::3:2'] to Vlan100
Restarting DHCP relay service...
```

#### Describe the results you expected:

CLI should reject ipv6 address with a relevant message.  

#### Output of `show version`:

Seen on all labels

PR #13039, opened 2022-12-13
[dhcpv6_relay] Fix DHCPv6 Relay config CLI to update DHCP_RELAY table

praveen-ramamoorthi:
#### Why I did it
Recently, DHCPv6 Relay DB table is changed from VLAN to DHCP_RELAY. The dhcp6relay process is spawned with the data fetched from this new table . The show CLIs were updated , but config CLIs were not updated to write to new table and still updated old VLAN table. Hence the DHCPv6 Relay was not working with CLI configurations.

#### How I did it
The config CLI script is modified to update new table DHCP_RELAY. The UT scripts are modified to test new table. 

#### How to verify it
Configure DHCPv6 Relay address for vlan using CLI and check if the config is updated in DHCP_RELAY table and verify if DHCPv6 Relay process is started for that vlan.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [x] 202205
- [x] 202211

Fixed configuration CLI for DHCPv6 Relay to update data in DHCP_RELAY table.

### DIFF
--- a/dockers/docker-dhcp-relay/cli-plugin-tests/conftest.py
+++ b/dockers/docker-dhcp-relay/cli-plugin-tests/conftest.py
@@ -10,6 +10,11 @@ def mock_cfgdb():
             'Vlan1000': {
                 'dhcp_servers': ['192.0.0.1']
             }
+        },
+        'DHCP_RELAY': {
+            'Vlan1000': {
+                'dhcpv6_servers': ['fc02:2000::100']
+            }
         }
     }
 
--- a/dockers/docker-dhcp-relay/cli-plugin-tests/test_config_dhcp_relay.py
+++ b/dockers/docker-dhcp-relay/cli-plugin-tests/test_config_dhcp_relay.py
@@ -154,7 +154,7 @@ class TestConfigVlanDhcpRelay(object):
             assert result.exit_code == 0
             assert result.output == config_vlan_add_dhcpv6_relay_output
             assert mock_run_command.call_count == 3
-            db.cfgdb.set_entry.assert_called_once_with('VLAN', 'Vlan1000', {'dhcp_servers': ['192.0.0.1'], 'dhcpv6_servers': ['fc02:2000::1']})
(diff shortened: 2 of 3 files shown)


praveen-ramamoorthi:
@kellyyeh  Can you please review this fix?

vivekrnv:
I don't think we should just redirect the dhcpv6_servers field to DHCP_RELAY table. We're also leaving out rfc6939_support field. 

i instead think we should update the VLAN cli to reject adding dhcpv6 config and just load the config with config load and restart the service. Related issue on that: https://github.com/sonic-net/sonic-buildimage/issues/12954. @kellyyeh , @yxieca what do you think

kellyyeh:
> I don't think we should just redirect the dhcpv6_servers field to DHCP_RELAY table. We're also leaving out rfc6939_support field.
> 
> i instead think we should update the VLAN cli to reject adding dhcpv6 config and just load the config with config load and restart the service. Related issue on that: #12954. @kellyyeh , @yxieca what do you think

@vivekrnv Agreed. Config add/del was a feature for VLAN table and we did not have plan to use config cli to update DHCP_RELAY. dhcp6relay does not support dynamic change in DHCP_RELAY table yet, so let's stick with config load and restart service only

praveen-ramamoorthi:
@vivekrnv @kellyyeh - This PR implements config load + service restart in Click command & thus avoid manual config load + service restart sequence. For #12954, I would prefer fixing the issue (update in proper redis tables) rather than rejecting the config.

From command definition perspective, VLAN related configurations are grouped under "config vlan <>" tree. This does not mean that all such configurations are stored in single redis DB table. For example, "config vlan member <>" & "config vlan proxy_arp <>" are being updated into VLAN_MEMBER & VLAN_INTERFACE tables respectively.

Regarding "rfc6939_support", no Click command is available yet which shall be treated as new command enhancement.

PR #13056, opened 2022-12-14
Keep custom rsyslog file for docker-sonic-vs

saiarcot895:
docker-sonic-vs doesn't have the infra needed for the syslog rate limit configuration, so it's not going to be rendering jinja templates to overwrite /etc/rsyslog.conf. This also means that syslog messages would get logged twice (because both the default /etc/rsyslog.conf file and /etc/rsyslog.d/50-default.conf are telling it to log to syslog).

Therefore, keep the custom static /etc/rsyslog.conf file for docker-sonic-vs.

Fixes sonic-net/sonic-swss#2570.

#### How to verify it

The logging rules aren't present twice in two different files:

```
root@13cb2ff3267e:/# grep -r authpriv /etc/rsyslog.*
/etc/rsyslog.d/50-default.conf:auth,authpriv.*                  /var/log/auth.log
/etc/rsyslog.d/50-default.conf:*.*;auth,authpriv.none           -/var/log/syslog
/etc/rsyslog.d/50-default.conf:#        auth,authpriv.none;\
/etc/rsyslog.d/50-default.conf:#        auth,authpriv.none;\
```

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211


saiarcot895:
FYI @Junchao-Mellanox

yxieca (approved):

stephenxs:
hI @saiarcot895 
The submodule `sonic-linux-kernel` was changed in this PR but it wasn't mentioned in the PR description. After that the submodule head is not correct. Probably by mistake or I missed something?
```
stephens@arc-build-server-2:/builds2/stephens/wip/tmp/sonic-buildimage/src$ git submodule update --init --recursive sonic-linux-kernel/
fatal: reference is not a tree: dbe40633f195fa8495f6d34d4c2048dfd9764830
Unable to checkout 'dbe40633f195fa8495f6d34d4c2048dfd9764830' in submodule path 'sonic-linux-kernel'
```
Can you please check it?
Thank you

PR #13059, opened 2022-12-14
Fix dhcp-relay and macsec dockers not being marked as local packages

saiarcot895:

#### Why I did it

When INSTALL_DEBUG_TOOLS=y is set (i.e. debug tools and containers are enabled), the non-debug versions of the dhcp-relay and macsec dockers are not added into SONIC_PACKAGES_LOCAL, and thus the tagging for those packages is not handled correctly when loading those dockers.

This issue happens when a regular build is done, and then the debug image is built with `INSTALL_DEBUG_TOOLS=y`. When this happens, the build system tries to load the docker image from the `target/`. However, this image was tagged as if it is a local package, but now in this build, on the debug dockers are considered local package. Because of this, there's a mismatch in the expected tag name, where the loaded docker uses the image tag, but the build system uses `latest` tag instead.

#### How I did it

To fix this, change the code that checks to see if something is part of a local package by searching for both just the package name and with the -dbg suffix to the name.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211


saiarcot895:
@stepanblyschak could you review this when you get a chance?

xumia:
Better to use the latest tag when saving the docker image.
A SONiC docker image does not have an individual version, it uses the image version. The image version varies in each of the commit. 

For example:
Commit 1, SONiC_VERSION=version 1, do an image build,  saving docker-dhcp-relay:<version 1>
Commit 2, do some changes not relative to docker-dhcp-relay, and build again, it will load the wrong version.

How about always saving docker-dhcp-relay:latest to docker-dhcp-relay.gz?

saiarcot895:
@xumia, that makes sense, dockers saved to disk can't have version-specific information.

I've modified my changes to save the docker with just the latest tag.

stepanblyschak:
@xumia @saiarcot895 Yes, that makes sense. I will rework the fix for https://github.com/sonic-net/sonic-buildimage/issues/11521

PR #13061, opened 2022-12-15
Generate the unique machine-id from mac address

kuanyu99:
#### Why I did it

1. The machine-id is generated during the building process and will be copied to each installed device. So each device will have the same id.
2. The machine-id will be different for each different image on the same device.

#### How I did it

1. Generate the machine-id from the eth0 MAC address during initializing the ram-file system
2. Use symbolic files target to the /run/machine-id to replace original /etc/machine-id and /var/lib/dbus/machine-id
3. Due to the existence of the symbolic files, the SONiC boot up process will be consistent after this commit. If the Linux system detect /etc/machine-id missing, it will trigger its first boot mechanism.

#### How to verify it

After the change, the behavior will be:
1. The machine-id will be different for each device with the same version image.
2. The machine-id will be the same for each different image on the same device.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012
- [x] 202106
- [x] 202111
- [x] 202205
- [x] 202211

This change can fit to any release which using build_debian.sh to build the SONiC image.

* Originally, the machine-id is generated randomly during the building process. And will be copied to each device during installing.
* After this, the machine-id will be generated according to mac address which is unique for each device. And will be consistent even installing new image.

### DIFF
--- a/build_debian.sh
+++ b/build_debian.sh
@@ -217,6 +217,9 @@ sudo chmod +x $FILESYSTEM_ROOT/etc/initramfs-tools/scripts/init-bottom/varlog
 #sudo chmod +x $FILESYSTEM_ROOT/etc/initramfs-tools/scripts/init-bottom/mgmt-intf-dhcp
 sudo cp files/initramfs-tools/union-fsck $FILESYSTEM_ROOT/etc/initramfs-tools/hooks/union-fsck
 sudo chmod +x $FILESYSTEM_ROOT/etc/initramfs-tools/hooks/union-fsck
+# Generate the machine-id from the MAC addr during initramfs setup
+sudo cp files/initramfs-tools/machineid $FILESYSTEM_ROOT/etc/initramfs-tools/scripts/init-bottom/machineid
+sudo chmod +x $FILESYSTEM_ROOT/etc/initramfs-tools/scripts/init-bottom/machineid
 pushd $FILESYSTEM_ROOT/usr/share/initramfs-tools/scripts/init-bottom && sudo patch -p1 < $OLDPWD/files/initramfs-tools/udev.patch; popd
 if [[ $CONFIGURED_ARCH == armhf || $CONFIGURED_ARCH == arm64 ]]; then
     sudo cp files/initramfs-tools/uboot-utils $FILESYSTEM_ROOT/etc/initramfs-tools/hooks/uboot-utils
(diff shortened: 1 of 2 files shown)


lguohan:
it is unclear what the machine id is used for, why unique machine id is needed.

kuanyu99:
From the man page, https://manpages.ubuntu.com/manpages/kinetic/en/man5/machine-id.5.html
It should be unique for the local system.
And it is also a requirement from our users.

lguohan:
i cannot agree with on current approach to use mac address and make it consistent across installation.

check the machine-id here.

https://wiki.debian.org/MachineId

> It's intended as an opaque, non-human-meaningful, persistent unique identifier for a machine (or more precisely an OS installation), used as a lookup key in state/configuration storage in the same sorts of places you might be tempted to use a hostname.

generating machine-id during installation makes sense

kuanyu99:
> i cannot agree with on current approach to use mac address and make it consistent across installation.
> 
> check the machine-id here.
> 
> https://wiki.debian.org/MachineId
> 
> > It's intended as an opaque, non-human-meaningful, persistent unique identifier for a machine (or more precisely an OS installation), used as a lookup key in state/configuration storage in the same sorts of places you might be tempted to use a hostname.
> 
> generating machine-id during installation makes sense

Regarding this page, I think the machine-id should be different for each version SONiC on a device.
So, instead of generating from MAC address, using a random 16 bytes id is a preferable way, right?

kuanyu99:
> i cannot agree with on current approach to use mac address and make it consistent across installation.
> 
> check the machine-id here.
> 
> https://wiki.debian.org/MachineId
> 
> > It's intended as an opaque, non-human-meaningful, persistent unique identifier for a machine (or more precisely an OS installation), used as a lookup key in state/configuration storage in the same sorts of places you might be tempted to use a hostname.
> 
> generating machine-id during installation makes sense

Regarding this page, I think the machine-id should be different for each version SONiC on a device.
So, instead of generating from MAC address, using a random 16 bytes id is a preferable way, right?

CharlieChenEC:
Hi @lguohan,

May I have your comment on the solution for machine-id generation on SONiC?

I think it's OK to revise the PR to generate machine-id during the installation stage.

What about the part on the content of 'machine-id'? If you think it should not use 'mac address' to generate 'machine-id'?

PR #13065, opened 2022-12-15
[dhcpmon] Fix dhcpmon socket filter and tx count issue

kellyyeh:
#### Why I did it
1. Fix issue caused by dualtor support PR https://github.com/sonic-net/sonic-buildimage/pull/11201
2. Improve code

#### How I did it
1. On single ToR, packets received count was duplicated due to socket filter set to "inbound"
2. Tx count not increasing due to filter set to "inbound". Added an outbound socket to count tx packets
3. Added vlan member interface mapping for Ethernet interface to vlan interface lookup in reference to PR https://github.com/sonic-net/sonic-dhcp-relay/pull/27
4. Exit when socket fails to initialize to allow dhcp_relay docker to restart

#### How to verify it
Tested on vstestbed single tor and dual tor, sent packets and verify printed out dhcpmon rx and tx counters is correct
- Correct number of tx increases
- Tx does not increase when ToR is on standby

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012
- [ ] 202106
- [ ] 202111
- [x] 202205
- [ ] 202211

### DIFF
--- a/src/dhcpmon/src/dhcp_device.cpp
+++ b/src/dhcpmon/src/dhcp_device.cpp
@@ -41,6 +41,7 @@
 #define DHCP_OPTIONS_HEADER_SIZE 240
 /** Offset of DHCP GIADDR */
 #define DHCP_GIADDR_OFFSET 24
+#define CLIENT_IF_PREFIX "Ethernet"
 
 #define OP_LDHA     (BPF_LD  | BPF_H   | BPF_ABS)   /** bpf ldh Abs */
 #define OP_LDHI     (BPF_LD  | BPF_H   | BPF_IND)   /** bpf ldh Ind */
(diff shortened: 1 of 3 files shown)


src/dhcpmon/src/dhcp_device.cpp
```diff
@@ -163,86 +213,102 @@ static void handle_dhcp_option_53(dhcp_device_context_t *context,
...
+                                    ethhdr->ether_shost[3] == context->mac[3] &&
+                                    ethhdr->ether_shost[4] == context->mac[4] &&
+                                    ethhdr->ether_shost[5] == context->mac[5]) ?
+                                    DHCP_TX : DHCP_RX;
+        int offset = 0;
+        int stop_dhcp_processing = 0;
+        while ((offset < (dhcp_option_sz + 1)) && dhcp_option[offset] != 255) {
+            switch (dhcp_option[offset])
+            {
+            case 53:
```
jcaiMR, line 252: Better to use enum for option53, like:
enum {
  OptionDHCPMessageType = 53,
}
If we only care about option53, maybe we don;t need switch case here, just use if and remove temp variable stop_dhcp_processing.

jcaiMR:
I think current diff still has duplicate packets issue, because listen socket bind on all interfaces.
So each time call dhcp_devman_add_intf( ) we will have a bind all input socket, and input packets will reach each all sockets.

Normally, for bind all interface, we just need one socket and give this socket big enough rcvd buffer size.
For TX socket as it bind to particular interface, so each interface a socket would be fine. 

And if we use one socket for all input interface binding, better to use batch process for better performance, that is one fd will trigger a batch recvfrom.

kellyyeh:
Each dhcpmon process will only monitor one vlan
https://github.com/sonic-net/sonic-buildimage/blob/3740f1efb7be8be325e93f37834adf7cc61e169f/dockers/docker-dhcp-relay/dhcp-relay.monitors.j2#L44

src/dhcpmon/src/dhcp_device.cpp
```diff
@@ -252,62 +308,28 @@ static void read_callback_dual_tor(int fd, short event, void *arg)
     while ((event == EV_READ) &&
```
jcaiMR, line 308: Here "event == EV_READ" will always true as we register this handler for EV_READ, we don't need to check it again.

src/dhcpmon/src/dhcp_device.cpp
```diff
@@ -163,86 +213,92 @@ static void handle_dhcp_option_53(dhcp_device_context_t *context,
...
  *
  * @return none
  */
-static void read_callback(int fd, short event, void *arg)
+static void read_tx_callback(int fd, short event, void *arg)
 {
     dhcp_device_context_t *context = (dhcp_device_context_t*) arg;
     ssize_t buffer_sz;
 
     while ((event == EV_READ) &&
```
jcaiMR, line 284: Here "event == EV_READ" will always true as we register this handler for EV_READ, we don't need to check it again.

src/dhcpmon/src/dhcp_device.cpp
```diff
@@ -495,18 +517,29 @@ static int init_socket(dhcp_device_context_t *context, const char *intf)
...
+        }
+
+        struct sockaddr_ll tx_addr;
+        memset(&tx_addr, 0, sizeof(tx_addr));
+        tx_addr.sll_ifindex = if_nametoindex(intf);
+        tx_addr.sll_family = AF_PACKET;
+        tx_addr.sll_protocol = htons(ETH_P_ALL);
+        if (bind(context->tx_sock, (struct sockaddr *) &tx_addr, sizeof(tx_addr))) {
             syslog(LOG_ALERT, "bind: failed to bind to interface '%s' with '%s'\n", intf, strerror(errno));
             break;
```
jcaiMR, line 544: If we failed to bind socket, we may just call exit() to let container restart.

src/dhcpmon/src/dhcp_device.cpp
```diff
@@ -495,18 +517,29 @@ static int init_socket(dhcp_device_context_t *context, const char *intf)
     int rv = -1;
 
     do {
-        context->sock = socket(AF_PACKET, SOCK_RAW | SOCK_NONBLOCK, htons(ETH_P_ALL));
-        if (context->sock < 0) {
+        context->rx_sock = socket(AF_PACKET, SOCK_RAW | SOCK_NONBLOCK, htons(ETH_P_ALL));
+        context->tx_sock = socket(AF_PACKET, SOCK_RAW | SOCK_NONBLOCK, htons(ETH_P_ALL));
+        if (context->rx_sock < 0 || context->tx_sock < 0) {
             syslog(LOG_ALERT, "socket: failed to open socket with '%s'\n", strerror(errno));
             break;
```
jcaiMR, line 524: if socket failed to create, we can just exit program.

src/dhcpmon/src/dhcp_device.cpp
```diff
@@ -666,21 +700,27 @@ int dhcp_device_start_capture(dhcp_device_context_t *context,
         }
         context->snaplen = snaplen;
 
-        if (setsockopt(context->sock, SOL_SOCKET, SO_ATTACH_FILTER, &dhcp_sock_bfp, sizeof(dhcp_sock_bfp)) != 0) {
+        if (setsockopt(context->rx_sock, SOL_SOCKET, SO_ATTACH_FILTER, &dhcp_inbound_sock_bfp, sizeof(dhcp_inbound_sock_bfp)) != 0) {
             syslog(LOG_ALERT, "setsockopt: failed to attach filter with '%s'\n", strerror(errno));
```
jcaiMR, line 704: Same like above, maybe we just exit program if any error happened in this function.

src/dhcpmon/src/dhcp_device.cpp
kellyyeh: event could be EV_PERIST or EV_READ

src/dhcpmon/src/dhcp_device.cpp
kellyyeh: Updated

src/dhcpmon/src/dhcp_device.cpp
kellyyeh: Updated

src/dhcpmon/src/dhcp_device.cpp
kellyyeh: Updated

src/dhcpmon/src/dhcp_device.cpp
jcaiMR: EV_PERIST controls the event is persistent, it makes you don't need to use event_add() again after callback. But when callback is called it must be a READ event, as you don't register read_tx_callback to any WRITE/DELETE events.

kellyyeh: Thanks I see. Removed event check

jcaiMR (approved):

yxieca (approved):

qiluo-msft:
What issue is fixed by this PR? could you explain in PR description?

yxieca:
@kellyyeh  is there a particular reason you didn't request to backport to 202211 branch?

yxieca:
@kellyyeh please create PR for 202205 branch for identify missing dependencies

kellyyeh:
@yxieca PR https://github.com/sonic-net/sonic-buildimage/pull/11201 is not cherry-picked to 202211 yet. Just added label for both PRs

qiluo-msft:

yxieca:
@kellyyeh this change cannot be cherry-picked cleanly to 20205 branch. can you raise separate PR?

liuh-80:
202012 branch PR merged: https://github.com/sonic-net/sonic-buildimage/pull/13441

Issue #428, opened 2017-03-23
Jenkins builds failing due to lack of free space


jleveque:
https://sonic-jenkins.westus.cloudapp.azure.com/job/common/job/buildimage-baseimage-pr/413/:
```
make[4]: Entering directory '/sonic/src/sonic-linux-kernel/linux-3.16.36'
dh_installdocs 
dh_installchangelogs
dh_strip
dh_compress
dh_fixperms
dh_installdeb
dh_gencontrol -- -Vkernel:Recommends=irqbalance,
dh_md5sums
dh_builddeb -- -Zxz 
dpkg-deb: building package `linux-image-3.16.0-4-amd64' in `../linux-image-3.16.0-4-amd64_3.16.36-1+deb8u2_amd64.deb'.
make[4]: Leaving directory '/sonic/src/sonic-linux-kernel/linux-3.16.36'
dpkg-deb (subprocess): compressing data member: lzma write error: No space left on device
dpkg-deb: error: subprocess <compress> from tar -cf returned error exit status 2
dh_builddeb: dpkg-deb -Zxz --build debian/linux-image-3.16.0-4-amd64-dbg .. returned exit code 2
debian/rules.real:173: recipe for target 'install-base' failed
```

https://sonic-jenkins.westus.cloudapp.azure.com/job/broadcom/job/buildimage-brcm-all-pr/83/:
```
running build
running build_ext
error: [Errno 28] No space left on device: 'build/temp.linux-x86_64-3.6-pydebug'
Makefile:627: recipe for target 'sharedmods' failed
make[3]: *** [sharedmods] Error 1
```

https://sonic-jenkins.westus.cloudapp.azure.com/job/cavium/job/buildimage-cavm-all-pr/410/:
```
/tmp/ccJe73Wu.s: Assembler messages:
/tmp/ccJe73Wu.s: Fatal error: can't write tests-converter_ut.o: No space left on device
```

stcheng:
https://sonic-jenkins.westus.cloudapp.azure.com/job/broadcom/job/buildimage-brcm-all/130/console
```
dpkg-deb: building package `swss' in `../swss_1.0.0_amd64.deb'.
dpkg-deb (subprocess): compressing control member: internal gzip write error: No space left on device
dpkg-deb: error: subprocess gzip -9c returned error exit status 2
dh_builddeb: dpkg-deb --build debian/swss .. returned exit code 2
[...]

PR #13067, opened 2022-12-15
[build] create a clean_docker_images profile

guillaumelambert:
SONiC image build process generates several docker images. They consume much disk space.
This space might need to be reclaimed once the SONiC image has been built to avoid disk full error messages.
Many of these docker images will not be reused at next build.

- create a shell script to detect docker images generated during build (with a blob pattern) in order to delete them and reclaim disk space
- declare in the Make a profile "clean_image" that points to this script

Issue  #428
for example

all branches are affected

 _______\)%%%%%%%%._ 
`''''-'-;   % % % % %'-._
        :b) \            '-. 
        : :__)'    .'    .' 
        :.::/  '.'   .' 
        o_i/   :    ; 
               :   .' 
                ''`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -98,3 +98,6 @@ clean reset showtag docker-cleanup sonic-slave-build sonic-slave-bash :
 # Freeze the versions, see more detail options: scripts/versions_manager.py freeze -h
 freeze:
 	@scripts/versions_manager.py freeze $(FREEZE_VERSION_OPTIONS)
+
+clean_docker_images:
+	@scripts/clean_docker_images.sh
--- a/scripts/clean_docker_images.sh
+++ b/scripts/clean_docker_images.sh
@@ -1,0 +1,41 @@
+#!/bin/sh
+
+DOCKER_CMD=docker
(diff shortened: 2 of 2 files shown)


clean_docker_images.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/sh
```
lguohan, line 1: can you move this script under scripts directory so that the structure becomes a little more cleaner?

guillaumelambert: sure

lguohan (approved):

clean_docker_images.sh
guillaumelambert: it seems some tests on the VM failed but I hardly believe it is due to this little modification...
Once logged in Azure pipelines website, is a bit weird.
The stage tests is in red alhough It is written "100% tests passed".
I clicked on "rerun the test stage" but I cannot see any effect.
Any hint about this ?

guillaumelambert: Let's try a rebase on current master and see.

PR #13070, opened 2022-12-15
[build] make apt installation quieter & faster

guillaumelambert:
Many apt commands are called inside chroot during SONiC build process. Though they are not interactive, they generate a lot of scrolling output in the console that slows down the whole SONiC installation process.

https://shallowsky.com/blog/linux/performance-of-scrolling-output.html

Most of this output has no added value and makes troubleshooting more complex.
It can be safely disabled by setting the right environment variables and options to apt-get. Note they cannot completely prevent underlying calls to dpkg from displaying a few output. Only a redirection to /dev/null can prevent the dpdk output from being displayed

Also since there are several wrappers to apt (apt-get/apt/aptitude) and since almost all of these apt calls are done inside chroot, it is preferable from a maintenance and efficiency perspective to
- create a shell function with all these parameters to install packages
- cascade contiguous chroot and apt commands to avoid mutiple context allocations

all branches are affected

[build] make apt installation quieter & faster

  //
 ('>
 /rr
*\))_


guillaumelambert:
@collivier @odd22
Hello
I pushed it as draft because I think there is still room for a few possible improvements.
and since I have rebased this old work on master, I would appreciate a double check.
Thanks in advance

PR #13084, opened 2022-12-16
[swss/syncd] remove dependency on interfaces-config.service

stepanblyschak:

#### Why I did it

Remove dependency on interfaces-config.service to speed up boot, because interfaces-config.service takes a lot of time on boot.

#### How I did it

Changed service files for swss, syncd.

#### How to verify it

Boot and check swss/syncd start time comparing to interfaces-config

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211

### DIFF
--- a/files/build_templates/per_namespace/swss.service.j2
+++ b/files/build_templates/per_namespace/swss.service.j2
@@ -11,7 +11,6 @@ Requires=opennsl-modules.service
 {% endif %}
 Requires=updategraph.service
 After=updategraph.service
-After=interfaces-config.service
 BindsTo=sonic.target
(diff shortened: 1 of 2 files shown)


lguohan:
in interfaces-config.sh file, there are some ZTP support which enable dhclient on front panel port as well. from current look, it seems it only try to generate correct /etc/network/interfaces file and restart networking service. i tend to agree that swss and syncd should not have dependency on interfaces-config.service.

lguohan (approved):

PR #13138, opened 2022-12-21
[azp] Add Semgrep - [actions] Remove Semgrep

maipbui:
#### Why I did it
Support Semgrep in Azure Pipelines instead Github Actions
Semgrep in Github Acions currently has a drawback which cannot customize rules (cannot exclude rule)
Semgrep in Azure Pipelines works as CLI, which can exclude rule using flag --exclude-rule. Semgrep CLI reference: https://semgrep.dev/docs/cli-reference/
#### How I did it
Remove Semgrep in actions
Add Semgrep in azp
#### How to verify it
azp build checker

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211


azure-pipelines.yml
```diff
@@ -44,6 +44,24 @@ variables:
   value: y
 
 stages:
+- stage: Analysis
```
qiluo-msft, line 47: Could you mark this stage not as a hard enforcement?

maipbui: I added continueOnError flag, which will continue to run even if Semgrep check is failed

qiluo-msft (approved):
LGTM

maipbui:
Could you help review this PR please? @xumia @liushilongbuaa

azure-pipelines.yml
```diff
@@ -44,6 +44,24 @@ variables:
...
+  jobs:
+  - job:
+    displayName: "Semgrep"
+    continueOnError: true
+    pool:
+      vmImage: ubuntu-latest
+    steps:
+    - ${{ if eq(variables['Build.Reason'], 'PullRequest') }}:
```
liushilongbuaa, line 55: I'm not sure if it will break CI triggerred job. Because when it is CI triggerred, it looks like no step.
It's better to add checkout step explicitly.

Issue #11352, opened 2022-07-06
[Radius] Fail to login to switch when Radius is configured "shell /usr/bin/sonic-launch-shell does not exist"


DavidZagury:
#### Steps to reproduce the issue:
1. Configure Radius Server
2. Configure aaa authentication on Sonic Switch
`config aaa authentication login radius local`

4. Config radius server on Sonic Switch
`config radius add {RadiusServerIP} -k {RadiusServerIP} `

5. Try to connect by ssh to Sonic Switch with user from Radius Database

#### Describe the results you received:
Authentication attempt is unsuccessful

Radius Server shows:
```
(0) Received Access-Request Id 70 from 10.210.25.145:53057 to 10.209.102.129:1812 length 100
(0)   User-Name = "david"
(0)   User-Password = "\010\n\r\177INCORRECT"
(0)   NAS-IP-Address = 127.0.0.1
(0)   NAS-Identifier = "r-lionfish-13"
(0)   NAS-Port = 98986
(0)   NAS-Port-Type = Virtual
(0)   Service-Type = Authenticate-Only
(0)   Calling-Station-Id = "10.209.102.129"
(0) # Executing section authorize from file /etc/raddb/sites-enabled/default
(0)   authorize {
(0)     policy filter_username {
(0)       if (&User-Name) {
(0)       if (&User-Name)  -> TRUE
(0)       if (&User-Name)  {
(0)         if (&User-Name =~ / /) {
(0)         if (&User-Name =~ / /)  -> FALSE
(0)         if (&User-Name =~ /@[^@]*@/ ) {
(0)         if (&User-Name =~ /@[^@]*@/ )  -> FALSE
(0)         if (&User-Name =~ /\.\./ ) {
(0)         if (&User-Name =~ /\.\./ )  -> FALSE
(0)         if ((&User-Name =~ /@/) && (&User-Name !~ /@(.+)\.(.+)$/))  {
(0)         if ((&User-Name =~ /@/) && (&User-Name !~ /@(.+)\.(.+)$/))   -> FALSE
(0)         if (&User-Name =~ /\.$/)  {
(0)         if (&User-Name =~ /\.$/)   -> FALSE
(0)         if (&User-Name =~ /@\./)  {
(0)         if (&User-Name =~ /@\./)   -> FALSE
(0)       } # if (&User-Name)  = notfound
(0)     } # policy filter_username = notfound
(0)     [preprocess] = ok
(0)     [chap] = noop
[...]

PR #13141, opened 2022-12-22
Login through RADIUS is not working due to invalid shell

saravanan-i-CLS:
#### Why I did it
RADIUS presumes that sonic launch shell (/usr/bin/sonic-launch-shell) is available by default to be authenticated users, which is not the case. This leads to failure in RADIUS authentication.

#### How I did it
Added valid bash shell  (/bin/bash) in RADIUS nss code

#### How to verify it
Verification of remote RADIUS user with valid credentials is authenticated without any failure - PASS

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [x] 202205
- [ ] 202211

Fixed RADIUS user authentication issue due to invalid shell (sonic-launch-shell)
Resolves

### DIFF
--- a/src/radius/nss/libnss-radius/nss_radius_common.c
+++ b/src/radius/nss/libnss-radius/nss_radius_common.c
@@ -159,11 +159,11 @@ static void init_rnm(RADIUS_NSS_CONF_B * conf) {
     rnm[0].gid   = 999;
     rnm[0].groups = "docker";
     rnm[0].gecos = "remote_user";
-    rnm[0].shell = "/usr/bin/sonic-launch-shell";
+    rnm[0].shell = "/bin/bash";
     rnm[RADIUS_MAX_MPL-1].gid   = 1000;
     rnm[RADIUS_MAX_MPL-1].groups = "admin,sudo,docker";
     rnm[RADIUS_MAX_MPL-1].gecos = "remote_user_su";
-    rnm[RADIUS_MAX_MPL-1].shell = "/usr/bin/sonic-launch-shell";
+    rnm[RADIUS_MAX_MPL-1].shell = "/bin/bash";
 
 }
 
--- a/src/sonic-host-services-data/templates/radius_nss.conf.j2
+++ b/src/sonic-host-services-data/templates/radius_nss.conf.j2
@@ -13,13 +13,13 @@ debug=on
 #
 # User Privilege:
 # Default:
-# user_priv=15;pw_info=remote_user_su;gid=1000;group=sudo,docker;shell=/usr/bin/sonic-launch-shell
-# user_priv=1;pw_info=remote_user;gid=999;group=docker;shell=/usr/bin/sonic-launch-shell
+# user_priv=15;pw_info=remote_user_su;gid=1000;group=sudo,docker;shell=/bin/bash
+# user_priv=1;pw_info=remote_user;gid=999;group=docker;shell=/bin/bash
 
 # Eg:
-# user_priv=15;pw_info=remote_user_su;gid=1000;group=sudo,docker;shell=/usr/bin/sonic-launch-shell
-# user_priv=7;pw_info=netops;gid=999;group=docker;shell=/usr/bin/sonic-launch-shell
-# user_priv=1;pw_info=operator;gid=100;group=docker;shell=/usr/bin/sonic-launch-shell
+# user_priv=15;pw_info=remote_user_su;gid=1000;group=sudo,docker;shell=/bin/bash
+# user_priv=7;pw_info=netops;gid=999;group=docker;shell=/bin/bash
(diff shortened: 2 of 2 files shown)


DavidZagury:
Looks good to me. However,  in the triage meeting where we first discussed this issue, we asked BRCM to check the reason it was first built with that shell.
I don't think we got a answer about that.

@zhangyanzhao did BRCM check that? can we conclude that the shell really is wrong and we can change it? if so, this PR should fix the issue.

dharmaraj-gurusamy (approved):
Looks good to me. @lguohan / @zhangyanzhao - Please review.

DavidZagury (approved):

keboliu:
@qiluo-msft  would you please help to review and approve?

keboliu:
> THIS PR also claims fix #11352
> 
> #14466
> 
> can we get clarification? which pr fix? which one to go?

@saravanan-i  would you please clarify?

saravanan-i:
> > THIS PR also claims fix #11352
> > #14466
> > can we get clarification? which pr fix? which one to go?
> 
> @saravanan-i would you please clarify?

This PR has corrected the invalid soft link ("sonic-launch-shell") to valid RADIUS login launch shell("/bin/bash"). Other authentication protocols like TACACS use this shell ("/bin/bash") in SoNIC.

zhangyanzhao:
@adyeung can you please help to review this PR? Thanks.

src/radius/nss/libnss-radius/nss_radius_common.c
```diff
@@ -159,11 +159,11 @@ static void init_rnm(RADIUS_NSS_CONF_B * conf) {
     rnm[0].gid   = 999;
     rnm[0].groups = "docker";
```
Yarden-Z, line 160: To reduce the attack surface, maybe change the default group given here.
MPL = 0 user does not require docker, but maybe it does require the redis group or adm group.

qiluo-msft, line 160: The readonly user should be in `users` group, and no `docker` group membership.

saravanan-i: Hi @qiluo-msft. User groups were defined based on existing RADIUS design document of SoNiC [RADIUS_Design](https://github.com/sonic-net/SONiC/blob/master/doc/aaa/radius_authentication.md). I haven't introduced them as part of this pull request. Can you please confirm if we have to modify the same ?

Yarden-Z: I understand that these values were introduced in the HLD design.
But given this item - I think we can re-visit this.
From my perspective, a non-privileged user requires redis (potentially) and the default adm group.
If I understood your question

dt-nexthop: @saravanan-i I think we should remove this docker group because the sudoer file already allows for the read-only docker commands. Adding docker group allows this user to potentially do non-read_only things.   I can push that change if you want to give me access to this branch so we can keep all the history of this PR without having to make a new one.

CovetingEpiphany2152: @dt-nexthop , removing the group caused another regression.

- [user_add()](https://github.com/sonic-net/sonic-buildimage/blob/bdd8010639dd5c8bb2453d94caac0e4fb0b4b542/src/radius/nss/libnss-radius/nss_radius_common.c#L170) always constructs execl(..., "-G", sec_grp, ...).
- In execl, NULL terminates argv, yielding a failed call:
  ```text
    useradd -U -G
  ```
- so if the user does not provide their own `/etc/radius_nss.conf`, user_add() will always fail to create an unconfirmed user

```text
Jul 14 01:27:17 sonic sshd-session[54953]: Connection from 10.110.10.86 port 65169 on 10.110.10.57 port 22 rdomain ""
Jul 14 01:27:17 sonic sshd-session[54953]: nss: "/var/cache/radius/user/r-viewer/Management-Privilege-Level": Absent.
Jul 14 01:27:17 sonic sshd-session[54953]: nss: sshd-session: [accepted]: Lookup r-viewer
Jul 14 01:27:17 sonic sshd-session[54953]: nss: Creating user "r-viewer"
Jul 14 01:27:17 sonic sshd-session[54953]: nss: /usr/sbin/useradd r-viewer failed
```

I believe @qiluo-msft was mentioning that the secondary groups should be `users` instead. But if we're using GID 999, shouldn't it be `systemd-journal` instead? Whichever it turns out to be, the [HLD](https://github.com/sonic-net/SONiC/blob/master/doc/aaa/radius_authentication.md#Mapping-Users) should be updated as well.

There is a PR attempting to fix this https://github.com/sonic-net/sonic-buildimage/pull/28192

PR #13142, opened 2022-12-22
Client Auth is not working in telemetry (dial-in mode)

saravanan-i:
#### Why I did it
Client Auth option is not working in telemetry (dial-in mode). Client Authentication value read from redis db is not passed to telemetry script in docker startup.

#### How I did it
Included client_auth option in telemetry arguments 

#### How to verify it
Verification of gnmi_get by passing certificate to the server - PASS

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [x] 202205
- [ ] 202211

When an valid client_auth is present in config db (json file), same is passed as an argument to dial-in server startup script

### DIFF
--- a/dockers/docker-sonic-telemetry/telemetry.sh
+++ b/dockers/docker-sonic-telemetry/telemetry.sh
@@ -60,6 +60,8 @@ TELEMETRY_ARGS+=" --port $PORT"
 CLIENT_AUTH=$(echo $GNMI | jq -r '.client_auth')
 if [ -z $CLIENT_AUTH ] || [ $CLIENT_AUTH == "false" ]; then
     TELEMETRY_ARGS+=" --allow_no_client_auth"
+else
+    TELEMETRY_ARGS+=" --client_auth $CLIENT_AUTH"
 fi
 
 LOG_LEVEL=$(echo $GNMI | jq -r '.log_level')


qiluo-msft:
Thanks for the contribution!
Could you provide more details of the verification you have done for original code and the new code in "How to verify it" in PR description?

dockers/docker-sonic-telemetry/telemetry.sh
```diff
@@ -60,6 +60,8 @@ TELEMETRY_ARGS+=" --port $PORT"
 CLIENT_AUTH=$(echo $GNMI | jq -r '.client_auth')
 if [ -z $CLIENT_AUTH ] || [ $CLIENT_AUTH == "false" ]; then
     TELEMETRY_ARGS+=" --allow_no_client_auth"
+else
+    TELEMETRY_ARGS+=" --client_auth $CLIENT_AUTH"
```
ganglyu, line 64: https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-yang-models/yang-models/sonic-telemetry.yang#L57
client_auth is a flag for requiring client auth, but you are using it as client auth type？

saravanan-i: Closing this PR. And will open a new PR addressing the comments

qiluo-msft: If you want to address PR comments, better to keep this PR open, and push new commit in the same PR. So the comment will have enough context and easy to track.

saravanan-i: Fine @qiluo-msft

saravanan-i: Client Authentication Argument in Telemetry dialin server (https://github.com/sonic-net/sonic-gnmi/blob/master/telemetry/telemetry.go) accepts jwt,password,certification. Made changes in telemetry.sh script to accept the different client authentication modes.

dharmaraj-gurusamy (approved):

PR #13150, opened 2022-12-23
[build] Add environment when using dpkg hook to avoid lock loop.

liushilongbuaa:
#### Why I did it
In some cases, dpkg will call dpkg to validate version.
dpkg hook will get stuck in a loop to lock.
#### How I did it
Use an env variable to skip duplicated lock.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012
- [ ] 202106
- [ ] 202111
- [x] 202205
- [x] 202211


saiarcot895 (approved):

src/sonic-build-hooks/hooks/dpkg
```diff
@@ -3,7 +3,10 @@
 . /usr/local/share/buildinfo/scripts/buildinfo_base.sh
 REAL_COMMAND=$(get_command dpkg)
 COMMAND_INFO="Locked by command: $REAL_COMMAND $@"
-lock_result=$(acquire_apt_installation_lock "$COMMAND_INFO" )
+if ! printenv LOCKED;then
```
saiarcot895, line 6: Would it be better/more robust to check here that `LOCKED` actually contains `y`?

liushilongbuaa: Added.

src/sonic-build-hooks/hooks/dpkg
```diff
@@ -3,7 +3,10 @@
 . /usr/local/share/buildinfo/scripts/buildinfo_base.sh
 REAL_COMMAND=$(get_command dpkg)
 COMMAND_INFO="Locked by command: $REAL_COMMAND $@"
-lock_result=$(acquire_apt_installation_lock "$COMMAND_INFO" )
+if ! printenv LOCKED > /dev/null || [[ "$(printenv LOCKED)" != "y" ]];then
```
kv-y, line 6:
```suggestion
if [ "$LOCKED" != "y" ]; then
```

kv-y: And maybe rename to something like DPKG_HOOK_LOCKED instead of LOCKED to exclude possible conflicts.

build_debian.sh
```diff
@@ -123,6 +123,8 @@ echo '[INFO] Install and setup eatmydata'
 sudo LANG=C chroot $FILESYSTEM_ROOT apt-get -y install eatmydata
 sudo LANG=C chroot $FILESYSTEM_ROOT ln -s /usr/bin/eatmydata /usr/local/bin/dpkg
 echo 'Dir::Bin::dpkg "/usr/local/bin/dpkg";' | sudo tee $FILESYSTEM_ROOT/etc/apt/apt.conf.d/00image-install-eatmydata > /dev/null
+## Note: dpkg hook conflict with eatmydata
+sudo LANG=C chroot $FILESYSTEM_ROOT rm /usr/local/sbin/dpkg -rf
```
xumia, line 127: Why conflict? Can we fix in dpkg command?

liushilongbuaa: sudo LANG=C chroot $FILESYSTEM_ROOT ln -s /usr/bin/eatmydata /usr/local/bin/dpkg
In line 124.
eatmydata will call dpkg. Then it will fall in a loop.
build_debian runs in one thread. It doesn't need lock.

saiarcot895 (approved):

liushilongbuaa:
It will break 202211 branch's build. Remove Request for 202211 Branch.

PR #13153, opened 2022-12-23
[install.sh] install different platform image should be blocked

inspurSDN:
#### How I did it
1. Check platform and asic before install the image, if the platform and asic version is different with running image, the install action should be blocked.

2. This commit can block wrong installation in ONIE and CLI env.

#### Why I did it
-  Install different platform-asic image will make kernel can't be reload correctly.

Steps to reproduce Bug

1. Install x86 image in CLI
2. Install arm64 image in CLI
3. Install x86 image in CLI
4. reboot
5. the kernel can't be loaded, user need to insatll image by ONIE to recover it

#### How to verify it
EX: Install build arm64 image in x86 platform, the install operation will execute fail.

- [O] 201811
- [O] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211

### DIFF
--- a/installer/install.sh
+++ b/installer/install.sh
@@ -83,15 +83,33 @@ fi
 # get running machine from conf file
 if [ -r /etc/machine.conf ]; then
     read_conf_file "/etc/machine.conf"
+    machine_conf_path=/etc/machine.conf
 elif [ -r /host/machine.conf ]; then
     read_conf_file "/host/machine.conf"
+    machine_conf_path=/host/machine.conf
 elif [ "$install_env" != "build" ]; then
     echo "cannot find machine.conf"
     exit 1
 fi
 
+onie_running_arch=`cat $machine_conf_path | grep onie_arch | sed -n "s/^onie_arch=\(.*\)$/\1/p"`
+onie_running_switch_asic=`cat $machine_conf_path | grep onie_switch_asic | sed -n "s/^onie_switch_asic=\(.*\)$/\1/p"`
+
+# bcm aisc is broadcom
+if [ "$onie_running_switch_asic" = "bcm" ]; then
+    onie_running_switch_asic=broadcom
+fi
+
+platform_prefix=$(echo $platform | cut -d"-" -f 1)-$(echo $platform | cut -d"-" -f 2)
+onie_running_platform=$onie_running_arch-$onie_running_switch_asic
 
 echo "onie_platform: $onie_platform"
+echo "onie_running_platform: $onie_running_platform"
+
+if [ "$onie_running_platform" != "$platform_prefix" ]; then
+    echo "Error: Image file is of a different platform than running image."
+    exit 1
+fi
 
 # Get platform specific linux kernel command line arguments
 ONIE_PLATFORM_EXTRA_CMDLINE_LINUX=""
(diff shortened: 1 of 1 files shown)


lguohan:
@qiluo-msft , can you help and assign someone review this?

installer/install.sh
```diff
@@ -83,15 +83,33 @@ fi
...
+    machine_conf_path=/host/machine.conf
 elif [ "$install_env" != "build" ]; then
     echo "cannot find machine.conf"
     exit 1
 fi
 
+onie_running_arch=`cat $machine_conf_path | grep onie_arch | sed -n "s/^onie_arch=\(.*\)$/\1/p"`
+onie_running_switch_asic=`cat $machine_conf_path | grep onie_switch_asic | sed -n "s/^onie_switch_asic=\(.*\)$/\1/p"`
+
+# bcm aisc is broadcom
```
qiluo-msft, line 98: >aisc [](http://example.com/codeflow?start=6&length=4)

typo?

qiluo-msft:
There is a similar feature https://github.com/sonic-net/sonic-buildimage/pull/8814. Could you check if this is duplicated or partially overlapped?

installer/install.sh
```diff
@@ -83,15 +83,33 @@ fi
...
+    machine_conf_path=/etc/machine.conf
 elif [ -r /host/machine.conf ]; then
     read_conf_file "/host/machine.conf"
+    machine_conf_path=/host/machine.conf
 elif [ "$install_env" != "build" ]; then
     echo "cannot find machine.conf"
     exit 1
 fi
 
+onie_running_arch=`cat $machine_conf_path | grep onie_arch | sed -n "s/^onie_arch=\(.*\)$/\1/p"`
```
qiluo-msft, line 95: >onie_running_arch [](http://example.com/codeflow?start=0&length=17)

`read_conf_file` function is a general parsing function for `etc/machine.conf`. So you do not need to parse these variables manually.

installer/install.sh
```diff
@@ -83,15 +83,33 @@ fi
...
+onie_running_switch_asic=`cat $machine_conf_path | grep onie_switch_asic | sed -n "s/^onie_switch_asic=\(.*\)$/\1/p"`
```
qiluo-msft, line 96: >onie_running_switch_asic [](http://example.com/codeflow?start=0&length=24)

I am not sure if the code is working for old version of ONIE. Suggest you test the scenarios that these variables are not existing.

installer/install.sh
```diff
@@ -83,15 +83,33 @@ fi
...
+
+# bcm aisc is broadcom
+if [ "$onie_running_switch_asic" = "bcm" ]; then
```
wen587, line 99: for kvm image, its onie_switch_asic is 'qemu'.

wen587:
I prefer to keep the original design.
We have some internal usages. We want to avoid the installation of the public image on internal usage though they are sharing the same ASIC.

qiluo-msft:
@inspurSDN I see you closed this issue. Is your original issue still open? You may create a Github issue.

Issue #12047, opened 2022-09-12
[multiasic][supervisor] sonic-db-cli crashes at boot up when execute sonic-db-cli PING command in database.sh on multiasic platform


mlok-nokia:
#### Description
On supervisor card,  sonic-db-cli crashes when executes the sonic-db-cli PING command in the database.sh.   The new implementation of the sonci-db-cli with PING  command calls initializeGlobalConfig() which will check all ASICs redis#/sonic-db/database_config.json files which are not ready yet. This cause crash and the following error log.   This function was used to wait for all database ready.  If sonic-db-cli tries to access redis#/sonic-db/database_config.json files, it will failed.
```
Sep  9 23:21:15 sonic sonic-db-cli: :- parseDatabaseConfig: Sonic database config file doesn't exist at /var/run/redis/sonic-db/../../redis0/sonic-db/database_config.json
Sep  9 23:21:15 sonic database.sh[4739]: terminate called after throwing an instance of 'std::runtime_error'
Sep  9 23:21:15 sonic database.sh[4739]:   what():  Sonic database config file syntax error >> Sonic database config file doesn't exist at /var/run/redis/sonic-db/../../redis0/sonic-db/database_config.json
Sep  9 23:21:15 sonic sonic-db-cli: :- initializeGlobalConfig: Sonic database config file syntax error >> Sonic database config file doesn't exist at /var/run/redis/sonic-db/../../redis0/sonic-db/database_config.json
 ```  
There are 16 ASICs on this supervisor cards.  This issue is similar to the isisue  https://github.com/sonic-net/sonic-buildimage/issues/10105.   If sonic-db-cli behavior is changed, we may need to change waitForAllInstanceDatabaseConfigJsonFilesReady 

#### Steps to reproduce the issue:
1. Reboot the the syatem with the new image.
2.
3.

#### Describe the results you received:
There are core files.  and the following error logs
```
Sep  9 23:21:15 sonic sonic-db-cli: :- parseDatabaseConfig: Sonic database config file doesn't exist at /var/run/redis/sonic-db/../../redis0/sonic-db/database_config.json
Sep  9 23:21:15 sonic database.sh[4739]: terminate called after throwing an instance of 'std::runtime_error'
Sep  9 23:21:15 sonic database.sh[4739]:   what():  Sonic database config file syntax error >> Sonic database config file doesn't exist at /var/run/redis/sonic-db/../../redis0/sonic-db/database_config.json
[...]

PR #13207, opened 2022-12-30
[Chassis][multiasic] Fix the sonic-db-cli core files issue on multiasic platform after the c++ implementation of sonic-db-cli

mlok-nokia:
#### Why I did it
Fixe #12047.  After the c++ implementation of the sonic-db-cli,  sonic-db-cli PING command tries to initialize the global database for all instances database starting.  If all instance database-config.json are not ready yet. it will crash and generate core file.  PR https://github.com/sonic-net/sonic-swss-common/pull/701 only fix the crash and the process abortion.  With this PR, many core files are still generated: 
The following error are logged in the syslog
```
Dec 28 16:44:34.811893 ixre-cpm-chassis8 ERR sonic-db-cli: :- parseDatabaseConfig: Sonic database config file doesn't exist at /var/run/redis/sonic-db/../../redis0/sonic-db/database_config.json
Dec 28 16:44:34.812099 ixre-cpm-chassis8 ERR sonic-db-cli: :- initializeGlobalConfig: Sonic database config file syntax error >> Sonic database config file doesn't exist at /var/run/redis/sonic-db/../../redis0/sonic-db/database_config.json
Dec 28 16:44:34.812162 ixre-cpm-chassis8 INFO database.sh[5764]: terminate called after throwing an instance of 'std::runtime_error'
Dec 28 16:44:34.812215 ixre-cpm-chassis8 INFO database.sh[5764]:   what():  Sonic database config file syntax error >> Sonic database config file doesn't exist at /var/run/redis/sonic-db/../../redis0/sonic-db/database_config.json
```
The following core files are generated
```
admin@supervisor:~$ ls /var/core -al
total 376
drwxr-xr-x 1 root root 4096 Nov 22 22:00 .
drwxr-xr-x 1 root root 4096 Nov 22 20:50 ..
-rw-r--r-- 1 root root 88525 Nov 22 21:42 sonic-db-cli.1669153338.6192.core.gz
-rw-r--r-- 1 root root 93392 Nov 22 21:42 sonic-db-cli.1669153339.6757.core.gz
-rw-r--r-- 1 root root 93413 Nov 22 21:42 sonic-db-cli.1669153339.6886.core.gz
-rw-r--r-- 1 root root 93284 Nov 22 21:42 sonic-db-cli.1669153339.7072.core.gz
```
#### How I did it
In the database.sh, call function  waitForAllInstanceDatabaseConfigJsonFilesReady  to make sure all instance database-config.json are present before the command sonic-db-cli PING.  This will avoid the database-config,json doesn't exist and cause errors and generate the core file.  

#### How to verify it
With this change, after the image install and boot up, check and verify the syslog, the following lines should not be in the syslog and this related core file should not be seen.
'''
Dec 28 16:44:34.811893 ixre-cpm-chassis8 ERR sonic-db-cli: :- parseDatabaseConfig: Sonic database config file doesn't exist at /var/run/redis/sonic-db/../../redis0/sonic-db/database_config.json
Dec 28 16:44:34.812099 ixre-cpm-chassis8 ERR sonic-db-cli: :- initializeGlobalConfig: Sonic database config file syntax error >> Sonic database config file doesn't exist at /var/run/redis/sonic-db/../../redis0/sonic-db/database_config.json
Dec 28 16:44:34.812162 ixre-cpm-chassis8 INFO database.sh[5764]: terminate called after throwing an instance of 'std::runtime_error'
Dec 28 16:44:34.812215 ixre-cpm-chassis8 INFO database.sh[5764]:   what():  Sonic database config file syntax error >> Sonic database config file doesn't exist at /var/run/redis/sonic-db/../../redis0/sonic-db/database_config.json
'''

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [x] 202205
- [x] 202211

### DIFF
--- a/files/build_templates/docker_image_ctl.j2
+++ b/files/build_templates/docker_image_ctl.j2
@@ -155,11 +155,13 @@ function waitForAllInstanceDatabaseConfigJsonFilesReady()
                     fi
 		done
             done
-	fi
+        fi
+        # Delay a second to allow all instance database_config.json files to be completely generated and fully accessible.
+        # This delay is needed to make sure that the database_config.json files are correctly rendered from j2 template 
+        # files ( renderning takes some time )
+        sleep 1 
     fi
 }
-# delay a second to allow the file to be fully accessible
-sleep 1
 {%- endif %}
 
 function postStartAction()
@@ -225,6 +227,10 @@ function postStartAction()
         # then we catch python exception of file not valid
         # that comes to syslog which is unwanted so wait till database
         # config is ready and then ping
+	# sonic-db-cli try to initialize the global database. If in multiasic platform, inital global
+        # database will try to access to all other instance database-config.json. If other instance
+        # database-config.json files are not ready yet, it will generate the sonic-db-cli core files.
+	waitForAllInstanceDatabaseConfigJsonFilesReady
         until [[ ($(docker exec -i database$DEV pgrep -x -c supervisord) -gt 0) && ($($SONIC_DB_CLI PING | grep -c PONG) -gt 0) &&
                  ($(docker exec -i database$DEV sonic-db-cli PING | grep -c PONG) -gt 0) ]]; do
           sleep 1;
@@ -234,11 +240,7 @@ function postStartAction()
             mv $WARM_DIR/dump.rdb $WARM_DIR/dump.rdb.old
         else
             # If there is a config_db.json dump file, load it.
-            if [ -r /etc/sonic/config_db$DEV.json ]; then
-
-		# For multi-asic, all /var/run/redis$DEV/sonic-db/database_config.json need to ready
-                # for loading config with --write-to-db
-		waitForAllInstanceDatabaseConfigJsonFilesReady
+            if [ -r /etc/sonic/config_db$DEV.json ]; then		
 
                 if [ -r /etc/sonic/init_cfg.json ]; then
                     $SONIC_CFGGEN -j /etc/sonic/init_cfg.json -j /etc/sonic/config_db$DEV.json --write-to-db


mlok-nokia:
@rlhui @judyjoseph @abdosi  After 202205 branch uses the c++ implementation of the sonic-db-cli, even with PR  https://github.com/sonic-net/sonic-swss-common/pull/701,  database.sh still generates many core files.  This PR will address this core files issue.  Please review it.

files/build_templates/docker_image_ctl.j2
```diff
@@ -221,6 +221,10 @@ function postStartAction()
         # then we catch python exception of file not valid
         # that comes to syslog which is unwanted so wait till database
         # config is ready and then ping
+	# sonic-db-cli try to initialize the global database. If in multiasic platform, inital global
+        # database will try to access to all other instance database-config.json. If other instance
+        # database-config.json files are not ready yet, it will generate the sonic-db-cli core files.
+	waitForAllInstanceDatabaseConfigJsonFilesReady
```
liuh-80, line 227: The crash is not cause by a behavior change, python version sonic-db-cli also crash in this case, access name space DB need name space config file ready.

liuh-80 (approved):

qiluo-msft (approved):

files/build_templates/docker_image_ctl.j2
```diff
@@ -151,11 +151,11 @@ function waitForAllInstanceDatabaseConfigJsonFilesReady()
                     fi
 		done
             done
-	fi
```
judyjoseph, line 154: Since we are moving the waitForAllInstanceDatabaseConfigJsonFilesReady call before the check 
```
until [[ ($(docker exec -i database$DEV pgrep -x -c supervisord) -gt 0) && ($($SONIC_DB_CLI PING | grep -c PONG) -gt 0) &&
```
can we change the if [ ! -z "$DEV" ]; then in **line 133** to if [ -z "$DEV" ]; then ?

Also do we really need to add sleep 1 delay here

judyjoseph: @mlok-nokia could you check this again

mlok-nokia: > Since we are moving the waitForAllInstanceDatabaseConfigJsonFilesReady call before the check
> 
> ```
> until [[ ($(docker exec -i database$DEV pgrep -x -c supervisord) -gt 0) && ($($SONIC_DB_CLI PING | grep -c PONG) -gt 0) &&
> ```
> 
> can we change the if [ ! -z "$DEV" ]; then in **line 133** to if [ -z "$DEV" ]; then ?
> 
> Also do we really need to add sleep 1 delay here

We cannot change the if { ! -z "$DEV" ] to if [ -z "$DEV" ].   The negative condition check is the main key for making sure the instance database# start need to wait for the Redis reday.   Although  instance database# depends on the host database. But when database is started, the instance database# will be started right after it.   This is the reason which we need to use this waitForAllInstanceDatabaseConfigJsonFilesReady  function,    

I think we can remove 1 second delay. which is at the end.  Let me verify it and I will update the PR

files/build_templates/docker_image_ctl.j2
```diff
@@ -151,11 +151,11 @@ function waitForAllInstanceDatabaseConfigJsonFilesReady()
...
+        fi
+        # delay a second to allow the file to be fully accessible
```
judyjoseph, line 155: @mlok-nokia could you add a comment here as discussed on to why we need this delay. This delay is needed to make sure that the database_config.json files are correctly rendered from j2 template files ( renderning takes some time )

mlok-nokia: > @mlok-nokia could you add a comment here as discussed on to why we need this delay. This delay is needed to make sure that the database_config.json files are correctly rendered from j2 template files ( renderning takes some time )

@judyjoseph  I have updated it with comment which you mentioned above.

judyjoseph (approved):

mlok-nokia:
This PR needs to be marked "request for 202205"

stepanblyschak:
@mlok-nokia @judyjoseph Please add request for 202211

PR #13224, opened 2023-01-01
remove /var/lib/apt/lists/* to reduce docker image size

kv-y:
test


kv-y:
Some cleanupt was implemented in src/sonic-build-hooks/scripts/post_run_cleanup one month ago.
For some reason I missed it.

TODO:
If we do cleanup in the separate script then probably we can remove some actions from dockerfiles.
I mean `apt-get clean, autoclean and autoremove`, `rm -rf /debs ~/.cache`, etc.

kv-y:
@Kalimuthu-Velappan 

In #12005 you added apt-get commands with dry-run flag:
https://github.com/sonic-net/sonic-buildimage/blob/aeaec4053ce01c3a5064fd1ef60200789bfa6bda/src/sonic-build-hooks/scripts/post_run_cleanup#L18-L20

Was it on purpose or maybe it's just a typo?
If no drawbacks were observed here then probably we can disable a simulation flag and move cleanup code from Dockerfiles to this script. Now it's duplicated in every Dockerfile like this:
https://github.com/sonic-net/sonic-buildimage/blob/aeaec4053ce01c3a5064fd1ef60200789bfa6bda/dockers/docker-lldp/Dockerfile.j2#L32-L38

kv-y:
It seems we still need to do apt cleanup before `post_run_buildiinfo`.
Otherwise we will collect info about additional deb packages that removed via apt-get autoremove.

PR #13246, opened 2023-01-04
[Mellanox] Skip the leftover hardware reboot cause in case of last boot is warm/fast reboot

keboliu:
#### Why I did it
In case of warm/fast reboot, the hardware reboot cause will NOT be cleared because CPLD will not be touched in this flow. To not confuse the reboot cause determine logic, the leftover hardware reboot cause shall be skipped by the platform API, platform API will return the 'REBOOT_CAUSE_NON_HARDWARE'  instead of the "hardware" reboot cause.

#### How I did it
Check the proc cmdline to see whether the last reboot is a warm or fast reboot, if yes skip checking the leftover hardware reboot cause.

#### How to verify it
a. Manual test:
1. Perform a power loss
2. Perform a warm/fast reboot
3. check the reboot cause should be "warm-reboot" or "fast-reboot" instead of "power loss" 

b. Run reboot cause related regression test.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211

### DIFF
--- a/platform/mellanox/mlnx-platform-api/sonic_platform/chassis.py
+++ b/platform/mellanox/mlnx-platform-api/sonic_platform/chassis.py
@@ -30,6 +30,7 @@ try:
     from .utils import extract_RJ45_ports_index
     from . import utils
     from .device_data import DeviceDataManager
+    import re
 except ImportError as e:
     raise ImportError (str(e) + "- required module not found")
 
@@ -740,7 +741,7 @@ class Chassis(ChassisBase):
         self.reboot_by_software = 'reset_sw_reset'
         self.reboot_cause_initialized = True
 
-    def _parse_warmfast_reboot_from_proc_cmdline():
+    def _parse_warmfast_reboot_from_proc_cmdline(self):
         if os.path.isfile(REBOOT_TYPE_KEXEC_FILE):
             with open(REBOOT_TYPE_KEXEC_FILE) as cause_file:
                 cause_file_kexec = cause_file.readline()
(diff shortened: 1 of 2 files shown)


prgeor:
@keboliu please captue the output of "show reboot-cause" and "show reboot-cause history" while doing this

1. Do warm-reboot
2. disconnect the power while system is coming up

stephenxs (approved):

platform/mellanox/mlnx-platform-api/sonic_platform/chassis.py
```diff
@@ -748,6 +765,14 @@ def get_reboot_cause(self):
             to pass a description of the reboot cause.
         """
         #read reboot causes files in the following order
+
+        # To avoid the leftover hardware reboot cause confusing the reboot cause determine service
+        # Skip the hardware reboot cause check if warm/fast reboot cause found from cmdline
+        if utils.is_host():
+            reboot_cause = self._parse_warmfast_reboot_from_proc_cmdline()
```
prgeor, line 772: reboot_cause will be None when system is booting after user issued "sudo reboot". ?

keboliu: New function _parse_warmfast_reboot_from_proc_cmdline() is designed to check the /proc/cmdline to see whether the last boot is a warm/fast reboot,  if Not, then the local variable 'reboot_cause' will be 'None', but this 'reboot_cause' is not the final reboot cause which will be returned by this platform API in this case, there will be further logic to determine the reboot cause from the sysfs if you check the reset of the code.

prgeor (approved):

prgeor:
@lguohan could you help merge this change?

liat-grozovik (approved):

PR #13277, opened 2023-01-05
[Arista] Rely on automatic flash size detection for Raven

grahamhayes:
Many of these switches have had flash upgraded beyond 2G however, in boot0 both were assigned 2GB for legacy reasons.

Remove the hardcoding of the flash size and let boot0 autodetect the available space.

Signed-off-by: Graham Hayes <gr@ham.ie>

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211

### DIFF
--- a/files/Aboot/boot0.j2
+++ b/files/Aboot/boot0.j2
@@ -483,7 +483,6 @@ write_platform_specific_cmdline() {
     if [ "$platform" = "raven" ]; then
         # Assuming sid=Cloverdale
         aboot_machine=arista_7050_qx32
-        flash_size=2000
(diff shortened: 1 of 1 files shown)


lguohan:
@Staphylo , can you review? why we have this flash_size hardcoded in the first place?

Staphylo (approved):

Staphylo:
@lguohan this change lgtm.
I don't recall why we initially hardcoded the flash_size as an attribute for each product instead of just relying on automatic detection.
On some products we do use this mechanism to ensure consistent behavior across all the switches which might have shipped with different flash sizes.

PR #13283, opened 2023-01-06
During build time mask only those feature/services that are disabled excplicitly

abdosi:
What I did:
Fix :  https://github.com/sonic-net/sonic-buildimage/issues/13117

How I did:

Build time behavior ::-

1. init_cfg.json.j2 during build time make state as disabled then mask the service 
2.  init_cfg.json.j2 during build time make state as another jinja2 template render string than do no mask the service if the service is one of `database,bgp,swss,sycnd,teamd` else mask the service
3. init_cfg.json.j2 during build time make state as enabled then do not mask the service

How I verify:
Manual Verification.


yxieca (approved):

stepanblyschak:
Request for 202211

PR #13294, opened 2023-01-06
Upgrade docker-sonic-vs and docker-syncd-vs to Bullseye

### LINKED ISSUE #11706: docker-sonic-vs.gz on master branch is on buster
#### Description

docker-sonic-vs.gz is buster on the master branch build

Build sonic-buildimage from master branch

cannot test swss changes with pytest because swss master is on bullseye and docker-sonic-vs.gz provides a buster image.

docker-sonic-vs.gz should be on bullseye\

saiarcot895:

#### Why I did it

Fixes #11706

#### How I did it

Upgrade docker-sonic-vs, docker-syncd-vs, and iproute2 to Bullseye/Bullseye versions.

Submodules that use docker-sonic-vs for their CI will need to be updated to pull/install Bullseye packages instead of Buster packages.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211


platform/vs/docker-sonic-vs.mk
```diff
@@ -2,9 +2,7 @@
 
 DOCKER_SONIC_VS = docker-sonic-vs.gz
 $(DOCKER_SONIC_VS)_PATH = $(PLATFORM_PATH)/docker-sonic-vs
-$(DOCKER_SONIC_VS)_DEPENDS += $(SWSS) \
```
mestery, line 5: Curious why `$(SWSS)` is removed here? Is it because it's now taken care of in the addition on line 46?

mestery (approved):
Not tested, but it looks good to me, thanks!

platform/vs/docker-sonic-vs.mk
saiarcot895: Yes, `docker-sonic-vs` will be based on top of `docker-swss-layer-bullseye`, which will contain SWSS.

The initial intention of that layer was to save space by having SWSS be in a single layer and then other containers can be built on top of this layer. In this case, since `docker-sonic-vs` won't actually be part of the SONiC image, there's effectively no impact.

yxieca (approved):

platform/vs/docker-sonic-vs/Dockerfile.j2
```diff
@@ -24,10 +24,6 @@ RUN apt-get install -y net-tools \
                        python-ply \
                        libqt5core5a \
                        libqt5network5 \
-                       libboost-program-options1.71.0 \
-                       libboost-serialization1.71.0 \
```
liushilongbuaa, line 28: Hi @saiarcot895 , why removed libboost package?

liushilongbuaa: https://github.com/sonic-net/sonic-swss-common/blob/e3d9024602847d5bb41e14f153099431ef35f291/debian/control#L5
swss-common still depends on libboost1.71.

saiarcot895: Boost 1.71 is not available in Bullseye, Bullseye has Boost 1.74. Because of this (and other differences), the pipelines for swss-common, swss, and sairedis were bound to fail. In addition, the base layer of this container has changed; it's now based on swss-layer-bullseye (which will contain both swss and swsscommon).

As for the pipeline breakages, I've already opened sonic-net/sonic-swss-common#716 to fix it for swss-common. Once that is fixed, I should be able to fix it for sairedis (sonic-net/sonic-sairedis#1225) and swss (PR to be opened).

PR #13310, opened 2023-01-09
[202205] sonic-host-services: Pin deepdiff to version 6.2.2

saiarcot895:
The deepdiff python package was recently updated to 6.2.3. As part of this, a dependency was introduced on orjson. There's no armv8l python wheel available for orjson, which means it needs to be built from source. However, building it requires rust (which Buster and Bullseye don't have a new enough version of) and maturin.

As a quick fix, pin this to version 6.2.2, before the orjson dependency is introduced.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211

### DIFF
--- a/src/sonic-host-services/setup.py
+++ b/src/sonic-host-services/setup.py
@@ -38,7 +38,7 @@ setup(
         'pytest',
         'pyfakefs',
         'sonic-py-common',
-        'deepdiff'
+        'deepdiff==6.2.2'
     ],
     classifiers = [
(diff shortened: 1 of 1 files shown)


qiluo-msft:
On 202012 branch, we have another fix merged https://github.com/sonic-net/sonic-buildimage/pull/13329, so no need to backport to 202012.

PR #13326, opened 2023-01-11
[yang] Change swss-event, dhcp-relay-event leafref to string

zbud-msft:
#### Why I did it

Do not require leafref as part of yang. Only need string to compare whether string received from event matches what is possible for ifname.

#### How to verify it

Run UT

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211


src/sonic-yang-models/yang-models/sonic-events-dhcp-relay.yang
```diff
@@ -40,16 +40,11 @@ module sonic-events-dhcp-relay {
...
             leaf ifname {
-                type union {
-                    type leafref {
-                        path "/port:sonic-port/port:PORT/port:PORT_LIST/port:name"; 
-                    }
-                    type leafref {
-                        path "/lag:sonic-portchannel/lag:PORTCHANNEL/lag:PORTCHANNEL_LIST/lag:name";
-                    }
+                type string {
+                    pattern 'Ethernet[0-9]{1,2}';
```
qiluo-msft, line 44: >Ethernet [](http://example.com/codeflow?start=29&length=8)

There is use case like `Ethernet124` not covered by this pattern.
PortChannel may be a valid use case if there is PortChannel in Vlan.

src/sonic-yang-models/yang-models/sonic-events-swss.yang
```diff
@@ -38,11 +38,11 @@ module sonic-events-swss {
                 The name of the flapping i/f and status are the only params.";
 
             leaf ifname {
-                type leafref {
-                    path "/port:sonic-port/port:PORT/port:PORT_LIST/port:name";
+                type string {
+                    pattern 'Ethernet[0-9]{1,2}';
```
qiluo-msft, line 42: >Ethernet [](http://example.com/codeflow?start=29&length=8)

There is use case like `Ethernet124` not covered by this pattern.

src/sonic-yang-models/yang-models/sonic-events-dhcp-relay.yang
zbud-msft: Changed pattern.

I checked with Kelly, for this event it should just be Ethernets, PortChannel is used for sending out packets

src/sonic-yang-models/yang-models/sonic-events-swss.yang
zbud-msft: Changed

ganglyu (approved):

kellyyeh (approved):

src/sonic-yang-models/yang-models/sonic-events-dhcp-relay.yang
qiluo-msft: Could you explain more about the disadvantage of using leafref PORT_LIST? The pattern string is not future proof.

zbud-msft: I initially had it as a leafref, but we can't use it for our use case, because we are getting events as a json format such as 

              "stdout_lines": [
                   "2023-01-10 01:02:45,053 [INFO] CURRENT JSON_DATA: {\"sonic-events-swss:sonic-events-swss\":{", 
                   "  \"sonic-events-swss:if-state\": {", 
                   "    \"ifname\": \"Ethernet8\",", 
                   "    \"status\": \"down\",", 
                   "    \"timestamp\": \"2023-01-10T01:02:30.349203Z\"", 
                   "  }", 
                   "}}", 

In order to use a leafref I would need a reference such as the following as part of my json, which is not possible 
      "sonic-port:sonic-port": {
            "sonic-port:PORT": {
                "PORT_LIST": [
                    {
                        "admin_status": "up",
                        "alias": "eth0",
                        "description": "Ethernet0",
                        "lanes": 65,
                        "mtu": 9000,
                        "name": "Ethernet0",
                        "speed": 25000
                    }
                ]
            }
        }, 

"2023-01-10 01:02:45,056 [INFO] Exception thrown: Leafref \"/sonic-port:sonic-port/sonic-port:PORT/sonic-port:PORT_LIST/sonic-port:name\" of value \"Ethernet8\" points to a non-existing leaf."

qiluo-msft (approved):

PR #13345, opened 2023-01-12
database-chassis python http-server implementation

anamehra:
Signed-off-by: anamehra [anamehra@cisco.com](mailto:anamehra@cisco.com)
```
  Python HTTP Server service.
  This service runs in database-chassis docker based on the configuration
  provided in device/<platform>/chassisdb.conf

  Configuration parameters provided by chassisdb.conf are as follows:

  start_http_server: if "yes" or "1", http-server will be started on this
      node
  chassis_db_address   : IP address for http-server. This is the same IP used
   for the chassis db server
  http_server_port : Port to bind to, default: 8000
  http_server_dir  : HTTP server home directory path, default: /var/www/

# Example of config:
'''
start_http_server=yes
chassis_db_address=127.0.0.10
http_server_dir=/var/www/tftp
'''
```

#### Why I did it
Enable http-server as database-chassis service on Chassis supervisor to enable LCs to download image and config files from Supervisor.

#### How I did it
Added python3 http-server based implementation in database-chassis as a docker service run by supervisord.

#### How to verify it
Add device/<platform>/chassisdb.conf with the following data:
start_http_server=yes
chassis_db_address=127.0.0.10
http_server_dir=/var/www/tftp

Boot with the image with the above data.

put an image file at /var/www/tftp/filename
check 'systemctl status http-service'
curl http://http-server:8000/filename -o out_file
compare file_out with /var/www/tftp/filename

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [x] 202205
- [ ] 202211


anamehra:
@abdosi , FYI-
Thanks

abdosi (dismissed):
LGTM

abdosi:
@arlakshm / @judyjoseph  please help review.

anamehra:
@judyjoseph , @arlakshm , please let me know if you have any comments. I resolved all comments from our last meeting discussion in this PR. moved the http-server in database-chassis docker for chassis.

dockers/docker-database/http-server
```diff
@@ -0,0 +1,120 @@
...
+            syslog.syslog(syslog.LOG_INFO, 'start_http_server is set, starting http-server')
+            if self.http_config:
+                with socketserver.TCPServer((self.http_server_ip, self.http_server_port), HTTPRequestHandler) as httpd:
+                    syslog.syslog(syslog.LOG_INFO, 'HTTP Server Port:{} home:{}'.format(self.http_server_port, http_dir))
+                    httpd.serve_forever()
+        else:
+            syslog.syslog(syslog.LOG_INFO, 'start_http_server is not set, exiting...')
+
+def main():
+    #import pdb; pdb.set_trace()
```
judyjoseph, line 112: We could remove this comment.

judyjoseph (dismissed):

dockers/docker-database/supervisord.conf.j2
```diff
@@ -56,3 +56,19 @@ stdout_logfile=syslog
 stderr_logfile=syslog
 dependent_startup=true
 dependent_startup_wait_for=rsyslogd:running
+
+{% if INSTANCES %}
+{%     for redis_inst, redis_items in INSTANCES.items() %}
+{% if redis_inst == 'redis_chassis' %}
```
qiluo-msft, line 62: >if [](http://example.com/codeflow?start=3&length=2)

Need 2 level indentation

dockers/docker-database/Dockerfile.j2
```diff
@@ -44,5 +44,6 @@ COPY ["files/sysctl-net.conf", "/etc/sysctl.d/"]
 COPY ["critical_processes", "/etc/supervisor"]
 COPY ["files/update_chassisdb_config", "/usr/local/bin/"]
 COPY ["flush_unused_database", "/usr/local/bin/"]
+COPY ["http-server", "/usr/local/bin/"]
```
qiluo-msft, line 47: >http-server [](http://example.com/codeflow?start=7&length=11)

Why put this service in database container? I do not see any dependency

dockers/docker-database/http-server
```diff
@@ -0,0 +1,119 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+###########################################################################
+
+# Example of config:
+'''
+start_http_server=yes
+chassis_db_address=127.0.0.10
```
qiluo-msft, line 40: >127.0.0.10 [](http://example.com/codeflow?start=19&length=10)

hard-coded IP?

dockers/docker-database/supervisord.conf.j2
```diff
@@ -56,3 +56,19 @@ stdout_logfile=syslog
...
+[program: http-server]
+command=python3 /usr/local/bin/http-server
+priority=3
+autostart=true
```
qiluo-msft, line 66: >true [](http://example.com/codeflow?start=10&length=4)

no way to disable it

dockers/docker-database/http-server
```diff
@@ -0,0 +1,119 @@
...
+import argparse
+import http.server
+import os
+import socketserver
+import subprocess
+import syslog
+
+from pathlib import Path
+
+HTTP_DEFAULT_BIND_PORT = "8000"
```
qiluo-msft, line 54: >8000 [](http://example.com/codeflow?start=26&length=4)

Do we have rules to protect this service?

dockers/docker-database/supervisord.conf.j2
anamehra: Thanks for the review @qiluo-msft !
I will fix this.

dockers/docker-database/Dockerfile.j2
anamehra: My initial PR was to run this as host side service as a feature but in the last review in Chassis group, it was decided to use it inside database-chassis for now as it's only for internal image/data hosting for chassis modules. It runs only on Supervisor on Chassis and binds to same internal ip as redis_chassis.server.

dockers/docker-database/http-server
anamehra: This is an example of the config. The chassis_db_address is used for redis_chassis.server and is hardcoded as of today, provided by chassisd.conf file by vendor in <platform> dir.
This http-serevr use the same IP to bind to service requests on internal midplane network.

dockers/docker-database/http-server
anamehra: I will investigate more on this and respond/implement. Thanks

dockers/docker-database/Dockerfile.j2
abdosi: @qiluo-msft yes we had discussion on this and if we want to run on host there was comment from Guohan to add this as Feature in Feature Table. But that was little deviation from current model where each feature is like docker with we can enable/disable the feature by doing `container stop/start`.  So in chassis subgroup we discussed and thought since that will be some infrastructure change for this scenario to unblock chassis use case we thought to make it part of `database-chassis` docker which is specific to chassis use case. Also in future if this is needed in host we can revisit.

dockers/docker-database/http-server
abdosi: @qiluo-msft your point is to have Control plane acl (ip table rules)   to explicitly permit this ?

qiluo-msft: yes, I believe you also want to limit the access, for example by some rules on 5-tuple

anamehra: @abdosi , @qiluo-msft , this port is on the internal network. Do we have any rules on other internal ip/ports which I can refer to?

anamehra: Hi @qiluo-msft , Abhishek and I had a discussion on this. The current Sonic implementation does not have any rule defined for the midplane subnet yet and we need to have a rule for all kinds of access in the midplane subnet. We will have to define platform subnet parameters that caclmgrd could use to define the rules for midplane access for all kinds of midplane traffic in the chassis. I will open a case and we can use a different PR to define the general subnet access rule. Let me know if that is fine.

abdosi: @qiluo-msft on chassis setup we have additional interface `eth1-midplane` both on supervisor and LC for internal traffic (eg: redis and http)  . We will add iptable rule to allow all traffic on that interface. That way http traffic will be protected from outside network (`eth0`) and will get permit only if coming from internal network. I will create separate for that

abdosi: Created PR : https://github.com/sonic-net/sonic-host-services/pull/42 . With this PR any port 8000 traffic that is not coming on eth1-midplane will get drop and thus protected from external network access.

abdosi:
@qiluo-msft can you please review this again.

dockers/docker-database/supervisord.conf.j2
anamehra: fixed.

gechiang:
@anamehra , can you resolve the conflict since the changes has been a while and require rebase/merge to resolve conflicts.
We can then ask @qiluo-msft  to approve/merge.
Thanks!

abdosi:
> @anamehra , can you resolve the conflict since the changes has been a while and require rebase/merge to resolve conflicts. We can then ask @qiluo-msft to approve/merge. Thanks!

we don;t need this for 202205. This will gets merged to master directly.

anamehra:
> @anamehra , can you resolve the conflict since the changes has been a while and require rebase/merge to resolve conflicts. We can then ask @qiluo-msft to approve/merge. Thanks!

resolved. Thanks!

arlakshm:
Lower priority now. Will revisit in the future

Issue #13281, opened 2023-01-06
[Build] Failed to build isc dhcp


Junchao-Mellanox:
#### Description

Build failed due to

```
16:51:17  make[1]: Entering directory '/sonic/src/isc-dhcp'
16:51:17  # Remove any stale files
16:51:17  rm -rf ./isc-dhcp-4.4.1
16:51:17  # Get isc-dhcp release, debian files
16:51:17  dget http://deb.debian.org/debian/pool/main/i/isc-dhcp/isc-dhcp_4.4.1-2.3.dsc
16:51:17  pushd ./isc-dhcp-4.4.1
16:51:17  # Create a git repository here for stg to apply patches
16:51:17  git init
16:51:17  git add -f *
16:51:17  git commit -qm "initial commit"
16:51:17  # Apply patches
16:51:17  stg init
16:51:17  stg import -s ../patch/series
16:51:17  # Build source and Debian packages
16:51:17  dpkg-buildpackage -rfakeroot -b -us -uc -j30 --admindir /sonic/dpkg/tmp.lJosSFVurg
16:51:17  popd
16:51:17  # Move the newly-built .deb packages to the destination directory
16:51:17  mv isc-dhcp-relay-dbgsym_4.4.1-2.3_amd64.deb isc-dhcp-relay_4.4.1-2.3_amd64.deb /sonic/target/debs/bullseye/
16:51:17  dget: retrieving http://deb.debian.org/debian/pool/main/i/isc-dhcp/isc-dhcp_4.4.1-2.3.dsc
16:51:17  get_url_version http://deb.debian.org/debian/pool/main/i/isc-dhcp/isc-dhcp_4.4.1-2.3.dsc failed
16:51:17  dget: curl isc-dhcp_4.4.1-2.3.dsc http://deb.debian.org/debian/pool/main/i/isc-dhcp/isc-dhcp_4.4.1-2.3.dsc failed
```

The root cause is that isc-dhcp_4.4.1-2.3.dsc is no longer available.

#### Steps to reproduce the issue:
Build sonic

#### Describe the results you received:

```
dget: curl isc-dhcp_4.4.1-2.3.dsc http://deb.debian.org/debian/pool/main/i/isc-dhcp/isc-dhcp_4.4.1-2.3.dsc failed
```

#### Describe the results you expected:

No error

#### Output of `show version`:

```
(paste your output here)
```
#### Output of `show techsupport`:

```
(paste your output here or download and attach the file here )
```

[...]

PR #13381, opened 2023-01-16
[build]  Check versions availability in rules/*.mk at init and use apt-get source instead of dget or wget in Makefiles

guillaumelambert:
- create an a apt alternate configuration directory during 'make init'
- use 'apt-get source' instead of dget and wget in makefiles
- adapt packages versions accordingly in rules/XX.mk files
- perform versions check at the make init stage  to avoid time and resources waste
  Available versions can be retrieved with 'apt-cache showsrc'
  If target version is no more available, there is no point to pursue the build.
  Cf https://github.com/sonic-net/sonic-buildimage/pull/13288
  
SONiC images compilation is currently designed to be performed with a
non-root user under Ubuntu systems.
And the build process recompiles packages for various GNU+Linux Debian
distributions.
As a result, there is no guarantee that debian repos are declared in
/etc/apt/sources.list on the Ubuntu host system used.
And 'apt-get source' can not be directly used out-of-the-box.
Currently, the packages sources are retrieved from more or less official URLs
and locations with dget and wget.
As discussed in the following URL
https://serverfault.com/questions/447457/use-apt-get-source-on-a-debian-repo-without-using-etc-apt-source-list
dget is the simplest solution but this approach has several drawbacks from a
maintenance and security standpoint, for example :
* maintaining manually these URLS since they can be removed by external
  sites owners
https://github.com/sonic-net/sonic-buildimage/issues/13281
* loose authentication of packages sources retrieved with HTTP instead of HTTPS potentially vulnerable to MITM at least for wget

A more complex solution is proposed (on the same URL at serverfault) and address these
drawbacks by using an alternate apt configuration directory that is writable
by non-root user. This PR leverages this solution and allows here to:
* identify packages sources with official Debian Keyring GPG signatures
* use 'apt-get source <pkgsrc_name>' to retrieve packages without maintaining
  several specific URLs
* use 'apt-cache showsrc <pkgsrc_name>' to retrieve a
  list of available versions to ckeck packages availability during the
  'make init' phase
  If no candidate version is available, the script will exit with an error code and stops the whole compilation process.
  This will prevent starting a long build process condemned to fail and thus spare precious time and computing resources.

Issue  #13281

potentially all

[build]  Check versions availability in rules/*.mk at init and use apt-get source instead of dget or wget in Makefiles 

                          .".
                         /  |
                        /  /
                       / ,"
           .-------.--- /
          "._ __.-/ o. o\  
             "   (    Y  )
                  )     /
                 /     (
                /       Y
            .-"         |
           /  _     \    \ 
          /    `. ". ) /' )
         Y       )( / /(,/
        ,|      /     )
       ( |     /     /
        " \_  (__   (__        [nabis]
            "-._,)--._,)


guillaumelambert:
Builds themselves were OK.
It seems previous CI run was KO because there was an issue in the queue to lock one of the testbed.
https://dev.azure.com/mssonic/build/_build/results?buildId=203972&view=logs&j=90badf3f-7a57-5c0e-d9bd-d2e7a4fde36a&t=bc4195ad-8694-52a1-21a3-01785e725d8d

PR #13386, opened 2023-01-17
[build] Check if patches are applied before applying patches

liushilongbuaa:
#### Why I did it
If make fails, we can't rerun the make process, because existing patches can't apply again.
#### How I did it
Check if patches are applied. if yes, don't apply patches again.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211

### DIFF
--- a/slave.mk
+++ b/slave.mk
@@ -623,7 +623,7 @@ $(addprefix $(FILES_PATH)/, $(SONIC_MAKE_FILES)) : $(FILES_PATH)/% : .platform $
 		# Remove target to force rebuild
 		rm -f $(addprefix $(FILES_PATH)/, $*)
 		# Apply series of patches if exist
-		if [ -f $($*_SRC_PATH).patch/series ]; then pushd $($*_SRC_PATH) && QUILT_PATCHES=../$(notdir $($*_SRC_PATH)).patch quilt push -a; popd; fi
+		if [ -f $($*_SRC_PATH).patch/series ]; then pushd $($*_SRC_PATH) && quilt applied || QUILT_PATCHES=../$(notdir $($*_SRC_PATH)).patch quilt push -a; popd; fi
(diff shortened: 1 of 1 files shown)


kv-y:
> If make fails, we can't rerun the make process, because existing patches can't apply again.

Maybe just need to fix a cleanup stage for some packages?

I mean usually we have the next steps:

1. Remove a directory with source code if it exists
2. Download and unpack some files if it's required
3. Apply patches
4. Build
5. Install

Some packages like ntp have incorrect 1st step:
NTP_VERSION is 4.2.8p15+dfsg and we try to remove ntp-$(NTP_VERSION) but in fact directory is named `ntp-4.2.8p15`

So if directory with source code is removed correctly we don't have any issues with applying patches when we rerun make. But it's just a suggestion.
Probably your PR fixes another bug.

qiluo-msft (approved):

liushilongbuaa:
> > If make fails, we can't rerun the make process, because existing patches can't apply again.
> 
> Maybe just need to fix a cleanup stage for some packages?
> 
> I mean usually we have the next steps:
> 
> 1. Remove a directory with source code if it exists
> 2. Download and unpack some files if it's required
> 3. Apply patches
> 4. Build
> 5. Install
> 
> Some packages like ntp have incorrect 1st step: NTP_VERSION is 4.2.8p15+dfsg and we try to remove ntp-$(NTP_VERSION) but in fact directory is named `ntp-4.2.8p15`
> 
> So if directory with source code is removed correctly we don't have any issues with applying patches when we rerun make. But it's just a suggestion. Probably your PR fixes another bug.

That will be a big change. This small change will work in a lot of scenarios.

PR #13408, opened 2023-01-17
[dualtor][active-active]Killing radv instead of stopping on `active-active` dualtor if config knob is on

zjswhhh:
`radv` sends a good-bye packet when the service is stopped, which causes a IPv6 route update on SoC side. And this update leads to an interface bouncing and causes traffic disruption even though the ToR device might already be isolated. 

This PR is to mitigate the traffic disruption issue during planned maintenance, by killing radv instead of stopping. So the cease packet won't be sent. 

sign-off: Jing Zhang zhangjing@microsoft.com

#### How to verify it
Verified on dev clusters: 
* Traffic disruption was no longer reproducible.
* `radv` took the killing path
* if knob was off, `radv` would take the stopping path 

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211

### DIFF
--- a/files/scripts/service_mgmt.sh
+++ b/files/scripts/service_mgmt.sh
@@ -39,7 +39,11 @@ function check_redundant_type()
     else
         ACTIVE_ACTIVE="false"
     fi
-    debug "Active-active cable type dualtor: ${ACTIVE_ACTIVE}"
+    CONFIG_KNOB=`$SONIC_DB_CLI CONFIG_DB hget "FEATURE|mux" kill_radv`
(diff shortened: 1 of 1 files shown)


yxieca:
Please make following updates:
1. make the default off.
2. put the configuration in a better table.

zjswhhh:
> Please make following updates:
> 
> 1. make the default off.
> 2. put the configuration in a better table.

Addressed.

yxieca (approved):

PR #13412, opened 2023-01-18
Fix sonic-slave-* build errors about sudo command not found

liushilongbuaa:
#### Why I did it
issue https://github.com/sonic-net/sonic-buildimage/issues/13395
1. Fix a bug about sudo failure.
```
/usr/local/share/buildinfo/scripts/buildinfo_base.sh: line 24: sudo: command not found
```
2. Fix an issue about warning message.
```
./scripts/run_with_retry: line 4: [: : integer expression expected
```

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211

### DIFF
--- a/scripts/run_with_retry
+++ b/scripts/run_with_retry
@@ -1,7 +1,9 @@
 #!/bin/bash
 
 run_with_retry(){
+    # set default value and change invalid param input to 0
     (( SONIC_BUILD_RETRY_COUNT > 0 )) || SONIC_BUILD_RETRY_COUNT=0
+    (( SONIC_BUILD_RETRY_INTERVAL > 0 )) || SONIC_BUILD_RETRY_INTERVAL=0
     [[ "$*" == "" ]] && { echo "run_with_retry: input command can't be empty." 1>&2;exit 1; }
     for ((i=0; i<=$SONIC_BUILD_RETRY_COUNT; i++))
     do
--- a/scripts/run_with_retry
+++ b/scripts/run_with_retry
@@ -3,7 +3,7 @@
 run_with_retry(){
     # set default value and change invalid param input to 0
     (( SONIC_BUILD_RETRY_COUNT > 0 )) || SONIC_BUILD_RETRY_COUNT=0
-    (( SONIC_BUILD_RETRY_INTERVAL > 0 )) || SONIC_BUILD_RETRY_INTERVAL=0
+    (( SONIC_BUILD_RETRY_INTERVAL > 0 )) || SONIC_BUILD_RETRY_INTERVAL=600
     [[ "$*" == "" ]] && { echo "run_with_retry: input command can't be empty." 1>&2;exit 1; }
     for ((i=0; i<=$SONIC_BUILD_RETRY_COUNT; i++))
     do
--- a/scripts/run_with_retry
+++ b/scripts/run_with_retry
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 run_with_retry(){
-    [ "$SONIC_BUILD_RETRY_COUNT" -gt 0 ] || SONIC_BUILD_RETRY_COUNT=0
+    (( SONIC_BUILD_RETRY_COUNT > 0 )) || SONIC_BUILD_RETRY_COUNT=0
     [[ "$*" == "" ]] && { echo "run_with_retry: input command can't be empty." 1>&2;exit 1; }
     for ((i=0; i<=$SONIC_BUILD_RETRY_COUNT; i++))
     do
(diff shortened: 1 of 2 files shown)


src/sonic-build-hooks/scripts/buildinfo_base.sh
```diff
@@ -15,26 +15,27 @@ REPR_MIRROR_URL_PATTERN='http:\/\/packages.trafficmanager.net\/'
...
+fi
+
 if [ -e /vcache ]; then
 	PKG_CACHE_PATH=/vcache/${IMAGENAME}
 else
 	PKG_CACHE_PATH=/sonic/target/vcache/${IMAGENAME}
 fi
 PKG_CACHE_FILE_NAME=${PKG_CACHE_PATH}/cache.tgz
-sudo chown $USER $(dirname $PKG_CACHE_PATH)
+$SUDO chown $USER $(dirname $PKG_CACHE_PATH)
```
nazariig, line 31: @liushilongbuaa do we need to put it AFTER `mkdir`?

nazariig (changes requested):

src/sonic-build-hooks/scripts/buildinfo_base.sh
liushilongbuaa: fixed.

scripts/run_with_retry
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 run_with_retry(){
-    [ "$SONIC_BUILD_RETRY_COUNT" -gt 0 ] || SONIC_BUILD_RETRY_COUNT=0
+    (( SONIC_BUILD_RETRY_COUNT > 0 )) || SONIC_BUILD_RETRY_COUNT=0
```
nazariig, line 4: @liushilongbuaa IMHO, this is a bad design. Here we are hiding an active build configuration from user, plus some potential issues (e.g., invalid param input). I would recommend adding these params (both `SONIC_BUILD_RETRY_COUNT` and `SONIC_BUILD_RETRY_INTERVAL`) to the `rules/config`.

Another option is:
```bash
SONIC_BUILD_RETRY_COUNT="${SONIC_BUILD_RETRY_COUNT:-0}"
SONIC_BUILD_RETRY_INTERVAL="${SONIC_BUILD_RETRY_INTERVAL:-600}"
```

Please check:
1. `sonic-net/sonic-buildimage/.azure-pipelines/template-variables.yml`
2. `sonic-net/sonic-buildimage/Makefile.work`
3. `sonic-net/sonic-buildimage/slave.mk`
4. `sonic-net/sonic-buildimage/rules/config`

nazariig (changes requested):

scripts/run_with_retry
liushilongbuaa: If we put it in rules/config, the change in Makefile seems to be heavy. We need to include rules/config and export variables.
It is not a SONiC related config. It is a build related config. We want to improve PR checks build success rate. I think it is not neccessary to put it in rules/config.

liushilongbuaa: If input is string. Line 4 will change it to 0.
SONIC_BUILD_RETRY_COUNT="${SONIC_BUILD_RETRY_COUNT:-0}"   this format will keep string variable.

liushilongbuaa: https://github.com/sonic-net/sonic-buildimage/pull/12325/files
This change didn't touch slave.mk and Makefile.work

nazariig: > If input is string. Line 4 will change it to 0. SONIC_BUILD_RETRY_COUNT="${SONIC_BUILD_RETRY_COUNT:-0}" this format will keep string input.

Exactly. As i wrote above - we are hiding an issues from user (e.g., expected integer - provided string).
Do we really want to do that?

nazariig: @liushilongbuaa got your point

nazariig (approved):

scripts/run_with_retry
liushilongbuaa: Developer maybe don't need this feature, because it will take long time to retry. But it will be good to PR validation and official build.

xumia (approved):

liat-grozovik:
@xumia as this is a blocking issue on  many builds we should take it as is but I think @nazariig  has a good point and suggest to re evaluate the fix moving forward.

liushilongbuaa:
@liat-grozovik , Thanks for merging it.
FYI. It is not blocking issue. If someone is debugging build break, this PR maybe don't help.

PR #13452, opened 2023-01-20
[202205] Update BRCM SAI version to 7.1.32.4

tjchadaga:
#### Why I did it
To pick up the below DNX fixes:
1. CS00012275689: DSCP->TC and TC->QUEUE mappings are not happening for packets received on LAG ports (SONIC-69367)
2. CS00012277618: Crash in _brcm_sai_dnx_irpp_port_core_get (SONIC-70001)

#### How I did it
Updated SAI branch with the above fixes

#### How to verify it
Ran basic sonic-mgmt tests with the SAI debian on XGS and DNX platforms

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211


vmittal-msft (approved):

lguohan:
are we updating xgs/dnx sai at the same time now?

gechiang (approved):

gechiang:
> are we updating xgs/dnx sai at the same time now?

This change is applicable to DNX only so as far as XGS is concerned it is still the same as the previous version.
I think it may be better to keep both version the same even though there are no changes.  We should only make it different if there is a need to make it different.

arlakshm (approved):

PR #13453, opened 2023-01-20
teamd: Add support for custom retry counts for LACP sessions

saiarcot895:

#### Why I did it

This is to add support for specifying custom retry counts for LACP sessions. This is to make warmboot easier on low-storage and low-memory platforms, by allowing more than 90 seconds of downtime.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211

### DIFF
--- a/src/libteam/patch/block-retry-count-changes.patch
+++ b/src/libteam/patch/block-retry-count-changes.patch
@@ -1,0 +1,60 @@
+Don't reset the retry count after setting it for 60 seconds
+
+From: Saikrishna Arcot <sarcot@microsoft.com>
+Date: 2023-01-18 14:26:36 -0800
+
+After setting the retry count to some custom value, if a normal LACP
+packet comes in without a custom retry count, don't reset it back to
+that custom retry count for the first 60 seconds.
+---
+ teamd/teamd_runner_lacp.c |   22 ++++++++++++++++------
+ 1 file changed, 16 insertions(+), 6 deletions(-)
+
+diff --git a/teamd/teamd_runner_lacp.c b/teamd/teamd_runner_lacp.c
+index f58ad3b..7df3bcb 100644
+--- a/teamd/teamd_runner_lacp.c
++++ b/teamd/teamd_runner_lacp.c
+@@ -180,6 +180,7 @@ struct lacp {
+ 	struct lacp_port *selected_agg_lead; /* leading port of selected aggregator */
+ 	bool carrier_up;
+ 	time_t warm_start_mode_timer;
++	time_t next_retry_count_change_time;
+ 	struct {
+ 		bool active;
+ #define		LACP_CFG_DFLT_ACTIVE true
+@@ -1457,6 +1458,8 @@ static int lacpdu_send(struct lacp_port *lacp_port)
+ 	return err;
+ }
+ 
++#define LACP_NEXT_RETRY_COUNT_CHANGE_TIMEOUT 60
++
+ static int lacpdu_process(struct lacp_port *lacp_port, struct lacpdu* lacpdu)
+ {
+ 	int err;
+@@ -1493,13 +1496,20 @@ static int lacpdu_process(struct lacp_port *lacp_port, struct lacpdu* lacpdu)
+ 					   lacp_port->partner_retry_count,
+ 					   lacpdu->v2.actor_retry_count);
+ 		lacp_port->partner_retry_count = lacpdu->v2.actor_retry_count;
++		lacp_port->lacp->next_retry_count_change_time = time(NULL) + LACP_NEXT_RETRY_COUNT_CHANGE_TIMEOUT;
+ 	} else {
+-		if (lacp_port->partner_retry_count != LACP_CFG_DFLT_RETRY_COUNT)
+-			teamd_log_dbg(lacp_port->ctx, "%s: retry count from partner changed from %u to %u",
+-					   lacp_port->tdport->ifname,
+-					   lacp_port->partner_retry_count,
+-					   LACP_CFG_DFLT_RETRY_COUNT);
+-		lacp_port->partner_retry_count = LACP_CFG_DFLT_RETRY_COUNT;
++		time_t currentTime = time(NULL);
++		if (currentTime < lacp_port->lacp->next_retry_count_change_time) {
++			teamd_log_dbg(lacp_port->ctx, "%s: retry count changed too recently",
++					lacp_port->tdport->ifname);
++		} else {
++			if (lacp_port->partner_retry_count != LACP_CFG_DFLT_RETRY_COUNT)
++				teamd_log_dbg(lacp_port->ctx, "%s: retry count from partner changed from %u to %u",
++						lacp_port->tdport->ifname,
++						lacp_port->partner_retry_count,
++						LACP_CFG_DFLT_RETRY_COUNT);
++			lacp_port->partner_retry_count = LACP_CFG_DFLT_RETRY_COUNT;
++		}
+ 	}
+ 
+ 	err = lacp_port_set_state(lacp_port, PORT_STATE_CURRENT);
--- a/src/libteam/patch/add-support-for-custom-retry.patch
+++ b/src/libteam/patch/add-support-for-custom-retry.patch
@@ -1,0 +1,358 @@
+Add support for custom retry counts for LACP sessions
+
+From: Saikrishna Arcot <sarcot@microsoft.com>
+Date: 2022-12-21 18:11:31 -0800
+
+Add support for using custom retry count (instead of the default of 3) for LACP
+sessions, to allow for sessions to stay up for more than 90 seconds.
+---
+ teamd/teamd_runner_lacp.c |  212 ++++++++++++++++++++++++++++++++++++++++++---
+ 1 file changed, 198 insertions(+), 14 deletions(-)
+
+diff --git a/teamd/teamd_runner_lacp.c b/teamd/teamd_runner_lacp.c
+index 82b8b86..f58ad3b 100644
+--- a/teamd/teamd_runner_lacp.c
++++ b/teamd/teamd_runner_lacp.c
+@@ -77,22 +77,45 @@ struct lacpdu {
+ 	uint8_t			collector_info_len;
+ 	uint16_t		collector_max_delay;
+ 	uint8_t			__reserved3[12];
+-	uint8_t			terminator_tlv_type;
+-	uint8_t			terminator_info_len;
+-	uint8_t			__reserved4[50];
++	union {
++		struct {
++			uint8_t	terminator_tlv_type;
++			uint8_t	terminator_info_len;
++			uint8_t __reserved4[8];
++		} __attribute__((__packed__)) v1;
++		struct {
++			uint8_t actor_retry_tlv_type;
++			uint8_t actor_retry_tlv_len;
++			uint8_t actor_retry_count;
++			uint8_t __reserved_a[1];
++			uint8_t partner_retry_tlv_type;
++			uint8_t partner_retry_tlv_len;
++			uint8_t partner_retry_count;
++			uint8_t __reserved_b[1];
++			uint8_t	terminator_tlv_type;
++			uint8_t	terminator_info_len;
++		} __attribute__((__packed__)) v2;
++	} __attribute__((__packed__));
++	uint8_t			__reserved5[42];
+ } __attribute__((__packed__));
+ 
+-static void lacpdu_init(struct lacpdu *lacpdu)
++static void lacpdu_init(struct lacpdu *lacpdu, uint8_t version)
+ {
+ 	memset(lacpdu, 0, sizeof(*lacpdu));
+ 	lacpdu->subtype			= 0x01;
+-	lacpdu->version_number		= 0x01;
++	lacpdu->version_number		= version;
+ 	lacpdu->actor_tlv_type		= 0x01;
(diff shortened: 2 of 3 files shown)


src/libteam/patch/block-retry-count-changes.patch
```diff
@@ -0,0 +1,60 @@
...
+ 	time_t warm_start_mode_timer;
++	time_t next_retry_count_change_time;
+ 	struct {
+ 		bool active;
+ #define		LACP_CFG_DFLT_ACTIVE true
+@@ -1457,6 +1458,8 @@ static int lacpdu_send(struct lacp_port *lacp_port)
+ 	return err;
+ }
+ 
++#define LACP_NEXT_RETRY_COUNT_CHANGE_TIMEOUT 60
```
yxieca, line 29: Is this in seconds? Suggest rename the macro to indication the unit.

yxieca:
@saiarcot895 in the patch file, there seems to be a mix use of tabs and whitespaces, which caused indentation issues even just see from this PR. Can you unify the spaces?

src/libteam/patch/block-retry-count-changes.patch
saiarcot895: Done

saiarcot895:
> @saiarcot895 in the patch file, there seems to be a mix use of tabs and whitespaces, which caused indentation issues even just see from this PR. Can you unify the spaces?

I converted some lines that appeared to be using spaces to tabs, hopefully I got all of them.

src/libteam/patch/0014-block-retry-count-changes.patch
```diff
@@ -0,0 +1,60 @@
...
+ 		lacp_port->partner_retry_count = lacpdu->v2.actor_retry_count;
++		lacp_port->lacp->next_retry_count_change_time = time(NULL) + LACP_NEXT_RETRY_COUNT_CHANGE_TIMEOUT_SECONDS;
+ 	} else {
+-		if (lacp_port->partner_retry_count != LACP_CFG_DFLT_RETRY_COUNT)
+-			teamd_log_dbg(lacp_port->ctx, "%s: retry count from partner changed from %u to %u",
+-					   lacp_port->tdport->ifname,
+-					   lacp_port->partner_retry_count,
+-					   LACP_CFG_DFLT_RETRY_COUNT);
+-		lacp_port->partner_retry_count = LACP_CFG_DFLT_RETRY_COUNT;
++		time_t currentTime = time(NULL);
```
yxieca, line 46: Are you taking the wall clock or monotonic clock?

saiarcot895: This is taking the wall clock time, but it should instead be using monotonic time. I'll make the change.

src/libteam/patch/0013-add-support-for-custom-retry.patch
```diff
@@ -0,0 +1,358 @@
...
++	}
+ 
+ 	err = teamd_send(lacp_port->sock, &lacpdu, sizeof(lacpdu), 0);
+ 	return err;
+@@ -1428,6 +1482,26 @@ static int lacpdu_process(struct lacp_port *lacp_port, struct lacpdu* lacpdu)
+ 			return err;
+ 	}
+ 
++	if (lacpdu->version_number == 0xf1) {
++		if (lacpdu->v2.actor_retry_count < 3) {
```
judyjoseph, line 174: Can we remove the hardcoded value here, use LACP_CFG_DFLT_RETRY_COUNT instead ?

src/libteam/patch/0013-add-support-for-custom-retry.patch
```diff
@@ -0,0 +1,358 @@
...
++			teamd_log_err("%s: retry count from partner (%u) out of its limits.", lacp_port->tdport->ifname, lacpdu->v2.actor_retry_count);
++			return -EINVAL;
++		}
++		if (lacp_port->partner_retry_count != lacpdu->v2.actor_retry_count)
++			teamd_log_dbg(lacp_port->ctx, "%s: retry count from partner changed from %u to %u",
++					   lacp_port->tdport->ifname,
++					   lacp_port->partner_retry_count,
++					   lacpdu->v2.actor_retry_count);
++		lacp_port->partner_retry_count = lacpdu->v2.actor_retry_count;
```
judyjoseph, line 183: Does this mean we allways expect the actor and partner to have same retry_count ?

src/libteam/patch/0013-add-support-for-custom-retry.patch
saiarcot895: Oh, yes this should use `LACP_CFG_DFLT_RETRY_COUNT` instead.

src/libteam/patch/0013-add-support-for-custom-retry.patch
saiarcot895: No. The actor fields in the `lacpdu` packet that we received refer to the partner fields in `lacp_port`. In other words, in the LACP PDU that we received, the actor information refers to our peer's information (aka our partner).

src/libteam/patch/0014-block-retry-count-changes.patch
saiarcot895: Changed.

src/libteam/patch/0013-add-support-for-custom-retry.patch
saiarcot895: Fixed.

yxieca (approved):

src/libteam/patch/0015-add-support-for-custom-retry.patch
```diff
@@ -0,0 +1,505 @@
...
++	if (err)
++		lacp->cfg.enable_retry_count = LACP_CFG_DFLT_ENABLE_RETRY_COUNT;
++	teamd_log_dbg(ctx, "Using enable_retry_count \"%d\".", lacp->cfg.enable_retry_count);
++
++	err = teamd_config_int_get(ctx, &tmp, "$.runner.retry_count");
++	if (err) {
++		lacp->cfg.retry_count = LACP_CFG_DFLT_RETRY_COUNT;
++	} else if (tmp < 3) {
++		teamd_log_err("\"retry_count\" value is out of its limits.");
++		return -EINVAL;
```
yxieca, line 138: We could probably generate the error message here and continue with retry count of LACP_CFG_DFLT_RETRY_COUNT, and proceed with LACP_CFG_DFLT_RETRY_COUNT?

saiarcot895: Updated.

src/libteam/patch/0016-block-retry-count-changes.patch
```diff
@@ -0,0 +1,141 @@
+Don't reset the retry count after setting it for 60 seconds
+
+From: Saikrishna Arcot <sarcot@microsoft.com>
+Date: 2023-01-18 14:26:36 -0800
+
+After setting the retry count to some custom value, if a normal LACP
+packet comes in without a custom retry count, don't reset it back to
```
vaibhavhd, line 7: Is this description correct? Should this be `don't reset the retry count back to old/default retry count`?

src/libteam/patch/0016-block-retry-count-changes.patch
```diff
@@ -0,0 +1,141 @@
...
+ 	enum lacp_port_state state;
++	int last_received_lacpdu_version;
+ 	bool lacpdu_saved;
+ 	struct lacpdu last_pdu;
+ 	struct {
+@@ -1351,6 +1353,7 @@ static int lacp_port_set_state(struct lacp_port *lacp_port,
+ 		if (err)
+ 			return err;
+ 		lacp_port->lacp->cfg.retry_count = LACP_CFG_DFLT_RETRY_COUNT;
++		lacp_port->last_received_lacpdu_version = 0x01;
```
vaibhavhd, line 37: Why is the last received version set to 0x01?

src/libteam/patch/0016-block-retry-count-changes.patch
```diff
@@ -0,0 +1,141 @@
...
+ 	if (!lacpdu_check(lacpdu)) {
+ 		teamd_log_warn("malformed LACP PDU came.");
+@@ -1523,17 +1529,55 @@ static int lacpdu_process(struct lacp_port *lacp_port, struct lacpdu* lacpdu)
+ 					lacp_port->partner_retry_count,
+ 					lacpdu->v2.actor_retry_count);
+ 			lacp_port->partner_retry_count = lacpdu->v2.actor_retry_count;
++			if (clock_gettime(CLOCK_MONOTONIC, &monotonic_time)) {
++				err = errno;
++				teamd_log_err("%s: unable to get current time: %s", lacp_port->tdport->ifname, strerror(err));
++				return -err;
```
vaibhavhd, line 70: Is this if block doing anything? I don't see `monotonic_time` that is collected here being used anywhere.

src/libteam/patch/0016-block-retry-count-changes.patch
```diff
@@ -0,0 +1,141 @@
...
++			}
++		}
++		if (lacp_port->partner_retry_count != LACP_CFG_DFLT_RETRY_COUNT) {
++			// Reset the change time every time a 0xf1 packet comes in
++			if (clock_gettime(CLOCK_MONOTONIC, &monotonic_time)) {
++				err = errno;
++				teamd_log_err("%s: unable to get current time: %s", lacp_port->tdport->ifname, strerror(err));
++				return -err;
```
vaibhavhd, line 78: What happens to retry count if getting time fails? Will/should this be fatal for teamd?

src/libteam/patch/0016-block-retry-count-changes.patch
```diff
@@ -0,0 +1,141 @@
...
++						lacp_port->tdport->ifname,
++						lacp_port->partner_retry_count,
++						LACP_CFG_DFLT_RETRY_COUNT);
++				lacp_port->partner_retry_count = LACP_CFG_DFLT_RETRY_COUNT;
++			}
+ 		}
+ 	}
+ 
++	if (lacp_port->last_received_lacpdu_version != lacpdu->version_number) {
++        teamd_log_dbg(lacp_port->ctx, "%s: LACPDU version changed from %u to %u",
```
vaibhavhd, line 108: Indentation here looks not consistent w/ rest of the indentation in this file.

src/libteam/patch/0016-block-retry-count-changes.patch
```diff
@@ -0,0 +1,141 @@
...
++                lacp_port->tdport->ifname,
++                lacp_port->last_received_lacpdu_version,
++                lacpdu->version_number);
++		lacp_port->last_received_lacpdu_version = lacpdu->version_number;
++		// Force-send a LACPDU packet acknowledging change in version
++		err = lacpdu_send(lacp_port);
```
vaibhavhd, line 114: The same packet w/ changed version is sent here. Don't you also want to change retry count in the sent packet?

src/libteam/patch/0016-block-retry-count-changes.patch
saiarcot895: This is in the case where the LAG has gone down (changed to either expired or defaulted state); at that point, it's better to assume that the peer doesn't understand 0xf1, in case the peer changed, or the peer OS version changed to some older version; otherwise, the LAG may never come up.

src/libteam/patch/0016-block-retry-count-changes.patch
saiarcot895: `clock_gettime` returns -1 for failure and 0 for success; that means if it gets into the if-block, then `clock_gettime` failed for some reason.

src/libteam/patch/0016-block-retry-count-changes.patch
saiarcot895: Hmm, with the current code, this would mean the LACPDU packet wouldn't even get processed, which would mean that the 90-second (or however many seconds based on the retry count) timer won't get refreshed, and the LAG would go down.

The timers that teamd uses also use `CLOCK_MONOTONIC`, so I'm inclined to think that if that call fails 3 times, something is messed up with the system.

src/libteam/patch/0016-block-retry-count-changes.patch
saiarcot895: The new retry count is in this packet (it would've been updated in the previous if-else block, the `lacp_port->partner_retry_count` field).

src/libteam/patch/0016-block-retry-count-changes.patch
saiarcot895: Fixed.

src/libteam/patch/0016-block-retry-count-changes.patch
saiarcot895: Fixed.

vaibhavhd (approved):

judyjoseph (approved):
LGTM, I see that the cutom retry is sent based on the RX of LACP PDU. Is there a way user can set this explicitly to override the default retry of 3 ?

saiarcot895:
The user can specify any custom retry count from the CLI (there's a sonic-utilities PR that was merged in with support for this). However, after some period of time, or some state changes (such as the link going down or the LACP session going down), it will reset back to 3 to comply with the current LACP standard.

PR #13484, opened 2023-01-23
Finalize fast-reboot in warmboot finalizer

arfeigin:
This PR should come along with sonic-utilities PR (https://github.com/sonic-net/sonic-utilities/pull/2621) removing the timer usage from fast-reboot script and all other submodule PRs using the fast-reboot state-db entry:
https://github.com/sonic-net/sonic-swss-common/pull/742
https://github.com/sonic-net/sonic-platform-daemons/pull/335
https://github.com/sonic-net/sonic-sairedis/pull/1196

This set of PRs solves the issue https://github.com/sonic-net/sonic-buildimage/issues/13251

#### Why I did it
To solve an issue with upgrade with fast-reboot including FW upgrade which has been introduced since moving to fast-reboot over warm-reboot infrastructure.
As well, this introduces fast-reboot finalizing logic to determine fast-reboot is done.
 
#### How I did it
Added logic to finalize-warmboot script to handle fast-reboot as well, this makes sense as using fast-reboot over warm-reboot this script will be invoked. The script will clear fast-reboot entry from state-db instead of previous implementation that relied on  timer. The timer could expire in some scenarios between fast-reboot finished causing fallback to cold-reboot and possible crashes.

As well this PR updates all services/scripts reading fast-reboot state-db entry to look for the updated value representing fast-reboot is active.

#### How to verify it
Run fast-reboot and check that fast-reboot entry exists in state-db right after startup and being cleared as warm-reboot is finalized and not due to a timer.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [X] 202205
- [X] 202211

### DIFF
--- a/files/image_config/warmboot-finalizer/finalize-warmboot.sh
+++ b/files/image_config/warmboot-finalizer/finalize-warmboot.sh
@@ -55,13 +55,11 @@ function check_warm_boot()
 function check_fast_reboot()
 {
     debug "Checking if fast-reboot is enabled..."
-    SYSTEM_FAST_REBOOT=`sonic-db-cli STATE_DB GET "FAST_RESTART_ENABLE_TABLE|system"`
-    if [[ ${SYSTEM_FAST_REBOOT} == "enable" ]]; then
+    FAST_REBOOT=`sonic-db-cli STATE_DB hget "FAST_RESTART_ENABLE_TABLE|system" enable`
+    if [[ x"${FAST_REBOOT}" == x"true" ]]; then
        debug "Fast-reboot is enabled..."
-       FAST_REBOOT='true'
     else
        debug "Fast-reboot is disabled..."
-       FAST_REBOOT='false'
     fi
 }
 
@@ -113,7 +111,7 @@ function finalize_warm_boot()
 function finalize_fast_reboot()
 {
     debug "Finalizing fast-reboot..."
-    sonic-db-cli STATE_DB SET "FAST_RESTART_ENABLE_TABLE|system" "disable" &>/dev/null
+    sonic-db-cli STATE_DB hset "FAST_RESTART_ENABLE_TABLE|system" "enable" "false" &>/dev/null
 }
 
 function stop_control_plane_assistant()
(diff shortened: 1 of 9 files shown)


stepanblyschak (approved):

files/image_config/warmboot-finalizer/finalize-warmboot.sh
```diff
@@ -52,6 +52,16 @@ function check_warm_boot()
     WARM_BOOT=`sonic-db-cli STATE_DB hget "WARM_RESTART_ENABLE_TABLE|system" enable`
 }
 
+function check_fast_reboot()
+{
+    SYSTEM_FAST_REBOOT=`sonic-db-cli STATE_DB GET "FAST_REBOOT|system"`
```
vaibhavhd, line 57: It might help to put a debug print here to announce if fast-reboot is enabled/disable. I think we should do same for warmreboot.

files/image_config/warmboot-finalizer/finalize-warmboot.sh
```diff
@@ -152,4 +169,7 @@ if [[ -n "${list}" ]]; then
     debug "Some components didn't finish reconcile: ${list} ..."
 fi
 
+if [ ${FAST_REBOOT} == "true" ]; then
+    finalize_fast_reboot
+fi
 finalize_warm_boot
```
vaibhavhd, line 175: `finalize_warm_boot` is being called regardless of system undergoing fast/warm boot. Is that needed? We can put that into conditional block for warm reboot.

vaibhavhd (commented):
Some minor comments on code.

files/image_config/warmboot-finalizer/finalize-warmboot.sh
arfeigin: As I wrote in the other comment in case warm-reboot is disabled it is being printed to log and script exits. In that case is it required to add also print for the case it is enabled? It will print when warm-reboot is finalized.

Let me know what you think. In the meantime I will add debug prints for fast-reboot check.

vaibhavhd: No change needed in that case. Thanks!

stepanblyschak (approved):

PR #13532, opened 2023-01-27
[Chassis][EVERFLOW][ACL_ATBLE] Fix for everflow ACL_TABLE in config_db not having the routed ports when no -ASIC in the asic_port_name

mlok-nokia:
#### Why I did it
After the renaming of the asic_port_name in port_config.ini file (PR:  https://github.com/sonic-net/sonic-buildimage/pull/13053  ), the asic_ifname in port_config.ini is changed from '<port>-ASIC<asic_id>' to just port. Example: 'Eth0-ASIC0' to 'Eth0'.

However, with this change a config_db generated via config load_minigraph would cause the EVERFLOW and EVERFLOWV6 tables under ACL_TABLE to not have any of non-LAG front panel interfaces. This was causing the EVERFLOW suite to fail.
  
#### How I did it
In parse_asic_external_neigbhors in minigraph.py there was a check that the asic_name.lower() (like asic0) is present in the port_alias_asic_map. However with -ASIC removed from the asic_ifname, the port_alias_asic_map would not have the asic_name and thus any non-LAG neighbor would not be included.

 Fix was the ignore the asic name change as the port_alias_asic_map is already only looking for ports in just the same asic as asic_name.

#### How to verify it

Execute "config load_minigraph" with the mingraph which is generated by sonic-mgmt gen-minigraph script. And confirm ono-lag interface are present in the Everfloe table in the config_dbs.
```
       "EVERFLOW": {
            "stage": "ingress", 
            "type": "MIRROR", 
            "policy_desc": "EVERFLOW", 
            "ports": [
                "PortChannel102", 
                "Ethernet64"
            ]
        }, 
        "EVERFLOWV6": {
            "stage": "ingress", 
            "type": "MIRRORV6", 
            "policy_desc": "EVERFLOWV6", 
            "ports": [
                "PortChannel102", 
                "Ethernet64"
            ]
        }

```

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [x] 202205
- [ ] 202211

### DIFF
--- a/src/sonic-config-engine/minigraph.py
+++ b/src/sonic-config-engine/minigraph.py
@@ -333,15 +333,13 @@ def parse_asic_external_link(link, asic_name, hostname):
     # if chassis internal is false, the interface name will be
     # interface alias which should be converted to asic port name
     if (enddevice.lower() == hostname.lower()):
-        if ((endport in port_alias_asic_map) and
-                (asic_name.lower() in port_alias_asic_map[endport].lower())):
+        if endport in port_alias_asic_map:
             endport = port_alias_asic_map[endport]
             neighbors[port_alias_map[endport]] = {'name': startdevice, 'port': startport}
             if bandwidth:
                 port_speeds[port_alias_map[endport]] = bandwidth
     elif (startdevice.lower() == hostname.lower()):
-        if ((startport in port_alias_asic_map) and
-                (asic_name.lower() in port_alias_asic_map[startport].lower())):
+        if startport in port_alias_asic_map:
             startport = port_alias_asic_map[startport]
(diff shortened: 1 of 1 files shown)


mlok-nokia:
@judyjoseph @sanmalho-hit,  This PR is required by the OC Test for the asic_port_name change in port_config.ini.  Please review it.  Thanks

judyjoseph (approved):

arlakshm (approved):

judyjoseph:
@yxieca @lguohan please could you help merge this, thanks

PR #13547, opened 2023-01-30
[Dual-ToR] add default value for ACL rule for mellanox platform (Standby ToR)

ayurkiv-nvda:
#### Why I did it
Need to add the possibility to choose between dropping packets (using ACL) on ingress or egress in Dual ToR scenario

#### How I did it
Add new attribute "mux_tunnel_ingress_acl" to SYSTEM_DEFAULTS table

#### How to verify it
check that new attribute exists in redis:
admin@sonic:~$ redis-cli -n 4
127.0.0.1:6379[4]> HGETALL SYSTEM_DEFAULTS|mux_tunnel_ingress_acl
1."state"
2."false"

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [x] 202205
- [x] 202211

### DIFF
--- a/files/build_templates/init_cfg.json.j2
+++ b/files/build_templates/init_cfg.json.j2
@@ -133,5 +133,16 @@
             "digits_class": "true",
             "special_class": "true"
         }
+    },
+    "SYSTEM_DEFAULTS" : {
+{%- if include_mux == "y" %}
+        "mux_tunnel_egress_acl": {
+{%- if sonic_asic_platform == "mellanox" %}
+            "status": "enabled"
+{% else %}
+            "status": "disabled"
+{% endif %}
+        }
+{% endif %}
     }
 }
--- a/src/sonic-yang-models/tests/files/sample_config_db.json
+++ b/src/sonic-yang-models/tests/files/sample_config_db.json
@@ -2109,6 +2109,10 @@
         "SYSTEM_DEFAULTS": {
             "tunnel_qos_remap": {
                 "status": "enabled"
+            },
+            "mux_tunnel_egress_acl":
+            {
+                "status": "enabled"
             }
          },
         "LOSSLESS_TRAFFIC_PATTERN": {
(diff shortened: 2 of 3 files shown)


files/build_templates/init_cfg.json.j2
```diff
@@ -133,5 +133,16 @@
             "digits_class": "true",
             "special_class": "true"
         }
+    },
+    "SYSTEM_DEFAULTS" : {
+{%- if include_mux == "y" %}
```
bingwang-ms, line 138: Is it better to generate the `SYSTEM_DEFAULTS` table only when `sonic_asic_platform == "mellanox"`, and `include_mux == "y"`? In this way, there will be no change for other platforms.

ayurkiv-nvda: Thank you for your remark.
I thought about it previously. But I decided that it may look confusing that **SYSTEM_DEFAULT** table (that is generic table) is under "mellanox" and "mux" condition.
What do you think about it?

liat-grozovik:
@bingwang-ms can you please review and approve?

files/build_templates/init_cfg.json.j2
bingwang-ms: Make sense. Please also consider when we are upgrading from previous version, the `mux_tunnel_egress_acl` is absent. Our code should be handle that exception.

bingwang-ms (approved):

liat-grozovik (approved):

liat-grozovik:
@ayurkiv-nvda please clarify to which branch the fix should be taken? 
and if not cleanly cherry picked please have a separated PR against to requested branch.

ayurkiv-nvda:
> @ayurkiv-nvda please clarify to which branch the fix should be taken? and if not cleanly cherry picked please have a separated PR against to requested branch.

It should be merged to 202205 also 
But it can be merged only after https://github.com/sonic-net/sonic-buildimage/pull/11117  will be merged to 202205

PR #13564, opened 2023-01-31
Add support for platform syncd pre shutdown plugin

anamehra:

#### Why I did it
Vendor platform may require running platform specific pre-shutdown routine before shutting down the syncd process which runs the SAI and vendor sdk instance.

#### How I did it
Added a platform script hook which will be executed if the plugin script is provided by the platform in device/<platform>/plugins/

#### How to verify it
Add platform plugin script and run systemctl stop syncd<>

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012
- [ ] 202106
- [ ] 202111
- [x] 202205
- [x] 202211

### DIFF
--- a/files/scripts/syncd.sh
+++ b/files/scripts/syncd.sh
@@ -105,6 +105,11 @@ function stopplatform1() {
     fi
 
     if [[ x$sonic_asic_platform != x"mellanox" ]] || [[ x$TYPE != x"cold" ]]; then
+        # Invoke platform specific pre shutdown routine.
+        PLATFORM=`$SONIC_DB_CLI CONFIG_DB hget 'DEVICE_METADATA|localhost' platform`
+        PLATFORM_PRE_SHUTDOWN="/usr/share/sonic/device/$PLATFORM/plugins/syncd_pre_shutdown.sh"
+        [ -f $PLATFORM_PRE_SHUTDOWN ] && $PLATFORM_PRE_SHUTDOWN start $DEV
+
         debug "${TYPE} shutdown syncd process ..."
         /usr/bin/docker exec -i syncd$DEV /usr/bin/syncd_request_shutdown --${TYPE}
 
--- a/files/scripts/syncd.sh
+++ b/files/scripts/syncd.sh
@@ -105,8 +105,11 @@ function stopplatform1() {
     fi
 
     if [[ x$sonic_asic_platform != x"mellanox" ]] || [[ x$TYPE != x"cold" ]]; then
-        # Invoke syncd pre shutdown routine
-        /usr/bin/docker exec -i syncd$DEV /usr/bin/syncd_request_pre_shutdown --${TYPE} > /dev/null
+        # Invoke platform specific pre shutdown routine.
+        PLATFORM=`$SONIC_DB_CLI CONFIG_DB hget 'DEVICE_METADATA|localhost' platform`
+        PLATFORM_PRE_SHUTDOWN="/usr/share/sonic/device/$PLATFORM/plugins/syncd_request_pre_shutdown"
+        [ -f $PLATFORM_PRE_SHUTDOWN ] && \
+            /usr/bin/docker exec -i syncd$DEV /usr/share/sonic/platform/plugins/syncd_request_pre_shutdown --${TYPE}
 
         debug "${TYPE} shutdown syncd process ..."
         /usr/bin/docker exec -i syncd$DEV /usr/bin/syncd_request_shutdown --${TYPE}
--- a/files/scripts/syncd.sh
+++ b/files/scripts/syncd.sh
@@ -105,10 +105,8 @@ function stopplatform1() {
     fi
 
     if [[ x$sonic_asic_platform != x"mellanox" ]] || [[ x$TYPE != x"cold" ]]; then
-        # Invoke platform specific pre shutdown routine.
-        PLATFORM=`$SONIC_DB_CLI CONFIG_DB hget 'DEVICE_METADATA|localhost' platform`
-        PLATFORM_PRE_SHUTDOWN="/usr/share/sonic/device/$PLATFORM/plugins/syncd_pre_shutdown.sh"
-        [ -f $PLATFORM_PRE_SHUTDOWN ] && $PLATFORM_PRE_SHUTDOWN start $DEV
+        # Invoke syncd pre shutdown routine
+        /usr/bin/docker exec -i syncd$DEV /usr/bin/syncd_request_pre_shutdown --${TYPE} > /dev/null
 
         debug "${TYPE} shutdown syncd process ..."
         /usr/bin/docker exec -i syncd$DEV /usr/bin/syncd_request_shutdown --${TYPE}


anamehra:
@abdosi , FYI. Thanks

files/scripts/syncd.sh
```diff
@@ -105,6 +105,11 @@ function stopplatform1() {
     fi
 
     if [[ x$sonic_asic_platform != x"mellanox" ]] || [[ x$TYPE != x"cold" ]]; then
+        # Invoke platform specific pre shutdown routine.
+        PLATFORM=`$SONIC_DB_CLI CONFIG_DB hget 'DEVICE_METADATA|localhost' platform`
+        PLATFORM_PRE_SHUTDOWN="/usr/share/sonic/device/$PLATFORM/plugins/syncd_pre_shutdown.sh"
+        [ -f $PLATFORM_PRE_SHUTDOWN ] && $PLATFORM_PRE_SHUTDOWN start $DEV
```
abdosi, line 111: can we make change to have script within syncd and then execute from syncd context. for exampe:  docker exec -i syncd0 /usr/share/sonic/device/plugins/pre_shutdown.sh.  

This way we need not have vendor to force creating script to take start and $DEV as argument

anamehra: Added the script as part of vendor syncd docker.
/usr/bin/syncd_request_pre_shutdown which vendor may provide from syncd docker.

abdosi: @anamehra what if given vendor do not have this script ?

anamehra: if the file is not present, docker throws error which we are redirecting to /dev/null. In case the file is present, error messaging shall be handled by the vendor process.

 docker exec -it syncd0 /usr/bin/test
OCI runtime exec failed: exec failed: container_linux.go:380: starting container process caused: exec: "/usr/bin/test": stat /usrbin/test: no such file or directory: unknown

abdosi (approved):
LGTM @yxieca can you also please review.

abdosi:
@yxieca can you please review this

files/scripts/syncd.sh
```diff
@@ -105,6 +105,9 @@ function stopplatform1() {
     fi
 
     if [[ x$sonic_asic_platform != x"mellanox" ]] || [[ x$TYPE != x"cold" ]]; then
+        # Invoke syncd pre shutdown routine
+        /usr/bin/docker exec -i syncd$DEV /usr/bin/syncd_request_pre_shutdown --${TYPE} > /dev/null
```
yxieca, line 109: Which PR introduced /usr/bin/syncd_request_pre_shutdown? Calling a non-existing script will cause failure.

anamehra: If the script is not present, the command just throws an error which we redirect to /dev/null.
It should not cause script failure.

yxieca (changes requested):
Seems like calling a non-existing script.

files/scripts/syncd.sh
anamehra: root@sonic:/home/abc# /usr/bin/docker exec -i syncd0 ls /usr/bin/syncd_request_pre_shutdown
ls: cannot access '/usr/bin/syncd_request_pre_shutdown': No such file or directory           <<<<< **file not present**
root@sonic:/home/abc# 
root@sonic:/home/abc# /usr/bin/docker exec -i syncd0 /usr/bin/syncd_request_pre_shutdown
OCI runtime exec failed: exec failed: container_linux.go:380: starting container process caused: exec: "/usr/bin/syncd_request_pre_shutdown": stat /usr/bin/syncd_request_pre_shutdown: no such file or directory
root@sonic:/home/abc# 

root@sonic:/home/abc# /usr/bin/docker exec -i syncd0 /usr/bin/syncd_request_pre_shutdown --cold > /dev/null
root@sonic:/home/abc#

anamehra: We have the required script added on Cisco platforms, other platforms may add it if required. if not present, it will be a no op.

anamehra:
> Seems like calling a non-existing script.

Shall I put Cisco platform specific check here:
[[ x$sonic_asic_platform != x"cisco-8000" ]]

files/scripts/syncd.sh
yxieca: Suggest:

1. call this script from platform plugin directory instead of /usr/bin.
2. Check if the script exists before calling.

yxieca:
@anamehra What is the impact of this change to fast/warm reboot?

anamehra:
> @anamehra What is the impact of this change to fast/warm reboot?

This should not impact the fast-reboot/warm-reboot. This script is called as part of systemd shutting down all services.
I can add Cisco-8000 specific check to run only on Cisco platforms and also it will execute the script only if present.

I fast-reboot/warm-reboot script calls the syncd_request_shutdown as part of pre-shutdown request. I am working on a change to call this pre-shutdown script there as well to help with warm reboots on Cisco platform.

files/scripts/syncd.sh
anamehra: There is no platform plugin available in this path before syncd_request_shutdown. I can add a plugin as in here if that is acceptable:
https://github.com/sonic-net/sonic-buildimage/pull/13564/commits/2288dbcd075dc54451b7b761d054cfc8ce637093

anamehra: @yxieca , please let me know if the following is fine which does the script check and run only if available:
        PLATFORM=`$SONIC_DB_CLI CONFIG_DB hget 'DEVICE_METADATA|localhost' platform`
        PLATFORM_PRE_SHUTDOWN="/usr/share/sonic/device/$PLATFORM/plugins/syncd_pre_shutdown.sh"
        [ -f $PLATFORM_PRE_SHUTDOWN ] && $PLATFORM_PRE_SHUTDOWN
        
 If you and @abdosi agree, I will update the PR

abdosi: this show work correct ?

`
PLATFORM_PRE_SHUTDOWN="/usr/share/sonic/device/$PLATFORM/plugins/syncd_pre_shutdown.sh"
[ -f $PLATFORM_PRE_SHUTDOWN ] && /usr/bin/docker exec -i syncd$DEV //usr/share/sonic/platform/plugins/syncd_request_pre_shutdown --${TYPE} `

anamehra: yes @abdosi , if @yxieca agrees, I can create the platform plugin like this or call PLATFORM_PRE_SHUTDOWN and let that handle the docker call.

abdosi: ok, please process with above change . it handle @yxieca  request and also we make sure to call from docker context.

anamehra: @abdosi , updated the code, please check.

abdosi (approved):
LGTM

abdosi:
@yxieca can you please review again

yxieca (approved):

anamehra:
Hi @abdosi , please rerun the pipeline. I see some issue suggesting Lock testbed failed.

anamehra:
Hi @abdosi , please help with merging this PR.

abdosi:
@yxieca / @rlhui / @prsunny  please help merge.

PR #13587, opened 2023-02-02
Use tmpfs for /var/log for Arista 7260

saiarcot895:

#### Why I did it

This is to reduce writes to disk, which might cause the SSD to get worn out faster.

#### How to verify it

Loaded it on an Arista 7260CX3, verified that /var/log was backed by tmpfs instead of disk.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012
- [ ] 202106
- [ ] 202111
- [x] 202205
- [x] 202211

### DIFF
--- a/files/Aboot/boot0.j2
+++ b/files/Aboot/boot0.j2
@@ -502,6 +502,7 @@ write_platform_specific_cmdline() {
     if [ "$sid" = "Gardena" ] || [ "$sid" = "GardenaE" ]; then
         aboot_machine=arista_7260cx3_64
         flash_size=28000
+        cmdline_add logs_inram=on
     fi
     if [ "$sid" = "Alhambra" ]; then
         aboot_machine=arista_7170_64c


prgeor (approved):

files/Aboot/boot0.j2
```diff
@@ -502,6 +502,7 @@ write_platform_specific_cmdline() {
     if [ "$sid" = "Gardena" ] || [ "$sid" = "GardenaE" ]; then
         aboot_machine=arista_7260cx3_64
         flash_size=28000
+        cmdline_add logs_inram=on
```
yxieca, line 505: What is the partition size? We might need to reduce the partition size when putting it on tmpfs

saiarcot895: It will be just under 800MB. Prior to this, on disk, it was close to 4GB.

yxieca: 800MB was calculated at run time?

saiarcot895: Yes, calculated between 5% and 10% of total RAM (unless there's an override for a certain hwsku in `files/initramfs-tools/union-mount.j2`).

yxieca (approved):

qiluo-msft:

PR #13595, opened 2023-02-02
[DX010 platform] fix dx010 platform testcase issues

qnos:
#### Why I did it
1. fix chassis test_set_fans_led case
2. fix chassis get_name case mismatch issue
3. fix fan_drawer test_set_fans_speed
4. fix component test_components test case

#### How I did it
Add corresponding configuration into chassis json file

#### How to verify it
Run platform tests cases to verify these failure cases

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012
- [x] 202106
- [x] 202111
- [x] 202205
- [x] 202211

### DIFF
--- a/device/celestica/x86_64-cel_seastone-r0/platform.json
+++ b/device/celestica/x86_64-cel_seastone-r0/platform.json
@@ -1,6 +1,6 @@
 {
     "chassis": {
-        "name": "Celestica-DX010-C32",
+        "name": "DX010",
         "status_led": {
             "controllable": true,
             "colors": ["green", "off"]
@@ -17,6 +17,9 @@
             },
             {
                 "name": "CPLD4"
+            },
+            {
+                "name": "CPLD5"
             },
             {
                 "name": "BIOS"
@@ -123,10 +126,22 @@
                 },
                 "fans": [
                     {
-                        "name": "FAN-1F"
+                        "name": "FAN-1F",
+                        "status_led": {
+                            "controllable": false
+                        },
+                        "speed": {
+                            "controllable": false
+                        }
                     },
                     {
-                        "name": "FAN-1R"
+                        "name": "FAN-1R",
+                        "status_led": {
+                            "controllable": false
+                        },
+                        "speed": {
+                            "controllable": false
+                        }
                     }
                 ]
             },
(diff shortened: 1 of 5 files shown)


assrinivasan (approved):

prgeor:
@qnos none of the fans are controllable on DX010? How is the platform then doing the thermal cooling as temperature warm up?

qnos:
> @qnos none of the fans are controllable on DX010? How is the platform then doing the thermal cooling as temperature warm up?

DX010 take the automatic thermal control policy. It uses the thermal sensor temperature to automatically adjust the fan speed, so I set the fan to not controllable.

assrinivasan:
@prgeor please help unblock this merge, thanks

prgeor:
> > @qnos none of the fans are controllable on DX010? How is the platform then doing the thermal cooling as temperature warm up?
> 
> DX010 take the automatic thermal control policy. It uses the thermal sensor temperature to automatically adjust the fan speed, and manually change the fan speed through fan API would be override by fancontrol daemon, so I set the fan to not controllable.

@qnos I see your platform has the API to set the fan speed [here](https://github.com/sonic-net/sonic-buildimage/blob/master/device/celestica/x86_64-cel_seastone-r0/sonic_platform/fan.py#L203) Why skip this test to check the fan set_speed API? Can you adhere to [this](https://github.com/sonic-net/SONiC/blob/master/doc/platform_json_enhancement.md) HLD

ikkzhu:
> > > @qnos none of the fans are controllable on DX010? How is the platform then doing the thermal cooling as temperature warm up?
> > 
> > 
> > DX010 take the automatic thermal control policy. It uses the thermal sensor temperature to automatically adjust the fan speed, and manually change the fan speed through fan API would be override by fancontrol daemon, so I set the fan to not controllable.
> 
> @qnos I see your platform has the API to set the fan speed [here](https://github.com/sonic-net/sonic-buildimage/blob/master/device/celestica/x86_64-cel_seastone-r0/sonic_platform/fan.py#L203) Why skip this test to check the fan set_speed API? Can you adhere to [this](https://github.com/sonic-net/SONiC/blob/master/doc/platform_json_enhancement.md) HLD

@prgeor Yes, DX010 provide the set_speed API. But as I described, DX010 takes the automatic fan control policy. Only if we disable all fancontrol daemons in pmon, otherwise the speed value set through set_speed api would be overrided by fancontrol daemon, and cause the speed value from get api not match what we set. So I set the fans not controllable in chassis.py.

prgeor:
@lguohan please help merge

prgeor:
@StormLiangMS @yxieca please cherry pick

PR #13611, opened 2023-02-03
[Mellanox] set select timeout to no more than 1 sec to make sure fast shutdown

Junchao-Mellanox:
#### Why I did it

Commit https://github.com/sonic-net/sonic-platform-daemons/commit/153ea47e6ecddf750d67978f6704f832afea0d5f changed SfpStateUpdateTask from Process to Thread. In this commit, it raises an exception in SfpStateUpdateTask to make shutdown flow fast. But it does not work on Nvidia platform as Nvidia platform is passing timeout parameter of get_change_event to select. Linux select function can not be interrupted by a Python exception. There is no such issue on Nvidia platform before that commit. However, in order to comply with the commit and make shutdown flow fast, we decided to change Nvidia platform API implementation.

To fix issue https://github.com/sonic-net/sonic-buildimage/issues/13591.

#### How I did it

1. The select call in get_change_event should use no more than 1 second as timeout parameter. 
2. Outside the select call, add a while loop to make sure timeout parameter of get_change_event work as expected

#### How to verify it

Manual test

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [x] 202205
- [x] 202211

### DIFF
--- a/platform/mellanox/mlnx-platform-api/sonic_platform/chassis.py
+++ b/platform/mellanox/mlnx-platform-api/sonic_platform/chassis.py
@@ -31,10 +31,10 @@ try:
     from . import utils
     from .device_data import DeviceDataManager
     import re
+    import time
 except ImportError as e:
     raise ImportError (str(e) + "- required module not found")
 
-MAX_SELECT_DELAY = 3600
 
 RJ45_TYPE = "RJ45"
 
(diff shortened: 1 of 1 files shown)


prgeor:
@Junchao-Mellanox do you still want to keep in draft?

liat-grozovik (approved):

liat-grozovik:
@keboliu  please review

platform/mellanox/mlnx-platform-api/sonic_platform/chassis.py
```diff
@@ -387,26 +387,30 @@ def get_change_event(self, timeout=0):
             self.sfp_event.initialize()
 
         wait_for_ever = (timeout == 0)
+        # select timeout should be no more than 1000ms to ensure fast shutdown flow
```
prgeor, line 390: @Junchao-Mellanox may be add more context in the comment why max timeout is kept 1sec?

platform/mellanox/mlnx-platform-api/sonic_platform/chassis.py
```diff
@@ -387,26 +387,30 @@ def get_change_event(self, timeout=0):
             self.sfp_event.initialize()
 
         wait_for_ever = (timeout == 0)
```
prgeor, line 389: not sure if you still want to keep this timeout=0 for future use

prgeor (approved):
approved with comments

PR #13613, opened 2023-02-03
Set 'origin' and 'AS Path' for T1 SLB routes

jcaiMR:
#### Why I did it
Add support for Set 'origin' and 'AS Path' for T1 SLB routes.

When SLB programs routes to T1 (overlay routes), T1 must advertise such routes to match with the existing SLB 'origin' and AS path length. Following changes to be done during T1 advertisement:

'origin' set to 'incomplete' (value 2)
AS path to be appended as "x" + "x" (two prepend) to match the current AS path at T1 layer

#### How I did it
Add origin and as-path prepend for routemap FROM_SDN_SLB_ROUTES.

#### How to verify it
Follow test steps in HLD.
image
image

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211

### DIFF
--- a/src/sonic-bgpcfgd/bgpcfgd/managers_rm.py
+++ b/src/sonic-bgpcfgd/bgpcfgd/managers_rm.py
@@ -75,11 +75,11 @@ class RouteMapMgr(Manager):
     def __update_rm(self, rm, data):
         cmds = []
         cmds.append("route-map %s permit 100" % ("%s_RM" % rm))
-        cmds.append(" set community %s" % data["community_id"])
-        cmds.append(" set origin incomplete")
         bgp_asn = \
             self.directory.get_slot("CONFIG_DB", swsscommon.CFG_DEVICE_METADATA_TABLE_NAME)['localhost']['bgp_asn']
         cmds.append(" set as-path prepend %s %s" % (bgp_asn, bgp_asn))
+        cmds.append(" set community %s" % data["community_id"])
+        cmds.append(" set origin incomplete")
 
         log_debug("BGPRouteMapMgr:: update route-map %s community %s origin incomplete as-path prepend %s %s" % \
                   ("%s_RM" % rm, data["community_id"], bgp_asn, bgp_asn))
--- a/src/sonic-bgpcfgd/bgpcfgd/managers_rm.py
+++ b/src/sonic-bgpcfgd/bgpcfgd/managers_rm.py
@@ -1,4 +1,5 @@
 from .manager import Manager
+from swsscommon import swsscommon
 from .log import log_err, log_debug
 
 ROUTE_MAPS = ["FROM_SDN_SLB_ROUTES"]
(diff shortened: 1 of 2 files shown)


src/sonic-bgpcfgd/bgpcfgd/managers_rm.py
```diff
@@ -72,7 +73,15 @@ def __del_handler_validate(self, key):
         return True
 
     def __update_rm(self, rm, data):
-        cmds = ["route-map %s permit 100" % ("%s_RM" % rm), " set community %s" % data["community_id"]]
-        log_debug("BGPRouteMapMgr:: update route-map %s community %s" % ("%s_RM" % rm, data["community_id"]))
+        cmds = []
```
StormLiangMS, line 76: could we add a check if it is FROM_SDN_SLB_ROUTES, to implement these attributes? by this way, we can make it more generic.

jcaiMR: There is already a check in set_handler() -> __set_handler_validate(), to make sure the key is "FROM_SDN_SLB_ROUTES". Is it possible in future that __set_handler_validate() will allow more route map name string ?

StormLiangMS: @jcaiMR yes, we could extend routemap for others too.

jcaiMR: ok，will apply in next diff.

StormLiangMS (approved):
LGTM

src/sonic-bgpcfgd/bgpcfgd/managers_rm.py
```diff
@@ -72,7 +73,16 @@ def __del_handler_validate(self, key):
...
     def __update_rm(self, rm, data):
-        cmds = ["route-map %s permit 100" % ("%s_RM" % rm), " set community %s" % data["community_id"]]
-        log_debug("BGPRouteMapMgr:: update route-map %s community %s" % ("%s_RM" % rm, data["community_id"]))
-        self.cfg_mgr.push_list(cmds)
+        cmds = []
+        if rm is "FROM_SDN_SLB_ROUTES":
+            cmds.append("route-map %s permit 100" % ("%s_RM" % rm))
+            bgp_asn = \
+                self.directory.get_slot("CONFIG_DB", swsscommon.CFG_DEVICE_METADATA_TABLE_NAME)['localhost']['bgp_asn']
+            cmds.append(" set as-path prepend %s %s" % (bgp_asn, bgp_asn))
```
prsunny, line 81: Prefer to have a fixed value for AS prepend instead of prepending the current device's AS value. This could be gotten from /etc/sonic/constants.yml : 2. Not sure if this is significant changes.

prsunny (approved):

prsunny:
@jcaiMR , likely to have conflict for 202012. Would you raise another PR for 202012?

jcaiMR:
> @jcaiMR , likely to have conflict for 202012. Would you raise another PR for 202012? @siqbal1986 for viz

Sure, I will open the PR today.

jcaiMR:
> @jcaiMR , likely to have conflict for 202012. Would you raise another PR for 202012? @siqbal1986 for viz

 @prsunny Hi Prince, I checked the code and also tried this diff with 202012 base, there is no gap from code. Maybe we can still waiting for the  'squash merge' into 202012.

src/sonic-bgpcfgd/bgpcfgd/managers_rm.py
jcaiMR: updated.

prsunny:
@jcaiMR , since this is a dependency on constants.yml, lets ensure a sonic-mgmt test is added so that any future change wont break this implementation.

jcaiMR:
> @jcaiMR , since this is a dependency on constants.yml, lets ensure a sonic-mgmt test is added so that any future change wont break this implementation.

Sure, Storm has write some test case for this SDN SLB routes and ready to PR, I will add some extra test cases for this commit.

PR #13726, opened 2023-02-09
[ci] Kill hanged docker build process to avoid build timeout issue.

liushilongbuaa:
#### Why I did it
Docker build has a low rate of hanging up.
It hangs on different steps. So, it looks like a bug in docker daemon.
#### How I did it
Start a daemon process to scan running time more than 1 hours, and kill the process.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211


.azure-pipelines/azure-pipelines-build.yml
```diff
@@ -110,6 +110,25 @@ jobs:
     buildSteps:
       - template: template-skipvstest.yml
       - bash: |
+          (
+            while true
+            do
+              sleep 120
+              now=$(date +%s)
+              pids=$(ps aux | grep -v grep | grep -E "^.{,100}docker build" | awk '{print$2}')
```
xumia, line 118: You want to list the docker PIDs, right? A small improvement.
ps -C docker -o pid,args | grep "docker build"

liushilongbuaa: I think 'ps -C' is good. But etime is not good to use.
etime has a format '[[dd-]hh:]mm:ss'. If I want to compare it with 1 hour, I need to compare the etime length.

.azure-pipelines/azure-pipelines-build.yml
```diff
@@ -110,6 +110,25 @@ jobs:
...
       - bash: |
+          (
+            while true
+            do
+              sleep 120
+              now=$(date +%s)
+              pids=$(ps -C docker -o pid,etime,args | grep "docker build" | cut -d" " -f1)
+              for pid in $pids
+              do
+                start=$(date --date="$(ls -dl /proc/$pid --time-style full-iso | awk '{print$6,$7}')" +%s)
```
xumia, line 121: ps -o "lstart" $pid

xumia (approved):

.azure-pipelines/azure-pipelines-build.yml
```diff
@@ -110,6 +110,25 @@ jobs:
     buildSteps:
       - template: template-skipvstest.yml
       - bash: |
+          (
```
xumia, line 113: Can we add the step in a template, and the release branch use the same template?

.azure-pipelines/template-daemon.yml
```diff
@@ -0,0 +1,24 @@
...
+            echo =========== $(date +%F%T) $time_s &>> target/daemon.log
+            ps $pid &>> target/daemon.log
+            sudo kill $pid
+          fi
+        done
+      done
+    ) &
+    daemon_pid=$!
+    ps $daemon_pid
+    echo "##vso[task.setvariable variable=DAEMON_PID]$daemon_pid"
```
xumia, line 23: Maybe the variable name is too common, can we give a more specific name? For example, PIPELINE_WATCHDOG_DEMON_PID?

.azure-pipelines/template-stop-daemon.yml
```diff
@@ -0,0 +1,5 @@
+steps:
+- bash: |
```
xumia, line 2: Is it possible to exit the daemon automatically?

.azure-pipelines/template-stop-daemon.yml
```diff
@@ -0,0 +1,5 @@
...
+    ps $(DAEMON_PID)
+    sudo kill $(DAEMON_PID) || true
+  displayName: kill daemon process
```
xumia, line 5: Always()?

.azure-pipelines/template-stop-daemon.yml
liushilongbuaa: If we don't exit the daemon. Each job will start one daemon. It looks not secure.
cleanup is used in sonic-mgmt, sonic-slave and sonic-buildimage pipeline.

xumia (approved):

PR #13786, opened 2023-02-13
[Build] Change the default mirror version config file

xumia:
#### Why I did it
Change the mirror config file
Use the files/build/versions/default/versions-mirror only when reproducible build enabled.
The config in files/build/versions is only for reproducible build, while snapshot mirror feature does not have the dependency on the reproducible build.

#### How I did it
Skip the mirror config in files/build/versions/default/versions-mirror if reproducible build not enabled.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [x] 202205
- [x] 202211

### DIFF
--- a/scripts/build_mirror_config.sh
+++ b/scripts/build_mirror_config.sh
@@ -6,7 +6,8 @@ export ARCHITECTURE=$2
 export DISTRIBUTION=$3
 
 DEFAULT_MIRROR_URL_PREFIX=http://packages.trafficmanager.net
-MIRROR_VERSION_FILE=files/build/versions/default/versions-mirror
+MIRROR_VERSION_FILE=
+[[ "$SONIC_VERSION_CONTROL_COMPONENTS" == *deb* || $SONIC_VERSION_CONTROL_COMPONENTS == *all* ] && MIRROR_VERSION_FILE=files/build/versions/default/versions-mirror
 [ -f target/versions/default/versions-mirror ] && MIRROR_VERSION_FILE=target/versions/default/versions-mirror
 
 # The default mirror urls
--- a/scripts/build_mirror_config.sh
+++ b/scripts/build_mirror_config.sh
@@ -7,7 +7,7 @@ export DISTRIBUTION=$3
 
 DEFAULT_MIRROR_URL_PREFIX=http://packages.trafficmanager.net
 MIRROR_VERSION_FILE=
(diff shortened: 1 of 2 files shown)


scripts/build_mirror_config.sh
```diff
@@ -6,7 +6,8 @@ export ARCHITECTURE=$2
 export DISTRIBUTION=$3
 
 DEFAULT_MIRROR_URL_PREFIX=http://packages.trafficmanager.net
-MIRROR_VERSION_FILE=files/build/versions/default/versions-mirror
+MIRROR_VERSION_FILE=
+[[ "$SONIC_VERSION_CONTROL_COMPONENTS" == *deb* || $SONIC_VERSION_CONTROL_COMPONENTS == *all* ]] && MIRROR_VERSION_FILE=files/build/versions/default/versions-mirror
 [ -f target/versions/default/versions-mirror ] && MIRROR_VERSION_FILE=target/versions/default/versions-mirror
```
StormLiangMS, line 11: @xumia if target/versions/default/versions-mirror exists, MIRROR_VERSION_FILE will be set to target/versions/default/versions-mirror anyway? It will overwrite the MIRROR_VERSION_FILE set by "MIRROR_VERSION_FILE=
[[ "$SONIC_VERSION_CONTROL_COMPONENTS" == *deb* || $SONIC_VERSION_CONTROL_COMPONENTS == *all* ]] && MIRROR_VERSION_FILE=files/build/versions/default/versions-mirror"

xumia: Yes, it is as expected, the file files/build/versions/default/versions-mirror will be copied to target folder if not exist.

StormLiangMS: @xumia can I understand before this change, files/build/versions/default/versions-mirror will be used by default. With the change, it would be optional, and target/versions/default/versions-mirror has higher priority than files/build/versions/default/versions-mirror?

xumia: The file target/versions/default/versions-mirror has higher priority than files/build/versions/default/versions-mirror, the behaviors is not changed.

files/build/versions/default/versions-mirror only used when reproducible enabled, because files/build/versions/ is designed for reproducible build.

StormLiangMS (approved):
LGTM

xumia:
@yxieca , please help approve for 202205, thanks.

PR #13789, opened 2023-02-13
[master]staticroutebfd process implementation

qiwang4:
#### Why I did it

To enable the BFD for static route

HLD: https://github.com/sonic-net/SONiC/pull/1216

#### How I did it

create a new process to handle static route with "bfd" tag

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [x] 202205
- [x] 202211


prsunny:
Can you please link the HLD for this feature to this PR?

jamesan47:
Hi @prsunny, @lguohan, @StormLiangMS, Cisco folks are not able to add other reviewers to this PR.  Could you help add @prsunny, @abdosi , Shahzad and Baorong to this PR?  Thanks!

abdosi:
cc @rlhui for viz

abdosi:
@jamesan47 what is Boarang github id ?

jamesan47:
@abdosi , Baorong's github ID is baorliu.

rlhui:
please correct spelling of the PR title

lguohan:
since the bgpcfgd already manages the static route, why do we introduce a new process to manage the bfd session for static route?

baorliu:
> 
the staticroutebfd process manages next hop reachability for static routes. it still use StaticRouteMgr to install/update static route by updating STATIC_ROUTE table in APPL_DB. 
we try to decouple it from the existing process, make the maintenance easier,  less interference on each other when design changes, also have the flexibility to run this process somewhere else (where can access redis DB). (That is the reason we don't use log package from bgpcfgd ).

src/sonic-bgpcfgd/staticroutebfd/main.py
```diff
@@ -0,0 +1,691 @@
+import os
```
venkatmahalingam, line 1: Any reason for adding this static route bfd under sonic-bgpcfgd folder? I believe, this is the generic module and no dependency on bgpcfgd. if you keep this module generic and keep it outside of sonic-bgpcfgd (may be one of the host services), we can make use of this for sonic-frr-mgmt-framework/frrcfgd as well.

src/sonic-bgpcfgd/staticroutebfd/main.py
```diff
@@ -0,0 +1,691 @@
...
+        :param key: key to split
+        :return: table name, vrf name, interface name and peer ip extracted from the key
+        """
+        if key.count("|") < 2:
+            return 'default', 'default', key
+        else:
+            return tuple(key.split('|'))
+
+    def append_to_sr_table_entry(self, sr_key, nh_info):
+        entry = self.get_local_db(LOCAL_SR_TABLE, sr_key)
```
venkatmahalingam, line 490: SR_TABLE seems like segment routing table, suggest to change the table name.

src/sonic-bgpcfgd/staticroutebfd/main.py
baorliu: the StaticRouteBfd design watches and updates redis db tables only, it could be in anywhere in the system where can access redis db tables. 
Yes. this module has no dependency on bgpcfgd. we need to put it into a container to start (restart in the case of crash)  the process. the reason we put the code under sonic-bgpcfgd just because the logic is related. 
(originally we were planning to design the new component as one of the "manager" in the runner/manager framework. but later we found that the current manager watches one table only, but the StaticRouteBfd needs to watch both config db and state db. so we moved it out, run as a separate process.)

abdosi:
can we have UT result here also ?

dockers/docker-fpm-frr/frr/supervisord/supervisord.conf.j2
```diff
@@ -141,6 +141,20 @@ stderr_logfile=syslog
 dependent_startup=true
 dependent_startup_wait_for=bgpd:running
 
+{% if DEVICE_METADATA.localhost.frr_mgmt_framework_config is defined and DEVICE_METADATA.localhost.frr_mgmt_framework_config == "true" %}
```
abdosi, line 144: why we have this if condition ? seems completely unrelated

dockers/docker-fpm-frr/frr/supervisord/supervisord.conf.j2
```diff
@@ -141,6 +141,20 @@ stderr_logfile=syslog
...
+{% else %}
+[program:staticroutebfd]
+command=/usr/local/bin/staticroutebfd
+priority=6
```
abdosi, line 148: is the priority and dependent_startup_wait_for setup correctly here ? Just want to make sure we have thought of any race-condition that can occur

src/sonic-bgpcfgd/staticroutebfd/main.py
abdosi: also in multi-asic where we have multiple bgp docker instance it aligns with that model

src/sonic-bgpcfgd/staticroutebfd/main.py
```diff
@@ -0,0 +1,750 @@
...
+        self.set_local_db(LOCAL_INTERFACE_TABLE, if_name, value)
+
+    def find_interface_ip(self, if_name, ip_example):
+        valid, is_ipv4, nh = check_ip(ip_example)
+        if not valid:
+            return False, ""
+
+        value = self.get_local_db(LOCAL_INTERFACE_TABLE, if_name)
+        ip = value.get(is_ipv4, "")
+        if len(ip)>0: #ip should be verified when add to local_db
```
abdosi, line 213: how do we make sure we process Static Route Config after processing Interface/Portchannel Interface table ?
it seems there is no dependency we have created ? for chassis we don;t want to use Loopback4096 or Loopback0 . if we don;t find ip in interface table can we put setting up that BFD session in hold state (in cache) and can we do retry once it is available ?

src/sonic-bgpcfgd/staticroutebfd/main.py
```diff
@@ -0,0 +1,750 @@
...
+            return True, ip
+
+        #if there is no ip for the interface, try loopback ip address
+        value = self.get_local_db(LOCAL_INTERFACE_TABLE, LOOPBACK4096_NAME)
```
abdosi, line 217: if intf_name is there we should not use Loopback4096/Loopback0 . We can add this logic as if..else

abdosi (changes requested):
As per comments.

src/sonic-bgpcfgd/staticroutebfd/main.py
baorliu: see the reconciliation function, we process interface table first before process static route table. after that, it will be event driven.

src/sonic-bgpcfgd/staticroutebfd/main.py
baorliu: see the above line# 211, if we can get IP from intf_name, will return that IP.

dockers/docker-fpm-frr/frr/supervisord/supervisord.conf.j2
baorliu: see the line# 128, because staticroutebfd need to work together with bgpcfgd, so use same condition with bgpcfgd

dockers/docker-fpm-frr/frr/supervisord/supervisord.conf.j2
baorliu: Here try to use same priority and dependency with program bgpcfgd program, because the static route update need to go through bgpd eventually (staticroutebfd update redis_db APPL_DB STATIC_ROUTE_TABLE, and bgpcfgd handle the table update and update static route through bgpd).

src/sonic-bgpcfgd/staticroutebfd/main.py
baorliu: updated.

src/sonic-bgpcfgd/staticroutebfd/main.py
baorliu: changed the design. wait for interface ip, not use loopback ip address.

abdosi (approved):
LGTM.

abdosi:
@prsunny can you please help with review/sign-off this PR.

abdosi:
@prsunny can you please help with review/sign-off this PR.

src/sonic-bgpcfgd/bgpcfgd/managers_static_rt.py
```diff
@@ -60,19 +71,47 @@ def set_handler(self, key, data):
...
         else:
-            log_debug("Nothing to update for static route {}".format(key))
+            log_debug("{} Nothing to update for static route {}".format(self.db_name, key))
 
         self.static_routes.setdefault(vrf, {})[ip_prefix] = (ip_nh_set, route_tag)
 
         return True
 
 
+    def skip_appl_del(self, vrf, ip_prefix):
```
prsunny, line 83: May be it is good to provide some comments as to what this function does and why it is skipped for some cases.

prsunny (approved):

src/sonic-bgpcfgd/bgpcfgd/managers_static_rt.py
prsunny: @baorliu , would you please address this for completion?

baorliu: > 

will update it soon...

abdosi:
@baorliu can you please resolve merge conflict ?

baorliu:
> @baorliu can you please resolve merge conflict ?

Done. thanks

prsunny (approved):

abdosi:
@gechiang needed for msft repo 202205 branch

abdosi:
@baorliu can you create direct PR for 202205 branch on msft repo ?

That way we need not have to cherry-pick.

baorliu:
> 

Sure. will do it, and  will include one fix https://github.com/sonic-net/sonic-buildimage/pull/15269

PR #13792, opened 2023-02-13
[Mellanox] Non upstream patches for hw-mgmt V.4.0020.4104

stephenxs:
#### Why I did it

Add non-upstream kernel patches for the Nvidia platforms
These patches are not yet upstream but needed for new technology.
A flow to upstream them is in progress and once they will be approved they will be moved officially to sonic-linux-kernel.
Till then to include them in the build (not must) the build option `INCLUDE_EXTERNAL_PATCH_TAR=y` should be included

#### How I did it

Zip all the patches in to a `tar.gz` tarball.

#### How to verify it

Manually test

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [x] 202211


stephenxs:
> I'd suggest either overriding this flag https://github.com/sonic-net/sonic-buildimage/blob/master/platform/mellanox/rules.mk#L55, or zip the patches.tar to patches.tar.gz and thus we don't have to provide EXTERNAL_KERNEL_PATCH_TAR while building, Also did you test compiling the kernel, did the patches apply without any conflicts?

Hi
Thanks for reviewing it.
I did test to compile it and manually checked that all the patches had been successfully applied.
Regarding the tarball, what's the difference between `.tar` and `.tar.gz`?

dgsudharsan:
> > I'd suggest either overriding this flag https://github.com/sonic-net/sonic-buildimage/blob/master/platform/mellanox/rules.mk#L55, or zip the patches.tar to patches.tar.gz and thus we don't have to provide EXTERNAL_KERNEL_PATCH_TAR while building, Also did you test compiling the kernel, did the patches apply without any conflicts?
> 
> Hi Thanks for reviewing it. I did test to compile it and manually checked that all the patches had been successfully applied. Regarding the tarball, what's the difference between `.tar` and `.tar.gz`?

Hi @stephenxs  I believe if you have .tar.gz you need not have to override EXTERNAL_KERNEL_PATCH_TAR  during build and just setting the flag INCLUDE_EXTERNAL_PATCH_TAR=y should be sufficient since the default value as Vivek pointed out is below
override EXTERNAL_KERNEL_PATCH_TAR := $(BUILD_WORKDIR)/$(PLATFORM_PATH)/non-upstream-patches/patches.tar.gz

I recommend updating it to gz for simplicity during building

dgsudharsan (changes requested):
Please update tar to tar.gz

stephenxs:
> > > I'd suggest either overriding this flag https://github.com/sonic-net/sonic-buildimage/blob/master/platform/mellanox/rules.mk#L55, or zip the patches.tar to patches.tar.gz and thus we don't have to provide EXTERNAL_KERNEL_PATCH_TAR while building, Also did you test compiling the kernel, did the patches apply without any conflicts?
> > 
> > 
> > Hi Thanks for reviewing it. I did test to compile it and manually checked that all the patches had been successfully applied. Regarding the tarball, what's the difference between `.tar` and `.tar.gz`?
> 
> Hi @stephenxs I believe if you have .tar.gz you need not have to override EXTERNAL_KERNEL_PATCH_TAR during build and just setting the flag INCLUDE_EXTERNAL_PATCH_TAR=y should be sufficient since the default value as Vivek pointed out is below override EXTERNAL_KERNEL_PATCH_TAR := (BUILDWORKDIR)/(PLATFORM_PATH)/non-upstream-patches/patches.tar.gz
> 
> I recommend updating it to gz for simplicity during building

I see. I think 

> EXTERNAL_KERNEL_PATCH_TAR

Got your point. I thought `tar -xf` doesn't work for `.gar.gz` file. Now I understand it does work.
Thank you for clarifying.

stephenxs:
> Please update tar to tar.gz

Updated.

dgsudharsan (approved):

vivekrnv (approved):

liat-grozovik (approved):

PR #13810, opened 2023-02-15
Update check of IP_TYPE from ACL Yang model

bingwang-ms:
#### Why I did it
This PR is to update the check of `IP_TYPE` from `sonic-acl.yang`.
It's because if the ACL rule is added by loading a json file with acl-loader, there is no `IP_TYPE` for ACL rule. If such rule exists in `ACL_RULE` table, the GCU (generic config updater) refuses to update any ACL rules because the existing one is invalid.
 This PR updates the yang model for ACL. If the `IP_TYPE` leaf doesn't exist, then we don't check the field.
#### How I did it
Accept the rule if `IP_TYPE` is absent.

#### How to verify it
The change is verified by UT.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211

Remove check of IP_TYPE from ACL Yang model


wen587 (approved):

ganglyu (approved):

ganglyu:
@bingwang-ms, can you check the failed test?

------------------- Test 636: Configure IP_TYPE as ARP and DST_IPV6 in ACL_RULE.---------------------
libyang[0]: Missing required element "lanes" in "PORT_LIST". (path: /sonic-port:sonic-port/PORT/PORT_LIST[name='Ethernet0'])
INFO:YANG-TEST:Missing required element "lanes" in "PORT_LIST".
ERROR:YANG-TEST: Exception >Mismatch ['When condition', 'not satisfied', 'IP_TYPE'] and Missing required element "lanes" in "PORT_LIST".< in /sonic/src/sonic-yang-models/tests/yang_model_tests/test_yang_model.py:213
INFO:YANG-TEST:Configure IP_TYPE as ARP and DST_IPV6 in ACL_RULE. Failed

INFO:YANG-TEST:

bingwang-ms:
> 

Checking

src/sonic-yang-models/yang-templates/sonic-acl.yang.j2
```diff
@@ -117,7 +116,6 @@ module sonic-acl {
 					}
 
 					case ip6_prefix {
-						when "boolean(IP_TYPE[.='ANY' or .='IP' or .='IPV6' or .='IPv6ANY'])";
```
praveen-li, line 120: without this condition, user can apply ip6_prefix with IP_TYPE as IPv4.

So is it possible to modify the condition such that if IP_TYPE does not exist, this results in True.
If exists then it check the value as today. 

Let me know if that is not possible then I will approve this PR with current Changes itself. Thanks

bingwang-ms: It's a good idea. I will investigate the reply shortly.

bingwang-ms: @praveen-li  I updated the yang model. Now if IP_TYPE does not exist, this results in True. Please help review. Thanks

wen587:
LGTM

qiluo-msft (approved):

PR #13850, opened 2023-02-17
Check SONiC dependencies before installation.

liushilongbuaa:
#### Why I did it
SONiC related packages shouldn't be intalled from Pypi.
It is security compliance requirement.
#### How I did it
Check SONiC related packages when using setup.py.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211

### DIFF
--- a/src/sonic-config-engine/setup.py
+++ b/src/sonic-config-engine/setup.py
@@ -53,7 +53,7 @@ if sys.version_info.major == 3:
     ]
 
 for package in sonic_dependencies:
-    r = subprocess.call([sys.executable, '-m', 'show', package.split("==")[0]], stdout=sys.stderr.fileno())
+    r = subprocess.call([sys.executable, '-m', 'pip', 'show', package], stdout=sys.stderr.fileno())
     if r != 0:
         sys.stderr.write("Please build and install SONiC python wheels dependencies from github.com/sonic-net/sonic-buildimage\n")
         exit(1)
--- a/src/sonic-config-engine/setup.py
+++ b/src/sonic-config-engine/setup.py
@@ -54,7 +54,7 @@ if sys.version_info.major == 3:
     ]
 
 for package in dependencies:
-    if package.find("sonic") == -1:
+    if "sonic" not in package:
         continue
     try:
         package_dist = pkg_resources.get_distribution(package.split(">=")[0])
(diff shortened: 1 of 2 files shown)


xumia (approved):

src/sonic-config-engine/setup.py
```diff
@@ -50,6 +55,14 @@
         'sonic_yang_cfg_generator'
     ]
 
+for package in sonic_dependencies:
+    try:
+        pkg_resources.get_distribution(package)
+    except pkg_resources.DistributionNotFound:
+        print(package+" is not found!")
```
qiluo-msft, line 62: >print [](http://example.com/codeflow?start=8&length=5)

Print 2 error messages to stderr.

src/sonic-config-engine/setup.py
```diff
@@ -50,6 +55,14 @@
...
+        print("Please build and install SONiC python wheels dependencies from github.com/sonic-net/sonic-buildimage")
```
qiluo-msft, line 63: >github.com/sonic-net/ [](http://example.com/codeflow?start=78&length=21)

You could remove url and only keep the repo name.

src/sonic-config-engine/setup.py
```diff
@@ -24,6 +28,7 @@
         'sonic-yang-mgmt>=1.0',
         'sonic-yang-models>=1.0'
     ]
+    sonic_dependencies += ['sonic-yang-mgmt', 'sonic-yang-models']
```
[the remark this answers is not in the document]
liushilongbuaa: We can't generate from dependencies. Something like swsssdk didn't have prefix sonic-*.

src/sonic-config-engine/setup.py
liushilongbuaa: It should be used by python2 and python3.
we can't use print. Use sys.stderr.write instead.

qiluo-msft: you can use print(msg, file=sys.stderr)

liushilongbuaa: python2 doesn't support it.

src/sonic-config-engine/setup.py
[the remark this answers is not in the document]
qiluo-msft: sonic_dependencies could be ground truth, including the package names and version comparison.

1. install_requires could take the package names and version comparison. No need to repeat each package.
2. get_distribution check could take the package names, and fetch local versions, and compare with expected condition.

qiluo-msft (approved):

Issue #13791, opened 2023-02-13
sonic-cfggen isn't able to render template sporadically when the image is installed for the first time from ONIE


stephenxs:
#### Steps to reproduce the issue:
1. Install a new image on the switch from ONIE
2. Reboot the switch
3.

#### Describe the results you received:

Error message `buffermgrd: ERROR (spawn error) ` observed which is caused `asic_table.json` not being able to rendered from the template.
But from the dump we see the `asic_table.json` is empty, which is not expected.

The `asic_table.json` should be rendered from the template `asic_table.j2` using `sonic-cfggen` when the `swss` docker is created.
The template `asic_table.j2` is built into the image, which means it should be available if the image is good. The image is able to start at most times, which means the image should be good. So, `asic_table.j2` should be available.
The only possible cause is that `sonic-cfggen` wasn't able to render the template and generate `asic_table.json`.
The command to generate the json is `sonic-cfggen -d -t /usr/share/sonic/templates/asic_table.j2 > /etc/sonic/asic_table.json`. Only `DEVICE_METADATA|localhost[platform]` is required in the template.
The table `DEVICE_METADATA` should be good because from the log we see both platform and hwsku, where are both in `DEVICE_METADATA`, are available and correct.
In `sonic-cfggen`, only two if branches are executed:
1. args.from_db - load all items from the CONFIG_DB
2. args.print_data - dump the rendered dict to a json file.

branch 2 is simple and branch 1 is complicated. I suspect it somehow exited the process so nothing was output and `asic_table.json` was empty.

#### Describe the results you expected:

`asic_table.json` should be available.

#### Output of `show version`:

The issue is observed on master and 202211 from last December. 4 times occurred in the past 2 monthes.

```
Build date: Fri Feb 10 10:08:46 UTC 2023
Built by: sw-r2d2-bot@r-build-sonic-ci03-241

Platform: x86_64-mlnx_msn2700-r0
HwSKU: Mellanox-SN2700
ASIC: mellanox
ASIC Count: 1
Serial Number: MT2020T04277
Model Number: MSN2700-CS2FO
Hardware Revision: A2
Uptime: 15:00:00 up 5 min,  1 user,  load average: 4.60, 3.40, 1.48
Date: Fri 10 Feb 2023 15:00:00

Docker images:
REPOSITORY                                         TAG                               IMAGE ID       SIZE
[...]

PR #13888, opened 2023-02-20
Temporary WA for the issue that asic_table.json can not be rendered

stephenxs:
#### Why I did it

We suspect the issue #13791 is caused by redis server being temporarily unavailable during system initialization so we do not use -d in sonic-cfggen, for now, to avoid accessing redis server

#### How I did it

Provide a string containing required json data when calling `sonic-cfggen`

#### How to verify it

Manually test it

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211


liat-grozovik:
@stephenxs was this approach should be taken in other places in the code?
do we have an issue in github to track the issue?

stephenxs:
> @stephenxs was this approach should be taken in other places in the code? do we have an issue in github to track the issue?

Hi @liat-grozovik 
We have an issue https://github.com/sonic-net/sonic-buildimage/issues/13791
Theoretically, the issue can occur wherever redis-server is accessed, but so far we have observed similar issue only in rsyslog.conf in teamd https://github.com/sonic-net/sonic-buildimage/issues/13674 and there is a WA for it as well.
So I do not think we need to change the code everywhere redis-server is accessed.
Meanwhile, I'm not 100% confident that the issue is caused by redis-server not being accessible. Another PR to collect other info into sonic-dump (https://github.com/sonic-net/sonic-utilities/pull/2686) is also opened

files/build_templates/docker_image_ctl.j2
```diff
@@ -418,13 +418,13 @@ start() {
     {%- if docker_container_name == "swss" %}
     # Generate the asic_table.json and peripheral_table.json
     if [ ! -f /etc/sonic/asic_table.json ] && [ -f /usr/share/sonic/templates/asic_table.j2 ]; then
-        sonic-cfggen -d -t /usr/share/sonic/templates/asic_table.j2 > /etc/sonic/asic_table.json
+        sonic-cfggen -a '{"DEVICE_METADATA":{"localhost":{"platform":"'$PLATFORM'"}}}' -t /usr/share/sonic/templates/asic_table.j2 > /etc/sonic/asic_table.json
     fi
     if [ ! -f /etc/sonic/peripheral_table.json ] && [ -f /usr/share/sonic/device/$PLATFORM/port_peripheral_config.j2 ]; then
-        sonic-cfggen -d -t /usr/share/sonic/device/$PLATFORM/port_peripheral_config.j2 > /etc/sonic/peripheral_table.json
```
qiluo-msft, line 424: >-d [](http://example.com/codeflow?start=21&length=2)

Will this change break `gearbox_type` feature?

files/build_templates/docker_image_ctl.j2
```diff
@@ -418,13 +418,13 @@ start() {
...
+        sonic-cfggen -t /usr/share/sonic/device/$PLATFORM/port_peripheral_config.j2 > /etc/sonic/peripheral_table.json
     fi
     if [ ! -f /etc/sonic/zero_profiles.json ] && [ -f /usr/share/sonic/templates/zero_profiles.j2 ]; then
-        sonic-cfggen -d -t /usr/share/sonic/device/$PLATFORM/zero_profiles.j2 > /etc/sonic/zero_profiles.json
+        sonic-cfggen -t /usr/share/sonic/device/$PLATFORM/zero_profiles.j2 > /etc/sonic/zero_profiles.json
```
qiluo-msft, line 427: >zero_profiles [](http://example.com/codeflow?start=58&length=13)

This is more like a static content, not a template. Should we stop rendering?

files/build_templates/docker_image_ctl.j2
stephenxs: No.
this template is for the dynamic buffer manager to calculate headroom size. It is not for `gearbox_type` feature.

files/build_templates/docker_image_ctl.j2
stephenxs: Originally, it is designed as a template for all vendors that will use it. Now, even it is used by Nvidia only as static content, let's keep it, for now, to align with the design.
In addition, this is a WA and can be reverted once the root cause has been identified.
We will consider converting to a static json in the future once the root casue has been identified.

files/build_templates/docker_image_ctl.j2
```diff
@@ -418,13 +418,28 @@ start() {
...
+            sonic-cfggen -a '{"DEVICE_METADATA":{"localhost":{"platform":"'$PLATFORM'"}}}' -t /usr/share/sonic/templates/asic_table.j2 > /etc/sonic/asic_table.json
+        fi
     fi
     if [ ! -f /etc/sonic/peripheral_table.json ] && [ -f /usr/share/sonic/device/$PLATFORM/port_peripheral_config.j2 ]; then
-        sonic-cfggen -d -t /usr/share/sonic/device/$PLATFORM/port_peripheral_config.j2 > /etc/sonic/peripheral_table.json
+        sonic-cfggen -d -t /usr/share/sonic/device/$PLATFORM/port_peripheral_config.j2 > /etc/sonic/peripheral_table.json 2> errorlog.txt
+        if [[ $? -ne 0 ]] ; then
+            echo "sonic-cfggen failed to render peripheral_table.json, try working around it"
+            echo "$(cat errorlog.txt)"
+            sonic-cfggen -t /usr/share/sonic/device/$PLATFORM/port_peripheral_config.j2 > /etc/sonic/peripheral_table.json
```
liuh-80, line 433: 1. The target file should be port_peripheral_config.json ?

2. The port_peripheral_config.j2 include peripheral_config.j2, so when render this template, maybe also need provide DEVICE_METADATA?
    https://github.com/sonic-net/sonic-buildimage/blob/master/device/mellanox/x86_64-mlnx_msn3800-r0/port_peripheral_config.j2

3. Maybe create a method to reuse dupe code?

files/build_templates/docker_image_ctl.j2
```diff
@@ -418,13 +418,28 @@ start() {
     {%- if docker_container_name == "swss" %}
     # Generate the asic_table.json and peripheral_table.json
     if [ ! -f /etc/sonic/asic_table.json ] && [ -f /usr/share/sonic/templates/asic_table.j2 ]; then
-        sonic-cfggen -d -t /usr/share/sonic/templates/asic_table.j2 > /etc/sonic/asic_table.json
+        sonic-cfggen -d -t /usr/share/sonic/templates/asic_table.j2 > /etc/sonic/asic_table.json 2> errorlog.txt
+        if [[ $? -ne 0 ]] ; then
+            echo "sonic-cfggen failed to render asic_table.json, try working around it"
+            echo "$(cat errorlog.txt)"
+            sonic-cfggen -a '{"DEVICE_METADATA":{"localhost":{"platform":"'$PLATFORM'"}}}' -t /usr/share/sonic/templates/asic_table.j2 > /etc/sonic/asic_table.json
```
liuh-80, line 425: Use -a can work around, but there is a potential issue: if these templates update in future to use some data in other CONFIG_DB tables, and who change it doesn't know this work around. then when this issue happens, the rendered file will break.

Is there any solution to prevent that happen?

stephenxs: Theoretically, you are correct. But this is a very short-term WR.
During this time, we do not expect any change in asic_table.json.

files/build_templates/docker_image_ctl.j2
stephenxs: Re:
1. not really. the name should be peripheral_table.json
2. good catch, thank you. will fix it.
3. I tend to keep it as it is since it's just a short term WA and will be reverted once the redis issue will be fixed.

liuh-80 (approved):

liat-grozovik:
@qiluo-msft could you please review/approve?

neethajohn (dismissed):

files/build_templates/docker_image_ctl.j2
qiluo-msft: I see it in the file platform/mellanox/peripheral_table.j2.

files/build_templates/docker_image_ctl.j2
```diff
@@ -418,13 +418,28 @@ start() {
     {%- if docker_container_name == "swss" %}
     # Generate the asic_table.json and peripheral_table.json
     if [ ! -f /etc/sonic/asic_table.json ] && [ -f /usr/share/sonic/templates/asic_table.j2 ]; then
-        sonic-cfggen -d -t /usr/share/sonic/templates/asic_table.j2 > /etc/sonic/asic_table.json
+        sonic-cfggen -d -t /usr/share/sonic/templates/asic_table.j2 > /etc/sonic/asic_table.json 2> errorlog.txt
+        if [[ $? -ne 0 ]] ; then
```
qiluo-msft, line 422: >if [[ $? -ne 0 ]] ; then [](http://example.com/codeflow?start=8&length=24)

The error handling is useful and generic, we can separate into another PR.
This PR is WA, and will be revert in future.

stephenxs: PR #14384 has been opened for the error-handling logic.
I'll adjust this PR once that one is merged.

qiluo-msft: It is merged.

stephenxs: Thanks. Conflict resolved.

files/build_templates/docker_image_ctl.j2
```diff
@@ -424,20 +424,23 @@ start() {
         if [[ $? -ne 0 ]] ; then
             echo "sonic-cfggen failed to render asic_table.json"
             echo "$(cat errorlog.txt)"
+            sonic-cfggen -a '{"DEVICE_METADATA":{"localhost":{"platform":"'$PLATFORM'"}}}' -t /usr/share/sonic/templates/asic_table.j2 > /etc/sonic/asic_table.json
```
qiluo-msft, line 427: >/usr/share/sonic/templates/asic_table.j2 [](http://example.com/codeflow?start=94&length=40)

If someone change this j2 and using extra var other than "platform", how do you detect the breaking change?

Currently it will fail the command and silently continue. Is it better to report the error and exit with an error code?

stephenxs: If a vendor updated the template with another variable introduced, the failure can be detected by analyzing the dump. In that case, the asic_table.json will be an empty list `[]` and the root cause can be easily understood. In this sense, it doesn't bring too much benefit to detect the failure as this is a WA.
In addition, `sonic-cfggen` won't return an error if a variable isn't defined because sometimes it is expected from the j2 template's PoV.

qiluo-msft (approved):

keboliu (approved):

PR #13913, opened 2023-02-22
[Ci] Support to use the same snapshot for all platform builds

xumia:
#### Why I did it
Support to use the same snapshot for all platform builds.
There are multiple builds for the different platforms, such as vs, broadcom, mellanox, etc. The jobs to build each of the platform are not triggered at the same time, depended on if the agents are ready or not. The builds may use different snapshots, and use different version of a package, it is not a desired behavior.

#### How I did it
Added another stage to prepare the snapshot version, make sure all the builds using the same snapshot.
Backport to 202205/202211 which support the snapshot based build.

#### How to verify it
Created a self pipeline to verify it, see https://dev.azure.com/mssonic/build/_build/results?buildId=222340&view=logs&j=cef3d8a9-152e-5193-620b-567dc18af272&t=b9010a6b-c6d1-5dc2-a48b-b94374f4e7e6

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [x] 202205
- [x] 202211

### DIFF
--- a/.azure-pipelines/azure-pipelines-UpgrateVersion.yml
+++ b/.azure-pipelines/azure-pipelines-UpgrateVersion.yml
@@ -56,6 +56,7 @@ stages:
       name: SetVersions
       displayName: 'Set snapshot versions'
 - stage: Build
+  dependsOn: Prepare
   variables:
     - name: CACHE_MODE
       value: none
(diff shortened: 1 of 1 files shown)


liushilongbuaa (approved):

xumia:
The commit message is not correct, caused by copy & paste issue.

PR #13935, opened 2023-02-22
[chassis][voq] 400g to100g speed changes for chassis linecards

arlakshm:
#### Why I did it
On SONiC VoQ chassis, the speed changes are done from 400G to 100G needs to be supported on 400G linecards. 
To enable this, along with speed change the port lanes need to be changed. This PR has the changes to update the port lanes when such speed change happens.

This PR is intended only for VoQ chassis linecards. These platforms today have 400g port with 8 serdes lines, and 100g will operate with 4 serdes lane. When the port speed changes from 400G to 100G the first 4 lanes will be used for 100G port.

Platforms which support 2x50g PAM4 or support 100G PAM4 serdes or other combinations are not handled in the PR.

#### How I did it
Updated the port lanes when the port speed is changed from 400g to 100g.

#### How to verify it
UT and test on sonic chassis

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [x ] 202205
- [x ] 202211

### DIFF
--- a/src/sonic-config-engine/minigraph.py
+++ b/src/sonic-config-engine/minigraph.py
@@ -1666,6 +1666,15 @@ def parse_xml(filename, platform=None, port_config_file=None, asic_name=None, hw
         if port_name in mgmt_alias_reverse_mapping.keys():
             continue
 
+        port_default_speed =  port_speeds_default.get(port_name, None)
+        port_png_speed = port_speed_png[port_name]
+
+        # when the port speed is changes from 400g to 100g 
+        # update the port lanes, use the first 4 lanes of the 400G port to support 100G port
+        if port_default_speed == '400000' and port_png_speed == '100000':
+            port_lanes =  ",".join(ports[port_name].get('lanes', '').split(',')[:4])
+            ports[port_name]['lanes'] = port_lanes
+
         ports.setdefault(port_name, {})['speed'] = port_speed_png[port_name]
 
     for port_name, port in list(ports.items()):
--- a/src/sonic-config-engine/minigraph.py
+++ b/src/sonic-config-engine/minigraph.py
@@ -1669,11 +1669,16 @@ def parse_xml(filename, platform=None, port_config_file=None, asic_name=None, hw
         port_default_speed =  port_speeds_default.get(port_name, None)
         port_png_speed = port_speed_png[port_name]
 
-        # when the port speed is changes from 400g to 100g 
-        # update the port lanes, use the first 4 lanes of the 400G port to support 100G port
-        if port_default_speed == '400000' and port_png_speed == '100000':
-            port_lanes =  ",".join(ports[port_name].get('lanes', '').split(',')[:4])
-            ports[port_name]['lanes'] = port_lanes
+        if switch_type == 'voq':
+            # when the port speed is changes from 400g to 100g 
+            # update the port lanes, use the first 4 lanes of the 400G port to support 100G port
+            if port_default_speed == '400000' and port_png_speed == '100000':
+                port_lanes =  ports[port_name].get('lanes', '').split(',')
+                # check if the 400g port has only 8 lanes
+                if len(port_lanes) != 8:
+                    continue
+                updated_lanes = ",".join(port_lanes[:4])
+                ports[port_name]['lanes'] = updated_lanes
 
         ports.setdefault(port_name, {})['speed'] = port_speed_png[port_name]
 
--- a/src/sonic-config-engine/tests/test_cfggen.py
+++ b/src/sonic-config-engine/tests/test_cfggen.py
@@ -39,6 +39,8 @@ class TestCfgGen(TestCase):
         self.packet_chassis_port_ini = os.path.join(self.test_dir, 'sample-chassis-packet-lc-port-config.ini')
         self.macsec_profile = os.path.join(self.test_dir, 'macsec_profile.json')
         self.sample_backend_graph = os.path.join(self.test_dir, 'sample-graph-storage-backend.xml')
+        self.voq_port_config_400g = os.path.join(self.test_dir, 'voq-sample-400g-port-config.ini')
+        self.voq_sample_masic_graph = os.path.join(self.test_dir, 'voq-sample-masic-graph.xml')
         # To ensure that mock config_db data is used for unit-test cases
         os.environ["CFGGEN_UNIT_TESTING"] = "2"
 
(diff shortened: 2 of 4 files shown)


prgeor (approved):
@arlakshm approved with comments if you an address

src/sonic-config-engine/minigraph.py
```diff
@@ -1666,6 +1666,15 @@ def parse_xml(filename, platform=None, port_config_file=None, asic_name=None, hw
         if port_name in mgmt_alias_reverse_mapping.keys():
             continue
 
+        port_default_speed =  port_speeds_default.get(port_name, None)
+        port_png_speed = port_speed_png[port_name]
+
+        # when the port speed is changes from 400g to 100g
```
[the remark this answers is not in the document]
arlakshm: @prgeor, in this case when the port speed is changed from 400g to 100g. 
the 400g port is removed from sai and 100g port with new lanes is created in sai.

arlakshm:
@kenneth-arista , @kartik-arista , @abohanyang , @mlok-nokia @sanmalho-git  for viz...

src/sonic-config-engine/minigraph.py
```diff
@@ -1666,6 +1666,15 @@ def parse_xml(filename, platform=None, port_config_file=None, asic_name=None, hw
...
             continue
 
+        port_default_speed =  port_speeds_default.get(port_name, None)
+        port_png_speed = port_speed_png[port_name]
+
+        # when the port speed is changes from 400g to 100g 
+        # update the port lanes, use the first 4 lanes of the 400G port to support 100G port
+        if port_default_speed == '400000' and port_png_speed == '100000':
+            port_lanes =  ",".join(ports[port_name].get('lanes', '').split(',')[:4])
+            ports[port_name]['lanes'] = port_lanes
```
judyjoseph, line 1676: Just a question -- Is this method of taking the first four Serdes lanes when moving from 400G --> 100G is ok to be applied to non-voq and pizzabox devices also ?

judyjoseph (approved):
LGTM

src/sonic-config-engine/minigraph.py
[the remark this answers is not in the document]
abdosi: wondering is this valid assumption that always first 4 lines will translate to 100G ? Also should we check if  number of lanes are 8 then we should make it 4 ? Consider in future serdes speed might change

abdosi:
cc @anamehra  for review

abdosi:
added one comment,

anamehra:
Thanks for tagging @abdosi !
@shyam77git , @rajann FYI, please check.

src/sonic-config-engine/minigraph.py
anamehra: we need to validate this condition for our platforms and will revert back. Shouldn't be platform specific?

arlakshm: added a check for` switch_type = 'voq'` in the latest commit

src/sonic-config-engine/minigraph.py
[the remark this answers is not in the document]
arlakshm: @abdosi, added the check the number of lanes == 8.  In the voq chassis the assumption is for this specific usecase .i.e. speed change from 400g to 100g we will use the first 4 lanes. This is the only case we support today.

src/sonic-config-engine/minigraph.py
anamehra: Do we need a check for voq only? Most of non voq platforms support the first 4 lanes for 100G speed. For the rest, we need to check if modifying lane ordering in port_config.ini will work. If not, we will have to add additional support. If we do not have this check for voq, this won't be breaking any functionality for supported platforms. So I think we should be good without this check.

anamehra:
Does this support the speed change only via minigraph or CLI as well?

prgeor (changes requested):
@arlakshm @abdosi please note that in future when we use 100G PAM4 serdes this change won't be able to handle 400G speed(4 lanes) to 100G speed(1 lane). There is an inherent assumption here that 400G speed will always have 8 lanes and 100G will always have 4 lanes. Can we think of a better generic solution? This solution looks only specific to a particular platform and won't stand future platforms that are going to be based on 100G PAM4.

gechiang:
> 100G PAM4 serdes

Excellent point.  I understand some vendors have started to use this but I would suggest we take on this 100G PAM4 serdes as a phase 2 change.  Also, the adaptation to this new optic which inherently has "higher" chance of noise can causing bad data at corrupting 2 bits at the same time may require additional time to see if it will be "widely" adopted by the community...

arlakshm:
> @arlakshm @abdosi please note that in future when we use 100G PAM4 serdes this change won't be able to handle 400G speed(4 lanes) to 100G speed(1 lane). There is an inherent assumption here that 400G speed will always have 8 lanes and 100G will always have 4 lanes. Can we think of a better generic solution? This solution looks only specific to a particular platform and won't stand future platforms that are going to be based on 100G PAM4.

@prgeor, as discussed offline, the scope of change to solve a specific use case: Support changing speed from 400g to 100g for voq linecards.  These platforms today have 400g port with 8 serdes lines, and 100g will operate with 4 serdes lane.
Currently we don't have platforms with 400g port with 4 lanes. When such platforms are introduced, we need to enhance this change in the future

arlakshm:
> Does this support the speed change only via minigraph or CLI as well?

The current use case is to support this speed change via minigraph

src/sonic-config-engine/minigraph.py
arlakshm: @anamehra, we are currently testing only voq linecards. so added the check to limit the scope of the change in this PR.
When we validate this feature on other platform when can relax the check

arlakshm:
> > 100G PAM4 serdes
> 
> Excellent point. I understand some vendors have started to use this but I would suggest we take on this 100G PAM4 serdes as a phase 2 change. Also, the adaptation to this new optic which inherently has "higher" chance of noise can causing bad data at corrupting 2 bits at the same time may require additional time to see if it will be "widely" adopted by the community...

@gechiang, yes, The current PR is not intended for all the platforms. We can enhance this change on other platforms, based testing and other use-cases which might arise in the future.

src/sonic-config-engine/minigraph.py
anamehra: @abdosi , do you know what are the other non voq platforms, other than chassis-packet, could be affected by this? I think this check for voq could be dropped. Any other way for a platform to decide whether to use or not this feature by checking compatibility?
Also, is it possible that the platform could provde the lane information to be used for 400G/100G instead of assuming the first 4 lanes?

prgeor (approved):

prgeor:
@lguohan please help merge

prgeor:
> > 100G PAM4 serdes
> 
> Excellent point. I understand some vendors have started to use this but I would suggest we take on this 100G PAM4 serdes as a phase 2 change. Also, the adaptation to this new optic which inherently has "higher" chance of noise can causing bad data at corrupting 2 bits at the same time may require additional time to see if it will be "widely" adopted by the community...

@gechiang not just that, we dont have means in minigraph to know whether the 100G speed is from 2x100G, 4x100G or 8x100G. We can come back later

abdosi (approved):
LGTM for current usecase

src/sonic-config-engine/minigraph.py
abdosi: @anamehra agree once we have use case we can remove voq check as needed. To have backward compatibility it is better to have voq check for now.  

Regarding platform change i think that is another option for this. We can think on this further in subsequent enhancement

lguohan:
@prgeor , we should test this and see if pizza box works with this lane selection.

lguohan:
the pr looks good, however, description is a little bit concise/brief. i suggest to pull some of the discussions into the pr description and give people a full story. for example, there is some limitation of this pr, like 2x50PAM 100g mode is not supported.

arlakshm:
> the pr looks good, however, description is a little bit concise/brief. i suggest to pull some of the discussions into the pr description and give people a full story. for example, there is some limitation of this pr, like 2x50PAM 100g mode is not supported.

@lguohan updated the description.

sanmalho-git:
@arlakshm - was this tested against a multi-asic linecard?

arlakshm:
@sanmalho-git, yes I am testing on a multi asic linecard

sanmalho-git:
@arlakshm  - thanks - please let us know once you are done with testing, and I can try it out on our cards as well. As is, it is failing on multi-asic cards.

PR #13497, opened 2023-01-24
system-health service_checker should check containers based on asic presence

spilkey-cisco:
#### Why I did it

On a supervisor card in a chassis, syncd/teamd/swss/lldp etc dockers are created for each Switch Fabric card. However, not all chassis would have all the switch fabric cards present. In this case, only dockers for Switch Fabrics present would be created.

system-health indicates errors in this scenario as it is expecting dockers for all Switch Fabrics (based on NUM_ASIC defined in asic.conf file).

system-health process error messages were also altered to indicate which container had the issue; multiple containers may run processes with the same name, which can result in identical system-health error messages, causing ambiguity.

#### How I did it

Port container_checker logic from https://github.com/sonic-net/sonic-buildimage/pull/11442 into service_checker for system-health.

#### How to verify it

Bringup Supervisor card with one or more missing fabric cards. Execute 'show system-health summary'. The command should not report failure due to missing dockers for the asics on the fabric cards which are not present.

#### Which release branch to backport (provide reason below if selected)

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [x] 202205
- [x] 202211

PR #13966, opened 2023-02-23
system-health service_checker should check containers based on asic presence (cherry-pick #13497)

spilkey-cisco:
**Why I did it**

On a supervisor card in a chassis, syncd/teamd/swss/lldp etc dockers are created for each Switch Fabric card. However, not all chassis would have all the switch fabric cards present. In this case, only dockers for Switch Fabrics present would be created.

system-health indicates errors in this scenario as it is expecting dockers for all Switch Fabrics (based on NUM_ASIC defined in asic.conf file).

system-health process error messages were also altered to indicate which container had the issue; multiple containers may run processes with the same name, which can result in identical system-health error messages, causing ambiguity.

**How I did it**

Cherry-pick https://github.com/sonic-net/sonic-buildimage/pull/13497 into 202205 and resolve conflicts

**How to verify it**

Bringup Supervisor card with one or more missing fabric cards. Execute 'show system-health summary'. The command should not report failure due to missing dockers for the asics on the fabric cards which are not present.

### DIFF
--- a/src/system-health/health_checker/service_checker.py
+++ b/src/system-health/health_checker/service_checker.py
@@ -67,6 +67,19 @@ class ServiceChecker(HealthChecker):
         """
         expected_running_containers = set()
         container_feature_dict = {}
+
+        # Get current asic presence list. For multi_asic system, multi instance containers
+        # should be checked only for asics present.
+        asics_id_presence = multi_asic.get_asic_presence_list()
+
+        # Some services may run all the instances irrespective of asic presence.
+        # Add those to exception list.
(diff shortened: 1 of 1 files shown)


spilkey-cisco:
@gechiang, @abdosi, opened this PR to resolve conflicts from cherry-pick of https://github.com/sonic-net/sonic-buildimage/pull/13497

gechiang (approved):

spilkey-cisco:
Conflict was from https://github.com/sonic-net/sonic-buildimage/pull/12563 in master.

Issue #13046, opened 2022-12-14
Time sync issue between SONiC and FRR stack


nmoray-ebay:
#### Description
There is a time sync issue between SONIC and the FRR stack. For instance, Inside vtysh, "Show interface status" CLI output and FRR log always shows the UTC timestamp which does not match with the SONIC in case a specific time zone is configured on the SONIC side.

#### Steps to reproduce the issue:
1. Configure specific timezone except UTC in SONIC via "timedatectl set-timezone <>" CLI
2. Flap the UP interface
3. Go inside vtysh and check the output of "show interface <>" for the flap timestamp
4. Alternatively check the zebra.log file

#### Describe the results you received:
LOGS:
root@SN2100-Leaf1:~# show clock
Wed 14 Dec 2022 09:47:17 AM UTC
root@SN2100-Leaf1:~# timedatectl set-timezone America/Los_Angeles  ========> configured as PST ( using set timezone)
root@SN2100-Leaf1:~# 
root@SN2100-Leaf1:~# 
root@SN2100-Leaf1:~# 
root@SN2100-Leaf1:~# show clock
Wed 14 Dec 2022 01:48:12 AM PST
root@SN2100-Leaf1:~# config interface shutdown Ethernet0
root@SN2100-Leaf1:~# config interface startup Ethernet0
root@SN2100-Leaf1:~# vtysh

Hello, this is FRRouting (version 7.5.1-sonic).
Copyright 1996-2005 Kunihiro Ishiguro, et al.

SN2100-Leaf1# date 
% Unknown command: date
SN2100-Leaf1# show int Ethernet0  
Interface Ethernet0 is up, line protocol is down
  Link ups:       1    last: 2022/12/14 06:35:42.89.  ====>  shows  UTC timestamp
  Link downs:     6    last: 2022/12/14 09:48:48.29
  vrf: default
  index 748 metric 0 mtu 9100 speed 0 
  flags: <UP,BROADCAST,MULTICAST>
  Type: Unknown
  HWaddr: 1c:34:da:36:f8:00
  Interface Type Other
  Interface Slave Type None

root@SN2100-Leaf1:/var/log/frr# tail -5f zebra.log
Dec 14 09:48:37.953850 SN2100-Leaf1 ERR bgp#zebra[35]: [EC 4043309093] netlink-dp (NS 0) error: Network is down, type=RTM_NEWNEXTHOP(104), seq=165, pid=2957759799
[...]

PR #14000, opened 2023-02-27
Timezone sync issue between the host and containers

nmoray:
#### Why I did it
To fix the timezone sync issue between the containers and the host. If a certain timezone has been configured on the host (SONIC) then the expectation is to reflect the same across all the containers.

This will fix [Issue:13046](https://github.com/sonic-net/sonic-buildimage/issues/13046).

For instance, a PST timezone has been set on the host and if the user checks the link flap logs (inside the FRR), it shows the UTC timestamp. Ideally, it should be PST. 

LOGS:
root@sonic:~# show clock
Mon Feb 7 12:01:16 PST 2022 ======== configured as PST ( using set timezone)

sonic# vtysh
sonic# show interface Ethernet0
Interface Ethernet0 is up, line protocol is up
Link ups: 4 last: 2022/02/03 20:17:38.34 ----> shows UTC timestamp
Link downs: 6 last: 2022/02/03 20:17:13.35 ---->
vrf: default
index 26 metric 0 mtu 9126 speed 0
flags: <UP,BROADCAST,RUNNING,MULTICAST>
Type: Unknown
HWaddr: 00:30:64:6a:fa:c7
inet [172.21.227.64/31](https://nam10.safelinks.protection.outlook.com/?url=http%3A%2F%2F172.21.227.64%2F31&data=05%7C01%7Cnmoray%40ebay.com%7C9b88b81e54ac402f3f4a08da3e1ed96e%7C46326bff992841a0baca17c16c94ea99%7C0%7C0%7C637890599471979401%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000%7C%7C%7C&sdata=lyPmxXhB8QOLP2JXtoCMMxDZ%2FuTFGzKCtIoJeukE%2BcA%3D&reserved=0)
inet6 fd0a:200💯:340/127
inet6 fe80::230:64ff:fe6a:fac7/64
Interface Type Other

#### How I did it
Mounted host's /etc/timezone file on all the containers to  find out the presently configured timezone during the docker initialisation phase. Later, based on the host's timezone, create a symbolic link at runtime inside each of the containers (/etc/localtime). This way, we can avoid mounting the symlink on the containers. 

#### How to verify it
root@sonic:~# timedatectl set-timezone America/Los_Angeles
root@sonic:~#
root@sonic:~#
root@sonic:~# date
Fri 28 Apr 2023 10:22:40 AM PDT

root@sonic:~# cat /etc/timezone
America/Los_Angeles
root@sonic:~# ls /etc/localtime -l
lrwxrwxrwx 1 root root 41 Apr 28 10:22 /etc/localtime -> ../usr/share/zoneinfo/America/Los_Angeles
root@sonic:~# docker exec -it bgp bash
root@sonic:/# date
Fri Apr 28 17:23:17 UTC 2023
root@sonic:/# exit
exit

root@sonic:~# config reload -f
Clear current config and reload config in config_db format from the default config file(s) ? [y/N]: y
Running command: rm -rf /tmp/dropstat-*
Stopping SONiC target ...
Running command: /usr/local/bin/sonic-cfggen  -j /etc/sonic/init_cfg.json  -j /etc/sonic/config_db.json  --write-to-db
Restarting SONiC target ...
Reinitializing monit daemon
root@sonic:~# docker ps
CONTAINER ID   IMAGE                                COMMAND                  CREATED              STATUS          PORTS     NAMES
b50ea6d53d4b   docker-snmp:latest                   "/usr/local/bin/supe…"   About a minute ago   Up 2 seconds              snmp
15b9c0860197   docker-sonic-telemetry:latest        "/usr/local/bin/supe…"   About a minute ago   Up 5 seconds              telemetry
9a8b5c0fed5d   docker-sonic-mgmt-framework:latest   "/usr/local/bin/supe…"   About a minute ago   Up 5 seconds              mgmt-framework
c302bdbeee41   docker-lldp:latest                   "/usr/bin/docker-lld…"   4 minutes ago        Up 5 seconds              lldp
c7e150c50494   docker-platform-monitor:latest       "/usr/bin/docker_ini…"   4 minutes ago        Up 14 seconds             pmon
b0be22f94e6c   docker-router-advertiser:latest      "/usr/bin/docker-ini…"   5 minutes ago        Up 4 seconds              radv
4298354779e0   docker-syncd-invm:latest             "/usr/local/bin/supe…"   5 minutes ago        Up 6 seconds              syncd
a03f8f8a300d   docker-teamd:latest                  "/usr/local/bin/supe…"   5 minutes ago        Up 6 seconds              teamd
c24c23066a6b   docker-orchagent:latest              "/usr/bin/docker-ini…"   5 minutes ago        Up 7 seconds              swss
89a0522ffc86   docker-fpm-frr:latest                "/usr/bin/docker_ini…"   5 minutes ago        Up 13 seconds             bgp
ecccce070e53   docker-database:latest               "/usr/local/bin/dock…"   5 minutes ago        Up 5 minutes              database

root@sonic:~# docker exec -it bgp bash
root@sonic:/# date
Fri Apr 28 10:24:07 PDT 2023
root@sonic:/# cat /etc/timezone
America/Los_Angeles
root@sonic:/# ls /etc/localtime -l
lrwxrwxrwx 1 root root 39 Apr 28 10:23 /etc/localtime -> /usr/share/zoneinfo/America/Los_Angeles
root@sonic:/# exit
exit

root@sonic:~# docker exec -it snmp bash
root@sonic:/# date
Fri Apr 28 10:29:40 PDT 2023
root@sonic:/# cat  /etc/timezone
America/Los_Angeles
root@sonic:/# ls /etc/localtime -l
lrwxrwxrwx 1 root root 39 Apr 28 10:24 /etc/localtime -> /usr/share/zoneinfo/America/Los_Angeles
root@sonic:/#

- [ ] 201811
- [x] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [x] 202111
- [x] 202205
- [ ] 202211

Respective builds will be consumed by the customer.


madhupalu (approved):
Code changes looks good. 
LGTM

madhupalu:
@gechiang
Can you take a look and approve these changes? 
Thanks,
Madhu

gechiang (dismissed):

madhupalu:
@lguohan @xumia and @qiluo-msft 
This issue being triaged as p1, couple of approvals already received. 
Can you take a look and approve it? 

Thanks,
-Madhu

lguohan:
the description is not clear, can you please update the description so we know clear the problem is? @nmoray

nmoray:
> the description is not clear, can you please update the description so we know clear the problem is? @nmoray

Sure. Updated the description with more insights of the actual issue. @lguohan

nmoray:
+ @nyanto: From eBay

@lguohan: Gentle reminder!

rules/docker-fpm-frr.mk
```diff
@@ -30,6 +30,8 @@ SONIC_DOCKER_DBG_IMAGES += $(DOCKER_FPM_FRR_DBG)
 $(DOCKER_FPM_FRR)_CONTAINER_NAME = bgp
 $(DOCKER_FPM_FRR)_RUN_OPT += --privileged -t
 $(DOCKER_FPM_FRR)_RUN_OPT += -v /etc/sonic:/etc/sonic:ro
+$(DOCKER_FPM_FRR)_RUN_OPT += -v /etc/timezone:/etc/timezone:ro
+$(DOCKER_FPM_FRR)_RUN_OPT += -v /etc/localtime:/etc/localtime:ro
```
lguohan, line 34: this is the symbol link, how does mount work here?

nmoray: > the problem seems not specific to frr, why other containers do not need this? @qiluo-msft , do you have any concern for mount them into the container?

This issue persist across all the containers. We can use the same technique to sync time/timezone on other containers too.

nmoray: > this is the symbol link, how does mount work here?

Yes "/etc/localtime" is a symlink and it works only when we restart the BGP container. For instance, if we have changed the timezone on the host and to sync that with the container, we need to restart the container to update the respective mount share inside the container. With this, we don't need to mount the absolute path which is "/usr/share/zoneinfo".

lguohan:
@qiluo-msft can you take a look?

nmoray:
> > 1. timedatectl set-timezone America/Los_Angeles
> 
> > @madhupalu should we apply this to all containers? Why bgp only?
> 
> Yes @lguohan, same logic can be applied to all other dockers as well to sync the time across all the containers. Just that the user needs to make sure it will do a "config reload" (and not just a specific container restart) after setting a specific timezone to update all the containers at the same time with the new timezone and local time.

As requested, I have updated *.mk files of all the dockers to get the time sync across all the dockers.

nyanto:
+1  This improvement if fairly straightforward and is deployed in our environment. 
It would be good if it can be merged in soon.

rules/docker-fpm-frr.mk
qiluo-msft: @nmoray Mounting a symlink on host without mounting the linked file at the same time is problematic. And this symlink is distro implementation detail, and it may break if any host/container distro change.

I am trying to understand your motivation at the very beginning. If your goal is to customize timezone used in BGP container, the best fix it to change timezone in BGP container only. Why you need to change timezone on the host and expect the FRR and SONiC time sync? The time sync is a nice-to-have feature, but overkill for your need.

qiluo-msft (changes requested):
As comments.

nmoray:
> As comments.

Sorry but are you asking me to add the comments?

rules/docker-fpm-frr.mk
nmoray: @qiluo-msft I agree that we should mount both symlink along with the target file. But why do you think the distro change will break the implementation? If the container distro changes, it will automatically update the mounted symlink too, right? 

Yes, we can manually update the timezone in each and every container. But the motive behind doing this is the auto time / zone sync between the container(s) and host.

madhupalu:
@nmoray thanks for taking care of all the docker services, @lguohan @qiluo-msft if there are no comments could you please approve the PR.  Also let us know if it needs to ported any other releases. Thanks

madhupalu:
@nmoray do you have any pending code changes to address this PR review comments?

nmoray:
> @nmoray do you have any pending code changes to address this PR review comments?

Yes, I do @madhupalu. I will soon update the PR to address @qiluo-msft comment.

nmoray:
@qiluo-msft As per your comment, I am no more mount the symlink (/etc/localtime). Alternatively, creating that at run time based on the in already mounted /etc/timezone. Please review and share your comments.

@madhupalu FYI

dockers/docker-database/docker-database-init.sh
```diff
@@ -95,4 +95,8 @@ do
     fi
 done
 
+TZ=$(cat /etc/timezone)
+rm -rf /etc/localtime
```
qiluo-msft, line 99: >rm [](http://example.com/codeflow?start=0&length=2)

This script is only run once when container starting. So the motivation "the auto time / zone sync between the container(s) and host" is not fully achieved. If after container staring, user change timezone on the host, it will not sync to the container.

qiluo-msft:
Please change the PR title, it is not a FRR specific PR.
Please add issue link in the PR description. So people can click the link.

dockers/docker-database/docker-database-init.sh
nmoray: If we consider an actual deployment scenario, timezone is configured only once before going to the Production. It is not the config which keeps getting changed again and again. But still, in case network admin wants to change the timezone post deployment, it can still be synced across the containers by triggering the config-reload command post changing the timezone on the host.

nmoray:
> Please change the PR title, it is not a FRR specific PR. Please add issue link in the PR description. So people can click the link.

nmoray:
> > Please change the PR title, it is not a FRR specific PR. Please add issue link in the PR description. So people can click the link.

Sure, I will do it.

madhupalu:
@qiluo-msft @lguohan 
@nmoray had addressed review comments, can you help to get approve this PR? 
Thanks

dockers/docker-database/docker-database-init.sh
qiluo-msft: I do not agree "timezone is configured only once before going to the Production". The feature name 'Time sync' seems not accurate. You only propagate it when container starting, not sync after.

nmoray: Once the initial time sync has been completed between the host and container (during docker-init), later there will not any mismatch until the timezone on the host has been changed (Which is sort of periodic sync only). And to tackle that user needs to do config-reload to propagate updated timezone inside the containers.

Can you please tell me the actual used case where we need to change the timezone frequently and config reload / container restart is not allowed?

In my opinion, user can at least get some sort of clean mechanism to sync the time between the host and containers with this approach.

qiluo-msft: We always use UTC timezone so we do not have such use case. You raised the original issue https://github.com/sonic-net/sonic-buildimage/issues/13046. In order to fix your use case, change timezone inside FRR container is good enough. For a timezone fix, there is no need to disruptive dataplane traffic.

nmoray: Initially I was focusing only on the FRR but later I found that this feature is required for other containers too. For instance, sairedis logs from swss where the timezone is synced correctly with the swss container and the host. 

------------------------------------------------------------------------------------
`root@sonic:/var/log/swss# tail -f sairedis.rec 
2023-06-14.21:23:20.593741|s|SAI_OBJECT_TYPE_PORT:oid:0x1000000000133|SAI_PORT_ATTR_QOS_DSCP_TO_TC_MAP=oid:0x140000000003d3
2023-06-14.21:23:20.594524|s|SAI_OBJECT_TYPE_PORT:oid:0x1000000000133|SAI_PORT_ATTR_QOS_TC_TO_QUEUE_MAP=oid:0x140000000003d5
2023-06-14.21:23:20.595066|s|SAI_OBJECT_TYPE_PORT:oid:0x1000000000133|SAI_PORT_ATTR_QOS_TC_TO_PRIORITY_GROUP_MAP=oid:0x140000000003d4
2023-06-14.21:23:20.595703|s|SAI_OBJECT_TYPE_PORT:oid:0x1000000000133|SAI_PORT_ATTR_PRIORITY_FLOW_CONTROL=24
2023-06-14.21:24:16.185792|c|SAI_OBJECT_TYPE_HOSTIF_TRAP_GROUP:oid:0x110000000003d7|SAI_HOSTIF_TRAP_GROUP_ATTR_QUEUE=2
2023-06-14.21:24:16.187074|c|SAI_OBJECT_TYPE_POLICER:oid:0x120000000003d8|SAI_POLICER_ATTR_CBS=1000|SAI_POLICER_ATTR_CIR=1000|SAI_POLICER_ATTR_METER_TYPE=SAI_METER_TYPE_PACKETS|SAI_POLICER_ATTR_MODE=SAI_POLICER_MODE_SR_TCM|SAI_POLICER_ATTR_RED_PACKET_ACTION=SAI_PACKET_ACTION_DROP
2023-06-14.21:24:16.187953|s|SAI_OBJECT_TYPE_HOSTIF_TRAP_GROUP:oid:0x110000000003d7|SAI_HOSTIF_TRAP_GROUP_ATTR_POLICER=oid:0x120000000003d8
2023-06-14.21:24:16.191987|c|SAI_OBJECT_TYPE_HOSTIF:oid:0xd0000000003d9|SAI_HOSTIF_ATTR_GENETLINK_MCGRP_NAME=packets|SAI_HOSTIF_ATTR_TYPE=SAI_HOSTIF_TYPE_GENETLINK|SAI_HOSTIF_ATTR_NAME=psample
2023-06-14.21:24:16.193059|c|SAI_OBJECT_TYPE_HOSTIF_TRAP:oid:0x220000000003da|SAI_HOSTIF_TRAP_ATTR_TRAP_TYPE=SAI_HOSTIF_TRAP_TYPE_SAMPLEPACKET|SAI_HOSTIF_TRAP_ATTR_TRAP_GROUP=oid:0x110000000003d7|SAI_HOSTIF_TRAP_ATTR_PACKET_ACTION=SAI_PACKET_ACTION_TRAP|SAI_HOSTIF_TRAP_ATTR_TRAP_PRIORITY=1
**2023-06-14.21:24:16**.200215|c|SAI_OBJECT_TYPE_HOSTIF_TABLE_ENTRY:oid:0x230000000003db|SAI_HOSTIF_TABLE_ENTRY_ATTR_TYPE=SAI_HOSTIF_TABLE_ENTRY_TYPE_TRAP_ID|SAI_HOSTIF_TABLE_ENTRY_ATTR_TRAP_ID=oid:0x220000000003da|SAI_HOSTIF_TABLE_ENTRY_ATTR_CHANNEL_TYPE=SAI_HOSTIF_TABLE_ENTRY_CHANNEL_TYPE_GENETLINK|SAI_HOSTIF_TABLE_ENTRY_ATTR_HOST_IF=oid:0xd0000000003d9
^C
root@sonic:/var/log/swss# date
**Wed Jun 14 21:25:40 PDT 2023**`
------------------------------------------------------------------------------------
Thus, this feature will fix [#13046](https://github.com/sonic-net/sonic-buildimage/issues/13046) and other similar issues too. 
BTW, this feature will be optional only. It is not going to harm or disrupt any exiting operation. If someone wants to sync the timezone across the containers, he / she can use it else others can live with the UTC timezone only. There is no force (for the ones who doesn't want to change the timezone) to disrupt the dataplane traffic and the ones who wants to change it, can plan it accordingly as per their deployment strategies. There are several customers, who prefers to have the timezone set as per their geographical location. Thus, in my opinion, it is a good to have feature. 

To answer to your previous comment "The feature name 'Time sync' seems not accurate. You only propagate it when container starting, not sync after", we can certainly achieve the periodic sync by simply adding a script per container which will periodically check for the change of TZ on the host but it will be an overhead just for syncing the timezone. Thus, I decided to sync it only at the time of docker init as usually that config is not being updated frequently.

qiluo-msft: Do you explore the solution of 
```
docker run -e TZ=HOSTTIMEZONE image command
docker create image -e TZ=HOSTTIMEZONE
```

The benefit is centralized code change. And it behaves the same way only running once when container starting.

nmoray: Yeah it works if we already knew the TIMEZONE. For instance, we can set ENV in Dockerfile to pre-set the TZ. But how are we going to make the TZ as variable field in SONIC?

qiluo-msft: Sorry, I realize above 2 commands only work during creating the container, not during container restart.

dockers/docker-sflow/start.sh
qiluo-msft, line 1: This file is empty.

Do you want to add
```
TZ=$(cat /etc/timezone)
rm -rf /etc/localtime
ln -sf /usr/share/zoneinfo/$TZ /etc/localtime
```

dockers/docker-sflow/supervisord.conf
```diff
@@ -28,6 +28,17 @@ stdout_logfile=syslog
 stderr_logfile=syslog
 dependent_startup=true
 
+[program:start]
+command=/usr/bin/start.sh
+priority=2
+autostart=false
```
qiluo-msft, line 34: >false [](http://example.com/codeflow?start=10&length=5)

true?

dockers/docker-database/docker-database-init.sh
nmoray: Yeah :-)

dockers/docker-sflow/start.sh
nmoray: yeah. I missed it. Thanks for pointing it out.

dockers/docker-sflow/supervisord.conf
nmoray: Correct me if I wrong but I guess it is not required as the dependent_startup flag is true. So it will eventually gets started post dependent-startup. Also, I have referred other supervisord.conf files and found that it is most of the times false only.

dockers/docker-sflow/start.sh
nmoray: Updated the file.

qiluo-msft (approved):

PR #14002, opened 2023-02-27
Add teamd patches to solve traffic loss issue when removing port from LAG

liorghub:
#### Why I did it
When removing port from LAG while traffic is running thorough LAG there is traffic disruption of 60 seconds.

#### How I did it
The patch I added introduces "port_removing" op and call it right before Kernel is asked to remove the port. 
Implement the op in LACP runner to disable the port which leads to proper LACPDU send.

#### How to verify it
Set LAG between 2 switches.
Set LAGs to be router port and set ip address.
In switch A send ping to ip address of LAG in switch B.
In switch B, while ping is running remove port from LAG.
Verify ping is not stopping.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [x] 202205
- [x] 202211

### DIFF
--- a/src/libteam/patch/0014-dont-move-the-port-state-from-disabled-when-admin-state-is-down
+++ b/src/libteam/patch/0014-dont-move-the-port-state-from-disabled-when-admin-state-is-down
@@ -1,0 +1,32 @@
+From 23ab49c4df0c06eb629ce2e3bb4c4dd7c527975a Mon Sep 17 00:00:00 2001
+From: Jiri Pirko <jiri@nvidia.com>
+Date: Thu, 2 Feb 2023 17:00:51 +0100
+Subject: [PATCH] teamd: lacp: don't move the port state from disabled when
+ admin state is down
+
+When the team admin state is down, the port should stay in disabled
+state, no matter what's happening. So check the admin state and bail out
+in that case.
+
+Signed-off-by: Jiri Pirko <jiri@nvidia.com>
+---
+ teamd/teamd_runner_lacp.c | 4 +++-
+ 1 file changed, 3 insertions(+), 1 deletion(-)
+
+diff --git a/teamd/teamd_runner_lacp.c b/teamd/teamd_runner_lacp.c
+index 51c7714..a76c372 100644
+--- a/teamd/teamd_runner_lacp.c
++++ b/teamd/teamd_runner_lacp.c
+@@ -956,9 +956,11 @@ static int lacpdu_send(struct lacp_port *lacp_port);
+ static int lacp_port_set_state(struct lacp_port *lacp_port,
+ 			       enum lacp_port_state new_state)
+ {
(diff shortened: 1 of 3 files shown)


liat-grozovik:
@stepanblyschak , @vaibhavhd  appreciate if you can review.

liat-grozovik:
@liorghub can you confirm this can be cleanly cherry picked to 202205?

liorghub:
@liat-grozovik 
> @liorghub can you confirm this can be cleanly cherry picked to 202205?

Yes, this PR can be cleanly cherry picked to 202205.

liorghub:
This PR fixes issue https://github.com/sonic-net/sonic-buildimage/issues/14381

src/libteam/patch/0014-dont-move-the-port-state-from-disabled-when-admin-state-is-down
saiarcot895, line 1: Can you add a `.patch` suffix to this?

src/libteam/patch/series
```diff
@@ -10,3 +10,5 @@
 0010-When-read-of-timerfd-returned-0-don-t-consider-this-.patch
 0011-Remove-extensive-debug-output.patch
 0012-Increase-min_ports-upper-limit-to-1024.patch
+0013-set-port-to-disabled-state-during-removal.patch
+0014-dont-move-the-port-state-from-disabled-when-admin-state-is-down
```
saiarcot895, line 14: This needs to be updated.

saiarcot895:
I tried to repro the issue described here on a t0-64-32 topo with sonic neighbors (with config_db.json modified to remove the `"min_links": 2` requirement), and didn't see any ping failures while I removed and added a LAG member from an existing LAG. I did see slightly higher RTT times (probably because the primary interface was being removed, and another interface had to be selected).

Is this happening only on physical hardware, or is it on a virtual setup as well?

src/libteam/patch/series
liorghub: Apologies, done.

liorghub:
> I tried to repro the issue described here on a t0-64-32 topo with sonic neighbors (with config_db.json modified to remove the `"min_links": 2` requirement), and didn't see any ping failures while I removed and added a LAG member from an existing LAG. I did see slightly higher RTT times (probably because the primary interface was being removed, and another interface had to be selected).
> 
> Is this happening only on physical hardware, or is it on a virtual setup as well?

It happens only on physical hardware when both sides are switches (when one side is host and we create bonding, issue is not reproduced).

saiarcot895 (approved):
I verified that this is sort-of happening on virtual switch as well. The ICMP response packet can arrive on the interface that was just removed from the port channel (and so outside of the port channel), if it happens to be the primary interface on the peer virtual switch. It doesn't manifest as a traffic drop because `rp_filter` checks are disabled, so the kernel accepts it as a valid packet, even though it's outside of the port channel.

PR #14005, opened 2023-02-27
[sonic-py-common] support displaying output while command is running (e.g., 'tail')

maipbui:
#### Why I did it
Previously, command output wasn't being displayed until the command finished running. This would cause programs like show logging (in this PR change https://github.com/sonic-net/sonic-utilities/pull/2666/files) to not display any output until the command finishes running.
#### How I did it
Support displaying output while command is running
#### How to verify it
Manual test show logging command

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211

### DIFF
--- a/src/sonic-py-common/sonic_py_common/general.py
+++ b/src/sonic-py-common/sonic_py_common/general.py
@@ -1,4 +1,5 @@
 import sys
+import click
 from subprocess import Popen, STDOUT, PIPE, CalledProcessError, check_output
 
 
@@ -43,7 +44,7 @@ def getstatusoutput_noshell(cmd):
     return exitcode, output
 
 
-def getstatusoutput_noshell_pipe(cmd0, *args):
+def getstatusoutput_noshell_pipe(cmd0, *args, return_cmd=True):
     """
     This function implements getstatusoutput API from subprocess module
     but using shell=False to prevent shell injection. Input command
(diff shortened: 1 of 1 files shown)


src/sonic-py-common/sonic_py_common/general.py
```diff
@@ -1,4 +1,5 @@
 import sys
+import click
```
qiluo-msft, line 2: >click [](http://example.com/codeflow?start=7&length=5)

Suggest not to depend on `click`. The user of the library has the freedom to use any framework other than `click`.

src/sonic-py-common/sonic_py_common/general.py
```diff
@@ -43,7 +44,7 @@ def getstatusoutput_noshell(cmd):
     return exitcode, output
 
 
-def getstatusoutput_noshell_pipe(cmd0, *args):
+def getstatusoutput_noshell_pipe(cmd0, *args, return_cmd=True):
```
qiluo-msft, line 47: >return_cmd [](http://example.com/codeflow?start=46&length=10)

Possible to fix the bug without new parameter?

qiluo-msft, line 47: >getstatusoutput_noshell_pipe [](http://example.com/codeflow?start=4&length=28)

Add a unit test for this bug fix?

maipbui: The user can choose to display or not to display the output. Without new parameter, the program forces to display the output?

qiluo-msft: I see you added a unit test. However this PR is fixing something specially for "support displaying output while command is running", not for a quick run command. Is it possible to test it? The test case should fail on current code.

PR #14034, opened 2023-03-01
Revert "[submodule] Advance sonic-swss-common pointer"

liushilongbuaa:
Reverts sonic-net/sonic-buildimage#13950
libswsscommon-dbg_1.0.0_amd64.deb package is not built in new submodule HEAD.
PR validation passed because it didn't build debug image.

13950 breaks debug image build.


liat-grozovik (approved):

liat-grozovik:
the main problem is that with this revert we are missing a lot of changes coming from swss-common which are needed just because debug image is not available.
@lguohan  is this the right approach?
what is the offended commit that cause it? maybe that one can be reverted on swss-common so we can have the rest in master?

qiluo-msft (approved):

PR #14051, opened 2023-03-02
[ci] Add debug image in PR checks to check debug packages.

liushilongbuaa:
#### Why I did it
Build debug image in PR checks to avoid official build failure in debug image.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -63,6 +63,7 @@ stages:
       jobGroups:
       - name: broadcom
         variables:
+          dbg_image: yes
           swi_image: yes
       - name: mellanox
         variables:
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -63,12 +63,16 @@ stages:
       jobGroups:
       - name: broadcom
         variables:
+          dbg_image: yes
```
xumia, line 66: How about using the variable INSTALL_DEBUG_TOOLS=y, do we want to build release version and the debug version for PR build, or the debug version is good enough for PR build?

PR #14085, opened 2023-03-06
[system-health] Make check interval more accurate

Junchao-Mellanox:
#### Why I did it

Healthd check system status every 60 seconds. However, running checker may take several seconds. Say checker takes X seconds, healthd takes (60 + X) seconds to finish one iteration. This implementation makes sonic-mgmt test case not so stable because the value X is hard to predict and different among different platforms. This PR introduces an interval 
compensation mechanism to healthd main loop.

#### How I did it

Introduces an interval compensation mechanism to healthd main loop: healthd should wait (60 - X) seconds for next iteration

#### How to verify it

Manual test 
Unit test

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [x] 202205
- [x] 202211

### DIFF
--- a/src/system-health/scripts/healthd
+++ b/src/system-health/scripts/healthd
@@ -7,6 +7,7 @@
 
 import signal
 import threading
+import time
 
 from sonic_py_common.daemon_base import DaemonBase
 from swsscommon.swsscommon import SonicV2Connector
@@ -79,17 +80,26 @@ class HealthDaemon(DaemonBase):
                 return
             sysmon = Sysmonitor()
             sysmon.task_run()
-            while 1:
-                stat = manager.check(chassis)
-                self._process_stat(chassis, manager.config, stat)
-
-                if self.stop_event.wait(manager.config.interval):
-                    break
+            while self._run_checker(manager, chassis):
+                pass
         except ImportError:
             self.log_warning("sonic_platform package not installed. Cannot start system-health daemon")
 
         self.deinit()
         sysmon.task_stop()
+
+    def _run_checker(self, manager, chassis):
+        begin = time.time()
+        stat = manager.check(chassis)
+        self._process_stat(chassis, manager.config, stat)
+        elapse = time.time() - begin
+        sleep_time_in_sec = manager.config.interval - elapse
+        if sleep_time_in_sec < 0:
+            self.log_notice(f'System health takes {elapse} seconds for one iteration')
+            sleep_time_in_sec = 1
+        if self.stop_event.wait(sleep_time_in_sec):
+            return False
+        return True
 
     def _process_stat(self, chassis, config, stat):
         from health_checker.health_checker import HealthChecker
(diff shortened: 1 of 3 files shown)


keboliu (approved):

prgeor:
@Junchao-Mellanox can you elaborate more on this `This implementation makes sonic-mgmt test case not so stable`

Junchao-Mellanox:
> @Junchao-Mellanox can you elaborate more on this `This implementation makes sonic-mgmt test case not so stable`

Currently, sonic-mgmt has a test case like this:
1. mock something, e.g. PSU high temperature
2. wait 70 seconds 
3. check redis database and verify healthd put PSU high temperature error to STATE DB

The case failed intermittently because sometimes 70 seconds is not enough. If we introduce a time compensation to healthd main loop, healthd loop interval will be more accurate and 70 seoncds is enough. BTW, time compensation is a common way in sonic.

src/system-health/scripts/healthd
```diff
@@ -79,18 +80,27 @@ class HealthDaemon(DaemonBase):
                 return
             sysmon = Sysmonitor()
             sysmon.task_run()
-            while 1:
-                stat = manager.check(chassis)
-                self._process_stat(chassis, manager.config, stat)
-
-                if self.stop_event.wait(manager.config.interval):
-                    break
+            while self._run_checker(manager, chassis):
```
prgeor, line 83: previously healthd was running continuously in while 1 loop. with this change the loop can break and healthd wont run. why this is changed?

src/system-health/scripts/healthd
```diff
@@ -79,18 +80,27 @@ class HealthDaemon(DaemonBase):
...
+        begin = time.time()
+        stat = manager.check(chassis)
+        self._process_stat(chassis, manager.config, stat)
+        elapse = time.time() - begin
+        sleep_time_in_sec = manager.config.interval - elapse
+        if sleep_time_in_sec < 0:
+            self.log_notice(f'System health takes {elapse} seconds for one iteration')
+            sleep_time_in_sec = 1
+        if self.stop_event.wait(sleep_time_in_sec):
+            return False
```
prgeor, line 101: if return 'False' healthd will stop. can we just break?

src/system-health/scripts/healthd
Junchao-Mellanox: Previous the loop still can be break, this logic does not change actually.
```
            while 1:
                stat = manager.check(chassis)
                self._process_stat(chassis, manager.config, stat)

                if self.stop_event.wait(manager.config.interval):
                    break
```

src/system-health/scripts/healthd
Junchao-Mellanox: if stop_event is set, it means that healthd got a signal, and it should stop.

src/system-health/scripts/healthd
Junchao-Mellanox: @prgeor

src/system-health/scripts/healthd
Junchao-Mellanox: @prgeor

prgeor (approved):

prgeor:
@liat-grozovik @lguohan please help merge

liat-grozovik (approved):

PR #14129, opened 2023-03-07
Syncd graceful shutdown should not be ignored

Junchao-Mellanox:
Depends on https://github.com/sonic-net/sonic-sairedis/pull/1216

Why I did this?

Due to a limitation in mellanox platform, sonic ignores syncs graceful shutdown in syncd.sh previously. However, ignoring graceful shutdown in syncd.sh also ignores closing other resources such as removing all counters. 
This PR changes syncd shutdown flow for mellanox platform to make sure it execute graceful shutdown flow. There is another PR for sonic-sairedis to make sure mellanox platform does not call remove_switch for cold shutdown.

How I test this?

Running config reload on many platforms and check result.


nazariig (dismissed):

liat-grozovik (approved):

Junchao-Mellanox:
sairedis PR has merged, pending submodule update PR https://github.com/sonic-net/sonic-buildimage/pull/14235

Junchao-Mellanox:
Hi @saiarcot895, @qiluo-msft , could you please help review this?

dgsudharsan (approved):

Junchao-Mellanox:
Hi @saiarcot895, @qiluo-msft , kindly reminder.

keboliu:
@saiarcot895 @qiluo-msft  would you please help to review?

files/scripts/syncd.sh
```diff
@@ -104,34 +104,32 @@ function stopplatform1() {
         debug "Stopped pmon service"
     fi
 
-    if [[ x$sonic_asic_platform != x"mellanox" ]] || [[ x$TYPE != x"cold" ]]; then
```
qiluo-msft, line 107: >[[ x$TYPE != x"cold" ]]; [](http://example.com/codeflow?start=53&length=24)

Why remove the condition? Maybe you intend only to remove the sonic_asic_platform check.

Junchao-Mellanox: Hi @qiluo-msft , we intend to remove the whole if block because:

- if platform is not mellanox, it will trigger graceful shutdown flow no matter what type it is. It means that removing the condition won't affect any other vendor. 
- if platform is mellanox, it will trigger graceful shutdown if type is not cold (old logic). This is exactly what we try to fix. We move the `type != cold` check to Syncd.cpp, see PR https://github.com/sonic-net/sonic-sairedis/pull/1216 .

keboliu: @qiluo-msft  would you please check the feedback from Junchao?

qiluo-msft (approved):

liat-grozovik:
@Junchao-Mellanox  IMO this fix is not relevant any more and the sairedis changes were reverted.
I am closing the PR, if you wish to have it back please reopen.

PR #14238, opened 2023-03-14
Finalize fast-reboot in warmboot finalizer

arfeigin:
This PR should come along with sonic-utilities PRs (https://github.com/sonic-net/sonic-utilities/pull/2621 https://github.com/sonic-net/sonic-utilities/pull/2734, already merged) removing the timer usage from fast-reboot script, and as well with advancing sonic-sairedis pointer to point merged https://github.com/sonic-net/sonic-sairedis/pull/1213.
This should be merged alongside with submodule update PR for sonic-utilities and sonic-sairedis with the above changes.

It will be followed by:
https://github.com/sonic-net/sonic-swss-common/pull/742
https://github.com/sonic-net/sonic-platform-daemons/pull/335

This set of PRs solves the issue https://github.com/sonic-net/sonic-buildimage/issues/13251

#### Why I did it
To solve an issue with upgrade with fast-reboot including FW upgrade which has been introduced since moving to fast-reboot over warm-reboot infrastructure.
As well, this introduces fast-reboot finalizing logic to determine fast-reboot is done.

#### How I did it
Added logic to finalize-warmboot script to handle fast-reboot as well, this makes sense as using fast-reboot over warm-reboot this script will be invoked. The script will clear fast-reboot entry from state-db instead of previous implementation that relied on timer. The timer could expire in some scenarios between fast-reboot finished causing fallback to cold-reboot and possible crashes.

As well this PR updates all services/scripts reading fast-reboot state-db entry to look for the updated value representing fast-reboot is active.

#### How to verify it
Run fast-reboot and check that fast-reboot entry exists in state-db right after startup and being cleared as warm-reboot is finalized and not due to a timer.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [X] 202211

### DIFF
--- a/files/image_config/ntp/ntp-config.sh
+++ b/files/image_config/ntp/ntp-config.sh
@@ -8,11 +8,11 @@ reboot_type='cold'
 function get_database_reboot_type()
 {
     SYSTEM_WARM_START=`sonic-db-cli STATE_DB hget "WARM_RESTART_ENABLE_TABLE|system" enable`
-    SYSTEM_FAST_START=`sonic-db-cli STATE_DB get "FAST_REBOOT|system"`
+    SYSTEM_FAST_START=`sonic-db-cli STATE_DB hget "FAST_RESTART_ENABLE_TABLE|system" enable`
 
     if [[ x"${SYSTEM_WARM_START}" == x"true" ]]; then
         reboot_type='warm'
-    elif [[ x"${SYSTEM_FAST_START}" == x"1" ]]; then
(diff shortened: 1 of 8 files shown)


vaibhavhd (approved):

vaibhavhd:
Please also rename the PR to reflect that this is also a submodule advance PR.

liat-grozovik (approved):

PR #14239, opened 2023-03-14
[dns] Add support for static DNS configuration.

oleksandrivantsiv:
#### Why I did it
Add support for static DNS configuration. According to https://github.com/sonic-net/SONiC/pull/1262 HLD.

#### How I did it
Add a new `resolv-config.service` that is responsible for transferring configuration from Config DB into `/etc/resolv.conf` file that is consumed by various subsystems in Linux to resolve domain names into IP addresses.

#### How to verify it
- Run the image compilation. Each component related to the static DNS feature is covered with the unit tests.
- Run sonic-mgmt tests. Static DNS feature will be covered with the system tests.
- Install the image and run manual tests.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211

### DIFF
--- a/device/nokia/armhf-nokia_ixs7215_52x-r0/platform.json
+++ b/device/nokia/armhf-nokia_ixs7215_52x-r0/platform.json
@@ -267,16 +267,16 @@
                 "name": "Ethernet47"
             },
             {
-                "name": "Ethernet48"
+                "name": "SFP48"
             },
             {
-                "name": "Ethernet49"
+                "name": "SFP49"
             },
             {
-                "name": "Ethernet50"
+                "name": "SFP50"
             },
             {
-                "name": "Ethernet51"
+                "name": "SFP51"
             }
        	]
     },
--- a/platform/marvell-armhf/sonic-platform-nokia/7215/sonic_platform/sfp.py
+++ b/platform/marvell-armhf/sonic-platform-nokia/7215/sonic_platform/sfp.py
@@ -1,19 +1,21 @@
-#############################################################################
-# Nokia
+# Name: sfp.py, version: 1.0
 #
-#############################################################################
-
-import subprocess
+# Description: Module contains the definitions of SFP related APIs
+# for Nokia IXR 7250 platform.
+#
+# Copyright (c) 2023, Nokia
+# All rights reserved.
+#
 
 try:
-    from sonic_platform_base.sfp_base import SfpBase
-    from sonic_platform_base.sonic_sfp.sff8472 import sff8472InterfaceId
-    from sonic_platform_base.sonic_sfp.sff8472 import sff8472Dom
-    from sonic_platform_base.sonic_sfp.sfputilhelper import SfpUtilHelper
-    from sonic_py_common import logger
-    from sonic_py_common.general import getstatusoutput_noshell
+    from sonic_platform_base.sonic_xcvr.sfp_optoe_base import SfpOptoeBase
+    from sonic_py_common.logger import Logger
+    from sonic_py_common import device_info
+
 except ImportError as e:
     raise ImportError(str(e) + "- required module not found")
+
+import subprocess as cmd
 
 smbus_present = 1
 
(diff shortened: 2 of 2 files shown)


ganglyu:
Does this feature cover DNS configuration inside container? For example, "docker exec telemetry cat /etc/resolv.conf"

oleksandrivantsiv:
> Does this feature cover DNS configuration inside container? For example, "docker exec telemetry cat /etc/resolv.conf"

Yes. During the container start the docker daemon copies "/etc/resolv.conf" from the host OS rootfs to the docker container rootfs.
But if the DNS configuration is changed in the run time when the container is running this change won't be reflected inside the docker container. In this case config reload will be required.

qiluo-msft:
> 

`config reload` will impact data plane, which is not necessary for DNS config. Could you improve this PR to update DNS config inside each container?

files/image_config/resolv-config/resolv-config.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash
+
+WD=/var/run/resolvconf/
+CONFIG_DIR=${WD}/interface/
+STATIC_CONFIG_FILE=mgmt.static
+DYNAMIC_CONFIG_FILE_TEMPLATE=*.dhclient
+
+start()
+{
+    redis-dump -d 4 -k "DNS_NAMESERVER*" -y > /tmp/dns.json
```
qiluo-msft, line 10: >-d 4 [](http://example.com/codeflow?start=15&length=4)

If your intention is to check if the table is empty, no need to dump.

4 is magic number, you may consider `sonic-db-cli` with `CONFIG_DB`.

@judyjoseph to check multi-asic and T2.

qiluo-msft:
> echo 'KUBELET_EXTRA_ARGS="--resolv-conf=/etc/resolv.conf --cgroup-driver=cgroupfs --node-ip=::"' | sudo tee -a  $FILESYSTEM_ROOT/etc/default/kubelet

Wondering if KUBELET_EXTRA_ARGS is impacted by this new feature?

@lixiaoyuner could you help review?

---
Refers to: files/build_templates/sonic_debian_extension.j2:477 in 4fd0180. [](commit_id = 4fd01802a87f2edfe5ac75651ac5e3c7dacf146e, deletion_comment = False)

files/image_config/resolv-config/resolv.conf.j2
```diff
@@ -0,0 +1,3 @@
+{% for ip in DNS_NAMESERVER|sort %}
+nameserver {{ ip }}
```
qiluo-msft, line 2: >nameserver [](http://example.com/codeflow?start=0&length=10)

In some use case, people are using "/etc/resolv.conf" to manually config DNS entries. In this PR, the content will be overwrite by j2 rendering.

Could you still allow old use case?

src/sonic-config-engine/tests/data/dns/static_dns.json
```diff
@@ -0,0 +1,6 @@
+{
+    "DNS_NAMESERVER": {
```
qiluo-msft, line 2: >DNS_NAMESERVER [](http://example.com/codeflow?start=5&length=14)

Need new yang model for this new ConfigDB table.
Also need yang unit test.

oleksandrivantsiv: Yang model and tests are added: https://github.com/sonic-net/sonic-buildimage/pull/13834

files/image_config/resolv-config/resolv.conf.j2
oleksandrivantsiv: Those are mutually exclusive approaches. If the user wants to edit /etc/resolv.conf manually it should not configure DNS via config DB and vice-versa. The default behavior is preserved if no DNS entries are specified in the config DB.

PR #14303, opened 2023-03-17
Fix telemetry.sh passing in null as log level value

zbud-msft:
#### Why I did it

Bug in script that was passing in null as log level value if missing from config_db

#### How I did it

Added more robust conditional statement

#### How to verify it

1) Remove log_level from config db
2) config reload -y
3) telemetry should not crash

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211

### DIFF
--- a/dockers/docker-sonic-telemetry/telemetry.sh
+++ b/dockers/docker-sonic-telemetry/telemetry.sh
@@ -63,10 +63,10 @@ if [ -z $CLIENT_AUTH ] || [ $CLIENT_AUTH == "false" ]; then
 fi
 
 LOG_LEVEL=$(echo $GNMI | jq -r '.log_level')
-if [ -z $LOG_LEVEL ] || [ $LOG_LEVEL == "null" ]; then
-    TELEMETRY_ARGS+=" -v=2"
-else
+if [[ $LOG_LEVEL =~ ^[0-9]+$ ]]; then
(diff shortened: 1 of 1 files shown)


dockers/docker-sonic-telemetry/telemetry.sh
```diff
@@ -63,10 +63,10 @@ if [ -z $CLIENT_AUTH ] || [ $CLIENT_AUTH == "false" ]; then
 fi
 
 LOG_LEVEL=$(echo $GNMI | jq -r '.log_level')
-if [ ! -z $LOG_LEVEL ]; then
-    TELEMETRY_ARGS+=" -v=$LOG_LEVEL"
-else
+if [ -z $LOG_LEVEL ] || [ $LOG_LEVEL == "null" ]; then
```
qiluo-msft, line 66: >$LOG_LEVEL [](http://example.com/codeflow?start=26&length=10)

If you expect only numbers, check for numbers as allowlist. ref: [https://stackoverflow.com/a/2210386/2514803](https://stackoverflow.com/a/2210386/2514803)

zbud-msft: Done

qiluo-msft (approved):

PR #14347, opened 2023-03-20
Update sonic-py-common, add missing dependency to redis-dump-load.

liuh-80:
Update sonic-py-common, add missing dependency to redis-dump-load.

#### Why I did it
The script sonic_db_dump_load.py in sonic-py-common is depends on redis-dump-load, however the dependency is missing.

#### How I did it
Add redis-dump-load dependency.

#### How to verify it
Pass all E2E test case.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205

Update sonic-py-common, add missing dependency to redis-dump-load.

### DIFF
--- a/src/sonic-py-common/setup.py
+++ b/src/sonic-py-common/setup.py
@@ -1,10 +1,30 @@
+from __future__ import print_function
+import sys
 from setuptools import setup
+import pkg_resources
+from packaging import version
+
+# sonic_dependencies, version requirement only supports '>='
+sonic_dependencies = ['redis-dump-load']
 
 dependencies = [
     'natsort==6.2.1', # 6.2.1 is the last version which supports Python 2
     'pyyaml',
-    'redis-dump-load',
 ]
+
+dependencies += sonic_dependencies
+for package in sonic_dependencies:
+    try:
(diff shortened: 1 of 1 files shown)


src/sonic-py-common/setup.py
```diff
@@ -3,6 +3,7 @@
 dependencies = [
     'natsort==6.2.1', # 6.2.1 is the last version which supports Python 2
     'pyyaml',
+    'redis-dump-load',
```
qiluo-msft, line 6: >redis-dump-load [](http://example.com/codeflow?start=5&length=15)

Thanks for the fixing!
Since we build redis-dump-load with patch, we never expect setup.py automatically download it from pypi. We enforce this rule by treating it as "sonic_dependencies". Could you follow the sample in [https://github.com/sonic-net/sonic-buildimage/pull/13850/files](https://github.com/sonic-net/sonic-buildimage/pull/13850/files)

liuh-80: Fixed, also create a manually cherry-pick PR for 202211 and 202205: https://github.com/sonic-net/sonic-buildimage/pull/14360

qiluo-msft (approved):

PR #14351, opened 2023-03-20
Always use the openssl backend of curl

saiarcot895:
This should make sure that the openssl backend of curl is used, instead of the gnutls or nss backend, for consistency reasons.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [x] 202205
- [x] 202211

### DIFF
--- a/sonic-slave-bullseye/Dockerfile.j2
+++ b/sonic-slave-bullseye/Dockerfile.j2
@@ -202,7 +202,7 @@ RUN apt-get update && apt-get install -y \
         libxml2-dev \
 # For BFN sdk build
         libusb-1.0-0-dev \
-        libcurl3-nss-dev \
+        libcurl4-openssl-dev \
         libunwind8-dev \
         telnet \
         libc-ares2 \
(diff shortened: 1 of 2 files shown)


ganglyu (approved):

xumia (approved):

sonic-slave-bullseye/Dockerfile.j2
```diff
@@ -202,7 +202,7 @@ RUN apt-get update && apt-get install -y \
         libxml2-dev \
 # For BFN sdk build
         libusb-1.0-0-dev \
-        libcurl3-nss-dev \
+        libcurl4-openssl-dev \
```
xumia, line 205: Can we change the one in sonic-slave-buster/Dockerfile.j2 as well? We are still using the docker image.

saiarcot895: Done

saiarcot895:
@k-v1 I think it would be better for feature/capability consistency (and some minor space savings) that everything uses the openssl version of curl, unless there's a very good reason that some other backend needs to be used.

bluecmd:
I filed https://github.com/sonic-net/sonic-mgmt-framework/issues/130 to investigate changing it to openssl as well instead of gnutls.

PR #14385, opened 2023-03-22
[202012] Check config file not empty after modify it in hostcfgd. 

liuh-80:
**What I did**
Check /etc/pam.d/sshd integrity after modify it in hostcfgd.

**Why I did it**
Found some incident that /etc/pam.d/sshd become empty file during OR upgrade. 

**How I verified it**
Pass all UT.
Add new UT to cover new code.

**Details if related**
This is a manually cherry-pick PR for https://github.com/sonic-net/sonic-host-services/pull/36


liuh-80:
Because we can only cherry pick from master branch to older branch, so will create manuallt chertry-pick PR for 201811 and 201911

PR #14405, opened 2023-03-24
[Build] disable DOCKER_BUILDKIT if SONIC_USE_DOCKER_BUILDKIT is not set

### LINKED ISSUE #14081: Disable Docker BuildKit explicitly
Docker 23.0.0 (1 Feb 2023) Release notes:

- Set Buildx and BuildKit as the default builder on Linux.
- The legacy builder can still be used by explicitly setting DOCKER_BUILDKIT=0.

Probably need to explicitly set DOCKER_BUILDKIT=0 in slave.mk if SONIC_USE_DOCKER_BUILDKIT is disabled.

k-v1:
#### Why I did it

Fix #14081
By default DOCKER_BUILDKIT is enabled after docker version 23.0.0
So we need to disable it explicitly if SONIC_USE_DOCKER_BUILDKIT is not set.
Otherwise it will produce larger installable images.

#### How I did it

set DOCKER_BUILDKIT=0 in slave.mk

### DIFF
--- a/slave.mk
+++ b/slave.mk
@@ -453,6 +453,8 @@ include Makefile.cache
 ifeq ($(SONIC_USE_DOCKER_BUILDKIT),y)
 $(warning "Using SONIC_USE_DOCKER_BUILDKIT will produce larger installable SONiC image because of a docker bug (more details: https://github.com/moby/moby/issues/38903)")
 export DOCKER_BUILDKIT=1
+else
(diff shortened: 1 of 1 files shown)


lguohan (approved):

lguohan:
thanks for explanation, approved, and will merge once test passed.

xumia (approved):

PR #14407, opened 2023-03-24
[Build] don't install rsync if vcache is disabled

k-v1:
#### Why I did it

We don't need to install rsync in every docker container if vcache is disabled.

#### How I did it

Install rsync in pre_run_buildinfo script only if vcache is enabled.


xumia (approved):

xumia:
LGTM, the build failure may be caused by not running the command "apt-get update". I sent another PR to fix it, https://github.com/sonic-net/sonic-buildimage/pull/14439. @k-v1, you also cherry-pick the commit in your PR, or rerun the test after the PR merged.

PR #14423, opened 2023-03-27
[201811] Check config file not empty after modify it in hostcfgd.

liuh-80:
**What I did**
Check /etc/pam.d/sshd integrity after modify it in hostcfgd.

**Why I did it**
Found some incident that /etc/pam.d/sshd become empty file during OR upgrade. 

**How I verified it**
Pass all UT.

**Details if related**
This is a manually cherry-pick PR for https://github.com/sonic-net/sonic-host-services/pull/36


liuh-80:
201811/201911 build break, seems because jessie not supported by all debian sources.

liuh-80:
Close and re-open to trigger validation

PR #14513, opened 2023-04-04
Add remote syslog configuration

iavraham:
* Add an ability to configure remote syslog servers
* Add an initial configuration for remote syslog
* Extend YANG module and add unit tests

#### Why I did it
Adding the following functionality to rsyslog feature:

- Configure remote syslog servers: protocol, filter, severity level
- Update global syslog configuration: severity level, message format

#### How I did it
added parameters to syslog server and global configuration.

#### How to verify it
create syslog server using CLI/adding to Redis-DB
verify server is added to file /etc/rsyslog.conf and server is functional.

extend rsyslog capabilities, added server and global configuration parameters.

https://github.com/iavraham/sonic-buildimage/blob/master/src/sonic-yang-models/yang-models/sonic-syslog.yang


src/sonic-yang-models/tests/yang_model_tests/tests_config/syslog.json
```diff
@@ -136,17 +136,6 @@
             }
         }
     },
-    "SYSLOG_SERVER_INVALID_IPADDR_TEST" : {
```
dgsudharsan, line 139: Can you please add this back?

dgsudharsan (changes requested):

files/image_config/rsyslog/rsyslog-config.sh
```diff
@@ -17,7 +17,10 @@ if [[ ($NUM_ASIC -gt 1) ]]; then
 else
     udp_server_ip=$(ip -j -4 addr list lo scope host | jq -r -M '.[0].addr_info[0].local')
 fi
+hostname=$(hostname)
```
qiluo-msft, line 20: >hostname [](http://example.com/codeflow?start=0&length=8)

There is DEVICE_METADATA/hostname in ConfigDB. Could you check which one is better?

iavraham: it is better to take from linux than DEVICE_METADATA/hostname. 
example: DB can be empty if not configured with CLI. hostname can also be dynamically assigned using DHCP.

dgsudharsan:
@iavraham Can you please update configuration.md to reflect the new schema changes?

fastiuk:
> @iavraham Can you please update configuration.md to reflect the new schema changes?

done

src/sonic-yang-models/tests/yang_model_tests/tests_config/syslog.json
dgsudharsan, line 139: Why is this test removed? Is it because syslog server can be a string and hence IP validation cannot be done?

dgsudharsan (approved):

files/image_config/rsyslog/rsyslog.conf.j2
```diff
@@ -91,25 +90,28 @@ $RepeatedMsgReduction on
...
+
+{% set port = conf.get('port', 514) -%}
+{% set proto = conf.get('protocol', 'udp') -%}
+{% set vrf = conf.get('vrf', 'default') -%}
+{% set severity = conf.get('severity', gconf.get('severity', 'notice')) -%}
+{% set filter = conf.get('filter') -%}
+{% set regex = conf.get('filter_regex') -%}
+
+{% set fmodifier = '!' if filter == 'exclude' else '' %}
+{% set device = 'mgmt' if vrf in ['mgmt', 'management'] else 'eth0' %}
```
dgsudharsan, line 109: This assumes that syslog server is reachable only through mgmt network or eth0. What if the server is reachable through the data network or OOB mgmt port?

dgsudharsan (changes requested):

files/image_config/rsyslog/rsyslog.conf.j2
fastiuk: oh, yes. I missed the `address` as well. I will fix it.
Good catch!

fastiuk: Updated

src/sonic-yang-models/tests/yang_model_tests/tests_config/syslog.json
fastiuk: yes

fastiuk:
> Commenter does not have sufficient privileges for PR 14513 in repo sonic-net/sonic-buildimage

lol :)

fastiuk:
Hey all, the test here will fail without changes in https://github.com/sonic-net/sonic-utilities/pull/2843
When that PR will be merged - that test should pass

dgsudharsan (approved):

fastiuk:
/azpw run Azure.sonic-buildimage (Test kvmtest-t0 by Elastictest)

fastiuk:
/azpw run "Azure.sonic-buildimage (Test kvmtest-t0 by Elastictest)"

fastiuk:
Wrong ip address in syslog tests:
```
Jun 17 15:16:06.266392 vlab-01 ERR rsyslogd: omfwd: could not get addrinfo for hostname '10.0.0.587':'514': Temporary failure in name resolution [v8.2302.0 try https://www.rsyslog.com/e/2007 ]
```
The address is `10.0.0.587` which is wrong. That is why test is failing

fastiuk:
I update the tests to pass here:
https://github.com/sonic-net/sonic-mgmt/pull/8668

But tests must be merged before this PR in order to pass here

dgsudharsan:
Depends on https://github.com/sonic-net/sonic-mgmt/pull/8668. @qiluo-msft  @yxieca  Can you please help to review and merge this sonic-mgmt PR?

files/build_templates/init_cfg.json.j2
```diff
@@ -110,7 +110,9 @@
     "SYSLOG_CONFIG": {
         "GLOBAL": {
             "rate_limit_interval" : "0",
-            "rate_limit_burst" : "0"
+            "rate_limit_burst" : "0",
```
qiluo-msft, line 113: >rate_limit_burst [](http://example.com/codeflow?start=13&length=16)

For each GLOBAL attribute, could you define default values in yang model, then you can completely remove SYSLOG_CONFIG table in this file.

files/image_config/rsyslog/rsyslog-config.sh
qiluo-msft: If DEVICE_METADATA/hostname has value, it should be preferred.

files/build_templates/init_cfg.json.j2
fastiuk: We are not generating defaults from the YANG model, so that is why we place it here. It cannot be removed

files/image_config/rsyslog/rsyslog-config.sh
fastiuk: The hostname can be dynamically assigned by DHCP, in that case, DB value may contain the wrong hostname.
Plus `DEVICE_METADATA` is a configuration and not a STATE, once data is placed to DEVICE_METADATA it will update the system hostname according to https://github.com/sonic-net/sonic-host-services/blob/bc08806b64002c506b8401eae5d9e1c760651e49/scripts/hostcfgd#L1568
So in my opinion it is better to take it from the `hostname` utility as the most appropriate one.

fastiuk:
@qiluo-msft all comments addressed. Please take a look and if everything looks good to you, please merge that PR

wen587 (approved):
LGTM

files/build_templates/init_cfg.json.j2
dgsudharsan: @qiluo-msft I think this is a generic comment which needs to be addressed for all fields (and may be for all tables in init_cfg.j2). I would suggest lets open a task and address this change independently rather than being part of this PR.

fastiuk:
@qiluo-msft updated according to the last comment. Could you please take a look?

files/build_templates/init_cfg.json.j2
qiluo-msft: I did not mean a generic comment. I only ask to stop adding new fields into this file:
            "format": "standard",
            "severity": "notice"

The quickest solution is to add hard-coded default value into rsyslog.config.j2, like default port which already in that file.

fastiuk: > The quickest solution is to add hard-coded default value into rsyslog.config.j2, like default port which already in that file.

@qiluo-msft Done

src/sonic-utilities
qiluo-msft, line 1: Most of sonic-utilities commits are not relevant. Could you move submodule in a separate PR?

qiluo-msft:
Hard coded `eth0` is fixed in a future https://github.com/sonic-net/sonic-buildimage/pull/17616

PR #14522, opened 2023-04-04
fix long delay of stopping syncd container in warm-reboot

jhli-cisco:
#### Why I did it
We have seen LACP flapping in warm-reboot testing. The reason is that the time between teamd sending the last LACPDU before going down and the 1st LACPDU after teamd coming up after warm-reboot is more than 90 seconds. One cause contributes to this long warm-reboot time is syncd container takes more than 20 seconds to be shut down:

Apr  4 20:42:01.710695 t0-yy38 NOTICE root: Stopping syncd service...
Apr  4 20:42:01.731929 t0-yy38 NOTICE root: Warm boot flag: syncd true.
Apr  4 20:42:01.735908 t0-yy38 NOTICE root: Fast boot flag: syncd false.
Apr  4 20:42:01.982858 t0-yy38 NOTICE root: warm shutdown syncd process ...
Apr  4 20:42:08.245574 t0-yy38 NOTICE root: Finished warm shutdown syncd process ...  <----- syncd process has been gracefully shut down up to this point. It takes ~7 seconds. 

Apr  4 20:42:09.532254 t0-yy38 INFO syncd#supervisord 2023-04-04 20:42:09,531 WARN received SIGTERM indicating exit request
Apr  4 20:42:09.532494 t0-yy38 INFO syncd#supervisord 2023-04-04 20:42:09,532 INFO waiting for supervisor-proc-exit-listener, rsyslogd, syncd to die
Apr  4 20:42:12.536981 t0-yy38 INFO syncd#supervisord 2023-04-04 20:42:12,536 INFO waiting for supervisor-proc-exit-listener, rsyslogd, syncd to die
Apr  4 20:42:15.541243 t0-yy38 INFO syncd#supervisord 2023-04-04 20:42:15,540 INFO waiting for supervisor-proc-exit-listener, rsyslogd, syncd to die
Apr  4 20:42:18.545492 t0-yy38 INFO syncd#supervisord 2023-04-04 20:42:18,545 INFO waiting for supervisor-proc-exit-listener, rsyslogd, syncd to die
Apr  4 20:42:21.865734 t0-yy38 INFO container: docker cmd: wait for syncd
Apr  4 20:42:21.866022 t0-yy38 INFO container: docker cmd: stop for syncd
Apr  4 20:42:21.894735 t0-yy38 NOTICE root: Stopped syncd service...

 It takes another 12 seconds to bring down the container after syncd process is gracefully shut down.

The root cause for the long delay is supervisord can't manage a run-to-complete process properly, in this case is the syncd process returned from main() after gracefully shutdown, but supervisord is still thinking syncd is RUNNING. When stopping syncd container, supervisord has to wait until timeout for SIGTERM and forcefully kill it because syncd process is no longer running.

#### How I did it
run stopping syncd container at background to not block the progress of warm-reboot. This is safe as syncd process has already gracefully shut down up to that point.

#### How to verify it
DO warm-reboot to verify lacp flapping is not seen.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012
- [ ] 202106
- [ ] 202111
- [x] 202205
- [x] 202211

### DIFF
--- a/files/scripts/syncd_common.sh
+++ b/files/scripts/syncd_common.sh
@@ -163,7 +163,7 @@ stop() {
 
     stopplatform1
 
-    /usr/bin/${SERVICE}.sh stop $DEV
+    /usr/bin/${SERVICE}.sh stop $DEV &
     debug "Stopped ${SERVICE}$DEV service..."
 
     stopplatform2
--- a/files/scripts/syncd_common.sh
+++ b/files/scripts/syncd_common.sh
@@ -163,7 +163,11 @@ stop() {
 
     stopplatform1
 
-    /usr/bin/${SERVICE}.sh stop $DEV &
+    if [[ x"$WARM_BOOT" == x"true" ]]; then
+        /usr/bin/${SERVICE}.sh stop $DEV &
+    else
+        /usr/bin/${SERVICE}.sh stop $DEV
+    fi
     debug "Stopped ${SERVICE}$DEV service..."
 
     stopplatform2
--- a/files/scripts/syncd_common.sh
+++ b/files/scripts/syncd_common.sh
@@ -163,7 +163,7 @@ stop() {
 
     stopplatform1
 
-    if [[ x"$WARM_BOOT" == x"true" ]]; then
+    if [[ x"$WARM_BOOT" == x"true" ]] && [[ x$sonic_asic_platform == x"cisco-8000" ]]; then
         /usr/bin/${SERVICE}.sh stop $DEV &
     else
         /usr/bin/${SERVICE}.sh stop $DEV


jhli-cisco:
@lguohan , can you please review ?

lguohan:
@vaibhavhd , can you check this one?

vaibhavhd:
> The root cause for the long delay is supervisord can't manage a run-to-complete process properly, in this case is the syncd process returned from main() after gracefully shutdown, but supervisord is still thinking syncd is RUNNING. When stopping syncd container, supervisord has to wait until timeout for SIGTERM and forcefully kill it because syncd process is no longer running.

I don't see this issue in my testing on a physical device. I understand the benefit of fix here, but rather then fixing the symptom, it will be better to fix the root cause, right?
Specifically, why ` supervisord is still thinking syncd is RUNNING. `? If syncd has exited from main, supervisor should act accordingly. The SIGTERM or forceful kill is not expected.

Hypothetically the issue that we see w/ syncd kill in your case can be seen w/ any other container too.
Then do you want to kill everything in the background?

jhli-cisco:
@vaibhavhd Thank you very much for taking time to review and try it out. Even the supervisord is notified with syncd exit, putting it into background still save a few seconds as there are other processes in the same container such as rsyslogd etc, which takes time to stop. During my warm-reboot testing, a few seconds is important to decide if port-channel flapping or not.

We dont need to change how other containers are stopped because syncd is the only container left to stop in warm-reboot path after teamd container is stopped.

vaibhavhd:
> @vaibhavhd Thank you very much for taking time to review and try it out. Even the supervisord is notified with syncd exit, putting it into background still save a few seconds as there are other processes in the same container such as rsyslogd etc, which takes time to stop. During my warm-reboot testing, a few seconds is important to decide if port-channel flapping or not.
> 
> We dont need to change how other containers are stopped because syncd is the only container left to stop in warm-reboot path after teamd container is stopped.

I agree that any benefit should be taken advantage of in this shutdown path. However, I still insist to fix the root cause _as well_.

> syncd is the only container left to stop in warm-reboot path after teamd container is stopped.

This is not true.

jhli-cisco:
Sure, will look into more as why supervisord is not notified when syncd process exit. But can I commit the current improvement?

There are more containers to stop after syncd, but they are not stopped individually, but all stopped together by stopping docker.service. They are non-blocking. The containers that need to be stopped individually and their order is defined in /etc/sonic/warm-reboot_order:
lldp mux nat radv sflow bgp swss teamd syncd

jhli-cisco:
@vaibhavhd Using most resent sonic-buildimage, supervisord does get notified when syncd program exit:

Apr 13 23:07:36.279778 t0-yy38 INFO syncd#supervisord 2023-04-13 23:07:36,279 INFO exited: syncd (exit status 0; expected)
Apr 13 23:07:36.428460 t0-yy38 NOTICE root: Finished warm shutdown syncd process ...
Apr 13 23:07:37.282340 t0-yy38 INFO syncd#supervisord 2023-04-13 23:07:37,281 WARN received SIGTERM indicating exit request
Apr 13 23:07:37.282340 t0-yy38 INFO syncd#supervisord 2023-04-13 23:07:37,281 INFO waiting for supervisor-proc-exit-listener, rsyslogd to die
Apr 13 23:07:41.397429 t0-yy38 INFO container: docker cmd: wait for syncd
Apr 13 23:07:41.397712 t0-yy38 INFO container: docker cmd: stop for syncd
Apr 13 23:07:41.426088 t0-yy38 NOTICE root: Stopped syncd service...
Apr 13 23:07:41.430155 t0-yy38 NOTICE root: Unlocking /tmp/swss-syncd-lock (10) from syncd service
Apr 13 23:07:41.435417 t0-yy38 INFO systemd[1]: syncd.service: Succeeded.
Apr 13 23:07:41.435923 t0-yy38 INFO systemd[1]: Stopped syncd service.
Apr 13 23:07:41.441506 t0-yy38 NOTICE cisco: Stopped syncd

But it still takes 4-5 seconds to stop syncd container/service. Putting it into background could save 4-5 seconds.

jhli-cisco:
@vaibhavhd , will you approve this PR?

vaibhavhd:
Requesting review from @kcudnik  and @yxieca too.

jhli-cisco:
@vaibhavhd limit the change to warm-reboot case only based on your concern.

This is safe for warm-reboot as syncd process already exit in stopplatform1 before container is even stopped. Swss is stopped even earlier.

vaibhavhd:
> * In this case what if syncd shutdown takes longer and dockerd is killed before even syncd warm shutdown is complete?
> * In the warmreboot case, we want to ensure that syncd warm shutdown completes before we kill database and dockerd. This is because warm shutdown of syncd involves an important difference of letting dataplane be up an running. If syncd can't complete warmshutdown part before dockerd is killed or kexec is issued, then we cannot guarantee that dataplane will be up and running.
> * What happens if warmboot.bin file is not completely generated as syncd shutdown is working in background, but kexec gets issued in foreground?

vaibhavhd:
> @vaibhavhd limit the change to warm-reboot case only based on your concern.
> 
> This is safe for warm-reboot as syncd process already exit in stopplatform1 before container is even stopped. Swss is stopped even earlier.

Thanks. My concerns are still open:

What if syncd shutdown takes longer and dockerd is killed before even syncd warm shutdown is complete?

In the warmreboot case, we want to ensure that syncd warm shutdown completes before we kill database and dockerd. This is because warm shutdown of syncd involves an important difference of letting dataplane be up an running. If syncd can't complete warmshutdown part before dockerd is killed or kexec is issued, then we cannot guarantee that dataplane will be up and running.

What happens if warmboot.bin file is not completely generated as syncd shutdown is working in background, but kexec gets issued in foreground?

jhli-cisco:
> > @vaibhavhd limit the change to warm-reboot case only based on your concern.
> > This is safe for warm-reboot as syncd process already exit in stopplatform1 before container is even stopped. Swss is stopped even earlier.
> 
> Thanks. My concerns are still open:
> 
> What if syncd shutdown takes longer and dockerd is killed before even syncd warm shutdown is complete?
> 
> In the warmreboot case, we want to ensure that syncd warm shutdown completes before we kill database and dockerd. This is because warm shutdown of syncd involves an important difference of letting dataplane be up an running. If syncd can't complete warmshutdown part before dockerd is killed or kexec is issued, then we cannot guarantee that dataplane will be up and running.
> 
> What happens if warmboot.bin file is not completely generated as syncd shutdown is working in background, but kexec gets issued in foreground?

The current warm-reboot waits 20 seconds for syncd process being shut down gracefully, which including dump of sai-warmboot.bin.  If this is not enough the syncd process will be shut down forcefully during container stop anyway. So this creates same problem irrespective of container stop is running in background or foreground. If 20 seconds is not enough to shut down syncd process, this will post serious issue with warm-reboot as the control planes will flap, and this really needs to be corrected at from syncd/sai.

From architecture pov, for warm-reboot to be successful, all components/steps involved need to behave within their allocated time boundary. If 20 seconds is the budget for shutting down syncd process, then it has to finish shutdown within 20 seconds. Otherwise there is no guarantee for a successful warmreboot.

jhli-cisco:
@vaibhavhd ???

kcudnik (approved):

jhli-cisco:
@lguohan can we please have this merged? It is required in 202205 and any other branches that warm-reboot are supported. Thanks.

files/scripts/syncd_common.sh
```diff
@@ -163,7 +163,11 @@ stop() {
 
     stopplatform1
 
-    /usr/bin/${SERVICE}.sh stop $DEV
+    if [[ x"$WARM_BOOT" == x"true" ]]; then
+        /usr/bin/${SERVICE}.sh stop $DEV &
```
kcudnik, line 167: why this goes to background ?

yxieca (changes requested):
I agree that pre-shutdown is done before this point. However, the shutdown is the place where the warm reboot date was finalized and persisted for the next boot up.

How do we know that sending syncd stop to background (potentially being killed later) won't affect the integrity of the warm reboot date?

jhli-cisco:
> 



> I agree that pre-shutdown is done before this point. However, the shutdown is the place where the warm reboot date was finalized and persisted for the next boot up.
> 
> How do we know that sending syncd stop to background (potentially being killed later) won't affect the integrity of the warm reboot date?

reboot reason/date is recorded after docker service being stopped at https://github.com/sonic-net/sonic-utilities/blob/master/scripts/fast-reboot#L790. I dont see how stopping syncd container in background/foreground could affect it.

files/scripts/syncd_common.sh
jhli-cisco: to warm boot time as explained in the whole thread

jhli-cisco:
@yxieca Any further comments?

yxieca:
> > 
> 
> > I agree that pre-shutdown is done before this point. However, the shutdown is the place where the warm reboot date was finalized and persisted for the next boot up.
> > How do we know that sending syncd stop to background (potentially being killed later) won't affect the integrity of the warm reboot date?
> 
> reboot reason/date is recorded after docker service being stopped at https://github.com/sonic-net/sonic-utilities/blob/master/scripts/fast-reboot#L790. I dont see how stopping syncd container in background/foreground could affect it.

The line you pointed out is to record the reboot cause. Which is not a conern.

The concern is that syncd, particularly, SAI needs to persist data so that on the boot up path, the data could be used to restore the asic states. The shutdown phase is when these data are fully generated and persisted. How can you guarantee the integrity of this data?

jhli-cisco:
> sai-warmboot.bin



> > > 
> > 
> > 
> > > I agree that pre-shutdown is done before this point. However, the shutdown is the place where the warm reboot date was finalized and persisted for the next boot up.
> > > How do we know that sending syncd stop to background (potentially being killed later) won't affect the integrity of the warm reboot date?
> > 
> > 
> > reboot reason/date is recorded after docker service being stopped at https://github.com/sonic-net/sonic-utilities/blob/master/scripts/fast-reboot#L790. I dont see how stopping syncd container in background/foreground could affect it.
> 
> The line you pointed out is to record the reboot cause. Which is not a conern.
> 
> The concern is that syncd, particularly, SAI needs to persist data so that on the boot up path, the data could be used to restore the asic states. The shutdown phase is when these data are fully generated and persisted. How can you guarantee the integrity of this data?

I see. I guess you refer to sai-warmboot.bin. This is generated as part of syncd graceful shutdown in pre-shutdown phase, which the current implementation waited for 20 seconds to finish before stopping syncd container. So at the time to stop syncd container, syncd process shall already gracefully exit, which means sai-warmboot.bin has been generated already.

jhli-cisco:
@yxieca any further comments?

yxieca:
> > sai-warmboot.bin
> 
> > > > 
> > > 
> > > 
> > > > I agree that pre-shutdown is done before this point. However, the shutdown is the place where the warm reboot date was finalized and persisted for the next boot up.
> > > > How do we know that sending syncd stop to background (potentially being killed later) won't affect the integrity of the warm reboot date?
> > > 
> > > 
> > > reboot reason/date is recorded after docker service being stopped at https://github.com/sonic-net/sonic-utilities/blob/master/scripts/fast-reboot#L790. I dont see how stopping syncd container in background/foreground could affect it.
> > 
> > 
> > The line you pointed out is to record the reboot cause. Which is not a conern.
> > The concern is that syncd, particularly, SAI needs to persist data so that on the boot up path, the data could be used to restore the asic states. The shutdown phase is when these data are fully generated and persisted. How can you guarantee the integrity of this data?
> 
> I see. I guess you refer to sai-warmboot.bin. This is generated as part of syncd graceful shutdown in pre-shutdown phase, which the current implementation waited for 20 seconds to finish before stopping syncd container. So at the time to stop syncd container, syncd process shall already gracefully exit, which means sai-warmboot.bin has been generated already.

This might be true for Cisco platform. But other platforms would persist data at final shutdown phase per warm reboot design. This change is not safe for all platforms.

jhli-cisco:
> > > sai-warmboot.bin
> > 
> > 
> > > > > 
> > > > 
> > > > 
> > > > > I agree that pre-shutdown is done before this point. However, the shutdown is the place where the warm reboot date was finalized and persisted for the next boot up.
> > > > > How do we know that sending syncd stop to background (potentially being killed later) won't affect the integrity of the warm reboot date?
> > > > 
> > > > 
> > > > reboot reason/date is recorded after docker service being stopped at https://github.com/sonic-net/sonic-utilities/blob/master/scripts/fast-reboot#L790. I dont see how stopping syncd container in background/foreground could affect it.
> > > 
> > > 
> > > The line you pointed out is to record the reboot cause. Which is not a conern.
> > > The concern is that syncd, particularly, SAI needs to persist data so that on the boot up path, the data could be used to restore the asic states. The shutdown phase is when these data are fully generated and persisted. How can you guarantee the integrity of this data?
> > 
> > 
> > I see. I guess you refer to sai-warmboot.bin. This is generated as part of syncd graceful shutdown in pre-shutdown phase, which the current implementation waited for 20 seconds to finish before stopping syncd container. So at the time to stop syncd container, syncd process shall already gracefully exit, which means sai-warmboot.bin has been generated already.
> 
> This might be true for Cisco platform. But other platforms would persist data at final shutdown phase per warm reboot design. This change is not safe for all platforms.

Would it be ok to limit the change to cisco platform only? We have tested it extensively on cisco-8000

jhli-cisco:
@yxieca limited the change to cisco-8000 only.

lguohan:
if we know this is the root cause, why not fix it?
> The root cause for the long delay is supervisord can't manage a run-to-complete process properly, in this case is the syncd process returned from main() after gracefully shutdown, but supervisord is still thinking syncd is RUNNING. When stopping syncd container, supervisord has to wait until timeout for SIGTERM and forcefully kill it because syncd process is no longer running.

jhli-cisco:
> if we know this is the root cause, why not fix it?
> 
> > The root cause for the long delay is supervisord can't manage a run-to-complete process properly, in this case is the syncd process returned from main() after gracefully shutdown, but supervisord is still thinking syncd is RUNNING. When stopping syncd container, supervisord has to wait until timeout for SIGTERM and forcefully kill it because syncd process is no longer running.

This is supervisor limitation. It's beyond platform or even sonic scope.

Aravind-Subbaroyan:
@lguohan - Are you good with the new change from @jhli-cisco ?

PR #14537, opened 2023-04-05
[Do Not Merge] Set no ecn marking for backplane ports

zhixzhu:
#### Why I did it
One option of ECN marking solution on Cisco T2 product, disable ECN marking on backplane ports, only do ECN marking on the chassis egress ports.

#### How I did it
Apply a different WRED profile to backplane ports.

#### How to verify it
Generate the qos config with the new qos template.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [x] 202205
--> 202205 branch is used for T2.
- [ ] 202211

### DIFF
--- a/files/build_templates/qos_config.j2
+++ b/files/build_templates/qos_config.j2
@@ -49,6 +49,11 @@
 
 {%- set require_global_dscp_to_tc_map = true -%}
 
+{%- set vars = {'backplane_port_exist': false} %}
+{% for port in PORT_ACTIVE if port.startswith('Ethernet-BP')%}
+{% if vars.update({'backplane_port_exist': true}) %} {% endif %}
+{% endfor %}
+
 {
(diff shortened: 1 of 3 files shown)


files/build_templates/qos_config.j2
```diff
@@ -49,6 +49,11 @@
 
 {%- set require_global_dscp_to_tc_map = true -%}
 
+{%- set vars = {'backplane_port_exist': false} %}
+{% for port in PORT_ACTIVE if port.startswith('Ethernet-BP')%}
```
neethajohn, line 53: There are other multi-asic platforms with port names starting with 'Ethernet-BP' and this change could affect those

zhixzhu: Added a flag backplane_port_no_ecn_marking for it.

PR #14547, opened 2023-04-06
[Mellanox] make sure shared storage with syncd is cleared on restarts

stepanblyschak:
#### Why I did it

Sharing the storage of syncd with other proprietary application extensions allows them to communicate with syncd in differnt ways.
If one container wants to pass some information to syncd then shared storage can be used. However, today the shared storage isn't cleaned on restarts making it possible for syncd to read out-of-date information generated in the past.

NOTE: No plans to use it for standard SONIC dockers and we are working on removing the SDK dependency from PMON docker

#### How I did it

Implemented new service to clean the shared storage.

#### How to verify it

Do reboot/fast-reboot/warm-reboot/config-reload/systemctl restart swss and verify /tmp/ is cleaned after each restart in syncd container.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211

### DIFF
--- a/files/build_templates/docker_image_ctl.j2
+++ b/files/build_templates/docker_image_ctl.j2
@@ -157,9 +157,9 @@ function waitForAllInstanceDatabaseConfigJsonFilesReady()
             done
         fi
         # Delay a second to allow all instance database_config.json files to be completely generated and fully accessible.
-        # This delay is needed to make sure that the database_config.json files are correctly rendered from j2 template 
+        # This delay is needed to make sure that the database_config.json files are correctly rendered from j2 template
         # files ( renderning takes some time )
-        sleep 1 
+        sleep 1
     fi
 }
 {%- endif %}
@@ -240,7 +240,7 @@ function postStartAction()
             mv $WARM_DIR/dump.rdb $WARM_DIR/dump.rdb.old
         else
             # If there is a config_db.json dump file, load it.
-            if [ -r /etc/sonic/config_db$DEV.json ]; then		
+            if [ -r /etc/sonic/config_db$DEV.json ]; then
 
                 if [ -r /etc/sonic/init_cfg.json ]; then
                     $SONIC_CFGGEN -j /etc/sonic/init_cfg.json -j /etc/sonic/config_db$DEV.json --write-to-db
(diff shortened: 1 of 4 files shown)


liat-grozovik (approved):

liat-grozovik:
@stepanblyschak does it have any impact of upgrades between different version on the same branch and between branches?

liat-grozovik:
@lguohan would you like to review or assign someone to review? LGTM

lguohan:
frankly, i do not like the idea of having additional channel for communication between dockers. there are lots of challenges of dependency between dockers and very hard to debug. I would really like to understand why this is needed, what is the data shared between syncd docker and other dockers.

lguohan:
@vaibhavhd  and @yxieca  for more comments.

stepanblyschak:
@lguohan This is not new communication channel. It exists for quite some time and used by pmon, what-just-happened to communicate with SDK directly as not all features are now supported by SAI. This PR is not about adding new communication channel but making the shared location is cleared on restarts.

liat-grozovik:
This is not new
This is available for years a
It is just rearrangement

pmon is already sharing it and this is how it can work on Nvidia platform
pmon and syncd are already aligned in systemd

stepanblyschak:
@yxieca @liat-grozovik @lguohan Clarified the purpose in PR description.

yxieca:
> @yxieca This is guernteed by systemd. What is the risk?

I am not sure that is fully guaranteed. e.g. if syncd is stopping while pmon is accessing the shared memory. Then the syncd won't be able to remove the shared memory folder. As result, it could see stale information after next start. Or the service stop/start could fail?

stepanblyschak:
Pmon is instructed to start After=syncd - https://github.com/sonic-net/sonic-buildimage/blob/fe773e2ed613a7b44e266df120c6e6ec0bd3c2ce/files/build_templates/pmon.service.j2#L6.

That means, on shutdown the reverse order is used. - 
```
When two units with an ordering dependency between them are shut down, the inverse of the start-up order is applied. I.e. if a unit is configured with After= on another unit, the former is stopped before the latter if both are shut down
```

If syncd service is about to stop (no matter if gracefully by request or abnormally) systemd first stops pmon service.

liat-grozovik:
@lguohan could you please help to review/approve?

lguohan:
@yxieca , i see this note. 

NOTE: No plans to use it for standard SONIC dockers and we are working on removing the SDK dependency from PMON docker

looks like it is only used between mellanox syncd docker and pmon docker, and they are working on removing that dependency, so that will be removed in the future. is this ok?

lguohan (approved):

yxieca:
> @yxieca , i see this note.
> 
> NOTE: No plans to use it for standard SONIC dockers and we are working on removing the SDK dependency from PMON docker
> 
> looks like it is only used between mellanox syncd docker and pmon docker, and they are working on removing that dependency, so that will be removed in the future. is this ok?

I believe @stepanblyschak is checking warm reboot scenario, we can merge on his signal.

stepanblyschak:
@yxieca We are good to merge it.

PR #14549, opened 2023-04-06
[dns] Add support for static DNS configuration.

oleksandrivantsiv:
#### Why I did it
Add support for static DNS configuration. According to https://github.com/sonic-net/SONiC/pull/1262 HLD.

#### How I did it
Add a new `resolv-config.service` that is responsible for transferring configuration from Config DB into `/etc/resolv.conf` file that is consumed by various subsystems in Linux to resolve domain names into IP addresses.

#### How to verify it
- Run the image compilation. Each component related to the static DNS feature is covered with the unit tests.
- Run sonic-mgmt tests. Static DNS feature will be covered with the system tests.
- Install the image and run manual tests.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211


oleksandrivantsiv:
This is the re-creation of https://github.com/sonic-net/sonic-buildimage/pull/14239, which was closed by mistake

ganglyu (dismissed):

files/image_config/resolv-config/update-containers
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+for container in $(docker ps -a --format=" {{ .ID }}"); do
```
ganglyu, line 3: Some containers like telemetry will start after 3 minutes.
If we upgrade SONiC image, and start the system for the first time, can we update /etc/resolv.conf in telemetry container?

oleksandrivantsiv: `docker ps -a` command returns all existing containers on the system. And we will update `/etc/resolv.conf` file for all containers including those that are not running.

For the case when the container is created in the runtime (for example app extension), when the dockerd starts the container it copies `/etc/resolv.conf` file from the host OS into the container filesystem. So, we will always have a valid DNS configuration inside the containers.

liat-grozovik:
@ganglyu would you please refer to comments and help to approve?

ganglyu (approved):

Yakiv-Huryk (approved):

liat-grozovik (approved):

StormLiangMS:
@oleksandrivantsiv @ganglyu For me, this looks like a feature other than bug fix, if so, not to 202305.

liat-grozovik:
The feature was discussed for 202305 and hld was available for long time including PRs

The history is that it was reported as bug
The solution was to add config and handle it as a feature

It is nice to have in 202305, not must.

oleksandrivantsiv:
@StormLiangMS can we finalize this? Should we backport the changes to 202305?

Thanks,
Oleksandr

PR #14601, opened 2023-04-11
Support to add SONiC OS Version in device info

xumia:
#### Why I did it
Support to add SONiC OS Version in device info.
It will be used to display the version info in the SONiC command "show version".

```
SONiC Software Version: SONiC.master-13812.218661-7d94c0c28
SONiC OS Version: 11
Distribution: Debian 11.6
Kernel: 5.10.0-18-2-amd64
```

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211


qiluo-msft (commented):
LGTM

lguohan (approved):

rules/config
```diff
@@ -294,3 +294,6 @@ SONIC_SLAVE_DOCKER_DRIVER ?= vfs
 # GZ_COMPRESS_PROGRAM - select pigz (a parallel implementation of gzip) to reduce a build time
 # and speed up a decompression of docker images on target system
 GZ_COMPRESS_PROGRAM ?= gzip
+
+# SONIC_OS_VERSION - sonic os version
+SONIC_OS_VERSION ?= 11
```
saiarcot895, line 299: Does this 11 correspond to the Debian version? If so, wouldn't it be better to read from `/etc/lsb_release.yml` and parse the version field there, rather than having it hardcoded here?

PR #14630, opened 2023-04-12
[sflow] Add egress sflow support.

rajkumar38:
- Updated sonic-yang-model
- change hsflowd version to 2.0.45

#### Why I did it
New enhancement added to current sflow to support egress sampling at port level.
https://github.com/sonic-net/SONiC/pull/1268
##### Work item tracking

#### How I did it
Added new command, yang and swss changes.
https://github.com/sonic-net/sonic-utilities/pull/2790
https://github.com/sonic-net/sonic-swss/pull/2731
#### How to verify it
Compile and loaded the sonic bin with all changes.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211

- [ ] 
- [ ] 

Egress Sflow support.

https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-yang-models/doc/Configuration.md#sflow


rules/sflow.mk
```diff
@@ -1,6 +1,6 @@
 # host-sflow package
 
-HSFLOWD_VERSION = 2.0.35
```
Gokulnath-Raja, line 3: Any reason for not using latest version 2.0.51 ? or stable previous versions such as 2.0.49?

src/sonic-yang-models/yang-models/sonic-sflow.yang
```diff
@@ -32,10 +32,28 @@ module sonic-sflow{
 
     description "SFLOW yang Module for SONiC OS";
 
+    revision 2023-04-11 {
```
Gokulnath-Raja, line 35: In https://github.com/openconfig/public/blob/master/release/models/sampling/openconfig-sampling-sflow.yang egress-sampling-rate supported at global level and per interface level ? How do we map sonic-sflow.yang to https://github.com/openconfig/public/blob/master/release/models/sampling/openconfig-sampling-sflow.yang ? or we don't support egress-sampling-rate configuration at all in SONIC?

src/sonic-yang-models/yang-models/sonic-sflow.yang
```diff
@@ -32,10 +32,28 @@ module sonic-sflow{
...
+        description "Add direction command to support egress sflow";
+    }
+
     revision 2021-04-26 {
         description "First Revision";
     }
 
+    typedef sample_direction {
```
Gokulnath-Raja, line 43: As per https://github.com/openconfig/public/blob/master/release/models/sampling/openconfig-sampling-sflow.yang it assumes that the interface will perform both ingress and egress sampling without the need of direction leaf? If that is the case when we support latest https://github.com/openconfig/public/blob/master/release/models/sampling/openconfig-sampling-sflow.yang in community sonic we should always populate direction as "both" in sonic-flow.yang?

rules/sflow.mk
rajkumar38: No specific reason. Egress sample support was added in 2.0.45. Hence updated to this version.

src/sonic-yang-models/yang-models/sonic-sflow.yang
rajkumar38: As mentioned in above comment, as per sflow spec, an interface must have same sample-rate on ingress and egress. Since initial support in SONiC was handling only Ingress, default value is set to "rx" for backward compatibility. Pls check HLD for more info.

vadymhlushko-mlnx (approved):

prsunny:
@Gokulnath-Raja , please signoff

Gokulnath-Raja (approved):

rajkumar38:
@qiluo-msft @xumia @lguohan 
Can you pls review/approve this PR .

prsunny (approved):

PR #14632, opened 2023-04-13
Install python-redis package to docker containers

liuh-80:
Install python-redis package to docker containers

#### Why I did it
This this bug: https://github.com/sonic-net/sonic-buildimage/issues/14531
The 'flush_unused_database' is part of docker-database, and docker-database does not install python-redis package by itself. it's using redis installed by sonic-py-swsssdk.
So after remove sonic-py-swsssdk from container, this script break.

To this this bug and avoid similer bug happen again, install python-redis to docker containers which removed sonic-py-swsssdk .

#### How I did it
Install python-redis to containers.

#### How to verify it
Pass all UT.
Create new UT to cover this scenario: https://github.com/sonic-net/sonic-mgmt/pull/8032

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012
- [x] 202106
- [x] 202111
- [x] 202205
- [x] 202211

Improve sudo cat command for RO user.

### DIFF
--- a/dockers/docker-config-engine-bullseye/Dockerfile.j2
+++ b/dockers/docker-config-engine-bullseye/Dockerfile.j2
@@ -23,6 +23,9 @@ RUN apt-get install -y        \
 RUN pip3 install pyangbind==0.8.1
 RUN pip3 uninstall -y enum34
 
+# Install python-redis
+RUN pip3 install redis==4.5.4
+
 {% if docker_config_engine_bullseye_debs.strip() %}
 # Copy locally-built Debian package dependencies
 {{ copy_files("debs/", docker_config_engine_bullseye_debs.split(' '), "/debs/") }}
--- a/dockers/docker-platform-monitor/Dockerfile.j2
+++ b/dockers/docker-platform-monitor/Dockerfile.j2
@@ -47,9 +47,6 @@ RUN pip3 install requests
 # We install the libpci module in order to be able to do PCI transactions
 RUN pip3 install libpci
 
-# Install python-redis
-RUN pip3 install redis==4.5.4
-
 {% if docker_platform_monitor_debs.strip() -%}
 # Copy locally-built Debian package dependencies
 {{ copy_files("debs/", docker_platform_monitor_debs.split(' '), "/debs/") }}
--- a/dockers/docker-platform-monitor/Dockerfile.j2
+++ b/dockers/docker-platform-monitor/Dockerfile.j2
@@ -47,6 +47,9 @@ RUN pip3 install requests
 # We install the libpci module in order to be able to do PCI transactions
 RUN pip3 install libpci
 
+# Install python-redis
+RUN pip3 install redis==4.5.4
+
 {% if docker_platform_monitor_debs.strip() -%}
 # Copy locally-built Debian package dependencies
 {{ copy_files("debs/", docker_platform_monitor_debs.split(' '), "/debs/") }}
--- a/dockers/docker-snmp/Dockerfile.j2
+++ b/dockers/docker-snmp/Dockerfile.j2
@@ -45,6 +45,9 @@ RUN pip3 install --no-cache-dir \
         pyyaml                              \
         smbus
 
+# Install python-redis
+RUN pip3 install redis==4.5.4
+
 {% if docker_snmp_whls.strip() -%}
 # Copy locally-built Python wheel dependencies
 {{ copy_files("python-wheels/", docker_snmp_whls.split(' '), "/python-wheels/") }}
--- a/dockers/docker-snmp/Dockerfile.j2
+++ b/dockers/docker-snmp/Dockerfile.j2
@@ -45,9 +45,6 @@ RUN pip3 install --no-cache-dir \
         pyyaml                              \
         smbus
 
-# Install python-redis
-RUN pip3 install redis==4.5.4
-
 {% if docker_snmp_whls.strip() -%}
 # Copy locally-built Python wheel dependencies
 {{ copy_files("python-wheels/", docker_snmp_whls.split(' '), "/python-wheels/") }}
--- a/dockers/docker-config-engine-buster/Dockerfile.j2
+++ b/dockers/docker-config-engine-buster/Dockerfile.j2
@@ -23,6 +23,9 @@ RUN apt-get install -y        \
 RUN pip3 install pyangbind==0.8.1
 RUN pip3 uninstall -y enum34
 
+# Install python-redis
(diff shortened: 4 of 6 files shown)


qiluo-msft:
Is it true that only flush_unused_database inside docker-database uses redis python package? If yes, we can change only docker-database Dockerfile.

Is it possible to replace this dependency by libswsscommon? If yes, create a new issue to track.

liuh-80:
> Is it true that only flush_unused_database inside docker-database uses redis python package? If yes, we can change only docker-database Dockerfile.
> 
> Is it possible to replace this dependency by libswsscommon? If yes, create a new issue to track.
The swsssdk been removed from all these containser by this PR: https://github.com/sonic-net/sonic-buildimage/pull/11469/files

To avoid potensial risk that other script may break because python-redis missing, maybe we install python-redis to all containers?

Also it's possible replace the dependency by libswsscommon, new issue:  https://github.com/sonic-net/sonic-buildimage/issues/14634

dockers/docker-snmp/Dockerfile.j2
```diff
@@ -45,6 +45,9 @@ RUN pip3 install --no-cache-dir \
         pyyaml                              \
         smbus
 
+# Install python-redis
+RUN pip3 install redis==4.5.4
```
nazariig, line 49: @liuh-80 why do we need this?

dockers/docker-platform-monitor/Dockerfile.j2
```diff
@@ -47,6 +47,9 @@ RUN pip3 install requests
 # We install the libpci module in order to be able to do PCI transactions
 RUN pip3 install libpci
 
+# Install python-redis
+RUN pip3 install redis==4.5.4
```
nazariig, line 51: @liuh-80 why do we need this?

xumia:
Hua, do we need to install all docker images, or only swss layer enough? Such as docker-swss-layer-bullseye, docker-swss-layer-buster.

dockers/docker-snmp/Dockerfile.j2
liuh-80: The swssdk depends on python-redis.
So before we remove swssdk by this PR, python-redis also install to docker-snmp when install swsssdk: https://github.com/sonic-net/sonic-buildimage/pull/11469/files

So to avoid potensial risk that some other script also using python-redis break because swsssdk removed from this docker, maybe we install python-redis to this container.

dockers/docker-platform-monitor/Dockerfile.j2
liuh-80: same reason as docker-snmp

liuh-80:
> Hua, do we need to install redis to all docker config engine images, or only swss layer enough? Such as docker-swss-layer-bullseye, docker-swss-layer-buster.

Seems those docker layer is based on docker config engine images?

https://github.com/sonic-net/sonic-buildimage/blob/d57de0987a26553e94b8c31caed385e8b156c70c/dockers/docker-swss-layer-buster/Dockerfile.j2
FROM docker-config-engine-buster-{{DOCKER_USERNAME}}:{{DOCKER_USERTAG}}

The original swsssdk and python-redis been installed to docker config engine images, so there are potensial gap if we only install python-redis to these layer.

nazariig:
> Seems those docker layer is based on docker config engine images?

@liuh-80 that is exactly my question: why do we need to have the same package in database docker when it is already installed in config engine? What is the motivation to have it in both places?

liuh-80:
> > Seems those docker layer is based on docker config engine images?
> 
> @liuh-80 that is exactly my question: why do we need to have the same package in database docker when it is already installed in config engine? What is the motivation to have it in both places?

@nazariig, I check these docker layer file, currently the python-redis does not installed in it.

Also the docker-database is based on config engine, not using swss layer.

dockers/docker-platform-monitor/Dockerfile.j2
qiluo-msft, line 51: This is already in base image docker-config-engine-bullseye*

dockers/docker-snmp/Dockerfile.j2
qiluo-msft, line 49: This is already in base image docker-config-engine-bullseye*

dockers/docker-config-engine-bullseye/Dockerfile.j2
```diff
@@ -23,6 +23,9 @@ RUN apt-get install -y        \
 RUN pip3 install pyangbind==0.8.1
 RUN pip3 uninstall -y enum34
 
+# Install python-redis
+RUN pip3 install redis==4.5.4
```
qiluo-msft, line 27: >==4.5.4 [](http://example.com/codeflow?start=22&length=7)

@xumia, do we need to specify the version?

xumia: > > ==4.5.4
> 
> @xumia, do we need to specify the version?
If python3-redis installed (redis=3.5.3 currently), then need to specify the version, or use pip3 install --upgrade redis. If the version not specified, then the installation will be skipped, use 3.5.3.

If we need the version 4.5.4 exactly, then "pip3 install redis==4.5.4" is good.
If we want to use the latest version, and at least 4.5.4, use "pip3 install redis>=4.5.4", prefer to auto version upgrade.

liuh-80: Hi, we need exactly 4.5.4, because swsssdk install 4.5.4:

https://github.com/sonic-net/sonic-py-swsssdk/blob/master/setup.py
dependencies = [
    'redis==4.5.4;python_version >= "3.0"',
    'redis>=3.5.3;python_version < "3.0"',
    'redis-dump-load',
]

dockers/docker-platform-monitor/Dockerfile.j2
liuh-80: Fixed, code removed.

liuh-80: Fixed, code removed.

dockers/docker-snmp/Dockerfile.j2
liuh-80: Fixed, code removed.

liuh-80: Fixed, code removed.

xumia (approved):

qiluo-msft (approved):

PR #14792, opened 2023-04-21
Adding comments for vlan import

Ndancejic:
#### Why I did it
vlan import for yang currently doesn't work, so was removed from the original PR: https://github.com/sonic-net/sonic-buildimage/pull/14521

Adds comments to remember to add later

### DIFF
--- a/src/sonic-yang-models/yang-models/sonic-neigh.yang
+++ b/src/sonic-yang-models/yang-models/sonic-neigh.yang
@@ -11,6 +11,11 @@ module sonic-neigh {
 		prefix yang;
 	}
 
+	// TODO: Uncomment the following lines when sonic-vlan.yang is available
+	// import sonic-vlan {
+	// 	prefix svlan;
+	// }
+
 	organization "SONiC";
 	contact "SONiC";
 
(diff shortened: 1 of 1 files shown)


ganglyu (approved):

src/sonic-yang-models/yang-models/sonic-neigh.yang
```diff
@@ -11,6 +11,11 @@ module sonic-neigh {
 		prefix yang;
 	}
 
+	// TODO: Uncomment the following lines when sonic-vlan.yang is available
+	// import sonic-vlan {
```
prsunny, line 15: how did it work for original PR? I don't think we should have dead code. Yang sub-group to check.

ganglyu: https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-yang-models/yang-models/sonic-vxlan.yang#L87-L90
We have other comments for back-links issue.

Ndancejic: I removed the import in the original PR, I'm just adding comments with import and ref so we can have the TODO for when we can enable the import. ask: https://github.com/sonic-net/sonic-buildimage/pull/14521#discussion_r1173194179

prsunny (approved):

PR #14804, opened 2023-04-22
Fixed the internal peer route-map policy

abdosi:
What I did:
In FRR command `update source <interface-name>` is not at address-family level. Because of this 
internal peer route-map for ipv6 were getting applied to ipv4 address family. As a result
TSA over iBGP for Ipv6 was not getting applied.

How I verify:

Manual Verification of TSA over both ipv4 and ipv6 after fix works fine.
Updated UT for this.

Added sonic-mgmt test gap: https://github.com/sonic-net/sonic-mgmt/issues/8170

### DIFF
--- a/src/sonic-bgpcfgd/tests/data/internal/peer-group.conf/result_chasiss_packet.conf
+++ b/src/sonic-bgpcfgd/tests/data/internal/peer-group.conf/result_chasiss_packet.conf
@@ -3,15 +3,15 @@
 !
   neighbor INTERNAL_PEER_V4 peer-group
   neighbor INTERNAL_PEER_V6 peer-group
-  address-family ipv4
     neighbor INTERNAL_PEER_V4 update-source Loopback4096 
+  address-family ipv4
     neighbor INTERNAL_PEER_V4 soft-reconfiguration inbound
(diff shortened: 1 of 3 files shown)


tjchadaga (dismissed):

abdosi:
@rlhui can you please help merge with this.

src/sonic-bgpcfgd/tests/data/internal/peer-group.conf/result_chasiss_packet.conf
```diff
@@ -3,15 +3,15 @@
 !
   neighbor INTERNAL_PEER_V4 peer-group
   neighbor INTERNAL_PEER_V6 peer-group
-  address-family ipv4
     neighbor INTERNAL_PEER_V4 update-source Loopback4096
```
lguohan, line 6: should we remove this indentation if neighbor update-source is not in address-family group.

abdosi: updated.

tjchadaga (approved):

abdosi:
@rlhui / @lguohan please help merge this.

lguohan (approved):

PR #14816, opened 2023-04-24
[202205] Added changes in caclmgrd for chassis

abdosi:
What/Why I did:

1. Allow traffic with source and destination as chassis eth1-midplane ip. Needed for Supervisor Redis-db connection (Redis packet has source and destination ip as eth1-midpane) after we load acl.json that has catch-all drop rule. Changes are generic and not specific to supervisor and applies on LC also.

2. Made multi_asic_ns_to_host_fwd as False for ACL service for External Client. This flag is needed for service SSH and SNMP where traffic can come in namespace over front-panel ports and we need to send the traffic in host where corresponding docker/service are running. There is no use-case of External client service for multi-asic as of now. Having flag as True creates failure when we try to load acl.json.

How I verify:
Manual Verification
UT updated accordingly.

### DIFF
--- a/src/sonic-host-services/scripts/caclmgrd
+++ b/src/sonic-host-services/scripts/caclmgrd
@@ -102,7 +102,7 @@ class ControlPlaneAclManager(daemon_base.DaemonBase):
         },
         "EXTERNAL_CLIENT": {
             "ip_protocols": ["tcp"],
-            "multi_asic_ns_to_host_fwd":True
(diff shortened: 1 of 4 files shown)


judyjoseph (approved):

rlhui:
Thanks. Please add comments why this PR is on 202205 only.  Is there another PR for master and how about 202211 branch?

PR #14843, opened 2023-04-25
[healthd] Use unix_socket_path instead of loopback ip

vivekrnv:
#### Why I did it

interfaces-config service restarts networking service, which in-turn results in loopback interface address is being removed and reassigned back

If the system-health happens to start during that instance expections and logs like this are seen:

```
Apr 15 18:14:49.357869 r-panther-20 ERR healthd: update system status exception:Unable to connect to redis: Cannot assign requested address
Apr 15 18:14:49.429778 r-panther-20 ERR healthd: subscribe_statedb exited- Unable to connect to redis: Cannot assign requested address
Apr 15 18:14:52.218594 r-panther-20 ERR healthd: system_service_Map_base::at
Apr 15 18:14:52.219714 r-panther-20 ERR healthd: system_service_Map_base::at
Apr 15 18:14:55.218636 r-panther-20 ERR healthd: system_service_Map_base::at
Apr 15 18:14:55.218722 r-panther-20 ERR healthd: system_service_Map_base::at
```

#### How I did it

use unix socket path

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [x] 202211

- [ ] 
- [ ]


liat-grozovik:
@qiluo-msft could you please help to review?

moshemos:
@qiluo-msft can you please help merge it, we need this fix for our 202211 release

liat-grozovik:
@Junchao-Mellanox could you please help to review the concept?

Junchao-Mellanox:
Is system-health the only service that will connect redis during loopback interface flapping?

dgsudharsan (dismissed):

vivekrnv:
> Try to understand the issue. Is system-health the only service that will connect redis during loopback interface flapping?
> 
> Maybe we should fix this line https://github.com/sonic-net/sonic-buildimage/blob/a738c39328b3e069a83162fe036ae693a8bd3d27/src/system-health/health_checker/sysmonitor.py#LL34C31-L34C42 . Use unix socket instead of tcp socket.

There were a few instances of this one reported. Using unix socket apparently causes some permission issues https://github.com/sonic-net/sonic-buildimage/pull/10179#discussion_r840977124

@qiluo-msft can clarify on that.

Junchao-Mellanox:
system health is a service managed by systemd, it is not a CLI. So I assume the permission issue is not relevant.

vivekrnv:
> system health is a service managed by systemd, it is not a CLI. So I assume the permission issue is not relevant.

Hmm, another instance of where this approach is used. https://github.com/sonic-net/sonic-buildimage/commit/c93716a142f47cc5cb684a5a76ace7470b5644a5

Junchao-Mellanox:
rsyslog service is a different case. It uses UDP which depends on LO interface. System health does not rely on UDP/TCP. So, I don't see an issue to use unix socket here.

vivekrnv:
> rsyslog service is a different case. It uses UDP which depends on LO interface. System health does not rely on UDP/TCP. So, I don't see an issue to use unix socket here.

So you suggest moving all DBconnector initialization in system-health to Unix socket?

dgsudharsan (approved):

Junchao-Mellanox (approved):

vivekrnv:
@Junchao-Mellanox, updated with your recommendation. Can you review?

src/system-health/tests/mock_connector.py
```diff
@@ -2,7 +2,7 @@ class MockConnector(object):
     STATE_DB = None
     data = {}
 
-    def __init__(self, host):
+    def __init__(self, use_unix_socket_path):
```
qiluo-msft, line 5: >use_unix_socket_path [](http://example.com/codeflow?start=23&length=20)

Does it need a default value?

qiluo-msft:
@judyjoseph to review for multi-asic.

src/system-health/tests/mock_connector.py
vivekrnv: No, it's just a mock. the variable has no significance when running tests

judyjoseph:
> @judyjoseph to review for multi-asic.

@qiluo-msft Since this service runs in host, it is same behavior for multi-asic device too.

One question I have is which user is system-health service running on ? If it is not "root" we will have permissions issues accessing the unix socket path ? 
```
cat ./system-health.service
[Unit]
Description=SONiC system health monitor
Requires=database.service updategraph.service
After=database.service updategraph.service

[Service]
ExecStart=/usr/local/bin/healthd
Restart=always

[Install]
WantedBy=multi-user.target
```

vivekrnv:
> > @judyjoseph to review for multi-asic.
> 
> @qiluo-msft Since this service runs in host, it is same behavior for multi-asic device too.
> 
> One question I have is which user is system-health service running on ? If it is not "root" we will have permissions issues accessing the unix socket path ?
> 
> ```
> cat ./system-health.service
> [Unit]
> Description=SONiC system health monitor
> Requires=database.service updategraph.service
> After=database.service updategraph.service
> 
> [Service]
> ExecStart=/usr/local/bin/healthd
> Restart=always
> 
> [Install]
> WantedBy=multi-user.target
> ```

This comment might help answer your question. https://github.com/sonic-net/sonic-buildimage/pull/14843#issuecomment-1539743553

vivekrnv:
@judyjoseph any other comments?

judyjoseph (approved):
LGTM. And please make sure it is validated in a JIT based environment -- so that there is no permission issues.

vivekrnv:
> LGTM. And please make sure it is validated in a JIT based environment -- so that there is no permission issues.

Can you expand on what JIT means?

src/system-health/tests/mock_connector.py
qiluo-msft: I see the caller code is `SonicV2Connector(use_unix_socket_path=True)`.

So ideally you need to mock for a optional parameter. Even the current works.

keboliu:
Hi @qiluo-msft would you please merge this since it's approved?

keboliu:
@StormLiangMS  would you please help to cherry-pick?

src/system-health/tests/mock_connector.py
qiluo-msft: Still applicable, even this PR is merged.
@liat-grozovik @vivekrnv

vivekrnv: I've ran the tests with the previous mock, and it failed. That's why i've updated the mock. 

Previously, we used to call `SonicV2Connector(host="127.0.0.1")` with host argument, so the ctor for mock is also expected to have the host argument. Similarly for `use_unix_socket_path` arg

vivekrnv:
Issue was also seen on 202205 as reported here https://github.com/sonic-net/sonic-buildimage/issues/15364 and thus request to backport to 202205

PR #14908, opened 2023-05-01
Update PG headroom settings ports based on port speed/cable length

vmittal-msft:
#### Why I did it
Update cable length for uplink/downlink ports for chassis and and update PG/pool headroom size accordingly.

##### Work item tracking
17880812

#### How I did it
Updated cable length as well as buffer config in HWSKU files.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [x] 202205
- [x] 202211

- [ ] 
- [ ]


src/sonic-config-engine/tests/test_j2files.py
```diff
@@ -293,7 +294,7 @@ def test_l2switch_template_dualtor(self):
     def test_qos_arista7050_render_template(self):
         self._test_qos_render_template('arista', 'x86_64-arista_7050_qx32s', 'Arista-7050-QX-32S', 'sample-arista-7050-t0-minigraph.xml', 'qos-arista7050.json')
 
-    def do_test_qos_and_buffer_arista7800r3_48cq2_lc_render_template(self, platform, hwsku):
+    def do_test_qos_and_buffer_arista7800r3_48cq2_lc_render_template(self, platform, hwsku, minigraph, qos_sample_output, buffer_sample_output):
```
neethajohn, line 297: Why do we need this function? Can we not reuse the function "do_test_qos_and_buffer_lc_render_template" which is used by other chassis hwskus?

vmittal-msft: @neethajohn  Since other SKUs has multiple asics and these SKU have single asic, there is difference in location of config files in both cases. Hence using different functions.

neethajohn: The only diff is in L334. You can extend the function arguments to include a multi_asic flag with a default value. If that is true, add the path to asic0

vmittal-msft: Commited the feeback.

neethajohn (approved):

vmittal-msft:
Thanks. @yxieca @lguohan @rlhui , please help merge this change for chassis.

arlakshm (approved):

rlhui (approved):

PR #14989, opened 2023-05-09
[sonic-bgpcfgd] replace yaml.load() and exit()

maipbui:
#### Why I did it
It is not safe to call yaml.load with any data received from an untrusted source.
sys.exit is better than exit, considered good to use in production code.
Ref:
https://stackoverflow.com/questions/6501121/difference-between-exit-and-sys-exit-in-python
https://stackoverflow.com/questions/19747371/python-exit-commands-why-so-many-and-when-should-each-be-used
##### Work item tracking
- Microsoft ADO **(number only)**: 15022050

#### How I did it
Replace yaml.load() with yaml.safe_load()
Replace exit() by sys.exit()
#### How to verify it
pass UT
test in DUT

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211

- [ ]  internal.0-f06bc9dd7e
- [ ]

### DIFF
--- a/src/sonic-bgpcfgd/bgpcfgd/utils.py
+++ b/src/sonic-bgpcfgd/bgpcfgd/utils.py
@@ -26,7 +26,7 @@ def run_command(command, shell=False, hide_errors=False):
 def read_constants():
     """ Read file with constants values from /etc/sonic/constants.yml """
     with open('/etc/sonic/constants.yml') as fp:
-        content = yaml.load(fp) # FIXME: , Loader=yaml.FullLoader)
+        content = yaml.safe_load(fp) # FIXME: , Loader=yaml.FullLoader)
         if "constants" not in content:
             log_crit("/etc/sonic/constants.yml doesn't have 'constants' key")
             raise Exception("/etc/sonic/constants.yml doesn't have 'constants' key")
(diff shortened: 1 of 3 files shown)


src/sonic-bgpcfgd/bgpcfgd/utils.py
```diff
@@ -26,8 +26,8 @@ def run_command(command, shell=False, hide_errors=False):
 def read_constants():
     """ Read file with constants values from /etc/sonic/constants.yml """
     with open('/etc/sonic/constants.yml') as fp:
-        content = yaml.load(fp) # FIXME: , Loader=yaml.FullLoader)
```
qiluo-msft, line 29: >FIXME [](http://example.com/codeflow?start=34&length=5)

You can remove the FIXME since it is not applicable anymore.

src/sonic-bgpcfgd/tests/util.py
```diff
@@ -15,6 +15,6 @@ def load_constants_dir_mappings():
 
 def load_constants(constants = CONSTANTS_PATH):
     with open(constants) as f:
-        data = yaml.load(f) # FIXME" , Loader=yaml.FullLoader)
+        data = yaml.safe_load(f) # FIXME" , Loader=yaml.FullLoader)
```
qiluo-msft, line 18: >FIXME [](http://example.com/codeflow?start=35&length=5)

remove FIXME

qiluo-msft (approved):

PR #15046, opened 2023-05-12
Implement zram compression for docker in RAM

Staphylo:
#### Why I did it

Some devices running SONiC have a small storage device (2G and 4G mainly)
The SONiC image growth over time has made it impossible to install 2 images on a single device.
Some mitigations have been implemented in the past for some devices but there is a need to do more.

One such mitigation is `docker_inram` which creates a `tmpfs` and extracts `dockerfs.tar.gz` in it.
This all happens in the SONiC initramfs and by ensuring the installation process does not extract `dockerfs.tar.gz` on the flash but keep the file as is.

This mitigation does a tradeoff by using more RAM to reduce the disk footprint.
It however creates new issues for devices with 4G of system memory since the extracted `dockerfs.tar.gz` nears the 1.6G. 
Considering debian upgrades (with dual base images) and the continuous stream of features this is only going to get bigger.

This change introduces an alternative to the `tmpfs` by allowing a system to extract the `dockerfs.tar.gz` inside a `zram` device thus bringing compression in play at the detriment of performance.

#### How I did it

Introduce 2 new optional kernel parameters to be consumed by SONiC initramfs.
 - `docker_inram_size` which represent the max physical size of the `zram` or `tmpfs` volume (defaults to DOCKER_RAMFS_SIZE)
 - `docker_inram_algo` which is the method to use to extract the `dockerfs.tar.gz` (defaults to `tmpfs`)
    other values are considered to be compression algorithm for `zram` (e.g `zstd`, `zlo-rle`, `lz4`)

Refactored the logic to mount the docker fs in the SONiC initramfs under the `union-mount` script.
Moved the code into a function to make it cleaner and separated the inram volume creation and docker extraction.

On Arista platform with a flash smaller or equal to 4GB set `docker_inram_algo` to `zstd` which produces the best compression ratio at the detriment of a slower write performance and a similar read performance to other `zram` compression algorithms.

Also fixed an issue in the `docker_inram` code for Arista platforms.

Here are some benchmarking data for the various `zram` algorithms using https://gist.github.com/Staphylo/9699128cf8a622ef1585c69ed39c053a
```
Algo     Real Size  Comp Size Mem Size  Xtract Xtract+Sync Read  CRatio MRatio
lzo      1633783808 764522544 798195712 28.39  29.31       83.34 3.14   3.05
lzo-rle  1633787904 769651487 803340288 27.88  28.95       83.58 3.12   3.03
zstd     1634697216 573782415 602689536 26.28  66.27       91.68 3.85   3.71
lz4      1633783808 801108972 835792896 25.47  26.39       83.86 3.04   2.95
lz4hc    1634447360 738901758 773656576 39.49  71.14       82.53 3.21   3.11
```

#### How to verify it

When booting on an Arista system with less than 4GB of flash, the following 2 messages should appear on the console
```
Creating zram to extract dockerfs.tar.gz                                                                  
Extracting dockerfs.tar.gz
```
Then once the system is booted you can run `zramctl` and `df -h /var/lib/docker` to make sure the zram volume is used.
```
root@sonic:~# zramctl
NAME       ALGORITHM DISKSIZE  DATA  COMPR  TOTAL STREAMS MOUNTPOINT
/dev/zram0 zstd          1.9G  1.5G 496.9M 521.9M       4 /var/lib/docker

root@sonic:~# df -h /var/lib/docker
Filesystem      Size  Used Avail Use% Mounted on
/dev/zram0      1.9G  1.5G  347M  82% /var/lib/docker
```

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012
- [ ] 202106
- [ ] 202111
- [x] 202205
- [x] 202211

Tests pending on other branches.
 - [ ] 202012 will likely require an independent PR with more changes
 - [ ] 202205 will likely have a cherry-pick conflict
 - [ ] 202211 should apply properly

Add optional zram compression for docker in RAM

Since it's been a while, here's a compressed cat
Compressed cat

### DIFF
--- a/build_image.sh
+++ b/build_image.sh
@@ -202,12 +202,12 @@ elif [ "$IMAGE_TYPE" = "aboot" ]; then
     zip -g $OUTPUT_ABOOT_IMAGE .platforms_asic
 
     if [ "$ENABLE_FIPS" = "y" ]; then
-        echo "sonic_fips=1" > kernel-cmdline
+        echo "sonic_fips=1" >> kernel-cmdline-append
     else
-        echo "sonic_fips=0" > kernel-cmdline
+        echo "sonic_fips=0" >> kernel-cmdline-append
     fi
-    zip -g $OUTPUT_ABOOT_IMAGE kernel-cmdline
-    rm kernel-cmdline
+    zip -g $OUTPUT_ABOOT_IMAGE kernel-cmdline-append
+    rm kernel-cmdline-append
 
     zip -g $OUTPUT_ABOOT_IMAGE $ABOOT_BOOT_IMAGE
     rm $ABOOT_BOOT_IMAGE
--- a/files/initramfs-tools/union-mount.j2
+++ b/files/initramfs-tools/union-mount.j2
@@ -12,6 +12,8 @@ case $1 in
 esac
 
 docker_inram=false
+docker_inram_algo=tmpfs
+docker_inram_size={{ DOCKER_RAMFS_SIZE }}
 logs_inram=false
 secureboot=false
 bootloader=generic
@@ -26,6 +28,12 @@ for x in $(cat /proc/cmdline); do
             ;;
         docker_inram=on)
             docker_inram=true
+            ;;
+        docker_inram_algo=*)
+            docker_inram_algo="${x#docker_inram_algo=}"
+            ;;
+        docker_inram_size=*)
+            docker_inram_size="${x#docker_inram_size=}"
             ;;
         logs_inram=on)
             logs_inram=true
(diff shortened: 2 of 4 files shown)


files/initramfs-tools/union-mount.j2
```diff
@@ -95,6 +103,63 @@ remove_not_in_allowlist_files()
     rm -f $allowlist_pattern_file
 }
 
+mount_docker_inram()
+{
+    if [ "$docker_inram_algo" = "tmpfs" ]; then
```
StormLiangMS, line 108: @Staphylo could we split this PR to 2 PRs? One is for the docker_inram, the other is to turn on zram. With smaller PR, it would be easier to manage if any issue we want to revert or turn off.

Staphylo: This PR is made of 2 commits, would you want these 2 commits as individual PRs?
 - fix docker_inram behavior on Arista systems
 - refactor docker mount to cleanly add zram support

Or would you rather see 3 commits (much longer due to dependencies):
 - fix docker_inram behavior on Arista systems
 - refactor union-mount.j2 with the introduction of `mount_docker` and only `tmpfs` support like it is today
 - add support for the zram compression?

Whatever split you decide, please bear in mind that these changes will take much longer to merge as all of these changes generate conflict with each other. So I'll have to open the PRs one by one and wait until the previous one is merged. Cherry-picking to older release branches, most of these PRs will conflict again.
I will also need to spend much more time getting builds and running testing on individual changes rather than testing the aggregated content.
I'll do as you prefer but this will take much longer for everything to merge and be available in 202012 and 202205.

Please let me know

lguohan: thanks for working on this. current two commits are ok for me. however, for current commits, i couldn't find the commit messages. we can do a rebase merge which keep your original commits. can you check if your original commits has proper commit messages?

Staphylo: I updated the commits of this PR so they have the correct description.

build_image.sh
```diff
@@ -202,12 +202,12 @@ elif [ "$IMAGE_TYPE" = "aboot" ]; then
     zip -g $OUTPUT_ABOOT_IMAGE .platforms_asic
 
     if [ "$ENABLE_FIPS" = "y" ]; then
-        echo "sonic_fips=1" > kernel-cmdline
+        echo "sonic_fips=1" >> kernel-cmdline-append
```
qiluo-msft, line 205: Do you want to empty the file "kernel-cmdline-append" before the first usage?

Staphylo: It's indeed better if the file is reset before starting appending to it to avoid duplicated content.
I addressed your comment here https://github.com/sonic-net/sonic-buildimage/pull/15343
This PR will need to be cherry-picked to 202205 where this change also merged.

PR #2835, opened 2019-04-29
DellEMC S6100 Watchdog support

paavaanan:
**- What I did**
- Added two scripts platform_watchdog_enable.sh and platform_watchdog_disable.sh. 

**- How I did it**
  - Before upgrading the box please do execute platform_watchdog_enable.sh. This enable the SMF Watchdog bit. Also, configure the watchdog timer to 180 seconds.
- Upgrade the box with new image.
- During upgrade or in the middle of upgrade, if the box hangs in middle,  watchdog kicks-in at 180 seconds interval and perform a warm-reboot. 
- This will avoid manual power-cycle interruption at the time of issue state.
- Instead, if there is no any issues with the box we disable watchdog in startup script [s6100_platform.sh]
- Enabling watchdog is not a in-build one. The user have to manually trigger the script to enable watchdog.

**- How to verify it**
- execute platform_watchdog_enable.sh
- Upgrade the box with latest image.

PR #1296, opened 2018-01-11
[libsaithrift-dev]: Enable building libsaithrift-dev and pythonthrift libraries

pavel-shirshov:
**- What I did**
I restored building of libsaithrift and python-saithrift libraries, which is required for tests with saithrift.

**- How I did it**
Updated sairedis with updated SAI. Restore commented out code, fix versions

**- How to verify it**
```
set -e

# "broadcom"
# "cavium" - failed
# "centec"
# "marvell" - failed
# "mellanox"
# "nephos"

for vendor in "broadcom" "centec" "mellanox" "nephos"
do
    rm -fr sonic-buildimage/
    git clone https://github.com/pavel-shirshov/sonic-buildimage.git
    cd sonic-buildimage
    git checkout pavelsh/saithrift
    make init
    make configure PLATFORM=$vendor
    make target/debs/libsaithrift-dev_1.0.4_amd64.deb
    make target/debs/python-saithrift_1.0.4_amd64.deb
    cd ..
done
```
**- Description for the changelog**

Restore libsaithrift library building

**- A picture of a cute animal (not mandatory but encouraged)**

PR #1254, opened 2017-12-20
[DHCP relay]: Add patch to prevent dropping VLAN-tagged packets

jleveque:

PR #15058, opened 2023-05-13
[NTP] Add NTP extended configuration

fastiuk:
hld [#1296](https://github.com/sonic-net/SONiC/pull/1296)
closes [#1254](https://github.com/sonic-net/SONiC/issues/1254)
depends-on [#60](https://github.com/sonic-net/sonic-host-services/pull/60), [#781](https://github.com/sonic-net/sonic-swss-common/pull/781), [#2835](https://github.com/sonic-net/sonic-utilities/pull/2835), [#10749](https://github.com/sonic-net/sonic-mgmt/pull/10749)

#### Why I did it
To cover the next AIs:
* Configure NTP global parameters
* Add/remove new NTP servers
* Change the configuration for NTP servers
* Show NTP status
* Show NTP configuration

##### Work item tracking

#### How I did it
* Add YANG model for a new configuration
* Extend configuration templates to support new knobs

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [x] 202311

- [ ] 
- [ ] 

* Add ability to configure NTP global parameters such as authentication, dhcp, admin state
* Change the configuration for NTP servers
* Add an ability to show NTP configuration

[NTP configuration](https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-yang-models/doc/Configuration.md#ntp-and-syslog-servers)

#### A picture of a cute animal (it is my cat Finn)
PXL_20230413_140941610 PORTRAIT


files/image_config/ntp/ntp.conf.j2
```diff
@@ -90,32 +138,24 @@ interface listen eth0
...
+{# {% if global.server_role == 'disabled' and global.dhcp == 'disabled' -%}
+    {% set options = options ~ ' ignore' -%}
+{% endif -%} #}
 
+# Access control configuration
 # By default, exchange time with everybody, but don't allow configuration.
-restrict -4 default kod notrap nomodify nopeer noquery
-restrict -6 default kod notrap nomodify nopeer noquery
+restrict -4 default kod limited notrap nomodify noquery{{ options }}
+restrict -6 default kod limited notrap nomodify noquery{{ options }}
```
saiarcot895, line 157: Why was `nopeer` removed here? Is it so that in theory, other devices can use this device as a source for the current time?

saiarcot895: Ah, never mind, I see it manually gets added above.

files/image_config/ntp/ntp.conf.j2
```diff
@@ -1,45 +1,93 @@
...
 interface ignore wildcard
 
-# set global variable for configured source interface name
-# set global boolean to indicate if the ip of the configured source interface is configured
-# if the source interface is configured but no ip on that interface, then listen on another
-# interface based on existing logic
+{# Set interface to listen on set global variable for configured source
+interface name set global boolean to indicate if the ip of the configured source
+interface is configured if the source interface is configured but no ip on that
+interface, then listen on another interface based on existing logic. -#}
```
saiarcot895, line 90: Because of the rearrangement of the line breaks, this now reads like a long run-on sentence. Please add periouds/capitalization to make the separate sentences more clear.

files/image_config/ntp/ntp.conf.j2
```diff
@@ -90,32 +138,24 @@ interface listen eth0
...
-#
-# Note that "restrict" applies to both servers and clients, so a configuration
-# that might be intended to block requests from certain clients could also end
-# up blocking replies from your own upstream servers.
+{# Access control options -#}
+{% set options = '' -%}
+
+{# Allow additional servers mobilization from the pool. Otherwise we don't need
+that -#}
+{% if is_pools == False -%}
```
saiarcot895, line 146: Note that because of Jinja's variable scoping rules, `is_pools` will always be `False`, and `nopeer` will always get added. See https://jinja.palletsprojects.com/en/2.10.x/templates/#assignments for more info on this.

As such, I think we would want `nopeer` to be present always in the default `restrict` configuration?

src/sonic-yang-models/yang-models/sonic-ntp.yang
```diff
@@ -112,10 +173,96 @@ module sonic-ntp {
                 leaf server_address {
                     type inet:host;
                 }
+
+                leaf association_type {
+                    type association-type;
+                    default server;
+                    description "NTP remote association type. Server, pool, or
+                                 peer.";
```
saiarcot895, line 181: Is `peer` allowed? I don't see any examples/handling for this.

files/image_config/ntp/ntp-systemd-wrapper
```diff
@@ -27,21 +28,29 @@ fi
 
 (
         flock -w 180 9
+        ntpEnabled=$(/usr/local/bin/sonic-cfggen -d -v 'NTP["global"]["admin_state"]' 2> /dev/null)
+        if [ "$ntpEnabled" = "disabled" ]
```
saiarcot895, line 32: Can `ExecCondition=` in the systemd service file be used to check this condition instead? That way the service status will accurately reflect that a condition check failed.

There also shoudn't be another ntp daemon running at this point of time, so there shouldn't be anything that needs to be stopped.

src/sonic-yang-models/tests/yang_model_tests/tests_config/ntp.json
```diff
@@ -4,13 +4,38 @@
...
+                        "iburst": "on",
+                        "admin_state": "enabled"
+                    }
+                ]
+            },
+            "sonic-ntp:NTP_KEY": {
+                "NTP_KEYS_LIST": [
+                    {
+                        "id": 10,
+                        "value": "lumos"
```
bsun-sudo, line 34: SONiC Yang shows value is encrypted, does the test only accept plain text value?

src/sonic-yang-models/tests/yang_model_tests/tests_config/ntp.json
```diff
@@ -237,5 +262,234 @@
...
+        }
+    },
+    "NTP_KEY_VALID": {
+        "sonic-ntp:sonic-ntp": {
+            "sonic-ntp:NTP_KEY": {
+                "NTP_KEYS_LIST": [
+                    {
+                        "id": 20,
+                        "type": "md5",
+                        "value": "jsdf88320fsd2@@445",
```
bsun-sudo, line 429: same here, value is not encrypted.

src/sonic-config-engine/sonic-cfggen
```diff
@@ -137,6 +138,31 @@ def ip_network(value):
         return "Invalid ip address %s" % value
     return r_v.network
 
+
+def b64encode(value):
```
bsun-sudo, line 142: Where will b64encode get called? Could you please provide an example?

files/image_config/ntp/ntp.keys.j2
```diff
@@ -0,0 +1,18 @@
...
+{% for server in NTP_SERVER if NTP_SERVER[server].trusted == 'yes' and
+                               NTP_SERVER[server].resolve_as -%}
+    {% set _ = trusted_arr.append(NTP_SERVER[server].resolve_as) -%}
+{% endfor -%}
+
+{# Define authentication keys inventory -#}
+{% set trusted_str = ' ' ~ trusted_arr|join(',') -%}
+{% for keyid in NTP_KEY if NTP_KEY[keyid].type and NTP_KEY[keyid].value %}
+{% set keyval = NTP_KEY[keyid].value | b64decode %}
+{{ keyid }} {{ NTP_KEY[keyid].type }} {{ keyval }}{{trusted_str}}
```
bsun-sudo, line 17: Could you please also add an example of ntp.keys where trusted_str is not empty?

Meir-renford:
Update: we are reviewing the comments, and will address by next week

files/image_config/ntp/ntp.conf.j2
fastiuk: ok, will fix

files/image_config/ntp/ntp.conf.j2
fastiuk: No, it shoudn't be there always. I will check condition and change if needed

src/sonic-yang-models/yang-models/sonic-ntp.yang
fastiuk: not allwoed, will fix description

files/image_config/ntp/ntp-systemd-wrapper
fastiuk: hmm. nice idea, I guess it will work. I will try it and let you know the result here

files/image_config/ntp/ntp.conf.j2
```diff
@@ -54,8 +102,8 @@ interface ignore wildcard
 
 {% set ns = namespace(source_intf = "") %}
 {% set ns = namespace(source_intf_ip = 'false') %}
-{% if (NTP) and (NTP['global']['src_intf'])  %}
-    {% set ns.source_intf = (NTP['global']['src_intf']) %}
+{% if global.src_intf  %}
```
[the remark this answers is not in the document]
fastiuk: Nothing was changed in that flow from the original code, so if it was supported - it still supported now

src/sonic-yang-models/tests/yang_model_tests/tests_config/ntp.json
fastiuk: it is for tests only. I will change it to encrypted value to avoid confusions

src/sonic-yang-models/tests/yang_model_tests/tests_config/ntp.json
fastiuk: will change

src/sonic-config-engine/sonic-cfggen
fastiuk: see `files/image_config/ntp/ntp.keys.j2` file from that PR. it used there

files/image_config/ntp/ntp.keys.j2
fastiuk: Yes, I will attach it here.

files/image_config/ntp/ntp-systemd-wrapper
saiarcot895: Any update here on whether this worked?

fastiuk:
The PRs were updated, and all submodules changes were approved.
Please review this PR and approve

saiarcot895:
See [my comment](https://github.com/sonic-net/sonic-buildimage/pull/15058#discussion_r1329376843) above

files/image_config/ntp/ntp.keys.j2
fastiuk: Here is the example:
```
# This file was AUTOMATICALLY GENERATED. DO NOT MODIFY.
# Controlled by ntp-config.service

1 md5 password1 1.2.3.45,216.239.35.4
2 md5 password2 1.2.3.45,216.239.35.4
3 sha1 password3 1.2.3.45,216.239.35.4
```

Ip addresses here are trusted servers

files/image_config/ntp/ntp-systemd-wrapper
fastiuk: Hey. I tried it with another daemon and it actually works fine. I tried to rework it here, but it won't work here in a couple of reasons:
1. It is not a one-line solution and will basically make this implementation more complex.
2. We are using lock here, which will be much harder to use in the service file.
3. We need to use `ExecStopPost` condition to stop the daemon.

Other reasons I don't like this option:
1. We need to modify ntp service file which is controlled by open source community and not modified by us.
2. That `ntp-systemd-wrapper` script was created exactly to customize the control of ntp daemon.

So I would like to avoid of such changes

fastiuk:
Hey there. Addressed comments. Could you please review it?

saiarcot895 (approved):

fastiuk:
Hey there, I updated sonic-mgmt tests to support a new code, please approve and merge: https://github.com/sonic-net/sonic-mgmt/pull/10749

ganglyu:
would you please update src/sonic-yang-models/tests/files/sample_config_db.json?

fastiuk:
> would you please update src/sonic-yang-models/tests/files/sample_config_db.json?

updated

qiluo-msft (approved):

PR #15072, opened 2023-05-15
[Mellanox] Update SAI to 2211.24.0.21 and SDK/FW to 4.5.5142/2010_5144

keboliu:
#### Why I did it

SDK/FW Fixed Issues:

> •	When a system has more than 256 ACL entries, on rare occasion, removing/adding entries may cause some ACL entries not to work.
> •	When using mirror session policer on spectrum-2, spectrum-3, the actual CIR was 1.28 times more than the configured CIR value
> •	After warm boot process, when enabling ECN marking and the port is in split mode, traffic sent to the port under congestion (for example, when connecting two ports with a total speed of 50GbE to a single 25GbE port) is not marked.
> •	Warm boot might fail if the key value SAI_KEY_ACCUMULATED_FLOW_COUNTER_UNITS_IN_KB is set
> •	If counters are bound to an next hop group, there is a probability the next API calls that modify the next-hop group members will fail.
> •	In Spectrum platforms Fastboot mode is not operational for Split port with Force mode in 50G speed
> •	When fine grain next hop group has a size of 2K or 4K members, and group is removed, FW will remove only (size % 2048) members, resulting in leakage of KVD resources
> •	When reading some port statistics, or bulk reading some Queue or PG statistics, and in parallel reading or writing other counters, FW may, in rare cases, get stuck
> •	SN2201 Module 1 is considered to be present/linked while no cable/module is plugged
> •	On Spectrum-3 when port configure to 400G FW might stuck after running mlxlink while 400G interface connected  and swap between upper and lower 4 lanes
 
SAI New features:
> •	ACL: Added support for an ACL match on the AETH field (SAI_ACL_TABLE_ATTR_FIELD_AETH_SYNDROME, SAI_ACL_ENTRY_ATTR_FIELD_AETH_SYNDROME) to count RoCE NAK and CNP packets.
> •	PLL Status: Added a new logging entry that alerts the user upon a PLL lock loss event.
> •	Dual ToR - Additional MAC Address: Added support for setting a MAC address for the router interface which is not part of the 10 bit MAC address available for RIFs on Spectrum-1, as part of the Dual ToR scenario.
> •	Dual ToR: DSCP Remapping Added support for tunnel QoS maps as part of the Dual TOR scenario.
 
SAI Fixed issues:
> •	When setting a WRED profile attribute for a color that was not enabled during the profile create time, an error would be returned. After the fix, a default profile is create on such scenario and the set attribute is applied on top of it
> •	When calling the flush FDB by using the SAI_FDB_FLUSH_ATTR_BRIDGE_PORT_ID attribute, the bridge bv_id value was filled on the notification callback where it should have been left empty.

##### Work item tracking

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [x] 202211

- [ ] 
- [ ]


liat-grozovik (dismissed):

Junchao-Mellanox (approved):

liat-grozovik (approved):

StormLiangMS:
@liat-grozovik is this for SAI 1.12?

PR #15137, opened 2023-05-18
[202205] Implement zram compression for docker in RAM

Staphylo:
Cherry pick of https://github.com/sonic-net/sonic-buildimage/pull/15046
Will complete description once testing is complete.

MS ADO: 21450507

### DIFF
--- a/build_image.sh
+++ b/build_image.sh
@@ -202,12 +202,12 @@ elif [ "$IMAGE_TYPE" = "aboot" ]; then
     zip -g $OUTPUT_ABOOT_IMAGE .platforms_asic
 
     if [ "$ENABLE_FIPS" = "y" ]; then
-        echo "sonic_fips=1" > kernel-cmdline
+        echo "sonic_fips=1" >> kernel-cmdline-append
     else
-        echo "sonic_fips=0" > kernel-cmdline
+        echo "sonic_fips=0" >> kernel-cmdline-append
     fi
-    zip -g $OUTPUT_ABOOT_IMAGE kernel-cmdline
-    rm kernel-cmdline
+    zip -g $OUTPUT_ABOOT_IMAGE kernel-cmdline-append
+    rm kernel-cmdline-append
 
     zip -g $OUTPUT_ABOOT_IMAGE $ABOOT_BOOT_IMAGE
     rm $ABOOT_BOOT_IMAGE
--- a/files/Aboot/boot0.j2
+++ b/files/Aboot/boot0.j2
@@ -651,6 +651,7 @@ write_platform_specific_cmdline() {
            varlog_size=256
            cmdline_add logs_inram=on
            cmdline_add docker_inram=on
+           cmdline_add docker_inram_algo=zstd
            if [ $flash_size -le 2000 ]; then
                # enable docker_inram for switches with less than 2G of flash
                varlog_size=128
--- a/files/Aboot/boot0.j2
+++ b/files/Aboot/boot0.j2
@@ -402,7 +402,7 @@ extract_image() {
 extract_image_secureboot() {
     info "Extracting necessary swi content"
     # NOTE: boot/ is not used by the boot process but only extracted for kdump
-    unzip -oq "$swipath" 'boot/*' .imagehash -d "$image_path"
+    unzip -oq "$swipath" 'boot/*' .imagehash kernel-cmdline-append -d "$image_path"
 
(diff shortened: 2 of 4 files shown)


lipxu:
Hi, @Staphylo , Thanks for the PR. 
We picked these changes and built the 202205 image for running the nightly pipelines. Based on the pass rate, the zram solution is working well. Some cases failed due to no space. 
There are some cases that need to install extra docker images during the test, so is it possible to introduce a flag that allows for the control of the features, we can use it to disable the docker_inram and zram features and create another pipeline to run these special cases in our nightly test, thanks a lot.
Alternatively, do you have any other suggestions or more effective methods? thank you very much.

Staphylo:
Hi @lipxu,

Regarding the no space left I can see 2 issues at play here:

1) The ram filesystem is not big enough to host the containers, but I don't believe this is the case since with zram, the size provided to the zram device is the maximum physical size `1500M` should be plenty given the numbers I provided.
Internally we use 1900M since we have some cases with only `tmpfs` and not zram.
See https://github.com/sonic-net/sonic-buildimage/blob/dd8b9f2502aaaec96c658b544f7b0761fd95bc35/onie-image.conf#L31

2) On systems with small flashes and 4GB of RAM there is another issue at play.
The first mitigation we introduced with regard to space usage was to move the SWI image to a tmpfs before exctracting its contents.
See https://github.com/sonic-net/sonic-buildimage/blob/dd8b9f2502aaaec96c658b544f7b0761fd95bc35/files/Aboot/boot0.j2#L348
The reason for this change was to avoid the flash spike usage during install (equivalent to 2.5 installations) by doing a trade off with system memory.
In the current model this is still a problem since at the worst point in time you need to store mainly:
 - previous SONiC installation
 - `sonic.swi` the new image to install until everything is extracted
 - `fs.squashfs` the new rootfs
 - `dockerfs.tar.gz` the new docker images

In terms of additional options we could use to minimize RAM and Flash usage here are some of my additional thoughts.

I was investigating booting the SWI the same way we do for our secureboot feature.
In this mode we don't extract the content of the SWI image at installation time.
Instead, the installation is simply moving the `sonic.swi` into its folder under `/host/image-XXX/sonic.swi`
The `fs.squashfs` can then be mounted directly from within the SWI by creating a loop device using the start offset of the `fs.squashfs` in the `sonic.swi`
The `dockerfs.tar.gz` is using the `docker_inram` feature and is extracted directly from the swi by using a pipe (`unzip -qop sonic.swi | tar -xf -`)
The implementation details can be found here https://github.com/sonic-net/sonic-buildimage/pull/4741
This would remove any spike of RAM/flash usage during installation time since there is almost no extraction taking place.
However this would be mutually exclusive with the feature where we extract `dockerfs.tar.gz` inside a `zfs`/`btrfs` compressed fs blob that I am currently investigating.
This feature would however require some changes in `sonic-utilities/sonic-installer` and `sonic-utilities/fast-reboot`

In my opinion mitigating the spike usage of flash/memory during installation is what's making things hard since it is artificially bumping the requirement from being able to install 2 images to 3 images (with the 3rd either in flash or ram)

With regard to enabling/disabling the `docker_inram` feature this should be possible.
It could be etiher a new config in `boot-config` (e.g `DOCKER_INRAM=none`, `DOCKER_INRAM=zstd`) or a file on the flash.

lipxu:
Hi, @Staphylo , thanks for your details description. 
There are two types of new issues after enabling docker_inram for 7060 and 7050qx.

1. One is teardown failure due to memory threshold check failure. Will sync with you one by one later, thanks
2. Other failures are related to setup failure caused by no space. For some cases, we need to install some specific docker image (like QOS/Sai) or install certain packages(like COPP) that are necessary for the test. It's important to note that these Docker images and packages are only used for testing purposes and will be removed after the test. Previously, without docker_inram, they would be installed on the disk, however, with docker_inram enabled, they failed due to the 1.5GB size limitation. Therefore, we need a workaround for these tests. One easy way to address this is by adding a flag to disable docker_inram for these specific tests. thanks a lot

lipxu:
Verified on nightly pipelines, the solution works fine. only COPP and QOS/Sai related case failure due to not enough space.
Enlarged the zram size to 3000M by adding docker_inram_size=3000M , the actual ram took about 1.1G during QOS/Sai tests and all case passed.

lipxu (approved):

Issue #15114, opened 2023-05-17
Many useless trailing blanks in source code


guillaumelambert:
Trailing blanks are generally and strongly dis-advised by most programming languages best practices.
It is commonly admitted that they complicate the code maintenance.
Though, many can be found in the sonic-buidimage repo source code  (about 17400 lines).
Also no dedicated linter to address this issue can be found in the SONiC Azure CI.

liushilongbuaa:
`grep -Eo " *$" ./ -r --exclude-dir .git | awk -F: '{print$1}' | xargs -i sudo sed -i "s/ *$//" {}`
How about bash command to remove tail spaces?
I tried locally and got 17477 lines added, 17481 lines removed.
`git diff --numstat  | awk '{s+=$1} END {print s}'`
But removing all tail spaces is a big PR. It is hard to review.
If we add it to CI pipeline. We need to push commits to original PR.

guillaumelambert:
> `grep -Eo " *$" ./ -r --exclude-dir .git | awk -F: '{print$1}' | xargs -i sudo sed -i "s/ *$//" {}` How about bash command to remove tail spaces? I tried locally and got 17477 lines added, 17481 lines removed. `git diff --numstat | awk '{s+=$1} END {print s}'` But removing all tail spaces is a big PR. It is hard to review. If we add it to CI pipeline. We need to push commits to original PR. There maybe permission issue.

Thanks for your feedback. Such shell scripts commands do work.
Though, as stated in  the commit  https://github.com/Orange-OpenSource/sonic-buildimage/commit/1aebc1646349e9d5389af2063f789de903f4c73c in one of the related PRs,  trailing blanks removal of some .conf files in the src folder  made some tests fail during build phase. And  jinja templates in docker folder use a known hack that requires trailing blanks... So we will have to deal with some exceptions at least temporarily. This means incorporating more commands and potentially complex regex in such a shell script.
I proposed in https://github.com/sonic-net/sonic-buildimage/pull/15115 a different solution based on tox & pre-commit, which is IMHO more flexible and maintainable in a CI. It embeds a tox profile dedicated to the CI that allows to check only files modified since the introduction of this PR.

I do agree that trailing blanks complete trimming would be a big PR.
 cf https://github.com/sonic-net/sonic-buildimage/pull/15115#issuecomment-1555068914
I already proposed and split it in several pieces to ease its review and merge. Hope this helps. I noticed some have been already merged.

[...]

PR #15160, opened 2023-05-19
[CI][doc][build] Trim rules folder files trailing blanks

guillaumelambert:
- run pre-commit tox profile to trim all trailing blanks from https://github.com/sonic-net/sonic-buildimage/pull/15115
- use several commits with a per-folder based strategy to ease their merge

Issue #15114


rules/bash.mk_
```diff
@@ -0,0 +1,18 @@
+# bash package
```
liushilongbuaa, line 1: Added by accident?

rules/ifupdown2.mk_
```diff
@@ -0,0 +1,11 @@
+# ifupdown2 package
```
liushilongbuaa, line 1: Added by accident?

guillaumelambert: Ah yes. It comes from another branch work. Thanks for the feedback. I will rebase and do a git clean before.

PR #15176, opened 2023-05-21
Container net host remove

ycoheNvidia:
submodule change to support snmp container hardening: https://github.com/sonic-net/sonic-snmpagent/pull/281
#### Why I did it
Net host removal for snmp and database as part of container hardening

#### How to verify it
Modified snmp related templates to support net host removal:

* SNMP Docker listening-address:port forwarding
* snmpd.conf listening address to include all addresses - since docker port forwarding is the actual routing table - and blocks all other addresses.
* $NET modified to bridge for snmp and database container
* snmp docker stop to rm container - since port forwarding can only happen on newly created image

snmp and database host net removal

       .
      ":"
    ___:____     |"\/"|
  ,'        `.    \  /
  |  O        \___/  |
~^~^~^~^~^~^~^~^~^~^~^~^~


dockers/docker-database/Dockerfile.j2
```diff
@@ -32,6 +32,8 @@ RUN apt-get clean -y                                  && \
              s/^# unixsocket/unixsocket/;                \
              s/redis-server.sock/redis.sock/g;           \
              s/^client-output-buffer-limit pubsub [0-9]+mb [0-9]+mb [0-9]+/client-output-buffer-limit pubsub 0 0 0/; \
+             s/^bind 127.0.0.1 ::1$/# bind 127.0.0.1 ::1/;       \
```
qiluo-msft, line 35: ># bind 127.0.0.1 [](http://example.com/codeflow?start=36&length=16)

If you comment the bind line, it will listen to all interfaces, which is a security concern.

dockers/docker-database/Dockerfile.j2
```diff
@@ -32,6 +32,8 @@ RUN apt-get clean -y                                  && \
...
+             s/^protected-mode yes/protected-mode no/;    \
```
qiluo-msft, line 36: >no [](http://example.com/codeflow?start=50&length=2)

This is not safe, right?

dockers/docker-database/supervisord.conf.j2
```diff
@@ -36,7 +36,7 @@ dependent_startup=true
 {%- else -%}
 {%- set LOOPBACK_IP = '' -%}
 {%- endif -%}
-command=/bin/bash -c "{ [[ -s /var/lib/{{ redis_inst }}/dump.rdb ]] || rm -f /var/lib/{{ redis_inst }}/dump.rdb; } && mkdir -p /var/lib/{{ redis_inst }} && exec /usr/bin/redis-server /etc/redis/redis.conf --bind {{ LOOPBACK_IP }} {{ redis_items['hostname'] }} --port {{ redis_items['port'] }} --unixsocket {{ redis_items['unix_socket_path'] }} --pidfile /var/run/redis/{{ redis_inst }}.pid --dir /var/lib/{{ redis_inst }}"
```
qiluo-msft, line 39: >bind [](http://example.com/codeflow?start=207&length=4)

why removing?

dockers/docker-snmp/snmpd.conf.j2
```diff
@@ -13,16 +13,10 @@
 #  AGENT BEHAVIOUR
 #
 
-#  Listen for connections on all ip addresses, including eth0, ipv4 lo
+#  Listen for connections on all ip addresses, including eth0, ipv4 and ipv6 lo
 #
-{% if SNMP_AGENT_ADDRESS_CONFIG %}
-{% for (agentip, port, vrf) in SNMP_AGENT_ADDRESS_CONFIG %}
```
qiluo-msft, line 19: >SNMP_AGENT_ADDRESS_CONFIG [](http://example.com/codeflow?start=31&length=25)

This feature is still useful, and there are some other PRs trying to extend it.

dockers/docker-database/Dockerfile.j2
qiluo-msft, line 35: ># bind 127.0.0.1 ::1 [](http://example.com/codeflow?start=36&length=20)

It will bind to all the interfaces visibile to database container. To be super cautious, could you add a sonic-mgmt testcase to ensure redis is not expose external port on sonic?

files/build_templates/docker_image_ctl.j2
```diff
@@ -538,9 +548,51 @@ start() {
...
+        ipc_var=""
+    else
+        pmon_sdk_sock="SX_API_SOCKET_FILE=/var/run/sx_sdk/sx_api_0.sock"
+        syncd_flag="--device"
+        syncd_var="/dev/sxdevs/sxcdev$(( $DEV + 1 )):/dev/sxdevs/sxcdev"
+        ipc_var="private"
+    fi
+{%- if docker_container_name == "snmp" %}
+# get snmp listening address and port list from redis
+addr_port_values=$(python3 -c 'from swsscommon.swsscommon import ConfigDBConnector; cfg_db = ConfigDBConnector();  cfg_db.connect(wait_for_init=True, retry_on=True); [print(k[0] + "|" + k[1]) for k in cfg_db.get_keys("SNMP_AGENT_ADDRESS_CONFIG|*")]')
```
SuvarnaMeenakshi, line 565: why is this logic moved here instead of doing in snmpd.conf.j2 file?
What is the value add we get?
This is done during docker create. If we are just restarting snmp server and there is a change in the SNMP_AGENT_ADDRESS_CONFIG in config_db, will the new change get picked up if we just restart running docker?

files/build_templates/docker_image_ctl.j2
```diff
@@ -523,6 +531,8 @@ start() {
         {%- if docker_container_name == "database" %}
         NET="bridge"
         DB_OPT=$DB_OPT" -v /var/run/redis$DEV:/var/run/redis:rw "
+	{%- elif docker_container_name == "snmp" %}
+        NET="bridge"
```
maipbui, line 535: >NET="bridge" [](http://example.com/codeflow?start=8&length=12)

Checking best practice ([https://docs.docker.com/network/network-tutorial-standalone/](https://docs.docker.com/network/network-tutorial-standalone/)), default bridge network is not best choice for production. User-defined bridge networks is best choice for production. Can we use user-defined bridge network here?

files/build_templates/docker_image_ctl.j2
ycoheNvidia: This was moved here because port and address forwarding can only be done during docker create. What it means that in order to harden this container's network - each address/port change for this service will require docker removal and creation

dockers/docker-snmp/snmpd.conf.j2
ycoheNvidia: I am sorry, I wasn't aware of the other PR's for this feature.
The logic of address/port/vrf was moved to the docker create port and address forwarding. Since only user defined ports and addresses will be forwarded, this logic will no longer be relevant here

files/build_templates/docker_image_ctl.j2
Yarden-Z: I agree, defining a user-defined bridge is the best case solution for this item.
Problem is - this FR wasn't scoped to define this user-bridge and therefor remained in the general bridge domain.
If such a bridge will be defined - then we can add it to the SNMP container and redis container (and any future container) as well.

dockers/docker-database/Dockerfile.j2
andriydnvd: Redis protected mode only prevents the Redis server from responding to queries outside the loopback interfaces.
Docker network mode was changed from host to bridged. This requires to be able to connect from host IP to Redis docker IP.

files/build_templates/docker_image_ctl.j2
```diff
@@ -483,7 +483,15 @@ start() {
     {%- endif %}
 
     if [ -z "$DEV" ]; then
-        NET="host"
+	{%- if docker_container_name == "database" %}
+        NET="bridge"
+	PORT_MAP="-p 127.0.0.1:6379:6379"
```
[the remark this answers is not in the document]
andriydnvd: If you mean multi-asic case. Its working. Each asic db docker has its own IP with mapped 6379 listening port and shared redis.sock socket file. 
If there are more use cases, that may require different ports please elaborate.

dockers/docker-snmp/snmpd.conf.j2
ycoheNvidia: I have updated the code to include the latest merged PR

ycoheNvidia:
@qiluo-msft we addressed the comments and I adjusted the code to the recent merged snmp code. Can we proceed please?

files/build_templates/docker_image_ctl.j2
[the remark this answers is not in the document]
qiluo-msft: You may need to fetch the ports for all redis instances from conf files: "/var/run/redis/sonic-db/database_config.json".

With multi-DB feature, there will be multiple-namespace, and multiple instance per namespace. ref: [https://github.com/sonic-net/SONiC/blob/master/doc/database/multi_namespace_db_instances.md](https://github.com/sonic-net/SONiC/blob/master/doc/database/multi_namespace_db_instances.md)

files/build_templates/docker_image_ctl.j2
qiluo-msft: This PR will kill the feature of SNMP_AGENT_ADDRESS_CONFIG/port. Do you want to implement the same feature here or in another PR?

ycoheNvidia: I am not sure I understand your question. This PR and https://github.com/sonic-net/sonic-snmpagent/pull/281 moves the logic of snmp networking management from the container to the host. The host will forward the relevant packets only (SNMP_AGENT_ADDRESS_CONFIG for example) and the container will take all packets sent to port 161

dockers/docker-snmp/snmpd.conf.j2
```diff
@@ -13,41 +13,11 @@
...
-{% if LOOPBACK_INTERFACE is defined %}
-{% for lo in LOOPBACK_INTERFACE %}
-{% if lo | length == 2 %}
-{% set agentip = lo[1].split('/')[0] %}
-agentAddress {{ protocol(agentip) }}:[{{ agentip }}]:161
-{% endif %}
-{% endfor %}
-{% endif %}
-{% else %}
+#  Listen for connections on all ip addresses, including eth0, ipv4 and ipv6 lo
```
qiluo-msft, line 16: >  [](http://example.com/codeflow?start=2&length=1)

Remove one extra blank after `#`.

files/build_templates/docker_image_ctl.j2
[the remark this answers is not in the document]
qiluo-msft: still applicable.

dockers/docker-snmp/snmpd.conf.j2
qiluo-msft: Could you explain "The logic of address/port/vrf was moved to the docker create port and address forwarding"?

ycoheNvidia: In the current implementation, snmpd has its configuration in snmpd.conf which defines which addresses and ports can be used for query, how and if vrf is being used, etc.
Removing snmp container from host network means that it is not exposed to any of these addresses by default,
What we did here was to use docker address:port forwarding as the method to implement the logic for it. When a user configures address and port to be used for snmp queries - this address:port tuple will be used when snmp docker is created, similar to a firewall. In what we offer here, the snmp demon inside the container will listen to all packets, and the docker networking logic (address:port forwarding) will forward only the relevant packets, as configured by user

files/build_templates/docker_image_ctl.j2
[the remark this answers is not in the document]
andriydnvd: Thanks, checking

qiluo-msft:
Please resolove conflict

files/build_templates/docker_image_ctl.j2
```diff
@@ -529,9 +539,58 @@ start() {
...
+        syncd_var="/dev/sxdevs/sxcdev$(( $DEV + 1 )):/dev/sxdevs/sxcdev"
+        ipc_var="private"
+    fi
+{%- if docker_container_name == "snmp" %}
+# get snmp listening address and port list from redis
+addr_port_values=$(python3 -c 'from swsscommon.swsscommon import ConfigDBConnector; cfg_db = ConfigDBConnector();  cfg_db.connect(wait_for_init=True, retry_on=True); [print(k[0] + "|" + k[1]) for k in cfg_db.get_keys("SNMP_AGENT_ADDRESS_CONFIG|*")]')
+NAMESPACE_COUNT=$NUM_ASIC
+if [ -z $addr_port_values ]; then
+if [ -z $NAMESPACE_COUNT ] || [ $NAMESPACE_COUNT -lt 2 ]; then
+addr_port_values=$(python3 -c 'from swsscommon.swsscommon import ConfigDBConnector; cfg_db = ConfigDBConnector();  cfg_db.connect(wait_for_init=True, retry_on=True); [print(k[1].split("/")[0].lower() + "%" + k[0]) if len(k) == 2 and k[1].split('/')[0].lower().startswith("fe80") else print(k[1].split("/")[0].lower()) for k in cfg_db.get_keys("LOOPBACK_INTERFACE|*")+cfg_db.get_keys("MGMT_INTERFACE|*") if len(k) == 2]')
```
SuvarnaMeenakshi, line 560: here currently only Loopback and Mgmt IP addresses are used.
There is SNMP_AGENT_ADDRESS_CONFIG table which can be configured with the IPs for snmpd to listen on.
This change does not include using that.

ycoheNvidia: @SuvarnaMeenakshi  you either missed line 556 which addresses the specific SNMP_AGENT_ADDRESS_CONFIG table, or I miss something. I followed the logic that was found in snmpd.conf.j2 and translated it here

yaqiangz:
Hi @ycoheNvidia, have you verified whether syslog generated inside bridge mode container can be written to host machine and update to remote syslog server?

files/build_templates/docker_image_ctl.j2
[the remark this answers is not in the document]
qiluo-msft: still applicable

files/build_templates/docker_image_ctl.j2
```diff
@@ -474,7 +474,15 @@ start() {
     {%- endif %}
 
     if [ -z "$DEV" ]; then
-        NET="host"
+	{%- if docker_container_name == "database" %}
+        NET="bridge"
+	PORT_MAP="-p 127.0.0.1:6379:6379"
```
qiluo-msft, line 479: >	 [](http://example.com/codeflow?start=0&length=1)

mix tabs and spaces.

files/build_templates/docker_image_ctl.j2
```diff
@@ -529,9 +539,58 @@ start() {
...
+        syncd_var="/dev/shm:/dev/shm:rw"
+        ipc_var=""
+    else
+        pmon_sdk_sock="SX_API_SOCKET_FILE=/var/run/sx_sdk/sx_api_0.sock"
+        syncd_flag="--device"
+        syncd_var="/dev/sxdevs/sxcdev$(( $DEV + 1 )):/dev/sxdevs/sxcdev"
+        ipc_var="private"
+    fi
+{%- if docker_container_name == "snmp" %}
+# get snmp listening address and port list from redis
```
qiluo-msft, line 555: ># get snmp [](http://example.com/codeflow?start=0&length=10)

This block is not well formatted.

files/build_templates/docker_image_ctl.j2
```diff
@@ -646,6 +705,9 @@ stop() {
     /usr/local/bin/container stop -t 60 $DOCKERNAME
     {%- else %}
     /usr/local/bin/container stop $DOCKERNAME
+    {%- if docker_container_name == "snmp" %}
+    docker rm $DOCKERNAME
```
qiluo-msft, line 709: >docker rm [](http://example.com/codeflow?start=4&length=9)

Why remove docker container?

files/build_templates/docker_image_ctl.j2
ycoheNvidia: removed

files/build_templates/docker_image_ctl.j2
```diff
@@ -474,7 +474,15 @@ start() {
     {%- endif %}
 
     if [ -z "$DEV" ]; then
-        NET="host"
+    {%- if docker_container_name == "database" %}
+        NET="bridge"
+        PORT_MAP="-p 127.0.0.1:6379:6379"
+    {%- elif docker_container_name == "snmp" %}
+    NET="bridge"
```
yaqiangz, line 481: I just tried image built by this PR yesterday, version: SONiC.master-15176.360116-6bb24c135. Looks like rsyslog inside snmp container cannot work well. I cannot find any syslog generated by snmp container (contains 'snmp#') in local host syslog file or our syslog server, and process `containercfgd` which is to dynamically generate rsyslog configuration file inside container doesn't startup. Not sure whether you have verified this, could you confirm that?

files/build_templates/docker_image_ctl.j2
qiluo-msft: From Line 544 to Line 573, this block is still not well formatted:
1. `if` not indented
2. use 2 spaces each indentation level, prefer 4 instead

dockers/docker-database/supervisord.conf.j2
```diff
@@ -36,7 +36,7 @@ dependent_startup=true
...
+command=/bin/bash -c "{ [[ -s /var/lib/{{ redis_inst }}/dump.rdb ]] || rm -f /var/lib/{{ redis_inst }}/dump.rdb; } && mkdir -p /var/lib/{{ redis_inst }} && exec /usr/bin/redis-server /etc/redis/redis.conf --port {{ redis_items['port'] }} --unixsocket {{ redis_items['unix_socket_path'] }} --pidfile /var/run/redis/{{ redis_inst }}.pid --dir /var/lib/{{ redis_inst }}"
```
qiluo-msft, line 39: >port [](http://example.com/codeflow?start=207&length=4)

Could you move database container change to another PR? we can focus on snmp container in this PR.

ycoheNvidia: Will do this week

ycoheNvidia: Done

files/build_templates/docker_image_ctl.j2
qiluo-msft: If there is hot patches inside docker container, original behavior is that the patches will survive config reload or reboot. But this PR lose the capabilities.

files/build_templates/docker_image_ctl.j2
```diff
@@ -486,7 +486,12 @@ start() {
     {%- endif %}
 
     if [ -z "$DEV" ]; then
-        NET="host"
+    {%- if docker_container_name == "snmp" %}
```
qiluo-msft, line 489: >{%- [](http://example.com/codeflow?start=4&length=3)

Indent these 5 lines.

files/build_templates/docker_image_ctl.j2
Yarden-Z: Are there real use-cases where a docker is being loaded with patches inside it?
There might also be some local files (such as temp config and such in every container), is this a real use-case?

qiluo-msft: Yes, it is possible in production environment.

files/build_templates/docker_image_ctl.j2
SuvarnaMeenakshi: @ycoheNvidia  could check if change like this is needed. https://github.com/sonic-net/sonic-buildimage/pull/4738

SuvarnaMeenakshi:
@ycoheNvidia  2 issues that I observed when using this diff with latest master branch image:
1.  redis-connection is not working from snmp_subagent even with the PR change in snmp submodule, I added a comment on that PR. https://github.com/sonic-net/sonic-snmpagent/pull/281/
2. SNMP walk is failing for single-asic device, I tried adding iptable NAT rule to modify destination address as eth0 IP address inside snmp docker and modify the SRC while sending out response to docker0 IP, with this I see that snmp query works fine. Wondering if you verified the snmp query with these changes?
sudo ip6tables -t nat -A PREROUTING -p udp -m udp --dport 161 -j DNAT --to-destination fd00::242:f07f:102 (eth0 IP of snmp docker )
sudo ip6tables -t nat -A POSTROUTING -p tcp -m tcp --dport 161 -j SNAT --to-source  fd00::1 (docker0 Ipv6)
sudo iptables -t nat -A PREROUTING -p udp -m udp --dport 161 -j DNAT --to-destination 240.127.1.2 (eth0 IP of snmp docker)
sudo iptables -t nat  -A POSTROUTING -p udp -m udp --dport 161 -j SNAT --to-source 240.127.1.1 (docker0 Ipv4)
We need to check if these rules have to added as postStartAction in /usr/bin/snmp.sh after starting docker or if they should be added in caclmgrd. Something similar is done for multi-asic.

files/build_templates/docker_image_ctl.j2
qiluo-msft: Still applicable.

files/build_templates/docker_image_ctl.j2
qiluo-msft: Still applicable.

files/build_templates/docker_image_ctl.j2
qiluo-msft: Still applicable.

PR #15184, opened 2023-05-22
[arm64] Update Marvell SAI to 1.11.0-8

Pavan-Nokia:
#### Why I did it
 Update Marvell SAI to 1.11.0-8

##### Work item tracking

#### How to verify it
Build marvell-arm64 target using below instructions
make init
make configure PLATFORM=marvell-arm64 PLATFORM_ARCH=arm64
make target/sonic-marvell-arm64.bin SONIC_BUILD_JOBS=2

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211

- [ ] 
- [ ]

### DIFF
--- a/platform/marvell-arm64/sai.mk
+++ b/platform/marvell-arm64/sai.mk
@@ -1,6 +1,6 @@
 # Marvell SAI
 
-export MRVL_SAI_VERSION = 1.11.0-1
+export MRVL_SAI_VERSION = 1.11.0-8
 export MRVL_SAI = mrvllibsai_$(MRVL_SAI_VERSION)_$(PLATFORM_ARCH).deb
 
 $(MRVL_SAI)_SRC_PATH = $(PLATFORM_PATH)/sai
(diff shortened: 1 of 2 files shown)


liushilongbuaa:
Which line to avoid race condition?

Pavan-Nokia:
> Which line to avoid race condition?

The upgrade of SAI from 1.11.0-1 to 1.11.0-8 helps with avoiding the race condition. 
•	libsaivs is added by sonic-buildimage in rules/sairedis.mk
•	In sai.mk these packages are mentioned as conflicting. 
    In a working install log for mrvllibsai is uninstalled before installing libsaivs-dev. 

And the cleanup in Makefile was removed to prevent false positive. The issue with the cleanup is the bash statement  "download || remove"  is , if either is successful, then the whole line is successful which is not what we want.

Also @liushilongbuaa  I had sent you an email with more findings yesterday. did you you get a chance to look at that

PR #15228, opened 2023-05-26
[arm64] Fix marvell-arm64 pipeline build

Pavan-Nokia:
#### Why I did it
When git clone -b xxx command is used the versions-git will reset the HEAD of the git to the commit ID in the versions-git file. Which causes incorrect commit to be checked out causing build errors. 

##### Work item tracking

#### How I did it
Split ‘git clone -b’ into two steps to avoid owerwrite
1.	Git clone
2.	cd mrvl-prestera; git checkout ; cd ..

#### How to verify it
Build marvell-arm64 target using below instructions
make init
make configure PLATFORM=marvell-arm64 PLATFORM_ARCH=arm64
make target/sonic-marvell-arm64.bin SONIC_BUILD_JOBS=2

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211

- [ ] 
- [ ]

### DIFF
--- a/platform/marvell-arm64/prestera/debian/rules
+++ b/platform/marvell-arm64/prestera/debian/rules
@@ -21,7 +21,8 @@ clean:
 build:
 	# get sources
 	rm -rf mrvl-prestera || true
-	git clone -b ${MRVL_PRESTERA_SRC_TAG} ${MRVL_PRESTERA_SRC_URL}
+	git clone ${MRVL_PRESTERA_SRC_URL}
+	cd mrvl-prestera && git checkout ${MRVL_PRESTERA_SRC_TAG} && cd ..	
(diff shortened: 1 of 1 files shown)


liushilongbuaa (approved):
build passed.

Pavan-Nokia:
@lguohan Please help merge this PR.

@liushilongbuaa has confirmed a successful marvell-arm64 pipeline build with this PR

xumia:
It is caused by the same url by different tag used in armhf and arm64.
In armhf, we set the tag to MRVL_PRESTERA_DRIVER_1.2, but in arm64 the tag is MRVL_PRESTERA_DRIVER_1.3.
Currently, the git reproducible build only supports one url with one commit, not take the branches/tags into account the commit id to lock.

We can merge the code to bypass the issue, it can be reverted once the git for multiple branches/tags supported.

PR #15269, opened 2023-05-31
[staticroutebfd]fix an issue on deleting a non-bfd static route

baorliu:
#### What I did it

Fix an issue for deleting a non-bfd static route
also remove the staticroutebfd from critical_processes list and make it auto restart in the case of crash. 

(Will provide github issue number and update this description later)

#### Why I did it
Current design access "None" object (nh_vrf_list) when delete a non-bfd static route

For supervisord, because staticroutebfd support restart and recovery. so it is not necessary to restart bgp container if there is any crash inside staticroutebfd. supervisord will restart it after crash.

##### Work item tracking

#### How I did it
check if the route is BFD protected when delete this static route, and check if the nh_vrf_list is None.

#### How to verify it

add a UT deleting a non-bfd static route to reproduce the issue:
```
self = <staticroutebfd.main.StaticRouteBfd object at 0x7f2f1add8b80>, key = '3.3.3.0/24', redis_del = {'ifname': 'if1, if2', 'nexthop': '192.168.1.2 , 192.168.2.2'}

    def static_route_del_handler(self, key, redis_del):
        valid, vrf, ip_prefix = static_route_split_key(key)
        if not valid:
            return True
        route_cfg_key = vrf + "|" + ip_prefix
    
        valid, is_ipv4, ip = check_ip(ip_prefix)
        if not valid:
            return True
    
        data = self.get_local_db(LOCAL_CONFIG_TABLE, route_cfg_key)
        if len(data) == 0:
            # this route is not handled by StaticRouteBfd, skip
            return True
    
        arg_list    = lambda v: [x.strip() for x in v.split(',')] if len(v.strip()) != 0 else None
        nh_list     = arg_list(data['nexthop']) if 'nexthop' in data else None
        nh_vrf_list = arg_list(data['nexthop-vrf']) if 'nexthop-vrf' in data else None
        for index in range(len(nh_list)):
            nh_ip = nh_list[index]
>           nh_vrf = nh_vrf_list[index]
E           TypeError: 'NoneType' object is not subscriptable

staticroutebfd/main.py:507: TypeError
```
fix the issue and run the UT:
```
staticroutebfd/__init__.py              0      0   100%
staticroutebfd/main.py                577    204    65%
staticroutebfd/vars.py                  5      0   100%
-------------------------------------------------------
TOTAL                                2367   1689    29%

================================================================================================ 32 passed in 0.74s =================================================================================================
```

For supervisord, after changed the critical_processes.j2 and supervisord.conf.j2, restart bgp (sudo systemctl restart bgp),
manually kill the staticroutebfd process inside bgp container, it can be restarted without restart the whole bgp container.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211

- [ ] 
- [ ]


src/sonic-bgpcfgd/staticroutebfd/main.py
```diff
@@ -502,18 +502,26 @@ def static_route_del_handler(self, key, redis_del):
...
+                    bfd_key = nh_vrf + ":default:" + nh_ip
+                    self.remove_from_local_db(LOCAL_BFD_TABLE, bfd_key)
+                    self.del_bfd_session_from_appl_db(bfd_key)
 
-            if len(self.get_local_db(LOCAL_NEXTHOP_TABLE, nh_key)) == 0:
-                bfd_key = nh_vrf + ":default:" + nh_ip
-                self.remove_from_local_db(LOCAL_BFD_TABLE, bfd_key)
-                self.del_bfd_session_from_appl_db(bfd_key)
+        # do not delete it from appl_db if the route is not bfd enabled
+        if bfd_enabled:
```
dgsudharsan, line 522: Why aren't we removing the static route if bfd not enabled? What was the previous behavior? Does this mean if the user removes static route via CLI, won't it get removed from APP_DB?

dgsudharsan (changes requested):

src/sonic-bgpcfgd/staticroutebfd/main.py
baorliu: If BFD is not enabled, staticroutebfd does not write static route entry to appl_db. so it should not delete it from appl_db. In general, prefix in config_db should not in appl_db. but to be safe, it is better that staticroutebfd does not touch that route in appl_db if it is not created by  staticroutebfd.

dgsudharsan (approved):

vivekrnv (changes requested):

dockers/docker-fpm-frr/frr/supervisord/critical_processes.j2
```diff
@@ -9,5 +9,4 @@ program:pimd
 program:frrcfgd
 {%- else %}
 program:bgpcfgd
-program:staticroutebfd
```
vivekrnv, line 12: But Isn't this a critical process?

baorliu: staticroutebfd itself supports recovery from restart/crash. So it is not necessary to set it to critical process. When a critical process crashes, the bgp container itself will restart, which has more impact. to reduce impact to the bgp container, removed it from critical process, and let supervisord restart staticroutebfd only (see change in supervisord.conf.j2 below).

src/sonic-bgpcfgd/tests/test_static_rt_bfd.py
```diff
@@ -169,6 +168,28 @@ def test_set_del():
...
+        }),
+        {},
+        {}
+    )
+
+    # test delete a non-bfd static route
+    set_del_test(dut, "srt",
+        "DEL",
+        ("3.3.3.0/24", {
+            "nexthop": "192.168.1.2 , 192.168.2.2",
```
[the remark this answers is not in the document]
baorliu: I will remove that to prevent DUT using these values in case

src/sonic-bgpcfgd/tests/test_static_rt_bfd.py
```diff
@@ -169,6 +168,28 @@ def test_set_del():
         {'del_default:2.2.2.0/24': {}}
     )
 
+    # test add a non-bfd static route
+    set_del_test(dut, "srt",
```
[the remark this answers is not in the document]
baorliu: new testcases are added for nexthop-vrf list. thanks

vivekrnv (approved):

dgsudharsan:
@prsunny Can we merge this?

dgsudharsan (approved):

prsunny (approved):

abdosi:
@baorliu can you create direct PR for 202205 branch on msft repo ?

That way we need not have to cherry-pick.

cc @anamehra @gechiang

PR #15284, opened 2023-05-31
Use execle instead of popen in tacas nss to avoid shell escape exploits

seiferteric:
#### Why I did it

Tacacs nss library uses popen to execute useradd and usermod commands. Popen executes using a shell (/bin/sh) which is passed the command string with "-c". This means that if untrusted user input is supplied, unexpected shell escapes can occur. In this case the username supplied can be untrusted user input when logging in via ssh or other methods when tacacs is enabled. Debian has very little limitation on usernames and as such characters such as quotes, braces, $, >, | etc are all allowed. Since the nss library is run by root, any shell escape will be ran as root.

In the current community version of tacacs nss library, the issue is mitigated by the fact that the useradd command is only ran if the user is found to exist on the tacacs server, so the bad username would have to already exists there which is unlikely. However, internally (at Dell) we had to modify this behavior to support other tacacs servers that do not allow authorization messages to verify user existence prior to a successful authentication. These servers include Cisco ISE and Aruba ClearPass. In order to support these tacacs+ servers, we have to create a temporary user immediately, which means this would be a much bigger issue. 

I also plan to supply the patch to support ISE and ClearPass and as such, I would suggest taking this patch to remediate this issue first. 

#### How I did it
Replace call to popen with fork/execl of the useradd/usermod binary directly.

#### How to verify it

Install patched version of libnss-tacplus and verify that tacacs+ user login still works as expected.

### DIFF
--- a/src/tacacs/nss/patch/0011-Replace-popen-shell-execution-with-safer-execle.patch
+++ b/src/tacacs/nss/patch/0011-Replace-popen-shell-execution-with-safer-execle.patch
@@ -1,0 +1,109 @@
+From 9e46c167abeb461456b0c432d6a75559013a79ce Mon Sep 17 00:00:00 2001
+From: Eric Seifert <seiferteric@gmail.com>
+Date: Wed, 31 May 2023 09:08:07 -0700
+Subject: [PATCH] Replace popen shell execution with safer execle
+
+---
+ nss_tacplus.c | 65 +++++++++++++++++++++++++++++++++++++++++----------
+ 1 file changed, 53 insertions(+), 12 deletions(-)
+
+diff --git a/nss_tacplus.c b/nss_tacplus.c
+index cd73870..8db8f58 100644
+--- a/nss_tacplus.c
++++ b/nss_tacplus.c
+@@ -34,6 +34,9 @@
+ #include <netdb.h>
+ #include <nss.h>
+ #include <limits.h>
++#include <unistd.h>
++#include <sys/wait.h>
++
+ 
+ #include <libtac/libtac.h>
+ 
+@@ -439,6 +442,39 @@ static int delete_conf_line(const char *name)
+     return 0;
+ }
+ 
++int user_mod_add(const char* cmd, const char* name, char* gid, char* sec_grp, char* gecos, char* home, char* shell) {
++
++    pid_t pid;
++    int wstatus;
++
++    pid = fork();
++
++    if(pid > 0) {
++        do {
++            if (waitpid(pid, &wstatus, WUNTRACED | WCONTINUED) == -1) {
++                int errsv = errno;
++                char serr[256] = {0};
++                strerror_r(errsv, serr, 256);
++                syslog(LOG_ERR, "%s: exec of %s failed with error %d: %s", nssname, cmd, errsv, serr);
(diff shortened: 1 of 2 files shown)


seiferteric:
@lguohan Can you review? Also looks like some testbed failure that has nothging to do with my changes, can we restart the failed test above? Thanks.

adyeung:
@a-barboza pls help review the change

a-barboza:
> @a-barboza pls help review the change

@adyeung @seiferteric @shdasari 

I recommend merging this patch. There are some minor enhancements (replace integer constants with macros which could be done). But, I recommend merging this patch. Please let me know if anything else needs to be done from my side. thank you

seiferteric:
Okay, the test was passing, I was just looking at the old test the first time it failed. Yes please review.

a-barboza (approved):

lguohan:
@liuh-80 to review

src/tacacs/nss/patch/0011-Replace-popen-shell-execution-with-safer-execle.patch
```diff
@@ -0,0 +1,109 @@
...
++            }
++        } while (!WIFEXITED(wstatus) && !WIFSIGNALED(wstatus));
++        if WIFEXITED(wstatus)
++            return WEXITSTATUS(wstatus);
++        else
++            return -1;
++    // Child
++    } else if(pid == 0) {
++        execl(cmd, cmd, "-G", sec_grp, name, "-g", gid, "-c", gecos, "-d", home, "-m", "-s", shell, NULL);
++        syslog(LOG_ERR, "%s: exec of %s failed!", nssname, cmd);
```
liuh-80, line 52: Add errno to error log?

src/tacacs/nss/patch/0011-Replace-popen-shell-execution-with-safer-execle.patch
```diff
@@ -0,0 +1,109 @@
...
+-                command, user->secondary_grp, name, user->gid, user->info, name, user->shell);
+-            if(debug) syslog(LOG_DEBUG, "%s", buf);
+-            fp = popen(buf, "r");
+-            if(!fp || -1 == pclose(fp)) {
+-                syslog(LOG_ERR, "%s: %s popen failed errno=%d %s",
+-                        nssname, command, errno, strerror(errno));
+-                delete_conf_line(name);
+-                return -1;
+-            }
+-            if(debug)
```
liuh-80, line 90: Please also keep the success log.

src/tacacs/nss/patch/0011-Replace-popen-shell-execution-with-safer-execle.patch
seiferteric: Okay added

src/tacacs/nss/patch/0011-Replace-popen-shell-execution-with-safer-execle.patch
seiferteric: Okay added

liuh-80 (approved):

maipbui (approved):

PR #15333, opened 2023-06-05
[sflow]Delay starting sflow service until ports are created

dgsudharsan:
#### Why I did it
Delaying sflow service until interfaces are created.  Starting sflow early leads to CPU contention and issues like [this](https://github.com/sonic-net/sonic-buildimage/issues/13407) where sflow daemon will try to access interfaces even before they are created.
 
##### Work item tracking

#### How I did it
Set delayed to true in init_cfg.j2

#### How to verify it
Perform config reload and reboot to ensure sflow is started only after interfaces are created

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211

- [ ] 
- [ ]

### DIFF
--- a/files/build_templates/init_cfg.json.j2
+++ b/files/build_templates/init_cfg.json.j2
@@ -51,7 +51,7 @@
 {%- if include_nat == "y" %}{% do features.append(("nat", "disabled", false, "enabled")) %}{% endif %}
 {%- if include_p4rt == "y" %}{% do features.append(("p4rt", "disabled", false, "enabled")) %}{% endif %}
 {%- if include_restapi == "y" %}{% do features.append(("restapi", "enabled", false, "enabled")) %}{% endif %}
-{%- if include_sflow == "y" %}{% do features.append(("sflow", "disabled", false, "enabled")) %}{% endif %}
+{%- if include_sflow == "y" %}{% do features.append(("sflow", "disabled", true, "enabled")) %}{% endif %}
 {%- if include_macsec == "y" %}{% do features.append(("macsec", "{% if 'type' in DEVICE_METADATA['localhost'] and DEVICE_METADATA['localhost']['type'] == 'SpineRouter' and DEVICE_RUNTIME_METADATA['MACSEC_SUPPORTED'] %}enabled{% else %}disabled{% endif %}", false, "enabled")) %}{% endif %}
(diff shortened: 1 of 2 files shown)


venkatmahalingam:
@Gokulnath-Raja Please take a look at the changes.

files/build_templates/init_cfg.json.j2
```diff
@@ -51,7 +51,7 @@
 {%- if include_nat == "y" %}{% do features.append(("nat", "disabled", false, "enabled")) %}{% endif %}
 {%- if include_p4rt == "y" %}{% do features.append(("p4rt", "disabled", false, "enabled")) %}{% endif %}
 {%- if include_restapi == "y" %}{% do features.append(("restapi", "enabled", false, "enabled")) %}{% endif %}
-{%- if include_sflow == "y" %}{% do features.append(("sflow", "disabled", false, "enabled")) %}{% endif %}
+{%- if include_sflow == "y" %}{% do features.append(("sflow", "disabled", true, "enabled")) %}{% endif %}
```
Gokulnath-Raja, line 54: Just cross check do we need to modify UT test script for config changes for sflow?
https://github.com/sonic-net/sonic-utilities/blob/1c1e22d3b6123f5635f3ea5f525ed7f1c1d97926/tests/feature_test.py

files/build_templates/sflow.service.j2
```diff
@@ -14,6 +14,3 @@ ExecStartPre=/usr/bin/{{docker_container_name}}.sh start
 ExecStart=/usr/bin/{{docker_container_name}}.sh wait
 ExecStop=/usr/bin/{{docker_container_name}}.sh stop
 RestartSec=30
-
```
Gokulnath-Raja, line 17: Any specific reason for removing this ?

dgsudharsan: In order for the daemon to be managed by hostcfgd, it needs to be removed from sonic.target. If we don't remove, config reload and reboot will implicitly start sflow since it is associated with sonic.target. Removing this means hostcfgd will listen to PortInitDone and only then spawn the daemon

files/build_templates/init_cfg.json.j2
dgsudharsan: The tests use mock files and hence its not required to modify UT

Gokulnath-Raja (approved):

prsunny (approved):

PR #15349, opened 2023-06-06
updated internal route policy for chassis-packet

abdosi:
What I did:

Workaround for the issue seen here : https://github.com/FRRouting/frr/issues/13682
It seems there is timing issue where there are multiple recursive lookup needed to resolve nexthop of the route it's possible that it does not happen correctly causing route to remain in `inactive` state

Issue is seen on chassis-packet as there 2 level of recursive lookup needed for a given e-BGP learnt route
     - Level1 to resolve e-BGP peer (connected route via bgp ) over Loopback4096 (i-BGP peering)
     - Level 2  Loopback4096 over backend port-channels next-hops

For VOQ chassis there is no `e-BGP peer (connected route via bgp ) ` resolution as route is added as Static route by orchagent over `Ethernet-IB`.

Also as part of this remove route-map policy from `instance.conf.j2` as same is define in `peer-group.j2`.

Microsoft ADO: https://msazure.visualstudio.com/One/_workitems/edit/24198507

How I verify:
Functional Verification manually
Updated UT.
We will be adding sanity check in sonic-mgmt to make sure none of route are in inactive state.


abdosi:
cc @anamehra / @vperumal

arlakshm (approved):

abdosi:
@yxieca  can you please help merge this. Also cheery-pick is needed for 2022205.

yxieca:
202305 branch hasn't been officially cut yet. Please use this flag after 6/16/2023.

PR #15357, opened 2023-06-06
[systemd-sonic-generator] replace sscanf with strtol

maipbui:
#### Why I did it
Avoid 'sscanf()' for number conversions. Its use can lead to undefined behavior, slow processing, and integer overflows. Instead prefer the 'strto*()' family of functions.
##### Work item tracking

#### How I did it
replace sscanf with strtol
#### How to verify it
Manual test

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211

- [ ] 
- [ ]

### DIFF
--- a/src/systemd-sonic-generator/systemd-sonic-generator.c
+++ b/src/systemd-sonic-generator/systemd-sonic-generator.c
@@ -569,13 +569,11 @@ int get_num_of_asic() {
                     str_num_asic = strtok_r(NULL, "=", &saveptr);
                     strip_trailing_newline(str_num_asic);
                     if (str_num_asic != NULL){
-                        char *endptr;
-                        num_asic = strtol(str_num_asic, &endptr, 10);
-                        if (*endptr != '\0') {
-                            fprintf(stderr, "Failed to get num asics.\n");
+                        num_asic = strtol(str_num_asic, NULL, 10);
+                        if (num_asic > 0) {
+                            break;
                         }
                     }
-                    break;
                 }
             }
             fclose(fp);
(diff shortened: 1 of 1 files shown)


src/systemd-sonic-generator/systemd-sonic-generator.c
```diff
@@ -569,7 +569,11 @@ int get_num_of_asic() {
                     str_num_asic = strtok_r(NULL, "=", &saveptr);
                     strip_trailing_newline(str_num_asic);
                     if (str_num_asic != NULL){
-                        sscanf(str_num_asic, "%d",&num_asic);
+                        char *endptr;
+                        num_asic = strtol(str_num_asic, &endptr, 10);
+                        if (*endptr != '\0') {
+                            fprintf(stderr, "Failed to get num asics.\n");
```
qiluo-msft, line 575: >fprintf [](http://example.com/codeflow?start=28&length=7)

You will treat any extra chars after num as failure. I am thinking original sscanf just ignore them, and it's safe to ignore them using strtol.

How about just checking `num_asic > 0` ?

qiluo-msft (approved):
LGTM

src/systemd-sonic-generator/systemd-sonic-generator.c
```diff
@@ -569,9 +569,11 @@ int get_num_of_asic() {
                     str_num_asic = strtok_r(NULL, "=", &saveptr);
                     strip_trailing_newline(str_num_asic);
                     if (str_num_asic != NULL){
-                        sscanf(str_num_asic, "%d",&num_asic);
+                        num_asic = strtol(str_num_asic, NULL, 10);
+                        if (num_asic > 0) {
+                            break;
```
SuvarnaMeenakshi, line 574: I did not understand the usage of break here?
If num_asic is not greater than 0, then what might happen?

In what scenario will this check be useful? 
Previously, we were breaking after parsing the line starting with "NUM_ASIC=", we could still do the same without check for the value?

SuvarnaMeenakshi (approved):
lgtm

PR #15432, opened 2023-06-12
Bypass the systemd service restart limit and do immediately restart when change to local mode

lixiaoyuner:
#### Why I did it
- During the upgrade process via k8s, the feature's systemd service will restart as well, all of the feature systemd service has restart number limit, and the limit number is too small, only three times. if fallback happens when upgrade, the start count will be 2, just once again, the systemd service will be down. So, need to bypass this. This restart function will be called when do local -> kube, kube -> kube, kube ->local, each time call this function, we indeed need to restart successfully, so do reset-failed every time we do restart.
- When need to go back to local mode, we do systemd restart immediately without waiting the default restart interval time so that we can reduce the container down time.
##### Work item tracking
24172368
#### How I did it
- Before every restart for upgrade, do reset feature's restart number. The restart number will be reset to 0 to bypass the restart limit.
- When need to go back to local mode, we do systemd restart immediately.

#### How to verif it
Feature's systemd service can be always restarted successfully during upgrade process via k8s.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [x] 202205
- [x] 202211

- [x] 20220531.28
- [ ]


rules/config
```diff
@@ -178,7 +178,7 @@ INCLUDE_ROUTER_ADVERTISER ?= y
 
 # INCLUDE_KUBERNETES - if set to y kubernetes packages are installed to be able to
 # run as worker node in kubernetes cluster.
-INCLUDE_KUBERNETES ?= n
+INCLUDE_KUBERNETES ?= y
```
losha228, line 181: is that for test purpose ?

src/sonic-ctrmgrd/ctrmgr/ctrmgrd.py
```diff
@@ -581,7 +588,7 @@ def on_state_update(self, key, op, data):
 
     def do_tag_latest(self, feat, docker_id, image_ver):
         ret = kube_commands.tag_latest(feat, docker_id, image_ver)
-        if ret != 0:
+        if ret == 1:
```
losha228, line 591: I suppose 0 means success, other values means failure, right ? what's other values besides 0 and 1 ? why not retry in that  case ?

src/sonic-ctrmgrd/ctrmgr/kube_commands.py
```diff
@@ -478,7 +478,7 @@ def tag_latest(feat, docker_id, image_ver):
         else:
             log_error(err)
     elif ret == -1:
-        ret = 0
+        pass
```
losha228, line 481: log_debug(out) for this case

rules/config
lixiaoyuner: Yes, will change it back to "n" after test

src/sonic-ctrmgrd/ctrmgr/kube_commands.py
lixiaoyuner: Okay

src/sonic-ctrmgrd/ctrmgr/ctrmgrd.py
lixiaoyuner: Here has another code -1, it means the scheduled container is not stable, for this case, we don't do tag latest and clean up images

losha228 (approved):

qiluo-msft:
Your code changes is more than what you said in PR title and PR description. Could you update them?

lixiaoyuner:
> Your code changes is more than what you said in PR title and PR description. Could you update them?

Thanks for your comments, have updated, could you please go ahead to review.

losha228 (approved):

losha228 (approved):

src/sonic-ctrmgrd/ctrmgr/ctrmgrd.py
```diff
@@ -151,6 +151,7 @@ def is_systemd_active(feat):
 def restart_systemd_service(server, feat, owner):
     log_debug("Restart service {} to owner:{}".format(feat, owner))
     if not UNIT_TESTING:
+        subprocess.call(["systemctl", "reset-failed", str(feat)])
```
lguohan, line 154: why don't we check the return code here?

lguohan:
for the first pr, i think it should be a separate pr. I spent quite some time to figure where which line code map to this reset failed. even for this feature, we need to explain on this "It's easy to meet this limit when upgrade and fallback happen at the same time." why? i couldn't figure out.

lixiaoyuner:
> for the first pr, i think it should be a separate pr. I spent quite some time to figure where which line code map to this reset failed. even for this feature, we need to explain on this "It's easy to meet this limit when upgrade and fallback happen at the same time." why? i couldn't figure out.

Let me explain first how the systemd sevice restart when k8s upgrade the container. For example v1(kube) --> v2(kube), k8s will stop v1 container first, this time the systemd service is doing "docker wait v1-id", once the v1 container stops, the "docker wait v1-id" will return error code and the systemd service will exit with error code, due to the restart policy, the systemd service will restart, and the failed number will +1, but the failed number limit is only three within 20 minutes, it means if we do three times upgrade or fallback, the systemd service will be never up. So, we need to reset-failed number before we do systemd restart.

src/sonic-ctrmgrd/ctrmgr/ctrmgrd.py
```diff
@@ -563,6 +564,12 @@ def on_state_update(self, key, op, data):
 
             log_debug("try to tag latest label after {} seconds @{}".format(
                     remote_ctr_config[TAG_IMAGE_LATEST], start_time))
+        
+        # This is for going back to local without waiting the sytemd restart time
```
qiluo-msft, line 568: >sytemd [](http://example.com/codeflow?start=62&length=6)

typo

lguohan:
> k8s will stop v1 container first, this time the systemd service is doing "docker wait v1-id", once the v1 container stops, the "docker wait v1-id" will return error code and the systemd service will exit with error code,

in this case, it is planned stop, so why docker wait will return error code? can we make this like planned stopped with error code = 0?

lixiaoyuner:
> in this case, it is planned stop, so why docker wait will return error code? can we make this like planned stopped with error code = 0?

The reason is that k8s will kill the container, so the docker wait result is not zero. There is one way that we can check whether the wait id is the feature name or not. Feature name means it's a local container. Not feature name means it's a kube container, if it's a kube container, after docker wait returns, we could not care the docker wait returns code and we can return 0 directly, this maybe a solution. I can try to implement to verify it's a feasible solution.

lguohan (approved):

PR #15440, opened 2023-06-13
Updated LLDP startup configuration for multi-asic

abdosi:
What I did:
Changes in LLDP Configuration to handle this:

1.  **Enable LLDP only on front panel interfaces.** 

Why I did:
`This is done to free up CPU cycle as we don't see use for that in multi-asic and also for Chassis we have disable lldp docker on supervisor so there is no lldp peering even if send LLDP BPDU on backplane ports`

2.  **For multi-asic do not generate lldpd.conf with eth0 related configuration**

Why I did:
`In namespace lldp docker there is no need to generate eth0 configuration. It gives not needed errror message as seen here https://github.com/sonic-net/sonic-buildimage/issues/15382`

How I did it:
Instead of relying on `DEVICE_METADTA` `sub_role` we are using `NAMESPACE_ID` environmental variable (passed to all docker instance) which will be > 0 for multiasic docker instance and not define for global/host docker instance.

How I verify:
Manual Verification on multi-asic Chassis system.


abdosi:
cc @anamehra for viz.

abdosi:
Might fix: https://github.com/sonic-net/sonic-buildimage/issues/15382

arlakshm:
@abdosi please help fix the test failures

mlok-nokia (approved):

judyjoseph (approved):
LGTM

PR #15472, opened 2023-06-15
Fix docker-sonic-mgmt build by specific version install of pyaml and azure-cli by apt-get

shbalaku-microsoft:
#### Why I did it
Current docker-sonic-mgmt build is broken. So below are two fixes which can help in mitigating the same.

1. PYAML - Download a specific version in python2 as after https://pypi.org/project/pyaml/23.5.5/ there was support only for python3. This update happened on May 5th. And consequently all daily builds after this changes https://dev.azure.com/mssonic/build/_build/results?buildId=266733&view=results (starting build to break) kept failing
2. Azure-CLI - this can be downloaded by apt-get repository. So modify as an improvement.

##### Work item tracking
- Microsoft ADO **(number only)**: #15567

#### How I did it
By manually checking the release notes of pyaml and install azure-cli in newly installed docker container using apt-get

#### How to verify it
You can run below commands to validate:

make configure PLATFORM=generic
make target/docker-sonic-mgmt.gz

Second line would fail without the commit.

### DIFF
--- a/dockers/docker-sonic-mgmt/Dockerfile.j2
+++ b/dockers/docker-sonic-mgmt/Dockerfile.j2
@@ -5,6 +5,7 @@ ENV DEBIAN_FRONTEND=noninteractive
 
 RUN apt-get update && apt-get install -y apt-transport-https \
                                          apt-utils \
+                                         azure-cli \
                                          build-essential \
                                          ca-certificates \
                                          cmake \
@@ -75,6 +76,7 @@ RUN pip install allure-pytest==2.8.22 \
                 pexpect \
                 prettytable \
                 psutil \
+                pyaml==21.10.1 \
                 pyasn1==0.1.9 \
                 pycryptodome==3.9.8 \
                 pyfiglet \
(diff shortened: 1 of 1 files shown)


dockers/docker-sonic-mgmt/Dockerfile.j2
```diff
@@ -5,6 +5,7 @@ ENV DEBIAN_FRONTEND=noninteractive
 
 RUN apt-get update && apt-get install -y apt-transport-https \
                                          apt-utils \
+                                         azure-cli \
```
wangxin, line 8: @shbalaku-microsoft Version of the azure-cli package installed from apt is much lower than the package installed by one command. 
`2.0.81` vs `2.49.0`

It is better to keep the current way of installing `azure-cli`.

wangxin: I see, this is to address the build issue https://github.com/sonic-net/sonic-buildimage/issues/15567

wangxin (approved):

aaronpayment:
Can this be backported to 202305 branch?

PR #15487, opened 2023-06-15
[SNMP][IPv6]: Fix SNMP IPv6 reachability issue in certain scenarios

SuvarnaMeenakshi:
Modify snmpd.conf to start snmpd to listen on specific management and loopback ips instead of listening on any ip.

#### Why I did it
SNMP over IPv6 is not working for all scenarios for a single asic platforms.
The expectation is that SNMP query over IPv6 should work over Management or Loopback0 addresses.
**Specific scenario where this issue is seen**
In case of Lab T0 device,  when SNMP request is sent from a directly connected T1 neighbor over Loopback IP, SNMP response was not received.
This was because the SRC IP address in SNMP response was not Loopback IP, it was the PortChannel IP connected to the neighboring device.
```
23:18:51.620897  In 22:26:27:e6:e0:07 ethertype IPv6 (0x86dd), length 105: fc00::72.41725 > **fc00:1::32**.161:  C="msft" **GetRequest**(28)  .1.3.6.1.2.1.1.1.0
23:18:51.621441 Out 28:99:3a:a0:97:30 ethertype IPv6 (0x86dd), length 241: **fc00::71**.161 > fc00::72.41725:  C="msft" **GetResponse**(162)  .1.3.6.1.2.1.1.1.0="SONiC Software Version: SONiC.xxx - HwSku: xx - Distribution: Debian 10.13 - Kernel: 4.19.0-12-2-amd64"
```
In case of IPv4, the SRC IP in SNMP response was correctly set to Loopback IP.
```
23:25:32.769712  In 22:26:27:e6:e0:07 ethertype IPv4 (0x0800), length 85: 10.0.0.57.56701 > **10.1.0.32**.161:  C="msft" **GetRequest**(28)  .1.3.6.1.2.1.1.1.0
23:25:32.975967 Out 28:99:3a:a0:97:30 ethertype IPv4 (0x0800), length 221: **10.1.0.32**.161 > 10.0.0.57.56701:  C="msft" **GetResponse**(162)  .1.3.6.1.2.1.1.1.0="SONiC Software Version: SONiC.xxx - HwSku: xx - Distribution: Debian 10.13 - Kernel: 4.19.0-12-2-amd64"
```

**Sequence of SNMP request and response**
1. SNMP request will be sent with SRC IP fc00::72 DST IP fc00:1::32
2. SNMP request is received at SONiC device is sent to snmpd which is listening on port 161 :::161/
3. snmpd process will parse the request create a response and sent to DST IP fc00::72. 
snmpd process does not track the DST IP on which the SNMP request was received, which in this case is Loopback IP.
snmpd process will only keep track what is tht IP to which the response should be sent to.
4. snmpd process will send the response packet.
5. Kernel will do a route look up on destination IP and find the best path.
ip -6 route get fc00::72
fc00::72 from :: dev PortChannel101 proto kernel src fc00::71 metric 256 pref medium
5. Using the "src" ip from about, the response is sent out. This SRC ip is that of the PortChannel and not the device Loopback IP.

The same issue is seen when SNMP query is sent from a remote server over Management IP.
SONiC device eth0 --------- Remote server
SNMP request comes with SRC IP <Remote_server> DST IP <Mgmt IP>
If kernel finds best route to Remote_server_IP is via BGP neighbors, then it will send the response via front-panel interface with SRC IP as Loopback IP instead of Management IP.

Main issue is that in case of IPv6, snmpd ignores the IP address to which SNMP request was sent, in case of IPv6.
In case of IPv4, snmpd keeps track of DST IP of SNMP request, it will keep track if the SNMP request was sent to mgmt IP or Loopback IP.
Later, this IP is used in ipi_spec_dst as SRC IP which helps kernel to find the route based on DST IP using the right SRC IP.
https://github.com/net-snmp/net-snmp/blob/master/snmplib/transports/snmpUDPBaseDomain.c#L300 
ipi.ipi_spec_dst.s_addr = srcip->s_addr
Reference: https://man7.org/linux/man-pages/man7/ip.7.html
```
If IP_PKTINFO is passed to sendmsg(2)
              and ipi_spec_dst is not zero, then it is used as the local
              source address for the routing table lookup and for
              setting up IP source route options.  When ipi_ifindex is
              not zero, the primary local address of the interface
              specified by the index overwrites ipi_spec_dst for the
              routing table lookup.
```

**This issue is not seen on multi-asic platform, why?**
on multi-asic platform, there exists different network namespaces.
SNMP docker with snmpd process runs on host namespace.
Management interface belongs to host namespace.
Loopback0 is configured on asic namespaces.
Additional inforamtion on how the packet coming over Loopback IP reaches snmpd process running on host namespace: https://github.com/sonic-net/sonic-buildimage/pull/5420
Because of this separation of network namespaces, the route lookup of destination IP is confined to routing table of specific namespace where packet is received.
if packet is received over management interface, SNMP response also is sent out of management interface. Same goes with packet received over Loopback Ip.

##### Work item tracking
- Microsoft ADO **17537063**:

#### How I did it
Have snmpd listen on specific Management and Loopback IPs specifically instead of listening on any IP for single-asic platform.

Before Fix
```
admin@xx:~$ sudo netstat -tulnp | grep 161   
udp        0      0 0.0.0.0:161             0.0.0.0:*                           15631/snmpd         
udp6       0      0 :::161                  :::*                                15631/snmpd  
```
After fix
```
admin@device:~$ sudo netstat -tulnp | grep 161
udp        0      0 10.1.0.32:161           0.0.0.0:*                           215899/snmpd        
udp        0      0 10.3.1.1:161             0.0.0.0:*                           215899/snmpd        
udp6       0      0 fc00:1::32:161          :::*                                215899/snmpd        
udp6       0      0 fc00:2::32:161          :::*                                215899/snmpd  
``` 

**How this change helps with the issue?**
To see snmpd trace logs, modify snmpd to start using the below parameters, in supervisord.conf file
```
/usr/sbin/snmpd -f -LS0-7i -Lf /var/log/snmpd.log
```
When snmpd listens on any IP, snmpd binds to IPv4 and IPv6 sockets as below:
```
netsnmp_udpbase: binding socket: 7 to UDP: [0.0.0.0]:0->[0.0.0.0]:161
trace: netsnmp_udp6_transport_bind(): transports/snmpUDPIPv6Domain.c, 303:
netsnmp_udpbase: binding socket: 8 to UDP/IPv6: [::]:161
```

When IPv4 response is sent, it goes out of fd 7 and IPv6 response goes out of fd 8.
When IPv6 response is sent, it does not have the right SRC IP and it can lead to the issue described.

When snmpd listens on specific Loopback/Management IPs, snmpd binds to different sockets:
```
trace: netsnmp_udpipv4base_transport_bind(): transports/snmpUDPIPv4BaseDomain.c, 207:
netsnmp_udpbase: binding socket: 7 to UDP: [0.0.0.0]:0->[10.250.0.101]:161
trace: netsnmp_udpipv4base_transport_bind(): transports/snmpUDPIPv4BaseDomain.c, 207:
netsnmp_udpbase: binding socket: 8 to UDP: [0.0.0.0]:0->[10.1.0.32]:161
trace: netsnmp_register_agent_nsap(): snmp_agent.c, 1261:
netsnmp_register_agent_nsap: fd 8
netsnmp_udpbase: binding socket: 10 to UDP/IPv6: [fc00:1::32]:161
trace: netsnmp_register_agent_nsap(): snmp_agent.c, 1261:
netsnmp_register_agent_nsap: fd 10
netsnmp_ipv6: fmtaddr: t = (nil), data = 0x7fffed4c85d0, len = 28
trace: netsnmp_udp6_transport_bind(): transports/snmpUDPIPv6Domain.c, 303:
netsnmp_udpbase: binding socket: 9 to UDP/IPv6: [fc00:2::32]:161
```
When SNMP request comes in via Loopback IPv4, SNMP response is sent out of fd 8
```
trace: netsnmp_udpbase_send(): transports/snmpUDPBaseDomain.c, 511:
netsnmp_udp: send 170 bytes from 0x5581f2fbe30a to UDP: [10.0.0.33]:46089->[10.1.0.32]:161 on fd 8
```
When SNMP request comes in via Loopback IPv6, SNMP response is sent out of fd 10
```
netsnmp_ipv6: fmtaddr: t = (nil), data = 0x5581f2fc2ff0, len = 28
trace: netsnmp_udp6_send(): transports/snmpUDPIPv6Domain.c, 164:
netsnmp_udp6: send 170 bytes from 0x5581f2fbe30a to UDP/IPv6: [fc00::42]:43750 on fd 10
```

#### How to verify it
Verified on single asic and multi-asic devices.
Single asic SNMP query with Loopback 
```
ARISTA01T1#bash snmpget -v2c -c xxx 10.1.0.32 1.3.6.1.2.1.1.1.0
SNMPv2-MIB::sysDescr.0 = STRING: SONiC Software Version: SONiC.xx - HwSku: Arista-7260xx - Distribution: Debian 10.13 - Kernel: 4.19.0-12-2-amd64
ARISTA01T1#bash snmpget -v2c -c xxx fc00:1::32 1.3.6.1.2.1.1.1.0
SNMPv2-MIB::sysDescr.0 = STRING: SONiC Software Version: SONiC.xx - HwSku: Arista-7260xxx - Distribution: Debian 10.13 - Kernel: 4.19.0-12-2-amd64
```

On multi-asic -- no change.
```
sudo netstat -tulnp | grep 161
udp        0      0 0.0.0.0:161             0.0.0.0:*                           17978/snmpd         
udp6       0      0 :::161                  :::*                                17978/snmpd 
```
Query result using Loopback IP from a directly connected BGP neighbor
```
ARISTA01T2#bash snmpget -v2c -c xxx 10.1.0.32 1.3.6.1.2.1.1.1.0
SNMPv2-MIB::sysDescr.0 = STRING: SONiC Software Version: SONiC.xx - HwSku: xx - Distribution: Debian 9.13 - Kernel: 4.9.0-14-2-amd64
ARISTA01T2#bash snmpget -v2c -c xxx fc00:1::32 1.3.6.1.2.1.1.1.0
SNMPv2-MIB::sysDescr.0 = STRING: SONiC Software Version: SONiC.xx - HwSku: xx - Distribution: Debian 9.13 - Kernel: 4.9.0-14-2-amd64  
```

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211

- [ ] 
- [ ]


dockers/docker-snmp/snmpd.conf.j2
```diff
@@ -13,12 +13,38 @@
...
-#  Listen for connections on all ip addresses, including eth0, ipv4 lo
+#  Listen for connections on ip addresses based on configuration
 #
+# if SNMP_AGENT_ADDRESS_CONFIG is present
+# use SNMP_AGENT_ADDRESS_CONFIG
+# elif either MGMT_INTERFACE or LOOPBACK_INTERFACE
+# use management and loopback ip addresses
+# use docker0 ipv4 and ipv6 address
+# else if none of the above config is present
+# listen on any IP
```
qiluo-msft, line 24: The logic is complex.
how about minigraph.py just generated correct SNMP_AGENT_ADDRESS_CONFIG?
1. single-asic, bind to mgmt and lo
2. multi-asic bind to docker0

dockers/docker-snmp/start.sh
```diff
@@ -16,11 +16,16 @@ mkdir -p /etc/ssw /etc/snmp
 # Parse snmp.yml and insert the data in Config DB
 /usr/bin/snmp_yml_to_configdb.py
 
+DOCKER0_V4=$(ip -4 addr show docker0 | grep -oP '(?<=inet\s)\d+(\.\d+){3}')
```
qiluo-msft, line 19: >docker0 [](http://example.com/codeflow?start=29&length=7)

More secure container will bring better isolation between host and container. There is no access to docker0 insider container in future.

You may consider passing env var while `docker create`.

dockers/docker-snmp/snmpd.conf.j2
qiluo-msft: I think docker0 IP is not part of user config, you can use template here.

However, mgmt IP, lo IP are from user config, it is natural to add them into ConfigDB SNMP_AGENT_ADDRESS_CONFIG table.

zhangyanzhao:
@ganglyu  @arlakshm @abdosi  can you please help to review and approve this PR? Thanks.

ganglyu (approved):

src/sonic-config-engine/minigraph.py
```diff
@@ -1700,6 +1700,17 @@ def parse_xml(filename, platform=None, port_config_file=None, asic_name=None, hw
             results['LOOPBACK_INTERFACE'][host_lo_intf[0]] = {}
 
     results['MGMT_VRF_CONFIG'] = mvrf
+    
+    # Set SNMP_AGENT_ADDRESS_CONFIG to Management IP and Loopback0 IP 
+    if asic_name is None:
+        results['SNMP_AGENT_ADDRESS_CONFIG'] = {}
+        for mgmt_if in results['MGMT_INTERFACE'].keys():
+            snmp_key = mgmt_if[1].split('/')[0] + '|161|'
```
qiluo-msft, line 1708: >1 [](http://example.com/codeflow?start=31&length=1)

You could not assume that index 1 must be a special one. What is your criteria? Please check each mgmt_if.

src/sonic-config-engine/minigraph.py
```diff
@@ -1700,6 +1700,17 @@ def parse_xml(filename, platform=None, port_config_file=None, asic_name=None, hw
...
+            results['SNMP_AGENT_ADDRESS_CONFIG'][snmp_key] = {}
+        for loip in results['LOOPBACK_INTERFACE']:
+            if len(loip) == 2 and loip[0] == 'Loopback0':
+                snmp_key = loip[1].split('/')[0] + '|161|'
```
qiluo-msft, line 1712: >1 [](http://example.com/codeflow?start=32&length=1)

The same,  do no assume the order.

src/sonic-config-engine/minigraph.py
SuvarnaMeenakshi: The MGMT_INTERFACE keys contains interface name and ip address.
For example: ('eth0', '10.250.0.101/24')
So used index 1 to get the IP address.

qiluo-msft: oh, you mean 1 is the index of tuple. Could you prevent using magic number? how about tuple assignment or define a constant for 1.

dockers/docker-snmp/snmpd.conf.j2
qiluo-msft: I see the challenge of determine single/multi-asic in minigraph.py. So it is okay to keep complex code here.

SuvarnaMeenakshi: I have modified the snmpd.conf.j2 template to use management and loopback IP for single asic platform.
With this, we could backport to previous branches to ensure that we could use ipv6 .

Next step is to:
1. If the change to modify to use bridge network for snmp docker is merged in, we could add NAT rule for single asic platform similar to multi-asic platform.
2. After that change, will update minigraph parser to update SNMP_AGENT_ADDRESS_CONFIG table with management IP and docker0 IP for both single and multi-asic platform that way minigraph parser need not have the logic to differentiate between single and multi-asic platform.
3. Remove the changes in snmpd.conf.j2 and directly use SNMP_AGENT_ADDRESS_CONFIG which will be updated by minigraph parser.

dockers/docker-snmp/snmpd.conf.j2
```diff
@@ -13,12 +13,36 @@
...
+{{ 'udp' }}
+{%- endif -%}
+{% endmacro %}
+
 {% if SNMP_AGENT_ADDRESS_CONFIG %}
 {% for (agentip, port, vrf) in SNMP_AGENT_ADDRESS_CONFIG %}
-agentAddress {{ agentip }}{% if port %}:{{ port }}{% endif %}{% if vrf %}%{{ vrf }}{% endif %}{{ "" }}
+agentAddress {{ protocol(agentip) }}:[{{ agentip }}]{% if port %}:{{ port }}{% endif %}{% if vrf %}%{{ vrf }}{% endif %}{{ "" }}
+{% endfor %}
+{% elif NAMESPACE_COUNT is not defined or NAMESPACE_COUNT|int <= 1 %}
```
abdosi, line 31: can we just check NAMESPACE_COUNT|int == 1 . Not sure how this cannot be define as we are passing explictly below . or value != 1

abdosi (approved):
LGTM

arlakshm (dismissed):

dockers/docker-snmp/snmpd.conf.j2
```diff
@@ -13,12 +13,36 @@
 #  AGENT BEHAVIOUR
 #
 
-#  Listen for connections on all ip addresses, including eth0, ipv4 lo
+# Listen for connections on all ip addresses, including eth0, ipv4 lo for multi-asic platform
+# Listen on managment and loopback0 ips for single asic platform
 #
+{% macro protocol(ip_addr) %}
+{%- if ':' in ip_addr -%}
```
qiluo-msft, line 20:
```suggestion
{%- if in ip_addr | ipv6 -%}
```

qiluo-msft (approved):

arlakshm (approved):

PR #15512, opened 2023-06-16
Use execl instead of popen in RADIUS NSS code to fix vulnerability.

shdasari:
#### Why I did it

https://github.com/sonic-net/sonic-buildimage/pull/15284 fixes a case of shell escape exploit for TACACS+. This applies to RADIUS as well. RADIUS creates an unconfirmed user locally on the switch while attempting authentication. popen() is used to execute useradd,usermod and userdel commands. This exposes a vulnerability where a tactically designed username (which could contain explicit linux commands) can lead to getting executed as root. 

An example of such a username could be  **"asd\";echo>remoteRCE2;#"**. This leads to remoteRCE2 getting created in "/".

##### Work item tracking

#### How I did it
All calls to popen() used to execute useradd, usermod and userdel are replaced with fork()/execl().

#### How to verify it

**Prior to the fix, following is the behavior:**

[s@i vm] ssh "asd\";echo>remoteRCE2;#"@1.1.1.1
asd";echo>remoteRCE2;#@1.1.1.1's password: 
Permission denied, please try again.

On the SONiC switch,

root@sonic:/# ls
accton_as7816_monitor.log  home		   lib64   **remoteRCE2**  sys
bin			   host		   libx32  root        tmp
boot			   initrd.img	   media   run	       usr
cache.tgz		   initrd.img.old  mnt	   sbin        var
dev			   lib		   opt	   sonic       vmlinuz
etc			   lib32	   proc    srv	       vmlinuz.old
root@sonic:/# ls -l

**With the fix:**

[s@i vm] ssh "asd\";echo>remoteRCE2;#"@1.1.1.1
asd";echo>remoteRCE2;#@1.1.1.1's password: 
Permission denied, please try again.

root@sonic:/# ls
accton_as7816_monitor.log  etc		   lib	   mnt	 sbin	usr
bin			   home		   lib32   opt	 sonic	var
boot			   host		   lib64   proc  srv	vmlinuz
cache.tgz		   initrd.img	   libx32  root  sys	vmlinuz.old
dev			   initrd.img.old  media   run	 tmp

Verified that RADIUS authentication works as expected for valid users as well.


a-barboza:
> #### Why I did it
> #15284 fixes a case of shell escape exploit for TACACS+. This applies to RADIUS as well. RADIUS creates an unconfirmed user locally on the switch while attempting authentication. popen() is used to execute useradd,usermod and userdel commands. This exposes a vulnerability where a tactically designed username (which could contain explicit linux commands) can lead to getting executed as root.
> 
> An example of such a username could be **"asd";echo>remoteRCE2;#"**. This leads to remoteRCE2 getting created in "/".
> 
> ##### Work item tracking
> * Microsoft ADO **(number only)**:
> 
> #### How I did it
> All calls to popen() used to execute useradd, usermod and userdel are replaced with fork()/execl().
> 
> #### How to verify it
> **Prior to the fix, following is the behavior:**
> 
> [s@i vm] ssh "asd";echo>remoteRCE2;#"@1.1.1.1 asd";echo>remoteRCE2;#@1.1.1.1's password: Permission denied, please try again.
> 
> On the SONiC switch,
> 
> root@sonic:/# ls accton_as7816_monitor.log home lib64 **remoteRCE2** sys bin host libx32 root tmp boot initrd.img media run usr cache.tgz initrd.img.old mnt sbin var dev lib opt sonic vmlinuz etc lib32 proc srv vmlinuz.old root@sonic:/# ls -l
> 
> **With the fix:**
> 
> [s@i vm] ssh "asd";echo>remoteRCE2;#"@1.1.1.1 asd";echo>remoteRCE2;#@1.1.1.1's password: Permission denied, please try again.
> 
> root@sonic:/# ls accton_as7816_monitor.log etc lib mnt sbin usr bin home lib32 opt sonic var boot host lib64 proc srv vmlinuz cache.tgz initrd.img libx32 root sys vmlinuz.old dev initrd.img.old media run tmp
> 
> Verified that RADIUS authentication works as expected for valid users as well.

@adyeung @seiferteric @shdasari 

The semgrep flagged failure appears to be a false +ve. The compiler cannot optimize that particular memset, since the target of the memset is modified inside the function, and returned from the function, and used else where.

I recommend merging this patch. There are some minor enhancements (replace integer constants with macros which could be done). But, I recommend merging this patch. Please let me know if anything else needs to be done from my side. thank you

a-barboza (approved):

lguohan:
can you check the semgrep failure?

src/radius/nss/libnss-radius/nss_radius_common.c
```diff
@@ -466,82 +552,54 @@ int radius_update_user(RADIUS_NSS_CONF_B * conf, const char * user, int mpl) {
...
+    char home[64] = {0};
+    snprintf(sgid, 10, "%d", rnm->gid);
+    snprintf(home, 63, "/home/%s", user);
 
-        snprintf(buf, sizeof(buf), "Unconfirmed-%ld", time(NULL));
-        written = snprintf(useradd, sizeof(useradd),
-            "%s -U -G %s -c \"%s\" -d \"/home/%s\" -m -s %s \"%s\"",
-            USERADD, rnm->groups, unconfirmed ? buf : user, user,
-            rnm->shell, user);
+    memset(buf, 0, sizeof(buf));
```
a-barboza, line 583: This memset seems redundant? snprintf() Man Page: "The functions snprintf() and vsnprintf() write at most size bytes (including the terminating null byte ('\0')) to str." semgrep appears to complain about it.

a-barboza (approved):
semgrep run passes.

a-barboza (approved):

adyeung:
@lguohan pls check if you are ok with the vulnerability fix

src/radius/nss/libnss-radius/nss_radius_common.c
```diff
@@ -167,6 +168,109 @@ static void init_rnm(RADIUS_NSS_CONF_B * conf) {
 
 }
 
+static int user_add(const char* cmd, const char* name, char* gid, char* sec_grp, char* gecos,
```
seiferteric, line 171: What is the purpose of accepting cmd as first argument if you have one function for each command (user_add, user_del, user_mod)?

src/radius/nss/libnss-radius/nss_radius_common.c
```diff
@@ -167,6 +168,109 @@ static void init_rnm(RADIUS_NSS_CONF_B * conf) {
...
+        if WIFEXITED(wstatus)
+            return WEXITSTATUS(wstatus);
+        else
+            return -1;
+
+    // Child
+
+    } else if(pid == 0) {
+
+        return execl(cmd, cmd, "-G", sec_grp, "-c", name, name, NULL);
```
seiferteric, line 263: Maybe we can add log in case of execl failure.

seiferteric (commented):
Added a couple minor comments, but otherwise looks good to me.

lguohan:
@adyeung , i see some comments, as soon as eric sign off, i can get it merged.

src/radius/nss/libnss-radius/nss_radius_common.c
```diff
@@ -167,6 +168,124 @@ static void init_rnm(RADIUS_NSS_CONF_B * conf) {
...
 }
 
+static int user_add(const char* name, char* gid, char* sec_grp, char* gecos, 
+                        char* home, char* shell, const char* unconfirmed_user, int many_to_one) {
+    pid_t pid, w;
+    int status = 0;
+    int wstatus;
+    char cmd[64];
+
+    snprintf(cmd, 63, "%s", USERADD);
```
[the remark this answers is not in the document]
shdasari: These changes have been tested and the user is added as expected. Note that the #define code is existing code and not added as a part of this code review. We can get rid of the defines and use the command directly however.

seiferteric (approved):

src/radius/nss/libnss-radius/nss_radius_common.c
```diff
@@ -466,82 +567,53 @@ int radius_update_user(RADIUS_NSS_CONF_B * conf, const char * user, int mpl) {
...
-        written = snprintf(useradd, sizeof(useradd),
-            "%s -g %d -G %s -c \"%s\" -m -s %s \"%s\"",
-            USERADD, rnm->gid, rnm->groups, rnm->gecos, rnm->shell, user);
+    char sgid[10] = {0};
+    char home[64] = {0};
+    snprintf(sgid, 10, "%d", rnm->gid);
+    snprintf(home, 63, "/home/%s", user);
 
-    } else {
+    snprintf(buf, sizeof(buf), "Unconfirmed-%ld", time(NULL));
```
Yarden-Z, line 598: Shouldn't we use sizeof(buf)-1 to leave a null termination place for the buf?

PR #15535, opened 2023-06-19
enable info log for graceful restart events

StormLiangMS:
#### Why I did it
Graceful restart is a key event for bgpd, related log print is debug level. To change it to info level to get more visibilities when this kind of event is triggered.

##### Work item tracking
- Microsoft ADO **(13875291)**:

#### How I did it
To create patch file to change from debug level to info level. 

#### How to verify it
To run PR test and capture the print.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012
- [ ] 202106
- [ ] 202111
- [x] 202205
- [ ] 202211
- [x] 202305

- [ ] 
- [ ]

### DIFF
--- a/src/sonic-frr/patch/0029-bgpd-Change-log-level-for-graceful-restart-events.patch
+++ b/src/sonic-frr/patch/0029-bgpd-Change-log-level-for-graceful-restart-events.patch
@@ -1,0 +1,122 @@
+From c423bce4db804c1d07d65ce3d06a9e62c4eceb2b Mon Sep 17 00:00:00 2001
+From: stormliang <stormliang@microsoft.com>
+Date: Mon, 19 Jun 2023 13:57:01 +0000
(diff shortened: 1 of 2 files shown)


gechiang (approved):
Looks good to me.
The only concern is whether any of these log will get triggered over and over again?
What I am asking is if whether these events is in some kind of a timer based loop where it will just keep printing?  if not, then I think it is a good change.

StormLiangMS:
> Looks good to me. The only concern is whether any of these log will get triggered over and over again? What I am asking is if whether these events is in some kind of a timer based loop where it will just keep printing? if not, then I think it is a good change.

@gechiang this log added for timer start and stop, which is to trace the GR procedure, it won't be always on for the timer.

StormLiangMS:
https://github.com/sonic-net/sonic-buildimage/pull/15863 cherry-pick PR to 202205

PR #15581, opened 2023-06-22
Add new BGP peer configuration for BGPSentinel

guangyao6:
#### Why I did it
For route registry service, in order to block hijacked routes, IBGP session needs to be set up from BGP sentinel service to SONiC, and BGP sentinel service advertise the same route with higher local-preference and no export community. So that SONiC takes the route from BGP sentinel as the best path and does not advertise the route to EBGP peers.
In order to do that, new route-maps are needed. So this change adds a new set of templates, keeping BGPSentinel peers out of the other templates.  
##### Work item tracking
- Microsoft ADO **(number only)**: 24300658

#### How I did it
1. Add sentinel_community in constants.yml, route from BGPSentinel do not match this community will be denied.
2. Add support to convert BGPSentinel related configuration in the BGPPeerPassive element of the minigraph to a new BGP_SENTINELS table in CONFIG_DB
3. Add a new set of "sentinels" templates to docker-fpm-frr
4. Add a new BGP peer manager to bgpcfgd, to add neighbors from the BGP_SENTINELS table using the "sentinels" templates 
5. Add a test case for minigraph.py, making sure the BGPSentinel and BGPSentinelV6 elements create BGP_SENTINELS DB entry.
6. Add a set of test cases for the new sentinels templates in sonic-bgpcfgd tests.
7. Add sonic-bgp-sentinel.yang and a set of testcases for the yang file.
#### How to verify it
1. Testcases newly added would pass.
2. Setup IPv4 and IPv6 BGPSentinel services in minigraph, and load minigraph, show CONFIG_DB and "show runningconfig bgp", configuration would be loaded successfully.
3.  Using t1-lag topo and setup IBGP session from BGPSentinel to SONiC loopback address, IBGP session would up.
4. Advertise route from BGPSentinel to T1 with sentinel_community, higher local-preference and no-export communiyt. In T1, show bgp route, the result is "Not advertise to any EBGP peer".
5. Withdraw the route in BGPSentinel, in T1, route would advertise to EBGP peers.
6. Advertise route from T1 that does not match sentinel_community, in T1, would not see the route in show bgp route.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [x] 202205
- [x] 202211
- [x] 202305

- [x] Locally build 202205 image based on 036b8d131593b57f2eb1fa9d4b8969bde7cc3333
- [ ]


dockers/docker-fpm-frr/frr/bgpd/templates/sentinels/policies.conf.j2
```diff
@@ -0,0 +1,18 @@
+!
+! template: bgpd/templates/sentinels/policies.conf.j2
+!
+{% if constants.bgp.sentinel_community is defined %}
```
tiejunzhang-ms, line 4: is it possible sentinel_community not defined?

tiejunzhang-ms (approved):

dockers/docker-fpm-frr/frr/bgpd/templates/sentinels/policies.conf.j2
guangyao6: sentinel_community is defined in constants.yaml file, this item should exist as it is also firstly added in this commit, add this for a safer check.

PR #15587, opened 2023-06-22
[YANG][SNMP]: Add YANG model for SNMP_AGENT_ADDRESS_CONFIG table

SuvarnaMeenakshi:
#### Why I did it
https://github.com/sonic-net/sonic-utilities/pull/472 Added SNMP_AGENT_ADDRESS_CONFIG table in config db.
This PR is to add corresponding YANG model for that table.
##### Work item tracking

#### How I did it
Added YANG modesl for SNMP_AGENT_ADDRESS_CONFIG.

keys: agent_ip, port number, vrf.
CLI implementaion checks if agent_ip, port number already exists in CONFIG_DB table, if it does, then new entry is not added.
So added another condition to ensure combination of agent_ip and port is unique.
Below is an example of how data looks like in DB:
```
127.0.0.1:6379[4]> HGETALL  "SNMP_AGENT_ADDRESS_CONFIG|10.1.1.1|161|foo"
1) "NULL"
2) "NULL"
127.0.0.1:6379[4]> HGETALL "SNMP_AGENT_ADDRESS_CONFIG|10.1.0.32|161|"
1) "NULL"
2) "NULL"
```
#### How to verify it
Added unit-test for various combinations and ensures that it passes.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/src/sonic-yang-models/yang-models/sonic-snmp.yang
+++ b/src/sonic-yang-models/yang-models/sonic-snmp.yang
@@ -3,6 +3,12 @@ module sonic-snmp {
   prefix ssnmp;
   yang-version 1.1;
 
+  import ietf-inet-types {
+      prefix inet;
+  }
+  import sonic-vrf {
+    prefix vrf;
+  }
 
   organization
     "SONiC";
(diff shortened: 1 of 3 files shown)


src/sonic-yang-models/yang-models/sonic-snmp.yang
```diff
@@ -156,5 +162,35 @@ module sonic-snmp {
...
+          description "SNMP agent listening IP";
+        }
+        leaf port {
+          type inet:port-number;
+          description "SNMP agent listening port number";
+        }
+        leaf vrf_name {
+          type union {
+            type string {
+              pattern '';
```
qiluo-msft, line 182: >pattern [](http://example.com/codeflow?start=14&length=7)

@ganglyu @wen587  I find empty string is weird. Is it going to impact gnmi path, impact json representation of ConfigDB?

SuvarnaMeenakshi: The reason for empty string is because config CLI adds key in CONFIG_TABLE with empty string for vrf if it is not specified in the CLI argument.
https://github.com/sonic-net/sonic-utilities/blob/49fc3896a0ef907ff337467a3e04d2501ef850fd/config/main.py#L3084

ganglyu: GNMI path can support it, need GCU to confirm.

wen587: GCU can support it.

qiluo-msft (approved):

isabelmsft (approved):

PR #15606, opened 2023-06-23
Add function to retrieve docker0 ipv4 and ipv6 addresses

SuvarnaMeenakshi:
#### Why I did it
https://github.com/sonic-net/sonic-buildimage/pull/15487 this PR modifies snmpd to listen on specific IPs.
By default, snmpd will listen on management and loopback0 ipv4 and ipv6 addresses for single asic platform.
snmpd will listen will listen on management and docker0 ip address for multi asic platforn.
This PR is to add function to retrieve docker0 ip addresses for multi-asic platform, which could be used generically.
##### Work item tracking

#### How I did it
Add a new function get_docker0_ip() in sonic_py_common to get docker0 ipv4 and ipv6 address on multi-asic platform.

#### How to verify it
on multi-asic platform
```
>>> from sonic_py_common.multi_asic import get_docker0_ip
>>> get_docker0_ip()
('240.127.1.1', 'fd00::1')
>>> 
```
on single-asic platform
```
>>> from sonic_py_common.multi_asic import get_docker0_ip
>>> get_docker0_ip()
(None, None)
```

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/src/sonic-py-common/sonic_py_common/multi_asic.py
+++ b/src/sonic-py-common/sonic_py_common/multi_asic.py
@@ -1,6 +1,7 @@
 import glob
 import os
 import subprocess
+import netifaces
 
 from natsort import natsorted
 from swsscommon import swsscommon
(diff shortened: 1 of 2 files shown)


src/sonic-py-common/sonic_py_common/multi_asic.py
```diff
@@ -484,3 +485,26 @@ def get_asic_presence_list():
                     # asic is asid id: asic0, asic1.... asicN. Get the numeric value.
                     asics_list.append(int(get_asic_id_from_name(asic)))
     return asics_list
+
+def get_docker0_ip():
```
qiluo-msft, line 489: >get_docker0_ip [](http://example.com/codeflow?start=4&length=14)

get_docker0_ips

src/sonic-py-common/sonic_py_common/multi_asic.py
```diff
@@ -484,3 +485,26 @@ def get_asic_presence_list():
...
+    @summary: This function will return docker0 ipv4 and ipv6 addresses
+              for multi_asic platform.
+              This function will return None for single asic platform.
+    @return: tuple containing docker0 ipv4 and ipv6 address
+    """
+    docker0_v4 = None
+    docker0_v6 = None
+
+    # return (None, None) for single asic platform
+    if not is_multi_asic():
```
qiluo-msft, line 500: >is_multi_asic [](http://example.com/codeflow?start=11&length=13)

Is it harmful to let single-asic return the proper ips?

SuvarnaMeenakshi: This helper function is added so that we could use this to get docker0 ip for multi-asic platform.
Most likely we will need docker0 information only for multi-asic platform.
For example, we want to update SNMP_AGENT_ADDRESS_CONFIG table in config_db with docker0 and mgmt IP in miniraph parser and we could avoid adding the single/multi-asic in minigraph parser if we ensure that required IP is passed from this helped function.

SuvarnaMeenakshi:
Do not plan to use this change.

PR #15633, opened 2023-06-27
Don't build libhiredis anymore

saiarcot895:
#### Why I did it

We're not adding any patch on top of hiredis, and there's no apparent reason to build this. Remove the build step here, and just install the package from the Debian repos.

##### Work item tracking
- Microsoft ADO **(number only)**: 24381590

#### How to verify it

Build the SONiC image, and load it. Verify that services come up.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [x] 202305

- [ ] master branch
- [ ] 20230531.02


sonic-slave-bullseye/Dockerfile.j2
```diff
@@ -406,6 +406,8 @@ RUN apt-get update && apt-get install -y \
         libgirepository1.0-dev \
         libsystemd-dev \
         pkg-config \
+# For sonic-swss-common build
+        libhiredis-dev \
```
qiluo-msft, line 410: >libhiredis [](http://example.com/codeflow?start=8&length=10)

also change sonic-slave-buster/Dockerfile.j2?

saiarcot895: Added.

qiluo-msft (approved):

PR #15660, opened 2023-06-29
[Build] Fix some of the patches not applied issue

xumia:
#### Why I did it
Fix some of the patches in .patches folder not applied issue.
The command "quilt applied" only lists the applied patches, if some of the patches have issues, then the patches will not be applied when you run the build command again.

##### Work item tracking
- Microsoft ADO **(number only)**: 24410730

#### How I did it
1. Run the command to apply the patches without any conditions.
2. If failed, check if the failure reason is "series fully applied".

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012
- [x] 202106
- [x] 202111
- [x] 202205
- [x] 202211
- [x] 202305

- [ ] 
- [ ]

### DIFF
--- a/slave.mk
+++ b/slave.mk
@@ -648,7 +648,7 @@ $(addprefix $(FILES_PATH)/, $(SONIC_MAKE_FILES)) : $(FILES_PATH)/% : .platform $
 		# Remove target to force rebuild
 		rm -f $(addprefix $(FILES_PATH)/, $*)
 		# Apply series of patches if exist
-		if [ -f $($*_SRC_PATH).patch/series ]; then pushd $($*_SRC_PATH) && QUILT_PATCHES=../$(notdir $($*_SRC_PATH)).patch quilt push -a || quilt push -a 2>&1 | grep -q "series fully applied"; popd; fi $(LOG)
+		if [ -f $($*_SRC_PATH).patch/series ]; then pushd $($*_SRC_PATH) && ( quilt pop -a -f 1>/dev/null 2>&1 || true ) && QUILT_PATCHES=../$(notdir $($*_SRC_PATH)).patch quilt push -a; popd; fi $(LOG)
 		# Build project and take package
 		make DEST=$(shell pwd)/$(FILES_PATH) -C $($*_SRC_PATH) $(shell pwd)/$(FILES_PATH)/$* $(LOG)
 		# Clean up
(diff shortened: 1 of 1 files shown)


slave.mk
```diff
@@ -648,7 +648,7 @@ $(addprefix $(FILES_PATH)/, $(SONIC_MAKE_FILES)) : $(FILES_PATH)/% : .platform $
 		# Remove target to force rebuild
 		rm -f $(addprefix $(FILES_PATH)/, $*)
 		# Apply series of patches if exist
-		if [ -f $($*_SRC_PATH).patch/series ]; then pushd $($*_SRC_PATH) && quilt applied || QUILT_PATCHES=../$(notdir $($*_SRC_PATH)).patch quilt push -a; popd; fi $(LOG)
+		if [ -f $($*_SRC_PATH).patch/series ]; then pushd $($*_SRC_PATH) && QUILT_PATCHES=../$(notdir $($*_SRC_PATH)).patch quilt push -a || quilt push -a 2>&1 | grep -q "series fully applied"; popd; fi $(LOG)
```
qiluo-msft, line 651: >quilt push -a [](http://example.com/codeflow?start=135&length=13)

Why retry?
Why removing the env var?
Is it simply reverting [https://github.com/sonic-net/sonic-buildimage/pull/13386](https://github.com/sonic-net/sonic-buildimage/pull/13386) ?

qiluo-msft: Alternatively, we may check if any patch already applied, if yes, revert back, and apply all again.

xumia: The second one only checks whether the patch applied. It is to keep the output information (STDOUT/ERROUT) the same with the original one.
The second one does not need the env QUILT_PATCHES, because the .pc file has already been created by the first command, so the variable is not required.

xumia: > Alternatively, we may check if any patch already applied, if yes, revert back, and apply all again.

Do you think it is good below? Looks like to remove and apply again is not necesary, right?
```
if [ -f $($*_SRC_PATH).patch/series ]; then pushd $($*_SRC_PATH) && { quilt pop -a 1>/dev/null 2>&1 || ture } && QUILT_PATCHES=../$(notdir $($*_SRC_PATH)).patch quilt push -a; popd; fi $(LOG)
```

qiluo-msft: I prefer no retry. You last full command has a `pushd` which is not needed.

xumia: Qi, thanks for your comment, will change it to remove all patches before applying patches.

qiluo-msft (approved):

liuh-80:
@xumia , ADO branch tag missing, will approve 202012 branch after update ADO

PR #15685, opened 2023-06-30
Fix CONFIG_DB_INITIALIZED flag check logic and set/reset flag for warmboot

vaibhavhd:
MSFT ADO: 24274591

#### Why I did it

Two changes:
### 1  Fix a day1 issue, where check to wait until `CONFIG_DB_INITIALIZED` is incorrect.
There are multiple places where same incorrect logic is used.

Current logic (`until [[ $($SONIC_DB_CLI CONFIG_DB GET "CONFIG_DB_INITIALIZED") ]];`) will always result in pass, irrespective of the result of GET operation.
```
root@str2-7060cx-32s-29:~# sonic-db-cli CONFIG_DB GET "CONFIG_DB_INITIALIZED"
1
root@str2-7060cx-32s-29:~# until [[ $(sonic-db-cli CONFIG_DB GET "CONFIG_DB_INITIALIZED") ]]; do echo "entered here"; done
root@str2-7060cx-32s-29:~# 

root@str2-7060cx-32s-29:~# 
root@str2-7060cx-32s-29:~# sonic-db-cli CONFIG_DB GET "CONFIG_DB_INITIALIZED"                                             
0
root@str2-7060cx-32s-29:~# until [[ $(sonic-db-cli CONFIG_DB GET "CONFIG_DB_INITIALIZED") ]]; do echo "entered here"; done
root@str2-7060cx-32s-29:~# 
```

Fix this logic by checking for value of flag to be "1".
```
root@str2-7060cx-32s-29:~# until [[ $(sonic-db-cli CONFIG_DB GET "CONFIG_DB_INITIALIZED") -eq 1 ]]; do echo "entered here"; done
entered here
entered here
entered here
```

This gap in logic was highlighted when another fix was merged: https://github.com/sonic-net/sonic-buildimage/pull/14933
The issue being fixed here caused warmboot-finalizer to not wait until config-db is initialized.

### 2 Set and unset CONFIG_DB_INITIALIZED for warm-reboot case

Currently, during warm shutdown `CONFIG_DB_INITIALIZED`'s value is stored in redis db backup. This is restored back when the dump is loaded during warm-recovery.
So the value of `CONFIG_DB_INITIALIZED` does not depend on config db's state, however it remain what it was before reboot.

Fix this by setting `CONFIG_DB_INITIALIZED` to 0 as when the DB is loaded, and set it to 1 after db_migrator is done.

##### Work item tracking

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]


vaibhavhd:
This PR needs to be merged after https://github.com/sonic-net/sonic-buildimage/pull/15684

yxieca (dismissed):

files/build_templates/docker_image_ctl.j2
```diff
@@ -248,28 +249,18 @@ function postStartAction()
                     $SONIC_CFGGEN -j /etc/sonic/config_db$DEV.json --write-to-db
                 fi
             fi
-
-            if [[ "$BOOT_TYPE" == "fast" ]]; then
```
vaibhavhd, line 252: This code section is being removed as it will never get called - the check for `fast` reboot will enter the prev parent if block and not this.

Moreoever, w/ new changes in fast-boot, we have a dedicated flag for fastboot that gets cleared as part of finalizer. The timed expiry mechanism is not needed.

files/build_templates/docker_image_ctl.j2
```diff
@@ -248,28 +249,18 @@ function postStartAction()
...
-            # Call db_migrator after config-setup loads the config (from old config or minigraph)
-            echo "Delaying db_migrator until config migration is over"
+            # for warmboot case, DB is loaded but migration is still pending
+            # For firstbboot/fast/cold reboot case, DB contains nothing at this point
+            # unset CONFIG_DB_INITIALIZED to indicate pending config load and migration
+            # This flag will be set to "1" after DB migration/initialization is completed as part of config-setup
+            $SONIC_DB_CLI CONFIG_DB SET "CONFIG_DB_INITIALIZED" "0"
         else
-            # this is not a first time boot to a new image. Datbase container starts w/ old pre-existing config
-            if [[ -x /usr/local/bin/db_migrator.py ]]; then
```
vaibhavhd, line 267: Note for reviwer:

I don't think migration is needed for same-image reboot.

For case when device reboots to a different image, but this is not first-time boot to this image:
Cold boot: Migration is not needed for cold reboot, as config_db file in target image already has correct version.

Warm/fast boot: Migration is needed, and old config dump will have old version. But, do we want to support this path, as ideally we expect that the upgrades are always first-time-boot. If this expectation is not true then I can enable the call here for warmboot case.

yxieca (approved):

StormLiangMS:
@vaibhavhd cherry pick to 202305 failed.

files/build_templates/docker_image_ctl.j2
arfeigin: Just saw this, this is actually crucial for upgrading from older versions prior fast-reboot enhancements.
For example, upgrading from 202012 images, in case it exceeded 180 seconds no indication of fast-reboot is found in STATE-DB and db-migrator will fail to set fast-reboot new notation.

vaibhavhd:
> @vaibhavhd PR conflicts with 202305 branch

I did not see any conflicts while cherrypicking. Anyway, here is the cherrypick PR: https://github.com/sonic-net/sonic-buildimage/pull/16217

vaibhavhd:
> @vaibhavhd cherry pick to 202305 failed.

Here is the cheerypick PR: https://github.com/sonic-net/sonic-buildimage/pull/16217

vaibhavhd:
> @vaibhavhd PR conflicts with 202012 branch

Here is 202012 PR: https://github.com/sonic-net/sonic-buildimage/pull/16225

liuh-80:
202012 PR created: https://github.com/sonic-net/sonic-buildimage/pull/16225

files/build_templates/docker_image_ctl.j2
vaibhavhd: You are right. Thanks for reporting this. The PR was merged when you added comment, and I did not check it.
Fixed this here: https://github.com/sonic-net/sonic-buildimage/pull/16733

PR #15711, opened 2023-07-04
Fix issue: set has_timer attribute to true for platform monitor service

Junchao-Mellanox:
#### Why I did it

There is a redundant line in init_cfg.json.j2. It would cause pmon service always has "has_timer=False". However, we know that PMON has a timer now. So, I try to fix it here.

##### Work item tracking

#### How I did it

Remove the redundant line

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [x] 202211

- [ ] 
- [ ]

### DIFF
--- a/files/build_templates/pmon.timer
+++ b/files/build_templates/pmon.timer
@@ -1,0 +1,12 @@
+[Unit]
+# This delay is for fast/warm reboot performance
+Description=Delays pmon docker until SONiC has started
+PartOf=pmon.service
+
+[Timer]
+OnUnitActiveSec=0 sec
+OnBootSec=1min 30 sec
+Unit=pmon.service
+
+[Install]
+WantedBy=timers.target sonic.target sonic-delayed.target
--- a/files/build_templates/sonic_debian_extension.j2
+++ b/files/build_templates/sonic_debian_extension.j2
@@ -902,6 +902,9 @@ sudo LANG=C cp $SCRIPTS_DIR/write_standby.py $FILESYSTEM_ROOT/usr/local/bin/writ
 # Copy mark_dhcp_packet script
 sudo LANG=C cp $SCRIPTS_DIR/mark_dhcp_packet.py $FILESYSTEM_ROOT/usr/local/bin/mark_dhcp_packet.py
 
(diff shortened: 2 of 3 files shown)


keboliu (approved):

prgeor:
@Junchao-Mellanox can you elaborate why this change is needed? why PMON need to start late?

files/build_templates/pmon.timer
```diff
@@ -0,0 +1,12 @@
+[Unit]
```
prgeor, line 1: @Junchao-Mellanox This will affect the fast-reboot time for link bring up on platform where Xcvrd is responsible for link bring up.

files/build_templates/sonic_debian_extension.j2
```diff
@@ -902,6 +902,9 @@ sudo LANG=C cp $SCRIPTS_DIR/write_standby.py $FILESYSTEM_ROOT/usr/local/bin/writ
 # Copy mark_dhcp_packet script
 sudo LANG=C cp $SCRIPTS_DIR/mark_dhcp_packet.py $FILESYSTEM_ROOT/usr/local/bin/mark_dhcp_packet.py
 
+sudo cp $BUILD_TEMPLATES/pmon.timer $FILESYSTEM_ROOT_USR_LIB_SYSTEMD_SYSTEM
+echo "pmon.timer" | sudo tee -a $GENERATED_SERVICE_FILE
```
prgeor, line 906: @Junchao-Mellanox wont this delay the PMON launch for platforms that depend upon Xcvrd to initialize optics during cold reboot and there by affect link bring up time?

arfeigin:
> @Junchao-Mellanox can you elaborate why this change is needed? why PMON need to start late?

Hi, this change was already merged in the past: https://github.com/sonic-net/sonic-buildimage/pull/10567
As mentioned the motivation was that profiling the system state on init after fast-reboot during create_switch function execution, it is possible to see few python scripts running at the same time.

Bootchart is attached in the mentioned PR.

Junchao-Mellanox:
i will close this PR since there is another PR https://github.com/sonic-net/sonic-buildimage/pull/15334. In that PR, it makes pmon delay based on fast reboot status.

PR #15716, opened 2023-07-05
Add k8s master code new

lixiaoyuner:
#### Some context
This PR is same with [Previous PR](https://github.com/sonic-net/sonic-buildimage/pull/14870), the reason why I closed the previous PR is that I don't know why new commits can't appear in the previous PR. Open this new PR to fix that issue. Threre are two comments in the previous PR, I paste the two comments and reply in this PR's comments.

Figure it out, got a message from github, This is caused by an ongoing incident. The team is currently working on resolving it, the incident status page: https://www.githubstatus.com/incidents/pr0ptqcw6q8d
#### Why I did it
Currently, k8s master image is generated from a separate branch which we created by ourselves, not release ones. We need to commit these k8s master related code to master branch for a better way to do k8s master image build out.
##### Work item tracking
19998138
#### How I did it
- Install k8s dashboard docker images
- Install geneva mds and mdsd and fluentd docker images and tag them as latest, tagging latest will help create container always with the latest version
- Install azure-storage-blob and azure-identity, this will help do etcd backup and restore.
- Install kubernetes python client packages, this will help read worker and container state, we can send these metric to Geneva.
- Remove mdm debian package, will replace it with the mdm docker image
- Add k8s master entrance script, this script will be called by rc-local service when system startup. we have some master systemd services in compute-move repo, when VMM service create master VM, VMM will copy all master service files inside VM, the entrance script will setup all services according to the service files.
- When the entrance script content changed, the PR build will set include_kubernetes_master=y to help do validation for k8s master related code change. The default value of include_kubernetes_master should be always n for public master branch. We will generate master image from internal master branch
#### How to verify it
Build with INCLUDE_KUBERNETES_MASTER = y

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211

- [ ] 
- [ ]


files/build_templates/sonic_debian_extension.j2
```diff
@@ -791,11 +791,20 @@ sudo https_proxy=$https_proxy LANG=C chroot $FILESYSTEM_ROOT docker $SONIC_NATIV
...
+sudo https_proxy=$https_proxy LANG=C chroot $FILESYSTEM_ROOT docker $SONIC_NATIVE_DOCKERD_FOR_DOCKERFS pull linuxgeneva-microsoft.azurecr.io/distroless/genevamdsd:${MASTER_MDS_VERSION}
+sudo https_proxy=$https_proxy LANG=C chroot $FILESYSTEM_ROOT docker $SONIC_NATIVE_DOCKERD_FOR_DOCKERFS tag linuxgeneva-microsoft.azurecr.io/distroless/genevamdsd:${MASTER_MDS_VERSION} linuxgeneva-microsoft.azurecr.io/distroless/genevamdsd:latest
+sudo https_proxy=$https_proxy LANG=C chroot $FILESYSTEM_ROOT docker $SONIC_NATIVE_DOCKERD_FOR_DOCKERFS pull linuxgeneva-microsoft.azurecr.io/distroless/genevafluentd_td-agent:${MASTER_FLUENTD_VERSION}
+sudo https_proxy=$https_proxy LANG=C chroot $FILESYSTEM_ROOT docker $SONIC_NATIVE_DOCKERD_FOR_DOCKERFS tag linuxgeneva-microsoft.azurecr.io/distroless/genevafluentd_td-agent:${MASTER_FLUENTD_VERSION} linuxgeneva-microsoft.azurecr.io/distroless/genevafluentd_td-agent:latest
 echo "kubernetes master docker images pull complete"
 # Install python package for mdm service usage
```
qiluo-msft, line 803: The comment is outdated. Since you will not install mdm debian package, are psutil and statsd not needed?

files/image_config/kubernetes/kubernetes_master_entrance.sh
```diff
@@ -1,2 +1,23 @@
...
+
+set -x
+
+# mount disk from host
+mount_point="/from_host"
+disk="/dev/sdb1"
+mkdir -p $mount_point
+mount $disk $mount_point
+# check whether it is the first time to boot
+first_boot_flag_file="/from_host/first_boot_flag"
```
qiluo-msft, line 14: >first_boot_flag [](http://example.com/codeflow?start=33&length=15)

Is this a new feature?
Wondering if you can use existing sonic feature:
```
FIRST_BOOT_FILE="/host/image-${SONIC_VERSION}/platform/firsttime"
```

files/image_config/kubernetes/kubernetes_master_entrance.sh
```diff
@@ -1,2 +1,23 @@
+#!/bin/bash
+
 # This script is for kubernetes master image usage
 # Will mount kubernetes master disk and execute kubernetes entrance script
+
+set -x
+
+# mount disk from host
+mount_point="/from_host"
+disk="/dev/sdb1"
```
qiluo-msft, line 10: >sdb1 [](http://example.com/codeflow?start=11&length=4)

why hard-coded?

.azure-pipelines/template-skipvstest.yml
```diff
@@ -3,6 +3,13 @@ steps:
   - script: |
       set -ex
       tar_branch=origin/$(System.PullRequest.TargetBranch)
+      # Check if k8s master entrance script is changed
+      k8s_master_changed=$(git diff $tar_branch..HEAD --name-only | grep files/image_config/kubernetes/kubernetes_master_entrance.sh)
```
qiluo-msft, line 7: >files/image_config/kubernetes/kubernetes_master_entrance.sh [](http://example.com/codeflow?start=73&length=59)

this is a regex, so `.` has special meaning.

files/build_templates/sonic_debian_extension.j2
lixiaoyuner: > The comment is outdated. Since you will not install mdm debian package, are psutil and statsd not needed?

Yes, I install mdm docker to replace mdm debian package. And I still need to start mdm service to send metric, still need psutil and statsd. And I change the comments to "# Install python package for mdm metric collection service usage"

.azure-pipelines/template-skipvstest.yml
lixiaoyuner: Use grep -F to fix

files/image_config/kubernetes/kubernetes_master_entrance.sh
lixiaoyuner: We have a contract with VMM service, when VMM create master VM, it will attach a disk to the VM, and the disk has a certain location, and the name will be certain (/dev/sdb1). So I do hard-coded here.

files/image_config/kubernetes/kubernetes_master_entrance.sh
lixiaoyuner: Just want to check whether it's the first boot here, if it's the first time, need to call the master script. If there is already one same feature, I can reuse the existing one. But after I checked, I don't find there is this flag ( "/host/image-${SONIC_VERSION}/platform/firsttime") both at first boot and subsequent restart

lixiaoyuner: I mean it's not good to use the existing feature. The reason is that we call the k8s master script in the last line of rc.local, but we don't change the rc.local file directly. We insert the call master script line by sed command when include_kubernetes_master=y. If we want to add the calling k8s master script to existing firsttime flag logic, we need to change the rc.local file directly, it's not good to add lines to rc.local by sed command any more. If we agree to change the rc.local file directly, I can try to use the existing flag.

qiluo-msft: I see the point of `first_boot_flag`, `kubernetes_master_entrance.sh` and `sed -i '/^exit 0/i\bash /usr/sbin/kubernetes_master_entrance.sh' $FILESYSTEM_ROOT/etc/rc.local`.

They are introduced for a deployment workflow, but this workflow is fragile. Suggest:
1. Create and start the Kubernetes master VM
2. Configure the VM IP by hyperv-daemons, or by console access
3. SSH into the VM, so you have full control
4. Copy some scripts into the VM
5. Enable/start some services.

files/build_templates/sonic_debian_extension.j2
```diff
@@ -790,15 +790,27 @@ sudo https_proxy=$https_proxy LANG=C chroot $FILESYSTEM_ROOT docker $SONIC_NATIV
...
-sudo LANG=C chroot $FILESYSTEM_ROOT systemctl enable mdm.service
+sudo https_proxy=$https_proxy LANG=C chroot $FILESYSTEM_ROOT pip3 install kubernetes
+# Install python packages to upload and download etcd backup files for backup and restore service usage
+sudo https_proxy=$https_proxy LANG=C chroot $FILESYSTEM_ROOT pip3 install azure-storage-blob azure-identity
 sudo LANG=C chroot $FILESYSTEM_ROOT systemctl enable kubelet.service
 # Add kubernetes master entrance
+sudo cp files/image_config/kubernetes/kubernetes_master_entrance.service ${FILESYSTEM_ROOT_USR_LIB_SYSTEMD_SYSTEM}/
 sudo cp files/image_config/kubernetes/kubernetes_master_entrance.sh $FILESYSTEM_ROOT/usr/sbin/
-sudo sed -i '/^exit 0/i\bash /usr/sbin/kubernetes_master_entrance.sh' $FILESYSTEM_ROOT/etc/rc.local
+sudo chmod +x $FILESYSTEM_ROOT/usr/sbin/kubernetes_master_entrance.sh
```
qiluo-msft, line 812: >/usr/sbin/ [](http://example.com/codeflow?start=30&length=10)

Why you use sbin folder? we prefer `/usr/sbin` for executable.

lixiaoyuner: Okay, I think you mean /usr/bin, let me move the file to /usr/bin

qiluo-msft (approved):

files/image_config/kubernetes/kubernetes_master_entrance.sh
qiluo-msft: I created https://github.com/sonic-net/sonic-buildimage/issues/15849, and please consider it in future PR.

qiluo-msft (dismissed):

files/image_config/kubernetes/kubernetes_master_entrance.sh
lixiaoyuner: You are right, actually we can't ensure the device name is sda or sdb, I updated this part of code. We can specify the disk's location number when attach the disk to the SCSI controller of the VM. Inside VM, we can find the disk by disk path file, because the path file name includes the location number (lun-x). Thus, we can always mount the correct disk.

TRISS708: Wow you might be good at hacking but your still stupid because I always find your online accounts...that's ok as you see I been on reporting things officially and you'll be out my phone off my internet and hopefully stop watching me on your 3D VM weirdo

TRISS708: > >sdb1 [](http://example.com/codeflow?start=11&length=4)
> 
> 
> 
> why hard-coded? #Closed

Because he's hacking me with it. May you must be too

qiluo-msft (approved):

xumia (approved):

rules/config
```diff
@@ -199,12 +199,23 @@ INCLUDE_KUBERNETES_MASTER ?= n
...
 MASTER_PAUSE_VERSION = 3.5
 MASTER_COREDNS_VERSION = v1.8.4
 MASTER_ETCD_VERSION = 3.5.0-0
-MASTER_CRI_DOCKERD = 0.2.5
+MASTER_CRI_DOCKERD = 0.3.1
+MASTER_UI_METRIC_VERSION = v1.0.8
+MASTER_UI_DASH_VERSION = v2.7.0
+MASTER_MDM_VERSION = 2.2023.505.1124-45da18-20230505t1700
+MASTER_MDS_VERSION = mariner_20230517.1
+MASTER_FLUENTD_VERSION = mariner_20230517.1
```
xumia, line 218: Better to create a makefile in rules/ folder, for instance rule/k8s.mk, for all the specific version variables.

ridahanif96:
@lixiaoyuner ,Hi, i am trying to test this code via custom build in rules file where i did "INCLUDE_KUBERNETES=y && INCLUDE_KUBERNETES_MASTER==y" but got build error in [PR](https://github.com/sonic-net/sonic-buildimage/pull/17634). Can you please help.

PR #15721, opened 2023-07-05
[minigraph] remove number of lanes check for changing speed from 400G to 100G and set speed setting before lane reconfiguration  

vdahiya12:
8111 800G interface, split to 2x400G (each has 4 lanes) fails to change interface speed from 400G to 100G during deploy mg. 
In minigraph.xml, the interface speed configuration is good, but fails to generate the right value to config_db.json.

In order to support this SKU the speed transitioning should support both 4 lanes and 8 lanes in the port_config.ini. 

#### Why I did it

before this change  for a 400G to 100G transition, in all cases except when lanes are 8, we would continue and the line 
`ports.setdefault(port_name, {})['speed'] = port_speed_png[port_name]`
would not be executed, hence the default speed will never be set for a case and config_db will not be updated, 
where speed is transitioning from 400G to 100G or 40G, but lanes are not equal to 8. 

In order for those cases to pass where lanes are not specifically 8, we need the change

##### Work item tracking
24242657

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]


abdosi (approved):

prgeor:
@vdahiya12 can you specify in "Why I did it" that 400G interface could be realized with 100G (x 4) PAM4 or 50G (x8)PAM4 and hence the check for 8 lanes will not work for platforms using 100G PAM4?

prgeor (approved):

prgeor:
@lguohan @yxieca please help merge

src/sonic-config-engine/minigraph.py
```diff
@@ -1775,9 +1775,7 @@ def parse_xml(filename, platform=None, port_config_file=None, asic_name=None, hw
         # update the port lanes, use the first 4 lanes of the 400G port to support 100G/40G port
         if port_default_speed == '400000' and (port_png_speed == '100000' or port_png_speed == '40000'):
             port_lanes =  ports[port_name].get('lanes', '').split(',')
-            # check if the 400g port has only 8 lanes
-            if len(port_lanes) != 8:
-                continue
+
             updated_lanes = ",".join(port_lanes[:4])
```
yxieca, line 1779: Would it better to update lane map when the number of lanes is 8? Is taking first 4 lanes always right?

lguohan:
@prgeor , i am not sure i understand the problem. in 8111, 2x400g mode, each 400G port only has 4 lanes, why this 8 lanes check poses issue?

vdahiya12:
> @prgeor , i am not sure i understand the problem. in 8111, 2x400g mode, each 400G port only has 4 lanes, why this 8 lanes check poses issue?

For this Cisco-8111-O64, the minigraph parser fails as we only consider ports with 8 lanes for the 400G to 100G transition, but since we can go from 400G to 100G on a 4 lanes port as well, we need to disable this check.

src/sonic-config-engine/minigraph.py
vdahiya12: in both cases when a port has 4 lanes or 8 lanes, if we want to transition from 400G to 100G, we need only first 4 lanes, so we are right in that regard.

lguohan:
> > @prgeor , i am not sure i understand the problem. in 8111, 2x400g mode, each 400G port only has 4 lanes, why this 8 lanes check poses issue?
> 
> For this Cisco-8111-O64, the minigraph parser fails as we only consider ports with 8 lanes for the 400G to 100G transition, but since we can go from 400G to 100G on a 4 lanes port as well, we need to disable this check.

if the lane is already 4, then there is no need to do updated_lanes. I still do not understand the code logic here.

prgeor:
> > > @prgeor , i am not sure i understand the problem. in 8111, 2x400g mode, each 400G port only has 4 lanes, why this 8 lanes check poses issue?
> > 
> > 
> > For this Cisco-8111-O64, the minigraph parser fails as we only consider ports with 8 lanes for the 400G to 100G transition, but since we can go from 400G to 100G on a 4 lanes port as well, we need to disable this check.
> 
> if the lane is already 4, then there is no need to do updated_lanes. I still do not understand the code logic here.

@lguohan agree. minigraph parser should get the 4 lanes from the port_config.ini. Checking with Kevin Wang what was the issue Wang

prgeor:
@vdahiya12 please update the description according to the latest PR changes

vdahiya12:
@lguohan @yxieca please help merge

src/sonic-config-engine/minigraph.py
```diff
@@ -1771,17 +1771,16 @@ def parse_xml(filename, platform=None, port_config_file=None, asic_name=None, hw
         port_default_speed =  port_speeds_default.get(port_name, None)
         port_png_speed = port_speed_png[port_name]
 
+        ports.setdefault(port_name, {})['speed'] = port_speed_png[port_name]
```
lguohan, line 1774: given 1772 set the port_png_speed, it is better to use it here as well.

src/sonic-config-engine/minigraph.py
```diff
@@ -1771,17 +1771,16 @@ def parse_xml(filename, platform=None, port_config_file=None, asic_name=None, hw
...
+
         # when the port speed is changes from 400g to 100g/40g
         # update the port lanes, use the first 4 lanes of the 400G port to support 100G/40G port
         if port_default_speed == '400000' and (port_png_speed == '100000' or port_png_speed == '40000'):
             port_lanes =  ports[port_name].get('lanes', '').split(',')
-            # check if the 400g port has only 8 lanes
-            if len(port_lanes) != 8:
-                continue
```
lguohan, line 1780: i think this is not the problem, we should keep this.

theoritical speaking, there about 2lane, each lane is 200G, total 400G configuration. in this case, you do not want to select 4 lanes since the port only has 2 lanes.

prgeor: @vdahiya12 
consider following cases:-
1. 400G using 8x50G
2. 400G using 4x100G
3. 400G using 2x200G (in the future)

A. We should skip in the #3 case because we don't have 4 lanes in port_config.ini to choose from for 100G

B. #2 In this case we don't need to select the lanes for 100G speed because its already 4 lanes

So, the only case which can support 100G NRZ is when 400G is 8*50G so we should skip updating the lanes to first 4 lanes if the number of lanes is not 8

src/sonic-config-engine/minigraph.py
vdahiya12: fixed

src/sonic-config-engine/minigraph.py
vdahiya12: fixed

PR #14788, opened 2023-04-21
Move k8s script to docker-config-engine

lixiaoyuner:
#### Why I did it
To reduce the container's dependency from host system
##### Work item tracking
- Microsoft ADO **(number only)**:
17713469
#### How I did it
Move the k8s container startup script to config engine container, other than mount it from host.
#### How to verify it
Check file path(/usr/share/sonic/scripts/container_startup.py) inside config engine container.

#### Which release branch to backport (provide reason below if selected)

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [x] 202205
- [x] 202211

#### Tested branch (Please provide the tested image version)

- [ ] 
- [ ]

PR #15740, opened 2023-07-07
Move k8s script to docker-config-engine (#14788)

lixiaoyuner:
Why I did it
To reduce the container's dependency from host system

Work item tracking
Microsoft ADO (number only):
17713469
How I did it
Move the k8s container startup script to config engine container, other than mount it from host.

How to verify it
Check file path(/usr/share/sonic/scripts/container_startup.py) inside config engine container.


yxieca:
This PR cherry-picks #14788

PR #15772, opened 2023-07-10
Container image clean up bug fix

lixiaoyuner:
#### Why I did it
When do clean up container images, current code has two bugs need to be fixed. And some variables' name maybe cause confused, change the variables' name.

##### Work item tracking
- Microsoft ADO **(number only)**: 24502294

#### How I did it
- We do clean up after tag latest successfully. But currently tag latest function only return 0 and 1, 0 means succeed and 1 means failed, when we get 1, we will retry, when we get 0, we will do clean up. Actually the code 0 includes another case we don't need to do clean up. The case is that when we are doing tag latest, the container image we want to tag maybe not running, so we can not tag latest and don't need to cleanup, we need to separate this case from 0, return -1 now.
- When local mode(v1) -> kube mode(v2) happens, one problem is how to handle the local image, there are two cases. one case is that there was one kube v1 container dry-run(cause we don't relace the local if kube version = local version), we will remove the kube v1 image and tag the local version with ACR prefix and remove local v1 local tag. Another case is that there was no kube v1 container dry-run, we remove the local v1 image directly, cause the local v1 image should not be the last desire version.
- About the docker_id variable, it may cause confused, it's actually docker image id, so rename the variable. About the two dicts and the list, rename them to be more readable.
#### How to verify it
Check tag latest and image clean up result.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [x] 202205
- [x] 202211
- [x] 202305

- [x] 20220531.28
- [ ]

### DIFF
--- a/src/sonic-ctrmgrd/ctrmgr/kube_commands.py
+++ b/src/sonic-ctrmgrd/ctrmgr/kube_commands.py
@@ -478,7 +478,7 @@ def tag_latest(feat, docker_id, image_ver):
         else:
             log_error(err)
     elif ret == -1:
-        ret = 0
+        log_debug(out)
     else:
         log_error(err)
     return ret
@@ -487,18 +487,18 @@ def _do_clean(feat, current_version, last_version):
     err = ""
     out = ""
     ret = 0
-    DOCKER_ID = "docker_id"
+    IMAGE_ID = "image_id"
     REPO = "repo"
     _, image_info, err = _run_command("docker images |grep {} |grep -v latest |awk '{{print $1,$2,$3}}'".format(feat))
     if image_info:
         version_dict = {}
         version_dict_default = {}
         for info in image_info.split("\n"):
-            rep, version, docker_id = info.split()
+            rep, version, image_id = info.split()
             if len(rep.split("/")) == 1:
-                version_dict_default[version] = {DOCKER_ID: docker_id, REPO: rep}
+                version_dict_default[version] = {IMAGE_ID: image_id, REPO: rep}
             else:
-                version_dict[version] = {DOCKER_ID: docker_id, REPO: rep}
+                version_dict[version] = {IMAGE_ID: image_id, REPO: rep}
 
         if current_version in version_dict:
             image_prefix = version_dict[current_version][REPO]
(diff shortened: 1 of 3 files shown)


src/sonic-ctrmgrd/ctrmgr/kube_commands.py
```diff
@@ -509,9 +509,16 @@ def _do_clean(feat, current_version, last_version):
...
         # should be only one item in version_dict_default
         for k, v in version_dict_default.items():
-            local_version, local_repo, local_docker_id = k, v[REPO], v[DOCKER_ID]
-            tag_res, _, err = _run_command("docker tag {} {}:{} && docker rmi {}:{}".format(
-                local_docker_id, image_prefix, local_version, local_repo, local_version))
+            local_version, local_repo, local_image_id = k, v[REPO], v[IMAGE_ID]
+            # if there is a kube image with same version, need to remove the kube version
+            # and tag the local version to kube version for fallback preparation
+            # and remove the local version
+            # if there is no kube image with same version, just remove the local version
```
losha228, line 516: it's better to put it in front of else.

src/sonic-ctrmgrd/ctrmgr/kube_commands.py
```diff
@@ -523,7 +530,7 @@ def _do_clean(feat, current_version, last_version):
         if last_version in version_dict:
             del version_dict[last_version]
 
-        versions = [item[DOCKER_ID] for item in version_dict.values()]
+        versions = [item[IMAGE_ID] for item in version_dict.values()]
```
losha228, line 533: the "versions" here means image id list ?

src/sonic-ctrmgrd/ctrmgr/kube_commands.py
```diff
@@ -487,18 +487,18 @@ def _do_clean(feat, current_version, last_version):
     err = ""
     out = ""
     ret = 0
-    DOCKER_ID = "docker_id"
+    IMAGE_ID = "image_id"
     REPO = "repo"
     _, image_info, err = _run_command("docker images |grep {} |grep -v latest |awk '{{print $1,$2,$3}}'".format(feat))
     if image_info:
         version_dict = {}
         version_dict_default = {}
```
losha228, line 495: one suggestion: use a meaningful name for var, it will be more readable.

src/sonic-ctrmgrd/ctrmgr/kube_commands.py
lixiaoyuner: > the "versions" here means image id list ?

Yes, it's image id list which we need to remove. Let me rename the variable

src/sonic-ctrmgrd/ctrmgr/kube_commands.py
lixiaoyuner: Okay.

src/sonic-ctrmgrd/ctrmgr/kube_commands.py
lixiaoyuner: Fixed

losha228 (approved):

lguohan (approved):

PR #15791, opened 2023-07-11
[202012] [TACACS+] When load-minigraph enable per-command accounting by default

liuh-80:
When load-minigraph enable per-command accounting by default.

##### Work item tracking
- Microsoft ADO **(number only)**: 24433713
- 
#### Why I did it
Enable per-command accounting by default.

#### How I did it
Add hard code config in sonic-cfggen.

#### How to verify it
Also pass all current UT.
Manually check the TACACS per-command accounting enabled after load minigraph.

N/A

When load-minigraph enable per-command accounting by default.


liuh-80:
Closed because create PR with wrong target branch.

PR #15792, opened 2023-07-11
[202012] [TACACS+] When load-minigraph enable per-command accounting by default

liuh-80:
When load-minigraph enable per-command accounting by default.

##### Work item tracking
- Microsoft ADO **(number only)**: 24433713
- 
#### Why I did it
Enable per-command accounting by default.

#### How I did it
Add hard code config in sonic-cfggen.

#### How to verify it
Also pass all current UT.
Manually check the TACACS per-command accounting enabled after load minigraph.

    N/A

Extract tacacs support functions into library, this will share TACACS config file parse code with other project.
Also fix memory leak issue in parse config code.

- [ ]  SONiC.202012-15723.312602-e230e2d3e

When load-minigraph enable per-command accounting by default.

### DIFF
--- a/src/sonic-config-engine/sonic-cfggen
+++ b/src/sonic-config-engine/sonic-cfggen
@@ -331,6 +331,14 @@ def main():
         else:
             deep_update(data, parse_xml(minigraph, port_config_file=args.port_config, asic_name=asic_name, hwsku_config_file=args.hwsku_config))
 
+        # enable TACACS per-command accounting by default on 202012 branch
+        tacacs_accounting = {
+                'AAA': {
+                    'accounting': {
+                        "login": "tacacs+"
+                }}}
+        deep_update(data, tacacs_accounting)
+
     if args.device_description is not None:
         deep_update(data, parse_device_desc_xml(args.device_description))
 


src/sonic-config-engine/sonic-cfggen
```diff
@@ -331,6 +331,14 @@ def main():
         else:
             deep_update(data, parse_xml(minigraph, port_config_file=args.port_config, asic_name=asic_name, hwsku_config_file=args.hwsku_config))
 
+        # enable TACACS per-command accounting by default on 202012 branch
+        tacacs_accounting = {
+                'AAA': {
+                    'accounting': {
+                        "login": "tacacs+"
```
qiluo-msft, line 338: What happen if tacacs (address, passkey, etc) is not configured properly?

src/sonic-config-engine/sonic-cfggen
```diff
@@ -331,6 +331,14 @@ def main():
         else:
             deep_update(data, parse_xml(minigraph, port_config_file=args.port_config, asic_name=asic_name, hwsku_config_file=args.hwsku_config))
 
+        # enable TACACS per-command accounting by default on 202012 branch
```
qiluo-msft, line 334: sonic-cfggen may be not proper place to add the default value. For example, in [L2-Switch-mode](https://github.com/sonic-net/SONiC/wiki/L2-Switch-mode), there is no minigraph.xml at all, and there is no need to set the accounting default value.

liuh-80: Offline discussion, this is OK

liuh-80:
According to offline discussion, will close this PR because 202012 need align with master branch.

PR #15829, opened 2023-07-13
[docker-teamd] limit privileged flag for teamd container

maipbui:
#### Why I did it
HLD implementation: Container Hardening (https://github.com/sonic-net/SONiC/pull/1364)
##### Work item tracking
- Microsoft ADO **(number only)**: 14807420

#### How I did it
Reduce linux capabilities in privileged flag, retain NET_ADMIN capability

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/rules/docker-teamd.mk
+++ b/rules/docker-teamd.mk
@@ -33,7 +33,7 @@ SONIC_INSTALL_DOCKER_DBG_IMAGES += $(DOCKER_TEAMD_DBG)
 endif
 
 $(DOCKER_TEAMD)_CONTAINER_NAME = teamd
-$(DOCKER_TEAMD)_RUN_OPT += --privileged -t
+$(DOCKER_TEAMD)_RUN_OPT += -t --cap-add=NET_ADMIN
 $(DOCKER_TEAMD)_RUN_OPT += -v /etc/sonic:/etc/sonic:ro
 $(DOCKER_TEAMD)_RUN_OPT += -v /etc/timezone:/etc/timezone:ro 
 $(DOCKER_TEAMD)_RUN_OPT += -v /host/warmboot:/var/warmboot


qiluo-msft (approved):

Yarden-Z (approved):

judyjoseph:
Looks ok here for teamd, one question on below capability.

> NET_RAW | Use RAW and PACKET sockets
```
    Is this needed for doing tcpdump inside the docker container ? 
    In the multi-asic platforms we usually go into one of these dockers in the "asic" namespace and do a tcpdump
    as the interfaces are present in the namespace. 
```
Another interesting capability is below
> NET_BIND_SERVICE | Bind a socket to internet domain privileged ports (port numbers less than 1024).
```
  Not significant to L2 protocols like LLDP, LACP, but would need to set this in case of frr docker, as we have BGP bould to port 179.
```

maipbui:
> Looks ok here for teamd, one question on below capability.
> 
> > NET_RAW | Use RAW and PACKET sockets
> 
> ```
>     Is this needed for doing tcpdump inside the docker container ? 
>     In the multi-asic platforms we usually go into one of these dockers in the "asic" namespace and do a tcpdump
>     as the interfaces are present in the namespace. 
> ```
> 
> Another interesting capability is below
> 
> > NET_BIND_SERVICE | Bind a socket to internet domain privileged ports (port numbers less than 1024).
> 
> ```
>   Not significant to L2 protocols like LLDP, LACP, but would need to set this in case of frr docker, as we have BGP bould to port 179.
> ```

@judyjoseph both NET_RAW and NET_BIND_SERVICE caps are granted by default if not use `--privileged` flag, check this ref: https://docs.docker.com/engine/reference/run/#runtime-privilege-and-linux-capabilities

judyjoseph (approved):

PR #15924, opened 2023-07-20
[build] add support for 2 stage rootfs build

Yakiv-Huryk:
This adds optimization for the SONiC image build by splitting the final build step into two stages. It allows running the first stage in parallel, improving build time.

The optimization is enabled via new rules/config flag _ENABLE_RFS_SPLIT_BUILD_ (disabled by default)

#### Why I did it
To improve a build time.

##### Work item tracking

#### How I did it
Added a logic to run build_debian.sh in two stages, transferring the progress via a new build artifact.

#### How to verify it
```bash
make ENABLE_RFS_SPLIT_BUILD=y SONIC_BUILD_JOBS=32 target/<IMAGE_NAME>.bin
```

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/Makefile.cache
+++ b/Makefile.cache
@@ -363,8 +363,8 @@ RFS_DEP_FILES := $(wildcard \
 	$(shell git ls-files files/image_config) \
 	$(shell git ls-files files/apparmor) \
 	$(shell git ls-files files/apt) \
-	$(wildcard files/sshd/*) \
-	$(wildcard files/dhcp/*) \
+	$(shell git ls-files files/sshd) \
+	$(shell git ls-files files/dhcp) \
 	src/sonic-build-hooks/buildinfo/trusted.gpg.d \
 	platform/$(CONFIGURED_PLATFORM)/modules  \
 	files/docker/docker.service.conf \
--- a/Makefile.cache
+++ b/Makefile.cache
@@ -360,10 +360,9 @@ endef
 RFS_DEP_FILES := $(wildcard \
 	$(addprefix scripts/, build_debian_base_system.sh prepare_debian_image_buildinfo.sh build_mirror_config.sh) \
 	$(wildcard files/initramfs-tools/*) \
-	$(addprefix files/image_config/, kdump/vmcore-sysctl.conf ntp/ntp-apparmor sonic_release ntp/ntp ntp/ntp-systemd-wrapper \
-		systemd/systemd-udevd/fix-udev-rule-may-fail-if-daemon-reload-command-runs.patch) \
-	$(addprefix files/sshd/, host-ssh-keygen.sh override.conf) \
-	$(addprefix files/dhcp/, rfc3442-classless-routes sethostname sethostname6 graphserviceurl snmpcommunity vrf) \
+	$(shell git ls-files files/image_config) \
+	$(wildcard files/sshd/*) \
+	$(wildcard files/dhcp/*) \
 	src/sonic-build-hooks/buildinfo/trusted.gpg.d \
 	platform/$(CONFIGURED_PLATFORM)/modules  \
 	files/docker/docker.service.conf \
--- a/Makefile.cache
+++ b/Makefile.cache
@@ -357,6 +357,20 @@ define SAVE_CACHE
 	$(if $(call CHECK_WCACHE_ENABLED,$(1)), $(call SAVE_INTO_CACHE,$(1),$(2)))
 endef
 
+RFS_DEP_FILES := $(wildcard \
+	$(addprefix scripts/, build_debian_base_system.sh prepare_debian_image_buildinfo.sh build_mirror_config.sh) \
+	$(wildcard files/initramfs-tools/*) \
+	$(addprefix files/image_config/, kdump/vmcore-sysctl.conf ntp/ntp-apparmor sonic_release ntp/ntp ntp/ntp-systemd-wrapper \
+		systemd/systemd-udevd/fix-udev-rule-may-fail-if-daemon-reload-command-runs.patch) \
+	$(addprefix files/sshd/, host-ssh-keygen.sh override.conf) \
+	$(addprefix files/dhcp/, rfc3442-classless-routes sethostname sethostname6 graphserviceurl snmpcommunity vrf) \
+	src/sonic-build-hooks/buildinfo/trusted.gpg.d \
+	platform/$(CONFIGURED_PLATFORM)/modules  \
+	files/docker/docker.service.conf \
+	files/build_templates/default_users.json.j2 \
+	files/build_scripts/generate_asic_config_checksum.py \
+	files/scripts/core_cleanup.py \
+	build_debian.sh onie-image.conf)
 
 
 # Set the target path for each target.
@@ -386,6 +400,9 @@ $(foreach pkg, $(SONIC_INSTALL_PKGS), \
 
 $(foreach pkg, $(SONIC_RFS_TARGETS), \
         $(eval $(pkg)_DST_PATH := $(if $($(pkg)_DST_PATH), $($(pkg)_DST_PATH), $(TARGET_PATH))) \
+        $(eval $(pkg)_CACHE_MODE := GIT_CONTENT_SHA) \
+        $(eval $(pkg)_DEP_FLAGS := $(SONIC_COMMON_FLAGS_LIST)) \
+        $(eval $(pkg)_DEP_FILES := $(SONIC_COMMON_BASE_FILES_LIST) $(RFS_DEP_FILES)) \
         $(eval $(TARGET_PATH)/$(pkg)_TARGET := $(pkg)) )
 
 # define the DEP files(.dep and .smdep) and SHA files (.sha and smsha) for each target
--- a/Makefile.cache
+++ b/Makefile.cache
@@ -359,8 +359,10 @@ endef
 
 RFS_DEP_FILES := $(wildcard \
 	$(addprefix scripts/, build_debian_base_system.sh prepare_debian_image_buildinfo.sh build_mirror_config.sh) \
-	$(wildcard files/initramfs-tools/*) \
+	$(shell git ls-files files/initramfs-tools) \
 	$(shell git ls-files files/image_config) \
+	$(shell git ls-files files/apparmor) \
+	$(shell git ls-files files/apt) \
 	$(wildcard files/sshd/*) \
 	$(wildcard files/dhcp/*) \
 	src/sonic-build-hooks/buildinfo/trusted.gpg.d \
--- a/Makefile.cache
+++ b/Makefile.cache
@@ -384,11 +384,14 @@ $(foreach pkg, $(SONIC_INSTALL_PKGS), \
         $(eval $(pkg)_DST_PATH := $(if $($(pkg)_DST_PATH), $($(pkg)_DST_PATH), $(FSROOT_PATH))) \
 		$(eval $(FSROOT_PATH)/$(pkg)_TARGET := $(pkg)) )
 
+$(foreach pkg, $(SONIC_RFS_TARGETS), \
+        $(eval $(pkg)_DST_PATH := $(if $($(pkg)_DST_PATH), $($(pkg)_DST_PATH), $(TARGET_PATH))) \
+        $(eval $(TARGET_PATH)/$(pkg)_TARGET := $(pkg)) )
 
 # define the DEP files(.dep and .smdep) and SHA files (.sha and smsha) for each target
 $(foreach pkg, $(SONIC_MAKE_DEBS) $(SONIC_DPKG_DEBS) $(SONIC_ONLINE_DEBS) $(SONIC_COPY_DEBS) \
         $(SONIC_MAKE_FILES) $(SONIC_PYTHON_STDEB_DEBS) $(SONIC_PYTHON_WHEELS) \
-        $(SONIC_DOCKER_IMAGES) $(SONIC_DOCKER_DBG_IMAGES) $(SONIC_INSTALL_PKGS), \
+        $(SONIC_DOCKER_IMAGES) $(SONIC_DOCKER_DBG_IMAGES) $(SONIC_INSTALL_PKGS) $(SONIC_RFS_TARGETS), \
         $(eval $(pkg)_MOD_SRC_PATH:=$(if $($(pkg)_SRC_PATH),$($(pkg)_SRC_PATH),$($(pkg)_PATH))) \
         $(eval $(pkg)_BASE_PATH:=$(if $($(pkg)_BASE_PATH),$($(pkg)_BASE_PATH),$(CURDIR))) \
         $(eval $(pkg)_DEP_FLAGS_FILE:=$($(pkg)_DST_PATH)/$(pkg).flags) \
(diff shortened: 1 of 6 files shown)


Yakiv-Huryk:
HLD:
https://github.com/sonic-net/SONiC/pull/1430

vivekrnv (approved):

build_debian.sh
```diff
@@ -616,6 +621,46 @@ if [[ ! -f './asic_config_checksum' ]]; then
...
+    ## Display all process details access /proc
+    sudo LANG=C chroot $FILESYSTEM_ROOT fuser -vm /proc
+    ## Kill the processes
+    sudo LANG=C chroot $FILESYSTEM_ROOT fuser -km /proc || true
+    ## Wait fuser fully kill the processes
+    sleep 15
+    sudo LANG=C chroot $FILESYSTEM_ROOT umount /proc
+
+    sudo rm -f $TARGET_PATH/$RFS_SQUASHFS_NAME
+    sudo mksquashfs $FILESYSTEM_ROOT $TARGET_PATH/$RFS_SQUASHFS_NAME
```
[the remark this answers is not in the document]
saiarcot895: Agreed. I _think_ a simple `tar` (without any compression) would serve the purpose here.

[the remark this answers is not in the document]
Yakiv-Huryk: Good point. I did some testing:
* default: 27 sec, 532M
* compression-level 1: 2 sec, 574M
* no compression: 1 sec, 1.4G

The "compression-level 1" seems to be the optimal tradeoff.

@saiarcot895 
Regarding `tar`, I'd stick with `mksquashfs` since it is already tested and (after compression tuning) performs very well.

I've updated the PR.

slave.mk
```diff
@@ -1208,6 +1210,60 @@ $(DOCKER_LOAD_TARGETS) : $(TARGET_PATH)/%.gz-load : .platform docker-start $$(TA
...
+$(foreach installer, $(SONIC_INSTALLERS), $(eval $(installer)_RFS_DEPENDS=$(call rfs_get_installer_dependencies,$(installer))))
+
+SONIC_RFS_TARGETS= $(foreach installer, $(SONIC_INSTALLERS), $(call rfs_get_installer_dependencies,$(installer)))
+SONIC_TARGET_LIST += $(addprefix $(TARGET_PATH)/, $(SONIC_RFS_TARGETS))
+endif
+
+$(addprefix $(TARGET_PATH)/, $(SONIC_RFS_TARGETS)) : $(TARGET_PATH)/% : \
+        .platform \
+        build_debian.sh \
+        $(addprefix $(IMAGE_DISTRO_DEBS_PATH)/,$(INITRAMFS_TOOLS) $(LINUX_KERNEL))
```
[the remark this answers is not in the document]
Yakiv-Huryk: Added in https://github.com/sonic-net/sonic-buildimage/pull/15924/commits/8e1e685e3133235c49af5648595e5eacdfbc9452

build_debian.sh
```diff
@@ -616,6 +621,46 @@ if [[ ! -f './asic_config_checksum' ]]; then
...
+if [[ $RFS_SPLIT_FIRST_STAGE == y ]]; then
+    echo '[INFO] Finished with RFS first stage'
+    echo '[INFO] Umount all'
+
+    ## Display all process details access /proc
+    sudo LANG=C chroot $FILESYSTEM_ROOT fuser -vm /proc
+    ## Kill the processes
+    sudo LANG=C chroot $FILESYSTEM_ROOT fuser -km /proc || true
+    ## Wait fuser fully kill the processes
+    sleep 15
```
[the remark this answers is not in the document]
Yakiv-Huryk: As you pointed out, it's derived from existing code. This is a separate issue, fixing it is out of the scope of this PR.

slave.mk
```diff
@@ -1208,6 +1210,62 @@ $(DOCKER_LOAD_TARGETS) : $(TARGET_PATH)/%.gz-load : .platform docker-start $$(TA
...
+endef
+
+$(foreach installer, $(SONIC_INSTALLERS), $(eval $(installer)_RFS_DEPENDS=$(call rfs_get_installer_dependencies,$(installer))))
+SONIC_RFS_TARGETS= $(foreach installer, $(SONIC_INSTALLERS), $(call rfs_get_installer_dependencies,$(installer)))
+
+ifeq ($(ENABLE_RFS_SPLIT_BUILD),y)
+SONIC_TARGET_LIST += $(addprefix $(TARGET_PATH)/, $(SONIC_RFS_TARGETS))
+endif
+
+$(addprefix $(TARGET_PATH)/, $(SONIC_RFS_TARGETS)) : $(TARGET_PATH)/% : \
```
qiluo-msft, line 1225: >SONIC_RFS_TARGETS [](http://example.com/codeflow?start=31&length=17)

Could you make sure the new target `SONIC_RFS_TARGETS` is reused if we build multiple vendors image in series?

We also have reproducible mechanism in rules/*.dep files. Is it possible to define this target's dep file, so the target is reused as cached for other later build jobs?

@xumia to review.

liat-grozovik:
who ever is reviewing this PR, we have a target to finish this review and enjoy the approved time of build in the next week.
who ever is tagged and wish to provide feedback please do it asap. 
@Yakiv-Huryk @xumia @Kalimuthu-Velappan

slave.mk
```diff
@@ -1210,21 +1210,23 @@ $(DOCKER_LOAD_TARGETS) : $(TARGET_PATH)/%.gz-load : .platform docker-start $$(TA
...
+
 define rfs_get_installer_dependencies
 $(1)__$($(1)_MACHINE)__rfs.squashfs $(if $($(1)_DEPENDENT_MACHINE),$(1)__$($(1)_DEPENDENT_MACHINE)__rfs.squashfs)
 endef
 
 $(foreach installer, $(SONIC_INSTALLERS), $(eval $(installer)_RFS_DEPENDS=$(call rfs_get_installer_dependencies,$(installer))))
-
 SONIC_RFS_TARGETS= $(foreach installer, $(SONIC_INSTALLERS), $(call rfs_get_installer_dependencies,$(installer)))
+
+ifeq ($(ENABLE_RFS_SPLIT_BUILD),y)
```
Kalimuthu-Velappan, line 1221: Can you split the build_debian.sh into two separate files?
1. build_debian_rootfs.sh  => which contains all the standard squashfs packages
2. build_debian_sonic.sh => Includes all the sonic packages into existing squashfs and creates the dockerfs.

slave.mk
```diff
@@ -1573,7 +1575,9 @@ SONIC_CLEAN_TARGETS += $(addsuffix -clean,$(addprefix $(TARGET_PATH)/, \
 		       $(SONIC_DOCKER_IMAGES) \
 		       $(SONIC_DOCKER_DBG_IMAGES) \
 		       $(SONIC_SIMPLE_DOCKER_IMAGES) \
-		       $(SONIC_INSTALLERS)))
+		       $(SONIC_INSTALLERS) \
+		       $(SONIC_RFS_TARGETS)))
```
Kalimuthu-Velappan, line 1579: Can you add caching support for rootfs preparation in squashfs file system?
You can refer DPKG caching framework for deb packages.

slave.mk
```diff
@@ -1210,21 +1210,23 @@ $(DOCKER_LOAD_TARGETS) : $(TARGET_PATH)/%.gz-load : .platform docker-start $$(TA
...
+ifeq ($(ENABLE_RFS_SPLIT_BUILD),y)
 SONIC_TARGET_LIST += $(addprefix $(TARGET_PATH)/, $(SONIC_RFS_TARGETS))
 endif
 
 $(addprefix $(TARGET_PATH)/, $(SONIC_RFS_TARGETS)) : $(TARGET_PATH)/% : \
         .platform \
         build_debian.sh \
-        $(addprefix $(IMAGE_DISTRO_DEBS_PATH)/,$(INITRAMFS_TOOLS) $(LINUX_KERNEL))
+        $(addprefix $(IMAGE_DISTRO_DEBS_PATH)/,$(INITRAMFS_TOOLS) $(LINUX_KERNEL)) \
+        $(addsuffix -install,$(addprefix $(IMAGE_DISTRO_DEBS_PATH)/,$(DEBOOTSTRAP)))
```
Kalimuthu-Velappan, line 1229: Can you add tmpfs support for generating the squashfs preparation?
This will improve the performance by 5x.

slave.mk
Yakiv-Huryk: The target is reused. I've also added a cache: https://github.com/sonic-net/sonic-buildimage/pull/15924/commits/a880d53d14d593f6739e0ca33d22354157503aec
Since the targets are defined dynamically, I could not add the '.dep' file. Instead, the relevant variables are also set dynamically.

slave.mk
Yakiv-Huryk: I'm not sure I understand why splitting the build_debian.sh is better. Also, with the current design, the feature is optional and configurable (via ENABLE_RFS_SPLIT_BUILD). If we split the build_debian.sh, the current way to build will no longer be available. 

Can you please help to understand why the _if..else_ is better avoided?

slave.mk
Yakiv-Huryk: Added in https://github.com/sonic-net/sonic-buildimage/pull/15924/commits/a880d53d14d593f6739e0ca33d22354157503aec

slave.mk
Yakiv-Huryk: It's a nice idea. Unfortunately, it's not in the scope of this PR.

Yakiv-Huryk:
> 
@k-v1 
Fixed the # 1.

Regarding the first stage dependencies (including many files from `/files`):
I've collected a list and added it to the _DEP_FILES. 
However, it's not a robust solution since there is no way to track what is used in the `build_debian.sh`, and the list should be manually maintained which is not optimal.
@xumia @Kalimuthu-Velappan  do you have any ideas/suggestions on how to handle this?

rules/config
```diff
@@ -300,3 +300,6 @@ GZ_COMPRESS_PROGRAM ?= gzip
 
 # SONIC_OS_VERSION - sonic os version
 SONIC_OS_VERSION ?= 11
+
+# ENABLE_RFS_SPLIT - enable 2-stage installer build
+ENABLE_RFS_SPLIT_BUILD ?= n
```
qiluo-msft, line 305: >ENABLE_RFS_SPLIT_BUILD [](http://example.com/codeflow?start=0&length=22)

Since it is an optimization, no need to enable it by config, just enable by default.

Yakiv-Huryk: I'd suggest to consider keeping the flag and having it on by default.
In this case, if there are any issues, we can just disable it and have the build working as it is now.

@qiluo-msft @xumia @Kalimuthu-Velappan what do you think?

qiluo-msft: If there are any issues, we can just revert this PR since the PR is small.

Yakiv-Huryk: Ok, done

qiluo-msft (approved):

Makefile.cache
```diff
@@ -357,6 +357,20 @@ define SAVE_CACHE
 	$(if $(call CHECK_WCACHE_ENABLED,$(1)), $(call SAVE_INTO_CACHE,$(1),$(2)))
 endef
 
+RFS_DEP_FILES := $(wildcard \
+	$(addprefix scripts/, build_debian_base_system.sh prepare_debian_image_buildinfo.sh build_mirror_config.sh) \
+	$(wildcard files/initramfs-tools/*) \
+	$(addprefix files/image_config/, kdump/vmcore-sysctl.conf ntp/ntp-apparmor sonic_release ntp/ntp ntp/ntp-systemd-wrapper \
+		systemd/systemd-udevd/fix-udev-rule-may-fail-if-daemon-reload-command-runs.patch) \
+	$(addprefix files/sshd/, host-ssh-keygen.sh override.conf) \
+	$(addprefix files/dhcp/, rfc3442-classless-routes sethostname sethostname6 graphserviceurl snmpcommunity vrf) \
```
saiarcot895, line 366: Would it be better to use a wildcard/glob expression here, so that if some file gets added/removed, it's automatically added to this list?

Yakiv-Huryk: Good idea, added in https://github.com/sonic-net/sonic-buildimage/pull/15924/commits/2d638f0755d103db0e979ee5b0265f692907e019

dgsudharsan:
@saiarcot895  @qiluo-msft Can you please review and signoff?

slave.mk
```diff
@@ -1210,6 +1238,61 @@ $(DOCKER_LOAD_TARGETS) : $(TARGET_PATH)/%.gz-load : .platform docker-start $$(TA
...
+		$(eval machine=$($*_MACHINE))
+
+		export debs_path="$(IMAGE_DISTRO_DEBS_PATH)"
+		export initramfs_tools="$(IMAGE_DISTRO_DEBS_PATH)/$(INITRAMFS_TOOLS)"
+		export linux_kernel="$(IMAGE_DISTRO_DEBS_PATH)/$(LINUX_KERNEL)"
+		export kversion="$(KVERSION)"
+		export image_type="$($(installer)_IMAGE_TYPE)"
+		export sonicadmin_user="$(USERNAME)"
+		export sonic_asic_platform="$(patsubst %-$(CONFIGURED_ARCH),%,$(CONFIGURED_PLATFORM))"
+		export RFS_SPLIT_FIRST_STAGE=y
```
saiarcot895, line 1265: Nitpick, but can you set `RFS_SPLIT_LAST_STAGE` to `n` here, so that in case in the future, some code gets added that references this variable before `RFS_SPLIT_FIRST_STAGE`, it will still be defined?

build_debian.sh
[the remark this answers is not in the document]
Yakiv-Huryk: Done in https://github.com/sonic-net/sonic-buildimage/pull/15924/commits/b6f90066a59c5171cfb38651cd19ac38a3609ede

slave.mk
Yakiv-Huryk: Sure, added in https://github.com/sonic-net/sonic-buildimage/pull/15924/commits/b6f90066a59c5171cfb38651cd19ac38a3609ede

slave.mk
```diff
@@ -1210,6 +1238,61 @@ $(DOCKER_LOAD_TARGETS) : $(TARGET_PATH)/%.gz-load : .platform docker-start $$(TA
...
+
+		j2 -f env files/initramfs-tools/union-mount.j2 onie-image.conf > files/initramfs-tools/union-mount
+		j2 -f env files/initramfs-tools/arista-convertfs.j2 onie-image.conf > files/initramfs-tools/arista-convertfs
```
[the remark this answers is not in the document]
Yakiv-Huryk: Make sense. Anyway, I've fixed it to use the `git ls-files`

Makefile.cache
```diff
@@ -357,6 +357,21 @@ define SAVE_CACHE
...
 endef
 
+RFS_DEP_FILES := $(wildcard \
+	$(addprefix scripts/, build_debian_base_system.sh prepare_debian_image_buildinfo.sh build_mirror_config.sh) \
+	$(shell git ls-files files/initramfs-tools) \
+	$(shell git ls-files files/image_config) \
+	$(shell git ls-files files/apparmor) \
+	$(shell git ls-files files/apt) \
+	$(wildcard files/sshd/*) \
+	$(wildcard files/dhcp/*) \
```
saiarcot895, line 367: Can these be changed to `git ls-files` as well?

Yakiv-Huryk: Sure, done.

liat-grozovik:
@saiarcot895 could you please take further look now that the comments were addressed?

saiarcot895 (approved):

liat-grozovik (approved):

liat-grozovik:
@Yakiv-Huryk many thanks for this great achievements. 
Thanks for all the rest who provided feedback and made it even better

skg-net:
@Yakiv-Huryk Can you please update the Quality Metric (Alpha/Beta/GA) for the feature either in this PR comments or in HLD itself based on https://github.com/sonic-net/SONiC/blob/master/doc/SONiC%20feature%20quality%20definition.md
Thanks

Yakiv-Huryk:
> @Yakiv-Huryk Can you please update the Quality Metric (Alpha/Beta/GA) for the feature either in this PR comments or in HLD itself based on https://github.com/sonic-net/SONiC/blob/master/doc/SONiC%20feature%20quality%20definition.md Thanks

It’s GA. Since it’s a build improvement, it doesn't really fit the Feature Quality Definition. There is no flag to enable/disable it and no sonic-mgmt tests.

PR #15925, opened 2023-07-20
[202012][Arista] Always clear reload cause after reading

byu343:
#### Why I did it
On 202012 images, there is an issue that certain reload causes are not cleared after reading and the uncleared causes may confuse the following reboot-cause reading after warm-reboot to master/202305.

##### Work item tracking
- Microsoft ADO **(number only)**: 24378183

#### How I did it
Clear the reload cause even in the case that it is determined as not important. 

#### How to verify it
1) Cold-boot to 202012
2) Warm-reboot to master
3) Check the reboot cause is 'warm-reboot' by 'show reboot-cause'

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012 This Arista driver change is only towards 202012 branch.
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]


vaibhavhd:
@byu343, thanks for the fix. Do we not need a similar fix on master? Last year this issue was fixed for 201811 to 202012, now I think this PR will fix it for 202012->202305.
It would be ideal to have a fix that would prevent us from revisiting this issue for every new cross-branch upgrade.

vaibhavhd:
MSFT ADO: 24378183

byu343:
> 

@vaibhavhd No, this is not needed for master. I have checked the code that the issue here is not affecting the later release branches and master branch.
For 201811, certain reload causes were not read at all, and here for 202012, such reload causes were read but not cleared, so there is slight difference.

prgeor (approved):

prgeor:
@qiluo-msft please merge this to 202012

vaibhavhd (approved):

qiluo-msft (approved):

prgeor (approved):

PR #16018, opened 2023-08-02
[memory_checker] Change log severity to WARNING in a case when the docker service is not running

vadymhlushko-mlnx:
#### Why I did it
To fix the logic introduced by [[memory_checker] Do not check memory usage of containers which are not created #11129](https://github.com/sonic-net/sonic-buildimage/pull/11129).
There could be a scenario before the reboot, where
1. The `docker service` has stopped
2. In a very short period of time, the monit service performs the `root@sonic:/home/admin# monit status container_memory_telemetry`

In such scenario, the `memory_checker` script will throw an error to the syslog:
```
ERR memory_checker: Failed to retrieve the running container list from docker daemon! Error message is: 'Error while fetching server API version: ('Connection aborted.', FileNotFoundError(2, 'No such file or directory'))'
```
But, actually, this scenario is a correct behavior, because when the docker service is stopped, the Unix socket is destroyed and that is why we could see the `FileNotFoundError(2, 'No such file or directory'` exception in the syslog.

##### Work item tracking

#### How I did it
Change the log severity to the warning and changed the return value.

#### How to verify it
It is really hard to catch the exact moment described in the `Why I did it` section.
In order to check the logic:
1. Change the Unix socket path to non-existing in [/usr/bin/memory_checker](https://github.com/sonic-net/sonic-buildimage/blob/47742dfc2c0d1fa27198d69c9183ddc044e11b22/files/image_config/monit/memory_checker#L139) file on the switch.
2. Execute the `root@sonic:/home/admin# monit restart container_memory_telemetry`
3. Check the syslog for such messages:
```
WARNING memory_checker: Failed to retrieve the running container list from docker daemon! Error message is: 'Error while fetching server API version: ('Connection aborte
d.', FileNotFoundError(2, 'No such file or directory'))'

INFO memory_checker: [memory_checker] Exits without checking memory usage since container 'telemetry' is not running!
```

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [x] 202205
- [x] 202211
- [x] 202305

- [ ] 
- [ ]

### DIFF
--- a/files/image_config/monit/memory_checker
+++ b/files/image_config/monit/memory_checker
@@ -140,10 +140,10 @@ def get_running_container_names():
         running_container_list = docker_client.containers.list(filters={"status": "running"})
         running_container_names = [ container.name for container in running_container_list ]
     except (docker.errors.APIError, docker.errors.DockerException) as err:
-        syslog.syslog(syslog.LOG_ERR,
+        syslog.syslog(syslog.LOG_WARNING,
                       "Failed to retrieve the running container list from docker daemon! Error message is: '{}'"
                       .format(err))
-        sys.exit(5)
+        return []
 
     return running_container_names
 


vadymhlushko-mlnx:
@yozhao101 could you please take a look at this PR?

vadymhlushko-mlnx:
@yozhao101 kind reminder

liat-grozovik:
@qiluo-msft @yozhao101 could you please help to review? this is a fix following a PR merged/reviewed by you so are the optimal reviewer for this one

files/image_config/monit/memory_checker
```diff
@@ -140,10 +140,10 @@ def get_running_container_names():
         running_container_list = docker_client.containers.list(filters={"status": "running"})
         running_container_names = [ container.name for container in running_container_list ]
     except (docker.errors.APIError, docker.errors.DockerException) as err:
-        syslog.syslog(syslog.LOG_ERR,
```
qiluo-msft, line 143: >LOG_ERR [](http://example.com/codeflow?start=29&length=7)

Is it possible to check first docker service is running, if Failed to retrieve the running container list, treat it still as ERR?

vadymhlushko-mlnx: That is the idea of `get_running_container_names()`, if the `docker service is not running it just gives us a warning message to the syslog and returns the empty list of docker containers.

qiluo-msft: To clarify:
1. check docker service is running, if not running, no ERR, no WARNING. Low frequent INFO may be good.
2. if docker service is running, retrieve the running container list. If failed to retrieve, treat it still as ERR.

vadymhlushko-mlnx: We already have the check if the `docker service is running, but in some cases (which I described in the PR description) is not enough.
This check is right before the `get_running_container_names()` function call.

qiluo-msft: Thanks for the explanation! Let me update my proposal:
1. when you catch an exception, if you are sure docker service is not running, log INFO in low frequency
2. if not a docker service issue, something really bad happens, let's still use ERR.

files/image_config/monit/memory_checker
```diff
@@ -140,7 +140,12 @@ def get_running_container_names():
...
+        if "('Connection aborted.', FileNotFoundError(2, 'No such file or directory'))" in str(err):
```
nazariig, line 143: @vadymhlushko-mlnx IMHO, better to switch to `systemd` direct calls or use `pystemd`

vadymhlushko-mlnx: Alinged

qiluo-msft (approved):

PR #16090, opened 2023-08-10
[Mellanox] Enhance FW upgrade mechanism

stephenxs:
#### Why I did it

1. Enhance the diagnosis information collecting mechanism
   - If the option `-v` is fed, it will pass additional diagnosis flags to mlxfwmanager
   - Collect all the output from mlxfwmanager and print them to syslog if it fails
2. Abort syncd in case waiting for device or upgrading firmware fails

##### Work item tracking

#### How to verify it

Regression and manual test

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/platform/mellanox/mlnx-fw-upgrade.j2
+++ b/platform/mellanox/mlnx-fw-upgrade.j2
@@ -162,7 +162,7 @@ function WaitForDevice() {
     local -i QUERY_RETRY_COUNT_MAX="10"
     local -i QUERY_RETRY_COUNT="0"
 
-    ${MFT_DIAGNOSIS_FLAGS} ${QUERY_CMD} > /dev/null
+    ${QUERY_CMD} > /dev/null
 
     while [[ ("${QUERY_RETRY_COUNT}" -lt "${QUERY_RETRY_COUNT_MAX}") && ("$?" -ne "${EXIT_SUCCESS}") ]]; do
         sleep 1s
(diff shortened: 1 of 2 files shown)


platform/mellanox/mlnx-fw-upgrade.j2
```diff
@@ -165,8 +167,16 @@ function WaitForDevice() {
     while [[ ("${QUERY_RETRY_COUNT}" -lt "${QUERY_RETRY_COUNT_MAX}") && ("$?" -ne "${EXIT_SUCCESS}") ]]; do
         sleep 1s
         ((QUERY_RETRY_COUNT++))
-        ${QUERY_CMD} > /dev/null
+        output=$(eval ${MFT_DIAGNOSIS_FLAGS} ${QUERY_CMD}) > /dev/null
```
qiluo-msft, line 170: >eval [](http://example.com/codeflow?start=17&length=4)

`eval` is dangerous, even this case there is no malicious content in variables. Could you try prevent `eval` usage?

stephenxs: Hi
We use it because we need to execute the environment variables along with the command. Without it, the environment variable can be treated as command, resulting in the following error
```
Fri 25 Aug 2023 10:46:38 AM IDT Rebooting with platform x86_64-mlnx_msn3420-r0 specific tool ...
/usr/bin/mlnx-fw-upgrade.sh: line 153: FLASH_ACCESS_DEBUG=1: command not found
/usr/bin/mlnx-fw-upgrade.sh: line 158: FLASH_ACCESS_DEBUG=1: command not found

```

PR #16115, opened 2023-08-11
[multi-asic] sudo not required explicitly as /bin/ip netns identify present in sudoers

judyjoseph:
#### Why I did it
Few commands in multiasic platforms when run with the "sudo ip netns exec asic0 <show cmd>" option was taking like 15 mins to get the o/p. This behavior of sudo getting hung was seen by just doing this 

jujoseph@svcstr-server-2:~ sudo ip netns exec asic0 bash
jujoseph@svcstr-server-2:~ sudo ls

##### Work item tracking

#### How I did it

The reason was found that when a show command is issued, there is an API to get the current namespace 
```
def get_current_namespace(pid=None):
154     """
155     This API returns the network namespace in which it is
156     invoked. In case of global namepace the API returns None
157     """
158 
159     net_namespace = None
160     command = ['sudo', '/bin/ip', 'netns', 'identify', "{}".format(os.getpid() if not pid else pid)]
161     proc = subprocess.Popen(command,
162                             stdout=subprocess.PIPE,
163                             universal_newlines=True,
164                             stderr=subprocess.STDOUT)
```
and  we were doing it  with sufo prefixed. This was ok with usually show commands which is issued in linux host, but when issued with "sudo ip netns exec asic0" prefix ( i.e the command is executed in a linux network namespace )  sudo gets hung adding to overall delay to command o/p

Ideally sudo is not needed as we have /bin/ip netns identify present in /etc/sudoers file. Hence removing it

#### How to verify it

Verified with both RO and RW user in a multi-asic platform, confirmed the show commands work, and "sudo ip netns exec asic0 <show cmd>" command returns result quicker.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/src/sonic-py-common/sonic_py_common/multi_asic.py
+++ b/src/sonic-py-common/sonic_py_common/multi_asic.py
@@ -157,7 +157,7 @@ def get_current_namespace(pid=None):
     """
 
     net_namespace = None
-    command = ["sudo", '/bin/ip', 'netns', 'identify', "{}".format(os.getpid() if not pid else pid)]
+    command = ['/bin/ip', 'netns', 'identify', "{}".format(os.getpid() if not pid else pid)]
     proc = subprocess.Popen(command,
                             stdout=subprocess.PIPE,
                             universal_newlines=True,


judyjoseph:
To add more details 

Have tested this change  **with/without** RW Jit, show commands are ok in multi-asic.
 
Checking why the sudo was added earlier : it was part of [ Pull Request #6100](https://github.com/sonic-net/sonic-buildimage/pull/6100) -- related to process_checker and monit. I see that currently we don't use the process checker, we use the supervisorctl  proc_listener to find a critical process exit.

arlakshm (approved):

rlhui:
@judyjoseph , please add ado, thanks.

gechiang:
In the interest of time, I created the MAFT ADO...
MSFT ADO: 24914440

gechiang:
@yxieca , @StormLiangMS please help cherry-pick to back port this fix to the corresponding releases.
Thanks!

judyjoseph:
> 24914440

Thank you @gechiang

PR #16116, opened 2023-08-11
Run db_migrator for non first-time reboots

vivekrnv:
#### Why I did it

The recent change https://github.com/sonic-net/sonic-buildimage/pull/15685#discussion_r1268400869 removed the db migration for non first reboots. 

This is problematic for many deployments which doesn't rely on ZTP and push a custom config_db.json

Port to older branches after https://github.com/sonic-net/sonic-buildimage/pull/15685 is ported back

##### Work item tracking

#### How I did it

Re-introduce the logic to run the db_migrator on non-first boots

#### How to verify it

Verified reboot and warm-reboot cases

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]


dgsudharsan (approved):

files/build_templates/docker_image_ctl.j2
```diff
@@ -259,6 +259,16 @@ function postStartAction()
             # This flag will be set to "1" after DB migration/initialization is completed as part of config-setup
             $SONIC_DB_CLI CONFIG_DB SET "CONFIG_DB_INITIALIZED" "0"
         else
+            if [[ "$BOOT_TYPE" != "warm" ]]; then
```
vaibhavhd, line 262: Why is this setting needed now, when it was not needed earlier? And, why is warm-reboot an exception?

files/build_templates/docker_image_ctl.j2
```diff
@@ -259,6 +259,16 @@ function postStartAction()
...
+                $SONIC_DB_CLI CONFIG_DB SET "CONFIG_DB_INITIALIZED" "0"
+            fi
+            # this is not a first time boot to a new image. Datbase container starts w/ old pre-existing config
+            if [[ -x /usr/local/bin/db_migrator.py ]]; then
```
vaibhavhd, line 266: I am thinking if we not follow the same flow as we had earlier. Rather than that, why not confine all the migration logic in one place - config-setup.service?

This would ensure that migration also happens after the DB is loaded w/ relevant config - from minigraph, old config, new pushed config, etc. 

Specifically, I am asking that this db_migrator call be moved here as part of new `else` block:
https://github.com/sonic-net/sonic-buildimage/blob/c05af46d02db315b3354a6f88eef45d49ff85b08/files/image_config/config-setup/config-setup#L370

and to https://github.com/sonic-net/sonic-buildimage/blob/c05af46d02db315b3354a6f88eef45d49ff85b08/files/image_config/config-setup/config-setup#L382

files/build_templates/docker_image_ctl.j2
vivekrnv: Because db_migrator busy waits when the CONFIG_DB_INITIALIZED flag is not set. Probably because of ConfigDBConnector https://github.com/sonic-net/sonic-swss-common/blob/master/common/configdb.cpp#L29.

I've only checked for warm-reboot since warm-reboot might already contain the  CONFIG_DB_INITIALIZED from the previous db.

I feel its better to update db_migrator to not depend on CONFIG_DB_INIT, what do you think?

files/build_templates/docker_image_ctl.j2
vivekrnv: We have to run it here since the config_db.json has to be loaded first. If we move this to config-setup, this can't be guaranteed.

https://github.com/sonic-net/sonic-buildimage/pull/16116/files#diff-130e5ab75471398db06b73a4c6e4b56517f42e3c1f17870b5309620fc0e2fafbR247

vivekrnv:
@vaibhavhd Handled the comments. the db_migrator fix is actually not required given we are setting the INIT flag to 0 in db. service. I will fix the db_migrator script later but since that is not required here i don't want to tie that optional fix with this PR.

vaibhavhd (approved):

vivekrnv:
Failing in kvmtest-multi-asic-t1-lag. i.e. on masic platforms. Will move the migration back to database.service.

liat-grozovik (approved):

dgsudharsan:
@yxieca Can you please help with the cherry-pick to 202205?

PR #16160, opened 2023-08-15
reset-factory more robust with exec reboot

MichaelWangSmci:
#### Why I did it
reset-factory dose not flush config_db.json when executed from a console terminal.

##### Work item tracking
- Microsoft ADO **(number only)**: None

#### How I did it
Replace "reboot" with "exec reboot" in the reset-factory script.

#### How to verify it
Create a VLAN and save the config.
Execute reset-factory from a console terminal.
After reboot check the VLAN should not exist.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [x] master 

reset-factory more robust with exec reboot


lguohan:
what is the difference between exec reboot and reboot. the explanation is not quite clear.

MichaelWangSmci:
> what is the difference between exec reboot and reboot. the explanation is not quite clear.

I found that if run "reset-factory" in a console terminal, then after reboot the config_db.json is not reset.
And if replace "reboot" with "exec reboot" in "reset-factory", then config_db.json is reset as expected.
I guess this is because "reset-factory" calls "reboot" which is actually "/usr/local/bin/reboot", and it uses exec to do reboot, makes "reset-factory" lost it's change on config_db.json.
"exec reboot" can trigger config_db.json flushed to disk when the process is replaced with "/usr/local/bin/reboot".

PR #16173, opened 2023-08-16
[Arista] Remove one pcie device from pcie.yaml accross platforms

Staphylo:
#### Why I did it

These devices will not reliabily report the proper devid and vendorid when reading it is read directly from the pci config space. It can be read but shouldn't be compared against some fixed value like the one stored in pcie.yaml.

##### Work item tracking

#### How I did it

Since this makes pcied unhappy, the simplest path forward is to just remove this device from monitoring.

#### How to verify it

The pcied periodic read from pci config space no longer complains.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [x] 202205
- [x] 202211
- [x] 202305

Remove one pcie device from pcie.yaml accross Arista platforms


prgeor:
@Staphylo can we say that SCD device in question is not actually PCIe device since these are sitting on LPC bus?

prgeor (approved):

prgeor:
@yxieca please help merge. the build failure is not related to this change.

lguohan (approved):

PR #16213, opened 2023-08-19
[chassis] Chassis DB cleanup when asic comes up

vganesan-nokia:
#### Why I did it

In an operational system, after some configuration is changed (that involves line cards generating entries to chassis db), if config reload or system reboot is done without saving the new configuration changes and when system comes up,  entries created by the new config are still present in the chassis db and hence the corresponding voq system entries (such as system interface, system neighbor and so on ) in all other line cards. These stale entries may affect the accuracy of the current entries and hence the intended operation of chassis. To fix this, we cleanup the chassis db when the an asic comes up. The chassis db is cleaned up for entries created by the asic that is coming up.

##### Work item tracking

N/A

#### How I did it

Cleanup the entries from the following tables in chassis app db in redis_chassis server in the supervisor

(1) SYSTEM_NEIGH
(2) SYSTEM_INTERFACE
(3) SYSTEM_LAG_MEMBER_TABLE
(4) SYSTEM_LAG_TABLE

As part of the clean up only those entries created by the asic that is coming up are deleted. The LAG IDs used by the asics are also de-allocated from SYSTEM_LAG_ID_TABLE and SYSTEM_LAG_ID_SET

#### How to verify it

- When chassis is up and running, change configuration that will create entries for the tables mentioned above. Make sure that the entries created are due to changes in the existing configuration.
- Note that chassis db has entries for new changed configutations.
- Without saving the changed configuration, do a config reload or reboot the line card or reboot the chassis.
- Observe that the chassis db does not have the entries created by the changed configuration.

N/A

master

- Changes swss.sh service script. Which is invoked with swss service is started. After cleaning up the critical tables in local redis database, the chassis db entries are cleaned up. The clean up use the hostname and asic name currently detected by the service initialization script. If this hostname and asic name are different than those used to create the entries in the chassis db, the entries will not be cleaned up.

None

### DIFF
--- a/files/scripts/swss.sh
+++ b/files/scripts/swss.sh
@@ -107,6 +107,92 @@ function clean_up_tables()
     end" 0
 }
 
+# This function cleans up the chassis db table entries created ONLY by this asic
+# This is used to do the clean up operation when the line card / asic reboots
+# When the asic/lc is RE-booting, the chassis db server is supposed to be running 
+# in the supervisor.  So the clean up is done when only the chassis db connectable. 
+# Otherwise no need to do the clean up since both the supervisor and line card may be 
+# rebooting (the whole chassis scenario)
+# The clean up operation is required to delete only those entries created by
+# the asic that is rebooted. Entries from the following tables are deleted in the order
+# given below
+#   (1) SYSTEM_NEIGH
+#   (2) SYSTEM_INTERFACE
+#   (3) SYSTEM_LAG_MEMBER_TABLE
+#   (4) SYSTEM_LAG_TABLE
+#   (5) The corresponding LAG IDs of the entries from SYSTEM_LAG_TABLE
+#       SYSTEM_LAG_ID_TABLE and SYSTEM_LAG_ID_SET are adjusted appropriately
+function clean_up_chassis_db_tables()
+{
+    if [[ !($($SONIC_DB_CLI CHASSIS_APP_DB PING | grep -c True) -gt 0) ]]; then
+        return
+    fi
+
+    lc=`$SONIC_DB_CLI CONFIG_DB  hget 'DEVICE_METADATA|localhost' 'hostname'`
+    asic=`$SONIC_DB_CLI CONFIG_DB  hget 'DEVICE_METADATA|localhost' 'asic_name'`
+
+    # First, delete SYSTEM_NEIGH entries
+    $SONIC_DB_CLI CHASSIS_APP_DB EVAL "
+    local host = string.gsub(ARGV[1], '%-', '%%-')
+    local dev = ARGV[2]
+    local ps = 'SYSTEM_NEIGH*|' .. host .. '|' .. dev
+    local keylist = redis.call('KEYS', 'SYSTEM_NEIGH*')
+    for j,key in ipairs(keylist) do
+        if string.match(key, ps) ~= nil then
+            redis.call('DEL', key)
+        end
+    end
+    return " 0 $lc $asic
+
+    sleep 30
+
+    # Next, delete SYSTEM_INTERFACE entries
+    $SONIC_DB_CLI CHASSIS_APP_DB EVAL "
+    local host = string.gsub(ARGV[1], '%-', '%%-')
+    local dev = ARGV[2]
+    local ps = 'SYSTEM_INTERFACE*|' .. host .. '|' .. dev
+    local keylist = redis.call('KEYS', 'SYSTEM_INTERFACE*')
+    for j,key in ipairs(keylist) do
+        if string.match(key, ps) ~= nil then
+            redis.call('DEL', key)
+        end
+    end
+    return " 0 $lc $asic
+
+    # Next, delete SYSTEM_LAG_MEMBER_TABLE entries
+    $SONIC_DB_CLI CHASSIS_APP_DB EVAL "
+    local host = string.gsub(ARGV[1], '%-', '%%-')
+    local dev = ARGV[2]
+    local ps = 'SYSTEM_LAG_MEMBER_TABLE*|' .. host .. '|' .. dev
+    local keylist = redis.call('KEYS', 'SYSTEM_LAG_MEMBER_TABLE*')
+    for j,key in ipairs(keylist) do
+        if string.match(key, ps) ~= nil then
+            redis.call('DEL', key)
+        end
+    end
+    return " 0 $lc $asic
+
+    sleep 15
+
+    # Finally, delete SYSTEM_LAG_TABLE entries and deallot LAG IDs
+    $SONIC_DB_CLI CHASSIS_APP_DB EVAL "
+    local host = string.gsub(ARGV[1], '%-', '%%-')
+    local dev = ARGV[2]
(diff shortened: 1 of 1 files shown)


files/scripts/swss.sh
```diff
@@ -107,6 +107,92 @@ function clean_up_tables()
...
+    local ps = 'SYSTEM_LAG_MEMBER_TABLE*|' .. host .. '|' .. dev
+    local keylist = redis.call('KEYS', 'SYSTEM_LAG_MEMBER_TABLE*')
+    for j,key in ipairs(keylist) do
+        if string.match(key, ps) ~= nil then
+            redis.call('DEL', key)
+        end
+    end
+    return " 0 $lc $asic
+
+    sleep 15
```
gechiang, line 175: Same question as above... can we eliminate unnecessary sleep if it is not really needed...

files/scripts/swss.sh
```diff
@@ -107,6 +107,92 @@ function clean_up_tables()
...
+    local ps = 'SYSTEM_NEIGH*|' .. host .. '|' .. dev
+    local keylist = redis.call('KEYS', 'SYSTEM_NEIGH*')
+    for j,key in ipairs(keylist) do
+        if string.match(key, ps) ~= nil then
+            redis.call('DEL', key)
+        end
+    end
+    return " 0 $lc $asic
+
+    sleep 30
```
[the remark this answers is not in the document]
vganesan-nokia: This sleep time is required. We need to make sure that neighbors associated with a system interface is removed in the syncd/SAI before deleting the system interface. If this delay was not there, we observed an issue that at the orchangent system interface refcount will be zero but the syncd meta layer will stil have non-zero.  Due to this out of sync issue which is in turn diue to time difference between the orchangent and syncd,  syncd returns error and hence orchagent exits. There is no specific criteria used to select this value. This is thie lowest rounded time we used with which we did not see the issue.

files/scripts/swss.sh
vganesan-nokia: This is for the similar reason as explained for the previous comment but for SYSTEM_LAG_TABLE and SYSTEM_LAG_MEMBER_TABLE entries. At the line card/asic, deletion of system lag will  happen only if all the members of the system lag are removed. Even though we send  system lag member delete before sending the system lag delete, due to the order of processing and other dependency (like system inteface delete and neighbours delete which introduces timing differences), by the time system lag is tried to be deleted, sometimes, the a system lag still has system interface pending to be deleted. Having this delay (the lowest rounded time we tested with which we did not see the issue) helps to make sure that all system lag members are deleted before deleting system lag

files/scripts/swss.sh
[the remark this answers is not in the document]
gechiang: Thanks for the explanation.  Please add a comment to disclose this so that in future someone may not try to remove it or at least when try to better tunning the value will know what to verify so it will not end up breaking something...

files/scripts/swss.sh
gechiang: Thanks for the explanation.  Please add a comment to disclose this so that in future someone may not try to remove it or at least when try to better tunning the value will know what to verify so it will not end up breaking something...

files/scripts/swss.sh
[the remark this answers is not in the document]
vganesan-nokia: Added comment.

files/scripts/swss.sh
vganesan-nokia: Added comment

gechiang (approved):

deepak-singhal0408 (approved):
Changes look good to me.. Thanks

files/scripts/swss.sh
```diff
@@ -107,6 +107,100 @@ function clean_up_tables()
...
+#   (3) SYSTEM_LAG_MEMBER_TABLE
+#   (4) SYSTEM_LAG_TABLE
+#   (5) The corresponding LAG IDs of the entries from SYSTEM_LAG_TABLE
+#       SYSTEM_LAG_ID_TABLE and SYSTEM_LAG_ID_SET are adjusted appropriately
+function clean_up_chassis_db_tables()
+{
+    if [[ !($($SONIC_DB_CLI CHASSIS_APP_DB PING | grep -c True) -gt 0) ]]; then
+        return
+    fi
+
```
arlakshm, line 130: can add a check to do the clean-up for VoQ Linecards. The code as it is now will add delay of about 45 seconds on supervisor and non-voq linecards.

files/scripts/swss.sh
```diff
@@ -107,6 +107,100 @@ function clean_up_tables()
...
+        end
+    end
+    return " 0 $lc $asic
+
+    # Wait for some time before deleting system interface so that the system interface's "object in use"
+    # is cleared in both orchangent and in syncd. Without this delay, the orchagent clears the refcount
+    # but the syncd (meta) still has no-zero refcount. Because of this, orchagent gets "object still in use"
+    # error and aborts.
+
+    sleep 30
```
judyjoseph, line 152: Can we also add a check to see if the number of SYSTEM_NEIGHBOR entries ( and similarly below for SYSTEM_LAG_MEMBER_TABLE) **are non-zero** before adding this sleep(). In case of retries when swss systemd service restarts multiple times etc we need not each time wait for 45 sec.

files/scripts/swss.sh
vganesan-nokia: Added check to run the clean up only for voq switches.

files/scripts/swss.sh
vganesan-nokia: The problem is not with number of system neighbors or system lag members in the chassis app db. The problem is with timing in processing in other remote asics and linecards. It depends on how much time the remote asics/linecards take to remove the referenced entries in that asic/linecard.

judyjoseph: That is true, I was pointing to a few cases as below where we can skip the sleep altogether 

(a) no SYSTEM_NEIGHBORS/LAG for that asic/LC ( so ideally nothing to clean locally and remote)
or in scenario below 
(b) where swss restarts, we check and cleanup SYSTEM_NEIGHBORS for that asic/LC in chassis_db. Wait for 30 secs and make sure that the processing in remote asic/linecard is done.
   > But not for some reason swss restarts again (assume orchgent crashed). Again we will do same cleanup SYSTEM_NEIGHBORS for that asic/LC in chassis_db -- but this time there are no entries as in last round we cleaned up all .
  > So now there is no need to wait for 30 sec and 15 sec as there is no SYSTEM_NEIGHBOR/LAG entries relating to my LC/asic present in CHASSIS_DB

But don't block merge for this comment -- we can fine tune this in subsequent commits too

arlakshm (approved):

gechiang:
BRM build failed... retrying the pipeline again...
Azure Pipelines / Azure.sonic-buildimage

files/scripts/swss.sh
vganesan-nokia: Thanks for the clarification. I'll fix this.

gechiang (approved):

gechiang:
@yxieca , @StormLiangMS 
MSFT ADO: 25038392
Appreciate approval for the corresponding branches.  Thanks!

PR #16230, opened 2023-08-22
Supports FRR-VRRP configuration

philo-micas:
##### Work item tracking

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/files/image_config/interfaces/interfaces.j2
+++ b/files/image_config/interfaces/interfaces.j2
@@ -79,21 +79,29 @@ iface {{ name }} {{ 'inet' if prefix | ipv4 else 'inet6' }} static
 {%     set vrf_table = '5000' %}
     vrf mgmt
 {% endif %}
+{% set force_mgmt_route_priority = 32764 %}
     ########## management network policy routing rules
     # management port up rules
     up ip {{ '-4' if prefix | ipv4 else '-6' }} route add default via {{ MGMT_INTERFACE[(name, prefix)]['gwaddr'] }} dev {{ name }} table {{ vrf_table }} metric 201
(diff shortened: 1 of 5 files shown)


src/sonic-yang-models/yang-models/sonic-vrrp.yang
```diff
@@ -0,0 +1,111 @@
...
+        container VRRP {
+
+            description "VRRP part of configdb.json"
+
+            list VRRP_LIST {
+
+                key "interface_name vrid";
+
+                leaf interface_name {
+                    type union {
```
prvattem, line 33: Please leafref to sonic-interface, sonic-port-channel, sonic-vlan

philo-micas: ok.

dockers/docker-fpm-frr/frr/supervisord/critical_processes.j2
```diff
@@ -1,6 +1,9 @@
 program:zebra
 program:staticd
 program:bgpd
+program:vrrpd
```
lguohan, line 4: what is the communication channel between vrrpd and zebra, I think we need to put vrrpd into a seperate container, not merging with bgp container. the vrrpd needs to enable/disable separately as feature level. this setup does not allow us to achieve that.

lguohan (changes requested):
put vrrpd into a separate container.

PR #16237, opened 2023-08-22
[SNMP]: Fix SNMP mgmt reachability when mgmt IP is obtained via DHCP

SuvarnaMeenakshi:
#### Why I did it
fixes https://github.com/sonic-net/sonic-buildimage/issues/16165

https://github.com/sonic-net/sonic-buildimage/pull/15487 made changes to ensure that SNMP works over IPv6 mgmt and loopback if snmpagentaddress configuration is not present in config_db. This change checks for MGMT_INTERFACE and LOOPBACK_INTERFACE tables in config_db and snmpd starts to listen on those IPs.
With this change, if MGMT_INTERFACE is not present in config_db and is configured via DHCP, snmpd will start to listen only on Loopback0 IP. So snmp query over DHCP configured mgmt IP does not work.

##### Work item tracking
24914478

#### How I did it
Check if both Loopback and MGMT interface configurations are present in config_db to be used as snmpagentaddress. If both are not present, fall to listening on any IP.
new logic:
For single-asic platform:
- if SNMP_AGENT_ADDRESS_CONFIG configuration is present, use that as agentaddress
- elif MGMT_INTERFACE and LOOPBACK_INTERFACE are configured use that as agentaddress
- else listen on any IP.

Caveat: If DHCP is used to configure MGMT interface, then snmpd will listen on any IP and SNMP IPv6 reachability will not work.
Workaround will be to use SNMP_AGENT_ADDRESS_CONFIG  to specifically configure specific IPs for snmpd to listen on.

#### How to verify it
Verified change on 202211 single asic VS image.
```
admin@vlab-01:~$ show ver

SONiC Software Version: SONiC.202211.342905-adb43ff1f
SONiC OS Version: 11
Distribution: Debian 11.7
Kernel: 5.10.0-18-2-amd64
Build commit: adb43ff1f
Build date: Sun Aug 20 15:25:16 UTC 2023
Built by: AzDevOps@vmss-soni001TDF
admin@vlab-01:~$ sudo netstat -tulnp | grep 161
tcp        0      0 127.0.0.1:3161          0.0.0.0:*               LISTEN      48242/snmpd         
udp        0      0 10.1.0.32:161           0.0.0.0:*                           48242/snmpd         
udp        0      0 10.250.0.101:161        0.0.0.0:*                           48242/snmpd         
udp6       0      0 fc00:1::32:161          :::*                                48242/snmpd         
udp6       0      0 fec0::ffff:afa:1:161    :::*                                48242/snmpd
admin@vlab-01:~$ docker exec -it snmp bash
root@vlab-01:/# cat /etc/snmp/snmpd.conf | grep agentAddress
agentAddress udp:[10.250.0.101]:161
agentAddress udp6:[fec0::ffff:afa:1]:161
agentAddress udp:[10.1.0.32]:161
agentAddress udp6:[fc00:1::32]:161
root@vlab-01:/# 
```

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/dockers/docker-snmp/snmpd.conf.j2
+++ b/dockers/docker-snmp/snmpd.conf.j2
@@ -29,12 +29,12 @@
 agentAddress {{ protocol(agentip) }}:[{{ agentip }}]{% if port %}:{{ port }}{% endif %}{% if vrf %}%{{ vrf }}{% endif %}{{ "" }}
 {% endfor %}
 {% elif NAMESPACE_COUNT is not defined or NAMESPACE_COUNT|int <= 1 %}
-{% if MGMT_INTERFACE is defined and LOOPBACK_INTERFACE is defined%}
+{% if MGMT_INTERFACE is defined and LOOPBACK_INTERFACE is defined %}
 {% for intf, ip in MGMT_INTERFACE %}
 {% set agentip = ip.split('/')[0]|lower %}
 {% set zoneid = '' %}
-# Use interface as zoneid for link local ipv6
 {% if agentip.startswith('fe80') %}
+# Use interface as zoneid for link local ipv6
 {% set zoneid = '%' + intf %}
 {% endif %}
 agentAddress {{ protocol(agentip) }}:[{{ agentip }}{{ zoneid }}]:161
@@ -44,8 +44,8 @@ agentAddress {{ protocol(agentip) }}:[{{ agentip }}{{ zoneid }}]:161
 {% set intf = lo[0] %}
 {% set agentip = lo[1].split('/')[0]|lower %}
 {% set zoneid = '' %}
-# Use interface as zoneid for link local ipv6
 {% if agentip.startswith('fe80') %}
+# Use interface as zoneid for link local ipv6
 {% set zoneid = '%' + intf %}
 {% endif %}
 agentAddress {{ protocol(agentip) }}:[{{ agentip }}{{ zoneid }}]:161
--- a/dockers/docker-snmp/snmpd.conf.j2
+++ b/dockers/docker-snmp/snmpd.conf.j2
@@ -29,7 +29,7 @@
 agentAddress {{ protocol(agentip) }}:[{{ agentip }}]{% if port %}:{{ port }}{% endif %}{% if vrf %}%{{ vrf }}{% endif %}{{ "" }}
 {% endfor %}
 {% elif NAMESPACE_COUNT is not defined or NAMESPACE_COUNT|int <= 1 %}
-{% if MGMT_INTERFACE is defined %}
+{% if MGMT_INTERFACE is defined and LOOPBACK_INTERFACE is defined%}
 {% for intf, ip in MGMT_INTERFACE %}
 {% set agentip = ip.split('/')[0]|lower %}
 {% set zoneid = '' %}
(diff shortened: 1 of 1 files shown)


dockers/docker-snmp/snmpd.conf.j2
```diff
@@ -29,25 +29,23 @@
 agentAddress {{ protocol(agentip) }}:[{{ agentip }}]{% if port %}:{{ port }}{% endif %}{% if vrf %}%{{ vrf }}{% endif %}{{ "" }}
 {% endfor %}
 {% elif NAMESPACE_COUNT is not defined or NAMESPACE_COUNT|int <= 1 %}
-{% if MGMT_INTERFACE is defined %}
+{% if MGMT_INTERFACE is defined and LOOPBACK_INTERFACE is defined %}
```
dgsudharsan, line 32: This check will ensure that if mgmt interface and loopback interface both defined, the agentip will be populated. Is this a temporary work around?  Can't we control this agentIP using a specific configuration knob in config_db.json rather than having some arbitrary checks?

dgsudharsan (changes requested):

dockers/docker-snmp/snmpd.conf.j2
SuvarnaMeenakshi: We could populate the SNMP_AGENT_ADDRESS table in config_db and if that is available in config_db that will be used.
This condition is to specifically use if we do not have any AGENT_ADDRESS configuration.
The main issue is that if we listen on any IP, then snmp IPv6 will not work.
Having specific IP configured as agentaddress is to solve this baseline ipv6 issue.

dgsudharsan: The issue here is the design enforces the user to configure both mgmt ip table and loopback interface table together in order to generate the agentIP and overcome this issue.
There can be deployments without mgmt ip table or loopback interface. So this will be a problem there. The solution needs to be generic. 
There is no logical reason to depend on mgmt interface or loopback interface to generate agentip. In my opinion, if we have a configuration option, the user should use it. Generating it implicitly or only in certain conditions seems to be a hack and the user might not be aware of the implications. @qiluo-msft  please comment your feedback here.

SuvarnaMeenakshi: We have plan to update SNMP_AGENT_ADDRESS table when loading configuration from minigraph.
But again, plan is to populate SNMP_AGENT_ADDRESS table with Mgmt and Loopback IP from minigraph and use that in snmpd.conf.
This could be overridden with new configuration as chosen by updating SNMP_AGENT_ADDRESS  in config_db.json or using CLI.

abdosi: > This check will ensure that if mgmt interface and loopback interface both defined, the agentip will be populated. Is this a temporary work around? Can't we control this agentIP using a specific configuration knob in config_db.json rather than having some arbitrary checks?

not sure how this will work ? Do we know DHCP will provide the IP address by the time SNMP starts ? For DHCP Based IP assignment and SNMP to work it would need to listen to 0.0.0.0 for v4 .

dgsudharsan: That's a good question too. However I don't think the solution should be an arbitrary check where if both mgmt and loopback are configured, then we add agentIP to the snmpd.conf.

abdosi: Above changes have made sure for v4/v6:

1> First use SNMP_AGENT_ADDRESS_CONFIG
2> if both MGMT and LOOPBACK then use them
3> else use catch all 0.0.0.0 / ::/0  ==> FOR DHCP case

Above change will make sure for v4 it works for all scenarios and for v6 DHCP base SNMP will not work because of the issue as define here: https://github.com/net-snmp/net-snmp/issues/630

SuvarnaMeenakshi: @dgsudharsan @moshemos Is it possible to add support in SNMP to listen to changes in mgmt ip obtaining from DHCP and restart if there is any change? Is there any SONiC table that gets change in mgmt IP due to DHCP?

PR #16248, opened 2023-08-23
Adjusting docker creation to include ports forwarding and docker network type

azmy98:
#### Why I did it
To support docker network configuration and support port forwarding from values in the manifest file

#### How I did it
update the docker_image_ctl.j2 file

#### How to verify it
In order to verify this feature, we need to have the complementary part in the sonic-utilities repo
and then need to install a docker image file with manifest that includes ports in it and network type as well using sonic-package-manager and validate that they are as expected using docker commands such as: 'docker ps' and 'docker inspect' commands 

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/files/build_templates/docker_image_ctl.j2
+++ b/files/build_templates/docker_image_ctl.j2
@@ -540,7 +540,9 @@ start() {
     # TODO: Mellanox will remove the --tmpfs exception after SDK socket path changed in new SDK version
 {%- endif %}
     docker create {{docker_image_run_opt}} \
+{%- if '--net' not in docker_image_run_opt %}
         --net=$NET \
+{%- endif %}
         -e RUNTIME_OWNER=local \
         --uts=host \{# W/A: this should be set per-docker, for those dockers which really need host's UTS namespace #}
 {%- if install_debug_image == "y" %}


files/build_templates/docker_image_ctl.j2
```diff
@@ -540,7 +540,9 @@ start() {
     # TODO: Mellanox will remove the --tmpfs exception after SDK socket path changed in new SDK version
 {%- endif %}
     docker create {{docker_image_run_opt}} \
+{%- if '--net' not in docker_image_run_opt %}
         --net=$NET \
```
stepanblyschak, line 544: I see the value fo ```$NET``` is ```host``` when the container is multi ASIC agnostic. So, I assume you may want to block having ```asic-service``` set to ```True``` and specifying custom ```network``` in the manifest. Could you please add such check in sonic-utilities PR?

stepanblyschak (changes requested):

lguohan:
what feature requires this, can you explain the motivation?

stepanblyschak (approved):

azmy98:
@lguohan 
> what feature requires this, can you explain the motivation?

the purpose of this change is to be able to support docker creation with specified network type (host, bridge) and the forwarding ports found in the docker manifest when installing docker image from sonic-package-manager. Please check the following PR: https://github.com/sonic-net/sonic-utilities/pull/2952 to understand the whole picture. the manifest attributes will finally be exported to dictionary in docker_image_run_opt including the net type of the docker, so it will help us containerize the docker image and block it from accessing all the host network.

files/build_templates/docker_image_ctl.j2
```diff
@@ -540,7 +540,9 @@ start() {
     # TODO: Mellanox will remove the --tmpfs exception after SDK socket path changed in new SDK version
 {%- endif %}
     docker create {{docker_image_run_opt}} \
+{%- if '--net' not in docker_image_run_opt %}
```
qiluo-msft, line 543: >if [](http://example.com/codeflow?start=4&length=2)

This `if` will possibly override NET related feature in this file. Please unify the solution like docker_image_run_opt and stop using NET.

azmy98: If a docker does not have network attribute, he will have a default value from this file. I used the` if` statement to make code backward compatible.

azmy98: @qiluo-msft, can you please confirm?

azmy98:
@qiluo-msft, there is a conflict also on this PR:

` docker create {{docker_image_run_opt}} \
<<<<<<< sonic_docker_image_ctl_update
{%- if '--net' not in docker_image_run_opt %}
=======
{%- if docker_container_name != "dhcp_server" %}
>>>>>>> master
        --net=$NET \
{%- endif %}`

can you please direct me to who wrote this new code:

`{%- if docker_container_name != "dhcp_server" %}`

since I want to know how to reslove this conflict.

qiluo-msft (approved):

fastiuk:
@qiluo-msft , could you please approve this PR? It was approved

fastiuk:
@qiluo-msft gentle reminder to merge this PR

PR #16297, opened 2023-08-27
[device] Adding platform support for Accton as5912-54x

fischerdouglas:
#### Why I did it
Add device and platform files for Accton as5912-54 - AGR100.

##### Work item tracking
No tracking yet.

#### How I did it
Add as SONiC Porting Guide.

#### How to verify it
I sincerely don't know!
Still a work in progress.

Until now I'm guessing no backport will be needed.

No tests yet.

Add device and platform files for Accton as5912-54 - AGR100.

No module changes were made.


fischerdouglas:
I'm not a programmer, and I don't really know what I'm doing.

But I used your PR for AS7315 (which is also Qumran), and decided to give it a try.

I have no idea how lanes are defined for interfaces in port_config.ini. I basically copied it from another model and it seemed to make sense.
I also added a 6th fan in sensors.conf, but I don't know if that makes sense.

@roylee123 could you take a look and see if what's been done here is a far cry from what really needs to be done?

akenliu:
AS7315 and AS5912 are two different devices, the hardware layout are also difference.
You should not copy the files from AS7315 and rename all as7315 to as5912, it is wrong.
I think you also did not test your codes on real AS5912.
If you buy the device, AS5912-54x, from Edgecore, I think you can contact with Edgecore.

fischerdouglas:
Hello @akenliu
> it is wrong.
I Actually know that it is "wrong"... But I needed to start somewhere.
Sorry if I exceeded the limit requesting PR.

>  I think you also did not test your codes on real AS5912.
Yep, I did not test it on any real device. Exactly because I knew that something would be wrong.
But we have 18 of those and we want to migrate to SONiC.

I need to reinforce the "sorry" part.
I was just following the docs on [Porting-Guide](https://github.com/sonic-net/SONiC/wiki/Porting-Guide).

>I think you can contact with Edgecore support website.
Well... I made some contacts in the past.
And at that moment they basically limited themselves to answering that such equipment is not supported in SONiC.
Maybe now the conversation with them can be more productive.

P.S.:
I felt that the orientations on Porting-Guide could be improved.
Maybe some fluorograms and more detailed instructions on the modules part.
Fans, BCM, and especially the Lanes mapping to the interfaces.
I'm just starting... But I will make an effort on that.

akenliu:
> I Actually know that it is "wrong"... But I needed to start somewhere. Sorry if I exceeded the limit requesting PR.
> 
> > I think you also did not test your codes on real AS5912.
> 
> Yep, I did not test it on any real device. Exactly because I knew that something would be wrong. But we have 18 of those and we want to migrate to SONiC.
> 
> I need to reinforce the "sorry" part. I was just following the docs on [Porting-Guide](https://github.com/sonic-net/SONiC/wiki/Porting-Guide).

Porting-Guide is a general guideline, it would not describe how to control the hardware components in the switch device. 
I think you understand the lane mapping should be modified, so you ask the questions.
However, the other hardware components are also all different, such as FAN, LED, PSU, thermal.
Therefore, the next step is that you need to understand the hardware layout for each component.
I mean we does not just modify the name/string from AS7315 to AS5912, we need to modify driver codes for real hardware design regarding the FAN, PSU, LED, thermal sensor, transceiver, EERPOM....

Lewis-Kang:
@fischerdouglas Could you please confirm if you will be verifying your PR modifications for as5912-54x?

fischerdouglas:
Hello @Lewis-Kang!

I will not have the opportunity to deal with that in the next few months.
You are probably asking that to remove the tests, right?
Sorry for that!

What are the instructions to remove that from the pipeline tests?

Lewis-Kang:
Hi @fischerdouglas 
I think it's improper to submit a PR  without verifying the code's functionality.

PR #16311, opened 2023-08-29
chassis-packet: Update arp_update script for FAILED and STALE check

anamehra:
Signed-off-by: anamehra anamehra@cisco.com

#### Why I did it
Fixes https://github.com/sonic-net/sonic-buildimage/issues/16312

1. Fixing an issue with FAILED entry resolution retry. Neighbor entries in arp table may sometimes enter a FAILED state when the far end is down and reports the state as follows:
```
2603:10e2:400:3::1 dev PortChannel19 router FAILED
```

While the arp_update script handles the entries for FAILED in the following format, the above was not handled due to the token location (extra router keyword at index 4):
```
2603:10e2:400:3::1 dev PortChannel19 FAILED
```

The former format may appear if an arp resolution is tried on a link that is known but the far end goes down, e.g., pinging a STALE entry while the far end is down.

2. Refreshing STALE entries to make sure the far end is reachable. STALE entries for some backend ports may appear in chassis-packet when no traffic is received for a while on the port. When the far end goes down, it is expected for BFD to stop sending packets on the session for which the far end is not reachable. But as the entry is known as stale, on the Cisco chassis, BFD keeps sending packets. Refreshing the stale entry will keep active links as reachable in the neighbor table while the entries for the far end down will enter a failed state. FAILED state entries will be retired and entered reachable when far end comes back up.

##### Work item tracking

#### How I did it
Modified arp_update state check to use grep for states in neighbor data string place of using the position. If a string contains any of the matching patterns from INCOMPLETE, FAILED or STALE, try ping to resolve.

#### How to verify it

On a chassis-packet Chassis LC:

1. Bring down far end asic or LC.
2. run arp_update on local namespace swss docker
3. check ip netns exec ip -4 neigh show
4. There will be FAILED entries
5. bringup far end and wait for links to be up
6. run arp_update on local namespace swss docker
7. Failed entries for this far end should change to reachable
8. Wait for couple of mins and check for STALE entries.
9. run arp_update and this should resolve STALE entries to reachable

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [x] 202205
- [ ] 202211
- [x] 202305

- [ ] 
- [ ]

### DIFF
--- a/files/scripts/arp_update
+++ b/files/scripts/arp_update
@@ -25,29 +25,35 @@ while /bin/true; do
       for i in ${!STATIC_ROUTE_NEXTHOPS[@]}; do
           nexthop="${STATIC_ROUTE_NEXTHOPS[i]}"
           if [[ $nexthop == *"."* ]]; then
-              neigh_state=( $(ip -4 neigh show | grep -w $nexthop | tr -s ' ' | cut -d ' ' -f 3,4) )
+              neigh_state=$(ip -4 neigh show | grep -w $nexthop | tr -s ' ')
               ping_prefix=ping
           elif [[ $nexthop == *":"* ]] ; then
-              neigh_state=( $(ip -6 neigh show | grep -w $nexthop | tr -s ' ' | cut -d ' ' -f 3,4) )
+              neigh_state=$(ip -6 neigh show | grep -w $nexthop | tr -s ' ')
               ping_prefix=ping6
           fi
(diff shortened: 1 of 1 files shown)


anamehra:
@abdosi , Please review.

abdosi (approved):

prsunny (approved):

abdosi:
@theasianpianist for viz.

gechiang:
@yxieca , @StormLiangMS ,
MSFT ADO: 25087618

anamehra:
> @anamehra PR conflicts with 202211 branch

Its already in 202211 via #16384

PR #16326, opened 2023-08-29
[database] make Redis process runs as non-root user

maipbui:
#### Why I did it
Running the Redis server as the "root" user is not recommended. It is suggested that the server should be operated by a non-privileged user.
##### Work item tracking
- Microsoft ADO **(number only)**: 15895240

#### How I did it
Ensure the Redis process is operating under the 'redis' user in supervisord and make redis user own REDIS_DIR inside db container.
#### How to verify it
Built new image, verify redis process is running as 'redis' user and all containers are up. 

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ]  SONiC.master.0-dirty-20230829.173048
- [ ]

### DIFF
--- a/dockers/docker-database/docker-database-init.sh
+++ b/dockers/docker-database/docker-database-init.sh
@@ -18,6 +18,7 @@ then
 fi
 
 REDIS_DIR=/var/run/redis$NAMESPACE_ID
+chown -R redis:redis $REDIS_DIR
 mkdir -p $REDIS_DIR/sonic-db
 mkdir -p /etc/supervisor/conf.d/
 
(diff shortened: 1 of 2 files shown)


dockers/docker-database/docker-database-init.sh
```diff
@@ -18,6 +18,7 @@ then
 fi
 
 REDIS_DIR=/var/run/redis$NAMESPACE_ID
+chown -R redis:redis $REDIS_DIR
```
qiluo-msft, line 21: >chown [](http://example.com/codeflow?start=0&length=5)

`chown` line is too early, files created later will be owned by root:root.

maipbui: fixed, moved to bottom of file

qiluo-msft (approved):

PR #16434, opened 2023-09-04
[202211][Mellanox] Update SDK/FW to 4.6.1062/2012.1062 Update SDK/FW/SAI to 4.6.1062/2012.1062/SAIBuild2211.25.1.4

dprital:
#### Why I did it

**SAI bug Fixes**
1.	When creating an ACL rule with SAI_ACL_ENTRY_ATTR_FIELD_SRC_IP/SAI_ACL_ENTRY_ATTR_FIELD_DST_IP enabled, and then disabling the field by setting enable=false, a match on L3_type=IPv4 will remain programmed for the rule Issue resolved after the fix
2.	Allow the max scale of virtual routers to be configure for SPC-1, SPC-2, SPC-3  which is 255 when fastboot enable and 511 when fastboot disable
3.	Remove default hash key of SRC_MAC, DST_MAC and ETH_TYPE
 
**SDK/FW bug fixes**
1.	When preforming fast boot from an old SDK version (currently installed) to a newer one (target version), and the system was initially loaded with a new SDK version (past version), and the system has not been wiped, under specific conditions, the fast boot would use the past version's data and may fail.

##### Work item tracking

#### How I did it
* Update SAI version to SAIBuild2211.25.1.4
* Update SDK/FW version to 4.6.1062/2012.1062

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]


platform/mellanox/fw.mk
```diff
@@ -21,7 +21,7 @@ MLNX_FW_BASE_PATH = $(MLNX_SDK_BASE_PATH)
 # Place an URL here to FW if you want to download FW instead
 MLNX_FW_BASE_URL =
 
-SIMX_VERSION = 23.7-1086
+SIMX_VERSION = 23.10-1025
```
keboliu, line 24: SIMX version shouldn't be changed, SONiC is still using 23.7-1086 regardless of SDK working with 23.10-1025.

liat-grozovik (approved):

PR #16452, opened 2023-09-05
[minigraph] remove number of lanes check for changing speed from 400G to 100G and set speed setting before lane reconfiguration 

vdahiya12:
…   (#15721)
cherry pick conflict https://github.com/sonic-net/sonic-buildimage/pull/15721
raising a separate PR

8111 800G interface, split to 2x400G (each has 4 lanes) fails to change interface speed from 400G to 100G during deploy mg. In minigraph.xml, the interface speed configuration is good, but fails to generate the right value to config_db.json.

In order to support this SKU the speed transitioning should support both 4 lanes and 8 lanes in the port_config.ini.

Why I did it

before this change for a 400G to 100G transition, in all cases except when lanes are 8, we would continue and the line ports.setdefault(port_name, {})['speed'] = port_speed_png[port_name] would not be executed, hence the default speed will never be set for a case and config_db will not be updated, where speed is transitioning from 400G to 100G or 40G, but lanes are not equal to 8.

In order for those cases to pass where lanes are not specifically 8, we need the change

Work item tracking
24242657

##### Work item tracking

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]


yxieca:
@vdahiya12 can you check the build fialures?

vdahiya12:
> @vdahiya12 can you check the build fialures?

seems all check passes now @yxieca

PR #16460, opened 2023-09-06
[TACACS] Improve per-command authorization performance by read passwd entry with getpwent

liuh-80:
Improve per-command authorization performance by read passwd entry with getpwent.

#### Why I did it
Currently per-command authorization will check if user is remote user with getpwnam API, which will trigger tacplus-nss for authentication with TACACS server.
But this is not necessary because when user login the user information already add to local passwd file.
Use getpwent API can directly read from passwd file, this will improve per-command authorization performance.

##### Work item tracking
- Microsoft ADO: 25104723

#### How I did it
Improve per-command authorization performance by read passwd entry with getpwent.

#### How to verify it
Pass all UT.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [x] 202205
- [x] 202211
- [x] 202305

- [x] SONiC.master-16460.356317-6c3424111
- [x] SONiC.202205-16659.368899-1c78e9b70

Improve per-command authorization performance by read passwd entry with getpwent.


qiluo-msft:
In "Tested branch (Please provide the tested image version)", please provide a version to indicate you have tested on earliest backport branch.

src/tacacs/bash_tacplus/bash_tacplus.c
```diff
@@ -350,40 +350,39 @@ int is_local_user(char *user)
     }
 
     struct passwd pwd;
-    struct passwd *pwdresult;
-    char *buf;
-    size_t bufsize = sysconf(_SC_GETPW_R_SIZE_MAX);
-    if (bufsize == -1) {
-        bufsize = DEFAULT_SC_GETPW_R_SIZE_MAX;
-    }
+    struct passwd *pwp;
```
qiluo-msft, line 353: >pwp [](http://example.com/codeflow?start=19&length=3)

Is it so easy to confuse `pwd`. how about `ppwd`?

src/tacacs/bash_tacplus/bash_tacplus.c
```diff
@@ -350,40 +350,39 @@ int is_local_user(char *user)
...
+            // no more pw entry
+            break;
+        }
 
-    buf = malloc(bufsize);
-    if (buf == NULL) {
-       output_error("failed to allocate getpwnam_r buffer.\n");
-       return ERROR_CHECK_LOCAL_USER;
-    }
+        if (strcmp(pwp->pw_name, user) != 0) {
```
qiluo-msft, line 365: >strcmp [](http://example.com/codeflow?start=12&length=6)

Is there an uplimit for the buffer size? If yes, prefer `strncmp`.

src/tacacs/bash_tacplus/bash_tacplus.c
```diff
@@ -350,40 +350,39 @@ int is_local_user(char *user)
...
+            continue;
+        }
 
-    int s = getpwnam_r(user, &pwd, buf, bufsize, &pwdresult);
-    int result = IS_LOCAL_USER;
-    if (pwdresult == NULL) {
-        if (s == 0)
-            output_error("get user information user failed, user: %s not found\n", user);
+        // compare passwd entry
+        if (strncmp(pwp->pw_gecos, REMOTE_USER_GECOS_PREFIX, strlen(REMOTE_USER_GECOS_PREFIX)) == 0) {
```
qiluo-msft, line 370: >strlen [](http://example.com/codeflow?start=61&length=6)

Suggest get the buf length at compilation time. Be careful, not easy to implement correctly.

qiluo-msft, line 370: >strncmp [](http://example.com/codeflow?start=12&length=7)

You need to use buf size, which at least string length + 1. Otherwise you may check string prefix instead of equality.

qiluo-msft:
> int is_local_user(char *user)

Suggest extensively unit test.

---
Refers to: src/tacacs/bash_tacplus/bash_tacplus.c:345 in ad1647d. [](commit_id = ad1647d0f86144c96274c4a6e57b3ea5770f9f0d, deletion_comment = False)

src/tacacs/bash_tacplus/bash_tacplus.c
liuh-80: No, both pw_name and user are char* return from system API.

src/tacacs/bash_tacplus/bash_tacplus.c
liuh-80: Fixed.

src/tacacs/bash_tacplus/bash_tacplus.c
liuh-80: Fixed by change to strcmp.

liuh-80: REMOTE_USER_GECOS_PREFIX is a pre-defined char*, not easy to get length in compilation unless we also define a length.
Fixed by change to strcmp.

liuh-80:
> Suggest extensively unit test.

Will add new UT and update later

liuh-80:
Change can't be clean cherry-pick to 202205, create a manually cherry-pick PR: https://github.com/sonic-net/sonic-buildimage/pull/16659

Will update PR description later.

liuh-80:
Close and open to trigger re-build.

qiluo-msft (approved):

liuh-80:
TACACS UT keeps failed on T0 device, every time failed on different place. test on local test bed all UT passed.

After check test failed log, seems UT check TACPLUS server log before server updates it.
Before this change, the per-command authorization need take longer time, so there is enough time to make tacplus server write data to log.

 create a new UT to fix this issue.:

https://github.com/sonic-net/sonic-mgmt/pull/10110

liuh-80:
UT fix merged, close and re-open to trigger validation.

liuh-80:
Master branch build image failed, waiting fix merge first: https://github.com/sonic-net/sonic-buildimage/pull/16859

liuh-80:
Cherry-pick to 202205 will have conflict, here is manually cherry-pick PR for 202205: https://github.com/sonic-net/sonic-buildimage/pull/16659

liuh-80:
UT still failed, seems found a corner case, will fix and update later.

qiluo-msft (approved):

liuh-80:
Close and reopen to trigger rebuild and validation.

src/tacacs/bash_tacplus/bash_tacplus.c
```diff
@@ -350,40 +351,39 @@ int is_local_user(char *user)
...
+            break;
+        }
 
-    buf = malloc(bufsize);
-    if (buf == NULL) {
-       output_error("failed to allocate getpwnam_r buffer.\n");
-       return ERROR_CHECK_LOCAL_USER;
-    }
+        if (strcmp(ppwd->pw_name, user) != 0) {
+            continue;
```
qiluo-msft, line 367: >continue [](http://example.com/codeflow?start=12&length=8)

This will lead to a loop, is it possible that the code flow always like this and keep burning CPU in an infinite loop?

liuh-80: This will not be an infinity loop, the getpwent_r API designed to return next entry in passwd, the loop will end when there is no next entry:

https://www.man7.org/linux/man-pages/man3/getpwent_r.3.html

PR #16504, opened 2023-09-09
[202205] Assign altname for bridge interface on chassis and iptables rules update to allow traffic on it.

abdosi:
What I did:
Fixes: https://github.com/sonic-net/sonic-buildimage/issues/16468

Why I did:
On Some chassis there is no dedicated `eth1-midplane` interface on supervisor for supervisor and LC communication but instead Linux bridge `br1` is used for that. Because of this changes that were done to white-list traffic over `eth1-midplane` would not work.

How I did:
To fix this we are using `altname` property of ip link command to set `eth1-midplane` as altname of `br1` interface.  This is done to keep design generic across chassis and between supervisor and LC also. IP-table rules are updated to get parent/base interface name of `eth1-midplane`.

How I verify:
UT Updated
Manual verification on chassis system


lguohan:
where is master branch pr?

lguohan:
is there a design doc for eth1-midplane? it is strange that we have a bridge name called eth1-midplane.

abdosi:
@lguohan can you check above comments and review it back

abdosi:
@lguohan can you review this PR and above comments

files/build_templates/docker_image_ctl.j2
```diff
@@ -180,6 +180,9 @@ function postStartAction()
                    ip link add name ns-eth1"$NET_NS" type veth peer name eth1@"$NET_NS"
                    ip link set dev eth1@"$NET_NS" master br1
                    ip link set dev eth1@"$NET_NS" up
+                   # For chassis system where Linux bridge is used on supervisor for midplane communication
+                   # assign alternate name as eth1-midplane for generic design
+                   ip link property add dev br1 altname eth1-midplane
```
lguohan, line 185: do you need to check if br1 exists before adding alias to br1?

src/sonic-host-services/scripts/caclmgrd
```diff
@@ -277,19 +277,24 @@ class ControlPlaneAclManager(daemon_base.DaemonBase):
...
+                                             " | awk '{print $0}' | cut -d' ' -f2"
+
+        midplane_dev_name = self.run_commands([chassis_midplane_dev_name_command])
+
         chassis_midplane_ip_command = "ip -4 -o addr show " + "eth1-midplane" +\
                                  " | awk '{print $4}' | cut -d'/' -f1 | head -1"
-        return self.run_commands([chassis_midplane_ip_command])
-
+        midplane_ip = self.run_commands([chassis_midplane_ip_command])
+        return midplane_dev_name, midplane_ip
```
lguohan, line 288: i am curious why we do not use this api?

https://github.com/sonic-net/sonic-platform-common/blob/master/sonic_platform_base/module_base.py#L537

files/build_templates/docker_image_ctl.j2
abdosi: `br` is created by the platform code. In that matter `eth1-midplane` is also created by vendor platform code. In the code above here there is already assumption `br1` is there since we are attaching all the Fabric ASIC Namespace to the `br1` for allowing them to talk to chassis db which is running over `br1` network.  Please note this all logic only applies to Arista Chassis as they have define `midplane_subnet`: https://github.com/sonic-net/sonic-buildimage/blob/a8af76f108893cc04fb472f4537bfc220626c5ef/device/arista/x86_64-arista_7800_sup/chassisdb.conf#L3.  Also since they are using 127.xx.xx.xx network for `br1` they had to create eth1 in Fabric namespace and enable local routing to talk to `br1`. All other vendor fabric namespace reaches chassis db over the docker bridge connected to `eth0` in each fabric namespace.

lguohan (approved):

PR #16539, opened 2023-09-13
Updated the API get_platform_info() to return running/detected ASIC's count

abdosi:
What I did:

Updated the API `get_platform_info() ` to return running/detected ASIC's count for given  system

ADO: 25158825

Why I did:

In Chassis for supervisor running asic count can be <= max asic possible.
`get_num_asics()` return max asic count that can be present based on static `asic.conf` file eg: https://github.com/sonic-net/sonic-buildimage/blob/fe24c2699681df2e3ef86199718b5e22651e6eac/device/nokia/x86_64-nokia_ixr7250e_sup-r0/asic.conf#L1. However in supervisor max_asic represnt number of asic where all Fabric cards are populated. However most of the time we have less than that inserted. `get_asic_presence_list()` returns how many Fabric asics are detected on Supervisor and for LC/Multi-asic system it still return max asic count as per `asic.conf` file

How I Verify:
Verified below commands on Chassis setup (multi-asic) and Pizza box with Single Asic

`show version`
`show platform summary`


gechiang:
What kind of issue does this fix?  Is this in particular to address issues such as some fabric card may not be present, so we need the running count instead of the max possible count?

abdosi:
> What kind of issue does this fix? Is this in particular to address issues such as some fabric card may not be present, so we need the running count instead of the max possible count?

yes that is the case. Also it's posssible some outside service can use the o/p from above commands to detect number of asic's on the system

src/sonic-py-common/sonic_py_common/device_info.py
```diff
@@ -479,15 +479,15 @@ def get_platform_info(config_db=None):
...
+    from .multi_asic import get_asic_presence_list
 
     version_info = get_sonic_version_info()
 
     hw_info_dict['platform'] = get_platform()
     hw_info_dict['hwsku'] = get_hwsku()
     if version_info:
         hw_info_dict['asic_type'] = version_info.get('asic_type')
-    hw_info_dict['asic_count'] = get_num_asics()
+    hw_info_dict['asic_count'] = len(get_asic_presence_list())
```
arlakshm, line 490: can we put this change in try/catch to ensure `show version` when chassis_db is not up?

abdosi: @arlakshm thanks. have updated accordingly.

gechiang:
@arlakshm , can you re-review with the latest change that Abhishek committed?
Thanks!

arlakshm (dismissed):

src/sonic-py-common/sonic_py_common/device_info.py
```diff
@@ -479,15 +479,18 @@ def get_platform_info(config_db=None):
     if hw_info_dict:
         return hw_info_dict
 
-    from .multi_asic import get_num_asics
+    from .multi_asic import get_asic_presence_list
```
lguohan, line 482: can you explain the reason for this change? why get_num_asics does not work, but get_asic_presence_list?

what is the scenario that was broken?

abdosi: @lguohan 
In Chassis for supervisor running asic count can be <= max asic possible.
`get_num_asics()` return max asic count that can be present based on static file eg: https://github.com/sonic-net/sonic-buildimage/blob/fe24c2699681df2e3ef86199718b5e22651e6eac/device/nokia/x86_64-nokia_ixr7250e_sup-r0/asic.conf#L1. However in supervisor max_asic represnt number of asic where all Fabric cards are populated. However most of the time we have less than that inserted. `get_asic_presence_list()` returns how many asics are detected on Supervisor and for LC/Multi-asic system it still return max asic count

src/sonic-py-common/sonic_py_common/device_info.py
```diff
@@ -479,15 +479,18 @@ def get_platform_info(config_db=None):
...
 
     hw_info_dict['platform'] = get_platform()
     hw_info_dict['hwsku'] = get_hwsku()
     if version_info:
         hw_info_dict['asic_type'] = version_info.get('asic_type')
-    hw_info_dict['asic_count'] = get_num_asics()
+    try:
+        hw_info_dict['asic_count'] = len(get_asic_presence_list())
+    except:
+        hw_info_dict['asic_count'] = 'N/A'
```
lguohan, line 493: is this compatiable with single asic? previously for single asic, the get_num_asics() return 1, in this case, it returns 
"N/A" now?

abdosi: @lguohan yes this change was breaking single asic as it was returning 0. Have fixed API `get_asic_presence_list()` to work  for single asic.

gechiang (approved):

src/sonic-py-common/sonic_py_common/device_info.py
lguohan: in the multi-asic case, what is in the asics_list? is that 0,1,2 or something else? @abdosi

abdosi: @lguohan : yes it 0,1,2

StormLiangMS:
@abdosi have you tested with 202305?

PR #16555, opened 2023-09-14
[build] Fix build isue caused by SECURE_UPGRADE_PROD_SIGNING_TOOL is empty.

liushilongbuaa:
#### Why I did it
Fix build error message:
basename: missing operand
Try 'basename --help' for more information.
##### Work item tracking
- Microsoft ADO **(number only)**:  25164528

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/slave.mk
+++ b/slave.mk
@@ -1287,7 +1287,9 @@ $(addprefix $(TARGET_PATH)/, $(SONIC_INSTALLERS)) : $(TARGET_PATH)/% : \
 	export sonic_su_dev_signing_key="$(SECURE_UPGRADE_DEV_SIGNING_KEY)"
 	export sonic_su_signing_cert="$(SECURE_UPGRADE_SIGNING_CERT)"
 	export sonic_su_mode="$(SECURE_UPGRADE_MODE)"
+ifneq ($(SECURE_UPGRADE_PROD_SIGNING_TOOL),)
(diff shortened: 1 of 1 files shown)


liushilongbuaa:
@DavidZagury , can you help review?

slave.mk
```diff
@@ -1287,7 +1287,9 @@ $(addprefix $(TARGET_PATH)/, $(SONIC_INSTALLERS)) : $(TARGET_PATH)/% : \
 	export sonic_su_dev_signing_key="$(SECURE_UPGRADE_DEV_SIGNING_KEY)"
 	export sonic_su_signing_cert="$(SECURE_UPGRADE_SIGNING_CERT)"
 	export sonic_su_mode="$(SECURE_UPGRADE_MODE)"
+ifneq ($(SECURE_UPGRADE_PROD_SIGNING_TOOL),)
```
DavidZagury, line 1290: Since SECURE_UPGRADE_PROD_SIGNING_TOOL is a necessary when  using prod signing, I think it would be better to check SECURE_UPGRADE_MODE and keep the error (or even increase its visibility) when it equals "prod"

liushilongbuaa: https://github.com/sonic-net/sonic-buildimage/blob/9ade881f87323f4b19703a3f8b6845df4d2d57a7/rules/config#L247
How about add default value in rules/config?

DavidZagury: This will work

PR #16561, opened 2023-09-14
[202205][Arista] Update arista platform submodules

Staphylo:
#### Why I did it

 - fix issue where platform debug info would no longer be in the dump
 - fix issue in scd-xcvr where active low bits couldn't be set
 - fix issue in scd-smbus where it perform an oob access

##### Work item tracking

Update arista platform library submodules


prgeor:
@Staphylo could you specify which all platforms have memory scrubber enabled?

Staphylo:
@prgeor I updated the PR description with the affected products

prgeor:
@Staphylo can we have the "enable DRAM ECC scrubbing on some platforms" is a feature so would like to be only in Master branch

Staphylo:
@prgeor I reverted the DRAM scrubbing change for 202205

prgeor:
> @prgeor I reverted the DRAM scrubbing change for 202205

@Staphylo could you update the PR description accordingly? Now the DRAM scrubbing is removed

Staphylo:
@prgeor I just updated the description

PR #16592, opened 2023-09-19
adjust the log rotation template to reduce the log size

sunhorus:
#### Why I did it
Modify the rsyslog template to reduce the logs files and size of each file
#### How I did it
Modify the rsyslog template to new smaller values to delete old logs periodically 
#### How to verify it
check [/etc/logrotate.d/rsyslog]

- [ x] 202111

some services like telemetry produce lots of logs and this can be challenging every time a techsupport extract is required as we have to delete manually a lot of log files

download


lguohan:
@prgeor to take a look

PR #16593, opened 2023-09-19
Common Authentication Template fix

Vignesh1405:
Issue Description: Local authentication succeeds (admin user) even when only tacacs+ is configured for login authentication Root-Cause: Authentication of Tacacs and tacacs,local was being handled in the same way in the template file common-auth-sonic.j2 
What i did: Separated the condition such that tacacs+,local is handled differently from tacacs+ configuration

#### Why I did it
Local authentication succeeds (admin user) even when only TACACS+  is configured for login authentication

#### How I did it
Separated the condition such that tacacs+,local is handled differently from tacacs+ configuration
#### How to verify it
Configure aaa authentication login as tacacs+. open a new ssh session, and try to login using local user credentials. The authentication should not succeed.

Separated the condition such that tacacs+,local is handled differently from tacacs+ configuration in common-auth-sonic.j2 file.

### DIFF
--- a/src/sonic-host-services-data/templates/common-auth-sonic.j2
+++ b/src/sonic-host-services-data/templates/common-auth-sonic.j2
@@ -22,11 +22,16 @@ auth	[success=done new_authtok_reqd=done default=ignore{{ ' auth_err=die' if not
 auth	[success=1 default=ignore]	pam_tacplus.so server={{ last_server.ip }}:{{ last_server.tcp_port }} secret={{ last_server.passkey }} login={{ last_server.auth_type }} timeout={{ last_server.timeout }} {% if last_server.vrf %} vrf={{ last_server.vrf }} {% endif %} {{ 'source_ip=%s' % src_ip if src_ip }} try_first_pass
 
 {% endif %}
-{% elif auth['login'] == 'tacacs+' or auth['login'] == 'tacacs+,local' %}
+{% elif auth['login'] == 'tacacs+' %}
 {% for server in servers %}
 auth	[success=done new_authtok_reqd=done default=ignore{{ ' auth_err=die' if not auth['failthrough'] }}]	pam_tacplus.so server={{ server.ip }}:{{ server.tcp_port }} secret={{ server.passkey }} login={{ server.auth_type }} timeout={{ server.timeout }} {%if server.vrf %} vrf={{ server.vrf }} {% endif %} {{ 'source_ip=%s' % src_ip if src_ip }} try_first_pass
 {% endfor %}
-auth	[success=1 default=ignore]	pam_unix.so nullok try_first_pass
+
+{% elif auth['login'] == 'tacacs+,local' %}
+{% for server in servers %}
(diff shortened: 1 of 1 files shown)


lguohan:
@qiluo-msft to review

src/sonic-host-services-data/templates/common-auth-sonic.j2
```diff
@@ -22,11 +22,16 @@ auth	[success=done new_authtok_reqd=done default=ignore{{ ' auth_err=die' if not
 auth	[success=1 default=ignore]	pam_tacplus.so server={{ last_server.ip }}:{{ last_server.tcp_port }} secret={{ last_server.passkey }} login={{ last_server.auth_type }} timeout={{ last_server.timeout }} {% if last_server.vrf %} vrf={{ last_server.vrf }} {% endif %} {{ 'source_ip=%s' % src_ip if src_ip }} try_first_pass
 
 {% endif %}
-{% elif auth['login'] == 'tacacs+' or auth['login'] == 'tacacs+,local' %}
+{% elif auth['login'] == 'tacacs+' %}
 {% for server in servers %}
 auth	[success=done new_authtok_reqd=done default=ignore{{ ' auth_err=die' if not auth['failthrough'] }}]	pam_tacplus.so server={{ server.ip }}:{{ server.tcp_port }} secret={{ server.passkey }} login={{ server.auth_type }} timeout={{ server.timeout }} {%if server.vrf %} vrf={{ server.vrf }} {% endif %} {{ 'source_ip=%s' % src_ip if src_ip }} try_first_pass
 {% endfor %}
-auth	[success=1 default=ignore]	pam_unix.so nullok try_first_pass
+
```
liuh-80, line 29: This line missing for 'tacacs+' only case:
auth [success=1 default=ignore] pam_unix.so nullok try_first_pass

I will verify the issue again.

src/sonic-host-services-data/templates/common-auth-sonic.j2
```diff
@@ -22,11 +22,16 @@ auth	[success=done new_authtok_reqd=done default=ignore{{ ' auth_err=die' if not
 auth	[success=1 default=ignore]	pam_tacplus.so server={{ last_server.ip }}:{{ last_server.tcp_port }} secret={{ last_server.passkey }} login={{ last_server.auth_type }} timeout={{ last_server.timeout }} {% if last_server.vrf %} vrf={{ last_server.vrf }} {% endif %} {{ 'source_ip=%s' % src_ip if src_ip }} try_first_pass
 
 {% endif %}
-{% elif auth['login'] == 'tacacs+' or auth['login'] == 'tacacs+,local' %}
+{% elif auth['login'] == 'tacacs+' %}
 {% for server in servers %}
```
liuh-80, line 26: The TACACS sever loop code are same.
2 place changed:

For TACACS+ only scenario, following line removed:
auth [success=1 default=ignore] pam_unix.so nullok try_first_pass

For TACACS+, LOCAL scenario, following line added:
auth [success=done new_authtok_reqd=done default=ignore{{ ' auth_err=die' if not auth['failthrough'] }}] pam_unix.so nullok try_first_pass

I will verify this again

liuh-80: The TACACS authorization code be there for long time, and in practice local user can't login when TACACS+ only enabled and remote server reachable.

So, need reproduce and verify this issue.

liuh-80:
@Vignesh1405 , I can't reproduce the issue in description with following steps:
1. Enable TACACS on device.
2. Check TACACS server side config, make sure 'admin' account not a remote user.
3. Login with device with 'admin' account.
My test result is: 'admin' user can't login with above steps.

Can you share me how to produce the issue you trying to fix?

Also, SONiC has following feature: When authentication set to 'tacacs+' and TACACS server not reachable, local user can login to device with local authorization.

PR #16691, opened 2023-09-26
fix ONIE partition size customization and add platform specific set ability

### LINKED ISSUE #15949: Unable to set custom ONIE disk partition size for OS
#### Description
When building the SONiC image, I'm unable to set the partition size to a custom value that is based on the HW configuration.
I was able to do it using some make file changes, but not sure if that is the way or if there is a already supported method to
customize the Disk partition size. If not supported yet, would be happy to do a pull request with my changes.

1. Try to set the ONIE build parameter ONIE_IMAGE_PART_SIZE to a custom value below.  
2. make USERNAME=test PASSWORD=test123 ONIE_IMAGE_PART_SIZE=130048 all
3. Check partition size on the final image built using "lsblk"

After booting the image and checking the OS partition size, It is still 32G

NAME   MAJ:MIN RM   SIZE RO TYPE MOUNTPOINT
loop0    7:0    0 419.4M  0 loop 
loop1    7:1    0     4G  0 loop /var/log
sda      8:0    0 119.2G  0 disk 
->sda1   8:1    0   256M  0 part 
->sda2   8:2    0   128M  0 part 
->sda3   8:3    0    32G  0 part /host

Expect the  the disk partition size to be 118.8G for sda3

This was tried on a 202205 branch, commit hash a66d4699f2ad75442ad146f9237717c01bfc45f3

pbudds:
#### Why I did it
fixes #15949
Problem 1: Setting ONIE_IMAGE_PART_SIZE using env variable or using "make ONIE_IMAGE_PART_SIZE=65536 USERNAME=test PASSWORD=test all" did not work.

Problem 2: The platform specific file for example "device/x86_64-8201_32fh_o-r0/installer.conf" cannot override it by setting value of ONIE_IMAGE_PART_SIZE in the file. change 2 adds support to do that. 

##### Work item tracking

#### How I did it
Change 1: when ONIE_IMAGE_PART_SIZE, the files Makefile.work and slave.mk should pass that setting along all the way to build_image.sh. Please see commit 1.

Change 2: In installer/install.sh, save the value set during build time string replace into a value and then let this value be overridden later when installer.conf get read which is platform specific. If platform does not override it, the original value will continue to work. Please see commit 2.

#### How to verify it
1: The below command works now
make ONIE_IMAGE_PART_SIZE=65536 USERNAME=test PASSWORD=test all"

The image properly was installed using ONIE and the partition size reflects what was passed in the above build command.
If the above value is not set, the default from "onie-image.conf" takes effect and still works.

2: Set ONIE_IMAGE_PART_SIZE in platform specific file like below example
--------------Diff----
device/x86_64-8201_32fh_o-r0/installer.conf
@@ -1 +1,2 @@
 ONIE_PLATFORM_EXTRA_CMDLINE_LINUX=" intel_iommu=off"
+ONIE_IMAGE_PART_SIZE=128000

and built the image using "make USERNAME=test PASSWORD=test all" and verified that the final installation properly partitioned the disk to the requested value from installer.conf file.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

master and 202205 branches

- [ ] 
- [ ]


lguohan (approved):

pbudds:
@lguohan - We are looking to have this fix as part of branch 202305.  Should I do a separate request for it? Could you please advise? Thank you.

PR #16702, opened 2023-09-26
[202305][Arista] Update platform library submodules

Staphylo:
#### Why I did it
- Convert `hw-dump` into `generate-dump` plugins
- Enable DRAM scrubber on some products
- Fix xcvr driver active low register bit logic
- Improve cooling algorithm (now considers xcvrs and modules)
- Add linecard graceful shutdown (disabled by default)
- Fix scd-smbus oob access

The scrubber was enabled for the following products:
- `DCS-7050QX-32S`
- `DCS-7050CX3-32S`
- `DCS-7060CX-32S`

##### Work item tracking

Update Arista platform library


Staphylo:
@prgeor @yxieca

StormLiangMS:
@prgeor  @Staphylo what's the known issue fixed in this PR? Is that necessary to have this PR in 202305?

Staphylo:
Hi Storm,

There are several issues addressed by this PR as described the description.
It fixes issues with the techsupport where we no longer have platform debugging available which is affecting our ability to debug production issues.
There are fixes related to chassis with the linecard graceful shutdown and the improvement of the cooling algorithm.
There is the DRAM scrubber issue which leave some old products without ECC protection.
And there are 2 fixes to our kernel modules.

StormLiangMS (approved):
lgtm

PR #16703, opened 2023-09-26
[Arista] Disable CPU C-States other than C1

Staphylo:
#### Why I did it

Networking devices need to be responsive. Such responsiveness is harmed when the CPU change state.
There is a latency penalty when a CPU is idle (e.g C2) and need to exit this state to come back to C1 state.
To prevent this from happening the CPU should be forced to remain in C1 state.

##### Work item tracking

#### How I did it

Generalize the cstate forcing to C1 to all Arista products.
This is done by adding `processor.max_cstate=1` to the kernel cmdline for all CPUs.
Additionally Intel CPUs also need `intel_idle.max_cstate=0` to fallback to the `acpi_idle` driver.

#### How to verify it

Check that `processor.max_cstate=1` is present on the cmdline for AMD CPUs
Check that both `processor.max_cstate=1` and `intel_idle.max_cstate=0` are present on the cmdline for Intel CPUs

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [x] 202211
- [x] 202305

Disable CPU C-States other than C1 for Arista products

### DIFF
--- a/files/Aboot/boot0.j2
+++ b/files/Aboot/boot0.j2
@@ -615,6 +615,14 @@ write_platform_specific_cmdline() {
         aboot_machine=arista_7280cr3mk_32p4
         flash_size=7382
     fi
+
+    # disable cpu c-state other than C1
+    local cpuvendor="$(sed -nr 's/vendor_id[\t ]*: (.*)/\1/p' /proc/cpuinfo | head -n 1)"
+    cmdline_add processor.max_cstate=1
+    if [ "$cpuvendor" = "GenuineIntel" ]; then
+        cmdline_add intel_idle.max_cstate=0
+    fi
+
     if in_array "$platform" "rook" "magpie" "woodpecker" "sprucefish"; then
         cmdline_add tsc=reliable
         cmdline_add pcie_ports=native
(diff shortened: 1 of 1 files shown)


Staphylo:
@StormLiangMS @lipxu

lipxu (commented):
Hi, @Staphylo , may you please let us know what's the motivation for this PR? Are we encountering any specific issues, or is there a performance concern driving this change? if so, how much improvement can be expected? thanks a lot

Staphylo:
Hi @lipxu this changes bring Arista products to parity with others using ONIE on the regard of CPU cstates.
See https://github.com/sonic-net/sonic-buildimage/pull/16339 and https://github.com/sonic-net/sonic-buildimage/pull/16371
This change removes the corner case where the CPU needs to spend extra time to process some data because it first has to exit a higher cstate.
A real world example in networking would be the CPU receiving the control plane packet but since it's idle it has to spend a few cycles more to actually start processing it.

lipxu (approved):

Blueve (approved):

Blueve:
@lguohan @yxieca I think we are ready to merge this PR, could you help?

PR #16776, opened 2023-10-04
Fix the dependency grpcio-tools version

qiluo-msft:
#### Why I did it
Fix the build break of marvell-armhf/sonic-ycabled

##### Work item tracking

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/sonic-slave-bullseye/Dockerfile.j2
+++ b/sonic-slave-bullseye/Dockerfile.j2
@@ -501,6 +501,11 @@ RUN patch -p1 -i /disable-non-manylinux.patch /usr/local/lib/python3.9/dist-pack
 # For building sonic-utilities
 RUN pip3 install fastentrypoints mock
 
+# For building sonic_ycabled
+# Note: upstream build breaks with old version of setuptools
+# ref: https://github.com/grpc/grpc/issues/34569
+RUN pip3 install grpcio==1.58.0 grpcio-tools==1.58.0
+
 # For running Python unit tests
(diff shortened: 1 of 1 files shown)


sonic-slave-bullseye/Dockerfile.j2
```diff
@@ -501,6 +501,11 @@ RUN patch -p1 -i /disable-non-manylinux.patch /usr/local/lib/python3.9/dist-pack
 # For building sonic-utilities
 RUN pip3 install fastentrypoints mock
 
+# For building sonic_ycabled
+# Note: upstream build breaks with old version of setuptools
+# ref: https://github.com/grpc/grpc/issues/34569
+RUN pip3 install grpcio==1.58.0 grpcio-tools==1.58.0
```
saiarcot895, line 507: Looks like the dockers and the base image are on GRPC 1.39.0. Do we want to use that version instead?

qiluo-msft: If the current situation is truly any version, it is using relative newer version until recent build broken. So I prefer not to change situation dramatically.

saiarcot895 (approved):

Blueve:
@StormLiangMS @yxieca please help backport this patch.

ADO: 25396344

PR #16777, opened 2023-10-04
[chassis/multi-asic] Make sure iBGP session established as directly connected 

abdosi:
What I did:
Make Sure for internal iBGP we are one-hop away (directly connected) by using Generic TTL security mechanism.

Why I did:
Without this change it's possible on packet chassis i-BGP can be established even if there no direct connection. Below is the example

- Let's say we have 3 LC's LC1/LC2/LC3 each having i-BGP session session with each other over Loopback4096
- Each LC's have static route towards other LC's Loopback4096 to establish i-BGP session
- LC1 learn default route `0.0.0.0/0` from it's e-BGP peers and send it over to LC2 and LC3 over i-BGP
- Now for some reason on LC2 static route towards LC3 is removed/not-present/some-issue we expect i-BGP session should go down between LC2 and LC3
- However i-BGP between LC2 and LC3 does not go down because of feature `ip nht-resolve-via-default`  where LC2 will use default route to reach Loopback4096 of LC3. As it's using default route BGP packets from LC2 towards LC3 will first route to LC1 and then go to LC3 from there.
- Above scenario can result in packet mis-forwarding on data plane 

How I fixed it:-
- To make sure BGP packets between i-BGP peers are not going with extra routing hop enable using `GTSM` feature
```
neighbor PEER ttl-security hops NUMBER
This command enforces Generalized TTL Security Mechanism (GTSM), as specified in RFC 5082. With this command, only neighbors that are the specified number of hops away will be allowed to become neighbors. This command is mutually exclusive with ebgp-multihop.
```
- We set hop count as 1 which makes FRR to reject BGP connection if we receive BGP packets if it's TTL < 255. Also setting this attribute make sure i-BGP frames are originated with IP TTL of 255.

How I verfiy:

Manual Verification of above scenario. See blow BGP packets receive with IP TTL 254 (additional routing hop) we are seeing FIN TCP flags as BGP is rejecting the connection

image


abdosi:
@arlakshm can you please help with review of this

arlakshm (approved):
lgtm

abdosi:
@lguohan / @StormLiangMS can you please help with merge of this.

judyjoseph (approved):

lguohan (approved):

gechiang:
@yxieca can you review if this can be approved for 202205?
Thanks!

saksarav-nokia:
@abdosi ,
In Broadcom based Chassis, the TTL for ibgp packets will be 254 since the packets are sent via recycle ports and the TTL will be decremented in the ingress asic.  So the ibgp neighbors are not coming up in our chassis after this PR is merged

abdosi:
> @abdosi PR conflicts with 202305 branch

@StormLiangMS : https://github.com/sonic-net/sonic-buildimage/pull/17237

PR #16782, opened 2023-10-05
[Nvidia] [Bookworm] Separate KERNEL_MFT into a new target

vivekrnv:
#### Why I did it

- KERNEL_MFT has dependency on kernel headers. But MFT is required in syncd (bullseye) docker. Thus separate them into two targets.
- MFT is no longer required in pmon and thus removed the dependency
- Update the linux-kernel submodule

##### Work item tracking

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]


platform/mellanox/mft/Makefile
```diff
@@ -47,6 +47,17 @@ $(addprefix $(DEST)/, $(MAIN_TARGET)): $(DEST)/% :
 	wget -O $(MFT_TGZ) $(MFT_TGZ_URL)
 	tar xzf $(MFT_TGZ)
 
+	# fix timestamp because we do not actually build tools, only kernel
+	touch $(MFT_NAME)/DEBS/*.deb
+	mv $(MFT_NAME)/DEBS/*.deb $(DEST)
+
+$(addprefix $(DEST)/, $(DERIVED_TARGETS)): $(DEST)/% : $(DEST)/$(MAIN_TARGET)
+
+$(addprefix $(DEST)/, $(KERNEL_MFT_TARGET)): $(DEST)/% :
```
saiarcot895, line 56: Unless I'm missing something, I think there's a chance that both this target and `MAIN_TARGET` could be built at the same time, in which case there would be issues here? Or would it not because `MAIN_TARGET`/mft is only built for Bullseye, whereas this target is built for Bookworm?

Either way, this would need to be fixed before the syncd container moves to Bookworm.

vivekrnv: Good catch, We need MFT for bookworm & bullseye. Whereas KERNEL_MFT only for bookworm. I'd move the processing of MFT into a tmp folder, extract there and copy the files

vivekrnv: Handled

saiarcot895 (approved):

PR #16879, opened 2023-10-13
[tph]: Detect LAG flaps from APPL_DB

theasianpianist:
#### Why I did it
- A race condition exists while the TPH is processing a netlink message - if a second netlink message arrives during processing it will be missed since TPH is not listening for other messages.
- Another bug was found where TPH was unnecessarily restarting since it was checking admin status instead of operational status of portchannels.

##### Work item tracking
- Microsoft ADO **(number only)**: 23622217

#### How I did it
- Subscribe to APPL_DB for updates on LAG operational state
- Track currently sniffed interfaces

#### How to verify it

- Send tunnel packets with destination IP of an unresolved neighbor, verify that `ping` commands are run
- Shut down a portchannel interface, verify that sniffer does not restart
- Send tunnel packets, verify `ping` commands are still run
- Bring up portchannel interface, verify that sniffer restarts

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/dockers/docker-orchagent/tunnel_packet_handler.py
+++ b/dockers/docker-orchagent/tunnel_packet_handler.py
@@ -8,6 +8,7 @@ packet is trapped to the CPU. In this case, we should ping the inner
 destination IP to trigger the process of obtaining neighbor information
 """
 import subprocess
+import sys
 import time
 from datetime import datetime
 from ipaddress import ip_interface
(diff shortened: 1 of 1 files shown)


dockers/docker-orchagent/tunnel_packet_handler.py
```diff
@@ -242,52 +253,36 @@ def get_inner_pkt_type(self, packet):
...
-        for msg in messages:
-            if self.netlink_msg_is_for_portchannel(msg):
-                if msg['state'] == 'up':
-                    logger.log_info('{} came back up, sniffer restart required'
-                                    .format(self.get_intf_name(msg)))
-                    return True
-        return False
+        oper_status = dict(fvs).get(OPER_STATUS_KEY)
+        if lag not in self.sniff_intfs and oper_status == 'up':
+            import pdb; pdb.set_trace()
```
prsunny, line 266: pdb? looks like test code?

prsunny (approved):

PR #16904, opened 2023-10-17
[build] Add DOCKER_USER_ENV option

jusherma:
#### Why I did it

To enable users to pass environmental variables into the SONiC build containers, especially values that cannot be safely written to disk, like authentication tokens.

The `DOCKER_USER_ENV` parameter supports any format allowed by Docker's `-e` flag. Values can be set explicitly or inherited from the host environment. For example, both of these are valid:

```
$ make DOCKER_USER_ENV="A=foo,B=bar" sonic-slave-bash

$ export A=foo
$ export B=bar
$ make DOCKER_USER_ENV="A,B" sonic-slave-bash
```

##### Work item tracking

#### How I did it

The optional Makefile parameter, `DOCKER_USER_ENV`, accepts comma-separated values to be passed as `-e` arguments to the `$(DOCKER_RUN)` call. This is the environmental variable equivalent of the `DOCKER_BUILDER_USER_MOUNT` parameter.

#### How to verify it

Build the `sonic-slave-bash` pseudotarget and check the environmental variables set inside of the build container(s):

1. Passing values explicitly:
```
$ make DOCKER_USER_ENV="A=foo,B=bar" sonic-slave-bash

BLDENV=buster make -f Makefile.work sonic-slave-bash
jusherma@630c70f71b19:/sonic$ echo ${A} ${B}
foo bar

BLDENV=bullseye make -f Makefile.work sonic-slave-bash
jusherma@02e2af7ed76c:/sonic$ echo ${A} ${B}
foo bar
```

2. Inheriting values from the host environment:
```
$ export A=foo
$ export B=bar
$ make DOCKER_USER_ENV="A,B" sonic-slave-bash

BLDENV=buster make -f Makefile.work sonic-slave-bash
jusherma@6a7ffcba48c5:/sonic$ echo ${A} ${B}
foo bar

BLDENV=bullseye make -f Makefile.work sonic-slave-bash
jusherma@e6a22c6175d4:/sonic$ echo ${A} ${B}
foo bar
```

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [x] 202205
- [ ] 202211
- [x] 202305

- [x] master 5ce2a71d 
- [x] 202205 8e945fb2

Enable users to pass environmental variables into the SONiC build containers

n/A

### DIFF
--- a/Makefile.work
+++ b/Makefile.work
@@ -348,6 +348,11 @@ ifneq ($(SIGNING_CERT),)
 endif
 endif
 
+# Allow user-defined environmental variables to be imported into the build containers
+ifneq ($(DOCKER_USER_ENV),)
+	DOCKER_RUN += $(foreach var,$(subst $(comma), ,$(DOCKER_USER_ENV)), $(addprefix -e , $(var)))
+endif
+
 # User name and tag for "docker-*" images created by native dockerd mode.
 ifeq ($(strip $(SONIC_CONFIG_USE_NATIVE_DOCKERD_FOR_BUILD)),y)
 DOCKER_USERNAME = $(USER_LC)


Makefile.work
```diff
@@ -348,6 +348,11 @@ ifneq ($(SIGNING_CERT),)
 endif
 endif
 
+# Allow user-defined environmental variables to be imported into the build containers
+ifneq ($(DOCKER_USER_ENV),)
```
xumia, line 352: How about using SONIC_BUILDER_EXTRA_CMDLINE to add the additional options?
https://github.com/sonic-net/sonic-buildimage/blob/7c3a063255c9e0914111ee89b09e2c12c58ec63b/Makefile.work#L304

jusherma: Yeah, it looks like `SONIC_BUILDER_EXTRA_CMDLINE` works for our use case. You can close this PR if you don't think it adds further value

PR #16907, opened 2023-10-17
Fix monit false alarm issue, which located in process_checker and missed "disk-sleep" status check, thus some 201911 SONiC box report "pmon|sensord" error coincidently.

FengPan-Frank:
Fix monit false alarm issue, which located in process_checker and it missed "disk-sleep" status check, thus some 201911 SONiC box report "pmon|sensord" error coincidently.

#### Why I did it
Currently psutil library returns below detail process status:
running: The process is currently running.
sleeping: The process is sleeping or waiting for an event to occur.
disk-sleep: The process is waiting for I/O operations to complete.
stopped: The process has been stopped (e.g. via the SIGSTOP signal).
zombie: The process has terminated but is still listed in the process table.
dead: The process has terminated and has been removed from the process table.

We should regard running/sleeping/disk-sleep as normal case and not alert in monit process.

Now once the disk-sleep occurs during monit cycle, below syslog will be paged, so get rid of syslog output meanwhile.

yslog.2.gz:Feb 24 06:12:17.394619 MEL23-0101-0301-04T1 ERR monit[6040]: 'pmon|sensord' status failed (1) -- '/usr/sbin/sensord -f daemon' is not running in host
syslog.2.gz:Feb 24 06:13:17.932531 MEL23-0101-0301-04T1 ERR monit[6040]: 'pmon|sensord' status failed (1) -- '/usr/sbin/sensord -f daemon' is not running in host
syslog.2.gz:Feb 24 06:14:18.502505 MEL23-0101-0301-04T1 ERR monit[6040]: 'pmon|sensord' status failed (1) -- '/usr/sbin/sensord -f daemon' is not running in host

Then I tried to reproduce the issue by triggering process_checker for sensord frequently and observed it's under "disk-sleep" status once the alert is raised.

##### Work item tracking
- Microsoft ADO **(number only)**:17663589

#### How I did it
Fix process_checker script code for adding "disk-sleep" case handling.

#### How to verify it
Verified in local DUT.

- [ ] 201811
- [x] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/files/image_config/monit/process_checker
+++ b/files/image_config/monit/process_checker
@@ -51,7 +51,7 @@ def check_process_existence(container_name, process_cmdline):
 
             for process in psutil.process_iter(["cmdline", "status", "pid"]):
                 try:
-                    if ((' '.join(process.cmdline())).startswith(process_cmdline) and process.status() in ["running", "sleeping"]):
+                    if ((' '.join(process.cmdline())).startswith(process_cmdline) and process.status() in ["running", "sleeping", "disk-sleep"]):
                          process_namespace_found_set.add(multi_asic.get_current_namespace(process.info['pid']))
                 except psutil.NoSuchProcess:
                     pass
@@ -72,10 +72,6 @@ def check_process_existence(container_name, process_cmdline):
                              namespace_display_str += ", " + ns
 
                  join_str = " and" if host_display_str and namespace_display_str else ""
-                 
-                 # If this script is run by Monit, then the following output will be appended to
-                 # Monit's syslog message.
-                 print("'{}' is not running{}{}{}".format(process_cmdline, host_display_str, join_str, namespace_display_str))
                  sys.exit(1)
     else:
         syslog.syslog(syslog.LOG_ERR, "container '{}' is not included in SONiC image or the given container name is invalid!"


qiluo-msft (approved):

abdosi (approved):

files/image_config/monit/process_checker
```diff
@@ -51,7 +51,7 @@ def check_process_existence(container_name, process_cmdline):
 
             for process in psutil.process_iter(["cmdline", "status", "pid"]):
                 try:
-                    if ((' '.join(process.cmdline())).startswith(process_cmdline) and process.status() in ["running", "sleeping"]):
```
prgeor, line 54: @FengPan-Frank one problem i see is that if the process is continuously in disk-sleep then its not expected and monit should raise alert. So, I would suggest, to remove the print from process_checker for process not running and let the monit check the return status of process_checker and raise the alert:-

```
check program pmon|ledd with path "/usr/bin/process_checker pmon /usr/bin/python /usr/bin/ledd"
    if status != 0 for 5 times within 5 cycles then alert repeat every 1 cycles
```

This way, we get alert if the process is in disk-sleep for 5 * 30 secs...

FengPan-Frank: Hi @prgeor , refer current logic https://github.com/sonic-net/sonic-buildimage/blob/36e65035bab6c0356ebcb9d0b1851553c9b5901c/dockers/docker-platform-monitor/base_image_files/monit_pmon.j2#L45C38-L45C38, all processes in monit will get alert if 5 consecutive times within 5 Monit cycles (1 minute per cycle)

Are you suggesting we adjust monit logic for sensord particularly?

prgeor: @FengPan-Frank remove this print https://github.com/sonic-net/sonic-buildimage/blob/8cf5b6827c8a1204ecd5ad0074cfb682d31bbd8e/files/image_config/monit/process_checker#L78

and let the monit alert print the syslog

FengPan-Frank: @prgeor Are you suggesting remove process_checker logging and when checking process status, do NOT skip "disk-sleep", as well as "sleeping" considering keep in "sleeping" state may also be somehow problematic, right?

prgeor: @FengPan-Frank for now I would suggest to keep the behavior same for 201911 except remove the extra print() and let monit watch over.

prgeor: @FengPan-Frank please capture the monit log which you see in your DUT in the PR description. I got confused, because of the printf()

prgeor (approved):

files/image_config/monit/process_checker
```diff
@@ -72,10 +72,6 @@ def check_process_existence(container_name, process_cmdline):
                              namespace_display_str += ", " + ns
 
                  join_str = " and" if host_display_str and namespace_display_str else ""
-                 
-                 # If this script is run by Monit, then the following output will be appended to
-                 # Monit's syslog message.
-                 print("'{}' is not running{}{}{}".format(process_cmdline, host_display_str, join_str, namespace_display_str))
                  sys.exit(1)
```
bingwang-ms, line 75: Why the print is removed?

bingwang-ms (approved):

PR #16956, opened 2023-10-20
Revert "[Mellanox] Align PSU temperature sysfs node name with hw-management change"

Junchao-Mellanox:
This PR depends on new hw-management version 7.0030.2000. Revert it for now and waiting for hw-management upgrading.


liat-grozovik:
@Junchao-Mellanox  please check if sonic-mgmt test should be reverted as well. if not, just followup on the checker failures please

liat-grozovik (approved):

PR #16969, opened 2023-10-23
Add LDAP feature support

davidpil2002:
#### Why I did it
To support LDAP feature
##### Work item tracking

#### How I did it
Similar to Radius/Tacacs authentication methods, the SONiC device is the LDAP client.
Installed the Debian LDAP packages related to making SONiC able to function as an LDAP client.

More description in the following HLD:
https://github.com/sonic-net/SONiC/pull/1487

#### How to verify it

Do LDAP configuration according to the HLD, then connect to the SONiC switch by using a user that exists in your LDAP server.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ] 

YANG Schema can be found attached in the HLD as well.

### DIFF
--- a/files/build_templates/sonic_debian_extension.j2
+++ b/files/build_templates/sonic_debian_extension.j2
@@ -298,7 +298,7 @@ sudo dpkg --root=$FILESYSTEM_ROOT -i $debs_path/bash-tacplus_*.deb || \
 # Install audisp-tacplus
 sudo dpkg --root=$FILESYSTEM_ROOT -i $debs_path/audisp-tacplus_*.deb || \
     sudo LANG=C DEBIAN_FRONTEND=noninteractive chroot $FILESYSTEM_ROOT apt-get -y install -f
-# Disable tacplus and LDAP y default
+# Disable tacplus and LDAP by default
 ## NOTE: this syntax of pam-auth-update is meant to be used when the package gets removed, not for specifying
 ## some local configuration of a PAM module. Currently, there's no clean way of noninteractively specifying
 ## whether some PAM module needs to be enabled or disabled on a system (there are hacky ways, though).
--- a/files/build_templates/sonic_debian_extension.j2
+++ b/files/build_templates/sonic_debian_extension.j2
@@ -272,6 +272,19 @@ sudo dpkg --root=$FILESYSTEM_ROOT -i $debs_path/sonic-device-data_*.deb || \
 # package for supporting password hardening
 sudo LANG=C DEBIAN_FRONTEND=noninteractive chroot $FILESYSTEM_ROOT apt-get -y install libpam-pwquality
 
+# Install pam-ldap, nss-ldap, ldap-utils
+sudo LANG=C DEBIAN_FRONTEND=noninteractive chroot $FILESYSTEM_ROOT apt-get -y install      \
+    libnss-ldapd \
+    libpam-ldapd \
+    ldap-utils
+
+# add networking.service dependancy to nslcd
+sudo LANG=C chroot $FILESYSTEM_ROOT sed -i '/# Required-Start:/ s/$/ networking.service/' /etc/init.d/nslcd
+
+# nslcd disable default
+sudo LANG=C chroot $FILESYSTEM_ROOT systemctl stop nslcd.service
+sudo LANG=C chroot $FILESYSTEM_ROOT systemctl mask nslcd.service
+
 # Install pam-tacplus and nss-tacplus
 sudo dpkg --root=$FILESYSTEM_ROOT -i $debs_path/libtac2_*.deb || \
     sudo LANG=C DEBIAN_FRONTEND=noninteractive chroot $FILESYSTEM_ROOT apt-get -y install -f
@@ -285,14 +298,14 @@ sudo dpkg --root=$FILESYSTEM_ROOT -i $debs_path/bash-tacplus_*.deb || \
 # Install audisp-tacplus
 sudo dpkg --root=$FILESYSTEM_ROOT -i $debs_path/audisp-tacplus_*.deb || \
     sudo LANG=C DEBIAN_FRONTEND=noninteractive chroot $FILESYSTEM_ROOT apt-get -y install -f
-# Disable tacplus by default
+# Disable tacplus and LDAP y default
 ## NOTE: this syntax of pam-auth-update is meant to be used when the package gets removed, not for specifying
 ## some local configuration of a PAM module. Currently, there's no clean way of noninteractively specifying
 ## whether some PAM module needs to be enabled or disabled on a system (there are hacky ways, though).
 ##
 ## If there is some PAM module that's installed/removed after this point, then this setting will end up having
 ## no impact, and there may be errors/test failures related to authentication.
-sudo LANG=C chroot $FILESYSTEM_ROOT pam-auth-update --remove tacplus
+sudo LANG=C chroot $FILESYSTEM_ROOT pam-auth-update --remove tacplus ldap
 sudo sed -i -e '/^passwd/s/ tacplus//' $FILESYSTEM_ROOT/etc/nsswitch.conf
 
 # Install pam-radius-auth and nss-radius
--- a/src/sonic-yang-models/yang-models/sonic-system-ldap.yang
+++ b/src/sonic-yang-models/yang-models/sonic-system-ldap.yang
@@ -1,30 +1,10 @@
 module sonic-system-ldap {
     yang-version 1.1;
     namespace "http://github.com/Azure/sonic-system-ldap";
-    prefix ssys;
+    prefix ssys-ldap;
 
     import ietf-inet-types {
         prefix inet;
-    }
-
-    import sonic-port {
-        prefix port;
-    }
-
-    import sonic-portchannel {
-        prefix lag;
-    }
-
-    import sonic-loopback-interface {
-        prefix loopback;
-    }
-
-    import sonic-interface {
-        prefix interface;
-    }
-
-    import sonic-mgmt_port {
-        prefix mgmt-port;
     }
 
     description "LDAP YANG Module for SONiC OS";
@@ -38,9 +18,9 @@ module sonic-system-ldap {
         container LDAP_SERVER {
             list LDAP_SERVER_LIST {
                 max-elements 8;
-                key "ipaddress";
+                key "hostname";
 
-                leaf ipaddress {
+                leaf hostname {
                     type inet:host;
                     description
                         "LDAP server's Domain name or IP address (IPv4 or IPv6)";
@@ -75,7 +55,7 @@ module sonic-system-ldap {
                     type string {
                         length "1..65";
                         pattern "[^ #,]*" {
-                            error-message 'TACACS shared secret (Valid chars are ASCII printable except SPACE, "#", and ",")';
+                            error-message 'LDAP shared secret (Valid chars are ASCII printable except SPACE, "#", and ",")';
                         }
                     }
                     description "Shared secret used for encrypting the communication";
@@ -111,7 +91,7 @@ module sonic-system-ldap {
                 leaf port {
                     type inet:port-number;
                     default 389;
-                    description "TCP port to communite with LDAP server";
+                    description "TCP port to communicate with LDAP server";
                 }
 
                 leaf timeout {
--- a/src/sonic-yang-models/yang-models/sonic-system-ldap.yang
+++ b/src/sonic-yang-models/yang-models/sonic-system-ldap.yang
@@ -1,0 +1,128 @@
+module sonic-system-ldap {
+    yang-version 1.1;
+    namespace "http://github.com/Azure/sonic-system-ldap";
+    prefix ssys;
+
+    import ietf-inet-types {
+        prefix inet;
+    }
+
+    import sonic-port {
+        prefix port;
+    }
+
+    import sonic-portchannel {
+        prefix lag;
+    }
+
+    import sonic-loopback-interface {
+        prefix loopback;
+    }
+
+    import sonic-interface {
+        prefix interface;
+    }
+
+    import sonic-mgmt_port {
+        prefix mgmt-port;
+    }
+
+    description "LDAP YANG Module for SONiC OS";
+
+    revision 2023-10-01 {
+        description "First Revision";
+    }
+
+    container sonic-system-ldap {
+
+        container LDAP_SERVER {
+            list LDAP_SERVER_LIST {
+                max-elements 8;
+                key "ipaddress";
+
+                leaf ipaddress {
+                    type inet:host;
+                    description
+                        "LDAP server's Domain name or IP address (IPv4 or IPv6)";
+                }
+
+                leaf priority {
+                    default 1;
+                    type uint8 {
+                        range "1..8" {
+                            error-message "LDAP server priority must be 1..8";
+                        }
+                    }
+                    description "Server priority";
+                }
+            }
+        }
+
+        container LDAP {
+
+            container global {
+
+
+                leaf bind_dn {
+                    type string {
+                        length "1..65";
+                    }
+                    description
+                            'LDAP global bind dn';
+                }
+
+                leaf bind_password {
+                    type string {
+                        length "1..65";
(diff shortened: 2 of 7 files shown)


liat-grozovik:
@lguohan who should review and provide feedback?

qiluo-msft:
@a-barboza Could you also help review this PR?

src/sonic-yang-models/yang-models/sonic-system-ldap.yang
```diff
@@ -0,0 +1,128 @@
...
+                    }
+                    description
+                            'LDAP global bind dn';
+                }
+
+                leaf bind_password {
+                    type string {
+                        length "1..65";
+                        pattern "[^ #,]*" {
+                            error-message 'TACACS shared secret (Valid chars are ASCII printable except SPACE, "#", and ",")';
```
madhupalu, line 78: Error message must be corrected from TACACS -> LDAP

src/sonic-yang-models/yang-models/sonic-system-ldap.yang
```diff
@@ -0,0 +1,128 @@
...
+                    type string {
+                        length "1..65";
+                    }
+                    description "Ldap user base dn";
+                }
+
+                leaf port {
+                    type inet:port-number;
+                    default 389;
+                    description "TCP port to communite with LDAP server";
```
madhupalu, line 114: Correct the typo - communite -> communicate

src/sonic-yang-models/yang-models/sonic-system-ldap.yang
```diff
@@ -0,0 +1,128 @@
+module sonic-system-ldap {
+    yang-version 1.1;
+    namespace "http://github.com/Azure/sonic-system-ldap";
+    prefix ssys;
+
+    import ietf-inet-types {
+        prefix inet;
+    }
+
+    import sonic-port {
```
madhupalu, line 10: Please remove unused imports such as port, lag, loopback, interface and mgmt-port.

src/sonic-yang-models/yang-models/sonic-system-ldap.yang
```diff
@@ -0,0 +1,128 @@
...
+                }
+            }
+        }
+
+        container LDAP {
+
+            container global {
+
+
+                leaf bind_dn {
```
a-barboza, line 66: In the ldap.conf file do these  correspond to the bind_dn --> binddn, bind_password --> bindpw, bind_timeout --> bind_timelimit, timeout --> timelimit, base_dn --> base  ?

src/sonic-yang-models/yang-models/sonic-system-ldap.yang
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    container sonic-system-ldap {
+
+        container LDAP_SERVER {
+            list LDAP_SERVER_LIST {
+                max-elements 8;
+                key "ipaddress";
+
+                leaf ipaddress {
```
a-barboza, line 43: ip address could be a hosname as well?  ipaddress --> name ?

files/build_templates/sonic_debian_extension.j2
```diff
@@ -280,14 +293,14 @@ sudo dpkg --root=$FILESYSTEM_ROOT -i $debs_path/bash-tacplus_*.deb || \
 # Install audisp-tacplus
 sudo dpkg --root=$FILESYSTEM_ROOT -i $debs_path/audisp-tacplus_*.deb || \
     sudo LANG=C DEBIAN_FRONTEND=noninteractive chroot $FILESYSTEM_ROOT apt-get -y install -f
-# Disable tacplus by default
+# Disable tacplus and LDAP y default
```
a-barboza, line 296: "by" default ?

a-barboza (changes requested):

src/sonic-yang-models/yang-models/sonic-system-ldap.yang
```diff
@@ -0,0 +1,128 @@
+module sonic-system-ldap {
+    yang-version 1.1;
+    namespace "http://github.com/Azure/sonic-system-ldap";
+    prefix ssys;
+
+    import ietf-inet-types {
+        prefix inet;
```
asha-behera, line 7: Are we not supporting ldap map configurations ? I do not see the corresponding yang fields 

map passwd loginShell "/bin/bash"
map group member uniqueMember

src/sonic-yang-models/yang-models/sonic-system-ldap.yang
```diff
@@ -0,0 +1,128 @@
...
+
+
+                leaf bind_dn {
+                    type string {
+                        length "1..65";
+                    }
+                    description
+                            'LDAP global bind dn';
+                }
+
```
asha-behera, line 73: Can we rename bind_dn to binddn, bind_password to bindpw, base_dn to base, version to ldap_version similar naming as in  nscld.conf

src/sonic-yang-models/yang-models/sonic-system-ldap.yang
```diff
@@ -0,0 +1,128 @@
...
+                    }
+                    description "Ldap bind timeout";
+                }
+
+                leaf version {
+                    default 3;
+                    type uint16 {
+                        range "1..3" {
+                            error-message "Ldap version must be 1..3";
+                        }
```
asha-behera, line 99: Are you not supporting "scope" field ?

src/sonic-yang-models/yang-models/sonic-system-ldap.yang
```diff
@@ -0,0 +1,128 @@
+module sonic-system-ldap {
+    yang-version 1.1;
+    namespace "http://github.com/Azure/sonic-system-ldap";
+    prefix ssys;
```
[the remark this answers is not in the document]
asha-behera: you may use "ssys-ldap" as prefix name

src/sonic-yang-models/yang-models/sonic-system-ldap.yang
```diff
@@ -0,0 +1,128 @@
...
+                    type string {
+                        length "1..65";
+                        pattern "[^ #,]*" {
+                            error-message 'TACACS shared secret (Valid chars are ASCII printable except SPACE, "#", and ",")';
+                        }
+                    }
+                    description "Shared secret used for encrypting the communication";
+                }
+
+                leaf bind_timeout {
```
asha-behera, line 84: Is bind_timeout same as bind_timelimit. If yes please rename it to nomenclature similar to conf file

src/sonic-yang-models/yang-models/sonic-system-ldap.yang
```diff
@@ -0,0 +1,128 @@
...
+                    description
+                            'LDAP global bind dn';
+                }
+
+                leaf bind_password {
+                    type string {
+                        length "1..65";
+                        pattern "[^ #,]*" {
+                            error-message 'TACACS shared secret (Valid chars are ASCII printable except SPACE, "#", and ",")';
+                        }
```
asha-behera, line 79: Do we not intend to add to support nss_initgroups_ignoreusers,  timelimit, idle_timelimit.

src/sonic-yang-models/yang-models/sonic-system-ldap.yang
```diff
@@ -0,0 +1,128 @@
...
+                    description "Ldap user base dn";
+                }
+
+                leaf port {
+                    type inet:port-number;
+                    default 389;
+                    description "TCP port to communite with LDAP server";
+                }
+
+                leaf timeout {
```
asha-behera, line 117: Do we intend to add SSL options 
"Ssl/Tls Options
T<sslT> on|off|start_tls
Specifies whether to use SSL/TLS or not (the default is not to). If start_tls is specified then StartTLS is used rather than raw LDAP over SSL. Not all LDAP client libraries support both SSL, StartTLS and all related configuration options."

src/sonic-yang-models/yang-models/sonic-system-ldap.yang
```diff
@@ -0,0 +1,128 @@
...
+    import sonic-mgmt_port {
+        prefix mgmt-port;
+    }
+
+    description "LDAP YANG Module for SONiC OS";
+
+    revision 2023-10-01 {
+        description "First Revision";
+    }
+
```
asha-behera, line 35: Do we intend to add filter options , which would be useful in optimizing queries
T<filterT> MAP FILTER
The FILTER is an LDAP search filter to use for a specific map. The default filter is a basic search on the objectClass for the map (e.g. T<(objectClass=posixAccount)T>).

src/sonic-yang-models/yang-models/sonic-system-ldap.yang
davidpil2002: correct

src/sonic-yang-models/yang-models/sonic-system-ldap.yang
davidpil2002: done

src/sonic-yang-models/yang-models/sonic-system-ldap.yang
davidpil2002: done

src/sonic-yang-models/yang-models/sonic-system-ldap.yang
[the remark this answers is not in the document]
davidpil2002: done

src/sonic-yang-models/yang-models/sonic-system-ldap.yang
davidpil2002: modify to host, since it can support domain as well

src/sonic-yang-models/yang-models/sonic-system-ldap.yang
davidpil2002: In this first LDAP PRs, we are covering the basic functionalities.
LDAP map and filter are not part of the HLD presented,
it can be part of future work, 
if you need you can raise a request and we can see who can lead this enhancement.

src/sonic-yang-models/yang-models/sonic-system-ldap.yang
davidpil2002: not in this first LDAP PRs, can be part of a future works

src/sonic-yang-models/yang-models/sonic-system-ldap.yang
davidpil2002: timelimit was part of the HLD.
The others can be raised as a request as well for future work

The motivation is to have first a stable feature version.
And then we can discuss who can lead advances in LDAP configurations

src/sonic-yang-models/yang-models/sonic-system-ldap.yang
davidpil2002: (as in prev comment)
Adance LDAP configuration can be raised for future work.
The motivation is to have first a stable feature version.
And then we can discuss who can lead advances in LDAP configurations

src/sonic-yang-models/yang-models/sonic-system-ldap.yang
davidpil2002: for now, we decided to remain with the current names since they are human readable as well.

src/sonic-yang-models/yang-models/sonic-system-ldap.yang
davidpil2002: yes
for now, we decided to remain with the current names since they are human readable as well.

timeout is more human readable

src/sonic-yang-models/yang-models/sonic-system-ldap.yang
davidpil2002: Adance LDAP configuration can be raised for future work.
The motivation is to have first a stable feature version.
And then we can discuss who can lead advances in LDAP configurations

src/sonic-yang-models/yang-models/sonic-system-ldap.yang
davidpil2002: done

files/build_templates/sonic_debian_extension.j2
davidpil2002: done

davidpil2002:
error look not related to the new code.
rerunning:
/azpw run Azure.sonic-buildimage

davidpil2002:
> yang-model tests are missing for LDAP, please add the same.

DONE

davidpil2002:
All comments was answered,
pls can you re-review/approve?
@shdasari @asha-behera @a-barboza @madhupalu

src/sonic-yang-models/yang-models/sonic-system-ldap.yang
a-barboza: > done - modified to hostname, since it can support domain as well

hostname per openconfig-system.yang has the oc-inet:domain-name type which is exclusively for domain name. Since the intent for the key is to also have ipv4|ipv6 addresses then it could be suggestion to use a domain name which is not the intent. Hence the suggestion to use "name" which is neutral.

src/sonic-yang-models/yang-models/sonic-system-ldap.yang
a-barboza: Why not use the same names as in nslcd.conf? It will be less of burden of mapping the nslcd.conf names to users.

davidpil2002: for now, we decided to remain with the current names since they are human readable as well.

src/sonic-yang-models/yang-models/sonic-system-ldap.yang
davidpil2002: If there is no technical problem in this yang model, I prefer to remain with the hostname, which is more descriptive of the use case.
probably someone should fix the already usage of hostname in the context of oc-inet to be domain_name to not block the usage of hostname in other cases like this one.

src/sonic-yang-models/yang-models/sonic-system-ldap.yang
a-barboza: Aren't the nslcd.conf names supposed to be human readable? It is a flat text file with descriptive comments.

src/sonic-yang-models/yang-models/sonic-system-ldap.yang
a-barboza: hostname utility actually refers to displaying the systems DNS name (Eg: hostname cli utility). I think the oc-inet:domain-name type usage is accurate for openconfig-system.yang.

fastiuk:
@a-barboza , could you please review? I saw you did HLD review as well

src/sonic-yang-models/yang-models/sonic-system-ldap.yang
davidpil2002: yes and no,
from Linux POV yes, but we are using an special sonic script `sonic-cli-gen`which is generating the CLI from the yang file
So for now, we decided to saved the code as is regarding naming,
If you find any functional bug, pls share.

files/build_templates/sonic_debian_extension.j2
```diff
@@ -272,6 +272,19 @@ sudo dpkg --root=$FILESYSTEM_ROOT -i $debs_path/sonic-device-data_*.deb || \
 # package for supporting password hardening
 sudo LANG=C DEBIAN_FRONTEND=noninteractive chroot $FILESYSTEM_ROOT apt-get -y install libpam-pwquality
 
+# Install pam-ldap, nss-ldap, ldap-utils
```
qiluo-msft, line 275: >Install [](http://example.com/codeflow?start=2&length=7)

Suggest to add build time config, so deployers not using LDAP could reduce image size, and reduce the security concerns.

Yarden-Z: Why should this be different than radius/tacacs? Why show different approaches for a feature in the same domain?

PR #16995, opened 2023-10-25
[Chassis] : Update PG/Q from 4=>7 for CPU bound packets

vmittal-msft:
#### Why I did it
This is needed to segregate CPU bound traffic from other PG/Q traffic. e.g. PG/Q=4

##### Work item tracking

#### How I did it
Updated copp_config.j2 to use queue7 group instead of queue4.

#### How to verify it
Verified on sonic chassis.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [x] 202205
- [x] 202211
- [x] 202305

- [ ] 
- [ ]

### DIFF
--- a/files/image_config/copp/copp-config.sh
+++ b/files/image_config/copp/copp-config.sh
@@ -1,3 +1,8 @@
 #!/bin/bash
 
-sonic-cfggen -d -t /usr/share/sonic/templates/copp_cfg.j2 > /etc/sonic/copp_cfg.json
+switch_type=`sonic-db-cli CONFIG_DB  hget 'DEVICE_METADATA|localhost' 'switch_type'`
+if [[$switch_type != 'voq']]; then
+    sonic-cfggen -d -t /usr/share/sonic/templates/copp_cfg.j2 > /etc/sonic/copp_cfg.json
+else
+    sonic-cfggen -d -t /usr/share/sonic/templates/copp_cfg_chassis.j2 > /etc/sonic/copp_cfg.json
+fi
--- a/files/image_config/copp/copp_cfg_chassis.j2
+++ b/files/image_config/copp/copp_cfg_chassis.j2
@@ -1,0 +1,111 @@
+{
+    "COPP_GROUP": {
+	    "default": {
+		    "queue": "0",
+		    "meter_type":"packets",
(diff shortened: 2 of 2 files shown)


files/image_config/copp/copp_cfg_chassis.j2
```diff
@@ -0,0 +1,111 @@
+{
```
arlakshm, line 1: do we need to create a new file for chassis? Can we update the existing j2  file  to modify the queue for voq switch-type?

files/image_config/copp/copp-config.sh
```diff
@@ -1,3 +1,8 @@
 #!/bin/bash
 
-sonic-cfggen -d -t /usr/share/sonic/templates/copp_cfg.j2 > /etc/sonic/copp_cfg.json
+switch_type=`sonic-db-cli CONFIG_DB  hget 'DEVICE_METADATA|localhost' 'switch_type'`
+if [[$switch_type != 'voq']]; then
+    sonic-cfggen -d -t /usr/share/sonic/templates/copp_cfg.j2 > /etc/sonic/copp_cfg.json
+else
+    sonic-cfggen -d -t /usr/share/sonic/templates/copp_cfg_chassis.j2 > /etc/sonic/copp_cfg.json
```
arlakshm, line 7: Do we need changes to copy `copp_cfg_chassis.j2`  to `/usr/share/sonic/templates/` ?

arlakshm (commented):
as comments

files/image_config/copp/copp_cfg_chassis.j2
```diff
@@ -0,0 +1,111 @@
...
+		    "red_action":"drop"
+	    },
+	    "queue7_group1": {
+		    "trap_action":"trap",
+		    "trap_priority":"4",
+		    "queue": "7"
+	    },
+	    "queue7_group2": {
+		    "trap_action":"copy",
+		    "trap_priority":"4",
```
judyjoseph, line 18: So here we are keeping the trap_priority as 4 itself, and queue as 7, any relation between queue number and trap priority?

PR #16827, opened 2023-10-10
Adding new secondary field to VLAN_INTERFACE table

shbalaku-microsoft:
#### Why I did it

This is change taken as part of the HLD: https://github.com/sonic-net/SONiC/pull/1470. 
In this PR we add the logic to parse the SecondarySubnets field in the minigraph and add a flag in "secondary" in the vlan_interface table of the config db.

##### Work item tracking
- Microsoft ADO **(number only)**: 16784946

#### How I did it

Made changes in the minigraph.py to parse the xml entry and add the parsed value to the config db 

#### How to verify it

Added python tests in the sonic-config-engine folder to test the config db entries.

#### Which release branch to backport (provide reason below if selected)

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

#### Tested branch (Please provide the tested image version)

- [ ] 
- [ ] 

#### Description for the changelog

Adding new secondary field to VLAN_INTERFACE table

PR #17012, opened 2023-10-25
Modify dhcp relay to pick primary address

shbalaku-microsoft:
#### Why I did it
This is change taken as part of the HLD: https://github.com/sonic-net/SONiC/pull/1470 and this is a follow up on the PR #16827 where in the docker-dhcp we pick the value of primary gateway of the interface from the VLAN_Interface table which has "secondary" flag set in the config_db
##### Work item tracking
- Microsoft ADO **(number only)**: 16784946

#### How I did it

1: Changes in the j2 file to add a new "-pg" parameter in the dhcpv4-relay.agents.j2, the ip would be retrieved from the config db's vlan_interface table such that the interface which are picked will have secondary field set.
2: Changes in isc-dhcp to re-order the addresses of the discovered interface and which has the ip which has the passed parameter.

#### How to verify it

Validated using the existing test cases.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ] 

Modify dhcp relay to pick primary address


src/sonic-config-engine/tests/test_j2files.py
```diff
@@ -19,6 +19,7 @@ def setUp(self):
         self.ztp_ip = os.path.join(self.test_dir, "sample-ztp-ip.json")
         self.ztp_inband_ip = os.path.join(self.test_dir, "sample-ztp-inband-ip.json")
         self.t0_minigraph = os.path.join(self.test_dir, 't0-sample-graph.xml')
+        self.t0_minigraph_secondary_subnets = os.path.join(self.test_dir, 't0-sample-graph-secondary-subnets.xml')
```
prsunny, line 22: Does this graph already exist?

src/isc-dhcp/patch/0015-option-to-set-primary-address-in-interface.patch
```diff
@@ -0,0 +1,114 @@
+diff --git a/common/discover.c b/common/discover.c
```
prsunny, line 1: @kellyyeh , could you please review this?

src/sonic-config-engine/sonic-cfggen
```diff
@@ -137,6 +138,27 @@ def ip_network(value):
         return "Invalid ip address %s" % value
     return r_v.network
 
+def get_primary_addr(value):
+    if not value:
+        return ""
+    table = pfx_filter(value)
+    returnTable = OrderedDict()
```
prsunny, line 145: Please use one consistent format for name. Rename this to `primary_subnets`

dockers/docker-dhcp-relay/dhcpv4-relay.agents.j2
```diff
@@ -24,6 +24,9 @@ command=/usr/sbin/dhcrelay -d -m discard -a %%h:%%p %%P --name-alias-map-file /t
 {% for (name, prefix) in PORTCHANNEL_INTERFACE|pfx_filter %}
 {% if prefix | ipv4 %} -iu {{ name }}{% endif -%}
 {% endfor %}
+{% for (name, gateway) in VLAN_INTERFACE|get_primary_addr %}
```
prsunny, line 27: Does this change the existing behavior if no secondary address is present? I thought we are setting `-pg` only if there is secondary address in the graph.

prsunny (commented):
Can you also add unit tests?

dockers/docker-dhcp-relay/dhcpv4-relay.agents.j2
shbalaku-microsoft: We only add -pg for the VLANs which has secondary subnets present. So the current behavior of no secondary subnet present would not add any -pg parameters

src/sonic-config-engine/tests/test_j2files.py
shbalaku-microsoft: It did not exist, so removed this.

kellyyeh (approved):

prsunny (approved):

PR #16684, opened 2023-09-25
Add special rsyslog filter for MSN2700 platform

keboliu:
##### Work item tracking
- Microsoft ADO **(number only)**:

Mellanox MSN2700 platforms have a non-functional error log:  "ERR pmon#sensord: Error getting sensor data: dps460/#10: Can't read".  This error is because of a firmware issue with some PSU, we are not able to upgrade the FW online. Since there is no functional impact, this error log can be ignored safely.

#### How I did it

Add a new rsyslog rule to the rsyslog-container.conf.j2, if the docker name is pmon and the platform name matches, the new rule will be inserted into the docker rsyslogd.conf

#### How to verify it

run regression on the MSN2700 platform to make the error log will not be printed to the syslog.

#### Which release branch to backport (provide reason below if selected)

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [x] 202205
- [x] 202211
- [x] 202305

#### Tested branch (Please provide the tested image version)

- [ ] 
- [ ]

PR #17020, opened 2023-10-26
[202205] Add special rsyslog filter for MSN2700 platform (#16684)

keboliu:
#### Why I did it

**backport https://github.com/sonic-net/sonic-buildimage/pull/16684**

Mellanox MSN2700 platforms have a non-functional error log:  "ERR pmon#sensord: Error getting sensor data: dps460/#10: Can't read".  This error is because of a firmware issue with some PSU, we are not able to upgrade the FW online. Since there is no functional impact, this error log can be ignored safely.

##### Work item tracking

#### How I did it

Add a new rsyslog rule to the rsyslog-container.conf.j2, if the docker name is pmon and the platform name matches, the new rule will be inserted into the docker rsyslogd.conf

#### How to verify it

run regression on the MSN2700 platform to make the error log will not be printed to the syslog.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]


keboliu:
@yxieca please help to approve and merge when the checker pass, this is to backport https://github.com/sonic-net/sonic-buildimage/pull/16684

PR #17041, opened 2023-10-30
VOQ: add dependent peer's stronger dependency with swss to avoid double stop/start and traffic loss

deepak-singhal0408:
#### Why I did it
Issue details and RCA here. https://github.com/sonic-net/sonic-buildimage/issues/16991
it fixes https://github.com/sonic-net/sonic-buildimage/issues/16991

##### Work item tracking
- Microsoft ADO **25422412**:

#### How I did it
In SWSS script, I have added stronger dependency with dependent peer services using "Requires=" command in unit files of dependent peer services. This will ensure that, syncd/gbsyncd dont start immediately and instead wait till swss service starts its init phase.
I have made this change for VOQ chassis for now, however this is also applicable to pizza boxes running SONiC(there is no functionality impact seen on pizza box however).

#### How to verify it
followed same steps as mentioned in https://github.com/sonic-net/sonic-buildimage/issues/16991 and confirmed that 
a. No traffic loss end to end
b. Dont see double stop/start of syncd service

Also Ran few more tests.
1. Tested on both single asic and multi-asic hw platforms.
2. Tested config-reload
3. sudo reboot
4. service restart
5. syncd process crash, orchagent process crash

Confirmed from logs, that in none of the scenario, syncd gets killed twice.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [x] 202205
- [ ] 202211
- [x] 202305

SONiC.20220532.45

- [ ] 
- [ ]

### DIFF
--- a/files/scripts/swss.sh
+++ b/files/scripts/swss.sh
@@ -18,6 +18,12 @@ function debug()
     /bin/echo `date` "- $1" >> ${DEBUGLOG}
 }
 
+function get_switch_type()
+{
+    switch_type=`$SONIC_DB_CLI CONFIG_DB  hget 'DEVICE_METADATA|localhost' 'switch_type'`
+    echo "${switch_type}"
+}
+
 function read_dependent_services()
 {
     # Update dependent list based on other packages requirements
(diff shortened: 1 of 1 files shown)


deepak-singhal0408:
@mlok-nokia for viz..

rlhui:
@anamehra also help review

deepak-singhal0408:
Opened another PR https://github.com/sonic-net/sonic-host-services/pull/83 to fix it via hosfcfgd/featured script to keep systemd service files override at one place.

PR #17042, opened 2023-10-30
Check if PSU files exists when getting psu_voltage properties

oleksandra-bella:
### Why I did it 
Error messages occured when trying to read PSU files on init: 
`ERR pmon#psud: Failed to read from file /var/run/hw-management/power/psu1_volt_out2_capability - FileNotFoundError(2, 'No such file or directory')`

This can happen when the power cord is disconnected from the PSU, so some PSU files may be absent, e.g.: 
/var/run/hw-management/power/psu2_volt_out2
/var/run/hw-management/power/psu2_volt_out2_capability

### How I did it 
Check if a file exists for a specific PSU parameter If not, return None so we can't read the PSU file any further

### How to verify it
1. Disconnect power cord from PSU and power supply from system
2. Wait few minutes and then connect power supply to system without power cord
3. Check logs for errors

### DIFF
--- a/platform/mellanox/mlnx-platform-api/sonic_platform/psu.py
+++ b/platform/mellanox/mlnx-platform-api/sonic_platform/psu.py
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2019-2023 NVIDIA CORPORATION & AFFILIATES.
+# Copyright (c) 2019-2024 NVIDIA CORPORATION & AFFILIATES.
 # Apache-2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


prgeor (dismissed):

stephenxs (changes requested):
IMO it should be the caller's responsibility to guarantee the PSU is available before calling such APIs.
So, I suggest fixing it by checking the PSU availability before calling the APIs.

prgeor:
> IMO it should be the caller's responsibility to guarantee the PSU is available before calling such APIs. So, I suggest fixing it by checking the PSU availability before calling the APIs.

@oleksandra-bella please address comment

oleksandra-bella:
> > IMO it should be the caller's responsibility to guarantee the PSU is available before calling such APIs. So, I suggest fixing it by checking the PSU availability before calling the APIs.
> 
> @oleksandra-bella please address comment

I've already fixed the comment by checking PSU file availability before calling the APIs

stephenxs (approved):

prgeor (approved):

prgeor:
@liat-grozovik please help merge

liat-grozovik (approved):

PR #17045, opened 2023-10-31
[SNMP]: Modify minigraph parser to update SNMP_AGENT_ADDRESS_CONFIG

SuvarnaMeenakshi:
table

#### Why I did it
SNMP query over IPv6 does not work due to issue in net-snmp where IPv6 query does not work on multi-nic environment.
To get around this, if snmpd listens on specific ipv4 or ipv6 address, then the issue is not seen.
We plan to configure Management IP and Loopback IP configured in minigraph.xml as SNMP_AGENT_ADDRESS in config_db., based on changes discussed in https://github.com/sonic-net/SONiC/pull/1457.

##### Work item tracking
- Microsoft ADO **(number only)**:26091228

#### How I did it
Modify minigraph parser to update SNMP_AGENT_ADDRESS_CONFIG with management and Loopback0 IP addresses.
Modify snmpd.conf.j2 to use SNMP_AGENT_ADDRESS_CONFIG table if it is present in config_db, if not listen on any IP.
Main change:
1. if minigraph.xml is used to configure the device, then snmpd will listen on mgmt and loopback IP addresses,
2. if config_db is used to configure the device, snmpd will listen IP present in SNMP_AGENT_ADDRESS_CONFIG  if that table is present, if table is not present snmpd will listen on any IP.
#### How to verify it
config_db.json created from minigraph.xml for single asic VS image with mgmt and Loopback IP addresses.
```
    "SNMP_AGENT_ADDRESS_CONFIG": {
        "10.1.0.32|161|": {},
        "10.250.0.101|161|": {},
        "FC00:1::32|161|": {},
        "fec0::ffff:afa:1|161|": {}
    },
 .....
 
 snmpd listening on the above IP addresses:
 admin@vlab-01:~$ sudo netstat -tulnp | grep 161
tcp        0      0 127.0.0.1:3161          0.0.0.0:*               LISTEN      71522/snmpd         
udp        0      0 10.250.0.101:161        0.0.0.0:*                           71522/snmpd         
udp        0      0 10.1.0.32:161           0.0.0.0:*                           71522/snmpd         
udp6       0      0 fec0::ffff:afa:1:161    :::*                                71522/snmpd         
udp6       0      0 fc00:1::32:161          :::*                                71522/snmpd  
```

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/src/sonic-config-engine/minigraph.py
+++ b/src/sonic-config-engine/minigraph.py
@@ -1723,14 +1723,13 @@ def parse_xml(filename, platform=None, port_config_file=None, asic_name=None, hw
     if not is_multi_asic() and asic_name is None:
         results['SNMP_AGENT_ADDRESS_CONFIG'] = {}
         port = '161'
-        for mgmt_intf in results['MGMT_INTERFACE'].keys():
+        for mgmt_intf in mgmt_intf.keys():
             snmp_key = mgmt_intf[1].split('/')[0] + '|' + port + '|'
             results['SNMP_AGENT_ADDRESS_CONFIG'][snmp_key] = {}
         # Add Loopback IP as agent address for single asic
-        for loip in results['LOOPBACK_INTERFACE']:
-            if len(loip) == 2:
-                snmp_key = loip[1].split('/')[0] + '|' + port + '|'
-                results['SNMP_AGENT_ADDRESS_CONFIG'][snmp_key] = {}
+        for loip in lo_intfs.keys():
+            snmp_key = loip[1].split('/')[0] + '|' + port + '|'
+            results['SNMP_AGENT_ADDRESS_CONFIG'][snmp_key] = {}
     else:
         results['SNMP_AGENT_ADDRESS_CONFIG'] = {}
 
--- a/src/sonic-config-engine/minigraph.py
+++ b/src/sonic-config-engine/minigraph.py
@@ -16,6 +16,7 @@ from natsort import natsorted, ns as natsortns
 
 from portconfig import get_port_config, get_fabric_port_config, get_fabric_monitor_config
 from sonic_py_common.interface import backplane_prefix
+from sonic_py_common.multi_asic import is_multi_asic
 
 # TODO: Remove this once we no longer support Python 2
 if sys.version_info.major == 3:
@@ -1719,17 +1720,17 @@ def parse_xml(filename, platform=None, port_config_file=None, asic_name=None, hw
     results['MGMT_VRF_CONFIG'] = mvrf
     # Update SNMP_AGENT_ADDRESS_CONFIG with Management IP and Loopback0 IP
     # if available.
-    results['SNMP_AGENT_ADDRESS_CONFIG'] = {}
-    port = 161
-    vrf = ''
-    for mgmt_if in results['MGMT_INTERFACE'].keys():
-        snmp_key = mgmt_if[1].split('/')[0] + '|' + str(port)  + '|'
-        results['SNMP_AGENT_ADDRESS_CONFIG'][snmp_key] = {}
-    # Add Loopback0 IP as agent address for single asic
-    for loip in results['LOOPBACK_INTERFACE']:
-        if len(loip) == 2 and loip[0] == 'Loopback0':
-            snmp_key = loip[1].split('/')[0] + '|161|'
+    if not is_multi_asic() and asic_name is None:
+        results['SNMP_AGENT_ADDRESS_CONFIG'] = {}
+        port = 161
+        for mgmt_if in results['MGMT_INTERFACE'].keys():
+            snmp_key = mgmt_if[1].split('/')[0] + '|' + str(port)  + '|'
             results['SNMP_AGENT_ADDRESS_CONFIG'][snmp_key] = {}
+        # Add Loopback0 IP as agent address for single asic
+        for loip in results['LOOPBACK_INTERFACE']:
+            if len(loip) == 2 and loip[0] == 'Loopback0':
+                snmp_key = loip[1].split('/')[0] + '|' + str(port)  + '|'
+                results['SNMP_AGENT_ADDRESS_CONFIG'][snmp_key] = {}
 
     phyport_intfs = {}
     vlan_intfs = {}
--- a/src/sonic-config-engine/minigraph.py
+++ b/src/sonic-config-engine/minigraph.py
@@ -1718,7 +1718,7 @@ def parse_xml(filename, platform=None, port_config_file=None, asic_name=None, hw
             results['LOOPBACK_INTERFACE'][host_lo_intf[0]] = {}
 
     results['MGMT_VRF_CONFIG'] = mvrf
-    # Update SNMP_AGENT_ADDRESS_CONFIG with Management IP and Loopback0 IP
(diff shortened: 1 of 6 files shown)


src/sonic-config-engine/minigraph.py
```diff
@@ -1717,6 +1718,19 @@ def parse_xml(filename, platform=None, port_config_file=None, asic_name=None, hw
...
+    # if available.
+    if not is_multi_asic() and asic_name is None:
+        results['SNMP_AGENT_ADDRESS_CONFIG'] = {}
+        port = 161
+        for mgmt_if in results['MGMT_INTERFACE'].keys():
+            snmp_key = mgmt_if[1].split('/')[0] + '|' + str(port)  + '|'
+            results['SNMP_AGENT_ADDRESS_CONFIG'][snmp_key] = {}
+        # Add Loopback0 IP as agent address for single asic
+        for loip in results['LOOPBACK_INTERFACE']:
+            if len(loip) == 2 and loip[0] == 'Loopback0':
```
arlakshm, line 1731: condition `len(loip)` seems very strict. This condition will break if an additional loopback interface is introduced in the minigraph.

SuvarnaMeenakshi: @arlakshm thanks for reviewing, the len(loip) is to make sure that the loip inside LOOPBACK_INTERFACE is actually <name of interface>|<IP address>.
For example, we have 
```
       "LOOPBACK_INTERFACE": {
        "Loopback0": {}, -- This does not contain the IP address, so we don't need to parse this.
        "Loopback0|10.1.0.32/32": {}, -- here the len(loip) will be 2 and we will be able to get the IP address.
        
(Pdb) loip
('Loopback0', '10.1.0.32/32')
(Pdb) len(loip)
2
```

arlakshm (dismissed):
lgtm.

arlakshm (dismissed):

src/sonic-config-engine/minigraph.py
```diff
@@ -1717,6 +1718,19 @@ def parse_xml(filename, platform=None, port_config_file=None, asic_name=None, hw
             results['LOOPBACK_INTERFACE'][host_lo_intf[0]] = {}
 
     results['MGMT_VRF_CONFIG'] = mvrf
+    # Update SNMP_AGENT_ADDRESS_CONFIG with Management IP and Loopback IP
+    # if available.
+    if not is_multi_asic() and asic_name is None:
```
qiluo-msft, line 1723: >asic_name [](http://example.com/codeflow?start=31&length=9)

why checking `asic_name`?

SuvarnaMeenakshi: For host namespace, asic_name will be None and SNMP_AGENT_ADDRESS_CONFIG is only required for host namespace in case of multi-asic devices. For single asic device asic_name will always be None, but added this check to ensure that we only generate for host namespace.
Minigraph parser generates config for host and each namespace separately.
This is done by passing the right asic_name/

src/sonic-config-engine/minigraph.py
```diff
@@ -1717,6 +1718,19 @@ def parse_xml(filename, platform=None, port_config_file=None, asic_name=None, hw
...
+        results['SNMP_AGENT_ADDRESS_CONFIG'] = {}
+        port = 161
+        for mgmt_if in results['MGMT_INTERFACE'].keys():
+            snmp_key = mgmt_if[1].split('/')[0] + '|' + str(port)  + '|'
+            results['SNMP_AGENT_ADDRESS_CONFIG'][snmp_key] = {}
+        # Add Loopback IP as agent address for single asic
+        for loip in results['LOOPBACK_INTERFACE']:
```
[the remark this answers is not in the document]
SuvarnaMeenakshi: the len(loip) is to make sure that the loip inside LOOPBACK_INTERFACE is actually in the format ```<name of interface>|<IP address>```
For example, we have 
```
       "LOOPBACK_INTERFACE": {
        "Loopback0": {}, ------------------ This does not contain the IP address, so we don't need to parse this.
        "Loopback0|10.1.0.32/32": {}, ---- here the len(loip) will be 2 and we will be able to get the IP address.
        
(Pdb) loip
('Loopback0', '10.1.0.32/32')
(Pdb) len(loip)
2
```

src/sonic-config-engine/minigraph.py
```diff
@@ -1717,6 +1718,19 @@ def parse_xml(filename, platform=None, port_config_file=None, asic_name=None, hw
             results['LOOPBACK_INTERFACE'][host_lo_intf[0]] = {}
 
     results['MGMT_VRF_CONFIG'] = mvrf
+    # Update SNMP_AGENT_ADDRESS_CONFIG with Management IP and Loopback IP
+    # if available.
+    if not is_multi_asic() and asic_name is None:
+        results['SNMP_AGENT_ADDRESS_CONFIG'] = {}
+        port = 161
+        for mgmt_if in results['MGMT_INTERFACE'].keys():
```
qiluo-msft, line 1726: >results['MGMT_INTERFACE'] [](http://example.com/codeflow?start=23&length=25)

Could you use `mgmt_intf`?

src/sonic-config-engine/minigraph.py
qiluo-msft, line 1723: >if [](http://example.com/codeflow?start=4&length=2)

If the condition does not meet, please initialize an empty `results['SNMP_AGENT_ADDRESS_CONFIG']`.

qiluo-msft, line 1723: >not [](http://example.com/codeflow?start=7&length=3)

I guess you want to check `is_multi_asic`, not single asic.

SuvarnaMeenakshi: I want to check if the device is single asic, then update SNMP_AGENT_ADDRESS_CONFIG.
Hence checking "not is_multi_asic()".
No change required for multi-asic.

src/sonic-config-engine/minigraph.py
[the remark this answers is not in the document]
qiluo-msft: `results['MGMT_INTERFACE']` is generated by this script itself, it is more efficient to parse from original data, not parsing from generated data.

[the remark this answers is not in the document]
SuvarnaMeenakshi: Makes sense, I have modified as suggested.

wen587:
Is it a feature in 202205 branch? If so, we will also require the change in NDM side.
Also, for new table we will require related YANG model to support the YANG validation.

SuvarnaMeenakshi:
> Is it a feature in 202205 branch? If so, we will also require the change in NDM side. Also, for new table we will require related YANG model to support the YANG validation.

This change will be only in master and not in 202205/2023 branches.
We have added the yang model with this PR: https://github.com/sonic-net/sonic-buildimage/pull/15587

qiluo-msft (approved):
LGTM. Please also check with reviewers.

wen587 (approved):

PR #17065, opened 2023-11-02
Issuing DEL command to remove APPL_DB:PORT_TABLE

mihirpat1:
#### Why I did it
This change is needed to issue a "DEL" command to remove the APPL_DB:PORT_TABLE. The current approach of issuing FLUSH command to remove the DB doesn't allow xcvrd to subscribe and act on PORT_TABLE deletion event.

The change set is part of implementation of the HLD [OA crash handling to reinitialize port through xcvrd](https://github.com/sonic-net/SONiC/pull/1432)

##### Work item tracking
24264739

#### How I did it
Calling clean_up_tables function to issue "DEL" command for removing APPL_DB:PORT_TABLE.

#### How to verify it
Please refer to https://github.com/sonic-net/sonic-platform-daemons/pull/403 for the test results.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [x] 202305

- [ ] 
- [ ]

### DIFF
--- a/files/scripts/swss.sh
+++ b/files/scripts/swss.sh
@@ -292,6 +292,7 @@ start() {
     # Don't flush DB during warm boot
     if [[ x"$WARM_BOOT" != x"true" ]]; then
         debug "Flushing APP, ASIC, COUNTER, CONFIG, and partial STATE databases ..."
+        clean_up_tables APPL_DB "'PORT_TABLE:E*'"
         $SONIC_DB_CLI APPL_DB FLUSHDB
         $SONIC_DB_CLI ASIC_DB FLUSHDB
         $SONIC_DB_CLI COUNTERS_DB FLUSHDB
--- a/files/scripts/swss.sh
+++ b/files/scripts/swss.sh
@@ -292,7 +292,7 @@ start() {
     # Don't flush DB during warm boot
     if [[ x"$WARM_BOOT" != x"true" ]]; then
         debug "Flushing APP, ASIC, COUNTER, CONFIG, and partial STATE databases ..."
-        clean_up_tables APPL_DB "'PORT_TABLE:E*'"
+        clean_up_tables APPL_DB "'PORT_TABLE:Ethernet*'"
         $SONIC_DB_CLI APPL_DB FLUSHDB
         $SONIC_DB_CLI ASIC_DB FLUSHDB
         $SONIC_DB_CLI COUNTERS_DB FLUSHDB


mihirpat1:
@prgeor @shyam77git @jaganbal-a - It will be great if you can help in reviewing this.

files/scripts/swss.sh
```diff
@@ -292,6 +292,7 @@ start() {
     # Don't flush DB during warm boot
     if [[ x"$WARM_BOOT" != x"true" ]]; then
         debug "Flushing APP, ASIC, COUNTER, CONFIG, and partial STATE databases ..."
+        clean_up_tables APPL_DB "'PORT_TABLE:E*'"
```
prgeor, line 295: @mihirpat1 is there a difference in behavior when swss start as part of cold reboot vs as part of swss/OA crash?

mihirpat1: @prgeor - No, I don't see a difference in behavior between between cold reboot v/s SWSS/OA crash

prgeor: @mihirpat1 can you add the regex for `EthernetXX`

mihirpat1: I have addressed this now.

files/scripts/swss.sh
```diff
@@ -292,6 +292,7 @@ start() {
     # Don't flush DB during warm boot
     if [[ x"$WARM_BOOT" != x"true" ]]; then
         debug "Flushing APP, ASIC, COUNTER, CONFIG, and partial STATE databases ..."
+        clean_up_tables APPL_DB "'PORT_TABLE:Ethernet*'"
```
prsunny, line 295: I think xcvrd should subscribe to STATE_DB instead of APP_DB. This should not be done here, IMO

prgeor: @mihirpat1 i see Xcvrd subscribing to these tables. What am I missing?

```
  port_tbl_map = [
        {'CONFIG_DB': swsscommon.CFG_PORT_TABLE_NAME},
        {'STATE_DB': 'TRANSCEIVER_INFO'},
        {'STATE_DB': 'PORT_TABLE', 'FILTER': ['host_tx_ready']},
    ]
```

mihirpat1: @prsunny - XCVRD already subscribes to APPL_DB since with this change, I am changing the way to delete entries from APPL_DB. Am I missing something here?
https://github.com/sonic-net/sonic-platform-daemons/blob/502c0b6622008363cb1ed6d1b7c85b4093997093/sonic-xcvrd/xcvrd/xcvrd.py#L1173

Also, as @prgeor mentioned, XCVRD already subscribes to STATE_DB as well.

mihirpat1:
Closing this PR since I am now using STATE_DB instead of APPL_DB for killing xcvrd during OA crash scenario.

https://github.com/sonic-net/sonic-utilities/pull/3111

PR #17080, opened 2023-11-02
Disable NETIF_F_HW_CSUM in KNET

byu343:
#### Why I did it
This is CSP CS00012280996.
The issue to fix is that the checksum was incorrect for all TCP packets leaving the system so that the BGP connection cannot be established.  We found the issue on BCM56993, and it is possible to affect all platforms using linux_ngknet.

##### Work item tracking

#### How I did it
Remove NETIF_F_HW_CSUM

#### How to verify it
We verified that the BGP connection is established according to 'show ip bgp summary'.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [x] 202305

- [ ] 
- [ ]


StormLiangMS:
ADO:25760023

PR #13969, opened 2023-02-23
[config reload]Config Reload Enhancement

dgsudharsan:
#### Why I did it
Implementing code changes for https://github.com/sonic-net/SONiC/pull/1203

#### How I did it
Removed the timers and delayed target since the delayed services would start based on event driven approach.
Cleared port table during config reload and cold reboot scenario.
Modified yang model, init_cfg.json to change has_timer to delayed

#### How to verify it
Running regression

#### Which release branch to backport (provide reason below if selected)

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211

PR #17133, opened 2023-11-10
[lldp] Clean up service start logic owing to port init start optimization

vivekrnv:
#### Why I did it

As part of service start optimization done, we've removed timer units and used PORT_INIT_DONE as the checkpoint for starting delayed services. Thus this logic is redundant

This fixes the following error log. 
```
Nov  3 14:16:54.281546 r-leopard-44 ERR systemctl[10553]: Failed to get unit file state for lldp.timer: No such file or directory
```

#### How I did it

Redundant lldp timer code is removed from syncd.sh. 

#### How to verify it

Build and test 

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/files/build_templates/lldp.timer.j2
+++ b/files/build_templates/lldp.timer.j2
@@ -1,1 +1,0 @@
-per_namespace/lldp.timer.j2
--- a/files/scripts/syncd.sh
+++ b/files/scripts/syncd.sh
@@ -105,17 +105,6 @@ function waitplatform() {
             debug "Started pmon service"
         fi
     fi
-    if [[ x"$BOOT_TYPE" = @(x"fast"|x"warm"|x"fastfast") ]]; then
-        debug "LLDP service is delayed by a timer for better fast/warm boot performance"
-    else
-        lldp_state=$(systemctl is-enabled lldp.timer)
-        if [[ $lldp_state == "enabled" ]]
-        then
-            debug "Starting lldp service..."
-            /bin/systemctl start lldp
-            debug "Started lldp service"
-        fi
-    fi
 }
 
 function stopplatform1() {


dgsudharsan (approved):

files/build_templates/lldp.timer.j2
```diff
@@ -1 +0,0 @@
-per_namespace/lldp.timer.j2
```
saiarcot895, line 1: Does the actual file need to be delelted as well, not just the symlink?

vivekrnv: Actual file doesn't exist

saiarcot895 (approved):

files/build_templates/lldp.timer.j2
yxieca: The file is here: https://github.com/sonic-net/sonic-buildimage/blob/master/files/build_templates/per_namespace/lldp.service.j2

Removing the symbolic link might break chassis? @arlakshm

yxieca (changes requested):
@arlakshm to assess the removal of lldp.timer.j2

files/build_templates/lldp.timer.j2
saiarcot895: That's the `lldp.service` file, not the `lldp.timer` file. It looks like the timer was removed in #13969.

arlakshm:
> @arlakshm to assess the removal of lldp.timer.j2

@SuvarnaMeenakshi @abdosi can help check if this change will have impact on chassis bootup?

vivekrnv:
@SuvarnaMeenakshi @abdosi  @yxieca Any update here?

SuvarnaMeenakshi:
I don't see this change impacting multi-asic or chassis apart from the fact that lldp will always start after the delay which was not the case earlier (not due to this PR but due to PR 13969)
This PR is more to clean up the change that was missed in the enhancement PR.
The script change here is checking if lldp.timer is enabled, if it is enabled then it would let the timer wait and start service upon warm-boot or fast-boot it will start immediately if for cold boot. 
With the enhancement changes in PR 13969, lldp is a delayed service and will start after the delay for all scenarios and all platforms. 
Is that acceptable? @yxieca

PR #17152, opened 2023-11-13
[UT] Fix eventd zmq code bug and UT flaky issue.

liushilongbuaa:
#### Why I did it
Fix eventd UT flaky error.  Here is zmq doc:  http://api.zeromq.org/2-1:zmq-setsockopt

int zmq_setsockopt (void *socket, int option_name, const void *option_value, size_t option_len);

Caution: All options, with the exception of ZMQ_SUBSCRIBE, ZMQ_UNSUBSCRIBE and ZMQ_LINGER, only take effect for subsequent socket bind/connects.

Typical use is to set this socket option ahead of each zmq_connect() attempt to a new host. Each connection MUST be assigned a unique name. Assigning a name that is already in use is not allowed.

##### Work item tracking

#### How I did it

Put setsockopt before connects.

#### How to verify it

In my local test. 
eventd.proxy passed 10 out of 10
eventd.capture passed 9 out of 10.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/src/sonic-eventd/src/eventd.cpp
+++ b/src/sonic-eventd/src/eventd.cpp
@@ -382,14 +382,14 @@ capture_service::do_capture()
     cap_sub_sock = zmq_socket(m_ctx, ZMQ_SUB);
     RET_ON_ERR(cap_sub_sock != NULL, "failing to get ZMQ_SUB socket");
 
+    rc = zmq_setsockopt(cap_sub_sock, ZMQ_RCVTIMEO, &block_ms, sizeof (block_ms));
+    RET_ON_ERR(rc == 0, "Failed to ZMQ_RCVTIMEO to %d", block_ms);
+
     rc = zmq_connect(cap_sub_sock, get_config(string(CAPTURE_END_KEY)).c_str());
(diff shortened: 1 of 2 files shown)


liushilongbuaa:
@zbud-msft , can you help review?
I'm not sure if it will fix the UT issue but it make sense according to zmq doc.

zbud-msft:
According to zmq documentation, the use of setting the socket option is only listed for the below option, which we are not setting.

ZMQ_CONNECT_RID: Assign the next outbound connection id
The ZMQ_CONNECT_RID option sets the peer id of the next host connected via the zmq_connect() call, and immediately readies that connection for data transfer with the named id. This option applies only to the first subsequent call to zmq_connect(), calls thereafter use default connection behaviour.

Typical use is to set this socket option ahead of each zmq_connect() attempt to a new host. Each connection MUST be assigned a unique name. Assigning a name that is already in use is not allowed.

PR #17161, opened 2023-11-14
[dpu]: Add DPU database service

Pterosaur:
Sub PRs:
https://github.com/sonic-net/sonic-host-services/pull/84
https://github.com/sonic-net/sonic-buildimage/pull/17191

#### Why I did it
According to the design, the database instances of DPU will be kept in the NPU host.

##### Work item tracking
- Microsoft ADO **(number only)**: 25072889

#### How I did it

To follow the multiple ASIC design, I assume a new platform environment variable `NUM_DPU` will be defined in the `/usr/share/sonic/device/$PLATFORM/platform_env.conf`. Based on this number, NPU host will launch a corresponding number of instances for the DPU database.

#### How to verify it

1. Defined `NUM_DPU` in the platform_env.conf
image

2. Declare a new field in the config_db for database feature.
image

3. remove the existing databse service by 'docker rm -f database' and reboot OS.

We should get multiple database services and global_config as following.
image

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/dockers/docker-database/Dockerfile.j2
+++ b/dockers/docker-database/Dockerfile.j2
@@ -32,7 +32,8 @@ RUN apt-get clean -y                                  && \
              s/^# unixsocket/unixsocket/;                \
              s/redis-server.sock/redis.sock/g;           \
              s/^client-output-buffer-limit pubsub [0-9]+mb [0-9]+mb [0-9]+/client-output-buffer-limit pubsub 0 0 0/; \
-             s/^notify-keyspace-events ""$/notify-keyspace-events AKE/ \
+             s/^notify-keyspace-events ""$/notify-keyspace-events AKE/; \
+             s/^databases [0-9]+$/databases 100/ \
             ' /etc/redis/redis.conf
 
 COPY ["supervisord.conf.j2", "/usr/share/sonic/templates/"]
--- a/dockers/docker-database/database_config.json.j2
+++ b/dockers/docker-database/database_config.json.j2
@@ -3,8 +3,9 @@
         "redis":{
             "hostname" : "{{HOST_IP}}",
             "port" : {{REDIS_PORT}},
-            "unix_socket_path" : "/var/run/redis{{NAMESPACE_ID}}/redis.sock",
-            "persistence_for_warm_boot" : "yes"
+            "unix_socket_path" : "/var/run/redis{{DEV}}/redis.sock",
+            "persistence_for_warm_boot" : "yes",
+            "database_type": "{{DATABASE_TYPE}}"
         },
         "redis_chassis":{
             "hostname" : "redis_chassis.server",
@@ -94,6 +95,29 @@
             "separator": ":",
             "instance" : "redis"
         }
+{% if DATABASE_TYPE is defined and DATABASE_TYPE == "dpudb" %}
+        ,
+        "DPU_APPL_DB" : {
+            "id" : 15,
+            "separator": ":",
+            "instance" : "redis"
+        },
+        "DPU_APPL_STATE_DB" : {
+            "id" : 16,
+            "separator": "|",
+            "instance" : "redis"
+        },
+        "DPU_STATE_DB" : {
+            "id" : 17,
+            "separator": "|",
+            "instance" : "redis"
+        },
+        "DPU_COUNTERS_DB" : {
+            "id" : 18,
+            "separator": ":",
+            "instance" : "redis"
+        }
+{% endif %}
     },
     "VERSION" : "1.0"
 }
--- a/dockers/docker-database/database_config.json.j2
+++ b/dockers/docker-database/database_config.json.j2
@@ -100,7 +100,8 @@
         "DPU_APPL_DB" : {
             "id" : 15,
             "separator": ":",
-            "instance" : "redis"
+            "instance" : "redis",
+            "format": "proto"
         },
         "DPU_APPL_STATE_DB" : {
             "id" : 16,
--- a/dockers/docker-database/database_config.json.j2
+++ b/dockers/docker-database/database_config.json.j2
@@ -2,7 +2,7 @@
     "INSTANCES": {
         "redis":{
             "hostname" : "{{HOST_IP}}",
-            "port" : 6379,
+            "port" : {{REDIS_PORT}},
             "unix_socket_path" : "/var/run/redis{{NAMESPACE_ID}}/redis.sock",
             "persistence_for_warm_boot" : "yes"
         },
--- a/dockers/docker-database/database_config.json.j2
+++ b/dockers/docker-database/database_config.json.j2
@@ -4,8 +4,10 @@
             "hostname" : "{{HOST_IP}}",
             "port" : {{REDIS_PORT}},
             "unix_socket_path" : "/var/run/redis{{DEV}}/redis.sock",
-            "persistence_for_warm_boot" : "yes",
-            "database_type": "{{DATABASE_TYPE}}"
+            "persistence_for_warm_boot" : "yes"
+            {% if DATABASE_TYPE is defined and DATABASE_TYPE != "" %}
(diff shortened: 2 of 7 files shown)


dockers/docker-database/database_config.json.j2
```diff
@@ -2,7 +2,7 @@
     "INSTANCES": {
         "redis":{
             "hostname" : "{{HOST_IP}}",
-            "port" : 6379,
+            "port" : {{REDIS_PORT}},
             "unix_socket_path" : "/var/run/redis{{NAMESPACE_ID}}/redis.sock",
```
ganglyu, line 6: DPU should use different unix socket path.

ganglyu:
Why do we have /var/run/redisdpu0/sonic-db/database_global.json?

ganglyu:
Please create a separate PR for sonic yang models

dockers/docker-database/database_config.json.j2
Pterosaur: Done

Pterosaur:
> Why do we have /var/run/redisdpu0/sonic-db/database_global.json?

Removed.

ganglyu (approved):

src/sonic-py-common/sonic_py_common/device_info.py
```diff
@@ -841,3 +844,24 @@ def is_frontend_port_present_in_host():
         if not namespace_id:
             return False
     return True
+
+
+def get_num_dpus():
+    num_dpus = 0
```
lguohan, line 850: let's add todo here. we should use platform api to get the dpu number instead of rely on the platform env config. @prgeor .

lguohan (dismissed):

dockers/docker-database/Dockerfile.j2
```diff
@@ -32,7 +32,8 @@ RUN apt-get clean -y                                  && \
              s/^# unixsocket/unixsocket/;                \
              s/redis-server.sock/redis.sock/g;           \
              s/^client-output-buffer-limit pubsub [0-9]+mb [0-9]+mb [0-9]+/client-output-buffer-limit pubsub 0 0 0/; \
-             s/^notify-keyspace-events ""$/notify-keyspace-events AKE/ \
+             s/^notify-keyspace-events ""$/notify-keyspace-events AKE/; \
+             s/^databases [0-9]+$/databases 100/ \
```
r12f, line 36: is this change expected? do we need 100 databases or it is changed for testing but later forget to remove

dockers/docker-database/database_config.json.j2
```diff
@@ -94,6 +95,30 @@
             "separator": ":",
             "instance" : "redis"
         }
+{% if DATABASE_TYPE is defined and DATABASE_TYPE == "dpudb" %}
+        ,
+        "DPU_APPL_DB" : {
```
r12f, line 100: maybe i missed some update, so just checking - have we decided to use DPU_APPL_DB or DASH_APPL_DB and etc.

dockers/docker-database/database_global.json.j2
```diff
@@ -15,7 +21,22 @@
         },
 {% endif %}
 {% endfor %}
+{% endif %}
+
+{% if dpu_cnt > 0 %}
+{% for dpu in range(dpu_cnt) %}
+        {
+            "database_type" : "dpudb",
+            "include" : "../../redisdpu{{dpu}}/sonic-db/database_config.json"
```
r12f, line 30: nit: i wonder if we could use "redis-dpu{{dpu}}"

r12f: and we can do it in another change, since this change is already approved.

r12f (approved):

dockers/docker-database/database_config.json.j2
```diff
@@ -2,9 +2,10 @@
     "INSTANCES": {
         "redis":{
             "hostname" : "{{HOST_IP}}",
-            "port" : 6379,
-            "unix_socket_path" : "/var/run/redis{{NAMESPACE_ID}}/redis.sock",
-            "persistence_for_warm_boot" : "yes"
+            "port" : {{REDIS_PORT}},
+            "unix_socket_path" : "/var/run/redis{{DEV}}/redis.sock",
+            "persistence_for_warm_boot" : "yes",
+            "database_type": "{{DATABASE_TYPE}}"
```
judyjoseph, line 8: should we add a check to add the field "database_type" if DATABASE_TYPE is defined?

dockers/docker-database/Dockerfile.j2
Pterosaur: The default value of databases is 16 which isn't enough for new databases of DPU.

dockers/docker-database/database_config.json.j2
Pterosaur: Yes, I want to introduce new databases like, `DPU_APPL_DB`, which follows chassis design. I feel it's clearer to keep the underlay and overlay objects in different databases.

dockers/docker-database/docker-database-init.sh
```diff
@@ -59,7 +71,7 @@ if [[ $DATABASE_TYPE == "chassisdb" ]]; then
 fi
 
 # copy/generate the database_global.json file if this is global database service in multi asic platform.
-if [[ $NAMESPACE_ID == "" ]] && [[ $NAMESPACE_COUNT -gt 1 ]]
+if [[ $DATABASE_TYPE == "" ]] && [[ $NAMESPACE_COUNT -gt 1 || $NUM_DPU -gt 1 ]]
```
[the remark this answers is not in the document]
Pterosaur: In the DPU scenario, all dpu databases are running in the host namespace which means the namespace_id is also empty to the dpu database instance. So, I use database_type == "" to identify the host db instance.

dockers/docker-database/database_global.json.j2
Pterosaur: It's not easy. Because the database folder consists of the prefix "redis" and suffix "$DEV" without the dash "-". This convention has been used in the multi-ASIC scenario.

dockers/docker-database/database_config.json.j2
Pterosaur: Done

src/sonic-py-common/sonic_py_common/device_info.py
Pterosaur: Done

dockers/docker-database/docker-database-init.sh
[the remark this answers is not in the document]
judyjoseph: From what I understand in case of dpu we set the DATABASE_TYPE as below, and for all other cases like single_asic/multi-asic DATABASE_TYPE will be ""

The earlier check below was applicable only to global db which is in host namespace in case of multi-asic/single asic 
```
if [[ $NAMESPACE_ID == "" ]] && [[ $NAMESPACE_COUNT -gt 1 ]]
```

So now with this change ```$NAMESPACE_ID == ""```  -------->  ```$DATABASE_TYPE == ""``` , for multi-asic platforms it will never go to the else part as DATABASE_TYPE is always "" ?.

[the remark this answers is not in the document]
Pterosaur: You are right, I will fix it.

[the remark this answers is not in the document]
Pterosaur: Fix it in PR: https://github.com/sonic-net/sonic-buildimage/pull/17221 Please help to check it.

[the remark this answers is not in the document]
judyjoseph: Taking a look.

PR #17171, opened 2023-11-14
Add SysLogger for new implementation.

xincunli-sonic:
This PR introduces the SysLogger to address an issue where the original Logger.py could override the SYSLOG_IDENTIFIER if a new Logger instance is created.

**Why I did it?**
This change was made because the SysLog was displaying `NOTICE pmon#CCmisApi:` instead of `NOTICE pmon#ycable:`. This discrepancy led to confusion during debugging and incorrect component identification in log analysis.

**How I did it**
I implemented an instance-level syslog using the `logging.handlers` module to prevent the SYSLOG_IDENTIFIER from being overridden.

**How did you verify/test it?**
1. Run code snippet

```python
import SysLogger
from daemon_base import DaemonBase

# 1. SysLogger
SYSLOG_IDENTIFIER = "TestSysLogger"

# Global logger instance
log = SysLogger.SysLogger(SYSLOG_IDENTIFIER)

log.log_info("This is a info message")
log.log_notice("This is a notice message")
log.log_warning("This is a warning message")
log.log_debug("This is a debug message")
log.log_error("This is a error message")

log.log_info("This is a info message", also_print_to_console=True)
log.log_notice("This is a notice message", also_print_to_console=True)
log.log_warning("This is a warning message", also_print_to_console=True)
log.log_debug("This is a debug message", also_print_to_console=True)
log.log_error("This is a error message", also_print_to_console=True)


# 2. Daemonbase with default syslog.

log2 = DaemonBase(SYSLOG_IDENTIFIER)

log2.log_info("log2: This is a info message")
log2.log_notice("log2: This is a notice message")
log2.log_warning("log2: This is a warning message")
log2.log_debug("log2: This is a debug message")
log2.log_error("log2: This is a error message")

log2.log_info("log2: This is a info message", also_print_to_console=True)
log2.log_notice("log2: This is a notice message", also_print_to_console=True)
log2.log_warning("log2: This is a warning message", also_print_to_console=True)
log2.log_debug("log2: This is a debug message", also_print_to_console=True)
log2.log_error("log2: This is a error message", also_print_to_console=True)

# 3. Daemonbase with SysLogger.

log3 = DaemonBase(SYSLOG_IDENTIFIER, use_syslogger=True)

log3.log_info("log3: This is a info message")
log3.log_notice("log3: This is a notice message")
log3.log_warning("log3: This is a warning message")
log3.log_debug("log3: This is a debug message")
log3.log_error("log3: This is a error message")

log3.log_info("log3: This is a info message", also_print_to_console=True)
log3.log_notice("log3: This is a notice message", also_print_to_console=True)
log3.log_warning("log3: This is a warning message", also_print_to_console=True)
log3.log_debug("log3: This is a debug message", also_print_to_console=True)
log3.log_error("log3: This is a error message", also_print_to_console=True)

```
3. Sample output

```log
Jan  5 20:28:18 xincun-dev-vm TestSysLogger: This is a info message
Jan  5 20:28:18 xincun-dev-vm TestSysLogger: This is a notice message
Jan  5 20:28:18 xincun-dev-vm TestSysLogger: This is a warning message
Jan  5 20:28:18 xincun-dev-vm TestSysLogger: This is a error message
Jan  5 20:28:18 xincun-dev-vm TestSysLogger: This is a info message
Jan  5 20:28:18 xincun-dev-vm TestSysLogger: This is a notice message
Jan  5 20:28:18 xincun-dev-vm TestSysLogger: This is a warning message
Jan  5 20:28:18 xincun-dev-vm TestSysLogger: This is a error message
Jan  5 20:28:18 xincun-dev-vm TestSysLogger[967382]: log2: This is a notice message
Jan  5 20:28:18 xincun-dev-vm TestSysLogger[967382]: log2: This is a warning message
Jan  5 20:28:18 xincun-dev-vm TestSysLogger[967382]: log2: This is a error message
Jan  5 20:28:18 xincun-dev-vm TestSysLogger[967382]: log2: This is a notice message
Jan  5 20:28:18 xincun-dev-vm TestSysLogger[967382]: log2: This is a warning message
Jan  5 20:28:18 xincun-dev-vm TestSysLogger[967382]: log2: This is a error message
Jan  5 20:28:18 xincun-dev-vm TestSysLogger: log3: This is a info message
Jan  5 20:28:18 xincun-dev-vm TestSysLogger: log3: This is a notice message
Jan  5 20:28:18 xincun-dev-vm TestSysLogger: log3: This is a warning message
Jan  5 20:28:18 xincun-dev-vm TestSysLogger: log3: This is a error message
Jan  5 20:28:18 xincun-dev-vm TestSysLogger: log3: This is a info message
Jan  5 20:28:18 xincun-dev-vm TestSysLogger: log3: This is a notice message
Jan  5 20:28:18 xincun-dev-vm TestSysLogger: log3: This is a warning message
Jan  5 20:28:18 xincun-dev-vm TestSysLogger: log3: This is a error message
```

**Which release branch to backport (provide reason below if selected)**
 - [ ] 202205
 - [ ] 202305
 - [ ] 202311

Signed-off-by: Xincun Li [xincun.li@microsoft.com](mailto:xincun.li@microsoft.com)


src/sonic-py-common/sonic_py_common/daemon_base.py
```diff
@@ -48,6 +65,20 @@ def __init__(self, log_identifier):
         if not signal.getsignal(signal.SIGTERM):
             signal.signal(signal.SIGTERM, self.signal_handler)
 
+    def log(self, priority, msg, also_print_to_console=False):
```
nazariig, line 68: @xincunli-sonic why do we need this? Where it is being used?

src/sonic-py-common/sonic_py_common/SysLogger.py
```diff
@@ -0,0 +1,70 @@
...
+    def set_min_log_priority(self, priority):
+        """
+        Sets the minimum log priority level. All log messages
+        with a priority lower than this will not be logged
+        """
+        self._min_log_level = self.LOG_LEVEL_MAP[priority]
+
+    # Methods for logging messages
+    def log(self, priority, msg, also_print_to_console=False):
+        log_level = self.LOG_LEVEL_MAP[priority]
```
nazariig, line 49: @xincunli-sonic can we use native interface instead of manually checking severity level?

xincunli-sonic: This was intentional for easily identify log level, but your suggestion is reasonable, hence removed.

src/sonic-py-common/sonic_py_common/daemon_base.py
xincunli-sonic: Current daemon_base is being used by most of PMON daemon process, and it is logger's subclass, not necessary to change them to inherit from syslogger, so this log function is going to be compatible between SysLogger and Logger.

src/sonic-py-common/sonic_py_common/daemon_base.py
```diff
@@ -48,6 +66,20 @@ def __init__(self, log_identifier):
...
```
nazariig, line 69: @xincunli-sonic IMHO, this is a hack and in general a bad architecture. The method you override, was not designed to be overridable. Also, the logger class is used in many other places, not just only a various python demons implementation. This means, that the issue appears to be in many places. Ideally, there should be two different implementations, which share the same interface (e.g., LoggerBase[defines logger interface API], SystemLogger[uses native syslog], SyslogLogger[uses syslog handler], FileLogger[uses file handler]). But that will lead up to a major code changes. I would suggest to abandon having a separate SyslogLogger implementation and just fix the original one, preserving all the defined set of APIs.

xincunli-sonic: This is intending to be reduce the scope of new change. Here is the consideration
1. Most of daemon process will inherit from Daemon_Base with old logger implementation. If we use the new SysLogger, then it could be used for all daemons.
2. What we want to do would be only apply new SysLogger to some new Daemon process, like ycabled etc.
3. The override function, you can think it like factory pattern, which depends on use_syslogger flag as optional parameter, if it is true then it will use new SysLogger, otherwise it keeps the old logger and behavior will not change.

nazariig (approved):

src/sonic-py-common/sonic_py_common/syslogger.py
```diff
@@ -0,0 +1,65 @@
...
+        for handler in self.logger.handlers[:]:
+            self.logger.removeHandler(handler)
+
+        # self.logger.setLevel(log_level)
+        handler = SysLogHandler(address="/dev/log", facility=log_facility, socktype=socket.SOCK_DGRAM)
+        formatter = logging.Formatter("%(name)s: %(message)s")
+        handler.setFormatter(formatter)
+        self.logger.addHandler(handler)
+
+        # Set the default minimum log priority to 'LOG_DEBUG'
```
prgeor, line 33: @xincunli-sonic comment says LOG_DEBUG, but its actually set to LOG_INFO

src/sonic-py-common/sonic_py_common/daemon_base.py
```diff
@@ -32,12 +36,16 @@ def db_connect(db_name, namespace=EMPTY_NAMESPACE):
 
 
 class DaemonBase(Logger):
-    def __init__(self, log_identifier):
-        super(DaemonBase, self).__init__(
-            log_identifier=log_identifier,
-            log_facility=Logger.LOG_FACILITY_DAEMON,
-            log_option=(Logger.LOG_OPTION_NDELAY | Logger.LOG_OPTION_PID)
-        )
+    def __init__(self, log_identifier, use_syslogger=False):
```
prgeor, line 39: @xincunli-sonic @nazariig can we have the default `use_syslogger`=True

nazariig: @prgeor totally agree. Better to force all the daemons/libraries to use a new implementation by default

src/sonic-py-common/sonic_py_common/syslogger.py
```diff
@@ -0,0 +1,64 @@
+import os
+import socket
+import sys
+import logging
```
prgeor, line 4: @xincunli-sonic is this import required. The next import is what we need?

src/sonic-py-common/sonic_py_common/daemon_base.py
```diff
@@ -32,12 +36,16 @@ def db_connect(db_name, namespace=EMPTY_NAMESPACE):
...
-    def __init__(self, log_identifier):
-        super(DaemonBase, self).__init__(
-            log_identifier=log_identifier,
-            log_facility=Logger.LOG_FACILITY_DAEMON,
-            log_option=(Logger.LOG_OPTION_NDELAY | Logger.LOG_OPTION_PID)
-        )
+    def __init__(self, log_identifier, use_syslogger=True):
+        super().__init__()
+        if use_syslogger:
+            self.logger_instance = SysLogger(log_identifier)
```
Junchao-Mellanox, line 42: it seems the new logger instance will lose PID in log message. Is it expected?

src/sonic-py-common/sonic_py_common/daemon_base.py
```diff
@@ -32,12 +36,16 @@ def db_connect(db_name, namespace=EMPTY_NAMESPACE):
 
 
 class DaemonBase(Logger):
```
Junchao-Mellanox, line 38: do we still need inherit from Logger?

src/sonic-py-common/sonic_py_common/syslogger.py
```diff
@@ -0,0 +1,62 @@
...
+from logging.handlers import SysLogHandler
+
+
+class SysLogger:
+    """
+    SysLogger class for Python applications using SysLogHandler
+    """
+
+    DEFAULT_LOG_FACILITY = SysLogHandler.LOG_USER
+    DEFAULT_LOG_LEVEL = SysLogHandler.LOG_INFO
```
Junchao-Mellanox, line 13: should default logger level be NOTICE? The old Logger class in logger.py is default to NOTICE level.

Junchao-Mellanox (changes requested):

Junchao-Mellanox:
What is the performance impaction of this change? Could you provide test data?

Junchao-Mellanox:
In XCVRD, there is a global logger instance, and there are also daemons. After your change, global logger instance will use log class `Logger`, daemons will use new log class "SysLogger". Will they affect each other?

Junchao-Mellanox:
I did some simple test with this new logger class. I found a few issues:
1. SysLogger.log_xxx does not work. It is probably caused by wrong log level used in the new log class "SysLogHandler.LOG_ERR". I guess we should not use it, but "logging.ERROR". (the same issue with INFO, NOTICE, WARNING)
```
if __name__ == "__main__":
    log1 = SysLogger(log_identifier="test1")
    log1.log_error("from log1")    #  DOES NOT WORK
    log1.logger.error('error')         # WORK
```

2. the new logger class cannot display hostname correctly. It only displays "localhost" instead of real hostname. I did not dig into it too much, no root cause found yet.

```
With old logger class, the message:
Mar 13 03:54:35.213887 sonic ERR test1: from log1

With new logger class, the message:
Mar 13 03:54:35.213887 localhost ERR test1: from log1
```

Junchao-Mellanox:
Hi @xincunli-sonic ,

I checked your latest code. It seems the host name issue has been fixed. But the following issue is still there:

1. SysLogger.log_xxx does not work. It is probably caused by wrong log level used in the new log class. For example, "SysLogHandler.LOG_ERR" is used for log_error. I guess we should not use it, but "logging.ERROR". (the same issue with INFO, NOTICE, WARNING). My test:
```
if __name__ == "__main__":
    log1 = SysLogger(log_identifier="test1")
    log1.log_error("from log1")    #  DOES NOT WORK
    log1.logger.error('error')         # WORK
    log1.log(logging.ERROR, "also work")  # WORK
```

Junchao-Mellanox (approved):

liat-grozovik:
@xincunli-sonic could you please help to prioritize this PR feedback and checkers?
this infra is required for additional PR related here as well and we should have it on 202405

Junchao-Mellanox:
Hi @xincunli-sonic , I suppose the checker failure is related to the code change. The unit test tried to connect to a unix socket of a logger server, but the unix socket does not exist. 
```
--- Logging error ---
  File "/usr/lib/python3.11/logging/handlers.py", line 1007, in emit
    self.socket.send(msg)
OSError: [Errno 9] Bad file descriptor

```

Junchao-Mellanox:
Hi @prgeor , could you please kindly review this?

prgeor (approved):

prgeor:
@liat-grozovik can you merge?

liat-grozovik:
i removed the backport ask for old branches as 202205 and 202305. if this is must lets discuss.
what about 202311? is it needed ? if so, lets add the label and discuss on it

PR #17203, opened 2023-11-16
Disable systemd auto-restart of dependent services for spineRouters

deepak-singhal0408:
#### Why I did it
This is cherry-pick PR.. Master branch PR. https://github.com/sonic-net/sonic-host-services/pull/83
Issue details and RCA here. https://github.com/sonic-net/sonic-buildimage/issues/16991

##### Work item tracking
- Microsoft ADO **25422412**:

#### How I did it
Currently hostcfgd script overrides the systemd service files of the features depending upon auto_restart enable/disable.
I am skipping dependent features(syncd, gbsyncd for now) to have "RESTART=Always"
for them to not start immediately, and instead get started by SWSS through swss.sh script.
The issue of syncd double stop is also applicable to pizza box platforms, however no traffic impact is seen there, whereas on VOQ chassis, we do see traffic impact due to early start of syncd service.
The change can be extended to pizza box platforms as well in future once it gets vetted on SpineRouters.

#### How to verify it
extended Unit tests to cover this scenario.
followed same steps as mentioned in https://github.com/sonic-net/sonic-buildimage/issues/16991 and confirmed that
a. No traffic loss end to end
b. Don't see double stop/start of syncd service
Also Ran few more tests.

Tested on both single asic and multi-asic hw platforms.
Tested config-reload
sudo reboot
service restart
syncd process crash, orchagent process crash
Confirmed from logs, that in none of the scenario, syncd gets killed twice

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

SONiC.20220532.45

- [ ] 
- [ ]

### DIFF
--- a/src/sonic-host-services/scripts/hostcfgd
+++ b/src/sonic-host-services/scripts/hostcfgd
@@ -364,7 +364,20 @@ class FeatureHandler(object):
         Returns:
             None.
         """
-        restart_field_str = "always" if "enabled" in feature_config.auto_restart else "no"
+        # As per the current code(due to various dependencies) SWSS service stop/start also stops/starts the dependent services(syncd, teamd, bgpd etc)
+        # There is an issue seen of syncd service getting stopped twice upon a critical process crash in syncd service due to above reason.
+        # Also early start of syncd service has traffic impact on VOQ chassis.
+        # to fix the above issue, we are disabling the auto restart of syncd service as it will be started by swss service.
+        # This change can be extended to other dependent services as well in future and also on pizza box platforms.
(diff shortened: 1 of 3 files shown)


arlakshm (approved):

deepak-singhal0408:
@lguohan @yxieca could you please help review/merge? Thanks!

deepak-singhal0408:
@gechiang FYI..

PR #17220, opened 2023-11-18
Add support to make determine/process reboot-cause services restartable

anamehra:
#### Why I did it
Fixes https://github.com/sonic-net/sonic-buildimage/issues/16990

Requires: https://github.com/sonic-net/sonic-host-services/pull/86

1. determine-reboot-cause  and process-reboot-cause service does not start If the database service fails to restart in the first attempt. Even if the Database service succeeds in next attempt, these reboot-cause services do not start.

2. The process-reboot-cause service also do not restart if the docker or database service restarts, which leads to an empty reboot-cause history

3. deploy-mg from sonic-mgmt also triggers the docker service restart. The restart of the docker service caused the issue stated in 2 above. The docker restart also triggers determine-reboot-cause to restart which creates an additional reboot-cause file in history and modifies the last reboot-cause.

This PR along with sonic-host-services PR 82 fixes these issues by making both processes start again when dependency meets after dependency failure, making both processes restart when the database service restarts, and preventing duplicate processing of the last reboot reason.

##### Work item tracking
- Microsoft ADO **25892856**

#### How I did it
1. Modified systemd unit files to make determine-reboot-cause and process-reboot-cause services restartable when the database service restarts.
2. On the restart, determine-reboot-cause service should not recreate a new reboot-cause entry in the database. Added check for first start or restart to skip entry for restart case.

#### How to verify it
On single ASIC pizza box:
1.  Installed the image and check reboot-cause history
2. restart the database service and verify that determine-reboot-cause and process-reboot-cause services also restart. Verify that reboot-cause shows correct data and no new entry is created for restart.

On Chassis:
1.  Installed the image and check reboot-cause history
2. restart the database service and verify that determine-reboot-cause and process-reboot-cause services also restart. Verify that reboot-cause shows correct data and no new entry is created for restart.
3. Reboot LC. On Supervicor, stop database-chassis service.
     Let database service on LC fail the first time. determine-reboot-cause and process-reboot-cause  would fail to start due to dependency failure
     start database-chassis on Supervisor. Database service on LC should now start successfully.
     Verify determine-reboot-cause and process-reboot-cause  also starts
     Verify show reboot-cause history output

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [x] 202205
- [ ] 202211
- [x] 202305

- [ ] 
- [ ] 

Add support to make determine/process reboot-cause services restartable


anamehra:
Hi @abdosi , @gechiang , please review. Thanks

gechiang:
MSFT ADO: 25892856

gechiang:
@anamehra , PR Tests are failing... Can you please take a look and address the failures.

prgeor (changes requested):
@anamehra we probably don't need this change provided we follow same approach of https://github.com/sonic-net/sonic-host-services/pull/86 in the unit service file of src/sonic-host-services-data/debian/sonic-host-services-data.determine-reboot-cause.service

prgeor:
@anamehra @gechiang lets discuss a better solution as suggested above before merging

anamehra:
> @anamehra we probably don't need this change provided we follow same approach of [sonic-net/sonic-host-services#86](https://github.com/sonic-net/sonic-host-services/pull/86) in the unit service file of src/sonic-host-services-data/debian/sonic-host-services-data.determine-reboot-cause.service

Hi @prgeor,  process-reboot-cause service is timer timer-based simple service, and that is why these changes are required to make it restartable. I was testing an approach by removing timer logic and making the process the same as determine-reboot-cause but systemd does not work to start the service when it fails due to dependency failure when the database service fails.

prgeor:
@anamehra you can use this unit file which is a cleaner approach than modifying the reboot cause script file 

image



Here is the systemd log showing skipping of additional runs of determine-reboot-cause service

image

anamehra:
The files are moved to host-services submodule. I will open a new PR in sonic-host-services repo

PR #17430, opened 2023-12-06
Change leaf value of used_cnt of sonic-events-swss:chk_crm_threshold

zbud-msft:
#### Why I did it

Current YANG model of sonic-events-swss:chk_crm_threshold has the type uint8 for leaf used_cnt which is too small of a range to hold values of used_cnt which can greatly exceed that. Updating leaf type of used_cnt and free_cnt to match defined definition.

Changed to uint32 as per defined here: https://github.com/sonic-net/sonic-swss/blob/master/orchagent/crmorch.h#L99

##### Work item tracking
- Microsoft ADO **(number only)**:26091912

#### How I did it

Update leaf value

#### How to verify it

UT and sonic-mgmt PR checker

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [x] 202305

- [ ] 
- [ ]

### DIFF
--- a/src/sonic-yang-models/yang-models/sonic-events-swss.yang
+++ b/src/sonic-yang-models/yang-models/sonic-events-swss.yang
@@ -102,11 +102,11 @@ module sonic-events-swss {
             }
 
             leaf used_cnt {
-                type uint64;
+                type uint32;
             }
 
             leaf free_cnt {
-                type uint64;
+                type uint32;
             }
 
             uses evtcmn:sonic-events-cmn;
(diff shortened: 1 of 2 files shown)


src/sonic-yang-models/yang-models/sonic-events-swss.yang
```diff
@@ -102,11 +102,11 @@ module sonic-events-swss {
             }
 
             leaf used_cnt {
-                type uint8;
+                type uint32;
```
qiluo-msft, line 105: How do you konw uint32 is large enough?

zbud-msft: Changed to uint32 as per defined here: https://github.com/sonic-net/sonic-swss/blob/master/orchagent/crmorch.h#L99

qiluo-msft (approved):

zbud-msft:
@StormLiangMS Can we backport to 202311?

PR #17616, opened 2023-12-25
Do not pass the option "device" in rsyslog.conf by default when syslog server's source address is configured

stephenxs:
#### Why I did it

An in-band syslog server will not receive any syslog if it is configured without a VRF specified, which is because `eth0` is always specified as the `device` of a syslog server and the syslog packets will be sent to `eth0` regardless of its destination IP address.

##### Work item tracking

#### How I did it

Pass the option "device" in rsyslog.conf only if when syslog server's source address is configured with a non-default VRF

#### How to verify it

Manually test:
1. Configuring a syslog server without VRF specified or with `default` as the VRF: no `device` passed in `rsyslog.conf`
2. Configuring a syslog server with non-default VRF: the configured VRF passed as `device` in `rsyslog.conf`

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/files/image_config/rsyslog/rsyslog.conf.j2
+++ b/files/image_config/rsyslog/rsyslog.conf.j2
@@ -107,7 +107,7 @@ $RepeatedMsgReduction on
 {% set regex = conf.get('filter_regex') -%}
 
 {% set fmodifier = '!' if filter == 'exclude' else '' %}
-{% set device = 'eth0' if vrf != '' and vrf != 'default' -%}
+{% set device = vrf if vrf != '' and vrf != 'default' -%}
 {% set template = 'WelfRemoteFormat' if format == 'welf' else 'SONiCFileFormat' -%}
 
 {# Server extra options -#}
--- a/files/image_config/rsyslog/rsyslog.conf.j2
+++ b/files/image_config/rsyslog/rsyslog.conf.j2
@@ -115,11 +115,16 @@ $RepeatedMsgReduction on
 
 {% if source -%}
     {% set options = options ~ ' Address="' ~ source ~ '"'-%}
+    {% set device = device if device != 'eth0' else '' -%}
+{% endif -%}
+
+{% if device -%}
+    {% set options = options ~ ' Device="' ~ device ~ '"'-%}
 {% endif -%}
 
 {% if filter %}
 :msg, {{ fmodifier }}ereregex, "{{ regex }}"
 {% endif %}
 *.{{ severity }}
-action(type="omfwd" Target="{{ server }}" Port="{{ port }}" Protocol="{{ proto }}" Device="{{ device }}" Template="{{ template }}"{{ options }})
+action(type="omfwd" Target="{{ server }}" Port="{{ port }}" Protocol="{{ proto }}" Template="{{ template }}"{{ options }})
 {% endfor %}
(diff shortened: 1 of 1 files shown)


StormLiangMS:
hi @qiluo-msft could you have expert to take a look at this PR? And there is an internal conflict.

qiluo-msft:
Is it covered by sonic-mgmt testcases?

stephenxs:
> Is it covered by sonic-mgmt testcases?

I do not think so. this was found by our internal QA test.

stephenxs:
> > > Is it covered by sonic-mgmt testcases?
> > 
> > 
> > I do not think so. this was found by our internal QA test.
> 
> Could you add a test for the behavior "an in-band syslog server will not receive any syslog if it is configured without a VRF specified"?

I'm not sure whether there is infra for such test cases. So, I need to check the details and get back to you
Is it possible to merge it first because it's a blocker to our 202305 release and then we will add the test case later?
Thanks

files/image_config/rsyslog/rsyslog.conf.j2
```diff
@@ -107,19 +107,24 @@ $RepeatedMsgReduction on
 {% set regex = conf.get('filter_regex') -%}
 
 {% set fmodifier = '!' if filter == 'exclude' else '' %}
-{% set device = 'eth0' if vrf == 'default' else vrf -%}
+{% set device = vrf if vrf != '' and vrf != 'default' -%}
```
qiluo-msft, line 110: >if [](http://example.com/codeflow?start=20&length=2)

There is no `else` after this `if`. Will it break?

qiluo-msft:
@prsunny Could you help review?

files/image_config/rsyslog/rsyslog.conf.j2
stephenxs: Hi
No.
If the condition `vrf != '' and vrf != 'default'` isn't satisfied, `device` is empty, which is expected.
In line [121](https://github.com/sonic-net/sonic-buildimage/pull/17616/files#diff-b5c3b035d28139ed77348578faf3d18d55884f61a98bb218fca31c993725a98eR121), the `device` is empty and option will be empty as well.
We do not pass `device` without `vrf` specified as a non-default one, which is the motivation of this PR.
This case is covered by the 1st manual test item `Configuring a syslog server without VRF specified or with default as the VRF: no device passed in rsyslog.conf`

qiluo-msft (approved):

qiluo-msft:
ADO 27353460
@StormLiangMS Please help backport.

StormLiangMS:
hi @stephenxs  @wen587  pls have test with 202305 on brcm/mellanox/cisco/marvell platform, would like to avoid regression.

stephenxs:
> hi @stephenxs @wen587 pls have test with 202305 on brcm/mellanox/cisco/marvell platform, would like to avoid regression.

@StormLiangMS this is platform independent feature. It was tested on 202305 on mellanox platform.

wen587:
Hi @StormLiangMS , I tested the rsyslog.conf generation in 202305 PROD. I tried with one MGMTdevice and one non-MGMTdevice patching the PR.
Both devices show the same difference as below: 
 
`action(type="omfwd" Target="some_ip" Port="514" Protocol="udp" Device="eth0")`
Changed to 
`action(type="omfwd" Target="some_ip" Port="514" Protocol="udp" Template="SONiCFileFormat")`
Since SONiCFileFormat is the default template. So it just eliminates the Device variable. There should be no regression.

PR #17701, opened 2024-01-08
Add the hostname field，When device/default_sku is set to l2

philo-micas:
#### Why I did it
According to the issue description（https://github.com/sonic-net/sonic-buildimage/issues/17700）, we also add the hostname field to the DEVICE_METADATA|localhost key in the configDB when configuring the l2 l3 mode.

##### Work item tracking

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]


theasianpianist:
our l2 config workflow includes a manually specified hostname (https://github.com/sonic-net/SONiC/wiki/L2-Switch-mode), could you add a check to see if the hostname is already set so that it's not overwritten.

PR #17767, opened 2024-01-12
[YANG] Add an infrastructure to write pytest tests for the YANG model.

oleksandrivantsiv:
#### Why I did it
The existing test cases and test data for YANG model tests are defined in JSON files. The tests are limited and require a huge amount of copy-paste to provide good coverage.
For the features that have a lot of similar attributes that need to be covered with the tests, the JSON files can be quite big and hard to maintain (for example CRM).
The debuggability of the JSON tests is almost zero. Which makes implementation of the new test cases much harder than it should be.  When the single test case fails it generates thousands of lines of output that make it hard to understand the issue. It is hard to find what failed, and what is the error message. It is impossible to re-run single test cases. Etc.

The Pytest provides an alternative way to implement the tests.
Both JSON tests and Pytests can co-exist. But it is suggested to use Pytest for the new tests.
The tests written in Pytest are compact, readable, and human-friendly. 
They give more flexibility in the testing of corner cases allow to leverage Python and Pytest capabilities to generate configuration and data for better coverage.

##### Work item tracking

#### How I did it
Add an infrastructure that allows to write model tests in Python.
Conver CRM, DASH CRM, and SmartSwitch tests to Pytest.

#### How to verify it
Build the YANG model package.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]


dgsudharsan (approved):

src/sonic-yang-models/tests/yang_model_pytests/test_dash_crm.py
```diff
@@ -0,0 +1,105 @@
...
+            "localhost": {
+                "switch_type": "dpu"
+            }
+        }
+    }
+}
+
+
+@pytest.fixture
+def data(request):
```
ganglyu, line 17: why do you need this fixture?

src/sonic-yang-models/tests/yang_model_pytests/test_crm.py
```diff
@@ -0,0 +1,92 @@
...
+    'ipv6_route',
+    'ipv4_nexthop',
+    'ipv6_nexthop',
+    'ipv4_neighbor',
+    'ipv6_neighbor',
+    'nexthop_group_member',
+    'nexthop_group',
+    'fdb_entry',
+    'mpls_nexthop',
+    'srv6_nexthop'
```
ganglyu, line 27: original test has acl_counter and dnat_entry, but this new test does not have them?

src/sonic-yang-models/tests/yang_model_pytests/conftest.py
```diff
@@ -0,0 +1,41 @@
+import pytest
+import yang as ly
+from json import dumps
+from glob import glob
+
+
+class YangModel:
+
+    model_dir = './yang-models'
```
qiluo-msft, line 9: >. [](http://example.com/codeflow?start=17&length=1)

Relative path means you assume the current folder must be some specific one. Could you not to assume the current folder?

src/sonic-yang-models/tests/yang_model_pytests/conftest.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def __init__(self) -> None:
+        self.ctx = None
```
qiluo-msft, line 12: >ctx [](http://example.com/codeflow?start=13&length=3)

This line can be removed.

src/sonic-yang-models/tests/yang_model_pytests/test_crm.py
oleksandrivantsiv: Good catch, thank you!

src/sonic-yang-models/tests/yang_model_pytests/test_dash_crm.py
oleksandrivantsiv: It is needed to pass a base "sonic-device_metadata" configuration into each test without copying it.

src/sonic-yang-models/tests/yang_model_pytests/conftest.py
oleksandrivantsiv: done

src/sonic-yang-models/tests/yang_model_pytests/conftest.py
oleksandrivantsiv: done

qiluo-msft (approved):

oleksandrivantsiv:
@qiluo-msft can we merge this PR?

qiluo-msft:
@ganglyu Do you have more concern or approve?

ganglyu (approved):

PR #17784, opened 2024-01-15
[dualtor] Disable zebra link-detect for vlan interfaces

lolyu:
#### Why I did it
If all vlan member ports are admin down, the vlan route is removed by FRR, so the packets to the vlan subnet will be forwarded with the default route, two issues:
1. route inconsistency between kernel and ASIC.
the vlan route in kernel still has nexthop as the vlan device, the vlan route in ASIC has nexthop as the default route.
2. no tunnel will be created.
As those traffic will be bounced back to T1s by the ASIC, they will not be able to be trapped to kernel to trigger the zero-mac tunnel route creation.

For dualtor, the downstream standby traffic should be tunneled to the active side.
To ensure the zero-mac tunnel route works as expected, the vlan route should always present in both the kernel and ASIC.
So let's disable the link-detect for vlan interfaces.

Fixes: https://github.com/sonic-net/sonic-buildimage/issues/16085

Signed-off-by: Longxiang Lyu <lolv@microsoft.com>

##### Work item tracking
- Microsoft ADO **(number only)**: 26325737

#### How I did it
Disable the `link-detect` flag in the frr zebra template.

#### How to verify it
UT and nightly run.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [x] 202205
- [x] 202211
- [x] 202305

- [ ] 
- [ ]


prsunny (approved):
lgtm. @lguohan , @theasianpianist for viz

prsunny:
@lolyu , do we want to backport to 202205/202305?

lolyu:
> @lolyu , do we want to backport to 202205/202305?

The vlan port admin down scenario is required for Libra, so 202205/202305 should have this.
@yxieca, any suggestions?

PR #17797, opened 2024-01-17
dhcrelay: Don't look up the ifindex for the fallback interface

saiarcot895:
#### Why I did it

Currently, whenever isc-dhcp-relay forwards a packet upstream, internally, it will try to send it on a "fallback" interface. My understanding is that this isn't meant to be a real interface, but instead is basically saying to use Linux's regular routing stack to route the packet appropriately (rather than having isc-dhcp-relay specify specifically which interface to use).

The problem is that on systems with a weak CPU, a large number of interfaces, and many upstream servers specified, this can introduce a noticeable delay in packets getting sent. The delay comes from trying to get the ifindex of the fallback interface. In one test case, it got to the point that only 2 packets could be processed per second. Because of this, dhcrelay will easily get backlogged and likely get to a point where packets get dropped in the kernel.

##### Work item tracking
- Microsoft ADO **(number only)**: 26273071

#### How I did it

Fix this by adding a check saying if we're using the fallback interface, then don't try to get the ifindex of this interface. We're never going to have an interface named this in SONiC.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [x] 202205
- [x] 202211
- [x] 202305

- [x] 20220531.40

### DIFF
--- a/src/isc-dhcp/patch/0016-Don-t-look-up-the-ifindex-for-fallback.patch
+++ b/src/isc-dhcp/patch/0016-Don-t-look-up-the-ifindex-for-fallback.patch
@@ -1,0 +1,34 @@
+From 079ff1bb570dae96c4ca513e210c9856e9cc75b0 Mon Sep 17 00:00:00 2001
+From: Saikrishna Arcot <sarcot@microsoft.com>
+Date: Wed, 10 Jan 2024 23:30:17 -0800
+Subject: [PATCH] Don't look up the ifindex for fallback
+
+If sending a packet on the "fallback" interface, then don't try to get the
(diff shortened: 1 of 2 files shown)


yxieca (approved):

kellyyeh (approved):

yaqiangz (approved):

StormLiangMS:
@saiarcot895  could you help to file separate PR for 202305?

StormLiangMS:
hi @yaqiangz  do we need this for 202405?

PR #17809, opened 2024-01-17
[201911][Arista] make fancontrol config more resilient on 7060CX-32S

Staphylo:
#### Why I did it

In some rare cases the fancontrol config is invalid for DCS-7060CX-32S
There is no determinism in the kernel allocation of i2c bus number nor hwmon ids
If a driver takes too long to initialize another driver might initialize first thus leading sysfs path changes

##### Work item tracking

#### How I did it

Updated the fancontrol configuration to use some wildcards where determinism might be a problem.

#### How to verify it

Run fancontrol with this new configuration

Make fancontrol config more resilient on DCS-7060CX-32S


Staphylo:
There's only one new change related `[build] Use public storage for public resources.` since I published this PR.
I can rebase this PR but I don't think it will change the outcome of the checks given the error log.

PR #17836, opened 2024-01-18
[chassis] update service_checker module to handle database-chassis service

abdosi:
What I did:

Added change service_checker.py to add `database-chassis` docker also as expected running container for chassis supervisor.  Change is inline with this chnage: https://github.com/sonic-net/sonic-buildimage/blob/master/files/image_config/monit/container_checker#L82

Why I did:
service_checker.py is used for critical process monitoring and `database-chassis` is critical service on Chasiss architecture on Supervisor.

How I verify:
Manual Veriifcation.


judyjoseph (approved):
LGTM

rlhui:
THis is needed for alert.

PR #17860, opened 2024-01-19
[chassis] [202305] Added support of isolating given LC in Chassis with TSA mode

abdosi:
Cherry-pick of PR: https://github.com/sonic-net/sonic-buildimage/pull/16732


StormLiangMS (approved):
LGTM

PR #17874, opened 2024-01-22
reduce the amount of logmessages

scoopex:
SONiC creates huge amount of logmessages caused by automated processes which perform "sudo" with the messages enlisted by the READ_ONLY_CMDS Cmnd_Alias.
This change prevents the logging of messages of the following type:
```
Jan 22 17:34:44.534159 st01-sw1g-r01-u32 NOTICE sudo:    \
       admin : TTY=pts/2 ; PWD=/home/admin ; USER=root ; \
       COMMAND=/usr/bin/docker ps
```
 If you use centralized logging for dozens and hundreds of switches, this always causes you to filter these messages out.


scoopex:
Ping

scoopex:
@qiluo-msft i will abandon this change at 2024-12-31 if i get no feedback for it.

PR #17907, opened 2024-01-26
Change tcp port range to support telemetry and gnmi

ganglyu:
#### Why I did it
A common cause of failure for the telemetry and gnmi server is port conflict. If other processes are using tcp ports 50051 and 50052, the server cannot start.

##### Work item tracking
- Microsoft ADO **(number only)**: 26538971

#### How I did it
Change local port range in sysctl, new port range is from 32768 to 50001.
And then service ports for telemetry and gnmi will not be used by automatic port assignments.
https://www.kernel.org/doc/html/latest/networking/ip-sysctl.html

#### How to verify it
Stop telemetry and gnmi service, and create new tcp connections, and they will not use 50051 and 50052.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]


files/image_config/sysctl/sysctl-net.conf
```diff
@@ -15,6 +15,7 @@ net.ipv4.conf.all.arp_announce=1
 net.ipv4.conf.all.arp_filter=0
 net.ipv4.conf.all.arp_notify=1
 net.ipv4.conf.all.arp_ignore=2
+net.ipv4.ip_local_reserved_ports=50051,50052
```
liuh-80, line 18: Here is current SONiC auto allocate port range: 32768    60999
If any other SONiC service in this range, also need add to here.

admin@vlab-01:~$ sudo sysctl -a | grep ip_local_
**net.ipv4.ip_local_port_range = 32768    60999**
net.ipv4.ip_local_reserved_ports =
admin@vlab-01:~$

liuh-80 (approved):

liuh-80 (approved):

zbud-msft (approved):

qiluo-msft (approved):

Issue #17945, opened 2024-01-30
sonic-db-cli  -n <asic-ns> CHASSIS_APP_DB EVAL  fails intermittently after config reload or load-minigraph 


saksarav-nokia:
#### Description
When the config reload or load-minigraph is done , swss dockers are restarted and when swss comes back up, swss.sh script cleans the SYSYTEM_NEIGH, SYSTEM_INTERFACE, SYSTEM_LAG_MEMBER_TABLE and SYSTEM_LAG_ID_TABLE entries added by this LC+asic from the CHASSIS_APP_DB table using sonic-db-cli -n <asic-ns> CHASSIS_APP_DB EVAL command.

Even though the EVAL command is called after  sonic-db-cli -n <asic-ns> CHASSIS_APP_DB PING is successful, the EVAL command fails sometimes.  

When we replaced the "sonic-db-cli -n <asic-ns> CHASSIS_APP_DB EVAL" with "redis-cli -h 10.6.0.100 -p 6380 -n 12 EVAL",  the issues are not seen.

#### Steps to reproduce the issue:
1. Run full OC test or write a  simple bash script which just does config reload or load-minigraph 100 times with 7 mins sleep between each config reload or load-migraph.
2. Check the syslog for "Invalid database name input : 'CHASSIS_APP_DB'  error
3. Check for orchagent crash in the remote LC's

#### Describe the results you received:
 1) Error in syslog with "Invalid database name input : 'CHASSIS_APP_DB' , with  Unable to connect to redis: Cannot assign requested address
 2) The orchagent crash in the remote LC's

#### Describe the results you expected:
No errors in syslog and no crash

#### Output of `show version`:
SONiC Software Version: SONiC.20220532.54
SONiC OS Version: 11
Distribution: Debian 11.8
Kernel: 5.10.0-23-2-amd64
Build commit: b9e6caad98
Build date: Tue Jan  9 00:13:06 UTC 2024
Built by: cloudtest@95bebd0dc000000

```
(paste your output here)
```
#### Output of `show techsupport`:

```
(paste your output here or download and attach the file here )
```

saksarav-nokia:
@arlakshm @judyjoseph @abdosi @deepak-singhal0408  @mlok-nokia  for viz

judyjoseph:
@qiluo-msft, below is the question from Shakthi 

```
[...]

PR #17962, opened 2024-01-31
Wait till CHASIS_APP_DB PING is successful,  host_name and asic_name are valid in CONIFG_DB before starting chassis-db-cleanup

saksarav-nokia:
#### Why I did it
This PR fixes the issue reported in Issu #17945 
We noticed that chassis db clean up is skipped sometimes when the CHASSIS_APP_DB PING fails. Also if host_name and asic_name are not written to CONIG_DB, it could pass the empty strings to CHASSIS_APP_DB EVAL commands.
The service hostname-config.service is restarted whenever the config-reload or load-minigraph is done and this services renames the file /etc/hosts to updates it with the new file. This interferes with swss@.service and when swss.sh script CHASSIS_APP_DPP when the /etc/hosts file is renamed, the error "Unable to connect to redis: Cannot assign requested address" is seen and the CHASSIS_APP_DB EVAL command fails. This causes the chassis db entries not getting cleaned up and causes orchagent crash in  remote LC's.

##### Work item tracking

#### How I did it
 Wait till CHASS_APP_DB PING is successful before checking for entries in CHASSIS_APP_DB table. Also wait till host_name and asic_name are valis in CONFIG_DB. 
Modified swss@.service to start after hostname-config.service

#### How to verify it
Ran a script with 200 times config reload & load-minigraph and verified that chassis db cleanup is done every time and the orchagent crash is not seen .

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [x ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/files/build_templates/per_namespace/swss.service.j2
+++ b/files/build_templates/per_namespace/swss.service.j2
@@ -11,7 +11,6 @@ Requires=opennsl-modules.service
 {% endif %}
 Requires=updategraph.service
 After=updategraph.service
-After=hostname-config.service
 BindsTo=sonic.target
 After=sonic.target
 Before=ntp-config.service
--- a/files/build_templates/per_namespace/swss.service.j2
+++ b/files/build_templates/per_namespace/swss.service.j2
@@ -11,6 +11,7 @@ Requires=opennsl-modules.service
 {% endif %}
 Requires=updategraph.service
 After=updategraph.service
+After=hostname-config.service
 BindsTo=sonic.target
 After=sonic.target
 Before=ntp-config.service
--- a/files/image_config/hostname/hostname-config.sh
+++ b/files/image_config/hostname/hostname-config.sh
@@ -11,10 +11,16 @@ fi
 echo $HOSTNAME > /etc/hostname
 hostname -F /etc/hostname
 
+#Don't update the /etc/hosts if hostname is not changed
+#This is to prevent intermittent redis_chassis.server reachability issue
+if [ $CURRENT_HOSTNAME == $HOSTNAME ] ;  then
+    exit 0
+fi
+
 # Remove the old hostname entry from hosts file.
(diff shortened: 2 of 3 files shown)


saksarav-nokia:
@judyjoseph for viz

saksarav-nokia:
@judyjoseph @arlakshm @abdosi , please review this PR

saksarav-nokia:
We ran the complete oc with this fix  and  the error "Unable to connect to redis: Cannot assign requested address" is not seen. Also  we didn't see the orchagent crash

judyjoseph (approved):
Change looks good to me

judyjoseph:
@arlakshm @abdosi for review as well

judyjoseph:
@saksarav-nokia, Trying to find an alternative solution here as this change to add hostname-config.service dependency will affect all platforms. 

I checked this script, can we add a specific check in this script to proceed with changes in /etc/hosts file only if HOSTNAME changes ?: https://github.com/sonic-net/sonic-buildimage/blob/a6a8d198b06d784a018c9a539b7428211974ebc3/files/image_config/hostname/hostname-config.sh#L10. 

That should help our case and we need not add this hostname-config.service dependency

saksarav-nokia:
@judyjoseph , I think that will also fix the issue. I will test it out and update the PR.

saksarav-nokia:
@judyjoseph , Addressed your comments and verified the changes and ensured the issue is not seen with current changes. Please review it.

judyjoseph (approved):
LGTM

judyjoseph:
@rlhui @lguohan please help merge this PR

gechiang:
MSFT ADO: 27704026

gechiang:
@yxieca , who can review/approve for 202311 for this PR?

Issue #17948, opened 2024-01-30
The "dhcp-server-ipv4:kea-dhcp4" process failed to start after a cold reboot


oleksandrivantsiv:
#### Description

The "dhcp-server-ipv4:kea-dhcp4" process failed to start after a cold reboot with the following message is the syslog:
```
Jan 30 20:39:37.819348 mtvr-leopard-01 INFO dhcp_server#supervisord: kea-dhcp4 kea-dhcp4 already running? Daemon::createPIDFile: PID: 31 exists, PID file: /usr/local/var/run/kea/kea-dhcp4.kea-dhcp4.pid
Jan 30 20:39:37.821479 mtvr-leopard-01 INFO dhcp_server#supervisord 2024-01-30 20:39:37,821 INFO exited: kea-dhcp4 (exit status 1; not expected)
Jan 30 20:39:37.822199 mtvr-leopard-01 INFO dhcp_server#supervisord 2024-01-30 20:39:37,822 INFO gave up: kea-dhcp4 entered FATAL state, too many start retries too quickly
Jan 30 20:39:37.887478 mtvr-leopard-01 INFO dhcp_server#supervisord 2024-01-30 20:39:37,887 INFO exited: dependent-startup (exit status 0; expected)
```
The status of the process reported by supervisor is "FATAL":
```
# supervisorctl status
dependent-startup                EXITED    Jan 30 08:39 PM
dhcp-server-ipv4:dhcpservd       RUNNING   pid 43, uptime 0:05:13
dhcp-server-ipv4:kea-dhcp4       FATAL     Exited too quickly (process log may have details)
rsyslogd                         RUNNING   pid 33, uptime 0:05:16
start                            EXITED    Jan 30 08:39 PM
supervisor-proc-exit-listener    RUNNING   pid 30, uptime 0:05:17
```

#### Steps to reproduce the issue:
1. Install and run the image with the DHCP server container enabled
2. Enable DHCP sever feature and apply configuration:
```
{
    "FEATURE": {
        "dhcp_relay": {
            "auto_restart": "enabled",
            "state": "enabled"
        },
        "dhcp_server": {
            "auto_restart": "enabled",
            "state": "enabled"
        }
    },
    "VLAN": {
        "Vlan4094": {
            "vlanid": "4094"
        }
    },
    "VLAN_INTERFACE": {
        "Vlan4094": {},
        "Vlan4094|10.255.255.254/24": {}
    },
    "VLAN_MEMBER": {
[...]

PR #17974, opened 2024-01-31
[dhcp-server] Change the kea-dhcp4 PID file directory to tmpfs.

oleksandrivantsiv:
#### Why I did it
Fix the issue with running "dhcp-server-ipv4:kea-dhcp4" process after a cold reboot: #17948

##### Work item tracking

#### How I did it
The root cause of the issue is that during the cold reboot, the PID file of the process (that is stored on the persisted FS) is not cleared. From the [KEA DHCP server documentation](https://kea.readthedocs.io/en/kea-2.2.0/arm/dhcp4-srv.html):

> If the file already exists and contains the PID of a live process, the server issues a DHCP4_ALREADY_RUNNING log message and exits. It is possible, though unlikely, that the file is a remnant of a system crash and the process to which the PID belongs is unrelated to Kea. In such a case, it would be necessary to manually delete the PID file.

To make sure that the PID file is always cleared during the dhcp-server container restart the PID file should be stored on the tmpfs. 

#### How to verify it

Go over the steps to reproduce described in #17948.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/dockers/docker-dhcp-server/supervisord.conf
+++ b/dockers/docker-dhcp-server/supervisord.conf
@@ -61,3 +61,4 @@ stdout_logfile=syslog
 stderr_logfile=syslog
 dependent_startup=true
 dependent_startup_wait_for=dhcpservd:running
+environment=KEA_PIDFILE_DIR=/tmp/


yaqiangz (approved):
Thanks for the fix!

yaqiangz:
MSFT ADO number: 26640026

yaqiangz:
@yxieca Could you please help to merge this PR?

PR #18041, opened 2024-02-05
[Celestica] Make 'sfputil show presence' 10x faster

dal00:
#### Why I did it
The is_host function is used when enumerating SFPs which was very slow when detecting if it was running inside docker. The following call makes the detection 10x faster.

#### How I did it
Instead of executing `docker` check for `/.dockerenv` which is present when inside a container. 

#### How to verify it
Before patch:
```
  $ time sudo sfputil show presence
  Port         Presence
  -----------  -----------
  Ethernet0    Present
  Ethernet1    Present
  Ethernet2    Present

  -- snip --

  real    0m13.521s
  user    0m10.239s
  sys     0m4.104s
```
After patch:
```  
  $ time sudo sfputil show presence
  Port         Presence
  -----------  -----------
  Ethernet0    Present
  Ethernet1    Present
  Ethernet2    Present

  -- snip --

  real    0m2.133s
  user    0m1.461s
  sys     0m0.347s
```

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [x] 202211
- [x] 202305

- [x] 202211

n/a

n/a

image

### DIFF
--- a/platform/broadcom/sonic-platform-modules-cel/services/platform_api/sonic_platform/common.py
+++ b/platform/broadcom/sonic-platform-modules-cel/services/platform_api/sonic_platform/common.py
@@ -25,7 +25,6 @@ class Common:
 
     SET_METHOD_IPMI = 'ipmitool'
     NULL_VAL = 'N/A'
-    HOST_CHK_CMD = ["docker"]
     REF_KEY = '$ref:'
 
     def __init__(self, conf=None):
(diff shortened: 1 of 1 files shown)


qnos (approved):

dal00:
@lguohan Is there still something needed before this can be reviewed?

bluecmd:
Ping @lguohan

bluecmd:
Ping @lguohan - one last attempt before fork!

bradh352:
The only thing I can think of why they haven't merged this is maybe they don't like this change in how the check works for fear issues without Celestica's approval?

Technically the more proper thing to do is look at `/proc/1/cgroup` and if it's "/" then its the host, otherwise you're running in a container of some sort.  But though more proper its probably not as easy to accept as keeping the check the same.

Perhaps more acceptable would be to cache the result instead inside of __init__, so you're actually still using the same container check.  Probably rename `is_host()` to `_is_host()` to fetch the result and create a new `is_host()` that just returns the cached variable.

Issue #17792, opened 2024-01-16
[systemd][teamsyncd] missing logs after stopping system logging service


ayurkiv-nvda:
#### Description
Sometimes some logs may be lost due to syslog restart

#### Steps to reproduce the issue:
1. run "fast-reboot" (most probably problem is generic, we just manged to catch it during advanced-reboot test (fast-reboot) because  this test is looking for specific logs in syslog)
2. check logs

#### Describe the results you received:
```
Jan 12 07:02:12.808502 arc-switch1004 NOTICE teamd#teamsyncd: :- applyState: Applying state
Jan 12 07:02:12.808502 arc-switch1004 NOTICE teamd#teamsyncd: :- dump: getting took 0.001219 sec
Jan 12 07:02:12.798364 arc-switch1004 INFO systemd[1]: Stopping System Logging Service...
Jan 12 07:02:12.809352 arc-switch1004 INFO rsyslogd: [origin software="rsyslogd" swVersion="8.2302.0" x-pid="385" x-info="https://www.rsyslog.com"] exiting on signal 15.
Jan 12 07:02:13.799523 arc-switch1004 DEBUG container: container_wait: BEGIN
```
According to logs, it looks like **tlm_teamd** started and all is up and running. **"INFO systemd[1]: Stopping System  Logging Service  message"** means that syslog is restarting in parallel and at this moment we can lose some messages.
Syslog restart is expected flow on boot because config is changed in runtime (and we need to restart it in order to apply changes).

#### Describe the results you expected:
```
Jan 16 00:25:01.751031 r-tigon-04 NOTICE teamd#teamsyncd: :- applyState: Applying state
Jan 16 00:25:01.751943 r-tigon-04 NOTICE teamd#teamsyncd: :- dump: getting took 0.000604 sec
Jan 16 00:25:01.752638 r-tigon-04 NOTICE teamd#teamsyncd: :- dump: getting took 0.000301 sec
Jan 16 00:25:01.753146 r-tigon-04 NOTICE teamd#teamsyncd: :- setWarmStartState: teamsyncd warm start state changed to reconciled
Jan 16 00:25:01.753475 r-tigon-04 NOTICE teamd#tlm_teamd: :- try_add_lag: The LAG 'PortChannel104' has been added.
Jan 16 00:25:01.754515 r-tigon-04 NOTICE teamd#tlm_teamd: :- try_add_lag: The LAG 'PortChannel103' has been added.
[...]

PR #17458, opened 2023-12-08
Optimize syslog rate limit feature for fast and warm boot

Junchao-Mellanox:
#### Why I did it

Optimize syslog rate limit feature for fast and warm boot

##### Work item tracking
- Microsoft ADO **(number only)**:

#### How I did it

1. Optimize redis start time
2. Don't render rsyslog.conf in container startup script
3. Disable containercfgd by default. There is a new CLI to enable it (in another PR)

#### How to verify it

Manual test
Regression test

#### Which release branch to backport (provide reason below if selected)

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [x] 202305
- [X] 202311 

#### Tested branch (Please provide the tested image version)

- [ ] 
- [ ]

PR #12490, opened 2022-10-25
[infra] Support syslog rate limit configuration

Junchao-Mellanox:
HLD: https://github.com/sonic-net/SONiC/pull/1049

Depends on https://github.com/sonic-net/sonic-buildimage/pull/12489 (Build dependency)

#### Why I did it

Support syslog rate limit configuration feature

#### How I did it

1. Remove unused rsyslog.conf from containers
2. Modify docker startup script to generate rsyslog.conf from template files
3. Add metadata/init data for syslog rate limit configuration

#### How to verify it

Manual test
New sonic-mgmt regression cases

#### Which release branch to backport (provide reason below if selected)

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205

PR #18113, opened 2024-02-16
[rsyslog]: Use RELP instead of UDP for forwarding from container to host

saiarcot895:
#### Why I did it

When the host's rsyslog is restarted (for example, to regenerate the config after some changes, or as part of some automated script), there is a chance that some syslog messages from the containers are lost. Most of the time, this isn't an issue. However, if there are test cases that expect all syslogs to be present (such as the advanced-reboot test case), then this can cause a problem. Additionally, this could affect debuggability of issues where a rsyslog restart happens in the middle.

There are two options for reliable message transport in rsyslog: TCP and RELP. With TCP, while the protocol knows whether a syslog message has been delivered or not, the application doesn't know, because there is no feedback from the remote side saying the message was received. This means that there is still a chance that messages could be lost when the connection is broken (if, for example, the host rsyslog gets restarted), because after the connection is established, the sender rsyslog (in the
container) doesn't know if the message has been received or not.

RELP instead adds a feedback mechanism where the remote side notifies the sender whether the message has actually been received or not. This makes it much less likely to lose a message. There is one known possible case where a message (or messages) could be lost: the network is down, and rsyslog gets restarted. This at least requires both the network and rsyslog to have an issue, rather than just one. There is also a slim possibility where a message could get duplicated; this should be mostly fine (hopefully).

RELP does require that both sides are using a recent version of rsyslogd (at least 7.3.16, which looks like it was released more than 10 years ago), but since we use Debian on both the container and the host, it should be fine.

Therefore, switch to using RELP when sending syslog messages from the container to the host.

##### Work item tracking

#### How I did it

Modify the rsyslog.conf file on the host and the container to use RELP instead of UDP.

#### How to verify it

Stop rsyslogd on the host, make sure that the containers generate some syslogs, restart rsyslogd on the host, and verify no logs were lost.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/dockers/docker-base-bullseye/etc/rsyslog.conf
+++ b/dockers/docker-base-bullseye/etc/rsyslog.conf
@@ -9,24 +9,23 @@
 #### MODULES ####
 #################
 
-$ModLoad imuxsock # provides support for local system logging
-
 #
 # Set a rate limit on messages from the container
 #
 $SystemLogRateLimitInterval 300
 $SystemLogRateLimitBurst 20000
 
-#$ModLoad imklog  # provides kernel logging support
-#$ModLoad immark  # provides --MARK-- message capability
+module(load="imuxsock") # provides support for local system logging
+#module(load="imklog")   # provides kernel logging support
+#module(load="immark")  # provides --MARK-- message capability
 
 # provides UDP syslog reception
-#$ModLoad imudp
-#$UDPServerRun 514
+#module(load="imudp")
+#input(type="imudp" port="514")
 
 # provides TCP syslog reception
-#$ModLoad imtcp
-#$InputTCPServerRun 514
+#module(load="imtcp")
+#input(type="imtcp" port="514")
 
 
 ###########################
@@ -37,7 +36,8 @@ set $.CONTAINER_NAME=getenv("CONTAINER_NAME");
 
 # Set remote syslog server
 template (name="ForwardFormatInContainer" type="string" string="<%PRI%>%TIMESTAMP:::date-rfc3339% %HOSTNAME% %$.CONTAINER_NAME%#%syslogtag%%msg:::sp-if-no-1st-sp%%msg%")
-*.* action(type="omfwd" target=`echo $SYSLOG_TARGET_IP` port="514" protocol="udp" Template="ForwardFormatInContainer")
+module(load="omrelp")
+*.* action(type="omrelp" target=`echo $SYSLOG_TARGET_IP` port="2514" Template="ForwardFormatInContainer")
 
 #
 # Use traditional timestamp format.
--- a/files/image_config/rsyslog/rsyslog.conf.j2
+++ b/files/image_config/rsyslog/rsyslog.conf.j2
@@ -13,8 +13,6 @@
 #### MODULES ####
 #################
 
-$ModLoad imuxsock # provides support for local system logging
-
 {% set gconf = (SYSLOG_CONFIG | d({})).get('GLOBAL', {}) -%}
 {% set rate_limit_interval = gconf.get('rate_limit_interval') %}
 {% set rate_limit_burst = gconf.get('rate_limit_burst') %}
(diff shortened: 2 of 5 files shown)


dockers/docker-base-bookworm/etc/rsyslog.conf
```diff
@@ -9,24 +9,20 @@
...
 # Set a rate limit on messages from the container
 #
-$SystemLogRateLimitInterval 300
-$SystemLogRateLimitBurst 20000
-
-#$ModLoad imklog  # provides kernel logging support
-#$ModLoad immark  # provides --MARK-- message capability
+module(load="imuxsock" SysSock.RateLimit.Interval="300" SysSock.RateLimit.Burst="20000") # provides support for local system logging
+#module(load="imklog")   # provides kernel logging support
+#module(load="immark")  # provides --MARK-- message capability
```
prgeor, line 17: @saiarcot895 can you mention this syntax change in PR description

dockers/docker-base-bookworm/etc/rsyslog.conf
```diff
@@ -37,7 +33,8 @@ set $.CONTAINER_NAME=getenv("CONTAINER_NAME");
 
 # Set remote syslog server
 template (name="ForwardFormatInContainer" type="string" string="<%PRI%>%TIMESTAMP:::date-rfc3339% %HOSTNAME% %$.CONTAINER_NAME%#%syslogtag%%msg:::sp-if-no-1st-sp%%msg%")
-*.* action(type="omfwd" target=`echo $SYSLOG_TARGET_IP` port="514" protocol="udp" Template="ForwardFormatInContainer")
+module(load="omrelp")
+*.* action(type="omrelp" target=`echo $SYSLOG_TARGET_IP` port="2514" action.resumeRetryCount="-1" queue.type="LinkedList" Template="ForwardFormatInContainer")
```
prgeor, line 37: @saiarcot895 port changed from 514 to 2514 for?

dockers/docker-base-bullseye/etc/rsyslog.conf
prgeor, line 1: @saiarcot895 are we still using bullseye?

prgeor (commented):
what about `platform/vs/docker-sonic-vs/etc/rsyslog.conf` don't need this change?

files/image_config/rsyslog/rsyslog.conf.j2
```diff
@@ -13,34 +13,28 @@
...
-$SystemLogRateLimitBurst {{ rate_limit_burst }}
-{% endif %}
-
-$ModLoad imklog   # provides kernel logging support
-#$ModLoad immark  # provides --MARK-- message capability
+module(load="imuxsock" {% if rate_limit_interval is not none %}SysSock.RateLimit.Interval="{{ rate_limit_interval }}"{% endif %} {% if rate_limit_burst is not none %}SysSock.RateLimit.Burst="{{ rate_limit_burst }}"{% endif %}) # provides support for local system logging
+module(load="imklog")   # provides kernel logging support
+#module(load="immark")  # provides --MARK-- message capability
 
 # provides UDP syslog reception
```
prgeor, line 24: @saiarcot895 This UDP syslog is for remote server?

dockers/docker-base-bookworm/etc/rsyslog.conf
saiarcot895: There's no standard port used by RELP. The default port that rsyslog uses is 514, but that can conflict with regular syslog forwarding over TCP. A couple of the examples in the documentation for this feature uses 2514.

dockers/docker-base-bullseye/etc/rsyslog.conf
saiarcot895: Yes, bullseye is still being used for a couple containers.

dockers/docker-base-bookworm/etc/rsyslog.conf
saiarcot895: Done.

files/image_config/rsyslog/rsyslog.conf.j2
saiarcot895: Yes, in the case of a remote syslog server sending over UDP.

saiarcot895:
> what about `platform/vs/docker-sonic-vs/etc/rsyslog.conf` don't need this change?

Strictly speaking, it doesn't need this change, because the logs aren't actually being forwarded anywhere. It'll forward it to localhost port 514, but there likely won't be anything listening on this port. That container doesn't end up on the device.

It would be nice to update the syntax there to have it use the new syntax, but I'll keep that separate for now.

prgeor (dismissed):

qiluo-msft:
This is kind of new feature. Is it possible to config for new behavior or old behavior?

liat-grozovik:
@saiarcot895  is there any plan to have this fix some how handled and merged?

saiarcot895:
> This is kind of new feature. Is it possible to config for new behavior or old behavior?

@qiluo-msft While it could be added as a configurable feature, this would require an application (or daemon) running in each of the containers to determine which method to use (RELP or UDP). The last time such a feature was added (#12490), it had to be reworked to be disabled by default (#17458). Given that there are currently issues seen on cold bootup (#25382) and warm reboot (#17792), I'd like this to be enabled by default for simplicity.

PR #18158, opened 2024-02-22
Fix missing gnupg package issue during installing k8s gpg file process

lixiaoyuner:
#### Why I did it
When try to install k8s gpg file, there reports an error.
- "E: gnupg, gnupg2 and gnupg1 do not seem to be installed, but one of them is required for this operation   curl: (23) Failed writing body". 

To fix error we need to install gunpg package. But apt-key add is going to be deprecated, mv the gpg file to /etc/apt/trusted.gpg.d/ directly
##### Work item tracking
- Microsoft ADO **(number only)**: 26901718

#### How I did it
Download k8s gpg file and put it in /etc/apt/trusted.gpg.d/
#### How to verify it
We can install k8s gpg file successfully, but no gnupg package installed inside sonic image.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/build_debian.sh
+++ b/build_debian.sh
@@ -271,11 +271,9 @@ sudo LANG=C chroot $FILESYSTEM_ROOT apt-get -y install docker-ce=${DOCKER_VERSIO
 
 install_kubernetes () {
     local ver="$1"
-    sudo LANG=C chroot $FILESYSTEM_ROOT apt-get -y install gnupg
     sudo https_proxy=$https_proxy LANG=C chroot $FILESYSTEM_ROOT curl -fsSL \
         https://packages.cloud.google.com/apt/doc/apt-key.gpg | \
-        sudo LANG=C chroot $FILESYSTEM_ROOT apt-key add -
-    sudo LANG=C chroot $FILESYSTEM_ROOT apt-get -y purge gnupg
+        sudo LANG=C chroot $FILESYSTEM_ROOT tee /etc/apt/trusted.gpg.d/kubernetes.asc
     ## Check out the sources list update matches current Debian version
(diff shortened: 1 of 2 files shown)


qiluo-msft (dismissed):

qiluo-msft:
@saiarcot895 Could you also help review?

build_debian.sh
```diff
@@ -271,9 +271,11 @@ sudo LANG=C chroot $FILESYSTEM_ROOT apt-get -y install docker-ce=${DOCKER_VERSIO
 
 install_kubernetes () {
     local ver="$1"
+    sudo LANG=C chroot $FILESYSTEM_ROOT apt-get -y install gnupg
     sudo https_proxy=$https_proxy LANG=C chroot $FILESYSTEM_ROOT curl -fsSL \
         https://packages.cloud.google.com/apt/doc/apt-key.gpg | \
         sudo LANG=C chroot $FILESYSTEM_ROOT apt-key add -
```
saiarcot895, line 277: `apt-key` is deprecated and may be removed soon. The suggestion is to just copy the file that is downloaded to `/etc/apt/trusted.gpg.d/`. (Note that you may need to change the suffix from `.gpg` to `.asc`, depending on the actual key format that is used.)

saiarcot895: This will also remove the gnupg dependency.

lixiaoyuner: Thanks, very helpful, changed the code as your comments.

saiarcot895 (approved):

qiluo-msft (approved):

PR #18161, opened 2024-02-22
[Nokia-7215-T1] Disable sysrq-trigger from platform init

Pavan-Nokia:
#### Why I did it
Disable sysrq  invocation by keyboard and terminal server to prevent accidently triggering it under console overload conditions and performing unintentional actions

##### Work item tracking.
- Microsoft ADO **(number only)**:17610243

#### How I did it
Disable sysrq by writing 0 into  "/proc/sys/kernel/sysrq" register

#### How to verify it

1) Verify the register value is 0. also verify that sysrq cannot be triggered by sending a break character via the terminal server.

```
root@sonic:# cat /proc/sys/kernel/sysrq
0
root@sonic:#
telnet> send brk
hh
-bash: hh: command not found
root@sonic:#
```

2) Verify sysrq can still be triggered from procfs

```
root@sonic:# echo u > /proc/sysrq-trigger 
root@sonic:#  sysrq: EmergeRemount R/O

root@sonic:# echo "hi" > /tmp/abc.txt
-bash: /tmp/abc.txt: Read-only file system
```

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [x] 202205
- [ ] 202211
- [x] 202305
- [x] 202311

- [ ] 
- [ ]

### DIFF
--- a/platform/marvell-armhf/sonic-platform-nokia/7215/scripts/nokia-7215init.sh
+++ b/platform/marvell-armhf/sonic-platform-nokia/7215/scripts/nokia-7215init.sh
@@ -30,6 +30,9 @@ nokia_7215_profile()
 # Install kernel drivers required for i2c bus access
 load_kernel_drivers
 
+# Disable sysrq-trigger
+echo 0 > /proc/sys/kernel/sysrq
+
 # LOGIC to enumerate SFP eeprom devices - send 0x50 to kernel i2c driver - initialize devices
 # the mux may be enumerated at number 4 or 5 so we check for the mux and skip if needed
 


Pavan-Nokia:
@jon-nokia @carl-nokia

prgeor:
@Pavan-Nokia How are you running the telnet command?  As you can see below, its not working on 7215

image

platform/marvell-armhf/sonic-platform-nokia/7215/scripts/nokia-7215init.sh
```diff
@@ -30,6 +30,9 @@ nokia_7215_profile()
 # Install kernel drivers required for i2c bus access
 load_kernel_drivers
 
+# Disable sysrq-trigger
+echo 0 > /proc/sys/kernel/sysrq
```
prgeor, line 34: @Pavan-Nokia As I understand, there is still a small window of opportunity during boot which may trigger sysrq commands?

Pavan-Nokia:
> @Pavan-Nokia How are you running the telnet command? As you can see below, its not working on 7215
> 
> image

the way we connect to the 7215 via the console is using a telnet command 
for example:
image



and as this message here says the escape character to get to telnet prompt is "ctrl+ ]" 
and from there I am sending a break

from the error message I see in your screenshot you have not connected to any device via telnet. so it doesn't understand where to send

platform/marvell-armhf/sonic-platform-nokia/7215/scripts/nokia-7215init.sh
prgeor: @Pavan-Nokia How did you rule out this window of time is NOT impacted on the MSFT prod scenario?

Pavan-Nokia: @prgeor What we have learnt from our debug sessions is that the MSFT scripts start sending data only when they see a prompt. This change is in the platform init which runs during postinstall, before the getty service which enables the prompt. 

The only way to get full protection from sysrq is to make a sonic-linux-kernel change to disable this feature completely this would remove any window / chance for failure on any platform

prgeor (approved):

prgeor:
@lguohan, @yxieca please help merge

Blueve (approved):

PR #18204, opened 2024-02-28
Adding a new field to CONFIG DB: "subport"

tshalvi:
#### Why I did it
The field 'subport' represents the index of the split port within a physical port. For example, if a port is split into 4, the subport of the first logical port is 1, the subport of the second logical port is 2, and so on.
In xcvrd, the CMIS manager uses the subport to calculate the lane mask, which is used to control the data path per lane. In Nvidia platform, the subport is missing and is always set to 0. According to the xcvrd code, when subport=0, it will always correspond to the first logical port. Therefore, if we shut down any logical port that is not the first one, we will see the operational status of the first logical port also becomes down.
This PR aims to add the subport field to CONFIG DB and prevent such scenarios. This is applicable only for static default breakout mode. For DPB, subport calculation will happen on the fly (changes are not in Sonic yet).
(Subport HLD: HLD of subport: [link to the HLD document])

##### Work item tracking

#### How I did it
I have added the 'subport' field to all relevant Nvidia hwsku.json files (minigraph generation is based on them). Additionally, I introduced the new 'subport' field to portconfig.py, so that sonic-cfggen will be able to generate the minigraph with it. In this file, I also fixed an error that caused all attributes from hwsku.json to be applied only to the first logical ports associated with a physical port.
Furthermore, I updated hwsku_json_checker to include the new field and applied a fix to the sample_hwsku.json file. sample_hwsku.json is the file that sonic-config-engine's unit tests rely on for its tests. Previously, it only included attributes for the first logical port of a split physical port. For example, if Ethernet4, a 4-lane port, was split into 2 ports, then sample_hwsku.json included only the entry for Ethernet4, with no entry for Ethernet6. This misalignment with the structure of other hwsku.json files has been corrected as well.

#### How to verify it
Ensure that each logical port has the correct value of 'subport' in CONFIG DB, and that shutting down a logical port affects only that port and not other ports in the split.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/device/mellanox/x86_64-mlnx_msn4700-r0/ACS-MSN4700/hwsku.json
+++ b/device/mellanox/x86_64-mlnx_msn4700-r0/ACS-MSN4700/hwsku.json
@@ -1,100 +1,132 @@
 {
     "interfaces": {
         "Ethernet0": {
-            "default_brkout_mode": "1x400G[200G,100G,50G,40G,25G,10G,1G]"
+            "default_brkout_mode": "1x400G[200G,100G,50G,40G,25G,10G,1G]",
+            "subport": "1"
         },
         "Ethernet8": {
-            "default_brkout_mode": "1x400G[200G,100G,50G,40G,25G,10G,1G]"
+            "default_brkout_mode": "1x400G[200G,100G,50G,40G,25G,10G,1G]",
+            "subport": "1"
         },
         "Ethernet16": {
-            "default_brkout_mode": "1x400G[200G,100G,50G,40G,25G,10G,1G]"
+            "default_brkout_mode": "1x400G[200G,100G,50G,40G,25G,10G,1G]",
+            "subport": "1"
         },
         "Ethernet24": {
-            "default_brkout_mode": "1x400G[200G,100G,50G,40G,25G,10G,1G]"
+            "default_brkout_mode": "1x400G[200G,100G,50G,40G,25G,10G,1G]",
+            "subport": "1"
         },
         "Ethernet32": {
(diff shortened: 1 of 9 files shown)


prgeor (approved):

liat-grozovik (approved):

dprital:
@yxieca - Can you please cherry pick to 202311 ?

yxieca:
@tshalvi @dprital do we have matching db_migrator change to cover warm reboot upgrade path?

tshalvi:
> @tshalvi @dprital do we have matching db_migrator change to cover warm reboot upgrade path?

Not yet, but there is an action item for @prgeor regarding this:
https://github.com/sonic-net/sonic-buildimage/issues/17862

tshalvi:
@yxieca 
This PR has already been merged, it just needs to be cherry-picked.
Could you please clarify the plan from this point?

yxieca:
> @yxieca This PR has already been merged, it just needs to be cherry-picked. Could you please clarify the plan from this point?

@tshalvi @prgeor please prioritize the db migrator change. we can only cherry-pick when both are ready to avoid warm reboot failures. This PR should not have been merged without matching db migrator change.

tshalvi:
PR to cherry pick to 202311:
https://github.com/sonic-net/sonic-buildimage/pull/18338

tshalvi:
I noticed that the 'Cherry Pick Conflict_202305' label was added. Why is this PR being considered for cherry-picking to 202305? This PR is intended to support the host CMIS management feature, which is only relevant for 202311.

PR #18225, opened 2024-02-29
Fix frrcfgd high CPU utilzation issue caused by small pubsub timeout.

liuh-80:
Fix frrcfgd high CPU utilization issue caused by small pubsub timeout.

#### Why I did it
Fix frrcfgd high CPU utilization issue caused by small pubsub timeout.

##### Work item tracking
- Microsoft ADO: 27002941

#### How I did it
Increase pubsub timeout to 10seconds.

#### How to verify it
Pass all UT.
Manually check CPU utilization back to normal.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211

Fix frrcfgd high CPU utilization issue caused by small pubsub timeout.

### DIFF
--- a/src/sonic-frr-mgmt-framework/frrcfgd/frrcfgd.py
+++ b/src/sonic-frr-mgmt-framework/frrcfgd/frrcfgd.py
@@ -1486,7 +1486,7 @@ class ExtConfigDBConnector(ConfigDBConnector):
         """Start listen Redis keyspace events and will trigger corresponding handlers when content of a table changes.
         """
         self.pubsub = self.get_redis_client(self.db_name).pubsub()
-        self.sub_thread = threading.Thread(target=self.listen_thread, args=(0.01,))
+        self.sub_thread = threading.Thread(target=self.listen_thread, args=(10,))
         self.sub_thread.start()
 
     def stop_listen(self):


src/sonic-frr-mgmt-framework/frrcfgd/frrcfgd.py
```diff
@@ -1486,7 +1486,7 @@ def listen(self):
         """Start listen Redis keyspace events and will trigger corresponding handlers when content of a table changes.
         """
         self.pubsub = self.get_redis_client(self.db_name).pubsub()
-        self.sub_thread = threading.Thread(target=self.listen_thread, args=(0.01,))
+        self.sub_thread = threading.Thread(target=self.listen_thread, args=(10,))
```
liuh-80, line 1489: This issue found by @bluecmd, for more info please check discussion of this PR: https://github.com/sonic-net/sonic-buildimage/pull/13836

bluecmd: Thanks for the fix!

I believe the timeout value should be a float according to the documentation I found, but maybe it doesn't matter. It's a `double` in swsscommon but maybe Python converts it.

juergen-weiss:
the root cause of the error that in src/sonic-swss-common/common/pubsub.cpp the function PubSub::get_message_internal expects the timeout as a double float in seconds and calls  m_select.select , which expects the timeout as an int in ms. That is the conversion should be int(timeout*1000.0) and not int(timeout). So currently  the with args=(0.01), that is a timeout of 10 ms, m_select.select is called with a timeout of 0 and returns immediately

liuh-80:
azpw run Azure.sonic-buildimage

src/sonic-frr-mgmt-framework/frrcfgd/frrcfgd.py
liuh-80: @bluecmd , Can you approve this PR if the fix is ok for you?

bluecmd (approved):

src/sonic-frr-mgmt-framework/frrcfgd/frrcfgd.py
bluecmd: Done!

qiluo-msft: We should fix the bug inside swss-common, so every application could benefit automatically.

liuh-80: I checked SONiC code, here is the only place using get_message() with timeout, so change code inside sonic-swss-common will not break existed code, I will close this PR and create new fix in sonic-swss-common

liuh-80: Will close this PR because PR in sonic-swss-common created: https://github.com/sonic-net/sonic-swss-common/pull/864

liuh-80:
Close because will fix in sonic-swss-common

liuh-80:
Fix PR in sonic-swss-common merged: https://github.com/sonic-net/sonic-swss-common/pull/864

PR #15856, opened 2023-07-15
Don't build ethtool from source

k-v1:
#### Why I did it

There is no reason to build deb package ethtool from source code.
We can install the same version from Debian bullseye mirror.

##### Work item tracking
- Microsoft ADO **(number only)**:

#### How I did it

1. Remove ethtool Makefiles from sonic-buildimage.
2. Install ethtool via apt-get in pmon container.

#### Which release branch to backport (provide reason below if selected)

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

#### Tested branch (Please provide the tested image version)

- [ ] 
- [ ]

PR #18241, opened 2024-03-03
[master][ethtool][chassis] Install ethtool  to localhost

mlok-nokia:
#### Why I did it
Platform Nokia-IXR7250E chassis requires the ethtool to be available in host to disable the Midplane interface autoneg for Linecard-Supervisor midplane communication.

##### Work item tracking

#### How I did it
Modify the build_debian.sh to Install the ethtool in host. 
 
#### How to verify it
Running the new images, verify the ethool is present in /sbin directory
```
dmin@ixre-cpm-chassis15:~$ ls /sbin/ethtool 
/sbin/ethtool
``` 

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]


mlok-nokia:
@rlhui @judyjoseph  This PR installs ethtool in host for midplane commuincation setup.  Please review.

lguohan:
i remember we removed ethtool on the host and use ethtool in the pmon. can you check the commit history to check why we did that?

judyjoseph:
@mlok-nokia the ethtool is not present even in 202205 branch of build_debian.sh. Were we installing it via sonic-platform https://github.com/nokia/sonic-platform/blob/e82caa40d4eec59e574bd4acffb0e92fd7187da4/debian/rules#L43 earlier ?

mlok-nokia:
> @mlok-nokia the ethtool is not present even in 202205 branch of build_debian.sh. Were we installing it via sonic-platform https://github.com/nokia/sonic-platform/blob/e82caa40d4eec59e574bd4acffb0e92fd7187da4/debian/rules#L43 earlier ?

In 202205 branch,  ethtool is built as ethtool_5.9-1_amd64.deb package by the common image build with source code in subdirectory src/ethtool.  What you highlight in Nokia platform is that Nokia platform dpkg and copies it to the Nokia folder /opt/srlinux/bin if the ethtoolxxx_amd64.deb exists.  

In the current master branch, ethtool is no longer built with the source code and it is installed in the PMON docker by PR#  https://github.com/sonic-net/sonic-buildimage/pull/15856. It appears to be he didn't know that ethtool is needed to be installed in localhost for chassis midplane interface configuration.

mlok-nokia:
> i remember we removed ethtool on the host and use ethtool in the pmon. can you check the commit history to check why we did that?

PR https://github.com/sonic-net/sonic-buildimage/pull/15856  moved the ethtool to PMON without much description.

mlok-nokia:
@lguohan  Please help to review and merge this PR. Thanks

lguohan:
@mlok-nokia , #15856 just install ethtool directly instead of building from source. there is some other PR that move ethtool into the PMON docker. can you check which PR is that and we can check with the owner on this

mlok-nokia:
> @mlok-nokia , #15856 just install ethtool directly instead of building from source. there is some other PR that move ethtool into the PMON docker. can you check which PR is that and we can check with the owner on this

It is in the same PR.  In the 

> @mlok-nokia , #15856 just install ethtool directly instead of building from source. there is some other PR that move ethtool into the PMON docker. can you check which PR is that and we can check with the owner on this

It is done in the same PR.    This PR adds the "ethtool" to docker/docker-platform-monitor/Dockerfile.j2 to install the ethtool to PMON docker.  And removes the rules/ethtool.mk to remove the build from source code.

mlok-nokia:
> @mlok-nokia , #15856 just install ethtool directly instead of building from source. there is some other PR that move ethtool into the PMON docker. can you check which PR is that and we can check with the owner on this

@lguohan It is done in the same PR. This PR adds the "ethtool" to docker/docker-platform-monitor/Dockerfile.j2 to install the ethtool to PMON docker. And removes the rules/ethtool.mk to remove the build from source code.

mlok-nokia:
@k-v1 For chassis, ethtool is required to be installed in localhost.  It is required to be localhost to setup midplane interfaces that happens before PMON is running.  Is there any reason why it has been moved to PMON?

prgeor:
@keboliu ethtool inside PMON came via your platform need, it will now be removed since your platform no longer needs inside PMON. it will be available in host only.

lguohan:
@dgsudharsan  let us know if you have any concern on this one?

dgsudharsan:
@keboliu @Junchao-Mellanox  Can you please check if we have any concern?

mlok-nokia:
@prgeor @k-v1 @lguohan @dgsudharsan   Per discussion, I have updated the PR with the change of removing ethtool from PMON.  Please review

keboliu:
I don't have concerns about removing Ethtool from PMON, but please note that there will be an impact on the way using Ethtool from the host side. 
Currently, on the host side, the Ethtool command is a wrapper script to call the Ethtool inside PMON, and it doesn't need root privilege, when we install Ethtool on the host side and use the native Ethtool command, it will require root privilege. This could impact the codes that are calling Ethtool and don't have root privileges.

mlok-nokia:
> I don't have concerns about removing Ethtool from PMON, but please note that there will be an impact on the way using Ethtool from the host side. Currently, on the host side, the Ethtool command is a wrapper script to call the Ethtool inside PMON, and it doesn't need root privilege, when we install Ethtool on the host side and use the native Ethtool command, it will require root privilege. This could impact the codes that are calling Ethtool and don't have root privileges.
> 
> btw, in this PR, I only see Ethtool is removed from PMON but doesn't install it on the host side, as I mentioned, currently from the host side, the Ethtool command is just a wrapper script to the Ethtool inside PMON, after we remove it, we need to install it on the host side, otherwise we will not have Ethtool, I assume this is not the intention.

The modification on the build_debian.sh is for installing ethtool on host.

mlok-nokia:
@liushilongbuaa @lguohan   Could you please follow up with this PR? Anything I need to do? Thanks.

rlhui:
@prgeor would you please review/comment? thanks

arlakshm:
@prgeor, please help review and signoff this PR

judyjoseph (approved):

PR #18270, opened 2024-03-05
Add boundary check before memset, memcpy, strncpy

maipbui:
#### Why I did it
Add boundary check before memset, memcpy, strncpy calls to prevent buffer overflow
##### Work item tracking
- Microsoft ADO **(number only)**: 27008041

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [x] 202106
- [x] 202111
- [x] 202205
- [x] 202211
- [x] 202305

- [ ] 
- [ ]

### DIFF
--- a/src/iccpd/src/iccp_cli.c
+++ b/src/iccpd/src/iccp_cli.c
@@ -117,7 +117,17 @@ int set_peer_link(int mid, const char* ifname)
                        csm->mlag_id, ifname);
     }
 
+    if (MAX_L_PORT_NAME > strlen(csm->peer_itf_name))
+    {
+        ICCPD_LOG_ERR(__FUNCTION__, "MAX=%d is greater than peer_itf_name length=%d", MAX_L_PORT_NAME, strlen(csm->peer_itf_name));
+        return MCLAG_ERROR;
+    }
     memset(csm->peer_itf_name, 0, MAX_L_PORT_NAME);
+    if (len > strlen(csm->peer_itf_name))
+    {
+        ICCPD_LOG_ERR(__FUNCTION__, "len=%d is greater than peer_itf_name length=%d", len, strlen(csm->peer_itf_name));
+        return MCLAG_ERROR;
+    }
     memcpy(csm->peer_itf_name, ifname, len);
 
     /* update peer-link link handler*/
@@ -208,8 +218,18 @@ int set_local_address(int mid, const char* addr)
 
     len = strlen(addr);
     memset(csm->sender_ip, 0, INET_ADDRSTRLEN);
+    if (len > strlen(csm->sender_ip))
+    {
+        ICCPD_LOG_ERR(__FUNCTION__, "len=%d is greater than sender_ip length=%d", len, strlen(csm->sender_ip));
+        return MCLAG_ERROR;
+    }
     memcpy(csm->sender_ip, addr, len);
     memset(csm->iccp_info.sender_name, 0, INET_ADDRSTRLEN);
+    if (len > strlen(csm->iccp_info.sender_name))
+    {
+        ICCPD_LOG_ERR(__FUNCTION__, "len=%d is greater than sender_name length=%d", len, strlen(csm->iccp_info.sender_name));
+        return MCLAG_ERROR;
+    }
     memcpy(csm->iccp_info.sender_name, addr, len);
 
     return 0;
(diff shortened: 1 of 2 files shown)


src/iccpd/src/iccp_cli.c
```diff
@@ -117,7 +117,17 @@ int set_peer_link(int mid, const char* ifname)
                        csm->mlag_id, ifname);
     }
 
+    if (MAX_L_PORT_NAME > strlen(csm->peer_itf_name))
```
gechiang, line 120: is this check correct?  I thought the protection is to ensure that the name is not greater than what is the allowed max value?  but the way it is checking it seems to be the other way around?  can you please check again this logic?

src/iccpd/src/iccp_cli.c
```diff
@@ -117,7 +117,17 @@ int set_peer_link(int mid, const char* ifname)
...
+    {
+        ICCPD_LOG_ERR(__FUNCTION__, "MAX=%d is greater than peer_itf_name length=%d", MAX_L_PORT_NAME, strlen(csm->peer_itf_name));
+        return MCLAG_ERROR;
+    }
     memset(csm->peer_itf_name, 0, MAX_L_PORT_NAME);
+    if (len > strlen(csm->peer_itf_name))
```
gechiang, line 126: same concern here...

src/iccpd/src/iccp_cli.c
```diff
@@ -208,8 +218,18 @@ int set_local_address(int mid, const char* addr)
 
     len = strlen(addr);
     memset(csm->sender_ip, 0, INET_ADDRSTRLEN);
+    if (len > strlen(csm->sender_ip))
```
gechiang, line 221: this needs to check not exceed INET_ADDRSTRLEN instead?

src/iccpd/src/iccp_cli.c
```diff
@@ -208,8 +218,18 @@ int set_local_address(int mid, const char* addr)
...
+    {
+        ICCPD_LOG_ERR(__FUNCTION__, "len=%d is greater than sender_ip length=%d", len, strlen(csm->sender_ip));
+        return MCLAG_ERROR;
+    }
     memcpy(csm->sender_ip, addr, len);
     memset(csm->iccp_info.sender_name, 0, INET_ADDRSTRLEN);
+    if (len > strlen(csm->iccp_info.sender_name))
```
gechiang, line 228: this one needs to check against INET_ADDRSTRLEN?

src/iccpd/src/iccp_cli.c
```diff
@@ -268,6 +288,11 @@ int set_peer_address(int mid, const char* addr)
     }
 
     memset(csm->peer_ip, 0, INET_ADDRSTRLEN);
+    if (len > strlen(csm->peer_ip))
```
gechiang, line 291: so is this check... INET_ADDRSTRLEN

src/iccpd/src/iccp_cli.c
```diff
@@ -117,7 +117,12 @@ int set_peer_link(int mid, const char* ifname)
                        csm->mlag_id, ifname);
     }
 
-    memset(csm->peer_itf_name, 0, MAX_L_PORT_NAME);
+    memset(csm->peer_itf_name, 0, IFNAMSIZ);
+    if (len > MAX_L_PORT_NAME)
```
gechiang, line 121: this also needs to change to check for IFNAMSIZ

src/iccpd/src/iccp_cli.c
```diff
@@ -208,8 +213,18 @@ int set_local_address(int mid, const char* addr)
 
     len = strlen(addr);
     memset(csm->sender_ip, 0, INET_ADDRSTRLEN);
+    if (len > INET_ADDRSTRLEN)
```
gechiang, line 216: for this one, did you confirm that the csm->sender_ip is created with the same "INET_ADDRSTRLEN" size, if not, you will need to change accordingly... if it is then it should be fine.

src/iccpd/src/iccp_cli.c
maipbui: fixed

src/iccpd/src/iccp_cli.c
maipbui: yes, I checked

gechiang (approved):

PR #18272, opened 2024-03-06
[FRR]Adding fix for memory leak seen with BGP community

dgsudharsan:
#### Why I did it
Porting fix for https://github.com/FRRouting/frr/issues/15459

Adding patch for PR https://github.com/FRRouting/frr/pull/15466
##### Work item tracking

#### How I did it
Ported the fix https://github.com/FRRouting/frr/pull/15466 to 8.5.1

#### How to verify it
Running the test_stress_route  and ensure no memory leak

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]


src/sonic-frr/patch/0037-bgp-community-memory-leak-fix.patch
```diff
@@ -0,0 +1,470 @@
...
+--- a/bgpd/bgp_updgrp_adv.c
++++ b/bgpd/bgp_updgrp_adv.c
+@@ -715,11 +715,14 @@ void subgroup_announce_table(struct update_subgroup *subgrp,
+ 						    &attr, NULL)) {
+ 				/* Check if route can be advertised */
+ 				if (advertise) {
+-					if (!bgp_check_withdrawal(bgp, dest))
++					if (!bgp_check_withdrawal(bgp, dest)) {
++						struct attr *adv_attr =
++							bgp_attr_intern(&attr);
```
StormLiangMS, line 179: hi @dgsudharsan by comparing the diff, we think there is one issue in the patch. after line 186 else, we think in that part, we also need a bgp_attr_flush(attr). because the leaked community allocated from subgroup_announce_check function, after this function, need to consider the flush. but in this patch at line 715, it only flush after !bgp_check_withdrawal, it may not enough. Could you help to check below snippet from Liping's patch. 

And one more thing, this attr is allocated without init, pls add struct attr attr = {0} at the beginning also. 

image

dgsudharsan: Hi @donaldsharp Can you please check this?

src/sonic-frr/patch/0037-bgp-community-memory-leak-fix.patch
```diff
@@ -0,0 +1,470 @@
...
++	return true;
+ }
+ 
+ /* The only time 'withdraw' will be false is if we are sending
+@@ -719,9 +721,10 @@ void subgroup_announce_table(struct update_subgroup *subgrp,
+ 						struct attr *adv_attr =
+ 							bgp_attr_intern(&attr);
+ 
+-						bgp_adj_out_set_subgroup(
++						if (!bgp_adj_out_set_subgroup(
```
StormLiangMS, line 372: @dgsudharsan  as we discussed, 
1. No need to check the return value of this function, either true  or false, we need to free. 
2. In the line 376, to replace bgp_attr_flush with bgp_attr_unintern, which is coupled with bgp_attr_intern.
3. In the else branch, to add the  bgp_attr_flush.

lipxu: Seems the latest patch still checking the function and no flush in the else patch, Could you please help to double confirm, thanks a lot.

dgsudharsan:
@StormLiangMS @lipxu Can you please review and signoff?

lipxu:
> @StormLiangMS @lipxu Can you please review and signoff?

Sorry for late response,  will verify it in these two days, and let you know if any update, thanks.

src/sonic-frr/patch/0037-bgp-community-memory-leak-fix.patch
```diff
@@ -1,200 +1,7 @@
...
- 		return true;
- 
--	if (peer->sort == BGP_PEER_EBGP
--	    && (ROUTE_MAP_OUT_NAME(filter) || PREFIX_LIST_OUT_NAME(filter)
--		|| FILTER_LIST_OUT_NAME(filter)
--		|| DISTRIBUTE_OUT_NAME(filter)))
-+	if (peer->sort == BGP_PEER_EBGP &&
-+	    (ROUTE_MAP_OUT_NAME(filter) || PREFIX_LIST_OUT_NAME(filter) ||
-+	     FILTER_LIST_OUT_NAME(filter) || DISTRIBUTE_OUT_NAME(filter) ||
-+	     UNSUPPRESS_MAP_NAME(filter)))
```
lipxu, line 29: Hi, @dgsudharsan , seems there was UNSUPPRESS_MAP_NAME(filter) in the first commit, but removed in the final version, is it expected? thanks.

vjardin:
For information, why does it require a specific patch instead of Sonic that could consume an official maintenance branch of FRR ?

src/sonic-frr/patch/0037-bgp-community-memory-leak-fix.patch
dgsudharsan: Hi @donaldsharp  can you please help answer this?

dgsudharsan:
> For information, why does it require a specific patch instead of Sonic that could consume an official maintenance branch of FRR ?

SONiC's methodology is to consume official release tags from FRR and then integrate any required fixes on top of it. In this case the last updated version was 8.5.1. This patch was made created only recently and can only be added as patch on top of 8.5.1. Later when SONiC picks in a stable official version of FRR which contains the patch, the patch would be removed.

src/sonic-frr/patch/0037-bgp-community-memory-leak-fix.patch
donaldsharp: it is missing, but the fix it came from `6814401c47b948a50` is a fix that is not related to the actual memory leak.  It is related to allowing a unsuppress-map as a valid outgoing policy.  Where it was not allowed before.

dgsudharsan:
@lipxu  @StormLiangMS Can you please help prioritize and close the review?

lipxu (approved):

dgsudharsan:
Thank you @lipxu . @StormLiangMS  Can you please help to merge this PR? I will create one for 202311 and will have this patch for master in 8.5.4 upgrade

jimmyzhai (approved):

StormLiangMS (approved):
LGTM

dgsudharsan:
Hi @StormLiangMS  Here is PR for 202311 https://github.com/sonic-net/sonic-buildimage/pull/18606

PR #18295, opened 2024-03-07
Remove PMON and XCVRD delay on system boot for modules controlled by SW

vadymhlushko-mlnx:
#### Why I did it
To support the `fastboot` feature alongside `module controlled by sw ` feature.
Because if the  feature is enabled we need the `XCVRD` to start as soon as possible to identify if the port is `SW` or `FW` controlled and bring the port to the UP state.

##### Work item tracking

#### How I did it
Remove the delays for the `PMON` docker and `XCVRD` daemon on the SONiC image start.

#### How to verify it
Run the [sonic-mgmt/tests/platform_tests/test_advanced_reboot.py - test_fast_reboot](https://github.com/sonic-net/sonic-mgmt/blob/89c4a1a0a560772c2ac02c5bd8c8c23d6883fd93/tests/platform_tests/test_advanced_reboot.py#L62)

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305
- [x] 202311

- [ ] 
- [ ]

### DIFF
--- a/files/scripts/syncd.sh
+++ b/files/scripts/syncd.sh
@@ -96,18 +96,9 @@ function waitplatform() {
 
     BOOT_TYPE=`getBootType`
     if [[ x"$sonic_asic_platform" == x"mellanox" ]]; then
-        if [[ x"$BOOT_TYPE" = @(x"fast"|x"warm"|x"fastfast") ]]; then
-            PMON_TIMER_STATUS=$(systemctl is-active pmon.timer)
-            if [[ x"$PMON_TIMER_STATUS" = x"inactive" ]]; then
-                systemctl start pmon.timer
-            else
-                debug "PMON service is delayed by a timer for better fast/warm boot performance"
-            fi
-        else
-            debug "Starting pmon service..."
-            /bin/systemctl start pmon
-            debug "Started pmon service"
-        fi
+        debug "Starting pmon service..."
+        /bin/systemctl start pmon
+        debug "Started pmon service"
     fi
 }
 
--- a/files/scripts/syncd.sh
+++ b/files/scripts/syncd.sh
@@ -92,16 +92,6 @@ function startplatform() {
     fi
 }
 
-function waitplatform() {
-
-    BOOT_TYPE=`getBootType`
-    if [[ x"$sonic_asic_platform" == x"mellanox" ]]; then
-        debug "Starting pmon service..."
-        /bin/systemctl start pmon
-        debug "Started pmon service"
-    fi
-}
-
 function stopplatform1() {
 
     if [[ x$sonic_asic_platform == x"mellanox" ]] && [[ x$TYPE == x"cold" ]]; then
--- a/device/mellanox/x86_64-mlnx_msn2700-r0/pmon_daemon_control.json
+++ b/device/mellanox/x86_64-mlnx_msn2700-r0/pmon_daemon_control.json
@@ -1,7 +1,7 @@
 {
     "skip_ledd": true,
     "skip_fancontrol": true,
-    "delay_xcvrd": true,
+    "delay_xcvrd": false,
     "skip_xcvrd_cmis_mgr": true
 }
 
--- a/files/build_templates/init_cfg.json.j2
+++ b/files/build_templates/init_cfg.json.j2
@@ -36,7 +36,7 @@
 {%- set features = [("bgp", "{% if not DEVICE_RUNTIME_METADATA['ETHERNET_PORTS_PRESENT'] or ('CHASSIS_METADATA' in DEVICE_RUNTIME_METADATA and DEVICE_RUNTIME_METADATA['CHASSIS_METADATA']['module_type'] in ['supervisor']) %}disabled{% else %}enabled{% endif %}", false, "enabled"),
                    ("database", "always_enabled", false, "always_enabled"),
                    ("lldp", "enabled", true, "enabled"),
-                   ("pmon", "enabled", true, "enabled"),
+                   ("pmon", "enabled", false, "enabled"),
                    ("snmp", "enabled", true, "enabled"),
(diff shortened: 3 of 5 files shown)


liat-grozovik:
@vadymhlushko-mlnx is this cleanly cherry picked to 202311 or we need a dedicate PR? if such is required we should have the label set properly and relate to the new PR against 202311.

files/build_templates/pmon.service.j2
```diff
@@ -11,6 +11,9 @@ Before=ntp-config.service
 StartLimitIntervalSec=1200
 StartLimitBurst=3
 
+[Install]
```
prgeor, line 14: @vadymhlushko-mlnx why is this change?

vadymhlushko-mlnx: Without the `[Install] WantedBy=sonic.target` the PMON service won't start at SONiC boot time

prgeor: @vadymhlushko-mlnx you mean today, without this change, PMON start is delayed ? PMON does not start?

vadymhlushko-mlnx: Yes, without this change - `PMON` start is delayed

prgeor: @vadymhlushko-mlnx Here is systemd-analyze dump for NON-Mellanox. As long as all systemd dependencies are statisfied, systemd will launch PMON service


image


There is only 1 second delay and that too if you check carefully, PMON started 1 second before syncd

I don't see any reason for this change needed for systemd or why systemd will delay PMON if all dependencies are satisfied.

vadymhlushko-mlnx: Okay, this `[Install]` section was here before we introduced a PR that `delays the PMON and XCVRD`.
So, okay, I'm going to remove it, if it is no longer needed

vadymhlushko-mlnx: @prgeor fixed, please approve and merge

vadymhlushko-mlnx:
@prgeor could you please help to merge

files/scripts/syncd.sh
```diff
@@ -92,18 +92,9 @@ function waitplatform() {
...
-                systemctl start pmon.timer
-            else
-                debug "PMON service is delayed by a timer for better fast/warm boot performance"
-            fi
-        else
-            debug "Starting pmon service..."
-            /bin/systemctl start pmon
-            debug "Started pmon service"
-        fi
+        debug "Starting pmon service..."
```
prgeor, line 95: @vadymhlushko-mlnx Why will Systemd not start the PMON early? Why not remove "After=Syncd" for your platform? https://github.com/sonic-net/sonic-buildimage/blob/6d0cd5519560f48cb2a7421588d100c617996eb7/files/build_templates/pmon.service.j2#L6

vadymhlushko-mlnx:
@prgeor kind reminder

files/scripts/syncd.sh
keboliu: @vadymhlushko-mlnx  would you please check the comments?

prgeor: @vadymhlushko-mlnx There is already a dependency between Xcvrd and Swss. Xcvrd waits for [PortInitDone](https://github.com/sonic-net/sonic-platform-daemons/blob/master/sonic-xcvrd/xcvrd/xcvrd.py#L2423) and untill portsyncd sets its after all host interfaces are created: https://github.com/sonic-net/sonic-swss/blob/19410232523283d99eaa211979cc1e58b6cb262f/portsyncd/portsyncd.cpp#L134

We are holding complete PMON init just for Xcvrd which doesn't seems right. Can we use PortInitDone as a synchronization?

vadymhlushko-mlnx: @prgeor kind reminder, I'm waiting for your comment or approve

prgeor: @vadymhlushko-mlnx Please explain more about:-
"@prgeor XCVRD should be started after we create_switch and initialize all the ports, that is why we have a dependency for the syncd."

vadymhlushko-mlnx: @prgeor if we remove the `After=syncd.service` from the `pmon.service` file - the `PMON` is going to start parallel to `SYNCD`. It could lead to such a scenario - if in the `PMON,` someone is going to do some ASIC-related operation while in parallel the `SYNCD` is doing `create_switch` and resetting the ASIC - such flow will introduce a degradation.

prgeor: @vadymhlushko-mlnx synchronization via systemd service file is fine. Why do we need to start pmon from synch.sh?

prgeor: @vadymhlushko-mlnx let systemd do the synchronization between syncd and pmon service

prgeor (changes requested):
@vadymhlushko-mlnx please use systemd for synchronization between PMON and syncd

files/scripts/syncd.sh
stepanblyschak: @prgeor Some services are started by shell scripts, some via systemd. Both approaches work in SONiC. Starting from shell script gives more flexibility

vadymhlushko-mlnx: @prgeor I've removed unused code from the `syncd.sh` as you requested

prgeor: @vadymhlushko-mlnx OK. is the PR still in draft or ready for review?

vadymhlushko-mlnx:
New PR will be opened by @stepanblyschak

stepanblyschak:
@prgeor Please have a look at https://github.com/sonic-net/sonic-buildimage/pull/19190

Issue #18137, opened 2024-02-20
[Yang] Incorrect restriction for ICMP/ICMPv6 type and code.


lizhijianrd:
#### Description

For ICMP, [sonic-acl.yang.j2](https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-yang-models/yang-templates/sonic-acl.yang.j2) require `type` between 1 and 44, and `code` between 1 and 16. However, this is incorrect. For example, according to [RFC 792](https://datatracker.ietf.org/doc/html/rfc792), type=0 and code=0 means echo reply message. These are valid ICMP type/code but cannot pass Yang validation.

For ICMPv6, [sonic-acl.yang.j2](https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-yang-models/yang-templates/sonic-acl.yang.j2) also require `type` between 1 and 44, and `code` between 1 and 16. This is incorrect, too. For example, according to [RFC 4443](https://datatracker.ietf.org/doc/html/rfc4443), type=129 and code=0 means echo reply message.

#### Steps to reproduce the issue:

Currently, this issue impacts GCU dataplane ACL update. Below is the repro steps:
1. Setup custom ACL table type on DUT.
    a. Write below content to file `acl_table_types.json` on DUT:
    ```json
    {
        "ACL_TABLE_TYPE": {
            "BMCDATAV6": {
                "MATCHES": "SRC_IPV6,DST_IPV6,ETHER_TYPE,IP_TYPE,IP_PROTOCOL,IN_PORTS,L4_SRC_PORT,L4_DST_PORT,L4_SRC_PORT_RANGE,L4_DST_PORT_RANGE",
                "ACTIONS": "PACKET_ACTION,COUNTER",
                "BIND_POINTS": "PORT"
            }
        }
    }
    ```
    b. Add custom ACL table type to config_db: `sonic-cfggen -j acl_table_types.json -w`
2. Create a data-plane ACL table: `sudo config acl add table SAMPLE_DATAPLANE_ACL_TABLE BMCDATAV6 -d SAMPLE -s ingress -p Ethernet1,Ethernet2`
3. Add data-plane ACL rule to config_db via `acl-loader`:
    a. Write below content to `acl_rules.json` file on DUT:
    ```json
    {
        "acl": {
            "acl-sets": {
                "acl-set": {
                    "SAMPLE_DATAPLANE_ACL_TABLE": {
                        "acl-entries": {
[...]

PR #18311, opened 2024-03-08
Fix yang model for ICMP and ICMPV6

bingwang-ms:
#### Why I did it
Fix[ #18137](https://github.com/sonic-net/sonic-buildimage/issues/18137)

This PR is to update the definition of Yang model for ICMP and ICMPv6.

|             | Before change | After change |
|-------------|---------------|--------------|
| ICMP type   |    [1, 44]    |    [0, 44]   |
| ICMP code   |    [1, 16]    |    [0, 16]   |
| ICMPV6 type |    [1, 44]    |   [0, 255]   |
| ICMPV6 code |    [1, 44]    |   [0, 255]   |

References
1. https://datatracker.ietf.org/doc/html/rfc792
2. https://datatracker.ietf.org/doc/html/rfc4443
##### Work item tracking
- Microsoft ADO **27118551**:

#### How to verify it
Verified by UT.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [x] 202305
- [x] 202311 

- [ ] 
- [ ] 

 Update the definition of Yang model for ICMP and ICMPv6.

### DIFF
--- a/src/sonic-yang-models/yang-templates/sonic-acl.yang.j2
+++ b/src/sonic-yang-models/yang-templates/sonic-acl.yang.j2
@@ -202,13 +202,13 @@ module sonic-acl {
 						when "not(IP_TYPE) or boolean(IP_TYPE[.='ANY' or .='IP' or .='IPV4' or .='IPv4ANY' or .='ARP'])";
 						leaf ICMP_TYPE {
 							type uint8 {
-								range 0..44;
+								range 0..255;
 							}
 						}
 
 						leaf ICMP_CODE {
 							type uint8 {
-								range 0..16;
+								range 0..255;
 							}
 						}
 					}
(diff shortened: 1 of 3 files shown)


Blueve (approved):

lizhijianrd (approved):

src/sonic-yang-models/yang-templates/sonic-acl.yang.j2
```diff
@@ -202,13 +202,13 @@ module sonic-acl {
 						when "not(IP_TYPE) or boolean(IP_TYPE[.='ANY' or .='IP' or .='IPV4' or .='IPv4ANY' or .='ARP'])";
 						leaf ICMP_TYPE {
 							type uint8 {
-								range 1..44;
+								range 0..44;
```
ganglyu, line 205: Maximum value is 44, why do we have this limit?

bingwang-ms: Yes, maximun value is 44. We have this limit.

ganglyu: https://www.iana.org/assignments/icmp-parameters/icmp-parameters.xhtml
https://www.iana.org/assignments/icmpv6-parameters/icmpv6-parameters.xml
Please check above links, do we need to modify the type range and code range?

bingwang-ms: Updated.

ganglyu (approved):

wen587 (approved):

PR #18340, opened 2024-03-12
[202305][Arista] Update platform library submodules

Staphylo:
#### Why I did it

 - fix reboot cause processing on linecards due to mgmt network restart during boot
 - fix reboot cause timestamp for some products
 - enhance out of spec transceiver support on wolverine linecard

##### Work item tracking

#### How I did it

Cherry-picked relevant master fixes to 202305 release branch 

Update Arista platform library submodules


lguohan:
is this only affecting chassis line card?

Staphylo:
This change is fixing issues affecting both linecards and fixed systems.
Respective to the changes listed in the description:
1) affecting all linecards
2) affecting a subset of fixed systems
3) affecting only the wolverine linecard

There are also 2 more fixes in the pipeline that we are validating internally.
 - One is affecting 7050CX3-32S following an escalation
 - One is affecting fabric cards initialization that we caught running sonic-mgmt

PR #18350, opened 2024-03-13
Fix to use IPv6 linklocal address as snmp agent address

SuvarnaMeenakshi:
#### Why I did it
1. fix https://github.com/sonic-net/sonic-buildimage/issues/16001
2. fix https://github.com/sonic-net/sonic-buildimage/issues/17807

https://github.com/sonic-net/sonic-buildimage/pull/17045 modified minigraph parser to use management and loopback IPs to support SNMP query over IPv6. With this fix, if mgmt or loopback IP contains link local IP, that will not work as link local IP has to be appended with scope id associating the IP address to a specific interface. This PR change is to ensure that snmp works with link local IPv6 address.

##### Work item tracking

#### How I did it
1. Modify minigraph parser to append the Ip address with % scope id if snmp agent address being used is link local IP address.
2. Modify snmpd.conf.j2 to take this change while checking if an IP address is ipv4 or ipv6.

#### How to verify it
Verified by configuring link local ipv6 address.
Last login: Wed Mar 13 01:45:09 2024 from 10.1.84.57
admin@<>:~$ sudo netstat -tulnp | grep 161
...    
udp6       0      0 fe80::f6ee:31ff:fe9:161 :::*                                70355/snmpd         
..

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/src/sonic-config-engine/minigraph.py
+++ b/src/sonic-config-engine/minigraph.py
@@ -1763,7 +1763,7 @@ def parse_xml(filename, platform=None, port_config_file=None, asic_name=None, hw
         results['SNMP_AGENT_ADDRESS_CONFIG'] = {}
         port = '161'
         for intf in list(mgmt_intf.keys()) + list(lo_intfs.keys()):
-            ip_addr = ipaddress.ip_address(intf[1].split('/')[0])
+            ip_addr = ipaddress.ip_address(UNICODE_TYPE(intf[1].split('/')[0]))
             if ip_addr.version == 6 and ip_addr.is_link_local:
                 agent_addr = str(ip_addr) + '%' + intf[0]
             else:
--- a/src/sonic-config-engine/minigraph.py
+++ b/src/sonic-config-engine/minigraph.py
@@ -1762,12 +1762,13 @@ def parse_xml(filename, platform=None, port_config_file=None, asic_name=None, hw
     if not is_multi_asic() and asic_name is None:
         results['SNMP_AGENT_ADDRESS_CONFIG'] = {}
         port = '161'
-        for mgmt_intf in mgmt_intf.keys():
-            snmp_key = mgmt_intf[1].split('/')[0] + '|' + port + '|'
-            results['SNMP_AGENT_ADDRESS_CONFIG'][snmp_key] = {}
-        # Add Loopback IP as agent address for single asic
-        for loip in lo_intfs.keys():
-            snmp_key = loip[1].split('/')[0] + '|' + port + '|'
+        for intf in list(mgmt_intf.keys()) + list(lo_intfs.keys()):
+            ip_addr = ipaddress.ip_address(intf[1].split('/')[0])
+            if ip_addr.version == 6 and ip_addr.is_link_local:
+                agent_addr = str(ip_addr) + '%' + intf[0]
+            else:
+                agent_addr = str(ip_addr)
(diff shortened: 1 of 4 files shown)


keboliu:
@SuvarnaMeenakshi seems the current unit test was broken, would you please check?

SuvarnaMeenakshi:
> 1. parser to append the Ip address with % scope id if snmp agent address being used is l

done, please help review the changes. thank you.

src/sonic-config-engine/minigraph.py
```diff
@@ -1762,12 +1762,13 @@ def parse_xml(filename, platform=None, port_config_file=None, asic_name=None, hw
...
-        for mgmt_intf in mgmt_intf.keys():
-            snmp_key = mgmt_intf[1].split('/')[0] + '|' + port + '|'
-            results['SNMP_AGENT_ADDRESS_CONFIG'][snmp_key] = {}
-        # Add Loopback IP as agent address for single asic
-        for loip in lo_intfs.keys():
-            snmp_key = loip[1].split('/')[0] + '|' + port + '|'
+        for intf in list(mgmt_intf.keys()) + list(lo_intfs.keys()):
+            ip_addr = ipaddress.ip_address(UNICODE_TYPE(intf[1].split('/')[0]))
+            if ip_addr.version == 6 and ip_addr.is_link_local:
+                agent_addr = str(ip_addr) + '%' + intf[0]
```
qiluo-msft, line 1768: >% [](http://example.com/codeflow?start=45&length=1)

1. Is it a config schema disruptive change?
2. Do you need to change yang model?
3. We try hard to prevent minigraph parser behavior change since we will deploy config_db.json directly in production. So backward-compatible config schema change is preferred.

SuvarnaMeenakshi: 1. This is not a config_db schema change
2. Yang change is already in place
3. this change is mainly to append %(zone) if the IP address is a link local IP Address, inside the SNMP_AGENT_ADDRESS_CONFIG table which is used here: https://github.com/sonic-net/sonic-buildimage/blob/master/dockers/docker-snmp/snmpd.conf.j2#L29
We cannot make this change in the template directly, because "zone" refers to the "interface name" to which the respective IP address belongs to, so we need to know what is the interface name while adding the IP address.

bingwang-ms:
@qiluo-msft Can you please review?

yxieca:
@qiluo-msft , @SuvarnaMeenakshi  is it safe to merge/cherry-pick this change?

bingwang-ms:
@SuvarnaMeenakshi Please raise a new PR for 202405 branch to address the conflict. Thanks

dprital:
@SuvarnaMeenakshi , Do we have already PR Vs. 202405 ?
Thanks.

SuvarnaMeenakshi:
> @SuvarnaMeenakshi , Do we have already PR Vs. 202405 ? Thanks.

@dprital yes, the changes are already merged to 202405.
@bingwang-ms fixed the tags to reflect that the changes are merged in 202405.
```
 git log --oneline origin/202405 | grep "linklocal"
a0ff20edf [snmpd]: Fix to use IPv6 linklocal address as snmp agent address (#18350)
```

PR #18393, opened 2024-03-19
[supervisor][sfm]Fix the issue of swss.sh shows backtrace when shutdown a SFM

mlok-nokia:
#### Why I did it
On a Supervisor card of a VOQ chassis, when remove or shutdown a Fabric card, swss.sh shows Stacktrace for all related empty SFM slots in the syslog file.  This PR fixes https://github.com/sonic-net/sonic-buildimage/issues/18384

##### Work item tracking

#### How I did it
In the asic_status.py, all empty SFM slots related swss.sh is in the waiting state to wait for the presence event of SFM -- SET operation.  The subscriber event handler also includes the "DEL" operation when a SFM is shutdown/removal.  When a SFM is shutdown, all empty slot's swss.sh also get the "DEL" event although it is not for them.  In the "DEL" operation, the current implementation doesn't check if this "DEL" operation for them, and then they exit the wait state and proceed to docker-wait-any with wrong operation in the wrong slot.  docker-wait0any raise the backtarce.           

#### How to verify it
1.   In a chassis which has some empty SMF slot, remove or shutdown a SFM.  There should not be related stacktrace shown in syslog 

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [x] 202205
- [x] 202211
- [x] 202305

- [ ] 
- [ ]

### DIFF
--- a/files/scripts/asic_status.py
+++ b/files/scripts/asic_status.py
@@ -69,8 +69,9 @@ def main():
                 logger.log_info('Detected asic{} is online'.format(global_asic_id))
                 sys.exit(0)
         elif asic_op == 'DEL':
-            logger.log_info('Detected asic{} is offline'.format(global_asic_id))
-            sys.exit(1)
+            if (global_asic_id == args_asic_id):
+                logger.log_info('Detected asic{} is offline'.format(global_asic_id))
+                sys.exit(1)
         else:
             continue
 


mlok-nokia:
@arlakshm  This PR fixed the backtarce which we saw in the syslog file when we test the shutdown a SFM.  Please review it

arlakshm (approved):
lgtm.

PR #18441, opened 2024-03-22
[PMON] Support to fetch pmon_daemon_control.json file from both platform and HWSKU folders

keboliu:
#### Why I did it

The pmon_daemon_control.json file currently has a platform-specific scope, resulting in all SKUs of the same platform sharing the same pmon daemon configuration. Consequently, this restricts the ability to have distinct pmon daemon configurations for different SKUs. The proposed change aims to enhance flexibility by allowing PMON to load this file from both the platform folder and the HWSKU folder. In case a pmon_daemon_control.json file exists in the HWSKU folder, it will take precedence over the one in the platform folder.

##### Work item tracking

#### How I did it

The pmon docker init script will be updated to prioritize the HWSKU folder when searching for the pmon_daemon_control.json file. If the file is present in the HWSKU folder, it will be utilized. However, if the file does not exist in the HWSKU folder, the script will then fallback to utilizing the one located in the platform folder.

#### How to verify it

Build image with this change, put the file in 
1. platform folder, 
2. in HWSKU folder
3. in both folders

make sure the pmon daemons started according to the configuration.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305
- [x] 202311

- [ ] 
- [ ]

### DIFF
--- a/dockers/docker-platform-monitor/docker_init.j2
+++ b/dockers/docker-platform-monitor/docker_init.j2
@@ -8,7 +8,6 @@ FANCONTROL_CONF_FILE="/usr/share/sonic/platform/fancontrol"
 
 SUPERVISOR_CONF_TEMPLATE="/usr/share/sonic/templates/docker-pmon.supervisord.conf.j2"
 SUPERVISOR_CONF_FILE="/etc/supervisor/conf.d/supervisord.conf"
-PMON_DAEMON_CONTROL_FILE="/usr/share/sonic/platform/pmon_daemon_control.json"
 MODULAR_CHASSISDB_CONF_FILE="/usr/share/sonic/platform/chassisdb.conf"
 
 HAVE_SENSORS_CONF=0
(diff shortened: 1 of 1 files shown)


dockers/docker-platform-monitor/docker_init.j2
```diff
@@ -17,6 +16,13 @@ IS_MODULAR_CHASSIS=0
 # Default use python2 version
 SONIC_PLATFORM_API_PYTHON_VERSION=2
 
+if [ -e /usr/share/sonic/hwsku/pmon_daemon_control.json ];
+then
+    PMON_DAEMON_CONTROL_FILE="/usr/share/sonic/hwsku/pmon_daemon_control.json"
```
prgeor, line 21: @keboliu is `hwsku` a standard known path? The HWSKU path is something like below
device/mellanox/x86_64-mlnx_msn2700-r0/Mellanox-SN2700/hwsku.json
or
device/mellanox/x86_64-mlnx_msn2700-r0/<HWSKU>/hwsku.json

keboliu: @prgeor HWSKU is a folder has been mounted to multiple dockers including PMON, please check the following code
https://github.com/sonic-net/sonic-buildimage/blob/master/files/build_templates/docker_image_ctl.j2#L633 

it's a folder which is identical to the path like "device/mellanox/x86_64-mlnx_msn2700-r0/Mellanox-SN2700/" on the host side.

prgeor (approved):

liat-grozovik (approved):

keboliu:
@yxieca would you please help to cherry-pick?

PR #18452, opened 2024-03-23
[multi-asic] Use asic_id/asic_name for APIs to get port config data

anamehra:
Signed-off-by: Anand Mehra anamehra@cisco.com

Fixes: https://github.com/sonic-net/sonic-buildimage/issues/18431

#### Why I did it
sonic-cfggen misses handling for namespace at couple of locations. Due to this, when port config is fetched, it returns data from global config and not the namespace.

The function calls for
get_path_to_port_config_file
get_port_config

needs to be fixed to handle namespace for multi-asic.

##### Work item tracking

#### How I did it
1.Pass asic_id as argument when calling get_path_to_port_config_file()
For single asic, asic_id is None and the function returns the correct port_config.ini file path.
For multi-asic, using asic_id provides the correct file path for the asic.

2. Pass asic_name as argument when calling get_port_config()

#### How to verify it
sonic-cfggen -H --print -n -k
This command should return the correct port config for the namespace.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [] 202211
- [x] 202305

- [ ] 
- [ ]


anamehra:
Hi @abdosi , please review.

anamehra:
Hi @abdosi , @rlhui , please help with the review. Thanks

anamehra:
I raised a PR to fix this https://github.com/sonic-net/sonic-buildimage/pull/18452
Please help to review.

anamehra:
Please help with the review. Thanks

mlok-nokia (approved):
Looks good

abdosi:
@SuvarnaMeenakshi : can you help with review of this.

SuvarnaMeenakshi:
Thank you for the fix.
Seems like was missed in this PR https://github.com/sonic-net/sonic-buildimage/pull/7632 and issue will NOT be seen if sonic-cfggen -m <minigraph.xml> is used.
Can we add a unit-test in sonic-config-engine multi-asic unit-test, so that if minigraph.xml is not passed, then the expected PORT from namespace is present in the output?

anamehra:
> Thank you for the fix. Seems like was missed in this PR #7632 and issue will NOT be seen if sonic-cfggen -m <minigraph.xml> is used. Can we add a unit-test in sonic-config-engine multi-asic unit-test, so that if minigraph.xml is not passed, then the expected PORT from namespace is present in the output?

Thanks for review @SuvarnaMeenakshi ! May I use another PR to add UT? Busy with a couple of other activities. This PR needs to help with nightly T2 runs.

SuvarnaMeenakshi:
> > Thank you for the fix. Seems like was missed in this PR #7632 and issue will NOT be seen if sonic-cfggen -m <minigraph.xml> is used. Can we add a unit-test in sonic-config-engine multi-asic unit-test, so that if minigraph.xml is not passed, then the expected PORT from namespace is present in the output?
> 
> Thanks for review @SuvarnaMeenakshi ! May I use another PR to add UT? Busy with a couple of other activities. This PR needs to help with nightly T2 runs.

It might fall through later, it appears to be a simple UT can do the test in this case.
similar to test done here: https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-config-engine/tests/test_multinpu_cfggen.py#L209, without -m option and could check for presence of BP interface.
Without this current fix, the BP interface would not be present in the result as it would pick from global config_db.

anamehra:
Hi @SuvarnaMeenakshi , added the test case, please review. Thanks

SuvarnaMeenakshi (approved):
lgtm

anamehra:
Hi @SuvarnaMeenakshi , @abdosi , please suggest what to be done for ms_conflict. Thanks

rlhui:
@anamehra  please help resolve conflicts

anamehra:
> @anamehra please help resolve conflicts

Hi @rlhui , I am not able to find details of the failure and what needs to be done. Waiting for inputs from @abdosi . Thanks

anamehra:
Hi @abdosi , please merge. Thanks

SuvarnaMeenakshi (approved):

abdosi:
@qiluo-msft : can you help with merge of this.

qiluo-msft (approved):

PR #18485, opened 2024-03-27
[build][reproducible]Fix version file access for dbg container build

baxia-lan:
#### Why I did it

Reproducible build fails when building dbg containers `make INSTALL_DEBUG_TOOLS=y target/docker-orchagent-dbg.gz`

#### How to verify it

```
export SONIC_VERSION_CONTROL_COMPONENTS=all
make configure PLATFORM=broadcom
make INSTALL_DEBUG_TOOLS=y target/docker-orchagent-dbg.gz
```

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

### DIFF
--- a/scripts/versions_manager.py
+++ b/scripts/versions_manager.py
@@ -419,6 +419,8 @@ class VersionModule:
             return os.path.join(source_path, 'files/build/versions', module_name)
         if module_name.startswith('build-sonic-slave-'):
             return os.path.join(source_path, 'files/build/versions/build', module_name)
+        if module_name.endswith('-dbg'):
+            return os.path.join(source_path, 'files/build/versions/dockers', module_name.removesuffix('-dbg'))
         return os.path.join(source_path, 'files/build/versions/dockers', module_name)
 
     def __repr__(self):


baxia-lan:
Oh typo there, should be `export SONIC_VERSION_CONTROL_COMPONENTS=all`

lguohan:
@liushilongbuaa , can you take a look?

liushilongbuaa:
We already get these dbg packages in
[This Line ](https://github.com/sonic-net/sonic-buildimage/blob/a56cf7905cdf114c2647723fec8235cfab8e9f06/files/build/versions/dockers/docker-orchagent/versions-deb-bullseye#L4)
all these packages:  gdb gdbserver vim openssh-client sshpass strace

When we copy version files in [this line](https://github.com/sonic-net/sonic-buildimage/blob/a56cf7905cdf114c2647723fec8235cfab8e9f06/scripts/prepare_docker_buildinfo.sh#L53), the version files in buildinfo is the same as normal docker.
docker-orchagent and docker-orchagent-dbg shares the same root folder. So they use the same version file.

baxia-lan:
Take docker-syncd-brcm-dbg.gz as an example, the dbg packages are saved to the dfiles/build/versions/dockers/docker-syncd-brcm/versions-deb-bullseye file, but these version files are not correct loaded when building. The version files copied to docker container are empty when inspecting it. 

Another evidence in docker-syncd-brcm-dbg.gz.log:
```
[91mWarning: the version of the package gdb is not specified.
Warning: the version of the package gdbserver is not specified.
Warning: the version of the package vim is not specified.
Warning: the version of the package openssh-client is not specified.
Warning: the version of the package sshpass is not specified.
Warning: the version of the package strace is not specified.
```
The warning is printed out at [this](https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-build-hooks/scripts/buildinfo_base.sh#L406) line. You can still build the dbg docker because the check_apt_version only give warnings never errors.

Speaking of that, I think the reproducible framework should  throw errors instead of warnings but continue with "latest" versions, this silent warning makes the build not reproducible... And pin ALL dependency packages during build, not only the incremental ones to avoid conflicts resolving dependencies versions. I would love to provide the possible solutions for debian packages if you agreed.

[docker-syncd-brcm-dbg.gz.log](https://github.com/sonic-net/sonic-buildimage/files/14832324/docker-syncd-brcm-dbg.gz.log)
[docker-syncd-brcm.gz.log](https://github.com/sonic-net/sonic-buildimage/files/14832327/docker-syncd-brcm.gz.log)

liushilongbuaa:
I checked the log. It didn't work for dbg docker image.

baxia-lan:
> It means that all deb packages for reproducible builds are installed from [snapshots](https://github.com/sonic-net/sonic-buildimage/blob/master/files/build/versions/default/versions-mirror) and `check_apt_version` can be ignored at all.

We don't use this public mirror to build as we host our own mirrors to have better control. But I assume the mirror repositories host multiple versions of packages as the version files might be different between different release branches and can be upgraded? The version files should be still needed to tell which versions of packages to install. Anyway, the behavior of the reproducible build should be consistent between non-dbg and dbg containers build. We should merge this PR and continue discussion for separate issues if needed.

baxia-lan:
> > Anyway, the behavior of the reproducible build should be consistent between non-dbg and dbg containers build.
> 
> I agree. It's just FYI. Also keep in mind warnings also exist for build_debian.sh:
> 
> https://sonic-build.azurewebsites.net/api/sonic/artifacts?branchName=master&platform=broadcom&buildId=514080&target=target%2Fsonic-broadcom.bin.log
> 
> ```
> Warning: the version of the package libcairo2-dev is not specified.
> Warning: the version of the package libdbus-1-dev is not specified.
> Warning: the version of the package libgirepository1.0-dev is not specified.
> Warning: the version of the package libsystemd-dev is not specified.
> Warning: the version of the package pkg-config is not specified.
> ```
> 
> And if you use some non-standard package installation like `apt-get install <package>/bookworm-backports` instead of `apt install -t bookworm-backports <package>` then version of this package probably also won't saved. That's why I prefer to use snapshots based on local mirror. Mirror has multiple versions but snapshot has only one for each package. And if we are talk about repositories like `https://download.docker.com/` then you can pin version of the package like [here](https://github.com/sonic-net/sonic-buildimage/blob/7b04c69896794e3234da6381e6c301ea79725389/sonic-slave-bookworm/Dockerfile.j2#L559)

Opened a thread to discuss the strategy. Please move discussions there. I also add my proposal in the issue description.

If all agreed, can this PR be approved and merged now?

baxia-lan:
@lguohan @qiluo-msft @xumia 

Kindly ping for review.

xumia (approved):
LGTM

PR #18499, opened 2024-03-29
Calulcate subport values and fix key error in hwsku.json

bobbymcgonigle:
Automatically populate the correct subport value for an interface and insert into config_db.json

### Problem 1
Breakouts of optic CMIS modules are broken. For example when in breakout mode of 4x100G on a port the /3, /5, /7 are not correctly associating with lanes (3,4), (5,6) and (7,8) of the transceiver respectively. The /3, /5, /7 are all trying to reprogram lanes (1,2).

In the xcvrd logs we see the following lane masks for all 4 interfaces for a port in 4x100G-2 mode:
```
EthernetX/X: Setting host_lanemask=0x3
EthernetX/X: Setting media_lanemask=0x1
```

As a result, we only see the /1 interface linking up correctly, the other 3 interfaces do not have the correct application advertisement + lanemasks applied. This is what we should expect to see:
```
lanes (1,2)
EthernetX/1: Setting host_lanemask=0x3
EthernetX/1: Setting media_lanemask=0x1

lanes (3,4)
EthernetX/3: Setting host_lanemask=0xc
EthernetX/3: Setting media_lanemask=0x2

lanes (5,6)
EthernetX/5: Setting host_lanemask=0x30
EthernetX/5: Setting media_lanemask=0x4

lanes (7,8)
EthernetX/7: Setting host_lanemask=0xc0
EthernetX/7: Setting media_lanemask=0x8
```

The problem is that  “subport” is expected and used in xcvrd, xcvrd unit tests and explained in the [Port Breakout HLD](https://github.com/sonic-net/SONiC/blob/master/doc/dynamic-port-breakout/sonic-dynamic-port-breakout-HLD.md) ; it was not implemented and does not end up in DB as expected, which defaults this value to 0.

```
(Pdb) self.port_dict["Ethernet194"]
{'index': 25, 'speed': '100000', 'lanes': '451,452', 'admin_status': 'up', 'cmis_state': 'INSERTED', 'cmis_retries': 0, 'cmis_expired': None}
```

### Problem 2
[This Pull Request](https://github.com/sonic-net/sonic-buildimage/pull/18204/files#diff-122f65421d488df19ed25bdb38aededeb48eb7753ff07cd6978ccbef34c1e51e) does not play well with all HWSKUs that define breakouts. It does a lookup for every child port in the hwsku.json file - There should be 1 default breakout mode for each front panel port - this change expects every child port in a front panel port to have an entry - which causes an exception. Right now as far as I can tell this will break a lot of HWSKUs in a breakout config.

For example this is what the Dynamic Port Breakout feature uses on an Arista SKU to define 2x200G on the first front panel port:
```
"Ethernet0": {
     "default_brkout_mode": "2x200G[100G,50G,40G,25G,10G]"
},        
```
but it currently expects:
```
"Ethernet0": {
    ...
},    
"Ethernet4": {
    ...
},     
```

#### How I did it
Calculate subport value automatically based on the number of child ports used per front panel port and insert it into CONFIG_DB for the interface. If that number is 1 for the front panel, subport 0 is used. If has more than 1 I.e. is in breakout mode use a 1 based subport number as the feature expects. For example:

1x400G-8:
```
Ethernet0: subport 0
```

4x100G-2:
```
Ethernet0: subport 1
Ethernet2: subport 2
Ethernet4: subport 3
Ethernet6: subport 4
```

 I have kept it compatible with Mellanox change - if you want to define a different subport number in your hwsku.json that is fine and it will override the dynamically generated default.

#### How to verify it
I verified this by checking the output of sonic-cfggen, updating config_db. and checking that both the host and media lane masks were correct for a 400GBASE-DR4 xcvr in both 2x200G-4 mode and 4x100G-2 mode.

All links came up and xcvrd prints the correct lanemasks in logs.

I also made sure that portconfig.py no longer causes an exception; it only checks for the optional HWSKU value if childports are present in hwsku.json file (this is to make sure we don't break Mellanox hwsku's)

I also changed the unit test for config_db generation and ensured the correct values match. The test passes.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [x] 202305
- [x] 202311

- [ ]  202305
- [ ]  202311
- [ ]  master


prgeor:
@bobbymcgonigle there is merge conflict

prgeor:
@liat-grozovik please have this reviewed for your platform

prgeor (approved):
@bobbymcgonigle Thanks for taking care of DPB and static DPB cases.

src/sonic-config-engine/portconfig.py
```diff
@@ -376,7 +378,8 @@ def get_config(self):
                     'alias': self._breakout_capabilities[alias_id],
                     'lanes': ','.join(lanes),
                     'speed': str(entry.default_speed),
-                    'index': self._indexes[lane_id]
+                    'index': self._indexes[lane_id],
+                    'subport': "0" if total_num_ports == 1 else str(alias_id + 1)
```
vivekrnv, line 382: So this field will be populated even if the hwsku.json doesn't define any subport?

src/sonic-config-engine/portconfig.py
```diff
@@ -423,9 +426,10 @@ def parse_platform_json_file(hwsku_json_file, platform_json_file):
 
         # take optional fields from hwsku.json
         for child_port in child_ports:
-            for key, item in hwsku_dict[INTF_KEY][child_port].items():
-                if key in OPTIONAL_HWSKU_ATTRIBUTES:
-                    child_ports.get(child_port)[key] = item
```
vivekrnv, line 428: Consider this case, Ethernet0 -> {Ethernet0 & Ethernet4}. Default computation would infer the value to be '1' for Ethernet0 but Ethernet0 according to the hwsku.json would be "0". So, i think letting the value of hwsku.json overriding the computed value wouldn't work.

src/sonic-config-engine/portconfig.py
vivekrnv, line 382: Will this flow be called for both Static and Dynamic breakouts? Like during load_minigraph etc?

Junchao-Mellanox, line 382: I got from HLD https://github.com/sonic-net/SONiC/blob/master/doc/dynamic-port-breakout/sonic-dynamic-port-breakout-HLD.md that subport number should be 0 if there is no breakout, but it will start from 1 if there is breakout. What is the benefit of it? Why not always set subport to 1 for the first port no matter there is breakout or not?
```
Configure a unique subport# for each broken-down (logical) port in platform's port_config.ini

subport# to start with 1 (and increment sequentailly for each logical port) under the same physcial port
subport# sequence may repeat for logical ports under another physcial port
subport# as 0 (on a port) implies physical port itself (i.e. no port breakout on it)
```

src/sonic-config-engine/portconfig.py
```diff
@@ -423,9 +426,10 @@ def parse_platform_json_file(hwsku_json_file, platform_json_file):
...
+            if child_port in hwsku_dict[INTF_KEY]:
```
Junchao-Mellanox, line 429: better to assign a variable for `hwsku_dict[INTF_KEY]` to avoid looking up it in every iteration

src/sonic-config-engine/portconfig.py
prgeor: @Junchao-Mellanox we want to keep it simple for Xcvrd to distinguish between a breakout port (subport >= 1) vs a non-breakout port(subport = 0)

prgeor:
@bobbymcgonigle please take care of the build failures

src/sonic-config-engine/portconfig.py
bobbymcgonigle: Are you suggesting I remove hwsku.json optional subport value altogether ?

src/sonic-config-engine/portconfig.py
bobbymcgonigle: Yes this is called both on config_db.json creation + dynamic breakouts.

For example
```
Ports to be deleted :
 {
    "Ethernet0": "400000"
}
Ports to be added :
 {
    "Ethernet0": "100000",
    "Ethernet2": "100000",
    "Ethernet4": "100000",
    "Ethernet6": "100000"
}
Breakout process got successfully completed.
Please note loaded setting will be lost after system reboot. To preserve setting, run `config save`.
root@bthd104:/# sonic-db-cli APPL_DB hgetall PORT_TABLE:Ethernet4
{'alias': 'Ethernet1/5', 'index': '1', 'lanes': '5,6', 'speed': '100000', 'subport': '3', 'mtu': '9100', 'admin_status': 'down'}
root@bthd104:/# sonic-db-cli APPL_DB hgetall PORT_TABLE:Ethernet0
{'alias': 'Ethernet1/1', 'index': '1', 'lanes': '1,2', 'speed': '100000', 'subport': '1', 'mtu': '9100', 'oper_status': 'down', 'admin_status': 'down'}
root@bthd104:/#
```

src/sonic-config-engine/portconfig.py
vivekrnv: Yes, if we can infer subport by looking at lanes directly i don't see a need for hwsku.json subport value.

bobbymcgonigle: Removed subport as an optional value

Junchao-Mellanox (approved):

prgeor (dismissed):

vivekrnv (approved):

lguohan:
@bobbymcgonigle , vstest keeps failing, can you check?

prsunny:
@bobbymcgonigle , please look into #19041

lguohan:
@bobbymcgonigle , there is conflict for picking to 202311, can you create a cherry-pick PR?

PR #18504, opened 2024-03-29
Update sonic-ztp submodule to 739470d

rajann:
#### Why I did it

Update sonic-ztp to SHA 739470d.
This updated is needed for 202305 ZTP based config_db plugin to successfully executed


StormLiangMS (approved):
lgtm

PR #18507, opened 2024-03-29
[sonic_platform] fix thermal.py for e1031

Xichen96:
Postpone get_temperature so call to Thermal() will not fail early in bootup.

#### Why I did it
When determine-reboot-cause is run early, there is very chance it will fail on call to Chassis(), which calls Thermal(), because it checks hwmon that was not yet initialized. However checking hwmon to get temperature can be postponed to later.

##### Work item tracking

#### How I did it
Postpone get_temperature to later

#### How to verify it
Make the change on 202305 and bootup many times to check if there are failures.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [x] 202305

202305.24

- [ ] 
- [ ]

### DIFF
--- a/device/celestica/x86_64-cel_e1031-r0/sonic_platform/thermal.py
+++ b/device/celestica/x86_64-cel_e1031-r0/sonic_platform/thermal.py
@@ -109,12 +109,16 @@ class Thermal(ThermalBase):
 
         self.name = self.get_name()
         self.postion = self._thermal_info["postion"]
-        self.minimum_thermal = self.get_temperature()
-        self.maximum_thermal = self.get_temperature()
+        self.minimum_thermal = None
+        self.maximum_thermal = None
 
     def _get_hwmon_path(self):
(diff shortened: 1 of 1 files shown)


prgeor:
@Xichen96 Why is this a new issue in 202305 but not seen in previous release? Also, please attach the process crash backtrace or the failure seen in bootup

prgeor:
@Xichen96 The platform driver has not completed initialization before determine-reboot-cause service could run. To avoid this add the following dependency in https://github.com/sonic-net/sonic-buildimage/blob/master/platform/broadcom/sonic-platform-modules-cel/haliburton/systemd/platform-modules-haliburton.service#L5

image

lizhijianrd:
Close this PR. Use #18526 instead.

PR #18517, opened 2024-04-01
Migrate the dpkg lock to container path

w1nda:
#### Why I did it
Because dpkg does not allow installing packages in parallel, we added a lock.

The lock is under repo path target/debs/$(BLDENV)/dpkg_lock, when the building progress is terminated by user, and the lock is not cleaned, the next time building progress will stuck at dpkg command.

As the lock should only exists when building, we raise this PR to migrate the lock to container path, so the lock will live on in the container.

##### Work item tracking

#### How I did it
Migrate the lock to folder /tmp/dpkg_lock which is in the container.

#### How to verify it
Triger the pipeline for build 

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/slave.mk
+++ b/slave.mk
@@ -27,6 +27,7 @@ SRC_PATH = src
 RULES_PATH = rules
 TARGET_PATH = target
 DOCKERS_PATH = dockers
+CONTAINER_TMP_PATH = /tmp
 BLDENV := $(shell lsb_release -cs)
 DEBS_PATH = $(TARGET_PATH)/debs/$(BLDENV)
 FILES_PATH = $(TARGET_PATH)/files/$(BLDENV)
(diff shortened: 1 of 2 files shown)


slave.mk
```diff
@@ -109,6 +110,7 @@ ifneq ($(CONFIGURED_PLATFORM),generic)
 endif
 
 configure :
+	$(Q)mkdir -p $(CONTAINER_TMP_PATH)
```
xumia, line 113: Looks like it is not necessary to create the dir /tmp, the /tmp path should always exist.

w1nda: Thanks, removed this lne.

saiarcot895:
`platform/mellanox/mft/Makefile` is also managing `dpkg_lock`,, this path may need to be updated (ideally, they shouldn't be doing any package installations at that stage at all).

w1nda:
Hi @xumia,
Hi @liushilongbuaa,
cloud you help review this PR?
thanks

w1nda:
> `platform/mellanox/mft/Makefile` is also managing `dpkg_lock`,, this path may need to be updated (ideally, they shouldn't be doing any package installations at that stage at all).

As the Makefile is not commonly used, I prefer to leave it as it was.

PR #18541, opened 2024-04-02
Add new HwSKUs for Arista-7060dx5-32

bobbymcgonigle:
#### Why I did it
New desired configurations on this Arista SKU

##### Work item tracking

#### How I did it
Added the hwsku folders with the necessary files to populate config_db

#### How to verify it
change default sku and generate a new config_db.json

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [x] 202305
- [x] 202311

- [x] 202305

### DIFF
--- a/device/arista/x86_64-arista_7060dx5_32/Arista-7060DX5-32-200Gx56-400Gx4/hwsku.json
+++ b/device/arista/x86_64-arista_7060dx5_32/Arista-7060DX5-32-200Gx56-400Gx4/hwsku.json
@@ -1,0 +1,105 @@
+{
+    "interfaces": {
+        "Ethernet0": {
+            "default_brkout_mode": "2x200G[100G,50G,40G,25G,10G]"
+        },
+        "Ethernet8": {
+            "default_brkout_mode": "2x200G[100G,50G,40G,25G,10G]"
+        },
+        "Ethernet16": {
+            "default_brkout_mode": "2x200G[100G,50G,40G,25G,10G]"
+        },
+        "Ethernet24": {
+            "default_brkout_mode": "2x200G[100G,50G,40G,25G,10G]"
+        },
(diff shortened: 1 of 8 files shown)


zhenggen-xu (commented):
Can you add buffer/qos config files similar to [Arista-7060DX5-32](https://github.com/sonic-net/sonic-buildimage/tree/master/device/arista/x86_64-arista_7060dx5_32/Arista-7060DX5-32) to the new HWKSUs?

zhenggen-xu (commented):
Is it possible to use the same th4XXX.bcm file for both HWSKUs?  We should make the bcm file agnostic to HWSKUs, and let SONiC SAI APIs to add/delete ports during the initialization, this will make future HWSKU creation straightforward.

bobbymcgonigle:
> Is it possible to use the same th4XXX.bcm file for both HWSKUs? We should make the bcm file agnostic to HWSKUs, and let SONiC SAI APIs to add/delete ports during the initialization, this will make future HWSKU creation straightforward.

Unfortunatly this is not available in 202305 (where we would like this to go) due to CSP case CS00012347969. This is fixed in SAI 9. 

Until 202305 moves to that SAI version we cannot support the single bcm file.

> Can you add buffer/qos config files similar to [Arista-7060DX5-32](https://github.com/sonic-net/sonic-buildimage/tree/master/device/arista/x86_64-arista_7060dx5_32/Arista-7060DX5-32) to the new HWKSUs?

We are planning to do this in a seperate PR. For now we would like to get these hwsku's merged to unblock testing with these front panel configurations.

zhenggen-xu (approved):
LGTM.

Issue #18358, opened 2024-03-13
Building sonic VS image : FAIL: test_buffers_dell6100_render_template (tests.test_j2files.TestJ2Files)


TafkaMax:
#### Description

Building Sonic VS image on a Ubuntu 22.04 VM with nested virt support.

#### Steps to reproduce the issue:
1.  Ubuntu 22.04 VM
2. 
```
stages:
  - build

variables:
  UNATTENDED: 1

.build_sonic:
  stage: build
  timeout: 5h
  before_script:
    - docker info
    - export PATH="~/.local/bin:$PATH"
    - pip3 install --user j2cli --quiet
    - git clone --recurse-submodules https://github.com/sonic-net/sonic-buildimage.git
    - cd sonic-buildimage
    # Not needed, if we do not need "patching" functionality.
    #- git restore .
    #- cd ..
    #- ./apply-patches.sh sonic-buildimage/sonic-slave-buster/Dockerfile.user.j2
    #- ./apply-patches.sh sonic-buildimage/sonic-slave-bullseye/Dockerfile.user.j2
    #- ./apply-patches.sh sonic-buildimage/sonic-slave-bookworm/Dockerfile.user.j2
    #- cd sonic-buildimage
    # We need to reset builds here. Resetting it later on, will reset our custom changes aswell.
    - git checkout "$CI_COMMIT_REF_NAME"
    - git restore .
    - make reset
    #- cp ../patches/sonic-config.mk rules/sonic-config.mk
    #- cp ../patches/sonic-platform-common.mk rules/sonic-platform-common.mk
    #- cp ../patches/device_info.py src/sonic-py-common/sonic_py_common/device_info.py
    - sed -i 's/INCLUDE_ICCPD = n/INCLUDE_ICCPD = y/' rules/config
    - sed -i 's/INCLUDE_RESTAPI = n/INCLUDE_RESTAPI = y/' rules/config
    - sed -i 's/INCLUDE_FIPS ?= y/INCLUDE_FIPS = n/' rules/config
    - sed -i 's/BUILD_MULTIASIC_KVM = n/BUILD_MULTIASIC_KVM = y/' rules/config
    - sed -i 's/SONIC_CONFIG_BUILD_JOBS = 1/SONIC_CONFIG_BUILD_JOBS = 4/g' rules/config
  script:
    - make init
    - make configure PLATFORM=$PLATFORM
    - make list
[...]

PR #18143, opened 2024-02-21
[Marvell] Build infrastructure enhancements 

pavannaregundi:
#### Why I did it
This change is about marvell build infrastructure enhancement by merging architecture specific marvell platform(marvell-arm64 and marvell-armhf) folders under 'marvell'.

##### Work item tracking
- Microsoft ADO **(number only)**:

#### How I did it
This is archived with following changes divided into following commits,
* Move mrvl-prestera and sonic-platform-marvell submodule 
   'mrvl-prestera' and 'sonic-platform-marvell' submodules are moved from 'marvell-arm64'to 'marvell'. Additionaly submodules are changed to support multi-architecture build. 

* Create sonic-platform-nokia from marvell-arm64/armhf
   'sonic-platform-nokia' is created from merge of 'marvell-armhf' and 'marvell-arm64' in 'marvell'. Additionally changes are done in debian/rules to support multi-architecture build.

* Change marvell platform makefile to support all arch
  Combine all the architecture targets in makefiles. Added CONFIGURED_ARCH check to build for specific architecture.
  Updated sai deb which matches latest SAI commit.

* Add support for architecture specific platform.conf
  Changed build_image.sh and onie-mk-demo.sh to read architecture specific platform.conf.
  Moved marvell-arm64/platform.conf to marvell/platform_arm64.conf.
  Moved marvell-armhf/platform.conf to marvell/platform_armhf.conf.

* [arm64] Adjust the path to read sonic_fit.its
   Use CONFIGURED_PLATFORM based path while reading sonic_fit.its.

* [arm64/armhf] Add architecure suffix for output image
  This change is needed to generate the output image with architecture suffix for arm64/armhf platform.
  For arm64, defined a marvell specific file to avoid affect on other vendors.

* Change platform_asic name to marvell
  Change 'marvell-arm64' and 'marvell-armhf' platform_asic name to 'marvell'.

* Remove marvell-armhf and marvell-arm64 folders
  Removed 'marvell-armhf' and 'marvell-arm64' as infra is combined with 'marvell' folder.

* Update readme to reflect new build commands

* Change azure pipeline commands for marvell

New build commands:
AMD64 - make configure PLATFORM=marvell && make  target/sonic-marvell.bin
ARM64 - make configure PLATFORM=marvell PLATFORM_ARCH=arm64 && make target/sonic-marvell-arm64.bin
ARMHF - make configure PLATFORM=marvell PLATFORM_ARCH=armhf && make  target/sonic-marvell-armhf.bin

[...]

PR #18566, opened 2024-04-04
Fix for sonic-config-engine build failure - test_buffers_dell6100_render_template 

krismarvell:
#### Why I did it
To fix sonic-config-engine build error at src/sonic-config-engine/tests/test_j2files.py unit test case while using "make PLATFORM=$PLATFORM target/sonic-$PLATFORM.bin" commands. This will fix the issue #18358 
Usage of PLATFORM=$PLATFORM is expected to work, and will be needed where the azure pipeline builds are executing make commands with PLATFORM parameter. 

##### Work item tracking

#### How I did it
By unsetting the PLATFORM make flag in python-wheel generator targets in slave.mk 

#### How to verify it
run the minimal build command make PLATFORM=marvell target/python-wheels/buster/sonic_config_engine-1.0-py2-none-any.whl --> To just build config-engine wheel package for buster as an example. 
Verify the errors in below test cases of src/sonic-config-engine/tests/test_j2files.py is NOT seen.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

master

### DIFF
--- a/slave.mk
+++ b/slave.mk
@@ -927,6 +927,9 @@ $(addprefix $(PYTHON_WHEELS_PATH)/, $(SONIC_PYTHON_WHEELS)) : $(PYTHON_WHEELS_PA
 	# Load the target deb from DPKG cache
 	$(call LOAD_CACHE,$*,$@)
 
+	# unset PLATFORM Flag for python wheel generation
+	PLATFORM=
+	export PLATFORM
 	# Skip building the target if it is already loaded from cache
 	if [ -z '$($*_CACHE_LOADED)' ] ; then
 


krismarvell:
@saiarcot895 Can you pls help with review and comment

liushilongbuaa:
I don't think it will work.
Official build didn't add 'PLATFORM=$PLATFORM' options when building target/sonic-vs.bin

krismarvell:
> Official build didn't add 'PLATFORM=$PLATFORM' options when building target/sonic-vs.bin

Yes I agree with you on the official build comment. current builds doesnt have this option, but still this fix will be a foolproof for two reasons.
1. this option may need to be used in make command for any specific use case (e.g. #18143 ) 
2. intuitively it looked to me like even if PLATFORM=$PLATFORM is used in the make command after a make configure PLATFORM=$PLATFORM ideally it shud not fail the build.
Pls let me know your comments. 

Also, when you refer as this wont work, do you refer to any builds/platforms/options where you think it may not work? Pls help me understand if i have missed any specific UT/BUild types, i can give a try with this fix.

slave.mk
```diff
@@ -927,6 +927,9 @@ $(addprefix $(PYTHON_WHEELS_PATH)/, $(SONIC_PYTHON_WHEELS)) : $(PYTHON_WHEELS_PA
 	# Load the target deb from DPKG cache
 	$(call LOAD_CACHE,$*,$@)
 
+	# unset PLATFORM Flag for python wheel generation
+	PLATFORM=
+	export PLATFORM
```
saiarcot895, line 932: Is it possible to have this be applied early in `slave.mk`, so that it applies for all builds, not just Python wheels?

liushilongbuaa:
> > Official build didn't add 'PLATFORM=$PLATFORM' options when building target/sonic-vs.bin
> 
> Yes I agree with you on the official build comment. current builds doesnt have this option, but still this fix will be a foolproof for two reasons.
> 
> 1. this option may need to be used in make command for any specific use case (e.g. [[Marvell] Build infrastructure enhancements  #18143](https://github.com/sonic-net/sonic-buildimage/pull/18143) )
> 2. intuitively it looked to me like even if PLATFORM=$PLATFORM is used in the make command after a make configure PLATFORM=$PLATFORM ideally it shud not fail the build.
>    Pls let me know your comments.
> 
> Also, when you refer as this wont work, do you refer to any builds/platforms/options where you think it may not work? Pls help me understand if i have missed any specific UT/BUild types, i can give a try with this fix.

I forgot it is for PR18143. [sonic-py-common](https://github.com/sonic-net/sonic-buildimage/blob/2c1f1aae6be55fbcb75ec8d4af580fd41f6deea3/src/sonic-py-common/sonic_py_common/device_info.py#L114) used env: PLATFORM. It makes sense to reset that variable when building image.

slave.mk
krismarvell: It looks like anywhere else unsetting this outside specific targets, have the effect of globally changing the flag even for "make configure" as well. Hence have added only in the specific python wheel targets where this is hitting into the error.

PR #18597, opened 2024-04-08
Yang model changes for Sequential IDF isolation

tjchadaga:
#### Why I did it
To define a field to track isolation state for the IDF to which the T2 device belongs

##### Work item tracking

#### How I did it
Added yang model definition, unit tests, sample config and documentation for the table

#### How to verify it
Validated config tree generation using "pyang -Vf tree -p /usr/local/share/yang/modules/ietf ./yang-models/sonic-voq-inband-interface.yang"

Built the below python-wheels to validate unit tests and other changes
target/python-wheels/bullseye/sonic_yang_mgmt-1.0-py3-none-any.whl
target/python-wheels/bullseye/sonic_yang_models-1.0-py3-none-any.whl
target/python-wheels/bullseye/sonic_config_engine-1.0-py3-none-any.whl

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ]  202205 (chassis)
- [ ]

### DIFF
--- a/src/sonic-yang-models/yang-models/sonic-bgp-device-global.yang
+++ b/src/sonic-yang-models/yang-models/sonic-bgp-device-global.yang
@@ -21,11 +21,25 @@ module sonic-bgp-device-global {
         container BGP_DEVICE_GLOBAL {
             container STATE {
                 description "BGP device-specific global data";
-                leaf tsa_enabled {
+                leaf tsa_enabled {                    
                     type boolean;
                     default "false";
                     description
                     "When set to true, Traffic is shifted away (TSA), i.e, BGP routes are not advertised to neighboring routers";
+                }
+                leaf idf_isolation_state {                    
+                    type enumeration {
+                        enum isolated_no_export {
+                            description "IDF isolated using no-export community tag";
+                        }
+                        enum isolated_withdraw_all {
+                            description "IDF isolated by withdrawing routes";
+                        }
+                        enum unisolated {
+                            description "IDF un-isolated";
+                        }
+                    }
+                    default unisolated;
                 }
             } /* end of STATE container */
         } /* end of BGP_DEVICE_GLOBAL container */
--- a/src/sonic-yang-models/yang-models/sonic-bgp-device-global.yang
+++ b/src/sonic-yang-models/yang-models/sonic-bgp-device-global.yang
@@ -21,13 +21,13 @@ module sonic-bgp-device-global {
         container BGP_DEVICE_GLOBAL {
             container STATE {
                 description "BGP device-specific global data";
-                leaf tsa_enabled {                    
+                leaf tsa_enabled {
                     type boolean;
                     default "false";
                     description
                     "When set to true, Traffic is shifted away (TSA), i.e, BGP routes are not advertised to neighboring routers";
                 }
-                leaf idf_isolation_state {                    
+                leaf idf_isolation_state {
                     type enumeration {
                         enum isolated_no_export {
                             description "IDF isolated using no-export community tag";
--- a/files/build_templates/init_cfg.json.j2
+++ b/files/build_templates/init_cfg.json.j2
@@ -30,7 +30,8 @@
     },
     "BGP_DEVICE_GLOBAL": {
         "STATE": {
-            "tsa_enabled": "false"
+            "tsa_enabled": "false",
+            "idf_isolation_state": "unisolated"
         }
     },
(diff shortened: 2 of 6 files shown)


src/sonic-yang-models/yang-models/sonic-bgp-device-global.yang
```diff
@@ -21,12 +21,26 @@ module sonic-bgp-device-global {
...
             container STATE {
                 description "BGP device-specific global data";
-                leaf tsa_enabled {
+                leaf tsa_enabled {                    
                     type boolean;
                     default "false";
                     description
                     "When set to true, Traffic is shifted away (TSA), i.e, BGP routes are not advertised to neighboring routers";
                 }
+                leaf idf_isolation_state {
```
qiluo-msft, line 30: >idf_isolation_state [](http://example.com/codeflow?start=21&length=19)

Is there any limitation of this field on M0/T0/T1 devices? If no, could we generalize the isolation for all roles, not to mention IDF?

tjchadaga: IDF isolation is only supported on T2 and the config will not have any effect on T0/T1 devices.

qiluo-msft: Then can we enforce a value for M0/T0/T1? The yang model could help the enforcement.

tjchadaga: Currently, the field is initialized to "unisolated" on all devices (including M0/T0/T1) in init_cfg.json.j2.  Changing this configuration through CLI/Script is blocked on non-T2 devices. Is that sufficient?

src/sonic-yang-models/yang-models/sonic-bgp-device-global.yang
```diff
@@ -21,12 +21,26 @@ module sonic-bgp-device-global {
         container BGP_DEVICE_GLOBAL {
             container STATE {
                 description "BGP device-specific global data";
-                leaf tsa_enabled {
+                leaf tsa_enabled {
```
ganglyu, line 24: Please remove space from end of line.

src/sonic-yang-models/yang-models/sonic-bgp-device-global.yang
```diff
@@ -21,12 +21,26 @@ module sonic-bgp-device-global {
...
+                leaf tsa_enabled {                    
                     type boolean;
                     default "false";
                     description
                     "When set to true, Traffic is shifted away (TSA), i.e, BGP routes are not advertised to neighboring routers";
                 }
+                leaf idf_isolation_state {                    
+                    type enumeration {
+                        enum isolated_no_export {
+                            description "IDF isolated using no-export community tag";
```
xincunli-sonic, line 33: >IDF isolated using no-export community tag [](http://example.com/codeflow?start=41&length=42)

Can we add DC size - Medium/Large

src/sonic-yang-models/yang-models/sonic-bgp-device-global.yang
```diff
@@ -21,12 +21,26 @@ module sonic-bgp-device-global {
...
+                        }
+                        enum isolated_withdraw_all {
+                            description "IDF isolated by withdrawing routes";
```
xincunli-sonic, line 36: >IDF isolated by withdrawing routes [](http://example.com/codeflow?start=41&length=34)

Same, also add DC size - Small/X-Small ?

src/sonic-yang-models/yang-models/sonic-bgp-device-global.yang
tjchadaga: This field is just meant to indicate that the isolation is to be done by advertising routes tagged with no-export BGP community and is independent of any DC size.

src/sonic-yang-models/yang-models/sonic-bgp-device-global.yang
tjchadaga: This field is just meant to indicate that the isolation is to be done by withdrawing routes and is independent of any DC size.

src/sonic-yang-models/yang-models/sonic-bgp-device-global.yang
tjchadaga: done

src/sonic-yang-models/yang-models/sonic-bgp-device-global.yang
qiluo-msft: What is the backend behavior if a T0 got config which is not unisolated? will it fail or ignore?

If the backend will treat this as a failure, we can also hardening yang models, so when ConfigDB is validated by yang models, the failure could be detected earlier, and stop the configuration process immediately.

tjchadaga: Only T2 will handle this config change. T0/T1 will ignore it

arlakshm (approved):

qiluo-msft (approved):
LGTM. Please check with other active reviewers.

PR #18609, opened 2024-04-09
[multi-asic] Fixed 13137 ERR python3: :- initializeGlobalConfig: SonicDBConfig Global config is already initialized 

wumiaont:
**Issue**: 
https://github.com/sonic-net/sonic-buildimage/issues/13137
If run "show interfaces status" we often see error in syslog "SonicDBConfig Global config is already initialized".

**Analysis**: 
During investigation it is found the swsscommon.load_sonic_global_db_config() has been triggered multiple times. Especially in a multi Asic scenario.

CLI triggers "intfutil -c status" with a separate process. Within the process it will trigger IntfStatus init --> MultiAsic() init-->load_db_config()-->load_sonic_global_db_config(). This time initializeGlobalConfig() will run to the end as m_global_init flag is false.

CLI will also run get_port_config() --> db_connect_configdb() --> load_sonic_global_db_config() (for different namepaces. In our case is asic0 and asic1). This time m_global_init flag is true which causes error log and return.

**Solution**: 
There was a PR in master already fixed this issue(https://github.com/sonic-net/sonic-buildimage/pull/10960). Cherry-picked this back to 202205. Added checks in 2 other places where load_sonic_global_db_config is called without check in the code and was not covered by PR 10960. 

**Testing**
Run on chassis with 202205 loaded. Loaded the python files modified with this PR into the chassis. Run "show interface status" and will not see the error log. Show command has interface status results correctly.

### DIFF
--- a/src/sonic-config-engine/sonic-cfggen
+++ b/src/sonic-config-engine/sonic-cfggen
@@ -138,11 +138,12 @@ def ip_network(value):
     return r_v.network
 
 def load_namespace_config(asic_name):
-    if not SonicDBConfig.isInit():
-        if is_multi_asic():
-            SonicDBConfig.load_sonic_global_db_config(namespace=asic_name)
-        else:
-            SonicDBConfig.load_sonic_db_config()
+    if is_multi_asic():
+        if not SonicDBConfig.isGlobalInit():
+            SonicDBConfig.initializeGlobalConfig()
+    else:
+        if not SonicDBConfig.isInit():
+            SonicDBConfig.initialize()
 
 class FormatConverter:
     """Convert config DB based schema to legacy minigraph based schema for backward capability.
--- a/src/sonic-config-engine/sonic-cfggen
+++ b/src/sonic-config-engine/sonic-cfggen
@@ -137,7 +137,7 @@ def ip_network(value):
         return "Invalid ip address %s" % value
     return r_v.network
 
-def load_namespace_config(asic_name):
+def load_namespace_config():
     if is_multi_asic():
         if not SonicDBConfig.isGlobalInit():
             SonicDBConfig.initializeGlobalConfig()
@@ -309,7 +309,7 @@ def main():
         deep_update(data, hardware_data)
         if args.port_config is None:
             args.port_config = device_info.get_path_to_port_config_file(hwsku)
-        load_namespace_config(asic_name)
+        load_namespace_config()
         (ports, _, _) = get_port_config(hwsku, platform, args.port_config, asic_id)
         if ports is None:
             print('Failed to get port config', file=sys.stderr)
@@ -335,7 +335,7 @@ def main():
 
     if args.minigraph is not None:
         minigraph = args.minigraph
-        load_namespace_config(asic_name)
+        load_namespace_config()
         if platform:
             if args.port_config is not None:
                 deep_update(data, parse_xml(minigraph, platform, args.port_config, asic_name=asic_name, hwsku_config_file=args.hwsku_config))
@@ -363,7 +363,7 @@ def main():
         if args.namespace is None:
             configdb = ConfigDBPipeConnector(use_unix_socket_path=use_unix_sock, **db_kwargs)
         else:
-            load_namespace_config(args.namespace)
+            load_namespace_config()
             configdb = ConfigDBPipeConnector(use_unix_socket_path=use_unix_sock, namespace=args.namespace, **db_kwargs)
 
         configdb.connect()
@@ -435,7 +435,7 @@ def main():
         if args.namespace is None:
             configdb = ConfigDBPipeConnector(use_unix_socket_path=True, **db_kwargs)
         else:
-            load_namespace_config(args.namespace)
+            load_namespace_config()
             configdb = ConfigDBPipeConnector(use_unix_socket_path=True, namespace=args.namespace, **db_kwargs)
 
         configdb.connect(False)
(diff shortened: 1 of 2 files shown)


wumiaont:
This is a cherry-pick fix from master. Added 2 more places with check where load_sonic_global_db_config() is called without check.

src/sonic-config-engine/sonic-cfggen
```diff
@@ -137,6 +137,13 @@ def ip_network(value):
         return "Invalid ip address %s" % value
     return r_v.network
 
+def load_namespace_config(asic_name):
+    if not SonicDBConfig.isInit():
+        if is_multi_asic():
+            SonicDBConfig.load_sonic_global_db_config(namespace=asic_name)
+        else:
+            SonicDBConfig.load_sonic_db_config()
```
qiluo-msft, line 145: >load_sonic_db_config [](http://example.com/codeflow?start=26&length=20)

Could you use `SonicDBConfig.initialize`? yours is for backward-compatible purpose only, should be deprecated in new code.

src/sonic-config-engine/sonic-cfggen
```diff
@@ -137,6 +137,13 @@ def ip_network(value):
         return "Invalid ip address %s" % value
     return r_v.network
 
+def load_namespace_config(asic_name):
+    if not SonicDBConfig.isInit():
+        if is_multi_asic():
+            SonicDBConfig.load_sonic_global_db_config(namespace=asic_name)
```
qiluo-msft, line 143: >load_sonic_global_db_config [](http://example.com/codeflow?start=26&length=27)

Could you use `SonicDBConfig.initializeGlobalConfig`? yours is for backward-compatible purpose only, should be deprecated in new code.

src/sonic-config-engine/sonic-cfggen
```diff
@@ -137,6 +137,13 @@ def ip_network(value):
         return "Invalid ip address %s" % value
     return r_v.network
 
+def load_namespace_config(asic_name):
+    if not SonicDBConfig.isInit():
```
qiluo-msft, line 141: >isInit [](http://example.com/codeflow?start=25&length=6)

You should check `isInit` or `isGlobalInit` based on single/multi-asic.

src/sonic-config-engine/sonic-cfggen
wumiaont: Changed.

src/sonic-config-engine/sonic-cfggen
wumiaont: Yes. Fixed.

src/sonic-config-engine/sonic-cfggen
wumiaont: Changed

wumiaont:
@qiluo-msft Please review. Thx

src/sonic-config-engine/sonic-cfggen
rlhui: @wumiaont - would you please also raise same change in the master PR version so that changes are consistent and not only in 202205? Thanks.

wumiaont: > @wumiaont - would you please also raise same change in the master PR version so that changes are consistent and not only in 202205? Thanks.

@rlhui This is a port fix from master. Yes. that fix has some issue. I am going to fix this in master also after this is got into 2205.

wumiaont: > > @wumiaont - would you please also raise same change in the master PR version so that changes are consistent and not only in 202205? Thanks.
> 
> @rlhui This is a port fix from master. Yes. that fix has some issue. I am going to fix this in master also after this is got into 2205.

Master fix was this one https://github.com/sonic-net/sonic-buildimage/issues/13137. Has a slight issue as checking the incorrect flag. Will correct this in master.

qiluo-msft:
The code lgtm. However you should fix the master completely and backport whatever commits to 202205. Not to submit PR to 202205 directly.

qiluo-msft:
The code lgtm. However you should fix the master completely and backport whatever commits to 202205. Not to submit PR to 202205 directly.

wumiaont:
> The code lgtm. However you should fix the master completely and backport whatever commits to 202205. Not to submit PR to 202205 directly.

OK. I will create PR against master soon.

wumiaont:
@qiluo-msft @rlhui Created PR against master https://github.com/sonic-net/sonic-buildimage/pull/18691. Please review. Unfortunately, that PR cannot be cherry-pick to 202205 as it's missing another important fix with original PR 10960 (change in portconfig.py). So best way is to let this in 202205 and have 18691 to handle fix for master.

qiluo-msft (approved):

deepak-singhal0408 (approved):

deepak-singhal0408:
MSFT ADO: 25516271

deepak-singhal0408:
@rlhui , could you pl help merge this change? Thanks!

gechiang:
@deepak-singhal0408 why is the label requested for 202305 is applied here but not from the master PR (https://github.com/sonic-net/sonic-buildimage/pull/10960) instead?  This is not usually how it works...  
Is the additional fix also needed for master?  If so, please raise another PR in master so that the master has all the fixes and request the back port from 10960 and the new PR that you will be raising.  I will remove the label "request for 202305" from this one for now.

deepak-singhal0408:
> @deepak-singhal0408 why is the label requested for 202305 is applied here but not from the master PR (#10960) instead? This is not usually how it works... Is the additional fix also needed for master? If so, please raise another PR in master so that the master has all the fixes and request the back port from 10960 and the new PR that you will be raising. I will remove the label "request for 202305" from this one for now.

sounds good.. actually earlier there was only one PR in 202205, and we were under the impression that nothing is needed in master.. thats why I added the label here.. but later on, we made a change for this in master as well.. so yes, ideally the label should be put on master branch PR.. thanks for fixing this.

PR #18620, opened 2024-04-10
PAC changes to receive config updates

vijaya-ops:
#### Why I did it
PAC changes to receive config updates


src/sonic-pac/paccfg/pac_authmgrcfg.cpp
```diff
@@ -0,0 +1,199 @@
...
+
+// Set learning mode for a port 
+bool PacCfg::intfLearningSet(string port, string learning)
+{
+   string key(port);
+   vector<FieldValueTuple> fvVector;
+  
+   // Configure port learning mode for FDB manager 
+   fvVector.emplace_back("learn_mode", learning);
+   m_stateOperPortTable.set(key, fvVector);
```
sutharsansr, line 52: This sets the value of the learning mode in the STATE_DB of PAC. Should we update this to the APPL_DB of port table also so that it is pushed down

sutharsansr (approved):

src/sonic-pac/paccfg/pac_authmgrcfg.cpp
vijaya-ops: PAC doesn't interact directly with the Orch agent. The Orch agent updates are done using VLAN mgr.  Hence PAC uses STATE_DB .

Praveen-Brcm (approved):

PR #18644, opened 2024-04-10
PAC infra libtool

vijaya-ops:
#### Why I did it
PAC infra libtool


src/sonic-pac/fpinfra/libtool
matiAlfaro, line 1: wouldn't it be better to download libtool? 
git clone https://git.savannah.gnu.org/git/libtool.git
or download a specific version from https://ftp.gnu.org/gnu/libtool/

vijaya-ops:
closing and re-opening the PR again.

src/sonic-pac/fpinfra/libtool
vijaya-ops: Yes, That is valid point. May be we can think of taking this as a future item.. For now we intend to keep the changes as is.

vijaya-ops:
closing and reopening the PR

PR #18178, opened 2024-02-25
[systemd/systemd-sonic-generator]: Systemd midplane network service of Smart Switch

Pterosaur:
#### Why I did it
Add systemd network service of Smart Switch for  configuring the midplane network.

##### Work item tracking
- Microsoft ADO **(number only)**:

#### How I did it
1. Add related systemd services and networkd configuration
2. According to paltform.json, the systemd-sonic-generator will install required services.

The changes are based on the HLD: https://github.com/sonic-net/SONiC/pull/1534 and https://github.com/sonic-net/SONiC/pull/1586

#### How to verify it
1. Check Azp that no failure to the existing testcases
2. Check NPU scenario:
2.1 Start a KVM and add the NPU platform.json as following
```
echo '{"DPUS":{"dpu0":{}, "dpu1":{}}}' | sudo tee /usr/share/sonic/device/x86_64-kvm_x86_64-r0/platform.json
```
2.2 You should get two specific database containers for DPU
```
admin@vlab-01:~$ docker ps -a
CONTAINER ID   IMAGE                                COMMAND                  CREATED          STATUS         PORTS     NAMES
4455f7ab611c   docker-database:latest               "/usr/local/bin/dock…"   5 minutes ago    Up 5 minutes             databasedpu0
8e983b5beb27   docker-database:latest               "/usr/local/bin/dock…"   5 minutes ago    Up 5 minutes             databasedpu1
```
2.3 Check the expected service midplane-network-npu.service is enabled but the midplane-network-dpu.service is disabled
```
admin@vlab-01:~$ sudo systemctl list-unit-files '*midplane*'
UNIT FILE                    STATE           PRESET
midplane-network-dpu.service disabled        enabled
midplane-network-npu.service enabled-runtime  enabled
```
2.4 Check the bridge-midplane has been created.
```
admin@vlab-01:~$ sudo brctl show
bridge name     bridge id               STP enabled     interfaces
...
bridge-midplane         8000.76c5a51a18f6       no              dummy-midplane
```
2.5 If we create two mock dpu interfaces, they should be added into the bridge-midplane automatically
```
admin@vlab-01:~$ sudo ip link add dpu0 type veth peer dpu1
admin@vlab-01:~$ sudo brctl show
bridge name     bridge id               STP enabled     interfaces
...
bridge-midplane         8000.76c5a51a18f6       no              dpu0 dpu1 dummy-midplane
```

4. Check DPU scenario:
3.1 Start a KVM and add the DPU platform.json as following
```
[...]

PR #18650, opened 2024-04-10
Added wireless-regdb as package to install for Debian

justin-wong-ce:
#### Why I did it

In Debian 6.1.38, `rfkill` is used, which uses `cfg80211`. `cfg80211` will attempt to load on boot and will look for `regulatory.db`, which is installed via `wireless-regdb`, a missing package.

##### Work item tracking

#### How I did it

Add `wireless-regdb` to packages to install for Debian.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/build_debian.sh
+++ b/build_debian.sh
@@ -402,11 +402,11 @@ sudo LANG=C DEBIAN_FRONTEND=noninteractive chroot $FILESYSTEM_ROOT apt-get -y in
     auditd                  \
     linux-perf              \
     resolvconf              \
-	lsof                    \
-	sysstat                 \
-	xxd                     \
+    lsof                    \
+    sysstat                 \
+    xxd                     \
     wireless-regdb          \
-	zstd
+    zstd
 
 # Have systemd create the auditd log directory
 sudo mkdir -p ${FILESYSTEM_ROOT}/etc/systemd/system/auditd.service.d
--- a/build_debian.sh
+++ b/build_debian.sh
@@ -405,6 +405,7 @@ sudo LANG=C DEBIAN_FRONTEND=noninteractive chroot $FILESYSTEM_ROOT apt-get -y in
 	lsof                    \
 	sysstat                 \
 	xxd                     \
+    wireless-regdb          \
 	zstd
 
 # Have systemd create the auditd log directory


build_debian.sh
```diff
@@ -405,6 +405,7 @@ sudo LANG=C DEBIAN_FRONTEND=noninteractive chroot $FILESYSTEM_ROOT apt-get -y in
 	lsof                    \
 	sysstat                 \
 	xxd                     \
+    wireless-regdb          \
```
StormLiangMS, line 408: could you fix this indent?

StormLiangMS:
hi @saiarcot895  could you help to review?

build_debian.sh
justin-wong-ce: Fixed

Weird that this happened, sorry for the inconvenience

saiarcot895:
I don't think `rfkill` is actually needed on any network switch, since this is usually used as a kill switch for wifi and bluetooth devices, and `cfg80211` is only for wifi devices. Instead, it would be better to not have the `rfkill` kernel module be loaded in the first place.

I see `systemd-rfkill.socket` is getting started on boot automatically, but initial searches online suggest that just disabling this might not prevent the module from being loaded. Something more might need to be done, possibly blacklisting the `rfkill` module itself.

justin-wong-ce:
> I don't think `rfkill` is actually needed on any network switch, since this is usually used as a kill switch for wifi and bluetooth devices, and `cfg80211` is only for wifi devices. Instead, it would be better to not have the `rfkill` kernel module be loaded in the first place.
> 
> I see `systemd-rfkill.socket` is getting started on boot automatically, but initial searches online suggest that just disabling this might not prevent the module from being loaded. Something more might need to be done, possibly blacklisting the `rfkill` module itself.

I have previously tried adding `rfkill` and `cfg80211` to `/etc/modprobe.d/blacklist` as an alternative solution, which seems to also work. However I opted not to do that since I am not 100% sure if any current or future package would become dependent on it on some edge case, especially when it is included in Debian. 

Installing `wireless-regdb` seems to be more robust.

StormLiangMS:
hi @saiarcot895  could you review the new response from Justin?

saiarcot895:
It looks like the cause for the `cfg80211` module getting automatically loaded is mentioned in systemd/systemd#27520 (specifically, systemd/systemd#20802). Some code refactoring occurred in systemd-networkd that causes the module to automatically get loaded, which then looks for the `regulatory.db` firmware file. (Also, the `cfg80211` module depends on the `rfkill` module, not the other way around.)

`systemd-networkd` isn't enabled by default by Debian, but looks like SONiC now uses `systemd-networkd` in one case, starting with #18178, which is probably when this issue started happening. I do see the advantages of using it more; network configuration could potentially become a bit easier with this, but it would take some effort in migrating everything over to it.

Given all of this, either the `cfg80211` module needs to be blacklisted, or the `wireless-regdb` package needs to be added. Considering there could be a theoretical use case of someone trying to use SONiC with wireless interfaces (although it would be clunky), it might be fine to add the `wireless-regdb` package.

saiarcot895 (approved):

PR #18670, opened 2024-04-12
Dns fix

gpunathilell:
#### Why I did it
To fix `resolv-config.service` being cancelled:
`resolv-config.service` is started from `hostcfgd` each time there is an [update to the dns configuration or during the start of hostcfgd.
](https://github.com/sonic-net/sonic-host-services/blob/6fce4781da508709f11be595e3c7048ffd3138cb/scripts/hostcfgd#L1406) When we perform `config reload` because `hostcfgd` is restarted `resolv-config.service` is also restarted, but due to dependency of [`resolv-config.service` on `sonic.target`](https://github.com/sonic-net/sonic-buildimage/blob/ed839c65055c3cbb446ddabdc9c15b0684988139/files/image_config/resolv-config/resolv-config.service#L5)  `systemd` also tries to restart `resolv-config.service` causing cancellation of one of the instances of `resolv-config.service`. This fix removes the dependency of `resolv-config.service` on `sonic.target`

##### Work item tracking

#### How I did it
By removing the `BindsTo` and `WantedBy` dependency to `sonic.target` the service is not restarted upon `config reload` by `systemd` anymore, which would solve the issue of the service being cancelled.

#### How to verify it
No occurrence of `Job for resolv-config.service canceled.` in `syslog` after this change is considered

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/files/image_config/resolv-config/resolv-config.service
+++ b/files/image_config/resolv-config/resolv-config.service
@@ -2,7 +2,6 @@
 Description=Update DNS configuration
 Requires=config-setup.service
 After=config-setup.service
-BindsTo=sonic.target
 After=sonic.target
 StartLimitIntervalSec=0
 
--- a/files/image_config/resolv-config/resolv-config.service
+++ b/files/image_config/resolv-config/resolv-config.service
@@ -11,5 +11,3 @@ Type=oneshot
 RemainAfterExit=yes
 ExecStart=/usr/bin/resolv-config.sh start
 
-[Install]
-WantedBy=sonic.target


dgsudharsan (approved):

dgsudharsan:
@qiluo-msft can you please review and signoff?

ganglyu:
Can you add end to end test?

dgsudharsan:
> Can you add end to end test?

I don't think this change requires E2E test case as the existing test case for DNS should cover the functionality.

ganglyu:
> > Can you add end to end test?
> 
> I don't think this change requires E2E test case as the existing test case for DNS should cover the functionality.

But we didn't capture this issue with existing test case, can you explain?

dgsudharsan:
> > > Can you add end to end test?
> > 
> > 
> > I don't think this change requires E2E test case as the existing test case for DNS should cover the functionality.
> 
> But we didn't capture this issue with existing test case, can you explain?

We did capture this issue through log-analyzer error logs. It is a statistical bug and depends on timing.

dgsudharsan:
@ganglyu @qiluo-msft can you please signoff?

ganglyu (approved):

dprital:
@qiluo-msft , Can you please merge ?

qiluo-msft (approved):

PR #18684, opened 2024-04-15
[Mellanox] add rshim package to mellanox image

Yakiv-Huryk:
#### Why I did it
To add rshim (BlueField Rshim Host Driver) package to the mellanox image.

##### Work item tracking

#### How I did it
* Added rshim.mk to the Mellanox platform.
* Disabled rshim.service in sonic_debian_extension.j2

#### How to verify it
Build sonic-mellanox.bin

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]


files/build_templates/sonic_debian_extension.j2
```diff
@@ -1028,6 +1028,8 @@ sudo LANG=C chroot $FILESYSTEM_ROOT systemctl enable nv-syncd-shared
 
 # Install minicom package
 sudo LANG=C DEBIAN_FRONTEND=noninteractive chroot $FILESYSTEM_ROOT apt-get -y install minicom
+
+sudo LANG=C chroot $FILESYSTEM_ROOT systemctl disable rshim.service
```
keboliu, line 1032: Is it possible to have the service disabled by default when building it? I mean this line is not needed on other vendor's images, right?

sonic-slave-bookworm/Dockerfile.j2
```diff
@@ -417,6 +417,9 @@ RUN apt-get update && apt-get install -y \
 # For audisp-tacplus
         libauparse-dev \
         auditd \
+# For nvidia
```
keboliu, line 420: Is it possible to only have them installed on our platform?

files/build_templates/sonic_debian_extension.j2
Yakiv-Huryk: I'm not aware of the way to build disabled by default. Since we build/install it only for Mellanox, this line is only needed for Mellanox.

sonic-slave-bookworm/Dockerfile.j2
Yakiv-Huryk: The slave docker is not platform-specific. In this "apt-get install" we install the packages for all the platforms.

keboliu (approved):

files/build_templates/sonic_debian_extension.j2
prgeor: @Yakiv-Huryk why is this line of change needed? seems this service will only be needed for Mellanox so why do we need to explicity disable?

saiarcot895: @Yakiv-Huryk If `--no-enable` is passed into `dh_installsystemd` when the deb package is built, then it will get installed as disabled by default, and won't get automatically enabled.

saiarcot895 (approved):

liat-grozovik (approved):

PR #18715, opened 2024-04-18
Introduce a SONiC-specific communication channel between FRR and SONiC

cscarpitta:
#### Why I did it

Introduce a SONiC-specific communication channel between FRR and SONiC.
HLD: https://github.com/sonic-net/SONiC/pull/1620

#### How I did it

Introduce a SONiC-specific communication channel between FRR and SONiC (FPM SONiC module).

#### How to verify it

By switching to the new FPM SONiC module, all existing test cases for all features are successfully passed.

* Add new FPM SONiC module
* Add a patch file "build-dplane-fpm-sonic-module.patch" to compile FRR with the new FPM SONiC module
* Change supervisord.conf.j2 of the docker-fpm-frr container to load the new FPM SONiC module on FRR startup

### DIFF
--- a/platform/vs/docker-sonic-vs/supervisord.conf.j2
+++ b/platform/vs/docker-sonic-vs/supervisord.conf.j2
@@ -164,7 +164,7 @@ environment=ASAN_OPTIONS="log_path=/var/log/asan/teammgrd-asan.log{{ asan_extra_
 {% endif %}
 
 [program:zebra]
-command=/usr/lib/frr/zebra -A 127.0.0.1 -s 90000000 -M dplane_fpm_nl --asic-offload=notify_on_offload
+command=/usr/lib/frr/zebra -A 127.0.0.1 -s 90000000 -M dplane_fpm_sonic --asic-offload=notify_on_offload
 priority=13
 autostart=false
 autorestart=false
--- a/dockers/docker-fpm-frr/frr/supervisord/supervisord.conf.j2
+++ b/dockers/docker-fpm-frr/frr/supervisord/supervisord.conf.j2
@@ -30,7 +30,7 @@ stderr_logfile=syslog
 dependent_startup=true
 
 [program:zebra]
-command=/usr/lib/frr/zebra -A 127.0.0.1 -s 90000000 -M dplane_fpm_nl -M snmp --asic-offload=notify_on_offload
+command=/usr/lib/frr/zebra -A 127.0.0.1 -s 90000000 -M dplane_fpm_sonic -M snmp --asic-offload=notify_on_offload
 priority=4
 autostart=false
 autorestart=false
--- a/src/sonic-frr/Makefile
+++ b/src/sonic-frr/Makefile
@@ -6,6 +6,7 @@ MAIN_TARGET = $(FRR)
 DERIVED_TARGET = $(FRR_PYTHONTOOLS) $(FRR_DBG) $(FRR_SNMP) $(FRR_SNMP_DBG)
 SUFFIX = $(shell date +%Y%m%d\.%H%M%S)
 STG_BRANCH = stg_temp.$(SUFFIX)
+DPLANE_FPM_SONIC_MODULE = dplane_fpm_sonic/dplane_fpm_sonic.c
 
 # DEBEMAIL required by gpb dch
 export DEBEMAIL := sonicproject@googlegroups.com
@@ -20,6 +21,7 @@ endif
 	stg branch --create $(STG_BRANCH) $(FRR_TAG)
 	stg import -s ../patch/series
 	gbp dch --ignore-branch --new-version=$(FRR_VERSION)-sonic-$(FRR_SUBVERSION) --dch-opt="--force-bad-version" --commit --git-author
+	cp ../$(DPLANE_FPM_SONIC_MODULE) zebra/
 
 ifeq ($(CROSS_BUILD_ENVIRON), y)
 	CFLAGS="-I $$CROSS_PERL_CORE_PATH" dpkg-buildpackage -rfakeroot -b -d -us -uc -Ppkg.frr.nortrlib -a$(CONFIGURED_ARCH) -Pcross,nocheck -j$(SONIC_CONFIG_MAKE_JOBS) --admindir $(SONIC_DPKG_ADMINDIR)
--- a/src/sonic-frr/dplane_fpm_sonic/dplane_fpm_sonic.c
+++ b/src/sonic-frr/dplane_fpm_sonic/dplane_fpm_sonic.c
@@ -1,0 +1,2253 @@
+/*
+ * Zebra dataplane plugin for Forwarding Plane Manager (FPM) using netlink.
+ *
+ * Copyright (C) 2019 Network Device Education Foundation, Inc. ("NetDEF")
+ *                    Rafael Zalamena
+ *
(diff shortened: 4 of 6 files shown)


platform/vs/docker-sonic-vs/supervisord.conf.j2
```diff
@@ -164,7 +164,7 @@ environment=ASAN_OPTIONS="log_path=/var/log/asan/teammgrd-asan.log{{ asan_extra_
 {% endif %}
 
 [program:zebra]
-command=/usr/lib/frr/zebra -A 127.0.0.1 -s 90000000 -M dplane_fpm_nl --asic-offload=notify_on_offload
+command=/usr/lib/frr/zebra -A 127.0.0.1 -s 90000000 -M dplane_fpm_sonic --asic-offload=notify_on_offload
```
lguohan, line 167: it is better if we do not change it by default, it is better to first introduce this as option and then wait till it becomes stable.

cscarpitta: Hi @lguohan

Many thanks for the review.

I added a compile-time option `ENABLE_DPLANE_FPM_SONIC`.

When this option is set, SONiC uses the new FPM module.
When this option is not set, SONiC uses the old FPM module and behaves exactly as it does today.

By default, this option is not set.

lguohan: i am fine that we build them into image by default, but do not enable it by default. if you can do it at runtime, it will be easier for you, and i am fine with that.

does this dplane_fpm_sonic support all features as dplane_fpm_nl?

cscarpitta: @lguohan

Yes, the `dplane_fpm_sonic` supports all features as `dplane_fpm_nl`. It is a copy of the old `dplane_fpm_nl`.

Indeed, it is easier to compile this by default and keep the old `dplane_fpm_nl` running by default.
I reverted the compile option and changed the supervisord config file to keep `dplane_fpm_nl` as default.

lguohan:
@cscarpitta , have we sync with frr community to see if we can add this plugin into frr repo and get maintainer there? I think this can be facilitated by the routing group? @kperumalbfn , and @eddieruan-alibaba ?

if we cannot get into frr repo, for this plugin, can we maintain as a patch format to fpm_nl? so that we can still keep track of upstream changes. What if there is some new changes to dplane_fpm_nl and we need those changes into dplane_fpm_sonic?

eddieruan-alibaba:
> @cscarpitta , have we sync with frr community to see if we can add this plugin into frr repo and get maintainer there? I think this can be facilitated by the routing group? @kperumalbfn , and @eddieruan-alibaba ?
> 
> if we cannot get into frr repo, for this plugin, can we maintain as a patch format to fpm_nl? so that we can still keep track of upstream changes. What if there is some new changes to dplane_fpm_nl and we need those changes into dplane_fpm_sonic?

Discussed this with Donald Sharp in Routing WG. The conclusion was that Donald didn't want to take it into FRR community since it is SONiC specific. This proposal was proposed via Mark Stepp from Cisco who worked in FRR fpm before and supported by Donald who owns FRR. The discussion meeting minutes could be found in https://lists.sonicfoundation.dev/g/sonic-wg-routing/wiki/36598. 

The basic idea is to let SONiC team own sonic fpm. The zebra/fpm apis have already been changed with versioning by Mark S. Since we own sonic fpm, I feel it is simple to keep dplane_fpm_sonic.c as a file instead of as a diff.

cscarpitta:
> @cscarpitta , have we sync with frr community to see if we can add this plugin into frr repo and get maintainer there? I think this can be facilitated by the routing group? @kperumalbfn , and @eddieruan-alibaba ?
> 
> if we cannot get into frr repo, for this plugin, can we maintain as a patch format to fpm_nl? so that we can still keep track of upstream changes. What if there is some new changes to dplane_fpm_nl and we need those changes into dplane_fpm_sonic?

@lguohan As Eddie mentioned, the option of adding the new changes as a patch was discussed in the Routing WG. However there were some concerns on this approach and the consensus was that this will be maintained as a separate module under the `sonic-buildimage` repository

Regarding the second point, in case of new features or bug fixes or improvements of `dplane_fpm_nl`, as part of the FRR upgrade process we will make sure that all the changes will be reflected in the `dplane_fpm_sonic`.

eddieruan-alibaba:
For the platform which wants to use dplane_fpm_sonic.so, how does this .so get picked up? Do you want to add a README file for that?

cscarpitta:
> For the platform which wants to use dplane_fpm_sonic.so, how does this .so get picked up? Do you want to add a README file for that?

Hi @eddieruan-alibaba Many thanks for the review. 

I added a README file to explain how to enable the `dplane_fpm_sonic` module.

cscarpitta:
@liushilongbuaa @StormLiangMS Can you please help fix the ms_conflict?

cscarpitta:
@lguohan 

The HLD for this feature has been merged: https://github.com/sonic-net/SONiC/pull/1620

We addressed all comments on this PR. Can you please take a look again and help merge?

eddieruan-alibaba (approved):

eddieruan-alibaba:
> > For the platform which wants to use dplane_fpm_sonic.so, how does this .so get picked up? Do you want to add a README file for that?
> 
> Hi @eddieruan-alibaba Many thanks for the review.
> 
> I added a README file to explain how to enable the `dplane_fpm_sonic` module.

Thanks Carmine. The readme explains the changes needed. As part of Phoenix wing project, we could change the .j2 based on the supported platform type.

eddieruan-alibaba:
@lguohan currently, he just commit the changes needed as infra. It would not be enabled on the current platforms. As part of phoenix wing project, we could enable them on the platforms verified in that project.

ysj-alibaba:
@cscarpitta I don't understand what does DPLAINE_OP_STARTUP_STAGE mean and it seems that there is no definition about this flag. Thanks.

eddieruan-alibaba:
This PR's changes are cherry picked to Phoenixwing Repro and get validated.

kperumalbfn:
@cscarpitta could you check the buildimage failures and VS test failurs?

kperumalbfn (approved):
LGTM. Will merge after resolving the conflict.

kperumalbfn:
@liushilongbuaa Could you check the conflict and trigger ms_conflict check

ahsalam:
@kperumalbfn could you please check again? the PR passed all tests. Thanks!

sudhanshukumar22:
Hi, is it possible to know that which of dplane_fpm_sonic or dplane_fpm_nl is running in the system loaded with sonic image ?

sudhanshukumar22:
root@sonic:/# ps -aux | grep zebra
frr           39  0.0  0.3 647216 24488 pts/0    Sl   08:42   0:00 /usr/lib/frr/zebra -A 127.0.0.1 -s 90000000 -M dplane_fpm_sonic -M snmp --asic-offload=notify_on_offload
root         299  0.0  0.0   3324  1388 pts/1    S+   09:05   0:00 grep zebra
Does this mean that dplane_fpm_sonic  is running in the system ?

ahsalam:
@sudhanshukumar22 dplane_fpm_sonic is running in the system.

PR #18727, opened 2024-04-19
[bgp] Add support to explicitly specify bgp router id for single asice device

yaqiangz:
#### Why I did it
HLD: https://github.com/sonic-net/SONiC/pull/1643
Add support to explictly specify bgp router id for single asic device

##### Work item tracking
- Microsoft ADO **(number only)**: 27674485

#### How I did it
1. When bgp_router_id configured in DEVICE_METADATA, use it as bgp router-id.
2. Remove the hard dependency on loopback0 ipv4 address when adding bgp peer.
3. Add UTs.

**Behavior of set bgp router-id**

To be clarified that when bgp router-id hasn't been explicitly set, bgp actions would totally like previous

|           | Loopback0 IPv4 address exists | Loopback0 IPv4 address doesn't exist |
|--------------|-------|------------|
| bgp_router_id configured | Honor bgp_router_id | Honor bgp_router_id |
| bgp_router_id doesn't be configured | Honor Loopback0 IPv4 address | FRR default router ID value is selected as the largest IP address of the device. When router zebra is not enabled bgpd can’t get interface information, so router-id is set to 0.0.0.0 |

**Behavior of add bgp peer**

To be clarified that when bgp router-id hasn't been explicitly set, bgp actions would totally like previous

|          | Loopback0 IPv4 address exists | Loopback0 IPv4 address doesn't exist |
|--------------|-------|------------|
| bgp_router_id configured | Add BGP peer | Add BGP peer |
| bgp_router_id doesn't be configured | Add BGP peer | Do not add BGP peer |

#### How to verify it
UT passed.
Run bgp related test in sonic-mgmt

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/src/sonic-bgpcfgd/bgpcfgd/managers_bgp.py
+++ b/src/sonic-bgpcfgd/bgpcfgd/managers_bgp.py
@@ -106,7 +106,6 @@ class BGPPeerMgrBase(Manager):
 
         deps = [
             ("CONFIG_DB", swsscommon.CFG_DEVICE_METADATA_TABLE_NAME, "localhost/bgp_asn"),
-            ("CONFIG_DB", swsscommon.CFG_LOOPBACK_INTERFACE_TABLE_NAME, "Loopback0"),
             ("CONFIG_DB", swsscommon.CFG_BGP_DEVICE_GLOBAL_TABLE_NAME, "tsa_enabled"),
             ("LOCAL", "local_addresses", ""),
             ("LOCAL", "interfaces", ""),
@@ -168,9 +167,11 @@ class BGPPeerMgrBase(Manager):
         bgp_asn = self.directory.get_slot("CONFIG_DB", swsscommon.CFG_DEVICE_METADATA_TABLE_NAME)["localhost"]["bgp_asn"]
         #
         lo0_ipv4 = self.get_lo_ipv4("Loopback0|")
-        if lo0_ipv4 is None:
+        if (lo0_ipv4 is None and "bgp_router_id"
+            not in self.directory.get_slot("CONFIG_DB", swsscommon.CFG_DEVICE_METADATA_TABLE_NAME)["localhost"]):
             log_warn("Loopback0 ipv4 address is not presented yet")
             return False
+
         #
         if self.peer_type == 'internal':
             lo4096_ipv4 = self.get_lo_ipv4("Loopback4096|")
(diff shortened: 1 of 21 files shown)


yaqiangz:
Hi @StormLiangMS Could you please help to review this PR?

StormLiangMS:
1. first one is a question, what's the issue this design is to address? To avoid bgp breakage due to the loopback0 ip is not configured? What's the use case? @yaqiangz Why we assume loopback0 ip is not configured, but bgp_router_id is configured?

StormLiangMS:
In this PR, no multi asic, it would be a separate PR?

src/sonic-bgpcfgd/bgpcfgd/managers_bgp.py
```diff
@@ -168,9 +167,11 @@ def add_peer(self, vrf, nbr, data):
         bgp_asn = self.directory.get_slot("CONFIG_DB", swsscommon.CFG_DEVICE_METADATA_TABLE_NAME)["localhost"]["bgp_asn"]
         #
         lo0_ipv4 = self.get_lo_ipv4("Loopback0|")
-        if lo0_ipv4 is None:
```
StormLiangMS, line 171: is that possible the loopback0 interface comes after bgpcfgd start? in that case, if we don't wait for this loopback interface up, it can't get the lo0 ip.

yaqiangz:
> In this PR, no multi asic, it would be a separate PR?

Yes

src/sonic-bgpcfgd/bgpcfgd/managers_bgp.py
yaqiangz: It get lo0 ip from config_db https://github.com/sonic-net/sonic-buildimage/blob/3db62a0c6f064cc1e53ecf5f223d5165010dc175/src/sonic-bgpcfgd/bgpcfgd/managers_bgp.py#L324
I don't think there is a scenario that we want loopback0 ipv4 but it is not shown in config_db

StormLiangMS: hi @yaqiangz I mean is there a race condition between bgp startup and the population of loopback config to the config_db? If we remove the dependency, and bgp startup before the loopback config populated into the config_db could cause issue.

yaqiangz: If I understand correctly, your concern is table subscription for interest table changed? The table change subscription is defined in [main.py](https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-bgpcfgd/bgpcfgd/main.py#L49) (DEVICE_METADATA and LOOPBACK_INTERFACE are both subscribed). We didn't change this part. The dependency is a condition for **whether to invoke handler function when subscription table changes (Checked by func `self.directory.available_deps`) [manager.py](https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-bgpcfgd/bgpcfgd/manager.py#L41)**. We removed this dependency is to **enable add bgp peer when Loopback v4 address doesn't exist**

yaqiangz:
> 1. first one is a question, what's the issue this design is to address? To avoid bgp breakage due to the loopback0 ip is not configured? What's the use case? @yaqiangz Why we assume loopback0 ip is not configured, but bgp_router_id is configured?

As HLD said, we want to remove the strong coupling of configuring router id and Loopback ipv4 address.

StormLiangMS (approved):
LGTM

PR #18785, opened 2024-04-24
Change global arp ignore level from "check ip within same subnet" to "all reply"

irene-pan1202:
#### Why I did it
arp_ignore has different modes for sending replies in response to received ARP requests that resolve the local target IP address.
Due to the strict settings of the arp_ignore global configuration, there are more restrictions.

##### Work item tracking
support a new field that allows dynamic modification of arp_ignore for different interfaces.
https://github.com/sonic-net/sonic-swss/issues/3122

#### How I did it
Change global arp ignore level from "check ip within same subnet" to "all reply".
#### How to verify it
```
admin@sonic:/$ sudo sysctl -a | grep all.arp_ignore
net.ipv4.conf.all.arp_ignore = 0
```

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]


prsunny:
Can you please share the motivation of this PR?

files/image_config/sysctl/sysctl-net.conf
```diff
@@ -14,7 +14,6 @@ net.ipv4.conf.all.arp_accept=0
 net.ipv4.conf.all.arp_announce=1
 net.ipv4.conf.all.arp_filter=0
 net.ipv4.conf.all.arp_notify=1
-net.ipv4.conf.all.arp_ignore=2
```
prsunny, line 17: Do not change default values that break backward compatibility.

irene-pan1202: According to linux documentation, https://github.com/torvalds/linux/blob/master/Documentation/networking/ip-sysctl.rst  
`The max value from conf/{all,interface}/arp_ignore is used when ARP request is received on the {interface}`

arp_ignore will take the max value, therefore all interfaces will use the global value. 
If this setting could be changed to be controlled via the interface, it would provide users with more flexibility.
Therefore, remove the global value and change the default arp_ignore setting of 2 to be configured per interface.
Related PR: [sonic-net/sonic-swss#3122](https://github.com/sonic-net/sonic-swss/pull/3122)

PR #18788, opened 2024-04-24
Modification for kdump for remote server dump

bilal-ismail:
#### Why I did it
For pre-boot initialization of network interfaces.
For my case, I had to test and implement the remote kdump file transfer using SSH and the crash kernel did not load configs from /etc/network/interfaces config file.

#### How I did it
1) New file created for a custom script, "network-interface-preboot-init"
2) Modified the "build_debian.sh" file for adding this to the initramfs-tools/scripts/ directory.

#### How to verify it
After this change the remote SSH kdump file transfer would be possible because network not reachable issue arose as the default "vmliuz" image did not change state of the interfaces to "UP" on crash kernel load. Neither did it load configs from /etc/network/interfaces config file.


bilal-ismail:
Build failed due to library versioning that i had not made any changes to. Will try to resolve this issue.

PR #18817, opened 2024-04-28
[System ready] Extend sysmonitor functionality to wait for host daemons

fastiuk:
#### Why I did it
To extend sysmonitor to report services states according to HLD: https://github.com/sonic-net/SONiC/pull/1363

#### How I did it
* Add timeout monitor task to sysmonitor to declare system ready state forcibly when timout happened.
* sysmonitor now has a state machine, so the status now doesn't get back to "Not OK" after it was declared as "OK"
* A new "irrel_for_sysready" configuration was added to "FEATURE" table to handle services for which sysmonitor shouldn't wait for.
* Add "wait_services" field to platform health config file to explicitly specify services sysmonitor has to wait for.

#### How to verify it

* Start NOS on the box
* Make sure it defines system ready status
* Make sure `sysreadyshow` shows all services as OK 

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ] 

[Device metadata config](https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-yang-models/doc/Configuration.md#device-metadata)

#### A picture of a cute animal (it is my cat Finn)
PXL_20230413_140941610 PORTRAIT


files/build_templates/docker_image_ctl.j2
```diff
@@ -108,6 +108,37 @@ function preStartAction()
     fi
 {%- elif docker_container_name == "snmp" %}
     $SONIC_DB_CLI STATE_DB HSET 'DEVICE_METADATA|localhost' chassis_serial_number $(decode-syseeprom -s)
+
+    # Wait for system ready. Continue after timeout if we won't hear anything
+    # from DB. Read timeout from config file and add two extra minutes on top of
+    # it.
+    CONFFILE=system_health_monitoring_config.json
+    PLATFORM=$(sonic-cfggen -d -v "DEVICE_METADATA['localhost']['platform']")
```
fastiuk, line 116: > Try to use cache rather than using sonic-cfggen as sonic-cfggen is a costly operation.

> Please refer to https://github.com/sonic-net/sonic-buildimage/pull/17343

@dgsudharsan ok, will change that

files/build_templates/docker_image_ctl.j2
```diff
@@ -108,6 +108,37 @@ function preStartAction()
...
+
+    # Set default if not found
+    if [[ -z $TIMEOUT ]]; then
+        MESSAGE="Failed to read timeout from config file, assuming default is \
+                10 minutes"
+        echo $MESSAGE
+        TIMEOUT=10
+    fi
+
+    # Add to extra minutes and convert to seconds and
```
fastiuk, line 129: > What is the rationale behind adding extra two minutes?

@dgsudharsan The system ready status should be declared in `TIMEOUT` time, but I am adding two extra minutes just to poll it for a bit longer time to make sure it will overlap timeout time.

files/build_templates/docker_image_ctl.j2
```diff
@@ -108,6 +108,37 @@ function preStartAction()
     fi
 {%- elif docker_container_name == "snmp" %}
     $SONIC_DB_CLI STATE_DB HSET 'DEVICE_METADATA|localhost' chassis_serial_number $(decode-syseeprom -s)
+
```
fastiuk, line 111: > Are you enabling this feature by default for snmp? Shouldn't it be based on configuration?

@dgsudharsan It is disabled for snmp by default: see `files/build_templates/init_cfg.json.j2`

files/build_templates/init_cfg.json.j2
```diff
@@ -79,11 +79,17 @@
...
+            "check_up_status" : "true",
+{%- endif %}
 {# For now, to support the infrastrucure, setting the check_up_status to false for bgp,swss,pmon  #}
 {# Once apps like bgp,synd supports app readiness, then bgp,syncd can set check_up_status to true #}
-{%- if feature in ["bgp", "swss", "pmon"] %}
+{%- if feature in ["bgp"] %}
             "check_up_status" : "false",
 {%- endif %}
+{%- if feature in ["ib-utils", "snmp"] %}
+            "irrel_for_sysready" : "true",
```
fastiuk, line 91: > Can you please rename it to ignore_for_sysready for better readability?

@dgsudharsan sorry, cannot. Cause it was approved in the HLD like that.

files/scripts/syncd.sh
```diff
@@ -44,6 +44,8 @@ function startplatform() {
         /usr/bin/mlnx-fw-upgrade.sh -v
         if [[ "$?" -ne "${EXIT_SUCCESS}" ]]; then
             debug "Failed to upgrade fw. " "$?" "Restart syncd"
+            sonic-db-cli STATE_DB HSET "FEATURE|$DEV_SRV" fail_reason \
```
fastiuk, line 47: > Why are we considering just the asic firmware update as sysready indication for syncd. Shouldn't it be the create switch success?

@dgsudharsan not only, we have `PortInitDone` event in swss and `SFP ready` in pmon.

src/system-health/health_checker/sysmonitor.py
```diff
@@ -1,30 +1,36 @@
...
+from . import utils
 from .config import Config
 import signal
 import jinja2
 
 SYSLOG_IDENTIFIER = "system#monitor"
 REDIS_TIMEOUT_MS = 0
-system_allsrv_state = "DOWN"
-spl_srv_list = ['database-chassis', 'gbsyncd']
+spl_srv_list = ['database-chassis', 'gbsyncd', 'e2scrub_reap']
```
fastiuk, line 23: > Can you clarify what is e2scrub_reap?

@dgsudharsan it is debian internal service. According to the manual: `e2scrub attempts to check (but not repair) all metadata in a mounted ext[234] file system if the file system resides on an LVM logical volume. The block device of the LVM logical volume can also be passed in.` So we shouldn't take it into concideration.

src/system-health/health_checker/sysmonitor.py
```diff
@@ -109,17 +128,72 @@ def task_notify(self, msg):
             return
         self.task_queue.put(msg)
 
+
+# Subprocess to monitor system ready timeout. If timeout will be exceeded,
+# send a message to queue and exit
+class MonitorTimeout(ProcessTaskBase):
```
fastiuk, line 134: > This thread needs to be spawned only when feature is enabled.

@dgsudharsan not only, it should be spawned also when feature is disabled, cause when it is disabled it still wait for one state - swss's `PortInitDone`

files/build_templates/docker_image_ctl.j2
dgsudharsan: My question is different.  In the init_cfg.j2 you are considering snmp as irrelevent for system ready but using system ready as a criteria to spawn snmp. Is this correct?

files/build_templates/docker_image_ctl.j2
dgsudharsan: Since you are polling every 15 seconds, shouldn't 1 minute be sufficient to handle overlap?

files/build_templates/docker_image_ctl.j2
dgsudharsan: Can you consider handling this logic through featured? Today we have a similar flow in featured where certain daemons are spawned after portinitdone. In this case the critera can be systemd. It can be much easy to extend this to multiple daemons

files/scripts/syncd.sh
dgsudharsan: My question is syncd up_status is just set base on startplatform completion. Is it possible that in the actual syncd flow, switch_create might fail and we will still see the up_status as true. Can we include the switch create in the definition of syncd up_status?

files/build_templates/docker_image_ctl.j2
fastiuk: > My question is different. In the init_cfg.j2 you are considering snmp as irrelevent for system ready but using system ready as a criteria to spawn snmp. Is this correct?

Yes, it is indeed correct. Irrelevant for system ready means that `system ready` feature won't just collect readiness status of snmp systemd service. But SNMP must be started after system is ready. The logic actually does't break each other.

fastiuk: > Can you consider handling this logic through featured? Today we have a similar flow in featured where certain daemons are spawned after portinitdone. In this case the critera can be systemd. It can be much easy to extend this to multiple daemons

I looked at the code in `featured` and I don't want to break anything there by adding wait for system ready. Most of the `delayed` services there depend on portinit done, and I don't want to add another wait point there as it will just make the logic there more complex and less clear.

files/scripts/syncd.sh
fastiuk: Yes, it can happen I think. Set system ready condition you see here is just an internal NV requirement, but at the same time it is very generic, so won't conflict with any other platform. I prefer to leave it as it is.

files/build_templates/init_cfg.json.j2
```diff
@@ -82,11 +82,17 @@
             "auto_restart": "{{autorestart}}",
             "support_syslog_rate_limit" : "true",
 {# Set check_up_status to true here when app readiness will be marked in state db  #}
+{%- if feature in ["swss", "syncd", "pmon"] %}
+            "check_up_status" : "true",
```
qiluo-msft, line 86: Please do not change image config generation behavior. We will rely on external service to generate golden config, and ask sonic to `config reload -f golden_config.json` to consume them.

qiluo-msft: @wen587 to review.

liat-grozovik:
@fastiuk i hope that you can handle conflict and we will be able to merge it finally before branch out this week.

PR #15734, opened 2023-07-06
[chassis] removed dependency between bgp and swss for chassis supervisor

arlakshm:
#### Why I did it
Fixes https://github.com/sonic-net/sonic-buildimage/issues/15667 and https://github.com/sonic-net/sonic-buildimage/issues/13293
##### Work item tracking
- Microsoft ADO **24472854**:

#### How I did it
On chassis supervisor `bgp` feature is disabled in hostcfgd. The dependency between `swss` and `bgp` causes the bgp containers to start even though the feature is disabled. 

#### How to verify it
Tests on chassis supervisor and LC

#### Which release branch to backport (provide reason below if selected)

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

#### Tested branch (Please provide the tested image version)

- [ ] 
- [ ]

PR #18829, opened 2024-04-30
[voq][chassis][dhcp_relay] swss.sh try to start the dhcp_relay service although it is masked

mlok-nokia:
#### Why I did it
on Master branch, dhcp_relay is not supported in VOQ chassis. It is disabled in the FEATURE table. But based on the dependency, swss.sh always call "systemctl start" it although it's service file has been masked/disabled. The following error is logged in syslog which causes the logAnalyze failed on some of the OC tests.  Fix issue https://github.com/sonic-net/sonic-buildimage/issues/18822
##### Work item tracking

#### How I did it
Added code to swss.sh to check if service is disabled or not.  If it is disabled, do not start the service although it is in the DPENDENT_LIST.  This avoids the ERROR log shown up in the syslog file

#### How to verify it
1)  Reboot the system or execut the config reload.  The following error messages should not be seen in the syslog file
```
Apr 27 23:15:14.472425 ixre-egl-board7 ERR systemctl[7299]: Failed to start dhcp_relay.service: Unit dhcp_relay.service is masked.
Apr 27 23:15:14.476897 ixre-egl-board7 ERR systemctl[7298]: Failed to start dhcp_relay.service: Unit dhcp_relay.service is masked.
```

This issue only exist in Master branch which is using the newer version of kernel.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305
- [x] 202405

tested on Master branch.

- [ ] 
- [ ]

### DIFF
--- a/files/scripts/swss.sh
+++ b/files/scripts/swss.sh
@@ -242,6 +242,17 @@ function clean_up_chassis_db_tables()
 
 }
 
+is_feature_enabled()
+{
+    s=$1
+    service=${s%@*}
+    state=$(sonic-db-cli CONFIG_DB hget "FEATURE|${service}" "state")
+    if [[ $state == "enabled" ]]; then
+        echo "true"
+    else
+        echo "false"
+    fi
+}
 start_peer_and_dependent_services() {
     check_warm_boot
 
@@ -254,7 +265,10 @@ start_peer_and_dependent_services() {
             fi
         done
         for dep in ${DEPENDENT}; do
-            /bin/systemctl start ${dep}
+            state=$(is_feature_enabled $dep)
+            if [[ $state == "true" ]]; then
+                /bin/systemctl start ${dep}
+            fi
         done
         for dep in ${MULTI_INST_DEPENDENT}; do
             if [[ ! -z $DEV ]]; then
--- a/files/scripts/swss.sh
+++ b/files/scripts/swss.sh
@@ -265,8 +265,12 @@ start_peer_and_dependent_services() {
             fi
         done
         for dep in ${DEPENDENT}; do
-            state=$(is_feature_enabled $dep)
-            if [[ $state == "true" ]]; then
+            if [[ $dep == "dhcp_relay" ]]; then
+                state=$(is_feature_enabled $dep)
+                if [[ $state == "true" ]]; then
+                    /bin/systemctl start ${dep}
+                fi
+            else
                 /bin/systemctl start ${dep}
             fi
         done


rlhui:
@kellyyeh  please review this at earliest, thanks.

wenyiz2021 (approved):
would like to keep an eye on this PR.

we have seen similar issue where sup has bgp disabled but not masked, and swss startup also starts bgp which caused an issue. there is https://github.com/sonic-net/sonic-buildimage/pull/15734 to fix that.

this PR LGTM as a more general fix.

mlok-nokia:
@lguohan @kellyyeh  Please help to review and merge this PR. Thanks

judyjoseph (dismissed):
LGTM, the only case is if the service is delayed and hostcfg sets the FEATURE state to enable/disable later -- which seems not the case with dhcp_relay service.

kellyyeh (approved):

wenyiz2021:
hi @lguohan / @yxieca , could you help merge?

yxieca:
> hi @lguohan / @yxieca , could you help merge?

This change is generally in the right direction. I am surprised that it only affected dhcp relay. @qiluo-msft , @saiarcot895  do you have other concerns for the change?

wenyiz2021:
> we have seen similar issue where sup has bgp disabled but not masked, and swss startup also starts bgp which caused an issue. there is #15734 to fix that.

we have seen similar issue where sup has bgp disabled but not masked, and swss startup also starts bgp which caused an issue. there was https://github.com/sonic-net/sonic-buildimage/pull/15734 to fix that.
current PR fix looks more general

wenyiz2021:
> > hi @lguohan / @yxieca , could you help merge?
> 
> This change is generally in the right direction. I am surprised that it only affected dhcp relay. @qiluo-msft , @saiarcot895 do you have other concerns for the change?

we have seen similar issue where sup has bgp disabled but not masked, and swss startup also starts bgp which caused an issue. there was https://github.com/sonic-net/sonic-buildimage/pull/15734 to fix that.
current PR fix looks more general

saiarcot895:
For swss, there's also the `docker-wait-any` script that gets started as part of the `wait` command (this basically checks to see if one of the dependent containers have exited; if so, it brings down this container). That might need to be modified as well hered/

files/scripts/swss.sh
```diff
@@ -242,6 +242,16 @@ function clean_up_chassis_db_tables()
 
 }
 
+is_feature_enabled()
+{
+    service=$1
```
wumiaont, line 247: There's an issue with this code. the $1 is passed in from DEPENDENT list. This could be something like this macsec@0. Needs to parse out the dev part. 
recommend code like this:  service=$1 changes to:
s=$1
service=${s%&*}
This will make config_db query to get the proper record.

mlok-nokia:
> For swss, there's also the `docker-wait-any` script that gets started as part of the `wait` command (this basically checks to see if one of the dependent containers have exited; if so, it brings down this container). That might need to be modified as well here?

For this particular issue, it only happens on the variable "DEPENDENT".  The docker-wait-any is using a variable "MULTI_INST_DEPENDENT".  No need to modify for docker-wait-any.

yaqiangz (approved):
LGTM

rlhui:
@prabhataravind would you be able to review/approve, so that this PR could be merged? thanks.

files/scripts/swss.sh
```diff
@@ -254,7 +265,10 @@ start_peer_and_dependent_services() {
             fi
         done
         for dep in ${DEPENDENT}; do
-            /bin/systemctl start ${dep}
+            state=$(is_feature_enabled $dep)
```
prabhataravind, line 268: @mlok-nokia, are there any dependent systemd services that are not tied to a feature? I guess most are but just wondering if you have confirmed it? CONFIG_DB hget would not return a valid state then and we will never start that service. Is that possible?

mlok-nokia: For the case which you mentioned about, as what we know, the change is good for now. But we could have a service which it is not associated with feature table.  Since this change is to address the dhcp-relay service which has this special dependency with swss,  I will modify the code to only check for the dhcp-relay service to avoid other impact.

prabhataravind (approved):
The latest change looks good to me.

prabhataravind:
> @prabhataravind would you be able to review/approve, so that this PR could be merged? thanks.

Reviewed and approved the latest diff.

judyjoseph (approved):

files/scripts/swss.sh
```diff
@@ -254,7 +265,14 @@ start_peer_and_dependent_services() {
             fi
         done
         for dep in ${DEPENDENT}; do
-            /bin/systemctl start ${dep}
+            if [[ $dep == "dhcp_relay" ]]; then
+                state=$(is_feature_enabled $dep)
+                if [[ $state == "true" ]]; then
```
kperumalbfn, line 270: After starting the service, should we enable the feature flag for this service?

yaqiangz:
Hi @yxieca are we good to merge this PR?

PR #18835, opened 2024-04-30
Close rsyslog plugin when rsyslog SIGTERM and EOF is sent to stream

zbud-msft:
##### Work item tracking
- Microsoft ADO **(number only)**:27882794

#### How I did it

Add signalOnClose for omprog as well as close rsyslog plugin when receives an EOF.

#### How to verify it

UT (will add sonic-mgmt testcase for storming events with logs)

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/src/sonic-eventd/rsyslog_plugin/rsyslog_plugin.cpp
+++ b/src/sonic-eventd/rsyslog_plugin/rsyslog_plugin.cpp
@@ -106,9 +106,8 @@ bool RsyslogPlugin::createRegexList() {
 void RsyslogPlugin::run() {
     lua_State* luaState = luaL_newstate();
     luaL_openlibs(luaState);
-    while(true) {
-        string line;
-        getline(cin, line);
+    string line;
+    while(getline(cin, line)) {
         if(line.empty()) {
             continue;
         }
--- a/files/build_templates/rsyslog_plugin.conf.j2
+++ b/files/build_templates/rsyslog_plugin.conf.j2
@@ -12,6 +12,7 @@ if re_match($programname, "{{ proc.name }}") then {
     action(type="omprog"
         binary="/usr/bin/rsyslog_plugin -r /etc/rsyslog.d/{{ proc.parse_json }} -m {{ yang_module }}"
         output="/var/log/rsyslog_plugin.log"
+        signalOnClose="SIGTERM"
         template="prog_msg")
 }
 {% endfor %}
--- a/src/sonic-eventd/rsyslog_plugin_tests/rsyslog_plugin_ut.cpp
+++ b/src/sonic-eventd/rsyslog_plugin_tests/rsyslog_plugin_ut.cpp
@@ -253,6 +253,16 @@ TEST(rsyslog_plugin, onMessage_noParams) {
     infile.close();
 }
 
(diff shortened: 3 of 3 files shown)


lguohan:
can you explain why?

keboliu:
@zbud-msft could you please check why the checker is skipped? would you please re-trigger it?

src/sonic-eventd/rsyslog_plugin/rsyslog_plugin.cpp
```diff
@@ -106,9 +106,8 @@ bool RsyslogPlugin::createRegexList() {
 void RsyslogPlugin::run() {
     lua_State* luaState = luaL_newstate();
     luaL_openlibs(luaState);
-    while(true) {
-        string line;
-        getline(cin, line);
+    string line;
+    while(getline(cin, line)) {
```
vivekrnv, line 110: I'd recommend adding a signal handler in rsyslog_plugin. something in the lines of 

```
trap SIGTERM signal_handler

void signal_handler(){
        stop_processing = true;
}

while(!stop_processing && !getline(cin,buf)){
       eventd logic.....
}
```

rsyslog dameon first sends a SIGTERM which would cause the program to terminate immediately. Meanwhile the plugin might be processing something it read just before from cin. So, to let it finish the pending task i'd suggest this approach

vivekrnv (changes requested):

src/sonic-eventd/rsyslog_plugin/rsyslog_plugin.cpp
zbud-msft: Added signal handler.

vivekrnv (approved):

qiluo-msft (approved):

zbud-msft:
@lguohan This change is to address 2 issues

1) When rsyslogd is terminated, no signal is sent to child process of rsyslog_plugin meaning that rsyslog_plugin will be constantly trying to read from cin with no writer on the other end of the pipe. This leads to rsyslog_plugin process will constantly be reading via getline infinitely.

2) Because rsyslog is terminated and the spawned rsyslog_plugin is still alive, when rsyslog starts backup again, and log is triggered, a new rsyslog_plugin will be spawned for that rsyslog process, which can lead to many "ghost" rsyslog_plugin processes that will be at high CPU usage.

vivekrnv:
Fixes https://github.com/sonic-net/sonic-buildimage/issues/18771

bingwang-ms:
Hi @zbud-msft, the change only handles `SIG_TERM`, but `EOF` of pipe is not handled yet. If `rsyslog` is killed by `SIG_KILL` or crash unexpectedly, the busy look is still there.

zbud-msft:
Discussed with Bing offline. From the code we are handling EOF. In loop we are checking getline(cin, line). If EOF is sent, getline(cin, line) will evaluate to false. UT also shows that we are replacing cin buffer with empty input stream, which will be treated as EOF when getline() is called by plugin->run.

PR #18846, opened 2024-05-01
[202311] Fix for DNS service being canceled 

gpunathilell:
Cherry Pick of #18670 
#### Why I did it
To fix `resolv-config.service` being cancelled:
`resolv-config.service` is started from `hostcfgd` each time there is an [update to the dns configuration or during the start of hostcfgd.
](https://github.com/sonic-net/sonic-host-services/blob/6fce4781da508709f11be595e3c7048ffd3138cb/scripts/hostcfgd#L1406) When we perform `config reload` because `hostcfgd` is restarted `resolv-config.service` is also restarted, but due to dependency of [`resolv-config.service` on `sonic.target`](https://github.com/sonic-net/sonic-buildimage/blob/ed839c65055c3cbb446ddabdc9c15b0684988139/files/image_config/resolv-config/resolv-config.service#L5)  `systemd` also tries to restart `resolv-config.service` causing cancellation of one of the instances of `resolv-config.service`. This fix removes the dependency of `resolv-config.service` on `sonic.target` and prevents double initialization of `resolv-config.service` since `hostcfgd` calls `resolv-config.service` during it's initialization so the initial configuration will not be missed after removing the dependency

##### Work item tracking

#### How I did it
By removing the `BindsTo` and `WantedBy` dependency to `sonic.target` the service is not restarted upon `config reload` by `systemd` anymore, which would solve the issue of the service being cancelled.

#### How to verify it
No occurrence of `Job for resolv-config.service canceled.` in `syslog` after this change is considered

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/files/image_config/resolv-config/resolv-config.service
+++ b/files/image_config/resolv-config/resolv-config.service
@@ -2,7 +2,6 @@
 Description=Update DNS configuration
 Requires=updategraph.service
 After=updategraph.service
-BindsTo=sonic.target
 After=sonic.target
 StartLimitIntervalSec=0
 
--- a/files/image_config/resolv-config/resolv-config.service
+++ b/files/image_config/resolv-config/resolv-config.service
@@ -11,5 +11,3 @@ Type=oneshot
 RemainAfterExit=yes
 ExecStart=/usr/bin/resolv-config.sh start
 
-[Install]
-WantedBy=sonic.target


dgsudharsan (approved):

keboliu:
@yxieca would you please help to approve and merge this to 202311? This is a backport bug fix PR.

liat-grozovik:
@yxieca please help to merge

ganglyu (approved):

yxieca:
@gpunathilell what is the consequence of missing the initial resolve config?

dgsudharsan:
> @gpunathilell what is the consequence of missing the initial resolve config?

Hi @yxieca I don't think the configuration will be missed as hostcfgd will process this during its initialization. This flow avoids double initialization as being tied to sonic.target will invoke it once followed by hostcfgd invoking it again.

PR #18850, opened 2024-05-02
[Nokia-7215] Enhance Watchdog service

Pavan-Nokia:
Mask Watchdog-control.service and make sure only one watchdog service starts on this platform

#### Why I did it
To resolve conflict between ordering of the 2 watchdog services.
1) watchdog-control.service -- common service designed to disable watchdog on all platforms
2) cpu_wdt.service -- enable Watchdog on nokia-7215 platform.

Is some cases service 1 was started after service 2 leaving the watchdog on the box disabled

##### Work item tracking

#### How I did it
Enhance service file to assure cpu_wdt.service always starts after watchdog-control.service

#### How to verify it
1) Try multiple upgrades and install scenario to make sure watchdog is always in enabled state 
2) Try multiple reboots to make sure watchdog is always in enabled state 

```
admin@sonic:~$ systemctl status watchdog-control.service 
â— watchdog-control.service
     Loaded: masked (Reason: Unit watchdog-control.service is masked.)
     Active: inactive (dead)
admin@sonic:~$ 
admin@sonic:~$ 
admin@sonic:~$ systemctl status cpu_wdt.service
â— cpu_wdt.service - CPU WDT
     Loaded: loaded (/etc/systemd/system/cpu_wdt.service; enabled; vendor prese>
     Active: active (running) since Mon 2024-05-20 14:57:15 UTC; 4min 57s ago
   Main PID: 635 (cpu_wdt.py)
      Tasks: 1 (limit: 4915)
     Memory: 13.9M
     CGroup: /system.slice/cpu_wdt.service
             â””â”€635 /usr/bin/python /usr/local/bin/cpu_wdt.py

admin@sonic:~$ 
```

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305
- [x] 202405

- [ ] 
- [ ]

### DIFF
--- a/platform/marvell/sonic-platform-nokia/7215/service/cpu_wdt.service
+++ b/platform/marvell/sonic-platform-nokia/7215/service/cpu_wdt.service
@@ -1,5 +1,6 @@
 [Unit]
 Description=CPU WDT
+Conflicts=watchdog-control.service
 After=nokia-7215init.service
 [Service]
 ExecStart=/usr/local/bin/cpu_wdt.py
--- a/platform/marvell/sonic-platform-nokia/debian/sonic-platform-nokia-7215.postinst
+++ b/platform/marvell/sonic-platform-nokia/debian/sonic-platform-nokia-7215.postinst
@@ -9,6 +9,8 @@ systemctl restart kmod
 systemctl enable nokia-7215init.service
 systemctl start nokia-7215init.service
 
+systemctl mask --now watchdog-control.service
+
 systemctl enable cpu_wdt.service
 systemctl start cpu_wdt.service
 


platform/marvell/sonic-platform-nokia/7215/service/cpu_wdt.service
```diff
@@ -1,6 +1,6 @@
 [Unit]
 Description=CPU WDT
-After=nokia-7215init.service
+After=nokia-7215init.service watchdog-control.service
```
prgeor, line 3: @Pavan-Nokia we don't need two watchdog services. Please override the watchdog-control.service completely

Pavan-Nokia: @prgeor the watchcgdog-control.service is part of common code and is triggered by the docker. There is no good way to disable this service from a platform context.
The only way I could think of is to crudely delete the script so that the watchdog-control.service fails, which I believe is not a great solution. that is the reason I chose this approach.
I do not understand why watchdog-control.service was introduced by SONiC.

prgeor: @saiarcot895 do you have a better way to override watchdog-control.service by platform specific watchdog service?

Pavan-Nokia: @saiarcot895 any feedback on this. I tried disabling the watchdog-control.service from the platform postinstall but it gets reenabled on next boot.
This is the only solution I could come up with that works reliably every time.

saiarcot895: You can use `/usr/local/lib/systemd/system/watchdog-control.service` to provide a platform-specific `watchdog-control.service` file. Specifically, systemctl treats files in `/usr/local/lib/systemd/system/` as higher priority than `/usr/lib/systemd/system/`, but lower than `/etc/systemd/system/` (which should still allow service mask/unmask to work).

Make sure that your platform-specific file has all of the ordering and conditions (Provides, After, WantedBy) of the original service file.

platform/marvell/sonic-platform-nokia/debian/sonic-platform-nokia-7215.postinst
```diff
@@ -10,7 +10,7 @@ systemctl enable nokia-7215init.service
 systemctl start nokia-7215init.service
 
 systemctl enable cpu_wdt.service
-systemctl start cpu_wdt.service
+systemctl start cpu_wdt.service &
```
saiarcot895, line 13: If this line is still needed, please use the `--no-block` argument instead of backgrounding this job.

platform/marvell/sonic-platform-nokia/7215/service/cpu_wdt.service
Pavan-Nokia: @saiarcot895
If I understand what you are saying correctly, You are suggesting that I eliminate the cpu_wdt.service completely and run my script under watchdog-control.service. I do not think this is a good Idea as the behavior of these 2 services are intended to be totally different. The watchdog-control is designed as a one shot to turn off the watchdog here as what the intent of cpu_wdt is to enable the watchdog as early as we can and keep running.

The best approach I think would be to have a different service for the platform watchdog and to disable the watchdog-control for our platform. Changing the functionality of an existing service is not good

saiarcot895: I'm fine with that approach, as long as on this platform, only one of the two services can be started. This would mean masking the service early in the bootup (perhaps by adding a symlink at `/usr/lib/systemd/system/watchdog-control.service` to point to `/dev/null`?).

Pavan-Nokia: @saiarcot895 Could you please review my final changes 
Thank you

platform/marvell/sonic-platform-nokia/debian/sonic-platform-nokia-7215.postinst
```diff
@@ -9,6 +9,8 @@ systemctl restart kmod
 systemctl enable nokia-7215init.service
 systemctl start nokia-7215init.service
 
+systemctl mask watchdog-control.service
```
saiarcot895, line 12: Is this guaranteed to happen before `watchdog-control.service` starts? If there's a chance it could be after, then add `--now` to have systemd stop the service as well.

Pavan-Nokia: I have verified all scenarios to make sure there is no race condition - fresh install, software upgrade, reboots. The `watchdog-control.service` is always masked correctly and the platform watchdog is active.

I have added the --now flag as recommended. I agree this is good to have in case something changes in the future

saiarcot895 (approved):

Pavan-Nokia:
@yxieca @Blueve Please help merge and backport to 202405 branch
also help merge PR for 2311 backport #18851 

Thank you

yxieca:
@bingwang-ms to tracking 202405 cherry-picking

PR #18855, opened 2024-05-02
[chassis][voq] Remove the leaf ref for the ifname in sonic-system-port.yang

### LINKED ISSUE #18854: [chassis] Yang validation fails for the system port yang
1.During the testing of incremental configuration of sonic chassis the application of the patch fails with below error
```
Failed to apply patch due to: Failed to apply patch on the following scopes:
- localhost: Given patch will produce invalid config. Error: Data Loading Failed
Invalid value "Ethernet-IB0" in "ifname" element.
- asic0: Given patch will produce invalid config. Error: Data Loading Failed
Invalid value "Ethernet-IB0" in "ifname" element.
- asic1: Given patch will produce invalid config. Error: Data Loading Failed
Invalid value "Ethernet-IB0" in "ifname" element.
```

202205
```
(paste your output here)
```
#### Output of `show techsupport`:

```
(paste your output here or download and attach the file here )
```

arlakshm:
##### Work item tracking
- Microsoft ADO **27934751**:

#### How I did it
Fixes #18854 
Update the `sonic-system-port.yang` to remove the check to have `ifname` as leafref in the port table
The system-port config table will have all the ports in the chassis not only to one linecard, so in case of the chassis with different linecards sku with different number of ports this leafref will not work.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/src/sonic-yang-models/yang-models/sonic-system-port.yang
+++ b/src/sonic-yang-models/yang-models/sonic-system-port.yang
@@ -40,13 +40,8 @@ module sonic-system-port {
                 }
 
                 leaf ifname {
-                    type union {
-                        type leafref {
-                            path /port:sonic-port/port:PORT/port:PORT_LIST/port:name;
-                        }
-                        type string {
-                            pattern "Cpu0";
-                        }
+                    type string {
+                        length 1..128;
                     }
                 }
 


kenneth-arista (approved):

xincunli-sonic (approved):

arlakshm:
test `(Test kvmtest-t0 by Elastictest)` failures are not related to this change

src/sonic-yang-models/yang-models/sonic-system-port.yang
```diff
@@ -40,13 +40,8 @@ module sonic-system-port {
...
-                    type union {
-                        type leafref {
-                            path /port:sonic-port/port:PORT/port:PORT_LIST/port:name;
-                        }
-                        type string {
-                            pattern "Cpu0";
-                        }
+                    type string {
+                        length 1..128;
                     }
```
matiAlfaro, line 45: Wouldn't it be better to add a reference to voq interface name?
```suggestion
			type leafref {
				path "/voq-inband-intf:sonic-voq-inband-interface/voq-inband-intf:VOQ_INBAND_INTERFACE/voq-inband-intf:VOQ_INBAND_INTERFACE_LIST/voq-inband-intf:name";
			}
                    }
```

you'll also need to add 
```
    import sonic-voq-inband-interface {
        prefix voq-inband-intf;
    }
```
at the import section of this file

arlakshm: This problem is not related only to Inband port.If we have a scenario where one Linecard on the chassis has 36 ports and another linecard has 48 ports, then this leafref does not hold true.

qiluo-msft: Why relax and remove the constrains?

src/sonic-yang-models/yang-models/sonic-system-port.yang
```diff
@@ -40,13 +40,8 @@ module sonic-system-port {
                 }
 
                 leaf ifname {
```
qiluo-msft, line 42: >ifname [](http://example.com/codeflow?start=21&length=6)

If you think this is a bugfix, could you also add testcases?

arlakshm: @qiluo-msft, is the not bugfix. It was an extra check which was added for this yang which is not required. The current tests cover the current yang model changes

PR #18864, opened 2024-05-03
[Nokia-IXR7250E] disable pcie aer

jon-nokia:
#### Why I did it
In Linux kernel v6.0, commit, f26e58bf6f5470 "PCI/AER: Enable error reporting when AER is native" changed the default behavior of Advanced Error Reporting (AER).  Reporting is now enabled by default, which was not the case prior.  There are custom FPGA devices on our PCI bus that are not yet configured nor enabled across reboot until the Sonic platform is initialized.  Therefore this is impacting AER adversely.  Since AER was not enabled in 202205, this is not a regression.

##### Work item tracking

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305
- [x] 202405

- [ ] 
- [ ]

### DIFF
--- a/device/nokia/x86_64-nokia_ixr7250e_36x400g-r0/installer.conf
+++ b/device/nokia/x86_64-nokia_ixr7250e_36x400g-r0/installer.conf
@@ -1,4 +1,4 @@
 CONSOLE_PORT=0x3f8
 CONSOLE_DEV=0
 CONSOLE_SPEED=115200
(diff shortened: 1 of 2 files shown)


lguohan:
in 202411, can we enable this AER reporting? this give the healthness of the pcie devices.

lguohan (approved):

lguohan:
@prgeor , this might be able affecting other platforms, i do not know if we need to disable this one other platform or not.

@jon-nokia , can you share the error log if there is any?

PR #18882, opened 2024-05-06
kdump fix, initializing network interface on crash kernel boot over 2023_11 release

bilal-ismail:
#### Why I did it
For pre-boot initialization of network interfaces.
For my case, I had to test and implement the remote kdump file transfer using SSH and the crash kernel did not load configs from /etc/network/interfaces config file.

#### How I did it
New file created for a custom script, "network-interface-state-init.sh"
Modified the "build_debian.sh" file for adding this to the /usr/sbin/network-interface-state-init.sh.

#### How to verify it
After this change the remote SSH kdump file transfer would be possible because network not reachable issue arose as the default crash kernel image did not change state of the interfaces to "UP" upon load. Neither did it load configs from /etc/network/interfaces config file.


bilal-ismail:
This fix had build issues in both "marvell_armhf" and "marvell_arm64" artifacts, therefore because of a conditional statement these changes have been skipped in build of those artifacts where kdump-tools packages is not installed on default location/path i.e for the script, "/usr/sbin/kdump-config"

PR #18979, opened 2024-05-16
Update save disable code for new Redis version

saiarcot895:
#### Why I did it

With the upgrade of docker-database to Bookworm in #18108 and the new Redis version that was brought in, database saves are enabled by default. However, we don't use (and maybe don't want, at this time) those database saves.

##### Work item tracking

#### How I did it

Restore the old behavior by modifying `/etc/redis/redis.conf` to not save at all.

#### How to verify it

Checked `/var/log/syslog` to make sure there weren't messages from redis every minute saying the database was being saved, and verified `/etc/redis/redis.conf` in the database container has a config line with `save ""`.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/dockers/docker-database/Dockerfile.j2
+++ b/dockers/docker-database/Dockerfile.j2
@@ -28,7 +28,7 @@ RUN apt-get clean -y                                  && \
     apt-get autoclean -y                              && \
     apt-get autoremove -y                             && \
     rm -rf /debs ~/.cache                             && \
-    sed -ri 's/^(save .*$)/# \1/g;                       \
+    sed -ri 's/^# save ""$/save ""/g;                    \
              s/^daemonize yes$/daemonize no/;            \
              s/^logfile .*$/logfile ""/;                 \
              s/^# syslog-enabled no$/syslog-enabled no/; \


abdosi:
@saiarcot895 : I assume this should fix this: **https://github.com/sonic-net/sonic-buildimage/issues/18667**

saiarcot895:
@abdosi Yes, this will fix that. I didn't realize there was already an impact here.

abdosi:
> @abdosi Yes, this will fix that. I didn't realize there was already an impact here.

can we add some UT to catch this as this is major behavior change

lguohan:
@saiarcot895 , can you check the conflict?

saiarcot895:
> @saiarcot895 , can you check the conflict?

It's saying that a conflict already exists in master, and to wait a few hours and rerun the conflict checker.

> can we add some UT to catch this as this is major behavior change

Opened sonic-net/sonic-mgmt#12877.

PR #19046, opened 2024-05-23
[Mellanox] Verify the value of the control type sysfs before changing it to FW control

tshalvi:
#### Why I did it
On Mellanox platforms, changing the control type sysfs value is not supported for ports that are set to firmware control. Therefore, attempting to change this value to firmware control will result in an error log, which we want to avoid.

##### Work item tracking

#### How I did it
I added a check to ensure that the control sysfs value is set to firmware control only if it is not already in firmware control.

#### How to verify it
Ensure that after a warm reboot on the Mellanox platform (an attempt to change the sysfs value from fw_control to fw_control is only applicable in this scenario), the following error does not appear:
sxd_kernel: [error] sx_core_set_module_control: Failed control change: module is dependent (FW control).

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/platform/mellanox/mlnx-platform-api/sonic_platform/sfp.py
+++ b/platform/mellanox/mlnx-platform-api/sonic_platform/sfp.py
@@ -1493,7 +1493,8 @@ class SFP(NvidiaSFPCommon):
     @classmethod
     def action_on_fw_control(cls, sfp):
         logger.log_info(f'SFP {sfp.sdk_index} is set to firmware control')
-        sfp.set_control_type(SFP_FW_CONTROL)
+        if sfp.get_control_type() != SFP_FW_CONTROL:
+            sfp.set_control_type(SFP_FW_CONTROL)
         
     @classmethod
     def action_on_cancel_wait(cls, sfp):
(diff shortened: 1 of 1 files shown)


Junchao-Mellanox (approved):

platform/mellanox/mlnx-platform-api/sonic_platform/sfp.py
```diff
@@ -1493,7 +1493,8 @@ def action_on_sw_control(cls, sfp):
     @classmethod
     def action_on_fw_control(cls, sfp):
         logger.log_info(f'SFP {sfp.sdk_index} is set to firmware control')
-        sfp.set_control_type(SFP_FW_CONTROL)
+        if sfp.get_control_type() != SFP_FW_CONTROL:
```
liat-grozovik, line 1496: if it is already in fw control why do we need the log in notice? maybe you should use debug as already in fw control?

tshalvi: I think we can just move this log inside the if statement because the actual set_control_type() mentioned in this log is performed only in that case.

liat-grozovik (approved):

dprital:
@yxieca , Can you please cherry pick to 202311 ?

bingwang-ms:
MSFT ADO: 28219305

PR #19088, opened 2024-05-27
[bgp] Re-add dependency in managers_bgp for bgpcfgd

yaqiangz:
#### Why I did it
These 2 dependencies were removed by https://github.com/sonic-net/sonic-buildimage/pull/18764 and https://github.com/sonic-net/sonic-buildimage/pull/18727 for decoupling hard dependency between Loopback IPv4 address and BGP.
Actually, these dependencies are to make sure existence of Loopback interface, no need to remove.

##### Work item tracking
- Microsoft ADO **(number only)**: 28210625

#### How I did it
Add dependency back

#### How to verify it
UT passed.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/src/sonic-bgpcfgd/bgpcfgd/managers_bgp.py
+++ b/src/sonic-bgpcfgd/bgpcfgd/managers_bgp.py
@@ -106,6 +106,7 @@ class BGPPeerMgrBase(Manager):
 
         deps = [
             ("CONFIG_DB", swsscommon.CFG_DEVICE_METADATA_TABLE_NAME, "localhost/bgp_asn"),
+            ("CONFIG_DB", swsscommon.CFG_LOOPBACK_INTERFACE_TABLE_NAME, "Loopback0"),
             ("CONFIG_DB", swsscommon.CFG_BGP_DEVICE_GLOBAL_TABLE_NAME, "tsa_enabled"),
             ("LOCAL", "local_addresses", ""),
             ("LOCAL", "interfaces", ""),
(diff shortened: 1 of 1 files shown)


StormLiangMS:
hi @yaqiangz I remember you mentioned this interface would be monitored in the bgp manager file? why we need to add these two back?

StormLiangMS:
> hi @yaqiangz I remember you mentioned this interface would be monitored in the bgp manager file? why we need to add these two back?

src/sonic-bgpcfgd/bgpcfgd/main.py

yaqiangz:
> hi @yaqiangz I remember you mentioned this interface would be monitored in the bgp manager file? why we need to add these two back?

@StormLiangMS Yes interface change will be monitored by https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-bgpcfgd/bgpcfgd/main.py#L50

The dependency is to make sure Loopback interface exist. With this dependency adding:
If interested table changed, but CONFIG_DB entry `LOOPBACK_INTERFACE|Loopback0` doesn't exist, it will not invoke handler for the table change, but record it until the entry `LOOPBACK_INTERFACE|Loopback0` appear.

StormLiangMS:
After discussion, if we configured IPv6 or IPv4 on the loopback interface, it will create LOOPBACK_INTERFACE|Loopback0 in the config_db, no matter a IPv6 or IPv4 addressed configured. In previous assumption, the device could be IPv6 only later, and in either case, there should at least one of them  (IPv6 or IPv4) available, no harm to add the dependency on LOOPBACK_INTERFACE|Loopback0 back.

StormLiangMS (approved):
LGTM.

bingwang-ms:
Hi @yaqiangz , @StormLiangMS  Is this change required for 202405 branch?

yaqiangz:
> Hi @yaqiangz , @StormLiangMS Is this change required for 202405 branch?

Hi @bingwang-ms it's included in 202405 by this commit https://github.com/sonic-net/sonic-buildimage/commit/d1875c66ea0d15b218c011f572ed8884a389176d

PR #19122, opened 2024-05-29
Revert "[Marvell-armhf] Update MRVL_PRESTERA_DRIVER"

pavannaregundi:
Reverts sonic-net/sonic-buildimage#18369

Reverting the SDK Driver change done to use reserved memory, as we saw issues related to delayed mac learning and ecmp tests.


lguohan:
can you add why revert?

pavannaregundi:
> can you add why revert?

@lguohan Have updated the description.

PR #19160, opened 2024-05-31
[minigraph] Added support to parse "AssociatedSliceStr" attribute of minigraph

abdosi:
What I did:
Added support to parse "AssociatedSliceStr" attribute of minigraph and save as `slice_type` as part of DEVICE_METADATA for Chassis Device Type.
Yang model are done as part of : https://github.com/sonic-net/sonic-buildimage/pull/19094

Also as part of this change fixed issue of current device not selected correctly as asic_hostname not being correct.

Why I did:
In BGP Templates we need to take different action based on this attribute.

How I verify:
UT updated.


arlakshm (approved):

lguohan:
@qiluo-msft , here is another minigraph parse change. can you check if you can accomodate this one?

abdosi:
@lguohan / @qiluo-msft  :  please help merge of this.

qiluo-msft (approved):

abdosi:
@bingwang-ms : can you please pick for 202405.Needed for chassis

PR #19163, opened 2024-05-31
Run logger CLI with parent script PID

theasianpianist:
#### Why I did it
When a script calls the `logger` utility, each individual call gets its own PID assigned. Rsyslogd creates an internal rate-limit  object per PID that sends a log message. If the `logger` utility is called repeatedly, it will cause rsyslogd memory usage to increase permanently since a rate-limit object is being created for each call.

##### Work item tracking
- Microsoft ADO **28263484**:

#### How I did it
Anytime the `logger` command is used, also use the `--id` argument to specify the parent script PID that rsyslogd should use for rate limiting.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/device/celestica/x86_64-cel_e1031-r0/thermal_overload_control.sh
+++ b/device/celestica/x86_64-cel_e1031-r0/thermal_overload_control.sh
@@ -42,12 +42,12 @@ usage() {
 }
 
 cpu_overload() {
-	logger "Enable CPU thermal overload control"
+	logger --id="$$" "Enable CPU thermal overload control"
 	set_reg=`echo ${TOVERREG} ${CPUOVER} > ${SETREG_FILE}`
 }
 
 asic_overload() {
-	logger "Enable ASIC thermal overload control"
+	logger --id="$$" "Enable ASIC thermal overload control"
 	set_reg=`echo ${TOVERREG} ${ASICOVER} > ${SETREG_FILE}`
 }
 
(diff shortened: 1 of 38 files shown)


Junchao-Mellanox:
what will be the pid in the log message after this change? The bash's PID?

theasianpianist:
> what will be the pid in the log message after this change? The bash's PID?

The PID of the script that is calling `logger`.

Junchao-Mellanox:
> > what will be the pid in the log message after this change? The bash's PID?
> 
> The PID of the script that is calling `logger`.

thanks

Junchao-Mellanox:
is it possible to use `alias logger=logger --id $$` ?

prgeor:
> is it possible to use `alias logger=logger --id $$` ?

@Junchao-Mellanox why?

Junchao-Mellanox:
we don't need to change most of the file if we use alias

prgeor:
@theasianpianist conflicts?

theasianpianist:
> is it possible to use `alias logger=logger --id $$` ?

Alias scope is generally limited to a shell process so I think we would still need to redefine it in every script. But we may be able to use a similar solution with a shell script wrapping the `logger` executable.

prgeor:
@theasianpianist https://github.com/sonic-net/sonic-buildimage/pull/19437 raised this PR instead. Awaiting to test.

PR #19179, opened 2024-06-03
[memory_monitoring] Enhance monitoring the memory usage of containers

FengPan-Frank:
#### Why I did it
We need to restrict memory usage of container specifically, and the reliable option is to read cgroup subsystem files instead of using "docker stats" commands, since the commands will be no response if containers hits hard limit.

##### Work item tracking
- Microsoft ADO **(number only)**:28334202

#### How I did it
Instead of depending on the output of docker stats, the background script memory_checker will calculate the memory usage of a container based on values reading from the cgroup subsystem files /sys/fs/cgroup/memory/docker/<container_name>/memory.usage_in_bytes and /sys/fs/cgroup/memory/docker/<container_name>/memory.stats.

Refer to this Docker official document (https://docs.docker.com/engine/reference/commandline/stats/#extended-description) to make sure the memory usage of a specific container reading from command output of docker stats is equal to the value subtracting cache usage from the total memory usage.

#### How to verify it
Local verified, since it's just internal enhancement for getting memory usage of container, below are comparison between new memory_check and previous implementation based on "docker stats --no-stream --format {{.MemUsage}} telemetry"

image

Added Unit test code, since there's no build time UT available in this repo currently so verified manually as below:
image

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]


files/image_config/monit/memory_checker
```diff
@@ -81,37 +165,33 @@ def check_memory_usage(container_name, threshold_value):
...
+        sys.exit(9)
+
+    total_memory_usage = memory_usage - cache_usage
+    syslog.syslog(syslog.LOG_INFO, "[memory_checker] Total memory usage of container '{}' is '{}' Bytes!"
+                  .format(container_name, total_memory_usage))
+
+    if total_memory_usage > threshold_value:
+        print("[{}]: Memory usage ({} Bytes) is larger than the threshold ({} Bytes)!"
+              .format(container_name, total_memory_usage, threshold_value))
+        sys.exit(3)
```
kperumalbfn, line 194: @FengPan-Frank could you add more details about the changes in the description?

files/image_config/monit/memory_checker
```diff
@@ -59,6 +62,87 @@ def get_command_result(command):
...
+
+    for line in command_stdout.splitlines():
+        if container_name in line:
+            container_id = line.split()[0].strip()
+            break
+
+    if not container_id:
+        syslog.syslog(syslog.LOG_ERR, "[memory_checker] Failed to get contianer ID of '{}'! Exiting ..."
+                      .format(container_name))
+        sys.exit(4)
```
zbud-msft, line 85: Can we define what certain error codes are and use those, such as for example, ERROR_CONTAINER_ID_NOT_FOUND? Maybe we can reuse certain error codes.

files/image_config/monit/memory_checker
```diff
@@ -59,6 +62,87 @@ def get_command_result(command):
...
+    get_memory_usage_cmd = "sudo cat {}".format(docker_memory_usage_file_path)
+    memory_usage_in_bytes = get_command_result(get_memory_usage_cmd)
+    if not memory_usage_in_bytes:
+        syslog.syslog(syslog.LOG_ERR, "[memory_checker] Failed to get the memory usage of container '{}'! Exiting ..."
+                      .format(container_id))
+        sys.exit(6)
+
+    return memory_usage_in_bytes
+
+def get_cache_usage(container_id):
```
zbud-msft, line 115: I think maybe we can refer to this as getting inactive cache usage since we want to calculate container memory usage as total memory usage - inactive cache usage

files/image_config/monit/memory_checker
```diff
@@ -81,37 +165,33 @@ def check_memory_usage(container_name, threshold_value):
...
+    container_id = get_container_id(container_name)
+    syslog.syslog(syslog.LOG_INFO, "[memory_checker] Container ID of '{}' is: '{}'."
+                  .format(container_name, container_id))
+
+    memory_usage_in_bytes = get_memory_usage(container_id)
+    syslog.syslog(syslog.LOG_INFO, "[memory_checker] The memory usage of container '{}' is '{}' Bytes!"
+                  .format(container_name, memory_usage_in_bytes))
 
+    cache_usage_in_bytes = get_cache_usage(container_id)
+    syslog.syslog(syslog.LOG_INFO, "[memory_checker] The cache usage of container '{}' is '{}' Bytes!"
```
zbud-msft, line 177: Maybe it would be beneficial to also see what the active cache usage is? Since here you are defining cache usage as only the inactive cache usage.

files/image_config/monit/memory_checker
FengPan-Frank: done.

files/image_config/monit/memory_checker
FengPan-Frank: Updated.

files/image_config/monit/memory_checker
FengPan-Frank: This should be just identical to the logic of "docker stats", refer https://docs.docker.com/reference/cli/docker/container/stats/#extended-description and below snippet calculation.

image

files/image_config/monit/memory_checker
FengPan-Frank: updated.

files/image_config/monit/memory_checker
```diff
@@ -59,6 +69,81 @@ def get_command_result(command):
...
+def get_container_id(container_name):
+    """Gets full container ID of the specified container
+    Args:
+        container_name: A string indicates the name of specified container.
+    Returns:
+        container_id: A string indicates the full ID of specified container.
+    """
+    container_id = ""
+
+    get_container_info_cmd = "docker ps --no-trunc"
```
maipbui, line 81: >"docker ps --no-trunc" [](http://example.com/codeflow?start=29&length=22)

I recall that in the `get_command_result()` function, when using subprocess with shell=False, the command should be provided as a list of strings rather than a single string. Could you double-check this? Same for other places

files/image_config/monit/memory_checker
```diff
@@ -59,6 +69,81 @@ def get_command_result(command):
...
+        memory_usage_in_bytes: A string indicates memory usage (Bytes) of a container.
+    """
+    memory_usage_in_bytes = ""
+
+    docker_memory_usage_file_path = CGROUP_DOCKER_MEMORY_DIR + container_id + "/memory.usage_in_bytes"
+    if not os.path.exists(docker_memory_usage_file_path):
+        syslog.syslog(syslog.LOG_ERR, ERROR_CGROUP_MEMORY_USAGE_NOT_FOUND.format(docker_memory_usage_file_path, container_id))
+        sys.exit(5)
+
+    get_memory_usage_cmd = "sudo cat {}".format(docker_memory_usage_file_path)
```
xincunli-sonic, line 111: Why do we not use python inbuild open function? Suppose this script should be running by sudoer?

files/image_config/monit/memory_checker
```diff
@@ -59,6 +69,81 @@ def get_command_result(command):
...
+        cache_usage_in_bytes: A string indicates the cache usage (Bytes) of a container.
+    """
+    cache_usage_in_bytes = ""
+
+    docker_memory_stat_file_path = CGROUP_DOCKER_MEMORY_DIR + container_id + "/memory.stat"
+    if not os.path.exists(docker_memory_stat_file_path):
+        syslog.syslog(syslog.LOG_ERR, ERROR_CGROUP_MEMORY_STATS_NOT_FOUND.format(docker_memory_stat_file_path, container_id))
+        sys.exit(7)
+
+    get_inactive_cache_usage_cmd = "sudo cat {}".format(docker_memory_stat_file_path)
```
xincunli-sonic, line 134: remove unnecessary sudo

files/image_config/monit/memory_checker
```diff
@@ -59,6 +69,81 @@ def get_command_result(command):
...
+    get_container_info_cmd = "docker ps --no-trunc"
+    command_stdout = get_command_result(get_container_info_cmd)
+
+    for line in command_stdout.splitlines():
+        if container_name in line:
+            container_id = line.split()[0].strip()
+            break
+
+    if not container_id:
+        syslog.syslog(syslog.LOG_ERR, ERROR_CONTAINER_ID_NOT_FOUND.format(container_name))
```
xincunli-sonic, line 90: Minor suggestion: Can we refactor this log usage to be a common function, like

def log_and_exit(error_message, exit_code):
    syslog.syslog(syslog.LOG_ERR, error_message)
    sys.exit(exit_code)

Also, given exit_code constants in top of script?

files/image_config/monit/memory_checker
```diff
@@ -81,37 +166,33 @@ def check_memory_usage(container_name, threshold_value):
...
+        cache_usage = int(cache_usage_in_bytes)
+    except ValueError as err:
+        syslog.syslog(syslog.LOG_ERR, "[memory_checker] Failed to convert the memory or cache usage in string to integer! Exiting ...")
+        sys.exit(9)
+
+    total_memory_usage = memory_usage - cache_usage
+    syslog.syslog(syslog.LOG_INFO, "[memory_checker] Total memory usage of container '{}' is '{}' Bytes!"
+                  .format(container_name, total_memory_usage))
+
+    if total_memory_usage > threshold_value:
```
xincunli-sonic, line 192: Can you help add a validation for input of args.threshold_value? Like negative value and zero?

files/image_config/monit/memory_checker
xincunli-sonic, line 81: Maybe we can use this command to filter specific container name?

get_container_info_cmd = "docker ps --no-trunc --filter name={}".format(container_name)

files/image_config/monit/memory_checker
```diff
@@ -33,6 +34,15 @@ from swsscommon import swsscommon
...
+CGROUP_DOCKER_MEMORY_DIR = "/sys/fs/cgroup/memory/docker/"
+
+# Define common error codes
+ERROR_CONTAINER_ID_NOT_FOUND = "[memory_checker] Failed to get container ID of '{}'! Exiting ..."
+ERROR_CGROUP_MEMORY_USAGE_NOT_FOUND = "[memory_checker] cgroup memory usage file '{}' of container '{}' does not exist on device! Exiting ..."
+ERROR_CONTAINER_MEMORY_USAGE_NOT_FOUND = "[memory_checker] Failed to get the memory usage of container '{}'! Exiting ..."
+ERROR_CONTAINER_CACHE_USAGE_NOT_FOUND = "[memory_checker] Failed to get the cache usage of container '{}'! Exiting ..."
+ERROR_CGROUP_MEMORY_STATS_NOT_FOUND = "[memory_checker] cgroup memory statistics file '{}' of container '{}' does not exist on device! Exiting ..."
+
 def get_command_result(command):
```
xincunli-sonic, line 46: Is it possible to add retry mechanism in case timing issue? Like below

```python
import time

def get_command_result_with_retries(command, retries=3, delay=5):
    for attempt in range(retries):
        try:
            return get_command_result(command)
        except Exception as e:
            if attempt < retries - 1:
                time.sleep(delay)
            else:
                syslog.syslog(syslog.LOG_ERR, f"[memory_checker] Command '{command}' failed after {retries} attempts. Error: {e}")
                sys.exit(2)
```

files/image_config/monit/memory_checker
```diff
@@ -33,6 +34,15 @@ from swsscommon import swsscommon
 EVENTS_PUBLISHER_SOURCE = "sonic-events-host"
 EVENTS_PUBLISHER_TAG = "mem-threshold-exceeded"
 
+CGROUP_DOCKER_MEMORY_DIR = "/sys/fs/cgroup/memory/docker/"
```
xincunli-sonic, line 37: Can we add UT for all key functions?

files/image_config/monit/memory_checker
FengPan-Frank: There is already retry used in monit config, refer https://github.com/FengPan-Frank/sonic-buildimage/blob/d0d8b9f84083e545857749605ea7cf9a342a1f39/dockers/docker-sonic-telemetry/base_image_files/monit_telemetry#L5

files/image_config/monit/memory_checker
FengPan-Frank: done

files/image_config/monit/memory_checker
FengPan-Frank: Good point. done.

files/image_config/monit/memory_checker
FengPan-Frank: In current code not all the syslog require exit actually, so we may keep it first as error codes have been defined on the top.

files/image_config/monit/memory_checker
FengPan-Frank: N/A

files/image_config/monit/memory_checker
```diff
@@ -59,6 +69,81 @@ def get_command_result(command):
...
+    get_container_info_cmd = "docker ps --no-trunc --filter name={}".format(container_name)
+
+    command_stdout = get_command_result(get_container_info_cmd)
+
+    for line in command_stdout.splitlines():
+        if container_name in line:
+            container_id = line.split()[0].strip()
+            break
+
+    if not container_id:
```
xincunli-sonic, line 90: >container_id [](http://example.com/codeflow?start=11&length=12)

The container id should be guid format, can we do assertion, only concern is what if command output returns something else it may not be empty.

FengPan-Frank: Made some check and it seems container-id is not GUID actually, which is just unique id created as 
https://github.com/docker-archive/docker-ce/blob/44a430f4c43e61c95d4e9e9fd6a0573fa113a119/components/engine/daemon/names.go#L41

files/image_config/monit/memory_checker
FengPan-Frank: done, since there's no build time UT in current repo, after discussion with Qi just verified test code manually and updated result into PR.

files/image_config/monit/tests/__init__.py
xincunli-sonic, line 1: Minor suggestion, remove this file.

xincunli-sonic (approved):
LGTM

files/image_config/monit/tests/__init__.py
FengPan-Frank: Thanks, I'll refactor this part later, briefly test code will be moved into internal repo, as well as python wheel/pytest built-in.

files/image_config/monit/memory_checker
```diff
@@ -59,6 +69,81 @@ def get_command_result(command):
...
+        syslog.syslog(syslog.LOG_ERR, ERROR_CONTAINER_ID_NOT_FOUND.format(container_name))
+
+        sys.exit(4)
```
qiluo-msft, line 93: >4 [](http://example.com/codeflow?start=17&length=1)

magic number. Could you defined a name constant?

files/image_config/monit/memory_checker
```diff
@@ -59,6 +69,81 @@ def get_command_result(command):
...
+def get_container_id(container_name):
+    """Gets full container ID of the specified container
+    Args:
+        container_name: A string indicates the name of specified container.
+    Returns:
+        container_id: A string indicates the full ID of specified container.
+    """
+    container_id = ""
+
+    get_container_info_cmd = "docker ps --no-trunc --filter name={}".format(container_name)
```
qiluo-msft, line 81: >container_name [](http://example.com/codeflow?start=76&length=14)

This is a obvious vulnerability for command line injection. Malicious user could easily give a bad container name with special char. Please valid input, and never concat a command lines. Keep tokens in an array. 
@maipbui to help review.

files/image_config/monit/memory_checker
```diff
@@ -59,6 +69,81 @@ def get_command_result(command):
...
+        container_id: A string indicates the full ID of a container.
+    Returns:
+        memory_usage_in_bytes: A string indicates memory usage (Bytes) of a container.
+    """
+    memory_usage_in_bytes = ""
+
+    docker_memory_usage_file_path = CGROUP_DOCKER_MEMORY_DIR + container_id + "/memory.usage_in_bytes"
+    if not os.path.exists(docker_memory_usage_file_path):
+        syslog.syslog(syslog.LOG_ERR, ERROR_CGROUP_MEMORY_USAGE_NOT_FOUND.format(docker_memory_usage_file_path, container_id))
+        sys.exit(5)
```
qiluo-msft, line 110: >5) [](http://example.com/codeflow?start=17&length=2)

magic number

files/image_config/monit/memory_checker
```diff
@@ -59,6 +69,81 @@ def get_command_result(command):
...
+    """Reads the container's memory usage from the control group subsystem's file
+    '/sys/fs/cgroup/memory/docker/<container_id>/memory.usage_in_bytes'.
+    Args:
+        container_id: A string indicates the full ID of a container.
+    Returns:
+        memory_usage_in_bytes: A string indicates memory usage (Bytes) of a container.
+    """
+    memory_usage_in_bytes = ""
+
+    docker_memory_usage_file_path = CGROUP_DOCKER_MEMORY_DIR + container_id + "/memory.usage_in_bytes"
```
qiluo-msft, line 107: >container_id [](http://example.com/codeflow?start=63&length=12)

you should validate the parameter. malicious user may try "../.." as container_id.

files/image_config/monit/memory_checker
```diff
@@ -59,6 +69,81 @@ def get_command_result(command):
...
+    """Reads the container's cache usage from the field 'total_inactive_file' in control
+    group subsystem's file '/sys/fs/cgroup/memory/docker/<container_id>/memory.stat'.
+    Args:
+        container_id: A string indicates the full ID of a container.
+    Returns:
+        cache_usage_in_bytes: A string indicates the cache usage (Bytes) of a container.
+    """
+    cache_usage_in_bytes = ""
+
+    docker_memory_stat_file_path = CGROUP_DOCKER_MEMORY_DIR + container_id + "/memory.stat"
```
qiluo-msft, line 131: >container_id [](http://example.com/codeflow?start=62&length=12)

the same

files/image_config/monit/memory_checker
```diff
@@ -59,6 +69,81 @@ def get_command_result(command):
...
+    if not os.path.exists(docker_memory_stat_file_path):
+        syslog.syslog(syslog.LOG_ERR, ERROR_CGROUP_MEMORY_STATS_NOT_FOUND.format(docker_memory_stat_file_path, container_id))
+        sys.exit(7)
+
+    try:
+        with open(docker_memory_stat_file_path, 'r') as file:
+            for line in file:
+                if "total_inactive_file" in line:
+                    cache_usage_in_bytes = line.split()[1].strip()
```
qiluo-msft, line 140: >[1] [](http://example.com/codeflow?start=55&length=3)

index out of range?

files/image_config/monit/memory_checker
FengPan-Frank: Added validate function as well as one test case test_get_memory_usage_invalid().

files/image_config/monit/memory_checker
FengPan-Frank: Added checker.

files/image_config/monit/memory_checker
FengPan-Frank: Updated and unified some exit code into internal_error with syslog in details, now monit only uses 3, refer https://github.com/search?q=repo%3Asonic-net%2Fsonic-buildimage%20memory_checker&type=code

files/image_config/monit/memory_checker
maipbui: I agree with Qi's comment. Need command to be a list of strings, instead of strings
e.g.
```
    command = ["docker", "stats", "--no-stream", "--format", "{{.MemUsage}}", container_name]
    command_stdout = get_command_result(command)
```

files/image_config/monit/memory_checker
```diff
@@ -45,20 +69,104 @@ def get_command_result(command):
     command_stdout = ""
 
     try:
-        proc_instance = subprocess.Popen(command, stdout=subprocess.PIPE, stderr=subprocess.PIPE,
+        proc_instance = subprocess.Popen(command.split(), stdout=subprocess.PIPE, stderr=subprocess.PIPE,
```
maipbui, line 72: >.split() [](http://example.com/codeflow?start=48&length=8)

Suggest not to use split and pass command as an array instead

files/image_config/monit/memory_checker
FengPan-Frank: Fixed

files/image_config/monit/memory_checker
FengPan-Frank: fixed.

maipbui (approved):

qiluo-msft (approved):

zbud-msft:
Hi @FengPan-Frank, publish_events is now deleted in this PR, which is not expected.

FengPan-Frank:
> Hi @FengPan-Frank, publish_events is now deleted in this PR, which is not expected.

@zbud-msft  sorry seems this publish_events was missed previously, may I ask which test case failed? so that I can further check it if update this part code later.

PR #19199, opened 2024-06-04
[Secureboot] Revert the changes done to only do Kernel Signnning

abdosi:
What I did:

Before this feature https://github.com/sonic-net/SONiC/pull/1028 got merged their was intermediate change done to just support Kernel Signing via: https://github.com/sonic-net/sonic-buildimage/pull/10557.  However once this feature is merged : https://github.com/sonic-net/SONiC/pull/1028  (Which support sign of all boot components not just Kernel) we do not need the Kernel only signing changes as it define new rules macro which just create confusion.

So as part of this PR i am reverting the Kernel -only sign PR https://github.com/sonic-net/sonic-buildimage/pull/10557


abdosi:
@xumia : Please help review.

xumia (commented):
LGTM

xumia (approved):

bmridul (approved):
LGTM

PR #19232, opened 2024-06-06
Support notice level log

Junchao-Mellanox:
#### Why I did it

SysLogger does not support NOTICE level log. The PR is to support it.

##### Work item tracking

#### How I did it

Add a new log level NOTICE to python logging package.

#### How to verify it

Manual test

```
Jun  3 03:09:44.353536 sonic NOTICE test[364076]: This is a notice
```

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/src/sonic-py-common/sonic_py_common/syslogger.py
+++ b/src/sonic-py-common/sonic_py_common/syslogger.py
@@ -4,13 +4,19 @@ import os
 import socket
 import sys
 
+# customize python logging to support notice logger
+logging.NOTICE = logging.INFO + 1
+logging.addLevelName(logging.NOTICE, "NOTICE")
+SysLogHandler.priority_map['NOTICE'] = 'notice'
+
+
 class SysLogger:
     """
     SysLogger class for Python applications using SysLogHandler
     """
 
     DEFAULT_LOG_FACILITY = SysLogHandler.LOG_USER
-    DEFAULT_LOG_LEVEL = SysLogHandler.LOG_NOTICE
+    DEFAULT_LOG_LEVEL = logging.NOTICE
 
     def __init__(self, log_identifier=None, log_facility=DEFAULT_LOG_FACILITY, log_level=DEFAULT_LOG_LEVEL):
         if log_identifier is None:
@@ -53,10 +59,11 @@ class SysLogger:
         self.log(logging.WARNING, msg, also_print_to_console)
 
     def log_notice(self, msg, also_print_to_console=False):
-        self.log(logging.INFO, msg, also_print_to_console)
+        self.log(logging.NOTICE, msg, also_print_to_console)
 
     def log_info(self, msg, also_print_to_console=False):
         self.log(logging.INFO, msg, also_print_to_console)
 
     def log_debug(self, msg, also_print_to_console=False):
         self.log(logging.DEBUG, msg, also_print_to_console)
+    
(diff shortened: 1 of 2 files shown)


keboliu (approved):

xincunli-sonic (approved):
LGTM

keboliu:
Hi @prgeor would you please help to merge?

src/sonic-py-common/sonic_py_common/syslogger.py
```diff
@@ -4,13 +4,19 @@
 import socket
 import sys
 
+# customize python logging to support notice logger
+logging.NOTICE = logging.INFO + 1
```
qiluo-msft, line 8: How do you ensure this integer is not used by vanilla `logging` library?

Junchao-Mellanox: Hi @qiluo-msft , vanilla `logger` library supports customized log level. Here is what I copied from python official code:
```python
#---------------------------------------------------------------------------
#
# Default levels and level names, these can be replaced with any positive set
# of values having corresponding names. There is a pseudo-level, NOTSET, which
# is only really there as a lower limit for user-defined levels. Handlers and
# loggers are initialized with NOTSET so that they will log all messages, even
# at user-defined levels.
#

CRITICAL = 50
FATAL = CRITICAL
ERROR = 40
WARNING = 30
WARN = WARNING
INFO = 20
DEBUG = 10
NOTSET = 0
```

prgeor, line 8: @Junchao-Mellanox how did you arrive at the formula?

prgeor (approved):

src/sonic-py-common/sonic_py_common/syslogger.py
Junchao-Mellanox: Python logging package support customizing log level. I read the source code of it and find that:
```python
#---------------------------------------------------------------------------
#
# Default levels and level names, these can be replaced with any positive set
# of values having corresponding names. There is a pseudo-level, NOTSET, which
# is only really there as a lower limit for user-defined levels. Handlers and
# loggers are initialized with NOTSET so that they will log all messages, even
# at user-defined levels.
#

CRITICAL = 50
FATAL = CRITICAL
ERROR = 40
WARNING = 30
WARN = WARNING
INFO = 20
DEBUG = 10
NOTSET = 0
```

As you can see, there are 10 spare numbers between INFO and WARN, so I take INFO+1 as the NOTICE level.

Junchao-Mellanox: Hi @qiluo-msft , any further comment?

qiluo-msft (dismissed):

src/sonic-py-common/sonic_py_common/syslogger.py
qiluo-msft: To ensure future proof, in case vanilla `logging` library suddently change behavior, could you add a testcase to check? Just logging a notice message and see it is captured as a notice syslog.

Junchao-Mellanox:
Hi @qiluo-msft @prgeor , I added a test case and all checker passed. Could you please help merge this?

qiluo-msft (approved):

PR #19309, opened 2024-06-14
add config_db entries for TELEMETRY|gnmi|save_on_set

rlucus:
#### Why I did it
https://github.com/sonic-net/SONiC/pull/1297

add config_db entries for TELEMETRY|gnmi|save_on_set

### DIFF
--- a/src/sonic-yang-models/yang-models/sonic-telemetry.yang
+++ b/src/sonic-yang-models/yang-models/sonic-telemetry.yang
@@ -71,6 +71,11 @@ module sonic-telemetry {
                     description "Port gnmi runs on.";
                 }
 
+                leaf save_on_set {
+                    type boolean;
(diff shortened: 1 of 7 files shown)


src/sonic-yang-models/yang-models/sonic-telemetry.yang
```diff
@@ -71,6 +71,11 @@ module sonic-telemetry {
                     description "Port gnmi runs on.";
                 }
 
+                leaf save_on_set {
```
ganglyu, line 74: would you please also update sonic-gnmi.yang?

rlucus: Done

ganglyu (approved):

isabelmsft (approved):

baxia-lan (approved):

PR #19315, opened 2024-06-14
Fix QoS configuations for Quicksilver

rick-arista:
Adds egress_lossy_pool and fixes SAI error caused by TH5 default config

#### Why I did it

This change is needed to fix a SAI crash on boot. These values zero out the reserved buffer space that by default would cause the pools defined in config_db.json to fail to create due to insufficient remaining room.

##### Work item tracking

#### How I did it

The buffer pool values were chosen by simply dividing the total available MMU memory into 3 pools.

#### How to verify it

Without this change device fails to boot correctly.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305
- [x] 202311
- [x] 202405

Fix QoS configuations for Quicksilver

### DIFF
--- a/device/common/profiles/th5/gen/BALANCED/buffers_defaults_t1.j2
+++ b/device/common/profiles/th5/gen/BALANCED/buffers_defaults_t1.j2
@@ -5,32 +5,37 @@
 {%- macro generate_buffer_pool_and_profiles() %}
     "BUFFER_POOL": {
         "ingress_lossless_pool": {
+            "size": "60000000",
             "type": "ingress",
             "mode": "dynamic",
-            "size": "60000000",
             "xoff": "22600000"
         },
-        "egress_lossless_pool": {
+        "egress_lossy_pool": {
+            "size": "41300000",
             "type": "egress",
-            "mode": "static",
-            "size": "82600000"
+            "mode": "dynamic"
+        },
+        "egress_lossless_pool": {
+            "size": "41300000",
+            "type": "egress",
+            "mode": "static"
         }
     },
     "BUFFER_PROFILE": {
         "ingress_lossy_profile": {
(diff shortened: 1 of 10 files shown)


device/common/profiles/th5/gen/BALANCED/buffers_defaults_t1.j2
```diff
@@ -5,32 +5,37 @@
...
-            "size": "60000000",
             "xoff": "22600000"
         },
+        "egress_lossy_pool": {
+            "size": "41300000",
+            "type": "egress",
+            "mode": "dynamic"
+        },
         "egress_lossless_pool": {
+            "size": "41300000",
```
lguohan, line 19: in the sai definition, we do not have ITM concept. we treat whole device as pool, and therefore i am not sure we need to cut this in half is satisfying the sai definition.

r12f:
hi @rick-arista , thanks a lot for sending this PR - somehow with this fix, the PFC frame still cannot be sent out with ixia test.  I can see the new pg profile is created now and the ports are associated with it, but no PFC frame will be generated:

image

device/common/profiles/th5/gen/BALANCED/buffers_defaults_t1.j2
rick-arista: I think you're right, but these are just placeholder values to fix a SAI crash anyways. We'll be updating these with tuned values shortly.

r12f (approved):
lgtm!

bingwang-ms:
@rick-arista Please file another PR for 202405 branch to address cherry-pick conflict.

PR #19335, opened 2024-06-18
[202012]: add new function: interface neighbor arp-keepalive

nandou1006:
#### Why I did it
From SONiC view,  arp/ndp entries of kernel have an aging time. After the aging of entries, only soft-forwarded packets in Linux trigger the new ARP learning process.

In SONiC, underlying packets are not sent to CPU, so ARP probe mechanism of linux kernel cannot be triggered immediately, which will cause chip hardware not updating arp/ndp entries in time and cause packet loss.

Specifically, in a test of 12K arp entries, only 4K+ entries left within a short period of time after the aging time is reached, which may result in sudden packet loss.

So, we implements a neighbor arp-keepalive feature to prevent packet loss caused by the deletion of neighbor entries during the refresh of neighbor entries after the aging time of ARP/NDP entries in the Linux kernel.

#### How I did it
1. arp-keepalive (arpd) maintains an active state of a neighbor entry, that is, the neighbor entry in stack will not be deleted due to aging. In addition, if a neighbor does not exist, arp-keepalive  will also delete a neighbor entry.
    1. Create a new docker-iproute container, and start arp-keepalive process (arpd) in the container to monitor changes of the state of kernel entries. So we modified the arpd source code in `iproute2-4.9.0` by patch files.
    2. If the current state of a neighbor entry is `NUD_STALE,` arpd will set its state as `NUD_DELAY` and sends it back to the kernel to ensure that the state mechine of a neighbor entry is changed to: `NUD_REACHABLE`->`NUD_STALE`- >`NUD_DELAY`- >`NUD_PROBE`, in order to skip the Entry Deletion phase.
2. Configuration 
    1. Modify the code for the show/config module in the sonic-utilities project, arp-keepalive (arpd) subscribes to the CONFIG_DB database for show and config configurations by commands.
      ```
      show interface neighbor keep-alive
      config interface neighbor keep-alive <mode> <interface_name> [OPTIONS]

      mode: [ enable/disable ] arp-keepalive
      -l --linklocal:  enable linklocal
      ```
3. In a scenario with a large number of arp entries, a large number of arp request/ns packets need to be sent. Arp-keepalive uses user space to take over the sending task instead of kernel space to relieve the kernel pressure
      1. This part is also modified by patch files. When the kernel  need to send arp request/ns packets , the action of sending packets will be token over by the user space.

- [x] 202012
- [x] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

Arp-keepalive (arpd) implemented above is modified by specified patch files and is the source file of `iproute2-4.9.0`, but iproute2-5.10.0 is used after 202106. We can do some work based on `iproute2-5.10.0` soon.
So, I submited two commits based on branch 202012 of sonic-buildimage and sonic-utilities.
1. This commit in sonic-buildimage: https://github.com/nandou1006/sonic-buildimage/commit/bab7af88560fb62f8096d3a5e70a4122f4f51788
2. The commit in sonic-utilities: https://github.com/nandou1006/sonic-utilities/commit/dde982e8456dd5eb6f3e8c90bf13eb3c3a22e9e2


lolyu (commented):
Please have a high level design PR to https://github.com/sonic-net/sonic to review first.

nandou1006:
> Please have a high level design PR to https://github.com/sonic-net/sonic to review first.

Sorry, I don't quite understand.
Since this is a new feature, there should be a feature related design document first?
Is it possible to give an example or a link to learn? Thans for your help.

lolyu:
> > Please have a high level design PR to https://github.com/sonic-net/sonic to review first.
> 
> Sorry, I don't quite understand. Since this is a new feature, there should be a feature related design document first? Is it possible to give an example or a link to learn? Thanks for your help.

Yes, you can take this [BGP supress fib pending HLD](https://github.com/sonic-net/SONiC/blob/master/doc/BGP/BGP-supress-fib-pending.md) as a guideline.

nandou1006:
Hi, Lyu.

I have just submitted the HLD documentation for interface neighbor arp keepalive feature.

https://github.com/sonic-net/SONiC/pull/1730

 Hope that we can give back to the community some ideas on how to solve related problems.

PR #19337, opened 2024-06-18
Revert "[master] Upgrade xgs SAI version to 10.1.21.0"

Gfrom2016:
Reverts sonic-net/sonic-buildimage#19302
master image with xgs SAI 10.1.21.0 would fail at switch initialization.


StormLiangMS (approved):
LGTM

PR #19437, opened 2024-06-30
Add wrapper around logger command to always pass id/pid argument

prgeor:
#### Why I did it
Rsyslogd daemon installs rate limiting per process id. If a process calls `logger` to log syslog messages, then if the process becomes too chatty, then rsyslogd memory usage increases over time because each logger message is generated with new PID. To prevent this increase of rsyslogd memory over time due to flooding of syslog messages, a wrapper script is installed which invokes `/usr/bin/logger` with `-i` so that the logger logs the message with same process id and hence rate limit will be done per process and not per log message.

##### Work item tracking

#### How I did it
Add wrapper around /usr/bin/logger 

#### How to verify it
1. Generate flood of syslog message in a script and see if rsyslogd memory is increasing
2. Ensure if the user passes `-id` or `--id` then this argument takes precd

- [x] 202305
- [x] 202311
- [x] 202405

- [ ] 
- [ ]

### DIFF
--- a/dockers/docker-base-bookworm/bin/wrapper_logger.sh
+++ b/dockers/docker-base-bookworm/bin/wrapper_logger.sh
@@ -1,0 +1,6 @@
+#!/bin/bash
+
+# This is a wrapper script for logger to take -i option as the default
+# If user provides -id or --id then that option takes precedence
+
+/usr/bin/logger -i "$@"
--- a/files/image_config/bash/wrapper_logger.sh
+++ b/files/image_config/bash/wrapper_logger.sh
@@ -1,0 +1,6 @@
+#!/bin/bash
+
+# This is a wrapper script for logger to take -i option as the default
+# If user provides -id or --id then that option takes precedence
+
+/usr/bin/logger -i "$@"
--- a/dockers/docker-base-bookworm/Dockerfile.j2
+++ b/dockers/docker-base-bookworm/Dockerfile.j2
@@ -120,6 +120,9 @@ RUN apt-get clean -y                     && \
 COPY ["etc/rsyslog.conf", "/etc/rsyslog.conf"]
 COPY ["etc/rsyslog.d/*", "/etc/rsyslog.d/"]
 COPY ["root/.vimrc", "/root/.vimrc"]
+COPY ["bin/wrapper_logger.sh", "/usr/local/bin/logger"]
+RUN chmod +x /usr/bin/logger
+
 
 RUN ln /usr/bin/vim.tiny /usr/bin/vim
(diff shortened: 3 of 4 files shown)


dockers/docker-base-bookworm/bin/wrapper_logger.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+# This is a wrapper script for logger to take -i option as the default
+# If user provides -id or --id then that option takes precedence
+
+/usr/bin/logger -i "$@"
```
theasianpianist, line 6: Using just `-i` by itself would run into the same problem that we currently have since each call of the wrapper script would have a unique PID. I think we need to call it as 
```
/usr/bin/logger --id="$PPID" "$@"
```
So that the parent script PID is used.

files/image_config/bash/wrapper_logger.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
```
theasianpianist, line 1: Do we need two copies of the same file? Can we reference the same file for both the host OS and docker containers, or symlink these two paths so that there is only one copy?

theasianpianist (changes requested):

dockers/docker-base-bookworm/bin/wrapper_logger.sh
prgeor: @theasianpianist `logger` has two options to specify the PID :- 
1. `-id`  This option specify that the logger will use the PID of the process which invoked the logger command. So two difference process A and B invoking logger will use their respective PID so the rsyslog will have rate limiting fairness for both

2. when you use `--d=$PPID`  two different process A and B will have same PPID which i think is the bash. Using this option is not recommended because it will not be fair for both process A and B

files/image_config/bash/wrapper_logger.sh
prgeor: @theasianpianist docker builds cannot refer to file anything but inside docker/ dir. So the only option is to refer to docker/ file in `files/build_templates/sonic_debian_extension.j2` which IMO is not neat. Keeping this file indepenent for host and docker build looks to be necessary.

lguohan:
this increases the overhead, i would suggest to develop a bash lib, and use source to include the function and so that multiple script can use the new logger function defined in the library.

lguohan (changes requested):
would like you to change the approach to use source to do it.

prgeor:
closing after discussion with Guohan.

PR #19444, opened 2024-07-02
fix non canonical format of IPv6 address string in bgpcfgd

chenkelly:
#### Why I did it
IPv6 addresses text used in cfgdb may not follow the canonical IPv6 text format. When performing IPv6 addresses equivalent check with string, need to make sure that both of them are in the shortest form defined in rfc 5952

### DIFF
--- a/src/sonic-bgpcfgd/bgpcfgd/managers_bgp.py
+++ b/src/sonic-bgpcfgd/bgpcfgd/managers_bgp.py
@@ -324,6 +324,15 @@ class BGPPeerMgrBase(Manager):
                  If the interface has not been set, return None
         """
         local_addresses = self.directory.get_slot("LOCAL", "local_addresses")
+
+        # Handle local_addr string that is not represented in canonical text format, especially for IPv6 addresses
+        try:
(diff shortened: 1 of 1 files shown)


src/sonic-bgpcfgd/bgpcfgd/managers_bgp.py
```diff
@@ -324,6 +324,15 @@ def get_local_interface(self, local_addr):
...
         local_addresses = self.directory.get_slot("LOCAL", "local_addresses")
+
+        # Handle local_addr string that is not represented in canonical text format, especially for IPv6 addresses
+        try:
+            ip = netaddr.IPAddress(str(local_addr))
+        except (netaddr.NotRegisteredError, netaddr.AddrFormatError, netaddr.AddrConversionError):
+            log_warn("IP Address '%s' format is wrong" % (local_addr))
+            return None
+        local_addr = str(ip)
+
```
lguohan, line 335: @chenkelly , can you also add unit test for this code?

chenkelly: Ok, I'll check and see how to add it.

PR #19482, opened 2024-07-05
Made pmon feature delayed flag as jinja template

abdosi:
What I did:
To fix: https://github.com/sonic-net/sonic-buildimage/issues/19126
Depends on: https://github.com/sonic-net/sonic-host-services/pull/135 and it's submodule update.

Why I did:
PMON was delayed originally for faster uptime of data plane related services in case of fast/warm reboot.
This is not needed for T2/SpineRouter . Infact we need PMON to be up asap because of:

-   pmon need to enable asap to detect ASIC's on Supervisor.
-   pmon need to enable asap for bring-up of 400G ports on LC's fast because of CMIS state machine present in PMON.

How I verify:
Manual Verification and UT has been added in  https://github.com/sonic-net/sonic-host-services/pull/135


abdosi:
@anamehra @mlok-nokia for viz.

arlakshm (approved):

abdosi:
@arlakshm : please reapprove this.

arlakshm (dismissed):

files/build_templates/init_cfg.json.j2
```diff
@@ -38,7 +38,7 @@
 {%- set features = [("bgp", "{% if not DEVICE_RUNTIME_METADATA['ETHERNET_PORTS_PRESENT'] or ('CHASSIS_METADATA' in DEVICE_RUNTIME_METADATA and DEVICE_RUNTIME_METADATA['CHASSIS_METADATA']['module_type'] in ['supervisor']) %}disabled{% else %}enabled{% endif %}", false, "enabled"),
                    ("database", "always_enabled", false, "always_enabled"),
                    ("lldp", "enabled", true, "enabled"),
-                   ("pmon", "enabled", true, "enabled"),
+                   ("pmon", "enabled", "{% if 'type' in DEVICE_METADATA['localhost'] and DEVICE_METADATA['localhost']['type'] == 'SpineRouter' %}false{% else %}true{% endif %}", "enabled"),
```
mlok-nokia, line 41: This PR works with PR https://github.com/sonic-net/sonic-host-services/pull/135
But delayed value "true/false" are the match the expected value which in the featured line 79.  The defined value should be "True/False".
This change causes the system fails to start PMON

yejianquan:
Fix the issue and advance sonic-host-service sub-module
https://github.com/sonic-net/sonic-buildimage/pull/19551

files/build_templates/init_cfg.json.j2
yejianquan: Good catch, I suppose it's because of the removal of 'lower()' function, not sure why Abhishek changed this behavior.
Let me draft another PR and see.

yejianquan: https://github.com/sonic-net/sonic-buildimage/pull/19657

yejianquan:
https://github.com/sonic-net/sonic-buildimage/pull/19657 merged, closing this one

PR #19492, opened 2024-07-05
Add netconf docker to build-image

hussamsh:
#### Why I did it
I wanted to add NETCONF server support for configuring SONiC switches. Similar to the work done on RESTCONF server.

#### How I did it
First, the changes themselves are on the [`Official sonic netconf repositroy`](https://github.com/sonic-net/sonic-netconf-server).
This work focuses on integrating the netconf server in build-image processes. I updated relevant build scripts , dockerfiles and git modules to include the netconf server in the final image.

#### How to verify it
1) Normally build the image
2) Install new image with netconf server on a switch. After switch stabilizes inspect the docker to see the netconf-server docker image is up.
3) Test NETCONF connection using `ssh <username>@<ip> -p 830 -s netconf`

This change adds support for a RFC compliant NETCONF server for SONiC switches.

Implements sonic-net/SONiC#1466
Implements sonic-net/SONiC#1498
Uses sonic-net/sonic-netconf-server#2


hussamsh:
Hello @zhangyanzhao @aseaudi 

This is the remaining part of the NETCONF server implementation, integration in the sonic-build-image.

For your info

hussamsh:
@lguohan 
Work on this is almost done, can we get some feedback if we can merge this feature ?

EmmaLin11:
Hi @hussamsh,

I was doing some local validation and encountered a couple of points I'd like to share and get your insights on.

**netconf-server Panic (with PR's specified version)**

Firstly, I noticed this PR seems to be based on sonic-netconf-server commit 24db77. When I compiled and tested with this older version as specified in the PR, the compilation process generally went smoothly. However, the netconf-server itself panics at runtime, which prevents the service from starting and being used effectively.

Specifically, the netconf-server starts successfully, but the panic occurs immediately after attempting to connect via ssh -s netconf.

Here's the relevant section from the logs showing the panic:
```
2025 Jul  1 08:29:28.065509 sonic INFO netconf-server#netconf-server#supervisord: netconf-server panic: runtime error: invalid memory address or nil pointer dereference
2025 Jul  1 08:29:28.065953 sonic INFO netconf-server#netconf-server#supervisord: netconf-server [signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x100352e]
2025 Jul  1 08:29:28.065953 sonic INFO netconf-server#netconf-server#supervisord: netconf-server
2025 Jul  1 08:29:28.065953 sonic INFO netconf-server#netconf-server#supervisord: netconf-server goroutine 69
2025 Jul  1 08:29:28.065953 sonic INFO netconf-server#netconf-server#supervisord: netconf-server  [running]:
2025 Jul  1 08:29:28.065953 sonic INFO netconf-server#netconf-server#supervisord: netconf-server orange/sonic-netconf-server/netconf/server.capabilitesXML()
2025 Jul  1 08:29:28.065953 sonic INFO netconf-server#netconf-server#supervisord: netconf-server #011/sonic/src/sonic-netconf-server/netconf/server/handler.go:95 +0x3ce
2025 Jul  1 08:29:28.065953 sonic INFO netconf-server#netconf-server#supervisord: netconf-server orange/sonic-netconf-server/netconf/server.SessionHandler({0x12dcec0, 0xc0001d5e10})
2025 Jul  1 08:29:28.065953 sonic INFO netconf-server#netconf-server#supervisord: netconf-server #011/sonic/src/sonic-netconf-server/netconf/server/handler.go:50 +0x8b
2025 Jul  1 08:29:28.065953 sonic INFO netconf-server#netconf-server#supervisord: netconf-server github.com/gliderlabs/ssh.(*session).handleRequests.func2()
2025 Jul  1 08:29:28.066020 sonic INFO netconf-server#netconf-server#supervisord: netconf-server #011/sonic/src/sonic-netconf-server/vendor/github.com/gliderlabs/ssh/session.go:296 +0x3c
2025 Jul  1 08:29:28.066020 sonic INFO netconf-server#netconf-server#supervisord: netconf-server created by github.com/gliderlabs/ssh.(*session).handleRequests
2025 Jul  1 08:29:28.066064 sonic INFO netconf-server#netconf-server#supervisord: netconf-server #011/sonic/src/sonic-netconf-server/vendor/github.com/gliderlabs/ssh/session.go:295 +0x1294
2025 Jul  1 08:29:28.070186 sonic INFO netconf-server#netconf-server#supervisord 2025-07-01 08:29:28,069 WARN exited: netconf-server (exit status 2; not expected)
2025 Jul  1 08:29:29.075018 sonic INFO netconf-server#netconf-server#supervisord 2025-07-01 08:29:29,074 INFO spawned: 'netconf-server' with pid 56
```
---
**Observations on profile Version and Latest sonic-netconf-server**

Additionally, I also tried to test this PR by aligning my environment with the latest sonic-netconf-server (commit ea7fc99). In this scenario, I encountered a profile module dependency issue. The error message indicated that github.com/Azure/sonic-mgmt-common requires github.com/pkg/profile@v1.7.0, but the necessary go.sum entry was missing, halting the build process.

Here's the relevant error log for your reference:
```
ubuntu$ make target/docker-sonic-netconf-server.gz
...
go: downloading github.com/godbus/dbus/v5 v5.1.0
go: downloading google.golang.org/genproto v0.0.0-20200319113533-08878b785e9c
go: github.com/Azure/sonic-mgmt-common@v0.0.0-20220120155510-515652700481 requires
        github.com/pkg/profile@v1.7.0: missing go.sum entry; to add it:
        go mod download github.com/pkg/profile
make[2]: *** [Makefile:43: vendor/.done] Error 1
make[2]: Leaving directory '/sonic/src/sonic-netconf-server'
dh_auto_build: error: make -j1 returned exit code 2
make[1]: *** [debian/rules:3: build] Error 25
...
```

As a preliminary test, I manually ran go mod download github.com/pkg/profile, which successfully pulled profile@v1.7.0, and this specific compilation error was resolved, allowing the build to proceed.
```
ubuntu@e676a50347e4:/sonic/src/sonic-netconf-server$ go mod download github.com/pkg/profile
ubuntu@e9b6dc745f29:/sonic/src/sonic-netconf-server$ go get github.com/Azure/sonic-mgmt-common/translib/ocbinds@v0.0.0-20220120155510-515652700481
ubuntu@e9b6dc745f29:/sonic/src/sonic-netconf-server$ go get github.com/Azure/sonic-mgmt-common/translib/transformer@v0.0.0-20220120155510-515652700481
```

Based on these observations, it might be worth considering updating this PR to use the latest sonic-netconf-server (commit ea7fc99). This would help avoid the runtime panic observed with the older 24db77 commit.

If we move to the latest sonic-netconf-server commit, we would then need to address the profile module dependency issue (e.g., by running go mod tidy to update go.mod and go.sum files). This seems like a more robust path forward, as it resolves the critical runtime issue and keeps the component up-to-date.

msosyak:
The HLD https://github.com/sonic-net/SONiC/pull/1466  states that NECONF SERVR will support the following operations:
```
###### 3.2.2.11.9 NETCONF Operations
```
```
Netconf server supports the following operations:
* <get> (RFC 6241 Section 7.7)
* <get-config> (RFC 6241 Section 7.1)
* <copy-config> (RFC 6241 Section 7.3)
* <lock> (RFC 6241 Section 7.5)
* <unlock> (RFC 6241 Section 7.6)
* <close-session> (RFC 6241 Section 7.8)
* <kill-session> (RFC 6241 Section 7.9)
* <get-schema> (RFC 6022 Section 3.1)
```
But in the implementation https://github.com/sonic-net/sonic-netconf-server/blob/master/netconf/server/handler.go#L167 only `get` and `close-session` are mentioned. 

Will the full implementation, as described in the HLD, be contributed to the community? If so, when?

Or do I miss something and the `urn:ietf:params:netconf:capability:writable-running:1.0` is implemented in some different way?

hussamsh:
Hi @EmmaLin11 ,

Thanks for the feedback , I will update the PR shortly to reflect the latest commit and test the new image. As for the module dependency issue it seems to be building correctly in the CI/CD pipeline, nevertheless I will test this locally and will create a PR on the sonic-mgmt-common if the issue appears for me as this is not related to the netconf itself but rather that the mgmt-common master branch is not building correctly.

hussamsh:
@msosyak 

I have already implemented all of these RPCs locally , what I am sharing today is a minimal implementation with `<get>` and `<close-session>` only to start the integration of NETCONF into SONiC. 

Once NETCONF is accepted into SONiC I will update the sonic-netconf-server submodule with all of the missing RPCs. 

Do you think it is better to share the entire NETCONF implementation from the start ?

msosyak:
@hussamsh  Thank you for the clarification. Since the implementation resides in a separate repository, its completeness doesn’t directly affect the review of this PR. However, including the full implementation from the start could help facilitate broader discussions and a more comprehensive assessment of the overall NETCONF integration into SONiC.

EmmaLin11:
Hi @hussamsh,

Thanks for the quick update and for looking into this!

That's great to hear you'll be updating the PR to the latest sonic-netconf-server commit. This should definitely resolve the runtime panic we observed with the older version.

On a related note regarding sonic-mgmt-common, I've observed that the later commits of sonic-netconf-server seem to call translib.GetYanglibInfo. This function might not be exported in the current sonic-mgmt-common master branch. So, if we move to the latest sonic-netconf-server, it looks like another PR might be needed in sonic-mgmt-common to export translib.GetYanglibInfo to ensure proper build and functionality.

hussamsh:
@EmmaLin11 
[PR](https://github.com/sonic-net/sonic-mgmt-common/pull/180) for the mgmt-common created. Pending review by the mgmt-common reviewers.

EmmaLin11:
> [PR](https://github.com/sonic-net/sonic-mgmt-common/pull/180) for the mgmt-common created. Pending review by the mgmt-common reviewers.

Hi @hussamsh,

Thanks for the update! 👍

hussamsh:
@lguohan @xumia can we get any updates about the merge of netconf on sonic ?

pmc8893:
@hussamsh Does sonic netconf support Netopeer2? I haven't seen any tests for it.

rules/docker-sonic-netconf-server.mk
```diff
@@ -0,0 +1,38 @@
...
+SONIC_INSTALL_DOCKER_IMAGES += $(DOCKER_NETCONF_SERVER)
+endif
+
+SONIC_DOCKER_DBG_IMAGES += $(DOCKER_NETCONF_SERVER_DBG)
+ifeq ($(INCLUDE_NETCONF_SERVER), y)
+SONIC_INSTALL_DOCKER_DBG_IMAGES += $(DOCKER_NETCONF_SERVER_DBG)
+endif
+
+$(DOCKER_NETCONF_SERVER)_CONTAINER_NAME = netconf-server
+$(DOCKER_NETCONF_SERVER)_RUN_OPT += -v /etc/sonic:/etc/sonic:rw
```
qiluo-msft: This is security concern. why you need it as rw?

rules/docker-sonic-netconf-server.mk
```diff
@@ -0,0 +1,38 @@
...
+$(DOCKER_NETCONF_SERVER)_RUN_OPT += -v /etc:/host_etc:ro
```
qiluo-msft: This is security concern. why you need it at all?

rules/docker-sonic-netconf-server.mk
```diff
@@ -0,0 +1,38 @@
...
+$(DOCKER_NETCONF_SERVER)_RUN_OPT += -v /var/run/dbus:/var/run/dbus:rw
```
qiluo-msft, line 33: Could you explain which functionality require `/var/run/dbus`?

rules/config
```diff
@@ -183,6 +183,9 @@ INCLUDE_SFLOW = y
 # INCLUDE_MGMT_FRAMEWORK - build docker-sonic-mgmt-framework for CLI and REST server support
 INCLUDE_MGMT_FRAMEWORK = y
 
+# INCLUDE_NETCONF_SERVER - build docker-sonic-netconf server for NETCONF support
+INCLUDE_NETCONF_SERVER = y
```
qiluo-msft: Could it defaulto to n?

.github/workflows/protect-file.yml
```diff
@@ -24,7 +24,7 @@ jobs:
       - name: Check protected files
         run: |
           echo "Starting file check step"
-          protected_files=('src/sonic-config-engine/minigraph.py' 'files/build_templates/init_cfg.json.j2')
+          protected_files=('src/sonic-config-engine/minigraph.py')
           base_branch=$(git rev-parse origin/${GITHUB_BASE_REF})
```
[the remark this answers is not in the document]
hussamsh: the `init_cfg.json.j2` needs to be updated so that the netconf feature can be added. I can re-add the protected file restriction after netconf is added.

rules/docker-sonic-netconf-server.mk
hussamsh: The RPC functionality calls functions exposed by the dbus - defined in the `sonic-host-services` .  This option opens communication with dbus on the host.

rules/docker-sonic-netconf-server.mk
hussamsh: When the NETCONF first starts it needs to generate SSH keys for authentication. One edge case we found in operation is that if the NETCONF server restarts for any reason it would regenerate the key and all the existing clients using NETCONF will have to re-add it locally. That is why when the server starts the first time it checks for the `/etc/sonic/netconf-key.pub` to see if there an existing key, if so use it, if not generate a new key and add there for future use. This makes sure that any client using NETCONF server needs to add the key once across any reboot that might occur.

rules/config
hussamsh: yes of course will update this to default `n`.

rules/docker-sonic-netconf-server.mk
hussamsh: It is needed so because the default authentication method is local authentication - meaning username / password of the host - so this needs to be mounted inside the docker such that local authentication could happen ? Can you suggest a better approach for this ?

hussamsh:
Hi @qiluo-msft , I just wanted to follow up on the comments you left. if you could take another look so I can move forward with the PR. Thanks a lot!

rookie-who:
Hi @hussamsh — this has been in progress for a while (127 comments). The 202605 branch cut is at end of May. What's the status? Is additional review needed?

CC: @vaibhavhd @radha-danda (branch managers)

hussamsh:
Hi @rookie-who , PR is done, I had a review from @qiluo-msft and I addressed his issues promptly. If you can nudge any reviewer to finalize the review I would be so appreciative. PR is ready to be merged. No pending point from my side. 

Thank you !

rules/docker-sonic-netconf-server.mk
qiluo-msft: A compromised or buggy NETCONF container would expose info, such as /etc/shadow. Could you use more secure authentication, such as cert based authen?

qiluo-msft: `/etc/sonic:rw` is also weird, why need to modify this folder? some subfolder has credential, could you not to mount all of them?

dockers/docker-sonic-netconf-server/netconf-server.sh
```diff
@@ -0,0 +1,34 @@
...
+
+NETCONF_SERVER_ARGS="-logtostderr"
+[ -n "$SERVER_PORT" ] && NETCONF_SERVER_ARGS+=" -port $SERVER_PORT"
+[ -n "$LOG_LEVEL"   ] && NETCONF_SERVER_ARGS+=" -v $LOG_LEVEL"
+
+echo "NETCONF_SERVER_ARGS = $NETCONF_SERVER_ARGS"
+
+export CVL_SCHEMA_PATH=/usr/sbin/schema
+
+exec /usr/sbin/netconf_server ${NETCONF_SERVER_ARGS}
```
qiluo-msft: The unquoted expansion of NETCONF_SERVER_ARGS undergoes word splitting. A CONFIG_DB writer can supply `port: "830 --any-flag value"` and inject arbitrary netconf_server flags.

files/build_templates/init_cfg.json.j2
```diff
@@ -78,6 +78,7 @@
 {%- if sonic_asic_platform == "vs" %}{% do features.append(("gbsyncd", "enabled", false, "enabled")) %}{% endif %}
 {%- if include_iccpd == "y" %}{% do features.append(("iccpd", "disabled", false, "enabled")) %}{% endif %}
 {%- if include_mgmt_framework == "y" %}{% do features.append(("mgmt-framework", "enabled", true, "enabled")) %}{% endif %}
+{%- if include_netconf_server == "y" %}{% do features.append(("netconf-server", "enabled", true, "enabled")) %}{% endif %}
```
qiluo-msft: Do not change image default behavior. You can use /etc/sonic/config_db.json FEATURE table.

files/build_templates/netconf-server.timer
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=Delays netconf-server container until SONiC has started
+PartOf=netconf-server.service
+
+[Timer]
+OnUnitActiveSec=0 sec
```
qiluo-msft: Recommend removing `OnUnitActiveSec`, which means tight restart loops on service failure

files/build_templates/netconf-server.service.j2
```diff
@@ -1,17 +1,13 @@
 [Unit]
-Description=Redfish container
+Description=Netconf Server container
+Requires=database.service
+After=database.service swss.service syncd.service
```
qiluo-msft: syncd.service doesn't exist on chassis supervisor cards / vs-platform (only on line cards/single-asic). The new feature only depends on database, right?

files/build_templates/netconf-server.service.j2
```diff
@@ -1,17 +1,13 @@
...
 BindsTo=sonic.target
 After=sonic.target
-StartLimitIntervalSec=1200
-StartLimitBurst=3
+Before=ntp-config.service
```
qiluo-msft: help me understand why before ntp?

hussamsh: No specific reason, removed it.

hussamsh:
@qiluo-msft , I addressed your review points , could you please take a look ? 
cc : @rookie-who, @vaibhavhd , @radha-danda

securely1g (commented):
Review on behalf of @lguohan:

This PR has been open since July 2024 (over a year). Questions:

1. Is the netconf server still being actively pursued for SONiC? The community has largely standardized on gNMI.
2. The PR modifies .gitmodules and adds a new docker — what is the maintenance commitment?
3. Has this been rebased recently? Given the age, there are likely significant merge conflicts.

Please provide a status update on relevance of this PR.

hussamsh:
Hello @lguohan, regarding your questions.

1) NETCONF is still one of the main configuration management servers in the telecom side of the world, with great support from industry tools such as CISCO NSO. We use it extensively in our production environments and I know many other telecom providers that exclusively use NETCONF for managing their production environments. I actually developed NETCONF because our network team didn't want to deploy it unless it had NETCONF so that their existing tools and workflows can talk with SONiC. 

2) Orange is committed to provide support for this module in the future. At least one developer will be responsible for keeping this management server updated.

3) Currently no merge conflicts and I am merging into sonic-net:master.

dockers/docker-sonic-netconf-server/Dockerfile.j2
```diff
@@ -0,0 +1,34 @@
...
+debs/{{ deb }}{{' '}}
+{%- endfor -%}
+debs/
+
+RUN dpkg -i \
+{% for deb in docker_sonic_netconf_server_debs.split(' ') -%}
+debs/{{ deb }}{{' '}}
+{%- endfor %}
+
+COPY ["start.sh", "netconf-server.sh", "/usr/bin/"]
```
nakano-omw, line 26: Heads-up @hussamsh — while bringing up the netconf-server container from this PR I hit a blocker: the container13:02:23 [49/1956]
 netconf_server never reaches its :830 listen socket, because the two helper scripts are installed without an executable bit.

 This is not specific to our downstream patches. It reproduces on a pristine baseline = this PR's head (070da4f96) + INCLUDE_NETCONF_SERVER=y, no other changes.

```
 $ docker exec netconf-server supervisorctl status
 netconf-server                   STOPPED   Not started
start                            FATAL     command at '/usr/bin/start.sh' is not executable

 $ docker exec netconf-server ls -la /usr/bin/start.sh /usr/bin/netconf-server.sh
 -rw-rw-r-- 1 root root 1037 /usr/bin/netconf-server.sh
-rw-rw-r-- 1 root root  121 /usr/bin/start.sh

$ docker logs netconf-server 2>&1 | grep "not executable"
 INFO spawnerr: command at '/usr/bin/start.sh' is not executable
 INFO gave up: start entered FATAL state, too many start retries too quickly
```

Root cause
dockers/docker-sonic-netconf-server/Dockerfile.j2 does:
```
COPY ["start.sh", "netconf-server.sh", "/usr/bin/"]
```

Both scripts are tracked in Git as mode 100644. docker build preserves the source mode on COPY, so the in-container files inherit
-rw-rw-r--, and supervisord (which checks the exec bit before spawning) refuses them.

 Suggested fix — one line after the COPY:
```
COPY ["start.sh", "netconf-server.sh", "/usr/bin/"]
RUN chmod 0755 /usr/bin/start.sh /usr/bin/netconf-server.sh
COPY ["mgmt_vars.j2", "/usr/share/sonic/templates/"]
```

This makes the in-container mode deterministic regardless of the source-tree file mode — the same pattern already used by other
SONiC dockers. After this, the container comes up on its own:

```
$ docker exec netconf-server ss -tlnp | grep 830
LISTEN 0 512 *:830 *:* users:(("netconf_server",pid=127,fd=7))
$ docker exec netconf-server ls /etc/sonic/netconf-key*
/etc/sonic/netconf-key  /etc/sonic/netconf-key.pub
```

rookie-who:
This PR/feature will not be included in the SONiC 202605 release. It can be retargeted for the 202611 release.

qiluo-msft (commented):
A few follow-ups — mostly on the `sonic-netconf-server` submodule (host key + auth), plus inline comments on the build-image files.

**Host key handling (`sonic-netconf-server`, `netconf/main/main.go`):**
- Permissions: the private key is written via `os.Create(...)` → effectively `0644` (world-readable). It must be `0600` (public key `0644`).
- Algorithm: currently hardcoded RSA-1024. Please make the algorithm/size configurable (e.g. a flag fed from `mgmt_vars.j2`, like `port`/`log_level`) with a strong default — ed25519 or RSA ≥ 2048.
- Path: no need to make it runtime-configurable, but please relocate the hardcoded `/etc/sonic/netconf-key` path into a dedicated subdir (e.g. `/etc/sonic/netconf/`) so the host mount can be scoped (see the `.mk` comment).
- Re: reusing the host's main sshd key — I'd avoid it. Mounting the host's primary `ssh_host_*_key` into this container widens the blast radius to all SSH on the box if the container is compromised. A dedicated, well-permissioned key is better.

**Auth model (`sonic-netconf-server`, `netconf/server/auth.go`):**
- Reachability: `Authenticate()` does `ssh.Dial("127.0.0.1:22", ...)`, but I don't see `--network host` in the run options or an sshd in `supervisord.conf`. How does this loopback reach the host sshd? Please clarify (and if it relies on host networking, call that out as a separate consideration).
- Side effects: authenticating by spawning a real SSH login means every NETCONF auth shows up as a localhost sshd login in the audit log plus added latency, and `Authorize()`/`Account()` both return `true` (no RBAC — any valid host user gets full `get`). Worth a follow-up plan even if not blocking v1.

rules/docker-sonic-netconf-server.mk
```diff
@@ -0,0 +1,37 @@
...
+SONIC_INSTALL_DOCKER_IMAGES += $(DOCKER_NETCONF_SERVER)
+endif
+
+SONIC_DOCKER_DBG_IMAGES += $(DOCKER_NETCONF_SERVER_DBG)
+ifeq ($(INCLUDE_NETCONF_SERVER), y)
+SONIC_INSTALL_DOCKER_DBG_IMAGES += $(DOCKER_NETCONF_SERVER_DBG)
+endif
+
+$(DOCKER_NETCONF_SERVER)_CONTAINER_NAME = netconf-server
+$(DOCKER_NETCONF_SERVER)_RUN_OPT += -v /etc/sonic:/etc/sonic:rw
```
qiluo-msft: `-v /etc/sonic:/etc/sonic:rw` is too broad — `/etc/sonic` holds credentials, and a compromised NETCONF container would get write access to all of it. The only thing needing persistence here is the SSH host key. Please scope this to a dedicated subdir, e.g. `-v /etc/sonic/netconf:/etc/sonic/netconf:rw`.

Note this is coupled with the server binary: the key path is currently hardcoded to `/etc/sonic/netconf-key` in `sonic-netconf-server` (`netconf/main/main.go`), so scoping the mount requires relocating that hardcoded path into the same subdir (it doesn't need to be runtime-configurable — just moved), then bumping the submodule pin.

dockers/docker-sonic-netconf-server/mgmt_vars.j2
```diff
@@ -0,0 +1,5 @@
+{
+    "netconf_server": {% if NETCONF_SERVER is defined and "default" in NETCONF_SERVER.keys() %}{{ NETCONF_SERVER['default'] }}{% else %}""{% endif %},
+    "x509" : {% if "x509" in DEVICE_METADATA.keys() %}{{ DEVICE_METADATA["x509"] }}{% else %}""{% endif %},
```
qiluo-msft: `x509` and `tacacs_plus` are rendered here but are not consumed anywhere in the server binary at the pinned commit (only `port` and `log_level` are read by `netconf-server.sh`). This implies a security capability (cert auth / TACACS) that doesn't exist. Please either wire them up or remove them until they're implemented, so the config isn't misleading.

files/build_templates/init_cfg.json.j2
```diff
@@ -78,6 +78,7 @@
 {%- if sonic_asic_platform == "vs" %}{% do features.append(("gbsyncd", "enabled", false, "enabled")) %}{% endif %}
 {%- if include_iccpd == "y" %}{% do features.append(("iccpd", "disabled", false, "enabled")) %}{% endif %}
 {%- if include_mgmt_framework == "y" %}{% do features.append(("mgmt-framework", "enabled", true, "enabled")) %}{% endif %}
+{%- if include_netconf_server == "y" %}{% do features.append(("netconf-server", "disabled", true, "enabled")) %}{% endif %}
```
qiluo-msft: No need to add `netconf-server` to `init_cfg.json.j2` — please revert this. The feature can be enabled at runtime via the `FEATURE` table in `/etc/sonic/config_db.json` without baking it into the image default.

Since `init_cfg.json.j2` then no longer changes, the removal of it from `.github/workflows/protect-file.yml` is also unnecessary — please revert that too and keep the guardrail intact (see the comment there).

.github/workflows/protect-file.yml
```diff
@@ -24,7 +24,7 @@ jobs:
       - name: Check protected files
         run: |
           echo "Starting file check step"
-          protected_files=('src/sonic-config-engine/minigraph.py' 'files/build_templates/init_cfg.json.j2')
+          protected_files=('src/sonic-config-engine/minigraph.py')
```
qiluo-msft: Removing `init_cfg.json.j2` from the protected list drops a review guardrail to land this change. Since the `init_cfg.json.j2` edit isn't needed (configure the feature via `/etc/sonic/config_db.json` instead), please revert this line and keep the protection intact.

hussamsh:
@qiluo-msft, regarding your review:
1) I have revamped the Host key handling in this PR, your comments should now be addressed and the security hardened in this new version
2) Auth model:
- Readability: I am relying on host networking since netconf docker is launched with the `NET="host"`  flag.
- Side effects: This PR does not introduce the authentication model or broaden authorization behavior. The `Authorize()` / `Authenticate()` can be tracked separately. Since I am using the password auth path unchanged, this should not block this PR. I could create a separate issue for this if you want.

rules/docker-sonic-netconf-server.mk
```diff
@@ -0,0 +1,37 @@
...
+SONIC_INSTALL_DOCKER_IMAGES += $(DOCKER_NETCONF_SERVER)
+endif
+
+SONIC_DOCKER_DBG_IMAGES += $(DOCKER_NETCONF_SERVER_DBG)
+ifeq ($(INCLUDE_NETCONF_SERVER), y)
+SONIC_INSTALL_DOCKER_DBG_IMAGES += $(DOCKER_NETCONF_SERVER_DBG)
+endif
+
+$(DOCKER_NETCONF_SERVER)_CONTAINER_NAME = netconf-server
+$(DOCKER_NETCONF_SERVER)_RUN_OPT += -v /etc/sonic/netconf:/etc/sonic/netconf:rw
```
qiluo-msft, line 32: the mount was narrowed from the full `/etc/sonic` to `/etc/sonic/netconf` — that's better. follow-up: does the server actually write to this directory, or is it read-only config? if the container only reads its config from here, `:ro` is sufficient and limits blast radius from a compromised container.

dockers/docker-sonic-netconf-server/netconf-server.sh
```diff
@@ -0,0 +1,41 @@
...
+[ -z "$HOST_KEY_ALGORITHM" ] && HOST_KEY_ALGORITHM=ed25519
+[ -z "$HOST_KEY_RSA_BITS"  ] && HOST_KEY_RSA_BITS=3072
+
+NETCONF_SERVER_ARGS=("-logtostderr")
+[ -n "$SERVER_PORT" ] && NETCONF_SERVER_ARGS+=("-port" "$SERVER_PORT")
+[ -n "$LOG_LEVEL"   ] && NETCONF_SERVER_ARGS+=("-v" "$LOG_LEVEL")
+[ -n "$HOST_KEY_ALGORITHM" ] && NETCONF_SERVER_ARGS+=("-host_key_algorithm" "$HOST_KEY_ALGORITHM")
+[ -n "$HOST_KEY_RSA_BITS"  ] && NETCONF_SERVER_ARGS+=("-host_key_rsa_bits" "$HOST_KEY_RSA_BITS")
+
+echo "NETCONF_SERVER_ARGS = ${NETCONF_SERVER_ARGS[*]}"
```
qiluo-msft, line 37: nit: `${NETCONF_SERVER_ARGS[*]}` unquoted — word-splits on spaces in any element. harmless here (just logging), but inconsistent with the correctly-quoted `"${NETCONF_SERVER_ARGS[@]}"` on line 41. use `"${NETCONF_SERVER_ARGS[*]}"` or `"${NETCONF_SERVER_ARGS[@]}"`.

rules/docker-sonic-netconf-server.mk
```diff
@@ -0,0 +1,37 @@
...
+DOCKER_NETCONF_SERVER_DBG = $(DOCKER_NETCONF_SERVER_STEM)-$(DBG_IMAGE_MARK).gz
+
+$(DOCKER_NETCONF_SERVER)_PATH = $(DOCKERS_PATH)/$(DOCKER_NETCONF_SERVER_STEM)
+
+$(DOCKER_NETCONF_SERVER)_DEPENDS += $(SONIC_MGMT_COMMON)
+$(DOCKER_NETCONF_SERVER)_DEPENDS += $(SONIC_NETCONF_SERVER)
+$(DOCKER_NETCONF_SERVER)_DBG_DEPENDS = $($(DOCKER_CONFIG_ENGINE_BOOKWORM)_DBG_DEPENDS)
+$(DOCKER_NETCONF_SERVER)_DBG_DEPENDS += $(SONIC_NETCONF_SERVER_DBG)
+
+SONIC_DOCKER_IMAGES += $(DOCKER_NETCONF_SERVER)
```
qiluo-msft, line 14: `SONIC_DOCKER_IMAGES += $(DOCKER_NETCONF_SERVER)` runs unconditionally, but `$(SONIC_NETCONF_SERVER)` (which the docker depends on, line 10) is only defined when `INCLUDE_NETCONF_SERVER = y` (it's wrapped in an `ifeq` in `sonic-netconf-server.mk`). when `INCLUDE_NETCONF_SERVER = n`, the docker is added to the build list but its dependent deb doesn't exist → the docker build will fail. either wrap this line in `ifeq ($(INCLUDE_NETCONF_SERVER), y)` alongside the INSTALL lines, or make the package always build (like `SONIC_MGMT_COMMON`).

Issue #18695, opened 2024-04-17
Unable to change PortChannel mtu size


mbze430:
#### Description

after creating a portchannel interface changing the portchannel mtu I get a message that the portchannel doesn't exist

#### Steps to reproduce the issue:
1. admin@sonic:~$ sudo config portchannel add PortChannel02
2. admin@sonic:~$ sudo config interfaces mtu Ethernet80 9216
3. admin@sonic:~$ sudo config interfaces mtu Ethernet84 9216
4. admin@sonic:~$ sudo config interface mtu PortChannel02 9216
admin@sonic:~$ show interface portchannel
Flags: A - active, I - inactive, Up - up, Dw - Down, N/A - not available,
       S - selected, D - deselected, * - not synced
  No.  Team Dev       Protocol     Ports
-----  -------------  -----------  -----------------------------
   01  PortChannel01  LACP(A)(Up)  Ethernet104(S) Ethernet108(S)
   02  PortChannel02  LACP(A)(Dw)

#### Describe the results you received:
Invalid port PortChannel02

#### Describe the results you expected:
The portchannel mtu to be changed to 9216

#### Output of `show version`:

```
SONiC Software Version: SONiC.202305.505816-7b32c39cc
SONiC OS Version: 11
Distribution: Debian 11.8
Kernel: 5.10.0-23-2-amd64
Build commit: 7b32c39cc
Build date: Sat Mar 23 13:42:14 UTC 2024
Built by: AzDevOps@vmss-soni003BY6

Platform: x86_64-cel_seastone-r0
HwSKU: Seastone-DX010
ASIC: broadcom
ASIC Count: 1
Serial Number: DX010B2F108423LK100045
Model Number: R0872-F0010-01
Hardware Revision: N/A
Uptime: 22:13:19 up 3 days,  8:10,  1 user,  load average: 1.56, 1.97, 2.09
Date: Tue 16 Apr 2024 22:13:19

Docker images:
REPOSITORY                    TAG                       IMAGE ID       SIZE
docker-gbsyncd-broncos        202305.505816-7b32c39cc   291d5fd0b380   350MB
docker-gbsyncd-broncos        latest                    291d5fd0b380   350MB
docker-gbsyncd-credo          202305.505816-7b32c39cc   6a015c2e8a0c   322MB
docker-gbsyncd-credo          latest                    6a015c2e8a0c   322MB
[...]

PR #19558, opened 2024-07-12
Preserve permissions when copying

linqingxuan:
#### What I did

Fix for [#18695](https://github.com/sonic-net/sonic-buildimage/issues/18695)

#### How I did it

When using cp, add the -p option to preserve permissions

#### How to verify it

can make it

#### Previous command output (if the output of a command-line utility has changed)

Before this fix, you may find that you may not have permission when compiling arm

lin@lin-arm64:~/build$ ls -lrt sonic-buildimage.centec-arm64/target/files/bullseye/update_chassisdb_config
-rw-r--r-- 1 lin lin 2467 Jul 12 06:41 sonic-buildimage.centec-arm64/target/files/bullseye/update_chassisdb_config

#### New command output (if the output of a command-line utility has changed)

After the repair, the permissions can retain the previous execution permissions

lin@lin-arm64:~/build$ ls -lrt sonic-buildimage.centec-arm64/target/files/bullseye/update_chassisdb_config
-rwxr-xr-x 1 lin lin 2467 Jun  6 15:33 update_chassisdb_config

### DIFF
--- a/slave.mk
+++ b/slave.mk
@@ -625,7 +625,7 @@ SONIC_TARGET_LIST += $(addprefix $(DEBS_PATH)/, $(SONIC_COPY_DEBS))
 #     SONIC_COPY_FILES += $(SOME_NEW_FILE)
 $(addprefix $(FILES_PATH)/, $(SONIC_COPY_FILES)) : $(FILES_PATH)/% : .platform
 	$(HEADER)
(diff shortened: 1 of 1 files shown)


linqingxuan:
'/azpw ms_conflict'

PR #19583, opened 2024-07-16
[Static DNS] Optimize DNS configuration update during interface-config service restart

oleksandrivantsiv:
#### Why I did it
The `resolvconfig` service updates the DNS configuration in the host OS and each running Docker container when a configuration change is detected. The DNS configuration update during the shutdown of the management interface is redundant, as the management interface is going down temporarily and, when it is back online, the DNS configuration will remain the same. The update of the DNS configuration adds a couple of seconds (depending on the CPU) to the interface-config service restart time when the management interface uses a dynamic IP address. This can affect the fast boot. To optimize the flow and execute the service restart faster, the update should be skipped.

##### Work item tracking

#### How I did it
Do not run DNS configuration update during the shutdown of the management interface.

#### How to verify it
Measure the execution time of the `service interfaces-config restart` command on the device with the static IP address configuration on the management interface and compare it to the execution time of the same command with the dynamic IP address. The difference should be insignificant.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/files/image_config/resolv-config/update-containers
+++ b/files/image_config/resolv-config/update-containers
@@ -1,6 +1,11 @@
 #!/bin/bash
 
-for container in $(docker ps -a --format=" {{ .ID }}"); do
+networking_status=$(systemctl is-active networking.service 2>/dev/null)
+if [[ $networking_status != "active" ]]; then
+    exit 0
+fi
+
+for container in $(docker ps -q); do
     docker cp -L /etc/resolv.conf ${container}:/_resolv.conf
     docker exec -t ${container} bash -c "cat /_resolv.conf > /etc/resolv.conf"
     docker exec -t ${container} bash -c "rm /_resolv.conf"
--- a/files/image_config/interfaces/interfaces-config.sh
+++ b/files/image_config/interfaces/interfaces-config.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
+resolvconf_updates=true
 
 function wait_networking_service_done() {
     local -i _WDOG_CNT="1"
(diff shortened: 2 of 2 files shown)


vivekrnv (approved):

files/image_config/resolv-config/update-containers
```diff
@@ -1,6 +1,11 @@
 #!/bin/bash
 
-for container in $(docker ps -a --format=" {{ .ID }}"); do
+networking_status=$(systemctl is-active networking.service 2>/dev/null)
```
ganglyu, line 3: why do you need this change?

files/image_config/resolv-config/update-containers
```diff
@@ -1,6 +1,11 @@
...
+if [[ $networking_status != "active" ]]; then
+    exit 0
+fi
+
+for container in $(docker ps -q); do
```
ganglyu, line 8: Why do you use "docker ps -q" to replace "docker ps -a"?

files/image_config/resolv-config/update-containers
oleksandrivantsiv: This is a performance optimization. We don't need to update the containers if the networking is inactive. When the networking is back online, one more update will be triggered. This reduces fast/warm reboot time on devices with weak CPUs.

files/image_config/resolv-config/update-containers
oleksandrivantsiv: `docker ps -a` returns a list of all available containers (active and inactive). `docker ps -q` returns a list of running containers. We want to update only running containers. For the inactive containers docker will uptade `/etc/resolv.conf` file automatically during the container start. So they will always have up-to-date configuration.

ganglyu (approved):

qiluo-msft (approved):

PR #19590, opened 2024-07-16
Prioritize configuration from config_db over constants.yml during BBR status initialization

Gfrom2016:
#### Why I did it
There are two places which we can configure the BBR disabled/enabled, one is constants.yml, the other is by config_db. During run time, config_db win. But at the init stage, constants.yml win. This is a wrong logic, constants.yml only win when there is no such config in the config_db.
##### Work item tracking
- Microsoft ADO **(number only)**: 28302790

#### How I did it
Initialize BBR status from constants.yml only when config_db doesn't have BBR entry.

#### How to verify it
Build image and run the following:
```
# bbr status in constants.yml is set to disabled
# change the bbr status in config_db to enabled and reload config
admin@str2-7050cx3-acs-01:~$ redis-cli -n 4 hset "BGP_BBR|all" "status" "enabled"
(integer) 0
admin@str2-7050cx3-acs-01:~$ redis-cli -n 4 hget "BGP_BBR|all" "status"
"enabled"
admin@str2-7050cx3-acs-01:~$ vtysh -c 'show run' | grep 'allowas'
  neighbor PEER_V4 allowas-in 1
  neighbor PEER_V6 allowas-in 1
admin@str2-7050cx3-acs-01:~$ sudo config save -y
admin@str2-7050cx3-acs-01:~$ sudo config reload -y

# check bgpcfgd log, bbr default status is enabled
2024 Aug 14 05:42:47.653216 str2-7050cx3-acs-01 INFO bgp#bgpcfgd: BBRMgr::Initialized and enabled from config_db. Default state: 'enabled'
```

- [ ] 202205
- [ ] 202211
- [x] 202305
- [x] 202311
- [x] 202405  

- [ ] 
- [ ]

### DIFF
--- a/src/sonic-bgpcfgd/bgpcfgd/managers_bbr.py
+++ b/src/sonic-bgpcfgd/bgpcfgd/managers_bbr.py
@@ -45,26 +45,39 @@ class BBRMgr(Manager):
 
     def __init(self):
         """ Initialize BBRMgr. Extracted from constructor """
-        if not 'bgp' in self.constants:
-            log_err("BBRMgr::Disabled: 'bgp' key is not found in constants")
-            return
-        if 'bbr' in self.constants['bgp'] \
-                and 'enabled' in self.constants['bgp']['bbr'] \
-                and self.constants['bgp']['bbr']['enabled']:
+        # Check BGP_BBR table from config_db first
+        bbr_status_from_config_db = self._get_bbr_status_from_config_db()
+
+        if bbr_status_from_config_db is None:
+            if not 'bgp' in self.constants:
+                log_err("BBRMgr::Disabled: 'bgp' key is not found in constants")
+                return
+            if 'bbr' in self.constants['bgp'] \
+                    and 'enabled' in self.constants['bgp']['bbr'] \
+                    and self.constants['bgp']['bbr']['enabled']:
+                self.bbr_enabled_pgs = self.__read_pgs()
+                if self.bbr_enabled_pgs:
+                    self.enabled = True
+                    if 'default_state' in self.constants['bgp']['bbr'] \
+                            and self.constants['bgp']['bbr']['default_state'] == 'enabled':
+                        default_status = "enabled"
+                    else:
+                        default_status = "disabled"
+                    self.directory.put(self.db_name, self.table_name, 'status', default_status)
+                    log_info("BBRMgr::Initialized and enabled from constants. Default state: '%s'" % default_status)
+                else:
+                    log_info("BBRMgr::Disabled: no BBR enabled peers")
+            else:
+                log_info("BBRMgr::Disabled: no bgp.bbr.enabled in the constants")
+        else:
             self.bbr_enabled_pgs = self.__read_pgs()
             if self.bbr_enabled_pgs:
                 self.enabled = True
-                if 'default_state' in self.constants['bgp']['bbr'] \
-                        and self.constants['bgp']['bbr']['default_state'] == 'enabled':
-                    default_status = "enabled"
-                else:
-                    default_status = "disabled"
-                self.directory.put(self.db_name, self.table_name, 'status', default_status)
-                log_info("BBRMgr::Initialized and enabled. Default state: '%s'" % default_status)
+                self.directory.put(self.db_name, self.table_name, 'status', bbr_status_from_config_db)
+                log_info("BBRMgr::Initialized and enabled from config_db. Default status: '%s'" % bbr_status_from_config_db)
             else:
                 log_info("BBRMgr::Disabled: no BBR enabled peers")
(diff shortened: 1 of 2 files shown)


src/sonic-bgpcfgd/tests/test_bbr.py
```diff
@@ -185,7 +185,40 @@ def test___init_8():
...
         }
     }
-    __init_common(constants, "BBRMgr::Initialized and enabled. Default state: 'enabled'", None, expected_bbr_entries, "enabled")
+    __init_common(constants, "BBRMgr::Initialized and enabled from constants. Default state: 'enabled'", None, expected_bbr_entries, "enabled")
+
+@patch('bgpcfgd.managers_bbr.swsscommon.DBConnector')
+def test___init_with_config_db(mock_db_connector):
+    mock_db_connector_instance = MagicMock()
+    mock_db_connector.return_value = mock_db_connector_instance
+    mock_db_connector_instance.get_table.return_value = {'all': {'status': 'enabled'}}
```
ZhaohuiS, line 194: this case covers the scenario of the bbr enabled in config_db, what about the scenario of the bbr disabled in config_db

src/sonic-bgpcfgd/tests/test_bbr.py
```diff
@@ -185,7 +186,32 @@ def test___init_8():
             "bbr": expected_bbr_entries,
         }
     }
-    __init_common(constants, "BBRMgr::Initialized and enabled. Default state: 'enabled'", None, expected_bbr_entries, "enabled")
+    __init_common(constants, "BBRMgr::Initialized and enabled from constants. Default state: 'enabled'", None, expected_bbr_entries, "enabled")
+
+@patch('bgpcfgd.managers_bbr.BBRMgr.get_bbr_status_from_config_db', return_value='disabled')
+def test___init_with_config_db_1(mocked_get_bbr_status_from_config_db):
```
ZhaohuiS, line 192: @Gfrom2016 suggest to use a meaningful case name, such as test___init_with_config_db_default_disable

src/sonic-bgpcfgd/bgpcfgd/managers_bbr.py
```diff
@@ -45,26 +45,38 @@ def del_handler(self, key):
...
     def __init(self):
         """ Initialize BBRMgr. Extracted from constructor """
-        if not 'bgp' in self.constants:
-            log_err("BBRMgr::Disabled: 'bgp' key is not found in constants")
-            return
-        if 'bbr' in self.constants['bgp'] \
-                and 'enabled' in self.constants['bgp']['bbr'] \
-                and self.constants['bgp']['bbr']['enabled']:
+        # Check BGP_BBR table from config_db first
+        bbr_status_from_config_db = self.get_bbr_status_from_config_db()
```
StormLiangMS, line 49: @Gfrom2016 could we validate the return value of status?

StormLiangMS:
hi @Gfrom2016 could we communicate with @prsunny to make sure the understanding is fully aligned that the BBR status would be effected by peers config in constants.yaml file?

And we may need to add a nightly test to cover this?

src/sonic-bgpcfgd/tests/test_bbr.py
Gfrom2016: refactor testcase name

src/sonic-bgpcfgd/bgpcfgd/managers_bbr.py
Gfrom2016: Add validating return value in get_bbr_status_from_config_db()

src/sonic-bgpcfgd/bgpcfgd/managers_bbr.py
```diff
@@ -82,6 +94,23 @@ def __read_pgs(self):
                 res[pg_name] = pg_afs
         return res
 
+    def get_bbr_status_from_config_db(self):
+        """
+        Read BBR status from CONFIG_DB
+        :return: BBR status from CONFIG_DB or None if not found
+        """
+        config_db = swsscommon.ConfigDBConnector()
+        if config_db is None:
```
siqbal1986, line 103: if at init time config DB connection fails, we are reverting to default state from constants.yml. I suggest we throw a log here before returning None.

siqbal1986 (changes requested):
everything else looks good but a log would help with early detection if configDB connection fails.

src/sonic-bgpcfgd/bgpcfgd/managers_bbr.py
Gfrom2016: fixed

prsunny (approved):
lgtm

PR #19606, opened 2024-07-17
[action] [Broadcom]: Update XGS SAI version to 10.1.35.0

r12f:
Why I did it
Upgrade the xgs SAI version to 10.1.35.0 to include the following changes:

10.1.21.0: Update the ACL impelementation to enable the ACL switch bind to support PFCWD on MACSEC devices
10.1.23.0: Handle the FDR stats get not supported more properly.
10.1.24.0: (CSP CS00012316286) Fix LPM Miss ACL with counters fails
10.1.25.0: Improve ECMP performance
10.1.28.0: Fix SAI is not honoring the bcm_linkscan_interval set in .bcm config file for Ramon and error with setting SAI_NEIGHBOR_ENTRY_ATTR_IS_LOCAL attribute for neighbor
10.1.29.0: Fix fabric switch initialization delayed if SW linkscan enabled during platform init
10.1.30.0: (CSP CS00012356911) Fix traffic drops on all priorities when PFC asserted on two priorities.
10.1.31.0: NHG performance fix
10.1.33.0: (CSP CS00012345242) High CPU due to SAI code calls bcm_port_resource_get for every fabric error counter
10.1.35.0: Support MMU configuration updates for TH5.

Work item tracking
Microsoft ADO (number only): 28724885


platform/broadcom/sai.mk
```diff
@@ -1,4 +1,4 @@
-LIBSAIBCM_XGS_VERSION = 10.1.20.0
+LIBSAIBCM_XGS_VERSION = 10.1.35.0
 LIBSAIBCM_DNX_VERSION = 7.1.111.1
 LIBSAIBCM_XGS_BRANCH_NAME = SAI_10.1.20.0
```
Gfrom2016, line 3: LIBSAIBCM_XGS_BRANCH_NAME should be 'SAI_10.1.0_GA'

Gfrom2016 (changes requested):

platform/broadcom/sai.mk
r12f: Thanks @Gfrom2016 ! @kperumalbfn as FYI. : D

lguohan:
replaced by #19617

PR #19634, opened 2024-07-19
[finalize-warmboot.sh] reset cpufreq governor to default

stepanblyschak:
#### Why I did it

Set cpufreq.default_governor to *performance* for faster boot time. We observe consistent 1 sec improvement across several devices.

The change in finalize-warmboot.sh restores the default governor after fast or warm boot is finished.

**NOTE**: This will apply to upgrades starting from 202405 since this is set in shutdown path to avoid any extra scripts running at boot time. Upgrade from older versions/branches will require a runtime patch to fast-reboot and warm-reboot script.

##### Work item tracking

#### How I did it

After fast or warm boot is finished restore to default governor.

#### How to verify it

Run fast-reboot or warm-reboot. Check:
```
admin@sonic:~$ cat /sys/devices/system/cpu/cpu0/cpufreq/scaling_governor
performance
```

After boot is finalized check that it is reset back to default:
```
admin@sonic:~$ cat /sys/devices/system/cpu/cpu0/cpufreq/scaling_governor
schedutil
```

Tested with https://github.com/sonic-net/sonic-utilities/pull/3435

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [x] 202405

- [ ] 
- [ ]

### DIFF
--- a/files/image_config/warmboot-finalizer/finalize-warmboot.sh
+++ b/files/image_config/warmboot-finalizer/finalize-warmboot.sh
@@ -116,7 +116,7 @@ function finalize_common() {
 
     if [[ "$asic_type" == "mellanox" ]]; then
         # Read default governor from kernel config
-        local -r default_governor=$(cat "/boot/config-$(uname -r)" | grep -o 'CONFIG_CPU_FREQ_DEFAULT_GOV_[^=]*=y')
+        local -r default_governor=$(cat "/boot/config-$(uname -r)" | grep -E 'CONFIG_CPU_FREQ_DEFAULT_GOV_.*=y' | sed -E 's/CONFIG_CPU_FREQ_DEFAULT_GOV_(.*)=y/\1/')
         set_cpufreq_governor "$default_governor"
     fi
 }
--- a/files/image_config/warmboot-finalizer/finalize-warmboot.sh
+++ b/files/image_config/warmboot-finalizer/finalize-warmboot.sh
@@ -2,6 +2,9 @@
 
 VERBOSE=no
 
+# read SONiC immutable variables
+[ -f /etc/sonic/sonic-environment ] && . /etc/sonic/sonic-environment
+
 # Define components that needs to reconcile during warm
 # boot:
 #       The key is the name of the service that the components belong to.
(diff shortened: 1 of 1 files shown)


bingwang-ms:
@saiarcot895 @vaibhavhd Can you please help review?

liat-grozovik:
@vaibhavhd kindly reminder for code review feedback/approval.

vaibhavhd:
Comments are open on utilities PR.

files/image_config/warmboot-finalizer/finalize-warmboot.sh
```diff
@@ -101,16 +101,30 @@ function check_list()
     echo ${RET_LIST}
 }
 
+function set_cpufreq_governor() {
+    local -r governor="$1"
+    echo "$governor" | tee /sys/devices/system/cpu/cpu*/cpufreq/scaling_governor 1> /dev/null \
```
vaibhavhd, line 106: Will this setting (or the way we are writing to a file) be effective at runtime?

files/image_config/warmboot-finalizer/finalize-warmboot.sh
```diff
@@ -101,16 +101,30 @@ function check_list()
...
+        && debug "Set CPUFreq scaling governor to $governor" \
+        || debug "Failed to set CPUFreq scaling governor to $governor"
+}
+
+function finalize_common() {
+    # Read default governor from kernel config
+    local -r default_governor=$(cat "/boot/config-$(uname -r)" | grep -o 'CONFIG_CPU_FREQ_DEFAULT_GOV_[^=]*=y')
```
vaibhavhd, line 113: What's the guarantee that `"/boot/config-$(uname -r)"` will have the default governor setting?

Do you need to sanity check on the value and type of `default_governor` here? What if the `cat` or `grep` result in a failure? Wouldn't the code proceed to assign invalid value to `/sys/devices/system/cpu/cpu*/cpufreq/scaling_governor`

stepanblyschak: It's part of our sonic linux kernel config, why would we remove it or set it to invalid value?

files/image_config/warmboot-finalizer/finalize-warmboot.sh
stepanblyschak: Yes

files/image_config/warmboot-finalizer/finalize-warmboot.sh
```diff
@@ -101,16 +101,30 @@ function check_list()
...
+    set_cpufreq_governor "$default_governor"
```
vaibhavhd, line 114: The utilities change is made for MLNX platforms only. I think it would be best to make this call only for MLNX platforms as well.

vaibhavhd (approved):
Approved with suggestions.

dgsudharsan:
@qiluo-msft can you please help merge?

qiluo-msft (approved):

bingwang-ms:
@stepanblyschak Can you raise a PR to 202405 to do the cherry-pick? The auto cherry-pick is not working for this PR.

PR #19716, opened 2024-07-26
[202405] BBR disabled by default.

siqbal1986:
The BGP Bounce back routing feature was enabled by default in 202305.  However this causes breakage in the overlay-ecmp feature.
Due to a bug in BGP BBR manager, this default configuration takes precendence on the config DB switch. As a result the BBR becomes enabled after restart even when it is disabled via configuration (ADO 28302790). Until this bug is fixed, the default state is being set to disabled to prevent accidental enablement.

##### Work item tracking
28790780

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [X] 202305
- [X] 202311

- [ ] 
- [ ]

### DIFF
--- a/files/image_config/constants/constants.yml
+++ b/files/image_config/constants/constants.yml
@@ -33,7 +33,7 @@ constants:
           - "deny 0::/0 le 59"
           - "deny 0::/0 ge 65"
     bbr:
-      enabled: true
+      enabled: false
       default_state: "disabled"
     peers:
       general: # peer_type


StormLiangMS:
hi @Gfrom2016 could you take a look? There is another PR we are working on, any conflict?

Gfrom2016:
> hi @Gfrom2016 could you take a look? There is another PR we are working on, any conflict?

no conflict.

Gfrom2016 (approved):
LGTM

StormLiangMS:
hi @siqbal1986  should we submit PR to master branch, then do the cherry pick to feature branch?

files/image_config/constants/constants.yml
```diff
@@ -33,7 +33,7 @@ constants:
           - "deny 0::/0 le 59"
           - "deny 0::/0 ge 65"
     bbr:
-      enabled: true
+      enabled: false
```
Gfrom2016, line 36: On second thought, changing enabled to false would cause BBRManager drop all set requests according to https://github.com/sonic-net/sonic-buildimage/blob/4e78b11c96192d3c1e4332d5bca9c415ad7c1c9d/src/sonic-bgpcfgd/bgpcfgd/managers_bbr.py#L31.

Gfrom2016:
FYI, PR for ADO 28302790 https://github.com/sonic-net/sonic-buildimage/pull/19590

files/image_config/constants/constants.yml
siqbal1986: There is a bug in this code which gives the following behaviour.
IF BBR is enabled in the constants.yml and the configuration of BGP_BBR|all etnry is disabled, The BBR remains disabled until the config reload. At SONIC reload, the managers_bbr reads the constants,yml and this configuration takes precedence over config_db BGP_BBR|all toggle. As a result the deivces comes up with BBR enabled. For overlay ECMP, this causes T1s to  learn  routes via BGP which inturn causes traffic blackholing of ECMP routes. ADO for the bug. 28302790
Since this feature is not being actively used, we are disabling it by default until this bug is fixed.

PR #19747, opened 2024-07-31
Add to support the pmon.service restart when when syncd.service restart

PeterTSW-EC:
#### Why I did it
The pmon.service is not restarted after restarting syncd.service. 
The pmon.service need to set the preemphasis and media-type after the syncd.service restart. It is needed to add supporting the pmon.service restart after the syncd.service restart.

##### Work item tracking

#### How I did it
Adds a condition into syncd.sh to restart the pmon.service when syncd restart.

#### How to verify it
Restart syncd.service and confirm the pmon.service is also restart.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [X] 202305
- [X] 202311

- [ ] 
- [ ]

### DIFF
--- a/files/build_templates/pmon.service.j2
+++ b/files/build_templates/pmon.service.j2
@@ -4,6 +4,9 @@ Requires=database.service config-setup.service
 After=database.service config-setup.service
 {% if sonic_asic_platform == 'mellanox' %}
 After=syncd.service
+{% elif sonic_asic_platform == 'broadcom' %}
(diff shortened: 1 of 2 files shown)


judyjoseph (changes requested):
This might affect other platforms, needs more discussion, if there is another way to achive this @prgeor

PR #19755, opened 2024-07-31
[VOQ] Fix the parsing of core_port_id and core_id for Inband and Recirc Port.

abdosi:
What I did:
After the PR: https://github.com/sonic-net/sonic-buildimage/pull/18704 parsing for `core_port_id` and `core_id` for Inband and Recicrc Port needs updates otherwise value of these 2 fields comes as `None` casuing SAI not configure this ports correctly and also yang validation failure.

Microsoft ADO: 28897483

How I verify:
This issue could have been caught via but UT also had bug. Fixed the UT also.
Manual verified the change for SAI and Yang Validation


gechiang (approved):

kperumalbfn (approved):

gechiang:
NOTE; this change will not be backported to any older releases other than 202405 per dicussion with @abdosi.  Just capturing here so we know we did discuss about it.

PR #19817, opened 2024-08-05
Fix for config replace issues seen with COPP configurations

Verma-Anukul:
==> Issue: config replace/yang-validation fails for COPP configurations.

==> Root cause:
• COPP's device initial config copp_cfg.json has missing yang mandatory leaf "trap_action" for "default" COPP_GROUP.
• "genetlink_name" & "genetlink_mcgrp_name" nodes are not added in sonic-copp.yang, which are used for "queue2_group1" in copp_cfg.json

==> Fix:
• In copp_cfg.json, mandatory leaf "trap_action" is added for "default" COPP_GROUP.

"COPP_GROUP": {
  "default": {
    + "trap_action":"trap", <<< this node needs to be added
       "queue": "0",
       "meter_type":"packets",

• "genetlink_name" and "genetlink_mcgrp_name" leaves are added in sonic-copp.yang.

• DB Migration script is updated to upgrade the DB config automatically.

==> Tests :
Config replace ->
i. Config load copp_cfg.json
ii. Config save
iii. Config replace -> Ensure no error seen

Upgrade ->
i. Save config in build without the fix
ii. upgrade to build with changes and ensure configurations are reflected in new config_db.json

==> PRs :
Sonic-buildimage : 
Sonic-utilities :

### DIFF
--- a/files/image_config/copp/copp_cfg.j2
+++ b/files/image_config/copp/copp_cfg.j2
@@ -1,6 +1,7 @@
 {
     "COPP_GROUP": {
 	    "default": {
+		    "trap_action":"trap",
 		    "queue": "0",
 		    "meter_type":"packets",
 		    "mode":"sr_tcm",
--- a/src/sonic-yang-models/yang-models/sonic-copp.yang
+++ b/src/sonic-yang-models/yang-models/sonic-copp.yang
@@ -11,6 +11,11 @@ module sonic-copp {
 
 	description "CoPP YANG Module for SONiC OS";
 
+	revision 2024-05-29 {
+		description
+			"Added genetlink name and mcgrp leaves";
+	}
+
 	revision 2021-03-31 {
 		description
 			"First Revision";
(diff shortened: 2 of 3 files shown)


Verma-Anukul:
@qiluo-msft @lguohan @ganglyu @prsunny @dgsudharsan 

Please help in reviewing these PRs- 
Sonic-buildimage : https://github.com/sonic-net/sonic-buildimage/pull/19817
Sonic-utilities : https://github.com/sonic-net/sonic-utilities/pull/3473
SONiC (docs) : https://github.com/sonic-net/SONiC/pull/1773

src/sonic-yang-models/yang-models/sonic-copp.yang
```diff
@@ -133,6 +138,16 @@ module sonic-copp {
 					default "forward";
 					description "Red action";
 				}
+
```
dgsudharsan, line 141: These are internal fields and shouldn't be user configurable. Please check the comment here https://github.com/sonic-net/sonic-buildimage/pull/7199#discussion_r610339266

dgsudharsan (changes requested):

files/image_config/copp/copp_cfg.j2
```diff
@@ -1,6 +1,7 @@
 {
     "COPP_GROUP": {
 	    "default": {
+		    "trap_action":"trap",
```
dgsudharsan, line 4: This file is not validated against yang. The trap action for default is hardcoded due to legacy code present in https://github.com/sonic-net/sonic-swss/blob/465391d1825ea3906d27a1285d2da5876f763cd7/orchagent/copporch.cpp#L185

By default the copp config doesn't go to config_db and thus we don't require yang validation on this file. However if the user need to override the copp config, user needs to set the appropriate fields along with correct yang validation.

dgsudharsan (changes requested):

PR #19836, opened 2024-08-06
Revert "Revert suppress FIB pending feature"

dgsudharsan:
Reverts sonic-net/sonic-buildimage#19027

The support for BGP zebra enhancements is merged in https://github.com/sonic-net/sonic-buildimage/pull/19717. Hence re-enabling BGP suppress fib pending feature.


StormLiangMS:
hi @lipxu  could you help to take a look?

dgsudharsan:
> hi @lipxu could you help to take a look? Or suggest which test we should have nVidia to run to take in this PR?

Hi @StormLiangMS . We have run the entire sonic-mgmt test suite with these changes + zebra bgp enhancements. We found no issues. We also find memory doesn't increase with test_stress_route with 50 iterations.

StormLiangMS (approved):
LGTM

PR #19914, opened 2024-08-15
Add an artificial dependency between two modules for centec-arm64

saiarcot895:
#### Why I did it

For centec-arm64, both platform-modules-fs and platform-modules-e530 try to compile the same directory of kernel sources. If both compilations are running at the same time, then this is going to lead to errors.

##### Work item tracking

#### How I did it

Fix this by introducing an artificial dependency between the two so that the compilations are forced to be serialized.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]


saiarcot895:
@philo-micas could you review this when you get a chance?

PR #19924, opened 2024-08-16
Add VPN related configuration in frrcfgd.py

eddieruan-alibaba:
#### Why I did it
1. Add VPN related configuration in frrcfgd.py. This is needed for Phoenix Wing initiative. 
2. Change timeout value from 0.01 sec to 10 sec. The timeout value for 0.0.1 sec doesn't sound correct. With this timeout value, frrcfgd would be 99% busy which leads vsonic failed to be brought up. Eddie changed it to 1 sec, which reduced cpu usage to 15%, then changed to 10 sec, which reduced cpu usage to 0.1 %.
3. Currently, the initial config comes from j2 file, while the following configuration handles in frrcfgd.py. It is not friendly for adding new configurations, since there are two places need to modify. I want to replay configuration in frrcfgd.py for "separated" mode.

##### Work item tracking

#### How I did it
Fix

#### How to verify it
Fix is verified in phoenix wing 7 node testbed. Both underlay BGP and overlay BGP are up.

Test plan : https://github.com/eddieruan-alibaba/sonic-mgmt/blob/phoenix_ptf_testplan/docs/testplan/srv6/SRv6-phoenixwing-ptf-testplan.md 

Tested config db : https://github.com/eddieruan-alibaba/sonic-mgmt/blob/eruan-frrcfgd-cfg/tests/srv6/7nodes_cisco/PE1.json

N/A

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

202311 based

- [ ] 
- [ ] 

Add VPN related configuration in frrcfgd.py

n/a


eddieruan-alibaba:
Discussed in routing WG. The meeting minutes could be found at

https://lists.sonicfoundation.dev/g/sonic-wg-routing/wiki/37416

src/sonic-frr-mgmt-framework/frrcfgd/frrcfgd.py
```diff
@@ -2258,7 +2269,7 @@ def __init__(self):
         syslog.syslog(syslog.LOG_DEBUG, 'Init Cached DB data')
         for key, entry in self.table_data_cache.items():
             syslog.syslog(syslog.LOG_DEBUG, '  %-20s : %s' % (key, entry))
-        if self.config_mode == "unified":
+        if self.config_mode == "unified" or self.config_mode == "separated":
```
venkatmahalingam, line 2272: Is "unified" config_mode not working for you use-case? seperated config_mode supposed to generate the individual daemon config files during boot-up as per design in docker_init.sh, we should avoid duplicate config generation.

src/sonic-frr-mgmt-framework/frrcfgd/frrcfgd.py
```diff
@@ -1791,9 +1802,9 @@ class BGPConfigDaemon:
                    (['local_asn', '+local_as_no_prepend',
                      '+local_as_replace_as'],               '{no:no-prefix}neighbor {} local-as {} {:no-prepend} {:replace-as}'),
                    (['admin_status', '+shutdown_message'],  '{no:no-prefix}neighbor {} shutdown {:shutdown-msg}', ['false', 'true']),
-                   ('local_addr',                           '{no:no-prefix}neighbor {} update-source {}'),
+                   ('update_source',                           '{no:no-prefix}neighbor {} update-source {}'),
```
venkatmahalingam, line 1805: This is a backward incompatible change; I would keep the local_addr unless this adds any major value.

src/sonic-frr-mgmt-framework/frrcfgd/frrcfgd.py
eddieruan-alibaba: That was my original question how "fpm address 127.0.0.1" is applied in "unified" mode?

src/sonic-frr-mgmt-framework/frrcfgd/frrcfgd.py
eddieruan-alibaba: But that would break "show ip interface" since we don't need to have "update source" for normal BGPs.

src/sonic-frr-mgmt-framework/frrcfgd/frrcfgd.py
prvattem, line 2272: As I understand it, 

1. The initial configuration generated via j2 files to individual proto conf files will populate frr initially.
2. As the configuration is loaded in the DB, frrcfgd will process each event and again populate frr (same configuration)
3. This change with addition of the check `"self.config_mode == "separated"` will make another attempt to configure frr

Step 1 & 2 is the current implementation
Step 1 - will make the configuration available to frr much faster than the other 2.
Step 3 - is redundant.

venkatmahalingam: I believe, we had config loading time issue when we push the configs one by one during boot-up incase of scaled configs instead of generating the static config file per FRR daemon, if you have any perf data, if would be useful. If you think, this new change gives better performance and avoids additional jinja file creation, I would like to have a new config_mode or a field (in DEVICE_METADATA) defined for this purpose to avoid overloading of the seperated mode design today.

src/sonic-frr-mgmt-framework/frrcfgd/frrcfgd.py
```diff
@@ -1791,9 +1802,9 @@ class BGPConfigDaemon:
...
                    ('name',                                 '{no:no-prefix}neighbor {} description {}'),
-                   (['ebgp_multihop', '+ebgp_multihop_ttl'],'{no:no-prefix}neighbor {} ebgp-multihop {}', ['true', 'false']),
+                   (['ebgp_multihop', '+ebgp_multihop_ttl'],'{no:no-prefix}neighbor {} ebgp-multihop {}'),
```
prvattem, line 1807: Could you please explain this change?

src/sonic-frr-mgmt-framework/frrcfgd/frrcfgd.py
venkatmahalingam: https://github.com/sonic-net/sonic-utilities/blob/317e649514c9b205849ffb5ea96a6a233e38290c/scripts/ipintutil#L51

For that, we need to make the following changes in the above file not the field name change.

 if 'local_addr' in data[neighbor_ip]:
     local_addr = data[neighbor_ip]['local_addr']
 else:
     local_addr = ""

src/sonic-frr-mgmt-framework/frrcfgd/frrcfgd.py
eddieruan-alibaba: The current usage is 

https://github.com/eddieruan-alibaba/sonic-mgmt/blob/eruan-frrcfgd-cfg/tests/srv6/7nodes_cisco/PE1.json#L95

I don't know how to add "true" and "false" and don't understand the need to have this "true" and "false"

src/sonic-frr-mgmt-framework/frrcfgd/frrcfgd.py
eddieruan-alibaba: That means the same config handling would be coded in j2 and frrcfgd.py. That requires changing in two different places. Do we really want to take this approach?

eddieruan-alibaba: Ok. In unified mode, how does "fpm address 127.0.0.1" get applied? I could use "unified" mode if I could apply this config.

src/sonic-frr-mgmt-framework/frrcfgd/frrcfgd.py
eddieruan-alibaba: Ok. I will revert it and open an issue to track this fix.

eddieruan-alibaba: https://github.com/sonic-net/sonic-buildimage/issues/19930 is opened for track this change.

src/sonic-frr-mgmt-framework/frrcfgd/frrcfgd.py
prvattem: are we ok with slow config load and 3 times config updates?

eddieruan-alibaba: I am open for suggestions. Here are the issues I have hit

1. When using "unified" mode, "fpm address 127.0.0.1" is not applied. Does anyone know how this should be applied?
2. When using "separated" mode, new added configurations would be not applied at init time. It seems the initial j2 files are used. My concern is if that means all configurations would be created two copies of handling, one in frrcfgd.py for run time changing, and one in j2. How do we guarantee the consistency in developing. 


If someone could help to figure out how "fpm address 127.0.0.1" get applied in "unified" mode, I would move back to that mode which seems cleaner.

venkatmahalingam: ok, I'll help you for the above one.

venkatmahalingam: You can add the fpm address command in the following j2 file to work for unified mode, please check.
https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-frr-mgmt-framework/templates/frr/frr.conf.j2

src/sonic-frr-mgmt-framework/frrcfgd/frrcfgd.py
venkatmahalingam: Please revert this change and use as following for ebgp

```
 "default|2064:300::1f": {
            "admin_status": "true",
            "update_source" :"2064:100::1d",
            **"ebgp_multihop": "true",
            "ebgp_multihop_ttl":  "255",**
            "capability_ext_nexthop": "true",
            "local_addr" :"2064:100::1d",
            "local_asn" : "64600",
            "asn": "64602",
            "name": "PE3"
        },

```

src/sonic-frr-mgmt-framework/frrcfgd/frrcfgd.py
eddieruan-alibaba: Changed to unified mode and set fpm address in frr.conf.

eddieruan-alibaba: moved to unified mode as suggested by Venkat.

src/sonic-frr-mgmt-framework/frrcfgd/frrcfgd.py
eddieruan-alibaba: Revert back previous change and updated json file.

dockers/docker-fpm-frr/docker_init.sh
```diff
@@ -86,13 +86,15 @@ elif [ "$CONFIG_TYPE" == "split-unified" ]; then
...
 elif [ "$CONFIG_TYPE" == "unified" ]; then
+    echo "" >  /etc/frr/frr.conf
     CFGGEN_PARAMS=" \
         -d \
         -y /etc/sonic/constants.yml \
         -t /usr/share/sonic/templates/gen_frr.conf.j2,/etc/frr/frr.conf \
     "
     sonic-cfggen $CFGGEN_PARAMS
     echo "service integrated-vtysh-config" > /etc/frr/vtysh.conf
+    echo "fpm address 127.0.0.1" >>  /etc/frr/frr.conf
```
venkatmahalingam, line 97: Can this added in the frr.conf.j2 itself to keep all frr.conf contents in one file?

dockers/docker-fpm-frr/docker_init.sh
```diff
@@ -86,13 +86,15 @@ elif [ "$CONFIG_TYPE" == "split-unified" ]; then
     rm -f /etc/frr/bgpd.conf /etc/frr/zebra.conf /etc/frr/staticd.conf
     write_default_zebra_config /etc/frr/frr.conf
 elif [ "$CONFIG_TYPE" == "unified" ]; then
+    echo "" >  /etc/frr/frr.conf
```
venkatmahalingam, line 89: Why do we need this? the below one should override the frr.conf file?

dockers/docker-fpm-frr/docker_init.sh
eddieruan-alibaba: ok. I will add it to frr.conf.j2

dockers/docker-fpm-frr/docker_init.sh
eddieruan-alibaba: To make sure it starts from  a clean file. During docker restart, the old frr.conf remains.

venkatmahalingam: The below code should override the frr.conf contents always, are we observing the stale content in the frr.conf during docker restart?
CFGGEN_PARAMS=" \
        -d \
        -y /etc/sonic/constants.yml \
        -t /usr/share/sonic/templates/gen_frr.conf.j2,/etc/frr/frr.conf

eddieruan-alibaba: Yes. I saw "fm add" got added twice in frr.conf.

eddieruan-alibaba: Add "fpm addr ..." to frr.conf.j2. 

Trigger "docker restart" and see proper output for "fpm addr ..."

eddieruan-alibaba: @venkatmahalingam this approach breaks sanity. 
json_path = 'tests/data/sonic-cfggen/frr.conf.j2/all.json'
match_path = 'tests/data/sonic-cfggen/frr.conf.j2/all.conf'

E         -  ['fpm address 127.0.0.1'],
ls ./src/sonic-bgpcfgd/tests/data/sonic-cfggen/frr.conf.j2/
all.conf        all.json

venkatmahalingam: Let's discuss in the Routing WG meeting 8/22

venkatmahalingam (approved):

eddieruan-alibaba:
@lguohan  can you help to review it? Venkat has already reviewed it and approved it.

Also, any recent breakage in dualtor/test_switchover_failure.py? My changes are not related to dualtor switchover. Not sure why it fails.

E       Failed: Database states don't match expected state active,incorrect APP_DB values {
E           "HW_MUX_CABLE_TABLE:Ethernet12": {
E               "state": "standby"
E           },
E           "MUX_CABLE_TABLE:Ethernet12": {
E               "state": "standby"
E           }
E       }

db         = 0
self       = <tests.common.dualtor.control_plane_utils.DBChecker object at 0x7f301cab8eb0>

https://dev.azure.com/mssonic/build/_build/results?buildId=627456&view=logs&jobId=d8871267-39af-5540-59d2-ff1a86c185cf&j=d8871267-39af-5540-59d2-ff1a86c185cf

prvattem (approved):

kperumalbfn (approved):

PR #19932, opened 2024-08-17
[monit][arp_update] Adding monit script to check if arp_update is stuck

Ndancejic:
adding a script arp_update_checker, which will run every 10 minutes to check if arp_update script is stuck. It will check if arp_update ping command is run for longer than 5 seconds. If ping is running for longer than 5 seconds, the script will patch arp_update to add a 0.2 second timeout to ping and restart the process. After 3 failed attempts to restart arp_update, trigger alert.

#### Why I did it
Alerting generated a lot of false positives for arp_update being stuck. This will attempt to restart arp_update if it gets stuck on ping command and if arp_update gets stuck 3 cycles in a row, monit will alert.

##### Work item tracking
- Microsoft ADO **(number only)**: 28645547

#### How I did it
Added a script arp_update_checker which will check for ping child processes that arp_update creates and attempt to restart arp_update if the ping process is up for longer than 5 seconds.

#### How to verify it
upload arp_update_checker to sonic device under /usr/bin/arp_update_checker.
add the following code chunk to /etc/monit/conf.d/sonic-host:
```
# arp_update_checker tool that verifies that arp_update script is not stuck on ping command every 10 minutes
check program arp_update_checker with path "/usr/bin/arp_update_checker" every 10 cycles
    if status != 0 for 3 times within 3 cycles then alert repeat every 1 cycles
``` 

reload monit
```
sudo monit reload
```

run arp_update_checker manually:
```
sudo monit -g arp_update_checker
```

check the status of arp_update_checker:
```
sudo monit status arp_update_checker
```

Also simulated arp_update script with ping command that does not resolve.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

202305

- [x] 20230531.22

Create arp_update_checker monit script to check if arp_update is stuck and restart the process if needed.

ACC5A8CB-9A05-4AF7-B85A-11A3EB72BCD8_1_105_c

### DIFF
--- a/files/image_config/monit/arp_update_checker
+++ b/files/image_config/monit/arp_update_checker
@@ -20,9 +20,6 @@ import sys
 
 TIMEOUT=5.0
 
-# Set up logging to syslog
-syslog.openlog(ident='arp_update_checker', logoption=syslog.LOG_PID, facility=syslog.LOG_DAEMON)
-
 def log_info(message):
     syslog.syslog(syslog.LOG_INFO, message)
 
--- a/files/image_config/monit/arp_update_checker
+++ b/files/image_config/monit/arp_update_checker
@@ -1,0 +1,109 @@
+#!/usr/bin/env python3
+
+"""
+arp_update_checker
+
+This script is intended to be run by Monit. It will write an alerting message into
+syslog if it found arp_update script stuck on ping command.
+Then it will attempt to restart arp_update script.
+
+The following is an example in Monit configuration file to show how Monit will run
+this script:
+
+check program arp_update_checker with path "/usr/bin/arp_update_checker" every 10 cycles
+    if status != 0 for 3 times within 3 cycles then alert repeat every 1 cycles
+"""
+
+import syslog
+import subprocess
+import sys
+
+TIMEOUT=5.0
+
+# Set up logging to syslog
+syslog.openlog(ident='arp_update_checker', logoption=syslog.LOG_PID, facility=syslog.LOG_DAEMON)
+
+def log_info(message):
+    syslog.syslog(syslog.LOG_INFO, message)
+
+def log_warning(message):
+    syslog.syslog(syslog.LOG_WARNING, message)
+
+def log_error(message):
+    syslog.syslog(syslog.LOG_ERR, message)
+
+def is_process_running(process_name):
+    """Check if there is any running process that contains the given name."""
+    try:
+        # Use pgrep to check if the process is running
+        output = subprocess.check_output(["pgrep", "-f", process_name])
+        return bool(output.strip())
+    except subprocess.CalledProcessError:
+        return False
+
+def is_arp_update_stuck(timeout=TIMEOUT):
+    """Check if arp_update is stuck by looking for a ping command running longer than the specified timeout. (Default 1sec)"""
+    try:
+        arp_update_pid = subprocess.check_output(["pgrep", "-x", "arp_update"]).strip().decode('utf-8')
+        child_pids = subprocess.check_output(["pgrep", "-P", arp_update_pid]).strip().decode('utf-8').split()
+        for pid in child_pids:
+            cmd = subprocess.check_output(["ps", "-p", pid, "-o", "comm="]).strip().decode('utf-8')
+
+            if "ping" in cmd:
+                start_time = subprocess.check_output(["ps", "-p", pid, "-o", "etime="]).strip().decode('utf-8')
+                elapsed_time = parse_elapsed_time(start_time)
+
+                # Check if the elapsed time is greater than the timeout
+                if elapsed_time > timeout:
+                    return True
+        return False
+    except subprocess.CalledProcessError:
+        return False
+
+def parse_elapsed_time(etime):
+    """Convert elapsed time from ps command to seconds."""
+    parts = etime.split('-')
+    if len(parts) == 2:
+        days = int(parts[0])
+        time_part = parts[1]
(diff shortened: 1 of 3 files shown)


files/image_config/monit/arp_update_checker
```diff
@@ -0,0 +1,106 @@
...
+def is_arp_update_stuck(timeout=TIMEOUT):
+    """Check if arp_update is stuck by looking for a ping command running longer than the specified timeout. (Default 1sec)"""
+    try:
+        arp_update_pid = subprocess.check_output(["pgrep", "-x", "arp_update"]).strip().decode('utf-8')
+        child_pids = subprocess.check_output(["pgrep", "-P", arp_update_pid]).strip().decode('utf-8').split()
+        for pid in child_pids:
+            cmd = subprocess.check_output(["ps", "-p", pid, "-o", "comm="]).strip().decode('utf-8')
+
+            if "ping" in cmd:
+                start_time = subprocess.check_output(["ps", "-p", pid, "-o", "etime="]).strip().decode('utf-8')
```
siqbal1986, line 50: if you use "etimes" instead of "etime" this would give you the value in seconds and you can get rid of the parse_elapsed_time function

files/image_config/monit/arp_update_checker
```diff
@@ -0,0 +1,106 @@
+#!/usr/bin/env python3
+
+"""
+arp_update_checker
+
+This script is intended to be run by Monit. It will write an alerting message into
+syslog if it found arp_update script stuck on ping command.
```
siqbal1986, line 7: finds

files/image_config/monit/arp_update_checker
```diff
@@ -0,0 +1,106 @@
...
+        minutes = 0
+        seconds = float(time_parts[0])
+
+    total_seconds = days * 86400 + hours * 3600 + minutes * 60 + seconds
+    return total_seconds
+
+def restart_arp_update():
+    """Restart the arp_update process."""
+    try:
+        subprocess.check_call(["docker", "exec", "swss", "supervisorctl", "restart", "arp_update"])
```
theasianpianist, line 87: Restarting arp_update on its own won't fix the issue, the process will likely just get stuck again. Suggest also patching the arp_update file so the issue doesn't reoccur, then restart the process after.:
```
docker exec  swss sed -i -s 's/ping6cmd=\"ping6/ping6cmd=\"timeout 0.2 ping6/' /usr/bin/arp_update
```

files/image_config/monit/arp_update_checker
```diff
@@ -0,0 +1,106 @@
...
+
+def main():
+    """
+    This function will check if arp_update script is stuck and restart the script if needed.
+    """
+    if is_process_running("arp_update"):
+        if is_arp_update_stuck():
+            log_warning("arp_update process is stuck. Restarting...")
+            restart_arp_update()
+            sys.exit(2)
```
theasianpianist, line 101: Are these exit codes standard for monit/do we even need to have distinct exit codes? Looks like the monit script is only checking for `!= 0`, so different exit codes wouldn't provide any extra information

files/image_config/monit/arp_update_checker
```diff
@@ -0,0 +1,106 @@
...
+    except subprocess.CalledProcessError:
+        return False
+
+def is_arp_update_stuck(timeout=TIMEOUT):
+    """Check if arp_update is stuck by looking for a ping command running longer than the specified timeout. (Default 1sec)"""
+    try:
+        arp_update_pid = subprocess.check_output(["pgrep", "-x", "arp_update"]).strip().decode('utf-8')
+        child_pids = subprocess.check_output(["pgrep", "-P", arp_update_pid]).strip().decode('utf-8').split()
+        for pid in child_pids:
+            cmd = subprocess.check_output(["ps", "-p", pid, "-o", "comm="]).strip().decode('utf-8')
```
theasianpianist, line 47: Could we also grab `etime` or `etimes` here and save a subprocess call?

files/image_config/monit/arp_update_checker
Ndancejic: updated in the latest revision

files/image_config/monit/arp_update_checker
Ndancejic: looking at the other scripts I don't notice a standard/pattern, I changed both to be '1'

files/image_config/monit/arp_update_checker
Ndancejic: yeah good catch, I combined both subprocess calls into one and output to a tuple

Ndancejic:
Pipeline is failing because the program is not in the expected location:
```
"stderr": "/etc/monit/conf.d/sonic-host:61: Program does not exist: '/usr/bin/arp_update_checker'
```
Will update the PR with a fix

Ndancejic:
I noticed while updating this, a lot of scripts are located in src/sonic-utilities/scripts/ ... would it be better to put this script there as well? that would require a sonic-utilities PR and a submodule update.

theasianpianist (approved):
lgtm. please update PR description to include patch.

siqbal1986 (approved):

prsunny (approved):

kperumalbfn:
@Ndancejic Required in 202405 and 202311 branches?

kperumalbfn (approved):

PR #19952, opened 2024-08-19
Update dockerfiles to do staged builds

saiarcot895:
#### Why I did it

On newer versions of Docker, only the buildkit builder is supported, and
cannot be disabled by setting DOCKER_BUILDKIT to 0. The side effect of
this is that the behavior of `--squash` is different (see
moby/moby#38903). This will result in the container sizes being
significantly higher.

##### Work item tracking
- Microsoft ADO **(number only)**: 29277538

#### How I did it

To work around this, make all of our builds two-stage builds, with the
`--squash` flag entirely removed. The way this works is in the first
stage, whatever new files/packages need to be added are added (along
with files/packages that need to be removed). Then, in the second stage,
use rsync to copy over the changed files as a single command/layer. In the
case of the base layer for each Debian version, the final result of the first
stage will be copied into an empty base layer.

As part of this, also consolidate the container cleanup code into
`post_run_cleanup`, and remove it from the individual containers (for
consistency). Also experiment a bit with not needing to explicitly
install library dependencies, and let apt install it as necessary. This
will help during upgrades in the case of ABI changes for packages.

Also, remove the `SONIC_USE_DOCKER_BUILDKIT` option, and don't
set `DOCKER_BUILDKIT` option. This option will eventually have no impact.
This also means that builds will now use buildkit, as that is the default now.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]


dockers/docker-base-bookworm/Dockerfile.j2
```diff
@@ -1,26 +1,14 @@
...
 {% from "dockers/dockerfile-macros.j2" import install_debian_packages, install_python_wheels, copy_files %}
 {% if CONFIGURED_ARCH == "armhf" and (MULTIARCH_QEMU_ENVIRON == "y" or CROSS_BUILD_ENVIRON == "y") %}
-FROM --platform=linux/arm/v7 {{ prefix }}debian:bookworm
+ARG BASE=--platform=linux/arm/v7 {{ prefix }}debian:bookworm
 {% elif CONFIGURED_ARCH == "arm64" and (MULTIARCH_QEMU_ENVIRON == "y" or CROSS_BUILD_ENVIRON == "y") %}
-FROM --platform=linux/arm64 {{ prefix }}debian:bookworm
+ARG BASE=--platform=linux/arm64 {{ prefix }}debian:bookworm
 {% else %}
-FROM {{ prefix }}{{DOCKER_BASE_ARCH}}/debian:bookworm
+ARG BASE={{ prefix }}{{DOCKER_BASE_ARCH}}/debian:bookworm
```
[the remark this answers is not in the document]
saiarcot895: Agreed. I want to keep the focus of this PR on unblocking Docker upgrades, but I had to bring in some space optimization stuff (see the `COPY` at the end of this file) to get things to work.

liushilongbuaa:
@saiarcot895 , how about sonic-slave-bookworm?

Makefile.work
```diff
@@ -310,6 +312,7 @@ DOCKER_RUN := docker run --rm=true --privileged --init \
     -e "https_proxy=$(https_proxy)" \
     -e "no_proxy=$(no_proxy)" \
     -i$(shell { if [ -t 0 ]; then echo t; fi }) \
+    --ulimit nofile=524288:524288 \
```
liushilongbuaa, line 315: Why bullseye needs this option but bookworm doesn't?

saiarcot895: This was needed with an older version of changes that included a Docker daemon version upgrade, but isn't needed anymore. It may be needed in the future when the upgrade is done.

There was previously a commit to upgrade Docker to version 25, for the purposes of using containerd image store. With that upgrade, docker-in-docker in the Bullseye slave container did not start up because of a ulimit error. In the Bullseye slave container, the ulimit for the max open files was set to 1048576:1048576 (as in, 1048576 for both the soft and hard limit). However, Docker 25 and newer lowered just the hard limit to 524288 in moby/moby@c8930105b. This caused the soft limit to be higher than the hard limit, which is an issue, and so starting up docker failed. Based on local testing, the Bookworm slave container there appeared to be not affected because this ulimit was 1024:1048576, and so even lowering just the hard limit to 524288 would be fine.

Rechecking my testing now, when starting up a slave container locally, I see that both Bullseye and Bookworm have 1048576:1048576; I'm not sure what changed to also cause Bookworm to be potentially affected.

This issue affects us only after we upgrade to Docker 25 or newer; we're still on Docker 24 for now. I can either keep this change or take it out of the PR.

dockers/docker-config-engine-bookworm/Dockerfile.j2
```diff
@@ -49,10 +51,17 @@ COPY ["files/readiness_probe.sh", "/usr/bin/"]
...
+        libz-dev
+{%- endif %}
+
+RUN apt-get purge -y        \
+        python3-dev         \
+        build-essential
+
+FROM $BASE
+
+RUN --mount=type=bind,from=base,target=/changes-to-image rsync -axAX --no-D --exclude=/sys --exclude=resolv.conf /changes-to-image/ /
```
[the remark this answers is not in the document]
saiarcot895: Taking a `sonic-broadcom.bin` image built by the pipeline, from around January 15th, the `sonic-broadcom.bin` file with this PR is about 80MB smaller than the `sonic-broadcom.bin` from January 15th official build (921MB vs 1002MB). The `docker` directory, after extraction, is about 170MB smaller (1592MB vs 1766MB).

dockers/docker-base-bookworm/Dockerfile.j2
```diff
@@ -98,17 +88,14 @@ RUN apt-get -y purge   \
...
 COPY ["etc/rsyslog.d/*", "/etc/rsyslog.d/"]
 COPY ["root/.vimrc", "/root/.vimrc"]
 
 RUN ln /usr/bin/vim.tiny /usr/bin/vim
 
 COPY ["etc/supervisor/supervisord.conf", "/etc/supervisor/"]
+
+FROM scratch
+
+COPY --from=base / /
```
liushilongbuaa, line 101: Why docker-base use COPY but other dockers use RUN rsync?

saiarcot895: `rsync` is run from the base/source layer, and the new contents of the new image is mounted in a directory. In the case of `docker-base-bookworm`, the base layer is `debian:bookworm`, which doesn't have `rsync` installed, so `rsync` can't be run from there. Since `rsync` is installed in `docker-base-bookworm`, and everything else is on top of this image, `rsync` can be used elsewhere.

While it might be technically possible to run `rsync` from that mounted directory, it's far easier and more reliable to just copy the final results of `docker-base-bookworm` into an empty layer. This results in two things:
* When this container gets built, `docker-base-bookworm` will be the top-level image for any container that gets built from this container, instead of being `debian:bookworm`.
* For files that get removed in this container, those files will not be present in the final container build at all. Currently, when any file/directory is removed, there is a "whiteout" file that gets added into the container indicating that the file/directory referenced in the base layer no longer exists, but the space by that file/directory is still taken up. Now, that file/directory will not exist at all, and thus not take up disk space.

liushilongbuaa (approved):

platform/vs/syncd-vs.mk
```diff
@@ -3,13 +3,6 @@ $(LIBSAIREDIS)_DEB_BUILD_PROFILES += syncd vs
 SYNCD_VS = syncd-vs_1.0.0_$(CONFIGURED_ARCH).deb
 $(SYNCD_VS)_RDEPENDS += $(LIBSAIREDIS) $(LIBSAIMETADATA) $(LIBSAIVS)
 
-ifeq ($(BLDENV),bookworm)
```
liushilongbuaa, line 6: Why remove dependency on DASH SAI?

saiarcot895: Good catch! This was a change I had made locally to speed up my local build and didn't mean to commit it. Removed.

slave.mk
```diff
@@ -518,13 +518,6 @@ $(shell DBGOPT='$(DBGOPT)' scripts/prepare_slave_container_buildinfo.sh $(SLAVE_
 endif
 include Makefile.cache
 
-ifeq ($(SONIC_USE_DOCKER_BUILDKIT),y)
```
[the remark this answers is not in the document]
saiarcot895: `DOCKER_BUILDKIT` will default to enabled for building SONiC containers. Note that for the slave containers, that will depend on the host's docker settings/defaults.

saiarcot895:
> ```
> RUN apt-get -y purge   \
>     exim4              \
>     exim4-base         \
>     exim4-config       \
>     exim4-daemon-light
> ```
> 
> I think you can also remove these lines from base-bookworm and base-bullseye:
> 
> ```
> Step 29/41 : RUN apt-get -y purge       exim4                  exim4-base             exim4-config           exim4-daemon-light
>  ---> Running in 5de49fe2783c
> Reading package lists...
> Building dependency tree...
> Reading state information...
> Package 'exim4' is not installed, so not removed
> Package 'exim4-base' is not installed, so not removed
> Package 'exim4-config' is not installed, so not removed
> Package 'exim4-daemon-light' is not installed, so not removed
> 0 upgraded, 0 newly installed, 0 to remove and 3 not upgraded.
> ```
> 
> But probably it's out of scope for this PR.

I've taken in this change for docker-base-bookworm, since it's small enough. I didn't update bullseye, partly because I'm hoping we're getting to the point where we won't need it.

platform/broadcom/docker-syncd-brcm-dnx-rpc/Dockerfile.j2
```diff
@@ -1,5 +1,7 @@
 {% from "dockers/dockerfile-macros.j2" import install_debian_packages, install_python_wheels, copy_files %}
-FROM docker-syncd-brcm-dnx-{{DOCKER_USERNAME}}:{{DOCKER_USERTAG}}
+ARG BASE=docker-syncd-brcm-dnx-{{DOCKER_USERNAME}}:{{DOCKER_USERTAG}}
+
+FROM $BASE as base
```
liushilongbuaa, line 4: Why dockers/**/Dockerfile is using 'FROM AS', platform/**/Dockerfile is using 'FROM as'?

saiarcot895: Functionally speaking, there's no difference. I think it was just a copy-paste difference. However, I _think_ newer versions of Docker will print a warning when seeing lowercase `as`, but won't fail the build.

src/sonic-build-hooks/scripts/post_run_cleanup
```diff
@@ -15,18 +15,17 @@ if [ ! -z "$(get_version_cache_option)" ]; then
...
-apt-get -s autoclean -y
-apt-get -s autoremove -y
-#apt-get -s autoremove -y --purge
+apt-get clean -y
+apt-get autoclean -y
+apt-get autoremove -y --purge
 rm -f /var/cache/apt/archives/*.deb /var/cache/apt/*.bin
 
 if [[ ! ${IMAGENAME} =~ -slave- ]]; then
 	rm -f /var/lib/apt/lists/*
```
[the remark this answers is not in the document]
saiarcot895: Added.

platform/broadcom/docker-syncd-brcm-dnx-rpc/Dockerfile.j2
saiarcot895: Fixed case.

Issue #20019, opened 2024-08-25
Two issues when using config node as "unified"


eddieruan-alibaba:
#### Description

When using config mode as "unified", two issues observed during testing.

1. frr.conf can't be generated via https://github.com/sonic-net/sonic-buildimage/blob/master/dockers/docker-fpm-frr/docker_init.sh#L91

The following error is seen
  File "/usr/local/bin/sonic-cfggen", line 502, in <module>
    main()
  File "/usr/local/bin/sonic-cfggen", line 466, in main
    template_data = template.render(data)
  File "/usr/local/lib/python3.9/dist-packages/jinja2/environment.py", line 1304, in render
    self.environment.handle_exception()
  File "/usr/local/lib/python3.9/dist-packages/jinja2/environment.py", line 939, in handle_exception
    raise rewrite_traceback_stack(source=source)
  File "/usr/share/sonic/templates/gen_frr.conf.j2", line 2, in top-level template code
    {% include "/usr/local/sonic/frrcfgd/frr.conf.j2" %}
  File "/usr/local/sonic/frrcfgd/frr.conf.j2", line 23, in top-level template code
    {% include "staticd.db.default_route.conf.j2" %}
  File "/usr/local/lib/python3.9/dist-packages/jinja2/loaders.py", line 207, in get_source
    raise TemplateNotFound(template)
jinja2.exceptions.TemplateNotFound: staticd.db.default_route.conf.j2

When using -t /usr/share/sonic/templates/gen_frr.conf.j2  sonic-cfggen would only add /usr/share/sonic/templates/ in j2 path. 

https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-config-engine/sonic-cfggen#L468

But when using unified mode, we pick frr.conf.j2 from 
    {% include "/usr/local/sonic/frrcfgd/frr.conf.j2" %}

which causes this issue.

The fix is to use -T to provide additional Template directory.

diff --git a/dockers/docker-fpm-frr/docker_init.sh b/dockers/docker-fpm-frr/docker_init.sh
index 0ed274ec7..662a78b5a 100755
--- a/dockers/docker-fpm-frr/docker_init.sh
+++ b/dockers/docker-fpm-frr/docker_init.sh
@@ -88,6 +88,7 @@ elif [ "$CONFIG_TYPE" == "unified" ]; then
     CFGGEN_PARAMS=" \
         -d \
         -y /etc/sonic/constants.yml \
[...]

PR #20020, opened 2024-08-25
Fix Two issues when using config node as unified #20019

eddieruan-alibaba:
#### Why I did it
Described in https://github.com/sonic-net/sonic-buildimage/issues/20019

##### Work item tracking

#### How to verify it
Manual run via 
sonic-cfggen -d         -y /etc/sonic/constants.yml         -T /usr/local/sonic/frrcfgd         -t /usr/share/sonic/templates/gen_frr.conf.j2,/etc/frr/frr.conf 

and run it with integrated image.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]


src/sonic-frr-mgmt-framework/templates/bgpd/bgpd.conf.db.nbr_af.j2
```diff
@@ -112,16 +112,12 @@
 {% endif %}
 {# ------- route-map in --------------------------- #}
 {% if 'route_map_in' in n_af_val %}
-{% for map in n_af_val['route_map_in'] %}
-  neighbor {{nbr_name}} route-map {{map}} in
```
venkatmahalingam, line 116: This code helps to associate multiple route-maps to a neighbor, what issue did you find?

dockers/docker-fpm-frr/docker_init.sh
```diff
@@ -88,6 +88,7 @@ elif [ "$CONFIG_TYPE" == "unified" ]; then
     CFGGEN_PARAMS=" \
         -d \
         -y /etc/sonic/constants.yml \
+        -T /usr/local/sonic/frrcfgd \
```
venkatmahalingam, line 91: This might break if anyone uses "unified" config type for bgpcfgd. Did you try making the necessary changes in https://github.com/sonic-net/sonic-buildimage/blob/master/dockers/docker-fpm-frr/frr/gen_frr.conf.j2#L2?

src/sonic-frr-mgmt-framework/templates/bgpd/bgpd.conf.db.nbr_af.j2
eddieruan-alibaba: This is the second issue listed in https://github.com/sonic-net/sonic-buildimage/issues/20019

In frrcfgd.py, it picks the one input as well

https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-frr-mgmt-framework/frrcfgd/frrcfgd.py#L1827

dockers/docker-fpm-frr/docker_init.sh
eddieruan-alibaba: This is the first issue listed in https://github.com/sonic-net/sonic-buildimage/issues/2001.

Without this -T, unified mode would not be able to generatee frr.conf. I explained the cause in the issue.

eddieruan-alibaba: -T provides an additional path from original path. 

a.k.a both 
/usr/local/sonic/frrcfgd
and 
/usr/share/sonic/templates

In j2 template env.

https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-config-engine/sonic-cfggen#L463

src/sonic-frr-mgmt-framework/templates/bgpd/bgpd.conf.db.nbr_af.j2
venkatmahalingam: This means that the data stored in the neighbor_af table is not in list. Refer the below YANG for syntax (use leaf-list)

https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-yang-models/yang-models/sonic-bgp-common.yang#L385

dockers/docker-fpm-frr/docker_init.sh
venkatmahalingam: I would suggest going with new config_type for better control. It may be unmanageble when unified is overloaded for multiple use-cases.

src/sonic-frr-mgmt-framework/templates/bgpd/bgpd.conf.db.nbr_af.j2
eddieruan-alibaba: How does frrcfgd.py handle it, not a list, right?

dockers/docker-fpm-frr/docker_init.sh
eddieruan-alibaba: This change is to fix existing unified mode issue. Currently frr.conf would not be able to generated in unfied mode.

src/sonic-frr-mgmt-framework/templates/bgpd/bgpd.conf.db.nbr_af.j2
venkatmahalingam: No, frrcfgd should handle the value in a list.

dockers/docker-fpm-frr/docker_init.sh
venkatmahalingam: As long as this change doesnt break anything, for example, bgpcfgd operation, I'm good.

src/sonic-frr-mgmt-framework/templates/bgpd/bgpd.conf.db.nbr_af.j2
eddieruan-alibaba: Ok. I will revert this change and change the original issue 2 to add list handling in frrcfgd.

dockers/docker-fpm-frr/docker_init.sh
eddieruan-alibaba: Yes. There is no issue. These paths just provide j2's include file's searching path. 
Since gen_frr.conf.j2 uses two template files, we need to include both of them.

venkatmahalingam (approved):

kperumalbfn:
@eddieruan-alibaba Could you update the branch and retry? Recent commits have all sanities passed.

eddieruan-alibaba:
> @eddieruan-alibaba Could you update the branch and retry? Recent commits have all sanities passed.

@kperumalbfn I still hit a failure.

The log I saw is at https://elastictest.org/scheduler/publictestplan/66da5d8f9ae803ee15b3c380?testcase=syslog%2Ftest_logrotate.py%7C%7C%7C2&type=console

which has the following error message. I don't feel this failure is related to my changes.

syslog/test_logrotate.py::test_orchagent_logrotate[vlab-03-None-1-5] 
-------------------------------- live log call ---------------------------------
03:32:59 utilities.wait_until                     L0142 ERROR  | Exception caught while checking no_pending_entries:Traceback (most recent call last):
  File "/var/src/sonic-mgmt/tests/common/utilities.py", line 138, in wait_until
    check_result = condition(*args, **kwargs)
  File "/var/src/sonic-mgmt/tests/syslog/test_logrotate.py", line 260, in no_pending_entries
    return not bool(get_pending_entries(duthost, ignore_list=ignore_list))
  File "/var/src/sonic-mgmt/tests/syslog/test_logrotate.py", line 243, in get_pending_entries
    pending_entries.remove(entry)
KeyError: '_ROUTE_TABLE:20c0:e9b0::/64'
, error:'_ROUTE_TABLE:20c0:e9b0::/64'
03:33:01 utilities.wait_until                     L0142 ERROR  | Exception caught while checking

kperumalbfn (approved):

PR #19510, opened 2024-07-09
Pin pip to version 24.2, and update disable-non-manylinux.patch

saiarcot895:
#### Why I did it

Pip has been recently updated to 24.2. For armhf, there is a patch file that gets applied to disable the use of manylinux wheels, to avoid the use of packages that might be using a newer glibc version than what is available on the system. This patch file was designed for 24.0, but can't be applied to 24.2.

Update the patch file to apply on 24.2, and make sure version 24.2 of pip gets installed.

This also fixes the [version upgrade pipeline errors for master branch](https://dev.azure.com/mssonic/build/_build?definitionId=40). 

##### Work item tracking
- Microsoft ADO **(number only)**: 28835746

#### Which release branch to backport (provide reason below if selected)

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

#### Tested branch (Please provide the tested image version)

- [ ] 
- [ ]

PR #20037, opened 2024-08-27
[202311] Pin pip to version 24.2, and update disable-non-manylinux.patch

saiarcot895:
Cherry-pick of #19510

#### Why I did it

Pip has been recently updated to 24.2. For armhf, there is a patch file that gets applied to disable the use of manylinux wheels, to avoid the use of packages that might be using a newer glibc version than what is available on the system. This patch file was designed for 24.0, but can't be applied to 24.2.

##### Work item tracking
- Microsoft ADO **(number only)**: 28835746

#### How I did it

Update the patch file to apply on 24.2, and make sure version 24.2 of pip gets installed.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/sonic-slave-bullseye/Dockerfile.j2
+++ b/sonic-slave-bullseye/Dockerfile.j2
@@ -487,7 +487,7 @@ RUN wget -O golang-go.deb 'https://sonicstorage.blob.core.windows.net/public/fip
  && rm golang-go.deb golang-src.deb
 {%- endif %}
 
-RUN pip3 install --upgrade pip
+RUN pip3 install --upgrade pip==24.2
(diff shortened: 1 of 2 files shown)


liushilongbuaa (approved):
LGTM

liushilongbuaa:
@saiarcot895 , It seems that this PR is not a blocker issue.

saiarcot895:
@liushilongbuaa Ah, you're right. The version upgrade pipeline pins the python package versions for release branches. However, builds that don't use version controlling at all would have been broken.

PR #20075, opened 2024-08-29
Sort the system port ID generation in minigraph parser for Chassis

judyjoseph:
#### Why I did it
The system_port_id generation was based on the loop below

for interface in **interface_metadata.findall**(str(QName(ns1, "DeviceInterfaceMetadata"))):

"DeviceInterfaceMetadata" defined in the minigraph in DeviceInfo section, which is per interface and in this loop we increment the system_port_id++ so that each interface will have a unique ID. 

The for loop was based on interface_metadata list extracted by findall() API matching tag **"DeviceInterfaceMetadata"** [lxml.etree._Element](https://lxml.de/api/lxml.etree._Element-class.html). findall() doesn't guarantee document order. Hence the interface list and the corresponding system_port ids generated - has a possibility of not matching across config_db's in different linecards of a chassis.

When SYSTEM_PORT table entries hve mismatch across linecards, a few line cards behaving erratically, resulting in continuous pkt error interrupts getting fired and the IBGP sessions not getting established with other peer ASIC's in other line cards.

##### Work item tracking
- Microsoft ADO **29260084**:

#### How I did it
Add logic to do a sort of the system_ports dictionary based on the key (eg: "str-sonic-lc03|ASIC0|Ethernet120") and assign the system_port_id in an incremental way. 

This makes sure the system_port_ids in SYSTEM_PORT table in config_db matches in all linecards/asic

Thanks to @abdosi and @vmittal-msft in triaging and coming to this solution.

#### How to verify it

Verified by manually patching this logic in the minigraph parser in the sonic T2 chassis and make sure the dockers, interfaces, IBGP, EBGP comes up in all linecards across the chassis.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/src/sonic-config-engine/minigraph.py
+++ b/src/sonic-config-engine/minigraph.py
@@ -341,14 +341,13 @@ def parse_chassis_deviceinfo_intf_metadata(device_info, chassis_linecards_info, 
             if linecard_name is not None:
                 key = "%s|%s" % (linecard_name, key)
             system_ports[key] = {
-                "system_port_id": system_port_id,
+                "system_port_id": 0,
                 "switch_id": switch_id,
                 "core_index": core_id,
                 "core_port_index": core_port_id,
                 "speed": intf_speed,
                 "num_voq": num_voq
             }
-            system_port_id += 1
 
         chassis_port_alias.setdefault(slot_index, {}).update(
             {(intf_sonic_name, intf_speed): intf_name})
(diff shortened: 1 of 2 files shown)


abdosi (approved):

vmittal-msft (approved):

judyjoseph:
Validated on Nokia/Arista voq chassis - IBGP/EBGP up. @arlakshm @rlhui please review and merge .

lguohan:
@qiluo-msft , this is problem we discussed, people keep adding logic on minigraph parser, and this is going to break your flow. we need to catch such case.

qiluo-msft:
For this one, it is good to merge. We will model the config generation later for this table.

---
In reply to: [2323018102](https://github.com/sonic-net/sonic-buildimage/pull/20075#issuecomment-2323018102) [](http://example.com/codeflow?ancestors=2323018102)

PR #20143, opened 2024-09-05
[POC] verify Improve routeorch to stop process routes when high priority notification coming.

liuh-80:
please ignore, verify this PR:
https://github.com/sonic-net/sonic-swss/pull/3278


liuh-80:
Orch agent is performance sensitive, the code change in this PR will make sonic-mgmt test case failed because the code change will greatly increase latency:

With code change:
2024 Sep 10 06:44:00.943713 vlab-01 INFO python[15294]: ansible-shell_cmds Invoked with cmds=['config interface  startup 
2024 Sep 10 06:48:08.014114 vlab-01 NOTICE swss#orchagent: :- doPortTask: Set port Ethernet84 admin status to up
    
With out change.
2024 Sep 10 09:19:15.972793 vlab-01 INFO python[63586]: ansible-shell_cmds Invoked with cmds=['config interface  startup 
2024 Sep 10 09:19:46.827653 vlab-01 NOTICE swss#orchagent: :- doPortTask: Set port Ethernet84 admin status to up

Need consider other solution:
1. select and pop data in another thread.
2. decrease the check frequency.

PR #20157, opened 2024-09-05
[Arista] No-op quicksilver qos buffer templates, remove bcm baseline configs

rick-arista:
#### Why I did it
These configs are known to be incorrect and break some testcases, e.g.
the zero-ed out baseline config will cause no multicast traffic egress.
This baseline config was provided to fix a SAI error caused by incorrect
values in the templates, and is unnecessary if the templates result in
no values being added to config db

##### Work item tracking

- [x] 202311
- [x] 202405

No-op quicksilver qos buffer templates, remove bcm baseline configs


r12f (approved):
lgtm! we need to get brcm to get the valid config for these platforms.

kperumalbfn:
Thanks @rick-arista We could update buffer configs for other SKUs after validation.

kperumalbfn (approved):
LGTM

ccroy-arista:
I believe this will conflict with 202405 branch, which I opened a separate PR for: https://github.com/sonic-net/sonic-buildimage/pull/20231

PR #20231, opened 2024-09-11
Platform quicksilver noop buffers

ccroy-arista:
#### Why I did it
These configs are known to be incorrect and break some testcases, e.g.
the zero-ed out baseline config will cause no multicast traffic egress.
This baseline config was provided to fix a SAI error caused by incorrect
values in the templates, and is unnecessary if the templates result in
no values being added to config db

This PR is specifc to 202405. Corresponding PR for master (and 202311) can be found here: https://github.com/sonic-net/sonic-buildimage/pull/20157

##### Work item tracking

#### How to verify it
Sanitize DUT and verify that interfaces come up.

- [x] 202405

202405

No-op quicksilver qos buffer templates, remove bcm baseline configs


StormLiangMS:
hi @ccroy-arista normally we should file PR to master branch then cherry pick back if approved.

StormLiangMS:
hi @bingwang-ms could you help to cherry pick this one? This is for TH5.

bingwang-ms (approved):

PR #20232, opened 2024-09-11
Replace straightforward sonic-cfggen calls with sonic-db-cli

vivekrnv:
#### Why I did it

Fix https://github.com/sonic-net/sonic-buildimage/issues/20048

It is explained in ticket above, how sonic-cfggen calls in the hostname.sh, pcie-check.sh and banner-config and hogging the CPU and slightly delaying the start of swss

**pcie-check.sh**
redis-cli call is also replaced with sonic-db-cli since redis-cli is a wrapper around the actual redis-cli under database container
image

**swss start**
If the SKU has create_only_config_db_buffers.json set to true, the only config that's updated today is. 
```
{
    "DEVICE_METADATA": {
        "localhost": {
            "create_only_config_db_buffers": "true"
        }
    }
}
```
We use sonic-cfggen which cause 1.5 sec delay in the start of swss container. Thus replace with sonic-db-cli. If a complex use case arises in future, this can be updated.
Screenshot 2024-09-16 at 12 45 05 PM

##### Work item tracking

#### How to verify it

**Note: Everything was tested on MSN2700 device, Intel Celeron CPU with 2 cores**

- 1.5 sec saved in the swss container start
image

- 1.6 sec saved in the start of swss service. Previously, it took almost 1.8 sec after config-setup is finished for the swss to start. After replacing the calls, it start almost 0.2 sec after config-setup is finished

In total, anywhere between 2.5 - 3.5 sec is saved

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]


oleksandrivantsiv (approved):

stepanblyschak (approved):

files/image_config/pcie-check/pcie-check.sh
```diff
@@ -54,7 +54,7 @@ function check_and_rescan_pcie_devices()
 
      done
      debug "PCIe check failed"
-     redis-cli -n 6 HSET $PCIE_STATUS_TABLE "status" "FAILED"
```
vaibhavhd, line 57: Is this change also needed for the reasons mentioned in the description? Or is this added for completeness (make all sonic-db-cli calls) reasons?

vaibhavhd (approved):

files/image_config/pcie-check/pcie-check.sh
vivekrnv: This might have a very slight performance improvement. not significant i think. So i'd put this in the completeness category

vivekrnv:
@qiluo-msft Can we merge this?

qiluo-msft (approved):

PR #20393, opened 2024-10-01
Telemetry Enabled for Network Monitoring

sabakram:
#### Why I did it
Enabled Telemetry to setup a network monitoring system for SONiC
##### Work item tracking

#### How I did it
Made changes in rules/config file.


zbud-msft:
Why can't gnmi docker be used?

sabakram:
> Why can't gnmi docker be used?

gNMI port 50051 not working therefore trying to use telemetry.
Please help if there is an alternative way to use it.

PR #20400, opened 2024-10-01
Add Dummy Docker for Kubernetes Deployment

isabelmsft:
#### Why I did it
Pilot Kubernetes daemonset using dummy docker
##### Work item tracking

#### How I did it
Add dummy docker to SONiC buildimage
#### How to verify it
Build image, docker target, and test

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]


dockers/docker-dummyk8s/Dockerfile.j2
```diff
@@ -0,0 +1,19 @@
...
+ENV IMAGE_VERSION=$image_version
+
+# Update apt's cache of available packages
+RUN apt-get update
+
+COPY ["readiness-probe.sh", "/usr/bin"]
+
+RUN chmod +x /usr/bin/readiness-probe.sh
+
+ENTRYPOINT ["tail", "-f", "/dev/null"]
```
make1980, line 19: should we just sleep instead? maybe just sleep for 15 minutes so that we can also verify that k8s can restart our container?

https://stackoverflow.com/questions/43843079/using-tail-f-dev-null-to-keep-container-up-fails-unexpectedly

isabelmsft: Thanks, updated to sleep instead

make1980 (approved):

dockers/docker-dummyk8s/readiness-probe.sh
lixiaoyuner, line 1: https://github.com/sonic-net/sonic-buildimage/pull/15223/files, we have a readiness-probe in the base container. You add this one change this probe name to /usr/bin/readiness_probe_hook

lixiaoyuner, line 1: If want to use the upgrade function(kube to local, local to kube), may need add a start service which is like this https://github.com/sonic-net/sonic-buildimage/blob/master/dockers/docker-sonic-telemetry/start.sh

PR #20404, opened 2024-10-02
Refactor DHCP sethostname|sethostname6 scripts

tomvil:
#### Why I did it
I noticed that /etc/hosts file is getting constantly filled with new records when dhclient is enabled.
```
root@test:~# cat /etc/hosts | wc -l
246
```

DHCP is providing FQDN, while we check only against short hostname and this will never match.
```
root@test:~# hostname
test.lab.com
root@test:~# hostname -s
test
```

And DHCP offer:
```
root@test:~# tcpdump -i eth0 -n -vv port 67 or port 68
tcpdump: listening on eth0, link-type EN10MB (Ethernet), snapshot length 262144 bytes
...
	    Hostname (12), length 30: "test.lab.com"
...
```

`sethostname6` also uses `hostname` instead of `hostname -s`

##### Work item tracking

#### How I did it
Refactored sethostname and sethostname6 scripts to handle option 15 and not to duplicate new entries

#### How to verify it

- Enable DHCP client on the switch
- Receive offer with FQDN hostname option
- Without this fix everytime new entry will be added to /etc/hosts
- With this fix no new entry will be added

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305
- [x] 202405

202405

Refactor sethostname and sethostname6 scripts

### DIFF
--- a/files/dhcp/sethostname
+++ b/files/dhcp/sethostname
@@ -1,12 +1,21 @@
 case $reason in
     BOUND|RENEW|REBIND|REBOOT)
-        current_host_name=`hostname -s`
-        if [ "$current_host_name" != "$new_host_name" ] && [ -n "$new_host_name" ]
-        then
-            echo $new_host_name > /etc/hostname
-            hostname -F /etc/hostname
-            sed -i "/\s$current_host_name$/d" /etc/hosts
-            echo "127.0.0.1 $new_host_name" >> /etc/hosts
+        if [ -n "$new_host_name" ]; then
+            current_full_hostname=$(hostname -f)
+
+            desired_full_hostname=$new_host_name
+            if [ "$new_host_name" = "${new_host_name%%.*}" ] && [ -n "$new_domain_name" ]; then
+                desired_full_hostname="${new_host_name}.${new_domain_name}"
+            fi
+
+            if [ "$current_full_hostname" != "$desired_full_hostname" ]; then
+                echo "127.0.0.1 $desired_full_hostname ${desired_full_hostname%%.*}" >> /etc/hosts
+
+                echo "$desired_full_hostname" > /etc/hostname
+                hostname -F /etc/hostname
+
+                sed -i "/\s$current_full_hostname\s\+${current_full_hostname%%.*}$/d" /etc/hosts
+            fi
         fi
         ;;
 esac
--- a/files/dhcp/sethostname6
+++ b/files/dhcp/sethostname6
@@ -1,14 +1,17 @@
 case $reason in
     BOUND6|RENEW6|REBIND6|REBOOT)
-        current_fqdn_fqdn=`hostname`
-        if [ "$current_fqdn_fqdn" != "$new_fqdn_fqdn" ] && [ -n "$new_fqdn_fqdn" ]
-        then
-            echo $new_fqdn_fqdn > /etc/hostname
-            hostname -F /etc/hostname
-            sed -i "/\s$current_fqdn_fqdn$/d" /etc/hosts
-            sed -i "/\s$new_fqdn_fqdn$/d" /etc/hosts
-            echo "127.0.0.1 $new_fqdn_fqdn" >> /etc/hosts
-            echo ":: $new_fqdn_fqdn" >> /etc/hosts
+        if [ -n "$new_fqdn_fqdn" ]; then
+            current_full_hostname=$(hostname -f)
+
+            if [ "$current_full_hostname" != "$new_fqdn_fqdn" ]; then
+                echo "127.0.0.1 $new_fqdn_fqdn ${new_fqdn_fqdn%%.*}" >> /etc/hosts
+                echo "::1 $new_fqdn_fqdn ${new_fqdn_fqdn%%.*}" >> /etc/hosts
+
+                echo "$new_fqdn_fqdn" > /etc/hostname
+                hostname -F /etc/hostname
+
+                sed -i "/\s$current_full_hostname\s\+${current_full_hostname%%.*}$/d" /etc/hosts
+            fi
         fi
         ;;
 esac


tomvil:
Hey @qiluo-msft, could you review this or perhaps know someone who can? :)

tomvil:
hey @saiarcot895 could you review this?

wdoekes:
Funny, I just debugged this myself and came up with:
```sh
case $reason in
    BOUND|RENEW|REBIND|REBOOT)
        if [ -n "$new_host_name" ]; then
            current_host_long=$(hostname -f)
            current_host_short=$(hostname -s)

            case "$new_host_name" in
            *.*)
                current_host_name=$current_host_long
                new_host_short=${new_host_name%%.*}
                new_host_long=$new_host_name
                ;;
            *)
                current_host_name=$current_host_short
                new_host_short=$new_host_name
                new_host_long=$new_host_short.$(hostname -d 2>/dev/null)
                new_host_long=${new_host_long%.}
                ;;
            esac
            if [ "$current_host_name" != "$new_host_name" ]; then
                echo $new_host_name > /etc/hostname
                hostname -F /etc/hostname
                echo "127.0.0.1 $new_host_long $new_host_short" >> /etc/hosts
                sed -i "/\s$current_host_long\s\+$current_host_short$/d" /etc/hosts
            fi
        fi
        ;;
esac
```
Maybe your simple fix is sufficient though.

See #20579

saiarcot895:
Are DHCP servers guaranteed to use the FQDN here? Reading RFC 2132, some DHCP servers might only return the host name in that field (option 12), whereas the domain name might be in option 15.

tomvil:
@saiarcot895 it may, but in my understanding these are two different things. Option 15 sets the `search` value in `resolv.conf`, so it's basically telling the client what to append (if needed) when resolving. It doesn't extend the Option 12, which tells what your hostname should be. They may be completely different values if you like. Unless you have something different in mind :)

wdoekes:
@tomvil: not entirely sure about that. See this ticket here:
https://github.com/systemd/systemd/issues/13903
There 119 is used for the search domain and 15 is explicitly for the hostname-domain. See also rfc3397. Can of worms :grimacing: 

I agree with @saiarcot895 that rfc2132 and its predecessor rfc1533 are clear in that 15 (3.17) should be used as the domain for 12 (3.14). _If the are no dots in 12, I think that's a safe assumption to make._

We can certainly adapt the script above to take those into account.

johndexter1:
@wdoekes Can you clarify where RFC 2132 specifically mentions combining these options? From my reading, it seems to indicate that the domain name should be used when resolving hostnames, rather than when constructing them. While I understand that this may be an assumption, I don't see any explicit reference to this in the document.

wdoekes:
> While I understand that this may be an assumption, I don't see any explicit reference to this in the document.

Explicit? No.

We have this "DHCP Options and BOOTP Vendor Extensions" https://datatracker.ietf.org/doc/html/rfc2132

### 3.14 (option 12):

> The name may or may not be qualified with the local domain name (see section 3.17 for the preferred way to retrieve the domain name).

### 3.17 (option 15):

> This option specifies the domain name that client should use when resolving hostnames via the Domain Name System.

If you _just_ read 3.17, then you will rightly say it goes into the "search" (previously "domain") setting in /etc/resolv.conf.

But if you read 3.14 first, it says:

- The name might have the local domain name,
- but if it doesn't, you should check option 15 for the local domain.

That leads me to the conclusion that: _shorthost (from 12) is in the domain (from 15) domain._

This also rhymes with RFC4702 behaviour:
https://datatracker.ietf.org/doc/html/rfc4702#section-4

> The server SHOULD send its notion of the complete FQDN for the client in the `Domain Name` field. The server MAY simply copy the `Domain Name` field from the `Client FQDN` option that the client sent to the server.

This ties the client FQDN together with the Domain Name.

Additionally this makes sense for the lookup-of-self: do a lookup of [shorthost]+[search-domain] to find [self] in dns.

And one can find documentation in various places that this is convention, like in these Cisco docs:
https://documentation.meraki.com/General_Administration/Cross-Platform_Content/Setting_Custom_DHCP_Options

image

---

### _Do we need to set the domain in /etc/hosts and or /etc/hostname?_

The convention has been to set the following in /etc/hosts:
```
127.0.0.1 shortname.domain.name shortname
```
As for `/etc/hostname` (and `hostname`) being `hostname -s` or `hostname -f`, **there is no consensus there.** I think we set it do the value of option 12. (Personally I prefer the FQDN in /etc/hostname.)

I think most compatible with expectations would be to set the hostname (and /etc/hostname) as received (short or FQDN), and set both FQDN and short in /etc/hosts (`127.0.0.1 shortname.domain.name shortname`). This ensures that gethostname() works.

wdoekes:
Example implentation:
```sh
case $reason in
    BOUND|RENEW|REBIND|REBOOT)
        if [ -n "${new_host_name:-}" ]; then
            # Check hostname for dots. Fall back to adding new_domain_name or
            # $(hostname -d).
            if [ "${new_host_name%.*}" != "$new_host_name" ]; then
                # It has dots. Assume FQDN.
                my_hostname=${new_host_name%%.*}
                my_domain=${new_host_name#*.}
                my_fqdn=${my_hostname}.${my_domain}
            elif [ -n "${new_domain_name:-}" ]; then
                # No dots in new_host_name, and new_domain_name exists. Add it.
                my_hostname=$new_host_name
                my_domain=$new_domain_name
                my_fqdn=${my_hostname}.${my_domain}
            else
                # No domain supplied. Take pre-existing domain.
                my_hostname=$new_host_name
                my_domain=$(hostname -d 2>/dev/null || true)
                if [ -n "$my_domain" ]; then
                    my_fqdn=${my_hostname}.${my_domain}
                else
                    my_fqdn=$my_hostname  # alas, no domain at all
                fi
            fi

            current_fqdn=$(hostname -f)  # might be short if no domain was set
            if [ "$current_fqdn" != "$my_fqdn" ]; then
                # Set the host name as received from DHCP; might be FQDN.
                echo $new_host_name > /etc/hostname
                hostname -F /etc/hostname
                # Store both the determined FQDN and the short hostname in
                # /etc/hosts.
                echo "127.0.0.1 $my_fqdn $my_hostname" >> /etc/hosts
                sed -i "/\s$current_fqdn\s\+${current_fqdn%%.*}$/d" /etc/hosts
            fi
        fi
        ;;
esac
```

tomvil:
Great discussion and information provided @wdoekes , thank you for that!

> Are DHCP servers guaranteed to use the FQDN here? Reading RFC 2132, some DHCP servers might only return the host name in that field (option 12), whereas the domain name might be in option 15.

So, getting back to this, as we discussed it's not guaranteed. But with this quick fix new records won't be added even if it's "shortname" provided. And we're kinda sure that it won't break anything else as `sethostname6` is using the same logic.

If we would like to go for completely rework of the logic maybe @wdoekes could open a new pull request?

tomvil:
Hey guys, I've had some free time over the weekend and came up with a (in my opinion) simplified version of what @wdoekes has. It takes Option 15 in to account as well. Works as expected with my test scenarios.

Looking forward to any feedback or further testing insights.

wdoekes:
Oh, and `\s*` is a no-op in that expression. If you want to guard for it being a whole word, you use `\b`:
```console
$ echo '127.0.0.1 osso.net' | sed -e '/\sosso.net\b/d'
```
```console
$ echo '127.0.0.1 osso.net osso' | sed -e '/\sosso.net\b/d'
```
vs.
```console
$ echo '127.0.0.1 osso.nets' | sed -e '/\sosso.net\b/d'
127.0.0.1 osso.nets
```
```console
$ echo '127.0.0.1 osso.nets osso' | sed -e '/\sosso.net\b/d'
127.0.0.1 osso.nets osso
```
^- revert `\b` to `\s*` here and observe how it incorrectly deletes the line

tomvil:
> You can move the current_short_hostname=$(hostname -s) into the nested if

Agree. Moved.

> And I would prefer doing the sed/d after the append, so that there's always at least one hostname available.

Hmm, this is a tricky situation. If hostname was `leaf1.a.com` and the new one is `leaf1.b.com` all new lines would get removed as the shortname would match. I think it's safer option to remove old records first and add new ones later. Although, I agree about that milisecond being not able to resolve the hostname. What do you think?

> If you want to guard for it being a whole word, you use \b:

Changed.

> maybe do the append before the hostname -F.

Agree on this. Changed.

wdoekes:
> Hmm, this is a tricky situation. If hostname was leaf1.a.com and the new one is leaf1.b.com all new lines would get removed as the shortname would match. I think it's safer option to remove old records first and add new ones later. Although, I agree about that milisecond being not able to resolve the hostname. What do you think?

I think the search-regex should be improved so it does not delete the new record =)

tomvil:
hehe :smile: Then I'll reuse the idea from your script (https://github.com/sonic-net/sonic-buildimage/pull/20404#issuecomment-2464801003) (if you don't mind) to always have two values near 127.0.0.1, so it's even easier and simpler to handle.

wdoekes:
So, I'd move the
```bash
                echo "$desired_full_hostname" > /etc/hostname
                hostname -F /etc/hostname
```
between the append and the delete.

Other than that, LGTM.

tomvil:
I did some testing in testing lab and noticed that IPv6 doesn't work as expected as it has different option for FQDN - Option 39 (https://www.rfc-editor.org/rfc/rfc4704.html), which refers to `new_dhcp6_fqdn` in dhcpd client. I've changed the logic for `sethostname6`.

> between the append and the delete.

Makes sense. Moved.

saiarcot895 (dismissed):

saiarcot895:
Thanks everyone for helping fix this!

tomvil:
@saiarcot895 thanks for the approve! Who could add labels for backport and merge this PR?

tomvil:
ping @saiarcot895 @vdahiya12

tomvil:
@qiluo-msft please help merge

tomvil:
Hello, @saiarcot895 @vdahiya12 could you help to merge this and backport to 202405?

tomvil:
@saiarcot895 @StormLiangMS I have resolved all conflicts, also updated my PR to include latest @saiarcot895 change to use `new_fqdn_fqdn` instead of `new_dhcp6_fqdn`

Can we merge this?

PR #20415, opened 2024-10-03
Add a patch for scapy to fix fd leak issue in AsyncSniffer

bingwang-ms:
#### Why I did it
This PR is to add a patch to fix potential fd leak issue in `AsyncSniffer` in `scapy` python library.
There are two fd leak scenarios.
1. When starting worker thread `_run`, if an interface is down, an `OSError` is thrown, and the sockets that have been created will  be leaked as it never got a chance to be closed.
2. When stopping the worker thread, same error can happen when calling `close`. The sockets not closed will be leaked.

Test code:
```
from scapy.all import *
import time


sniffing_iface = ["PortChannel101", "PortChannel102", "PortChannel106", "PortChannel1024"]
sniffstring = "udp"

sniffer = None

def process_packet(packet):
    print("Got a packet {}".format(len(packet)))

while True:
    sniffer = AsyncSniffer(iface=sniffing_iface, filter=sniffstring, prn=process_packet, store=False)
    try:
        sniffer.start()
        RETRY = 30
        while RETRY >= 0 and not hasattr(sniffer, 'stop_cb'):
            time.sleep(0.1)
            RETRY -= 1
        if RETRY < 0:
            print("Sniffer failed to start")
        else:
            print("Sniffer started")
        try:
            sniffer.stop()
            time.sleep(1)
            print("Sniffer stopped")
        except:
            pass
    except:
        time.sleep(1)
        pass
```
Run above code while toggling portchannels on the device. The opened fd is increasing.

##### Work item tracking
- Microsoft ADO **29735068**:

#### How I did it
Catch `OSError` when creating sockets, and catch any exception when closing socket to ensure all sockets are closed.

#### How to verify it
Verified by the testing code above. No fd leak happened.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305
- [x] 202311
- [x] 202405 

The change is tested on 202311 and 202405 branches.

- [ ] 
- [ ] 

 fix potential fd leak issue in `AsyncSniffer` in `scapy` python library.

No config change.

### DIFF
--- a/src/scapy.patch/0004-Fix-fd-leak-in-worker-thread.patch
+++ b/src/scapy.patch/0004-Fix-fd-leak-in-worker-thread.patch
@@ -1,0 +1,47 @@
+diff --git a/scapy/sendrecv.py b/scapy/sendrecv.py
+index f97fc415..dbcc372f 100644
+--- a/scapy/sendrecv.py
++++ b/scapy/sendrecv.py
+@@ -1111,17 +1111,19 @@ class AsyncSniffer(object):
+             # The _RL2 function resolves the L2socket of an iface
+             _RL2 = lambda i: L2socket or resolve_iface(i).l2listen()  # type: Callable[[_GlobInterfaceType], Callable[..., SuperSocket]]  # noqa: E501
+             if isinstance(iface, list):
+-                sniff_sockets.update(
+-                    (_RL2(ifname)(type=ETH_P_ALL, iface=ifname, **karg),
+-                     ifname)
+-                    for ifname in iface
+-                )
++                for ifname in iface:
++                    try:
++                        sniff_sockets.update({_RL2(ifname)(type=ETH_P_ALL, iface=ifname, **karg): ifname})
++                    except OSError:
++                        # Ignore OSError when opening the socket
++                        # The error can happen when the port goes down during the creation of the socket
++                        pass
+             elif isinstance(iface, dict):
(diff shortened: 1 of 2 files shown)


src/scapy.patch/0004-Fix-fd-leak-in-worker-thread.patch
```diff
@@ -0,0 +1,47 @@
...
+                 iface = iface or conf.iface
+                 sniff_sockets[_RL2(iface)(type=ETH_P_ALL, iface=iface,
+@@ -1221,7 +1223,11 @@ class AsyncSniffer(object):
+         self.running = False
+         if opened_socket is None:
+             for s in sniff_sockets:
+-                s.close()
++                try:
++                    s.close()
++                except Exception:
```
theasianpianist, line 42: Should we limit this to `OSError` as well? Have you seen other errors thrown in this step?

theasianpianist (approved):

src/scapy.patch/0004-Fix-fd-leak-in-worker-thread.patch
bingwang-ms: I only saw `OSError` in my debugging. But I think it doesn't hurt to catch all exceptions, since it's in the teardown path.

PR #20506, opened 2024-10-15
[FC] Put default counters in init_cfg.json

stepanblyschak:
#### Why I did it

Simplify approach to delaying counters on warm boot and fast boot. Removed FLEX_COUNTER_DELAY_STATUS_FIELD and instead postpone all FC processing to happen after apply view to not delay data plane configuration.

The CONFIG_DB should not be updated in runtime anymore for counters to be delayed.

To address https://github.com/sonic-net/sonic-buildimage/issues/20302.

##### Work item tracking

#### How I did it

Put FC into init_cfg.json

#### How to verify it

Boot the system, verify counters are enabled.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/files/build_templates/init_cfg.json.j2
+++ b/files/build_templates/init_cfg.json.j2
@@ -26,6 +26,33 @@
             "FLEX_COUNTER_STATUS": "enable",
             "FLEX_COUNTER_DELAY_STATUS": "true",
             "POLL_INTERVAL": "10000"
+        },
+        "PORT": {
+            "FLEX_COUNTER_STATUS": "enable"
+        },
+        "RIF": {
+            "FLEX_COUNTER_STATUS": "enable"
+        },
+        "QUEUE": {
+            "FLEX_COUNTER_STATUS": "enable"
+        },
+        "PFCWD": {
+            "FLEX_COUNTER_STATUS": "enable"
+        },
+        "PG_WATERMARK": {
+            "FLEX_COUNTER_STATUS": "enable"
+        },
+        "PG_DROP": {
+            "FLEX_COUNTER_STATUS": "enable"
+        },
+        "QUEUE_WATERMARK": {
+            "FLEX_COUNTER_STATUS": "enable"
+        },
+        "BUFFER_POOL_WATERMARK": {
+            "FLEX_COUNTER_STATUS": "enable"
+        },
+        "PORT_BUFFER_DROP": {
+            "FLEX_COUNTER_STATUS": "enable"
         }
     },
     "BGP_DEVICE_GLOBAL": {
--- a/dockers/docker-orchagent/enable_counters.py
+++ b/dockers/docker-orchagent/enable_counters.py
@@ -39,24 +39,13 @@ def enable_rates():
 def enable_counters():
     db = swsscommon.ConfigDBConnector()
     db.connect()
-    default_enabled_counters = ['PORT', 'RIF', 'QUEUE', 'PFCWD', 'PG_WATERMARK', 'PG_DROP', 
-                                'QUEUE_WATERMARK', 'BUFFER_POOL_WATERMARK', 'PORT_BUFFER_DROP', 'ACL']
     dpu_counters = ["ENI"]
-
-    # Enable those default counters
-    for key in default_enabled_counters:
-        enable_counter_group(db, key)
 
     platform_info = device_info.get_platform_info(db)
     if platform_info.get('switch_type') == 'dpu':
         for key in dpu_counters:
             enable_counter_group(db, key)
 
-    # Set FLEX_COUNTER_DELAY_STATUS to false for those non-default counters
-    keys = db.get_keys('FLEX_COUNTER_TABLE')
-    for key in keys:
-        if key not in default_enabled_counters:
-            enable_counter_group(db, key)
     enable_rates()
 
 


wen587 (approved):
lgtm

files/build_templates/init_cfg.json.j2
```diff
@@ -26,6 +26,33 @@
...
+            "FLEX_COUNTER_STATUS": "enable"
+        },
+        "QUEUE_WATERMARK": {
+            "FLEX_COUNTER_STATUS": "enable"
+        },
+        "BUFFER_POOL_WATERMARK": {
+            "FLEX_COUNTER_STATUS": "enable"
+        },
+        "PORT_BUFFER_DROP": {
+            "FLEX_COUNTER_STATUS": "enable"
```
qiluo-msft, line 55: Let's treat this as a new feature, feature enabled by changining config_db.json. Please keep all the exsiting init_cfg.json / minigraph.py behavior unchanged, we are moving away from minigraph based config flow.

It is acceptable to define new field in config_db.json schema, indicating static/dynamic flex_counter mode, if config_db.json missing the new field, using default value (dynamic), and behavior is current behavior. So we also maintain backward compatible.

qiluo-msft (dismissed):
As comment

files/build_templates/init_cfg.json.j2
stepanblyschak: @qiluo-msft Why this change is not backward compatible? Not having these entries in init_cfg and not adding them in runtime on the other hand is backward incompatible (no counters enabled after upgrade). Why defining a new field? How moving away from minigraph based flow is related?

stepanblyschak: @qiluo-msft This is the flow of `config reload`:

```
admin@sonic:~$ sudo config reload -y /home/admin/config_db.json
Stopping SONiC target ...
Running command: /usr/local/bin/sonic-cfggen -j /etc/sonic/init_cfg.json -j /home/admin/config_db.json --write-to-db
Restarting SONiC target ...
Reinitializing monit daemon
```

The config load 1) merges config with init_cfg.json 2) runs db_migrator to handle breaking changes.

1. If `/home/admin/config_db.json` does not have any FLEX_COUNTER configuration it will be taken from `init_cfg.json` so that counters are going to be enabled. No breaking changes, same config_db.json would produce exactly the same FC configuration via `enable_counters.py` script in the old image.
2. If `/home/admin/config_db.json` has FLEX_COUNTER configuration (e.g. some counters were disabled before) it won't be overwritten by `init_cfg.json`. No breaking change.

qiluo-msft: @stepanblyschak As we discussed, share the command we are using to config device `sudo config load_minigraph --override_config -y`

stepanblyschak: @qiluo-msft I reviewed the command `config load_minigraph --override_config -y`. It calls `config override-config-table /etc/sonic/golden_config_db.json` command which for every table in `/etc/sonic/golden_config_db.json` removes table from CONFIG_DB and replaces with the content from the file.

I find that the behavior of `config override-config-table` is inconsistent with `config reload`. In order to handle breaking changes correctly it should do the same as `config reload`:
1.  Produce the config `/usr/local/bin/sonic-cfggen -j /etc/sonic/init_cfg.json -j /etc/sonic/golden_config_db.json`
2. Run the config replacement logic
3. Run `/usr/local/bin/db_migrator.py -o migrate`

So I believe this is a bug in `config override-config-table` command.

wen587: config override-config-table will not run db_migrator since the config source is gonna provide bt other source of truth.

stepanblyschak: @wen587 I don't follow, could you please clarify? Why is `override-config-table` different from `reload`?

wen587:
Hi @stepanblyschak , 
I saw not all device are with enable field on devices. 
Can we have some unit test for below scenarios? Here is what we expect in production: 

If mgmt device, which is `If DEVICE_METADATA['localhost']['type'] is in ['MgmtTsToR', 'MgmtToRRouter', 'BmcMgmtToRRouter'])`
Then:
```
    },
    "FLEX_COUNTER_TABLE": {
        "ACL": {
            "FLEX_COUNTER_DELAY_STATUS": "false",
            "FLEX_COUNTER_STATUS": "enable",
            "POLL_INTERVAL": "10000"
        },
        "BUFFER_POOL_WATERMARK": {
            "FLEX_COUNTER_DELAY_STATUS": "false",
            "FLEX_COUNTER_STATUS": "disable"
        },
        "PFCWD": {
            "FLEX_COUNTER_DELAY_STATUS": "false",
            "FLEX_COUNTER_STATUS": "disable"
        },
        "PG_DROP": {
            "FLEX_COUNTER_DELAY_STATUS": "false",
            "FLEX_COUNTER_STATUS": "disable"
        },
        "PG_WATERMARK": {
            "FLEX_COUNTER_DELAY_STATUS": "false",
            "FLEX_COUNTER_STATUS": "disable"
        },
        "PORT": {
            "FLEX_COUNTER_STATUS": "enable"
        },
        "PORT_BUFFER_DROP": {
            "FLEX_COUNTER_DELAY_STATUS": "false",
            "FLEX_COUNTER_STATUS": "disable"
        },
        "QUEUE": {
            "FLEX_COUNTER_DELAY_STATUS": "false",
            "FLEX_COUNTER_STATUS": "disable"
        },
        "QUEUE_WATERMARK": {
            "FLEX_COUNTER_DELAY_STATUS": "false",
            "FLEX_COUNTER_STATUS": "disable"
        },
        "RIF": {
            "FLEX_COUNTER_STATUS": "enable"
        }
    },
```

If not mgmt device: 
```
    "FLEX_COUNTER_TABLE": {
        "ACL": {
            "FLEX_COUNTER_DELAY_STATUS": "false",
            "FLEX_COUNTER_STATUS": "enable",
            "POLL_INTERVAL": "10000"
        },      
        "BUFFER_POOL_WATERMARK": {
            "FLEX_COUNTER_STATUS": "enable"
        },      
        "PFCWD": {
            "FLEX_COUNTER_STATUS": "enable"
        },      
        "PG_DROP": {
            "FLEX_COUNTER_STATUS": "enable"
        },      
        "PG_WATERMARK": {
            "FLEX_COUNTER_STATUS": "enable"
        },      
        "PORT": {
            "FLEX_COUNTER_STATUS": "enable"
        },      
        "PORT_BUFFER_DROP": {
            "FLEX_COUNTER_STATUS": "enable"
        },      
        "QUEUE": {
            "FLEX_COUNTER_STATUS": "enable"
        },      
        "QUEUE_WATERMARK": {
            "FLEX_COUNTER_STATUS": "enable"
        },  
        "RIF": {
            "FLEX_COUNTER_STATUS": "enable"
        }
    },
```

Can we also add some unit test to guarantee the output?

wen587:
BTW, image PR for this logic: https://github.com/sonic-net/sonic-buildimage/pull/16937

files/build_templates/init_cfg.json.j2
wen587: You can see the db_migrator is executed earlier than override: https://github.com/sonic-net/sonic-utilities/blob/master/config/main.py#L2086-L2103
That's because we want the golden config to be the single source of truth for configuration. It will help us to incremental configthe device(such as using GCU:generic config updater)  if there is any issue with the device.

wen587 (approved):

qiluo-msft (approved):

PR #20522, opened 2024-10-16
Adding custom create switch timeout for Cisco platforms

maunnikr-cisco:
##### Work item tracking

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/dockers/docker-orchagent/orchagent.sh
+++ b/dockers/docker-orchagent/orchagent.sh
@@ -76,6 +76,12 @@ elif [ "$platform" == "marvell" ]; then
     if [[ ! -z $CREATE_SWITCH_TIMEOUT ]]; then
         ORCHAGENT_ARGS+=" -t $CREATE_SWITCH_TIMEOUT"
     fi
+elif [ "$platform" == "cisco-8000" ]; then
+    ORCHAGENT_ARGS+="-m $MAC_ADDRESS"
+    CREATE_SWITCH_TIMEOUT=`cat $HWSKU_DIR/sai.profile | grep "createSwitchTimeout" | cut -d'=' -f 2`
+    if [[ ! -z $CREATE_SWITCH_TIMEOUT ]]; then
+        ORCHAGENT_ARGS+=" -t $CREATE_SWITCH_TIMEOUT"
+    fi
 else
     # Should we use the fallback MAC in case it is not found in Device.Metadata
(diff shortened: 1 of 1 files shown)


dockers/docker-orchagent/orchagent.sh
```diff
@@ -76,6 +76,12 @@ elif [ "$platform" == "marvell" ]; then
     if [[ ! -z $CREATE_SWITCH_TIMEOUT ]]; then
         ORCHAGENT_ARGS+=" -t $CREATE_SWITCH_TIMEOUT"
     fi
+elif [ "$platform" == "cisco-8000" ]; then
+    ORCHAGENT_ARGS+="-m $MAC_ADDRESS"
+    CREATE_SWITCH_TIMEOUT=`cat $HWSKU_DIR/sai.profile | grep "createSwitchTimeout" | cut -d'=' -f 2`
```
kevinskwang, line 81: could you add a comment to limit the hwsku to 8111?

maunnikr-cisco: Hi Kevin, we will specify the new timeout only in the sai.profile of the 8111 hwsku, so there's no need to specify that here. If ever there is a need for a higher timeout for any other hwsku, we can just add the new timeout to the sai.profile specific to that hwsku, and this code will work for that too. Specifying 8111 hwsku here will prevent that.

kperumalbfn: @maunnikr-cisco What is the range of new timeout value? Could you provide the reason for the delay.

maunnikr-cisco: The required timeout value seems to be around 75 seconds. With a buffer, 90s seems to be the appropriate time. We will be specifying this timeout in the sai.profile specific to the hwsku that needs this new timeout (8111). We don't think it's a good use of time to investigate this delay for the short term, and prefer to increase the delay for the hwsku instead, since there is no other functional impact.

kevinskwang:
Error:
tests/acl/test_acl.py:811:21: E129 visually indented line with same indent as next logical line

flake8...............................................(no files to check)Skipped
check conditional mark sort..........................(no files to check)Skipped

PR #20524, opened 2024-10-16
dhcp_v4_relay_issue

ypcisco:
#### Why I did it

If DHCP Servers are reachable only over RFC 5549 enabled links (that is V4 destinations over V6 neighbors) BOOTP responses are dropped by the relay switch and thus client DHCP requests are never complete.

#### How I did it

- Added Flag "-ih" for interfaces configured with only ipv6 address.
- Added check to allow packets received (with no ipv4 address) on above interfaces.

#### How to verify it

- Have a DHCP Server reachable over RFC 5549 enabled links.
- Capture BOOTP resposes via tcpdump.
- Check the DHCP relay address learnt via "show dhcp_relay ipv4 helper".

- Adding the "-ih" flag to distinguish UPSTREAM interface with ipv6 address only.
- For above case added a check to process the DHCP replies.


Blueve:
@ypcisco do we have sonic-mgmt test case to cover this issue?

@yaqiangz for vis

PR #20585, opened 2024-10-23
Fixing SRv6 SID uninstall

cscarpitta:
This PR fixes the SRv6 SID uninstall introduced in this PR: https://github.com/sonic-net/sonic-buildimage/pull/18715
This fix has been already merged in the FRR mainline: https://github.com/FRRouting/frr/pull/16835


ahsalam (approved):
This is needed correctly uninstall SRv6 SID when the SRv6 locator is deleted in FRR.

PR #20619, opened 2024-10-25
sonic-buildimage: Re-enable checking the pcie firmware is loaded in syncd for 720DT

aaronpayment:
#### Why I did it
Remove the check as the platform is strapped to load the pcie firmware and we should be able to keep the SDK check to ack the firmware was loaded instead of skipping the verification.

This is similar to https://github.com/sonic-net/sonic-buildimage/pull/13080 where we do not want to keep this check disabled if we do not need to.

#### How I did it
Removed the bcm config that enabled the skip behavior.

#### How to verify it
Ensure that the device is able to boot successfully and perform the check and not crash syncd.

- [x] 202405

- [x] 202405
- [x] master


aaronpayment:
@StormLiangMS for review

Blueve (approved):

Blueve:
@aaronpayment do you have platform test result uploaded for above change? 
Will any platform tests impact by this change?

aaronpayment:
> @aaronpayment do you have platform test result uploaded for above change? Will any platform tests impact by this change?

Hi Jing,

We have removed this internally and all uploads since this PR has been opened would have run with the change.   The reason to remove this is to catch if there is any future change that would cause issue and the result would be a SAI crash. The current testing we do on 202405 and master with the sonic mgmt tests do not have any have any crash so enabling this is a nop right now as there are no incompatible changes.

aaronpayment:
> hi @aaronpayment could you help to clarify why this is a must for 202405?

Hi Strom,
This is not a mustfix for 202405.

bingwang-ms:
Removing cherry-pick request since this is not a must-have for 202405

PR #20706, opened 2024-11-05
[supervisord.conf] enable neighsyncd to autorestart when exit code is 0

Ndancejic:
handle exception where neighsyncd exited normally but did not autorestart. enables autorestart for neighsyncd if exit code is 0.

#### Why I did it
auto-restart was not enabled for syncd

##### Work item tracking
- Microsoft ADO **(number only)**: 29962787

#### How I did it
edited supervisord.conf to set neighsyncd autorestart=unexpected and exitcodes=0. this will auto-restart neighsyncd if the exit code is 0 but not otherwise.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [x] 20231110.11

enable neighsyncd to autorestart when exit code is 0

### DIFF
--- a/dockers/docker-orchagent/supervisord.conf.j2
+++ b/dockers/docker-orchagent/supervisord.conf.j2
@@ -135,7 +135,8 @@ environment=ASAN_OPTIONS="log_path=/var/log/asan/coppmgrd-asan.log{{ asan_extra_
 command=/usr/bin/neighsyncd
 priority=7
 autostart=false
-autorestart=false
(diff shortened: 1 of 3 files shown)


prsunny (changes requested):
@Ndancejic , this must be fixed in neighsyncd. The change here is not generic since all other processes are not autorestarted with exit code 0.

Issue #20579, opened 2024-10-22
/etc/hosts regulary updated with same records


johndexter1:
#### Description
`sethostname` hook compares `hostname -s` instead of `hostname` and when dhcp provides FQDN it never matches the IF statement and new records are appended to `/etc/hosts` on every dhcp renew

#### Steps to reproduce the issue:
1. Set up dhcp client on mgmt port
2. Provide DHCP with FQDN as hostname option
3. On each renew new entry will be appended to `/etc/hosts`
4. Over some time `/etc/hosts` becomes huge. Depends on lease times, of course.

#### Describe the results you received:
A lot of identical lines in `/etc/hosts`

#### Describe the results you expected:
Only new (only if hostname has really changed) are added to `/etc/hosts` from dhcp

#### Output of `show version`:
202405

#### Additional information you deem important (e.g. issue happens only occasionally):
I have tested this PR (https://github.com/sonic-net/sonic-buildimage/pull/20404) and it solved the problem.

wdoekes:
The cause:
`/etc/dhcp/dhclient-exit-hooks.d/sethostname` (`./files/dhcp/sethostname`):
```sh
case $reason in
    BOUND|RENEW|REBIND|REBOOT)
        current_host_name=`hostname -s`
        if [ "$current_host_name" != "$new_host_name" ] && [ -n "$new_host_name" ]
        then
            echo $new_host_name > /etc/hostname
            hostname -F /etc/hostname
            sed -i "/\s$current_host_name$/d" /etc/hosts
            echo "127.0.0.1 $new_host_name" >> /etc/hosts
        fi
        ;;
esac
```
We have fqdn's, so we get this:
```console
# reason=RENEW new_host_name=leaf1.dostno.systems sh -x /etc/dhcp/dhclient-exit-hooks.d/sethostname
+ hostname -s
+ current_host_name=leaf1
+ [ leaf1 != leaf1.dostno.systems ]
+ [ -n leaf1.dostno.systems ]
+ echo leaf1.dostno.systems
+ hostname -F /etc/hostname
+ sed -i /\sleaf1$/d /etc/hosts
+ echo 127.0.0.1 leaf1.dostno.systems
```
It would stay fixed for short hostnames:
```console
# reason=RENEW new_host_name=leaf1 sh -x /etc/dhcp/dhclient-exit-hooks.d/sethostname
+ hostname -s
+ current_host_name=leaf1
+ [ leaf1 != leaf1 ]
```
[...]

PR #20740, opened 2024-11-08
[dhcp] Fix that /etc/hosts is not appended with the same record

wdoekes:
#### Work item tracking

Ticket: #20579
Alternative-PR: #20404

#### Why I did it

When a dhcp server hands out a fully qualified hostname in option 12, the previous sethostname script would keep appending to /etc/hosts.

#### How I did it

Replace DHCP sethostname and sethostname6 so that:

- hostname takes precedence if it is a FQDN
- if domain name is set and hostname is not FQDN, combine them
- if only hostname is set, merge it with a previously set domain name

Then set both the FQDN and the short hostname in /etc/hosts:

```
127.0.0.1 shortname.domain.com shortname
```

And make sure that the old one is removed when the hostname is updated through DHCP.

#### How to verify it

```
$ reason=BOUND new_host_name=leaf1 sh -x sethostname
$ reason=BOUND new_host_name=leaf2 new_domain_name=domain.com sh -x sethostname
$ reason=BOUND new_host_name=leaf3.domain.com sh -x sethostname
```
Check that `/etc/hosts` and `/etc/hostname` are as expected.

Rework how dhcp client sets hostname from dhcp host and domain options and keep it from adding duplicates to /etc/hosts.

### DIFF
--- a/files/dhcp/sethostname
+++ b/files/dhcp/sethostname
@@ -1,12 +1,40 @@
 case $reason in
     BOUND|RENEW|REBIND|REBOOT)
-        current_host_name=`hostname -s`
-        if [ "$current_host_name" != "$new_host_name" ] && [ -n "$new_host_name" ]
-        then
-            echo $new_host_name > /etc/hostname
-            hostname -F /etc/hostname
-            sed -i "/\s$current_host_name$/d" /etc/hosts
-            echo "127.0.0.1 $new_host_name" >> /etc/hosts
+        if [ -n "${new_host_name:-}" ]; then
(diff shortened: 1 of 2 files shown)


johndexter1:
I tested this, and when a non-FQDN hostname is provided, it results in the creation of a duplicate entry. For example: `127.0.0.1 leaf1 leaf1`. Is this behavior expected?

wdoekes:
Yes, that's expected and benign.

wdoekes:
Converted back to draft. See if #20404 gets merged first.

PR #20743, opened 2024-11-08
[Mellanox] Enable get_rx_los API support in CMIS cable host mgmt mode

mihirpat1:
#### Why I did it
The `show int transceiver status PORT` CLI always displays Rx LOS as False for all the lanes.

```
root@sonic:/home/admin# show int transceiver status Ethernet0 | grep "Rx loss"
        Rx loss of signal flag on media lane 1: False
        Rx loss of signal flag on media lane 2: False
        Rx loss of signal flag on media lane 3: False
        Rx loss of signal flag on media lane 4: False
        Rx loss of signal flag on media lane 5: False
        Rx loss of signal flag on media lane 6: False
        Rx loss of signal flag on media lane 7: False
        Rx loss of signal flag on media lane 8: False
root@sonic:/home/admin# 
```
##### Work item tracking
- Microsoft ADO **(number only)**: 30180403

#### How I did it
On Mellanox platforms, the `get_rx_los` API is overridden through the platform implementation (https://github.com/sonic-net/sonic-buildimage/blob/679997639ede79a7beef53654be101b14d04b006/platform/mellanox/mlnx-platform-api/sonic_platform/sfp.py#L1069) rather than using the implementation defined through cmis.py (https://github.com/sonic-net/sonic-platform-common/blob/59babf59f9aa611b84082896391b01a39ffcb866/sonic_platform_base/sonic_xcvr/api/public/cmis.py#L404).

The Mellanox platform is overriding the `get_rx_los` due to the below code
https://github.com/sonic-net/sonic-buildimage/blob/679997639ede79a7beef53654be101b14d04b006/platform/mellanox/mlnx-platform-api/sonic_platform/sfp.py#L1188-L1189

The current platform implementation assigns Rx LOS as False for all the lanes (irrespective of the actual status on the module) which in turn causes the `show int transceiver status PORT` CLI to display the Rx LOS as False for all the lanes.

Hence, removing the overriding part to ensure that the API is not overridden by the platform.

#### How to verify it
Verified the CLI on first 2 lanes which are in admin disabled state from the peer side.

```
root@sonic:/home/admin# show int transceiver status Ethernet0 | grep "Rx loss"
        Rx loss of signal flag on media lane 1: False
        Rx loss of signal flag on media lane 2: False
        Rx loss of signal flag on media lane 3: True
        Rx loss of signal flag on media lane 4: True
        Rx loss of signal flag on media lane 5: True
        Rx loss of signal flag on media lane 6: True
        Rx loss of signal flag on media lane 7: True
        Rx loss of signal flag on media lane 8: True
root@sonic:/home/admin# 
```

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/platform/mellanox/mlnx-platform-api/sonic_platform/sfp.py
+++ b/platform/mellanox/mlnx-platform-api/sonic_platform/sfp.py
@@ -927,8 +927,6 @@ class SFP(NvidiaSFPCommon):
         """
         if self._xcvr_api is None:
             self.refresh_xcvr_api()
-            if self._xcvr_api is not None:
-                self._xcvr_api.get_rx_los = self.get_rx_los
         return self._xcvr_api
 
     def is_sw_control(self):
--- a/platform/mellanox/mlnx-platform-api/sonic_platform/sfp.py
+++ b/platform/mellanox/mlnx-platform-api/sonic_platform/sfp.py
@@ -815,6 +815,11 @@ class SFP(NvidiaSFPCommon):
             list: [False] * channels
         """
         api = self.get_xcvr_api()
+        try:
+            if self.is_sw_control():
+                return api.get_rx_los() if api else None
+        except Exception as e:
+            print(e)
         return [False] * api.NUM_CHANNELS if api else None
 
     def get_tx_fault(self):


prgeor (dismissed):

prgeor:
@liat-grozovik @Junchao-Mellanox can you also review

Junchao-Mellanox (approved):

keboliu (changes requested):
There is a limitation on the Mellanox platform to support RX LOS reading in FW control mode, this change will also enable the RX LOS reading in FW control mode, we should have a logic to differentiate between FW control mode and host mgmt mode. @Junchao-Mellanox

Junchao-Mellanox:
> There is a limitation on the Mellanox platform to support RX LOS reading in FW control mode, this change will also enable the RX LOS reading in FW control mode, we should have a logic to differentiate between FW control mode and host mgmt mode. @Junchao-Mellanox

Thanks for catching this. Indeed, we need do something like following:
```
def get_rx_los(self):
        """Accessing rx los is not supproted, return all False

        Returns:
            list: [False] * channels
        """
        api = self.get_xcvr_api()
        if self.is_sw_control():
                return api.get_rx_los()
        return [False] * api.NUM_CHANNELS if api else None
```

Junchao-Mellanox (approved):

mihirpat1:
> we should have a logic to differentiate between FW control mode and host mgmt mode. @Junchao-Mellanox

@Junchao-Mellanox @keboliu I have now addressed this. Can you please help in reviewing it?

keboliu (approved):

prgeor (approved):

mihirpat1:
@bingwang-ms @yxieca Can you please help in merging this to 202405 and 202311?
MSFT ADO - 30180403

PR #20744, opened 2024-11-09
[FRR]Setting multipath size to 514 and disabling bgp-vnc for optimization

dgsudharsan:
#### Why I did it
Increased the max multipath to 512. In addition removed bgp-vnc as this feature is not used by SONiC .

https://docs.frrouting.org/en/latest/vnc.html#vnc-and-vnc-gw

This feature adds overhead in general and removing this improves bgp convergence time for scale.

##### Work item tracking

#### How I did it
Added a patch modify debian rules.

#### How to verify it
Run BGP tests in scaled topology.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]


dgsudharsan:
@eddieruan-alibaba @hasan-brcm  FYI. Please review the changes.

hasan-brcm:
@dgsudharsan , frr 10.0.1 upgrade ([PR#20269](https://github.com/sonic-net/sonic-buildimage/pull/20269)) is already in progress. Please wait for that to get merged.

src/sonic-frr/patch/0054-Updating-rules-to-set-multipath-to-512-and-removing-.patch
```diff
@@ -0,0 +1,23 @@
...
+diff --git a/debian/rules b/debian/rules
+index aa13106fec..e280b6c45b 100755
+--- a/debian/rules
++++ b/debian/rules
+@@ -68,8 +68,7 @@ override_dh_auto_configure:
+ 		--disable-protobuf \
+ 		--disable-zeromq \
+ 		--enable-ospfapi \
+-		--enable-bgp-vnc \
+-		--enable-multipath=256 \
```
eddieruan-alibaba, line 16: what is the use case to have 512 ECMP paths from FRR. 

Currently FRR uses uint8 for num paths.

dgsudharsan: This is fixed by FRR https://github.com/FRRouting/frr/pull/16967. We will be taking this fix once FRR upgrade is done. We do have some specific use case scenarios where the number of nexthops might scale more than 256. So we are setting to 512 here.

dgsudharsan: You can refer to the topology here https://github.com/sonic-net/sonic-mgmt/pull/15355

eddieruan-alibaba (approved):

liat-grozovik (approved):

PR #20895, opened 2024-11-22
Add bmp dynamic feature switch in frr side config.

FengPan-Frank:
#### Why I did it
Add bmp dynamic feature switch in frr side config, later will update liquid template to control bmp feature based on version.

##### Work item tracking
- Microsoft ADO **(number only)**:27588893

#### How I did it
Update bgpd.conf and supervisord.conf template to monitor the feature switch, and turn on corresponding section. Later once we want to turn on bmp feature, we will assign specific version on NDM liquid template, and on the rollout-ed host it will require a new config reload to take effect.

#### How to verify it
Added unit test cases.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]


FengPan-Frank:
Hi @StormLiangMS , we have new feature "bmp" which would like to be enabled in runtime later per NDM liquid config. The feature requires frr side config + cmd change, could you help on review if there's any concern? Thanks

dockers/docker-fpm-frr/frr/supervisord/supervisord.conf.j2
```diff
@@ -77,7 +77,11 @@ dependent_startup_wait_for=zebra:running
 {% endif %}
 
 [program:bgpd]
+{% if FEATURE.bmp is defined and FEATURE.bmp.state is defined and FEATURE.bmp.state == "enabled" %}
```
qiluo-msft, line 80: >FEATURE [](http://example.com/codeflow?start=6&length=7)

Could you make your 2 if statement consistent? Why did not you check "FEATURE is defined" here?

FengPan-Frank: done, previously just follow other below section style.

qiluo-msft (approved):

FengPan-Frank:
> hi @FengPan-Frank do we have test case to cover this in nightly? I would be curious about the memory and cpu utilization of this feature.

@StormLiangMS I'm working on nightly test cases for this feature continually, for this flag we want to create placeholder on image first which will be controlled by Sonic_FEATURE_Template.liquid dynamically, it will not be enabled unless nightly is in good shape later.

PR #20897, opened 2024-11-22
[database]: Fix "program:redisprogram:redis_bmp" ERR from supervisor …

wdoekes:
## Why I did it

Backport of 4807ac5109c2f5f20d92b33041ac45b9b9b11c89 for 202405.

The problem isn't there yet, but this fix is correct. And the problem might crop up once someone starts using two DBs.

## Original issue

Issue: https://github.com/sonic-net/sonic-buildimage/issues/20636
PR-for-master: https://github.com/sonic-net/sonic-buildimage/pull/20726

@FengPan-Frank can you merge this one as well?

### DIFF
--- a/dockers/docker-database/critical_processes.j2
+++ b/dockers/docker-database/critical_processes.j2
@@ -1,5 +1,5 @@
-{% if INSTANCES %}
-{%     for redis_inst, redis_items in INSTANCES.items() %}
+{% if INSTANCES -%}
+{%     for redis_inst, redis_items in INSTANCES.items() -%}
 program:{{ redis_inst }}
-{%-     endfor %}
-{%- endif %}
+{%     endfor -%}
+{% endif -%}


FengPan-Frank (approved):
I remember we only need to add a label to backport.
btw, I'm not sure if we have real multiple redis instance use case for 202405, since redis_bmp was not in 202405.

FengPan-Frank:
> > I remember we only need to add a label to backport.
> 
> Not sure what you mean. Do labels fix backports automatically? I have no permissions to add them either way.
> 
> > btw, I'm not sure if we have real multiple redis instance use case for 202405, since redis_bmp was not in 202405
> 
> Indeed we don't. But someone might a second instance in their stable branch. And by merging this they won't run into this problem.

OK, I've added label

bingwang-ms:
PR https://github.com/sonic-net/sonic-buildimage/pull/20898 is already merged. Do we still need this PR?

wdoekes:
https://github.com/sonic-net/sonic-buildimage/commit/2544e485bc9f6e5ae8bdfc78cabed73773680fc9 (202405) and https://github.com/sonic-net/sonic-buildimage/commit/4807ac5109c2f5f20d92b33041ac45b9b9b11c89 (202411 and master) are both included. So yes, this can be closed.

PR #20916, opened 2024-11-25
Mounting /host/reboot-cause on pmon so that the modules can persist their reboot cause history

rameshraghupathy:
Mounting /host/reboot-cause on pmon so that the modules can persist their reboot cause history

#### Why I did it
In SmartSwitch each DPU module's reboot-cause history needs to be persisted on the host

#### How I did it
By mounting  "/host/reboot-cause" volume  on pmon

#### How to verify it
Once the host boot up check the path on the pmon container


prgeor (approved):

prgeor:
@yxieca can you merge

PR #20603, opened 2024-10-24
[Arista] Update platform library submodules

Staphylo:
#### Why I did it

 - Add support for new psus
 - Add support for new fan modules
 - Report fan information from all fabric cards
 - Fix psud warning due to invalid pmbus threshold reported
 - Disable lpmode for chassis

##### Work item tracking
- Microsoft ADO **(number only)**:

#### How I did it

Updated our platform driver submodules

#### How to verify it

Changes validated against sonic-mgmt 202405 and master

#### Description for the changelog

Update Arista platform library submodules

PR #20929, opened 2024-11-26
[Arista] Update platform library submodules

patrickmacarthur:
#### Why I did it

(this is a replacement for PR #20603; this PR includes a revert for the "Disable lpmode for chassis" change)

* Add support for new psus
* Add support for new fan modules
* Report fan information from all fabric cards
* Fix psud warning due to invalid pmbus threshold reported

##### Work item tracking

#### How I did it

Updated our platform driver submodules

#### How to verify it

Changes validated against sonic-mgmt 202405 and master

Update Arista platform library submodules


arlakshm (approved):

gechiang:
@patrickmacarthur should we also port this change to 202405?  especially if we are to back port this PR to 202405:
https://github.com/sonic-net/sonic-platform-daemons/pull/555
If so, can you prepare the same for 202405 when the above thermalctld PR under platform-daemons merges in master?

patrickmacarthur:
> @patrickmacarthur should we also port this change to 202405? especially if we are to back port this PR to 202405: [sonic-net/sonic-platform-daemons#555](https://github.com/sonic-net/sonic-platform-daemons/pull/555) If so, can you prepare the same for 202405 when the above thermalctld PR under platform-daemons merges in master?

Yes we would like to port this to 202405, we can go ahead and open a PR against 202405.

PR #20938, opened 2024-11-27
Sonic Profiling: Docker Options in rules/config

amazor:
Update rules/config file to include default options for enabling, delay, and autorestart.
Standardize what "including" means in the make options.

Link to HLD: https://github.com/sonic-net/SONiC/pull/1796


rules/config
matiAlfaro, line 1: Consider using a macro

``` makefile
define DECLARE_FEATURE
INCLUDE_$(1) = $(2)
ENABLE_$(1) = $(3)
DELAY_$(1) = $(4)
AUTORESTART_$(1) = $(5)
endef
```
then for for example NAT we'll have

``` makefile
$(eval $(call DECLARE_FEATURE,NAT,y,n,n,y))
``````

This will also make it easier to maintain this form for future additions

slave.mk
matiAlfaro, line 1: suggest using a macro also for this

PR #21054, opened 2024-12-05
[yang] bgp address family l2vpn advertise-all-vni

bradh352:
#### Why I did it

Cannot configure unified bgp for vxlan evpn without specifying advertise-all-vpn.  The setting appears to have been introduced as part of PR #5142, can be seen it is already honored as an option here: https://github.com/sonic-net/sonic-buildimage/blob/8e0f1c66b1d49463b984b18dd417ec37fb5af24a/src/sonic-frr-mgmt-framework/templates/bgpd/bgpd.conf.db.addr_family.evpn.j2#L1-L3

#### How I did it

Added basic yang rule

#### How to verify it

Configure
```
"BGP_GLOBALS_AF": {
        "default|l2vpn_evpn": {
            "advertise-all-vni": "true"
        }
    }
```
and run `config replace`.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [X] 202106
- [X] 202111
- [X] 202205
- [X] 202211
- [X] 202305
- [X] 202311
- [X] 202405
- [X] 202411

master as of 20241205

[yang] bgp address family l2vpn advertise-all-vni

N/A

### DIFF
--- a/src/sonic-yang-models/yang-models/sonic-bgp-global.yang
+++ b/src/sonic-yang-models/yang-models/sonic-bgp-global.yang
@@ -445,6 +445,11 @@ module sonic-bgp-global {
                     }
                     description "Maximum duration to suppress a stable route";
                 }
+
+                leaf advertise-all-vni {
+                    type boolean;
+                    description "L2VPN advertise all VNIs";
+                }
             }
         }
 
--- a/src/sonic-yang-models/tests/files/sample_config_db.json
+++ b/src/sonic-yang-models/tests/files/sample_config_db.json
@@ -1771,6 +1771,9 @@
             "default|ipv6_unicast": {
                 "max_ebgp_paths": "2",
                 "max_ibgp_paths": "2"
+            },
+            "default|l2vpn_evpn": {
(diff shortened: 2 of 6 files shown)


bradh352:
@qiluo-msft please review, this should be an easy one

src/sonic-yang-models/yang-models/sonic-bgp-global.yang
```diff
@@ -445,6 +445,11 @@ module sonic-bgp-global {
                     }
                     description "Maximum duration to suppress a stable route";
                 }
+
+                leaf advertise-all-vni {
```
qiluo-msft, line 449: Thanks for the bugfix! Could you also add a unit test, in order to prevent future regression?

bradh352: I guess, but there are currently ZERO tests for BGP_DEVICE_GLOBAL_AF.  Seems odd for this to be the only test.

ganglyu (approved):

qiluo-msft (approved):

PR #21055, opened 2024-12-05
[yang] frr-mgmt-framework: admin_status should be up/down but wants true/false

bradh352:
#### Why I did it

YANG uses `stypes:admin_status;` instead of `boolean` here:
https://github.com/sonic-net/sonic-buildimage/blob/f7014a3cbab56d1c7e1d0f078565378b2684085a/src/sonic-yang-models/yang-models/sonic-bgp-common.yang#L320-L323
https://github.com/sonic-net/sonic-buildimage/blob/f7014a3cbab56d1c7e1d0f078565378b2684085a/src/sonic-yang-models/yang-models/sonic-bgp-common.yang#L344-L347
https://github.com/sonic-net/sonic-buildimage/blob/f7014a3cbab56d1c7e1d0f078565378b2684085a/src/sonic-yang-models/yang-models/sonic-bgp-common.yang#L576-L579

But the code expects `true/false` as seen in:
https://github.com/sonic-net/sonic-buildimage/blob/8e0f1c66b1d49463b984b18dd417ec37fb5af24a/src/sonic-frr-mgmt-framework/templates/bgpd/bgpd.conf.db.nbr_af.j2#L5-L9
and
https://github.com/sonic-net/sonic-buildimage/blob/8e0f1c66b1d49463b984b18dd417ec37fb5af24a/src/sonic-frr-mgmt-framework/templates/bgpd/bgpd.conf.db.nbr_or_peer.j2#L33-L38
and also relied upon in
https://github.com/sonic-net/sonic-buildimage/blob/b25a06fbb5acfc2a6b8c27d10d24d2aba6745495/src/sonic-frr-mgmt-framework/frrcfgd/frrcfgd.py#L1873-L1875

Test cases not only in sonic-frr-mgmt-framework, but also externally such as minigraph conversions all expect up/down.  Also for consistency admin_state in other areas of sonic  such as INTERFACES all use up/down for admin_status.

true/false compatibility was left in jinja2 in case someone is using that functionality without yang validation.

#### How I did it

Modified the jinja2 templates and python code.

- [x] 202411

master as of 20241205

[yang] frr-mgmt-framework: admin_status should be up/down but wants true/false

### DIFF
--- a/src/sonic-frr-mgmt-framework/frrcfgd/frrcfgd.py
+++ b/src/sonic-frr-mgmt-framework/frrcfgd/frrcfgd.py
@@ -1843,7 +1843,7 @@ class BGPConfigDaemon:
     cmn_key_map = [('asn&peer_type',                        '{no:no-prefix}neighbor {} remote-as {}'),
                    (['local_asn', '+local_as_no_prepend',
                      '+local_as_replace_as'],               '{no:no-prefix}neighbor {} local-as {} {:no-prepend} {:replace-as}'),
-                   (['admin_status', '+shutdown_message'],  '{no:no-prefix}neighbor {} shutdown {:shutdown-msg}', ['false', 'true']),
+                   (['admin_status', '+shutdown_message'],  '{no:no-prefix}neighbor {} shutdown {:shutdown-msg}', ['down', 'up']),
                    ('local_addr',                           '{no:no-prefix}neighbor {} update-source {}'),
                    ('name',                                 '{no:no-prefix}neighbor {} description {}'),
                    (['ebgp_multihop', '+ebgp_multihop_ttl'],'{no:no-prefix}neighbor {} ebgp-multihop {}', ['true', 'false']),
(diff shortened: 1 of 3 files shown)


bradh352:
rebased to force rebuild to see if general sonic CI tests are working yet

bradh352:
@qiluo-msft @lguohan please review

bradh352:
rebased to try to force rebuild since last build hung

bradh352:
@qiluo-msft @lguohan please review, this is a pretty trivial fix.  BGP with the frr mgmt framework won't work if doing YANG validation without this fix.

bradh352:
@qiluo-msft @lguohan can you merge this fix?  Or have any other feedback?

venkatmahalingam:
Please assign all frr-mgmt-framework changes to @venkatmahalingam for review.

src/sonic-frr-mgmt-framework/frrcfgd/frrcfgd.py
```diff
@@ -1843,7 +1843,7 @@ class BGPConfigDaemon:
     cmn_key_map = [('asn&peer_type',                        '{no:no-prefix}neighbor {} remote-as {}'),
                    (['local_asn', '+local_as_no_prepend',
                      '+local_as_replace_as'],               '{no:no-prefix}neighbor {} local-as {} {:no-prepend} {:replace-as}'),
-                   (['admin_status', '+shutdown_message'],  '{no:no-prefix}neighbor {} shutdown {:shutdown-msg}', ['false', 'true']),
```
venkatmahalingam, line 1846: I understand you are making this change for consistency; did you try loading old configs with this change? see if this breaks backward compatibility?

bradh352: Yes, I did verify that when a config is loaded at boot or via config reload, a value of true/false still works as it previously did.  However, it does not work when the config is updated at runtime, which is to be expected.  Infact when I tested that I got yang validation failures that prevented it due to yang expecting up/down.

venkatmahalingam (approved):

src/sonic-frr-mgmt-framework/frrcfgd/frrcfgd.py
venkatmahalingam: OK, thanks!

bradh352:
closing in favor of #21697 which is more robust

PR #9545, opened 2021-12-15
Fix libyang backlinks issue:

li-pingmao:
Fix libyang backlinks and sonic-yang-mgmt find_data_dependency issues:
    1) changes of find_data_dependency():  
        - remove call to libyang backlinks()
        - find data dependencies/backlinks based on schema and data tree python APIs
    2)  Add test of yang model with union of leafrefs

#### Why I did it
       
    1.  libyang dropped python binding backlinks() API with newer version
    2. old 1.0.173 python backlinks() API does not find the data dependencies/backlinks in case of union of leafrefs
    
#### How I did it
   - remove call to libyang backlinks()
   - changes of find_data_dependency():  
       find data dependencies/backlinks based on schema and data tree python APIs
 
#### How to verify it
    Add test cases of yang model with union of leafrefs
    
#### Which release branch to backport (provide reason below if selected)

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106

PR #21078, opened 2024-12-06
[libyang bug] workaround leaf-list via uses bug in BGP route-map

bradh352:
#### Why I did it

There is a bug in libyang when attempting to validate table `BGP_PEER_GROUP_AF` and likely `BGP_NEIGHBOR_AF` in respect to `route_map_in` and `route_map_out`.

The issue is with the use of `leaf-list` specifically when it is pulled in via a `uses` clause of `bgpcmn:sonic-bgp-cmn-af`.

The error message resembles that of if a child is specified that isn't recognized at all:
```
All Keys are not parsed in BGP_PEER_GROUP_AF
dict_keys(['default|PEERS|ipv4_unicast'])
exceptionList:["'route_map_in'"]
```

#### How I did it

Moving the leaf-list to the parent rather than being imported through the `uses` clause works around this issue.

Again, this is a bug in libyang itself, and the same block in `sonic-bgp-cmn-af` was simply moved to the parents and it magically fixes the issue.

#### How to verify it

Relevant config section to cause the issue (probably included too much here, but none-the-less effective):
```json
{
    "DEVICE_METADATA": {
        "localhost": {
           {# ... #}
            "docker_routing_config_mode": "unified",
            "frr_mgmt_framework_config": "true"
           {# ... #}
        }
    },
    "BGP_GLOBALS": {
        "default": {
            "load_balance_mp_relax": "true",
            "local_asn": "4210000001",
            "log_nbr_state_changes": "true",
            "router_id": "172.16.0.1"
        }
    },
    "BGP_GLOBALS_AF": {
        "default|ipv4_unicast": {
            "max_ebgp_paths": "2"
        },
        "default|ipv6_unicast": {
            "max_ebgp_paths": "2"
        },
        "default|l2vpn_evpn": {
            "advertise-all-vni": "true"
        }
    },
    "BGP_NEIGHBOR": {
        "default|Ethernet72": {
            "peer_group_name": "PEERS"
        }
    },
    "BGP_PEER_GROUP": {
        "default|PEERS": {
            "bfd": "true",
            "capability_ext_nexthop": "true",
            "ebgp_multihop": "true",
            "holdtime": "9",
            "keepalive": "3",
            "min_adv_interval": "5",
            "peer_type": "external"
        }
    },
    "BGP_PEER_GROUP_AF": {
        "default|PEERS|ipv4_unicast": {
            "admin_status": "up",
            "route_map_in": [
                "ALLOW"
            ],
            "route_map_out": [
                "ALLOW"
            ]
        },
        "default|PEERS|ipv6_unicast": {
            "admin_status": "up",
            "route_map_in": [
                "ALLOW"
            ],
            "route_map_out": [
                "ALLOW"
            ]
        },
        "default|PEERS|l2vpn_evpn": {
            "admin_status": "up",
            "route_map_in": [
                "ALLOW"
            ],
            "route_map_out": [
                "ALLOW"
            ],
            "unchanged_nexthop": "true"
        }
    },
    "ROUTE_MAP": {
        "ALLOW|1": {
            "route_operation": "permit"
        }
    },
    "ROUTE_MAP_SET": {
        "ALLOW": {}
    }
}
```

- [X] 202411

master as of 20241206

[libyang bug] workaround leaf-list via uses bug in BGP route-map

### DIFF
--- a/src/sonic-yang-models/yang-models/sonic-bgp-neighbor.yang
+++ b/src/sonic-yang-models/yang-models/sonic-bgp-neighbor.yang
@@ -28,6 +28,10 @@ module sonic-bgp-neighbor {
     //    prefix vlan;
     //}
 
+    import sonic-route-map {
+        prefix rmap;
+    }
+
     import sonic-bgp-global {
         prefix bgpg;
     }
@@ -130,7 +134,26 @@ module sonic-bgp-neighbor {
                     }
                     description "BGP Neighbor, it will be neighbor address or interface name";
                 }
+
                 uses bgpcmn:sonic-bgp-cmn-af;
+
+                // These should logically be in sonic-bgp-cmn-af, however due to a libyang2 bug when including
+                // a leaf-list entry, there are validation failures.
+                leaf-list route_map_in {
+                    type leafref {
+                        path "/rmap:sonic-route-map/rmap:ROUTE_MAP_SET/rmap:ROUTE_MAP_SET_LIST/rmap:name";
+                    }
+                    description "Route-map filter for incoming routes";
+                    max-elements 1;
+                }
+
+                leaf-list route_map_out {
+                    type leafref {
+                        path "/rmap:sonic-route-map/rmap:ROUTE_MAP_SET/rmap:ROUTE_MAP_SET_LIST/rmap:name";
+                    }
+                    description "Route-map filter for outgoing routes";
+                    max-elements 1;
+                }
             }
         }
     }
--- a/src/sonic-yang-models/yang-models/sonic-bgp-common.yang
+++ b/src/sonic-yang-models/yang-models/sonic-bgp-common.yang
@@ -382,22 +382,6 @@ module sonic-bgp-common {
             description "Interval after which connection will get re-established";
         }
 
-        leaf-list route_map_in {
-            type leafref {
-                path "/rmap:sonic-route-map/rmap:ROUTE_MAP_SET/rmap:ROUTE_MAP_SET_LIST/rmap:name";
-            }
-            description "Route-map filter for incoming routes";
-            max-elements 1;
-        }
-
-        leaf-list route_map_out {
-            type leafref {
-                path "/rmap:sonic-route-map/rmap:ROUTE_MAP_SET/rmap:ROUTE_MAP_SET_LIST/rmap:name";
-            }
-            description "Route-map filter for outgoing routes";
-            max-elements 1;
-        }
-
         leaf soft_reconfiguration_in {
             type boolean;
             description "Inbound soft reconfiguration";
--- a/src/sonic-yang-models/yang-models/sonic-bgp-peergroup.yang
+++ b/src/sonic-yang-models/yang-models/sonic-bgp-peergroup.yang
@@ -15,6 +15,10 @@ module sonic-bgp-peergroup {
         prefix bgpg;
     }
 
+    import sonic-route-map {
+        prefix rmap;
+    }
+
     organization
         "SONiC";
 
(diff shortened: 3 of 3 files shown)


bradh352:
rebased to force rebuild to see if general sonic CI tests are working yet

bradh352:
@qiluo-msft @lguohan please review

bradh352:
rebased to try to force rebuild since last build hung

bradh352:
rebased again since last build failed in unrelated test

bradh352:
rebased against master to force rebuild

bradh352:
@qiluo-msft can this be merged?

bradh352:
@qiluo-msft any reason this can't be merged?

src/sonic-yang-models/yang-models/sonic-bgp-neighbor.yang
```diff
@@ -130,7 +134,26 @@ module sonic-bgp-neighbor {
                     }
                     description "BGP Neighbor, it will be neighbor address or interface name";
                 }
+
                 uses bgpcmn:sonic-bgp-cmn-af;
+
+                // These should logically be in sonic-bgp-cmn-af, however due to a libyang2 bug when including
```
venkatmahalingam, line 140: Can you create an issue for tracking?

venkatmahalingam (approved):

src/sonic-yang-models/yang-models/sonic-bgp-neighbor.yang
bradh352: done in #21525 21525

ganglyu:
would you please add unit test?

bradh352:
@ganglyu
> would you please add unit test?

I'll come back to this issue once I know if there's any appetite to fix this the ***right*** way by upgrading to libyang3 https://lists.sonicfoundation.dev/g/sonic-dev/message/938

I'm offering to do the heavy lifting on it, but if I can't get guarantees on it ever actually getting reviewed and merged (in a timely manner), then I may need to drop it.

ganglyu:
@bradh352 

I don't think we can upgrade libyang. SONiC relies on a feature that requires backlinks from libyang, which were removed in versions after 1.0.73

bradh352:
@ganglyu
> I don't think we can upgrade libyang. SONiC relies on a feature that requires backlinks from libyang, which were removed in versions after 1.0.73

I agree that I do not see a replacement for what the leaf backlinks node previously provided.  It may be there of course, but not really documented.  From what I can tell, we can just use `lysc_module_dfs_full()` to scan the entire tree, use `lysc_node_lref_target()` and `lysc_path()` to generate the paths and make our own backlink structure and build a cache.... or still use that function but only store results for our desired base.  

How fast would this be?  Not sure, but I can't imagine it being that bad just walking the entire tree once ... granted this would be happening in Python not C code.  Also supposedly libyang2/3 are much faster than libyang1, so we may see a performance increase still even if we have to add some additional overhead.

wen587 (approved):
lgtm

ganglyu:
@bradh352 
You are welcome to explore and contribute.
LinkedIn tried this for several months but stopped. Their PR might still be useful for you:
https://github.com/sonic-net/sonic-buildimage/pull/9545

bradh352:
@ganglyu 
> @bradh352 You are welcome to explore and contribute. LinkedIn tried this for several months but stopped. Their PR might still be useful for you: #9545

Thanks for the link, I hadn't seen that ticket.

Months?? That seems ... strange.   Granted, they were trying to do it with libyang1 so that could have been their roadblock.

I will have to send some patches to libyang-python upstream to support calling into the library functions to accomplish this, but they accepted my memory leak patch pretty quickly already so I'm not foreseeing any issues.

Is there a team I need to be in communication with to review my PRs when ready?  I sent an email to the sonic-dev mailing list about this project and it was met with silence.

bradh352:
@ganglyu
> request Qiluo and me, and we will add the necessary reviewers.

I'll take you up on that, you should have a standalone PR which is step 1 to pull libyang3 and libyang3-py3 into sonic-buildimage today.  I'm working on splitting out those commits into a separate repo since just that migration is fairly sizeable and would make it easier to review the other stuff if that's not in the same PR.

bradh352:
@ganglyu please review #21679 which is the first in a set of PRs for what I was mentioning :)

bradh352:
@ganglyu interestingly, this bug is still present after upgrading to libyang3.  

After further research, it appears this is due to a bug in sonic-yang-mgmt.  It tries to translate config_db.json format to a valid yang data format, and in doing so it uses the raw parsed yang schema to try to determine how it needs to be formatted.   Being that it is a raw format (rather than the compiled schema), it then has to merge things like 'uses' clauses, and is only merging leaf nodes and not leaf-list nodes, hence the issue.

It shouldn't be a big deal to fix this oversight, but honestly the overall approach it is taken probably needs to be rethought at some point.

sonic-mgmt-common also does this conversion, but in a different way, and in Golang.  As of the pending libyang3 PRs, at least that is occurring on the compiled schema format so it doesn't have to do things like uses clause evaluation.

Considering how complex this process is, it would probably be beneficial to have this conversion done in one place that is shared .... that said being one is in python and one is in Golang, that is easier said than done.

ganglyu:
@bradh352 
I'm not familiar with sonic-yang-mgmt, could you please discuss this issue in the UMF work group meeting?

ganglyu (approved):

bradh352:
closing, proper fix in #21907 ganglyu you previously approved this PR.  I'm going to close it since an actual fix is now available.  Please review that PR and approve.

PR #18512, opened 2024-03-29
T2-VOQ-Chassis: VS support

deepak-singhal0408:
#### Why I did it
Changes to ensure sonic_vs.img has everything required for it to be emulated as T2-VOQ-Chassis: Supervisor/Linecard..
With this change, 
1. The image could be used to emulate Supervisor or Linecard
2. the image could be used to emulate different Linecard HWSKus.

##### Work item tracking
- Microsoft ADO **(number only)**:
27402561
#### How I did it
Following Changes are made as part of this PR:
1. On VS image, Database containers to be recreated again upon reboot. This will help change single asic VS image to multi-asic(docker_image_ctl.j2).
2. Copy all Sup and Linecard HWSKUs directories under kvm_platform directory. Create lanemap.ini, coreportindexmap.ini, fabriclanemap.ini under each hwsku/Asic directory. Copy asic.conf file from each platform directories to their child HW_SKU directories under kvm platform directory (This will help emulate different HW-Sku's with their respective num_asic asics):  (sonic-device-data/Makefile). 
3. New sonic-platform package for VS platforms. This is needed
 3.1:  To ensure, that database containers bring up goes through. The database containers on linecards fetch the supervisor slot_num, current_slot num etc.
PS: This package will be available on pizza box vs platforms as well. However, it will be noop there, as the package expects a metadata file, which will only be available on chassis VS platforms.
4. Changes to generate unique mac-address on VS platforms. This is done by using the device_hostname string (which will always be unique). Change to provide unique MAC address per asic on multi-asic VOQ VS platforms(sonic-cfggen, device_info.py, minigraph.py)
5. topology.service file to be dependent on sonic.target so that as part of config load_minigraph/config reload this service gets invoked.
6. topology.sh changes to move ports to their respective namespace (only applicable on multi-asic platforms)

#### How to verify it
1. Verified that sonic-vs.img.gz gets built succesfully and single asic VS DUT comes up fine(deployed vms-kvm-t0 topology) and  could see that all containers come up fine. The show commands work as expected.
2. Verified bringup on VOQ chassis with different flavor of linecard emulation.

#### Which release branch to backport (provide reason below if selected)

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

#### Tested branch (Please provide the tested image version)

- [ ] 
- [ ]

PR #21089, opened 2024-12-07
[Bug Fix] docker_image_ctl.j2: remove the logic that recreates database containers on VS platform

BYGX-wcr:
#### Why I did it

To support the emulation of VS chassis, we need to remove the existing critical service containers before transforming the HwSKU of the VS device. A previous PR [#18512](https://github.com/sonic-net/sonic-buildimage/pull/18512) introduced a change to the docker_image_ctl.j2 that forces VS images to recreate database containers every time the OS is cold started while the behaviors of other containers(swss/bgp/teamd/syncd) remained unchanged. As a consequence, when the VS device is rebooted without proper human intervention, the database containers will be recreated while the other services will reuse existing containers. That can cause the swss/bgp/syncd containers to become invalid if the database containers get recreated with a different container ID, because swss/bgp/syncd containers are configured to use the database containers as the underlying networking stack.

By further investigation, we have found that it is not necessary to recreate the database containers in /usr/bin/database.sh to perform HwSKU transformation. So, we should remove this logic.

##### Work item tracking
- Microsoft ADO **(number only)**: 30454307

#### How I did it
Remove the code that removes existing database containers in /usr/bin/database.sh

#### How to verify it

* Build VS image and install it on a KVM sonic-vs
* Reboot the sonic-vs multiple times and check if all the containers are started successfully every time.

The original change was introduced into 202205 branch so we need to backport this fix.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [X] 202205
- [ ] 202211
- [ ] 202305
- [x] 202405

- [X] 20220532.72

docker_image_ctl.j2: change to not remove existing database containers on VS platform

### DIFF
--- a/files/build_templates/docker_image_ctl.j2
+++ b/files/build_templates/docker_image_ctl.j2
@@ -451,7 +451,7 @@ start() {
 
     DOCKERCHECK=`docker inspect --type container ${DOCKERNAME} 2>/dev/null`
     if [ "$?" -eq "0" ]; then
-        {%- if docker_container_name == "database" and sonic_asic_platform == "vs" %}
+        {%- if sonic_asic_platform == "vs" %}
         echo "Removing existing ${DOCKERNAME} container"
         docker rm -f ${DOCKERNAME}
         {%- else %}
--- a/files/build_templates/docker_image_ctl.j2
+++ b/files/build_templates/docker_image_ctl.j2
@@ -451,37 +451,28 @@ start() {
 
     DOCKERCHECK=`docker inspect --type container ${DOCKERNAME} 2>/dev/null`
     if [ "$?" -eq "0" ]; then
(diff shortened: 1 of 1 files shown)


arlakshm (approved):
lgtm

deepak-singhal0408:
@BYGX-wcr Have you tried single asic to multi-asic transition?

deepak-singhal0408:
@BYGX-wcr , as discussed, please update the description, on whats needed for multi-asic transition (remove database container before reboot).?  Also make sure the change works on our internal emulation. thanks

BYGX-wcr:
@deepak-singhal0408 , I have updated the description of the PR.

deepak-singhal0408 (approved):

PR #21114, opened 2024-12-10
Enable FRR side bmp option by default and fix UT issue

FengPan-Frank:
#### Why I did it
we need to enable bmp feature from frr side by default, so that later we're able to rollout bmp container via KubeSonic path.

##### Work item tracking
- Microsoft ADO **(number only)**:30471294

#### How I did it
Turn on bmp config option and fix UT cases.

#### How to verify it
Before bmp service start, frr will try to connect bmp service every 1 minute and leave some logging.
After bmp service start, frr will connect to bmp service successfully.

image

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]


StormLiangMS:
hi @FengPan-Frank 
1. 
Do we have data for the cpu/memory load increase when enable this feature? could you take a try with mellanox/broadcom/cisco platforms? for example, when disable this feature, what's the cpu/memory usage, and compare with data enabled this feature?

2. could you address the ms_conflict?

PR #21174, opened 2024-12-13
PVST Feature enable

divyachandralekha:
What I did
Added new STP docker and rules to compile STP
Added libevent required for STP

How to verify it
After this PR STP docker will be available


src/libevent/Makefile
```diff
@@ -0,0 +1,20 @@
+.ONESHELL:
+SHELL = /bin/bash
+.SHELLFLAGS += -e
+
+MAIN_TARGET = libevent_$(LIBEVENT_VERSION)_$(CONFIGURED_ARCH).deb
+
+$(addprefix $(DEST)/, $(MAIN_TARGET)): $(DEST)/% :
+	# Obtaining the libevent
```
yaqiangz, line 8: For mellanox build, seems some dependency for libevent in docker-dhcp-relay is broken by the new libevent explicit building. Because macro `install_debian_packages` would invoke `apt-get -y install -f`, hence libevent required in docker-dhcp-relay would be install automatically (Right side of below pic). But with current PR (Left side of below pic), version of libevent is missing and there is not automatic installing.
image

Why does libevent need to be built separately here? I didn't find there is any customized patches for it, from my side, it should be installed by apt-get in Docker file of docker-stp, and it should avoid such issue in other container

divyachandralekha: Hi @yaqiangz 
I have removed the libevent dependency , dhcprelay compilation issue is not seen now. 
Thanks for the help.

PR #21200, opened 2024-12-17
[eventd] Ignore control character event if it comes in UT

zbud-msft:
#### Why I did it

Fixes https://github.com/sonic-net/sonic-buildimage/issues/21140

In https://github.com/sonic-net/sonic-buildimage/pull/20024 and https://github.com/sonic-net/sonic-swss-common/pull/906, we made the change that when a control character is read, zmq_message_read will return with rc 0, which will create an empty internal event. 

As part of eventd design, empty structured events are dropped, which leads to control characters being a no-op which is the expected behavior.

In UT, we are still always expecting a control character to be the first message to be read by zmq which is not the case as described in https://github.com/sonic-net/sonic-buildimage/pull/20024. We are also not ignoring empty events that are read which is being done by eventd. 

With this change, control characters are properly ignored if it does after the first test event.

##### Work item tracking
- Microsoft ADO **(number only)**:28728116

#### How I did it

Ignore empty structured events and not expect first zmq_message_read to be control character.

#### How to verify it

Manual test/pipeline

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/src/sonic-eventd/tests/eventd_ut.cpp
+++ b/src/sonic-eventd/tests/eventd_ut.cpp
@@ -164,16 +164,12 @@ void run_cap(void *zctx, bool &term, string &read_source,
     EXPECT_EQ(0, zmq_setsockopt(mock_cap, ZMQ_SUBSCRIBE, "", 0));
     EXPECT_EQ(0, zmq_setsockopt(mock_cap, ZMQ_RCVTIMEO, &block_ms, sizeof (block_ms)));
 
-    zmq_msg_t msg;
-    zmq_msg_init(&msg);
-    int rc = zmq_msg_recv(&msg, mock_cap, 0);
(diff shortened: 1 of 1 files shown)


dgsudharsan:
@lguohan Can we merge this PR. This will help us reduce instability in builds because of eventd UT failure

zbud-msft:
Ran ~25 image builds, unable to see failure a single time building eventd

qiluo-msft (approved):

r12f:
hi @kperumalbfn , for manual merged PRs, we will need to manually update the tag here. otherwise, it will create confusion in PR tracking.

PR #21209, opened 2024-12-18
[Fix 20284]: Enhance smartswitch environment variables parsing

### LINKED ISSUE #20284: Checking for DPUs in platform.json is adding delay to WR/FR reconciliation
#### Description

In 202405 and above, two extra steps are added before the start of every container which checks NUM_DPU and IS_DPU_DEVICE by parsing the platform.json file using the jq tool. This is only relevant for Smartswitch. However, this is adding some delay during the reconciliation phase of WR/FR resulting

```
    # Parse the platform.json file to get the platform specific information
    PLATFORM_JSON=/usr/share/sonic/device/$PLATFORM/platform.json
    if [ -f "$PLATFORM_JSON" ]; then
        NUM_DPU=$(jq -r '.DPUS | length' $PLATFORM_JSON 2>/dev/null)
        jq -e '.DPU' $PLATFORM_JSON >/dev/null
        if [[ $? -eq 0 ]]; then
            IS_DPU_DEVICE="true"
        else
            IS_DPU_DEVICE="false"
        fi
    fi
```

When there is load on CPU, both the jq calls are adding > 1 sec to the start of swss and almost 0.5 sec to the start of syncd. There are also present in teamd and bgp container start flow which may cause extra contention on the CPU.

swss start
image

syncd start
image

These 1.5 - 2 sec can be avoided as this can be optimized. Ideally, this should be read once maybe in the rc.local script and the values must be saved in some file like sonic-environment etc and should be read during the start of every container

**Note: All the values were taken from MSN2700 platform, Intel Celeron 2 core CPU**

Pterosaur:
#### Why I did it
Fix #20284

In 202405 and above, two extra steps are added before the start of every container which checks NUM_DPU and IS_DPU_DEVICE by parsing the platform.json file using the jq tool. This is only relevant for Smartswitch. However, this is adding some delay during the reconciliation phase of WR/FR resulting

##### Work item tracking

#### How I did it

Set the environment variables for systemd by systemd-sonic-generator.

#### How to verify it

1. Check there is no `jq` command under the swss.sh start and syncd.sh start from the sonic-bootchart
2. Regression test:

```
echo '{"DPUS":{"dpu0":{}, "dpu1":{}}}' | sudo tee /usr/share/sonic/device/x86_64-kvm_x86_64-r0/platform.json
docker rm -f database
sudo reboot

...

admin@vlab-01:~$ sudo systemctl list-unit-files '*midplane*'
UNIT FILE                    STATE           PRESET
midplane-network-dpu.service disabled        enabled
midplane-network-npu.service enabled-runtime  enabled


# The DPU databases should be created
admin@vlab-01:~$ docker ps -a
CONTAINER ID   IMAGE                                COMMAND                  CREATED          STATUS         PORTS     NAMES
4455f7ab611c   docker-database:latest               "/usr/local/bin/dock…"   5 minutes ago    Up 5 minutes             databasedpu0
8e983b5beb27   docker-database:latest               "/usr/local/bin/dock…"   5 minutes ago    Up 5 minutes             databasedpu1

# Environment variables should be inserted to the service file
admin@vlab-01:~$ cat $(sudo find /etc -name 'swss*service')
[Unit]
Description=switch state service
Requires=database@dpu0.service

Requires=database@dpu1.service

After=database@dpu0.service

After=database@dpu1.service

Requires=config-setup.service
After=config-setup.service
BindsTo=sonic.target
After=sonic.target
Before=ntp-config.service
StartLimitIntervalSec=1200

StartLimitBurst=3

[Service]
Environment="NUM_DPU=2"
Environment="IS_DPU_DEVICE=false"
User=root
Environment=sonic_asic_platform=vs
ExecStartPre=/usr/local/bin/swss.sh start
ExecStart=/usr/local/bin/swss.sh wait
ExecStop=/usr/local/bin/swss.sh stop
RestartSec=30

[Install]
WantedBy=sonic.target

```

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/src/systemd-sonic-generator/systemd-sonic-generator.cpp
+++ b/src/systemd-sonic-generator/systemd-sonic-generator.cpp
@@ -11,6 +11,9 @@
 #include <string>
 #include <sstream>
 #include <unordered_set>
+#include <fstream>
+#include <unordered_map>
+#include <regex>
 
 #define MAX_NUM_TARGETS 48
 #define MAX_NUM_INSTALL_LINES 48
@@ -383,6 +386,56 @@ static void replace_multi_inst_dep(const char *src) {
      */
     remove(src);
     rename(tmp_file_path, src);
+}
+
+static void update_environment(const std::string &src)
+{
+    std::ifstream src_file(src);
+    std::ofstream tmp_file(src + ".tmp");
+    bool has_service_section = false;
+    std::string line;
+
+    // locate the [Service] section
+    while (std::getline(src_file, line)) {
+        tmp_file << line << std::endl;
+        if (line.find("[Service]") != std::string::npos) {
+            has_service_section = true;
+            break;
+        }
+    }
+
+    if (!has_service_section) {
+        tmp_file << "[Service]" << std::endl;
+    }
+
+    std::unordered_map<std::string, std::string> env_vars;
+    env_vars["IS_DPU_DEVICE"] = (smart_switch_dpu ? "true" : "false");
+    env_vars["NUM_DPU"] = std::to_string(num_dpus);
+
+    for (const auto& [key, value] : env_vars) {
+        tmp_file << "Environment=\"" << key << "=" << value << "\"" << std::endl;
+    }
+
+    // copy the rest of the file
+    if (has_service_section) {
+        const static std::regex env_var_regex("Environment=\"?(\\w+)");
+        while (std::getline(src_file, line)) {
+            // skip the existing environment variables
+            std::smatch match;
+            if (std::regex_search(line, match, env_var_regex)) {
+                if (env_vars.find(match[1]) != env_vars.end()) {
+                    continue;
+                }
+            }
+            tmp_file << line << std::endl;
+        }
+    }
+
+    src_file.close();
+    tmp_file.close();
+    // remove the original file and rename the temporary file
+    remove(src.c_str());
+    rename((src + ".tmp").c_str(), src.c_str());
 }
 
 int get_install_targets(std::string unit_file, char* targets[]) {
@@ -1154,6 +1207,8 @@ int ssg_main(int argc, char **argv) {
 
             free(targets[j]);
         }
+
+        update_environment(get_unit_file_prefix() + unit_instance);
 
         free(unit_files[i]);
     }
--- a/files/build_templates/docker_image_ctl.j2
+++ b/files/build_templates/docker_image_ctl.j2
@@ -372,18 +372,6 @@ start() {
         source $PLATFORM_ENV_CONF
     fi
 
-    # Parse the platform.json file to get the platform specific information
-    PLATFORM_JSON=/usr/share/sonic/device/$PLATFORM/platform.json
-    if [ -f "$PLATFORM_JSON" ]; then
-        NUM_DPU=$(jq -r '.DPUS | length' $PLATFORM_JSON 2>/dev/null)
-        jq -e '.DPU' $PLATFORM_JSON >/dev/null
-        if [[ $? -eq 0 ]]; then
-            IS_DPU_DEVICE="true"
-        else
-            IS_DPU_DEVICE="false"
-        fi
-    fi
-
     {%- if sonic_asic_platform == "broadcom" %}
     {%- if docker_container_name == "syncd" %}
     # Set the SYNCD_SHM_SIZE if this variable not defined
--- a/src/systemd-sonic-generator/ssg-test.cc
+++ b/src/systemd-sonic-generator/ssg-test.cc
@@ -231,6 +231,7 @@ class SsgMainTest : public SsgFunctionTest {
             while (getline(file, line) && !found) {
                 if (str == line) {
                     found = true;
+                    break;
                 }
             }
         return found;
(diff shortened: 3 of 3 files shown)


src/systemd-sonic-generator/systemd-sonic-generator.cpp
```diff
@@ -385,6 +388,56 @@ static void replace_multi_inst_dep(const char *src) {
...
+
+    if (!has_service_section) {
+        tmp_file << "[Service]" << std::endl;
+    }
+
+    std::unordered_map<std::string, std::string> env_vars;
+    env_vars["IS_DPU_DEVICE"] = (smart_switch_dpu ? "true" : "false");
+    env_vars["NUM_DPU"] = std::to_string(num_dpus);
+
+    for (const auto& [key, value] : env_vars) {
```
vivekrnv, line 415: I think services files such as database, swss, syncd are enabled at build time. 
Are we certain that, these will be started only after systemd-sonic-generator is run atleast once?

src/systemd-sonic-generator/systemd-sonic-generator.cpp
```diff
@@ -385,6 +388,56 @@ static void replace_multi_inst_dep(const char *src) {
...
+        }
+    }
+
+    if (!has_service_section) {
+        tmp_file << "[Service]" << std::endl;
+    }
+
+    std::unordered_map<std::string, std::string> env_vars;
+    env_vars["IS_DPU_DEVICE"] = (smart_switch_dpu ? "true" : "false");
+    env_vars["NUM_DPU"] = std::to_string(num_dpus);
```
vivekrnv, line 413: This is a good approach to solve this issue but i see one setback. If we need to add new env variables, ssg code should be updated. which IMO is not very flexible. I have an idea for generic solution, let me know what you think

1) Add a oneshot service very early in the boot. Read static env variables (Eg: $PLATFORM, $NUM_ASIC, $NUM_DPU, $SONIC_BOOT_TYPE etc) and write them to a common file `/etc/sonic/static-env-variables` 
2) We can leverage ssg to write `EnvironmentFile=/etc/sonic/static-env-variables `option to all the services. making the ssg code minimal and flexible. 
3) We can potentially clean a lot of code under docker_image_ctl with this approach

src/systemd-sonic-generator/systemd-sonic-generator.cpp
vivekrnv: Please also verify the changes on smartswitch platform and make sure multiple database instances are created

Pterosaur: Thanks for your comments.
Yes, the generator will be run only once and before all services started.
I did confirm the smartswitch scenario and pasted the results on the PR description.

src/systemd-sonic-generator/systemd-sonic-generator.cpp
Pterosaur: Good idea, I can have a try.

Pterosaur: I tried it, but I feel it wasn't easy yet.

1. The shell of oneshot service you mentioned might look like the following. But at this time, the database service hasn't ready, so I have to parse the `/host/machine.conf` as same as what systemd-sonic-generator did if I would like to get the variable. The SSG has done this by an efficient function, C function, why should I do it again by a shell?

```shell

SYSTEMD_ENV_FILE="/etc/sonic/static_env"

# Load platform from sonic-cfggen

PLATFORM=${PLATFORM:-`sonic-cfggen -H -v DEVICE_METADATA.localhost.platform`}

echo "PLATFORM='${PLATFORM}'" >> "$SYSTEMD_ENV_FILE"

# Parse environment from platform.json

PLATFORM_JSON=/usr/share/sonic/device/$PLATFORM/platform.json

if [ -f "$PLATFORM_JSON" ]; then

# Environment variables for Smart Switch

    NUM_DPU=$(jq -r '.DPUS | length' $PLATFORM_JSON 2>/dev/null)
    if [[ -z "$NUM_DPU" ]]; then
        NUM_DPU=0
    fi

    jq -e '.DPU' $PLATFORM_JSON >/dev/null
    if [[ $? -eq 0 ]]; then
        IS_DPU_DEVICE="true"
    else
        IS_DPU_DEVICE="false"
    fi

    echo "NUM_DPU='${NUM_DPU}'" >> "$SYSTEMD_ENV_FILE"
    echo "IS_DPU_DEVICE='${IS_DPU_DEVICE}'" >> "$SYSTEMD_ENV_FILE"
fi

```
2. As your proposal, if we want to add new env variables, we have to update the shell script. I don't see any difference or challenge in doing this via SSG. If you feel that C code isn't flexible, I have to say we had an old SSG with python code previously, But we discarded it and rewrote it via C due to the efficiency issues.

vivekrnv: Yes, script couldn't use DB just like SSG

Hmm, difference is we need to run the script once. Since the oneshot service runs before other services are even started, CPU should be fairly free and should be executed quickly. we do need to benchmark this solution to measure impact. 

Advantage being load on SSG is less, all it does it to add EnvironmentFIle= and with some optimization we don't even need to edit and write to the .service file after the first

src/systemd-sonic-generator/systemd-sonic-generator.cpp
vivekrnv: Thanks for testing

src/systemd-sonic-generator/systemd-sonic-generator.cpp
vivekrnv: In any case, i'm okay with the current solution. We might move to the generic oneshot service if required in the future

src/systemd-sonic-generator/systemd-sonic-generator.cpp
```diff
@@ -385,6 +388,56 @@ static void replace_multi_inst_dep(const char *src) {
...
+
+    for (const auto& [key, value] : env_vars) {
+        tmp_file << "Environment=\"" << key << "=" << value << "\"" << std::endl;
```
vivekrnv, line 416: Do we need to add Environment= option if the file doesn't have`[Service]` section in the file

Pterosaur: I worry that the `Service` section might be introduced by other functions or modules in the future. And I don't see any side-effects if I define some environment variables at this point.

src/systemd-sonic-generator/systemd-sonic-generator.cpp
Pterosaur: If this PR looks good to you, could you please help to approve it.

src/systemd-sonic-generator/systemd-sonic-generator.cpp
vivekrnv: Please make sure, systemd won't throw an error if we add Environment= values after the last section without [Service]

Pterosaur: If there is no Service section, I will add it.

image

vivekrnv: Sounds good to me

vivekrnv (approved):

Pterosaur:
Hi @lguohan , Please help to review or merge this PR.

lguohan:
@Pterosaur , can you put delay numbers in the description?

PR #21255, opened 2024-12-21
Enhanced minigraph parser for packet-chasiss production graph 

abdosi:
What I did:
Fixed 2 issue while parsing production minigraph:
1.  `bfd` flag not being set for STATIC_ROUTE table
2.   Different `mac-address` not getting generated for each ASIC

How I fixed:
In production `switch_type` is derived using `ForwardingMethod` attribute. Modified code to use `chassis_type` which is derived using `ForwardingMethod` 

How I verify:
Added UT
Manual Verification.


abdosi:
@rlhui for viz.  @qiluo-msft this is bug fix so we should be able to get exception on this.

arlakshm (approved):

qiluo-msft (approved):

abdosi:
@rlhui : can you help with merge of this

PR #21291, opened 2024-12-27
Fix the issue of no INFO level syslog for caclmgrd

ZhaohuiS:
#### Why I did it
Fix issue https://github.com/sonic-net/sonic-buildimage/issues/21290
1. No info level log found for systemctl caclmgrd service in syslog file in 202405 images, only warning log found in /var/log/syslog
2. set_min_log_priority_info fuction doesn't work

##### Work item tracking
30611546
#### How I did it
After https://github.com/sonic-net/sonic-buildimage/pull/17171/, it added those log functions in `daemon_base.py`, actually, in those function, it didn't set priority, by default, it sets default level to `NOTICE`(https://github.com/sonic-net/sonic-buildimage/blame/9588a40abc27911a82aa2f116c17ce06c2cd7093/src/sonic-py-common/sonic_py_common/logger.py#L39), so even in `log_info`, without priority parameter, the default level is still `NOTICE`, the log which is printed with `INFO` level can't be printed into syslog.
Because both `src/sonic-py-common/sonic_py_common/syslogger.py` and `src/sonic-py-common/sonic_py_common/logger.py` implement these functions, we don't need to add those functions in `daemon_base.py`
The solution is remove them.
And by the way, in `caclmgrd`, it calls `caclmgr.set_min_log_priority_info()`, `caclmgrd` is an instance of `DaemonBase` whose parent class is `Logger`, `set_min_log_priority_info` both works for `use_syslogger=True` or `use_syslogger=Flase` in daemon_base.py init function.

```
  def log(self, priority, message, also_print_to_console=False):
        self.logger_instance.log(priority, message, also_print_to_console)

    def log_error(self, message, also_print_to_console=False):
        self.logger_instance.log_error(message, also_print_to_console)

    def log_warning(self, message, also_print_to_console=False):
        self.logger_instance.log_warning(message, also_print_to_console)

    def log_notice(self, message, also_print_to_console=False):
        self.logger_instance.log_notice(message, also_print_to_console)

    def log_info(self, message, also_print_to_console=False):
        self.logger_instance.log_info(message, also_print_to_console)

    def log_debug(self, message, also_print_to_console=False):
        self.logger_instance.log_debug(message, also_print_to_console)
```

https://github.com/sonic-net/sonic-buildimage/blame/9588a40abc27911a82aa2f116c17ce06c2cd7093/src/sonic-py-common/sonic_py_common/logger.py#L39
```
       # Set the default minimum log priority to LOG_PRIORITY_NOTICE
        self.set_min_log_priority(self.LOG_PRIORITY_NOTICE)
```
#### How to verify it

Test evidence:

```
from sonic_py_common import daemon_base

SYSLOG_IDENTIFIER = "TestSysLogger"
log3 = daemon_base.DaemonBase(SYSLOG_IDENTIFIER, use_syslogger=False)
log3.set_min_log_priority_info()

log3.log_info("log3: This is a info message")
log3.log_notice("log3: This is a notice message")
log3.log_warning("log3: This is a warning message")
log3.log_debug("log3: This is a debug message")
log3.log_error("log3: This is a error message")

log3.log_info("log3: This is a info message, console", also_print_to_console=True)
log3.log_notice("log3: This is a notice message, console", also_print_to_console=True)
log3.log_warning("log3: This is a warning message,console", also_print_to_console=True)
log3.log_debug("log3: This is a debug message,console", also_print_to_console=True)
log3.log_error("log3: This is a error message,consle", also_print_to_console=True)
```

Console output:
```
log3: This is a info message, console
log3: This is a notice message, console
log3: This is a warning message,console
log3: This is a error message,consle
```

Syslog file output:
```
2025 Jan  6 10:16:49.077345 str2-7050cx3-acs-13 INFO TestSysLogger[723065]: log3: This is a info message
2025 Jan  6 10:16:49.077797 str2-7050cx3-acs-13 NOTICE TestSysLogger[723065]: log3: This is a notice message
2025 Jan  6 10:16:49.077910 str2-7050cx3-acs-13 WARNING TestSysLogger[723065]: log3: This is a warning message
2025 Jan  6 10:16:49.077994 str2-7050cx3-acs-13 ERR TestSysLogger[723065]: log3: This is a error message
2025 Jan  6 10:16:49.078074 str2-7050cx3-acs-13 INFO TestSysLogger[723065]: log3: This is a info message, console
2025 Jan  6 10:16:49.086310 str2-7050cx3-acs-13 NOTICE TestSysLogger[723065]: log3: This is a notice message, console
2025 Jan  6 10:16:49.086508 str2-7050cx3-acs-13 WARNING TestSysLogger[723065]: log3: This is a warning message,console
2025 Jan  6 10:16:49.086608 str2-7050cx3-acs-13 ERR TestSysLogger[723065]: log3: This is a error message,consle
```

set `use_syslogger=True`, same test results with `use_syslogger=False`.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305
- [x] 202405

- [ ] 
- [ ]

### DIFF
--- a/src/sonic-py-common/sonic_py_common/daemon_base.py
+++ b/src/sonic-py-common/sonic_py_common/daemon_base.py
@@ -1,7 +1,5 @@
-import logging
 import signal
 import sys
-import syslog
 
 from . import device_info
 from .general import load_module_from_source
--- a/src/sonic-py-common/sonic_py_common/daemon_base.py
+++ b/src/sonic-py-common/sonic_py_common/daemon_base.py
@@ -55,25 +55,6 @@ class DaemonBase(Logger):
         if not signal.getsignal(signal.SIGTERM):
             signal.signal(signal.SIGTERM, self.signal_handler)
 
-    def log(self, priority, message, also_print_to_console=False):
-        self.logger_instance.log(priority, message, also_print_to_console)
-
-    def log_error(self, message, also_print_to_console=False):
-        self.logger_instance.log_error(message, also_print_to_console)
-
-    def log_warning(self, message, also_print_to_console=False):
-        self.logger_instance.log_warning(message, also_print_to_console)
-
-    def log_notice(self, message, also_print_to_console=False):
-        self.logger_instance.log_notice(message, also_print_to_console)
-
-    def log_info(self, message, also_print_to_console=False):
-        self.logger_instance.log_info(message, also_print_to_console)
-
-    def log_debug(self, message, also_print_to_console=False):
-        self.logger_instance.log_debug(message, also_print_to_console)
-
-    
     # Default signal handler; can be overridden by subclass
     def signal_handler(self, sig, frame):
         if sig == signal.SIGHUP:
--- a/src/sonic-py-common/sonic_py_common/syslogger.py
+++ b/src/sonic-py-common/sonic_py_common/syslogger.py
@@ -21,7 +21,7 @@ class SysLogger:
     """
 
     DEFAULT_LOG_FACILITY = SysLogHandler.LOG_USER
-    DEFAULT_LOG_LEVEL = logging.INFO
+    DEFAULT_LOG_LEVEL = logging.NOTICE
 
     def __init__(self, log_identifier=None, log_facility=DEFAULT_LOG_FACILITY, log_level=DEFAULT_LOG_LEVEL, enable_runtime_config=False):
         self.log_identifier = log_identifier if log_identifier else os.path.basename(sys.argv[0]) 
@@ -42,7 +42,7 @@ class SysLogger:
         
         if enable_runtime_config:
             self.update_log_level()
-                            
+
     def update_log_level(self):
         """Refresh log level.
 
--- a/src/sonic-py-common/sonic_py_common/syslogger.py
+++ b/src/sonic-py-common/sonic_py_common/syslogger.py
@@ -21,7 +21,7 @@ class SysLogger:
     """
 
     DEFAULT_LOG_FACILITY = SysLogHandler.LOG_USER
-    DEFAULT_LOG_LEVEL = logging.NOTICE
+    DEFAULT_LOG_LEVEL = logging.INFO
 
     def __init__(self, log_identifier=None, log_facility=DEFAULT_LOG_FACILITY, log_level=DEFAULT_LOG_LEVEL, enable_runtime_config=False):
         self.log_identifier = log_identifier if log_identifier else os.path.basename(sys.argv[0]) 


ZhaohuiS:
@Junchao-Mellanox  @xincunli-sonic could you please help review?

Junchao-Mellanox:
Hi @ZhaohuiS , changing the default log level will affect many other daemons. I am not sure it is what we want. I suppose there is a few options:
1. In cfgmgrd, set use_syslogger=False, or
2. Fix daemonbase to support setting log level

ZhaohuiS:
> Hi @ZhaohuiS , changing the default log level will affect many other daemons. I am not sure it is what we want. I suppose there are a few options:
> 
> 1. In cfgmgrd, set use_syslogger=False, or
> 2. Fix daemonbase to support setting log level

1. By default, it chooses SysLogger, which is not consistent as before. caclmgrd is just one of them. Or suggest to set use_syslogger=False
2. Why can't set INFO by default is not reasonable? INFO is a default level in syslog. If you set default level to NOTICE, INFO will lost by default, which doesn't make sense.

Junchao-Mellanox:
Hi, I am ok with your first option. For the second one:
```
INFO is a default level in syslog. If you set default level to NOTICE, INFO will lost by default, which doesn't make sense. I strongly recommend to set it back to INFO, not NOTICE, which is a new one, probably not be aware of by others.
```
SONiC python logger default level is NOTICE since day 0, so, changing it will affect many other daemon, such as xcvrd, thermalctld, etc.

ZhaohuiS:
@Junchao-Mellanox 
Could you provide the evidence? From what I found, the default level is INFO. Please check:
https://docs.python.org/3/library/syslog.html
image

Junchao-Mellanox:
Please check this line: https://github.com/sonic-net/sonic-buildimage/blob/9588a40abc27911a82aa2f116c17ce06c2cd7093/src/sonic-py-common/sonic_py_common/logger.py#L39. Please also review the history of this file.

ZhaohuiS:
@Junchao-Mellanox @xincunli-sonic  could you please review it again?

bingwang-ms:
@xincunli-sonic Can you help review this change?

ZhaohuiS:
caclmgr.set_min_log_priority_info() it sets min log priority to info, this function is in Logger class, SysLogger doesn't have this function. But DaemonBase still inherits Logger which implements set_min_log_priority_info, that's why even caclmgrd called this function, it didn't throw exception. But it didn't make INFO level effect in SysLogger which is actually used in caclmgrd

Even change to use Logger by setting use_syslogger=False, it still doesn't work.
The root cause is that it added a new instance for logger, self.logger_instance, any instance inherited from DaemonBase class can't change the debug level, the level they changed is their own instance, not the self.logger_instance's level.

The current fix is still not valid. It forces to use log function from Logger class, which  still doesn't make sense.
But it's very complicated to implement set_min_log_priority these functions in daemon_base.py, since SysLogger and Logger have total different level values.

For addressing no syslog issue for caclmgrd, fix it in the following PR:
https://github.com/sonic-net/sonic-host-services/pull/200

There  is still a problem for DaemonBase to set log level, track it in https://github.com/sonic-net/sonic-buildimage/issues/21361

Issue #21227, opened 2024-12-19
Rsyslog Plugin not available in containers in Arista non slim images


zbud-msft:
#### Steps to reproduce the issue:
1. Run test events in latest image in Arista non slim SKU, will fail in bgp/dhcp-relay since there is no rsyslog plugin in containers
2.
3.

#### Output of `show version`:

```
(paste your output here)
```
#### Output of `show techsupport`:

```
(paste your output here or download and attach the file here )
```

PR #21338, opened 2025-01-07
Remove rsyslog plugin from all non eventd containers

zbud-msft:
#### Why I did it

Fixes Issue #21227

- As of current, we have rsyslog_plugin and corresponding regex and configuration files in containers like bgp, swss, dhcp_relay. These binaries and files are built as part of the docker container itself. 

- This change removes that dependency by storing all rsyslog_plugin regex and conference files in the eventd docker container, and at runtime copying them to the host. 

- Fixes the issue that some images will not be able to publish certain events like "bgp-state", or "dhcp-relay-discard" due to reuse of docker containers during where eventd was disabled.

##### Work item tracking
- Microsoft ADO **(number only)**:30569670

#### How I did it

- Removed all rsyslog plugin configuration and regex files from containers such as bgp, swss, dhcp_relay. All files for configuration and regexes are built inside eventd docker container. When eventd docker is started, it will copy these files to the host and restart rsyslog.
- sshd regex and zebra regex are removed because those events are currently not being supported.
- Possible race condition with rsyslog-config is avoided due to "After" and since rsyslog-config.service is type oneshot, rsyslog-config.sh is guaranteed to complete before eventd.service is called.

#### How to verify it

Tested on all elastic pipeline for 6 different HWSKUs

Arista-7260CX, Arista720DT, Mellanox2700, Cisco 8102, Nokia M0, Arista 7060CX

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]


files/build_templates/events_info.json
```diff
@@ -5,10 +5,6 @@
             "name": "monit",
             "parse_json": "monit_regex.json"
         },
-        {
-            "name": "sshd",
```
qiluo-msft, line 9: >sshd [](http://example.com/codeflow?start=21&length=4)

why remove? I cannot find info in PR description.

zbud-msft: Updated PR description, but that event is not being supported.

qiluo-msft: I will close this comment. but still could you help me understand "sshd regex and zebra regex are removed because those events are currently not being supported." sshd does not support or regex does not support or eventd does not support?

qiluo-msft (approved):
with one question in comment

files/build_templates/events_info.json
zbud-msft: eventd will not support sshd
zebra regex is no longer supported.

qiluo-msft (approved):

PR #21346, opened 2025-01-07
[teamd] Optimize number of netlink calls by increasing buffer size and specifying max message size

saiarcot895:
#### Why I did it

Partial fix for https://github.com/sonic-net/sonic-buildimage/issues/19310.

On setups with many LAGs configured, if the teamd container is getting shut down, then there may be many messages sent from the kernel about LAGs/LAG members being removed. This may cause the netlink socket buffers to fill up.

NetworkManger (on the desktop side) uses 2MB socket buffers, whereas teamd is currently configured to use 960kB buffers. Additionally, because libnl doesn't know the size of each netlink message, it makes two `recvmsg` syscalls for each message: one to know the size of the message, and one to get the actual message. This can result in inefficiencies on weaker systems and potentially dropped netlink messages.

##### Work item tracking
- Microsoft ADO **(number only)**: 33931175

#### How I did it

Increase the size of the netlink buffers from 960kB to 2048kB. Additionally, specify the maximum size of each netlink message, so that libnl doesn't need to do a `MSG_PEEK` first to get the actual length, allocate the buffer for it, and then get the actual message.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]


judyjoseph (approved):
LGTM

PR #21377, opened 2025-01-10
Fix docker auto restart issue

FengPan-Frank:
#### Why I did it
if critical process crashes or killed, bmp docker container will not be auto-restarted.

##### Work item tracking
- Microsoft ADO **(number only)**:30807821

#### How I did it
/usr/bin/supervisor-proc-exit-listener takes in charge of critical process monitor and event publish, thus it should be autorestar-ted in any case, otherwise there might be issue if supervisor-proc-exit-listener crashes, or in some test cases like
"docker exec bmp kill -SIGKILL -1" critical processes may not work correctly in some race condition (depends on whether supervisor-proc-exit-listener is the last one to be killed)

When a container receives the SIGKILL signal to terminate its processes, the order in which the processes are actually terminated can depend on the scheduling and resource availability within the container.

Scheduling: Within a container, processes are scheduled by the operating system or container runtime. The order in which the processes are scheduled to run can impact the order of termination. The scheduler determines which process gets executed first, and this can vary depending on factors such as process priorities, resource availability, and the scheduling algorithm used.
Resource Availability: Containers share resources such as CPU, memory, and disk I/O. When a SIGKILL signal is sent to all processes, the available resources might be limited or constrained. The order in which processes get terminated can be affected by resource contention. If resources are heavily utilized, some processes might be prioritized for termination over others due to resource constraints.

as a result of this, if supervisor-proc-exit-listener is killed first before critical process, container auto restart will not be launched as expected.

#### How to verify it
image

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/dockers/docker-sonic-bmp/supervisord.conf
+++ b/dockers/docker-sonic-bmp/supervisord.conf
@@ -16,7 +16,7 @@ buffer_size=1024
 command=/usr/bin/supervisor-proc-exit-listener --container-name bmp
 events=PROCESS_STATE_EXITED,PROCESS_STATE_RUNNING
 autostart=true
-autorestart=false
+autorestart=unexpected
 buffer_size=1024
 
 [program:rsyslogd]
--- a/dockers/docker-sonic-gnmi/supervisord.conf
+++ b/dockers/docker-sonic-gnmi/supervisord.conf
@@ -16,7 +16,7 @@ buffer_size=1024
 command=/usr/bin/supervisor-proc-exit-listener --container-name gnmi
 events=PROCESS_STATE_EXITED,PROCESS_STATE_RUNNING
 autostart=true
-autorestart=false
+autorestart=unexpected
 buffer_size=1024
 
 [program:rsyslogd]
--- a/dockers/docker-sonic-restapi/supervisord.conf
+++ b/dockers/docker-sonic-restapi/supervisord.conf
@@ -16,7 +16,7 @@ buffer_size=1024
 command=/usr/bin/supervisor-proc-exit-listener --container-name restapi
 events=PROCESS_STATE_EXITED,PROCESS_STATE_RUNNING
 autostart=true
-autorestart=false
+autorestart=unexpected
 buffer_size=1024
 
 [program:rsyslogd]
--- a/dockers/docker-sonic-telemetry/supervisord.conf
+++ b/dockers/docker-sonic-telemetry/supervisord.conf
@@ -16,7 +16,7 @@ buffer_size=1024
 command=/usr/bin/supervisor-proc-exit-listener --container-name telemetry
 events=PROCESS_STATE_EXITED,PROCESS_STATE_RUNNING
 autostart=true
-autorestart=false
+autorestart=unexpected
 buffer_size=1024
 
 [program:rsyslogd]


ganglyu (approved):

qiluo-msft:
Just curios about PR description, you mentioned "some race condition". How bad is the bug? high chance or low change to repro? Please update PR description.

zbud-msft:
How can we reproduce this race condition issue? If we have steps to reproduce, can we add sonic-mgmt testcase? Do other containers need this change?

qiluo-msft (approved):

ganglyu:
> How can we reproduce this race condition issue? If we have steps to reproduce, can we add sonic-mgmt testcase? Do other containers need this change?

For instance, if we kill the supervisor-proc-exit-listener first and then kill the telemetry process, the telemetry container will not restart.

FengPan-Frank:
> Just curios about PR description, you mentioned "some race condition". How bad is the bug? high chance or low change to repro? Please update PR description.

add more comments in description.

The race issue is in mgmt test case https://github.com/sonic-net/sonic-mgmt/blob/a354b1e5d665bfc2fd5d9a4b3b22fc3fa2f50592/tests/autorestart/test_container_autorestart.py#L309, "docker exec container_name kill -SIGKILL -1" is sent to every container.

When a container receives the SIGKILL signal to terminate its processes, the order in which the processes are actually terminated can depend on the scheduling and resource availability within the container.

Scheduling: Within a container, processes are scheduled by the operating system or container runtime. The order in which the processes are scheduled to run can impact the order of termination. The scheduler determines which process gets executed first, and this can vary depending on factors such as process priorities, resource availability, and the scheduling algorithm used.
Resource Availability: Containers share resources such as CPU, memory, and disk I/O. When a SIGKILL signal is sent to all processes, the available resources might be limited or constrained. The order in which processes get terminated can be affected by resource contention. If resources are heavily utilized, some processes might be prioritized for termination over others due to resource constraints.

as a result of this, if supervisor-proc-exit-listener is killed first before critical process, container auto-restart will not be launched as expected if we don't config autorestart=unexpected there.

zbud-msft:
@FengPan-Frank If this is a generic issue within all containers, will we make this change in all containers?

zbud-msft (approved):

hdwhdw:
In general LGTM. Is there a consistent repro for the issue and verification? Say running `test_container_autorestart.py` X times? (I wonder what the practical value of X is, i imagine maybe 10 if you don't have that many critical process). Then we can verify the same flake does not happen again.

hdwhdw (approved):
In general LGTM. Is there a consistent repro for the issue and verification? Say running test_container_autorestart.py X times? (I wonder what the practical value of X is, i imagine maybe 10 if you don't have that many critical process). Then we can verify the same flake does not happen again.

FengPan-Frank:
> @FengPan-Frank If this is a generic issue within all containers, will we make this change in all containers?

only some of containers have this issue, not all containers

FengPan-Frank:
> In general LGTM. Is there a consistent repro for the issue and verification? Say running test_container_autorestart.py X times? (I wonder what the practical value of X is, i imagine maybe 10 if you don't have that many critical process). Then we can verify the same flake does not happen again.

For consistent repo, we can kill supervisor-proc-exit-listener first, then critical process exit will not restart docker container, I can add one additional test step for this into test_container_autorestart.py

PR #21462, opened 2025-01-17
[dhcp_server] Support scenario that gatway is not configured

yaqiangz:
#### Why I did it
DHCP default route shoule be an optional config to DHCP client

##### Work item tracking
- Microsoft ADO **(number only)**: 30877295

#### How I did it
Support to do not send default route to dhcp client

#### How to verify it
1. UT
2. Install new image to test

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/src/sonic-dhcp-utilities/tests/test_data/kea-dhcp4.conf.j2
+++ b/src/sonic-dhcp-utilities/tests/test_data/kea-dhcp4.conf.j2
@@ -63,10 +63,12 @@
                         "always-send": {{ config["always_send"] }}
                     },
     {%- endfor %}
+    {%- if "gateway" in subnet_info %}
                     {
                         "name": "routers",
-                        "data": "{{ subnet_info["gateway"] if "gateway" in subnet_info else subnet_info["server_id"] }}"
+                        "data": "{{ subnet_info["gateway"] }}"
                     },
+    {%- endif %}
                     {
                         "name": "dhcp-server-identifier",
                         "data": "{{ subnet_info["server_id"] }}"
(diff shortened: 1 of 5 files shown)


src/sonic-dhcp-utilities/tests/test_data/kea-dhcp4.conf.j2
```diff
@@ -63,10 +63,12 @@
                         "always-send": {{ config["always_send"] }}
                     },
     {%- endfor %}
+    {%- if "gateway" in subnet_info %}
                     {
                         "name": "routers",
-                        "data": "{{ subnet_info["gateway"] if "gateway" in subnet_info else subnet_info["server_id"] }}"
+                        "data": "{{ subnet_info["gateway"] }}"
```
Blueve, line 69: When it could be `subnet_info["server_id"]`? Seems like new code lost that feature

yaqiangz: In lin 72 to line 75 of this file. It's for clients to know which server is offering the IP address lease. For now we treated as non-configurable and always use DHCP-interface ip (Generally, it's vlan interface ip)

Blueve: Why we trying to configure `subnet_info["server_id"]` in routers entry previously?

yaqiangz: It's a invalid condition. In previous,gateway would always appear in dhcp_config and subnet_obj would always have "gateway", which would be pass to j2 template. It means that router always was configured as gateway rather than "server_id" in previous.
S50126-16461381_mark via

Blueve (approved):

Issue #21516, opened 2025-01-23
IPinIP tunnel table full error


sdszhang:
When there are a lot of interfaces in the system, SAI report IPinIP table full. This is seen more system with 200+ interfaces.

```
2025 Jan  8 00:54:15.857936 tb-1 NOTICE swss#orchagent: :- addDecapTunnelTermEntry: Created tunnel decap term entry fc00::71.
2025 Jan  8 00:54:15.859186 tb-1 ERR syncd#syncd: [none] SAI_API_TUNNEL:mptnl_xgs_ipinip_create_terminator:564 SDK bcm_tunnel_terminator_add:  sip(0xfc)
2025 Jan  8 00:54:15.859186 tb-1 ERR syncd#syncd: [none] SAI_API_TUNNEL:mptnl_xgs_ipinip_create_terminator:567 bcm_tunnel_terminator_add failed with error Table full (0xfffffffa).
2025 Jan  8 00:54:15.859212 tb-1 ERR syncd#syncd: [none] SAI_API_TUNNEL:mptnl_xgs_flexflow_create_terminator:1688 mptnl_xgs_ipinip_create_terminator failed with error -13.
2025 Jan  8 00:54:15.859212 tb-1 ERR syncd#syncd: [none] SAI_API_TUNNEL:_brcm_sai_mptnl_terminator_create:2739 mptnl_hw_create_terminator failed with error -13.
2025 Jan  8 00:54:15.859392 tb-1 ERR syncd#syncd: [none] SAI_API_TUNNEL:brcm_sai_tnl_mp_create_tunnel_term_table_entry:5273 _brcm_sai_mptnl_terminator_create failed with error -13.
2025 Jan  8 00:54:15.859686 tb-1 ERR swss#orchagent: :- create: create status: SAI_STATUS_TABLE_FULL
2025 Jan  8 00:54:15.859732 tb-1 ERR swss#orchagent: :- addDecapTunnelTermEntry: Failed to create tunnel decap term entry fc00::75/128.
2025 Jan  8 00:54:15.859785 tb-1 ERR swss#orchagent: :- handleSaiCreateStatus: Encountered failure in create operation, exiting orchagent, SAI API: SAI_API_TUNNEL, status: SAI_STATUS_TABLE_FULL
[...]

PR #21517, opened 2025-01-23
skip ipinip tunnel creation if many interfaces

sdszhang:
#### Why I did it
Temporary workaround for #21516. Issue is seen with 200+ interface. So, used >128 as condition instead of >256.

##### Work item tracking

#### How I did it
Reset the address list if port count is over 128.  So no more IPv4/IPv6 IPinIP tunnel will be created for interfaces.
tunnel interfaces on loopback and vlan MP2MP are still there.

#### How to verify it
Loaded the image with more than 200 interfaces, not seeing TABLE_FULL issue anymore.

- [ ] 
- [ ]

### DIFF
--- a/dockers/docker-orchagent/ipinip.json.j2
+++ b/dockers/docker-orchagent/ipinip.json.j2
@@ -52,11 +52,10 @@
         {%- set ipv6_vlan_addresses = ipv6_vlan_addresses.append(prefix) %}
     {%- endif %}
 {% endfor %}
-{# FIXME: SAI report tunnel TABLE_FULL with large interfaces. Disable it temporarily if over 128 ports. #}
-{% if PORT|length <= 128 %}
 {%- set ipv4_addresses = ipv4_addresses + ipv4_vlan_addresses %}
 {%- set ipv6_addresses = ipv6_addresses + ipv6_vlan_addresses %}
-{% else %}
+{# FIXME: SAI report tunnel TABLE_FULL for large topo. Disable it temporarily if over 128 IPv4/IPv6 tunnels. #}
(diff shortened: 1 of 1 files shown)


sdszhang:
@davidm-arista @r12f for review.

r12f (approved):
lgtm! thanks Shawn!

dockers/docker-orchagent/ipinip.json.j2
```diff
@@ -52,8 +52,14 @@
...
     {%- endif %}
 {% endfor %}
+{# FIXME: SAI report tunnel TABLE_FULL with large interfaces. Disable it temporarily if over 128 ports. #}
+{% if PORT|length <= 128 %}
 {%- set ipv4_addresses = ipv4_addresses + ipv4_vlan_addresses %}
 {%- set ipv6_addresses = ipv6_addresses + ipv6_vlan_addresses %}
+{% else %}
+{%- set ipv4_addresses = [] %}
+{%- set ipv6_addresses = [] %}
+{% endif %}
```
kperumalbfn, line 62: Why just the port length? IPinIP tunnel term table entries are created for all interface(port or port-channel)+vlan IPs. may be if the total ipv4/ipv6 addresses are more than 128, we could temporarily bypass

r12f: this might be even better, because vlan interface is indeed in play as well.

sdszhang: updated using address length instead.

kperumalbfn (approved):

PR #21593, opened 2025-02-01
[Mellanox] Query PSU fan speed only on presence

gpunathilell:
#### Why I did it
The `PsuFan` in Mellanox platform queries for fan speed even though the relevant sysfs files do not exist. In `psud` the speed is already overwritten on presence, since the current implementation generates an error log, it is changed to query speed only on PSU presence and return 0 if there is no presence (Previously returned default value)
##### Work item tracking

#### How I did it
Check presence using `get_presence` api and return fan speed only if it returns True

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/platform/mellanox/mlnx-platform-api/sonic_platform/fan.py
+++ b/platform/mellanox/mlnx-platform-api/sonic_platform/fan.py
@@ -258,11 +258,10 @@ class PsuFan(MlnxFan):
             return False
 
     def get_speed(self):
-        if self.get_presence():
-            return super().get_speed()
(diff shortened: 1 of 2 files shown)


oleksandrivantsiv (approved):

liat-grozovik (approved):

liat-grozovik:
@keboliu fyi

PR #21606, opened 2025-02-04
Revert unix socket for chassis db

tjchadaga:
#### Why I did it
Reverting part of https://github.com/sonic-net/sonic-buildimage/pull/21345, since Chassis DB connection from remote host (linecard) cannot use unix socket

##### Work item tracking

#### How I did it
Continue to use TCP socket for chassis DB connection

#### How to verify it
Bring up BGP on chassis, which currently fails after change in https://github.com/sonic-net/sonic-buildimage/pull/21345

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]


BYGX-wcr:
@tjchadaga , why this PR is still in draft mode?

BYGX-wcr (approved):
LGTM. Tested on virtual chassis.

src/sonic-bgpcfgd/bgpcfgd/runner.py
```diff
@@ -40,7 +40,8 @@ def add_manager(self, manager):
         db = swsscommon.SonicDBConfig.getDbId(db_name)
         if db not in self.db_connectors:
             if db_name == "CHASSIS_APP_DB":
-                self.db_connectors[db] = swsscommon.DBConnector(db_name, 0, False, '')
+                 # isTcpConn is True since we cannot use unix socket for Chassis DB connection from LC
```
qiluo-msft, line 43: ># isTcpConn is True since we cannot use unix socket for Chassis DB connection from LC [](http://example.com/codeflow?start=17&length=85)

Could you exactly revert the 2 files, not adding extra logic/comments?

tjchadaga: done

qiluo-msft (approved):

liuh-80 (approved):
Can we enable T2 test case to prevent similar regression?

PR #21634, opened 2025-02-05
[Arista] [202405] Update platform library submodules

patrickmacarthur:
#### Why I did it
- led: fix amber led for pali driver
- Make linecard temp sensor names unique per slot id

##### Work item tracking

#### How I did it

Updated both submodules to update the driver code

#### How to verify it

Boot the image on affected devices and validate the aforementioned changes

Update Arista platform driver submodules


arlakshm (approved):

bingwang-ms:
@patrickmacarthur Can you confirm this change is only for chassis?
@StormLiangMS FYI

patrickmacarthur:
> @patrickmacarthur Can you confirm this change is only for chassis? @StormLiangMS FYI

There was a fixed system patch included, but after discussion with Arvind I have reverted that patch so the only patch that is included now is the chassis linecard patch.

PR #21671, opened 2025-02-07
[Mellanox][Smartswitch] Fix for platform chassis initialization   

gpunathilell:
#### Why I did it
When a user uses TACACS credentials without permissions this causes `show interfaces status` failure, because the chassis is initialized in this case, and without permissions we get the following error: 
`RuntimeError: Unable to connect to redis (unix-socket) - Permission denied(1): Cannot assign requested address`
Tested fix on a different image, 
##### Work item tracking

#### How I did it
Changed path to not use unix socket path
#### How to verify it
As a TACACS user, try `show interfaces status` the command shouldnt fail

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/platform/mellanox/mlnx-platform-api/sonic_platform/module.py
+++ b/platform/mellanox/mlnx-platform-api/sonic_platform/module.py
@@ -267,7 +267,7 @@ class DpuModule(ModuleBase):
         self.CONFIG_DB_NAME = "CONFIG_DB"
         self.DHCP_SERVER_HASH = f"DHCP_SERVER_IPV4_PORT|bridge-midplane|{self._name.lower()}"
         self.DHCP_IP_ADDRESS_KEY = "ips@"
-        self.config_db = ConfigDBConnector(use_unix_socket_path=True)
+        self.config_db = ConfigDBConnector(use_unix_socket_path=False)
         self.config_db.connect()
         self.midplane_ip = None
         self.midplane_interface = None


oleksandrivantsiv (approved):

platform/mellanox/mlnx-platform-api/sonic_platform/module.py
```diff
@@ -267,7 +267,7 @@ def __init__(self, dpu_id):
         self.CONFIG_DB_NAME = "CONFIG_DB"
         self.DHCP_SERVER_HASH = f"DHCP_SERVER_IPV4_PORT|bridge-midplane|{self._name.lower()}"
         self.DHCP_IP_ADDRESS_KEY = "ips@"
-        self.config_db = ConfigDBConnector(use_unix_socket_path=True)
+        self.config_db = ConfigDBConnector(use_unix_socket_path=False)
```
kperumalbfn, line 270: @prgeor Do we need this change in other SKUs?

kperumalbfn (approved):

platform/mellanox/mlnx-platform-api/sonic_platform/module.py
prgeor: no

PR #21719, opened 2025-02-12
libyang: upgrade to v3 step 2 -- port sonic-yang-models test cases

bradh352:
#### Why I did it

libyang3 porting. See master tracking ticket https://github.com/sonic-net/sonic-buildimage/issues/22385

Port of sonic-yang-model test cases

Fixes https://github.com/sonic-net/sonic-buildimage/issues/22385

#### How I did it

Ported code

#### How to verify it

See test cases succeed in CI/CD

N/A

libyang: upgrade to v3 step 2 - port sonic-yang-models test cases

N/A

### DIFF
--- a/src/sonic-yang-models/tests/yang_model_pytests/test_smart_switch.py
+++ b/src/sonic-yang-models/tests/yang_model_pytests/test_smart_switch.py
@@ -381,9 +381,9 @@ class TestSmartSwitch:
 
     @pytest.mark.parametrize(
         "dpu_id, swbus_port, error_message", [
-            (0, 23606, None),
-            (1, 23607, None),
-            (7, 23613, None)]
+            ("0", 23606, None),
+            ("1", 23607, None),
+            ("7", 23613, None)]
         )
     def test_remote_dpu_swbus_port(self, yang_model, dpu_id, swbus_port, error_message):
         data = {


bradh352:
All checks passed, this is ready for review and merge.  This only migrates YANG model tests to libyang3.  It doesn't migrate any actual code.

src/sonic-yang-models/tests/yang_model_tests/test_yang_model.py
```diff
@@ -222,9 +235,12 @@ def runExceptionTest(self, test):
                 log.info(desc + " Passed\n")
                 return PASS
             else:
-                raise Exception("Mismatch {} and {}".format(eStr, s))
+                errstr = "{}: Mismatch {} and {}".format(test, eStr, s)
+                self.FailedTests.append(errstr)
+                raise Exception(errstr)
```
qiluo-msft, line 240: >Exception [](http://example.com/codeflow?start=22&length=9)

Should we raise more specific exception type, like ValueError?

src/sonic-yang-models/yang-models/sonic-bgp-allowed-prefix.yang
```diff
@@ -96,11 +96,13 @@ module sonic-bgp-allowed-prefix {
 
                 leaf-list prefixes_v4 {
                     type bgp-allowed-ipv4-prefix;
+                    ordered-by user;
```
qiluo-msft, line 99: >ordered-by user; [](http://example.com/codeflow?start=20&length=16)

This PR seems include too many topics than the title mentioned. For example "ordered-by" is not explained. If it is not related to libyang v3 upgrade, we may consider a separate PR for it.

src/sonic-yang-models/yang-templates/sonic-policer.yang.j2
```diff
@@ -66,7 +66,6 @@ module sonic-policer {
 						error-message "cbs must be greater than or equal to cir";
 					}
 					type uint64;
-					default 0;
```
qiluo-msft, line 69: >default [](http://example.com/codeflow?start=5&length=7)

@dgsudharsan, Could you review this change?

src/sonic-yang-models/yang-models/sonic-bgp-allowed-prefix.yang
bhouse-nexthop: Please read the individual commit messages :)

```
yang: leaf-list needs to use ordered-by user
RFC 7950 defines an ordered-by clause when a list needs to maintain
its order:
https://datatracker.ietf.org/doc/html/rfc7950#section-7.7.7

The default ordering is "system", which means the user specified
order doesn't need to be preserved.  Prior to this libyang commit
there was no difference between system and user ordering:
https://github.com/CESNET/libyang/commit/6cf1d16873f7a795f37658af58a8eb086a95e6d5

Since SONiC depends on list ordering being maintained, either explicitly
due to the order actually mattering, or implicitly due to the reverse
translation performed during config valiation, we need to tag all
leaf-list references as `ordered-by user`.
```

src/sonic-yang-models/tests/yang_model_tests/test_yang_model.py
bhouse-nexthop: Since this is only tests and NOT a library function, it honestly doesn't matter.  I should also note this isn't a logic change here, the original code did this.  

The only logic change I made was to queue the failures in the list so they could be dumped back out later as I found it really hard to find the failed test cases!

src/sonic-yang-models/yang-templates/sonic-policer.yang.j2
bhouse-nexthop: fyi, there is a commit message that explains this as well:
```
yang: policer should not define cbs default value of 0
It looks like there is a behavior difference between libyang1 and
libyang3 where a default value of 0 was actually treated the same
as not set, so now the `POLICER_PIR_WITHOUT_CIR` test case is
failing with:

ERROR    YANG-TEST:test_yang_model.py:303 POLICER_PIR_WITHOUT_CIR: Mismatch ["pir can't be configured without cir."] and failed to parse data tree: cbs can't be configured without cir.: Data path: /sonic-policer:sonic-policer/POLICER/POLICER_LIST[name='e3policer']/cbs

Get rid of this bogus default.
```

src/sonic-yang-models/yang-models/sonic-mgmt_interface.yang
```diff
@@ -55,7 +55,7 @@ module sonic-mgmt_interface {
                         type stypes:sonic-ip-prefix;
                         type inet:ip-address;
                     }
-
+                    ordered-by user;
                     description
                         "This configuration allows addtional routes to be added to default VRF table
```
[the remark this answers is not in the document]
bhouse-nexthop: I could make this change, but its not a line I touched in this PR and isn't relevant to the effort itself.

src/sonic-yang-models/tests/yang_model_tests/test_yang_model.py
```diff
@@ -254,7 +270,9 @@ def runVlanSpecialTest(self, test):
                 log.debug(jInput)
                 s = self.loadConfigData(json.dumps(jInput))
                 if s!="":
-                    raise Exception("{} in not empty".format(s))
+                    errstr="{}[{}]: {} in not empty".format(test,i,s)
```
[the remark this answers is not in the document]
bhouse-nexthop: I'm failing to see how is undefined.  I see this within a `for i in range (2,4095):` block

bhouse-nexthop:
@dgsudharsan @qiluo-msft any chance we can get this merged?  Once that's done I can rework step #3 of this process that might take a bit more time due to significant changes made there.

src/sonic-yang-models/yang-templates/sonic-policer.yang.j2
dgsudharsan: In such case can we remove the default from cir as well. 
On a note, considering default value of 0 as not set seems to be a wrong interpretation and there would be cases where we may need to enforce 0.

bhouse-nexthop: done

bradh352:
@lguohan @qiluo-msft any chance this can be merged before yet another merge conflict or otherwise failing test is introduced?  This is a high impact area that constantly causes this PR to need to be rebased.  The remaining PRs in this series don't have that same issue.

lguohan (approved):

lguohan:
@bradh352 , overall the pr looks good. in fact i'd like to merge pr as it is, meaning not squash merge, but use individual pr as it is. however, i found the last pr description is not good. is it possible to rebase the last pr (the rebase pr) into previous six prs. 

image

lguohan (dismissed):

src/sonic-yang-models/tests/yang_model_pytests/test_smart_switch.py
```diff
@@ -381,9 +381,9 @@ def test_dpu_orchagent_zmq_port(self, yang_model, orchagent_zmq_port, error_mess
 
     @pytest.mark.parametrize(
         "dpu_id, swbus_port, error_message", [
-            (0, 23606, None),
-            (1, 23607, None),
-            (7, 23613, None)]
+            ("0", 23606, None),
+            ("1", 23607, None),
+            ("7", 23613, None)]
```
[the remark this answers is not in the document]
bradh352: The yang model defines this as strings, but looks like it restricts it to numeric values.  So the yang model very well may be wrong, but I'm not modifying the yang model here, just making sure the tests pass with libyang3.

src/sonic-yang-models/tests/yang_model_tests/test_yang_model.py
```diff
@@ -180,26 +179,40 @@ def logStartTest(self, desc):
...
+                        log.info("Verify path value {} is {} in {}".format(verify['key'], verify['value'], data))
+                        assert (verify['key'] in data)
                         assert (data[verify['key']] == verify['value'])
                         s = 'verified'
         except Exception as e:
-            s = str(e)
-            log.info(s)
+            printExceptionDetails()
+
+        node.free()
```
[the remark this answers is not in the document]
bradh352: This suggestion is wrong since if the exception is caught the function returns so will never hit this.

src/sonic-yang-models/tests/yang_model_tests/test_yang_model.py
```diff
@@ -284,6 +302,17 @@ def test_run_tests(self):
...
+        if len(self.FailedTests):
+            print("{} Failures:".format(len(self.FailedTests)))
+            log.error("{} Failures:".format(len(self.FailedTests)))
+            for x in self.FailedTests:
+                print(x)
+                log.error(x)
+
+        if self.ctx:
+            self.ctx.destroy()
+            self.ctx = None
```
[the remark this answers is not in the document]
bradh352: a single instance does not make a pattern

bradh352:
@lguohan 
> @bradh352 , overall the pr looks good. in fact i'd like to merge pr as it is, meaning not squash merge, but use individual pr as it is. however, i found the last pr description is not good. is it possible to rebase the last pr (the rebase pr) into previous six prs.
> 

done

src/sonic-yang-models/tests/yang_model_tests/test_yang_model.py
```diff
@@ -1,11 +1,12 @@
 # This script is used to
 
-import yang as ly
+import libyang as ly
 import logging
 import argparse
 import sys
 import ijson
 import json
+import pytest
```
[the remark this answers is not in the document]
bradh352: done

PR #21771, opened 2025-02-18
[chassis-packet] Change the Orchagent redis pop batch size to 128 to handle link notification faster.

abdosi:
What I did:
Change the Orchagent redis pop batch size to 128 to handle link notification faster.

Why I did:
As part of Ixia BGP convergence test https://github.com/sonic-net/sonic-mgmt/blob/master/tests/snappi_tests/multidut/bgp/test_bgp_outbound_uplink_multi_po_flap.py we found that because of SAI programming slowness which is around 1500 Routes/sec. [Takes about approx 40sec+/- to program 60K routes across multiple iteration]
Because of above slowness even if we have Link Notification available Orchagent will not process it immediately as current OA will process 1024 Entries (Route entries in our case) before it can pick Link Notification for processing. Now 1K entries can take about 2 sec+/-  and if link notification are little spread out [not back 2 back] we can have batch of 1K entries which accumulate SAI delay of 2 sec.
 
 To optimize Link processing and give more chance to OA to pick Link Notification we reduce OA processing to 128 entries and this helped to reduce convergence time to about overall 2 sec.
 
Changing OA processing from 1K to 128 entries  does not have any impact of Route Programming as SAI slowness seems be tied with sequential processing at 1500 Routes per/sec.   However this is helpful in processing Link Notification quicker. For our test it reduce Convergence time from about 12-15sec to 2 sec.

How I verify:

Ran above ixia test in multiple iteration with both 1024 and 128 pop batch size and compared the performance.


abdosi:
@anamehra for viz.

abdosi:
@yejianquan for viz.

anamehra (approved):

liuh-80 (approved):

abdosi:
@rlhui : Can you help merge this.

rlhui (approved):

PR #21772, opened 2025-02-18
Bump up /dev/shm size for Broadcom platforms to 512MB by default

saiarcot895:
#### Why I did it

In the case of ASIC detection failures on Broadcom (or if the ASIC couldn't be detected in time), the `/dev/shm` partition in the syncd container will be only 64MB, which might cause issues if syncd/Broadcom SAI library needs more space than that.

##### Work item tracking

#### How I did it

Since using a larger `/dev/shm` on its own doesn't cause any issues, bump up the default to 512MB. This should be enough for most platforms.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/files/build_templates/docker_image_ctl.j2
+++ b/files/build_templates/docker_image_ctl.j2
@@ -662,7 +662,7 @@ start() {
 {%- endif %}
 {%- if sonic_asic_platform == "broadcom" %}
 {%- if docker_container_name == "syncd" %}
-        --shm-size=${SYNCD_SHM_SIZE:-64m} \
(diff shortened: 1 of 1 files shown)


saiarcot895:
> @saiarcot895 we have few platforms with very less memory. have we studied the impact on them?

Increasing the `/dev/shm` on its own doesn't automatically take memory away from the system. Memory will be taken only if something is using it.

Gfrom2016 (approved):
LGTM

PR #21775, opened 2025-02-18
[chassis-packet] Allow Fallback Route to get programmed on Downstream LC

abdosi:
What/Why I did:
Initially I made the change that Fallback Route [routes from AH on Upstream LC] not to get programmed [mark them as deny]  on Downstream LC's as reasoning was that on Downstream LC's can forward them based on Default-route. However that assumption is correct for example we have topology like this:

RH is connected to ASIC0 of Upstream LC
AH is connected to ASIC1 of Upstream LC

Downstream LC will learn route from RH [including default-route] and will only forward to ASIC0 of upstream LC .  Above assumption was fine if we always have AH and RH connected to same ASIC of Upstream LC.

How I verify:
UT updated
Manual Verification.


abdosi:
@arlakshm : can you help to approve this.

arlakshm (approved):

PR #21827, opened 2025-02-21
[Bgpcfgd] Add dependency callback logic in SRv6 Manager to handle out-of-order table processing

### LINKED ISSUE #21826: SRv6: ERR bgp#bgpcfgd: Found a SRv6 SID config entry with a locator that does not exist
**Description**
Adding a static SRv6 configuration to _config_db.json_ and doing `config reload` leads to an error in bgpcfgd, and the configuration is not applied to the appl_db.

**Steps to reproduce**
* _config_db.json_:
```
"SRV6_MY_LOCATORS": {
    "loc1": {
        "prefix": "3000:1:10::"
    }
},
"SRV6_MY_SIDS": {
    "loc1|3000:1:10::/48": {
        "action": "uN",
        "decap_dscp_mode": "pipe"
    }
}
```
* `config reload`

Error in syslog:
```
ERR bgp#bgpcfgd: Found a SRv6 SID config entry with a locator that does not exist: loc1|3000:1:10::/48 | {'action': 'uN', 'decap_dscp_mode': 'pipe'}
```

BYGX-wcr:
#### Why I did it

Previously, if the SRv6 manager received a config update that added a SRV6_MY_SID entry before its corresponding LOCATOR entry added, the SRv6 manager would not process the SRV6_MY_SID entry and print an ERR syslog. However, during testing, NVIDIA team found that it is possible to have such a scenario in normal operations, especially during config reload. So, we decide to add a dependency-based caching mechanism to handle out-of-order table processing.

This is the first use case where we will dynamically add/remove the dependency, so we also define a new unsubscribe function for the directory class. The reason why we need to remove the dependency in managers_srv6 is because the on_deps_change function will only process the cache when all dependencies are satisfied. If a locator is removed and the corresponding dependency is not removed, then the on_deps_change will never get to the satisfied condition.

Fix #21826 (Issue)

##### Work item tracking

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305
- [x] 202412

- [x] 202412

### DIFF
--- a/src/sonic-bgpcfgd/bgpcfgd/directory.py
+++ b/src/sonic-bgpcfgd/bgpcfgd/directory.py
@@ -157,3 +157,10 @@ class Directory(object):
         for db, table, path in deps:
             slot = self.get_slot_name(db, table)
             self.notify[slot][path].append(handler)
+
+    def unsubscribe(self, deps):
+        for db, table, path in deps:
+            slot = self.get_slot_name(db, table)
+            if slot in self.notify:
+                if path in self.notify[slot]:
+                    del self.notify[slot][path]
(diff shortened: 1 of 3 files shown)


Yakiv-Huryk:
@BYGX-wcr is it possible to add a test for it?

BYGX-wcr:
> @BYGX-wcr is it possible to add a test for it?

Added. And it has passed in the build checker.

Yakiv-Huryk (approved):

BYGX-wcr:
@abdosi , can you approve this one?

abdosi (approved):

PR #21848, opened 2025-02-25
Yang model additions for PFC Stat History

peterbailey-arista:
New pfc_stat_history leaf attribute for toggling history estimation per-interface in CONFIG_DB

#### Why I did it
PFC Historical Statistics feature https://github.com/sonic-net/SONiC/issues/1904

##### Work item tracking
https://github.com/sonic-net/sonic-buildimage/issues/21847

#### How to verify it
- Run the test cases in tests/ and ensure they pass.
- Deploy the changes to a SONiC device and verify the configuration and monitoring functionality using CLI commands.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [x] 202411

Addition of PFC_STAT_HISTORY table and Flex Counter group.

### DIFF
--- a/src/sonic-yang-models/tests/yang_model_tests/tests_config/pfc.json
+++ b/src/sonic-yang-models/tests/yang_model_tests/tests_config/pfc.json
@@ -329,5 +329,51 @@
                 ]
             }
         }
+    },
+    "PFC_WDOG_WITH_VALID_PFC_STAT_HISTORY": {
+        "sonic-port:sonic-port": {
+            "sonic-port:PORT": {
+                "PORT_LIST": [
+                {
+                    "admin_status": "up",
+                    "alias": "eth0",
+                    "description": "Ethernet4",
+                    "lanes": "65",
+                    "mtu": "9000",
+                    "name": "Ethernet4",
+                    "tpid": "0x8100",
+                    "speed": "25000"
+                }
+                ]
+            }
+        },
+        "sonic-pfcwd:sonic-pfcwd": {
+            "sonic-pfcwd:PFC_WD": {
+                "PFC_WD_LIST": [
+                    {
+                        "ifname": "Ethernet4",
+                        "action": "drop",
+                        "detection_time": 400,
+                        "restoration_time": 800,
+                        "pfc_stat_history": "enable"
+                    }
+                ]
+            }
+        }
+    },
+    "PFC_WDOG_WITH_INVALID_PFC_STAT_HISTORY": {
+        "sonic-pfcwd:sonic-pfcwd": {
+            "sonic-pfcwd:PFC_WD": {
+                "PFC_WD_LIST": [
+                    {
+                        "ifname": "Ethernet4",
+                        "action": "drop",
+                        "detection_time": 400,
+                        "restoration_time": 800,
+                        "pfc_stat_history": "down"
+                    }
+                ]
+            }
+        }
     }
 }
--- a/src/sonic-yang-models/tests/files/sample_config_db.json
+++ b/src/sonic-yang-models/tests/files/sample_config_db.json
@@ -1612,7 +1612,8 @@
             "Ethernet9": {
                "action": "drop",
                "detection_time": "100",
-               "restoration_time": "400"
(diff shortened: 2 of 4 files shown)


src/sonic-yang-models/yang-models/sonic-pfc-stat-history.yang
```diff
@@ -0,0 +1,41 @@
...
+                leaf port {
+                    type leafref {
+                        path "/port:sonic-port/port:PORT/port:PORT_LIST/port:name";
+                    }
+                    description "Interface name.";
+                }
+
+                leaf status {
+                    type string {
+                        pattern "enabled";
```
ganglyu, line 35: Why is it necessary to have a fixed status?

peterbailey-arista: The reason was because anything could be inserted as the value, the important data is the key holding the EthernetXYZ name. It's efficient to have the presence of the port in the table determine if it is being monitored by PFC Stat History. I couldn't find another valid way to define the Yang model

At the moment I am looking into whether the feature can instead be incorporated into the PFC Watchdog based on feedback from yesterday's community HLD review

peterbailey-arista: The Yang model has changed to reflect the new implementation and the fixed status is no longer used so I will resolve this conversation.

src/sonic-yang-models/tests/yang_model_tests/tests_config/pfc.json
```diff
@@ -329,5 +329,51 @@
...
+                ]
+            }
+        }
+    },
+    "PFC_WDOG_WITH_INVALID_PFC_STAT_HISTORY": {
+        "sonic-pfcwd:sonic-pfcwd": {
+            "sonic-pfcwd:PFC_WD": {
+                "PFC_WD_LIST": [
+                    {
+                        "ifname": "Ethernet4",
```
wen587, line 369: Why doesn't it catch error for missing leafref in ifname: Ethernet4?

peterbailey-arista: While I was simply following the example of the other tests, I believe it is because Test_yang_models::runExceptionTest only verifies that each eStr defined in the test case is found in the exceptions from parsing the input.
So from my understanding of the code, it's likely that there would be a missing leafref and possibly any number of other exceptions but the test case passes so long as all exceptions defined in eStr were found at least once. 

In my case I am only verifying that "pfc_stat_history must be either enable or disable" is shown and not any other exceptions.

peterbailey-arista:
@ganglyu or @wen587 are you able to review this PR or suggest reviewers I can ping for this and for the other related PRs for this feature? They are listed at https://github.com/sonic-net/SONiC/issues/1904#issuecomment-2680123338

Let me know, thanks!

vmittal-msft:
@peterbailey-arista As per my understanding, this is verifying port level configuration. How about global config validation ?

peterbailey-arista:
> @peterbailey-arista As per my understanding, this is verifying port level configuration. How about global config validation ?

@vmittal-msft The feature configuration is all at the port level, you enable history on ports individually in the same way the PFCWD is started on ports individually. Hopefully I've understood your question but there is no global configuration to do - global configurations are handled through the PFCWD. For example, polling time is simply the PFCWD's polling time

vmittal-msft (approved):

vmittal-msft:
@rlhui @yxieca  please help merge.

peterbailey-arista:
@vmittal-msft Are we able to follow up on merging this? Thanks!

vmittal-msft:
@yxieca @rlhui please help merge.

yejianquan:
Hi @vmittal-msft , I can see the feature https://github.com/sonic-net/SONiC/issues/1904 is moved to 202511 release.
Hence delete the request label for 202505 release, 
please let me know this is needed for 202505

PR #21852, opened 2025-02-25
Use pzstd to compress the Docker in SWI slim images

lipxu:
#### Why I did it
Same with https://github.com/sonic-net/sonic-buildimage/pull/21824

Enabled the docker inram feature for slim image.
It would extract the docker image to ram during the boot, so it would take extra times during boot.

##### Work item tracking
31323281

#### How I did it
Use pzstd which is more efficient tool to compress and decompress the docker file to reduce the boot time.
Currently, we do not modify the "FILESYSTEM_DOCKERFS=dockerfs.tar.gz" in onie-image.conf, so for slim image, we still use dockerfs.tar.gz as file name but actually with zstd compressed. 
We tried to find out a way to adjust the file name in onie-image.conf, but it seems not easy to do that.
So we use the file cmd to determine the compressing type in union-mount.j2, then use the related cmd to extract the docker file.
Plan to support zstd for all types of images in the future to unify the docker file image .

#### How to verify it
Boot swi slim image and normal image, boot mellanox bin image, all boot successfully.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]


saiarcot895 (approved):

StormLiangMS (approved):
LGTM

andywongarista:
@lipxu @StormLiangMS shouldn't this also be cherry picked to 202411?

lipxu:
> @lipxu @StormLiangMS shouldn't this also be cherry picked to 202411?

Thanks for reminder, @andywongarista , should be in 202411, let me add the label. And I think we should consider to support both zstd and gz for all images in the further.

PR #459, opened 2017-03-30
[sonic-slave]: Ignore old certificates in apt.

marian-pritsak:
Signed-off-by: marian-pritsak <marianp@mellanox.com>

PR #21863, opened 2025-02-26
Add BMP to init config in default disabled state

vivekverma-arista:
#### Why I did it

The new feature was introduced in the this set of pull requests: https://github.com/sonic-net/SONiC/pull/1621

This feature runs inside a new docker container. In order to enable it through CLI `sudo config feature state` it should be a part of base config or `init_cfg.json` at `/etc/sonic`.

Some sonic-mgmt tests were added in https://github.com/sonic-net/sonic-mgmt/pull/15749 which are currently failing in full test suite runs. We need this so that we can enable BMP while running them.

##### Work item tracking
[#459](https://github.com/aristanetworks/sonic-qual.msft/issues/459])

#### How I did it

Update `files/build_templates/init_cfg.json.j2`

#### How to verify it
Ran all BMP sonic-mgmt tests with changes in this PR as well as https://github.com/sonic-net/sonic-mgmt/pull/17179.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305
- [x] 202411

- [ ] 
- [ ]

### DIFF
--- a/files/build_templates/init_cfg.json.j2
+++ b/files/build_templates/init_cfg.json.j2
@@ -67,7 +67,8 @@
                    ("pmon", "enabled", "{% if 'type' in DEVICE_METADATA['localhost'] and DEVICE_METADATA['localhost']['type'] == 'SpineRouter' %}False{% else %}True{% endif %}", "enabled"),
                    ("snmp", "enabled", true, "enabled"),
                    ("swss", "enabled", false, "enabled"),
-                   ("syncd", "enabled", false, "enabled")] %}
+                   ("syncd", "enabled", false, "enabled"),
+                   ("bmp", "disabled", "False", "enabled")] %}
 {%- if include_router_advertiser == "y" %}{% do features.append(("radv", "enabled", false, "enabled")) %}{% endif %}
 {%- if include_teamd == "y" %}{% do features.append(("teamd", "{% if not DEVICE_RUNTIME_METADATA['ETHERNET_PORTS_PRESENT'] %}disabled{% else %}enabled{% endif %}", false, "enabled")) %}{% endif %}
 {% do features.append(("dhcp_relay", "{% if not (DEVICE_METADATA is defined and DEVICE_METADATA['localhost'] is defined and DEVICE_METADATA['localhost']['type'] is defined and DEVICE_METADATA['localhost']['type'] is not in ['ToRRouter', 'EPMS', 'MgmtTsToR', 'MgmtToRRouter', 'BmcMgmtToRRouter']) %}enabled{% else %}disabled{% endif %}", false, "enabled")) %}


files/build_templates/init_cfg.json.j2
```diff
@@ -67,7 +67,8 @@
                    ("pmon", "enabled", "{% if 'type' in DEVICE_METADATA['localhost'] and DEVICE_METADATA['localhost']['type'] == 'SpineRouter' %}False{% else %}True{% endif %}", "enabled"),
                    ("snmp", "enabled", true, "enabled"),
                    ("swss", "enabled", false, "enabled"),
-                   ("syncd", "enabled", false, "enabled")] %}
+                   ("syncd", "enabled", false, "enabled"),
+                   ("bmp", "disabled", "False", "enabled")] %}
```
FengPan-Frank, line 71: Shouldn't we move to golden config, instead of init_cfg.json change in the future? @qiluo-msft

vivekverma-arista: As far as I understand, golden config is not an alternative to init_cfg. https://github.com/sonic-net/sonic-mgmt/blob/master/ansible/library/generate_golden_config_db.py

It possibly just keeps track of last known sane config in sonic-mgmt.

I understand ini_cfg.json.j2 is protected and we want to exercise discretion while updating it but I believe this change is necessary. In case I am missing something please point me to the appropriate file where I should put this.

qiluo-msft:
Could you explain details of "Some sonic-mgmt tests were added in https://github.com/sonic-net/sonic-mgmt/pull/15749 which are currently failing in full test suite runs" ?

files/build_templates/init_cfg.json.j2
wen587: Hi @vivekverma-arista , why is it necessary? Cannot we use deploy-mg to deploy bmp initial config? I saw it is already being set to enable when bmp is supported:
https://github.com/sonic-net/sonic-mgmt/blob/master/ansible/library/generate_golden_config_db.py#L421-L429

vivekverma-arista: deploy-mg runs `sudo config load_minigraph -y` which itself relies on `/etc/sonic/init_cfg.json` to generate `FEATURE` config. We don't put any `FEATURE` config in minigraph.xml. (All minigraph templates: https://github.com/sonic-net/sonic-mgmt/tree/master/ansible/templates)

```
admin@gd333:/etc/sonic$ sudo config load_minigraph -y                                                                                            
Acquired lock on /etc/sonic/reload.lock
Stopping SONiC target ...
Running command: /usr/local/bin/sonic-cfggen -H -m -j /etc/sonic/init_cfg.json --write-to-db
Running command: config qos reload --no-dynamic-buffer --no-delay
Running command: /usr/local/bin/sonic-cfggen -d -t /usr/share/sonic/device/x86_64-arista_7260cx3_64/Arista-7260CX3-D108C8/buffers.json.j2,/tmp/cfg_buffer.json -t /usr/share/sonic/device/x86_64-arista_7260cx3_64/Arista-7260CX3-D108C8/qos.json.j2,/tmp/cfg_qos.json -y /etc/sonic/sonic_version.yml
Running command: /usr/local/bin/sonic-cfggen -j /tmp/cfg_buffer.json -j /tmp/cfg_qos.json --write-to-db
Running command: pfcwd start_default
Restarting SONiC target ...
Reinitializing monit daemon
Please note setting loaded from minigraph will be lost after system reboot. To preserve setting, run `config save`.
Released lock on /etc/sonic/reload.lock
```
Even if we somehow enable it through deploy-mg, it gets enabled by default. I don't think that is desired. 

But since you rightly pointed out that the support is there in `golden_config_db.json` we can follow an approach similar to this: https://github.com/sonic-net/sonic-mgmt/pull/17356

What do you think?

wen587: Hi @vivekverma-arista , the deploy-mg will run `sudo config load_minigraph --override-config -y` at first, if the osversion doesn't support this command, it will fall back to `sudo config load_minigraph -y`: https://github.com/sonic-net/sonic-mgmt/blob/master/ansible/config_sonic_basedon_testbed.yml#L628-L636

I had checked with @FengPan-Frank, and the decision is to enable bmp by default. In example of https://github.com/sonic-net/sonic-mgmt/blob/master/ansible/library/generate_golden_config_db.py#L421-L429, it will by default add bmp to FEATURE and enable it as long as the os version supports `--override-config`, and `--override-config` should be supported since 2022+.

wen587: Example log in our test env:
```
TASK [Generate golden_config_db.json] ******************************************
Monday 17 March 2025  07:50:21 +0000 (0:00:00.045)       0:00:13.992 ********** 
ok: [vlab-01]

TASK [execute cli "config load_minigraph --override_config -y" to apply new minigraph] ***
Monday 17 March 2025  07:50:22 +0000 (0:00:00.494)       0:00:14.487 ********** 
changed: [vlab-01]

TASK [execute cli "config load_minigraph -y" to apply new minigraph] ***********
Monday 17 March 2025  07:50:53 +0000 (0:00:31.360)       0:00:45.847 ********** 
skipping: [vlab-01]
```
You can see `config load_minigraph -y` is skipped since `config load_minigraph --override_config -y` is already executed.

vivekverma-arista: Okay I tested this and I can see bmp enabled by default after deploy-mg. Probably at the time of raising this change I didn't have golden db config change in my repo.

PR #21865, opened 2025-02-26
Add link requirement per sku before enabling traffic on a fap

jfeng-arista:
With the current fap initialization sequence, multiple faps in a modular system could come up with  stand alone mode. This is not a recommended configuration.  It might mess up the global clock and might ends up with traffic issue.  
https://github.com/sonic-net/sonic-buildimage/issues/21866

This change adds a required minimum number of links have to be ready before enabling traffic on a fap, which based on the guidance should be able to guarantee the faps in a modular system can form a proper topology.

We manually tested image on a modular system and the topology after boot up looks correct, and there is no more than 1 fap come up with stand alone mode.  We also tested traffic on our system setup and all looks good so far. 

**This change also need to go to 202405.** 

##### Work item tracking

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/device/arista/x86_64-arista_7800r3_48cq2_lc/Arista-7800R3-48CQ2-C48/jr2-a7280cr3-32d4-40x100G.config.bcm
+++ b/device/arista/x86_64-arista_7800r3_48cq2_lc/Arista-7800R3-48CQ2-C48/jr2-a7280cr3-32d4-40x100G.config.bcm
@@ -863,3 +863,5 @@ dma_desc_aggregator_enable_specific_MDB_FEC.BCM8869X=1
 sai_pfc_dlr_init_capability=0 
 sai_default_cpu_tx_tc=7
 sai_disable_srcmacqedstmac_ctrl=1
+appl_param_active_links_thr_high=58
+appl_param_active_links_thr_low=1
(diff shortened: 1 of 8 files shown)


jfeng-arista:
@kenneth-arista @arista-nwolfe @arlakshm @vmittal-msft

vmittal-msft (commented):
@jfeng-arista  Can you please share the logic behind picking these thresholds ? I believe we always assume certain number of fabric cards to be present in the system for this to work ? Also, any reason its different for 100g and 400g.

jfeng-arista:
> @jfeng-arista Can you please share the logic behind picking these thresholds ? I believe we always assume certain number of fabric cards to be present in the system for this to work ? Also, any reason its different for 100g and 400g.

The different is not come from 100g or 400g nif port. they are depending on totally number of fabric links that expected to be up .  

The number needs to be greater than half of the links need to up.  I also rounded up for the number, e.g  Arista-7800R3A-36D2-C36 , depends on the type of fabric, each fap designed to use 144 links,  180 links  or 179 links. so I put the number 91 as more than half of number in 80 case, and  that should be even safe for all other cases.

rlhui:
> We need this for 202405 and msft-202405

wondering why both branches are needed, is it ok if only msft-202405 takes it?

jfeng-arista:
> > We need this for 202405 and msft-202405
> 
> wondering why both branches are needed, is it ok if only msft-202405 takes it?

Hello Rita, 
Technically, all images/releases need this change for FAPs to initialize properly with a recommended config. I'm not familiar with the release merging process, so I'll rely on you or others to determine which releases can incorporate this change so the others can be synced accordingly.

kenneth-arista:
> > We need this for 202405 and msft-202405
> 
> wondering why both branches are needed, is it ok if only msft-202405 takes it?

If 202405 will not be deployed on chassis, then just `msft-202405` is fine

arlakshm (approved):

arlakshm:
This PR manually ported to msft-202405 branch in this PR https://github.com/Azure/sonic-buildimage-msft/pull/729

vmittal-msft (approved):

jfeng-arista:
the failures do not look like related to the change, I kick off another build

vmittal-msft:
@rlhui @yxieca please help merge.

PR #21897, opened 2025-03-01
Add mgmt_type into device metadata YANG model.

r12f:
#### Why I did it

Enhance yang model for networking-metadata.

Adding a new field in device metadata YANG model to distinguish the management type of the device. Depends on the users, this field can be used to tell the purpose of deployment or who/how the device is being managed.

##### Work item tracking

#### How I did it

This change adds a new field called mgmt_type for this purpose.

#### How to verify it

Unit test.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305
- [X] 202405
- [X] 202411

- [ ] 
- [ ]

### DIFF
--- a/src/sonic-yang-models/yang-models/sonic-device_metadata.yang
+++ b/src/sonic-yang-models/yang-models/sonic-device_metadata.yang
@@ -159,6 +159,7 @@ module sonic-device_metadata {
 
                 leaf mgmt_type {
                     type string;
+                    description "Indicates the management type of this device.";
                 }
 
                 leaf cluster {
--- a/src/sonic-yang-models/yang-models/sonic-device_metadata.yang
+++ b/src/sonic-yang-models/yang-models/sonic-device_metadata.yang
@@ -157,6 +157,10 @@ module sonic-device_metadata {
                     type string;
                 }
 
+                leaf mgmt_type {
+                    type string;
+                }
+
                 leaf cluster {
                     type string;
                     description "The switch is a member of this cluster.";
--- a/src/sonic-yang-models/tests/yang_model_tests/tests/device_metadata.json
+++ b/src/sonic-yang-models/tests/yang_model_tests/tests/device_metadata.json
@@ -94,6 +94,9 @@
     "DEVICE_METADATA_RESOURCE_TYPE_CONFIG": {
         "desc": "Verifying resource type configuration."
     },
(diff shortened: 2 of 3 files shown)


src/sonic-yang-models/yang-models/sonic-device_metadata.yang
```diff
@@ -157,6 +157,10 @@ module sonic-device_metadata {
                     type string;
                 }
 
+                leaf mgmt_type {
```
qiluo-msft, line 160: What does it mean? It is explained in HLD?

r12f: I can add a description field, if this helps.

r12f: done

src/sonic-yang-models/yang-models/sonic-device_metadata.yang
```diff
@@ -157,6 +157,11 @@ module sonic-device_metadata {
...
+                    type string;
+                    description "Indicates the management type of this device.";
```
ganglyu, line 162: Could you please provide example?

r12f: sure thing! which file should I update?

ganglyu (approved):
lgtm

src/sonic-yang-models/yang-models/sonic-device_metadata.yang
```diff
@@ -157,6 +157,11 @@ module sonic-device_metadata {
                     type string;
                 }
 
+                leaf mgmt_type {
```
xincunli-sonic, line 160: Would this field be enum, or it is arbitrary value? Also, is there any restrcition? like length bound, lastly, is it mandatory?

r12f: it is just a user-specified string, no pattern, not enum. the way to use it is in the template evaluation.

this field is optional, not mandatory.

xincunli-sonic (approved):
LGTM

kperumalbfn (approved):
LGTM

PR #21906, opened 2025-03-04
Add ACL yang changes for INNER SRC MAC REWRITE 

kavyavanibedida:
#### Why I did it
Yang changes for inner source mac rewrite feature in Smart switch(a new action and inner_src_ip and vni)

##### Work item tracking
- Microsoft ADO **(number only)**:https://msazure.visualstudio.com/One/_workitems/edit/31634070

#### How I did it
Added ACL table INNERSRCMACREWRITE, action type INNER_SRC_MAC_REWRITE_ACTION for this ACL type, VNI_ID
 
#### How to verify it
yang model tests
addition to sample config_db.json

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/src/sonic-yang-models/tests/files/sample_config_db.json
+++ b/src/sonic-yang-models/tests/files/sample_config_db.json
@@ -345,14 +345,8 @@
             "SRC-MAC-REWRITE-TABLE|RULE0": {
                 "PRIORITY": "999960",
                 "INNER_SRC_MAC_REWRITE_ACTION": "60:45:BD:08:19:68",                     
-                "INNER_SRC_IP": "10.176.0.0/15",
+                "INNER_SRC_IP": "1.1.1.1/32",
                 "VNI": "45"
-            },
-            "SRC-MAC-REWRITE-TABLE|RULE1": {
-                "PRIORITY": "999970",
-                "INNER_SRC_MAC_REWRITE_ACTION": "60:46:BD:08:19:68",                     
-                "INNER_SRC_IP": "10.177.0.0/15",
-                "VNI": "48"
             }
         },
         "SWITCH_HASH": {
--- a/src/sonic-yang-models/tests/files/sample_config_db.json
+++ b/src/sonic-yang-models/tests/files/sample_config_db.json
@@ -341,6 +341,18 @@
                 "PACKET_ACTION": "FORWARD",
                 "TUNNEL_TERM": "true",
                 "DST_IP": "1.1.1.1/32"
+            },
+            "SRC-MAC-REWRITE-TABLE|RULE0": {
+                "PRIORITY": "999960",
+                "INNER_SRC_MAC_REWRITE_ACTION": "60:45:BD:08:19:68",                     
+                "INNER_SRC_IP": "10.176.0.0/15",
+                "VNI": "45"
+            },
+            "SRC-MAC-REWRITE-TABLE|RULE1": {
+                "PRIORITY": "999970",
+                "INNER_SRC_MAC_REWRITE_ACTION": "60:46:BD:08:19:68",                     
+                "INNER_SRC_IP": "10.177.0.0/15",
+                "VNI": "48"
             }
         },
         "SWITCH_HASH": {
@@ -1104,9 +1116,42 @@
                     "Ethernet0",
                     "PortChannel0003"
                 ]
+            },
+            "SRC-MAC-REWRITE-TABLE": {
+                "ports": [
+                    "Ethernet14",
+                    "Ethernet15",
+                    "Ethernet23",
+                    "Ethernet30",
+                    "Ethernet31",
+                    "Ethernet18",
+                    "Ethernet19",
+                    "Ethernet25",
+                    "Ethernet24",
+                    "PortChannel0003"
+                ],
+                "stage": "EGRESS",
+                "type": "INNER_SRC_MAC_REWRITE_TABLE_TYPE"
             }
-
         },
+        "ACL_TABLE_TYPE": {
+            "ACL_TABLE_TYPE_LIST": [
+                {
+                    "ACL_TABLE_TYPE_NAME": "INNER_SRC_MAC_REWRITE_TABLE_TYPE",
+                    "ACTIONS": [
+                        "INNER_SRC_MAC_REWRITE_ACTION"
+                    ],
+                    "MATCHES": [
+                        "INNER_SRC_IP",
+                        "VNI"
+                    ],
+                    "BIND_POINTS": [
+                        "PORT",
+                        "PORTCHANNEL"
+                    ]
+                }
+            ]
+        }, 
         "PBH_HASH_FIELD": {
             "inner_ip_proto": {
                 "hash_field": "INNER_IP_PROTOCOL",
--- a/src/sonic-yang-models/tests/files/sample_config_db.json
+++ b/src/sonic-yang-models/tests/files/sample_config_db.json
@@ -1133,24 +1133,6 @@
                 "stage": "EGRESS",
                 "type": "INNER_SRC_MAC_REWRITE_TABLE_TYPE"
             }
-        },
-        "ACL_TABLE_TYPE": {
-            "ACL_TABLE_TYPE_LIST": [
-                {
-                    "ACL_TABLE_TYPE_NAME": "INNER_SRC_MAC_REWRITE_TABLE_TYPE",
-                    "ACTIONS": [
-                        "INNER_SRC_MAC_REWRITE_ACTION"
-                    ],
-                    "MATCHES": [
-                        "INNER_SRC_IP",
-                        "VNI"
-                    ],
-                    "BIND_POINTS": [
-                        "PORT",
-                        "PORTCHANNEL"
-                    ]
-                }
-            ]
         }, 
         "PBH_HASH_FIELD": {
             "inner_ip_proto": {
@@ -2652,7 +2634,20 @@
                      "PORT",
                      "PORTCHANNEL"
                  ]
-             }
+             },
+             "INNER_SRC_MAC_REWRITE_TABLE_TYPE": {
+                "ACTIONS": [
+                    "INNER_SRC_MAC_REWRITE_ACTION"
+                ],
+                "MATCHES": [
+                    "INNER_SRC_IP",
+                    "VNI"
+                ],
+                "BIND_POINTS": [
+                    "PORT",
+                    "PORTCHANNEL"
+                ]
+            }
         },
 
         "SNMP": {
--- a/src/sonic-yang-models/tests/files/sample_config_db.json
+++ b/src/sonic-yang-models/tests/files/sample_config_db.json
@@ -343,8 +343,8 @@
                 "DST_IP": "1.1.1.1/32"
             },
             "SRC-MAC-REWRITE-TABLE|RULE0": {
-                "PRIORITY": "999960",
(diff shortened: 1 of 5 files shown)


src/sonic-yang-models/yang-templates/sonic-types.yang.j2
```diff
@@ -267,6 +268,12 @@ module sonic-types {
         }
     }
 
+    typedef mac-address {
```
wen587, line 271: https://github.com/sonic-net/sonic-buildimage/blob/a1e2c1f2d4be9f45efcaf585de9ab57b1145d1a7/src/sonic-yang-models/yang-models/sonic-neigh.yang#L53

Yang has mac-address defined .

src/sonic-yang-models/yang-templates/sonic-acl.yang.j2
```diff
@@ -255,6 +265,15 @@ module sonic-acl {
 					}
 				}
 
+				leaf VNI_ID {
```
ganglyu, line 268: Please add description

src/sonic-yang-models/yang-templates/sonic-acl.yang.j2
```diff
@@ -84,6 +84,13 @@ module sonic-acl {
 					}
 				}
 
+				leaf INNER_SRC_MAC_REWRITE_ACTION {
```
ganglyu, line 87: please add description

kavyavanibedida: added a description

src/sonic-yang-models/yang-templates/sonic-acl.yang.j2
kavyavanibedida: added a description

SoumyaMishra18 (approved):

src/sonic-yang-models/yang-templates/sonic-types.yang.j2
kavyavanibedida: Removed this part of the code

src/sonic-yang-models/yang-templates/sonic-types.yang.j2
```diff
@@ -95,6 +95,7 @@ module sonic-types {
             enum MIRRORV6;
             enum MIRROR_DSCP;
             enum CTRLPLANE;
+            enum INNERSRCMACREWRITE;
```
prsunny, line 98: Why call this as a specific table INNERSRCMACREWRITE? Why not simply L3INNER and can be used for future inner match cases? @bingwang-ms what do you think?

anish-n, line 98: why do we need this change? I think we can skip this change.

src/sonic-yang-models/yang-templates/sonic-acl.yang.j2
```diff
@@ -84,6 +88,14 @@ module sonic-acl {
...
+					description "Inner Source Mac-adress rewrite action mapped to mac address";
+					type yang:mac-address;
+				}
+
+				/* Validating 'INNER_SRC_MAC_REWRITE_ACTION' exist if ACL type is 'INNERSRCMACREWRITE' */
```
anish-n, line 96: don't think we need to mandate the naming of ACL_TABLE associated with ACL_RULE, as we are not defining a new pre-created ACL_TABLE_TYPE in SONIC

src/sonic-yang-models/yang-templates/sonic-acl.yang.j2
```diff
@@ -128,6 +145,9 @@ module sonic-acl {
 					}
 				}
 
+				/* Validating 'SRC_IP' exist if ACL type is 'INNERSRCMACREWRITE' */
```
anish-n, line 148: not sure we need this, INNER_SRC_IP is a generic match type, can be applicable to other use cases for acl_rule, we probably don't need to mandate/limit it to innersrcmacrewrite

src/sonic-yang-models/yang-templates/sonic-acl.yang.j2
```diff
@@ -255,6 +275,16 @@ module sonic-acl {
 					}
 				}
 
+				leaf VNI_ID {
```
anish-n, line 278: How about just VNI or TUNNEL_VNI as is used for other use cases instead of VNI_ID, since the I in VNI stands for ID.

src/sonic-yang-models/yang-templates/sonic-acl.yang.j2
```diff
@@ -255,6 +275,16 @@ module sonic-acl {
...
+					description "VXLAN Network Identifier";
+					type uint32 {
+						range "1..16777215";
+					}
+				}
+
+				/* Validating 'VNI_ID' exist if ACL type is 'INNERSRCMACREWRITE' */
```
anish-n, line 285: VNI match type is something other uses cases outside of innersrcmacrewrite could leverage as well, it is a generic match condition as defined in the SAI.

src/sonic-yang-models/yang-templates/sonic-acl.yang.j2
```diff
@@ -355,6 +385,9 @@ module sonic-acl {
 					default "INGRESS";
 				}
 
+				/* Validating stage is EGRESS exist if ACL type is 'INNERSRCMACREWRITE' */
```
anish-n, line 388: don't think we need to mandate it to egress, that would be hw platform and use case dependent

src/sonic-yang-models/yang-templates/sonic-acl.yang.j2
```diff
@@ -255,6 +275,16 @@ module sonic-acl {
 					}
 				}
 
+				leaf VNI_ID {
+					description "VXLAN Network Identifier";
+					type uint32 {
+						range "1..16777215";
```
anish-n, line 281: there is an existing type for vni lets use that stypes:vnid_type

anish-n (changes requested):

src/sonic-yang-models/yang-templates/sonic-types.yang.j2
kavyavanibedida: Removed the new type

src/sonic-yang-models/yang-templates/sonic-acl.yang.j2
kavyavanibedida: Removed mandate for this case

src/sonic-yang-models/yang-templates/sonic-acl.yang.j2
kavyavanibedida: Updated the leaf name

src/sonic-yang-models/yang-templates/sonic-acl.yang.j2
kavyavanibedida: Removed mandate

src/sonic-yang-models/yang-templates/sonic-acl.yang.j2
kavyavanibedida: removed mandate

src/sonic-yang-models/yang-templates/sonic-acl.yang.j2
kavyavanibedida: Updated to existing vnid_type

src/sonic-yang-models/tests/yang_model_tests/tests_config/acl.json
```diff
@@ -1748,5 +1748,355 @@
...
+                "ACL_TABLE_LIST": [
+                    {
+                        "ACL_TABLE_NAME": "SRC-MAC-REWRITE-TABLE",
+                        "policy_desc":"Replace src MAC with ENI MAC",
+                        "ports": [
+                            "Ethernet0",
+                            "Ethernet1"
+                        ],
+                        "stage": "EGRESS",
+                        "type": "L2"
```
anish-n, line 1784: in the ut can you create a new ACL_TABLE_TYPE configuration instead of using a predefined/sw created config called L2?

anish-n (changes requested):

src/sonic-yang-models/tests/yang_model_tests/tests_config/acl.json
```diff
@@ -1748,5 +1748,355 @@
...
+                "ACL_TABLE_LIST": [
+                    {
+                        "ACL_TABLE_NAME": "SRC-MAC-REWRITE-TABLE",
+                        "policy_desc":"Replace src MAC with ENI MAC",
+                        "ports": [
+                            "Ethernet0",
+                            "Ethernet1"
+                        ],
+                        "stage": "EGRESS",
+                        "type": "L3"
```
anish-n, line 1784: this is still using a pre-defined table type, please create a new ACL_TABLE_TYPE configuration with some name like src-mac-rewrite and use that ACL_TABLE_TYPE in ACL_TABLE

anish-n (changes requested):

src/sonic-yang-models/tests/yang_model_tests/tests_config/acl.json
kavyavanibedida: updated table type to INNER_SRC_MAC_REWRITE_TABLE_TYPE

wen587:
Hi @kavyavanibedida , is it possible to add a config sample in https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-yang-models/tests/files/sample_config_db.json?

src/sonic-yang-models/tests/files/sample_config_db.json
```diff
@@ -1104,9 +1116,42 @@
...
+                    "Ethernet25",
+                    "Ethernet24",
+                    "PortChannel0003"
+                ],
+                "stage": "EGRESS",
+                "type": "INNER_SRC_MAC_REWRITE_TABLE_TYPE"
+            }
+        },
+        "ACL_TABLE_TYPE": {
+            "ACL_TABLE_TYPE_LIST": [
```
anish-n, line 1138: I don't think there exists a ACL_TABLE_TYPE_LIST, it should be one or more members under ACL_TABLE_TYPE itself

anish-n (changes requested):

src/sonic-yang-models/tests/files/sample_config_db.json
kavyavanibedida: Have used the existing ACL_TABLE_TYPE_LIST in sonic-acl.yang. Should this be removed?

anish-n (approved):

bingwang-ms (approved):

prsunny (dismissed):
lgtm,

qiluo-msft (approved):

prsunny (approved):

prsunny:
lgtm, @qiluo-msft to merge

anish-n:
@kavyavanibedida upon looking at the SWSS code, I see that the Yang model has a discrepancy with the schema followed by Swss for the VNI field. In SWSS the field is called TUNNEL_VNI, https://github.com/sonic-net/sonic-swss/blob/master/orchagent/aclorch.h#L49, but here it is being implemented as VNI, can you please raise a follow up PR to fix the same.

PR #21915, opened 2025-03-04
Add build time config for change database service from Redis to Valkey

liuh-80:
Add build time flag for replace Redis with Valkey 

#### Why I did it
Redis changed license from BSD to Redis Source Available License v2 (RSALv2) and the Server Side Public License v1 (SSPLv1).
Valkey forked source code in day 1 after Redis change license (Redis 7.3), and still keep BSD license.

##### Work item tracking
- Microsoft ADO **(number only)**: 31887132

#### How I did it
Add a build time flag ENABLE_VALKEY_DB to replace Redis database with valkey database.
There are following reason can't use a runtime flag:
1. Some external scripts may depend on redis-server redis-cli, so need create symbol link to link these commands to valkey-server/valkey-cli
2. Database container initializes with docker-database-init.sh, which is only run 1 time when create database container. which means we can't change the link by reload config

#### How to verify it
Pass all test case.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ] 

Replace Redis with Valkey

### DIFF
--- a/dockers/docker-base-bookworm/Dockerfile.j2
+++ b/dockers/docker-base-bookworm/Dockerfile.j2
@@ -111,6 +111,6 @@ COPY ["etc/rsyslog.d/*", "/etc/rsyslog.d/"]
 COPY ["root/.vimrc", "/root/.vimrc"]
 
 RUN ln /usr/bin/vim.tiny /usr/bin/vim
-RUN mv /usr/bin/redis-cli /usr/bin/redis-cli.ori
+RUN cp /usr/bin/redis-cli /usr/bin/redis-cli.ori
 
 COPY ["etc/supervisor/supervisord.conf", "/etc/supervisor/"]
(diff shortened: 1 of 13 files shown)


dockers/docker-base-bookworm/Dockerfile.j2
```diff
@@ -47,8 +47,8 @@ RUN apt update &&            \
         python-is-python3    \
         vim-tiny             \
         rsyslog              \
-# Install redis-tools
-        redis-tools          \
+# Install valkey-tools
+        valkey-tools         \
```
qiluo-msft, line 51: Is it possible to enable both, and use a runtime config to pick one. No need to warm switch between them, but a /etc/sonic/config_db.json could change the config by a `config reload`.

liuh-80: Will improve the PR, there may be some script will check the process name, so need figure out how dynamically create symbol link during config reload.

liuh-80:
close because this change not necessary

PR #21923, opened 2025-03-04
[rsyslog]: Remote logging with the highest rule priority

nazariig:
#### Why I did it
This fix resolves issue with missing certain types of messages while forwarding to the remote server.

Basically, this moves remote syslog forwarding rules to be executed before any other instructions take place,
so everything received by a local syslog instance will be forwarded to the remote machine.

The fix allows remote forwarding of messages defined as:
https://github.com/sonic-net/sonic-buildimage/blob/master/files/image_config/rsyslog/rsyslog.d/00-sonic.conf.j2

##### Work item tracking

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/files/image_config/rsyslog/rsyslog.conf.j2
+++ b/files/image_config/rsyslog/rsyslog.conf.j2
@@ -65,41 +65,8 @@ template(name="WelfRemoteFormat" type="string" string="%TIMESTAMP% id=firewall t
 :::date-second%\" fw=\"{{ fw_name }}\" pri=%syslogpriority% msg=\"%syslogtag%%msg:::sp-if-no-1st-sp%%msg:::drop-last-lf%\"\n")
 
 #
-# Set the default permissions for all log files.
-#
-$FileOwner root
-$FileGroup adm
-$FileCreateMode 0640
-$DirCreateMode 0755
-$Umask 0022
-
-#
-# Where to place spool and state files
-#
-$WorkDirectory /var/spool/rsyslog
-
-#
-# Include all config files in /etc/rsyslog.d/
-#
-$IncludeConfig /etc/rsyslog.d/*.conf
-
-#
-# Suppress duplicate messages and report "message repeated n times"
-#
-$RepeatedMsgReduction on
-
-###############
-#### RULES ####
-###############
-
-#
 # Remote syslog logging
 #
-
-# The omfwd plug-in provides the core functionality of traditional message
-# forwarding via UDP and plain TCP. It is a built-in module that does not need
-# to be loaded.
-
 {% set servers = SYSLOG_SERVER | d({}) -%}
 {% for server in servers %}
 {% set conf = servers[server] | d({}) -%}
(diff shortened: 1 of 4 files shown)


files/image_config/rsyslog/rsyslog.conf.j2
```diff
@@ -64,42 +64,9 @@ template(name="WelfRemoteFormat" type="string" string="%TIMESTAMP% id=firewall t
...
-$FileOwner root
-$FileGroup adm
-$FileCreateMode 0640
-$DirCreateMode 0755
-$Umask 0022
-
-#
-# Where to place spool and state files
-#
-$WorkDirectory /var/spool/rsyslog
```
liuh-80, line 79: Suggest keep permission and working directory config before remote syslog config.

According to doc, syslog may buffer massive data on disk with spool, not sure if move these after the remote syslog config will have impact:
https://www.rsyslog.com/doc/tutorials/high_database_rate.html

files/image_config/rsyslog/rsyslog.conf.j2
```diff
@@ -64,42 +64,9 @@ template(name="WelfRemoteFormat" type="string" string="%TIMESTAMP% id=firewall t
...
-
-#
-# Include all config files in /etc/rsyslog.d/
-#
-$IncludeConfig /etc/rsyslog.d/*.conf
-
-#
-# Suppress duplicate messages and report "message repeated n times"
-#
-$RepeatedMsgReduction on
```
liuh-80, line 89: RepeatedMsgReduction also need keep here? move this to end of file may impact remove syslog server-side storage.

files/image_config/rsyslog/rsyslog.conf.j2
```diff
@@ -64,42 +64,9 @@ template(name="WelfRemoteFormat" type="string" string="%TIMESTAMP% id=firewall t
...
-
-###############
-#### RULES ####
-###############
-
 #
 # Remote syslog logging
 #
-
-# The omfwd plug-in provides the core functionality of traditional message
```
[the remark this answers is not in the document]
nazariig: @zbud-msft this comment doesn't contain any valuable information from the configuration point of view, still i don't have a strong opinion here so adding back

files/image_config/rsyslog/rsyslog.conf.j2
nazariig: @liuh-80 duplicate message reduction is also a policy of the remote syslog instance and configuration from the day 1. Do you prefer to apply it before forwarding to the remote?

files/image_config/rsyslog/rsyslog.conf.j2
liuh-80: If this is from day one then this is OK

files/image_config/rsyslog/rsyslog.conf.j2
liuh-80: SAA

zbud-msft (approved):

zbud-msft:
@nazariig Can you please raise PR in sonic-mgmt that provides testcase for this fix?

qiluo-msft (approved):

kperumalbfn:
@nazariig could you please create a PR for 202411 due to conflict.

PR #21928, opened 2025-03-04
[FRR]: Fix `no srv6` command

cscarpitta:
Bring PR https://github.com/FRRouting/frr/pull/18289 from FRR mainline to SONiC to fix the `no srv6` command.

---

A user can configure static SIDs as follows:

```
[...]
segment-routing
 srv6
  static-sids
   sid fcbb:bbbb:1::/48 locator MAIN behavior uN
   sid fcbb:bbbb:1:fe00::/64 locator MAIN behavior uDT46
[...]
```
When the user runs vtysh and executes the no srv6 command, the expectation is that staticd will deallocate all SIDs.
However, currently FRR does not behaves as expected. After the user executes no srv6, the SIDs are still present.

This PR fixes the issue and makes the no srv6 command remove all SIDs as expected.


ahsalam (approved):
Merged in FRR mainline.

PR #21995, opened 2025-03-11
Fix SSH crash when started with -C  argument  /usr/sbin/sshd -T -C user=root -C host=nd-e013 -C addr=127.0.0.1 -C lport=22

asha-behera:
Fixes issue https://github.com/sonic-net/sonic-buildimage/issues/21997
#### Why I did it

SSH was crashing due to the patch 0003-Export-remote-info-for-authorization.patch when ssh is started with -C  argumnets .
Running debian packaged sshd with  '-T -C user=root -C host=nd-e013 -C addr=127.0.0.1 -C lport=22' arguments doesn't cause crash.
ssh NULL pointer was being accessed hence crash occured .
We are calling export_remote_info before ssh is assigned using ssh_alloc_session_state API later in code

Recreation of the crash:
 (gdb) file /usr/sbin/sshd
 (gdb) run -T -C user=root -C host=nd-e013 -C addr=127.0.0.1 -C lport=22
 Starting program: /usr/sbin/sshd -T -C user=root -C host=nd-e013 -C addr=127.0.0.1 -C lport=22
 [Thread debugging using libthread_db enabled]
 Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".

 Program received signal SIGSEGV, Segmentation fault.
 ssh_remote_ipaddr (ssh=ssh@entry=0x0) at ../../packet.c:532
 Download failed: Invalid argument.  Continuing without source file ./debian/build-deb/../../packet.c.
 532     ../../packet.c: No such file or directory.
 (gdb) bt
 #0  ssh_remote_ipaddr (ssh=ssh@entry=0x0) at ../../packet.c:532
 #1  0x0000555555572d6e in export_remote_info (ssh=ssh@entry=0x0) at ../../auth.c:1111
 #2  0x00005555555620e2 in main (ac=10, av=0x555555633960) at ../../sshd.c:1672...

#### How I did it

Add NULL check in export_remote_info

#### How to verify it

Start ssh with  arguments and check no crash 
 
admin#  /usr/sbin/sshd -T -C user=root -C host=nd-e013 -C addr=127.0.0.1 -C lport=22
addressfamily any
listenaddress 0.0.0.0:22
listenaddress [::]:22
usepam yes
logingracetime 120
x11displayoffset 10
maxauthtries 6
maxsessions 10
clientaliveinterval 600
....

### DIFF
--- a/src/openssh/patch/0003-Export-remote-info-for-authorization.patch
+++ b/src/openssh/patch/0003-Export-remote-info-for-authorization.patch
@@ -1,40 +1,42 @@
-From 51b3d58afef6796fe0568deb4c3765e24cc828c9 Mon Sep 17 00:00:00 2001
-From: liuh-80 <liuh@microsoft.com>
-Date: Fri, 30 Sep 2022 16:57:03 +0800
+From 191df972f207774f8703ebf7fffea1129c11ca68 Mon Sep 17 00:00:00 2001
+From: ab952469 <asha.behera@broadcom.com>
+Date: Thu, 13 Mar 2025 14:46:03 +0530
 Subject: [PATCH] Export remote info for authorization. authorization.
 
 ---
- auth.c    | 11 +++++++++++
+ auth.c    | 14 ++++++++++++++
  auth.h    |  3 +++
  session.c |  3 +++
  sshd.c    |  5 +++++
- 4 files changed, 22 insertions(+)
+ 4 files changed, 25 insertions(+)
 
 diff --git a/auth.c b/auth.c
-index c3693ba3f..96d551922 100644
+index 6653b9e..d9ceac4 100644
 --- a/auth.c
 +++ b/auth.c
-@@ -764,3 +764,14 @@ auth_authorise_keyopts(struct ssh *ssh, struct passwd *pw,
+@@ -764,3 +764,17 @@ auth_restrict_session(struct ssh *ssh)
  		fatal_f("failed to restrict session");
  	sshauthopt_free(restricted);
  }
 +
 +/* Export remote IP address and port for authorization. */
-+void
++	void
 +export_remote_info(struct ssh *ssh)
 +{
-+	const char *remote_ip = ssh_remote_ipaddr(ssh);
-+	const int remote_port = ssh_remote_port(ssh);
-+	const char remote_addr_port[32 + INET6_ADDRSTRLEN];
-+	snprintf(remote_addr_port, sizeof(remote_addr_port), "%s %d", remote_ip, remote_port);
-+	setenv("SSH_CLIENT_IPADDR_PORT", remote_addr_port, 1);
++	if (ssh != NULL)
++	{
++		const char *remote_ip = ssh_remote_ipaddr(ssh);
++		const int remote_port = ssh_remote_port(ssh);
++		const char remote_addr_port[32 + INET6_ADDRSTRLEN];
++		snprintf(remote_addr_port, sizeof(remote_addr_port), "%s %d", remote_ip, remote_port);
++		setenv("SSH_CLIENT_IPADDR_PORT", remote_addr_port, 1);
++	}
 +}
-\ No newline at end of file
 diff --git a/auth.h b/auth.h
-index 3cfce0eaf..3a34742b1 100644
+index d16dc66..8b87656 100644
 --- a/auth.h
 +++ b/auth.h
-@@ -229,6 +229,9 @@ struct passwd *fakepw(void);
+@@ -241,6 +241,9 @@ FILE	*auth_openprincipals(const char *, struct passwd *, int);
  
  int	 sys_auth_passwd(struct ssh *, const char *);
  
@@ -45,7 +47,7 @@ index 3cfce0eaf..3a34742b1 100644
  krb5_error_code ssh_krb5_cc_gen(krb5_context, krb5_ccache *);
  #endif
 diff --git a/session.c b/session.c
-index a638ceef1..c615cb3d0 100644
+index 768d3b3..40f3c41 100644
 --- a/session.c
 +++ b/session.c
 @@ -619,6 +619,9 @@ do_exec_pty(struct ssh *ssh, Session *s, const char *command)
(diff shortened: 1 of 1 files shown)


asha-behera:
@liuh-80 Please check this PR . Related to your changes in https://github.com/sonic-net/sonic-buildimage/commit/1f88a3ee0a6b7a5f3d7fb7d7177095e40f09a57e

src/openssh/patch/0004-ssh-crash-fix.patch
```diff
@@ -0,0 +1,107 @@
...
+ #ifndef HAVE_OSF_SIA
+diff --git a/sshd.c b/sshd.c
+index a2761f1..e14290a 100644
+--- a/sshd.c
++++ b/sshd.c
+@@ -1658,7 +1658,6 @@ main(int ac, char **av)
+ 			break;
+ 		case 'C':
+ 			/* Export remote IP address and port for authorization. */
+-			export_remote_info(ssh);
```
asha-behera, line 89: @liuh-80 : The call of export_remote_info(ssh) is not right as ssh is assigned after call to ssh_packet_set_connection at line 2169 in file sshd.c .
I want to remove this API call to export_remote_info as it is a NOOP as ssh pointer will always be NULL here

liuh-80: We need export_remote_info to export remote address for TACACS feature.
I will investigation and fix the issue, but we need keep the export_remote_info code.

liuh-80: Please make sure all TACACS related test case does not break.
We need export_remote_info for TACACS per-command authorization.
I think you can keeps those export_remote_info invoke code no change, because the fix in export_remote_info already return when ssh is empty.

src/openssh/patch/0004-ssh-crash-fix.patch
```diff
@@ -0,0 +1,107 @@
...
+ 			if (parse_server_match_testspec(connection_info,
+ 			    optarg) == -1)
+@@ -2182,7 +2181,10 @@ main(int ac, char **av)
+ 	remote_ip = ssh_remote_ipaddr(ssh);
+ 
+ 	/* Export remote IP address and port for authorization. */
+-	export_remote_info(ssh);
++        if (export_remote_info(ssh) != 0)
++        {
++          exit(1);
```
liuh-80, line 100: Suggest don't exit here, this will change sshd behavior.

src/openssh/patch/0004-ssh-crash-fix.patch
```diff
@@ -0,0 +1,107 @@
...
+diff --git a/packet.c b/packet.c
+index 00e3180..f53ce95 100644
+--- a/packet.c
++++ b/packet.c
+@@ -529,7 +529,7 @@ ssh_remote_ipaddr(struct ssh *ssh)
+ 	int sock;
+ 
+ 	/* Check whether we have cached the ipaddr. */
+-	if (ssh->remote_ipaddr == NULL) {
++	if (ssh != NULL && ssh->remote_ipaddr == NULL) {
```
liuh-80, line 61: This part of code belong to sshd, suggest don't change it's behavior.

src/openssh/patch/0004-ssh-crash-fix.patch
asha-behera: But if it will always ssh argument to export_remote_info   will always be NULL, then the call to export_remote_info  is of no use. What is the usecase to keep the call here ?

liuh-80:
I suggest fix the issue in 0003-Export-remote-info-for-authorization.patch:

+void
+export_remote_info(struct ssh *ssh)
+{
+      if (ssh == NULL)
+      {
+        return;
+      }
+	const char *remote_ip = ssh_remote_ipaddr(ssh);
+	const int remote_port = ssh_remote_port(ssh);
+	const char remote_addr_port[32 + INET6_ADDRSTRLEN];
+	snprintf(remote_addr_port, sizeof(remote_addr_port), "%s %d", remote_ip, remote_port);
+	setenv("SSH_CLIENT_IPADDR_PORT", remote_addr_port, 1);
+}

src/openssh/patch/0004-ssh-crash-fix.patch
liuh-80: For remove export_remote_info at line 89, if the ssh always be NULL, I agree with you, this can be removed.

src/openssh/patch/0004-ssh-crash-fix.patch
asha-behera: @liuh-80 : Do you see any issue with the check ? It is good to have check of ssh NULL before accessing its contents remote_ipaddr.

src/openssh/patch/0004-ssh-crash-fix.patch
```diff
@@ -0,0 +1,107 @@
...
++++ b/auth.c
+@@ -1191,12 +1191,17 @@ auth_authorise_keyopts(struct ssh *ssh, struct passwd *pw,
+ }
+ 
+ /* Export remote IP address and port for authorization. */
+-void
++int
+ export_remote_info(struct ssh *ssh)
+ {
++        if (ssh == NULL)
```
asha-behera, line 26: @liuh-80 If we do not intend to do anything with return status of export_remote_info as it will change behavior ...I will chnage as below :   Hope it is ok ?
/* Export remote IP address and port for authorization. */
void
export_remote_info(struct ssh *ssh)
{
        if (ssh != NULL)
		{
			const char *remote_ip = ssh_remote_ipaddr(ssh);
			const int remote_port = ssh_remote_port(ssh);
			const char remote_addr_port[32 + INET6_ADDRSTRLEN];
			snprintf(remote_addr_port, sizeof(remote_addr_port), "%s %d", remote_ip, remote_port);
			setenv("SSH_CLIENT_IPADDR_PORT", remote_addr_port, 1);
		}
}

src/openssh/patch/0004-ssh-crash-fix.patch
liuh-80: This part of code never crashes in official openssh, so I think ssh will never be NULL here, if we make the change, the risk is this patch may break with future version.

I also agree with you that the new code does not have issue. so just a suggestion.

src/openssh/patch/0004-ssh-crash-fix.patch
asha-behera: keeping in intact, to keep changes minimal

src/openssh/patch/0004-ssh-crash-fix.patch
asha-behera: removing this change for minimal change

src/openssh/patch/0004-ssh-crash-fix.patch
liuh-80: #Closed

liuh-80 (approved):

liuh-80:
@asha-behera build failed because apply patch failed, you may need upgrade your patch

Praveen-Brcm (approved):

PR #21999, opened 2025-03-11
Removed check to allow BGP peer in a vnet

sudarshankumar4893:
#### Why I did it
This change allows BGP peer creation inside a VNET through bgpcfgd. The support already exists for peers to be created in VRF. Removing the check which blocks peers to be created inside a VNET. This can already be done directly through FRR configs. However, that method does not persist the configuration. Letting peer addition inside vnet go through via bgpcfgd ensures that the configurations persists after reload or reboot.

We have proper support for handling VNETs and VRFs in sonic now. Hence it is safe to remove the check and allow peers to be created inside a VNET. We are already able to create peers in side VRFs since the support for VRF was added. From FRR perspective, both VNET and VRF translate to a VRF.

##### Work item tracking

#### How I did it
Removed check to allow BGP peer in a vnet.

#### How to verify it
Create a peer inside a VNET through bgpcfgd to verify that the peer gets added. Also used FRR CLI commands to validate details about the peer.

#### Sample configuration
Adding the following configuration to config_db.json results in creation of peer and dynamic peer in the Vnet.
"BGP_NEIGHBOR": {
        "Vnet1|10.0.0.57": {
            "admin_status": "up",
            "asn": "64600",
            "holdtime": "10",
            "keepalive": "3",
            "local_addr": "10.0.0.56",
            "name": "ARISTA01T1",
            "nhopself": "0",
            "rrclient": "0"
        },
        "Vnet2|fc00::7e": {
            "admin_status": "up",
            "asn": "64600",
            "holdtime": "10",
            "keepalive": "3",
            "local_addr": "fc00::7d",
            "name": "ARISTA04T1",
            "nhopself": "0",
            "rrclient": "0"
        }
    },
    "BGP_PEER_RANGE": {
        "Vnet2|BGPSLBPassive": {
            "ip_range": [
                "10.0.0.60/30"
            ],
            "peer_asn": "64600",
            "src_address": "10.0.0.60",
            "name": "BGPSLBPassive"
        }
    }
"VNET": {
        "Vnet1": {
            "vni": "799999",
            "vxlan_tunnel": "vtep_v4"
        },
        "Vnet2": {
            "vni": "799999",
            "vxlan_tunnel": "vtep_v4"
        }
    },
    "VXLAN_TUNNEL": {
        "vtep_v4": {
            "src_ip": "10.1.0.32"
        }
    }

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/src/sonic-bgpcfgd/bgpcfgd/managers_bgp.py
+++ b/src/sonic-bgpcfgd/bgpcfgd/managers_bgp.py
@@ -187,19 +187,12 @@ class BGPPeerMgrBase(Manager):
         if "local_addr" not in data:
             log_warn("Peer %s. Missing attribute 'local_addr'" % nbr)
         else:
-            # The bgp session that belongs to a vnet cannot be advertised as the default BGP session.
-            # So we need to check whether this bgp session belongs to a vnet.
             data["local_addr"] = str(netaddr.IPNetwork(str(data["local_addr"])).ip)
             interface = self.get_local_interface(data["local_addr"])
             if not interface:
                 print_data = nbr, data["local_addr"]
                 log_debug("Peer '%s' with local address '%s' wait for the corresponding interface to be set" % print_data)
                 return False
-            vnet = self.get_vnet(interface)
-            if vnet:
-                # Ignore the bgp session that is in a vnet
-                log_info("Ignore the BGP peer '%s' as the interface '%s' is in vnet '%s'" % (nbr, interface, vnet))
-                return True
 
         kwargs = {
             'CONFIG_DB__DEVICE_METADATA': self.directory.get_slot("CONFIG_DB", swsscommon.CFG_DEVICE_METADATA_TABLE_NAME),
--- a/src/sonic-bgpcfgd/bgpcfgd/managers_bgp.py
+++ b/src/sonic-bgpcfgd/bgpcfgd/managers_bgp.py
@@ -242,6 +242,8 @@ class BGPPeerMgrBase(Manager):
         """
         if "admin_status" in data:
             self.change_admin_status(vrf, nbr, data)
+        elif "ip_range" in data and self.peer_type == 'dynamic':
+            self.change_ip_range(vrf, nbr, data)
         else:
             log_err("Peer '(%s|%s)': Can't update the peer. Only 'admin_status' attribute is supported" % (vrf, nbr))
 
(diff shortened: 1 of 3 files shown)


src/sonic-bgpcfgd/bgpcfgd/managers_bgp.py
```diff
@@ -249,6 +242,8 @@ def update_peer(self, vrf, nbr, data):
         """
         if "admin_status" in data:
             self.change_admin_status(vrf, nbr, data)
+        elif "ip_range" in data:
+            self.change_ip_range(vrf, nbr, data)
```
anish-n, line 246: I know that ip_range is specific to dynamic neighbors, but can we add a 2 level check so that this only runs for dynamic neighbor types?

src/sonic-bgpcfgd/bgpcfgd/managers_bgp.py
```diff
@@ -285,6 +280,83 @@ def apply_admin_status(self, vrf, nbr, template_name, admin_state):
...
+        :param vrf: vrf name. Name is equal "default" for the global vrf
+        :param nbr: neighbor ip address (name for dynamic peer type)
+        :param new_ip_range: new ip range
+        :param ip_ranges_to_del: ip ranges to delete
+        """
+        cmd = ""
+        for ip_range in ip_ranges_to_del:
+            cmd += "no bgp listen range %s peer-group %s\n" % (ip_range, nbr)
+        for ip_range in new_ip_range:
+            cmd += "bgp listen range %s peer-group %s\n" % (ip_range, nbr)
```
anish-n, line 354: should we do a bgp listen range command from here or should we rather render the template which results in the addition?

sudarshankumar4893: If I use the existing peer add template, it would render some unnecessary configuration statements. If we want to render this via template, should I create a new template for IP range update handling?

src/sonic-bgpcfgd/bgpcfgd/managers_bgp.py
sudarshankumar4893: Thanks. Added the check for peer type.

src/sonic-bgpcfgd/bgpcfgd/managers_bgp.py
anish-n: I would suggest passing down kwargs with added and removed peer group ranges, and I think that it be used to handle all 3: create, update, and delete with the same template. Though if we really need separate templates for these cases that is fine too, but I think its doable to put all the handling in a single template.

sudarshankumar4893: Changed the "add" template to handle the update as well. Thanks.

prsunny:
@sudarshankumar4893 , can you please provide some example configs in the description with your code changes? Please check the build failures.

prsunny:
@StormLiangMS to review as well.

StormLiangMS:
hi @sudarshankumar4893  
1. could you split this PR to two PRs? The 1st is for vnet, and the 2nd is to allow dynamic configuration of ip range?
2. for the dynamic configuration of ip range, could you have a test plan how to test with sonic-mgmt? Since the dynamic change of configuration could lead to issues, like reset or restart the BGP sessions. We need to make sure the behavior is as we expected.

StormLiangMS (changes requested):
hi @sudarshankumar4893  
1. could you split this PR to two PRs? The 1st is for vnet, and the 2nd is to allow dynamic configuration of ip range?
2. for the dynamic configuration of ip range, could you have a test plan how to test with sonic-mgmt? Since the dynamic change of configuration could lead to issues, like reset or restart the BGP sessions. We need to make sure the behavior is as we expected.

prsunny (approved):
lgtm

anish-n (approved):

StormLiangMS (approved):
hi @sudarshankumar4893 Approved with one comment. To dynamic add configuration, which could cause BGP session flap, pls make sure this scenario is addressed by adding test case in sonic-mgmt repo.

PR #22022, opened 2025-03-12
[bgp] Fix the error from mixed internal and external member in bgp peer-group

byu343:
#### Why I did it
The issue is seen on t0-d18u8s4. The peer switches (internal) and t1 switches (external) were added to the same BGP peer group, which is not allowed.

The error is from bgpcfgd:
Peer-group members must be all internal or all external

##### Work item tracking

#### How I did it
Skip adding the internal neighbors (same ASN) to peer group.

#### How to verify it
Tested with sonic-mgmt topo t0-d18u8s4

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/dockers/docker-fpm-frr/frr/bgpd/templates/general/instance.conf.j2
+++ b/dockers/docker-fpm-frr/frr/bgpd/templates/general/instance.conf.j2
@@ -14,13 +14,18 @@
   neighbor {{ neighbor_addr }} shutdown
 {% endif %}
 !
+{% set skip_internal_group_member_hwskus = ['Arista-7050CX3-32S-C28S4'] %}
 {% if neighbor_addr | ipv4 %}
   address-family ipv4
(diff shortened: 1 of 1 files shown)


r12f:
hi @byu343 , in this case, should we fix the topology? it sounds like a topology issue.

byu343:
> hi @byu343 , in this case, should we fix the topology? it sounds like a topology issue.

Hi @r12f That may be possible. The topo used is t0-d18u8s4. Do you think changing the ASN of peer device so that they are not no longer internal bgp peers (https://github.com/sonic-net/sonic-mgmt/blob/master/ansible/vars/topo_t0-d18u8s4.yml#L381) or you have some other ideas?

r12f:
@sdszhang  we need to update the topology to use 4-bytes for the PT0 VMs.

```
  ARISTA04PT0:
    properties:
    - common
    bgp:
      asn: 65100
      peers:
        65100:
        - 10.0.0.162
        - FC00::17D
```

PR #22107, opened 2025-03-21
[vs]: Add script to build individual boost libraries

theasianpianist:
#### Why I did it
When running Docker VS tests in sonic-swss, the host machine needs to have libswsscommon installed which requires the Boost serialization library. If the machine where libswsscommon was built has a different OS from the host machine, the required Boost version may not be available from package repositories.

##### Work item tracking

#### How I did it
Add a script to build individual boost libraries from source.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/platform/vs/docker-sonic-vs/scripts/build_boost_lib.sh
+++ b/platform/vs/docker-sonic-vs/scripts/build_boost_lib.sh
@@ -1,0 +1,73 @@
+#!/bin/bash
+
+# Run script with desired library and version, e.g. to get libboost-serialization1.71.0, run:
+# build_boost_lib.sh serialization 1.71.0
+
+script=$(basename "$0")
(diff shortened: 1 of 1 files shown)


saiarcot895:
Instead of this, it would be more efficient (and maybe future-proofing) to build swsscommon for newer/other versions of Ubuntu.

theasianpianist:
> Instead of this, it would be more efficient (and maybe future-proofing) to build swsscommon for newer/other versions of Ubuntu.

This is more to run swsscommon on older versions of Ubuntu where the available libboost version is too old

PR #22190, opened 2025-03-31
Move software_bfd to system_defaults and enable for DPU.

dypet:
#### Why I did it
The software_bfd flag is currently under FEATURE; it is more appropriate to put it under SYSTEM_DEFAULTS as it is a simple enabled/disabled flag. Additionally, enabling it by default on t1-smartswitch DPU.

##### Work item tracking

#### How I did it
Moved software bfd checks from FEATURE to SYSTEM_DEFAULTS, and enabled it by default in src/sonic-config-engine/config_samples.py. 

#### How to verify it
Build and install DPU image (following https://github.com/sonic-net/sonic-buildimage/pull/22058), check that software_bfd flag is set in config_db.json and runningconfig by default. Unit tests during build also check this using src/sonic-config-engine/tests/sample_output/t1-smartswitch-dpu.json

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [x] master (953af37)

Move software_bfd to system_defaults and enable for DPU.

### DIFF
--- a/dockers/docker-fpm-frr/frr/supervisord/supervisord.conf.j2
+++ b/dockers/docker-fpm-frr/frr/supervisord/supervisord.conf.j2
@@ -206,7 +206,7 @@ dependent_startup_wait_for=bgpd:running
 
 {% endif %}
 
-{% if FEATURE is defined and FEATURE.software_bfd is defined and FEATURE.software_bfd.state is defined and FEATURE.software_bfd.state == "enabled" %}
+{% if SYSTEM_DEFAULTS is defined and SYSTEM_DEFAULTS.software_bfd is defined and SYSTEM_DEFAULTS.software_bfd.status is defined and SYSTEM_DEFAULTS.software_bfd.status == "enabled" %}
 [program:bfdmon]
 command=/usr/local/bin/bfdmon
 priority=6
--- a/src/sonic-config-engine/config_samples.py
+++ b/src/sonic-config-engine/config_samples.py
@@ -151,6 +151,12 @@ def generate_t1_smartswitch_dpu_sample_config(data, ss_config):
     data['DEVICE_METADATA']['localhost']['subtype'] = 'SmartSwitch'
     data['DEVICE_METADATA']['localhost']['bgp_asn'] = '65100'
 
+    data['SYSTEM_DEFAULTS'] = {
+        "software_bfd": {
+            "status": "enabled"
+        }
+    }
+
     for port in natsorted(data['PORT']):
         data['PORT'][port]['admin_status'] = 'up'
         data['PORT'][port]['mtu'] = '9100'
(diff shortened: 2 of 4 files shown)


dockers/docker-fpm-frr/frr/supervisord/supervisord.conf.j2
```diff
@@ -206,7 +206,7 @@ dependent_startup_wait_for=bgpd:running
 
 {% endif %}
 
-{% if FEATURE is defined and FEATURE.software_bfd is defined and FEATURE.software_bfd.state is defined and FEATURE.software_bfd.state == "enabled" %}
+{% if SYSTEM_DEFAULTS is defined and SYSTEM_DEFAULTS.software_bfd is defined and SYSTEM_DEFAULTS.software_bfd.status is defined and SYSTEM_DEFAULTS.software_bfd.status == "enabled" %}
```
prsunny, line 209: Can we check for both System default or feature table? Thinking on how this can be enabled for non-dpu cases.

dypet: For non-dpu case the flag can still be set under SYSTEM_DEFAULTS in config_db.json, either by user or by platform

KrisNey-MSFT: > For non-dpu case the flag can still be set under SYSTEM_DEFAULTS in config_db.json, either by user or by platform

For SW BFD on an NPU, this would be a breaking change?  Introduce SW BFD into the Feature table (prince) which reverts the prior change.  

Add to both System Default and Feature to ensure BFD will work.  (Usually Feature table is associated with a service).

Check the email w/ @dypet  and team to ensure we are aligned.

prsunny (approved):
lgtm. @kperumalbfn , could you please merge?

dockers/docker-fpm-frr/frr/supervisord/supervisord.conf.j2
prsunny: agree that FEATURE table enablement is not a fully working solution.

kperumalbfn (approved):

src/sonic-config-engine/config_samples.py
```diff
@@ -151,6 +151,12 @@ def generate_t1_smartswitch_dpu_sample_config(data, ss_config):
     data['DEVICE_METADATA']['localhost']['subtype'] = 'SmartSwitch'
     data['DEVICE_METADATA']['localhost']['bgp_asn'] = '65100'
 
+    data['SYSTEM_DEFAULTS'] = {
+        "software_bfd": {
```
kperumalbfn, line 155: @dypet could you confirm "software_bfd" system default is not set by default for T0s and non-smartswitch T1s?

dypet: It is only set by default on Smartswitch DPU for now

prsunny:
@vvolam for viz

PR #22229, opened 2025-04-03
Add delete template for dynamic BGP peers

NavdhaJindal:
#### Why I did it

This change is to support enhanced CRUD operations on dynamic bgp peers  using a standard template-based mechanism. 
Here is the High level design document explaining this feature:
https://github.com/sonic-net/SONiC/pull/1963

##### Work item tracking

#### How I did it
Upon a delete peer ocurring, we render the delete template(instead of executing the current behavior of ```no neighbor {{ neighbor addr}}```), on the other hand if a delete template is not defined then the default behavior of ```no neighbor {{ neighbor addr}}``` applies as usual, thereby making this change backward compatible.

#### How to verify it
Create a dynamic peer like so in config_db.json:

"BGP_NEIGHBOR": {
        "10.0.0.62": {
            "admin_status": "up",
            "asn": "65100",
            "holdtime": "10",
            "keepalive": "3",
            "local_addr": "10.0.0.63",
            "name": "vlab-01",
            "nhopself": "0",
            "rrclient": "0"
        }
}

Once the peer is up, delete the neighbor (eg. config bgp neighbor remove <neighbor name>). Verify that the neighbor is deleted. 

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/src/sonic-bgpcfgd/bgpcfgd/managers_bgp.py
+++ b/src/sonic-bgpcfgd/bgpcfgd/managers_bgp.py
@@ -109,6 +109,9 @@ class BGPPeerMgrBase(Manager):
 
         if (os.path.exists(self.fabric.env.loader.searchpath[0] + "/" + base_template + "update.conf.j2")):
             self.templates["update"] = self.fabric.from_file(base_template + "update.conf.j2")
+        if os.path.exists(self.fabric.env.loader.searchpath[0] + "/" + base_template + "delete.conf.j2"):
+            self.templates["delete"] = self.fabric.from_file(base_template + "delete.conf.j2")
+            log_info("Using delete template found at %s" % (base_template + "delete.conf.j2"))
 
         deps = [
             ("CONFIG_DB", swsscommon.CFG_DEVICE_METADATA_TABLE_NAME, "localhost/bgp_asn"),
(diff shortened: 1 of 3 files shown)


src/sonic-bgpcfgd/bgpcfgd/managers_bgp.py
```diff
@@ -100,7 +100,7 @@ def __init__(self, common_objs, db_name, table_name, peer_type, check_neig_meta)
         base_template = "bgpd/templates/" + self.constants["bgp"]["peers"][peer_type]["template_dir"] + "/"
         self.templates = {
             "add":         self.fabric.from_file(base_template + "instance.conf.j2"),
-            "delete":      self.fabric.from_string('no neighbor {{ neighbor_addr }}'),
+            "delete":      self.fabric.from_file(base_template + "instance_del.conf.j2"),
```
sudarshankumar4893, line 103: We need to check if the file is present first. If it is not, we continue with same from_string method. if it is, we use the file.

anish-n (approved):

prsunny (approved):

sudarshankumar4893 (approved):

StormLiangMS:
hi @NavdhaJindal  could you come up test plan and add test case in sonic-mgmt repo?
1. When delete the dynamic, it won't cause impact on other neighbors like BGP flap on existing neighbors. 
2. When delete, no crash. 
3. Stress tests, adding/deleting for at least 20 cycles, no crash / BGP flap.

StormLiangMS (approved):
LGTM

PR #22252, opened 2025-04-07
[Yang model] Support empty string in pfc_enable and pfc_wd_enable fields in PORT_QOS_MAP table

stephenxs:
#### Why I did it

Support empty string in pfc_enable and pfc_wd_enable fields in PORT_QOS_MAP table to represent "disable PFC ".

##### Work item tracking

#### How to verify it

Unit/manul test

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/src/sonic-yang-models/yang-models/sonic-port-qos-map.yang
+++ b/src/sonic-yang-models/yang-models/sonic-port-qos-map.yang
@@ -91,16 +91,18 @@ module sonic-port-qos-map {
 
                 leaf pfc_enable {
                     type string {
-                        pattern "[0-7](,[0-7])*";
+                        pattern "([0-7](,[0-7])*)?";
                     }
+                    description
+                        "Specify the queue(s) on which PFC is enabled. Empty string is allowed to disable PFC on all queues.";
                 }
 
                 leaf pfcwd_sw_enable {
                     type string {
-                        pattern "[0-7](,[0-7])*";
+                        pattern "([0-7](,[0-7])*)?";
                     }
                     description
-                        "Specify the queue(s) on which software pfc watchdog are enabled.";
+                        "Specify the queue(s) on which software pfc watchdog are enabled. Empty string is allowed to disable watchdog on all queues.";
                 }
 
                 leaf pfc_to_queue_map {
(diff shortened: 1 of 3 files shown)


liat-grozovik:
@qiluo-msft  @r12f can you please help to review and merge?

src/sonic-yang-models/yang-models/sonic-port-qos-map.yang
```diff
@@ -91,16 +91,18 @@ module sonic-port-qos-map {
...
                     type string {
-                        pattern "[0-7](,[0-7])*";
+                        pattern "([0-7](,[0-7])*)?";
                     }
                     description
-                        "Specify the queue(s) on which software pfc watchdog are enabled.";
+                        "Specify the queue(s) on which software pfc watchdog are enabled. Empty string is allowed to disable watchdog on all queues.";
                 }
 
                 leaf pfc_to_queue_map {
```
r12f, line 108: Is it possible to also help see if pfc_to_queue and pfc_to_pg map can be ignored too?

stephenxs: Hi @r12f 
Thanks for the comments 
I do not suggest doing so, because 
1. It suffices to disable PFC on both directions by clearing pfc_enable
2. The ASIC always have PFC to queue and PFC to PG mappings. Not setting them means keep them as the default, the customer won't understand what the mappings are, which is not good

What do you think?

r12f: ah, Stephen, what I mean is to allow the empty list. and looks like it is ok.

r12f (approved):

src/sonic-yang-models/yang-models/sonic-port-qos-map.yang
r12f: @kperumalbfn for viz to make sure it matches with our generated policies.

stephenxs: @kperumalbfn 
Already done here by providing [generate_lossless_tc_list](https://github.com/sonic-net/sonic-buildimage/blob/master/device/mellanox/x86_64-nvidia_sn5600-r0/Mellanox-SN5600-C256S1/qos.json.j2#L214) generating empty PFC enable and PFC watchdog enable bitmaps.

liat-grozovik:
@r12f  @kperumalbfn can you sign in so we can merge?

PR #22254, opened 2025-04-07
sonic-yang-mgmt: Generic Config Updater - performance dependencies

bhouse-nexthop:
#### Why I did it

***Important***: Please review each commit (including commit message) individually.  Looking at the patch-set as a whole may cause confusion.

Generic Config Updater (GCU) is notoriously slow.  These patches add some helpers for the GCU overhaul (mostly in sonic-utilities) in order to facilitate the optimizations.  These changes are in sonic-yang-mgmt plus a patch to libyang 1.

Changes include:
 * Libyang v1 was not exposing `must` data for leaf nodes (like it does for other node types).  Patch to correct this oversight.
 * Loading of sonic configuration data should not mutate the user-provided data, this forces callers to know they need to deepcopy the data, plus in most instances data won't be mutated.
 * sonic-yang-mgmt test models should more closely mimic real sonic yang models as we can't implement other features that assume this otherwise (sonic mandates that the top-level container has the same name as the module).
 * Import `sonic configdb`<>`yang xpath` conversion from sonic-utilities as this should be shared code.  90% of this code is copied from the original source but does contain some bugfixes and enhancements including caching.
 * Make `find_schema_dependencies()` public, plus add the ability to find dependencies recursively.  This implementation is caching.
 * Add new `find_schema_must_count()`, with recursive capabilities to find if a node (and its children) have must clauses. This implementation is caching.

sonic-utilities PR: https://github.com/sonic-net/sonic-utilities/pull/3831

Stats below ... (stats need both this and the sonic-utilities PR to be relevant)...

<ins>**Original Performance:**</ins>
Dry Run:
```
time sudo config replace -d ./config_db.json
...
real	2m51.588s
user	2m23.777s
sys	0m25.300s
```

Full:
```
time sudo config replace ./config_db.json
...
real	14m53.772s
user	12m2.376s
sys	2m8.908s
```

<ins>**With Patch**:</ins>
Dry Run:
```
time sudo config replace -d ./config_db.json
...
real	0m59.602s
user	0m56.434s
sys	0m2.110s
```

Full:
```
time sudo config replace ./config_db.json
...
real	1m54.303s
user	0m58.482s
sys	0m2.545s
```

So that's roughly 3x improvement for dry-run, and 7.5x improvement for full commit.  There is room for improvement on the full commit due to a `sleep(1)` being used between each patch because of a race condition found in the prior code (that was hidden due to a costly sanity check that has been removed).

#### How I did it

Gathered profiling data using cProfile and evaluated where the largest gains could be had.

#### How to verify it

This patch is standalone as it will not cause any issues in other projects which use sonic-yang-mgmt or libyang, however the performance benefits are in sonic-utilities.   Apply both this commit and the sonic-utilities PR to a local branch, build and run sonic-utilities tests.  Then create a full image, load it onto a DUT (with default configuration), and use the attached 
[config_db.json](https://github.com/user-attachments/files/19635712/config_db.json) to attempt a `config replace` operation (tested on Dell S5248F).

- [x] 202411

master as of 20250521

sonic-yang-mgmt: Generic Config Updater - performance dependencies

N/A

Fixes https://github.com/sonic-net/sonic-buildimage/issues/22372

### DIFF
--- a/src/libyang/patch/libyang-leaf-must.patch
+++ b/src/libyang/patch/libyang-leaf-must.patch
@@ -1,0 +1,29 @@
+diff --git a/swig/cpp/src/Tree_Schema.cpp b/swig/cpp/src/Tree_Schema.cpp
+index 3587320..8da206a 100644
+--- a/swig/cpp/src/Tree_Schema.cpp
++++ b/swig/cpp/src/Tree_Schema.cpp
+@@ -344,6 +344,7 @@ S_Schema_Node Schema_Node_Choice::dflt() {
+ Schema_Node_Leaf::~Schema_Node_Leaf() {};
+ S_Set Schema_Node_Leaf::backlinks() LY_NEW_CASTED(lys_node_leaf, node, backlinks, Set);
+ S_When Schema_Node_Leaf::when() LY_NEW_CASTED(lys_node_leaf, node, when, When);
++std::vector<S_Restr> Schema_Node_Leaf::must() LY_NEW_LIST_CASTED(lys_node_leaf, node, must, must_size, Restr);
+ S_Type Schema_Node_Leaf::type() {return std::make_shared<Type>(&((struct lys_node_leaf *)node)->type, deleter);}
+ S_Schema_Node_List Schema_Node_Leaf::is_key() {
+     uint8_t pos;
+diff --git a/swig/cpp/src/Tree_Schema.hpp b/swig/cpp/src/Tree_Schema.hpp
+index d506891..f8ecc50 100644
+--- a/swig/cpp/src/Tree_Schema.hpp
++++ b/swig/cpp/src/Tree_Schema.hpp
+@@ -683,8 +683,12 @@ public:
+     ~Schema_Node_Leaf();
+     /** get backlinks variable from [lys_node_leaf](@ref lys_node_leaf)*/
+     S_Set backlinks();
++    /** get must_size variable from [lys_node_leaf](@ref lys_node_leaf)*/
++    uint8_t must_size() {return ((struct lys_node_leaf *)node)->must_size;};
+     /** get when variable from [lys_node_leaf](@ref lys_node_leaf)*/
+     S_When when();
++    /** get must variable from [lys_node_leaf](@ref lys_node_leaf)*/
++    std::vector<S_Restr> must();
+     /** get type variable from [lys_node_leaf](@ref lys_node_leaf)*/
+     S_Type type();
+     /** get units variable from [lys_node_leaf](@ref lys_node_leaf)*/
--- a/src/libyang/patch/series
+++ b/src/libyang/patch/series
@@ -3,3 +3,4 @@ libyang_mgmt_framework.patch
 swig.patch
 large_file_support_arm32.patch
 debian-packaging-files.patch
+libyang-leaf-must.patch
--- a/src/sonic-yang-mgmt/setup.py
+++ b/src/sonic-yang-mgmt/setup.py
@@ -30,6 +30,7 @@ setup(
     install_requires = [
         'xmltodict==0.12.0',
         'ijson==3.2.3',
+        'jsonpointer>=1.9',
         'jsondiff>=1.2.0',
         'tabulate==0.9.0'
     ],
@@ -46,7 +47,7 @@ setup(
     include_package_data=True,
     keywords='sonic-yang-mgmt',
     name='sonic-yang-mgmt',
-    py_modules=['sonic_yang', 'sonic_yang_ext'],
+    py_modules=['sonic_yang', 'sonic_yang_ext', 'sonic_yang_path'],
     packages=find_packages(),
     version='1.0',
     zip_safe=False,
--- a/src/sonic-yang-mgmt/sonic_yang.py
+++ b/src/sonic-yang-mgmt/sonic_yang.py
@@ -4,13 +4,14 @@ import syslog
 from json import dump
 from glob import glob
 from sonic_yang_ext import SonicYangExtMixin, SonicYangException
+from sonic_yang_path import SonicYangPathMixin
 
 """
 Yang schema and data tree python APIs based on libyang python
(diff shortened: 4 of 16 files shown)


src/libyang/patch/libyang-leaf-must.patch
```diff
@@ -0,0 +1,29 @@
...
+@@ -683,8 +683,12 @@ public:
+     ~Schema_Node_Leaf();
+     /** get backlinks variable from [lys_node_leaf](@ref lys_node_leaf)*/
+     S_Set backlinks();
++    /** get must_size variable from [lys_node_leaf](@ref lys_node_leaf)*/
++    uint8_t must_size() {return ((struct lys_node_leaf *)node)->must_size;};
+     /** get when variable from [lys_node_leaf](@ref lys_node_leaf)*/
+     S_When when();
++    /** get must variable from [lys_node_leaf](@ref lys_node_leaf)*/
++    std::vector<S_Restr> must();
```
qiluo-msft, line 26: >must [](http://example.com/codeflow?start=26&length=4)

If the new fields are only used in python, is it possible not to patch cpp files?

bhouse-nexthop: No, libyang1 used SWIG to generate python bindings off the C++ bindings, so this is the only way to do it.

qiluo-msft: Sorry let clarify my question: If the new fields are only used in python? If yes, we do not need to patch cpp, we could just add this functionality inside python code, without worrying about SWIG/CFFI at all.

Let me know if it is not true - ie, cpp also need this functionality.

qiluo-msft:
Thanks for the contribution in this big PR! Is it possible to split into smaller and focusing PR? It also help the reviewing and merging.

Suggest several PRs:
- Performance improvement only
   * Loading of sonic configuration data should not mutate the user-provided data, this forces callers to know they need to deepcopy the data, plus in most instances data won't be mutated.

- Test improvement
   * sonic-yang-mgmt test models should more closely mimic real sonic yang models as we can't implement other features that assume this otherwise (sonic mandates that the top-level container has the same name as the module).

- Refactor for better code sharing
  * Import `sonic configdb`<>`yang xpath` conversion from sonic-utilities as this should be shared code.  90% of this code is copied from the original source but does contain some bugfixes and enhancements including caching.

- New functionality, ideally with even smaller PR focus on one feature per PR
  * Libyang v1 was not exposing `must` data for leaf nodes (like it does for other node types).  Patch to correct this oversight.
  * Make `find_schema_dependencies()` public, plus add the ability to find dependencies recursively.  This implementation is caching.
  * Add new `find_schema_must_count()`, with recursive capabilities to find if a node (and its children) have must clauses. This implementation is caching.

src/libyang/patch/libyang-leaf-must.patch
bhouse-nexthop: No, unfortunately that's not true.  libyang is a C library. Then they generate python bindings off the C++ wrapper code using SWIG, so it *must* exist in the C++ code since the python bindings are generated off that automatically.  That's why you don't see anything python-specific in the changeset for that.   The C++ code actually is never used anywhere in sonic except to facilitate the python bindings.

Its a different story for CFFI that libyang3 uses, there is no C++ there at all

bhouse-nexthop:
It would be easy enough to do that, but that's why

> Thanks for the contribution in this big PR! Is it possible to split into smaller and focusing PR? It also help the reviewing and merging.
> 
> Suggest several PRs:
> 
> * Performance improvement only
>   
>   * Loading of sonic configuration data should not mutate the user-provided data, this forces callers to know they need to deepcopy the data, plus in most instances data won't be mutated.
> * Test improvement
>   
>   * sonic-yang-mgmt test models should more closely mimic real sonic yang models as we can't implement other features that assume this otherwise (sonic mandates that the top-level container has the same name as the module).
> * Refactor for better code sharing
>   
>   * Import `sonic configdb`<>`yang xpath` conversion from sonic-utilities as this should be shared code.  90% of this code is copied from the original source but does contain some bugfixes and enhancements including caching.
> * New functionality, ideally with even smaller PR focus on one feature per PR
>   
>   * Libyang v1 was not exposing `must` data for leaf nodes (like it does for other node types).  Patch to correct this oversight.
>   * Make `find_schema_dependencies()` public, plus add the ability to find dependencies recursively.  This implementation is caching.
>   * Add new `find_schema_must_count()`, with recursive capabilities to find if a node (and its children) have must clauses. This implementation is caching.

That's why I left it as separate commits in the PR so it would be easy to review each commit individually, perhaps you're looking at the file changes directly instead of per-commit.  Also this PR passes all sonic-mgmt tests, so we know its not breaking anything.

If you require I break it into multiple smaller PRs, I will.  Just they will *all* have to be merged before https://github.com/sonic-net/sonic-utilities/pull/3831 as it needs them.
-Brad

I can

bhouse-nexthop:
@qiluo-msft I didn't see that you responded to my last comments.  Are you requiring I break this into separate PRs?  Like I said, I left the commits themselves split for that reason in this PR.

qiluo-msft (approved):

bhouse-nexthop:
@qiluo-msft thanks for the approval.  Can this be merged so that the tests on the sister ticket https://github.com/sonic-net/sonic-utilities/pull/3831 can be run?  Thanks!

yejianquan:
Hi @bhouse-nexthop , please resolve the conflict and raise a PR to 202505 branch

bhouse-nexthop:
> Hi @bhouse-nexthop , please resolve the conflict and raise a PR to 202505 branch

@yejianquan looks like this got merged before the 202505 branch was created, so its already included.

bhouse-nexthop:
> @bhouse-nexthop cherry pick PR didn't pass PR checker. Please check!!Azure/sonic-buildimage-msft#1711

doesn't look like the build spawned off properly to me ... are you seeing something else ?

PR #22261, opened 2025-04-08
Handling modification of ip range for dynamic BGP peers

sudarshankumar4893:
#### Why I did it
There is a requirement to be able to modify the listen range of dynamic peers in a peer group and persist the changes through bgpcfgd during run time. This PR adds support for being able to modify the ip_range for a dynamic BGP peer through bgpcfgd.

Here is the High level design document explaining this feature:
[Dynamic BGP peer update and delete support by anish-n · Pull Request #1963 · sonic-net/SONiC](https://github.com/sonic-net/SONiC/pull/1963)

##### Work item tracking

#### How I did it

The ranges from the set handler will be applied. Any older ranges which are not in the new ranges or any older ranges which overlap with the new ranges will be removed and then the new ranges will be applied. There will be no impact to existing ranges which are present in new range as well. If an older range gets deleted, the peers from that range will flap.

#### How to verify it
Create a dynamic peer range like so in config_db.json:
"BGP_PEER_RANGE": {
    "BGPSLBPassive": {
        "ip_range": [
            "10.0.0.60/30"
        ],
        "peer_asn": "64600",
        "src_address": "10.0.0.60",
        "name": "BGPSLBPassive"
    }
}
Once the peer is up, add/delete/modify the range to see that the corresponding FRR commands are generated and peers from new range come up. To parse the new changes, run sonic-cfggen like so:
sonic-cfggen -j /etc/sonic/config_db.json --write-to-db
Sample updates:
1. Updating the existing range.
"BGP_PEER_RANGE": {
    "BGPSLBPassive": {
        "ip_range": [
            "10.0.0.60/29"
        ],
        "peer_asn": "64600",
        "src_address": "10.0.0.60",
        "name": "BGPSLBPassive"
    }
}
2. Adding a new range:
"BGP_PEER_RANGE": {
    "BGPSLBPassive": {
        "ip_range": [
            "10.0.0.60/30","10.0.0.56/30"
        ],
        "peer_asn": "64600",
        "src_address": "10.0.0.60",
        "name": "BGPSLBPassive"
    }
}

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/src/sonic-bgpcfgd/bgpcfgd/managers_bgp.py
+++ b/src/sonic-bgpcfgd/bgpcfgd/managers_bgp.py
@@ -242,6 +242,8 @@ class BGPPeerMgrBase(Manager):
         """
         if "admin_status" in data:
             self.change_admin_status(vrf, nbr, data)
+        elif "ip_range" in data and self.peer_type == 'dynamic':
+            self.change_ip_range(vrf, nbr, data)
         else:
             log_err("Peer '(%s|%s)': Can't update the peer. Only 'admin_status' attribute is supported" % (vrf, nbr))
 
@@ -278,6 +280,94 @@ class BGPPeerMgrBase(Manager):
             log_info("Peer '%s|%s' admin state is set to '%s'" % print_data)
         else:
             log_err("Can't set peer '%s|%s' admin state to '%s'." % print_data)
+
+    def change_ip_range(self, vrf, nbr, data):
+        """
+        Change ip range of a peer
+        :param vrf: vrf name. Name is equal "default" for the global vrf
+        :param nbr: neighbor ip address (name for dynamic peer type)
+        :param data: associated data
+        :return: True if this adding was successful, False otherwise
+        """
+        if data['ip_range']:
+            log_info("Peer '(%s|%s)' ip range is going to be updated with range: %s" % (vrf, nbr, data['ip_range']))
+            new_ip_range = data["ip_range"].split(",")
+            ip_ranges_to_add = new_ip_range
+            ip_ranges_to_del = []
+            existing_ipv4_range, existing_ipv6_range = self.get_existing_ip_ranges(vrf, nbr)
+            if existing_ipv4_range:
+                for ipv4_range in existing_ipv4_range:
+                    if ipv4_range not in new_ip_range:
+                        ip_ranges_to_del.append(ipv4_range)
+                    else:
+                        ip_ranges_to_add.remove(ipv4_range)
+            if existing_ipv6_range:
+                for ipv6_range in existing_ipv6_range:
+                    if ipv6_range not in new_ip_range:
+                        ip_ranges_to_del.append(ipv6_range)
+                    else:
+                        ip_ranges_to_add.remove(ipv6_range)
+            if ip_ranges_to_del or ip_ranges_to_add:
+                log_info("Peer '(%s|%s)' ip range is going to be updated. Ranges to delete: %s Ranges to add: %s" % (vrf, nbr, ip_ranges_to_del, ip_ranges_to_add))
+                self.apply_range_changes(vrf, nbr, ip_ranges_to_add, ip_ranges_to_del, data)
+
+    def get_existing_ip_ranges(self, vrf, nbr):
+        """
+        Get existing ip range of a peer
+        :param vrf: vrf name. Name is equal "default" for the global vrf
+        :param nbr: neighbor ip address (name for dynamic peer type)
+        :return: existing ip range of a peer
+        """
+        ipv4_ranges = []
+        ipv6_ranges = []
+        if vrf == 'default':
+            command = ["vtysh", "-c", "show bgp peer-group %s json" % (nbr)]
+        else:
+            command = ["vtysh", "-c", "show bgp vrf %s peer-group %s json" % (vrf, nbr)]
+        try:
+            ret_code, out, err = run_command(command)
+            if ret_code == 0:
+                js_bgp = json.loads(out)
+                if nbr in js_bgp and 'dynamicRanges' in js_bgp[nbr] and 'IPv4' in js_bgp[nbr]['dynamicRanges'] and 'ranges' in js_bgp[nbr]['dynamicRanges']['IPv4']:
+                    ipv4_ranges= js_bgp[nbr]['dynamicRanges']['IPv4']['ranges']
+                    log_info("Peer '(%s|%s)' already has ipV4 range: %s" % (vrf, nbr, ipv4_ranges))
+                if nbr in js_bgp and 'dynamicRanges' in js_bgp[nbr] and 'IPv6' in js_bgp[nbr]['dynamicRanges'] and 'ranges' in js_bgp[nbr]['dynamicRanges']['IPv6']:
+                    ipv6_ranges= js_bgp[nbr]['dynamicRanges']['IPv6']['ranges']
+                    log_info("Peer '(%s|%s)' already has ipV6 range: %s" % (vrf, nbr, ipv6_ranges))
+                return ipv4_ranges, ipv6_ranges
+            else:
+                log_err("Can't read ip range of peer '%s|%s': %s" % (vrf, nbr, str(err)))
+                return ipv4_ranges, ipv6_ranges
+        except Exception as e:
+            log_err("Error in parsing ip range: %s" % str(e))
+            return ipv4_ranges, ipv6_ranges
+
+    def apply_range_changes(self, vrf, nbr, new_ip_range, ip_ranges_to_del, data):
+        """
+        Apply changes of ip range of a peer
+        :param vrf: vrf name. Name is equal "default" for the global vrf
+        :param nbr: neighbor ip address (name for dynamic peer type)
+        :param new_ip_range: new ip range
+        :param ip_ranges_to_del: ip ranges to delete
+        """
+        print_data = vrf, nbr, data
+        kwargs = {
+            'operation': 'update',
+            'vrf': vrf,
+            'neighbor_addr': nbr,
+            'bgp_session': data,
+            'delete_ranges': ip_ranges_to_del,
+            'add_ranges': new_ip_range
+        }
+        try:
+            cmd = self.templates["add"].render(**kwargs)
+        except jinja2.TemplateError as e:
+            msg = "Peer '(%s|%s)'. Error in rendering the template for 'SET' command '%s'" % print_data
+            log_err("%s: %s" % (msg, str(e)))
+            return True
+        if cmd is not None:
+            self.apply_op(cmd, vrf)
+            log_info("Peer '(%s|%s)' ip range has been scheduled to be updated with new range '%s'" % (vrf, nbr, new_ip_range))
 
     def del_handler(self, key):
         """
--- a/src/sonic-bgpcfgd/bgpcfgd/managers_bgp.py
+++ b/src/sonic-bgpcfgd/bgpcfgd/managers_bgp.py
@@ -101,11 +101,13 @@ class BGPPeerMgrBase(Manager):
         base_template = "bgpd/templates/" + self.constants["bgp"]["peers"][peer_type]["template_dir"] + "/"
         self.templates = {
             "add":         self.fabric.from_file(base_template + "instance.conf.j2"),
-            "update":      self.fabric.from_file(base_template + "instance_update.conf.j2"),
             "delete":      self.fabric.from_string('no neighbor {{ neighbor_addr }}'),
             "shutdown":    self.fabric.from_string('neighbor {{ neighbor_addr }} shutdown'),
             "no shutdown": self.fabric.from_string('no neighbor {{ neighbor_addr }} shutdown'),
         }
+
+        if (os.path.exists(self.fabric.env.loader.searchpath[0] + "/" + base_template + "update.conf.j2")):
+            self.templates["update"] = self.fabric.from_file(base_template + "update.conf.j2")
 
         deps = [
             ("CONFIG_DB", swsscommon.CFG_DEVICE_METADATA_TABLE_NAME, "localhost/bgp_asn"),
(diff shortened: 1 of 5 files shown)


src/sonic-bgpcfgd/bgpcfgd/managers_bgp.py
```diff
@@ -286,6 +288,94 @@ def apply_admin_status(self, vrf, nbr, template_name, admin_state):
...
+        kwargs = {
+            'operation': 'update',
+            'vrf': vrf,
+            'neighbor_addr': nbr,
+            'bgp_session': data,
+            'delete_ranges': ip_ranges_to_del,
+            'add_ranges': new_ip_range
+        }
+        try:
+            cmd = self.templates["add"].render(**kwargs)
```
anish-n, line 370: In my opinion this could be a breaking change such that community members will be mandated to change their internal templates to account for the new call/change, do you agree @prsunny, @StormLiangMS. I think a better way to do this could be to define a new template type for update, only IF an update template exists in the image then we call into the apply_range_changes function and do the render on the special update template itself

anish-n (changes requested):

StormLiangMS (changes requested):
hi @sudarshankumar4893  since this is for phase 2, could we have test cases ready for review and in place at same time?

src/sonic-bgpcfgd/bgpcfgd/managers_bgp.py
sudarshankumar4893: Added a new update template and only handling range changes if the file exists.

src/sonic-bgpcfgd/bgpcfgd/managers_bgp.py
```diff
@@ -279,6 +290,91 @@ def apply_admin_status(self, vrf, nbr, template_name, admin_state):
...
+        """
+        Apply changes of ip range of a peer
+        :param vrf: vrf name. Name is equal "default" for the global vrf
+        :param nbr: neighbor ip address (name for dynamic peer type)
+        :param new_ip_range: new ip range
+        :param ip_ranges_to_del: ip ranges to delete
+        """
+        print_data = vrf, nbr, data
+        kwargs = {
+            'bgp_session': data,
```
anish-n, line 364: CONFIG_DB__DEVICE_METADATA should be passed in here.

src/sonic-bgpcfgd/bgpcfgd/managers_bgp.py
```diff
@@ -279,6 +290,91 @@ def apply_admin_status(self, vrf, nbr, template_name, admin_state):
...
+            'delete_ranges': ip_ranges_to_del,
```
anish-n, line 365: VRF should also be passed in

anish-n (changes requested):

src/sonic-bgpcfgd/bgpcfgd/managers_bgp.py
```diff
@@ -279,6 +290,91 @@ def apply_admin_status(self, vrf, nbr, template_name, admin_state):
...
+        }
+        try:
+            cmd = self.templates["update"].render(**kwargs)
+        except jinja2.TemplateError as e:
+            msg = "Peer '(%s|%s)'. Error in rendering the template for 'SET' command '%s'" % print_data
+            log_err("%s: %s" % (msg, str(e)))
+            return True
+        if cmd is not None:
+            self.apply_op(cmd, vrf)
+            log_info("Peer '(%s|%s)' ip range has been scheduled to be updated with new range '%s'" % (vrf, nbr, new_ip_range))
```
anish-n, line 376: are you planning to add state db from HLD in this PR, or raise a separate PR for it?

src/sonic-bgpcfgd/bgpcfgd/managers_bgp.py
```diff
@@ -104,6 +105,11 @@ def __init__(self, common_objs, db_name, table_name, peer_type, check_neig_meta)
             "shutdown":    self.fabric.from_string('neighbor {{ neighbor_addr }} shutdown'),
             "no shutdown": self.fabric.from_string('no neighbor {{ neighbor_addr }} shutdown'),
         }
+        if (os.path.exists("/usr/share/sonic/templates/" + base_template + "instance_update.conf.j2")):
+            self.templates["update"] = self.fabric.from_file(base_template + "instance_update.conf.j2")
```
anish-n, line 109: name to be changed to update.conf.j2 since update.conf is design to modify config across all 3 types: instance.conf, peer-group.conf and policies.conf

anish-n (changes requested):

src/sonic-bgpcfgd/bgpcfgd/managers_bgp.py
```diff
@@ -104,6 +105,11 @@ def __init__(self, common_objs, db_name, table_name, peer_type, check_neig_meta)
...
+            self.update_template_exists = True
```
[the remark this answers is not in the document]
sudarshankumar4893: Changed to just check the existence of template in dictionary. Thanks.

src/sonic-bgpcfgd/bgpcfgd/managers_bgp.py
```diff
@@ -242,6 +248,11 @@ def update_peer(self, vrf, nbr, data):
         """
         if "admin_status" in data:
             self.change_admin_status(vrf, nbr, data)
+        elif "ip_range" in data and self.peer_type == 'dynamic':
+            if (self.update_template_exists):
```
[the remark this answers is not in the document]
sudarshankumar4893: done. thanks.

src/sonic-bgpcfgd/bgpcfgd/managers_bgp.py
sudarshankumar4893: Done. Thanks.

src/sonic-bgpcfgd/bgpcfgd/managers_bgp.py
sudarshankumar4893: done. thanks.

src/sonic-bgpcfgd/bgpcfgd/managers_bgp.py
sudarshankumar4893: I will create a separate PR for the state DB changes from HLD

src/sonic-bgpcfgd/bgpcfgd/managers_bgp.py
sudarshankumar4893: modified the name. Thanks.

src/sonic-bgpcfgd/tests/test_bgp.py
```diff
@@ -152,6 +152,29 @@ def test_add_peer_ipv6():
...
+        res = m.set_handler("BGPSLBPassive", {"peer_asn": "65200", "ip_range": "10.250.0.0/27"})
+        assert res, "Expect True return value"
+
+def test_add_dynamic_peer_ipv6():
+    for constant in load_constant_files():
+        m = constructor(constant, peer_type="dynamic")
+        res = m.set_handler("BGPSLBPassive", {"peer_asn": "65200", "ip_range": "fc00:20::/64"})
+        assert res, "Expect True return value"
+
+def test_modify_dynamic_peer_range():
```
anish-n, line 167: i feel that checking for just a true return value provides very little coverage, can you please look into improving the coverage so that we check more of what is happening internally in the code.

anish-n (changes requested):

src/sonic-bgpcfgd/bgpcfgd/managers_bgp.py
```diff
@@ -279,6 +327,94 @@ def apply_admin_status(self, vrf, nbr, template_name, admin_state):
...
+        if vrf == 'default':
+            command = ["vtysh", "-c", "show bgp peer-group %s json" % (nbr)]
+        else:
+            command = ["vtysh", "-c", "show bgp vrf %s peer-group %s json" % (vrf, nbr)]
+        try:
+            ret_code, out, err = run_command(command)
+            if ret_code == 0:
+                js_bgp = json.loads(out)
+                if nbr in js_bgp and 'dynamicRanges' in js_bgp[nbr] and 'IPv4' in js_bgp[nbr]['dynamicRanges'] and 'ranges' in js_bgp[nbr]['dynamicRanges']['IPv4']:
+                    ipv4_ranges= js_bgp[nbr]['dynamicRanges']['IPv4']['ranges']
```
StormLiangMS, line 378: could we have space between =?

src/sonic-bgpcfgd/bgpcfgd/managers_bgp.py
```diff
@@ -279,6 +327,94 @@ def apply_admin_status(self, vrf, nbr, template_name, admin_state):
...
+                    log_info("Peer '(%s|%s)' already has ipV4 range: %s" % (vrf, nbr, ipv4_ranges))
+                if nbr in js_bgp and 'dynamicRanges' in js_bgp[nbr] and 'IPv6' in js_bgp[nbr]['dynamicRanges'] and 'ranges' in js_bgp[nbr]['dynamicRanges']['IPv6']:
+                    ipv6_ranges= js_bgp[nbr]['dynamicRanges']['IPv6']['ranges']
```
StormLiangMS, line 381: same here.

src/sonic-bgpcfgd/bgpcfgd/managers_bgp.py
```diff
@@ -279,6 +327,94 @@ def apply_admin_status(self, vrf, nbr, template_name, admin_state):
...
+        :param nbr: neighbor ip address (name for dynamic peer type)
+        :param data: associated data
+        :return: True if this adding was successful, False otherwise
+        """
+        if data['ip_range']:
+            log_info("Peer '(%s|%s)' ip range is going to be updated with range: %s" % (vrf, nbr, data['ip_range']))
+            new_ip_range = data["ip_range"].split(",")
+            ip_ranges_to_add = new_ip_range
+            ip_ranges_to_del = []
+            existing_ipv4_range, existing_ipv6_range = self.get_existing_ip_ranges(vrf, nbr)
```
StormLiangMS, line 343: so if empty returned, how we know something went wrong or it is real empty?

sudarshankumar4893: we are logging in the previous function get_existing_ip_ranges in case there is any trouble fetching the ip ranges. There is no action to do except log and return if something went wrong. Do we want to also return a status and log here as well?

src/sonic-bgpcfgd/bgpcfgd/managers_bgp.py
sudarshankumar4893: Fixed. Thanks.

src/sonic-bgpcfgd/bgpcfgd/managers_bgp.py
sudarshankumar4893: Fixed. Thanks.

src/sonic-bgpcfgd/tests/test_bgp.py
sudarshankumar4893: Added a check for log that gets emitted as well.

src/sonic-bgpcfgd/bgpcfgd/managers_bgp.py
```diff
@@ -104,6 +105,8 @@ def __init__(self, common_objs, db_name, table_name, peer_type, check_neig_meta)
             "shutdown":    self.fabric.from_string('neighbor {{ neighbor_addr }} shutdown'),
             "no shutdown": self.fabric.from_string('no neighbor {{ neighbor_addr }} shutdown'),
         }
+        if (os.path.exists("/usr/share/sonic/templates/" + base_template + "update.conf.j2")):
```
anish-n, line 108: we should use the path predefined in template.py vs hardcoding. Refer to https://github.com/sonic-net/sonic-buildimage/pull/22229 on how to do it

src/sonic-bgpcfgd/bgpcfgd/managers_bgp.py
```diff
@@ -227,10 +230,53 @@ def add_peer(self, vrf, nbr, data):
...
+            key = nbr
+        else:
+            key = vrf + "|" + nbr
+        # Update the peer in the STATE_DB table
+        try:
+            state_db = swsscommon.DBConnector("STATE_DB", 0)
+            state_peer_table = swsscommon.Table(state_db, swsscommon.STATE_BGP_PEER_CONFIGURED_TABLE_NAME)
+            data_to_update = {}
+            if (op == "SET"):
+                if ("ip_range" in data and self.peer_type == 'dynamic'):
```
anish-n, line 257: why do we need these explicit checks and set the same value, could we just directly write the data?

src/sonic-bgpcfgd/tests/test_bgp.py
```diff
@@ -162,7 +164,40 @@ def test_add_peer_ipv6_in_vnet():
...
+
+        #Add a new range to the existing config.
+        res = m.set_handler("DynNbr1", {"peer_asn": "65200", "ip_range": "10.255.0.0/24,10.255.1.0/24"})
+        assert res, "Expect True return value"
+        mocked_log_info.assert_called_with("Peer '(default|DynNbr1)' ip range has been scheduled to be updated with range '10.255.0.0/24,10.255.1.0/24'")
+
+        #Modify one of the IP ranges to decrease the subnet size.
+        res = m.set_handler("DynNbr1", {"peer_asn": "65200", "ip_range": "10.255.0.0/26"})
+        assert res, "Expect True return value"
+        mocked_log_info.assert_called_with("Peer '(default|DynNbr1)' ip range has been scheduled to be updated with range '10.255.0.0/26'")
```
anish-n, line 195: UT should check for state db update

src/sonic-bgpcfgd/tests/test_bgp.py
```diff
@@ -162,7 +164,40 @@ def test_add_peer_ipv6_in_vnet():
...
+def test_modify_dynamic_peer_range(mocked_log_info):
+    for constant in load_constant_files():
+        m = constructor(constant, peer_type="dynamic")
+        base_template = "bgpd/templates/" + m.constants["bgp"]["peers"]["dynamic"]["template_dir"] + "/"
+        m.templates["update"] = m.fabric.from_file(base_template + "update.conf.j2")
+
+        #Add a new range to the existing config.
+        res = m.set_handler("DynNbr1", {"peer_asn": "65200", "ip_range": "10.255.0.0/24,10.255.1.0/24"})
+        assert res, "Expect True return value"
+        mocked_log_info.assert_called_with("Peer '(default|DynNbr1)' ip range has been scheduled to be updated with range '10.255.0.0/24,10.255.1.0/24'")
```
anish-n, line 190: are we certain that we cannot test commands called with ```self.cfg_mgr.push(cmd)```, we should ideally check that the required update occurred by checking the final output, example where this is being done for push_list, push should be similar though I suppose. https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-bgpcfgd/tests/test_bbr.py . If this is not possible/difficult to do then the log ```ip range has been scheduled``` should be updated to include the added range and deleted range vs passing in the new range. That way you can test if your logic for detecting range changes is working e2e.

StormLiangMS (approved):
LGTM

StormLiangMS:
hi @sudarshankumar4893 I would like to have the test case in sonic-mgmt repo to test this new feature, do we have an ETA we can have test cases in sonic-mgmt repo done?

src/sonic-bgpcfgd/bgpcfgd/managers_bgp.py
sudarshankumar4893: Fixed it. Thanks.

src/sonic-bgpcfgd/tests/test_bgp.py
sudarshankumar4893: Added the check for the generated commands for cfg_mgr.push(). Validating that it gets called with the expected commands when adding/deleting or modifying an IP range for dynamic neighbor.

src/sonic-bgpcfgd/bgpcfgd/managers_bgp.py
sudarshankumar4893: Thanks for clarifying. Writing the whole data now to the db.

StormLiangMS (approved):
LGTM

StormLiangMS:
hi @sudarshankumar4893 
Could you check the build error? Seems related the PR change. 

https://dev.azure.com/mssonic/build/_build/results?buildId=839050&view=logs&jobId=e2a1d004-6151-514d-dd3b-ba225f574bd9&j=e2a1d004-6151-514d-dd3b-ba225f574bd9&t=84643f3d-b87b-5e33-4d3a-7ffd9da7d441&s=66ecdfa4-8cac-587d-af5a-a4218cba5b68

src/sonic-bgpcfgd/bgpcfgd/managers_bgp.py
```diff
@@ -279,6 +321,94 @@ def apply_admin_status(self, vrf, nbr, template_name, admin_state):
...
+        }
+        try:
+            cmd = self.templates["update"].render(**kwargs)
+        except jinja2.TemplateError as e:
+            msg = "Peer '(%s|%s)'. Error in rendering the template for 'SET' command '%s'" % print_data
+            log_err("%s: %s" % (msg, str(e)))
+            return True
+        if cmd is not None:
+            self.apply_op(cmd, vrf)
+            self.update_state_db(vrf, nbr, data, "SET")
```
anish-n, line 409: note since we have added the state db logic here, the other update operation will not result in modification of state db. ie: If admin status is changed state db won't be updated with that

sudarshankumar4893: Added the update of state db after admin state change as well.

anish-n (approved):

StormLiangMS (approved):
LGTM

PR #22297, opened 2025-04-10
Moved the FRR docker_routing_config_mode value to unified by default

vidyac86:
#### Why I did it
The docker_routing_config_mode value which is set to default of separated is not working properly with the latest FRR.

#### How I did it
The changes are done in the docker_init.sh to modify the default behavior of considering the docker_routing_config_mode value as 'separated' when it is not seen in the config_db to modifying the value to 'unified'.

#### How to verify it
These values are significant to restore the values in FRR. When the value is separated, each process in FRR will have its own config file from where the configs are played back after save and restore, with the value of 'unified' there will be only one file frr.conf from where all the configs will be played back.

After doing this change, I manually verified the presence of frr.conf after save and restore and all the configs are restored properly.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

master

#### Tested image version
20250407.145357

Modified the default value of docker_routing_config_mode to unified.

UT Logs
----------

[UT_Logs.txt](https://github.com/user-attachments/files/19689684/UT_Logs.txt)


venkatmahalingam:
@yxieca Please review the changes.

dockers/docker-fpm-frr/docker_init.sh
```diff
@@ -56,7 +56,7 @@ if [[ ! -z "$NAMESPACE_ID" ]]; then
    update_default_gw 6
 fi
 
-if [ -z "$CONFIG_TYPE" ] || [ "$CONFIG_TYPE" == "separated" ]; then
+if  [ "$CONFIG_TYPE" == "separated" ]; then
```
dgsudharsan, line 59: Should we completely remove "separated" mode? I believe with db_migrator, any old configuration will be migrated to unified.
Keeping separated would result in "non integrated" mode in vtysh which wouldn't be supported by FRR community going forward.

dgsudharsan (changes requested):

PR #22299, opened 2025-04-10
update memory monit threshold for bmp container according to test.

FengPan-Frank:
#### Why I did it
BMP feature is enabled by-default currently, but there's config cli to enable dataset dump into redis-bmp dynamically. Previously we added test https://github.com/sonic-net/sonic-mgmt/pull/16854, added memory checker and thought that's the extremely case which will hit peak usage for both bgp and bmp memory. By recent Nightly, there're other cases like nhop and test_pfc_counters which reports that bmp exceeds the pre-defined monit threshold 400MB from time to time, and usually unable to be reproduced stably. 
Thus we need to update the monit memory threshold for bmp container based on the recent Nightly test result, meanwhile will turn on bmp for cover all existing test cases so that we can measure whether the threshold value makes sense.

##### Work item tracking
- Microsoft ADO **(number only)**:32246488

#### How I did it
Updated the memory monit threshold based on kusto queries, and turn on bmp as pre-condition step https://github.com/sonic-net/sonic-mgmt/pull/17963

#### How to verify it
need observe some days via Nightly on master.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/dockers/docker-sonic-bmp/base_image_files/monit_bmp
+++ b/dockers/docker-sonic-bmp/base_image_files/monit_bmp
@@ -1,5 +1,5 @@
 ###############################################################################
 ## Monit configuration for bmp container
 ###############################################################################
-check program container_memory_bmp with path "/usr/bin/memory_checker bmp 419430400"
+check program container_memory_bmp with path "/usr/bin/memory_checker bmp 838860800"
     if status == 3 for 10 times within 20 cycles then exec "/usr/bin/docker exec bmp supervisorctl restart openbmpd"


xwjiang-ms:
@FengPan-Frank could you explain why bmp container would exceeded the memory threshold in PR test?

FengPan-Frank:
> @FengPan-Frank could you explain why bmp container would exceeded the memory threshold in PR test?

pls refer description of this PR, previously we only measured bgp flap case as extreme stress test, monit threshold may not precise. seems nhop/pfc_counter also adds stress but I checked kusto it's only hits memory from time to time, most runs are good. I'll enable bmp before all test cases as pre-condition via https://github.com/sonic-net/sonic-mgmt/pull/17963, and we can observe some days for the master run.

qiluo-msft (approved):

StormLiangMS:
hi @qiluo-msft @FengPan-Frank  is that 800MB too much? It could be an issue for device which has 8G memory.

PR #22307, opened 2025-04-11
[sonic-mgmt-docker-image] Add packet mask counters in ptf dataplane

w1nda:
#### Why I did it
We need to count packets we care about on ptf dataplane.

##### Work item tracking

#### How I did it
Add a specific counter in ptf datplane, user can create counter with packet mask, so ptf dataplane can count packets that matching specific mask.

#### How to verify it
Tested on testbed
Sent 5 packets for each 100 ports and read counters.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/dockers/docker-sonic-mgmt/0003-add-dataplane-mask-counters-to-avoid-dataplane-noise.patch
+++ b/dockers/docker-sonic-mgmt/0003-add-dataplane-mask-counters-to-avoid-dataplane-noise.patch
@@ -1,0 +1,57 @@
+From a1e0c25218b5bac0adeebcbd95b29618b4cadf3d Mon Sep 17 00:00:00 2001
+From: wenda <wendachu@microsoft.com>
+Date: Fri, 11 Apr 2025 09:57:23 +0000
+Subject: [PATCH] 0003-add-dataplane-mask-counters-to-avoid-dataplane-noise
+
+---
+ src/ptf/dataplane.py | 17 +++++++++++++++++
+ 1 file changed, 17 insertions(+)
+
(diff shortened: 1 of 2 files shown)


dockers/docker-sonic-mgmt/0003-add-dataplane-mask-counters-to-avoid-dataplane-noise.patch
```diff
@@ -0,0 +1,57 @@
...
+index a5d4664..0aee7ab 100644
+--- a/src/ptf/dataplane.py
++++ b/src/ptf/dataplane.py
+@@ -574,6 +574,12 @@ class DataPlane(Thread):
+         # counters of transmited packets
+         self.tx_counters = defaultdict(int)
+ 
++        # rx/tx counters of packets that matching the packet mask
++        # the key of dict is mask
++        # the value is a dict of device number, port number to count
```
r12f, line 20: Will this filter be enough? Do we need ability to check the packet type and etc?

w1nda: We can filter packets by type, header, payload, or any byte in any position.

By leverage the mask class in ptf: https://github.com/p4lang/ptf/blob/main/src/ptf/mask.py


and my test code will use this feature to filter and **count** icmpv6 packets with special icmp type:
image

r12f (approved):

yxieca (approved):

PR #22308, opened 2025-04-11
SSH hardening configuration options

bhouse-nexthop:
#### Why I did it
The SSH configuration does not contain many of the hardening requirements by the various standards bodies.

This PR depends on https://github.com/sonic-net/sonic-host-services/pull/XXX

#### How I did it

 This adds support for:
 * password_authentication - ability to disable password auth
 * permit_root_login - ability to prevent root logins
 * ciphers - ability to specify available ciphers
 * kex_algorithms - ability to specify key exchange algorithms
 * macs - ability to specify macs
 
#### How to verify it

sonic-yang-models runs tests during build, as does sonic-host-services which validates the behavior.

- [ ] 202411

master as of 20250410

SSH hardening configuration options

https://github.com/bhouse-nexthop/sonic-buildimage/blob/bhouse-nexthop/ssh-config/src/sonic-yang-models/doc/Configuration.md#ssh_server

### DIFF
--- a/src/sonic-yang-models/yang-models/sonic-ssh-server.yang
+++ b/src/sonic-yang-models/yang-models/sonic-ssh-server.yang
@@ -59,6 +59,70 @@ module sonic-ssh-server {
                         range 0..100;
                     }
                 }
+                leaf permit_root_login {
+                    description "Specifies whether root can log in using ssh.";
+                    type boolean;
+                }
+                leaf password_authentication {
+                    description "Specifies whether password authentication is enabled.";
+                    type boolean;
+                    default true;
+                }
+                leaf-list ciphers {
+                    description "Specifies the ciphers allowed.";
+                    type enumeration {
+                        enum "3des-cbc";
+                        enum "aes128-cbc";
+                        enum "aes192-cbc";
+                        enum "aes256-cbc";
+                        enum "aes128-ctr";
+                        enum "aes192-ctr";
+                        enum "aes256-ctr";
+                        enum "aes128-gcm@openssh.com";
+                        enum "aes256-gcm@openssh.com";
+                        enum "chacha20-poly1305@openssh.com";
+                    }
+                }
+                leaf-list kex_algorithms {
+                    description "Specifies the available Key Exchange algorithms.";
+                    type enumeration {
+                        enum "diffie-hellman-group1-sha1";
+                        enum "diffie-hellman-group14-sha1";
+                        enum "diffie-hellman-group14-sha256";
+                        enum "diffie-hellman-group16-sha512";
+                        enum "diffie-hellman-group18-sha512";
+                        enum "diffie-hellman-group-exchange-sha1";
+                        enum "diffie-hellman-group-exchange-sha256";
+                        enum "ecdh-sha2-nistp256";
+                        enum "ecdh-sha2-nistp384";
+                        enum "ecdh-sha2-nistp521";
+                        enum "curve25519-sha256";
+                        enum "curve25519-sha256@libssh.org";
+                        enum "sntrup761x25519-sha512";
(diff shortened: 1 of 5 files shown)


src/sonic-yang-models/yang-models/sonic-ssh-server.yang
```diff
@@ -59,6 +59,70 @@ module sonic-ssh-server {
                         range 0..100;
                     }
                 }
+                leaf permit_root_login {
+                    description "Specifies whether root can log in using ssh.";
+                    type boolean;
```
jeff-yin, line 64: should we default to `false` here?

jeff-yin:
The original issue states:
> The current SONiC SSH default configuration will not pass most security scanners

Are we updating the SSH default config? I don't see code in this PR or https://github.com/sonic-net/sonic-host-services/pull/238 that indicates changes to the default SSH configs.

If we need to update the default cipher/kex/mac algorithms, it may be helpful to highlight the exact "hardening requirements by the various standards bodies" we're trying to satisfy. 

Or are we just going with the existing OpenSSH defaults (i.e., whatever settings are done when no value is specified for the given setting)?

bhouse-nexthop:
No defaults are updated at all.  Just enable the configuration knobs so that a user *can* comply with whatever security standard they're using.  I took the approach of not potentially breaking users.

src/sonic-yang-models/yang-models/sonic-ssh-server.yang
bhouse-nexthop: actually, this shouldn't be a boolean, good catch.  The official possible values are:

- yes
- prohibit-password
- forced-commands-only
- no

The default is prohibit-password.  I'll modify this to support all the options.  That said, only a value of 'no'/'false' is acceptable from a security standpoint.

I'll update these PRs to better represent this field.

bhouse-nexthop: permit_root_login is now an enum with the appropriate values and tests

jeff-yin (approved):

bhouse-nexthop:
I just rebased against latest master to force a rebuild since there were spurious failures.

bradh352:
I rebased again to force tests to run, they are all passing now.

src/sonic-yang-models/yang-models/sonic-ssh-server.yang
```diff
@@ -59,6 +59,75 @@ module sonic-ssh-server {
                         range 0..100;
                     }
                 }
+                leaf permit_root_login {
+                    description "Specifies whether root can log in using ssh.";
+                    type enumeration {
+                        enum "yes";
+                        enum "prohibit-password";
```
qiluo-msft, line 66: >enum [](http://example.com/codeflow?start=24&length=4)

Do you have a default value? Is it backward-compatible? Same questions for all the fields added in this PR. #Closed

bhouse-nexthop: The default value is "prohibit-password" in modern versions of openssh.  Not specifying this key will leave it at whatever the default is or whatever the user may have set it to in the configuration manually.

prabhataravind:
@bhouse-nexthop could you fix conflicts on this so @qiluo-msft can merge?

qiluo-msft:
@bhouse-nexthop This branch has conflicts that must be resolved

bhouse-nexthop:
@qiluo-msft merge conflicts have been resolved

bradh352:
another spurious test failure unrelated to the changes.  Rebased to force a rebuild.

bhouse-nexthop:
@lguohan this is the yang model update for the ticket you just merged :)

bhouse-nexthop:
@lguohan can this be merged?

PR #22310, opened 2025-04-11
DNS configuration option enhancements

### LINKED ISSUE #22311: Enhancement: Additional DNS configuration options
### Is it platform specific

generic

### Importance or Severity

Medium

Common DNS configuration options are missing from SONiC, we need to add these to support more deployments.

### Current Behavior

Missing additional configuration knobs

### Proposed Behavior

Add knobs for:
 * search
 * ndots
 * timeout
 * attempts

bhouse-nexthop:
#### Why I did it

Some needed DNS configuration options are not available such as search suffixes, ndots, query timeout (s), and query attempts.

Also depends on https://github.com/sonic-net/sonic-host-services/pull/239
Fixes #22311

#### How I did it

Added yang models, tests, and documentation.

#### How to verify it

During build it will run through the test suite.

- [ ] 202411

master as of 20250410

DNS configuration option enhancements

https://github.com/bhouse-nexthop/sonic-buildimage/blob/bhouse-nexthop/dns-config/src/sonic-yang-models/doc/Configuration.md#static-dns

### DIFF
--- a/src/sonic-yang-models/yang-models/sonic-dns.yang
+++ b/src/sonic-yang-models/yang-models/sonic-dns.yang
@@ -40,6 +40,41 @@ module sonic-dns {
 
         } /* end of container DNS_NAMESERVER */
 
+        container DNS_OPTIONS {
+            description "DNS_OPTIONS requires at least one DNS_NAMESERVER to be set.";
+
+            leaf-list search {
+                description "Configure the DNS search suffix list";
+                type inet:host;
+            }
+
+            leaf ndots {
+                description "Sets a threshold for the number of dots which must appear in a name given before an initial absolute query will be made";
+                type uint8 {
+                    range "0..15";
+                }
+                default 1;
+            }
+
+            leaf timeout {
+                description "Sets the amount of time in seconds the resolver will wait for a response from a remote name server before retrying the query via a different name server.";
+                type uint8 {
+                    range "1..30";
+                }
+                default 5;
+            }
+
+            leaf attempts {
+                description "Sets the number of times the resolver will send a query to its name servers before giving up and returning an error to the calling application.";
+                type uint8 {
+                    range "1..5";
+                }
+                default 2;
+            }
+
+            when "count(../DNS_NAMESERVER/DNS_NAMESERVER_LIST/ip) > 0";
+        } /* end of container DNS_OPTIONS */
+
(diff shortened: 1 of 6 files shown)


bhouse-nexthop:
@qiluo-msft can you review the PRs associated with this issue?  Its pretty simple.

src/sonic-yang-models/yang-models/sonic-dns.yang
```diff
@@ -40,6 +40,41 @@ module sonic-dns {
...
+
+            leaf-list search {
+                description "Configure the DNS search suffix list";
+                type inet:host;
+            }
+
+            leaf ndots {
+                description "Sets a threshold for the number of dots which must appear in a name given before an initial absolute query will be made";
+                type uint8 {
+                    range "0..15";
```
qiluo-msft, line 54: How do you figure out the range and default values?

bhouse-nexthop: from `man resolv.conf`:
```
              ndots:n
                     Sets a threshold for the number of dots which must appear in a name given to res_query(3) (see resolver(3)) before an initial  absolute
                     query  will be made.  The default for n is 1, meaning that if there are any dots in a name, the name will be tried first as an absolute
                     name before any search list elements are appended to it.  The value for this option is silently capped to 15.

              timeout:n
                     Sets the amount of time the resolver will wait for a response from a remote name server before retrying the query via a different  name
                     server.   This  may not be the total time taken by any resolver API call and there is no guarantee that a single resolver API call maps
                     to a single timeout.  Measured in seconds, the default is RES_TIMEOUT (currently 5, see <resolv.h>).  The  value  for  this  option  is
                     silently capped to 30.

              attempts:n
                     Sets  the number of times the resolver will send a query to its name servers before giving up and returning an error to the calling ap‐
                     plication.  The default is RES_DFLRETRY (currently 2, see <resolv.h>).  The value for this option is silently capped to 5.
```

bradh352:
Thanks for merging, please don't forget to merge the sister ticket to actually implement the functionality since the yang model alone doesn't do anything :    https://github.com/sonic-net/sonic-host-services/pull/239

I also updated the HLD here for documentation purposes: https://github.com/sonic-net/SONiC/pull/1975

PR #22342, opened 2025-04-16
[TACACS] Fix TACACS config revert to old config when device reboot issue.

liuh-80:
Fix TACACS config revert to old config when device reboot issue.

#### Why I did it
When SONiC TACACS config changed after OS upgrade, and device reboot, the TACACS config been reverted back to old config in /etc/sonic/old_config/tacacs.json

##### Work item tracking
- Microsoft ADO **(number only)**:32338799

#### How I did it
Move /etc/sonic/old_config/tacacs.json to /etc/sonic/old_config/tacacs.json_backup

#### How to verify it
Pass all test case.
Manually verify.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ] 

Fix TACACS config revert to old config when device reboot issue.

### DIFF
--- a/files/image_config/config-setup/config-setup
+++ b/files/image_config/config-setup/config-setup
@@ -139,6 +139,9 @@ apply_tacacs()
         sonic-cfggen -j /etc/sonic/old_config/${TACACS_JSON_BACKUP} --write-to-db
         echo "Applied tacacs json to restore tacacs credentials"
         config save -y
+
+        # Change tacacs config file name, so tacacs config will not be revert when device reboot.
+        mv /etc/sonic/old_config/${TACACS_JSON_BACKUP}_backup
     else
         echo "Missing tacacs json to restore tacacs credentials"
     fi


files/image_config/config-setup/config-setup
```diff
@@ -139,6 +139,9 @@ apply_tacacs()
         sonic-cfggen -j /etc/sonic/old_config/${TACACS_JSON_BACKUP} --write-to-db
         echo "Applied tacacs json to restore tacacs credentials"
         config save -y
+
+        # Change tacacs config file name, so tacacs config will not be revert when device reboot.
```
qiluo-msft, line 143: >Change tacacs config file name, so tacacs config will not be revert when device reboot [](http://example.com/codeflow?start=10&length=86)

Should we just remove tacacs-config.service completely and simplified tacacs just as simple as part of the config_db.json?

liuh-80: Complete remove the service is danger, because remove this service may cause SONiC missing TACACS config.
The OS upgrade scenario does not cover by sonic-mgmt test.

qiluo-msft (approved):

qiluo-msft:
> apply_tacacs()

Actually we should check running config first, if running config already has TACACS key, no need to read old_config file at all.

---
Refers to: files/image_config/config-setup/config-setup:136 in 5407540. [](commit_id = 54075404fff32bacb42f2d22a1700f6254f12936, deletion_comment = False)

Issue #22344, opened 2025-04-16
Bug: staticroutebfd crashes on static blackhole route configuration


Yakiv-Huryk:
### Is it platform specific

generic

### Importance or Severity

High

### Description of the bug

When adding a static route type blackhole to the config db, staticroutebfd crashes:

```
INFO bgp#supervisord: staticroutebfd Traceback (most recent call last):
INFO bgp#supervisord: staticroutebfd   File "/usr/local/bin/staticroutebfd", line 8, in <module>
INFO bgp#supervisord: staticroutebfd     sys.exit(main())
INFO bgp#supervisord: staticroutebfd              ^^^^^^
INFO bgp#supervisord: staticroutebfd   File "/usr/local/lib/python3.11/dist-packages/staticroutebfd/main.py", line 768, in main
INFO bgp#supervisord: staticroutebfd     do_work()
INFO bgp#supervisord: staticroutebfd   File "/usr/local/lib/python3.11/dist-packages/staticroutebfd/main.py", line 760, in do_work
INFO bgp#supervisord: staticroutebfd     sr_bfd.run()
INFO bgp#supervisord: staticroutebfd   File "/usr/local/lib/python3.11/dist-packages/staticroutebfd/main.py", line 747, in run
INFO bgp#supervisord: staticroutebfd     self.reconciliation()
INFO bgp#supervisord: staticroutebfd   File "/usr/local/lib/python3.11/dist-packages/staticroutebfd/main.py", line 288, in reconciliation
INFO bgp#supervisord: staticroutebfd     self.static_route_set_handler(key_new, data)
INFO bgp#supervisord:  File "/usr/local/lib/python3.11/dist-packages/staticroutebfd/main.py", line 415, in static_route_set_handler
INFO bgp#supervisord: staticroutebfd     nh_vrf_list = [vrf] * len(nh_list) if len(nh_list) > 0 else None
INFO bgp#supervisord: staticroutebfd                                           ^^^^^^^^^^^^
INFO bgp#supervisord: staticroutebfd TypeError: object of type 'NoneType' has no len()
```

### Steps to Reproduce

```bash
sonic-db-cli CONFIG_DB HSET STATIC_ROUTE\|default\|fcbb:bbbb::/32 blackhole true
```

### Actual Behavior and Expected Behavior

No crash, the route is configured successfully

### Relevant log output

```shell

```

### Output of `show version`, `show techsupport`

```shell

```

[...]

PR #22347, opened 2025-04-16
Fix staticroutebfd's handling on blackhole static routes and static routes that only have partial info correctly

BYGX-wcr:
#### Why I did it
Previously, the staticroutebfd module assume that static routes always have both ifname and nexthop configured, which is not true. First, a static route can have only either ifname or nexthop. Second, for blackhole routes, usually it is not configured with any ifname or nexthop. The staticroutebfd will encounter runtime error when those static routes are configured even if bfd is not enabled.

Fixes issue: #22344

##### Work item tracking

#### How I did it
Change the staticroutebfd's implementation to skip the processing if the route is a blackhole route or it only has partial info. I also added UT to verify it.

#### How to verify it

* Run UT.
* Test the new implementation on a SONiC devie.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/src/sonic-bgpcfgd/tests/test_static_rt_bfd.py
+++ b/src/sonic-bgpcfgd/tests/test_static_rt_bfd.py
@@ -114,7 +114,7 @@ def intf_setup(dut):
         ),
         {},
         {}
-    )    
+    )
 
 def test_set_del_ipv6():
     dut = constructor()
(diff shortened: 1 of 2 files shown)


src/sonic-bgpcfgd/tests/test_static_rt_bfd.py
```diff
@@ -695,4 +695,61 @@ def test_set_bfd_change_no_hold():
     )
 
 
+def test_set_del_blackhole_route():
+    dut = constructor()
+    intf_setup(dut)
+
+    test_set_del_blackhole_route.logs = []
+
+    _log_info = log_info
```
Yakiv-Huryk, line 704: Please consider the `@patch('staticroutebfd.main.log_info')` instead

BYGX-wcr: fixed.

prsunny (approved):

PR #22361, opened 2025-04-17
Add admin status to loopback interface yang model

Leyza:
#### Why I did it

Loopback interface is missing admin status therefore `config interface startup/shutdown` does not work for it. Admin status has been added to loopback interface in the below PRs. This PR adds it to the corresponding yang model.

Related PRs: https://github.com/sonic-net/sonic-swss/pull/3565 ; https://github.com/sonic-net/sonic-utilities/pull/3827

##### Work item tracking

#### How I did it

Added admin_status to loopback interface for yang models.

#### How to verify it

image

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/src/sonic-yang-models/yang-models/sonic-loopback-interface.yang
+++ b/src/sonic-yang-models/yang-models/sonic-loopback-interface.yang
@@ -58,6 +58,7 @@ module sonic-loopback-interface {
 
                 leaf admin_status {
                     type stypes:admin_status;
+                    default up;
                 }
             }
             /* end of LOOPBACK_INTERFACE_LIST */
--- a/src/sonic-yang-models/yang-models/sonic-loopback-interface.yang
+++ b/src/sonic-yang-models/yang-models/sonic-loopback-interface.yang
@@ -55,6 +55,10 @@ module sonic-loopback-interface {
                     }
                     default "0";
                 }
+
+                leaf admin_status {
+                    type stypes:admin_status;
+                }
             }
             /* end of LOOPBACK_INTERFACE_LIST */
 
--- a/src/sonic-yang-models/yang-models/sonic-loopback-interface.yang
+++ b/src/sonic-yang-models/yang-models/sonic-loopback-interface.yang
@@ -58,7 +58,6 @@ module sonic-loopback-interface {
 
                 leaf admin_status {
                     type stypes:admin_status;
-                    default up;
                 }
             }
             /* end of LOOPBACK_INTERFACE_LIST */
--- a/src/sonic-yang-models/tests/files/sample_config_db.json
+++ b/src/sonic-yang-models/tests/files/sample_config_db.json
@@ -1311,7 +1311,8 @@
         },
         "LOOPBACK_INTERFACE": {
             "Loopback0": {
-               "nat_zone": "2"
+               "nat_zone": "2",
+               "admin_status": "up"
             },
             "Loopback0|2a04:5555:40:4::4e9/128": {
                 "scope": "global",
--- a/src/sonic-yang-models/tests/yang_model_tests/tests/loopback.json
+++ b/src/sonic-yang-models/tests/yang_model_tests/tests/loopback.json
@@ -13,5 +13,14 @@
     "LOOPBACK_INVALID_INTERFACE_NAME_LENGTH": {
         "desc": "Configure invalid value for loopback interface name length.",
         "eStr" : "Invalid interface name length, it must not exceed 16 characters."
+    },
+    "LOOPBACK_DEFAULT_ADMIN_STATUS_UP": {
+        "desc": "Verify admin status is default up for loopback interface",
+        "eStrKey": "Verify",
+        "verify": {
+            "xpath": "/sonic-loopback-interface:sonic-loopback-interface/LOOPBACK_INTERFACE/LOOPBACK_INTERFACE_LIST[name='lo1']/name",
+            "key": "sonic-loopback-interface:admin_status",
+            "value": "up"
+        }
     }
 }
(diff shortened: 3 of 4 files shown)


src/sonic-yang-models/yang-models/sonic-loopback-interface.yang
```diff
@@ -55,6 +55,11 @@ module sonic-loopback-interface {
                     }
                     default "0";
                 }
+
+                leaf admin_status {
+                    type stypes:admin_status;
+                    default up;
```
anish-n, line 61: Can we just keep it as a optional field with no default value. I believe if you put a default here, it will actually write that to config_db. In the backend SWSS code we allow for no value to be present for admin_status and make it default up in the backend directly. So it may be better to keep this as optional+no value at all, thereby making it 100% backwards compatible.

Leyza: Updated to allow admin_status to be optional.

wen587:
Is `admin_status` a mandatory field? What would be default behavior if not set?

anish-n:
> Why do you want to shutdown loopback interface? Could you please provide any scenario?
We have a need to withdraw advertisement for a loopback IP in some conditions, but don't want to touch the BGP network statements, in those conditions we will shutdown the loopback interface, resulting in the removal of the interface IP from the kernel route table, and BGP not advertising the IP address as a result.

anish-n:
> Is `admin_status` a mandatory field? What would be default behavior if not set?

admin_status is optional, the default behavior if it is not set is to consider the interface as up https://github.com/sonic-net/sonic-swss/pull/3565

anish-n (approved):

ganglyu:
> We have a need to withdraw advertisement for a loopback IP in some conditions, but don't want to touch the BGP network statements, in those conditions we will shutdown the loopback interface, resulting in the removal of the interface IP from the kernel route table, and BGP not advertising the IP address as a result.

Is there any side effect? For example, lldp does not work at the same time

prsunny (dismissed):
@dgsudharsan for viz

anish-n:
> > We have a need to withdraw advertisement for a loopback IP in some conditions, but don't want to touch the BGP network statements, in those conditions we will shutdown the loopback interface, resulting in the removal of the interface IP from the kernel route table, and BGP not advertising the IP address as a result.
> 
> Is there any side effect? For example, lldp does not work at the same time

Thats something which the user of the feature needs to be aware of, if the user shuts off loopback0 for instance then all connectivity patterns associated with Loopback0 will be impacted, note it is quite possible for a user to shutdown Loopback today as well but needs to be done with a Linux Kernel command(```ip link set LoopbackX down```), we are only moving the ability to do this to config DB and SONiC CLIs(enabling persistence for the setting as well). In our use case we have a different loopback(not Loopback0), thereby allowing us to keep the rest of the TOR functionality to keep working.

prsunny:
> > > We have a need to withdraw advertisement for a loopback IP in some conditions, but don't want to touch the BGP network statements, in those conditions we will shutdown the loopback interface, resulting in the removal of the interface IP from the kernel route table, and BGP not advertising the IP address as a result.
> > 
> > 
> > Is there any side effect? For example, lldp does not work at the same time
> 
> Thats something which the user of the feature needs to be aware of, if the user shuts off loopback0 for instance then all connectivity patterns associated with Loopback0 will be impacted, note it is quite possible for a user to shutdown Loopback today as well but needs to be done with a Linux Kernel command(`ip link set LoopbackX down`), we are only moving the ability to do this to config DB and SONiC CLIs(enabling persistence for the setting as well). In our use case we have a different loopback(not Loopback0), thereby allowing us to keep the rest of the TOR functionality which depends on Loopback0 to keep functioning.

agree with @anish-n , there is no change to existing behavior. this is extending the config support to user.

src/sonic-yang-models/yang-models/sonic-loopback-interface.yang
```diff
@@ -55,6 +55,10 @@ module sonic-loopback-interface {
                     }
                     default "0";
                 }
+
+                leaf admin_status {
+                    type stypes:admin_status;
```
qiluo-msft, line 60: >admin_status [](http://example.com/codeflow?start=32&length=12)

Should we define the default value for admin_status?

Leyza: We decided to not have a default value so it is backwards compatible> 
image

qiluo-msft: It is not true that "In the backend SWSS code we allow for no value to be present for admin_status and make it default up in the backend directly".

"In the backend SWSS code we allow for no value to be present for admin_status and make it default up in the backend directly". This is exactly the point of a default value. If backend behave as there is a default value "up", yang model should 
explicitly define a default value.

Leyza: I see. @prsunny , do you have any thoughts on which design would be more preferred? Whether admin_status field in yang models should be default up or optional?

ganglyu (approved):

wen587 (approved):

src/sonic-yang-models/yang-models/sonic-loopback-interface.yang
Leyza: Hi @qiluo-msft , we've decided to go with your suggestion and use default up for the admin_status.

PR #22439, opened 2025-04-24
[resolv-config] Improve container resolv.conf update mechanism

oleksandrivantsiv:
#### Why I did it
Fix a possible race condition during the config reload caused by the concurrent restart of the Docker containers and the resolv-config service. The update of the DNS configuration inside the container is triggered by the container's post-start action.

##### Work item tracking

#### How I did it
- Add support for single container updates with container name argument
- Implement parallel updates for bulk operations
- Add comprehensive logging with syslog integration
- Add container state handling (start/wait for stopped containers)
- Add proper error handling and status reporting
- Remove temporary file usage during resolv.conf updates
- Add networking service check for bulk updates only

Integration changes:
- Add automatic DNS update call in docker_image_ctl post-start action

#### How to verify it
Run sonic-mgmt DNS tests

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/files/image_config/resolv-config/update-containers
+++ b/files/image_config/resolv-config/update-containers
@@ -1,12 +1,120 @@
 #!/bin/bash
 
+SCRIPT_NAME=$(basename "$0")
+DEBUG=${DEBUG:-false}
+WAIT_TIMEOUT=5  # 5 seconds timeout for waiting containers to start
+
+# Function to log messages to syslog
+log_message() {
+    local level=$1
+    local message=$2
+    local caller=${FUNCNAME[1]}
+    # Skip debug messages if DEBUG is not true
+    if [[ "$level" == "debug" && "$DEBUG" != "true" ]]; then
+        return
+    fi
+    logger -t "resolv-config" -p "user.${level}" "[${SCRIPT_NAME}:${caller}] ${message}"
+}
+
+# Function to wait for container to start
+# Used only when updating a specific container that is not running
+wait_for_container() {
+    local container=$1
+    local start_time=$(date +%s)
+    local container_name=$(docker inspect --format '{{.Name}}' ${container} | sed 's/^\///')
+    while [[ $(($(date +%s) - start_time)) -lt $WAIT_TIMEOUT ]]; do
+        if docker inspect --format '{{.State.Status}}' ${container} | grep -q "running"; then
+            log_message "info" "Container ${container_name} (${container}) is now running"
+            return 0
+        fi
+        sleep 1
+    done
+    return 1
+}
+
+# Function to update resolv.conf for a single container
+# Parameters:
+#   $1: container ID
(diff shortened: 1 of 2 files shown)


vivekrnv (approved):

qiluo-msft:
How does "waits up to 5 seconds for containers to start" help solving a possible race condition? Is it just reduce the possibility but not completely resolve it?

Can we add a testcase to prevent future regression?

ganglyu:
Is it possible to add a hook to run the update_container script after Docker starts?
Can we use a timer to run the update_container script?

oleksandrivantsiv:
@qiluo-msft, @ganglyu I updated the implementation to remove 5seconds wait period. Please review

qiluo-msft:
Did you ever try this easier solution https://unix.stackexchange.com/a/348406/161486?

files/image_config/resolv-config/update-containers
```diff
@@ -1,12 +1,120 @@
...
+        log_message "error" "Container with name '${container_name}' not found"
+        exit 1
+    fi
+    log_message "info" "Updating resolv.conf for container ${container_name}"
+    # For single container updates, attempt to start and wait if container is stopped
+    update_container_resolv "$container_id" "true"
+    exit $?
+fi
+
+# Check if networking service is active (only for bulk updates)
```
qiluo-msft, line 90: >networking service is active [](http://example.com/codeflow?start=11&length=28)

If "networking service is active" is a dependency to fix the issue, we need to worry about some container start during "networking service is not active", but later on the service started.

oleksandrivantsiv:
The solution described in https://unix.stackexchange.com/a/348406/161486 is redundant. This is a default behavior Docker has. When the container starts dockerd copies "/etc/resolv.conf" file from the host OS. The issue with this behavior is described in my previous comment:
```
There is a small time window when the race condition can occur. It always happens only during the config reload when the DNS configuration changes.

1. Dockerd starts the container with the current content of the "resolv.conf" (copies it from the host). It takes some time for the container to start and move into the "running" state.
2. Meanwhile, concurrently, hostcfgd receives an event with the DNS configuration. It runs resolv-config.service, which writes the DNS configuration into the "/etc/resolv.conf" file and triggers the update of the containers.
3. When the "update-containers" script starts, some Docker containers might not be in the "running" state yet, so the update for these containers can be missed.
```

files/image_config/resolv-config/update-containers
oleksandrivantsiv: I'm not sure I fully understand this comment

ganglyu (approved):

PR #22451, opened 2025-04-27
Add Orchagent ZMQ table feature flag: orch_dash_zmq_enabled and orch_route_zmq_enabled.

liuh-80:
Add Orchagent ZMQ table feature flag: orch_dash_zmq_enabled and orch_route_zmq_enabled

#### Why I did it
On current image, orchagent only enable ZMQ on Dash tables, so use -q option to control the ZMQ feature is enough.

However, we plan to improve route performance by enable ZMQ between fpmsyncd and orchagent. https://github.com/sonic-net/SONiC/pull/1659

Then there will be 2 features on orchagent share same ZMQ address, so using -q to control the Dash ZMQ feature is not enough.

For Dash, we plan to migrate to enable/disable with the 'orch_dash_zmq_enabled' feature flag. There will be another PR in orchagent to make Dash enable/disable ZMQ with this flag.

And for improve route performance by enable ZMQ between fpmsyncd and orchagent, it will be controlled by the 'orch_route_zmq_enabled' flag.

##### Work item tracking
- Microsoft ADO **(number only)**:32582830

#### How I did it
Add orch_dash_zmq_enabled and orch_route_zmq_enabled field to DEVICE_METADATA table in config_db.

orch_dash_zmq_enabled:
    Control enable/disable ZMQ on Dash feature.

orch_route_zmq_enabled:
    Control enable/disable ZMQ between fpmsyncd and orchagent for improve route performance.

#### How to verify it
Pass all test case.

Manually verified orch_zmq_tables.conf can generate correctly according to CONFIG_DB

admin@vlab-01:~$ sonic-db-cli CONFIG_DB hset "DEVICE_METADATA|localhost" "orch_route_zmq_enabled" "true"
1
admin@vlab-01:~$ sonic-db-cli CONFIG_DB hset "DEVICE_METADATA|localhost" "orch_dash_zmq_enabled" "true"
1
admin@vlab-01:~$ sudo config save -y
admin@vlab-01:~$ sudo config reload -y
Acquired lock on /etc/sonic/reload.lock
Disabling container and routeCheck monitoring ...
Stopping SONiC target ...
Running command: /usr/local/bin/sonic-cfggen -j /etc/sonic/init_cfg.json -j /etc/sonic/config_db.json --write-to-db
Restarting SONiC target ...
Enabling container and routeCheck monitoring ...
Reinitializing monit daemon
Released lock on /etc/sonic/reload.lock
admin@vlab-01:~$ docker exec -it swss bash
root@sonic:/# cat /etc/swss/orch_zmq_tables.conf
DASH_VNET_TABLE
DASH_QOS_TABLE
DASH_ENI_TABLE
DASH_ACL_IN_TABLE
DASH_ACL_OUT_TABLE
DASH_ACL_GROUP_TABLE
DASH_ACL_RULE_TABLE
DASH_HA_SET_TABLE
DASH_HA_SCOPE_TABLE
DASH_PREFIX_TAG_TABLE
DASH_ROUTING_TYPE_TABLE
DASH_APPLIANCE_TABLE
DASH_ROUTE_TABLE
DASH_ROUTE_RULE_TABLE
DASH_VNET_MAPPING_TABLE
DASH_ENI_ROUTE_TABLE
DASH_ROUTE_GROUP_TABLE
DASH_TUNNEL_TABLE
DASH_PA_VALIDATION_TABLE
DASH_METER_POLICY_TABLE
DASH_METER_RULE_TABLE
DASH_ROUTING_APPLIANCE_TABLE
DASH_ENI_FORWARD_TABLE
ROUTE_TABLE
LABEL_ROUTE_TABLE

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ] 

Add Orchagent ZMQ table feature flag: orch_dash_zmq_enabled and orch_route_zmq_enabled


src/sonic-yang-models/yang-models/sonic-device_metadata.yang
```diff
@@ -299,6 +299,11 @@ module sonic-device_metadata {
                     type string;
                     description "Set source of anchor route";
                 }
+
+                leaf-list orch_zmq_tables {
```
qiluo-msft, line 303: >orch_zmq_tables [](http://example.com/codeflow?start=26&length=15)

I understand orch_zmq_tables config is super flexible and could dynamically change orchagent behavior. Do we really need such flexibility? If not, a static config file on disk may be good enough.

liuh-80: We need at least 2 feature flags:
Enable/disable improve Route performance with ZMQ.
Enable/disable improve Dash performance with ZMQ.

Currently there is no feature flag for enable/disable Dash ZMQ, it's controlled by the 'subtype' flag, if subtype is 'SmartSwitch' then orchagent will start with ZMQ enabled and Dash tables will use ZMQ.

However, after add Route ZMQ support, orchagent can't use ZMQ flag to identify Dash ZMQ feature status, so we need a Dash ZMQ feature flag.

qiluo-msft (approved):

qiluo-msft (approved):

prabhataravind (changes requested):
This can potentially cause issues because we are not planning to have zmq enabled for non-dash tables in orchagent. If the throughput b/w fpmsyncd and orchagent is very high and orchangent<->syncd interface can't keep up with it, we will have issues. @prsunny  @liuh-80

qiluo-msft:
This feature is configurable by "orch_route_zmq_enabled" in ConfigDB.

src/sonic-yang-models/yang-models/sonic-device_metadata.yang
```diff
@@ -300,6 +300,18 @@ module sonic-device_metadata {
                     description "Set source of anchor route";
                 }
 
+                leaf orch_dash_zmq_enabled {
+                    type boolean;
+                    description "Enable ZMQ feature on APPL_DB DASH tables.";
```
prabhataravind, line 305: Now that you have a dedicate flag for dash orchs to use zmq, can we deprecate the current '-q' option with zmq_server_address that is currently passed to orchagent in smartswitch to enable zmq b/w gnmi and orchagent and use this instead? Why not have a consistent way to enable zmq b/w orchagent and all other components like gnmi, fpmsyncd etc..?

Ref: https://github.com/sonic-net/sonic-buildimage/pull/16661/

prabhataravind:
> This feature is configurable by "orch_route_zmq_enabled" in ConfigDB.

@qiluo-msft will the def

> This feature is configurable by "orch_route_zmq_enabled" in ConfigDB.

@qiluo-msft I see the yang model specifies a default value of false for these, does that mean that this will be set as "false" by default in config_db? As long as config_db will have this disabled by default, it should be ok.

dockers/docker-orchagent/docker-init.j2
```diff
@@ -11,6 +11,7 @@ CFGGEN_PARAMS=" \
      -a "{\"ENABLE_ASAN\":\"{{ENABLE_ASAN}}\"}" \
 {% endif %}
     -y /etc/sonic/constants.yml \
+    -t /usr/share/sonic/templates/orch_zmq_tables.conf.j2,/etc/swss/orch_zmq_tables.conf \
```
prabhataravind, line 14: suggest using /etc/swss/config.d/orch_zmq_tables.conf as per the convention.

src/sonic-yang-models/yang-models/sonic-device_metadata.yang
liuh-80: For Dash, we plan to migrate to enable/disable with the orch_dash_zmq_enabled feature flag.

But the -q option will keep, because the ZMQ address is generated according to the mgmt_ip and eth0-midplane interface status.

There will be another PR in orchagent to make it enable/disable ZMQ with DECICE_METADATA feature flags.

prabhataravind (approved):
Discussed offline

src/sonic-yang-models/yang-models/sonic-device_metadata.yang
prabhataravind: okay, please link all relevant zmq PRs together for convenience when you get a chance.

vivekkarri:
@liuh-80, @prsunny, @prabhataravind This change would break backward compatibility for SmartSwitchDpu types that are already using ZMQ for GNMI <-> orchagent communication

Please update the t1-smartswitch preset to have this config (`orch_dash_zmq_enabled`) as part of "DEVICE_METADATA|localhost" table.

https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-config-engine/config_samples.py#L156

prabhataravind:
> @liuh-80, @prsunny, @prabhataravind This change would break backward compatibility for SmartSwitchDpu types that are already using ZMQ for GNMI <-> orchagent communication
> 
> Please update the t1-smartswitch preset to have this config (`orch_dash_zmq_enabled`) as part of "DEVICE_METADATA|localhost" table.
> 
> https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-config-engine/config_samples.py#L156

@qiluo-msft for viz

liuh-80:
> @liuh-80, @prsunny, @prabhataravind This change would break backward compatibility for SmartSwitchDpu types that are already using ZMQ for GNMI <-> orchagent communication
> 
> Please update the t1-smartswitch preset to have this config (`orch_dash_zmq_enabled`) as part of "DEVICE_METADATA|localhost" table.
> 
> https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-config-engine/config_samples.py#L156

Hi @vivekrnv , This change will not break the feature because:
1. the orch_dash_zmq_enabled is for SmartSwitchDpu GNMI <-> orchagent. it's same feature, just improve to enable/disable by this feature flag, and also the feature is enabled by default when pass ZMQ address with port.
2. there is another draft PR will rename this feature flag to orch_northbond_dash_zmq_enabled: https://github.com/sonic-net/sonic-buildimage/pull/23028

vivekkarri:
> 1. orch_dash_zmq_enabled

Hi @liuh-80, 

I checked the orchagent code. It is enabled by default. Thanks for the clarification

PR #22518, opened 2025-05-02
Including platform_utils in make and removing default initialization of $SMARTSWITCH in TSA, TSB, TSC

### LINKED ISSUE #21721: TSC command doesn't work on the Smart Switch
**Steps to reproduce**

Can be manually reproduced:

```
admin@sonic:~$ sudo TSC
This commmand is only supported Chassis
```

PR https://github.com/sonic-net/sonic-buildimage/pull/19729 introduced an env variable SMARTSWITCH which is exported in by new script /usr/bin/platform_utils.

Only with the file existing and the SMARTSWITCH is true, the TSA/TSB/TSC can work on a smartswitch properly, otherwise it will be considered as a chassis with linecards:

https://github.com/sonic-net/sonic-buildimage/blob/master/dockers/docker-fpm-frr/base_image_files/TSC#L14

```
#!/bin/bash

# Source the platform_utils script
source "$(dirname "$0")/platform_utils" > /dev/null 2>&1

${SMARTSWITCH:=false}

# Restrict command to sudo users
if [ "$EUID" -ne 0 ] ; then
  echo "Root priveleges are needed for this operation"
  exit 1
fi

if [ -f /etc/sonic/chassisdb.conf ] && [ "$SMARTSWITCH" = false ] ; then
  if [[ $1 == "no-stats" ]]; then
    rexec all -c "sudo TSC no-stats"
  else
    rexec all -c "sudo TSC"
  fi
  exit 0
fi

/usr/bin/TS TSC
[[ $1 != "no-stats" ]] && portstat -p 5

if [[ "$(sonic-cfggen -d -v DEVICE_METADATA.localhost.subtype | tr '[:upper:]' '[:lower:]')" == *"dualtor"* ]]
then
  echo
  show mux status
fi
```

But the platform_utils is not there:

```
admin@sonic:/$ ls /usr/bin/platform_utils
ls: cannot access '/usr/bin/platform_utils': No such file or directory

admin@sonic:/$ whereis platform_utils
platform_utils:
```

`sonic-buildimage/rules/docker-fpm-frr.mk` should be updated to include `/usr/bin/platform_utils`

https://github.com/sonic-net/sonic-buildimage/blob/master/rules/docker-fpm-frr.mk

**Observed behavior**

Test failed:

Failed: TSC return unexpected state ERROR

**Expected behavior**

Test should pass on smart switch.

rameshraghupathy:
Including platform_utils in make and removing default initialization of $SMARTSWITCH in TSA, TSB, TSC

Fixes #21721 

#### Why I did it
Forgot to add platform_utils into make when I made the commit. Due to this TSC command doesn't work on the Smart Switch

##### Work item tracking

#### How I did it
Included platform_utils in rules/docker-fpm-frr.mk
Removed default initialization of $SMARTSWITCH in TSA, TSB, TSC

#### How to verify it
Run "sudo TSC"  and it should output 
sudo TSC
System Mode: No external neighbors
The rates are calculated within 5 seconds period
#### 
The command should not output
admin@sonic:~$ sudo TSC
This commmand is only supported Chassis

- [x] 202405
- [x] 202411


dockers/docker-fpm-frr/base_image_files/TSC
```diff
@@ -1,10 +1,9 @@
 #!/bin/bash
+set -x
```
oleksandrivantsiv, line 2: Can this be removed?

rameshraghupathy: Done

oleksandrivantsiv (approved):

vvolam (approved):
LGTM, Thanks!

kperumalbfn (approved):

PR #22526, opened 2025-05-03
sonic-cfggen should return boolean values as lowercase true and false

bhouse-nexthop:
#### Why I did it
YANG mandates that boolean values are true and false, having the first letter be capital as is represented by Python would cause yang validation failures inside of libyang.  Because of this, it appears a different yang type in `sonic-types.yang` has a secondary `boolean_types` that is used that allows more formats.

This can be seen, for instance, when setting:
```
"MGMT_VRF_CONFIG": {
  "vrf_global": {
    "mgmtVrfEnabled": "true"
  }
}
```

Then running `config save -y`.  The written value is then "True", if then just running `config replace /etc/sonic/config_db.json` it would result in an error of:

```
Invalid value "True" in "mgmtVrfEnabled" element.
```

#### How I did it

Capture native yang boolean values by looking up the yang schema when writing the config_db.json via sonic-cfggen and ensure they are written as lowercase.

#### How to verify it

See above for reproduction method

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202411

master as of 20250502

sonic-cfggen should return boolean values as lowercase true and false

N/A

Fixes https://github.com/sonic-net/sonic-buildimage/issues/22527

### DIFF
--- a/src/sonic-config-engine/sonic-cfggen
+++ b/src/sonic-config-engine/sonic-cfggen
@@ -27,6 +27,7 @@ import sys
 import yaml
 import ipaddress
 import base64
+import sonic_yang
 
 from collections import OrderedDict
 from config_samples import generate_sample_config, get_available_config
@@ -39,6 +40,7 @@ from swsscommon.swsscommon import ConfigDBConnector, SonicDBConfig, ConfigDBPipe
 from asic_sensors_config import get_asic_sensors_config
 
 PY3x = sys.version_info >= (3, 0)
+YANG_MODELS = "/usr/local/yang-models"
 
 # TODO: Remove STR_TYPE, FILE_TYPE once SONiC moves to Python 3.x
 # TODO: Remove the import SonicYangCfgDbGenerator once SONiC moves to python3.x
@@ -212,7 +214,13 @@ TODO(taoyl): Current version of config db only supports BGP admin states.
         return db_data
 
     @staticmethod
-    def to_serialized(data, lookup_key = None):
+    def to_serialized(data, lookup_key = None, sy = None, path_tokens = None):
+        if path_tokens is None:
+            path_tokens = []
+        if sy is None:
+          sy = sonic_yang.SonicYang(YANG_MODELS, print_log_enabled = False)
+          sy.loadYangModel()
+
         if type(data) is dict:
             if lookup_key is not None:
                 newData = {}
@@ -227,7 +235,25 @@ TODO(taoyl): Current version of config db only supports BGP admin states.
                 new_key = ConfigDBConnector.serialize_key(key)
                 if new_key != key:
                     data[new_key] = data.pop(key)
-                data[new_key] = FormatConverter.to_serialized(data[new_key])
+                path_tokens.append(new_key)
+                data[new_key] = FormatConverter.to_serialized(data[new_key], sy = sy, path_tokens = path_tokens)
+                path_tokens.pop()
+
+        if isinstance(data, bool):
+            data = 'true' if data else 'false'
+
+        # If we have a string that we know looks like true/false, but differs by
+        # Case, look up from the yang schema to see if we should convert into
+        # a properly formatted boolean for yang.
+        if isinstance(data, str) and data.lower() in [ "true", "false" ] and data not in [ "true", "false" ]:
+            try:
+                schema_xpath = sy.configdb_path_to_xpath(sy.configdb_path_join(path_tokens), schema_xpath = True)
+                datatype = sy.get_data_type(schema_xpath)
+                if datatype == "BOOL":
+                    data = 'true' if data.lower() == "true" else 'false'
+            except:
+                pass
+
         return data
 
     @staticmethod
@@ -504,7 +530,8 @@ def main():
 
     if args.var_json is not None and args.var_json in data:
         if args.key is not None:
-            print(json.dumps(FormatConverter.to_serialized(data[args.var_json], args.key), indent=4, cls=minigraph_encoder))
+            print(json.dumps(FormatConverter.to_serialized(data[args.var_json], args.key), indent=4,
+                             cls=minigraph_encoder))
         else:
             print(json.dumps(FormatConverter.to_serialized(data[args.var_json]), indent=4, cls=minigraph_encoder))
 
--- a/src/sonic-yang-mgmt/sonic_yang.py
+++ b/src/sonic-yang-mgmt/sonic_yang.py
@@ -53,6 +53,30 @@ class SonicYang(SonicYangExtMixin, SonicYangPathMixin):
         # element path for CONFIG DB. An example for this list could be:
         # ['PORT', 'Ethernet0', 'speed']
         self.elementPath = []
+        self.mapped_types = {
+            "DER":ly.LY_TYPE_DER,
+            "BINARY":ly.LY_TYPE_BINARY,
+            "BITS":ly.LY_TYPE_BITS,
+            "BOOL":ly.LY_TYPE_BOOL,
+            "DEC64":ly.LY_TYPE_DEC64,
+            "EMPTY":ly.LY_TYPE_EMPTY,
+            "ENUM":ly.LY_TYPE_ENUM,
+            "IDENT":ly.LY_TYPE_IDENT,
+            "INST":ly.LY_TYPE_INST,
+            "LEAFREF":ly.LY_TYPE_LEAFREF,
+            "STRING":ly.LY_TYPE_STRING,
+            "UNION":ly.LY_TYPE_UNION,
+            "INT8":ly.LY_TYPE_INT8,
+            "UINT8":ly.LY_TYPE_UINT8,
+            "INT16":ly.LY_TYPE_INT16,
+            "UINT16":ly.LY_TYPE_UINT16,
+            "INT32":ly.LY_TYPE_INT32,
+            "UINT32":ly.LY_TYPE_UINT32,
+            "INT64":ly.LY_TYPE_INT64,
+            "UINT64":ly.LY_TYPE_UINT64,
+            "UNKNOWN":ly.LY_TYPE_UNKNOWN
+        }
+
         try:
             self.ctx = ly.Context(yang_dir, sonic_yang_options)
         except Exception as e:
@@ -686,47 +710,39 @@ class SonicYang(SonicYangExtMixin, SonicYangPathMixin):
     output:   type
     """
     def _str_to_type(self, type_str):
-           mapped_type = {
-                "LY_TYPE_DER":ly.LY_TYPE_DER,
-                "LY_TYPE_BINARY":ly.LY_TYPE_BINARY,
-                "LY_TYPE_BITS":ly.LY_TYPE_BITS,
-                "LY_TYPE_BOOL":ly.LY_TYPE_BOOL,
-                "LY_TYPE_DEC64":ly.LY_TYPE_DEC64,
-                "LY_TYPE_EMPTY":ly.LY_TYPE_EMPTY,
-                "LY_TYPE_ENUM":ly.LY_TYPE_ENUM,
-                "LY_TYPE_IDENT":ly.LY_TYPE_IDENT,
-                "LY_TYPE_INST":ly.LY_TYPE_INST,
-                "LY_TYPE_LEAFREF":ly.LY_TYPE_LEAFREF,
-                "LY_TYPE_STRING":ly.LY_TYPE_STRING,
-                "LY_TYPE_UNION":ly.LY_TYPE_UNION,
-                "LY_TYPE_INT8":ly.LY_TYPE_INT8,
-                "LY_TYPE_UINT8":ly.LY_TYPE_UINT8,
-                "LY_TYPE_INT16":ly.LY_TYPE_INT16,
-                "LY_TYPE_UINT16":ly.LY_TYPE_UINT16,
-                "LY_TYPE_INT32":ly.LY_TYPE_INT32,
-                "LY_TYPE_UINT32":ly.LY_TYPE_UINT32,
-                "LY_TYPE_INT64":ly.LY_TYPE_INT64,
-                "LY_TYPE_UINT64":ly.LY_TYPE_UINT64,
-                "LY_TYPE_UNKNOWN":ly.LY_TYPE_UNKNOWN
-           }
+       if type_str not in self.mapped_types:
+           return ly.LY_TYPE_UNKNOWN
 
-           if type_str not in mapped_type:
-               return ly.LY_TYPE_UNKNOWN
+       return self.mapped_types[type_str]
 
-           return mapped_type[type_str]
+    def _type_to_str(self, type_int):
+       name=""
+       try:
+           name = list(self.mapped_types.keys())[list(self.mapped_types.values()).index(type_int)]
+       except ValueError:
+           name = "UNKNOWN"
+
+       return name
 
     def _get_data_type(self, schema_xpath):
         try:
             schema_node = self._find_schema_node(schema_xpath)
         except Exception as e:
             self.sysLog(msg="get_data_type(): Failed to find schema node from xpath: {}".format(schema_xpath), debug=syslog.LOG_ERR, doPrint=True)
-            self.fail(e)
             return None
 
         if (schema_node is not None):
            return schema_node.subtype().type().base()
 
         return ly.LY_TYPE_UNKNOWN
+
+    """
+    get_data_type: Retrieve the data type of the node
+    input: schema_xpath
+    output: string representing type of node this points to
+    """
+    def get_data_type(self, schema_xpath):
+        return self._type_to_str(self._get_data_type(schema_xpath))
 
     """
     get_leafref_type:   find the type of node that leafref references to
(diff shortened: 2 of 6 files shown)


src/sonic-config-engine/sonic-cfggen
```diff
@@ -228,6 +228,13 @@ TODO(taoyl): Current version of config db only supports BGP admin states.
                 if new_key != key:
                     data[new_key] = data.pop(key)
                 data[new_key] = FormatConverter.to_serialized(data[new_key])
+
+        # Ensure boolean values are returned as lowercase 'true' and 'false'
+        # to prevent yang validation failures.
+        if type(data) is bool:
+            return 'true' if data else 'false'
+        if type(data) is str and data.lower() in [ "true", "false" ]:
```
qiluo-msft, line 236: >str [](http://example.com/codeflow?start=25&length=3)

If it is string, there may be expectation that keep the case.

src/sonic-config-engine/sonic-cfggen
```diff
@@ -228,6 +228,13 @@ TODO(taoyl): Current version of config db only supports BGP admin states.
                 if new_key != key:
                     data[new_key] = data.pop(key)
                 data[new_key] = FormatConverter.to_serialized(data[new_key])
+
+        # Ensure boolean values are returned as lowercase 'true' and 'false'
+        # to prevent yang validation failures.
+        if type(data) is bool:
```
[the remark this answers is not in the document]
bradh352: I can change it, but a few lines above in the existing code it used `if type(data) is dict:` so I kept the same form for consistency.

src/sonic-config-engine/sonic-cfggen
bhouse-nexthop: I evaluated all the yang models I could for this fact, and only found one exception and corrected that in https://github.com/sonic-net/sonic-host-services/pull/252

If yang mandates boolean values as lowercase, NOTHING should ever try to use otherwise, that's just broken.

qiluo-msft: Just to clarify, a string "TrUE" may have nothing to do with boolean, it is just a valid string. Your code does not allow such string.

bhouse-nexthop: So you're worried about modification of actual string fields that someone just so happens to put the words true or false even though they're not boolean fields, being case-converted.  The only way to solve that would be to load the yang schema, evaluate the path and determine the datatype in the yang schema.  

If that's the ask, I can do that, but it will definitely cause a performance hit (though I'm not sure of if the performance of this is actually relevant).

Or if it would be acceptable to only case convert `True` and `False` to `true` and `false` ignoring cases like `TrUE` as being acceptable then I can do that too.

I just need guidance so this can move forward.

qiluo-msft: I feel "sonic-cfggen" has no responsibility to enforce yang models, the purpose of this tool is just to render json. So manipulating the string and guess it is a boolean should not be a valid problem to solve. I suggest only focus on boolean type, not checking string type.

bhouse-nexthop:
@qiluo-msft Hopefully the changes to this are now acceptable to you.  Only fields that are actually tagged as boolean are ever converted to lowercase true and false.  I added test cases around this particular purpose to ensure the conversion happens when it should and doesn't when it shouldn't.

bhouse-nexthop:
I have no idea why azure pipelines keeps marking the prepare phase for tests as failed.  Not seeing anything in the logs to indicate what the failure actually is.  Temporarily trying to revert all changes in this PR to see if its still unhappy.  If it is, then at least I know its not related to this PR.

AND that didn't work as it expects path changes to determine what to run.

bhouse-nexthop:
Ok, hopeful this build should work and test properly.  

Looks like some containers weren't starting due to featured related issues expecting True/False since https://github.com/sonic-net/sonic-host-services/pull/252 hasn't merged. 

I added a workaround that should allow this to function regardless if the above is merged by preserving the pythonic True/False when the YANG model doesn't indicate its really a boolean even though it was stored in CONFIG_DB as a boolean field. 

This workaround should really be removed in the future as that concept is just broken.

src/sonic-config-engine/sonic-cfggen
```diff
@@ -27,6 +27,7 @@ import sys
 import yaml
 import ipaddress
 import base64
+import sonic_yang
```
qiluo-msft, line 30: >sonic_yang [](http://example.com/codeflow?start=7&length=10)

config-cfggen currently does not depend on sonic_yang. It is used in some critical time window and need high performance, for example during warm-boot.

if introducing sonic_yang, the runtime speed will hurt a lot.

bhouse-nexthop: Do you have measurable numbers to show this to be the case?

The prior solution I presented wasn't sufficient for you because it would convert potentially string values into boolean values.

You suggested I only handle values that were ***stored*** as booleans in CONFIG_DB to output 'true' / 'false', but there were 2 issues with that:
  1. Sometimes in "FEATURES" the config_db  value is stored as a boolean, but it really wants True or False (you can see that from a test case I added here).  This is fixed in https://github.com/sonic-net/sonic-host-services/pull/252 that you just merged, thank you.
  2. Sometimes in config_db the value is stored as a string of "True" or "False" but needs to be output as 'true' or 'false', as is the case with Management VRF enablement that this ticket was created for.

Shall I go back to my original PR where I just properly case-convert all string and boolean values that case-insensitively match "true" and "false" to their lowercase variants regardless of if they are boolean or string?

Without using the YANG models to know for sure the proper datatype, there is really no other option.  I tried to limit the exposure on this by just loading the yang models once and only validating when necessary, the overhead in theory should be very low with this methodology in the PR as it currently sits.

bhouse-nexthop: @qiluo-msft need your guidance to move forward here since no solution I've presented has met your requirements.

bhouse-nexthop:
@qiluo-msft I still need more feedback here if you feel what I have done is not acceptable.  Regarding your warm-reboot scenario we're talking less than 100ms overhead with my approach, and I don't believe this would be within the outage window anyhow.

bhouse-nexthop:
@qiluo-msft please help!

lguohan:
@bhouse-nexthop , please take a look at the copilot review

src/sonic-config-engine/sonic-cfggen
```diff
@@ -212,7 +214,11 @@ TODO(taoyl): Current version of config db only supports BGP admin states.
         return db_data
 
     @staticmethod
-    def to_serialized(data, lookup_key = None):
+    def to_serialized(data, lookup_key = None, sy = None, path_tokens = []):
+        if sy is None:
+          sy = sonic_yang.SonicYang(YANG_MODELS, print_log_enabled = False)
+          sy.loadYangModel()
```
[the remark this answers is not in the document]
bhouse-nexthop: the none check is specifically to enable caching, so this is an invalid suggestion

src/sonic-config-engine/sonic-cfggen
```diff
@@ -212,7 +214,11 @@ TODO(taoyl): Current version of config db only supports BGP admin states.
         return db_data
 
     @staticmethod
-    def to_serialized(data, lookup_key = None):
+    def to_serialized(data, lookup_key = None, sy = None, path_tokens = []):
```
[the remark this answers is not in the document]
bhouse-nexthop: this suggestion seems odd and I don't understand the difference.  But sure, I'll accept it to make copilot happy

src/sonic-yang-mgmt/sonic_yang.py
```diff
@@ -686,48 +710,40 @@ def _get_module_prefix(self, module_name):
...
+           return ly.LY_TYPE_UNKNOWN
+
+       return self.mapped_types[type_str]
+
+    def _type_to_str(self, type_int):
+       name=""
+       try:
+           name = list(self.mapped_types.keys())[list(self.mapped_types.values()).index(type_int)]
+       except:
+           name = "UNKNOWN"
```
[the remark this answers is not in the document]
bhouse-nexthop: iterating across small lists are very fast.  There'd be very little benefit of creating a reverse lookup from a performance standpoint.  Also, this is only called when the value is "true" or "false" to verify if it is a boolean, so this limits the exposure even further.  In short, I don't think this is worth the effort.

bhouse-nexthop:
Thanks @lguohan I made a couple minor modifications and replied to the rest.

bradh352:
these spurious test failures in completely unrelated code paths are getting bad.  I just rebased to force a rebuild.

TafkaMax:
> these spurious test failures in completely unrelated code paths are getting bad. I just rebased to force a rebuild.

Also probably needs a rebuild again?

bradh352:
> > these spurious test failures in completely unrelated code paths are getting bad. I just rebased to force a rebuild.
> 
> Also probably needs a rebuild again?

$!*% yep

TafkaMax:
> > > these spurious test failures in completely unrelated code paths are getting bad. I just rebased to force a rebuild.
> > 
> > 
> > Also probably needs a rebuild again?
> 
> $!*% yep

Second time it worked :-)

bradh352:
@lguohan any chance you can look at this and let me know if you have similar concerns to @qiluo-msft ?  It at most loads the yang models once per sonic-cfggen execution to do proper field type validation to prevent an invalid configdb yang schema from being written out.

lguohan:
let me talk to qi when he is back tomorrow. do you have some measurable data to show that this won't add much overhead?

lguohan:
btw, the pr description/title probably should update saying that introducing yang model validation in sonic-cfggen?

bhouse-nexthop:
> btw, the pr description/title probably should update saying that introducing yang model validation in sonic-cfggen?

The description does say it under the "How I did it" section.

bhouse-nexthop:
> let me talk to qi when he is back tomorrow. do you have some measurable data to show that this won't add much overhead?

Considering this is only loading the yang models and not actually performing the validation step, I'd assume the hit would be negligible.  That said, I don't have hard numbers.  Let me see if I can collect some before / after numbers for you on real hardware.

src/sonic-config-engine/sonic-cfggen
```diff
@@ -227,7 +235,35 @@ TODO(taoyl): Current version of config db only supports BGP admin states.
                 new_key = ConfigDBConnector.serialize_key(key)
                 if new_key != key:
                     data[new_key] = data.pop(key)
-                data[new_key] = FormatConverter.to_serialized(data[new_key])
+                path_tokens.append(new_key)
+                data[new_key] = FormatConverter.to_serialized(data[new_key], sy = sy, path_tokens = path_tokens)
+                path_tokens.pop()
+
+        if isinstance(data, bool):
+            # TODO: once https://github.com/sonic-net/sonic-host-services/pull/252 merges, this should *always*
```
qiluo-msft: >once https://github.com/sonic-net/sonic-host-services/pull/252 merges [](http://example.com/codeflow?start=20&length=69)

[https://github.com/sonic-net/sonic-host-services/pull/252](https://github.com/sonic-net/sonic-host-services/pull/252) is already merged. Is it possible just always rewrite as lowercase true/false if it is a bool. I did not get the compatibility you mentioned.

bradh352: sure I can remove that logic, I think your main concern though was using the yang models at all to determine if it is a boolean or not.

I'll get some performance numbers for using yang models during the serialization of the redis to configdb json file as you were worried about warm reboot perf.

qiluo-msft: I am looking forward to remove yang dependency completely. Let's see how it is going.

bradh352: @qiluo-msft that wouldn't remove the yang dependency.  

I'd need guidance then on what is acceptable.  My first revision I submitted took any case-independent string of 'true' or 'false' and converted to lowercase and you didn't like that because non-boolean fields would be affected (meaning a true string field, like a description).  

So I added logic to pull in the yang models to *only* convert true/false to lowercase *if* it was really a boolean field as far as yang was concerned.  And you don't like that because of possible performance impacts. 

I'm willing to go either way but I don't think there's any middle ground.  Either you are going to "possibly" (though very very very unlikely) convert a user provided "string", e.g. description, to lowercase if they happen to use non-lowercase "true"/"false" as a value, or you need to validate against the yang models to only convert those that are truly booleans.

qiluo-msft:
I am trying to repro the original issue mentioned in PR description, but failed to repro. I could not observe 'The written value is then "True"'
```shell
admin@sonic:~$ sonic-db-cli CONFIG_DB hset "MGMT_VRF_CONFIG|vrf_global" "mgmtVrfEnabled" true
1
admin@sonic:~$ sonic-db-cli CONFIG_DB hgetall "MGMT_VRF_CONFIG|vrf_global"
{'mgmtVrfEnabled': 'true'}
admin@sonic:~$ sudo config save -y
admin@sonic:~$ grep -C 2 "mgmtVrfEnabled" /etc/sonic/config_db.json
    "MGMT_VRF_CONFIG": {
        "vrf_global": {
            "mgmtVrfEnabled": "true"
        }
    },
```

bhouse-nexthop:
> I am trying to repro the original issue mentioned in PR description, but failed to repro. I could not observe 'The written value is then "True"'
> 
> ```shell
> admin@sonic:~$ sonic-db-cli CONFIG_DB hset "MGMT_VRF_CONFIG|vrf_global" "mgmtVrfEnabled" true
> 1
> admin@sonic:~$ sonic-db-cli CONFIG_DB hgetall "MGMT_VRF_CONFIG|vrf_global"
> {'mgmtVrfEnabled': 'true'}
> admin@sonic:~$ sudo config save -y
> Running command: /usr/local/bin/sonic-cfggen -d --print-data > /etc/sonic/config_db.json
> admin@sonic:~$ grep -C 2 "mgmtVrfEnabled" /etc/sonic/config_db.json
>     "MGMT_VRF_CONFIG": {
>         "vrf_global": {
>             "mgmtVrfEnabled": "true"
>         }
>     },
> ```

I modified the config_db.json and didn't play with the sonic-db-cli.  I haven't looked at how sonic-db-cli writes that, but I'm guessing its writing the 'true' as a string.  A string would be preserved in the original case.  It actually has to be in Redis as a boolean value for this issue to be observed.

bhouse-nexthop:
Performance numbers below.

Platform: x86_64-mlnx_msn2700-r0
HwSKU: ACS-MSN2700

Pretty slow cpu:
```
root@sw1:~# cat /proc/cpuinfo | grep "model name" | head -n1
model name	: Intel(R) Celeron(R) CPU 1047UE @ 1.40GHz
```


with patch:
```
root@sw1:~# time config save -y

real	0m1.754s
user	0m1.374s
sys	0m0.328s
root@sw1:~# time config save -y

real	0m1.605s
user	0m1.341s
sys	0m0.237s
root@sw1:~# time config save -y

real	0m1.616s
user	0m1.362s
sys	0m0.215s
```

without patch:
```
root@sw1:~# time config save -y

real	0m1.239s
user	0m1.044s
sys	0m0.152s
root@sw1:~# time config save -y

real	0m1.204s
user	0m1.004s
sys	0m0.170s
root@sw1:~# time config save -y

real	0m1.197s
user	0m1.014s
sys	0m0.164s
```

src/sonic-config-engine/sonic-cfggen
bhouse-nexthop: I removed the temporary workaround since https://github.com/sonic-net/sonic-host-services/pull/252 merged and rebased.

qiluo-msft:
> > I am trying to repro the original issue mentioned in PR description, but failed to repro. I could not observe 'The written value is then "True"'
> > ```shell
> > admin@sonic:~$ sonic-db-cli CONFIG_DB hset "MGMT_VRF_CONFIG|vrf_global" "mgmtVrfEnabled" true
> > 1
> > admin@sonic:~$ sonic-db-cli CONFIG_DB hgetall "MGMT_VRF_CONFIG|vrf_global"
> > {'mgmtVrfEnabled': 'true'}
> > admin@sonic:~$ sudo config save -y
> > Running command: /usr/local/bin/sonic-cfggen -d --print-data > /etc/sonic/config_db.json
> > admin@sonic:~$ grep -C 2 "mgmtVrfEnabled" /etc/sonic/config_db.json
> >     "MGMT_VRF_CONFIG": {
> >         "vrf_global": {
> >             "mgmtVrfEnabled": "true"
> >         }
> >     },
> > ```
> 
> I modified the config_db.json and didn't play with the sonic-db-cli. I haven't looked at how sonic-db-cli writes that, but I'm guessing its writing the 'true' as a string. A string would be preserved in the original case. It actually has to be in Redis as a boolean value for this issue to be observed.

Could you provide your repro steps for the original issue? I do not understand "It actually has to be in Redis as a boolean value" since redis only support string as value. If you modify config_db.json and trigger a bug, I prefer to treat it as invalid input bug and just fix the input validation logic and fail sonic-cfggen immediately, so yang dependency is not needed.

bhouse-nexthop:
steps to reproduce are:
1) Edit `/etc/sonic/config_db.json` and insert this *valid* section:
```
"MGMT_VRF_CONFIG": {
  "vrf_global": {
    "mgmtVrfEnabled": "true"
  }
}
```
2. Run `config replace /etc/sonic/config_db.json` which applies only that diff to redis successfully
3. Run `config save -y` to rewrite the redis database to `/etc/sonic/config_db.json`
4. Run `config replace /etc/sonic/config_db.json`, this will FAIL because step #3 wrote "True" for `mgmtVrfEnabled`

The trigger to this behavior is the fact that somewhere the Pythonic True / False are being written when serializing from Redis.

PR #22533, opened 2025-05-04
Unified changes done for FRR

vidyac86:
Why I did it
The docker_routing_config_mode value which is set to default of separated is not working properly with the latest FRR.

How I did it
The changes are done in the docker_init.sh to modify the default behavior of considering the docker_routing_config_mode value as 'separated' when it is not seen in the config_db to modifying the value to 'unified'.

How to verify it
These values are significant to restore the values in FRR. When the value is separated, each process in FRR will have its own config file from where the configs are played back after save and restore, with the value of 'unified' there will be only one file frr.conf from where all the configs will be played back.

After doing this change, I manually verified the presence of frr.conf after save and restore and all the configs are restored properly.

Tested branch (Please provide the tested image version)
master

Tested image version
20250407.145357

Description for the changelog
Modified the default value of docker_routing_config_mode to unified.

[FRR_Unified_UT Logs.txt](https://github.com/user-attachments/files/19873691/FRR_Unified_UT.Logs.txt)

Link to config_db schema for YANG module changes
https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-yang-models/doc/Configuration.md#device-metadata

Related PRs
https://github.com/sonic-net/sonic-utilities/pull/3840
https://github.com/sonic-net/sonic-mgmt/pulls/vidyac86


venkatmahalingam:
The build is failing due to https://github.com/sonic-net/sonic-buildimage/pull/22538

dgsudharsan (dismissed):
we need to have full sonic-mgmt coverage to confirm these changes. Previously sonic had to move away from unified because of some issues found. We have to ensure that moving back to unified introduces no degradation.

StormLiangMS:
We do see regression with unified mode, which lost partial config during startup in a low possibilities. The PR testing may not enough to verify this change. We should have a solid testing before get this merged.

venkatmahalingam:
> We do see regression with unified mode, which lost partial config during startup in a low possibilities. The PR testing may not enough to verify this change. We should have a solid testing before get this merged.

As I requested earlier, Dell doesnt have an infra. to run the regression for bgpcfgd changes, need help from Microsoft to run the regression with this PR changes.

src/sonic-config-engine/minigraph.py
```diff
@@ -1627,7 +1627,7 @@ def parse_macsec_profile(val_string):
 def parse_global_info(root):
 
     hwsku = hostname = None
-    docker_routing_config_mode = "separated"
+    docker_routing_config_mode = "unified"
```
qiluo-msft: >docker_routing_config_mode [](http://example.com/codeflow?start=4&length=26)

We should not change behavior of minigraph.py. Today sonic relies on /etc/sonic/config_db.json for ground truth of golden config, and we should keep the behavior backward compatible.

dockers/docker-fpm-frr/docker_init.sh
```diff
@@ -56,41 +56,30 @@ if [[ ! -z "$NAMESPACE_ID" ]]; then
    update_default_gw 6
 fi
 
-if [ -z "$CONFIG_TYPE" ] || [ "$CONFIG_TYPE" == "separated" ]; then
```
qiluo-msft, line 59: >separated [](http://example.com/codeflow?start=49&length=9)

If "separated" is not supported in this image, we should fail the j2 rendering. So the user understand /etc/sonic/config_db.json contains wrong config, so they could take action to fix config.

qiluo-msft (dismissed):
as comments

src/sonic-config-engine/minigraph.py
vidyac86: Reverted the changes in minigraph.

dockers/docker-fpm-frr/docker_init.sh
vidyac86: Added prints to say the mode is not supported

qiluo-msft: Is "echo" statement enough? Should we fail this script completely instead of keep running as normal?

prsunny:
@lguohan for viz

dockers/docker-fpm-frr/docker_init.sh
```diff
@@ -56,41 +56,32 @@ if [[ ! -z "$NAMESPACE_ID" ]]; then
...
     write_default_zebra_config /etc/frr/frr.conf
-elif [ "$CONFIG_TYPE" == "unified" ]; then
-    CFGGEN_PARAMS=" \
-        -d \
-        -y /etc/sonic/constants.yml \
-        -T /usr/local/sonic/frrcfgd \
-        -t /usr/share/sonic/templates/gen_frr.conf.j2,/etc/frr/frr.conf \
-    "
+elif [ "$CONFIG_TYPE" == "separated" ]; then
+    echo "Config Type separated is not supported"
```
qiluo-msft: >separated is not supported [](http://example.com/codeflow?start=22&length=26)

This breaking change will break warm-reboot, because the old image config_db.json will be carried over to new image. Please update config in src/sonic-utilities/scripts/db_migrator.py, to use a supported "docker_routing_config_mode"

vidyac86: To solve this, db_migrator changes are added 
https://github.com/sonic-net/sonic-utilities/pull/3840

StormLiangMS:
hi @qiluo-msft  @lipxu could you help to review again?
@vidyac86  could you check on the conflict?

prvattem:
@vidyac86 - Is anything pending from the code changes? Can this be merged?

prvattem (approved):

eddieruan-alibaba:
Discussed in 0312's routing WG meeting, the conclusion is we need to change to unified mode, since FRR community is no longer support it. We are 3 releases behind. 

AI; @Deepak-Pandey  will check it internally to build a private image for an overnight regression test.

deepak-singhal0408 (dismissed):

venkatmahalingam:
@deepak-singhal0408  Are we good to merge this PR?

deepak-singhal0408:
> @deepak-singhal0408 Are we good to merge this PR?

@venkatmahalingam thanks for the reminder.
Quick update: We ran the longivity tests with this PR, and the results seems to be ok.. There was a memory jump initially which stayed throughout the run.. We have started another test without this change for comparison sake.. will have results available in next 2-3 days..

With that, if all looks good, we could hopefully get this PR merged by this Friday.

deepak-singhal0408:
> > @deepak-singhal0408 Are we good to merge this PR?
> 
> @venkatmahalingam thanks for the reminder. Quick update: We ran the longivity tests with this PR, and the results seems to be ok.. There was a memory jump initially which stayed throughout the run.. We have started another test without this change for comparison sake.. will have results available in next 2-3 days..
> 
> With that, if all looks good, we could hopefully get this PR merged by this Friday.

Just a quick update on this @venkatmahalingam .. The longivity test results on 202511 without/with this change are pretty similar.. so no concerns from that side..
But the partial config load(that was observed in MSFT production) is still a concern with unified frr.conf. We are adding a new sonic-mgmt testcase to catch such issues.. https://github.com/sonic-net/sonic-mgmt/pull/24061. We are discussing internally on this test PR and once we converge on that, will go ahead and get all relevant PRs merged (hopefully within this week)

venkatmahalingam:
> > > @deepak-singhal0408 Are we good to merge this PR?
> > 
> > 
> > @venkatmahalingam thanks for the reminder. Quick update: We ran the longivity tests with this PR, and the results seems to be ok.. There was a memory jump initially which stayed throughout the run.. We have started another test without this change for comparison sake.. will have results available in next 2-3 days..
> > With that, if all looks good, we could hopefully get this PR merged by this Friday.
> 
> Just a quick update on this @venkatmahalingam .. The longivity test results on 202511 without/with this change are pretty similar.. so no concerns from that side.. But the partial config load(that was observed in MSFT production) is still a concern with unified frr.conf. We are adding a new sonic-mgmt testcase to catch such issues.. [sonic-net/sonic-mgmt#24061](https://github.com/sonic-net/sonic-mgmt/pull/24061). We are discussing internally on this test PR and once we converge on that, will go ahead and get all relevant PRs merged (hopefully within this week)

Great! thanks @deepak-singhal0408

StormLiangMS:
Thanks @vidyac86 — I went through the diff and the prior threads. This PR has a good motivation (align SONiC with upstream FRR's preferred unified flow), but it changes the default `docker_routing_config_mode` for **every device that doesn't explicitly set the field**, so it's effectively a fleet-wide silent behavior change. Given the scope, several open concerns from prior reviews, and the fact that this has been waiting on hardware regression for ~11 months, **I'd like to suggest this be reviewed in the SONiC-FRR subgroup meeting before merge** so we can also discuss the deprecation/fallback strategy collectively. Detailed comments below.

## 🟥 Blocking issues

### 1. Dead-code bug in the `docker_init.sh` elif cascade

```bash
elif [ "$CONFIG_TYPE" == "separated" ]; then
    echo "Config Type separated is deprecated"
    # ... renders frr.conf via gen_frr.conf.j2 (i.e. unified layout)
elif [ -z "$CONFIG_TYPE" ] || [ "$CONFIG_TYPE" == "unified" ] || [ "$CONFIG_TYPE" == "separated" ]; then
    # ... unreachable for "separated" — already consumed by the previous branch
```

The third `|| [ "$CONFIG_TYPE" == "separated" ]` in the final `elif` is dead code. Either the OR is misleading (should be removed) or the cascade structure is wrong (the two branches were intended to be merged). Please clean up and add a comment explaining the intent.

### 2. `separated` is silently rewritten as `unified`, with only an `echo`

@qiluo-msft asked back on Jun 3 2025: *"Is 'echo' statement enough? Should we fail this script completely instead of keep running as normal?"* — that thread is still **unresolved**. A user who deliberately set `separated` (e.g. for tooling that scrapes per-daemon config files) will boot into `unified` with no failure, no syslog warning, just a docker-stdout `echo` that nobody reads.

Options, from strongest to weakest:
- Hard-fail with `exit 1` (qiluo's preference).
- Log to syslog via `logger -t docker-fpm-frr -p user.warning ...` so it shows up in `show logging`.
- At minimum, raise an event/alert in the changelog and release notes.

A docker-stdout `echo` is the weakest possible signal for a behavior change of this magnitude.

### 3. Default-flip can race with `db_migrator`

The PR depends on sonic-net/sonic-utilities#3840 to migrate existing configs. But `docker_init.sh` runs on every container restart, including before `db_migrator` runs on first boot of the new image. Across vendor images / fast-reboot / warm-reboot the ordering is not always strict. If `CONFIG_TYPE=""` is observed once before migration, the device silently switches to unified, and the user only later notices when their tooling breaks.

Mitigations:
- Keep the current default behavior (`separated` semantics) until `db_migrator` has explicitly written the field; only then flip the default.
- Or merge the `db_migrator` change first and gate this PR on it being in the same release branch.

### 4. `frrcfgd.py` flips the in-cache replay default — and the variable name is misleading

```python
if self.config_mode == "unified" and self.mgmt_framework_cfg_template_restore == 'false':
    for table, _ in self.table_handler_list:
        # replay cached DB tables to FRR
```

- **Old behavior:** in unified mode, **always replay cached tables**.
- **New behavior:** in unified mode, replay **only if** `mgmt_framework_cfg_template_restore == 'false'`.
- **New default:** `'true'` → replay is **skipped by default**.

Two problems:

a) The variable name reads as "restore = true ⇒ do restore"; the actual semantics are inverted. Either rename (e.g. `use_template_render_for_restore`) or invert the default. As-is it's a footgun for the next maintainer.

b) This is exactly the regression scenario I [flagged on May 7 2025](https://github.com/sonic-net/sonic-buildimage/pull/22533#issuecomment-2857698802) — *"We do see regression with unified mode, which lost partial config during startup in a low possibilities."* By disabling the table-cache replay, you remove the safety net that the old per-daemon flow used to provide. If the template render misses any field (e.g. due to template-include paths — see issue 5), the running config is silently incomplete and there's no recovery loop.

### 5. Template-path renames break the include resolver for `separated` mode

Several j2 includes were rewritten to add a subdirectory prefix:
```jinja
- {% include "bgpd.main.conf.j2" %}
+ {% include "bgpd/bgpd.main.conf.j2" %}
- {% include "staticd.loopback_route.conf.j2" %}
+ {% include "staticd/staticd.loopback_route.conf.j2" %}
- {% include "zebra.interfaces.conf.j2" %}
+ {% include "zebra/zebra.interfaces.conf.j2" %}
```

These prefix-prepended paths only resolve correctly if the cfggen template root is `/usr/share/sonic/templates`. In old `separated` mode, sonic-cfggen was invoked with `-t /usr/share/sonic/templates/bgpd/gen_bgpd.conf.j2,...`, which set the search root at `/usr/share/sonic/templates/bgpd/` — making `bgpd.main.conf.j2` (no prefix) the right path and `bgpd/bgpd.main.conf.j2` invalid. So:

- These renames irreversibly break the old `separated` rendering path even if a user manually invokes it for debugging.
- They work for unified because `gen_frr.conf.j2` resolves under the parent `/usr/share/sonic/templates/`.
- This means there is **no longer a real `separated` mode** — which makes the warning in issue 2 even more misleading.

If `separated` is truly gone, delete the branch entirely and have the warning point to release notes. If it's "deprecated but still works", these template renames need to be conditional or backward-compatible aliases (e.g. duplicate the templates under both paths during transition).

### 6. `gen_frr.conf.j2` rewrite changes the include surface

Old:
```jinja
{% include "/usr/share/sonic/templates/frr.conf.j2" %}
```
New:
```jinja
{% include "/usr/share/sonic/templates/zebra/zebra.conf.j2" %}
{% include "/usr/share/sonic/templates/staticd/staticd.conf.j2" %}
{% include "/usr/share/sonic/templates/bgpd/bgpd.conf.j2" %}
```

If `frr.conf.j2` ever included anything other than exactly these three (e.g. `bfdd.conf.j2`, `pimd.conf.j2`, `ospfd.conf.j2`, vendor-specific overrides), those fragments are now silently dropped from unified rendering. PR description doesn't show a diff of `frr.conf.j2` justifying that this enumeration is complete. Please add a commit message / PR note documenting that `frr.conf.j2` was confirmed to consist of exactly these three includes, and that no vendor j2 patch in the build chain injects more.

## 🟨 Significant but not strictly blocking

### 7. PR scope vs testing gap

PR has been open ~11 months, 79 comments, 2 commits. Touches the FRR boot path on master, with downstream effect on every SONiC image. CI is green only on KVM-VS / kvmtest-t0/t1-lag/t2 — none of which exercise:
- ChassisFrontendRouter / SpineChassisFrontendRouter (which has its own `bgpd.spine_chassis_frontend_router.conf.j2` include path renamed)
- Multi-asic with namespace + bgp container per ASIC
- BGP graceful-restart / warm-reboot config persistence
- Large config (the silent partial-load failure mode) — and there's now an open sonic-mgmt PR ([#24061](https://github.com/sonic-net/sonic-mgmt/pull/24061)) specifically designed to catch this failure mode

Strongly recommend gating merge on running #24061's `test_frr_large_config_load_stress` plus a warm-reboot regression on real T0/T1 hardware.

### 8. YANG addition lacks scope/applicability

The new `frr_mgmt_framework_cfg_template_restore` field has no `must` constraint or description note that it only takes effect in `unified` mode (and arguably only when `frr_mgmt_framework_config == 'true'`). Either add a `must` constraint, or at minimum extend the description to say so explicitly. And — see issue 4 — the name is inverted from its effect.

### 9. `Configuration.md` doc update is one-line; no migration note

The doc adds the new field to the example but doesn't tell users:
- That the default mode silently changed.
- What `frr_mgmt_framework_cfg_template_restore: 'true'` actually does (and that the default is the counter-intuitive direction).
- What action a user with `separated` in config_db today should take.

Needs a migration / behavior-change paragraph in both `Configuration.md` and the release notes.

## ✅ What looks fine
- YANG / sample_config_db / test config additions are consistent.
- CI green.
- The intent (consolidate on unified to match upstream FRR's preferred config flow) is reasonable.
- `supervisord.conf.j2` update for `vtysh -b` to also cover `separated` is correct given that branch now produces a unified layout.

---

## 📅 Suggested next steps

**Please bring this to the SONiC-FRR subgroup meeting** before merge. Two specific topics worth discussing collectively:

1. **Deprecation strategy.** Should we keep `separated` mode as a real fallback for the next 1–2 release cycles (e.g. 202511 + 202611) — i.e.:
   - Phase 1 (this release): change the *default* to `unified` when the field is unset, but leave `separated` fully functional for users who explicitly set it. Add a deprecation warning to syslog. Document in release notes.
   - Phase 2 (one release later): emit warning + hard-fail in `docker_init.sh` if `separated` is explicitly set.
   - Phase 3 (one more release later): remove the `separated` code path entirely.

   This gives operators with custom tooling around per-daemon config files time to migrate, and gives us a real rollback path if a regression surfaces in production. Today's PR effectively jumps straight to "phase 3 with `echo` warning" while still leaving dead code that *looks* like phase 2 — which is the worst of both worlds.

2. **Validation plan.** Who runs the hardware regression and the large-config stress https://github.com/sonic-net/sonic-mgmt/pull/24061 before this merges? The thread has been blocked on this question since May 2025.

### Minimum I'd want before approving (post-meeting alignment notwithstanding)

1. Resolve @qiluo-msft's open thread about hard-failing on `separated` (issue 2).
2. Fix the dead-code in the elif cascade (issue 1).
3. Either rename `frr_mgmt_framework_cfg_template_restore` or invert its default so the name matches the behavior (issue 4).
4. Confirm `frr.conf.j2` content equivalence and add a note in the PR description (issue 6).
5. Run [#24061's](https://github.com/sonic-net/sonic-mgmt/pull/24061) large-config stress test against this PR (issue 7).
6. Confirm `db_migrator` (sonic-utilities#3840) lands first or in the same release (issue 3).

vidyac86:
@StormLiangMS , Thanks for your detailed review and I have addressed wherever its applicable, Kindly review it.

Issue 1 - Dead-code in docker_init.sh: Removed the redundant || [ "$CONFIG_TYPE" == "separated" ] condition from the last elif branch (line 88). This condition was unreachable because the "separated" case is already handled by the preceding elif branch at line 67.

Issue 2 - Separated mode warning: Added logger -t docker-fpm-frr -p user.warning to log the deprecation warning to syslog. This ensures the warning appears in show logging output, making it visible to operators who may not monitor container stdout. The existing echo statement was retained for console visibility during boot.

Issue 3 - Variable naming: Renamed frr_mgmt_framework_cfg_template_restore to use_template_render_for_restore for clarity. The new name directly describes the behavior:

When true (default): use template rendering for FRR config generation
When false: restore from cached DB tables
This change was applied consistently across:

frrcfgd.py (Python code)
sonic-device_metadata.yang (YANG model)
sample_config_db.json (sample config)
Test configuration files
Configuration.md documentation
Issue 4 - frr.conf.j2 content equivalence: Added a comment in gen_frr.conf.j2 documenting that the new approach (including main daemon config files) provides equivalent configuration to the old frr.conf.j2 (which included sub-components directly). The main daemon configs (zebra.conf.j2, staticd.conf.j2, bgpd.conf.j2) each include the same sub-components, maintaining functional equivalence while improving modularity.

Items not addressed (require external coordination):

Issue 5 (large-config stress test) - Issue 6 (db_migrator coordination) - @deepak-singhal0408 has tested it, could you please upload the logs any if you have.

prvattem (approved):

StormLiangMS:
@vidyac86 thanks for the latest updates. I reviewed the current diff again. Several previous concerns look addressed now: the dead `separated` condition was cleaned up, explicit `separated` now logs via `logger`, the restore knob was renamed to `use_template_render_for_restore`, the YANG description now says it only applies in unified mode, and `gen_frr.conf.j2` has a template-equivalence note.

I still think this PR needs changes before merge. Could you please address the following?

### Blocking / high-priority

1. **Missing `docker_routing_config_mode` renders unified config but may not run `vtysh -b`.**

   `frr_vars.j2` emits an empty `docker_routing_config_mode` when the key is absent. `docker_init.sh` then treats empty mode as unified, renders `/etc/frr/frr.conf`, writes `service integrated-vtysh-config`, and removes per-daemon configs.

   However, `supervisord.conf.j2` only emits `[program:vtysh_b]` when `DEVICE_METADATA.localhost.docker_routing_config_mode is defined` and explicitly equals `unified`, `split-unified`, or `separated`. For the key-absent/default path, `vtysh -b` is not scheduled, so FRR may boot without replaying the generated unified config.

   Please align `supervisord.conf.j2` with the new effective default, or materialize the default before rendering supervisord.

2. **`frrcfgd.py` still defaults missing mode to `separated`.**

   `docker_init.sh` now treats missing mode as unified, but `frrcfgd.py` still sets `self.config_mode = "separated"` when the field is absent. This makes the effective-mode contract inconsistent. In particular, `use_template_render_for_restore=false` will not trigger cached DB replay unless the mode is explicitly set to `unified`.

   Please align the missing-mode default in `frrcfgd.py` with `docker_init.sh`, and add/confirm test coverage for missing mode + restore flag behavior.

3. **Handle `sonic-cfggen` failure before deleting fallback config files.**

   In the explicit `separated` and missing/unified paths, the script runs `sonic-cfggen $CFGGEN_PARAMS` and then unconditionally switches to integrated mode and removes `bgpd.conf`, `zebra.conf`, `staticd.conf`, `bfdd.conf`, `ospfd.conf`, and `pimd.conf`.

   Please check the `sonic-cfggen` result and fail safely before changing mode/removing per-daemon config files if rendering fails.

### Other items to close

4. Please provide rendered-output comparison for old `frr.conf.j2` vs the new `gen_frr.conf.j2` path on representative topologies. The new path includes full daemon templates (`zebra.conf.j2`, `staticd.conf.j2`, `bgpd.conf.j2`), which is cleaner but not exactly the same include surface.

5. Please gate this with the db migrator (`sonic-utilities#3840`) or confirm the warm-reboot/upgrade ordering guarantees. This PR should not land independently if existing config can still carry explicit `separated` or missing mode through upgrade.

6. Please get the current required checks green. The latest rollup still showed DCO action required and Azure / impacted-area failures (`kvmtest-t0`, `kvmtest-t0-sonic`, `kvmtest-t1-lag`).

7. Please add operator-facing migration/deprecation docs explaining:
   - missing `docker_routing_config_mode` now defaults to unified;
   - explicit `separated` is deprecated and rendered as unified;
   - what operators should do if tooling depends on per-daemon config files;
   - when to use `use_template_render_for_restore=false`.

The direction still looks reasonable, and the latest update addressed several earlier review points, but the default-mode startup path and cross-component mode consistency need to be fixed before this is safe to merge.

vidyac86:
> @vidyac86 thanks for the latest updates. I reviewed the current diff again. Several previous concerns look addressed now: the dead `separated` condition was cleaned up, explicit `separated` now logs via `logger`, the restore knob was renamed to `use_template_render_for_restore`, the YANG description now says it only applies in unified mode, and `gen_frr.conf.j2` has a template-equivalence note.
> 
> I still think this PR needs changes before merge. Could you please address the following?
> 
> ### Blocking / high-priority
> 1. **Missing `docker_routing_config_mode` renders unified config but may not run `vtysh -b`.**
>    `frr_vars.j2` emits an empty `docker_routing_config_mode` when the key is absent. `docker_init.sh` then treats empty mode as unified, renders `/etc/frr/frr.conf`, writes `service integrated-vtysh-config`, and removes per-daemon configs.
>    However, `supervisord.conf.j2` only emits `[program:vtysh_b]` when `DEVICE_METADATA.localhost.docker_routing_config_mode is defined` and explicitly equals `unified`, `split-unified`, or `separated`. For the key-absent/default path, `vtysh -b` is not scheduled, so FRR may boot without replaying the generated unified config.
>    Please align `supervisord.conf.j2` with the new effective default, or materialize the default before rendering supervisord.
> 2. **`frrcfgd.py` still defaults missing mode to `separated`.**
>    `docker_init.sh` now treats missing mode as unified, but `frrcfgd.py` still sets `self.config_mode = "separated"` when the field is absent. This makes the effective-mode contract inconsistent. In particular, `use_template_render_for_restore=false` will not trigger cached DB replay unless the mode is explicitly set to `unified`.
>    Please align the missing-mode default in `frrcfgd.py` with `docker_init.sh`, and add/confirm test coverage for missing mode + restore flag behavior.
> 3. **Handle `sonic-cfggen` failure before deleting fallback config files.**
>    In the explicit `separated` and missing/unified paths, the script runs `sonic-cfggen $CFGGEN_PARAMS` and then unconditionally switches to integrated mode and removes `bgpd.conf`, `zebra.conf`, `staticd.conf`, `bfdd.conf`, `ospfd.conf`, and `pimd.conf`.
>    Please check the `sonic-cfggen` result and fail safely before changing mode/removing per-daemon config files if rendering fails.
> 
> ### Other items to close
> 4. Please provide rendered-output comparison for old `frr.conf.j2` vs the new `gen_frr.conf.j2` path on representative topologies. The new path includes full daemon templates (`zebra.conf.j2`, `staticd.conf.j2`, `bgpd.conf.j2`), which is cleaner but not exactly the same include surface.
> 5. Please gate this with the db migrator (`sonic-utilities#3840`) or confirm the warm-reboot/upgrade ordering guarantees. This PR should not land independently if existing config can still carry explicit `separated` or missing mode through upgrade.
> 6. Please get the current required checks green. The latest rollup still showed DCO action required and Azure / impacted-area failures (`kvmtest-t0`, `kvmtest-t0-sonic`, `kvmtest-t1-lag`).
> 7. Please add operator-facing migration/deprecation docs explaining:
>    
>    * missing `docker_routing_config_mode` now defaults to unified;
>    * explicit `separated` is deprecated and rendered as unified;
>    * what operators should do if tooling depends on per-daemon config files;
>    * when to use `use_template_render_for_restore=false`.
> 
> The direction still looks reasonable, and the latest update addressed several earlier review points, but the default-mode startup path and cross-component mode consistency need to be fixed before this is safe to merge.

@StormLiangMS , for the review comments (1 to 3) which requires changes in the code, I have updated it wherever necessary to resolve all the inconsistencies. Kindly review it. Meanwhile will work on items (4 to 7) 
1)Fixed supervisord.conf.j2 to handle undefined docker_routing_config_mode as unified
2) Aligned frrcfgd.py default mode to unified (was separated)
3) Added sonic-cfggen error handling before deleting fallback config files

venkatmahalingam:
> we need to have full sonic-mgmt coverage to confirm these changes. Previously sonic had to move away from unified because of some issues found. We have to ensure that moving back to unified introduces no degradation.

Yes, we'll run the regression before this code merge.

vidyac86:
**Summary of changes Done in the conf files:**
1. docker_init.sh — Add separated mode template rendering support
In separated mode, FRR requires individual daemon config files (zebra.conf, bgpd.conf, etc.) instead of a unified frr.conf
This change renders the correct individual templates when docker_routing_config_mode is "separated"
Unified mode continues to use frr.conf.j2 as before

2. staticd.conf.j2 — Fix include path for loopback route template
Changed {% include "staticd.loopback_route.conf.j2" %} to {% include "staticd/staticd.loopback_route.conf.j2" %}
Required because the Jinja2 template root is frr/, so relative includes need the subdirectory prefix

3. supervisord.conf.j2 — Enable vtysh_b in all routing modes
Extended the condition to include "separated" mode (and undefined mode)
The vtysh_b program (vtysh batch mode) is needed to load FRR configurations in all modes, not just unified/split-unified

4. zebra.conf.j2 — Fix include path for interfaces template
Changed {% include "zebra.interfaces.conf.j2" %} to {% include "zebra/zebra.interfaces.conf.j2" %}
Same reason as staticd.conf.j2 — template root requires the subdirectory prefix

StormLiangMS:
@vidyac86 a few things I'd like to clear up before another review pass:

1. **sonic-net/sonic-utilities#3840 (db_migrator) is still OPEN / CHANGES_REQUESTED**, last touched 2026-03-29. I asked this in both my 2026-04-23 and 2026-04-29 reviews — without it, warm-reboot from an existing image carrying `docker_routing_config_mode = "separated"` is unsafe. Can you confirm this PR is gated on sonic-net/sonic-utilities#3840 landing in the same release branch?

2. **The Loopback0 wait loop in `7437ecd8`** confirms the partial-config race I flagged on 2025-05-07 is real — but it only guards Loopback0. Other late-arriving ConfigDB tables (VLAN, INTERFACE, PORTCHANNEL, BGP_NEIGHBOR) will still produce a partial `frr.conf` because unified rendering is one-shot. Can the guard be generalized to a ConfigDB convergence signal?

3. **CI on head**: `impacted-area-kvmtest-t0-sonic by Elastictest` is failing. 

4. **sonic-net/sonic-mgmt#24061 large-config stress test** @vidyac86  @deepak-singhal0408 could we have a test result for this PR?

Once these are answered I'll do another full review pass. Thanks.

rookie-who:
Hi @vidyac86 — this has approval from @prvattem and is very active. Is this ready to merge? 202605 branch cut is at end of May.

CC: @vaibhavhd @radha-danda (branch managers)

venkatmahalingam (dismissed):

venkatmahalingam:
> Hi @vidyac86 — this has approval from @prvattem and is very active. Is this ready to merge? 202605 branch cut is at end of May.
> 
> CC: @vaibhavhd @radha-danda (branch managers)

Yes, we’re currently waiting for the build checks to complete before merging this PR..

venkatmahalingam:
@StormLiangMS
1. https://github.com/sonic-net/sonic-utilities/pull/3840 (db_migrator) is still OPEN / CHANGES_REQUESTED, last touched 2026-03-29. I asked this in both my 2026-04-23 and 2026-04-29 reviews — without it, warm-reboot from an existing image carrying docker_routing_config_mode = "separated" is unsafe. Can you confirm this PR is gated on https://github.com/sonic-net/sonic-utilities/pull/3840 landing in the same release branch?
>> We don’t need any DB migration changes, since we still honor the separated mode in the configuration; however, the internal behavior now relies on a unified frr.conf. I don’t see any changes related to warm boot, am I missing something?
 
2. The Loopback0 wait loop in 7437ecd8 confirms the partial-config race I flagged on 2025-05-07 is real — but it only guards Loopback0. Other late-arriving ConfigDB tables (VLAN, INTERFACE, PORTCHANNEL, BGP_NEIGHBOR) will still produce a partial frr.conf because unified rendering is one-shot. Can the guard be generalized to a ConfigDB convergence signal?

>> Isn’t this essentially an existing race condition? Ideally, sonic-cfggen should dump the Config DB data and generate frr.conf as part of docker_init.sh. Any subsequent dynamic config-db updates should be handled by bgpcfgd/frrcfgd, which derive the FRR configuration and apply changes via vtysh.

vidyac86:
> @vidyac86 a few things I'd like to clear up before another review pass:
> 
> 1. **[Changes to make the docker_device_config_mode to unified sonic-utilities#3840](https://github.com/sonic-net/sonic-utilities/pull/3840) (db_migrator) is still OPEN / CHANGES_REQUESTED**, last touched 2026-03-29. I asked this in both my 2026-04-23 and 2026-04-29 reviews — without it, warm-reboot from an existing image carrying `docker_routing_config_mode = "separated"` is unsafe. Can you confirm this PR is gated on [Changes to make the docker_device_config_mode to unified sonic-utilities#3840](https://github.com/sonic-net/sonic-utilities/pull/3840) landing in the same release branch?
> 2. **The Loopback0 wait loop in `7437ecd8`** confirms the partial-config race I flagged on 2025-05-07 is real — but it only guards Loopback0. Other late-arriving ConfigDB tables (VLAN, INTERFACE, PORTCHANNEL, BGP_NEIGHBOR) will still produce a partial `frr.conf` because unified rendering is one-shot. Can the guard be generalized to a ConfigDB convergence signal?
> 3. **CI on head**: `impacted-area-kvmtest-t0-sonic by Elastictest` is failing.
> 4. **[[test] Add FRR large config load stress test sonic-mgmt#24061](https://github.com/sonic-net/sonic-mgmt/pull/24061) large-config stress test** @vidyac86  @deepak-singhal0408 could we have a test result for this PR?
> 
> Once these are answered I'll do another full review pass. Thanks.

@StormLiangMS , @venkatmahalingam , has answered your queries for 1) and 2), I have initiated rerun to address (3) issue. and for issue (4) @deepak-singhal0408 mentioned it was passing

vidyac86:
@StormLiangMS , @venkatmahalingam - The checks are failing due to orchagent.1779188534.54.core.gz seen during the verification of  bgp/test_frr_config_check.py and it is not related to our run.

lguohan:
the pr messed up with git history.

securely1g (commented):
Review on behalf of @lguohan:

This PR changes the default `docker_routing_config_mode` from `separated` to `unified` for FRR. A few concerns:

1. **Breaking change** — This changes default behavior for all existing deployments that haven't explicitly set `docker_routing_config_mode`. Have you verified this doesn't break config restore for existing separated-mode deployments during upgrade?
2. **Scope** — The PR has been open for over a year. Is the full changeset still needed, or can it be broken into smaller pieces?
3. **Testing** — What topology/test coverage validates that unified mode works correctly with all FRR daemons (zebra, bgpd, staticd, etc.)?

Please clarify the upgrade path for existing deployments.

StormLiangMS:
> Review on behalf of @lguohan:
> 
> This PR changes the default `docker_routing_config_mode` from `separated` to `unified` for FRR. A few concerns:
> 
> 1. **Breaking change** — This changes default behavior for all existing deployments that haven't explicitly set `docker_routing_config_mode`. Have you verified this doesn't break config restore for existing separated-mode deployments during upgrade?
> 2. **Scope** — The PR has been open for over a year. Is the full changeset still needed, or can it be broken into smaller pieces?
> 3. **Testing** — What topology/test coverage validates that unified mode works correctly with all FRR daemons (zebra, bgpd, staticd, etc.)?
> 
> Please clarify the upgrade path for existing deployments.

hi @vidyac86 Could you check this comment? And the test run into failure with this PR. 
image

venkatmahalingam:
> > Review on behalf of @lguohan:
> > This PR changes the default `docker_routing_config_mode` from `separated` to `unified` for FRR. A few concerns:
> > 
> > 1. **Breaking change** — This changes default behavior for all existing deployments that haven't explicitly set `docker_routing_config_mode`. Have you verified this doesn't break config restore for existing separated-mode deployments during upgrade?
> > 2. **Scope** — The PR has been open for over a year. Is the full changeset still needed, or can it be broken into smaller pieces?
> > 3. **Testing** — What topology/test coverage validates that unified mode works correctly with all FRR daemons (zebra, bgpd, staticd, etc.)?
> > 
> > Please clarify the upgrade path for existing deployments.
> 
> hi @vidyac86 Could you check this comment? And the test run into failure with this PR. image

For bgpcfgd:
--------------
It looks like we need to update the test scripts to use frr.conf for unified mode, since the current scripts don’t support migration to unified mode.
We should first modify the test scripts so they work with unified mode as well, and once that is in place, we can proceed with raising the PR for unified mode migration.

For frrcfgd:
-------------
We’ve already verified that both separated and unified modes can use frr.conf.
@eddieruan-alibaba  @venkit-nexthop - if you’re okay with aligning the backend behavior to use frr.conf for both separated and unified routing config modes, we can proceed with the necessary updates now and target merging them for the 202605 release. https://github.com/sonic-net/sonic-buildimage/pull/27547

rookie-who:
This PR/feature will not be included in the SONiC 202605 release. It can be retargeted for the 202611 release.

deepak-singhal0408:
Discussion from routing-wg PM meeting(05/21)..

This PR has been very long running and has some changes that can be simplified, and another PR https://github.com/sonic-net/sonic-buildimage/pull/27547 has been merged to move frrcfgd to Unified mode..  We decided, its a good idea to close this PR and can submit a fresh PR for bgpcfgd move to Unified mode.

cc: @vidyac86 @venkatmahalingam @dgsudharsan @rlhui

vidyac86:
> Discussion from routing-wg PM meeting(05/21)..
> 
> This PR has been very long running and has some changes that can be simplified, and another PR #27547 has been merged to move frrcfgd to Unified mode.. We decided, its a good idea to close this PR and can submit a fresh PR for bgpcfgd move to Unified mode.
> 
> cc: @vidyac86 @venkatmahalingam @dgsudharsan @rlhui

Closing this PR, will raise a simplified PR for bgpcfgd

PR #22600, opened 2025-05-12
[TACACS]|Fix memory leak when authenticating using tacacs

DavidZagury:
#### Why I did it
A small memory leak was found when running TACACS

##### Work item tracking

#### How I did it
A patch was created for the libnss-tacplus repository that will fix the memory leak.

#### How to verify it
Configure TACACS
Use the python script that reproduce the consumption and monitor its memory consumption.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/src/tacacs/nss/patch/0013-Partial-memleak-fix-due-to-unfreed-strdup.patch
+++ b/src/tacacs/nss/patch/0013-Partial-memleak-fix-due-to-unfreed-strdup.patch
@@ -1,0 +1,65 @@
+From ebbb87841b9e9e8b9ae7d2dc4c44afa73e332496 Mon Sep 17 00:00:00 2001
+From: Tal Berlowitz <talber@nvidia.com>
+Date: Thu, 8 May 2025 03:40:39 +0300
+Subject: [PATCH] Partial memleak fix due to unfreed strdup
+
+---
+ nss_tacplus.c | 24 +++++++++++++++++++++++-
+ 1 file changed, 23 insertions(+), 1 deletion(-)
+
+diff --git a/nss_tacplus.c b/nss_tacplus.c
+index 8e9eede..400b3c6 100644
+--- a/nss_tacplus.c
++++ b/nss_tacplus.c
+@@ -245,12 +245,32 @@ static void init_useradd_info()
+     user->shell = strdup("/bin/bash");
+ }
+ 
++static void free_useradd_info()
++{
++    useradd_info_t *user;
++
++    user = &useradd_grp_list[MIN_TACACS_USER_PRIV];
++    if(user->info)
++        free(user->info);
++    if(user->secondary_grp)
++        free(user->secondary_grp);
++    if(user->shell)
++        free(user->shell);
++
++    user = &useradd_grp_list[MAX_TACACS_USER_PRIV];
++    if(user->info)
++        free(user->info);
++    if(user->secondary_grp)
++        free(user->secondary_grp);
++    if(user->shell)
++        free(user->shell);
++}
++
+ static int parse_config(const char *file)
+ {
+     FILE *fp;
+     char buf[512] = {0};
+ 
+-    init_useradd_info();
+     fp = fopen(file, "r");
+     if(!fp) {
+         syslog(LOG_ERR, "%s: %s fopen failed", nssname, file);
+@@ -901,6 +921,7 @@ enum nss_status _nss_tacplus_getpwnam_r(const char *name, struct passwd *pw,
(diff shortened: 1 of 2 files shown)


qiluo-msft (approved):

src/tacacs/nss/patch/0013-Partial-memleak-fix-due-to-unfreed-strdup.patch
```diff
@@ -0,0 +1,65 @@
...
+index 8e9eede..400b3c6 100644
+--- a/nss_tacplus.c
++++ b/nss_tacplus.c
+@@ -245,12 +245,32 @@ static void init_useradd_info()
+     user->shell = strdup("/bin/bash");
+ }
+ 
++static void free_useradd_info()
++{
++    useradd_info_t *user;
```
liuh-80, line 20: useradd_grp_listis an array with 16 elements, current fix only handles MIN_TACACS_USER_PRIV and MAX_TACACS_USER_PRIV, it's OK for now because SONiC only using these 2 privilege level:
    static useradd_info_t useradd_grp_list[MAX_TACACS_USER_PRIV + 1];

But if SONiC config to use other privilege level in /etc/tacplus_nss.conf, the code in parse_user_priv() method will parse and initialize other elements in this array, and those initialized elements will not released by free_useradd_info().

My suggestion is: initialize and free every element in useradd_grp_list:

static void init_useradd_info()
{
    useradd_info_t *user;

    memset(useradd_grp_listm 0, sizeof(useradd_grp_list)); <== initialize every element

    user = &useradd_grp_list[MIN_TACACS_USER_PRIV];
    user->gid = 100;
    user->info = strdup("remote_user");
    user->secondary_grp = strdup("users");
    user->shell = strdup("/bin/bash");

    user = &useradd_grp_list[MAX_TACACS_USER_PRIV];
    user->gid = 1000;
    user->info = strdup("remote_user_su");
    user->secondary_grp = strdup("sudo,docker");
    user->shell = strdup("/bin/bash");
}

static void free_useradd_info()
{
    useradd_info_t *user;
    for (int idx = 0; idx <= MAX_TACACS_USER_PRIV; idx++) <== free every element
    {
        user = &useradd_grp_list[idx];
        if(user->info)
            free(user->info);
        if(user->secondary_grp)
            free(user->secondary_grp);
        if(user->shell)
            free(user->shell);
    }
}

isaacun: Hi @liuh-80 :
The problem we have is that in function init_useradd_info we allocate (with strdup) only in fields MIN_TACACS_USER_PRIV and MAX_TACACS_USER_PRIV without checking before if they are already allocated, so if they are we will leak. That will not happen with other elements because for the other elements we release before allocating

    if(priv && gid && info && group && shell) {
        useradd_info_t *user = &useradd_grp_list[priv];
        if(user->info)
            free(user->info);
        if(user->secondary_grp)
            free(user->secondary_grp);
        if(user->shell)
            free(user->shell);

        user->gid = gid;
        user->info = info;
        user->secondary_grp = group;
        user->shell = shell;
        
     So, the leak will happen only in those fields even if we will use more privilege levels.

liuh-80: There are 2 issue:
1. The useradd_grp_list is an array of c structure, which means the data member could be none zero value, because c run time will not do the initialize, it will keep the data in the original memory, unless we initialize id explicitly. if we don't initialize it, the free in your example will make the app crash because segmentation fault, this does not happen today because SONiC does not using other element.

3. As you said "for the other elements we release before allocating", then where release those new allocated memory?

liuh-80 (approved):
Having this fix is better than not having it.

liat-grozovik:
@qiluo-msft can you go a head and merge it? need to decide if to take to all the branches listed or not. please update the label accordingly.

bingwang-ms:
@yejianquan  for 202505 cherry-pick
@kperumalbfn  for 202411 cherry-pick

qiluo-msft:
@DavidZagury Could you raise a PR for 202311?

DavidZagury:
> Cherry-pick PR to 202411: #23150

@qiluo-msft 202311 PR: https://github.com/sonic-net/sonic-buildimage/pull/23226

PR #22616, opened 2025-05-14
Remove PFC/TC_TO_PG and PFC_TO_QUEUE map for lossy SKUs

kperumalbfn:
#### Why I did it
For LOSSY SKUs, PFC_TO_PG, TC_TO_PG and PFC_TO_QUEUE maps are not valid, so removing them from the default templates.

##### Work item tracking
- Microsoft ADO **(number only)**: 32883774

#### How I did it
Updated qos_config.j2 to remove PFC/TC_TO_PG and PFC_TO_QUEUE maps for lossy SKUs.

#### How to verify it
Verified on lossy SKUs with the latest j2 file.

### DIFF
--- a/files/build_templates/qos_config.j2
+++ b/files/build_templates/qos_config.j2
@@ -466,6 +466,7 @@
 {% if LOSSLESS_TC|length > 0 %}
             "pfc_to_queue_map": "AZURE"
 {% endif %}
+{% endif %}
         }{% if not loop.last %},{% endif %}
 
 {% endfor %}


r12f (approved):

r12f:
hi @kperumalbfn , looks like CI failures might be real, do you mind to help check it?

image

kperumalbfn:
@stephenxs could you please check and let me know if you have any comments.

files/build_templates/qos_config.j2
```diff
@@ -387,11 +391,19 @@
             "pfcwd_sw_enable" : "{{ LOSSLESS_TC|join(',') }}",
 {% endif %}
 {% if port not in PORT_DPC %}
+{% if LOSSLESS_TC|length > 0 %}
             "tc_to_pg_map"    : "AZURE",
+{% else %}
```
stephenxs: Why do you remove TC TO PG map?

kperumalbfn: What is the default PG for all the traffic, will that be "0"? Do we still need TC_TO_PG_MAP, please confirm.

stephenxs: The default is 0, that is correct 
But why do you rely on default values? Any issue to configure them explicitly?

kperumalbfn: Updated the PR to include TC_TO_PG to map all TCs to default PG(0)

r12f:
@kperumalbfn do you mind to help with the merge?

files/build_templates/qos_config.j2
```diff
@@ -441,9 +441,11 @@
 {% if port in port_names_list_extra_queues %}
             "pfc_to_pg_map"   : "AZURE_DUALTOR",
 {% else %}
+{% if LOSSLESS_TC|length > 0 %}
```
qiluo-msft, line 444: The changes in this PR will be overwritten by golden config in prod? Better to keep image behavior, and change golden config generation (not in this repo).

kperumalbfn: Thanks @qiluo-msft. Prod config will be from the tool

kperumalbfn:
Closing this PR for now to absorb this change later.

PR #22622, opened 2025-05-15
Introduce Xsight X2-Based Device for SONiC using Accton platform

deran1980:
#### Why I did it

Adding a new device platform support in sonic-buildimage for Xsight Labs devices (based on Accton platforms).
The new device platform name is ES9618XX and currently supports a single SKU O16x400G and it is based on the latest
chip from Xsight - X2.

##### Work item tracking

#### How I did it

- Added all Accton-relevant platform drivers and code.
- Uploaded Xsight-based platform support 
- Uploaded relevant precompiled Debian packages to the Xsight SONiC repository.

#### How to verify it

Run the compiled SONiC image for the Xsight platform on a Xsight Labs switch and verify successful boot and functionality using Accton ES9618 plaftorm based on the X2 chip.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

In order to build a SONiC image for the Xsight platform, use the following commands:

```bash
SONIC_BUILD_JOBS=8 SONIC_DPKG_CACHE_METHOD=none make target/sonic-xsight.bin

### CODE REVIEW

File: device/accton/x86_64-accton_es9618xx-r0/Accton-ES9618XX-O16x400G/copp_custom.json
Comment (CharlieChenEC) [inline device/accton/x86_64-accton_es9618xx-r0/Accton-ES9618XX-O16x400G/copp_custom.json:1]: What is the timing that the setting in the file takes effect?
What is the purpose to specify "snmp" trap in this file?

File: device/accton/x86_64-accton_es9618xx-r0/custom.json
Comment (CharlieChenEC) [inline device/accton/x86_64-accton_es9618xx-r0/custom.json:1]: With the enhancement in https://github.com/sonic-net/sonic-buildimage/pull/20826, I think these settings can be added to the file 'device/accton/x86_64-accton_es9618xx-r0/platform.json'

Comment (MichaelZaidman) [reply device/accton/x86_64-accton_es9618xx-r0/custom.json:1]: Thanks for the suggestion! That makes sense - with the enhancement in [#20826](https://github.com/sonic-net/sonic-buildimage/pull/20826), adding these settings to `device/accton/x86_64-accton_es9618xx-r0/platform.json` seems appropriate. I'll update the PR accordingly.

File: device/accton/x86_64-accton_es9618xx-r0/Accton-ES9618XX-O16x400G/copp_custom.json
Comment (aparkhomenko-xsight) [reply device/accton/x86_64-accton_es9618xx-r0/Accton-ES9618XX-O16x400G/copp_custom.json:1]: This file is not needed any more, so I deleted it

File: device/accton/x86_64-accton_es9618xx-r0/fast-reboot_plugin [hunk 1]
```diff
@@ -0,0 +1,24 @@
...
+if [ $? -eq 0 ]; then
+	echo "$fname: Unload xpci module" | tee /dev/kmsg
+	rmmod xpci
+	if [ $? -eq 1 ]; then
+		echo "$fname: Cannot reset X2 while xpci is loaded" | tee /dev/kmsg
+		exit 1
+	fi
+fi
+
+/opt/xplt/utils/es9618x_reset_x2.sh
```
CharlieChenEC, line 19: According to the design of fast-reboot (https://github.com/sonic-net/SONiC/blob/master/doc/fast-reboot/fastreboot.pdf), the data plane remains operational when this file is invoked. I'm wondering if the execution of "es9618x_reset_x2.sh" will stop the data plane from working?

vincentchiang-ec:
Looks good to me. Approved, as most platform-related tests have passed.

platform/xsight/sonic-platform-modules-accton/debian/changelog
```diff
@@ -0,0 +1,6 @@
+sonic-accton-platform-modules (1.1) unstable; urgency=low
+
+  * Add support for Accton ES9632XX
```
CharlieChenEC: "ES9632XX" --> Should be ES9618XX?

platform/xsight/sonic-platform-modules-accton/common/udev/70-persistent-net.rules
CharlieChenEC, line 1: These rules seem to be board-dependent ones. Suggest to move the file under "platform/xsight/sonic-platform-modules-accton/es9618xx/udev/"

platform/xsight/files/link/70-xsight-eth10g0.link
CharlieChenEC, line 1: What is the purpose of adding this file?
It seems that the files under "platform/xsight/files/link/" will be copied to $FILESYSTEM_ROOT/lib/systemd/network/ in "sonic_debian_extension.j2". The "Path" in this file contains the pcie device address to match, and the pcie device address could be changed among different hardware boards. It would be better to move the installation of the file into platform deb post-install script.

device/accton/x86_64-accton_es9618xx-r0/fast-reboot_plugin
MichaelZaidman: Yes, the script does stop the data plane earlier than expected.

This is a temporary workaround implemented by our engineer to pass the fast-reboot test in the SONiC management test suite:
[sonic-mgmt test_reboot.py#L213](https://github.com/sonic-net/sonic-mgmt/blob/master/tests/platform_tests/test_reboot.py#L213),
which was consistently failing with the following error:
AssertionError: Not all transceivers are detected or interfaces are up in 300 seconds

Here’s the related internal commit message for context:

> SON-1984: es9618xx: fast-reboot_plugin: fix interfaces oper state after running fast-reboot
> Added a fast-reboot_plugin script to unload the XPCI module and reset the chip, as specified in
> Clause 6 of the [Fast-Reboot implementation details](https://github.com/sonic-net/SONiC/wiki/Fast-Reboot#implementation-details).
> Without this script, interfaces remained in the 'down' state after a fast reboot, as shown by
> the show interfaces description command.
> 
> The addition of this script allowed the fast-reboot test to pass by ensuring the interfaces
> transitioned to the 'up' state, thereby avoiding a timeout in the test.

To clarify: full fast-reboot support is not yet implemented. We plan to address it properly in a future PR with a design that aligns with the official fast-reboot flow and minimizes data plane disruption.

platform/xsight/sonic-platform-modules-accton/debian/changelog
MichaelZaidman: Good catch, thanks! You're absolutely right - it should be ES9618XX. I'll correct it in the next update.

platform/xsight/sonic-platform-modules-accton/common/udev/70-persistent-net.rules
MichaelZaidman: Thanks for spotting this - it’s actually a leftover from a previous project and isn’t needed anymore. I’ll remove it in the next update.

platform/xsight/files/link/70-xsight-eth10g0.link
MichaelZaidman: You're right - these files are platform-specific, and moving their installation to the platform's Debian post-install script makes sense.

The reason we introduced the 70-xsight-eth10g0.link file is to address an intermittent issue where Intel's 10G hardware is enumerated on the PCI bus before the 1G hardware. In such cases, the ice driver is probed before the igb driver, and the system incorrectly assigns eth0 to the 10G interface. This happens in roughly 10% of reboots.

To avoid this misassignment, we use systemd.link file, which is processed by systemd-udevd earlier than traditional udev rules. This ensures a consistent and deterministic interface naming, regardless of enumeration order.

Thanks for pointing it out - we'll move this logic into the platform’s Debian post-install script accordingly.

device/accton/x86_64-accton_es9618xx-r0/fast-reboot_plugin
CharlieChenEC: Understood.Thanks for your explanation.

device/accton/x86_64-accton_es9618xx-r0/custom.json
MichaelZaidman: Done in https://github.com/sonic-net/sonic-buildimage/pull/22622/commits/9d22e6df515ab734dbb3836972ae4057c1c78bb9 commit.

platform/xsight/sonic-platform-modules-accton/debian/changelog
MichaelZaidman: Fixed in https://github.com/sonic-net/sonic-buildimage/pull/22622/commits/269112cbe100ced577fe0c46d73f963b1135362f commit.

platform/xsight/files/link/70-xsight-eth10g0.link
MichaelZaidman: Done in https://github.com/sonic-net/sonic-buildimage/pull/22622/commits/269112cbe100ced577fe0c46d73f963b1135362f commit.

platform/xsight/sonic-platform-modules-accton/common/udev/70-persistent-net.rules
MichaelZaidman: Removed in https://github.com/sonic-net/sonic-buildimage/pull/22622/commits/9b47b5232c65bcaecea1b3f7256ad1853907f686 commit.

CharlieChenEC (approved):
All of my review comment has been revised. I think the PR is ready for merge.

device/accton/x86_64-accton_es9618xx-r0/pmon_daemon_control.json
```diff
@@ -0,0 +1,4 @@
+{
+    "skip_xcvrd_cmis_mgr": true
```
prgeor: @deran1980 Why not use community based Xcvrd to manage?

platform/xsight/xpci/Makefile
```diff
@@ -0,0 +1,34 @@
+.ONESHELL:
+SHELL = /bin/bash
+.SHELLFLAGS += -e
+
+XPCI_DEB = xpci-modules-$(KVERSION)_$(XPCI_VERSION)_amd64.deb
+XPCI_URL_PREFIX = "https://raw.githubusercontent.com/xsightlabs/SONiC/main/amd64/kernel"
```
prgeor: @deran1980 Why this is not part of sonic-linux-kernel? @saiarcot895 for vs.

ktsym: Hi @prgeor. This Debian package includes the full Xsight XPCI driver, pre-compiled, as part of the Xsight SDK and X2 chip functionality. The driver is separate from sonic-linux-kernel because it is specific to the Xsight switch ASIC. We believe this modular approach helps avoid kernel rebuilds and keeps the base kernel clean and generic. It is our understanding that this is also the common approach used by other vendors.

device/accton/x86_64-accton_es9618xx-r0/pmon_daemon_control.json
MichaelZaidman: You’re absolutely right, and we are actively transitioning to SONiC’s native xcvrd-based transceiver management. I plan to submit the relevant commit shortly as part of this migration.

Originally, we used the /opt/xplt/utils/xcvrs tools in our lab for signal integrity (SI) characterization of modules on our Ubuntu-based XsightOS NOS. When we began working with SONiC, these tools helped fill the gap caused by limited CMIS and SI configuration support in earlier versions.

Given their reliability across different environments, we extended their use to the SONiC 202311 release and, eventually, into this PR. This has been a temporary measure, and we’re now aligning with the community's standard approach moving forward.

platform/xsight/xpci/Makefile
saiarcot895: If I'm reading this right, this suggests that the kernel driver is precompiled, and that precompiled version is being downloaded here. That will not work for SONiC's purposes. If there is a kernel version change or kernel config change, then the kernel modules will need to be recompiled separately from all of the other kernel modules.

platform/xsight/kvm-image.mk
saiarcot895, line 1: Is there a use case for having a KVM image for xsight, separate from the `vs` platform?

platform/xsight/docker-syncd-xsight/Dockerfile.j2
```diff
@@ -0,0 +1,31 @@
+{% from "dockers/dockerfile-macros.j2" import install_debian_packages, install_python_wheels, copy_files %}
+FROM docker-config-engine-bullseye-{{DOCKER_USERNAME}}:{{DOCKER_USERTAG}}
```
saiarcot895: Why is bullseye being used here? Why is bookworm not being used?

platform/xsight/xplt.mk
```diff
@@ -0,0 +1,9 @@
+# Xsight xplt tools package for es9618xx
+
+# Add platform specific tools
+
+ACCTON_ES9618XX_XPLT_PLATFORM_MODULE = xplt-es9618xx-sonicnos_3.3.3-51-g1a65a01_amd64.deb
+ES9618XX_XPLT_SRV = "https://raw.githubusercontent.com/xsightlabs/SONiC/main/amd64/utils/"
```
saiarcot895: This (along with `XPLT_FLAGS` in the dep file) feels messy, having binary packages be downloaded from a git repo from the main branch without any sort of versioning. This will make reproducible builds hard, and older versions of the SONiC repo cannot be easily rebuilt with the same content. This applies to all of the packages that are being downloaded from that repo.

I would recommend one of two approaches:
1. Add that repo as a submodule here, so that a specific tag is checked out here and so the same content will be used always.
2. Make an official tagged release in Github, and download the binaries from that release.

This also removes the `wget` calls from the `.dep` files, which can slow down the start of the build and cause false positives w.r.t rebuilds.

platform/xsight/xpci/Makefile
ktsym: Hi @saiarcot895. Ok, we will prepare a DKMS-based Debian package that includes the driver sources and compiles automatically for each kernel version. To our understanding, this approach aligns with SONiC requirements and ensures independence from any specific kernel build.

platform/xsight/docker-syncd-xsight/Dockerfile.j2
deran1980: when we started this PR the bullseye was still relevant for the version we relied on, but we are now in the process of moving to bookworm and we will fix this one also shortly.

platform/xsight/kvm-image.mk
deran1980: hi @saiarcot895 we actually have an internal VM version that we do not intend to open it currently. 
this file is not needed and we will remove it.

platform/xsight/xplt.mk
MichaelZaidman: Thanks for the detailed feedback.

We maintain versioned Debian packages under the xsightlabs/SONiC GitHub repo using a naming convention that encodes the platform, NOS, and package version (e.g., xplt-es9618xx-sonicnos_3.3.3-51-g1a65a01_amd64.deb). This ensures backward compatibility and tight coupling between each SONiC build and the corresponding xplt package.

Regarding your proposals:
**Using a git submodule** - our packaging model relies on distinct .deb filenames per target platform/NOS version. A submodule approach that builds from source or reuses the same package name across versions isn't compatible with this setup.

**Publishing versioned packages as GitHub releases** - this approach aligns well with our model. We can move the .deb artifacts into the Releases section of the GitHub repo and reference them via immutable version tags. This will decouple us from main branch volatility and improve reproducibility of builds.

We’ll proceed with implementing this release-based approach unless you have further recommendations. If there are examples in SONiC that follow a similar pattern, we’d appreciate it if you could point us to them for alignment.

device/accton/x86_64-accton_es9618xx-r0/pmon_daemon_control.json
MichaelZaidman: Done in https://github.com/sonic-net/sonic-buildimage/pull/22622/commits/5f3a3e0a3c23898edde3de5d533534f0cc443938 commit.

platform/xsight/xplt.mk
MichaelZaidman: Hi @saiarcot895,

Thank you again for your valuable feedback.

Following your comments, we held an internal discussion to evaluate the available approaches for managing binary artifacts. We specifically considered the two options you proposed, along with current practices in similar upstream contributions.

We reviewed two common methods:

1. **Binaries as Regular Git Files** (as used by Marvell)
   *Pros:*

   * Simplicity in implementation and usage
   * Fine-grained control: each binary can evolve independently
   * Straightforward per-file tracking and updating
  
   *Cons:*
   * Rapid repository size growth
   * Lacks native release/version management

2. **GitHub Releases for Binary Artifacts** (as used by Mellanox)
   *Pros:*

   * Purpose-built for managing binary deliverables
   * Clear versioning and release visibility
   * Suitable for formal artifact grouping
 
   *Cons:*
   * Coarse-grained updates: changes require retagging or creating a new release
   * Updating a single binary necessitates a new full release

While we recognize the strengths of the GitHub Releases model and your valid concerns about reproducibility and version traceability, the team has agreed to proceed with the "Binaries as Regular Git Files" model for now, primarily due to its flexibility and simplicity during the current development phase.

That said, we acknowledge and aim to address the critical issues you raised:

* We introduced explicit versioning for key binary components such as the `xpci` kernel module, to avoid ambiguous or floating references.
* We aligned our repository layout with Marvell’s approach for components like `sai-plugin` to improve clarity and consistency.
* We removed wget calls from .dep files.

We believe these steps will significantly improve reproducibility and maintainability, even within the chosen approach.

platform/xsight/kvm-image.mk
aparkhomenko-xsight: Done in https://github.com/sonic-net/sonic-buildimage/pull/22622/commits/60508346d505f780d173a84bf9c843fb73c73b14 commit.

platform/xsight/docker-syncd-xsight/Dockerfile.j2
aparkhomenko-xsight: Done in https://github.com/sonic-net/sonic-buildimage/pull/22622/commits/d01d74512881bf67c34f73d216f00c64e21db705 commit.

platform/xsight/xpci/Makefile
aparkhomenko-xsight: Done in https://github.com/sonic-net/sonic-buildimage/pull/22622/commits/d8e9d091518a7aad00fda518e59abf0b9c2ad69d commit.

platform/xsight/xplt.mk
aparkhomenko-xsight: .dep files have been updated in https://github.com/sonic-net/sonic-buildimage/pull/22622/commits/bda6dd21d9600623fabe5d3d4989c58f8f7c7cc9 and https://github.com/sonic-net/sonic-buildimage/pull/22622/commits/10a30d527d22470344a3667a29566261abfeaf00 commits.

vincentchiang-ec:
@lguohan Could you please help review?

files/build_templates/docker_image_ctl.j2
```diff
@@ -740,12 +744,17 @@ start() {
         --env "CONTAINER_NAME"=$DOCKERNAME \
         --env "SYSLOG_TARGET_IP"=$SYSLOG_TARGET_IP \
         --env "PLATFORM"=$PLATFORM \
+{%- if sonic_asic_platform == "xsight" %}
+{%- if docker_container_name == "syncd" %}
+        -p 0.0.0.0:31000:31000/tcp \
+        -p 0.0.0.0:31002:31002/tcp \
+        -p 0.0.0.0:31001:31001/tcp \
```
lguohan: why do we need all these tcp socket? i believe the only channel syncd expose should be SAI interface to the rest of the world.

deran1980: Hi @lguohan 
We are using these sockets as a debug sockets for an internal tool to debug our sdk. 
it is not needed for non developers so you are right and we decided to remove it from our code.

lguohan (dismissed):

deran1980:
Hi @lguohan @prgeor PR is rebased, passed all tests and ready to be merged. 
Appreciate your support on this.

PR #22628, opened 2025-05-15
[warm reboot] Using sonic-db-cli to optimize DB access time for warm reboot to come up on time

mlok-nokia:
#### Why I did it
With sonic-cfggen, it introduces more time for system to come up during wram reboot.  This will cause LACP session terminated.  This PR modified the code to use sonic-db-cli instead of sonic-cfggen to optimize the DB base access to reduce the bootup time. It fixes issue https://github.com/sonic-net/sonic-buildimage/issues/22438

##### Work item tracking

#### How I did it
Modify the asic_status.sh to use sonic-db-cli to get the platform name.  Also only does when the /etc/sonic/chassisdb.conf exist.  This will avoid to add more time to the reboot when the system is not VOQ system (Supervisor and DisaggregatedT2).  
Modify the chronyd-starter.sh to use sonic-db-cli instead of sonic-cfggen. 

#### How to verify it

The OC test should pass.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202503

- [ ] 
- [ ]


tirupatihemanth (approved):

dgsudharsan:
@vaibhavhd @bingwang-ms  Can you please review this fix?

rlhui:
@mlok-nokia  this is not for chassis, is it?

files/image_config/chrony/chronyd-starter.sh
```diff
@@ -1,8 +1,8 @@
 #!/bin/sh
 
-VRF_ENABLED=$(/usr/local/bin/sonic-cfggen -d -v 'MGMT_VRF_CONFIG["vrf_global"]["mgmtVrfEnabled"]' 2> /dev/null)
+VRF_ENABLED=$(sonic-db-cli CONFIG_DB HGET "MGMT_VRF_CONFIG|vrf_global" "mgmtVrfEnabled")
```
ymd-arista: Nit: We don't seem to be using error message from the `sonic-db-cli` command anywhere else if it errors out. Might as well redirect that to `/dev/null` as before.

mlok-nokia: Fixed as suggested.

ymd-arista: Forgot to mention that same is needed for `VRF_CONFIGURED`.

Apart from that, LGTM.

ymd-arista (approved):

files/image_config/chrony/chronyd-starter.sh
mlok-nokia: Fixed as suggested

mlok-nokia:
@arlakshm  Hi Arvind,  please help to review this PR.  Thanks

deepak-singhal0408:
@vaibhavhd  @arlakshm could you sign-off this PR.. thanks.,

files/scripts/asic_status.sh
```diff
@@ -1,13 +1,14 @@
...
-PLATFORM="$(sonic-cfggen -H -v DEVICE_METADATA.localhost.platform)"
-PLATFORM_ENV_CONF=/usr/share/sonic/device/$PLATFORM/platform_env.conf
-[ -f $PLATFORM_ENV_CONF ] && . $PLATFORM_ENV_CONF
-
 is_chassis_supervisor() {
-    if [ -f /etc/sonic/chassisdb.conf ] && [[ $disaggregated_chassis -ne 1 ]]; then
-        true
-        return
+    if [ -f /etc/sonic/chassisdb.conf ]; then
+        PLATFORM=$(sonic-db-cli CONFIG_DB HGET "DEVICE_METADATA|localhost" "platform")
```
arlakshm, line 5: on first boot the config db is empty, will it work in this case?

mlok-nokia: For the first time boot without config_db#.json file, it works/behaves the same as the before --- for multi-asic,  swss service will be in the failure state without config.

vaibhavhd (approved):

PR #22631, opened 2025-05-15
Bound eventd data cache vector and map and call malloc trim after emptied

zbud-msft:
##### Work item tracking

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]

### DIFF
--- a/src/sonic-eventd/src/eventd.cpp
+++ b/src/sonic-eventd/src/eventd.cpp
@@ -28,9 +28,11 @@ using namespace std;
 using namespace swss;
 
 #define MB(N) ((N) * 1024 * 1024)
-#define EVT_SIZE_AVG 150
+#define EVT_SIZE_AVG 250
 
-#define MAX_CACHE_SIZE (MB(100) / (EVT_SIZE_AVG))
+#define MAX_CACHE_SIZE (MB(50) / (EVT_SIZE_AVG))
+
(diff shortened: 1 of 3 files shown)


qiluo-msft:
Do you have a sonic-mgmt case to prevent future regression?

src/sonic-eventd/src/eventd.cpp
```diff
@@ -343,8 +346,10 @@ capture_service::init_capture_cache(const event_serialized_lst_t &lst)
             sequence_t seq;
 
             if (validate_event(event, rid, seq)) {
-                m_pre_exist_id[rid] = seq;
-                m_events.push_back(*itc);
+                if (itc->find("heartbeat") == string::npos && VEC_SIZE(m_events) < m_cache_max) { // filter out heartbeat events and bound m_events
+                    m_pre_exist_id[rid] = seq;
```
qiluo-msft, line 350: >m_pre_exist_id [](http://example.com/codeflow?start=20&length=14)

Since m_pre_exist_id is also possible increasing size, would you set uplimit, and check size before this line?

zbud-msft: Done

PR #22658, opened 2025-05-21
Refactor auditd rules reload check

xincunli-sonic:
#### Why I did it
Auditd_watchdog reloaded rules during the health check, which generates a lot of CONFIG_CHANGE, SYSCALL and SOCKADD noises log.

##### Work item tracking
- Microsoft ADO **(number only)**: 32301731

#### How I did it

Move reload logic from watchdog to auditd self.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ]


dockers/docker-auditd/config_checker.py
```diff
@@ -31,76 +44,122 @@ def run_command(cmd):
...
+            return hwsku
+    except Exception as e:
+        logger.log_error("Failed to get HW SKU: {}".format(str(e)))
     return None
 
 
 def is_auditd_rules_configured():
-    EXPECTED_HASH = "f88174f901ec8709bacaf325158f10ec62909d13"
-    NOKIA_EXPECTED_HASH = "bd574779fb4e1116838d18346187bb7f7bd089c9"
     hwsku = get_hwsku()
```
qiluo-msft, line 66: >hwsku [](http://example.com/codeflow?start=4&length=5)

It may be None, could you double check the following code still behave gracefully if this happens?

dockers/docker-auditd/config_checker.py
```diff
@@ -31,76 +44,122 @@ def run_command(cmd):
...
-    if "DEVICE_METADATA" in config_db and "hwsku" in config_db["DEVICE_METADATA"]["localhost"]:
-        return str(config_db['DEVICE_METADATA']['localhost']['hwsku'])
+    try:
+        with open("/etc/sonic/config_db.json") as fp:
+            config_db = json.load(fp)
+        if "DEVICE_METADATA" in config_db and "hwsku" in config_db["DEVICE_METADATA"]["localhost"]:
+            hwsku = str(config_db['DEVICE_METADATA']['localhost']['hwsku'])
+            logger.log_info("Detected HW SKU: {}".format(hwsku))
+            return hwsku
+    except Exception as e:
```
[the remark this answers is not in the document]
xincunli-sonic: My understanding is that here we just logged error, will not throw or raise again, so, does not matter for exception type.

dockers/docker-auditd/config_checker.py
xincunli-sonic: In most cases, hwsku is a very important field, it should exist, this was a protection, and I did add more check on None case.

dockers/docker-auditd/config_checker.py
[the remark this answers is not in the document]
qiluo-msft: If there is no try/except, the whole script will fail clearly and trigger external awareness. However return a None and keep this script going will be more harmful.

dockers/docker-auditd/config_checker.py
```diff
@@ -31,76 +44,122 @@ def run_command(cmd):
...
     return p.returncode, error
 
 
 def get_hwsku():
-    with open("/etc/sonic/config_db.json") as fp:
-        config_db = json.load(fp)
-    if "DEVICE_METADATA" in config_db and "hwsku" in config_db["DEVICE_METADATA"]["localhost"]:
-        return str(config_db['DEVICE_METADATA']['localhost']['hwsku'])
+    try:
+        with open("/etc/sonic/config_db.json") as fp:
```
qiluo-msft, line 54: Seems parsing json file directly does not work for multi-asic. Suggest use `sonic-cfggen -d -v DEVICE_METADATA.localhost.hwsku`.

dockers/docker-auditd/config_checker.py
qiluo-msft: Still applicable

PR #22711, opened 2025-05-27
Security Cipher Module: Enabled Backup and Restore support for NOS Upgrades

nmoray:
#### Why I did it
This is required majorly for NOS upgrade scenario to backup cipher_pass file stored at /etc dir.

#### How I did it
Added pre/post migration scripts

#### How to verify it
NOS upgrade

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

- [x] 202311

https://github.com/Azure/sonic-buildimage/blob/master/src/sonic-yang-models/doc/Configuration.md

### DIFF
--- a/files/image_config/config-setup/01-post-security-cipher
+++ b/files/image_config/config-setup/01-post-security-cipher
@@ -1,0 +1,11 @@
+#!/bin/bash
+
+set -e
+
+# Restore cipher_pass file from persistent storage
+if [ -f /host/security_cipher/cipher_pass ]; then
+    cp /host/security_cipher/cipher_pass /etc/cipher_pass
+    chmod 640 /etc/cipher_pass
+    echo "Restored /host/security_cipher/cipher_pass  to /etc/"
+fi
+
--- a/files/image_config/config-setup/01-pre-security-cipher
+++ b/files/image_config/config-setup/01-pre-security-cipher
@@ -1,0 +1,13 @@
+#!/bin/bash
+
+set -e
+
+# Ensure old_config directory exists
+mkdir -p /host/security_cipher
+
+# Copy cipher_pass file to persistent storage
+if [ -f /etc/cipher_pass ]; then
+    cp /etc/cipher_pass /host/security_cipher/cipher_pass 
+    echo "Saved /etc/cipher_pass to /host/security_cipher/"
+fi
+
--- a/src/sonic-yang-models/yang-models/sonic-system-tacacs.yang
+++ b/src/sonic-yang-models/yang-models/sonic-system-tacacs.yang
@@ -131,10 +131,11 @@ module sonic-system-tacacs {
                     default 5;
                 }
 
-		leaf key_encrypt {
-		    type boolean;
-		    description "Indicates if the passkey is encrypted.";
-		}
+                leaf key_encrypt {
+                    type boolean;
+                    default false;
+                    description "Indicates if the passkey is encrypted.";
+                }
 
                 leaf passkey {
                     type string {
--- a/files/build_templates/sonic_debian_extension.j2
+++ b/files/build_templates/sonic_debian_extension.j2
@@ -642,6 +642,12 @@ j2 files/build_templates/config-setup.service.j2 | sudo tee $FILESYSTEM_ROOT_USR
 sudo cp $IMAGE_CONFIGS/config-setup/config-setup $FILESYSTEM_ROOT/usr/bin/config-setup
 sudo mkdir -p $FILESYSTEM_ROOT/etc/config-setup
 sudo cp $IMAGE_CONFIGS/config-setup/config-setup.conf $FILESYSTEM_ROOT/etc/config-setup/config-setup.conf
+sudo mkdir -p $FILESYSTEM_ROOT/etc/config-setup/config-migration-pre-hooks.d
+sudo cp $IMAGE_CONFIGS/config-setup/01-pre-security-cipher $FILESYSTEM_ROOT/etc/config-setup/config-migration-pre-hooks.d/01-pre-security-cipher
+chmod +x $FILESYSTEM_ROOT/etc/config-setup/config-migration-pre-hooks.d/01-pre-security-cipher
+sudo mkdir -p $FILESYSTEM_ROOT/etc/config-setup/config-migration-post-hooks.d
+sudo cp $IMAGE_CONFIGS/config-setup/01-post-security-cipher $FILESYSTEM_ROOT/etc/config-setup/config-migration-post-hooks.d/01-post-security-cipher
+chmod +x $FILESYSTEM_ROOT/etc/config-setup/config-migration-post-hooks.d/01-post-security-cipher
 echo "config-setup.service" | sudo tee -a $GENERATED_SERVICE_FILE
 sudo LANG=C chroot $FILESYSTEM_ROOT systemctl enable config-setup.service
 


nmoray:
@qiluo-msft please help me understand [ms_conflict](https://github.com/sonic-net/sonic-buildimage/pull/22711/checks?check_run_id=42970042984) failure. I don't know why this is failing. 
Additionally, please approve the PR in case there are no new comments.

anders-nexthop:
@qiluo-msft @lguohan are there any further comments on this PR? If not, can we merge it and unblock this feature?

@nmoray this is what the MS-Conflict says:
```
failure: @StormLiangMS,@liushilongbuaa Conflict already exists in master
Please wait a few hours to run ms_conflict again!
'/azpw ms_conflict'
```
I don't know what this check is for, but maybe just rebase and push again to re-run it and see if that resolves things?

nmoray:
> @qiluo-msft @lguohan are there any further comments on this PR? If not, can we merge it and unblock this feature?
> 
> @nmoray this is what the MS-Conflict says:
> 
> ```
> failure: @StormLiangMS,@liushilongbuaa Conflict already exists in master
> Please wait a few hours to run ms_conflict again!
> '/azpw ms_conflict'
> ```
> 
> I don't know what this check is for, but maybe just rebase and push again to re-run it and see if that resolves things?

Yeah @anders-nexthop, I have already tried this option but no luck.

nmoray:
@anders-nexthop @qiluo-msft Please review the updated code and share the comments if any.

src/sonic-py-common/sonic_py_common/security_cipher.py
```diff
@@ -12,89 +12,140 @@
...
-        if feature_type == 'NA':
-            syslog.syslog(syslog.LOG_ERR, "__write_passwd_file: Invalid feature type: {}".format(feature_type))
+    def _load_registry(self):
+        if not os.path.exists(CIPHER_PASS_FILE):
+            return {}
+        try:
+            with open(CIPHER_PASS_FILE, 'r') as f:
+                return json.load(f)
+        except Exception as e:
+            syslog.syslog(syslog.LOG_ERR, "del_cipher_passwd: Exception occurred: {}".format(e))
```
anders-nexthop: "del_cipher_passwd" seems like maybe a copy/paste error? should that be "load_registry" instead?

src/sonic-py-common/sonic_py_common/security_cipher.py
```diff
@@ -12,89 +12,140 @@
...
+    def set_feature_password(self, feature_type, password):
+        """
+        Set a new password for a feature type.
+        It will not update if already exist.
+        """
+        data = self._load_registry()
+        if feature_type not in data:
+            data[feature_type] = {"callbacks": [], "password": None}
+        if data[feature_type]["password"] is not None:
+            syslog.syslog(syslog.LOG_INFO, "set_feature_password: Password already set for feature {}, not updating the new password.".format(feature_type))
```
anders-nexthop, line 178: is the idea that the password should be rotated, rather than re-set? should we mention something about it in the log?

src/sonic-py-common/sonic_py_common/security_cipher.py
```diff
@@ -12,89 +12,140 @@
...
+        """
+        On each call, read JSON data fresh from disk. Update password if provided,
+        and call all registered callbacks with the latest password.
+        """
+        data = self._load_registry()
+        if feature_type not in data:
+            syslog.syslog(syslog.LOG_ERR, "rotate_feature_passwd: No callbacks registered for {}".format(feature_type))
+            return
+
+        if new_password is not None:
```
anders-nexthop: so if new_password is None, we will still rotate but using the existing password (so we will still trigger all the callbacks). Do I understand that correctly?

anders-nexthop (commented):
Looks good, the rotation code is a good change.

src/sonic-py-common/sonic_py_common/security_cipher.py
nmoray: yeah

src/sonic-py-common/sonic_py_common/security_cipher.py
nmoray: Sure.

src/sonic-py-common/sonic_py_common/security_cipher.py
nmoray: You are right. This is help in synchronising the passkeys for a given feature in case one of them got out of sync.

files/image_config/config-setup/01-post-security-cipher
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+set -e
+
+# Restore cipher_pass file from persistent storage
+if [ -f /host/security_cipher/cipher_pass ]; then
+    cp /host/security_cipher/cipher_pass.json /etc/cipher_pass.json
+    chmod 640 /etc/cipher_pass.json
+    echo "Restored /host/security_cipher/cipher_pass.json  to /etc/"
```
anders-nexthop: nit: extra space after ".json"

src/sonic-py-common/sonic_py_common/security_cipher.py
```diff
@@ -101,12 +102,13 @@ def set_feature_password(self, feature_type, password):
             data[feature_type] = {"callbacks": [], "password": None}
         if data[feature_type]["password"] is not None:
             syslog.syslog(syslog.LOG_INFO, "set_feature_password: Password already set for feature {}, not updating the new password.".format(feature_type))
+            syslog.syslog(syslog.LOG_INFO, "set_feature_password: Note: Make use of rotate_feature_passwd() method for udpating the existing pass")
```
anders-nexthop: nit: s/udpating/updating

src/sonic-py-common/tests/test_security_cipher.py
```diff
@@ -30,9 +30,9 @@
     "LDAP": {"callbacks": [], "password": None}
 }
 
-def dummy_cb(password): pass
-def cb1(password): pass
-def cb2(password): pass
+def dummy_cb(table_info): pass
```
anders-nexthop: shouldn't these have 'secret' as well?

src/sonic-py-common/tests/test_security_cipher.py
```diff
@@ -114,23 +115,6 @@ def test_encrypt_raises_if_no_password(self):
             with pytest.raises(ValueError):
                 temp.encrypt_passkey("RADIUS", "plaintext")
 
-    def test_is_key_encrypt_enabled(self):
```
anders-nexthop: why remove this testcase? I think it's a good thing to test.

src/sonic-py-common/tests/test_security_cipher.py
nmoray: yeah

src/sonic-py-common/sonic_py_common/security_cipher.py
```diff
@@ -12,89 +12,141 @@
...
-           except FileNotFoundError:
-                syslog.syslog(syslog.LOG_ERR, "__read_passwd_file: File {} no found".format(self._file_path))
-           except PermissionError:
-                syslog.syslog(syslog.LOG_ERR, "__read_passwd_file: Read permission denied: {}".format(self._file_path))
-
-        return passwd
-    
-
-    def encrypt_passkey(self, feature_type, secret: str, passwd: str) -> str:
+    def rotate_feature_passwd(self, feature_type, table_info, secret, new_password=None):
```
anders-nexthop: why are we passing the secret here? Isn't the secret per-instance, while only the password is shared between users of a feature? I don't think we want to combine `rotate()` with any actual instance secret updates.

src/sonic-py-common/sonic_py_common/security_cipher.py
```diff
@@ -12,89 +12,141 @@
...
+        cb_names = data[feature_type].get("callbacks", [])
+        callbacks = [self.callback_lookup[name] for name in cb_names if name in self.callback_lookup]
+
+        if not callbacks:
+            syslog.syslog(syslog.LOG_ERR, "rotate_feature_passwd: No callbacks registered for {}".format(feature_type))
+            return
+
+        syslog.syslog(syslog.LOG_INFO, "rotate_feature_passwd: Rotating password for feature {} and notifying callbacks...".format(feature_type))
+        for cb in callbacks:
+            cb(table_info, secret)
```
anders-nexthop: I wonder if there is a more reliable way of handling this callback functionality. We expect the encrypted secrets to be stored in ConfigDB, correct? Would we ever want the callback to do anything other than update entries in specified table paths? maybe instead of actual callbacks, we can store table paths that need to be updated and in this function we can either iterate through them and update each one or maybe build a patch and use GCU apply-patch to update them all cleanly in one operation.

src/sonic-py-common/sonic_py_common/security_cipher.py
```diff
@@ -12,89 +12,141 @@
...
+        On each call, read JSON data fresh from disk. Update password if provided,
+        and call all registered callbacks with the latest password.
+        """
+        data = self._load_registry()
+        if feature_type not in data:
+            syslog.syslog(syslog.LOG_ERR, "rotate_feature_passwd: No callbacks registered for {}".format(feature_type))
+            return
+
+        if new_password is not None:
+            data[feature_type]["password"] = new_password
```
anders-nexthop: I think we are missing something here. we expect the user callback to update the encrypted secrets stored in ConfigDB, but those secrets have been encrypted with the old password. we can't just encrypt them again with the new one, we have to first decrypt each secret using the old password, then re-encrypt it using the new one and store it back in ConfigDB. Otherwise we will lose the original secret and things won't work properly.

src/sonic-yang-models/yang-models/sonic-system-tacacs.yang
```diff
@@ -91,6 +91,12 @@ module sonic-system-tacacs {
                     description "Authentication type";
                 }
 
+                leaf key_encrypt {
```
anders-nexthop, line 100: Would this be better as a typedef, so that multiple features can use the same type?

src/sonic-py-common/sonic_py_common/security_cipher.py
nmoray: Yeah. While implementing the callback for TACACS, I was thinking that there will be a single callback only. But yes, you are right. I guess, maintaining the table_info in place of callbacks will be a better option so that each and every table can be updated by decrypting the existing passkey with old pass and again re-encrypt them with a new one.

src/sonic-py-common/sonic_py_common/security_cipher.py
nmoray: As per https://github.com/sonic-net/sonic-utilities/pull/3936 implementation, user is not allowed to rotate the password without specifying the passkey or secret. So we don't need to worry about decryption and re-encryption.

src/sonic-py-common/tests/test_security_cipher.py
nmoray: Added the testcase back.

src/sonic-py-common/sonic_py_common/security_cipher.py
nmoray: As per the discussion, now replaced the callbacks with just table info so that whenever any feature registers, the respective table_info will be attached to the Feature_id and same will be used while rotating the password. The rotate method will sequentially updates the registered tables with the new encrypted passkey.

src/sonic-yang-models/yang-models/sonic-system-tacacs.yang
nmoray: AFAIK, we can not typedef the leaf itself. But can typedef its parameters.
e.g.
```
typedef key-encrypt-template {
    type boolean;
    default false;
    description "Indicates if the passkey is encrypted.";
}

leaf key_encrypt {
    type key-encrypt-template;
}
```

anders-nexthop: The advantage to a template would be that we would have some measure of certainty that implementers would all be using the same format when adding the key_encrypt leaf. The disadvantage would be that every feature would have to use the same default, but I think given the implementation that would be okay.

This is subjective. I would have a slight preference for having a shared template as the base, but I'm not that familiar with SONiC's YANG best practices. Up to you if you think this is helpful to add or not.

src/sonic-py-common/sonic_py_common/security_cipher.py
```diff
@@ -146,39 +99,123 @@ def decrypt_passkey(self, feature_type,  b64_encoded: str) -> str:
...
+        If, after removal, there are no more table_info entries for that feature,
+        remove the respective password as well.
+        """
+        data = self._load_registry()
+        if feature_type in data:
+            if table_info in data[feature_type]["table_info"]:
+                data[feature_type]["table_info"].remove(table_info)
+                if not data[feature_type]["table_info"]:
+                    # No more table_info left; remove password as well
+                    data[feature_type]["password"] = None
```
anders-nexthop, line 160: should we instead just remove the entire feature entry? Otherwise we will have to have handling for having password=None in other places.

Or do we need to remove the password at all here? It's possible to add a feature password without having any table_info. Do we really want to force the user to add a new password if they remove all the table_info entries? Maybe we should instead decouple the presence/absence of table_info entries from the password handling entirely.

src/sonic-py-common/sonic_py_common/security_cipher.py
```diff
@@ -146,39 +99,123 @@ def decrypt_passkey(self, feature_type,  b64_encoded: str) -> str:
...
+            encrypted_passkey = db_entry.get("passkey")
+            if encrypted_passkey:
+                # Decrypt with old password
+                plain_passkey = self._decrypt_passkey(feature_type, encrypted_passkey, old_password)
+                # Re-encrypt with new password
+                new_encrypted_passkey = self._encrypt_passkey(feature_type, plain_passkey, new_password)
+                # Update DB
+                db_entry["passkey"] = new_encrypted_passkey
+                # Make sure key_encrypt should be set true
+                db_entry["key_encrypt"] = 'True'
```
anders-nexthop: I think we should check key_encrypt before encrypting, and skip the table if it is set to false. If a user has disabled key_encrypt, they may have changed the passkey as well so we run the risk of mangling the passkey. In theory, if we have table_info stored then key_encrypt should always be true for that table,  but there's nothing stopping a user from directly modifying ConfigDB. I think if we detect that key_encrypt has been disabled, we should print an error message and continue without modifying that particular table.

That brings up another question though. How does this code handle the encryption feature being turned off for a given passkey? I think the caller would at a minimum have to deregister the table and set key_encrypt=false. What should happen to the passkey they have stored? It will be invalid once it is no longer associated with the encryption key. Should we assume the caller has to deal with that? Or should we add a mechanism to handle that in this code? Maybe `deregister()` should decrypt the passkey associated with that path before removing the table_info element?

src/sonic-py-common/sonic_py_common/security_cipher.py
```diff
@@ -146,39 +99,123 @@ def decrypt_passkey(self, feature_type,  b64_encoded: str) -> str:
...
+
+        # Update stored password
+        data[feature_type]["password"] = new_password
+        self._save_registry(data)
+        syslog.syslog(syslog.LOG_INFO, f"rotate_feature_passwd: Password for {feature_type} updated.")
+
+    def encrypt_passkey(self, feature_type, secret: str) -> str:
         """
-        Removes only the password for the given feature_type while keeping the file structure intact.
+        Encrypts the plaintext and returns the result as a hex string.
```
anders-nexthop, line 220: nit: the description comment for decrypt_passkey includes information about the encryption algorithm, should we add that here as well?

src/sonic-py-common/sonic_py_common/security_cipher.py
nmoray: There is no harm in removing the feature entry itself if no single registered entry exist. But we need to delete the cipher_pass file as well if there are no registered features too. Just to avoid that I kept the footprints of the registered Features. 
If we decide to have only the password without table_info then rotate feature won't be possible. IMO, now the new design with Rotate feature has been settled down. :-) 

Why do you think the coupling of password and table_info will cause an issue? If so, what will it be?

src/sonic-py-common/sonic_py_common/security_cipher.py
nmoray: Please refer to https://github.com/sonic-net/sonic-utilities/pull/3936/files. I have handled this from the Feature side and added this explicitly to avoid any failure due to a corner case. From the feature side, I have made sure that this method has to be called if Encrypt is enabled. So I am explicitly adding key_encrypt flag to avoid a case where there is no key_encrypt flag for a registered TABLE.

src/sonic-py-common/sonic_py_common/security_cipher.py
nmoray: As it's a wrapper, I have not added the Algorithm related details in this. _encrypt_passkey() is the actual API where everything happens so I have added those details there.

src/sonic-py-common/sonic_py_common/security_cipher.py
anders-nexthop: To be clear, I think that we absolutely want the rotation mechanism, and I think the overall design is very close to correct at this point!

Thinking ahead to how this service might be used in the field, I can imagine having a startup script that configures the encryption password for a feature when the device is first provisioned, with some rotation schedule set after that initial setup. The feature itself may be enabled or disabled many times (or have the passkey added/removed many times) before that rotation schedule is up. I believe with the current design, we will remove the table_info for a feature when that feature is globally disabled, and each time we do so we will force the user to either re-enter the password or use a new one. Managing this may be challenging; if the user wants to use the same password as they had before (which they might want if they are using the same encryption password on several devices, so that they can use a common config with a single encrypted feature password for all devices), they now need to keep track of that password and re-apply it as needed.

My feeling is that we would rather have the encryption password be something that is not reset very frequently (only when you explicitly want to rotate), while the table_info can come and go as needed.

Here's how it could look (very similar to what you have already):

* register()/deregister(): adds/removes tables associated with a given feature. No change to feature password. table_info can be empty and password will remain set.
* add_feature_password()/rotate_feature_password(): sets/rotates feature password. On rotate, any passkeys from associated table_info(s) will be updated. If there are no associated tables, the feature password is still updated.
* del_feature_password(): if we need, we can add back the option to remove the feature password entirely. This would be like a rotation, where we would go and decrypt all the associated passkeys, but without the re-encryption step. This would disable the encryption service, but would preserve configured passkeys. The list of table_info would remain intact, so if the service is re-enabled at some point the configured passkeys will be updated accordingly. I don't know if we actually need or want this method; if the user has a passkey with key_encrypt set but the service is disabled, they will have an unencrypted passkey even though they have requested an encrypted one, which doesn't seem good. I include a description of how this would work mainly just to demonstrate the mechanics of how we could maintain the table_info and the feature passkey separately.

To me, this is slightly better as it makes the contract with the caller simpler. You can update the feature password, or you can update the list of tables associated with a feature, or you can encrypt/decrypt. Each operation is separate. We can let the caller decide if they want to rotate the key when the feature is enabled/disabled. I think that this encryption service may end up being pretty generally useful in sonic, so having fewer internal assumptions is beneficial.

Does this suggest make sense or is it still not clear what I'm getting at? Do you see any downsides of the suggested approach that I am not considering?

src/sonic-py-common/sonic_py_common/security_cipher.py
anders-nexthop: I don't think we want to push logic into the caller if we can avoid it, since that makes this service less general. At the moment, we are just thinking about AAA users, but there's no reason we couldn't use this mechanism to store any type of sensitive identifying information in ConfigDB. We'd be better off to solve corner cases in the API code if we can, so that we don't have to re-implement the same logic for every caller.

src/sonic-py-common/sonic_py_common/security_cipher.py
nmoray: As per the current design, the user has to enter the password only when he/ she wants to encrypt the passkey for the first time and second time when he / she wants to rotate or change the feature password.  If the user wants to use the same password, he / she can keep that password and keep encrypting the new passkey as many times as he / she wants. 

Please refer to the following unit test logs.
```
1. Encrypt passkey:
sonic# config tacacs passkey nikhil --encrypt
sonic# show run al | jq .TACPLUS
{
  "global": {
    "auth_type": "login",
    "key_encrypt": "True",
    "passkey": "U2FsdGVkX19YTUcG+0r3+d78A5E0zTXEukS3ZNrfyFk=",
    "src_intf": "Loopback0"
  }
}
sonic# cat /etc/pam.d/common-auth-sonic
#THIS IS AN AUTO-GENERATED FILE
#
# /etc/pam.d/common-auth- authentication settings common to all services
# This file is included from other service-specific PAM config files,
# and should contain a list of the authentication modules that define
# the central authentication scheme for use on the system
# (e.g., /etc/shadow, LDAP, Kerberos, etc.). The default is to use the
# traditional Unix authentication mechanisms.
#
# here are the per-package modules (the "Primary" block)

auth	[success=done new_authtok_reqd=done default=ignore auth_err=die]	pam_tacplus.so server=<> secret=nikhil login=login timeout=5   try_first_pass
auth	[success=done new_authtok_reqd=done default=ignore auth_err=die]	pam_tacplus.so server=<> secret=nikhil login=login timeout=5   try_first_pass
auth	[success=1 default=ignore]	pam_unix.so nullok try_first_pass

#
# here's the fallback if no module succeeds
auth    requisite                       pam_deny.so
# prime the stack with a positive return value if there isn't one already;
# this avoids us returning an error just because nothing sets a success code
# since the modules above will each just jump around
auth    required                        pam_permit.so
# and here are more per-package modules (the "Additional" block)
sonic# cat /etc/cipher_pass.json
{
  "TACPLUS": {
    "table_info": [
      "TACPLUS|global"
    ],
    "password": "TEST1"
  }

2. Encrypt passkey and update existing password
sonic# config tacacs passkey nikhil --encrypt --rotate
Password:
sonic# cat /etc/cipher_pass.json
{
  "TACPLUS": {
    "table_info": [
      "TACPLUS|global"
    ],
    "password": "TEST3"
  }
sonic# cat /etc/pam.d/common-auth-sonic
#THIS IS AN AUTO-GENERATED FILE
#
# /etc/pam.d/common-auth- authentication settings common to all services
# This file is included from other service-specific PAM config files,
# and should contain a list of the authentication modules that define
# the central authentication scheme for use on the system
# (e.g., /etc/shadow, LDAP, Kerberos, etc.). The default is to use the
# traditional Unix authentication mechanisms.
#
# here are the per-package modules (the "Primary" block)

auth	[success=done new_authtok_reqd=done default=ignore auth_err=die]	pam_tacplus.so server=<> secret=nikhil login=login timeout=5   try_first_pass
auth	[success=done new_authtok_reqd=done default=ignore auth_err=die]	pam_tacplus.so server=<> secret=nikhil login=login timeout=5   try_first_pass
auth	[success=1 default=ignore]	pam_unix.so nullok try_first_pass

#
# here's the fallback if no module succeeds
auth    requisite                       pam_deny.so
# prime the stack with a positive return value if there isn't one already;
# this avoids us returning an error just because nothing sets a success code
# since the modules above will each just jump around
auth    required                        pam_permit.so
# and here are more per-package modules (the "Additional" block)
sonic# show run al | jq .TACPLUS
{
  "global": {
    "auth_type": "login",
    "key_encrypt": "True",
    "passkey": "U2FsdGVkX18/8nFrAiCe01pOqgPoUUZFaTpmtawdi8I=",
    "src_intf": "Loopback0"
  }
}

3. Revert back to plaintext passkey
sonic# config tacacs passkey ravi
sonic# cat /etc/pam.d/common-auth-sonic
#THIS IS AN AUTO-GENERATED FILE
#
# /etc/pam.d/common-auth- authentication settings common to all services
# This file is included from other service-specific PAM config files,
# and should contain a list of the authentication modules that define
# the central authentication scheme for use on the system
# (e.g., /etc/shadow, LDAP, Kerberos, etc.). The default is to use the
# traditional Unix authentication mechanisms.
#
# here are the per-package modules (the "Primary" block)

auth	[success=done new_authtok_reqd=done default=ignore auth_err=die]	pam_tacplus.so server=<> secret=ravi login=login timeout=5   try_first_pass
auth	[success=done new_authtok_reqd=done default=ignore auth_err=die]	pam_tacplus.so server=<> secret=ravi login=login timeout=5   try_first_pass
auth	[success=1 default=ignore]	pam_unix.so nullok try_first_pass

#
# here's the fallback if no module succeeds
auth    requisite                       pam_deny.so
# prime the stack with a positive return value if there isn't one already;
# this avoids us returning an error just because nothing sets a success code
# since the modules above will each just jump around
auth    required                        pam_permit.so
# and here are more per-package modules (the "Additional" block)
sonic# show run al | jq .TACPLUS
{
  "global": {
    "auth_type": "login",
    "key_encrypt": "False",
    "passkey": "ravi",
    "src_intf": "Loopback0"
  }
}
sonic# cat /etc/cipher_pass.json
{
  "TACPLUS": {
    "table_info": [],
    "password": null
  }
```

anders-nexthop: Right, what I'm wondering about is that last case. When the user reverts back to using a plaintext password, they lose the configured encryption password. If they want to re-enable passkey encryption, they will have to enter the encryption password again. Consider the case where we have multiple tacacs servers set up (with no global passkey). These servers can come and go, and the corresponding config entry for them will come and go along with them. If at any time all servers are disabled, then all of the table_info entries will be removed and we will discard the encryption password. The next time a server is configured, the encryption password will have to be entered again.

This seems like a minor hassle, but I think there are plausible use cases where this would be problematic. If you are using automated deployment, you might have an initial startup script that sets the encryption password, and have no other reference to it on the device. If later actions cause the configured password to be discarded, the encryption password will have to be added again. In a network where devices are mostly handled through automation, this can be an issue (the password is nowhere on the device, so a remote connection is needed to set things back up).

Here's an example:
* I have an init_cfg.json which does not define any tacacs servers.
* During initial device bringup in the network, an automated script configures a default tacacs server and sets the encryption password.
* During regular operation, an admin adds a new server and removes the default one
   - Here, the admin MUST add the new server before removing the old one, or the admin must re-enter the encryption password
* The device reboots, losing the configured servers
   - The next time a server is configured, the encryption password must be added again. Either a script needs to run on startup (as was done on device bringup) to configure a default server, or an administrator must re-enter the encryption password

In the above scenarios, the encryption password ends up getting set multiple times, both by scripts and by any number of administrator users. This feels like a potential security concern to me. Password proliferation should be avoided; the more times the password is set, and the more admins and scripts that need to know it, the less secure it will be.

src/sonic-py-common/sonic_py_common/security_cipher.py
```diff
@@ -164,24 +164,23 @@ def rotate_feature_passwd(self, feature_type, new_password):
             table, entry = table_info.split("|")
             db_entry = self._config_db.get_entry(table, entry)
             encrypted_passkey = db_entry.get("passkey")
-            if encrypted_passkey:
+            #Rotate only if valid passkey is present and 'key_encyrpt' flag is True
+            if encrypted_passkey and db_entry.get("key_encrypt") == 'True':
```
anders-nexthop: I always struggle with the yang boolean type. Isn't the key_encrypt string supposed to 'true'?

anders-nexthop (approved):
I am approving this change, as I think this is a workable implementation and would like to see it move forward. However, please consider my point about separating encryption password handling from table registration.

@qiluo-msft can you take another look, or assign another reviewer to get more eyes on this latest version?

@bhouse-nexthop your thoughts would be welcome here given your security background.

src/sonic-py-common/sonic_py_common/security_cipher.py
bhouse-nexthop: should definitely make it a case-insensitive comparison.  yang wants lowercase, python wants uppercase first char.  But when comparing as strings always compare lower.

src/sonic-py-common/sonic_py_common/security_cipher.py
nmoray: <When the user reverts back to using a plaintext password, they lose the configured encryption password>. It is expected, isn't it! :-) If the user reverts back to plaintext passkey. There should not be any password for the respective feature. Later if the user wants to encrypt that passkey, he / she must provide the passkey along with a new password. 
If we keep the old password and used that for the new encryption, then it won't be right as we will be using stale data in that case and the user may lose the track of the old password too.

In case of a reboot, the configs gets removed (unsaved ones) but the password saved in cipher_pass.json file won't be removed unless and until the user tries to reconfigure the tacplus server(s) with explicit passkey encryption as false.

src/sonic-py-common/sonic_py_common/security_cipher.py
nmoray: Agreed!

nmoray:
Thanks @anders-nexthop for the diligent review.
@qiluo-msft @bhouse-nexthop I have addressed the comments. Please approve in case there are no more comments.

nmoray:
@qiluo-msft please approve in case there are no more comments so that other dependent PRs can also be merged.
https://github.com/sonic-net/sonic-utilities/pull/3936
https://github.com/sonic-net/sonic-host-services/pull/274

src/sonic-py-common/sonic_py_common/security_cipher.py
anders-nexthop: The reboot case you mentioned points out the issue I'm describing perfectly. I can configure a passkey on a running system and pass in an encryption key. If I don't save the running config and reboot the system, it will start up again without the passkey, but the encryption key will still be saved. We are then in the case that we are trying to avoid, where there is a stale encryption key being used to encrypt passkeys. So no matter what, the encryption key handling is somewhat disconnected from the passkey handling, and we rely on the caller knowing about this case and handling it correctly. I think we would be better served to make that difference explicit by providing a `delete_feature_password()` that can be called explicitly by the user to disable the encryption for that feature. This way, the contract provided to the caller is very clear; they must explicitly `set` and `delete` the feature password manually when they want to enable/disable the service for a feature, and can use `encrypt` and `decrypt` only on an enabled feature.

This seems to me to be a cleaner approach, with less corner-case ambiguity. However I am happy to advocate for this change as a later enhancement, should the benefit become more clear as we implement callers. Let's keep pushing the feature forward and enhance it as needed, rather than try to solve for every eventuality here.

nmoray: Yeah, but reboot scenario will be same even with the new delete_feature_password() api as it won't be called in case of a reboot without a "config save".

anders-nexthop: Right, but if it's understood that the feature password is not managed through ConfigDB then that seems more predictable from a user standpoint. I set the encryption passkey for a feature on the device, and I have not unset it, therefore I am not surprised when it still exists.

Since we added the migration logic, which preserves configured passwords even across upgrades, it seems like we generally want to keep the encryption passwords around regardless of current configuration. Otherwise why bother with the migration at all?

nmoray:
@qiluo-msft will you please help in merging this PR in case there are no more comments?

anders-nexthop:
@qiluo-msft and @lguohan can we get this PR merged if there are no more comments?

prabhataravind:
hi @qiluo-msft is this good to be merged?

arlakshm:
@qiluo-msft, please help approve if no other comments.

anders-nexthop:
@qiluo-msft @lguohan Besides wanting this for TACACS, there are two outstanding feature requests that are dependent on this basic infrastructure getting merged (https://github.com/sonic-net/sonic-buildimage/issues/22276 and https://github.com/sonic-net/sonic-buildimage/issues/22277), can we please get this reviewed by an authorized reviewer and merged up?

lguohan:
do we have sonic-mgmt test to test this?

lguohan:
@anders-nexthop , i will ask qiluo to take a look this PR. Thanks for your patience.

files/image_config/config-setup/01-post-security-cipher
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+set -e
+
+# Restore cipher_pass file from persistent storage
+if [ -f /host/security_cipher/cipher_pass ]; then
+    cp /host/security_cipher/cipher_pass.json /etc/cipher_pass.json
+    chmod 640 /etc/cipher_pass.json
```
qiluo-msft: >640 [](http://example.com/codeflow?start=10&length=3)

Is it better to use `600` if the only reader is root? #Closed

src/sonic-py-common/sonic_py_common/security_cipher.py
```diff
@@ -28,69 +31,33 @@ def __new__(cls):
...
+    def _load_registry(self):
+        """
+        Read cipher_pass.json file
+        """
+        if not os.path.exists(CIPHER_PASS_FILE):
+            return {}
+        try:
+            with open(CIPHER_PASS_FILE, 'r') as f:
+                return json.load(f)
+        except Exception as e:
```
qiluo-msft: >Exception [](http://example.com/codeflow?start=15&length=9)

Could you use more specific exception type? #Closed

nmoray:
> do we have sonic-mgmt test to test this?

@lguohan No, but once this PR is merged, I will create a new PR to cover the sonic-mgmt testcases. That is part of sonic-mgmt submodule so anyway a new PR is required. So IMO, it would be better not to block this PR solely due to the sonic-mgmt test cases.

src/sonic-py-common/sonic_py_common/security_cipher.py
nmoray: @qiluo-msft Like? Any specific exception that you are looking for?

Is "except json.JSONDecodeError" okay? or "except (json.JSONDecodeError, OSError)"? 

IMO, "Exception as e" will catch all the runtime exceptions and also print the respective error in the form of 'e'. What do you think?

nmoray:
> As commented

@qiluo-msft I replied to your comment. Please check and share your views on the same.

files/image_config/config-setup/01-post-security-cipher
nmoray: sure.

src/sonic-py-common/sonic_py_common/security_cipher.py
nmoray: @qiluo-msft comments addressed. Please check and approve if no more comments.

nmoray:
@qiluo-msft comments addressed. Please check and approve if no more comments.

files/image_config/config-setup/01-pre-security-cipher
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+set -e
+
+# Ensure old_config directory exists
+mkdir -p /host/security_cipher
+
+# Copy cipher_pass file to persistent storage
+if [ -f /etc/cipher_pass ]; then
```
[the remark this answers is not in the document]
nmoray: done

files/image_config/config-setup/01-post-security-cipher
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+set -e
+
+# Restore cipher_pass file from persistent storage
+if [ -f /host/security_cipher/cipher_pass ]; then
```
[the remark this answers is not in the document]
nmoray: done

src/sonic-py-common/sonic_py_common/security_cipher.py
```diff
@@ -146,39 +134,122 @@ def decrypt_passkey(self, feature_type,  b64_encoded: str) -> str:
...
+                self._save_registry(data)
+                syslog.syslog(syslog.LOG_INFO, "deregister: table_info {} removed from feature {}".format(table_info, feature_type))
+            else:
+                syslog.syslog(syslog.LOG_ERR, "deregister: table_info {} not found for feature {}".format(table_info, feature_type))
+        else:
+            syslog.syslog(syslog.LOG_ERR, "deregister: No table_info registered for {}".format(feature_type))
+
+    def set_feature_password(self, feature_type, password):
+        """
+        Set a new password for a feature type.
```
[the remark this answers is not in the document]
nmoray: done

nmoray:
@qiluo-msft I have addressed all of your comments. Please check and approve if there are no more comments.

nmoray:
@qiluo-msft Please approve the PR. All of your comments have been addressed.

PR #22326, opened 2025-04-14
Install pygobject from Debian instead of pip

saiarcot895:
#### Why I did it

This package is already available in Debian and doesn't require any compilation on SONiC's end for using this.

This also removes compilation errors related to not having `girepository` installed when upgrading to a newer version of this package from pip.

##### Work item tracking
- Microsoft ADO **(number only)**:

#### Which release branch to backport (provide reason below if selected)

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

#### Tested branch (Please provide the tested image version)

- [ ] 
- [ ]

PR #22722, opened 2025-05-27
[build] Install python3-gi in docker-sonic-vs as pygobject's dependency

liushilongbuaa:
#### Why I did it
1. We already use apt to install python3-gi in SONiC image according to #22326 
But docker-sonic-vs.gz change is missed in that PR.
##### Work item tracking
- Microsoft ADO **(number only)**:  32869072

#### How I did it
1. Install python3-gi in docker-sonic-vs.gz image.

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

- [ ] 
- [ ]

### DIFF
--- a/platform/vs/docker-sonic-vs/Dockerfile.j2
+++ b/platform/vs/docker-sonic-vs/Dockerfile.j2
@@ -36,6 +36,7 @@ RUN apt-get install -y net-tools \
                        # (these can be uninstalled after installation)
                        build-essential \
                        python3-dev \
+                       python3-gi \
                        libssl-dev \
                        swig \
                        # For using Python m2crypto package


yijingyan2:
Why needed for 202505 Branch:
When upgrade package versions for 202505, the build would fail on sonic-host-services wheel installation in sonic-vs image. This is because sonic-host-services requires package pygobject. The latest version 3.56.2 requires girepository-2.0. but the girepository-2.0 is only available in Trixie. 
With this PR change, the pygobject will be installed by Debian instead of pip, so the version is 3.42.2, and only requires girepository-1.0, which is available in bookworm

PR #22769, opened 2025-05-29
[Mellanox] Update SN4700 media_settings to have another needed module type

noaOrMlnx:
#### Why I did it
To have QSFP+C module type in Mellanox media_settings.json

#### How I did it
- Added needed module with its supported applicationsץ
- Update QSFP module to contain up to 4 lanes, as this is Quad SFP and supports up to 4 lanes.
- Update all module types with '+' sign, with regex escape, so they would be detected in media_settings_parser.
#### How to verify it
- deploy image on SN4700 Mellanox switch. 
- enable cmis host mgmt.
- make sure all Serdes SI parameters are valid. 

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

- [ ] 
- [ ]


tshalvi (approved):

prgeor (approved):

device/mellanox/x86_64-mlnx_msn4700-r0/Mellanox-SN4700-O8C48/media_settings.json
```diff
@@ -1013,120 +1013,282 @@
...
-                      "lane4": "0x0000008c",
-                      "lane5": "0x0000008c",
-                      "lane6": "0x0000008c",
-                      "lane7": "0x0000008c"
+                      "lane3": "0x0000008c"
                   }
-              }
-          }
+               },
+              "speed:200GAUI-4": {
```
prgeor, line 1085: @noaOrMlnx What about a case where we have to enable 2x200G AOC?

noaOrMlnx: @prgeor as elaborated in mail, this 2x200G AOC is a splitter cable - QSFP-DD 400Gx4 lanes. 2 QSFP+ 200Gx4 lanes.
each QSFP+ side has 4 lanes and supports 200G/100G. 

for QSFP-DD we indeed need 8 lanes and this what media_settings contains. 
for QSFP+ sides we support only 4 lanes so we must have 4 lanes only in media seetings.

prgeor: @noaOrMlnx can you share the output of following command ? 

```
dump state port EthernetXX -t
```

device/mellanox/x86_64-mlnx_msn4700-r0/Mellanox-SN4700-O8C48/media_settings.json
```diff
@@ -1013,120 +1013,282 @@
...
+                      "lane0": "0x00000000",
+                      "lane1": "0x00000000",
+                      "lane2": "0x00000000",
+                      "lane3": "0x00000000"
+                  },
+                  "pre1": {
+                      "lane0": "0xfffffffe",
+                      "lane1": "0xfffffffe",
+                      "lane2": "0xfffffffe",
+                      "lane3": "0xfffffffe"
```
prgeor, line 1096: @noaOrMlnx I see that you have removed the lanes 4 to 7. What is the harm if we keep these? It will be useful for cases where we use 2x200G i.e there is a second subport at 200G.

noaOrMlnx: @prgeor 
Hi, 
QSFP means Quad SFP. meaning, there cannot be more than 4 lanes for this module. 
The harm we will have if we keep the 8 lanes in media_settings, is passing 8 lanes to SAI while trying to configure Serdes SI for the only number of lanes this modules supports - 4, and it will casue orchagent to crash - SAI recognizes 4 lanes in the port, but 8 parameters trying to be configured. 
reagrding 2x200G cable - 2x200G means 2 **different** modules, each one 200Gx4 lanes. since they are different, each one will configure serdes SI differently and regardless of the other QSFP+ side. 
4 lanes is what we need for 2x200G cable, as this is the only option this cable supports.

prgeor: @noaOrMlnx thanks. since the type is QSFP+, I agree, it should work only for the 4 lanes. 
`QSFP\\+-active_cable_media_interfac`

PR #22771, opened 2025-05-29
Improve auditd rate limit to 100/second

liuh-80:
Improve auditd rate limit to 100/second

#### Why I did it
Improve auditd event rate limit.

##### Work item tracking
- Microsoft ADO **(number only)**: 33069921

#### How I did it
Improve rate limit to 100

#### How to verify it
Manually verify
Pass all unit test cases.

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

- [ ] 
- [ ] 

Improve auditd rate limit to 100/second

### DIFF
--- a/dockers/docker-auditd/auditd_config_files/audit.rules
+++ b/dockers/docker-auditd/auditd_config_files/audit.rules
@@ -12,4 +12,4 @@
 -f 1
 
 ## Set auditd rate limit
--r 1000
+-r 100


yejianquan:
Just a thinking, should we disable auditd(or reduce the limitation) in our testing(maybe pretest) since it generates huge logs, and slows down the loganalyzes signaficantly, especially is the test modules if running for long times, like in T2, the test could run for 7 hours.
@liuh-80  @qiluo-msft  @StormLiangMS

liuh-80:
> Just a thinking, should we disable auditd(or reduce the limitation) in our testing(maybe pretest) since it generates huge logs, and slows down the loganalyzes signaficantly, especially is the test modules if running for long times, like in T2, the test could run for 7 hours. @liuh-80 @qiluo-msft @StormLiangMS

These auditd rules will not impact nightly test, because they are not including in image. these files will apply to image by auditd docker container rollout, which is done separately.

PR #22772, opened 2025-05-29
[7215] Disable /tmp in RAM on Nokia-7215 ARMHF

lizhijianrd:
##### Work item tracking

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [x] 202505

- [ ] 
- [ ]

### DIFF
--- a/files/initramfs-tools/union-mount.j2
+++ b/files/initramfs-tools/union-mount.j2
@@ -218,8 +218,10 @@ mkdir -p ${rootmnt}/boot
 mkdir -p ${rootmnt}/host/$image_dir/boot
 mount --bind ${rootmnt}/host/$image_dir/boot ${rootmnt}/boot
 
+{% if ARCH | default("x86_64") != "armhf" %}
 ## Mount the /tmp directory as tmpfs
 mount -t tmpfs -o rw,nosuid,nodev,size=25% tmpfs ${rootmnt}/tmp
+{% endif %}
 
 ## Mount loop device or tmpfs for /var/log
 if $logs_inram; then
--- a/slave.mk
+++ b/slave.mk
@@ -56,11 +56,14 @@ ifeq ($(PLATFORM_ARCH),)
 	override PLATFORM_ARCH = $(CONFIGURED_ARCH)
 endif
 DOCKER_BASE_ARCH := $(CONFIGURED_ARCH)
+ONIE_IMAGE_CONF_PATH := onie-image.conf
 ifeq ($(CONFIGURED_ARCH),armhf)
 	override DOCKER_BASE_ARCH = arm32v7
+	override ONIE_IMAGE_CONF_PATH = onie-image-armhf.conf
 else
 ifeq ($(CONFIGURED_ARCH),arm64)
 	override DOCKER_BASE_ARCH = arm64v8
+	override ONIE_IMAGE_CONF_PATH = onie-image-arm64.conf
 endif
 endif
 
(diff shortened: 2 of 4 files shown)


files/initramfs-tools/union-mount.j2
```diff
@@ -218,8 +218,10 @@ mkdir -p ${rootmnt}/boot
 mkdir -p ${rootmnt}/host/$image_dir/boot
 mount --bind ${rootmnt}/host/$image_dir/boot ${rootmnt}/boot
 
+{% if ARCH | default("x86_64") != "armhf" %}
 ## Mount the /tmp directory as tmpfs
 mount -t tmpfs -o rw,nosuid,nodev,size=25% tmpfs ${rootmnt}/tmp
+{% endif %}
```
saiarcot895, line 224: Instead of all these changes, `DPKG_ARCH` can be used (please test with this, I'm not 100% certain `armhf` is the correct string.

```suggestion
if [ $DPKG_ARCH != "armhf" ]; then
    ## Mount the /tmp directory as tmpfs
     mount -t tmpfs -o rw,nosuid,nodev,size=25% tmpfs ${rootmnt}/tmp
 fi
```

Better than this, though, is if we can do this based on available memory, that would be the best, because that is the underlying reason that this change is being made.

lizhijianrd: Hi @saiarcot895, the j2 cli take `onie-image.conf` as the data source. Any variable not defined in `onie-image.conf` can't be use here. That's why I added a new variable in `onie-image.conf`.

slave.mk
```diff
@@ -1350,8 +1354,8 @@ $(addprefix $(TARGET_PATH)/, $(SONIC_RFS_TARGETS)) : $(TARGET_PATH)/% : \
 		export RFS_SPLIT_FIRST_STAGE=y
 		export RFS_SPLIT_LAST_STAGE=n
 
-		j2 -f env files/initramfs-tools/union-mount.j2 onie-image.conf > files/initramfs-tools/union-mount
```
lizhijianrd, line 1353: @saiarcot895 Here the j2 command use `onie-image.conf` as data source, so the env variables can't be use in j2 template now...

saiarcot895 (approved):

files/initramfs-tools/union-mount.j2
```diff
@@ -218,8 +218,10 @@ mkdir -p ${rootmnt}/boot
 mkdir -p ${rootmnt}/host/$image_dir/boot
 mount --bind ${rootmnt}/host/$image_dir/boot ${rootmnt}/boot
 
-## Mount the /tmp directory as tmpfs
-mount -t tmpfs -o rw,nosuid,nodev,size=25% tmpfs ${rootmnt}/tmp
+if [ $DPKG_ARCH != "armhf" ]; then
```
vvolam, line 221: @lizhijianrd Is the change only effective for Nokia 7215 platform? If not, what are the platforms that are disabling this change?

lizhijianrd: Yes, this change only effect on Nokia-7215.
Nokia-7215 is the only armhf platform.

Blueve:
@StormLiangMS agree.
We will discover more solid solution and let platform owner to decide whether enable this feature and how much space reserved for /tmp.

@lizhijianrd please open an issue to track the long term improvement.

PR #22806, opened 2025-05-31
[build] fix dockers build context

k-v1:
#### Why I did it

The Docker client sends the entire “build context” to the Docker daemon. That build context (by default) is the entire directory the Dockerfile is in. Since sonic-buildimage supports parallel builds, there can be situations where a deb package and a docker image are being built at the same time. In such a case, Docker sometimes cannot pack the build context using tar, because the deb package being built is actively writing to its log file, which is also located in the debs directory.

You can see messsages like this in the build log of the docker image:

```
time="2025-05-28T14:13:26Z" level=error msg="Can't add file /sonic/platform/broadcom/docker-syncd-brcm/debs/sonic-mgmt-framework_1.0-01_amd64.deb.log to tar: archive/tar: write too long"
```

##### Work item tracking

#### How I did it

There are two solutions to this problem:
1. Switch to using BuildKit. In this case, external files that are not referenced in the Dockerfile are not sent to the Docker daemon.
2. Use a .dockerignore that includes only the dependencies (debs, wheels and files) needed for this docker image.

In this PR, I propose the second solution. It can also be used after switching to BuildKit, since it allows detecting incorrectly specified dependencies for docker images.

#### How to verify it

Verify that the builds completed successfully.

Refs:
https://docs.docker.com/build/concepts/context/
https://stackoverflow.com/questions/26600769/build-context-for-docker-image-very-large


k-v1:
After additional testing, I discovered that this bug (when docker tries to add a log file of some deb package in /debs to the build context) can still be reproduced even with a .dockerignore. Switching to BuildKit might be the best solution then.

PR #22808, opened 2025-06-01
[chassis] Cisco-8000: Use syseeprom for mac only as last option

anamehra:
syseepromg access is slower than getting the mac from eth0 interface and profile.ini file when available. On Chassis supervisor as many swss services spawns around same time and each try to access DEVICE info via sonic-cfggen, these make parallel calls to syseepromg to fetch mac. This sometimes lows the response during high cpu utilization. 
   For multi-asic namespace locations, mac is available from profile.ini and for others, eth0 should provide the correct mac.
Limiting the syseeprom access only when the mac is not retrieved from these options.

#### Why I did it
On supervisor, sometimes syncd process start get delayed during boot when parallel swss processes try to access syseeprom to get mac which doing DEVICE_METADATA.localhost.platform call via sonic-cfggen. This in some rare instances causes swss failure when syncd is not ready within 1 min and responds back to INIT_VIEW.

##### Work item tracking

#### How I did it
   For multi-asic namespace locations, mac is available from profile.ini and for others, eth0 should provide the correct mac.
Limiting the syseeprom access only when the mac is not retrieved from these options.

#### How to verify it
```
root@yy39-lc5:/home/cisco# time sonic-cfggen -H -v DEVICE_METADATA.localhost.mac
8c:84:42:90:97:20

real    0m0.532s
user    0m0.360s
sys     0m0.064s
root@yy39-lc5:/home/cisco# ifconfig  eth0
eth0: flags=4163<UP,BROADCAST,RUNNING,MULTICAST>  mtu 1500
        inet 1.75.46.11  netmask 255.255.0.0  broadcast 1.75.255.255
        inet6 fc0b::11  prefixlen 64  scopeid 0x0<global>
        inet6 fe80::8e84:42ff:fe90:9720  prefixlen 64  scopeid 0x20<link>
        ether 8c:84:42:90:97:20  txqueuelen 1000  (Ethernet)
        RX packets 7501239  bytes 889985239 (848.7 MiB)
        RX errors 0  dropped 0  overruns 0  frame 0
        TX packets 8052387  bytes 646097128 (616.1 MiB)
        TX errors 0  dropped 0 overruns 0  carrier 0  collisions 0

```

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [x] 202505

- [ ] 
- [ ]

### DIFF
--- a/src/sonic-py-common/sonic_py_common/device_info.py
+++ b/src/sonic-py-common/sonic_py_common/device_info.py
@@ -857,9 +857,17 @@ def get_system_mac(namespace=None, hostname=None):
             profile_cmd = ["false"]
             (mac, err) = run_command(profile_cmd)
         hw_mac_entry_outputs.append((mac, err))
-        (mac, err) = run_command(syseeprom_cmd)
-        hw_mac_entry_outputs.append((mac, err))
         (mac, err) = run_command_pipe(iplink_cmd0, iplink_cmd1, iplink_cmd2)
+        hw_mac_entry_outputs.append((mac, err))
+        for (mac, err) in hw_mac_entry_outputs:
+            if err:
+                continue
+            mac = mac.strip()
+            if _valid_mac_address(mac):
+                return mac
+        # If mac not found, fetch from syseeprom
+        hw_mac_entry_outputs = []
+        (mac, err) = run_command(syseeprom_cmd)
         hw_mac_entry_outputs.append((mac, err))
     elif (version_info['asic_type'] == 'pensando'):
         iplink_cmd0 = ["ip", 'link', 'show', 'eth0-midplane']


anamehra:
Hi @abdosi @yejianquan , for you viz.

anamehra:
Hi @abdosi @yejianquan , please review. Thanks

src/sonic-py-common/sonic_py_common/device_info.py
```diff
@@ -857,10 +857,18 @@ def get_system_mac(namespace=None, hostname=None):
...
-        hw_mac_entry_outputs.append((mac, err))
         (mac, err) = run_command_pipe(iplink_cmd0, iplink_cmd1, iplink_cmd2)
         hw_mac_entry_outputs.append((mac, err))
+        for (mac, err) in hw_mac_entry_outputs:
+            if err:
+                continue
+            mac = mac.strip()
+            if _valid_mac_address(mac):
+                return mac
+        # If mac not found, fetch from syseeprom
```
yejianquan, line 870: Hi @anamehra , Look like the code logic between 862 and 867 is the same with 883 and 888,
but the return logic in 867 doesn't looks very clean because most platforms put mac addr into hw_mac_entry_outputs, this could cause we lost post-actions(like in line 890, it doesn't look to be an issue for now, but in the future it may do more transofrmation),
Let's change to put the valid mac into hw_mac_entry_outputs,
and on 869,870 ,assert whether hw_mac_entry_outputs is not empty, only when it's empty, we fetch mac from syseeprom

anamehra: Hi @yejianquan , we need to do validity check here and not just empty check. If valid not found, then only fetch syseeprom.
I can remove return and let the function return at the end but will need validity check to decide if syseeprom need to be used. Is that fine?

yejianquan: Yes, that's what I mean

yejianquan (approved):
LGTM

PR #22888, opened 2025-06-08
YANG: PREFIX list implementation does not take sequence_id

bhouse-nexthop:
#### Why I did it

The sequence id is used as a key in the yang model and is therefor mandatory, however the implementation does not expect a sequence id at all and just uses the order in which the prefixes appear.

If passing value as YANG requires, the jinja2 renderer and frrcfgd both fail.

This affects unified mode with frr_mgmt_framework_config.

#### How I did it

This updates the YANG model to match the implementation, including test cases.

#### How to verify it

Configure prefix lists as per new yang model and see frr configuration is properly generated.

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [x] 202411
- [x] 202505

master as of 20250608

YANG: PREFIX list implementation does not take sequence_id

N/A


src/sonic-yang-models/yang-models/sonic-routing-policy-sets.yang
```diff
@@ -73,16 +73,6 @@ module sonic-routing-policy-sets {
                     description "Prefix name";
                 }
 
-                leaf sequence_number {
```
venkatmahalingam, line 76: FRR internally assumes the seq number when it's not used in the CLI configuration and running config shows the internally generated seq number, however it is good to expose the seq number to the user to ensure the order of execution. we may have to enhance the frrcfgd to call vtysh commands with seq number always,
ip prefix-list pref1 **seq <number>** permit 35.0.0.0/8

bradh352: ok, so we need to keep compatibility with the implementation itself, which means the sequence_number can't be part of the key otherwise we break users.

We can make it an optional field, and just emit it *if* it is provided.  

Would that be an acceptable change?  

I guess I'd also need to update the HLD to reference this was added as optional.

venkatmahalingam: Since the sequence number is generated internally by FRR when not provided, users are unaware of it unless they dump the prefix list commands from vtysh. We need to ensure that the sequence number is not configured in the prefix table or we should ensure non-conflicitng sequence number is used, as conflicting sequence numbers in the prefix list will lead to errors in vtysh.

I believe we just need to update frrcfgd to honor the sequence number. The front-end YANG was defined correctly, so there should not be any behavioral changes as long as users follow the YANG definitions correctly.

Do you see any issue?

bhouse-nexthop: @venkatmahalingam That is an invalid assumption.  MOST people don't use YANG validation at all since it is not enabled by default in SONiC, except for 'config replace'. 

I don't believe its ever valid to break users.  If we implement it the way you suggest, anyone currently using this feature (albeit without yang validation) will break.  This is also in violation of the HLD which is likely what someone would have used to implement said feature.

My suggestion is a happy middleground, where someone who wants to specify the sequence rather than depend on the auto-generated sequence *can*.  But doesn't mandate it thus any existing users of this feature will continue to work.

venkatmahalingam: I understand your concern, looks like YANG model was contributed without updating the HLD and frrcfgd for sequence number change.
You could add a following additional list inside a container PREFIX to satisfy the current HLD and frrcfgd code, we'll update the HLD and frrcfgd to have seq number also as an option in the future.

container PREFIX {
..............
list PREFIX_LIST { **-> Existing list**
                key "name sequence_number ip_prefix masklength_range";
                ................
                }
                
list PREFIX_NO_SEQ_LIST {    **-> New list**
                key "name ip_prefix masklength_range";
                ..............
                
                ....

bhouse-nexthop: thanks for your suggestion, however that still appears to fail to address the need to not break existing users of this feature.

I will move forward with implementing the sequence_number as an optional field instead and update this PR with that.  We can more forward with review/caveats after that point.

venkatmahalingam: I would suggest not to move forward with seq_number as a field, the below list (PREFIX_NO_SEQ_LIST) will align to current HLD and frrcfgd code, that's all needed now to not break existing users of this feature.
Ignore "list PREFIX_LIST" for now until we introduce the sequence number to the PREFIX container.

list PREFIX_NO_SEQ_LIST { -> New list
key "name ip_prefix masklength_range";

above new list will be aligned to the following schema (without seq-number),

"PREFIX": {
            "prefix1|10.0.0.0/8|8..16": {
            }
        },

bhouse-nexthop: Ok, I guess I see where there are heuristics in YANG processing to determine which internal list in the yang model things are mapped from CONFIG_DB to the YANG model list based on the number of keys in the list.

That said, I'm still failing to see why it should be done in that manner.  There is no benefit in having the sequence number contained within the keys as far as I can tell, especially if it means its optional since there are 2 forms of lists.  Why not just make the sequence_number an optional member?

venkatmahalingam: Here is the behavior of FRR VTYSH,

leaf1(config)# **ip prefix-list pref1 permit 37.0.0.0/8**
leaf1(config)# do show ip prefix-list pref1
ZEBRA: ip prefix-list pref1: 1 entries
   seq 5 permit 37.0.0.0/8
OSPF: ip prefix-list pref1: 1 entries
   seq 5 permit 37.0.0.0/8
BGP: ip prefix-list pref1: 1 entries
   seq 5 permit 37.0.0.0/8
PIM: ip prefix-list pref1: 1 entries
   seq 5 permit 37.0.0.0/8
leaf1(config)#
leaf1(config)#
leaf1(config)# **ip prefix-list pref1 seq 6 permit 37.0.0.0/8**
% Configuration failed.

Error type: validation
Error description: duplicated prefix list value: 37.0.0.0/8
leaf1(config)# ip prefix-list pref1 permit 37.0.0.0/8
leaf1(config)# do show ip prefix-list pref1
ZEBRA: ip prefix-list pref1: 1 entries
   seq 5 permit 37.0.0.0/8
OSPF: ip prefix-list pref1: 1 entries
   seq 5 permit 37.0.0.0/8
BGP: ip prefix-list pref1: 1 entries
   seq 5 permit 37.0.0.0/8
PIM: ip prefix-list pref1: 1 entries
   seq 5 permit 37.0.0.0/8
leaf1(config)# ip prefix-list pref1 seq 6 permit 37.0.0.0/8
% Configuration failed.

Error type: validation
Error description: duplicated prefix list value: 37.0.0.0/8
leaf1(config)# ip prefix-list pref1 seq 7 permit 37.0.0.0/8
% Configuration failed.

Error type: validation
Error description: duplicated prefix list value: 37.0.0.0/8
leaf1(config)# ip prefix-list pref1 seq 5 permit 38.0.0.0/8
leaf1(config)# do show ip prefix-list pref1
ZEBRA: ip prefix-list pref1: 1 entries
   seq 5 permit 38.0.0.0/8
OSPF: ip prefix-list pref1: 1 entries
   seq 5 permit 38.0.0.0/8
BGP: ip prefix-list pref1: 1 entries
   seq 5 permit 38.0.0.0/8
PIM: ip prefix-list pref1: 1 entries
   seq 5 permit 38.0.0.0/8
leaf1(config)# ip prefix-list pref1 seq 7 permit 38.0.0.0/8
% Configuration failed.

Error type: validation
Error description: duplicated prefix list value: 38.0.0.0/8
leaf1(config)# ip prefix-list pref1 seq 7 permit 39.0.0.0/8
leaf1(config)# do show ip prefix-list pref1
ZEBRA: ip prefix-list pref1: 2 entries
   seq 5 permit 38.0.0.0/8
   seq 7 permit 39.0.0.0/8
OSPF: ip prefix-list pref1: 2 entries
   seq 5 permit 38.0.0.0/8
   seq 7 permit 39.0.0.0/8
BGP: ip prefix-list pref1: 2 entries
   seq 5 permit 38.0.0.0/8
   seq 7 permit 39.0.0.0/8
PIM: ip prefix-list pref1: 2 entries
   seq 5 permit 38.0.0.0/8
   seq 7 permit 39.0.0.0/8
leaf1(config)#

bradh352: correct, FRR does validate duplicate prefixes.  I don't believe the yang model is checking that at this point.  I think it also prevents the same sequence from being specified (or just acts like a replacement if it is).  That check too would need to be added to the YANG model.

Those, however, are specific to the prefix list name.  Neither yang model currently prevents those issues.  

I should be able to add MUST clauses to check for that on the sequence_number, and its already enforced for the prefix itself in my updated model key since keys can't be duplicated.

venkatmahalingam: Here is the FRR YANG model for prefix list. Take a look at this (sequence number is being as the key), doesnt look like we're converging on the solution, we can push the current changes as is and provide sequence number support to prefix list as an enhancement later.
(https://github.com/FRRouting/frr/blob/master/yang/frr-filter.yang#L310)

bhouse-nexthop: Please look at the current commits I've done to see if its heading in the right path.  I'm going to close this particular discussion since its gotten quite long and isn't really relevant to the current implementation.

`sequence_number` is there again, its just not a Key.  But I did add it as an optional field with a `must` clause to enforce uniqueness within the same prefix set.

The only thing that you have asked for that isn't there is for it to be in the key, which I don't think is a necessary thing to be true as it is optional as you pointed out ... FRR will auto-generate one if not provided.

selva-nexthop:
Looks good to me.  Making it as an optional leaf node would fulfill the backend and also would be consistent with legacy way.

selva-nexthop (approved):

src/sonic-yang-models/yang-models/sonic-routing-policy-sets.yang
```diff
@@ -109,6 +99,20 @@ module sonic-routing-policy-sets {
                                   masklength_range: exact";
                 }
 
+                leaf sequence_number {
```
venkatmahalingam, line 112: If you keep this sequence_number as a field, you should allow to change it after the entry creation, but FRR doesn't allow it, how are you planning to address it?

bhouse-nexthop: the only solution there would be to delete the existing entry and then re-add it within frrcfgd ... i'll need to make sure we have all the data necessary to do that.

venkatmahalingam: yes, that's why we added the sequence number as one of the key to the PREFIX table to get the explicit delete of the entry itself from the user. My recommendation is the following,

   ```
grouping prefix-common-fields {
             leaf name {
                    type leafref {
                        path "../../../PREFIX_SET/PREFIX_SET_LIST/name";
                    }
                    description "Prefix name";
                }
          

                leaf ip_prefix {
                    /* TODO - Validate ip-prefix based on PREFIX_SET/mode */
                    type inet:ip-prefix;
                    description
                        "The prefix member in CIDR notation -- while the
                        prefix may be either IPv4 or IPv6, most
                        implementations require all members of the prefix set
                        to be the same address family.  Mixing address types in
                        the same prefix set is likely to cause an error.";
                }

                leaf masklength_range {
                    type string;
                    description
                        "Defines a range for the masklength, or 'exact' if
                        the prefix has an exact length.

                        Example: 10.3.192.0/21 through 10.3.192.0/24 would be
                        expressed as prefix: 10.3.192.0/21,
                                  masklength_range: 21..24.

                                      Example: 10.3.192.0/21 would be expressed as
                                      prefix: 10.3.192.0/21,
                                  masklength_range: exact";
                }

                leaf action {
                    type routing-policy-action-type;
                    description
                        "Permit/Deny action for the prefix list.";
                }
            }


} 

  container PREFIX {
 
           list PREFIX_NO_SEQ_LIST {
                key "name ip_prefix masklength_range";
                //ext:custom-validation ValidateIpPrefixListCfg;

                uses prefix-common-fields;                
            }

            list PREFIX_LIST {
                key "name sequence_number ip_prefix masklength_range";
                //ext:custom-validation ValidateIpPrefixListCfg;

                leaf sequence_number {
                    type uint32 {
                        range "1..4294967295";
                    }
                    description
                        "The sequence number determines the order in which prefix-list entries are applied. It must be unique
                        for each prefix in a prefix-list. Target devices should apply the prefixes in ascending order determined
                            by sequence no (low to high), rather than relying only on order in the list.";
                }

                uses prefix-common-fields; 

           }
      }


```

Comment (bradh352) [reply src/sonic-yang-models/yang-models/sonic-routing-policy-sets.yang:112]: Ok, I can see based on that why you made that recommendation.  However I'll need to see how to deal with a different number of keys in the various generators to make that happen.  One of the other problems is I don't see any tests for  frrcfgd  that take the db format and test the output it generates.  That's a bit concerning to me.  I wasn't planning on bringing into scope significant test framework additions.

Comment (selva-nexthop) [reply src/sonic-yang-models/yang-models/sonic-routing-policy-sets.yang:112]: it is getting complicated.  It is better just to have only 'list PREFIX_LIST' with the sequence_number in the key and handle it in the backend.

Comment (bhouse-nexthop) [reply src/sonic-yang-models/yang-models/sonic-routing-policy-sets.yang:112]: @selva-nexthop while we could do that in the yang model ... we'd have to support both in the backend implementation to prevent breaking users.

@venkatmahalingam I've implemented your suggestion.

File: src/sonic-frr-mgmt-framework/frrcfgd/frrcfgd.py [hunk 3]
```diff
@@ -2223,13 +2225,17 @@ def __init__(self):
                 self.prefix_set_list[key] = MatchPrefixList(entry['mode'].lower())
         pfx_table = self.config_db.get_table('PREFIX')
         for key, entry in pfx_table.items():
-            pfx_set_name, ip_pfx, len_range = key
+            if len(key) == 4:
+                pfx_set_name, seq, ip_pfx, len_range = key
```
venkatmahalingam, line 2229: Do you see any issue in extending the support for ipv6 also?

src/sonic-yang-models/tests/yang_model_tests/tests_config/route_filter.json
```diff
@@ -412,7 +429,7 @@
                 "PREFIX_SET_LIST": [
                 {
                     "name": "prefix1",
-                    "mode": "IPv6"
```
venkatmahalingam, line 438: If we support for ipv6 prefix list also, we dont need this change.

bhouse-nexthop: No, there were 2 errors here, but only supposed to be testing the invalid sequence number.  This change is still needed as the ip address provided is ipv4, so you can't set a mode of ipv6.

src/sonic-frr-mgmt-framework/frrcfgd/frrcfgd.py
bhouse-nexthop: I'm not seeing where ipv6 support is missing here.  The ip_pfx set here can be an ipv4 or ipv6 address.  The `len(key) == 4` is is because it includes the sequence_number in the key ... otherwise it is `len(key) == 3`

bhouse-nexthop:
just performed a rebase to re-run all tests.  Looks like an unrelated spurious failure occurred on the last attempt.

src/sonic-frr-mgmt-framework/frrcfgd/frrcfgd.py
venkatmahalingam: ok, thanks!

bradh352:
@qiluo-msft can this be merged?  I worked with @venkatmahalingam on an acceptable solution. Thanks!

bhouse-nexthop:
I just rebased again to try to rerun tests ... looks like it timed out (nothing to do with this PR).

bhouse-nexthop:
@lguohan @qiluo-msft any chance this can be merged? it has had reviews and all fixes made to review comments.

PR #22910, opened 2025-06-10
[FRR]: Fix a BGP check that verifies if the nexthop is invalid

cscarpitta:
Currently, in `bgp_update_martian_nexthop` we do a check to evaluate whether a nexthop is valid or not. But then the result of this check is not saved and therefore it gets lost.

```
		if (!ipv4_unicast_valid(&attr->nexthop) ||
			      bgp_nexthop_self(bgp, afi, type, stype, attr, dest));
```

This commit fixes the issue by saving the result of the check in the `nh_invalid` variable that already exists.


src/sonic-frr/patch/0002-Allow-BGP-attr-NEXT_HOP-to-be-0.0.0.0-due-to-allevia.patch
```diff
@@ -16,7 +16,7 @@ index 39172cc08..9d96a3e43 100644
  	if (CHECK_FLAG(attr->flag, ATTR_FLAG_BIT(BGP_ATTR_NEXT_HOP)))
 -		nh_invalid = (attr->nexthop.s_addr == INADDR_ANY ||
 -			      !ipv4_unicast_valid(&attr->nexthop) ||
-+		if (!ipv4_unicast_valid(&attr->nexthop) ||
```
BYGX-wcr, line 19: Why do we remove the if here?

BYGX-wcr: @cscarpitta, can you please explain this change?

cscarpitta: Hi @BYGX-wcr I'm fixing an issue in the `bgp_update_martian_nexthop()` function.

---

**Before my fix:**

```
/* Check if received nexthop is valid or not. */
bool bgp_update_martian_nexthop(struct bgp *bgp, afi_t afi, safi_t safi,
				uint8_t type, uint8_t stype, struct attr *attr,
				struct bgp_dest *dest)
{
	bool nh_invalid = false;

	...

	/* If NEXT_HOP is present, validate it:
	 * The route can have both nexthop + mp_nexthop encoded as multiple NLRIs,
	 * and we MUST check if at least one of them is valid.
	 * E.g.: IPv6 prefix can be with nexthop: 0.0.0.0, and mp_nexthop: fc00::1.
	 */
	if (CHECK_FLAG(attr->flag, ATTR_FLAG_BIT(BGP_ATTR_NEXT_HOP)))
		if (!ipv4_unicast_valid(&attr->nexthop) ||
			      bgp_nexthop_self(bgp, afi, type, stype, attr, dest));

	/* If MP_NEXTHOP is present, validate it. */
	...

	return nh_invalid;
}
```

**After my fix:**

```
/* Check if received nexthop is valid or not. */
bool bgp_update_martian_nexthop(struct bgp *bgp, afi_t afi, safi_t safi,
				uint8_t type, uint8_t stype, struct attr *attr,
				struct bgp_dest *dest)
{
	bool nh_invalid = false;

	...

	/* If NEXT_HOP is present, validate it:
	 * The route can have both nexthop + mp_nexthop encoded as multiple NLRIs,
	 * and we MUST check if at least one of them is valid.
	 * E.g.: IPv6 prefix can be with nexthop: 0.0.0.0, and mp_nexthop: fc00::1.
	 */
	if (CHECK_FLAG(attr->flag, ATTR_FLAG_BIT(BGP_ATTR_NEXT_HOP)))
		nh_invalid = !ipv4_unicast_valid(&attr->nexthop) ||
			      bgp_nexthop_self(bgp, afi, type, stype, attr, dest);

	/* If MP_NEXTHOP is present, validate it. */
	...

	return nh_invalid;
}
```

As you can see, before my fix, the if statement was having no effect, because we were doing the check to evaluate whether the nexthop is valid or not, but then we did not take any action - we ignored the result of the check.

---

In other words, I am replacing this:

```
if (!ipv4_unicast_valid(&attr->nexthop) ||
	      bgp_nexthop_self(bgp, afi, type, stype, attr, dest));
```

with this:

```
nh_invalid = !ipv4_unicast_valid(&attr->nexthop) ||
			      bgp_nexthop_self(bgp, afi, type, stype, attr, dest);
```

The function then returns `nh_invalid`.

BYGX-wcr: got it

BYGX-wcr (approved):
LGTM

PR #22911, opened 2025-06-10
[sonic-package-manager] Use tag to start application

DavidZagury:
This is done in order to be able to start using tag which will provide a user-friendly identifier when running docker ps.

Needed PR: https://github.com/sonic-net/sonic-utilities/pull/3917
#### Why I did it
To fix https://github.com/sonic-net/sonic-buildimage/issues/22124 while keeping the wanted ability to get real image while using docker ps

##### Work item tracking

#### How I did it
- Modified `docker_image_ctl.j2` to add support for using `docker_image_tag` when starting containers
- Added a new condition to check if `docker_image_tag` is defined and not none
- If a tag is specified, it will be used instead of the default `latest` tag

#### How to verify it

1. Install a package using sonic-package-manager
2. Run docker ps to verify that the container is running with the correct tag
3. Check the container configuration to ensure the tag is properly saved

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

- [ ] 
- [ ]

### DIFF
--- a/files/build_templates/docker_image_ctl.j2
+++ b/files/build_templates/docker_image_ctl.j2
@@ -744,8 +744,8 @@ start() {
 {%- if docker_container_name == "gbsyncd" %}
         -v /var/run/docker-syncd$DEV:/var/run/sswsyncd \
         "docker-$GBSYNCD_PLATFORM":latest \
-{%- elif docker_image_tag is defined and docker_image_tag is not none %}
-        {{docker_image_tag}} \
+{%- elif docker_image_reference is defined and docker_image_reference is not none %}
+        {{docker_image_reference}} \
 {%- elif docker_image_name is defined %}
         {{docker_image_name}}:latest \
 {%- else %}
--- a/files/build_templates/docker_image_ctl.j2
+++ b/files/build_templates/docker_image_ctl.j2
@@ -744,6 +744,8 @@ start() {
 {%- if docker_container_name == "gbsyncd" %}
         -v /var/run/docker-syncd$DEV:/var/run/sswsyncd \
         "docker-$GBSYNCD_PLATFORM":latest \
+{%- elif docker_image_tag is defined and docker_image_tag is not none %}
+        {{docker_image_tag}} \
 {%- elif docker_image_name is defined %}
         {{docker_image_name}}:latest \
 {%- else %}


liushilongbuaa (approved):
LGTM

files/build_templates/docker_image_ctl.j2
```diff
@@ -744,6 +744,8 @@ start() {
 {%- if docker_container_name == "gbsyncd" %}
         -v /var/run/docker-syncd$DEV:/var/run/sswsyncd \
         "docker-$GBSYNCD_PLATFORM":latest \
+{%- elif docker_image_tag is defined and docker_image_tag is not none %}
+        {{docker_image_tag}} \
```
qiluo-msft: >{{docker_image_tag}} [](http://example.com/codeflow?start=8&length=20)

Is the intention "{{docker_image_name}}:{{docker_image_tag}}" ?
If this is not your current implementation, suggest change design this way. #Closed

DavidZagury: @qiluo-msft , docker_image_tag already contains the full name used to install the docker image.
For example, in dhcp-relay it would contain _"docker-dhcp-relay:latest"_

qiluo-msft: This is confusing. Could you change the design to "{{docker_image_name}}:{{docker_image_tag}}"?

DavidZagury: It won't always have the {{docker_image_name}}:{{docker_image_tag}} schema, for example, the user can install using a image digest and then this is what it will get in in that variable.

How about changing the name to docker_image_reference?
This could mean both the both the image name with the tag and also image digest

qiluo-msft: docker_image_reference looks good name to me.

qiluo-msft (approved):

files/build_templates/docker_image_ctl.j2
DavidZagury: @qiluo-msft 
Since the sonic-utilities PR was already merged, we needed another PR to change the name:
https://github.com/sonic-net/sonic-utilities/pull/3998

PR #22970, opened 2025-06-13
update into kubesonic rollout check logic for bmp container daemon existence

FengPan-Frank:
#### Why I did it
when bmp container is rollout via kubeSonic path, its name is not "bmp", but "ksdatatest.azurecr.io/docker-sonic-bmp"

refer below

image

##### Work item tracking
- Microsoft ADO **(number only)**:33256759

#### How I did it
Update docker health check logic correspondingly, using "ps" instead of relying on container name

#### How to verify it
image

via KubeSonic path after build image.

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

- [ ] 
- [ ]

### DIFF
--- a/dockers/docker-bmp-watchdog/watchdog/src/main.rs
+++ b/dockers/docker-bmp-watchdog/watchdog/src/main.rs
@@ -11,9 +11,10 @@ struct HealthStatus {
     check_bmp_port: String,
 }
  
-fn check_bmp_supervisorctl() -> String {
-    let output = Command::new("docker")
-        .args(["exec", "-i", "bmp", "supervisorctl", "status"])
+fn check_bmp_processes_on_host() -> String {
+    let output = Command::new("sh")
+        .arg("-c")
+        .arg("ps aux | grep -E '/usr/local/bin/bmpcfgd|/usr/bin/openbmpd' | grep -v grep")
         .output();
 
     match output {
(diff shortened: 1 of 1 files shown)


dockers/docker-bmp-watchdog/watchdog/src/main.rs
```diff
@@ -13,7 +13,7 @@ struct HealthStatus {
  
 fn check_bmp_supervisorctl() -> String {
     let output = Command::new("docker")
-        .args(["exec", "-i", "bmp", "supervisorctl", "status"])
+        .args(["exec", "-i", "ksdatatest.azurecr.io/docker-sonic-bmp", "supervisorctl", "status"])
```
qiluo-msft: We should not hard-code the registery server name #Closed

FengPan-Frank: I changed it to "ps" instead of relying on docker container name.

FengPan-Frank: Discussed with @make1980 , it is unnecessary to detect docker health in watchdog since kubelete will do that. Thus get rid of this probe function directly.

make1980 (approved):
:shipit:

PR #23005, opened 2025-06-17
Add retry for cgroup memory usage file check

FengPan-Frank:
#### Why I did it
Add retry for cgroup memory usage file check since there's very rarely race occurs, during config reload/reboot.

##### Work item tracking
https://github.com/sonic-net/sonic-buildimage/issues/21880

#### How I did it
Add retry for cgroup memory usage file check

#### How to verify it
Sanity check with PR verification.

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

- [ ] 
- [ ]


qiluo-msft:
>         with open(docker_memory_usage_file_path, 'r') as file:

Is there extreme corner case, that the path exists when you check above, but suddently removed when you open?
If possible, then could you handle file not existing exception with open statement? #Closed

---
Refers to: files/image_config/monit/memory_checker:137 in 80039ab. [](commit_id = 80039ab1f59cb1449b2f970cc10a08a04e84ce5b, deletion_comment = False)

FengPan-Frank:
> > ```
> >     with open(docker_memory_usage_file_path, 'r') as file:
> > ```
> 
> Is there extreme corner case, that the path exists when you check above, but suddently removed when you open? If possible, then could you handle file not existing exception with open statement?
> 
> Refers to: files/image_config/monit/memory_checker:137 in [80039ab](https://github.com/sonic-net/sonic-buildimage/commit/80039ab1f59cb1449b2f970cc10a08a04e84ce5b). [](commit_id = [80039ab](https://github.com/sonic-net/sonic-buildimage/commit/80039ab1f59cb1449b2f970cc10a08a04e84ce5b), deletion_comment = False)

To avoid this potential race, change it into open with retry directly.

make1980:
>         with open(docker_memory_usage_file_path, 'r') as file:

agree with the concern. ideally all the IO operations should have similar retry logic - could you also add that for open()? we should have a generic function to handle the retry as well.

---
In reply to: [2989566848](https://github.com/sonic-net/sonic-buildimage/pull/23005#issuecomment-2989566848) [](http://example.com/codeflow?ancestors=2989566848)

---
Refers to: files/image_config/monit/memory_checker:137 in 80039ab. [](commit_id = 80039ab1f59cb1449b2f970cc10a08a04e84ce5b, deletion_comment = False)

files/image_config/monit/memory_checker
```diff
@@ -121,8 +121,16 @@ def get_memory_usage(container_id):
...
     docker_memory_usage_file_path = CGROUP_DOCKER_MEMORY_DIR + container_id + "/memory.usage_in_bytes"
-    if not os.path.exists(docker_memory_usage_file_path):
-        syslog.syslog(syslog.LOG_ERR, ERROR_CGROUP_MEMORY_USAGE_NOT_FOUND.format(docker_memory_usage_file_path, container_id))
+
+    # Retry once with delay if path does not exist
+    for attempt in range(3):
+        if os.path.exists(docker_memory_usage_file_path):
+            break
+        if attempt == 0:
+            time.sleep(0.1)  # 100 milliseconds delay
```
make1980: shouldn't we sleep for the second attempt as well? #Closed

make1980 (changes requested):
:clock1:

files/image_config/monit/memory_checker
```diff
@@ -109,30 +109,25 @@ def get_container_id(container_name):
...
 
 def get_memory_usage(container_id):
-    """Reads the container's memory usage from the control group subsystem's file
-    '/sys/fs/cgroup/memory/docker/<container_id>/memory.usage_in_bytes'.
-    Args:
-        container_id: A string indicates the full ID of a container.
-    Returns:
-        memory_usage_in_bytes: A string indicates memory usage (Bytes) of a container.
-    """
-    memory_usage_in_bytes = ""
```
make1980, line 119: why is this removed? #Closed

files/image_config/monit/memory_checker
```diff
@@ -109,30 +109,25 @@ def get_container_id(container_name):
...
-            memory_usage_in_bytes = file.read().strip()
-    except IOError as err:
-        syslog.syslog(syslog.LOG_ERR, ERROR_CONTAINER_MEMORY_USAGE_NOT_FOUND.format(container_id))
-        sys.exit(INTERNAL_ERROR)
+    for attempt in range(3):
+        try:
+            with open(docker_memory_usage_file_path, 'r') as file:
+                return file.read().strip()
+        except FileNotFoundError:
+            if attempt == 0:
```
make1980: should we add a unit test for this case? #Closed

files/image_config/monit/memory_checker
```diff
@@ -109,30 +109,25 @@ def get_container_id(container_name):
...
+                time.sleep(0.1)  # retry after short delay
+            else:
+                break
+        except IOError:
+            syslog.syslog(syslog.LOG_ERR, ERROR_CONTAINER_MEMORY_USAGE_NOT_FOUND.format(container_id))
+            sys.exit(INTERNAL_ERROR)
 
-    return memory_usage_in_bytes
+    syslog.syslog(syslog.LOG_ERR, ERROR_CGROUP_MEMORY_USAGE_NOT_FOUND.format(docker_memory_usage_file_path, container_id))
+    sys.exit(INTERNAL_ERROR)
```
make1980, line 137: are you missing an else: here? #Closed

files/image_config/monit/memory_checker
```diff
@@ -109,30 +109,25 @@ def get_container_id(container_name):
...
-        syslog.syslog(syslog.LOG_ERR, ERROR_CGROUP_MEMORY_USAGE_NOT_FOUND.format(docker_memory_usage_file_path, container_id))
-        sys.exit(INTERNAL_ERROR)
 
-    try:
-        with open(docker_memory_usage_file_path, 'r') as file:
-            memory_usage_in_bytes = file.read().strip()
-    except IOError as err:
-        syslog.syslog(syslog.LOG_ERR, ERROR_CONTAINER_MEMORY_USAGE_NOT_FOUND.format(container_id))
-        sys.exit(INTERNAL_ERROR)
+    for attempt in range(3):
```
qiluo-msft, line 123: >range [](http://example.com/codeflow?start=19&length=5)

seems you want to try at most 3 times (including first time), but you actually at most try 2 times. Is it a mistake?

Will be this script called frequently? do we really need retry? #Closed

files/image_config/monit/memory_checker
FengPan-Frank: updated.

files/image_config/monit/memory_checker
FengPan-Frank: previously since return logic has changed, add it back with some updates.

files/image_config/monit/memory_checker
FengPan-Frank: refer https://github.com/sonic-net/sonic-buildimage/blob/c151c73c9eae380ecf9533b5e938c6ab1542c474/files/image_config/monit/tests/test_memory_checker.py#L64, this doesn't change the functionality thus no impact to unit test.

files/image_config/monit/memory_checker
FengPan-Frank: no else here, if not return with file length it should raise error.

files/image_config/monit/memory_checker
FengPan-Frank: refer https://github.com/sonic-net/sonic-buildimage/issues/21880, as pointed in issue it's very rarely occurred, I guess retry should be best effort for this script.

PR #23010, opened 2025-06-17
fix arp_update template rendering when STATIC_ROUTE has blackhole attribute

abdosi:
What I did:

arp_update jinja2 template rendering fails when we have STATIC_ROUTE with only blackhole attribute. Added check to parse nexthop only if it is present.

How I verify:

Manual Verification.


BYGX-wcr (approved):
LGTM

PR #23011, opened 2025-06-17
Updated Static route yang model

abdosi:
What I did:

Updated Static Route Yang model to have `advertise` field in STATIC_ROUTE container with keys as `{vrf,prefix}`

How I verify:

Manual Verification
UT updated.


src/sonic-yang-models/tests/yang_model_tests/tests/static_route.json
```diff
@@ -36,6 +39,9 @@
     "STATIC_ROUTE_TEST_BLACKHOLE_INVALID": {
         "desc": "Configure with invalid value for blackhole",
         "eStrKey": "Pattern"
+    },
+    "STATIC_ROUTE_TEST_WITH_ADVERTISE_INVALID": {
+        "desc": "Configure static route with invalid advertise flag value"
```
BYGX-wcr: Are you sure we don't need an eStrKey here?

BYGX-wcr:
As I remembered, the YANG model will also be tested by sonic-yang-mgmt. So, you may want to update tests/files/sample_config_db.json which is consumed by sonic-yang-mgmt as I did in https://github.com/sonic-net/sonic-buildimage/pull/21560/files

src/sonic-yang-models/tests/yang_model_tests/tests_config/static-route.json
```diff
@@ -688,6 +737,55 @@
             }
         }
     },
+        "STATIC_ROUTE_TEST_WITH_ADVERTISE_INVALID": {
```
qiluo-msft: Could you fix the json indentation starting from this line?

prsunny (approved):

src/sonic-yang-models/yang-models/sonic-static-route.yang
```diff
@@ -95,6 +95,12 @@ module sonic-static-route {
           description
             "When interface is specified, forwarding happens through it";
         }
+        leaf advertise {
+          type string {
```
ganglyu, line 99: Can we use boolean?

https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-yang-models/yang-templates/sonic-types.yang.j2#L259-L263

wen587: same ask here.

PR #23054, opened 2025-06-23
Solve the error problem when the MTU configuration exceeds 9100

KroosMicas:
#### Why I did it
 As described in this issue : https://github.com/sonic-net/sonic-buildimage/issues/23045
The broadcom TH4/TH5 chip can not config MTU over 9100

##### Work item tracking

#### How I did it
Modify the knet configuration file to enable the MTU to be greater than 9100

#### How to verify it
The MTU can be configured to be greater than 9100.

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [x] 202405
- [x] 202411
- [x] 202505

- [ ] 
- [ ] 

image

### DIFF
--- a/platform/broadcom/saibcm-modules/debian/opennsl-modules.init
+++ b/platform/broadcom/saibcm-modules/debian/opennsl-modules.init
@@ -47,7 +47,7 @@ function load_kernel_modules()
     if [[ $is_ltsw_chip -eq 1 ]]; then
         modprobe psample
         modprobe linux_ngbde
-        modprobe linux_ngknet
+        modprobe linux_ngknet rx_buffer_size=$rx_buffer_size
         modprobe linux_ngknetcb
     else
         modprobe linux-kernel-bde dmasize=$dmasize maxpayload=128 debug=4 dma_debug=1 usemsi=$usemsi
@@ -93,6 +93,7 @@ function load_platform_env()
     dmasize=32M
     usemsi=0
     is_ltsw_chip=0
+    rx_buffer_size=9216
 
     # Source the platform env file
     env_file="/usr/share/sonic/device/$platform/platform_env.conf"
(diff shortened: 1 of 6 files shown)


bingwang-ms:
@saiarcot895 Can you help review this PR?

platform/broadcom/saibcm-modules/debian/opennsl-modules.init
```diff
@@ -47,7 +47,7 @@ function load_kernel_modules()
     if [[ $is_ltsw_chip -eq 1 ]]; then
         modprobe psample
         modprobe linux_ngbde
-        modprobe linux_ngknet
+        modprobe linux_ngknet rx_buffer_size=9238
```
anders-nexthop: Can we make `rx_buffer_size` configurable instead? Otherwise we will see this same issue with higher MTU than 9238, and also we may want to use a smaller buffer size if users intent to only use smaller MTU.

micas-net: > Can we make `rx_buffer_size` configurable instead? Otherwise we will see this same issue with higher MTU than 9238, and also we may want to use a smaller buffer size if users intent to only use smaller MTU.

Hi anders, have fixed as your comments, thanks!

anders-nexthop: Thanks! The rest looks okay to me but I am not an expert in this area.

micas-net:
@anders-nexthop Do me a favor to help approved it if no other question, thanks!

AndrewQu-Micas (approved):
Approved.  Thanks

anders-nexthop (commented):
Looks okay to me as well. @lguohan can you take as look as well as a code-owner?

anders-nexthop (approved):

Praveen-Brcm (approved):

PR #23058, opened 2025-06-23
fix issue in logger

muhammadalihussnain:
This PR addresses two issues encountered with the dhcp_dos_logger.service in the PR [18947](https://github.com/sonic-net/sonic-buildimage/pull/18947)

    Permission Issue:
        The dhcp_dos_logger.py script was missing execute permissions, service to fail with a Permission denied  error.
        Fixed by ensuring the script has the correct permissions using chmod 755.
    TypeError:
        The script attempted to use a string pattern on a bytes-like object, causing a runtime error.
        Fixed by decoding the bytes object to a string before using it with re.search.

Testing:

    Verified that the service starts correctly and logs DHCP DoS events as expected after applying these changes locally.

Image of Test:
Screenshot from 2025-06-23 14-35-30


muhammadalihussnain:
@qiluo-msft @yaqiangz @lguohan
I have updated the code PR . Kindly help to review and merge. Thanks

yaqiangz:
Would this service start automatically by default even if there is not any ddos configuration in interface?

muhammadalihussnain:
> Would this service start automatically by default even if there is not any ddos configuration in interface?

  service will start automatically, but will log only if there are drop counters on the ports.

yaqiangz (approved):
LGTM

muhammadalihussnain:
hi @lguohan , can you please approve the changes. Thanks

qiluo-msft (approved):

PR #23106, opened 2025-06-26
[logrotate] Add archived file clean up on log rotation

yejianquan:
##### Work item tracking

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

- [ ] 
- [ ]


yejianquan:
Hi @zbud-msft @liuh-80 could you help to review this PR?

yejianquan:
Tested the change on Cisco chassis.
https://elastictest.org/scheduler/testplan/685fb77b80f9d4ac435dfb35
20/20 passed 
image

files/image_config/logrotate/rsyslog.j2
```diff
@@ -123,5 +77,17 @@
...
+        # The clean up removes archived files till log partition has enough space
+        # So one time running during a logrotate is enough.
+        # Supposedly, the clean up happens on the firstaction which before the rotations.
+        # But there's a corner case, that on firstaction, no archived log files found(all files are origin logs, e.g. syslog, swss.rec)
+        # After rotated all log files are empty since log partition is full, then logrotate skips since it asserts notifempty and file size.
+        # On that condition, we need to clean up on postrotate
+        # Again, one time clean up during a logrotate round is enough, more running just add useless scanning.
+        # Hook and trigger clean up on rotation on syslog(last rotation log in the configuration), to avoid useless running on other log files
+        if [ $(echo $1 | grep -c "/var/log/syslog") -gt 0 ]; then
+{% include 'on_demand_archived_log_clean_up.j2' %}
```
zbud-msft, line 90: Is there some check that we need to invoke before doing the clean up a second time?

files/image_config/logrotate/rsyslog.j2
```diff
@@ -37,6 +36,7 @@
 /var/log/swss/sairedis*.rec
 /var/log/swss/swss*.rec
 /var/log/swss/responsepublisher.rec
+/var/log/syslog
```
qiluo-msft, line 39: Is there a reason to change the order of this line? #Closed

files/image_config/logrotate/rsyslog.j2
yejianquan: Hi @zbud-msft, no check is needed, the check/cleanup in on_demand_archived_log_clean_up.j2 is idempotent.
It firstly checks '$USED_KB -lt $THRESHOLD_KB' , so it will be skipped if no removal is needed.
But it still spends some time to calculate $USED_KB and $THRESHOLD_KB, so I only bind this second time to syslog rotation and skip on all other log file rotations.

files/image_config/logrotate/rsyslog.j2
yejianquan: Hi @qiluo-msft, you find the blind point, very nice question and there's a deisgn behind.
I bind the second rotation only on the syslog files to reduce meaningless removal attempts(the calculation of USED_KB and THRESHOLD_KB costs few seconds)
image
Actually any random file is bind-able, but I choose syslog file since it is supposed the most stable log file we will have and always be not empty hence enable to trigger rotation.
One tricky thing during my investigation is, the rotation happens following the sequence of log file configuration:
image
With original sequence, only auth.log/cron.log/syslog finished rotation and archive, so at most 3 files will be removed in the corner and faulty case(more details in https://github.com/sonic-net/sonic-buildimage/issues/23053)
But it won't help, because most of the storage is used by swss logs.
Hence need to adjust the sequence of syslog to the end, so the syslog-rotation-triggered on-demand-remove can always delete the archived files for all logs.

files/build_templates/sonic_debian_extension.j2
```diff
@@ -443,6 +443,7 @@ echo "system-health.service" | sudo tee -a $GENERATED_SERVICE_FILE
 # Copy logrotate.d configuration files
 sudo cp -f $IMAGE_CONFIGS/logrotate/logrotate.d/* $FILESYSTEM_ROOT/etc/logrotate.d/
 sudo cp $IMAGE_CONFIGS/logrotate/rsyslog.j2 $FILESYSTEM_ROOT_USR_SHARE_SONIC_TEMPLATES/
+sudo cp $IMAGE_CONFIGS/logrotate/on_demand_archived_log_clean_up.j2 $FILESYSTEM_ROOT_USR_SHARE_SONIC_TEMPLATES/
```
qiluo-msft, line 446: >on_demand_archived_log_clean_up [](http://example.com/codeflow?start=33&length=31)

Where is this j2 file be used? #Closed

yejianquan: in rsyslog.j2, 
firstaction and postrotate

yejianquan:
Hi @qiluo-msft @zbud-msft , comments replied, thanks

files/build_templates/sonic_debian_extension.j2
qiluo-msft: My question is about where this j2 file is used to render a config file?

yejianquan: during service logrotate-config restart
https://github.com/sonic-net/sonic-buildimage/blob/master/files/image_config/logrotate/logrotate-config.service
https://github.com/sonic-net/sonic-buildimage/blob/master/files/image_config/logrotate/logrotate-config.sh

logrotate-config.sh will render rsyslog.j2(which reuses on_demand_archived_log_clean_up.j2)

Issue #23117, opened 2025-06-27
Bug: SNMP not work when eth0 using ipv6 link local address and under mgmt vrf


mhchann:
### Is it platform specific

generic

### Importance or Severity

Medium

### Description of the bug

With enable MGMT VRF, and set link local address on eth0
SNMP will not work with following error messages
```
netsnmp_udp6: recvfrom fd 8 got 43 bytes (from UDP/IPv6: [fe80::4a91:3eb8:1118:4e
86%mgmt]:35596)
netsnmp_udp6: send 48 bytes from 0x55c63a044b44 to UDP/IPv6: [fe80::4a91:3eb8:111
8:4e86%mgmt]:35596 on fd 8
netsnmp_udp6: sendto error, rc -1 (errno 101)
send response: Failure in sendto
```

### Steps to Reproduce

Enable MGMT VRF with command `config vrf add mgmt`
and set ipv6 link local address on eth0

### Actual Behavior
SNMP not work when eth0 using ipv6 link local address and under mgmt vrf

### Expected Behavior
SNMP can work

### Relevant log output

```shell

```

### Output of `show version`, `show techsupport`

```shell

```

PR #23148, opened 2025-06-30
[SNMP] Fix IPv6 link local failed when enable management VRF

mhchann:
#### Why I did it
Fix issue #23117
When SNMP binds to `eth0` using a link-local address (`<link-local>%eth0`), there's a mismatch in interface scope handling:
1. SNMP correctly binds its socket to `eth0`
2. When `recvfrom()` retrieves the sender's address, the kernel returns the link-local address scoped to the `mgmt` interface (`<remote_addr>%mgmt`) instead of `eth0`
3. SNMP tries to send responses using `<remote_addr>%mgmt` as the destination, but this creates a routing/interface mismatch since the original socket was bound to `eth0`

The problem occurs because `eth0` is enslaved to the `mgmt` interface. In Linux networking, when an interface is enslaved, the kernel may report the master interface (`mgmt`) in address resolution rather than the actual physical interface (`eth0`) where the packet was received.

#### How I did it
Involving and correcting the IPv6 scope ID at the transport layer:
* Add a `sin6_scope_id` field to the `netsnmp_transport_s` structure to remember the original scope ID used during socket binding
* In `netsnmp_udp6_transport_bind()`, capture and store the scope ID from the bind address (`addr->sin6_scope_id`) into the transport structure (`t->sin6_scope_id`)
* Correct scope id on receive. In `netsnmp_udp6_recv()`, after receiving a packet via `recvfrom()`, if received address's scope ID differ from the stored original scope ID, override the received scope ID with the original one

This ensures that when SNMP later uses the received address for `sendto()` operations, it uses the same interface scope (`%eth0`) that was originally bound, rather than the master interface scope (`%mgmt`) returned by the kernel due to interface enslaving.

#### How to verify it
snmp can work with ipv6 link local even eth0 under mgmg VRF 

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

- [ ] 
- [ ]

### DIFF
--- a/src/snmpd/patch-5.9.3+dfsg/series
+++ b/src/snmpd/patch-5.9.3+dfsg/series
@@ -3,3 +3,4 @@
 0012-agent-Makefile.in-Unbreak-the-enable-minimalist-buil.patch  
 0013-enable-parallel-build-for-net-snmp.patch  
 cross-compile-changes.patch 
+0014-SNMP-ipv6-link-local-under-MGMT-VRF.patch
--- a/src/snmpd/patch-5.9.3+dfsg/0014-SNMP-ipv6-link-local-under-MGMT-VRF.patch
+++ b/src/snmpd/patch-5.9.3+dfsg/0014-SNMP-ipv6-link-local-under-MGMT-VRF.patch
@@ -1,0 +1,55 @@
+From cd79a528dee3348c65cbf501cec4010e2e22d717 Mon Sep 17 00:00:00 2001
+From: mhchann <mhchann082@gmail.com>
+Date: Mon, 30 Jun 2025 16:37:28 +0800
+Subject: [PATCH 1/1] 0014-SNMP-ipv6-link-local-under-MGMT-VRF
+
+---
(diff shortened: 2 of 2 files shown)


bingwang-ms:
@qiluo-msft Can you help review this PR?

src/snmpd/patch-5.9.3+dfsg/series
```diff
@@ -3,3 +3,4 @@
 0012-agent-Makefile.in-Unbreak-the-enable-minimalist-buil.patch  
 0013-enable-parallel-build-for-net-snmp.patch  
 cross-compile-changes.patch 
+0014-SNMP-ipv6-link-local-under-MGMT-VRF.patch
```
qiluo-msft, line 6: >0014 [](http://example.com/codeflow?start=0&length=4)

Could you add a testcase in sonic-mgmt repo for this issue/bugfix?

PR #23153, opened 2025-06-30
Fix process_audit rule in auditd

maipbui:
##### Work item tracking

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

- [ ] 
- [ ]

### DIFF
--- a/dockers/docker-auditd/auditd_config_files/30-audisp-tacplus.rules
+++ b/dockers/docker-auditd/auditd_config_files/30-audisp-tacplus.rules
@@ -22,8 +22,8 @@
 # We assume here that user 1000 is the first local user, and therefore
 # should not be looked up for tacacs.  You may need to change this for
 # your local configuration
--a always,exit -F arch=b32 -S execve -S exit -S exit_group -F auid>1000 -F auid!=4294967295 -k tacplus
--a always,exit -F arch=b64 -S execve -S exit -S exit_group -F auid>1000 -F auid!=4294967295 -k tacplus
+-a always,exit -F arch=b32 -S exit -S exit_group -F auid>1000 -F auid!=4294967295 -k tacplus
+-a always,exit -F arch=b64 -S exit -S exit_group -F auid>1000 -F auid!=4294967295 -k tacplus
 
 # In newer distributions (such as debian jessie), a number of auditing events
 # are logged by default even after the -D initialization.  If you are using
--- a/dockers/docker-auditd-watchdog/watchdog/src/main.rs
+++ b/dockers/docker-auditd-watchdog/watchdog/src/main.rs
@@ -7,8 +7,8 @@ static NSENTER_CMD: &str = "nsenter --target 1 --pid --mount --uts --ipc --net";
 
 // Expected hash values
 static AUDITD_CONF_HASH: &str = "7cdbd1450570c7c12bdc67115b46d9ae778cbd76";
-static AUDITD_RULES_HASH_DEFAULT: &str = "3042c67383dccff079755c5f3daa2820e2ca392a";
-static AUDITD_RULES_HASH_NOKIA: &str = "fffb00199104c493d04a621b257c579c48d77225";
+static AUDITD_RULES_HASH_DEFAULT: &str = "99bf4b5a80ac2b8d03fa69d1b8e7f7b2a1423ba8";
+static AUDITD_RULES_HASH_NOKIA: &str = "6002dcd7ef2cbeabc7a60925bd603ebe901d58be";
 
 // Helper to run commands
 fn run_command(cmd: &str) -> Result<String, String> {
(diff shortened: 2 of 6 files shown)


dockers/docker-auditd/auditd_config_files/30-audisp-tacplus.rules
```diff
@@ -22,8 +22,8 @@
 # We assume here that user 1000 is the first local user, and therefore
 # should not be looked up for tacacs.  You may need to change this for
 # your local configuration
--a always,exit -F arch=b32 -S execve -S exit -S exit_group -F auid>1000 -F auid!=4294967295 -k tacplus
--a always,exit -F arch=b64 -S execve -S exit -S exit_group -F auid>1000 -F auid!=4294967295 -k tacplus
```
qiluo-msft, line 26: >execve [](http://example.com/codeflow?start=30&length=6)

What is the purpose of this change? how necessary is this change?

qiluo-msft: since you modified dockers/docker-auditd/auditd_config_files/30-audisp-tacplus.rules, it possibly impact the per-command auditing feature. Did you run sonic-mgmt container upgrade testcase, especially auditing testcases?

qiluo-msft: @liuh-80 could you also review this part?

maipbui: if we leave 30-audisp-tacplus.rules unchanged, auditd will match with rule 1 and rule 2, and stop there, it won’t reach rule 3 or 4, so nothing gets logged under the process_audit key.
 
1.-a always,exit -F arch=b32 -S exit,execve,exit_group -F auid>1000 -F auid!=-1 -F key=tacplus
2. -a always,exit -F arch=b64 -S execve,exit,exit_group -F auid>1000 -F auid!=-1 -F key=tacplus
3. -a always,exit -F arch=b64 -S execve -F auid>=1000 -F auid!=-1 -F key=process_audit -F key=tacplus
4. -a always,exit -F arch=b32 -S execve -F auid>=1000 -F auid!=-1 -F key=process_audit -F key=tacplus

maipbui: with SYSCALL execve, key=tacplus will be key=70726F636573735F617564697401746163706C7573 in log message, key=70726F636573735F617564697401746163706C7573 covers both process_audit and tacplus. with SYSCALL exit, exit_group, the behavior is unchanged

liuh-80:
Hi @maipbui , you need also add a patch for this change in https://github.com/sonic-net/sonic-buildimage/tree/master/src/tacacs/audisp/patches

Because 30-audisp-tacplus.rules was installed by tacacs auditd plugin, if this plugin upgrade, your change may conflict or overwrite by updated rules file.

maipbui:
> Hi @maipbui , you need also add a patch for this change in https://github.com/sonic-net/sonic-buildimage/tree/master/src/tacacs/audisp/patches
> 
> Because 30-audisp-tacplus.rules was installed by tacacs auditd plugin, if this plugin upgrade, your change may conflict or overwrite by updated rules file.

@liuh-80, auditd container will remove and overwrite rules files https://github.com/sonic-net/sonic-buildimage/blob/master/dockers/docker-auditd/config_checker.py#L111-L115. if this plugin upgrade, I think we also need to update rules file in auditd container https://github.com/sonic-net/sonic-buildimage/tree/master/dockers/docker-auditd/auditd_config_files?

liuh-80:
> > Hi @maipbui , you need also add a patch for this change in https://github.com/sonic-net/sonic-buildimage/tree/master/src/tacacs/audisp/patches
> > Because 30-audisp-tacplus.rules was installed by tacacs auditd plugin, if this plugin upgrade, your change may conflict or overwrite by updated rules file.
> 
> @liuh-80, auditd container will remove and overwrite rules files https://github.com/sonic-net/sonic-buildimage/blob/master/dockers/docker-auditd/config_checker.py#L111-L115. if this plugin upgrade, I think we also need to update rules file in auditd container https://github.com/sonic-net/sonic-buildimage/tree/master/dockers/docker-auditd/auditd_config_files?

Here is my concern: If the audisp-tacplus.rules file in the auditsp-tacplus project changes in the future, the auditd container might overwrite or misinterpret the file, potentially breaking the TACACS accounting feature.

To prevent this, it would be better to add a patch that explicitly depends on the current version of audisp-tacplus.rules. This way, if the file changes, the patch will fail—alerting us to the need for an update and helping avoid silent failures.

maipbui:
@liuh-80 I added a patch in this PR https://github.com/sonic-net/sonic-buildimage/pull/23196

PR #23185, opened 2025-07-02
Correctly fetch port config when HWSKU doesn't match between sonic-cfggen and configDB

ram25794:
#### Why I did it

Incorrect golden config was being generated for non-default SKUs when using minigraph. This issue occurred because sonic-cfggen was reading port configuration from the config DB, even when the hardware SKU in the config DB differed. Incorrect ports and port_alias_map data from config db was causing the issue.

##### Work item tracking

#### How I did it
Modified the logic to ensure that port configuration is fetched from the config DB only when the hardware SKU in the config DB matches the hardware SKU for which sonic-cfggen is generating the config.

#### How to verify it

Generated golden configurations for non-default SKUs and confirmed that the output was correct.

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

- [ ] 
- [ ]


qiluo-msft:
It is not clear what is actually not incorrect by "Incorrect golden config was being generated for non-default SKUs when using minigraph."

Do you have repro steps? what is the command line used?

vivekkarri:
This is the command used to generate golden config_db from minigraph. Ref: https://github.com/sonic-net/sonic-mgmt/blob/4beb7a7a3df8cc74c7dc9a38ff6b21226827ff20/ansible/library/generate_golden_config_db.py#L244

`sonic-cfggen -H -m -j /etc/sonic/init_cfg.json --print-data`

When the switch boots from ONIE, it has default_sku for the platform. Minigraph can have any legal SKU. 

sonic-cfggen reads the PORT config from DB. However, it should originally read from hwsku.json/port_config.init of that SKU to generate correct config when the SKU in DB and SKU in minigraph differs

This PR fixes that bug in sonic-cfggen

vivekkarri:
> It is not clear what is actually not incorrect by "Incorrect golden config was being generated for non-default SKUs when using minigraph."
> 
> Do you have repro steps? what is the command line used?

@qiluo-msft Can you signoff on this?

qiluo-msft:
I think the right fix should be inside generate_smartswitch_golden_config_db(). We can first generate some valid config for the expected hwsku first. for example `sudo sonic-cfggen --preset l2 -p -H -k Mellanox-SN2700-D48C8`

vivekkarri:
> I think the right fix should be inside generate_smartswitch_golden_config_db(). We can first generate some valid config for the expected hwsku first. for example `sudo sonic-cfggen --preset l2 -p -H -k Mellanox-SN2700-D48C8`

This will not work. hwsku is anyway inferred from the minigraph, so argument '-k Mellanox-SN2700-D48C8'  doesn't add any value .

This is the code path today

https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-config-engine/sonic-cfggen#L393 
https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-config-engine/minigraph.py#L2064 
https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-config-engine/portconfig.py#L174
  
```
-> deep_update(data, parse_xml(minigraph, platform, asic_name=asic_name)) 
-> hwsku_qn = QName(ns, "HwSku")
-> (ports, alias_map, alias_asic_map) = get_port_config(hwsku=hwsku, platform=platform, port_config_file=port_config_file, asic_name=asic_name, hwsku_config_file=hwsku_config_file)
-> if config_db is not None and port_config_file is None:
```

vivekkarri:
Config Deploy flow on SmartSwitch:
1) Load from ONIE
2) Default SKU loads (Mellanox-SN4280-028)
3) In deploy-mg stage, it uses get_golden_config generate to get the config_db.json from minigraph.xml (It was added here https://github.com/sonic-net/sonic-mgmt/pull/15379, In this step the folowing command is run https://github.com/sonic-net/sonic-mgmt/blob/4e4e4c5df51b8decd7e7a1663a778236e4bdeab9/ansible/library/generate_golden_config_db.py#L395)
4) Now, when the SKU has a different port split from the default SKU the config generated is wrong. We are fixing that in this PR.

qiluo-msft (approved):

PR #23200, opened 2025-07-03
Add rsyslog plugin for update rsyslog count to COUNTERS_DB

liuh-80:
Add rsyslog plugin for update rsyslog count to COUNTERS_DB

#### Why I did it
Improve SONiC syslog feature.

##### Work item tracking
- Microsoft ADO **(number only)**:25112710

#### How I did it
Add a new rsyslog plugin, which will update rsyslog counter every 1 minutes.

admin@vlab-01:~$ sonic-db-cli COUNTERS_DB hgetall SYSLOG_COUNTER
{'COUNT': '12122'}

#### How to verify it
Pass all test case.
Manually verified the feature works:
// enable feature and reload config
admin@vlab-01:~$ sonic-db-cli CONFIG_DB hset "DEVICE_METADATA|localhost" syslog_counter true

// check syslog counter
admin@vlab-01:~$ sonic-db-cli COUNTERS_DB hgetall SYSLOG_COUNTER
{'COUNT': '12122'}
...
admin@vlab-01:~$ sonic-db-cli COUNTERS_DB hgetall SYSLOG_COUNTER
{'COUNT': '12151'}

// CPU and memory utilzation
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root       35507  0.1  0.1  88400  8044 ?        Sl   02:13   0:00 /usr/bin/syslog-counter

admin@vlab-01:~$ ps -o pid,rss,comm -p 35507 | awk 'NR==2 {printf "PID: %s, Memory: %.2f MB, Command: %s\n", $1, $2/1024, $3}'
PID: 35507, Memory: 7.86 MB, Command: syslog-counter

Test case PR: https://github.com/sonic-net/sonic-mgmt/pull/19749

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

- [ ] 
- [ ] 

Add rsyslog plugin for update rsyslog count to COUNTERS_DB

### DIFF
--- a/files/image_config/rsyslog/rsyslog.d/00-sonic.conf.j2
+++ b/files/image_config/rsyslog/rsyslog.d/00-sonic.conf.j2
@@ -5,6 +5,12 @@ template(name="prog_msg" type="list") {
     constant(value="\n")
 }
 
+{% if include_system_eventd == "y" and build_reduce_image_size != "y" %}
+
+$ModLoad omprog
+
+{% endif %}
+
 if re_match($programname, "bgp[0-9]*#(frr|zebra|staticd|watchfrr)") then {
     /var/log/frr/zebra.log
     stop


files/scripts/syslog_counter.py
```diff
@@ -0,0 +1,69 @@
...
+        # update counter every 1 minutes
+        current_time = time.time()
+        if current_time - last_update_time >= 60:
+            counters_db.set('COUNTERS_DB', 'SYSLOG_COUNTER', 'COUNT', syslog_count)
+
+        # Send the status code to rsyslog:
+        print("OK", flush=True)
+        line = sys.stdin.readline()
+except Exception:
+    logging.exception("syslog counter plugin unrecoverable error, exiting program")
```
qiluo-msft: >logging [](http://example.com/codeflow?start=4&length=7)

could you also catch the exception as var, and logging its details? #Closed

files/scripts/syslog_counter.py
```diff
@@ -0,0 +1,69 @@
...
+        syslog_count += 1
+
+        # update counter every 1 minutes
+        current_time = time.time()
+        if current_time - last_update_time >= 60:
+            counters_db.set('COUNTERS_DB', 'SYSLOG_COUNTER', 'COUNT', syslog_count)
+
+        # Send the status code to rsyslog:
+        print("OK", flush=True)
+        line = sys.stdin.readline()
```
qiluo-msft: >readline [](http://example.com/codeflow?start=25&length=8)

If syslog volume is large, will this loop burn CPU cycles? Did you test extreme large volume situation? Could you get some uplimit volume number? Should we protect this process to burn CPU is volume is larger than the limit? #Closed

files/scripts/syslog_counter.py
```diff
@@ -0,0 +1,69 @@
+#!/usr/bin/env python3
```
qiluo-msft: >python3 [](http://example.com/codeflow?start=15&length=7)

For high performance, and low memory consumption, suggest use Rust for new daemon/long running process. #Closed

files/scripts/syslog_counter.py
liuh-80: Fixed

files/scripts/syslog_counter.py
liuh-80: This plugin takes 10MB memory and not increase when syslog volume is large.

         0.00 Mb PID COMMAND
        10.41 Mb 9206 python3 /usr/bin/syslog_counter.py

But the CPU utilization will increase according syslog volume:

10 per-second:
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root        3301  1.2  0.3  45636 22836 ?        S    08:43   0:07 python3 /usr/bin/syslog_counter.py

100 per-second:
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root        9206  2.8  0.3  45644 23300 ?        R    10:06   1:58 python3 /usr/bin/syslog_counter.py

1000 per-second:
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
         root        9206  9.6  0.3  45644 22852 ?        S    10:06   0:34 python3 /usr/bin/syslog_counter.py

And rsyslogd take 20% CPU when 1000 syslog per-second.

qiluo-msft: To reduce CPU overhead, is it possible to read multiple lines and reduce the invocation of time()? For example using select with timeout. Currently in high volume, you keep calling time() per log line.

liuh-80: The plugin cannot process multiple lines at once. After reading each line, it must respond with an "OK" message. If this response is missing, rsyslog will automatically restart the plugin.

qiluo-msft:
Is there a sonic-mgmt testcase? #Closed

liuh-80:
> Is there a sonic-mgmt testcase?

Test case PR is here: https://github.com/sonic-net/sonic-mgmt/pull/19749

files/scripts/syslog_counter.py
```diff
@@ -0,0 +1,77 @@
...
+    logging.debug("syslog counter plugin start initialize")
+
+    # Start background thread
+    update_thread = threading.Thread(target=update_counter_db, daemon=True)
+    update_thread.start()
+
+    # plugin initialzied, send OK to rsyslog
+    print("OK", flush=True)
+
+    last_update_time = time.time()
```
saiarcot895: Is this variable used anywhere?

files/image_config/rsyslog/rsyslog.conf.j2
```diff
@@ -122,6 +123,11 @@ template(name="WelfRemoteFormat" type="string" string="%TIMESTAMP% id=firewall t
 action(type="omfwd" Target="{{ server }}" Port="{{ port }}" Protocol="{{ proto }}" Template="{{ template }}"{{ options }})
 {% endfor %}
 
+# update rsyslog counter
+{% if syslog_counter and syslog_counter == "true" %}
+action(type="omprog" binary="/usr/bin/syslog_counter.py" output="/var/log/syslog_counter.log" confirmMessages="on")
```
saiarcot895: If there is an issue with starting this script and keeping it running, will other processing of syslogs (i.e. going to `/var/log/syslog`, other ST collection) be impacted? Does this need a separate queue in rsyslog?

saiarcot895: Does `/var/log/syslog_counter.log` need to be added to logrotate?

files/image_config/rsyslog/rsyslog.conf.j2
```diff
@@ -122,6 +123,11 @@ template(name="WelfRemoteFormat" type="string" string="%TIMESTAMP% id=firewall t
 action(type="omfwd" Target="{{ server }}" Port="{{ port }}" Protocol="{{ proto }}" Template="{{ template }}"{{ options }})
 {% endfor %}
 
+# update rsyslog counter
```
saiarcot895: For template purposes, I'd recommend moving the comment to the inside of the if-block, so that it's not rendered if the syslog counter feature is not enabled at all.

files/image_config/rsyslog/rsyslog.conf.j2
qiluo-msft: Will this new log file double the log file disk usage? Can we have this feature without  adding new log file on disk?

liuh-80: This log file is very small, only update when plugin crash, will update it to logrotate.

files/scripts/syslog_counter.py
liuh-80: Will rewrite this plugin with rust and fix this issue

files/scripts/syslog_counter.py
liuh-80: Fixed, convert to rust

files/scripts/syslog_counter.py
liuh-80: Fixed, convert to rust with AI

files/image_config/rsyslog/rsyslog.d/00-sonic.conf.j2
```diff
@@ -5,12 +5,6 @@ template(name="prog_msg" type="list") {
     constant(value="\n")
 }
 
-{% if include_system_eventd == "y" and build_reduce_image_size != "y" %}
```
qiluo-msft: >include_system_eventd [](http://example.com/codeflow?start=6&length=21)

@zbud-msft, Could you review this part? #Closed

qiluo-msft (approved):

src/syslog-counter/Makefile
```diff
@@ -0,0 +1,14 @@
+.ONESHELL:
+SHELL = /bin/bash
+.SHELLFLAGS += -e
+
+MAIN_TARGET = syslog-counter_$(SYSLOG_COUNTER_VERSION)_$(CONFIGURED_ARCH).deb
+
+$(addprefix $(DEST)/, $(MAIN_TARGET)): $(DEST)/% :
+	# build debian package with cargo deb
+	cargo install cargo-deb
+    # sonic install rust and cargo with rustup, so can't build debian package with dpkg-buildpackage
```
saiarcot895: Non-blocking comment: Technically, dpkg-buildpackage can still be used, see https://github.com/sonic-net/sonic-buildimage/pull/23688/files#diff-eea666dcf398c6487761d0d854098338fa9382640ed1cb8d884cb3eda6185ade.

src/syslog-counter/src/lib.rs
```diff
@@ -0,0 +1,134 @@
...
+            .map_err(|e| format!("Failed to hget COUNT: {:?}", e))?;
+        convert_to_u64(raw.as_ref())
+            .ok_or_else(|| "Failed to parse COUNT as u64".to_string())
+    }
+
+    pub fn set_count(&self, count: u64) -> Result<(), String> {
+        let cxx_val = convert_to_CxxString(Some(count))
+            .ok_or_else(|| "Failed to convert count to CxxString".to_string())?;
+        self.connector.hset("SYSLOG_COUNTER", "COUNT", cxx_val.as_cxx_str())
+            .map_err(|e| format!("Failed to set COUNT: {:?}", e))
```
saiarcot895: In the error path, isn't this going to set `COUNT` to `Failed to convert count to CxxString`?

src/syslog-counter/src/lib.rs
```diff
@@ -0,0 +1,134 @@
...
+impl CountersDb {
+    pub fn new(connector: Box<dyn DbConnectorTrait + Send + 'static>) -> Self {
+        Self { connector }
+    }
+
+    pub fn get_count(&self) -> Result<u64, String> {
+        let raw = self.connector.hget("SYSLOG_COUNTER", "COUNT")
+            .map_err(|e| format!("Failed to hget COUNT: {:?}", e))?;
+        convert_to_u64(raw.as_ref())
+            .ok_or_else(|| "Failed to parse COUNT as u64".to_string())
```
saiarcot895:
```suggestion
        self.connector.hget("SYSLOG_COUNTER", "COUNT")
            .map_err(|e| format!("Failed to hget COUNT: {:?}", e))?;
            .and_then(|u| convert_to_u64(u))
```

This assumes `convert_to_u64` is modified.

src/syslog-counter/src/lib.rs
```diff
@@ -0,0 +1,134 @@
...
+pub fn convert_to_CxxString(opt: Option<u64>) -> Option<CxxString> {
+    opt.map(|num| CxxString::new(num.to_string()))
+}
+
+pub fn convert_to_u64(opt: Option<&CxxString>) -> Option<u64> {
+    opt.and_then(|s| match s.to_str() {
+        Ok(text) => text.parse::<u64>().ok(),
+        Err(_) => None,
+    })
+}
```
saiarcot895: To me, it doesn't make sense for these conversion functions to take in `Option<T>`, considering for these functions to do any work, that argument is required. I'd recommend having them take in the actual value, and then either return the actual target type, or `Result<T, E>` if the conversion could fail.

@r12f for his thoughts on this.

liuh-80: I will try fix it.

src/syslog-counter/Makefile
liuh-80: Fixed

src/syslog-counter/src/lib.rs
liuh-80: Fixed, convert_to_CxxString changed to return CxxString

src/syslog-counter/src/lib.rs
liuh-80: Fixed, code is different with suggestion because convert_to_u64 also changed

src/syslog-counter/src/lib.rs
liuh-80: Fixed

src/syslog-counter/src/lib.rs
```diff
@@ -0,0 +1,134 @@
...
+
+pub fn convert_to_u64(str: CxxString) -> Result<u64, String> {
+    match str.to_str() {
+        Ok(text) => text.parse::<u64>().map_err(|e| format!("Parse error: {}", e)),
+        Err(_) => Err("Failed to convert CxxString to &str".to_string()),
+    }
+}
+
+pub trait DbConnectorTrait: Send  {
```
yue-fred-gao, line 20: If I understand correctly, DbConnectorTrait is introduced to add Send so you can pass dbconnector to another thread/task. Alternatively, can you pass db_name, is_tcp_conn and timeout to the new thread and create dbconnector from there?

src/syslog-counter/src/lib.rs
```diff
@@ -0,0 +1,134 @@
...
+
+    println!("OK");
+
+    let stdin = io::stdin();
+    for line in stdin.lock().lines() {
+        match line {
+            Ok(_) => {
+                let mut count = counter.lock().unwrap();
+                *count += 1;
+                println!("OK");
```
yue-fred-gao, line 125: It seems too much to print ok for every log message received

src/syslog-counter/src/lib.rs
liuh-80: DbConnectorTrait is add for test case, so I can create a mock DbConnector and test CountersDb.

src/syslog-counter/src/lib.rs
liuh-80: For rsyslog plugin, every log message need reply an OK via stdout.

src/syslog-counter/debian/syslog-counter/DEBIAN/control
saiarcot895, line 1: Please remove the `DEBIAN` folder, these are generated files.

saiarcot895: Never mind, I saw it was removed later.

src/syslog-counter/src/lib.rs
yue-fred-gao: I see. Should we add the trait to swss-common? I think it will be useful for other applications

files/image_config/rsyslog/rsyslog.conf.j2
```diff
@@ -122,6 +123,11 @@ template(name="WelfRemoteFormat" type="string" string="%TIMESTAMP% id=firewall t
 action(type="omfwd" Target="{{ server }}" Port="{{ port }}" Protocol="{{ proto }}" Template="{{ template }}"{{ options }})
 {% endfor %}
 
+{% if syslog_counter and syslog_counter == "true" %}
+# update rsyslog counter
+action(type="omprog" binary="/usr/bin/syslog-counter" output="/var/log/syslog_counter.log" confirmMessages="on")
```
zbud-msft: Can you look into if using signalOnClose is helpful here? It can be helpful to ensure that if rsyslog parent process is killed or restarted that spawned child process like syslog-counter is also stopped, ensuring no memory leak or infinite pipe issue.

src/syslog-counter/src/lib.rs
liuh-80: That need add in another PR, will check how to improve sonic-swss-common for unit test, maybe add mockall support to that lib is better

files/image_config/rsyslog/rsyslog.conf.j2
liuh-80: Fixed
Enabling signalOnClose allows rsyslog to terminate unresponsive plugins automatically.

zbud-msft (approved):
can you confirm in e2e testing that there is no rsyslog service impact when 

1) $Modload omprog is included but there are no omprog actions specified when eventd is false and when syslog_counter is also false?
2) Both omprog actions are added/enabled as well as under a heavy syslog storm

liuh-80:
> can you confirm in e2e testing that there is no rsyslog service impact when
> 
> 1. $Modload omprog is included but there are no omprog actions specified when eventd is false and when syslog_counter is also false?
> 2. Both omprog actions are added/enabled as well as under a heavy syslog storm

This will use 6MB memory, no other performance impact, improved to not load omprog when feature disabled.

qiluo-msft (dismissed):

qiluo-msft (approved):

PR #23217, opened 2025-07-04
Check if fpmsyncd crash caused by ZMQ change.

liuh-80:


liuh-80:
pfcwd/test_pfcwd_warm_reboot.py passed in test result.
There are 2 test failed because ntp not installed to ptf host, which is not related with sonic-swss issue.

PR #23245, opened 2025-07-07
[FRR]: Fix enum value for FPM_ROUTE_ENCAP_SRV6_PIC_ID

GaladrielZhao:
#### Why I did it
In dplane_fpm_sonic.c, enum value 2 is wrongly assigned to both FPM_ROUTE_ENCAP_SRV6_ENCAP_SRC_ADDR and FPM_ROUTE_ENCAP_SRV6_PIC_ID.

#### How I did it
Let's fix FPM_ROUTE_ENCAP_SRV6_PIC_ID value as 3.

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [x] 202505

image


LARLSN (approved):
lgtm

cscarpitta (approved):
LGTM Thanks for fixing this issue.

BYGX-wcr (approved):
LGTM

src/sonic-frr/dplane_fpm_sonic/dplane_fpm_sonic.c
```diff
@@ -126,7 +126,7 @@ enum custom_rtattr_encap_srv6 {
 	FPM_ROUTE_ENCAP_SRV6_ENCAP_UNSPEC		= 0,
 	FPM_ROUTE_ENCAP_SRV6_VPN_SID			= 1,
 	FPM_ROUTE_ENCAP_SRV6_ENCAP_SRC_ADDR		= 2,
-	FPM_ROUTE_ENCAP_SRV6_PIC_ID			= 2,
+	FPM_ROUTE_ENCAP_SRV6_PIC_ID			= 3,
```
kperumalbfn, line 129: @GaladrielZhao Do we need to set explicit enum values if it has the default values from 0 and sequential?

GaladrielZhao: > @GaladrielZhao Do we need to set explicit enum values if it has the default values from 0 and sequential?

Thanks for your comment. You're right that we don't need to set the explicit values. However, since other enum definitions in this file also explicitly set their values, it might make sense to keep the style consistent. And this way the changes would be minimal and the code is easier to read.

kperumalbfn (approved):

Issue #21590, opened 2025-01-31
[SmartSwitch] Gnmi resets due to memory exceeding threshold when scaled DASH config is applied


vivekrnv:
- Apply big config via gnmi
- Eg: Apply 5M DASH_ROUTE_TABLE in chunks of 20K

while pushing dash_config to gnmi , the gnmi exceed memory threshold and is reset by monit. Config wont be applied as expected in this case. 

`2024 Dec 19 09:46:24.706341 smartswitch ERR monit[2176]: 'container_memory_gnmi' status failed (3) -- [gnmi]: Memory usage (588218368 Bytes) is larger than the threshold (419430400 Bytes)!`

**syslog:**
```
2024 Dec 19 09:46:24.567180 smartswitch NOTICE gnmi#telemetry: :- dbUpdateThread: dbUpdateThread for table: DASH_ROUTE_TABLE is exiting
2024 Dec 19 09:46:24.706341 smartswitch ERR monit[2176]: 'container_memory_gnmi' status failed (3) -- [gnmi]: Memory usage (588218368 Bytes) is larger than the threshold (419430400 Bytes)!
2024 Dec 19 09:46:24.706617 smartswitch INFO monit[2176]: 'container_memory_gnmi' exec: '/usr/bin/restart_service gnmi'
2024 Dec 19 09:46:24.778007 smartswitch INFO restart_service: Resetting failed status of service 'gnmi' ...
2024 Dec 19 09:46:24.809153 smartswitch INFO restart_service: Succeeded to reset failed status of service 'gnmi.service'.
2024 Dec 19 09:46:24.809302 smartswitch INFO restart_service: Restarting service 'gnmi' ...
2024 Dec 19 09:46:24.874363 smartswitch INFO systemd[1]: Stopping gnmi.service - GNMI container...
2024 Dec 19 09:46:24.894671 smartswitch NOTICE admin: Stopping gnmi service...
2024 Dec 19 09:46:24.971303 smartswitch NOTICE gnmi#telemetry: :- dbUpdateThread: dbUpdateThread begin
2024 Dec 19 09:46:24.990763 smartswitch NOTICE admin: Warm boot flag: gnmi false.
2024 Dec 19 09:46:24.997049 smartswitch NOTICE admin: Fast boot flag: gnmi false.
[...]

PR #23297, opened 2025-07-10
Remove SMARTSWITCH build flag across platforms.

rameshraghupathy:
Remove SMARTSWITCH build flag across platforms.  Include dhcp_server in the build for all platforms.  The dhcp_server is enabled only if the config exists in config_db.json  which is the case in SmartSwitch

#### Why I did it
There is a new request to have a single image across platforms and avoid a separate image for SmartSwitch.  Originally this was done to exclude the dhcp_server from the build for non-smartswitch platforms as it would increase the image size. Now that the benefit of having a single image is more compared to the image size. So making this change.

##### Work item tracking

#### How I did it
Unconditionally including of DHCP_SERVER in the build

#### How to verify it
Build an image for any platform and should build and run as it was prior to the change.  Run the same image on a smartswitch platform and it should work seamlessly.

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ x] 202405
- [x ] 202411
- [ x] 202505
- [ x] 202506

#### Description
Please refer to #21590   which also has relevance to this change

### DIFF
--- a/src/sonic-bgpcfgd/bgpcfgd/main.py
+++ b/src/sonic-bgpcfgd/bgpcfgd/main.py
@@ -13,8 +13,9 @@ from .config import ConfigMgr
 from .directory import Directory
 from .log import log_notice, log_crit
 from .managers_advertise_rt import AdvertiseRouteMgr
+from .managers_aggregate_address import AggregateAddressMgr, BGP_AGGREGATE_ADDRESS_TABLE_NAME
 from .managers_allow_list import BGPAllowListMgr
-from .managers_bbr import BBRMgr
+from .managers_bbr import BBRMgr, BGP_BBR_TABLE_NAME
 from .managers_bgp import BGPPeerMgrBase
 from .managers_db import BGPDataBaseMgr
 from .managers_intf import InterfaceMgr
@@ -34,6 +35,7 @@ from .utils import read_constants
 from .frr import FRR
 from .vars import g_debug
 
+
 def do_work():
     """ Main function """
     st_rt_timer = StaticRouteTimer()
@@ -47,6 +49,7 @@ def do_work():
         'cfg_mgr':   ConfigMgr(frr),
         'tf':        TemplateFabric(),
         'constants': read_constants(),
+        'state_db_conn': swsscommon.DBConnector("STATE_DB", 0)
     }
     managers = [
         # Config DB managers
@@ -71,7 +74,7 @@ def do_work():
         # AllowList Managers
         BGPAllowListMgr(common_objs, "CONFIG_DB", "BGP_ALLOWED_PREFIXES"),
         # BBR Manager
-        BBRMgr(common_objs, "CONFIG_DB", "BGP_BBR"),
+        BBRMgr(common_objs, "CONFIG_DB", BGP_BBR_TABLE_NAME),
         # Static Route Managers
         StaticRouteMgr(common_objs, "CONFIG_DB", "STATIC_ROUTE"),
         StaticRouteMgr(common_objs, "APPL_DB", "STATIC_ROUTE"),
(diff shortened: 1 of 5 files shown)


prabhataravind:
@ganglyu @vivekrnv @prsunny @oleksandrivantsiv for viz

vivekkarri:
LGTM

prabhataravind:
@rameshraghupathy we need to align platform/checkout/cisco-8000.ini and cisco-8000-smartswitch.ini as well. Currently both of them are pointing to different platform versions. Without that we cannot have a single image across the board.

prabhataravind:
@ganglyu for viz

rameshraghupathy:
> @rameshraghupathy we need to align platform/checkout/cisco-8000.ini and cisco-8000-smartswitch.ini as well. Currently both of them are pointing to different platform versions. Without that we cannot have a single image across the board.

@prabhataravind   Done

rameshraghupathy:
> @rameshraghupathy we need to align platform/checkout/cisco-8000.ini and cisco-8000-smartswitch.ini as well. Currently both of them are pointing to different platform versions. Without that we cannot have a single image across the board.

Done

rules/config
```diff
@@ -147,13 +144,8 @@ INCLUDE_NAT = y
...
 
-# INCLUDE_DHCP_SERVER - build and install dhcp-server package
-INCLUDE_DHCP_SERVER ?= n
-
-# Conditionally enable DHCP_SERVER when SMARTSWITCH is set to 1
-ifeq ($(SMARTSWITCH), 1)
-    INCLUDE_DHCP_SERVER = y
-endif
+# INCLUDE_DHCP_SERVER - build dhcp-server package and enable based on config_db.json
+INCLUDE_DHCP_SERVER = y
```
prabhataravind: Can we not set this globally? At this time, we want to include dhcp server only for mellanox and cisco images.

prabhataravind: @saiarcot895 do you have any comments on this? I am thinking that we don't want to package dhcp server in the image for all platforms, especially the ones where memory/disk is a concern? @prsunny for viz.

.azure-pipelines/azure-pipelines-build.yml
```diff
@@ -183,7 +183,7 @@ jobs:
               mv target/sonic-$(GROUP_NAME).bin target/sonic-$(GROUP_NAME)-dbg.bin
             fi
             if [ $(ss_image) == yes ]; then
-              make $BUILD_OPTIONS SMARTSWITCH=1 INCLUDE_DHCP_SERVER=y target/sonic-$(GROUP_NAME).bin
+              make $BUILD_OPTIONS target/sonic-$(GROUP_NAME).bin
```
liushilongbuaa: Please remove this if expression. It is duplicated with Line212

rules/config
liushilongbuaa: INCLUDE_DHCP_SERVER ?= y
We can override this options in pipeline build.

prabhataravind:
@liushilongbuaa could you help with the ms_conflict?

liushilongbuaa:
Please use '?=' instead of '='.
?= allows us to  override it in build options.

liushilongbuaa (approved):

prabhataravind:
hi @rameshraghupathy could you also remove ss_image variable entirely from line 49 and 118 as well? @vivekrnv, @oleksandrivantsiv can one of you sign-off from Nvidia side?

rameshraghupathy:
> hi @rameshraghupathy could you also remove ss_image variable entirely from line 49 and 118 as well? @vivekrnv, @oleksandrivantsiv can one of you sign-off from Nvidia side?

@prabhataravind Removed ss_image

prabhataravind:
@rameshraghupathy could you fix conflicts?

prabhataravind:
@rameshraghupathy could you pls fix conflicts?

PR #23299, opened 2025-07-10
fix auditd to periodically check and apply configuration changes

maipbui:
#### Why I did it
Currently, auditd only checks and updates the configuration once. This PR modifies it to perform the check and update periodically.
##### Work item tracking

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

- [ ] 
- [ ]

### DIFF
--- a/dockers/docker-auditd/config_checker.py
+++ b/dockers/docker-auditd/config_checker.py
@@ -155,4 +155,4 @@ def main():
 if __name__ == "__main__":
     while True:
         main()
-        time.sleep(6000)
+        time.sleep(900)
--- a/dockers/docker-auditd/config_checker.py
+++ b/dockers/docker-auditd/config_checker.py
@@ -1,7 +1,6 @@
 #!/usr/bin/env python3
 
 import json
-import time
 import subprocess
 import sys
 from sonic_py_common import logger as log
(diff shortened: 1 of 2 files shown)


dockers/docker-auditd/config_checker.py
```diff
@@ -152,4 +153,6 @@ def main():
 
 
 if __name__ == "__main__":
-    main()
+    while True:
+        main()
+        time.sleep(6000)
```
qiluo-msft: Since this script is inside a container, it fits a container ENV, and your code just use as environment variable. The benefit is easy to change by container start parameters. You can still give it a default value if not set in container start parameters. #Pending

maipbui: no longer use loop in config_checker script

qiluo-msft: Could you remove magic number and follow above suggestion in another PR?

qiluo-msft (approved):

dockers/docker-auditd/config_checker.py
maipbui: Sure, will address this comment in another PR

PR #23373, opened 2025-07-17
Add EVPN MH Yang Config Models

tbgowda:
##### Work item tracking

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

- [ ] 
- [ ]

### DIFF
--- a/src/sonic-yang-models/yang-models/sonic-evpn.yang
+++ b/src/sonic-yang-models/yang-models/sonic-evpn.yang
@@ -1,0 +1,146 @@
+module sonic-evpn {
+    yang-version 1.1;
+
+    namespace "http://github.com/sonic-net/sonic-evpn";
+    prefix sevpn;
+
+    import sonic-interface {
+        prefix sinterface;
+    }
+
+    organization
+        "SONiC";
+
+    contact
+        "SONiC";
+
+    description
+        "SONIC EVPN";
+
+    revision 2024-05-08 {
+        description
+            "First revision.";
+    }
+
+    revision 2024-06-14 {
+        description
+            "Add global EVPN Multihoming configuration.";
+    }
+
+    typedef esi-type {
+        type enumeration {
+            enum TYPE_0_OPERATOR_CONFIGURED {
+                value 0;
+                description "Type 0 ESI - Operator Configured";
+            }
+            enum TYPE_3_MAC_BASED {
+                value 3;
+                description "Type 3 ESI - MAC-Based ESI";
+            }
+        }
+    }
+
+    container sonic-evpn {
+
+        container EVPN_ETHERNET_SEGMENT {
+            description "CONFIG_DB EVPN_ETHERNET_SEGMENT Table";
+
+            list EVPN_ETHERNET_SEGMENT_LIST {
+
+                key "ifname";
+
+                /*
+                 * It's not possible to have a leafref pointing to two separate models
+                 * (INTERFACE and PORTCHANNEL), so a raw interface name is used instead.
+                 */
+                leaf ifname {
+                    type string {
+                        length 1..128;
+                    }
+                }
+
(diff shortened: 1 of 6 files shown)


src/sonic-yang-models/yang-models/sonic-evpn.yang
```diff
@@ -0,0 +1,146 @@
...
+
+        container EVPN_ETHERNET_SEGMENT {
+            description "CONFIG_DB EVPN_ETHERNET_SEGMENT Table";
+
+            list EVPN_ETHERNET_SEGMENT_LIST {
+
+                key "ifname";
+
+                /*
+                 * It's not possible to have a leafref pointing to two separate models
```
wen587: https://github.com/sonic-net/sonic-buildimage/blob/6f110acf2b9cd18c93ada95cae24ea367791bcb9/src/sonic-yang-models/yang-models/sonic-system-tacacs.yang#L149-L156

how about union to point to both?

src/sonic-yang-models/yang-models/sonic-portchannel.yang
```diff
@@ -120,6 +124,11 @@ module sonic-portchannel {
                     type stypes:boolean_type;
                 }
 
+                leaf system_mac {
```
wen587, line 127: Can we add one example in sample_config_db.json with this field?

tbgowda: done.

src/sonic-yang-models/yang-models/sonic-evpn.yang
tbgowda: Thank you for the suggestion. I have made this change.

ganglyu (approved):

tbgowda:
Hi @wen587 , could you please review the changes again? I have addressed your comments.

wen587 (approved):

tbgowda:
Hi @qiluo-msft , could you please review as well? thanks.

src/sonic-yang-models/yang-models/sonic-evpn.yang
```diff
@@ -0,0 +1,151 @@
...
+                            error-app-tag esi-invalid;
+                        }
+                    }
+                }
+
+                leaf type {
+                    type esi-type;
+                }
+
+                leaf df_pref {
```
selva-nexthop, line 85: is it better to have df-election container to select the supported df election algorithms.  I think we support both default and priority based.  df_pref can be one of those.

The HLD mentions  "modulo based scheme - (default)"

src/sonic-yang-models/yang-models/sonic-evpn.yang
```diff
@@ -0,0 +1,151 @@
...
+                    type uint16 {
+                        range "1..65535" {
+                            error-message "DF Preference value out of range";
+                            error-app-tag df-pref-invalid;
+                        }
+                    }
+                    default 32767;
+                }
```
selva-nexthop, line 93: Please check other parameters for this algo like revertive, wait-time and etc.

src/sonic-yang-models/yang-models/sonic-evpn.yang
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    container sonic-evpn {
+
+        container EVPN_ETHERNET_SEGMENT {
+            description "CONFIG_DB EVPN_ETHERNET_SEGMENT Table";
+
+            list EVPN_ETHERNET_SEGMENT_LIST {
+
+                key "ifname";
```
selva-nexthop, line 54: suggestion: we can use grouping for repeated leaf sets.

PR #22199, opened 2025-04-01
Add missing dbg / dbgsym pkgs to rules files

Ryangwaite:
#### Why I did it
A number of dbg/dbsym packages are missing from the rules files and the relevant containers, therefore debugging is harder.

By adding these missing packages to the rules files, they get copied to the target/ directory and installed into the debug containers.

##### Work item tracking
- Microsoft ADO **(number only)**: 30052188

#### How I did it
Found all dbg/dbgsym debs below src/ that weren't copied to target/ and created rules to copy them

#### How to verify it
Build the image and the dbg/dbgsym packages will be included in the debug containers.

#### Which release branch to backport (provide reason below if selected)

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

#### Tested branch (Please provide the tested image version)

- [ ] 
- [ ] 

#### Description for the changelog

Include missing dbg/dbsym debs in debug containers.

PR #23378, opened 2025-07-18
Increase the installer disk size by 1GB and switch to ext2 for KVM

saiarcot895:
#### Why I did it

Following #22199, the debug version of `sonic-vs.bin` file is at 4093MB (to be precise, one particular build was at 4291790831 bytes). This, in theory, should've fit fine in the 4096MB FAT32 installer disk that is created for building the sonic-vs.img QCOW2 image. However, it's short by 5MB. My guess for this reason is because FAT32 itself needs some space for metadata, and so we don't get the full 4096MB. This results in a failure to build the debug KVM image.

##### Work item tracking

#### How I did it

As a fix, increase the installer disk size that we create by 1GB, which should be enough for now, and switch to ext2 filesystem instead of FAT32. This switch is needed because FAT32 only allows files up to 4GB (minus one byte), and the latest debug images are at around 4.1GB.

#### How to verify it

I locally made a loopback file of 4104MB, copied the sonic-vs.bin debug installer file generated in https://dev.azure.com/mssonic/build/_build/results?buildId=900622&view=results into it (which is 4093MB), and verified that it did get copied into there.

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

- [ ] 
- [ ]

### DIFF
--- a/scripts/build_kvm_image.sh
+++ b/scripts/build_kvm_image.sh
@@ -36,9 +36,9 @@ create_disk()
 
 prepare_installer_disk()
 {
-    fallocate -l 4104M $INSTALLER_DISK
+    fallocate -l 5120M $INSTALLER_DISK
 
-    mkfs.vfat $INSTALLER_DISK
+    mkfs.ext2 $INSTALLER_DISK
 
     tmpdir=$(mktemp -d)
 
--- a/scripts/build_kvm_image.sh
+++ b/scripts/build_kvm_image.sh
@@ -36,7 +36,7 @@ create_disk()
 
 prepare_installer_disk()
 {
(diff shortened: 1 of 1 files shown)


saiarcot895:
...and I just saw that the debug installer file in https://dev.azure.com/mssonic/build/_build/results?buildId=902172&view=results is at 4115MB. Looking at the diff between the prior build and this build, I don't see an obvious cause for the increase in size, but this might mean that the change to ext2/ext4 is needed now.

qiluo-msft (approved):

PR #23402, opened 2025-07-21
Use Rust docker client to replace python docker wait to save memory

qiluo-msft:
#### Why I did it
TBD

##### Work item tracking

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

- [ ] 
- [ ]


src/sonic-ctrmgrd-rs/crates/docker-rs/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "docker-rs"
```
saiarcot895, line 2: Non-blocking comment: It may make sense in the future to move this up to the "parent", if there aren't going to be any other components here.

saiarcot895 (approved):

src/sonic-ctrmgrd-rs/crates/docker-rs/Cargo.toml
qiluo-msft: I anticipate there will be multiple apps in sonic-ctrmgrd-rs, so multiple crates.

r12f:
omg!!! the long waiting updating!!! love it!!

rules/sonic-ctrmgrd-rs.dep
```diff
@@ -0,0 +1,10 @@
+SPATH       := $($(SONIC_CTRMGRD_RS)_SRC_PATH)
+DEP_FILES   := $(SONIC_COMMON_FILES_LIST) rules/sonic-ctrmgrd-rs.mk rules/sonic-ctrmgrd-rs.dep
+DEP_FILES   += $(SONIC_COMMON_BASE_FILES_LIST)
+SMDEP_FILES := $(addprefix $(SPATH)/,$(shell cd $(SPATH) && git ls-files))
```
k-v1: smdep?

src/sonic-ctrmgrd-rs/Makefile
```diff
@@ -0,0 +1,78 @@
...
+# Top level targets
+#
+all: format lint build test
+release: format lint build-release test-release
+
+#
+# Install dependencies
+#
+install-deps:
+	sudo apt install -y protobuf-compiler libprotobuf-dev
```
k-v1: Probably wrong deps.

src/sonic-ctrmgrd-rs/debian/control
```diff
@@ -0,0 +1,11 @@
+Source: sonic
+Maintainer: SONiC Team <sonic-dev@lists.sonicfoundation.dev>
+Section: net
+Priority: optional
+Build-Depends: dh-exec (>=0.3), debhelper (>= 12), autotools-dev
```
k-v1, line 5: autotools?

src/sonic-ctrmgrd-rs/Cargo.toml
```diff
@@ -0,0 +1,82 @@
...
+serde = { version = "1", features = ["derive", "rc"] }
+serde_json = "1"
+serde_yaml = "0.9"
+serde_with = "3.12"
+
+# Command line utils
+clap = { version = "4", features = ["derive", "cargo", "wrap_help", "unicode", "string", "unstable-styles"] }
+color-eyre = "0.6"
+
+# gRPC
```
k-v1: Probably wrong deps.

saiarcot895: These files were ported over from [another repo](https://github.com/sonic-net/sonic-dash-ha/blob/master/Cargo.toml) that was building more applications than just this one, so many of the dependencies here aren't relevant to this one application, and some cleanup can be done. But in the short term, it shouldn't cause any negative impact, except for slightly increased build times and slightly larger binary.

src/sonic-ctrmgrd-rs/debian/control
qiluo-msft: autotools-dev is the correct Build-Depends. Could you double check?

rules/sonic-ctrmgrd-rs.dep
qiluo-msft: Could you clarify? or suggest new code?

k-v1: @qiluo-msft 
I mean we use SMDEP_FILES and SMDEP_PATHS for submodules, but SONIC_CTRMGRD_RS is not a submodule.
As I understand, this file should be 

```
SPATH       := $($(SONIC_CTRMGRD_RS)_SRC_PATH)
DEP_FILES   := $(SONIC_COMMON_FILES_LIST) rules/sonic-ctrmgrd-rs.mk rules/sonic-ctrmgrd-rs.dep
DEP_FILES   += $(SONIC_COMMON_BASE_FILES_LIST)
DEP_FILES   += $(shell git ls-files $(SPATH))

$(SONIC_CTRMGRD_RS)_CACHE_MODE  := GIT_CONTENT_SHA 
$(SONIC_CTRMGRD_RS)_DEP_FLAGS   := $(SONIC_COMMON_FLAGS_LIST)
$(SONIC_CTRMGRD_RS)_DEP_FILES   := $(DEP_FILES)
```

qiluo-msft: thanks! fixed.

k-v1:
Maybe it's also necessary to add a .gitignore to src/sonic-ctrmgrd-rs for the target directory, but I'm not sure. Other Rust packages like sonic-nettools and sonic-dash-ha have .gitignore files.

PR #23446, opened 2025-07-23
Use archive.debian.org for bullseye-backports

jmlu-arista:
#### Why I did it

bullseye-backports is end of life and has been removed from the standard mirrors recently. This causes the bullseye sonic-slave Docker image to fail to build.

Fixes https://github.com/sonic-net/sonic-buildimage/issues/23430

##### Work item tracking

#### How I did it

Update the sources.list.j2 template to use archive.debian.org for bullseye-backports

#### How to verify it

Check that `BLDENV=bullseye make -f Makefile.work sonic-slave-bash` works with these changes

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [x] 202405
- [x] 202411
- [x] 202505

This issue affects every branch using bullseye, as the upstream mirrors have removed bullseye-backports

- [x] 202405
- [x] master

The patch does not apply cleanly on releases older than 202405

Use archive.debian.org for bullseye-backports


jusherma:
We're hitting this issue as well. @lguohan any ETA on when this will be approved/merged? We will definitely need it in 202405 as well

GaladrielZhao (approved):
Thanks for updating the source. I also ran into this problem today, and this solution works for me :D

PR #23466, opened 2025-07-24
[DASH] Add support to enable Metering counters

sundar-pds:
What I did
This PR enables Sonic-buildimage support for DASH Metering counters.

Related to the following PR's:
https://github.com/sonic-net/sonic-swss/pull/3631

Why I did it
To enable support for metering counters .


KrisNey-MSFT:
@prsunny for a Reviewer please

PR #23497, opened 2025-07-28
Verify gnmi container

ganglyu:
##### Work item tracking

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

- [ ] 
- [ ]


rules/docker-gnmi.mk
```diff
@@ -34,6 +35,7 @@ $(DOCKER_GNMI)_RUN_OPT += -t
 $(DOCKER_GNMI)_RUN_OPT += -v /etc/sonic:/etc/sonic:ro
 $(DOCKER_GNMI)_RUN_OPT += -v /etc/localtime:/etc/localtime:ro
 $(DOCKER_GNMI)_RUN_OPT += -v /var/run/dbus:/var/run/dbus:rw
+$(DOCKER_GNMI)_RUN_OPT += --privileged --pid=host
```
vaibhavhd, line 38: @qiluo-msft , can you sign off on this `privileged` flag being added to the gnmi container?

This is needed for several reasons - including that in this PR, GCU may require to do restart of host services, or services outside of gnmi. Secondly, gnoi APIs are also located within gnmi, and they also need to talk to host services, and do rw operations.

CC - @hdwhdw

rules/docker-gnmi.mk
```diff
@@ -9,6 +9,7 @@ $(DOCKER_GNMI)_PATH = $(DOCKERS_PATH)/$(DOCKER_GNMI_STEM)
 $(DOCKER_GNMI)_DEPENDS += $(SONIC_MGMT_COMMON)
 $(DOCKER_GNMI)_DEPENDS += $(SONIC_TELEMETRY)
 $(DOCKER_GNMI)_DBG_DEPENDS = $($(DOCKER_CONFIG_ENGINE_BOOKWORM)_DBG_DEPENDS)
+$(DOCKER_GNMI)_PYTHON_WHEELS = $(SONIC_PLATFORM_COMMON_PY3) $(SONIC_UTILITIES_PY3)
```
vaibhavhd, line 12: @wen587 , FYI:

I don't think that it is a good idea to import/duplicate the sonic-utlitities packages as a whole within gnmi. If our goal is to move GCU to GNMI then we should just import GCU specific files to docker-gnmi, rather than all of sonic-utilities package.

What do you think? Gang suggested that it may be better, but the challenge is that we need to build GCU package separately. I think that's a good investment.

PR #23500, opened 2025-07-28
[frr]: Force disable next hop group support

lipxu:
[202411][frr]: Force disable next hop group support

#### Why I did it
Netscan loss observed on production devices

##### Work item tracking
33406776

#### How I did it
Force disable next hop group support

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

- [ ] 
- [ ]

### DIFF
--- a/dockers/docker-fpm-frr/docker_init.sh
+++ b/dockers/docker-fpm-frr/docker_init.sh
@@ -49,6 +49,10 @@ write_default_zebra_config()
         echo "no fpm use-next-hop-groups" >> $FILE_NAME
         echo "fpm address 127.0.0.1" >> $FILE_NAME
     }
+
+    grep -q '^no zebra nexthop kernel enable' $FILE_NAME || {
+        echo "no zebra nexthop kernel enable" >> $FILE_NAME
+    }
 }
 
 if [[ ! -z "$NAMESPACE_ID" ]]; then
--- a/src/sonic-bgpcfgd/tests/data/sonic-cfggen/zebra/zebra.conf
+++ b/src/sonic-bgpcfgd/tests/data/sonic-cfggen/zebra/zebra.conf
@@ -4,6 +4,8 @@
 ! file: zebra.conf
 !
 !
+! Force disable next hop group support
+no zebra nexthop kernel enable
 ! Uses the old known FPM behavior of including next hop information in the route (e.g. RTM_NEWROUTE) messages
 no fpm use-next-hop-groups
 !
--- a/dockers/docker-fpm-frr/frr/zebra/zebra.conf.j2
+++ b/dockers/docker-fpm-frr/frr/zebra/zebra.conf.j2
@@ -7,6 +7,8 @@
 {% endblock banner %}
 !
 {% block fpm %}
+! Force disable next hop group support
+no zebra nexthop kernel enable
 {% if ( ('localhost' in DEVICE_METADATA) and ('nexthop_group' in  DEVICE_METADATA['localhost']) and
         (DEVICE_METADATA['localhost']['nexthop_group'] == 'enabled') ) %}
 ! enable next hop group support
--- a/platform/vs/docker-sonic-vs/frr/zebra.conf
+++ b/platform/vs/docker-sonic-vs/frr/zebra.conf
@@ -1,5 +1,4 @@
+no zebra nexthop kernel enable
 no fpm use-next-hop-groups
-
 fpm address 127.0.0.1
 zebra nexthop-group keep 1
-
--- a/src/sonic-config-engine/tests/sample_output/py2/t2-chassis-fe-vni-zebra.conf
+++ b/src/sonic-config-engine/tests/sample_output/py2/t2-chassis-fe-vni-zebra.conf
@@ -4,6 +4,8 @@
 ! file: zebra.conf
 !
 !
+! Force disable next hop group support
+no zebra nexthop kernel enable
 ! Uses the old known FPM behavior of including next hop information in the route (e.g. RTM_NEWROUTE) messages
 no fpm use-next-hop-groups
 !
--- a/src/sonic-config-engine/tests/sample_output/py2/t2-chassis-fe-zebra.conf
+++ b/src/sonic-config-engine/tests/sample_output/py2/t2-chassis-fe-zebra.conf
@@ -4,6 +4,8 @@
 ! file: zebra.conf
 !
 !
+! Force disable next hop group support
+no zebra nexthop kernel enable
 ! Uses the old known FPM behavior of including next hop information in the route (e.g. RTM_NEWROUTE) messages
 no fpm use-next-hop-groups
 !
--- a/src/sonic-config-engine/tests/sample_output/py2/zebra_frr.conf
+++ b/src/sonic-config-engine/tests/sample_output/py2/zebra_frr.conf
@@ -4,6 +4,8 @@
 ! file: zebra.conf
 !
 !
+! Force disable next hop group support
+no zebra nexthop kernel enable
 ! Uses the old known FPM behavior of including next hop information in the route (e.g. RTM_NEWROUTE) messages
 no fpm use-next-hop-groups
 !
--- a/src/sonic-config-engine/tests/sample_output/py2/zebra_frr_dualtor.conf
+++ b/src/sonic-config-engine/tests/sample_output/py2/zebra_frr_dualtor.conf
@@ -4,6 +4,8 @@
 ! file: zebra.conf
 !
 !
+! Force disable next hop group support
+no zebra nexthop kernel enable
 ! Uses the old known FPM behavior of including next hop information in the route (e.g. RTM_NEWROUTE) messages
 no fpm use-next-hop-groups
 !
--- a/src/sonic-config-engine/tests/sample_output/py3/t2-chassis-fe-vni-zebra.conf
+++ b/src/sonic-config-engine/tests/sample_output/py3/t2-chassis-fe-vni-zebra.conf
@@ -4,6 +4,8 @@
 ! file: zebra.conf
 !
 !
+! Force disable next hop group support
+no zebra nexthop kernel enable
 ! Uses the old known FPM behavior of including next hop information in the route (e.g. RTM_NEWROUTE) messages
 no fpm use-next-hop-groups
 !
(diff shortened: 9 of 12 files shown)


dgsudharsan:
Hi @lipxu Can you please clarify why this is needed in master branch? Is there any issue observed?
nexthop group support in SONiC is being currently worked on. @hasan-brcm  @eddieruan-alibaba FYI

eddieruan-alibaba:
Should we discuss this PR in this week's WG meeting to understand the motivation for this change ? 

@lipxu @hasan-brcm @dgsudharsan

src/sonic-bgpcfgd/tests/data/sonic-cfggen/zebra/zebra.conf
```diff
@@ -4,6 +4,8 @@
 ! file: zebra.conf
 !
 !
+! Force disable next hop group support
```
eddieruan-alibaba, line 7: "Force disable next hop group support" Need to understand the motivation.

lipxu: Thanks for your comments @eddieruan-alibaba , we have discussed it in the meeting. We hit the issue in production. So far we cannot determine the root cause, but we are sure it is related to this configuration.
After synced it in the meeting, we plan to add this workaround first to disable the next hop group with metadata config, thanks.

lipxu:
> Hi @lipxu Can you please clarify why this is needed in master branch? Is there any issue observed? nexthop group support in SONiC is being currently worked on. @hasan-brcm @eddieruan-alibaba FYI

Yes, @dgsudharsan , as we synced in the meeting, there is an issue in production. 
As an agreement in the meeting, we plan to use a metadata config, and init the nexthop group support based on this configuration.
Please help to review the PR, thanks a lot.

dockers/docker-fpm-frr/docker_init.sh
```diff
@@ -49,6 +49,10 @@ write_default_zebra_config()
         echo "no fpm use-next-hop-groups" >> $FILE_NAME
         echo "fpm address 127.0.0.1" >> $FILE_NAME
     }
+
+    grep -q '^no zebra nexthop kernel enable' $FILE_NAME || {
```
dgsudharsan: This logic will create problem when zebra_nexthop is enabled

src/sonic-yang-models/yang-models/sonic-device_metadata.yang
```diff
@@ -317,6 +317,15 @@ module sonic-device_metadata {
                     description "Enable syslog with OS version feature.";
                     default "false";
                 }
+
+                leaf zebra_nexthop {
```
dgsudharsan: Why can't we reuse nexthop_group that is already available in device metadata?

dgsudharsan (changes requested):

dockers/docker-fpm-frr/docker_init.sh
lipxu: Thanks for reminder @dgsudharsan , this script should only be called during initialization. and the enable->disable case should not occur in production. Please correct me if anything wrong, thanks

src/sonic-yang-models/yang-models/sonic-device_metadata.yang
lipxu: Thanks, I am not sure what the nexthop_group flag is for, does it indicate the whole feature status? Can we also use it for the zebra disable nexthop case？ thanks a lot

dgsudharsan: @eddieruan-alibaba  can clarify this but i believe it is to indicate the whole feature https://github.com/sonic-net/SONiC/pull/1425

lipxu: @dgsudharsan @eddieruan-alibaba , could you please help to confirm why we could reuse the nexthop_group , thanks a lot

dgsudharsan: @ntt-omw @eddieruan-alibaba  can you please clarify the above question

dockers/docker-fpm-frr/docker_init.sh
dgsudharsan: But what if we have the config in /etc/sonic/config_db.json
DEVICE_METADATA['localhost']['zebra_nexthop'] == 'enabled'

This logic will override will result in having both 'no zebra nexthop kernel enable' and 'zebra nexthop kernel enable' in the file. Can you please confirm?

src/sonic-yang-models/yang-models/sonic-device_metadata.yang
nakano-omw: @dgsudharsan The nexthop_group feature requires FRR’s fpm use-next-hop-groups to be enabled. For backward compatibility, this feature is disabled by default. The enable/disable control was implemented upon request from the Routing WG to allow more flexible control of the feature status.

> How to enable / disable using NHG. The conclusion is to use zebra command line arguments to enable/disable this feature. No run time change is allow. For SONiC deployment, the configuration would be driven from config db to provide proper launch commands for zebra and fpmsyncd.
https://lists.sonicfoundation.dev/g/sonic-wg-routing/wiki/34834

dockers/docker-fpm-frr/docker_init.sh
lipxu: thanks @dgsudharsan , the initial config is generated based on the metadata, so I don't think there should be a chance for both commands to appear in the config. It should behave the same way as the nexthop_group , thanks

src/sonic-yang-models/yang-models/sonic-device_metadata.yang
lipxu: Thanks @dgsudharsan @nakano-omw ,
So could I understand we still need to keep the separate metadata parameter to control the FRR command, correct? thanks
fpm use-next-hop-groups
no zebra nexthop kernel enable

dockers/docker-fpm-frr/docker_init.sh
lipxu: @dgsudharsan Could you please help to review the PR again when free, thank you very much

lipxu:
@dgsudharsan , I've updated the comments and merged the latest master。 Could you please help review it again, thanks a lot

lipxu:
> @dgsudharsan , I've updated the comments and merged the latest master。 Could you please help review it again, thanks a lot

Hi, @dgsudharsan , Just want to know if you have a chance to review the PR again, thanks a lot.

dockers/docker-fpm-frr/docker_init.sh
dgsudharsan: Just to understand, this script will run at init irrespective of the value in device metadata after the zebra.json is rendered.
This means the file before this step will have "zebra nexthop kernel enable"
Now this check will return false for the grep and the 2nd part would execute appending "no zebra nexthop kernel enable".  Can you please test this flow?

dgsudharsan (changes requested):

dockers/docker-fpm-frr/docker_init.sh
StormLiangMS: hi @lipxu kind of agree with Sud, we may consider to use only zebra.conf.j2 to generate the configuration we need other than having it in the docker_init.sh also?

lipxu: Thanks @dgsudharsan  and @StormLiangMS , You are absolutely right. I checked the history, and this code appears to be legacy from the cherry-pick “Force disable next hop group support.” The tests all passed because we didn’t have metadata now. I’ve removed this code. Thanks for your reminder.

By the way, "no fpm use-next-hop-groups" likely has the same issue and should also be removed after syncing with the owner. Thanks.

```
    grep -q '^no fpm use-next-hop-groups' $FILE_NAME || {
        echo "no fpm use-next-hop-groups" >> $FILE_NAME
        echo "fpm address 127.0.0.1" >> $FILE_NAME
    }
```

StormLiangMS:
hi @eddieruan-alibaba would you help to take a look? We'd like to have one option to unblock current issue on high density BGP sessions platform and Chassis, to buy us some time to root cause this one.

eddieruan-alibaba:
@StormLiangMS 

when no zebra nexthop kernel enable  is configured, would we still need to allow enable nexthop group for fpm? 

{% block fpm %}
{% if ( ('localhost' in DEVICE_METADATA) and ('nexthop_group' in  DEVICE_METADATA['localhost']) and
        (DEVICE_METADATA['localhost']['nexthop_group'] == 'enabled') ) %}

StormLiangMS:
hi @eddieruan-alibaba 

> @StormLiangMS
> 
> when no zebra nexthop kernel enable is configured, would we still need to allow enable nexthop group for fpm?
> 
> {% block fpm %} {% if ( ('localhost' in DEVICE_METADATA) and ('nexthop_group' in DEVICE_METADATA['localhost']) and (DEVICE_METADATA['localhost']['nexthop_group'] == 'enabled') ) %}

@eddieruan-alibaba From what nakano-omw, to have nexthop_group would allow finer granularity on the control and required from routing WG.  We'd like to keep it as separate for now, since the "no zebra kernel nexthop enable" should be removed eventually after we finalized the solution to fix the rejected routes issue.

eddieruan-alibaba:
> hi @eddieruan-alibaba
> 
> > @StormLiangMS
> > when no zebra nexthop kernel enable is configured, would we still need to allow enable nexthop group for fpm?
> > {% block fpm %} {% if ( ('localhost' in DEVICE_METADATA) and ('nexthop_group' in DEVICE_METADATA['localhost']) and (DEVICE_METADATA['localhost']['nexthop_group'] == 'enabled') ) %}
> 
> @eddieruan-alibaba From what nakano-omw, to have nexthop_group would allow finer granularity on the control and required from routing WG. We'd like to keep it as separate for now, since the "no zebra kernel nexthop enable" should be removed eventually after we finalized the solution to fix the rejected routes issue.

Sounds good. So we treat 

"no zebra kernel nexthop enable" 

as a workaround until we finalize the solution.

lipxu:
Create a PR for 202511 manually due to conflict https://github.com/sonic-net/sonic-buildimage/pull/24997

r12f:
Fixing tag as manual pick is merged.

PR #23511, opened 2025-07-29
Blacklist amd_sfh for Shearwater and Redstart in Aboot

dayouliu1:
#### Why I did it

amd_sfh is a SOC that handles human interface devices using PCIe (ex: gyroscopic sensor, etc.). This is not used on the switch and can be blacklisted.

This module is throwing error messages at boot time due to the lack of devices to manage, where modprobe is failing to probe this device. This is failing mgmt tests (SRv6 tests). Blacklisting this device will resolve this issue.

##### Work item tracking

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [x] 202505
- [x] 202503
- [x] 202412  

- [ ] 
- [ ]

### DIFF
--- a/files/Aboot/boot0.j2
+++ b/files/Aboot/boot0.j2
@@ -702,7 +702,7 @@ write_platform_specific_cmdline() {
         read_system_eeprom
     fi
     if in_array "$platform" "shearwater" "redstart"; then
-        cmdline_add modprobe.blacklist=snd_hda_intel,hdaudio
+        cmdline_add modprobe.blacklist=snd_hda_intel,hdaudio,amd_sfh
         if [ -f /tmp/.switch-prefdl ]; then
             read_switch_eeprom
         else


vvolam (commented):
@dayouliu1 Just wanted to confirm this is intentional change to avoid probing PCIe devices which are supported by amd_sfh. What are devices that are supported by shearwater or Redstart?

Staphylo:
@vvolam yes this is the intended behavior as per our investigation the pcie endpoint of `amd_sfh` ends up failing with `-EOPNOTSUPP` as part of the probe logic and ends up printing some kernel error messages.
This creates a cosmetic issue that usually raises some eyebrows on top of failing sonic-mgmt tests due to log checking.
Given that this driver is not relevant on the `shearwater` and `redstart` supervisors we are disabling it with this PR.
These 2 supervisors are only relevant for `Quicksilver` variants for now. They might apply to other products reusing the same supervisors in the future, but `amd_sfh` will most likely remain irrelevant in those cases.

vvolam (approved):

sdszhang (approved):

StormLiangMS (approved):
LGTM

PR #23563, opened 2025-08-01
Don't attempt building the user container if the main slave fails

saiarcot895:
#### Why I did it

If building the main slave container fails, then don't continue and try to build the user slave container (or try to collect the versions used in the failed slave container). This reduces the number of errors logged after the actual error message, and makes it a bit easier for reporting the actual cause of failures, besides saying that the user slave container failed to build because the main slave container wasn't available.

##### Work item tracking

#### How I did it

Set the `-o pipefail` setting in the group command used to build the slave container, which will report the command as having failed if any command in a piped command fails. The bash shell already has `-e` set; however, in a piped command, the exit of the last command in the pipe is used in determining if the command failed. For example, with `command1 | command2`, without `-o pipefail`, the exit code of `command1` is completely ignored. The docker build command is piped through `| tee ...` to save the output to a log file, which means `-o pipefail` needs to be set here.

#### How to verify it

Add a `RUN false` near the beginning of a slave container's Dockerfile to force a failure, and observe the behavior.

Before fix:

image

After fix:

image

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

- [ ] 
- [ ]

### DIFF
--- a/Makefile.work
+++ b/Makefile.work
@@ -477,6 +477,7 @@ SONIC_SLAVE_BASE_BUILD = \
 		$(DOCKER_SLAVE_BASE_PULL_REGISTRY); \
 	} || \
 	{ \
+		set -o pipefail ; \
 		echo Image $(SLAVE_BASE_IMAGE):$(SLAVE_BASE_TAG) not found. Building... ; \
 		$(PREPARE_DOCKER) ; \
 		$(DOCKER_SLAVE_BASE_BUILD) ; \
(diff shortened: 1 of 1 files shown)


Makefile.work
```diff
@@ -477,6 +477,7 @@ SONIC_SLAVE_BASE_BUILD = \
 		$(DOCKER_SLAVE_BASE_PULL_REGISTRY); \
 	} || \
 	{ \
+		set -o pipefail ; \
```
liushilongbuaa, line 480: 'ls non-exist-file | grep xxx'      It will not run grep command. This is pipefail.
```
cat a.sh 
echo HH && { set -o pipefail; echo AA; ls oieaofijefojaoij; echo BB; }
bash a.sh
HH
AA
ls: cannot access 'oieaofijefojaoij': No such file or directory
BB
```

saiarcot895: The `$(DOCKER_SLAVE_BASE_BUILD)` command uses `$(SPLIT_LOG)`, which expands to `2>&1 | tee`. The pipefail is for that specific command. Same thing for `$(DOCKER_USER_BUILD)`.

Besides the pipefail itself, `-e` is set in `.SHELLFLAGS` in this file, which gets carried through to subshells, so it'll fail on the first failing command.

PR #23571, opened 2025-08-02
[libnss-radius] Handle empty secondary groups for useradd

manoharan-nexthop:
Fixes https://github.com/sonic-net/sonic-buildimage/issues/23570

#### How I did it
When a user login is attempted, the user is created locally and then the user is processed for authentication. With the issue, the user creation fails and the successive authentication happens with wrong information. When the user was created locally and then the authentication was attempted, we dont hit the issue.

the user addition fails as the secondary group list is empty and when it is, useradd fails with invalid -G argument. Check for secondary group string validity before using the same.

#### How to verify it
Verified the user authentication is successful.

The issue exists only in `master`.

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

[libnss-radius] Handle empty secondary groups for useradd

### DIFF
--- a/src/radius/nss/libnss-radius/nss_radius_common.c
+++ b/src/radius/nss/libnss-radius/nss_radius_common.c
@@ -193,10 +193,19 @@ static int user_add(const char* name, char* gid, char* sec_grp, char* gecos,
 
     } else if(pid == 0) {
 
-        if (many_to_one)
-          execl(cmd, cmd, "-g", gid, "-G", sec_grp, "-c", gecos, "-m", "-s", shell, name, NULL);
-        else
-          execl(cmd, cmd, "-U", "-G", sec_grp, "-c", unconfirmed_user, "-d", home, "-m", "-s", shell, name, NULL);
+        if (many_to_one) {
+            if (sec_grp && sec_grp[0] != '\0') {
+                execl(cmd, cmd, "-g", gid, "-G", sec_grp, "-c", gecos, "-m", "-s", shell, name, NULL);
+            } else {
+                execl(cmd, cmd, "-g", gid, "-c", gecos, "-m", "-s", shell, name, NULL);
+            }
+        } else {
+            if (sec_grp && sec_grp[0] != '\0') {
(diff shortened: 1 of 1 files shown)


dt-nexthop:
@qiluo-msft, would you mind reviewing this when you get a chance?

qiluo-msft:
Could you add sonic-mgmt testcase?

dt-nexthop:
Yes, there are in this PR - https://github.com/sonic-net/sonic-mgmt/pull/16475 - but since the current master build in sonic has this bug, all the tests will fail on the sonic-mgmt PR.  We have run them locally, though, with this patch, and they do pass. @manoharan-nexthop, can you paste in the local results to this PR?

manoharan-nexthop:
> Yes, there are in this PR - [sonic-net/sonic-mgmt#16475](https://github.com/sonic-net/sonic-mgmt/pull/16475) - but since the current master build in sonic has this bug, all the tests will fail on the sonic-mgmt PR. We have run them locally, though, with this patch, and they do pass. @manoharan-nexthop, can you paste in the local results to this PR?

With this change, all the tests passes..

1	test_radius_rw_user	PASS
2	test_radius_ro_user	PASS
3	test_radius_command_auth	PASS
4	test_radius_fallback	PASS
5	test_radius_failed_auth	PASS
6	test_radius_source_ip	PASS

dt-nexthop:
@qiluo-msft   please give another review to this PR when you get a chance

dt-nexthop:
@Yarden-Z, not sure about TACACS, can we take that up in a different PR?

manoharan-nexthop:
> Looks good - do we have a fix for tacacs as well? Since I think the same item applies there as well

I dont see similar issue in TACACS.. The issue is happening only with RADIUS.

dt-nexthop:
@ Yarden-Z I wanted to see if there is anything left we need to do merge this one in.  I would like to get this merged in so the RADIUS test in sonic-mgmt will also start passing.

PR #23577, opened 2025-08-04
Revert "Extend router types BackendLeafRouter and BackendToRRouter"

stephenxs:
Reverts sonic-net/sonic-buildimage#23528


r12f:
hi Stephen, may I know why submitting this PR?

r12f:
hi @stephenxs , I don't think we need this change, since it will break the qos config. I am going to remove the tag now and please let me know if you think otherwise.

PR #23600, opened 2025-08-05
Fix subport of Moby backplane ports

kewei-arista:
#### Why I did it

As per the issue reported in https://github.com/aristanetworks/sonic-qual.msft/issues/594, Moby backplane ports always have the subport to be 0, but we would like them to be indexed from 0 to 12. This PR fixes various issue with existing subport computation so the correct subport can be used for these Moby backplane ports. 

##### Work item tracking

#### How I did it

- Update `portconfig` script to make sure it can use the values defined in `hwsku.json` when they are available, and compute the subports based on the new port naming so that the values can also be computed correctly in the future for the products like Moby instead of always giving 0 by default
- Update `hwsku.json` with the new subport for Moby backplane ports. Note that based on https://github.com/sonic-net/sonic-buildimage/pull/23375 and https://github.com/sonic-net/sonic-buildimage/pull/23599, we only need to update a single `hwsku.json` file for each platform
- Update yang model accordingly so that subport 9-12 can be accepted

#### How to verify it

Tested on the Moby switch in our lab to make sure the correct subports are created in config file and config reload can also preserve these correct subport configuration.

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [x] 202505
- [x] msft-202412
- [x] msft-202503

- [x] msft-202412


sdszhang (approved):
lgtm

sdszhang:
Hi @kewei-arista , the build failed with the following error. Can you check it please?

```
------------------- Test 175: Out of range subport number---------------------
ERROR    YANG-TEST:test_yang_model.py:28  Exception >Mismatch ['does not satisfy', 'range', '0', '.', '.', '1', '2'] and < in /sonic/src/sonic-yang-models/tests/yang_model_tests/test_yang_model.py:225
INFO     YANG-TEST:test_yang_model.py:228 Out of range subport number Failed
```

sdszhang:
@kewei-arista Can you fix the PR feature on priority?  Without this fix, one of the interface will be down on Moby if optics is used on the physical port. which is affecting our nightly run.

@davidm-arista @aaronpayment @r12f for viz.

kewei-arista:
@sdszhang Sorry I missed your message earlier. Just fixed the issue.

sdszhang:
@kewei-arista there is conflict with 202505 branch, can you check?

kewei-arista:
> @kewei-arista there is conflict with 202505 branch, can you check?

@sdszhang I just created a new PR https://github.com/sonic-net/sonic-buildimage/pull/23685 for 202505

StormLiangMS:
hi @yxieca would you help to take a look a this change related sub port?

r12f (approved):

DavidZagury:
@kewei-arista
You have added the sub port to OPTIONAL_HWSKU_ATTRIBUTES, and now in parse_platform_json_file when there is a subport configured for the child ports, the same subport is applied to all the child ports from this port. Which is wrong.

PR #23615, opened 2025-08-06
Add patch to teamd to exit immediately if signalling the parent fails

saiarcot895:
#### Why I did it

When teamd is starting, on weaker overloaded systems, it may take a long time for it to fully initialize. When that happens, teammgrd might get a non-zero exit code from the parent teamd process (because it hasn't gotten a success signal from the child teamd process), and start the process of restarting teamd for that port channel. Meanwhile, the child teamd process might finish initialization, send a success signal, and stay up and process other requests. This can cause issues/race conditions if other applications start their configuration because they think that teamd is up.

This is related to https://github.com/sonic-net/sonic-swss/pull/3984.

##### Work item tracking

#### How I did it

Avoid the above issue by exiting immediately if it fails to signal the parent process that it has completed initialization. This suggests that it took too long and a failure signal to teammgrd was already sent before this point.

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

- [ ] 
- [ ]


saiarcot895:
@judyjoseph Please review.

judyjoseph (dismissed):
LGTM

bingwang-ms:
⚠️ **This review was performed by an AI agent** on behalf of @bingwang-ms. Please treat this as a preliminary review and verify independently.

---

## PR Review: Add patch to teamd to exit immediately if signalling the parent fails

### Summary
The fix is directionally correct and directly addresses the race condition where a "zombie" teamd enters its event loop after the parent (teammgrd) has already timed out and declared it failed. The 7-line change is minimal and well-isolated.

### Does This Fix the Race Condition?
**Yes.** When `daemon_retval_send(0)` fails (because the parent has already closed the pipe after timing out), teamd now exits immediately instead of proceeding into `teamd_run_loop_run()`. This eliminates the window where a zombie teamd runs while teammgrd is actively restarting it.

### Issues / Requested Changes

1. **Typo in patch header** — The patch description says `teamdmgr` but the correct daemon name in SONiC is `teammgrd`. Patch headers are permanent in git history and should be corrected.

2. **Speculative error log message** — The log message "unable to send response, likely timed out" is a guess. The failure could also be a broken pipe or closed fd. Please include the actual return value or `strerror(errno)` for better debuggability:
   ```c
   teamd_log_err("unable to send response to parent (err=%d: %s).", err, strerror(errno));
   ```

3. **Error value propagation** — The raw non-zero return value from `daemon_retval_send` is propagated as `teamd_start`'s return value. Upstream teamd convention uses negative errno values. Consider mapping this to a consistent negative errno (e.g., `-EPIPE`) to avoid confusing callers.

4. **No test coverage documented** — Understandable given the nature of a race condition, but worth noting in the PR description how this was verified (e.g., stress testing, manual reproduction).

5. **Backport conflict on msft-202405** — The cherry-pick conflict on `msft-202405` will need manual resolution before that backport can land.

### Overall
None of the above are blockers for the core fix. Items 1 and 2 in particular are straightforward cleanups worth doing before merge. The approach is sound and the change is safe.

bingwang-ms (dismissed):

PR #23664, opened 2025-08-10
Switch to use KILL as stop signal of openbmpd

FengPan-Frank:
#### Why I did it
To fix upgrade reboot core issue, simply use stopsignal=KILL so that supervisord will use SIGKILL instead of the default SIGTERM when stopping container, thus the process is terminated immediately by the kernel and skip signal handler. Existing openbmpd code has potential race issue in handling graceful termination.

image

which has already been used in bgpd/bfpd 
https://github.com/sonic-net/sonic-buildimage/blob/5cedc4f771a02df232b0b6dadd4c5c20e46b9527/dockers/docker-fpm-frr/frr/supervisord/supervisord.conf.j2#L88

##### Work item tracking
- Microsoft ADO **(number only)**: 33441494

#### How I did it
Switch to use stopsignal=KILL in supervisord.conf

#### How to verify it
supervisord.conf verified on DUT 
image

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

- [ ] 
- [ ]

### DIFF
--- a/dockers/docker-sonic-bmp/supervisord.conf
+++ b/dockers/docker-sonic-bmp/supervisord.conf
@@ -57,6 +57,7 @@ dependent_startup=true
 [program:openbmpd]
 command=/usr/bin/openbmpd -f -c /etc/bmp/openbmpd.conf
 priority=3
+stopsignal=KILL
 autostart=false
 autorestart=false
 stdout_logfile=NONE


dockers/docker-sonic-bmp/supervisord.conf
```diff
@@ -57,6 +57,7 @@ dependent_startup=true
 [program:openbmpd]
 command=/usr/bin/openbmpd -f -c /etc/bmp/openbmpd.conf
 priority=3
+stopsignal=KILL
```
make1980, line 60: could you open an ADO to fix the termination handling in bmp? k8s likely will send terminate instead of kill when restarting container?

make1980 (approved):

dockers/docker-sonic-bmp/supervisord.conf
FengPan-Frank: https://github.com/sonic-net/sonic-bmp/issues/21   created a public issue since this will be openbmpd repo issue probably.

qiluo-msft (approved):

PR #23681, opened 2025-08-12
[Mellanox] Stop SDK monitor thread before syncd shutdown

volodymyrsamotiy:
#### Why I did it
To avoid error message is syslog during shutdown flow related to SDK monitor thread.
Error message was observed only during shutdown flow, hence there was no functional impact, just error log.
It was not always reproducible because the reason for it was race condition related to monitor thread.

##### Work item tracking

#### How I did it
During shutdown flow stop SDK monitor thread before actual SDK shutdown in order to avoid race condition.

#### How to verify it
By running any type of reboot or config reload

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

- [ ] 
- [ ]

### DIFF
--- a/files/scripts/syncd.sh
+++ b/files/scripts/syncd.sh
@@ -97,6 +97,10 @@ function stopplatform1() {
         debug "Stopped pmon service"
     fi
 
+    if [[ x$sonic_asic_platform == x"mellanox" ]]; then
+        echo "health_check_trigger del_dev 1" > /proc/mlx_sx/sx_core
+    fi
+
     if [[ x$sonic_asic_platform != x"mellanox" ]] || [[ x$TYPE != x"cold" ]]; then
         # Invoke platform specific pre shutdown routine.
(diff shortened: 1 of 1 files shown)


bingwang-ms (approved):

bingwang-ms:
LGTM. The error log is also noticed in 202505 qual.
If there is no functionality impact, I think we can skip 202505 branch and merge it to master only. How do you think?

keboliu:
> LGTM. The error log is also noticed in 202505 qual. If there is no functionality impact, I think we can skip 202505 branch and merge it to master only. How do you think?

Hi @bingwang-ms although the error log is not functional but it will be captured by the log analyzer and fail some test, so it sill good to cherry-pick to avoid log analyzer failure.

bingwang-ms:
@yejianquan Can you please evaluate the cherry-pick request to 202505? Thanks

PR #23696, opened 2025-08-13
Default LEGACY_SONIC_MGMT_DOCKER to "n"

wangxin:
#### Why I did it
If the LEGACY_SONIC_MGMT_DOCKER parameter is not explicitly set during the build, it currently defaults to "y".

Since all active SONiC releases now use Python 3 for running tests, it is preferable to build docker-sonic-mgmt with Python 3 by default. Therefore, the default value of LEGACY_SONIC_MGMT_DOCKER should now be set to "n".

##### Work item tracking

#### How I did it
Set default value of LEGACY_SONIC_MGMT_DOCKER to "n".

#### How to verify it
Build docker-sonic-mgmt.

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

- [ ] 
- [ ]

### DIFF
--- a/Makefile.work
+++ b/Makefile.work
@@ -161,7 +161,7 @@ ifeq ($(ENABLE_DOCKER_BASE_PULL),)
 endif
 
 ifeq ($(LEGACY_SONIC_MGMT_DOCKER),)
-	override LEGACY_SONIC_MGMT_DOCKER = y
+	override LEGACY_SONIC_MGMT_DOCKER = n
 endif
 
 ifeq ($(CONFIGURED_ARCH),amd64)


r12f (approved):

ymd-arista:
This change opens a can of worms.

With https://github.com/sonic-net/sonic-mgmt/pull/11457, in python3 sonic-mgmt docker environments, we are forced to use yaml modules to load files and this has repercussions.

If we are using [TestbedProcessin.py](https://github.com/sonic-net/sonic-mgmt/blob/master/ansible/TestbedProcessing.py#L1121) to generate veos (which has ansible inventory entries for test server) and lab (which has inventory entries for dut, ptf, and fanout switch), we would be forced to use it with yaml flag as with https://github.com/sonic-net/sonic-mgmt/pull/11457, we can only open yaml files.

If you take a look at [makeLab](https://github.com/sonic-net/sonic-mgmt/blob/master/ansible/TestbedProcessing.py#L504) and [makeLabYaml](https://github.com/sonic-net/sonic-mgmt/blob/master/ansible/TestbedProcessing.py#L794), there are so many required fields that are not at all handled in current `makeLabYaml` function in its current state. So, we would have to fix those functions (`makeLabYAML`, `makeVeosYAML`) **first** before proceeding with this change. Else, we would be looking at a lot of sanitization failures.

qiluo-msft (approved):

ymd-arista (changes requested):
Leaving this review in so we can discuss these changes before merging them in.

wangxin:
Change of this PR is included in https://github.com/sonic-net/sonic-buildimage/pull/24306
Closing this PR.

Issue #23406, opened 2025-07-21
Enhancement: Secureboot production signing of kernel modules using ephemeral keys + bug fixes


bhouse-nexthop:
### Is it platform specific

generic

### Importance or Severity

Critical

## Kernel module ephemeral key signing support
The current sonic secureboot implementation assumes all assets are signed with the DB Key. There are roughly 3600 kernel modules that get signed in the process.

Organizations may have varying security policy requirements where any signing requests may require signoff by one or more parties, this may simply be infeasible for that many requests.

Given that the DB key may be a long-lived key, and only RSA 2048 is often used or available, it is a security best-practice to sign as few assets as possible with such a key.

## Sonic Image Signature Verification
The current signature verification of sonic images assumes the DB Keys are all Root CAs. The secureboot standard says nothing about this, the DBKeys are explicitly trusted by signing them with the KEK, and that signing method does not follow the standard X.509 PKI architecture. Therefore the DB Key is not guaranteed to be a CA Root (aka not self-signed). It is possible the DB Key was created as an intermediate, but since it is explicitly trusted that is ok.

## Bugs to be fixed
* Loading of signed platform modules fails due to `-f` flag passed to `modprobe` as it invalidates the signature
* `SECURE_UPGRADE_MODE` not over-writable by environment like other secureboot variables
* `SECURE_UPGRADE_PROD_TOOL_ARGS` not quoted properly so only passes through first argument specified

### Current Behavior

Described above

### Proposed Behavior

PRs:

- https://github.com/sonic-net/sonic-buildimage/pull/23405
- https://github.com/sonic-net/sonic-utilities/pull/3989
- https://github.com/sonic-net/sonic-linux-kernel/pull/499

bhouse-nexthop:
@davidpil2002 since you contributed the dev secureboot signing script, could you review these secureboot related changes?

bradh352:
looks like this autoclosed because one of the eight linked PRs merged.  @lguohan can you please reopen?

lguohan:
reopened

antonio-nexthop:
### Appending to this issue:

#### Current Behaviour:

Currently, `sonic-installer` does not install grub and other secure boot components. Therefore, if we have a non-secure boot image and we move to a secure boot image via `sonic-installer`, and then turn on secure boot enforcement, the boot process will fail. Further, even if we have a secure boot image, signatures may differ between images, even if grub itself is identical.

#### Proposed Behaviour:

[...]

PR #23732, opened 2025-08-17
Secureboot: PDDF: Do not always use -f to load custom PDDF modules

bhouse-nexthop:
#### Why I did it

When in kernel modules are signed for secureboot, passing `-f` to modprobe will fail with a module signing issue even though the module is signed properly.   This appears to be due to some internal module modification to handle mismatched kernel versions which invalidates the signature.

##### Work item tracking

#### How I did it

This modification first tries to load custom modules without the `-f` flag, then if (and only if) that fails, try again with the `-f` flag.  No attempt is made to determine if the kernel module is signed as it was not necessary and would only complicate the code further.

It is not clear what the valid use-case is for using `modprobe -f`, however this behavior has been maintained in this modification.

#### How to verify it

Build a SONiC secure-boot enabled build and observe custom modules (like PDDF modules) load proper now, and previously failed without this patch.

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [x] 202411
- [x] 202505

master as of 20250817

Secureboot: PDDF: Do not always use -f to load custom PDDF modules

Fixes: #23406

### DIFF
--- a/platform/pddf/i2c/utils/pddf_util.py
+++ b/platform/pddf/i2c/utils/pddf_util.py
@@ -36,8 +36,9 @@ DEBUG = False
 args = []
 ALL_DEVICE = {}               
 FORCE = 0
-kos = []
 perm_kos = []
+std_kos = []
+custom_kos = []
 devs = []
 
 # Instantiate the class pddf_obj
@@ -73,16 +74,13 @@ def main():
     
     # generate the KOS list from pddf device JSON file
     if 'std_perm_kos' in pddf_obj.data['PLATFORM'].keys():
-       kos.extend(pddf_obj.data['PLATFORM']['std_perm_kos'])
        perm_kos.extend(pddf_obj.data['PLATFORM']['std_perm_kos'])
-    kos.extend(pddf_obj.data['PLATFORM']['std_kos'])
-    kos.extend(pddf_obj.data['PLATFORM']['pddf_kos'])
 
-    kos = ['modprobe '+i for i in kos]
+    std_kos.extend(pddf_obj.data['PLATFORM']['std_kos'])
+    std_kos.extend(pddf_obj.data['PLATFORM']['pddf_kos'])
 
     if 'custom_kos' in pddf_obj.data['PLATFORM']:
-        custom_kos = pddf_obj.data['PLATFORM']['custom_kos']
-        kos.extend(['modprobe -f '+i for i in custom_kos])
+        custom_kos.extend(pddf_obj.data['PLATFORM']['custom_kos'])
 
     for opt, arg in options:
         if opt in ('-h', '--help'):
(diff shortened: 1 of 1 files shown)


bhouse-nexthop:
@DavidZagury @davidpil2002  @qiluo-msft please review

bhouse-nexthop:
@Yarden-Z any comments?

lguohan:
Here’s a review of the changes in [PR #23732 for sonic-net/sonic-buildimage](https://github.com/sonic-net/sonic-buildimage/pull/23732/files):

## Summary

This PR refactors the handling of kernel object modules (KOS) in `platform/pddf/i2c/utils/pddf_util.py`. It separates kernel modules into three lists: `perm_kos`, `std_kos`, and `custom_kos`, and adjusts how modules are loaded and unloaded during driver install/uninstall.

---

## Key Changes

### 1. Variable Refactoring

- **Before:**  
  A single list, `kos`, was used for all kernel modules, including permanent, standard, and custom modules.
- **After:**  
  Introduces three lists:
  - `perm_kos`: For permanent kernel modules.
  - `std_kos`: For standard kernel modules.
  - `custom_kos`: For custom kernel modules.

### 2. Module Loading Logic

- **Before:**  
  All modules in `kos` were loaded using `modprobe` or `modprobe -f` for custom modules.
- **After:**  
  - Loads `perm_kos` and `std_kos` with `modprobe`.
  - Loads `custom_kos` with `modprobe`, and if that fails, retries with `modprobe -f`.
  - Skips force-loading signed modules on the first try.

### 3. Module Unloading Logic

- **Before:**  
  Unloads all modules in `kos` (except for those in `perm_kos` or named `i2c-i801`) using string replacements.
- **After:**  
  - Unloads `std_kos` and `custom_kos` (not `perm_kos`).
  - Skips only modules containing `i2c-i801`.
  - Uses explicit `"modprobe -rq"` command for unloading.

---

## Pros

- **Improved Clarity:**  
  Separating module categories makes the code easier to understand and maintain.
- **Safer Module Loading:**  
  The logic to retry loading custom modules with `-f` only after a standard attempt is cleaner and safer.
- **Simplified Unloading:**  
  The new approach is more explicit and less prone to string substitution errors.
- **Better Extensibility:**  
  Adding or modifying module handling logic in the future will be easier.

---

## Cons / Suggestions

- **Backward Compatibility:**  
  If external scripts or tools depend on the old `kos` list structure, there could be integration issues. Review all callers.
- **Variable Initialization:**  
  Ensure all new lists (`std_kos`, `custom_kos`) are always initialized as lists, even if empty, to avoid `NoneType` issues.
- **Logging:**  
  Consider more detailed logging, especially when a module fails to load or unload.

---

## Overall Assessment

**This is a solid refactor that improves the clarity and reliability of kernel module handling in the PDDF utilities.**  
The code is easier to follow, less error-prone, and provides a better framework for future enhancements. Just ensure that all references to the old `kos` variable are updated and that empty list cases are handled.

**Recommendation:**  
Looks good to merge after confirming backward compatibility and list initialization.

PR #23735, opened 2025-08-17
Secureboot: install secureboot shim/grub if image is built with secureboot

bhouse-nexthop:
#### Why I did it

During an ONIE install, secureboot shim/grub were only being installed if ONIE was booted in secureboot mode.  This is undesirable as it would prevent the user from enabling secureboot after the fact.  The secureboot signed shim and grub will operate normally with secureboot disabled.

#### How I did it

Update installer script to always install the signed secureboot grub and shim if found in the relevant paths.

#### How to verify it

Build a secureboot enabled image.  Boot into a machine with secureboot disabled in the BIOS.  Install SONiC via ONIE.  Reboot and enable secureboot in the BIOS.  Observe SONiC boots as expected with secureboot enabled.  Prior to this patch it would not.

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [x] 202411
- [x] 202505

master as of 20250817

Secureboot: install secureboot shim/grub if image is built with secureboot

Fixes: https://github.com/sonic-net/sonic-buildimage/issues/23406

### DIFF
--- a/installer/default_platform.conf
+++ b/installer/default_platform.conf
@@ -438,21 +438,23 @@ bootloader_menu_config()
 
         echo "firmware=$firmware"
         if [ "$firmware" = "uefi" ] ; then
-            secure_boot_state=0
-            reg_sb_guid=""
+            secure_boot=no
             ENABLED=1
-            echo "checking secure boot state"
-            reg_sb_guid=$(efivar -l | grep "SecureBoot$") || echo "Secure Boot GUID not found in efivar list"
-            echo "Secure Boot GUID=$reg_sb_guid"
-            if [ -n "$reg_sb_guid" ]; then
-                secure_boot_state=$(efivar -d --name $reg_sb_guid) || echo "Could not read Secure Boot state from efivar"
+            # We know this image was built with secureboot support if we have the shim and grub EFI images in the
+            # expected paths.  There is currently no other indicator.  We could detect if Secureboot is enabled
+            # (via $(mokutil --sb | grep "enabled") or $(efivar -l | grep "SecureBoot$"), but that would mean the user
+            # would not be able to enable secureboot at a later point in time without a reinstall from ONIE with
+            # Secureboot enabled which could cause confusion.
+            echo "checking if image was built with secure boot support"
+            if [ -f "$demo_mnt/$image_dir/boot/shimx64.efi" -a -f "$demo_mnt/$image_dir/boot/grubx64.efi" ]; then
+                secure_boot=yes
             fi
-            echo secure_boot_state=$secure_boot_state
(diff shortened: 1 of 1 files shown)


bhouse-nexthop:
@DavidZagury @davidpil2002 @qiluo-msft please review

bhouse-nexthop:
@Yarden-Z any comments?

lguohan:
This pull request makes significant changes to the logic for detecting and handling UEFI Secure Boot in the installer script default_platform.conf. Here’s a review of the PR:

## Summary of Changes
- **Old Logic:**  
  The installer tried to detect whether Secure Boot was enabled on the system by checking efivar and extracting the Secure Boot GUID and state. If Secure Boot was enabled (state matched), it installed the shim bootloader; otherwise, it installed the regular grub bootloader.

- **New Logic:**  
  Instead of checking the system’s Secure Boot state, the script now checks whether the image itself was built with Secure Boot support. This is done by verifying the existence of specific EFI files (`shimx64.efi` and `grubx64.efi`) in the image’s boot directory.  
  - If both files exist, it treats the image as Secure Boot-enabled and installs the shim bootloader.
  - If not, it installs the regular grub bootloader.
  - The script does not check the current system’s Secure Boot status anymore.

## Motivation (from code comments)
- The PR notes that checking the runtime Secure Boot status (with mokutil or efivar) would force users to reinstall from ONIE with Secure Boot enabled if they want to enable it later, which could be confusing.
- Instead, the installer relies on the presence of the appropriate images to determine Secure Boot support.

## Pros
- **Simplicity:** The new approach is simpler and relies on the build artifacts rather than system state, which may be more predictable for automated deployments.
- **User Experience:** Avoids a scenario where users must reinstall the OS just to enable Secure Boot.

## Cons / Considerations
- **Potential for Misconfiguration:** If the image is Secure Boot-capable but Secure Boot is not enabled in firmware, the shim bootloader will be installed regardless. This might be fine, but should be documented.
- **No Dynamic Detection:** If a user enables Secure Boot after installation, the bootloader may not match the firmware’s state unless the image is rebuilt/reinstalled.
- **Documentation:** The rationale is explained in the comments, but release notes and user documentation should be updated to reflect this change in behavior.

## Suggestions
- Consider extending documentation to explain the new detection method and its implications for both users and maintainers.
- If possible, add a warning or note in the installer output clarifying why the regular Secure Boot state is not checked, to avoid confusion.
- Ensure that downstream tools/scripts that depend on the old behavior are updated accordingly.

## Conclusion
This PR streamlines Secure Boot image detection by focusing on image content rather than system state. It improves predictability and reduces the risk of confusing reinstall requirements. However, it trades off dynamic detection for build-time certainty, so clear documentation and communication are important.

Let me know if you want an in-depth review of code style, security, or potential edge cases!

PR #23768, opened 2025-08-20
[FRR]: Add fix for the issue reported in https://github.com/sonic-net/sonic-buildimage/issues/23121

cscarpitta:
This PR fixes an issue with routes installation in FRR caused by the ordering of events.
 
* First, a kernel route is received.
* Then, a connected route for the same destination arrives.
* The connected route should win over the kernel route. As a result, FRR is expected to remove the kernel route and keep only the connected route.
 
However, there are cases where, due to the order in which events are processed, FRR ends up with both a connected route and a kernel route for the same destination.
 
This PR fixes the issue by ensuring that the kernel route is removed from the early route queue before the connected route is installed.

Fixes https://github.com/sonic-net/sonic-buildimage/issues/23121


ahsalam (approved):

ahsalam:
Nvidia team confirmed this PR fixed the issue. https://github.com/sonic-net/sonic-buildimage/issues/23121#issuecomment-3323147963

@dgsudharsan @Yakiv-Huryk @BYGX-wcr

BYGX-wcr:
According to the latest routing group's consensus, we need to assign a unique id for each FRR patch. Is this PR covered by the new policy?

cscarpitta:
@BYGX-wcr

Hi Changrong, thanks for the review.
Yes, I'm assigning the unique ID 0062 to the new patch.

----

Screenshot 2025-09-30 at 21 45 40

BYGX-wcr:
@dgsudharsan , can you sign-off?

dgsudharsan (approved):

dgsudharsan:
@StormLiangMS  Can you please help to merge?

PR #23842, opened 2025-08-29
[Arista] Enable kdump by default for the first time of booting on some Arista platforms

byu343:
#### Why I did it
This is to enable kdump by default for the first time booting of a new image.
We enable it on the platforms defined in write_kdump_cmdline().

##### Work item tracking

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [x] 202405
- [x] 202411
- [x] 202505

- [ ] 
- [ ]

### DIFF
--- a/files/Aboot/boot0.j2
+++ b/files/Aboot/boot0.j2
@@ -488,6 +488,7 @@ get_eeprom_value() {
 write_kdump_cmdline() {
     local sid="$(cmdline_get sid | sed 's/Ssd$//')"
     case "$sid" in
+        Lodoga*|*Quicksilver*|*Moby|Shearwater*|Moranda*|\
         Wolverine*|Clearwater2*|OtterLake*|QuartzDd*|Redstart8Mk2Quartz4*|CitrineDd*)
             if ! cmdline_has crashkernel; then
                 cmdline_add crashkernel=0M-2G:256M,2G-4G:320M,4G-8G:384M,8G-16G:448M,16G-32G:768M,32G-:1G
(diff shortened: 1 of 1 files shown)


files/Aboot/boot0.j2
```diff
@@ -488,6 +488,7 @@ get_eeprom_value() {
 write_kdump_cmdline() {
     local sid="$(cmdline_get sid | sed 's/Ssd$//')"
     case "$sid" in
+        Lodoga*|*Quicksilver*|*Moby|Shearwater*|Moranda*|\
```
saiarcot895: Would this always enable kdump, with no way to disable it?

byu343: No, with this change, it is always enabled for the first time of booting of newly installed image and can be disabled through the existing CLI "config kdump disable" after the first time of booting.

This function is called by regular_install(), adding crashkernel to kernel-cmdline file of the new image. After the first time booting, the CLI to disable kdump could modify kernel-cmdline and regular_install() will not be called again when rebooting.

abdosi:
@judyjoseph to make sure we enable by kdump by default for other platforms.

saiarcot895 (approved):

StormLiangMS:
hi @byu343 master merged, since this is more like an enhancement other than bug fix, I would keep it at master and later branches other than cherrypick to old ones unless strong justification to do so.

deepak-singhal0408:
tag @arlakshm to ppick up,.

judyjoseph:
@byu343 can you confirm this PR enables for all Arista platforms including Arista-7060X6* skus ?

PR #23864, opened 2025-09-01
Add `preserve_fqdn` to remote syslog messages

spandan-nexthop:
Add `preserve_fqdn` to `SYSLOG_CONFIG` and update `hostname` in `rsyslog-config.sh` such that the remote syslog messages have the long/short hostname accordingly.
Fixes https://github.com/sonic-net/sonic-buildimage/issues/23863

#### Why I did it
Currently, in SONiC, syslog messages are only sent with the shortened hostname, not the FQDN. It would be good to conditionally be able to toggle the hostname to the long (FQDN) or short form in remote syslog messages.

##### Work item tracking

#### How I did it
- Added a new leaf - `preseve_fqdn` - to `SYSLOG_CONFIG` in `sonic-syslog.yang`
- Since any change inside `SYSLOG_CONFIG|GLOBAL` automatically triggers an `rsyslog-config` restart, I just added a check for `preserve_fqdn` in `rsyslog-config.sh` such that it sets the `hostname` to the long/short form correctly.
- Note that, adding `PreserveFQDN on|off` in the syslog config would have been an alternate solution, but that approach requires an update and restart of rsyslog in the host AS WELL AS all the containers separately since each run their own syslog daemon. This was a more disruptive solution and we chose to not go down that path.

#### How to verify it

Enabling `preserve_fqdn` and doing a tcpdump for the syslog messages shows the correct FQDN.
```
host$ show syslog
SERVER IP    SOURCE IP    PORT    VRF
-----------  -----------  ------  -----
10.11.0.5    N/A          N/A     N/A
10.11.0.6    N/A          N/A     N/A

host$ sonic-db-cli CONFIG_DB HSET "SYSLOG_CONFIG|GLOBAL" "preserve_fqdn" "true"
1

(snmp)root@host:/# logger --priority CRIT "Test syslog message from snmp container MSG1"

23:50:54.629484 lo    In  IP localhost.38482 > localhost.syslog: SYSLOG user.critical, length: 104
E.....@.@..H.........R...p..<10>2025-07-31T23:50:54.629330+00:00 <host_FQDN> snmp#root: Test syslog message from snmp container MSG1
23:50:54.629610 eth0  Out IP 10.250.15.2.36470 > 10.11.0.5.syslog: SYSLOG user.critical, length: 130
E....!@.@.M"
```

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

- [ ] 
- [ ] 

Adding `preserve_fqdn` to `SYSLOG_CONFIG` to conditionally toggle the full/short hostname sent in syslog messages.

https://github.com/spandan-nexthop/sonic-buildimage/blob/spandan-nexthop.add-fqdn-syslog/src/sonic-yang-models/yang-models/sonic-syslog.yang#L184

red-panda


bingwang-ms:
@qiluo-msft Can you help review this PR? Thanks

PR #23917, opened 2025-09-06
[Fix]: Allow empty object lists in high-frequency-telemetry YANG model

Pterosaur:
#### Why I did it
We must have a dummy HFT table to support dynamically configuring HFT. This Yang model change is to support the dummy table.

##### Work item tracking

#### How I did it
- Remove mandatory constraint for object_names list
- Add test cases for empty lists and validation scenarios
- Include high-frequency telemetry model
- 
#### How to verify it

Check Azp status. This PR includes its test.

- [x] 202412

master

### DIFF
--- a/src/sonic-yang-models/setup.py
+++ b/src/sonic-yang-models/setup.py
@@ -62,6 +62,7 @@ yang_files = [
     'sonic-feature.yang',
     'sonic-fips.yang',
     'sonic-hash.yang',
+    'sonic-high-frequency-telemetry.yang',
     'sonic-trimming.yang',
     'sonic-system-defaults.yang',
     'sonic-interface.yang',
@@ -146,7 +147,6 @@ yang_files = [
     'sonic-smart-switch.yang',
     'sonic-spanning-tree.yang',
     'sonic-srv6.yang',
-    'sonic-high-frequency-telemetry.yang',
 ]
 
 class my_build_py(build_py):
--- a/src/sonic-yang-models/setup.py
+++ b/src/sonic-yang-models/setup.py
@@ -146,6 +146,7 @@ yang_files = [
     'sonic-smart-switch.yang',
     'sonic-spanning-tree.yang',
     'sonic-srv6.yang',
+    'sonic-high-frequency-telemetry.yang',
 ]
 
 class my_build_py(build_py):
--- a/src/sonic-yang-models/tests/yang_model_tests/tests_config/high_frequency_telemetry.json
+++ b/src/sonic-yang-models/tests/yang_model_tests/tests_config/high_frequency_telemetry.json
@@ -69,6 +69,16 @@
                         "name": "high_frequency_counters",
                         "stream_state": "enabled",
                         "poll_interval": 100
+                    },
+                    {
+                        "name": "partial_objects",
+                        "stream_state": "disabled",
+                        "poll_interval": 100000
+                    },
+                    {
+                        "name": "dummy_profile",
+                        "stream_state": "disabled",
+                        "poll_interval": 100000
                     }
                 ]
             },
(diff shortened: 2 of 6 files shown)


src/sonic-yang-models/setup.py
```diff
@@ -146,6 +146,7 @@
     'sonic-smart-switch.yang',
     'sonic-spanning-tree.yang',
     'sonic-srv6.yang',
+    'sonic-high-frequency-telemetry.yang',
```
wen587: let's put by alphabet order

src/sonic-yang-models/tests/yang_model_tests/tests_config/high_frequency_telemetry.json
```diff
@@ -120,6 +130,110 @@
...
+                    },
+                    {
+                        "name": "Ethernet4",
+                        "lanes": "4",
+                        "speed": 25000
+                    }
+                ]
+            }
+        },
+        "sonic-high-frequency-telemetry:sonic-high-frequency-telemetry": {
```
wen587, line 214: could you add a test for empty object list? I don't think current YANG allowed empty list becasue there are still constraints here if group_name is PORT:
                leaf-list object_names {
                   type string;
                   must "( ../group_name = 'PORT' and current() = /port:sonic-port/port:PORT/port:PORT_LIST/port:name )"
                        + " or ( ../group_name = 'BUFFER_POOL' and current() = /bpl:sonic-buffer-pool/bpl:BUFFER_POOL/bpl:BUFFER_POOL_LIST/bpl:name )"
                        + " or ( ../group_name = 'INGRESS_PRIORITY_GROUP' and substring-before(current(), '|') = /bpg:sonic-buffer-pg/bpg:BUFFER_PG/bpg:BUFFER_PG_LIST/bpg:port and re-match(substring-after(current(), '|'), '[0-9]+') )"
                        + " or ( ../group_name = 'QUEUE' and substring-before(current(), '|') = /bql:sonic-buffer-queue/bql:BUFFER_QUEUE/bql:BUFFER_QUEUE_LIST/bql:port and re-match(substring-after(current(), '|'), '[0-9]+') )";
                }

src/sonic-yang-models/yang-models/sonic-high-frequency-telemetry.yang
```diff
@@ -93,8 +93,6 @@ module sonic-high-frequency-telemetry {
                         + " or ( ../group_name = 'QUEUE' and substring-before(current(), '|') = /bql:sonic-buffer-queue/bql:BUFFER_QUEUE/bql:BUFFER_QUEUE_LIST/bql:port and re-match(substring-after(current(), '|'), '[0-9]+') )";
                 }
 
-                must "count(object_names) > 0";
-
                 leaf-list object_counters {
```
wen587, line 96: please also include such config in sample_config_db.json

src/sonic-yang-models/tests/yang_model_tests/tests_config/high_frequency_telemetry.json
Pterosaur: https://github.com/sonic-net/sonic-buildimage/pull/23917/files#diff-e4744a17a9c8acd94344ef9dfdec05bee6b87ce5506dd69e790fba84e7b46eb5R145-R148

This test is for empty object list.

src/sonic-yang-models/yang-models/sonic-high-frequency-telemetry.yang
Pterosaur: Done

src/sonic-yang-models/setup.py
Pterosaur: Done

src/sonic-yang-models/tests/yang_model_tests/tests_config/high_frequency_telemetry.json
wen587: The test in the link still have elements in object_names
```
                 "profile_name": "partial_objects",
                        "group_name": "PORT",
                        "object_names": [
                            "Ethernet0"
                        ],
```
I mean empty object_names what ndm will generates. such as:
```
   "profile_name": "partial_objects",
                        "group_name": "PORT",
                        "object_names": [],
```
or
```
 "profile_name": "partial_objects",
                        "group_name": "PORT",
                        "object_names": "",
```

Pterosaur: It seems that our SONiC YANG parser doesn't support the structure like:
```
"object_names": [],
```
It will raise an error from Yang parser.
```
libyang[0]: Invalid JSON data (unexpected value). (path: /sonic-high-frequency-telemetry:sonic-high-frequency-telemetry/HIGH_FREQUENCY_TELEMETRY_GROUP/HIGH_FREQUENCY_TELEMETRY_GROUP_LIST[profile_name='dummy_profile'][group_name='PORT']/object_names)
 Exception >Invalid JSON data (unexpected value). in not empty< in /sonic/src/sonic-yang-models/tests/yang_model_tests/test_yang_model.py:278
```
IMHO, its schema is defined by us, so we can define we only support a json without these fields.

wen587: Hi @Pterosaur , right. That's what i am concerned. e.g. if NDM liquid provides "" for object_names and object_counters. Such config will still fail yang validation after we remove `must "count(object_names) > 0";` . 

liquid is using "" for field value:
https://msazure.visualstudio.com/One/_git/Networking-Metadata?path=/src/data/Network/DeviceConfigTemplates/Sonic_HIGH_FREQUENCY_TELEMETRY_GROUP_Template.liquid&version=GBmaster&_a=contents

To support such field for NDM, should we remove the constraints? such as 
```
                leaf-list object_names {
                   type string;
                   must "( ../group_name = 'PORT' and current() = /port:sonic-port/port:PORT/port:PORT_LIST/port:name )"
```
 I am not sure if "" applies to leaf-list. or we should pick [] instead for NDM

Pterosaur: Gotch, I add two negative test cases, please check it. We need to update the NDM code later.

src/sonic-yang-models/yang-models/sonic-high-frequency-telemetry.yang
```diff
@@ -93,8 +93,6 @@ module sonic-high-frequency-telemetry {
                         + " or ( ../group_name = 'QUEUE' and substring-before(current(), '|') = /bql:sonic-buffer-queue/bql:BUFFER_QUEUE/bql:BUFFER_QUEUE_LIST/bql:port and re-match(substring-after(current(), '|'), '[0-9]+') )";
                 }
 
-                must "count(object_names) > 0";
```
ganglyu, line 96: https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-yang-models/yang-models/sonic-pbh.yang#L119

Can we use min-elements?

Pterosaur: Use min-elements for what? I removed this condition because I want to support no explicate fields in HFT_GROUP object. such as:
```
"dummy_table|PORT": {}
```

ganglyu (approved):

qiluo-msft (approved):

PR #23932, opened 2025-09-09
Add support for use-route-replace in sonic dataplane provider

venkit-nexthop:
The FRR's zebra has two dataplane providers. One of them programs routes into the kernel, while the other one sends routes towards sonic. This one sends routes to fpmsyncd. Currently, this sonic dataplane provider does not make use of the route replace semantics. This means that any change to an existing route is first sent as a delete and then a new add notification is sent later. 
The sonic dataplane provider code currently sends a RTM_DELROUTE followed by a RTM_NEWROUTE for an update operation as well. This will lead to churn even when the ECMP set is growing.
This PR enables the "fpm use-route-replace" configuration in zebra. Additionally it also fixes the "show fpm status [json]" show cli.

Refer to the corresponding PR in fpmsyncd to be able to accept this https://github.com/sonic-net/sonic-swss/pull/3872

#### Why I did it

This will allow zebra to be configured with "fpm use-route-replace", which will cause it to update a route instead of delete followed by new.  This also helps when there are ECMP changes (ecmp growth or shrink) and allows for zero traffic drops. 

##### Work item tracking

#### How I did it
Added support for route replace semantics in zebra's sonic dataplane provider.

#### How to verify it
Verified with ECMP routes with growth or shrink of ECMP.

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

- [ ] 
- [ ] 

Add support for use-route-replace in sonic dataplane provider

### DIFF
--- a/dockers/docker-fpm-frr/frr/zebra/zebra.conf.j2
+++ b/dockers/docker-fpm-frr/frr/zebra/zebra.conf.j2
@@ -15,6 +15,13 @@ fpm use-next-hop-groups
 ! Uses the old known FPM behavior of including next hop information in the route (e.g. RTM_NEWROUTE) messages
 no fpm use-next-hop-groups
 {% endif %}
+{% if ( ('localhost' in DEVICE_METADATA) and ('route_replace' in  DEVICE_METADATA['localhost']) and
+        (DEVICE_METADATA['localhost']['route_replace'] == 'enabled') ) %}
+! enable route replace semantics
+fpm use-route-replace
+{% else %}
+no fpm use-route-replace
+{% endif %}
 !
 fpm address 127.0.0.1
 {% endblock fpm %}
--- a/src/sonic-frr/dplane_fpm_sonic/dplane_fpm_sonic.c
+++ b/src/sonic-frr/dplane_fpm_sonic/dplane_fpm_sonic.c
@@ -45,6 +45,7 @@
 #include "lib/network.h"
 #include "lib/ns.h"
 #include "lib/frr_pthread.h"
+#include "lib/termtable.h"
 #include "zebra/debug.h"
 #include "zebra/interface.h"
 #include "zebra/zebra_dplane.h"
@@ -176,6 +177,7 @@ struct fpm_nl_ctx {
 	bool disabled;
 	bool connecting;
 	bool use_nhg;
+	bool use_route_replace;
 	struct sockaddr_storage addr;
 
 	/* data plane buffers. */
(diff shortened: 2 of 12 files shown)


prabhataravind:
@dgsudharsan suggests to bring this up in routing subgroup.

dockers/docker-fpm-frr/frr/zebra/zebra.conf.j2
```diff
@@ -15,6 +15,13 @@ fpm use-next-hop-groups
 ! Uses the old known FPM behavior of including next hop information in the route (e.g. RTM_NEWROUTE) messages
 no fpm use-next-hop-groups
 {% endif %}
+{% if ( ('localhost' in DEVICE_METADATA) and ('route_replace' in  DEVICE_METADATA['localhost']) and
```
dgsudharsan: Lets have the knob in constants.yml rather than making it a user configuration. In case if there are issue we can use that knob to disable the feature.

src/sonic-frr/dplane_fpm_sonic/dplane_fpm_sonic.c
```diff
@@ -2207,6 +2304,20 @@ static int fpm_nl_enqueue(struct fpm_nl_ctx *fnc, struct zebra_dplane_ctx *ctx)
...
+	 * If route replace is enabled then directly encode the install which
+	 * is going to use `NLM_F_REPLACE` (instead of delete/add operations).
+	 */
+	if (fnc->use_route_replace && op == DPLANE_OP_ROUTE_UPDATE) {
+		nexthop = dplane_ctx_get_ng(ctx)->nexthop;
+		if (nexthop && nexthop->nh_srv6) {
+			// Dont change op for srv6 yet. Not sure if update
+			// semantics will work or not
+		} else {
+			op = DPLANE_OP_ROUTE_INSTALL;
```
eddieruan-alibaba, line 2317: You could make the same change for srv6 routes as well. Previously, there is no UPDATE event, therefore, UPDATE is treated as delete. 

You could change UPDATE as INSTALL once use_route_replace is true for all cases.

dockers/docker-fpm-frr/frr/zebra/zebra.conf.j2
venkit-nexthop: Thanks for the suggestion. I will take a look at contants.yml

src/sonic-frr/dplane_fpm_sonic/dplane_fpm_sonic.c
venkit-nexthop: You are right. I had already taken care of resetting segment source  (field "seg_src") if it is not present in fpmsyncd.
I will make the change for srv6 as well.

venkit-nexthop: I have made this change.

dockers/docker-fpm-frr/frr/zebra/zebra.conf.j2
venkit-nexthop: I removed all the knobs now and have enabled this by default.

venkit-nexthop:
@StormLiangMS @qiluo-msft @lguohan 
Now that https://github.com/sonic-net/sonic-swss/pull/3872 has merged, this PR makes things more efficient and also avoids packet drops during ECMP growth scenario (refer to issue https://github.com/sonic-net/sonic-buildimage/issues/23940).
Appreciate it if you can review the change and approve it.

venkit-nexthop:
@lguohan could you take a look at this one and approve it. It is one of 2 PRs to resolve ISSUE: https://github.com/sonic-net/sonic-buildimage/issues/23940
The other PR (in fpmsyncd) has merged => https://github.com/sonic-net/sonic-swss/pull/3872

venkit-nexthop:
> do you know if we need to wait till the swss submodule to be advanced first?

The swss submodule has already advanced and it now contains my fix.

venkit-nexthop:
@lguohan all the checks have passed. Please take a look and approve if you are ok with the changes.

PR #23957, opened 2025-09-10
Change incorrect component name for Arista-7060-CX32S platform

rgarofano-arista:
#### Why I did it
platform_tests/api/test_component.py was failing due to the hard coded component name in the corresponding platform.json file not matching the component name retrieved from the Arista platform information API.

##### Work item tracking

#### How I did it
Changed the component name in the platform.json file to match our platform information API.

#### How to verify it
Ran the test with and without the change, confirmed that it passes with the change.

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [x] 202505

- [ ] 
- [ ]

### DIFF
--- a/device/arista/x86_64-arista_7060_cx32s/platform.json
+++ b/device/arista/x86_64-arista_7060_cx32s/platform.json
@@ -15,7 +15,7 @@
                 "name": "Ucd90120A(addr=8-004e)"
             },
             {
-                "name": "CrowSysCpld(addr=2-0023)"
+                "name": "KoiSysCpld(addr=2-0023)"
             }
         ],
         "fans": [],


device/arista/x86_64-arista_7060_cx32s/platform.json
```diff
@@ -15,7 +15,7 @@
                 "name": "Ucd90120A(addr=8-004e)"
             },
             {
-                "name": "CrowSysCpld(addr=2-0023)"
+                "name": "KoiSysCpld(addr=2-0023)"
```
lipxu, line 18: Thanks for the fix. Is this related to an image issue? if so, could you please share the impact, thank you

vvolam, line 18: @rgarofano-arista Is this a regression due to platform API changes? or something else causing this issue?

StormLiangMS:
hi @rgarofano-arista could you check the comment?

device/arista/x86_64-arista_7060_cx32s/platform.json
rgarofano-arista: This is not an image issue, we changed our platform API a few weeks ago and did not update platform.json on sonic-buildimage.

rgarofano-arista: We made some changes to the platform API recently to prevent SEU triggered reboots on this platform.
Caused by submodule update: https://github.com/sonic-net/sonic-buildimage/pull/23646
Our Change: https://github.com/aristanetworks/sonic/commit/8b72dcdaf6aba78f41b7c90dc13ce28081ea5693

vvolam: Understood. Thanks

lipxu:
@StormLiangMS @yejianquan Could you please help review the PR?  thanks a lot.

device/arista/x86_64-arista_7060_cx32s/platform.json
lipxu: Thanks, @rgarofano-arista , may I know what's the impact for production devices if without this PR, thanks

rgarofano-arista: I suppose it depends how often SCD SEU occurs, but the FPGA is susceptible to it. When these errors occur the system will automatically reboot. Our change makes it so there will be a warning message in `/var/log/arista.daemon` asking the user to manually reboot (it won't just automatically reboot).

rgarofano-arista: In terms of this PR specifically where the component name is wrong in platform.json, this will not affect reboot behavior as that is at the hardware/driver level. I think the impact is negligible.

lipxu: Thanks for your clarification, @rgarofano-arista , if the device hit the issue, then reboot in production without this log, could we determine the reboot cause? thanks

rgarofano-arista: SCD SEU triggered reboot would be shown as a gpi 3 fault on the switch card. This would be visible in `/var/log/arista.log` and `/host/reboot-cause/platform/causes.json`.

PR #23974, opened 2025-09-11
Remove SMARTSWITCH build flag across platforms

rameshraghupathy:
Remove SMARTSWITCH build flag across platforms. Include dhcp_server in the build for all platforms. The dhcp_server is enabled only if the config exists in config_db.json which is the case in SmartSwitch

Why I did it
There is a new request to have a single image across platforms and avoid a separate image for SmartSwitch. Originally this was done to exclude the dhcp_server from the build for non-smartswitch platforms as it would increase the image size. Now that the benefit of having a single image is more compared to the image size. So making this change.

Work item tracking
Microsoft ADO (number only):
How I did it
Unconditionally including of DHCP_SERVER in the build

How to verify it
Build an image for any platform and should build and run as it was prior to the change. Run the same image on a smartswitch platform and it should work seamlessly.

Which release branch to backport (provide reason below if selected)
 202205
 202211
 202305
 202311
[ x] 202405
[x ] 202411
[ x] 202505
[ x] 202506
Description
Replaces [Remove SMARTSWITCH build flag across platforms. by rameshraghupathy · Pull Request #23297 ](https://github.com/sonic-net/sonic-buildimage/pull/23297)
Please refer to https://github.com/sonic-net/sonic-buildimage/issues/21590 which also has relevance to this change


prabhataravind:
@oleksandrivantsiv @vivekrnv, could one of you please approve from Nvidia side?

liushilongbuaa:
LGTM, but I'm not sure if it is good to enable dhcp server by default.

PR #23985, opened 2025-09-12
add loopback&vlan into the ipinip tunnel list for large topo

sdszhang:
#### Why I did it
Currently there is no tunnel being created when more than 128 addresses in the system. 
This PR enables for loopback addresses in this scenario as it's still needed.

##### Work item tracking
- Microsoft ADO **34827134**:

#### How I did it
Update the tunnel address to include loopback IPv4 and IPv6 addresses.

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

- [ ] 
- [ ] 
202412

### DIFF
--- a/dockers/docker-orchagent/ipinip.json.j2
+++ b/dockers/docker-orchagent/ipinip.json.j2
@@ -61,10 +61,10 @@
 {% endfor %}
 {%- set ipv4_addresses = ipv4_addresses + ipv4_vlan_addresses %}
 {%- set ipv6_addresses = ipv6_addresses + ipv6_vlan_addresses %}
-{# FIXME: SAI report tunnel TABLE_FULL for large topo. Disable it temporarily if over 128 IPv4/IPv6 tunnels. #}
+{# SAI report tunnel TABLE_FULL for large topo. Only generating for loopback if over 128 routed interfaces.#}
 {% if ipv4_addresses|length + ipv6_addresses|length > 128 %}
-{%- set ipv4_addresses = [] %}
-{%- set ipv6_addresses = [] %}
+{%- set ipv4_addresses = ipv4_loopback_addresses %}
+{%- set ipv6_addresses = ipv6_loopback_addresses %}
 {% endif %}
 
 {# Check if AZURE QoS map exists #}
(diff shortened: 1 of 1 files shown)


dockers/docker-orchagent/ipinip.json.j2
```diff
@@ -61,10 +61,10 @@
 {% endfor %}
 {%- set ipv4_addresses = ipv4_addresses + ipv4_vlan_addresses %}
 {%- set ipv6_addresses = ipv6_addresses + ipv6_vlan_addresses %}
-{# FIXME: SAI report tunnel TABLE_FULL for large topo. Disable it temporarily if over 128 IPv4/IPv6 tunnels. #}
+{# SAI report tunnel TABLE_FULL for large topo. Only generating for VLAN and loopback if over 128 routed interfaces.#}
 {% if ipv4_addresses|length + ipv6_addresses|length > 128 %}
-{%- set ipv4_addresses = [] %}
-{%- set ipv6_addresses = [] %}
+{%- set ipv4_addresses = ipv4_loopback_addresses + ipv4_vlan_addresses %}
+{%- set ipv6_addresses = ipv6_loopback_addresses + ipv6_vlan_addresses %}
```
kperumalbfn, line 67: Please check if IPINIP_SUBNET_V6 entries are created with this change.

sdszhang: IPINIP_SUBNET_V6 entry won't be created if subnet_decap is not enabled, which is what we tested.

kperumalbfn: Thanks

r12f: Will this cause real impact on production usage? @sdszhang

sdszhang: @r12f  It should not.  The subnet_decap behavior is related to ipv4_vlan_addresses,  here we only changed ipv4_addresses value, so it will not cause any behavior change.

markx-arista:
202505 needs this fix. M1-108/128 topos failed test_decap because no tunnel is created for loopback address by default and decap config is not applied during the test for vxlan=set_unset case.

PR #24010, opened 2025-09-16
Validate yang files completeness

wen587:
##### Work item tracking

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

- [ ] 
- [ ]

### DIFF
--- a/src/sonic-yang-models/setup.py
+++ b/src/sonic-yang-models/setup.py
@@ -2,6 +2,7 @@ import shutil
 import os
 import glob
 import jinja2
+import sys
 from setuptools import setup, find_packages
 from setuptools.command.build_py import build_py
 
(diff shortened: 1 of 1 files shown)


ganglyu (approved):

vaibhavhd:
About the already missing YANG models `sonic-bgp-aggregate-address.yang, sonic-heartbeat.yang,sonic-stormond-config.yang` -- can we have the original authors of these YANG models review this PR too? I want to be double-sure that the miss was accidental and not intentional. We would need to have different conversation if this was an intentional miss.

src/sonic-yang-models/setup.py
```diff
@@ -2,74 +2,154 @@
...
+        if extra_files:
+            print(f"\nEXTRA in yang_files list ({len(extra_files)} files):")
+            for file in sorted(extra_files):
+                print(f"  - {file}")
+
+    return is_valid
+
 # List of yang file names to be included in the wheel.
 # Specify only the file basenames here; directory prefixes will be added automatically.
 yang_files = [
```
vaibhavhd, line 79: What's the reason for sorting?

src/sonic-yang-models/setup.py
```diff
@@ -78,82 +158,80 @@
...
-    'sonic-serial-console.yang',
-    'sonic-smart-switch.yang',
-    'sonic-spanning-tree.yang',
-    'sonic-srv6.yang',
 ]
 
 class my_build_py(build_py):
     def run(self):
         if not self.dry_run:
             print("hehe")
```
vaibhavhd: NIT: Can we update this too, please? `hehe` does not seem right for the logging purposes. :)

src/sonic-yang-models/setup.py
vaibhavhd, line 79: NIT - can the yang_files list can move to another file?  That would make setup.py script more readable, and potentially allow for the hardcoded list of yang_files to be consumed by other scripts (for whatever reason). What do you think?

vaibhavhd, line 79: Alternative approach - if we want to enforce that all files in `yang-models` and `yang-templates` are present in `yang_files` list then do we really need authors to define the `yang_files` in this script? Can we not self pick ALL the files in `yang-templates` and `yang-models` and put them in the build process (and to image)? That could remove this gap from the root - every new yang model will be auto picked. What do you think?

src/sonic-yang-models/setup.py
```diff
@@ -2,74 +2,154 @@
...
+        tuple: (is_valid, expected_files, missing_files, extra_files)
+    """
+    # Get all template files and convert to .yang names
+    template_files = set()
+    if os.path.exists('./yang-templates/'):
+        for template_path in glob.glob('./yang-templates/*.yang.j2'):
+            template_name = os.path.basename(template_path).replace('.j2', '')
+            template_files.add(template_name)
+
+    # Get all actual YANG files (excluding auto-generated ones)
```
vaibhavhd, line 32: How do you exclude the auto-generated ones?

src/sonic-yang-models/setup.py
wen587: The sorting is for better order that future developer can find or add yang model in alphabet order.

src/sonic-yang-models/setup.py
wen587: It was legacy code contributed by guohan. :D

src/sonic-yang-models/setup.py
wen587: It seems yang_files serves as a middle layer for collecting YANG models. With this PR adding protection, we can ensure that no YANG files are missed in the list

src/sonic-yang-models/setup.py
wen587: The auto generated are ending with .j2. We have collected it here: https://github.com/sonic-net/sonic-buildimage/blob/303b733c08a59341f94a274eb520ad32f898ed75/src/sonic-yang-models/setup.py#L28-L30

and joined in line40.

w1nda (approved):

wen587:
> About the already missing YANG models `sonic-bgp-aggregate-address.yang, sonic-heartbeat.yang,sonic-stormond-config.yang` -- can we have the original authors of these YANG models review this PR too? I want to be double-sure that the miss was accidental and not intentional. We would need to have different conversation if this was an intentional miss.

Added yang owner to reviewer to get approval.

src/sonic-yang-models/setup.py
wen587: It should be fine. This is yang model file list which should only be used in the completeness test.

src/sonic-yang-models/setup.py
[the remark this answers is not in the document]
qiluo-msft: this is valid comment, please fix.

src/sonic-yang-models/setup.py
```diff
@@ -2,74 +2,154 @@
...
+    Validate that all YANG files are included in the yang_files list.
+    
+    This function checks:
+    1. All .yang files in yang-models/ directory (excluding auto-generated ones)
+    2. All .yang.j2 template files in yang-templates/ directory
+    3. Compares the total count with the yang_files list
+    
+    Returns:
+        tuple: (is_valid, expected_files, missing_files, extra_files)
+    """
```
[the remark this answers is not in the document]
qiluo-msft: @wen587 How did you "excluding" by code?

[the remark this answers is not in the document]
wen587: excluding is j2 fiels. So just pick up yang-models/ dir will exluding auto generated ones.

ashwnsri:
@wen587 Thanks for the catch, the stormond YANG file miss was unintentional.

ashwnsri (approved):

vaibhavhd:
Thank you, @ashwnsri, for the confirmation and to be back!

Issue #24004, opened 2025-09-15
Regression: hwsku.json subport and optional attributes incorrectly applied to all child ports instead of specific child port


DavidZagury:
### Is it platform specific

generic

### Importance or Severity

Critical

### Previous Working Version

202411,Master - d58f286799bdfb35ea06b16fb0477841ce330c91

### Steps to Reproduce

Use sonic-cfggen to create a DB which have subports configured on the hwsku.json file

### Impact of this regression

DB is created with wrong values for the subport field for each port.
e.g.
 ```
   "PORT": {
        "Ethernet0": {
            "alias": "etp1a",
            "lanes": "0,1,2,3",
            "speed": "200000",
            "index": "1",
            "subport": "2",
            "autoneg": "off"
        },
        "Ethernet4": {
            "alias": "etp1b",
            "lanes": "4,5,6,7",
            "speed": "200000",
            "index": "1",
            "subport": "2",
            "autoneg": "off"
        },
        "Ethernet8": {
            "alias": "etp2a",
            "lanes": "8,9,10,11",
            "speed": "200000",
            "index": "2",
            "subport": "2",
            "autoneg": "off"
        },
        "Ethernet12": {
            "alias": "etp2b",
            "lanes": "12,13,14,15",
            "speed": "200000",
            "index": "2",
            "subport": "2",
            "autoneg": "off"
        },
...
}
```
As we can see, all subport are with a value of "2".

This was started after the merge of https://github.com/sonic-net/sonic-buildimage/pull/23600

### Relevant log output

```shell

```

### Output of `show version`, `show techsupport`

```shell

```

bingwang-ms:
@kewei-arista Can you help take a look at this issue?

kewei-arista:
What HWSKU does this affect? This was introduced to FIX the subport configuration for some HWSKUs.

DavidZagury:
@kewei-arista I believe it should happen to any SKU that has splitted ports and has subport hardcoded in the hwsku.json file
For example I can see it in [Mellanox-SN4700-C128
](https://github.com/sonic-net/sonic-buildimage/tree/master/device/mellanox/x86_64-mlnx_msn4700-r0/Mellanox-SN4700-C128)

kewei-arista:
@DavidZagury 

I think my change makes sense as it fixed the issue where `subport` in `hwsku.json` was not honored. For example, if we have following subports in `hwsku.json` for some special usage:

```
        "Ethernet0": {
[...]

PR #24034, opened 2025-09-17
Fix how the optional fields are handled in portconfig.py

kewei-arista:
#### Why I did it

As tracked in https://github.com/sonic-net/sonic-buildimage/issues/24004, there's an issue with handling optional fields in `portconfig.py` that may cause the value of one child port applied to all the child ports in the same group, even each child port has its own individual value.

##### Work item tracking

#### How I did it

Updated logic to make sure the optional values are applied in the following order:
- if a child ports has its own value defined, force using that value
- otherwise, fallback to the existing behavior that apply the value of a child port to all the child ports in the same group if they have no corresponding values defined

I also updated the test case to make sure we have such coverage, which was missing initially.

#### How to verify it

Confirmed the original issue is fixed, e.g.

```
{
    "DEVICE_METADATA": {
        "localhost": {
            "hwsku": "Mellanox-SN4700-C128"
        }
    },
    "PORT": {
        "Ethernet0": {
            "alias": "etp1a",
            "lanes": "0,1",
            "speed": "100000",
            "index": "1",
            "subport": "1"
        },
        "Ethernet2": {
            "alias": "etp1b",
            "lanes": "2,3",
            "speed": "100000",
            "index": "1",
            "subport": "2"
        },
        "Ethernet4": {
            "alias": "etp1c",
            "lanes": "4,5",
            "speed": "100000",
            "index": "1",
            "subport": "3"
        },
        "Ethernet6": {
            "alias": "etp1d",
            "lanes": "6,7",
            "speed": "100000",
            "index": "1",
            "subport": "4"
        },
<---- output truncated ---->
```

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [x] 202505
- [x] msft-202412
- [x] msft-202503

- [x] 202505


sdszhang:
@prgeor @mihirpat1 can you help to review? Thanks!

kewei-arista:
> This works and fixed the problem we have, with it we can now use different values to the different subports.
> 
> The only behavioral question is, with this change, if some subports have an optional field and others don't - we force the ones that were not set with a value to have a value and there is still an uncertainty on what the value would be on the other ports.
> 
> If there are two subport, and one has autoneg set to off, the other will automatically be set to off. If there are three subports, and one is set to off, the other set to on, the two other ports will be set according to the first port. Is that the behavior we want?
> 
> Should we just not force other subport to have the optional value just because one has it? I am not sure that this would be the better behavior so I tend to approve this as is, but wanted to raise this question any way. @qiluo-msft @kewei-arista

@DavidZagury 

> we force the ones that were not set with a value to have a value 

I think this behavior was introduced in https://github.com/sonic-net/sonic-buildimage/pull/20313 so I kept it there for backward compatibility. Regarding uncertainty, we always can fix `hwsku.json` to make sure values are correctly added to all child ports in that group.

sdszhang (approved):
lgtm

kewei-arista:
Can someone help merge this PR?

bingwang-ms:
@prgeor Can you also help review? Thanks

src/sonic-config-engine/portconfig.py
```diff
@@ -441,8 +441,13 @@ def parse_platform_json_file(hwsku_json_file, platform_json_file):
             if child_port in hwsku_entry:
                 for key, item in hwsku_entry[child_port].items():
                     if key in OPTIONAL_HWSKU_ATTRIBUTES:
+                        # Always force to the values in hwsku.json if they are available.
```
prgeor, line 444: @kewei-arista @DavidZagury 
The logic for how the `subport` is derived has NOT changed since `202311`.

The _**starting**_ subport was always used to assign the subsequent subport values for remaining breakout ports belonging to the SAME physical port

202311: 
1. https://github.com/sonic-net/sonic-buildimage/blob/0318f8ac89a31745db107884e9ab502688f6abf3/src/sonic-config-engine/portconfig.py#L433
2. https://github.com/sonic-net/sonic-buildimage/blob/0318f8ac89a31745db107884e9ab502688f6abf3/src/sonic-config-engine/portconfig.py#L382

kewei-arista: `subport` was not added to `OPTIONAL_HWSKU_ATTRIBUTES` in:
https://github.com/sonic-net/sonic-buildimage/blob/0318f8ac89a31745db107884e9ab502688f6abf3/src/sonic-config-engine/portconfig.py#L40

but now it is: 
https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-config-engine/portconfig.py#L40

This changed the behavior when `subport` is defined in hwsku.json.

DavidZagury:
> @kewei-arista @DavidZagury I don't see we need this PR as explained here:
> 
> #24004

@prgeor This PR is needed, I have added my comments on the original issue.

dprital:
@prgeor , This fix is required for 202511 as well.

bingwang-ms:
@prgeor Kindly remind on this review request. Thanks

kewei-arista:
Superseded by https://github.com/sonic-net/sonic-buildimage/pull/25335

PR #24053, opened 2025-09-19
Integrate SHOW API probe into telemetry watchdog

lixiaoyuner:
#### Why I did it
Integrate SHOW API probe into telemetry watchdog. We want the watchdog to catch the bad SHOW command implementation.

##### Work item tracking
- Microsoft ADO **(number only)**: 34793098

#### How I did it
Add gnmi_get into telemetry-watchdog, and call gnmi_get in the watchdog http service for each probe, as long as one xpath fails, the probe will return 500. 
- The watchdog will read xpath list from /cmd_list.json and environment variables.

#### How to verify it
Run the telemetry-watchdog container, call curl http://127.0.0.1:50080/ to check if the output is expected or not.
```
root@bjw-can-2700-2:/# curl -s http://127.0.0.1:50080/ |jq .
{
  "check_telemetry_port": "OK",
  "xpath_commands": [
    {
      "xpath": "clock",
      "success": true,
      "error": null,
      "duration_ms": 200
    }
  ]
}
```

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

- [ ] 
- [ ]


dockers/docker-telemetry-watchdog/Dockerfile.j2
```diff
@@ -1,5 +1,5 @@
 {% from "dockers/dockerfile-macros.j2" import install_debian_packages, install_python_wheels, copy_files %}
-ARG BASE=docker-config-engine-bookworm-{{DOCKER_USERNAME}}:{{DOCKER_USERTAG}}
+ARG BASE=docker-sonic-gnmi-{{DOCKER_USERNAME}}:{{DOCKER_USERTAG}}
```
make1980, line 2: nit:

Integrate SHOW API probe into telemetry watchdog. We want the watch dog can catch the bad SHOW command implement.

->

Integrate SHOW API probe into telemetry watchdog. We want the watchdog to catch the bad SHOW command implementation.

FengPan-Frank, line 2: better to add DEPENDS in rule to make sure GNMI is built first for this.

dockers/docker-telemetry-watchdog/watchdog/src/main.rs
```diff
@@ -10,6 +14,221 @@ const DEFAULT_TELEMETRY_SERVICE_PORT: u16 = 50051;
...
+            #[derive(serde::Deserialize)]
+            struct JsonCfg { xpaths: Option<Vec<String>> }
+            match serde_json::from_str::<JsonCfg>(&content) {
+                Ok(cfg) => {
+                    if let Some(mut xs) = cfg.xpaths { list.append(&mut xs); }
+                },
+                Err(e) => eprintln!("Failed to parse {}: {}", CMD_LIST_JSON, e),
+            }
+        },
+    Err(e) => eprintln!("Could not read {}: {} (will continue with env var only)", CMD_LIST_JSON, e),
```
FengPan-Frank: should this kind of error be visible to user as well? return in json resp

dockers/docker-telemetry-watchdog/watchdog/src/main.rs
```diff
@@ -10,6 +14,221 @@ const DEFAULT_TELEMETRY_SERVICE_PORT: u16 = 50051;
...
+    }
+
+    if let Ok(env_val) = env::var(XPATH_ENV_VAR) {
+        for part in env_val.split(',') { let trimmed = part.trim(); if !trimmed.is_empty() { list.push(trimmed.to_string()); } }
+    }
+
+    // dedupe while preserving order
+    let mut seen = std::collections::HashSet::new();
+    list.retain(|x| seen.insert(x.clone()));
```
make1980: why don't insert to a hashset above directly? #Closed

dockers/docker-telemetry-watchdog/watchdog/src/main.rs
```diff
@@ -10,6 +14,221 @@ const DEFAULT_TELEMETRY_SERVICE_PORT: u16 = 50051;
...
+    if let Ok(blacklist) = env::var(XPATH_ENV_BLACKLIST) {
+        let mut blk = std::collections::HashSet::new();
+        for part in blacklist.split(',') {
+            let trimmed = part.trim();
+            if !trimmed.is_empty() { blk.insert(trimmed.to_string()); }
+        }
+        if !blk.is_empty() {
+            list.retain(|x| !blk.contains(x));
+        }
+    }
```
make1980, line 104: do we really need this? #Closed

dockers/docker-telemetry-watchdog/watchdog/src/main.rs
```diff
@@ -10,6 +14,221 @@ const DEFAULT_TELEMETRY_SERVICE_PORT: u16 = 50051;
...
+
+struct TelemetrySecurityConfig {
+    use_client_auth: bool,
+    ca_crt: String,
+    server_crt: String,
+    server_key: String,
+}
+
+fn run_gnmi_for_xpath(xpath: &str, port: u16, sec: &TelemetrySecurityConfig, target_name: &str, timeout: Duration) -> CommandResult {
+    // Build full command: gnmi_get -xpath_target SHOW -xpath <xpath> -target_addr 127.0.0.1:<port> -logtostderr -insecure true
```
make1980: longer than 120 characters #Closed

dockers/docker-telemetry-watchdog/watchdog/src/main.rs
```diff
@@ -10,6 +14,221 @@ const DEFAULT_TELEMETRY_SERVICE_PORT: u16 = 50051;
...
+
+fn get_security_config() -> TelemetrySecurityConfig {
+    // Redis DB 4 hashes:
+    // TELEMETRY|certs: ca_crt, server_crt, server_key
+    // TELEMETRY|gnmi: client_auth, target_name (target_name new; if absent we still proceed)
+    let client = match redis::Client::open("redis://127.0.0.1:6379/4") {
+        Ok(c) => c,
+        Err(e) => {
+            eprintln!("Redis client error (security): {e}");
+            return TelemetrySecurityConfig { use_client_auth: false, ca_crt: DEFAULT_CA_CRT.to_string(), server_crt: DEFAULT_SERVER_CRT.to_string(), server_key: DEFAULT_SERVER_KEY.to_string() };
```
make1980: >TelemetrySecurityConfig { use_client_auth: false, ca_crt: DEFAULT_CA_CRT.to_string(), server_crt: DEFAULT_SERVER_CRT.to_string(), server_key: DEFAULT_SERVER_KEY.to_string() }; [](http://example.com/codeflow?start=19&length=175)

nit: we should just construct this once and use it in all the failure cases #Closed

dockers/docker-telemetry-watchdog/watchdog/src/main.rs
```diff
@@ -10,6 +14,221 @@ const DEFAULT_TELEMETRY_SERVICE_PORT: u16 = 50051;
...
+        Err(e) => {
+            eprintln!("Redis connection error (security): {e}");
+            return TelemetrySecurityConfig { use_client_auth: false, ca_crt: DEFAULT_CA_CRT.to_string(), server_crt: DEFAULT_SERVER_CRT.to_string(), server_key: DEFAULT_SERVER_KEY.to_string() };
+        }
+    };
+
+    let mut get_field = |hash: &str, field: &str| -> Option<String> {
+        let r: redis::RedisResult<Option<String>> = conn.hget(hash, field);
+        match r { Ok(v) => v, Err(e) => { eprintln!("Redis HGET error {hash}.{field}: {e}"); None } }
+    };
```
make1980: nit: this pattern has appeared in this file multiple times - can we refactor it to a common function? #Closed

dockers/docker-telemetry-watchdog/watchdog/src/main.rs
```diff
@@ -108,22 +327,32 @@ fn main() {
...
+                            let port = get_gnmi_port();
+                            let sec_cfg = get_security_config();
+                            let timeout = read_timeout();
+                            let target_name = get_target_name();
+                            for xp in xpaths {
+                                let res = run_gnmi_for_xpath(&xp, port, &sec_cfg, &target_name, timeout);
+                                if !res.success { http_status = "HTTP/1.1 500 Internal Server Error"; }
+                                cmd_results.push(res);
+                            }
+                        }
```
make1980, line 384: could you introduce another env var to enable/disable this? #Closed

make1980 (changes requested):
:clock1:

dockers/docker-telemetry-watchdog/Dockerfile.j2
lixiaoyuner: > nit:
> 
> Integrate SHOW API probe into telemetry watchdog. We want the watch dog can catch the bad SHOW command implement.
> 
> ->
> 
> Integrate SHOW API probe into telemetry watchdog. We want the watchdog to catch the bad SHOW command implementation.

Thanks, fixed.

lixiaoyuner: > better to add DEPENDS in rule to make sure GNMI is built first for this.

Thanks, fixed

dockers/docker-telemetry-watchdog/watchdog/src/main.rs
lixiaoyuner: > do we really need this?

I'd like to keep this, we may encounter a case that need to ignore some API, we can update the environment variables via k8s yaml file which will be very quick, no need to wait for code change.

dockers/docker-telemetry-watchdog/watchdog/src/main.rs
lixiaoyuner: Added a SHOW API probe switch and enable probe as default.

dockers/docker-telemetry-watchdog/watchdog/src/main.rs
lixiaoyuner: > should this kind of error be visible to user as well? return in json resp

Put the error into JSON resp

dockers/docker-telemetry-watchdog/watchdog/src/main.rs
lixiaoyuner: > why don't insert to a hashset above directly?

Yes, it's better to use a HashSet directly, updated the code.

dockers/docker-telemetry-watchdog/watchdog/src/main.rs
lixiaoyuner: > longer than 120 characters

Thanks, fixed all the lines which are too long.

dockers/docker-telemetry-watchdog/watchdog/src/main.rs
lixiaoyuner: Thanks, fixed.

> > TelemetrySecurityConfig { use_client_auth: false, ca_crt: DEFAULT_CA_CRT.to_string(), server_crt: DEFAULT_SERVER_CRT.to_string(), server_key: DEFAULT_SERVER_KEY.to_string() };
> 
> nit: we should just construct this once and use it in all the failure cases

dockers/docker-telemetry-watchdog/watchdog/src/main.rs
lixiaoyuner: > nit: this pattern has appeared in this file multiple times - can we refactor it to a common function?

Thanks, refactored.

dockers/docker-telemetry-watchdog/watchdog/src/cmd_list.json
```diff
@@ -0,0 +1,60 @@
+{
+    "xpaths": [
```
zbud-msft, line 2: Issue with this is that 

1) We arent testing options with the xpath in this case, user can pass in options for example --interface.
2) There are required arguments for some commands that need to be passed in, and these are usually like interface name. I dont know if the best way would to be hardcode an interface to check for all devices like Ethernet0, or if we can grab an oper up/admin up one from the host?

dockers/docker-telemetry-watchdog/watchdog/src/main.rs
```diff
@@ -10,36 +15,294 @@ const DEFAULT_TELEMETRY_SERVICE_PORT: u16 = 50051;
...
+                if let Some(xs) = cfg.xpaths { for x in xs { if !x.is_empty() { set.insert(x); } } }
+            },
+            Err(e) => {
+                let msg = format!("Failed to parse {CMD_LIST_JSON}: {e}");
+                errors.push(msg);
+            },
+        }
+    } else {
+        if let Err(e) = fs::read_to_string(CMD_LIST_JSON) {
+            let msg = format!(
```
make1980, line 82: why do we need to read the file again?

dockers/docker-telemetry-watchdog/watchdog/src/main.rs
```diff
@@ -10,36 +15,294 @@ const DEFAULT_TELEMETRY_SERVICE_PORT: u16 = 50051;
...
+            if status.success() {
+                CommandResult {
+                    xpath: xpath.to_string(),
+                    success: true,
+                    error: None,
+                    duration_ms: dur,
+                }
+            } else {
+                // Possibly large stderr; truncate if huge (optional threshold 16KB).
+                let mut truncated = if stderr_string.len() > 16 * 1024 {
```
make1980: >16 * 1024 [](http://example.com/codeflow?start=61&length=9)

please define this as a constant

dockers/docker-telemetry-watchdog/watchdog/src/main.rs
lixiaoyuner: > why do we need to read the file again?

That's indeed a bad implementation, fixed.

dockers/docker-telemetry-watchdog/watchdog/src/main.rs
lixiaoyuner: > > 16 * 1024
> 
> please define this as a constant

Fixed.

dockers/docker-telemetry-watchdog/watchdog/src/cmd_list.json
lixiaoyuner: > Issue with this is that
> 
> 1. We arent testing options with the xpath in this case, user can pass in options for example --interface.
> 2. There are required arguments for some commands that need to be passed in, and these are usually like interface name. I dont know if the best way would to be hardcode an interface to check for all devices like Ethernet0, or if we can grab an oper up/admin up one from the host?

You are right, the commands with parameters are complicated, we don't plan to support for now, we support these common xpath at this stage. We can add more xpath with specific fixed parameters as you mentioned if we need later.

r12f:
hi @lixiaoyuner , do you mind to help with a manual pick to 202412? the auto pick detected some conflicts on the code:

image

make1980:
@Feng ***@***.***> - could you help with this?

Thanks,
Michael

FengPan-Frank:
@r12f  @make1980  For this PR, I have already cherrypicked it into 202412 manually

[202412] [PR:24053] Integrate SHOW API probe into telemetry watchdog by FengPan-Frank · Azure/sonic-buildimage-msft<https://github.com/Azure/sonic-buildimage-msft/pull/1708>


Thanks,
Feng


r12f:
thanks! updated tag.

PR #24056, opened 2025-09-19
Add host filesystem access to gnmi container for upgrade monitoring

hdwhdw:
Mount entire host filesystem at /mnt/host:ro to enable upgrade handler disk space monitoring functionality. This allows the upgrade handler to check available disk space on any filesystem path before firmware updates.

##### Work item tracking

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

- [ ] 
- [ ]

### DIFF
--- a/rules/docker-gnmi.mk
+++ b/rules/docker-gnmi.mk
@@ -34,5 +34,6 @@ $(DOCKER_GNMI)_RUN_OPT += -t
 $(DOCKER_GNMI)_RUN_OPT += -v /etc/sonic:/etc/sonic:ro
 $(DOCKER_GNMI)_RUN_OPT += -v /etc/localtime:/etc/localtime:ro
 $(DOCKER_GNMI)_RUN_OPT += -v /var/run/dbus:/var/run/dbus:rw
+$(DOCKER_GNMI)_RUN_OPT += -v /:/mnt/host:ro
 
 $(DOCKER_GNMI)_BASE_IMAGE_FILES += monit_gnmi:/etc/monit/conf.d


hdwhdw:
@ndas7 Hey Neha FYI for the change which might be helpful for other gNOI API.

ganglyu (approved):

rules/docker-gnmi.mk
```diff
@@ -34,5 +34,10 @@ $(DOCKER_GNMI)_RUN_OPT += -t
 $(DOCKER_GNMI)_RUN_OPT += -v /etc/sonic:/etc/sonic:ro
 $(DOCKER_GNMI)_RUN_OPT += -v /etc/localtime:/etc/localtime:ro
 $(DOCKER_GNMI)_RUN_OPT += -v /var/run/dbus:/var/run/dbus:rw
+$(DOCKER_GNMI)_RUN_OPT += -v /:/mnt/host:ro
```
vaibhavhd, line 38: Can you add some comments here to specify purpose of requiring mount of individual paths here?
1. / - To check disk space and monitor usage
2. /tmp - To download the image binary and other upgrade package.
3. ....

vaibhavhd: Why is `/var/tmp/` needed to be mounted?

hdwhdw: Done. 

for /var/tmp, this is in case when we don't want to use /tmp for download and need something persistent. /var/tmp is better than /host.

rules/docker-gnmi.mk
```diff
@@ -34,5 +34,20 @@ $(DOCKER_GNMI)_RUN_OPT += -t
...
+# For disk space monitoring.
+$(DOCKER_GNMI)_RUN_OPT += -v /:/mnt/host:ro
+# For sonic binary image download.
+$(DOCKER_GNMI)_RUN_OPT += -v /tmp:/mnt/host/tmp:rw
+# For sonic binary image downloads to persistent file system.
+$(DOCKER_GNMI)_RUN_OPT += -v /var/tmp:/mnt/host/var/tmp:rw
+# For host command execution in gnoi.
+$(DOCKER_GNMI)_RUN_OPT += --pid=host
+# For host command execution in gnoi.
+$(DOCKER_GNMI)_RUN_OPT += --privileged
```
qiluo-msft, line 46: >privileged [](http://example.com/codeflow?start=28&length=10)

Could you check [https://github.com/sonic-net/SONiC/pull/1364](https://github.com/sonic-net/SONiC/pull/1364) ? Is it possible to use fine granularity linux capacities?

PR #24061, opened 2025-09-20
Update telemetry entry sh to support k8s start and keep compatible with systemd start flow

FengPan-Frank:
#### Why I did it
Update telemetry entry sh to support k8s start and keep compatible with systemd start flow

#### How I did it
With k8s start, it will contains below env

sudo docker run -d --rm -it --privileged \
  --net=host --pid=host --ipc=host \
  --tmpfs /usr/share/sonic:rw \
  -v /usr/share/sonic/device:/usr/share/sonic/device:ro \
  -v /etc/sonic:/etc/sonic:ro \
  -v /etc/localtime:/etc/localtime:ro \
  -v /host/reboot-cause:/host/reboot-cause \
  -v /var/run/redis:/var/run/redis \
  -v /var/run/redis-chassis:/var/run/redis-chassis:ro \
  -v /usr/share/sonic/templates/rsyslog-container.conf.j2:/usr/share/sonic/templates/rsyslog-container.conf.j2:ro \
  -e RUNTIME_OWNER=kube \
  -e SYSLOG_TARGET_IP=127.0.0.1 \
  -e NAMESPACE_PREFIX=asic \
  -e launch_by=k8s \
  docker-sonic-telemetry:latest

by default it will not be used so that existing rollout mgmt test will not be break.

#### How to verify it
image

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

- [ ] 
- [ ]


dockers/docker-sonic-telemetry/docker-telemetry-entry.sh
```diff
@@ -1,8 +1,44 @@
 #!/usr/bin/env bash
 
+# Control with env var launch_by, only used if container is launched via k8s
+if [[ "${launch_by:-}" == "k8s" ]]; then
```
make1980, line 4: how do you pass in the environment variable? this is started by supervisord but I assume the environment variable set by k8s will also be available? #Closed

FengPan-Frank: This env will not be set by default so that existing mgmt test will keep working as originally, when k8s rollout telemetry image this env will be set via daemonset config

dockers/docker-sonic-telemetry/docker-telemetry-entry.sh
```diff
@@ -1,8 +1,44 @@
...
+
+  mkdir -p /usr/share/sonic
+
+  DST="/usr/share/sonic/platform"
+  rm -rf -- "$DST"
+  ln -s -- "$SRC_PLATFORM" "$DST"
+
+  DST="/usr/share/sonic/hwsku"
+  rm -rf -- "$DST"
+  ln -s -- "$SRC_HWSKU" "$DST"
```
qiluo-msft, line 30: why you need to create symlinks? Is it relating to k8s features? #Closed

FengPan-Frank: not for k8s feature, this was used in currently telemetry.sh https://github.com/sonic-net/sonic-buildimage/blob/1571dd06c24981acddd9465d396c77bbbfaba591/files/build_templates/docker_image_ctl.j2#L664, we only try to launch telemetry with the same parameter/mount/env but from k8s instead of systemd path.

make1980: k8s can only have the volume name as static instead of generating volume based on label of the node - so we create the symlink at run time based on the node's sku/platform.

qiluo-msft: The 2 DST folders above are RO mounted folder. I do not understand why you need to `rm` them and create symlink.

FengPan-Frank: added this logic to first remove the symlink/file/directory at DST in case they're not expected and then run "ln -s" just once.

make1980: @qiluo-msft  - the DST folders are not mounted. the mounted folder is /usr/share/sonic/device/

dockers/docker-sonic-telemetry/docker-telemetry-entry.sh
```diff
@@ -1,8 +1,44 @@
...
+
+  # --- Create /usr/share/sonic/hwsku symlink (single-ASIC) ---
+  ENV_FILE="/etc/sonic/sonic-environment"
+  [[ -r "$ENV_FILE" ]] || { echo "ERROR: $ENV_FILE missing or unreadable."; exit 1; }
+  # shellcheck source=/dev/null
+  . "$ENV_FILE"
+  [[ -n "${PLATFORM:-}" && -n "${HWSKU:-}" ]] || { echo "ERROR: PLATFORM/HWSKU not set in $ENV_FILE."; exit 1; }
+
+  resolve() { for p in "$@"; do [[ -e "$p" ]] && { printf '%s\n' "$p"; return 0; }; done; return 1; }
+  SRC_PLATFORM="$(resolve "/usr/share/sonic/device/${PLATFORM}" "/usr/share/sonic/device/device/${PLATFORM}")" \
```
make1980: >/usr/share/sonic/device/device/${PLATFORM} [](http://example.com/codeflow?start=65&length=42)

what's this? shouldn't we remove it? #Closed

FengPan-Frank: seems indeed redundant check, removed.

qiluo-msft (approved):

PR #24064, opened 2025-09-21
[Mellanox] Add CPO support and new sensor

tshalvi:
#### Why I did it
1. Add platform API level support for CPO
2. Add new sensors support

##### Work item tracking

#### How I did it
1. Create a CPO class to represent CPO module
2. Add new sensors support

#### How to verify it
1. Unit test passed
2. Manual test passed

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

- [ ] 
- [ ]

### DIFF
--- a/platform/mellanox/mlnx-platform-api/sonic_platform/device_data.py
+++ b/platform/mellanox/mlnx-platform-api/sonic_platform/device_data.py
@@ -229,13 +229,22 @@ class DeviceDataManager:
         version = check_output_pipe(["lspci", "-vv"], ["grep", "SimX"])
         parsed_version = re.search("([0-9]+\\.[0-9]+-[0-9]+)", version)
         return parsed_version.group(1) if parsed_version else "N/A"
+    
+    @classmethod
+    @utils.read_only_cache()
+    def get_fan_drawer_sysfs_count(cls):
+        return len(glob.glob('/run/hw-management/thermal/fan*_status'))
 
     @classmethod
     @utils.read_only_cache()
     def get_fan_drawer_count(cls):
         # Here we don't read from /run/hw-management/config/hotplug_fans because the value in it is not
         # always correct.
-        return len(glob.glob('/run/hw-management/thermal/fan*_status')) if cls.is_fan_hotswapable() else 1
+        fan_status_count = cls.get_fan_drawer_sysfs_count()
+        if fan_status_count == 0:
+            # For system with no fan, for example, liquid cooling system.
+            return 0
+        return fan_status_count if cls.is_fan_hotswapable() else 1
 
     @classmethod
     @utils.read_only_cache()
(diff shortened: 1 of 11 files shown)


Junchao-Mellanox (approved):

platform/mellanox/mlnx-platform-api/sonic_platform/device_data.py
```diff
@@ -229,13 +229,22 @@ def get_simx_version(cls):
...
+    
+    @classmethod
+    @utils.read_only_cache()
+    def get_fan_drawer_sysfs_count(cls):
+        return len(glob.glob('/run/hw-management/thermal/fan*_status'))
 
     @classmethod
     @utils.read_only_cache()
     def get_fan_drawer_count(cls):
         # Here we don't read from /run/hw-management/config/hotplug_fans because the value in it is not
```
keboliu, line 241: Please remove these comments

platform/mellanox/mlnx-platform-api/sonic_platform/sfp.py
```diff
@@ -1811,3 +1816,159 @@ def get_module_status(self):
...
+        return present == 1
+
+    def reinit(self):
+        """
+        Nothing to do for cpo. Just provide it to avoid exception
+        :return:
+        """
+        return
+
+    def read_eeprom(self, offset, num_bytes):
```
keboliu: Why not reuse the existing implementation? Is there anything special for CPO that would require a different EEPROM implementation? The same comment goes to the functions below in this new class.

platform/mellanox/mlnx-platform-api/sonic_platform/utils.py
```diff
@@ -272,6 +273,42 @@ def extract_RJ45_ports_index():
     return RJ45_port_index_list if bool(RJ45_port_index_list) else None
 
 
+def extract_cpo_ports_index():
```
keboliu: The logic of this function is mostly the same as extract_RJ45_ports_index(), can we have a common function with some parameters so it can serve both extract RJ45 ports and CPO ports?

platform/mellanox/mlnx-platform-api/sonic_platform/sfp.py
tshalvi: The original separation was intentional since CpoPort isn’t conceptually an SFP.
Following your suggestion, I’ve now moved the EEPROM access functions to the parent class, NvidiaSFPCommon, and removed them from both SFP and CpoPort, so the common logic is reused while keeping the class hierarchy semantically correct.

platform/mellanox/mlnx-platform-api/sonic_platform/utils.py
tshalvi: This function is planned to be updated as part of the upcoming multi-ASIC work, so it will be refactored anyway. I’ll also check whether the multi-ASIC addressing is relevant for RJ45 ports, and make sure the common logic is consolidated as part of that change if needed.

PR #24068, opened 2025-09-22
Modified platform to support MtFuji rev 2 board and bug fixes

SahilChaudhari:
- Added get_revision support for mtfuji board types
- Added sensor mapping for mtfuji rev 2
- Fixed fwutil show status output for sonic image version
- Fixed show system-health output for unused containers by disabling state in config_db.json FEATURE table
- Fixed ssh connectivity issue while doing sonic installation via ssh

##### Work item tracking

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

- [ ] 
- [ ]

### DIFF
--- a/files/dsc/dpu.init
+++ b/files/dsc/dpu.init
@@ -60,6 +60,8 @@ function start_polaris()
     echo $IMAGE_NAME:$TAG > $DPU_DOCKER_INFO_DIR/image
     /usr/sbin/ethtool -K Ethernet0 tx off sg off tso off rx off
 
+    echo 50 > $HOST_DIR_POLARIS/data/free_mem_threshold
+
     docker ps -a --format "{{.ID}}\t{{.Image}}" | grep "$IMAGE_NAME:$TAG" | awk '{print $1}' | xargs -I {} docker rm {}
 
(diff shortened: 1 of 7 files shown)


platform/pensando/sonic-platform-modules-dpu/dpu/utils/dpu_pensando_util.py
```diff
@@ -84,6 +88,19 @@ def replace_keys(obj):
     except Exception as e:
         log_err("failed to setup platform_components.json due to {}".format(e))
 
+def disable_unused_containers():
+    try:
+        import redis
+
+        r = redis.StrictRedis(host=LOCALHOST, port=REDIS_PORT, db=CHSSIS_DB, decode_responses=True)
```
vvolam: Is it CHASSIS_DB?

SahilChaudhari: It is CONFIG_DB, fixed it. Thanks

files/dsc/dpu.init
```diff
@@ -75,11 +77,8 @@ function start_polaris()
     sleep 5
     INTERFACE="eth0-midplane"
     if ip link show "$INTERFACE" &> /dev/null; then
-        echo "dhclient -r $INTERFACE"
```
prabhataravind, line 78: hi @SahilChaudhari could you add some details on why are we changing from dhclient to networkctl? Could you also confirm this works with upgrades from previous image version to a version with this change?

seveerap79: We tested the upgrade from previous version to this new version; upgrade was successful and eth0-midplane DHCP lease allocation was proper.

PR #24155, opened 2025-09-29
Enable kubectl execution in sidecar

FengPan-Frank:
#### Why I did it
1. Existing systemd script has issue in triggering kubectl for restarting container service, because telemetry systemd service is launched via admin user, which cannot access to kubectl related config file and cert.
2. container_checker has issue in triggering kubectl for filter running_container to append the container rollouted via kubernetes, change the implementation logic to fix it

##### Work item tracking

#### How I did it
1. Change telemetry.service to be run by root user.
2. Use sidecar container to overwrite /lib/systemd/system/telemetry.service. so that kubectl could be triggered in systemd script, which is used for container restart via pod grep/delete.
4. Change container_checker logic to add more container into running_containers set, the container rollouted via kubernetes has io.kubernetes.pod.namespace=sonic, this change is to make sure monit can still return succeed for periodical FEATURE table and running container check.

#### How to verify it
Verified on testbed.
image

image

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

- [ ] 
- [ ]


dockers/docker-telemetry-sidecar/systemd_stub.py
```diff
@@ -45,8 +43,22 @@ class SyncItem:
...
+cat > /etc/sudoers.d/telemetry-kubectl <<'EOF'
+User_Alias TELEMETRY=admin
+Cmnd_Alias KCTL=/usr/bin/kubectl --kubeconfig=/etc/kubernetes/kubelet.conf *
+TELEMETRY ALL=(root) NOPASSWD: KCTL
+Defaults!KCTL !requiretty
+EOF
+chown root:root /etc/sudoers.d/telemetry-kubectl
+chmod 0440 /etc/sudoers.d/telemetry-kubectl
+visudo -cf /etc/sudoers.d/telemetry-kubectl
+"""],
```
make1980: why don't we just change the owner of /etc/kubernetes/kubelet.conf and kubectl?

dockers/docker-telemetry-sidecar/systemd_stub.py
```diff
@@ -2,7 +2,6 @@
 from __future__ import annotations
```
make1980, line 2: please add more details in PR description - for example why do we need changes in container_checker? we don't have the label for k8s container? #Closed

files/image_config/monit/container_checker
```diff
@@ -160,27 +163,55 @@ def get_current_running_from_DB(always_running_containers):
...
-        print(f"Failed to retrieve the running container list. Error: '{err}'")
-        pass
+        print(f"Failed to retrieve Docker containers. Error: '{err}'")
+    except docker.errors.DockerException as err:
+        print(f"Docker client error: '{err}'")
+
+    # ---- Kubernetes ----
+    try:
+        node_name = socket.gethostname()
+        cmd = [
```
make1980: we can use this command to get the raw container name:

docker ps -q | xargs docker inspect --format '{{.Id}} {{ index .Config.Labels "io.kubernetes.container.name"}}'

should be much easier.

dockers/docker-telemetry-sidecar/systemd_stub.py
FengPan-Frank: updated to use chown

files/image_config/monit/container_checker
FengPan-Frank: Updated with test as well.


image

dockers/docker-telemetry-sidecar/systemd_stub.py
```diff
@@ -45,20 +43,43 @@ class SyncItem:
...
+  awk -F':' '/client-(certificate|key)/{val=$2; sub(/^[ \t]*/,"",val); gsub(/"/,"",val); print val}' "$KC" 2>/dev/null \
+  | sed 's/#.*$//' | tr -d '\r' | sort -u
+)"
+for p in $paths; do
+  [ -e "$p" ] || continue
+  sudo chown "$U:$U" "$p" || true
+  sudo chmod 600 "$p" || true
+done
+"""],
         ["sudo", "docker", "stop", "telemetry"],
```
make1980, line 58: nit: we should put this into a shell script file and include that into the docker so that we can easily test it. #Closed

dockers/docker-telemetry-sidecar/systemd_stub.py
```diff
@@ -45,20 +43,43 @@ class SyncItem:
...
+if [ -f "$KC" ]; then
+  sudo chown "$U:$U" "$KC" || true
+  sudo chmod 600 "$KC" || true
+fi
+
+# extract unique client cert/key paths from kubeconfig
+paths="$(
+  awk -F':' '/client-(certificate|key)/{val=$2; sub(/^[ \t]*/,"",val); gsub(/"/,"",val); print val}' "$KC" 2>/dev/null \
+  | sed 's/#.*$//' | tr -d '\r' | sort -u
+)"
```
make1980: one potential issue is that the certs are pulled by ACMS and the owner can be reset.

dockers/docker-telemetry-sidecar/systemd_scripts/telemetry.sh
```diff
@@ -1,85 +1,104 @@
...
+  for p in "${names[@]}"; do
+    [[ -n "$p" ]] && delete_pod_with_retry "$p" || rc=1
+  done
+  return "$rc"
 }
 
-cmd_start()   { kill_pods; }     # start == kill (DS restarts)
+cmd_start() {
+  # fire-and-forget deletion so systemd doesn't block on it
+  ( kill_pods >/dev/null 2>&1 & )
```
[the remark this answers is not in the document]
FengPan-Frank: fixed.

make1980:
> ARG BASE=docker-config-engine-bookworm-{{DOCKER_USERNAME}}:{{DOCKER_USERTAG}}

in PR description:

container_checker has issue intrigger kubectl for filter running_container to append the container rollouted via kubernetes, change the implementation logic to fix it.
=>
container_checker has issue in triggering kubectl for filter running_container to append the container rollouted via kubernetes, change the implementation logic to fix it.

---
Refers to: dockers/docker-telemetry-sidecar/Dockerfile.j2:2 in c809059. [](commit_id = c8090596bcf9b665692ca0e49fbff679e6da9002, deletion_comment = False)

FengPan-Frank:
> > ARG BASE=docker-config-engine-bookworm-{{DOCKER_USERNAME}}:{{DOCKER_USERTAG}}
> 
> in PR description:
> 
> container_checker has issue intrigger kubectl for filter running_container to append the container rollouted via kubernetes, change the implementation logic to fix it. => container_checker has issue in triggering kubectl for filter running_container to append the container rollouted via kubernetes, change the implementation logic to fix it.
> 
> Refers to: dockers/docker-telemetry-sidecar/Dockerfile.j2:2 in [c809059](https://github.com/sonic-net/sonic-buildimage/commit/c8090596bcf9b665692ca0e49fbff679e6da9002). [](commit_id = [c809059](https://github.com/sonic-net/sonic-buildimage/commit/c8090596bcf9b665692ca0e49fbff679e6da9002), deletion_comment = False)

updated

dockers/docker-telemetry-sidecar/systemd_scripts/telemetry.sh
```diff
@@ -1,85 +1,133 @@
...
-cmd_start()   { kill_pods; }     # start == kill (DS restarts)
+cmd_start() {
+  if command -v systemd-cat >/dev/null 2>&1; then
+    # background + pipe to journald with distinct priorities
+    ( kill_pods ) \
+      > >(systemd-cat -t telemetry-start -p info) \
+      2> >(systemd-cat -t telemetry-start -p err)
+  else
+    # background + pipe to syslog via logger
+    ( kill_pods ) \
```
make1980, line 103: in which case will systemd-cat not exist? could you add comment for that?

dockers/docker-telemetry-sidecar/systemd_scripts/telemetry.sh
```diff
@@ -1,85 +1,133 @@
...
+      > >(logger -t "telemetry-start" -p user.info) \
+      2> >(logger -t "telemetry-start" -p user.err)
+  fi &
+  disown
```
make1980, line 107: why do we need to run kill_pods in background?

make1980: shouldn't we use the return value of this function to reflect the result of kill_pods?

dockers/docker-telemetry-sidecar/systemd_scripts/telemetry.sh
```diff
@@ -1,85 +1,133 @@
...
+  exit 0
+}
 cmd_stop()    { kill_pods; }
```
make1980, line 111: we can't kill all the pods on the node - we should only kill the pod on the node that's containing telemetry, right? maybe add a label to the pod so that we can kill the pod based on the label. #Closed

dockers/docker-telemetry-sidecar/systemd_stub.py
```diff
@@ -136,15 +133,46 @@ def run_host_actions_for(path_on_host: str) -> None:
...
-
-# ───────────── file Sync logic ─────────────
 def sha256_bytes(b: Optional[bytes]) -> str:
     if b is None:
         return ""
     h = hashlib.sha256()
     h.update(b)
     return h.hexdigest()
 
+def render_telemetry_service_to_file(dst_path: str = TMP_RENDERED_SERVICE) -> bool:
```
make1980: do we need this to be invoked in sidecar container? we can just copy the generated telemetry.service file to the sidecar container and then use it to overwrite the file on the host, right? #Closed

make1980 (changes requested):

dockers/docker-telemetry-sidecar/systemd_scripts/telemetry.sh
FengPan-Frank: per my local testing, kubectl is usually slow which may hang the systemctl command for a while if using blocking mode here, thus just make it non-blocking mode and we can check result via "journalctl -t"

dockers/docker-telemetry-sidecar/systemd_scripts/telemetry.sh
FengPan-Frank: added

dockers/docker-telemetry-sidecar/systemd_scripts/telemetry.sh
FengPan-Frank: image

seems now we're using only one pod, do you mean later we may use multiple pods for containers?

dockers/docker-telemetry-sidecar/systemd_stub.py
FengPan-Frank: yes I tried but seems it's difficult to update image build flow to render telemetry service first before sidecar docker uses it.

dockers/docker-telemetry-sidecar/systemd_scripts/telemetry.sh
make1980: yes

dockers/docker-telemetry-sidecar/systemd_stub.py
make1980: there is just some build time dependency that we'll need to figure out - I don't see why that can't be done. could you try more? or we can look at the issue together.

make1980: maybe use something like this? # Stage 1: Build the artifact
FROM golang:1.22 AS builder
WORKDIR /src
COPY . .
RUN make build  # generates build/output.bin

# Stage 2: Final image
FROM alpine:3.19
COPY --from=builder /src/build/output.bin /usr/local/bin/myapp
CMD ["myapp"]

dockers/docker-telemetry-sidecar/systemd_scripts/telemetry.sh
make1980: we'll need to change this code to restart only the telemetry related pod - actually this is already problematic because we put telemetry together with other containers in the same pod today.

in any case let's only restart the telemetry related pod

FengPan-Frank: updated, will append label into daemonset

dockers/docker-telemetry-sidecar/systemd_stub.py
FengPan-Frank: refer https://github.com/sonic-net/sonic-buildimage/blob/6adbe35ec169bd6cf7d2bb518a595a27a6d2b92e/slave.mk#L1529

here is generic j2 render procedure, seems it's not clean change if we want to render telemetry.service.j2 separately and put it into sidecar container.

For your concern that telemetry.service.j2 is not verified during build should not be a concern, since the original render step will be blocked if there's any issue in j2 file change, telemetry.service.j2 is still one of the generic template file. How do you think?

make1980: it seems that we just need to use makefile to add a dependency from our docker image to build target $(addprefix $(TARGET_PATH)/, $(SONIC_INSTALLERS))?

at run time we do not build the same thing as what's build in the build process so there can be failure that's not caught by the build system.

alternatively I'd be OK to just have the systemd config file hardcoded in the repo - we hack systemd service definition for telemetry already, I don't feel it's a big problem of bypassing the template based config file generation.

make1980 (approved):
:shipit:

PR #24158, opened 2025-09-30
frr bgp conf template handling switch_type == voq

saravanan-nexthop:
when switch_type is voq, it could either be a single asic or multi-asic or even a voq_chassis
bgpd.main.conf.j2 needs to handle the single asic case differently than the multi-asic or voq_chassis.

#### Why I did it
when switch_type is voq, it could either be a single asic or multi-asic or even a voq_chassis
bgpd.main.conf.j2 needs to handle the single asic voq case differently than the multi-asic or voq_chassis.

##### Work item tracking

#### How I did it
Modifying bgpd.main.conf.j2

#### How to verify it
Verified sonic-mgmt bgp tests are passing on single asic voq sku 
Also verified the sonic-bgpcfgd/tests' unit tests pass 

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

- [ ] 
- [ ]

### DIFF
--- a/dockers/docker-fpm-frr/frr/bgpd/bgpd.main.conf.j2
+++ b/dockers/docker-fpm-frr/frr/bgpd/bgpd.main.conf.j2
@@ -42,7 +42,10 @@ ipv6 prefix-list LOCAL_VLAN_IPV6_PREFIX seq {{ loop.index * 5 }} permit {{ prefi
 {% if DEVICE_METADATA['localhost']['sub_role'] == 'FrontEnd' or DEVICE_METADATA['localhost']['sub_role'] == 'BackEnd' %}
 {% set multi_asic = True %}
 {% endif %}
-{% if multi_asic is defined or DEVICE_METADATA['localhost']['switch_type'] == 'voq' or DEVICE_METADATA['localhost']['switch_type'] == 'chassis-packet' %}
+{% if DEVICE_METADATA['localhost']['switch_type'] == 'voq' and ('chassis_db_address' in DEVICE_METADATA['localhost']) %}
+{% set voq_chassis = True %}
+{% endif %}
+{% if multi_asic is defined or voq_chassis is defined or DEVICE_METADATA['localhost']['switch_type'] == 'chassis-packet' %}
 ip prefix-list V4_P2P_IP permit 0.0.0.0/0 ge 31 le 31
 !
 ipv6 prefix-list V6_P2P_IP permit ::/0 ge 126 le 126
@@ -104,7 +107,7 @@ router bgp {{ DEVICE_METADATA['localhost']['bgp_asn'] }}
 !
 {% endif %}
 {# set router-id #}
-{% if DEVICE_METADATA['localhost']['sub_role'] == 'BackEnd' or DEVICE_METADATA['localhost']['switch_type'] == 'voq' or DEVICE_METADATA['localhost']['switch_type'] == 'chassis-packet' %}
+{% if DEVICE_METADATA['localhost']['sub_role'] == 'BackEnd' or voq_chassis is defined or DEVICE_METADATA['localhost']['switch_type'] == 'chassis-packet' %}
 {% if 'bgp_router_id' in DEVICE_METADATA['localhost'] %}
   bgp router-id {{ DEVICE_METADATA["localhost"]["bgp_router_id"] }}
 {% elif lo4096_ipv4 is not none %}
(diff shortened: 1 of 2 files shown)


dockers/docker-fpm-frr/frr/bgpd/bgpd.main.conf.j2
```diff
@@ -42,7 +42,10 @@ ipv6 prefix-list LOCAL_VLAN_IPV6_PREFIX seq {{ loop.index * 5 }} permit {{ prefi
 {% if DEVICE_METADATA['localhost']['sub_role'] == 'FrontEnd' or DEVICE_METADATA['localhost']['sub_role'] == 'BackEnd' %}
 {% set multi_asic = True %}
 {% endif %}
-{% if multi_asic is defined or DEVICE_METADATA['localhost']['switch_type'] == 'voq' or DEVICE_METADATA['localhost']['switch_type'] == 'chassis-packet' %}
+{% if DEVICE_METADATA['localhost']['switch_type'] == 'voq' and ('chassis_db_address' in DEVICE_METADATA['localhost']) %}
```
tjchadaga: @saravanan-nexthop, where is chassis_db_address added to DEVICE_METADATA? Is there another PR to check before this?

saravanan-nexthop: If the DEVICE_METADATA has `chassis_db_address` then the device is part of a chassis voq system.
If the DEVICE_METADATA has no `chassis_db_address` then the device is part of a single asic non-chassis voq system. 
There is no other PR to add `chassis_db_address` to DEVICE_METADATA.  Those devices would already be deployed with a valid `chassis_db_address` in their DEVICE_METADATA

saravanan-nexthop: If `chassis_db_address` is not the right field, please let me know if we can use a different field in DEVICE_METADATA to differentiate the system is part of a chassis system or a non-chassis system.

tjchadaga: Afaik, there is no single field in DEVICE_METADATA for this. One suggestion is to use "chassis_hostname" field in DEVICE_METADATA to identify if the device is part of a chassis
https://github.com/sonic-net/sonic-buildimage/blob/5149e6e9ff7e2fb70c7de0483e841275bc65fd8b/src/sonic-yang-models/yang-models/sonic-device_metadata.yang#L268

Basically, the check would be
if voq and (multi_asic or chassis), then set voq_chassis

saksarav-nokia: @tjchadaga , Is chssis_hostname newly introduced field in DEVICE_METADATA?. I know only hostname.

saravanan-nexthop: If we have no chassis specific filed in DEVICE_METADATA, then can I change the logic to look for chassisdb.conf file similar to the https://github.com/sonic-net/sonic-buildimage/pull/24163/ 's `is_voq_chassis()` 
```
{% if CONFIG_DB__DEVICE_METADATA['localhost']['switch_type'] == 'voq' and ('/usr/share/sonic/platform/chassisdb.conf' | file_exists) %}
{% set voq_chassis = True %}
{% endif %}
```

saravanan-nexthop: the latest diff is using `DEVICE_METADATA` lookup only for unit-test scenarios. Incase of unit test there is no `/usr/share/sonic` directory. hence need a way to input it through the `DEVICE_METADATA` 
The real deployment will be using the `file_exist of  /usr/share/sonic/platform/chassisdb.conf` and enable the voq_chassis

tjchadaga (approved):

dockers/docker-fpm-frr/frr/bgpd/bgpd.main.conf.j2
tjchadaga: discussed this internally and decided to proceed with this approach for now. We will add support for a CONFIG_DB field in sonic-mgmt as well and then update this section in the bgp template in a separate PR

saravanan-nexthop: thank you

PR #24163, opened 2025-09-30
Support for single ASIC VOQ Fixed-Systems

lakshmi-nexthop:
#### Why I did it
Fixed system VOQs do not have chassis app DB

##### Work item tracking

#### How I did it
- updated a function that checks for chassis to include if the voq is a fixed system
- swss.sh cleanup function is updated to check for chassis app db first

#### How to verify it
- bring up the FS in voq mode 

master

Support for single ASIC FS

PR#3847 under sonic-swss


abdosi (dismissed):

files/scripts/swss.sh
```diff
@@ -126,6 +126,12 @@ function clean_up_chassis_db_tables()
 {
 
     switch_type=`$SONIC_DB_CLI CONFIG_DB  hget 'DEVICE_METADATA|localhost' 'switch_type'`
+    platform=`$SONIC_DB_CLI CONFIG_DB  hget 'DEVICE_METADATA|localhost' 'platform'`
+    chassis_config="/usr/share/sonic/device/$platform/chassisdb.conf"
+    if [ ! -e $chassis_config ]; then
+       debug "No chassis config found"
+       return
+    fi
```
mlok-nokia: For efficiency and logical, I think this new section of code should be moved after line 139 which is at the end of if condition block)

src/sonic-py-common/sonic_py_common/device_info.py
```diff
@@ -590,9 +614,19 @@ def is_multi_npu():
     return (num_npus > 1)
 
 
+def is_single_voq():
+    chassis_db_conf_file_path = get_chassis_db_conf_file_path()
+    if chassis_db_conf_file_path is None:
+        return True
+
+    return False
```
mlok-nokia: Function name is_single_voq() is misleading.  Based on the function contents,  the checking inside the function only identifies If there is "chassisdb.conf" file.   All pizzabox (include the Non-VOQ platforms ) which doesn't contains chassisdb.conf will return True for this function.

lakshmi-nexthop: since this function is_single_voq() is used in combination with switch_type == voq, it will work fine but this puts the restriction on the usage of the function.
I can change function name to is_chassis_config_present?
This will change return codes etc, but should be only local impact as this function is not called anywhere else.

lakshmi-nexthop: changed the name to is_chassis_config_absent. so the logic does not change.

tjchadaga (approved):

saiarcot895:
@lakshmi-nexthop This is causing sonic-utilities tests to break, see https://dev.azure.com/mssonic/build/_build/results?buildId=985195&view=results

lakshmi-nexthop:
> @lakshmi-nexthop This is causing sonic-utilities tests to break, see https://dev.azure.com/mssonic/build/_build/results?buildId=985195&view=results

Ack. I am looking into how to fix this issue.

@saiarcot895 Sent PR to fix the tests. https://github.com/sonic-net/sonic-utilities/pull/4134

PR #24166, opened 2025-09-30
[Nvidia-bluefield][DPU]Fixed allowing PF interface to update MAC address

gpunathilell:
#### Why I did it

Previously the PF interface MAC address could not be set directly due to driver issues, now the need for a bridge network is not present since we can set the PF interface MAC address directly.  
Therefore the additional service and bridge network is not required since the mac address can be set directly using ip link command

##### Work item tracking

#### How I did it
Removed `pf-manager.service` This service was responsible to configure the pf0 interface as part of the eth0-midplane bridge and set the mac for eth0-midplane interface
Changed configure_midplane_iface command to only set the ip address of the midplane interface
Changed the udev rule to call the configure_midplane_iface function directly instead of using a service

#### How to verify it
Manual check with ifconfig the confirm that the changes are in effect (ifconfig command to confirm the ip address is updated)
Run tests to confirm that there are no regressions

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

- [ ] 
- [ ]


oleksandrivantsiv:
@prabhataravind, @kperumalbfn can you please review and merge this PR?

PR #24211, opened 2025-10-04
Remove version restriction for snappi and update OTEL packages to be compatible with snappi.

r12f:
#### Why I did it

The latest OTEL package has dependency on fairly recent gRPC and protobuf packages, which the latest snappi package does not support.

##### Work item tracking

#### How I did it

This commit fix the OTEL package version to a relatively old version, so it doesn't have the dependency on the gRPC and protobuf package. We cannot use newer version than 1.27.0 until snappi packages are updated with more loose dependency.

This change also removes the snappi package version, as we always prefer to use the latest version. This is one of the directions that mgmt container build is moving towards, in order to reduce maintenance effort.

#### How to verify it

Run pip install locally without error.

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

- [ ] 
- [ ]

### DIFF
--- a/dockers/docker-sonic-mgmt/Dockerfile.j2
+++ b/dockers/docker-sonic-mgmt/Dockerfile.j2
@@ -104,8 +104,8 @@ RUN python3 -m pip install aiohttp \
                  scapy==2.5.0 \
                  setuptools-rust \
                  six \
-                 snappi==1.33.0 \
-                 snappi-ixnetwork==1.33.1 \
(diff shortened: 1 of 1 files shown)


kperumalbfn (approved):

kperumalbfn:
@r12f Is this change required in other branches?

PR #24214, opened 2025-10-05
lldpmgrd: Interrupt select on signal

saiarcot895:
#### Why I did it

When trying to stop the lldp docker, the SIGTERM sinal is sent to lldpmgrd, but the handling of this signal is delayed because most of the time, lldpmgrd is going to be on the sel.select line, which goes to swss-common's C++ code where it calls `select()`. When the SIGTERM arrives, the `select()` calls returns early with EINTR, but because `interrupt_on_signal` is false, swss-common goes back into the `select()` call.

Because the SIGTERM handler is in Python, but the current call is in C++, the SIGTERM handling doesn't take effect until the call stack returns to Python code, which will take 10 seconds. By that time, docker will kill the docker with SIGKILL.

##### Work item tracking

#### How I did it

To avoid that, set `interrupt_on_signal` to true, which tells swss-common to return to Python code upon a signal, which will let any signal handlers run quickly.

#### How to verify it

Restart the lldp container, and check to see if docker had to send a SIGKILL to docker because it was blocked for 10 seconds, or if SIGTERM worked.

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

- [ ] 
- [ ]

### DIFF
--- a/dockers/docker-lldp/lldpmgrd
+++ b/dockers/docker-lldp/lldpmgrd
@@ -309,7 +309,7 @@ class LldpManager(daemon_base.DaemonBase):
 
         # Listen for changes to the PORT table in the CONFIG_DB and APP_DB
         while True:
-            (state, selectableObj) = sel.select(SELECT_TIMEOUT_MS)
+            (state, selectableObj) = sel.select(SELECT_TIMEOUT_MS, interrupt_on_signal=True)
 
             if state == swsscommon.Select.OBJECT:
                 if selectableObj.getFd() == sst_mgmt_ip_confdb.getFd():


qiluo-msft:
Is it a generic issue applicable to all the python code using select function?

saiarcot895:
It probably is a generic issue, particularly for ones that spend most of their time in the select function. If the application is getting events frequently or if their select timeout is low (maybe 5 seconds or less), then it's not much of an issue, because control would be returning back to Python frequently. A low select timeout will add to system load, though (frequent syscalls and wakeups).

I don't know off the top of my head which applications would fall into this bucket. bgpcfgd has a 1-second timeout, so it'll wake up frequently and also process the SIGTERM fairly quickly.

PR #24219, opened 2025-10-07
retry the chassis db cleanup operations

arlakshm:
#### Why I did it
When running load_minigraph or reloading configuration on the linecards, the `interface-config.service` restarts, which causes the midplane interface to flap. If swss.sh on the linecards deletes state from chassis_db, some states may not be cleaned up correctly, while others are successfully removed. For example, cleanup for `SYSTEM_NEIGHBOR` or `SYSTEM_INTF` may fail, but SYSTEM_LAG cleanup might succeed. This can lead to inconsistent lag IDs for the remote LC.

##### Work item tracking
- Microsoft ADO **35454463**

#### How I did it
Add logic to retry in swss.sh script. 

#### How to verify it
Run test to do load_minigraph on all the linecards and check for the logs to for remove lag failure for Lags on remote LC.

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

- [ ] 
- [ ]

### DIFF
--- a/files/scripts/swss.sh
+++ b/files/scripts/swss.sh
@@ -107,6 +107,120 @@ function clean_up_tables()
     end" 0
 }
 
+# This function attempts to delete entries using a specified delete function, retrying up to a maximum number of times if the deletion fails.
+# Arguments:
+#   $1 - delete_func: Name of the function to call for deleting the entry.
+#   $2 - host: Host identifier for the deletion operation.
+#   $3 - asic: ASIC identifier for the deletion operation.
+#   $4 - entry_name: Name of the entry being deleted (used for logging).
+#
+# Behavior:
+#   - Calls the delete function with host and asic as arguments.
+#   - If the deletion succeeds (exit code 0), prints the result and exits.
+#   - If the deletion fails, logs the attempt and retries up to max_retries times (default: 6).
+#   - Waits 10 seconds between retries.
+#   - If all attempts fail, logs an error and returns 1.
+function retry_delete_entries() {
+    local delete_func=$1
+    local host=$2
+    local asic=$3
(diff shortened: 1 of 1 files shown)


arlakshm:
@saksarav-nokia @ysmanman, @mlok-nokia, @arista-nwolfe, please help review...

saksarav-nokia:
@arlakshm , The code changes LGTM. However, i am unable to understand when we could hit this issue. If IMM can't reach the CHASSIS_APP_DB, then the following code waits till it is reachable right?
  until [[ $($SONIC_DB_CLI CHASSIS_APP_DB PING | grep -c True) -gt 0 ]]; do
        sleep 1
    done
    
    Also we fixed something in this area while ago with PR https://github.com/sonic-net/sonic-buildimage/pull/18756

arlakshm:
> @arlakshm , The code changes LGTM. However, i am unable to understand when we could hit this issue. If IMM can't reach the CHASSIS_APP_DB, then the following code waits till it is reachable right? until [[ $($SONIC_DB_CLI CHASSIS_APP_DB PING | grep -c True) -gt 0 ]]; do sleep 1 done
> 
> ```
> Also we fixed something in this area while ago with PR https://github.com/sonic-net/sonic-buildimage/pull/18756
> ```

Hi @saksarav-nokia, during load_minigraph the networking service is restarted, which happens parallely so the midplane interface can flap after this code is executed `$($SONIC_DB_CLI CHASSIS_APP_DB PING | grep -c True) -gt 0 ]]; do sleep 1 done`. In this case some the clean up in the chassis_db is not done. This PR to fix this case

PR #24220, opened 2025-10-07
Apparmor blocks chrony from reading tacacs config file

sarabjeetsingh1509:
Error:

ERR chronyd: nss_tacplus: /etc/tacplus_nss.conf fopen failed
ERR chronyd: nss_tacplus: bad config or server line for nss_tacplus"

Issue:

Chronyd is not able to access tacacs config file when tacacs is enabled.
This leads to above mentioned error.

RCA:

The issue arises because Debian includes a patch that enforces resource access control through AppArmor profiles. By default, the AppArmor profile for chrony does not permit access to /etc/tacplus_nss.conf. Since libnss-tacplus relies on this configuration file for TACACS+ user resolution, chronyd encounters access errors. To resolve this, the AppArmor profile must be updated to explicitly allow chrony to read /etc/tacplus_nss.conf.

Test case failing:

mvrf/test_mgmtvrf.py::TestServices::test_ntp

Fix:

Added a new AppArmor profile for chronyd to grant read access to the TACACS configuration file
/etc/tacplus_nss.conf r,

Updated the build script to copy the new AppArmor profile during filesystem construction

Back port request:
- [ ] 202405
- [ ] 202411
- [x] 202505
- [x] 202511


sarabjeetsingh1509:
@saiarcot895 Can you please help with the code review? Thanks.

saiarcot895 (dismissed):

sarabjeetsingh1509:
Co-pilot's suggestion merged was not signed-off. DCO check was failing. Updated the commit.

saiarcot895 (approved):

files/apparmor/usr.sbin.chronyd
```diff
@@ -0,0 +1,2 @@
+# Chrony needs to hook its API to the NSS plugin nss_tacplus and needs access to tacacs config file.
+/etc/tacplus_nss.conf r,
```
qiluo-msft, line 2: The PR fixes a non-bug

nsswitch is `passwd: tacplus files` (hostcfgd:759). chronyd calls getpwnam("_chrony") → nss_tacplus fails open, logs ERR, returns UNAVAIL → glibc falls through to files → lookup succeeds. NTP works fine. The cited test_mgmtvrf.py::test_ntp failure is almost certainly loganalyzer matching ERR chronyd: in syslog, not functional breakage.

Recommendation: whitelist chronyd.*nss_tacplus.*fopen failed in loganalyzer's ignore list. Zero production code change,

This is the root cause fix.

sarabjeetsingh1509: Hi @qiluo-msft 
thanks for the review.
I will investigate in this direction.
Will raise a sonic-mgmt PR if needed.

PR #24223, opened 2025-10-07
[initramfs] Add initramfs hook to log more info for blockdev wait issue in booting

byu343:
#### Why I did it
This change helps to the debug the issue when the block device for /host sometimes fails to initialize in booting.
When the issue happened before, the issue was not persistent so hard to reproduce, and the watchdog was triggered so there was no enough log to understand the cause.

##### Work item tracking

#### How I did it
Add an Arista-specific initramfs hook to wait for block dev for /host
 - Skip it for non-Arista hardware
 - Skip it if UUID is not in /proc/cmdline
 - Skip if in kdump enviroment
 - Wait for the device based on UUID and the timeout is around 10 seconds
 - If failed, print the last 100 lines of dmesg, ls /dev and blkid to the console. 

#### How to verify it
Verified that the expected debug info can be printed on a device with modified wrong UUID
Verified that the normal device can boot without issue.

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [x] 202505

- [ ] 
- [ ]

### DIFF
--- a/files/initramfs-tools/arista-wait-blockdev
+++ b/files/initramfs-tools/arista-wait-blockdev
@@ -1,0 +1,50 @@
+#!/bin/sh
+
+case $1 in
+    prereqs)
+        exit 0
+        ;;
+esac
+
+info() { printf "%04.2f: $@\n" "$(cut -f1 -d' ' /proc/uptime)"; }
+warn() { info "Warning: $@"; }
+
+set -e
+
+aboot_flag=''
+root_val=''
+
+# Extract kernel parameters
+set -- $(cat /proc/cmdline)
+for x in "$@"; do
+    case "$x" in
+        Aboot=*)
+            aboot_flag="${x#Aboot=}"
+            ;;
+        root=UUID=*)
+            root_val="${x#root=UUID=}"
+            ;;
+        systemd.unit=kdump-tools.service)
+            # In kdump environment, skip hooks
+            exit 0
+            ;;
+    esac
+done
+
+[ -z "$aboot_flag" ] && exit 0
+[ -z "$root_val" ] && exit 0
+
+wait_blockdev_retry_count=1
+while [ "$wait_blockdev_retry_count" -le 10 ]; do
+    info "Waiting for UUID=${root_val}: attempt ${wait_blockdev_retry_count}"
+    if blkid --uuid $root_val > /dev/null 2>&1; then
+        info "Finished waiting for UUID=${root_val}"
+        exit 0
+    fi
+    wait_blockdev_retry_count=$((wait_blockdev_retry_count + 1))
+    sleep 1
+done
+warn "Failed to wait for UUID=${root_val}"
+ls /dev
+dmesg | tail -n 100
+blkid
--- a/files/initramfs-tools/arista-wait-blockdev
+++ b/files/initramfs-tools/arista-wait-blockdev
@@ -9,8 +9,6 @@ esac
 info() { printf "%04.2f: $@\n" "$(cut -f1 -d' ' /proc/uptime)"; }
 warn() { info "Warning: $@"; }
 
-set -e
-
 aboot_flag=''
 root_val=''
 
@@ -35,7 +33,7 @@ done
 [ -z "$root_val" ] && exit 0
 
 wait_blockdev_retry_count=1
-while [ "$wait_blockdev_retry_count" -le 10 ]; do
+while [ "$wait_blockdev_retry_count" -le 30 ]; do
     info "Waiting for UUID=${root_val}: attempt ${wait_blockdev_retry_count}"
     if blkid --uuid $root_val > /dev/null 2>&1; then
         info "Finished waiting for UUID=${root_val}"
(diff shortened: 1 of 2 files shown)


files/initramfs-tools/arista-wait-blockdev
```diff
@@ -0,0 +1,50 @@
...
+            exit 0
+            ;;
+    esac
+done
+
+[ -z "$aboot_flag" ] && exit 0
+[ -z "$root_val" ] && exit 0
+
+wait_blockdev_retry_count=1
+while [ "$wait_blockdev_retry_count" -le 10 ]; do
```
Ryangwaite: In the worst case scenario, it looks like this will delay LACP/BGP restoration by 10s in the case of warmboot.

Can we avoid this change for warmboot? Or alternatively cap its runtime to ~1s worst case?

Good change for coldboot though.

byu343: This change repeatedly probes for /host every second. If the device does not appear, the warm-boot process would already be unable to proceed even without this modification.  In essence, the change only adds debug logging when the wait for /host takes too long, and it should not negatively impact cases where the device is slow to appear.

It helps to print some debug info we want to collect but could not because of the watchdog to power-cycle would kick in when block device not showing up for 180 seconds.
 - ls /dev
 - dmesg | tail -n 100
 - blkid

I also made the change less intrusive by:
- Removing set -e so the script continues even if a command like dmesg fails
- Ensuring the script always exits with status 0
- Changing the maximum wait time to 30 seconds. This means: If the device for /host take X (less than 30) seconds to appear, the script would block booting by X seconds for waiting for it. If it takes longer than 30 seconds, the script prints debug information, and then there is some existing system-level initramfs mechanism continues waiting for the device until it shows up.

Ryangwaite (approved):

davidm-arista (approved):

StormLiangMS (approved):
LGTM — net-positive: silently masks the slow-probe case on healthy boots, captures real evidence (`dmesg`/`ls /dev`/`blkid`) before the watchdog fires on the failure path. `exit 0` on every path means it cannot make boot worse than today.

One question on backport scope before we widen the blast radius:

**Do we actually need this on 202511 and 202505?** The 202605 backport makes sense — that's where ongoing diagnosis happens. But for the older release branches:

1. Have we **seen this bug occur on 202505 / 202511 in fleet**, or is the backport just precautionary? If it's only ever reproduced on master/202605, dropping the older backports keeps the change surface tight.
2. If we do backport, please also mirror `arista-hook`'s existing skip for warm/fast reboot so we don't add up to 30s to warmboot LACP/BGP recovery on a slow-probe box:
   ```sh
   SONIC_BOOT_TYPE=warm*|SONIC_BOOT_TYPE=fast*)
       exit 0
       ;;
   ```
   The earlier thread argued "warmboot is already toast if blockdev is missing" — true only if the device **never** appears; if it appears at second 15 of the wait, this script adds 15s to warmboot wall-clock that wasn't there before. Worth gating, especially for the 202505/202511 branches that are actively running warmboot in production.

vmittal-msft:
@byu343 please help answer @StormLiangMS  question so we can make decison for 202511

byu343:
> LGTM — net-positive: silently masks the slow-probe case on healthy boots, captures real evidence (`dmesg`/`ls /dev`/`blkid`) before the watchdog fires on the failure path. `exit 0` on every path means it cannot make boot worse than today.
> 
> One question on backport scope before we widen the blast radius:
> 
> **Do we actually need this on 202511 and 202505?** The 202605 backport makes sense — that's where ongoing diagnosis happens. But for the older release branches:
> 
> 1. Have we **seen this bug occur on 202505 / 202511 in fleet**, or is the backport just precautionary? If it's only ever reproduced on master/202605, dropping the older backports keeps the change surface tight.
> 2. If we do backport, please also mirror `arista-hook`'s existing skip for warm/fast reboot so we don't add up to 30s to warmboot LACP/BGP recovery on a slow-probe box:
>    ```shell
>    SONIC_BOOT_TYPE=warm*|SONIC_BOOT_TYPE=fast*)
>        exit 0
>        ;;
>    ```
>    
>    
>        
>          
>        
>    
>          
>        
>    
>        
>      
>    The earlier thread argued "warmboot is already toast if blockdev is missing" — true only if the device **never** appears; if it appears at second 15 of the wait, this script adds 15s to warmboot wall-clock that wasn't there before. Worth gating, especially for the 202505/202511 branches that are actively running warmboot in production.

Hi @StormLiangMS,

The issue to deal with here has not been seen on 202505. It was noticed on earlier release, but due to the watchdog triggered and no storage device ready, there is not much information to look into about why the storage device could not show up. 

Regarding the question of skipping the hook during a warm reboot, that doesn't seem correct. The issue only applies to warm reboots; the warm reboot process would still need to wait for the storage device to become available even without the wait introduced by this change; and the wait through the hook will finish right after the storage device shows up.

PR #24254, opened 2025-10-13
[alpinevs] Add platform specific configuration files for Alpine

sreeiyer-1:
#### Why I did it

Add the files for the new platform Alpine that supports a virtual dataplane (HLD - https://github.com/sonic-net/SONiC/blob/master/doc/alpine/alpine_hld.md ) 

##### Work item tracking

#### How I did it

Add the platform specific configuration files for Alpine. It broadly follows the structure of platform/vs.

#### How to verify it

This includes only a part of the changes for Alpine. A few more commits in the sonic-buildimage are required to build the full image. 

sonic-buildimage/platform/alpinevs$ keyword_check.sh .
Keyword check Passed.

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

- [ ] 
- [ ]


platform/alpinevs/docker-syncd-vs.mk
```diff
@@ -0,0 +1,19 @@
+# docker image for vs syncd
```
sonikajindal: alpinevs instead of vs?

platform/alpinevs/docker-syncd-vs.dep
```diff
@@ -0,0 +1,11 @@
+#DPKG FRK
+DPATH       := $($(DOCKER_SYNCD_BASE)_PATH)
+DEP_FILES   := $(SONIC_COMMON_FILES_LIST) platform/vs/docker-syncd-vs.mk platform/vs/docker-syncd-vs.dep
```
sonikajindal: Should this path be platform/alpinevs instead of platform/vs?

platform/alpinevs/docker-syncd-vs.mk
sreeiyer-1: changed

platform/alpinevs/docker-syncd-vs.dep
sreeiyer-1: changed path

platform/alpinevs/platform.conf
sonikajindal, line 1: Empty file? Needed?

sreeiyer-1: Kept it to be on the safer side. The script in https://github.com/sonic-net/sonic-buildimage/blob/master/build_image.sh#L91 refers to it but build did work even after deleting it. There is an identical empty file in platform/vs also.

sonikajindal: Ack, thanks

sonikajindal (approved):

sreeiyer-1:
Hi  @qiluo-msft ,  @lguohan  - can you please help to review this PR

lguohan:
most of the file seems duplicate of kvm vs, can we reuse them?

kazinator-arista (approved):
I've been using this base patch (among others) to do Alpine builds.

It looks identical to what I have.

This one is just adding the platform/alpinevs directory, not changing anything.

sreeiyer-1:
@lguohan  - updated the commit. For the nine files in platform/alpinevs that are identical to the files in platform/vs, replaced the files with softlinks to platform/vs

sreeiyer-1:
@lguohan  - Gentle reminder about this. Your comments are hopefully addressed, can you please take another look at this. 

Added the other part of the sonic-buildimage changes for alpine in https://github.com/sonic-net/sonic-buildimage/pull/24329  Please take a look at that also when possible.

lguohan:
i still have concerns on this pr due to duplication, this makes our future maintenance more difficult. I think we can explain vs image to build different flavors instead of introducing a new platform. what's the main difference between vs and alpinevs? mainly the syncd? maybe we can build to syncd in the vs image, included both of them in the same image, and depending on the platform, we can load either one. in this case, we can save our build pipleine to build one vs image instead of two vs image. let me know if you need more sync offline?

sonikajindal:
Hi @lguohan, we are introducing a new `platform=alpinevs` to support Alpine Virtual Switch capabilities as discussed in the HLD (https://github.com/sonic-net/SONiC/blob/master/doc/alpine/alpine_hld.md). This PR mainly has syncd changes, and there are follow up PRs where more code paths are added as per the platform type. There is definitely possibility of some code reuse, but for now there are a few things added only for Alpine Virtual Switch, so it would be better to keep them separate. We can revisit for optimization in later phases. Please let us know if this sounds ok.

sreeiyer-1:
Restructuring the code. Hence closing this PR.

PR #24262, opened 2025-10-14
[202505][Moby] Enable xcvrd SFF manager

kewei-arista:
#### Why I did it

High power mode is not correctly enabled for non CMIS QSFP modules, and we plan to use an AOI modules in the Moby QSFP port that requires enabling high power mode in 202505.

##### Work item tracking

#### How I did it

This change enables SFF manager so that high power mode can be correctly set for the API opics (and other affected modules). Note that we want to limit the impact in 202505 so we only enable it on Moby.

#### How to verify it

- Confirm the same AOI optics, which couldn't link up, can work with this change
- Run sonicmgmt tests to make sure no regressions are introduced

This change only targets 202505

- [x] 202505

### DIFF
--- a/device/arista/x86_64-arista_7060x6_16pe_384c_b/pmon_daemon_control.json
+++ b/device/arista/x86_64-arista_7060x6_16pe_384c_b/pmon_daemon_control.json
@@ -1,1 +1,4 @@
-../x86_64-arista_common/pmon_daemon_control.json
+{
+    "skip_fancontrol": true,
+    "enable_xcvrd_sff_mgr": true
+}


xwjiang-ms:
@kewei-arista we got an issue in PR test and now fixed, please run `/azpw run` to re-trigger PR test and get correct result

kewei-arista:
I ran `/azpw run` and looks like I have a successful build now. Not sure why the new successful checks don't override the old failing checks. Anything I can do to merge this PR?

xwjiang-ms:
> I ran `/azpw run` and looks like I have a successful build now. Not sure why the new successful checks don't override the old failing checks. Anything I can do to merge this PR?

let me try close and reopen

PR #24265, opened 2025-10-15
Revert "[submodule][master] Update submodule sonic-utilities to the latest HEAD automatically"

xwjiang-ms:
Reverts sonic-net/sonic-buildimage#23557 since it would cause tech_support failure:

test_techsupport_commands[vlab-03] failed for 
`Failed: Commands not found for 'frr_cmds': vtysh -c 'show thread cpu'; vtysh -c 'show thread poll'`


dgsudharsan:
@xwjiang-ms I have raised a PR to address this https://github.com/sonic-net/sonic-mgmt/pull/21007

PR #24311, opened 2025-10-21
FRR patch in zebra to fix a bug in zebra_ns_link_ifp()

markx-arista:
#### Why I did it

There is a race condition in zebra after bootup, where if_zebra_speed_update timer fires 15 seconds after interface creation. If speed is available in kernel, it calls if_add_update -> zebra_ns_link_ifp to link the interface to a namespace. But if this happens before RTM_NEWLINK event, it links the default ifindex IFINDEX_INTERNAL 0 and makes following if_lookup_by_index_per_ns fail. If one interface is affected, the BGP neighbor on that interface will remain permanently down. If two are affected, it causes assertion in zebra_ns_link_ifp because it tries to link both interfaces with ifindex 0.

#### How I did it

This workaround just skips linking if the ifindex is IFINDEX_INTERNAL. A complete solution requires careful redesign for the timer.

#### How to verify it

Reboot device with large amount of interfaces, there is 50% of chance that a random BGP neighbor goes down permanently on m1-108 and small chance that zebra asserts on m1-128.

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [x] 202505

Probably more branches?

- [x] 202505

This function should first check whether an interface is valid (i.e. an RTM_NEWLINK has been processed for the interface and it has a valid ifindex) before updating zebra_ns.ifp_tree.

### DIFF
--- a/src/sonic-frr/patch/0062-zebra-if-speed-change-check-fix.patch
+++ b/src/sonic-frr/patch/0062-zebra-if-speed-change-check-fix.patch
@@ -1,0 +1,28 @@
+Zebra patch to fix a bug in zebra_ns_link_ifp()
+
+From: dileep <dileep@arista.com>
+
+This function should first check whether the given interface is valid (i.e. an
+RTM_NEWLINK has been processed for the interface and it has a valid ifindex) before
+updating zebra_ns.ifp_tree.
+---
+ zebra/zebra_ns.c |    6 ++++++
+ 1 file changed, 6 insertions(+)
+
+diff --git a/zebra/zebra_ns.c b/zebra/zebra_ns.c
+index 8edfc9f3f..e9e83ae33 100644
+--- a/zebra/zebra_ns.c
++++ b/zebra/zebra_ns.c
+@@ -49,6 +49,12 @@ void zebra_ns_link_ifp(struct zebra_ns *zns, struct interface *ifp)
+ 	struct zebra_if *zif;
+ 	struct ifp_tree_link *link, tlink = {};
+
++	if (ifp->ifindex == IFINDEX_INTERNAL) {
++		if (IS_ZEBRA_DEBUG_EVENT)
++			zlog_debug("%s: interface %s not ready, ignoring", __func__, ifp->name);
++		return;
++	}
++
+ 	zif = ifp->info;
+ 	assert(zif != NULL);
+
--- a/src/sonic-frr/patch/series
+++ b/src/sonic-frr/patch/series
@@ -59,3 +59,4 @@
 0059-zebra-Ensure-that-the-dplane-can-send-the-full-packe.patch
 0060-bgpd-Convert-bmp-path_info-tracking-from-hash-to-rbt.patch
 0061-bgpd-Fix-JSON-wrapper-brace-consistency-in-neighbor.patch
+0062-zebra-if-speed-change-check-fix.patch


markx-arista:
This problem was introduced in FRR in 2018.
I have created an issue and PR for this workaround in FRR:
https://github.com/FRRouting/frr/issues/19792
https://github.com/FRRouting/frr/pull/19794
There is also an ongoing discussion for a more complete solution:
https://github.com/FRRouting/frr/pull/19804

StormLiangMS:
hi @markx-arista could you help to clarify why return can workaround this issue? There would be a retry followed up?

markx-arista:
> hi @markx-arista could you help to clarify why return can workaround this issue? There would be a retry followed up?

Hi @StormLiangMS, this has been merged in FRR: https://github.com/FRRouting/frr/pull/19794
I personally think if_zebra_speed_update shouldn't call if_add_update -> zebra_ns_link_ifp at all, zebra_interface_add_update should be called instead.
Only that timer function will pass a IFINDEX_INTERNAL in a race condition, so no retry should be done for this. If other correct code paths call if_add_update, it should have the correct ifindex.
This workaround is a confined solution that makes sure that if_zebra_speed_update doesn't link the wrong ifindex and also doesn't change the timer behavior. The timer has more issues than this, a more complete solution about the timer is being discussed in https://github.com/FRRouting/frr/pull/19804

StormLiangMS:
hi @markx-arista related to this issue, do you get some time today we can have a call to go through this quickly? We do see similar issue on 202411, but didn't find the code for zebra_ns_link_ifp in 10.0.1. 

vhttps://github.com/aristanetworks/sonic-qual.msft/issues/897

markx-arista:
> hi @markx-arista related to this issue, do you get some time today we can have a call to go through this quickly? We do see similar issue on 202411, but didn't find the code for zebra_ns_link_ifp in 10.0.1.
> 
> vhttps://github.com/[aristanetworks/sonic-qual.msft/issues/897](https://github.com/aristanetworks/sonic-qual.msft/issues/897)

Hi @StormLiangMS, sonic-frr is patched during build time, zebra_ns_link_ifp  is added in this patch: https://github.com/sonic-net/sonic-buildimage/blob/202411/src/sonic-frr/patch/0062-zebra-lib-use-internal-rbtree-per-ns.patch
But thanks for pointing this out, zebra_ns_link_ifp was actually added last year, previously if_link_per_ns was used and it actually checks ifp->ifindex == IFINDEX_INTERNAL. So this problem only happens to branches that include this patch.

StormLiangMS (approved):
LGTM

StormLiangMS:
hi @markx-arista thanks a lot for the quick fix, could you help to create a cherrypick PR to the 202411 branch?

StormLiangMS:
hi @lizhijianrd  @lipxu I'm wondering if this issue could be mitigated with a BGP restart? Or we have to do the config reload? I'm asking because it would be much heavier to have a config reload in prod environment.

yejianquan:
Hi @markx-arista , could you please fix the conflict and raise PR to 202505?

lizhijianrd:
202509: https://github.com/Azure/sonic-buildimage-msft/pull/1758

markx-arista:
> hi @lizhijianrd @lipxu I'm wondering if this issue could be mitigated with a BGP restart? Or we have to do the config reload? I'm asking because it would be much heavier to have a config reload in prod environment.

I can confirm a BGP service restart after all interfaces become ready can fix the issue.

markx-arista:
> Hi @markx-arista , could you please fix the conflict and raise PR to 202505?

202505: https://github.com/sonic-net/sonic-buildimage/pull/24356

markx-arista:
> hi @markx-arista thanks a lot for the quick fix, could you help to create a cherrypick PR to the 202411 branch?

202411: https://github.com/sonic-net/sonic-buildimage/pull/24358

markx-arista:
> hi @markx-arista thanks a lot for the quick fix, could you help to create a cherrypick PR to the 202411 branch?

Hi, can you review this?
https://github.com/sonic-net/sonic-buildimage/pull/24358

PR #24358, opened 2025-10-27
[202411] FRR patch in zebra to fix a bug in zebra_ns_link_ifp() (#24311)

markx-arista:
#### Why I did it

There is a race condition in zebra after bootup, where if_zebra_speed_update timer fires 15 seconds after interface creation. If speed is available in kernel, it calls if_add_update -> zebra_ns_link_ifp to link the interface to a namespace. But if this happens before RTM_NEWLINK event, it links the default ifindex IFINDEX_INTERNAL 0 and makes following if_lookup_by_index_per_ns fail. If one interface is affected, the BGP neighbor on that interface will remain permanently down. If two are affected, it causes assertion in zebra_ns_link_ifp because it tries to link both interfaces with ifindex 0.

#### How I did it

This workaround just skips linking if the ifindex is IFINDEX_INTERNAL. A complete solution requires careful redesign for the timer.

#### How to verify it

Reboot device with large amount of interfaces, there is 50% of chance that a random BGP neighbor goes down permanently on m1-108 and small chance that zebra asserts on m1-128.

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

Probably more branches?

- [x] 202505

This function should first check whether an interface is valid (i.e. an RTM_NEWLINK has been processed for the interface and it has a valid ifindex) before updating zebra_ns.ifp_tree.

### DIFF
--- a/src/sonic-frr/patch/0065-zebra-if-speed-change-check-fix.patch
+++ b/src/sonic-frr/patch/0065-zebra-if-speed-change-check-fix.patch
@@ -1,0 +1,28 @@
+Zebra patch to fix a bug in zebra_ns_link_ifp()
+
+From: dileep <dileep@arista.com>
+
+This function should first check whether the given interface is valid (i.e. an
+RTM_NEWLINK has been processed for the interface and it has a valid ifindex) before
+updating zebra_ns.ifp_tree.
+---
+ zebra/zebra_ns.c |    6 ++++++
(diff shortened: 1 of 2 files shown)


markx-arista:
@StormLiangMS Hi, can you approve this?

PR #24394, opened 2025-10-31
Hardcode the hostname to use for syslogs

saiarcot895:
#### Why I did it

On the host rsyslog, hardcode the hostname to use for logging syslogs, and don't use the hostname that is sent from rsyslog in the container. The reason for this is that when the container is started, the hostname might not be set correctly (which is the case for the database container); rsyslog reads the hostname once at process startup, but doesn't reread it for each syslog message. This means that syslogs from those containers might have the default hostname or an old hostname if the rsyslogd process wasn't restarted after a hostname change.

##### Work item tracking

#### How I did it

To work around this, on the host rsyslog, when we write it to `/var/log/syslog` or forward it out, use the hardcoded hostname that gets set in the config_db file in place of rsyslog determining the hostname or using the hostname passed down from the container rsyslog.

#### How to verify it

From the database container, generate a syslog. The easiest way is `logger "Test message"`. Check `/var/log/syslog` on the host and verify that it has the device's hostname and not `sonic`.

```
admin@vlab-01:~$ docker exec -it database bash
root@vlab-01:/# logger "Test log message"
root@vlab-01:/# exit
exit
admin@vlab-01:~$ sudo grep 'Test log' /var/log/syslog
2025 Nov 10 21:38:46.354173 vlab-01 NOTICE database#root: Test log message
```

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

- [ ] 
- [ ]

### DIFF
--- a/src/sonic-config-engine/tests/sample_output/py3/rsyslog.conf
+++ b/src/sonic-config-engine/tests/sample_output/py3/rsyslog.conf
@@ -40,10 +40,10 @@ $UDPServerRun 514
 #$ActionFileDefaultTemplate RSYSLOG_TraditionalFileFormat
 
 # Define a custom template
-$template SONiCFileFormat,"%timegenerated:::date-year% %timegenerated%.%timegenerated:::date-subseconds% %HOSTNAME% %syslogseverity-text:::uppercase% %syslogtag%%msg:::sp-if-no-1st-sp%%msg:::drop-last-lf%\n"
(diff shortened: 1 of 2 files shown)


arlakshm:
@saiarcot895, can you please create the PR for msft/202405 for chassis

deepak-singhal0408:
> @saiarcot895, can you please create the PR for msft/202405 for chassis

@saiarcot895 , could you create a PR for 202405? thanks,

PR #24401, opened 2025-10-31
Patch PTF gnxi for telemetry tests failures in mgmt IPv6 only setups

markx-arista:
py_gnmicli.py from gnxi library in PTF container doesn't handle IPv6 + port properly e.g. [addr]:port when passing it to grpc functions.

Fixes: https://github.com/sonic-net/sonic-buildimage/issues/24400

#### Why I did it

Telemetry tests failed for mgmt IPv6 only setups

#### How I did it

Patch gnxi library for PTF container, add square brackets to IPv6 address before combining with port.

#### How to verify it

Telemetry tests should pass.

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [x] 202505

- [x] 202505

Fix IPv6 addr + port format for grpc.

### DIFF
--- a/dockers/docker-ptf/gnxi-patches/0008-Fix-ipv6-addr-port-format-for-grpc.patch
+++ b/dockers/docker-ptf/gnxi-patches/0008-Fix-ipv6-addr-port-format-for-grpc.patch
@@ -1,0 +1,24 @@
+From a3b6293c8f9f5e0368a174711c981e9992a315d8 Mon Sep 17 00:00:00 2001
+From: markxiao <markxiao@arista.com>
+Date: Thu, 30 Oct 2025 13:54:47 -0700
+Subject: [PATCH] Fix IPv6 addr + port format for grpc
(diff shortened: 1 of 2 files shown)


StormLiangMS:
hi @vaibhavhd  could you have gnmi expert to take a look?

xincunli-sonic:
Hi @markx-arista, question about target self, what if target has square bracket, so, could we check target first then append. Thanks!

markx-arista:
> Hi @markx-arista, question about target self, what if target has square bracket, so, could we check target first then append. Thanks!

Hi, I originally tried to workaround this problem by passing the ipv6 addr with brackets from sonic-mgmt, but there were failures in _build_creds() function because socket library calls failed with brackets.
The whole py_gnmicli.py doesn't have any input check, so we should just expect users to input correct formats.

PR #24406, opened 2025-11-02
Issue 24405: [ztp] Support ZTP on eth0 over v6-only network

### LINKED ISSUE #24405: Bug: [ztp] ZTP cannot find IP6 default routes on eth0
### Is it platform specific

generic

### Importance or Severity

High

### Description of the bug

Currently, all interfaces in a sonic device are set to be IPv6 forwarding interfaces. This causes an issue because they won't honor RA advertisements. Also, currently, the IPv6 `dhclient` gets a prefix allocation, so by default, the interface receives a /128 address from the DHCPv6 server with no default route. The default behavior of ZTP is to try to find a DHCP server using the default route, if it exists. If there is no IPv6 default route, ZTP will fail unless some alternative is provided.

To address this, `eth0` can be set to host mode in the ZTP config script, so that it will accept RA, perform SLAAC, and retrieve the remaining information from the DHCPv6 server.

### Steps to Reproduce

Try to configure ZTP using the default address method when `eth0` is configured with an IPv6 address only.

Expected: ZTP finds a DHCP server over the `eth0` IPv6 default route and completes successfully.

Actual: ZTP fails to find a DHCP server and hangs waiting for a server to be found, never completing.

### Relevant log output

```shell

```

```shell

```

### Attach files (if any)

_No response_

anders-nexthop:
#### Why I did it

ZTP needs to use Router Advertisements to detect DHCP default routes in V6 networks.

Closes #24405 

##### Work item tracking

#### How I did it

Modify the interfaces.j2 template to support ZTP over v6 by using RA messages. This is done by setting accept_ra to 2, which allows the interface to accept RA messages.

#### How to verify it

Device is configurable using ZTP over ipv6 on eth0.

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

- [ ] 
- [ ] 

ZTP needs to use Router Advertisements to detect DHCP default routes in V6 networks.

### DIFF
--- a/files/image_config/interfaces/interfaces.j2
+++ b/files/image_config/interfaces/interfaces.j2
@@ -42,7 +42,7 @@ iface eth0 inet dhcp
 {% endif %}
 {% if ZTP['mode']['ipv6'] == 'true' %}
 iface eth0 inet6 dhcp
-    up sysctl net.ipv6.conf.eth0.accept_ra=1
+    up sysctl net.ipv6.conf.eth0.accept_ra=2
     down sysctl net.ipv6.conf.eth0.accept_ra=0
 {% endif %}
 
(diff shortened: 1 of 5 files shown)


anders-nexthop:
@lguohan can you take a look at this one please?

prgeor:
@kperumalbfn can you help review and merge?

PR #24441, opened 2025-11-05
[yang-model] support reserving mem in high region for crashkernel

ysmanman:
#### Why I did it

By default, crashekernel is reserved in low mem region used by DMA. Support reserving high mem region for crashkernel to avoid occupying DMA mem.

##### Work item tracking

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

- [ ] 
- [ ]


ysmanman:
@kenneth-arista @deepak-singhal0408 @YatishSVC @arlakshm

deepak-singhal0408:
@ysmanman can you update the testings done with this change.

arlakshm:
@ysmanman do we need to update the yang model?

ysmanman:
> @ysmanman can you update the testings done with this change.

Sure @deepak-singhal0408  Will do.

ysmanman:
> @ysmanman do we need to update the yang model?

Hi @arlakshm good point. I saved the new kdump config in configDb file, and then did config reload. The config reload went through without any issue. config-reload should validate config against yang model, right?

src/sonic-yang-models/yang-models/sonic-kdump.yang
```diff
@@ -33,7 +33,7 @@ module sonic-kdump {
 
                 leaf memory {
                     type string {
-                        pattern "(((([0-9]+[MG]?)?(-([0-9]+[MG])?):)?[0-9]+[MG],?)+)";
+                        pattern "(((([0-9]+[MG]?)?(-([0-9]+[MG])?):)?[0-9]+[MG],?)+(@[0-9]+[MG])?)";
```
deepak-singhal0408: I think we should have this optional offset inside the big pattern group(currently this regex only will support offset at the end).. What if in future we want to add offsets at other  ranges too..

ysmanman: Good suggestion. Updated the change.

deepak-singhal0408 (dismissed):
LGTM. thanks Song.

deepak-singhal0408:
@ysmanman please look at the PR checker failure.

ysmanman:
> @ysmanman please look at the PR checker failure.

Sure @deepak-singhal0408 . This is failure. Doesn't look like caused by my change. I will kick off build again.
 
```
2025-11-07T03:29:33.3468943Z Error type: RUN_TEST_CASE_FAILED
2025-11-07T03:29:33.3469268Z Error message: route/test_duplicate_route.py|||2 failed for RUN_TEST_CASE_FAILED, stop whole test plan
2025-11-07T03:29:33.3469821Z Operation failed with exception: Exception('Test plan id: 690d4ca98ad8ec69c27e4924, status: FAILED, result: FAILED, Elapsed 5438 seconds. Check ```

File: files/Aboot/boot0.j2 [hunk 2]
```diff
@@ -488,7 +488,12 @@ get_eeprom_value() {
 write_kdump_cmdline() {
     local sid="$(cmdline_get sid | sed 's/Ssd$//')"
     case "$sid" in
-        Wolverine*|Clearwater2*|OtterLake*|QuartzDd*|Redstart8Mk2Quartz4*|CitrineDd*)
+        OtterLake*)
+            if ! cmdline_has crashkernel; then
+                cmdline_add crashkernel=1G,high
```
judyjoseph: general question, is this 1G high applicable only for this platform. Should me make crashkernel allocation into high memory default ?

Also I see another PR from Arista : https://github.com/sonic-net/sonic-buildimage/pull/24382. May be we should close one of them ?

ysmanman: Hi @judyjoseph , as of now, we observed BRCM kernel module failed to allocate DMA mem on this platform, on which BRCM kernel module requires 512MB contiguous DMA memory.  So it does not have proactively make high mem by default if most platform does not request a big chunk of DMA mem.

https://github.com/sonic-net/sonic-buildimage/pull/24382 was created to enable KDUMP on different platforms. Both should be required.

src/sonic-yang-models/tests/yang_model_tests/tests_config/kdump.json
```diff
@@ -55,27 +55,27 @@
             }
         }
     },
-    "KDUMP_WITH_VALID_MEM_OFFSET": {
+    "KDUMP_WITH_VALID_MEM_HIGH": {
```
deepak-singhal0408, line 58: @ysmanman  could you add a testcase with the full memory paramaters..
"memory": "0M-2G:256M,2G-4G:256M,4G-8G:384M,8G-:448M, 16G:1G,high",

ysmanman: Hi @deepak-singhal0408 , `,high` does not support memory ranges at all. Only the size is allowed with `,high`.

deepak-singhal0408: got it. thanks

rlhui:
@ysmanman  please rebase? thanks.

arlakshm (approved):

arlakshm:
@ysmanman, can you please sign the commits for the DCO checker to pass

mlok-nokia (approved):

deepak-singhal0408 (approved):

ysmanman:
Hi @arlakshm @deepak-singhal0408 , this PR cast the fix from msft-202405 to master.  Can you help to merge this PR?

PR #24449, opened 2025-11-06
[gNMI] - Add command_whitelist.yaml for gNOI Debug RPC

matthew-soulsby:
#### Why I did it

This pull request introduces support for a gNOI Debug RPC command whitelist in the build process. The main change is the addition of a whitelist YAML file that specifies which commands are permitted for read and write operations, and ensures this configuration is included in the built image.

##### Work item tracking
- Microsoft ADO **(number only)**: 35562559

#### How I did it

* Added a new file `files/image_config/gnoi/debug/command_whitelist.yaml` containing lists of allowed commands for read and write operations, improving control over permitted debug actions.
* Modified `build_debian.sh` to copy the `command_whitelist.yaml` file into the image filesystem at `/etc/sonic/command_whitelist.yaml`, ensuring the whitelist is present in the final build.

#### How to verify it
* Image builds successfully
* From both host and gnmi container, the whitelist is present and populated as expected:
```
root@vlab-01:/# cat /etc/sonic/command_whitelist.yaml
read_whitelist:
  - ls
  - cat
  - echo
  - ping
  - grep
  - dmesg
  - zgrep
  - tail
  - teamshow
  - ps
  - uptime
  - awk
  - xargs
  - show
  - TSC

write_whitelist:
  - docker
  - reboot
  - mv
  - systemctl
  - ip
  - ifconfig
  - TSA
  - TSB
  - sonic-installer
  - sonic_installer
  - config
  - tcpdump
  - acl-loader
  - counterpoll
  - bcmcmd
  - redis-cli
  - sonic-db-cli
  - sonic-clear
  - wc
  - portstat
  - pfcwd
  - lspci
  - pcieutil
  - include
  - vtysh
```

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

- [x] master 
- [ ] 

* Add command whitelist YAML configuration file for the gNOI Debug RPC

### DIFF
--- a/files/image_config/gnoi/debug/command_whitelist.yaml
+++ b/files/image_config/gnoi/debug/command_whitelist.yaml
@@ -30,7 +30,6 @@ read_whitelist:
   - ps
   - uptime
   - awk
-  - xargs
   - show
   - TSC
 


files/image_config/gnoi/debug/command_whitelist.yaml
hdwhdw, line 1: Can we add some metadata here. For example, what does this file contain, who uses the list, etc? Basically add some introduction to someone who don't really know the project.

matthew-soulsby: Good catch, added a description with a goal, overview, and reference - how does that look?

saiarcot895: How/where is file consumed?

matthew-soulsby: @saiarcot895 it's consumed here: https://github.com/sonic-net/sonic-gnmi/pull/524/files
I've also added that link to the description in the file, hopefully to provide context to future readers as well

files/image_config/gnoi/debug/command_whitelist.yaml
```diff
@@ -0,0 +1,62 @@
...
+  - echo
+  - ping
+  - grep
+  - dmesg
+  - zgrep
+  - tail
+  - teamshow
+  - ps
+  - uptime
+  - awk
```
saiarcot895, line 32: `awk 'BEGIN { cmd = "ip link"; while ( ( cmd | getline result ) > 0 ) { print result }; close(cmd); }'`

files/image_config/gnoi/debug/command_whitelist.yaml
```diff
@@ -0,0 +1,62 @@
...
+  - xargs
```
saiarcot895: What's a use case for xargs?

files/image_config/gnoi/debug/command_whitelist.yaml
matthew-soulsby: This one's caught by the [validation logic](https://github.com/matthew-soulsby/sonic-gnmi/blob/180375ed5edc8f5c2b3c1255f9ee338392d2c390/pkg/gnoi/debug/validate.go) (which constructs an AST and traverses, rejecting anything potentially unsafe):
```
>>> client.run_command("awk 'BEGIN { cmd = \"ip link\"; while ( ( cmd | getline result ) > 0 ) { print result }; close(cmd); }'")
  File "<stdin>", line 1, in <module>
  File "/home/mattsoulsby/debug-service/.venv/lib/python3.11/site-packages/sonic_debug_client/client.py", line 22, in wrapper
    result = func(self, *args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mattsoulsby/debug-service/.venv/lib/python3.11/site-packages/sonic_debug_client/client.py", line 133, in run_command
    for response in responses:
  File "/home/mattsoulsby/debug-service/.venv/lib/python3.11/site-packages/grpc/_channel.py", line 543, in __next__
    return self._next()
           ^^^^^^^^^^^^
  File "/home/mattsoulsby/debug-service/.venv/lib/python3.11/site-packages/grpc/_channel.py", line 972, in _next
    raise self
grpc._channel._MultiThreadedRendezvous: <_MultiThreadedRendezvous of RPC that terminated with:
        status = StatusCode.PERMISSION_DENIED
        details = "command failed validation: command rejected by policy: word 1 is not a plain literal (contains expansions/substs/globs)"
        debug_error_string = "UNKNOWN:Error received from peer  {grpc_status:7, grpc_message:"command failed validation: command rejected by policy: word 1 is not a plain literal (contains expansions/substs/globs)"}"
>
```
That said, if you have any other ideas of how to break it, I'd definitely like to try, just to make sure it's as robust as expected!

files/image_config/gnoi/debug/command_whitelist.yaml
matthew-soulsby: My thought process is that for relatively simple cases like `find /etc/audit/rules.d/ -name '*.rules' -type f | sort | xargs cat` it would save on time and communication overhead to have the ability to do as much as possible on the device, but anything more complicated would be rejected (such as `echo "<interface name>" | xargs -I{} bash -c "sonic-db-cli APPL_DB hgetall 'MUX_CABLE_TABLE:{}'; sonic-db-cli STATE_DB hgetall 'MUX_CABLE_TABLE|{}'"`, which would be conservatively rejected by the validation policies).

Happy to remove it though, it's definitely not a necessity - what do you think?

saiarcot895: Ah, got it. As long as the command being run as part of the xargs is validated, this should be fine.

matthew-soulsby: Sorry, you're right, that's an oversight:
```
>>> client.run_command("xargs rm /home/admin/txt.txt")
0
```
I'll remove it, definitely not worth the added complexity for the small benefits that it brings

PR #24458, opened 2025-11-06
Add COPY packet_action support to ACL yang model

devpatha:
#### Why I did it
swss has support for ACL COPY action: https://github.com/sonic-net/sonic-swss/blob/master/orchagent/aclorch.h#L85
This action is missing in yang model.

##### Work item tracking

#### How I did it
Added COPY action to types template file/

#### How to verify it
Use COPY action with ACL rule and make sure yang validator pass.

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [x] 202505

- [ ] 
- [ ]


src/sonic-yang-models/yang-templates/sonic-types.yang.j2
```diff
@@ -66,6 +66,7 @@ module sonic-types {
             enum REDIRECT;
             enum DO_NOT_NAT;
             enum DISABLE_TRIM;
+            enum COPY;
```
kperumalbfn, line 69: @devpatha Could you add yang test for COPY ACL action

devpatha: @kperumalbfn I added tests for COPY ACL action. Could you please review it?

PR #24463, opened 2025-11-07
pmon: add missing devices due to unprivileged mode

Staphylo:
This change refactors the current mechanism to mount stuff in the containers. It lists all the top level content under `/dev` and mount them inside docker. Block and character devices are mounted using the --device parameter while other files are mounted using --mount.
Symlinks populated by udev or files containing : are not elegible for --device

#### Why I did it

The transition of `pmon` from a privileged container to unprivileged now prevents it from accessing host devices.
The "solution" consists of mounting individual host devices into at container creation time.
Some devices are currently missing including many `i2c-*`, `mmcblk*` and `uio*` for our platforms. 

##### Work item tracking

#### How I did it

Refactored the part of `docker_image_ctl.j2` that generates the `--device` arguments for `pmon`.
Instead of having many different ways of handling and figuring what to mount it makes more sense to  aggregate them into a single logic.

This logic iterates over all the top level entries of the host `/dev` folder.
When an entry matches the regex of devices to handle (i2c, sd, nvme, mmcblk, ipmi, ...) it generates the appropriate argument for the `docker container create` cli.
Some extra logic was added to mount block and character devices using  the `--device` argument and others such as symlinks using the `--mount` option.

The symlink handling was added because some platform use udev scripts to create symlinks for devices.
They have to be managed differently.
Also note that the `--device` argument only accepts devices that do not have a colon within.

#### How to verify it

Run the following command and make sure pmon is running and healthy
`systemctl stop pmon; docker rm pmon; systemctl start pmon`

Then ensure all desired devices/mounts are present by checking `docker inspect pmon`

pmon: add missing devices due to unprivileged mode

### DIFF
--- a/files/build_templates/docker_image_ctl.j2
+++ b/files/build_templates/docker_image_ctl.j2
@@ -29,8 +29,9 @@ link_namespace() {
 
 {%- if docker_container_name == "pmon" %}
 get_pmon_device_mounts() {
-    local devregex="$(echo 'i2c-[0-9]+|ipmi[0-9]+|sd[a-z]+|mmcblk[0-9].*|nvme[0-9].*|uio.*|watchdog[0-9]*' | sed -E 's#(^|\|)#\1/dev/#g')"
-    for device in $(find /dev/ -maxdepth 1 | grep -E "$devregex"); do
+    local devregex='i2c-[0-9]+|ipmi[0-9]+|sd[a-z]+|mmcblk[0-9].*|nvme[0-9].*|uio.*|watchdog[0-9]*'
+    local devpathregex="$(echo "$devregex" | sed -E 's#(^|\|)#\1/dev/#g')"
+    for device in $(find /dev/ -maxdepth 1 | grep -E "$devpathregex"); do
         if [ ! -L "$device" ] && [ -b "$device" -o -c "$device" ]; then
             echo "--device=$device"
         else


StormLiangMS:
hi @prgeor  could you help to review?

judyjoseph (dismissed):
LGTM. @Staphylo can we add a test case in sonic-mgmt to make sure we are able to access these devices from pmon docker/container - doesn't fail

files/build_templates/docker_image_ctl.j2
```diff
@@ -714,14 +714,7 @@ start() {
...
-    $(if [ -e "/dev/sdb" ]; then echo "--device=/dev/sdb:/dev/sdb"; fi) \
-    $(if [ -e "/dev/watchdog" ]; then echo "--device=/dev/watchdog:/dev/watchdog"; fi) \
-    $(for watchdog in /sys/class/watchdog/*; do if [ -d "$watchdog" ]; then device_name=$(basename "$watchdog"); dev_file="/dev/$device_name"; if [ -e "$dev_file" ]; then echo "--device=$dev_file:$dev_file"; fi; fi; done) \
-    $(for blockdev in /sys/block/*; do if [ -d "$blockdev" ]; then device_name=$(basename "$blockdev"); if [[ "$device_name" =~ ^(sd[a-z]+|nvme[0-9]+n[0-9]+)$ ]] && [[ ! "$device_name" =~ (boot|loop) ]]; then dev_file="/dev/$device_name"; if [ -e "$dev_file" ]; then echo "--device=$dev_file:$dev_file"; fi; fi; fi; done) \
-    $(for dev_file in /dev/nvme*; do if [ -e "$dev_file" ]; then echo "--device=$dev_file:$dev_file"; fi; done) \
+    $(for device in $(find /dev/ -maxdepth 1 | grep -E "$(echo 'i2c-[0-9]+|ipmi[0-9]+|sd[a-z]+|mmcblk[0-9].*|nvme[0-9].*|uio.*|watchdog[0-9]*' | sed -E 's#(^|\|)#\1/dev/#g')"); do if [ ! -L "$device" ] && [ -b "$device" -o -c "$device" ]; then echo "--device=$device"; else echo --mount=type=bind,source=$device,destination=$device; fi; done) \
```
r12f: hi @Staphylo I am so sorry... but do you mind to help add a few line breaks here :D

r12f:
functional-wise lgtm as Judy signed off. @prgeor , since this is i2c related and somewhat related to optics, do you also want to take another look?

files/build_templates/docker_image_ctl.j2
qiluo-msft: quite reasonable.

Staphylo: I don't believe I can add line breaks easily here because this is part of an argument list.
The solution I explored initially was to create a dedicated function only for `pmon` and call it here
```bash
{%- if docker_container_name == "pmon" %}
get_pmon_device_mounts() {
    local devregex="$(echo 'i2c-[0-9]+|ipmi[0-9]+|sd[a-z]+|mmcblk[0-9].*|nvme[0-9].*|uio.*|watchdog[0-9]*' | sed -E 's#(^|\|)#\1/dev/#g')"
    for device in $(find /dev/ -maxdepth 1 | grep -E "$devregex"); do
        if [ ! -L "$device" ] && [ -b "$device" -o -c "$device" ]; then
            echo "--device=$device"
        else
            echo "--mount=type=bind,source=$device,destination=$device"
        fi
    done
}
{%- endif %}
```
And then calling `$(get_pmon_device_mounts)` instead of existing long oneliner

Let me know if that sounds better

files/build_templates/docker_image_ctl.j2
```diff
@@ -27,6 +27,19 @@ link_namespace() {
...
 {%- endif %}
 
+{%- if docker_container_name == "pmon" %}
+get_pmon_device_mounts() {
+    local devregex="$(echo 'i2c-[0-9]+|ipmi[0-9]+|sd[a-z]+|mmcblk[0-9].*|nvme[0-9].*|uio.*|watchdog[0-9]*' | sed -E 's#(^|\|)#\1/dev/#g')"
+    for device in $(find /dev/ -maxdepth 1 | grep -E "$devregex"); do
+        if [ ! -L "$device" ] && [ -b "$device" -o -c "$device" ]; then
+            echo "--device=$device"
+        else
+            echo "--mount=type=bind,source=$device,destination=$device"
```
saiarcot895, line 38: Does a space need to be added to the end of each mount/device?

Staphylo: This is unnecessary since each mount option will be printed on its own line.
Because the `get_pmon_device_mounts` function is invoked from a subshell, the newlines will replaced by spaces.

StormLiangMS:
@judyjoseph could you help to take a look again?

judyjoseph:
@vmittal-msft could you cherry-pick this to 202511

r12f:
hi @Staphylo , do you mind to help pick this change into 202412?

PR #21475, opened 2025-01-20
[yang][dhcp_server] Make gateway in dhcp_server to be an un-mandatory config

yaqiangz:
#### Why I did it
DHCP default route shoule be an optional config to DHCP client

##### Work item tracking
- Microsoft ADO **(number only)**: 30877295

#### How I did it
Make the configuration to be optional in yang model

#### How to verify it
UTs

#### Which release branch to backport (provide reason below if selected)

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

#### Tested branch (Please provide the tested image version)

- [ ] 
- [ ]

PR #24495, opened 2025-11-11
Disable default route installed in dpu

yue-fred-gao:
#### Why I did it
DPU in smartswitch uses dhcp to get ip for its midplane interface. The interface is only for management traffic. Data traffic should go through Ethernet interface, which is where the default route should be. So we should not install a default route by dhcp over the midplane interface. It will disrupt data traffic.
##### Work item tracking

#### How I did it
Change dhcp server in smart-switch not to send "routers" option to dhcp client.
#### How to verify it
Reboot dpu or renew dhcp lease and verify no default route installed into dpu.

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

- [x]  SONiC.azure_cisco_master.0-dirty-20251106.125621
- [ ]


KrisNey-MSFT:
hi @prsunny and @prabhataravind may we have a reviewer added please?

prabhataravind:
@yue-fred-gao I recall this was fixed in the past. See below PRs. 

https://github.com/sonic-net/sonic-buildimage/pull/21462
https://github.com/sonic-net/sonic-buildimage/pull/21475

You can just remove the gateway config from smartswitch DHCP config and the default route should not get programmed. Could you check that instead of adding a smartswitch specific check?

yue-fred-gao:
> @yue-fred-gao I recall this was fixed in the past. See below PRs.
> 
> #21462 #21475
> 
> You can just remove the gateway config from smartswitch DHCP config and the default route should not get programmed. Could you check that instead of adding a smartswitch specific check?

Thanks. Let me explore the other option.

PR #24531, opened 2025-11-15
[yang models] Add vnet/vrf to bgp peer range

Leyza:
#### Why I did it
Bgp peer range yang model currently has no field to associate with a vnet or vrf, even though this feature is supported.

##### Work item tracking
- Microsoft ADO **(number only)**: 31626685

#### How I did it
Added vnet/vrf name field to the key for bgp peer range.

#### How to verify it
Tested on a physical testbed and UTs
image

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

- [ ] 
- [ ]

### DIFF
--- a/src/sonic-yang-models/yang-models/sonic-bgp-peerrange.yang
+++ b/src/sonic-yang-models/yang-models/sonic-bgp-peerrange.yang
@@ -59,7 +59,7 @@ module sonic-bgp-peerrange {
                 leaf name {
                     type string;
                     must "(current() = current()/../peer_range_name)" {
-                        error-message "Invalid name";
+                        error-message "Invalid name, name must match peer_range_name";
                     }
                 }
 
@@ -94,7 +94,7 @@ module sonic-bgp-peerrange {
                 leaf name {
                     type string;
                     must "(current() = current()/../peer_range_name)" {
-                        error-message "Invalid name";
+                        error-message "Invalid name, name must match peer_range_name";
                     }
                 }
 


src/sonic-yang-models/yang-models/sonic-bgp-peerrange.yang
```diff
@@ -28,6 +36,52 @@ module sonic-bgp-peerrange {
     container sonic-bgp-peerrange {
         container BGP_PEER_RANGE {
             list BGP_PEER_RANGE_LIST {
+                key "vrf_name peer_range_name";
```
anish-n, line 40: nit: could you add a description, something like BGP_PEER_RANGE_LIST_TEMPLATE with vnet and vrf support.

src/sonic-yang-models/yang-models/sonic-bgp-peerrange.yang
```diff
@@ -28,6 +36,52 @@ module sonic-bgp-peerrange {
...
+                }
+
+                leaf-list ip_range {
+                    type stypes:sonic-ip-prefix;
+                    ordered-by user;
+                    description "A range of addresses";
+                }
+            }
+
+            list BGP_PEER_RANGE_TEMPLATE_LIST {
```
anish-n, line 85: add a description here too

src/sonic-yang-models/yang-models/sonic-bgp-peerrange.yang
Leyza: Added description

src/sonic-yang-models/yang-models/sonic-bgp-peerrange.yang
Leyza: Added description

anish-n (approved):

StormLiangMS:
@vaibhavhd could you take a look?

src/sonic-yang-models/yang-models/sonic-bgp-peerrange.yang
```diff
@@ -28,6 +36,54 @@ module sonic-bgp-peerrange {
...
+
+                leaf peer_range_name {
+                    type string;
+                    description "Peer range name";
+                }
+
+                leaf name {
+                    type string;
+                    must "(current() = current()/../peer_range_name)" {
+                        error-message "Invalid name";
```
vaibhavhd: I agree with the Copilot suggestion here to make the error message more specific.

vaibhavhd:
I don't have any concerns with this change. However, @Leyza could you describe how this gap was detected? Is there a testcase to detect the gap here? Was there a config application failure due to the lack of this change on YANG model.

Also trying to understand why this was missed to begin with. How did we not catch this gap when the feature that added support for "Bgp peer range to associate with a vnet or vrf" was added?

Leyza:
@vaibhavhd We have a use case where we configure vnet bgp peer ranges, and that's where we noticed that the configuration did not pass yang validations, but functionality-wise everything is working. There will be test cases in the future that we are working on that will be applying vnet bgp peer ranges so this case will be covered there.

src/sonic-yang-models/yang-models/sonic-bgp-peerrange.yang
Leyza: Updated message

PR #24538, opened 2025-11-16
Introduce a build option SONIC_DPKG_CACHE_METHOD_OVERRIDE

yijingyan2:
Introduce a build option SONIC_DPKG_CACHE_METHOD_OVERRIDE that specifying whether to override the method used by SONIC_DPKG_CACHE_METHOD. It is designed to reading from cache when building linux-kernel while SONIC_DPKG_CACHE_METHOD=wcache

#### Why I did it
When SONIC_DPKG_CACHE_METHOD is set to wcache, we can use the new build option SONIC_DPKG_CACHE_METHOD_OVERRIDE and set it to rcache to read from cache when building linux-kernel to reduce the building time. 

##### Work item tracking
35882451

#### How I did it
By introducing a new parameter and refining the logic for checking if rcache is enabled.


yijingyan2:
Hi @liushilongbuaa, could you please review this pr?

Makefile.cache
```diff
@@ -310,13 +310,29 @@ endef
 # Read from the cache
 RCACHE_OPTIONS := cache rcache rwcache
 define CHECK_RCACHE_ENABLED
-$(if $(and $(filter $(RCACHE_OPTIONS),$(SONIC_DPKG_CACHE_METHOD)),$(filter-out none,$($(1)_CACHE_MODE))),enabled)
+$(strip \
```
liushilongbuaa: I guess we only need to change filter function here.

Makefile.cache
```diff
@@ -310,13 +310,17 @@ endef
 # Read from the cache
 RCACHE_OPTIONS := cache rcache rwcache
 define CHECK_RCACHE_ENABLED
-$(if $(and $(filter $(RCACHE_OPTIONS),$(SONIC_DPKG_CACHE_METHOD)),$(filter-out none,$($(1)_CACHE_MODE))),enabled)
+$(if $(and $(filter $(RCACHE_OPTIONS), \
+					$(or $(filter-out none,$($(1)_CACHE_OVERRIDE)),$(SONIC_DPKG_CACHE_METHOD))), \
```
liushilongbuaa: It is override not or. override has higher priority than SONIC_DPKG_CACHE_METHOD.
```
$(SONIC_DPKG_CACHE_METHOD)  ->  $(if $($(1)_CACHE_OVERRIDE),$($(1)_CACHE_OVERRIDE),$(SONIC_DPKG_CACHE_METHOD))
```

ygindn: (or condition1,condition2,…) returns the first non-empty argument. So in this case, override method has a higher priority. if override method is not none, we use override method, else we use sonic_dpkg_cache_method

Makefile.cache
```diff
@@ -310,13 +310,17 @@ endef
...
+			$(filter-out none,$($(1)_CACHE_MODE))),enabled)
 endef
 
 # Write into the cache
 WCACHE_OPTIONS := cache wcache rwcache
 define CHECK_WCACHE_ENABLED
-$(if $(and $(filter $(WCACHE_OPTIONS),$(SONIC_DPKG_CACHE_METHOD)),$(filter-out none,$($(1)_CACHE_MODE))),enabled)
+$(if $(and $(filter $(WCACHE_OPTIONS), \
+					$(or $(filter-out none,$($(1)_CACHE_OVERRIDE)),$(SONIC_DPKG_CACHE_METHOD))), \
```
saiarcot895: Since `SONIC_DPKG_CACHE_METHOD_OVERRIDE` is defined to be only `none` or `rcache`, do we even want to check it for cache writes?

rules/linux-kernel.mk
```diff
@@ -23,6 +23,7 @@ export EXTERNAL_KERNEL_PATCH_LOC
 
 LINUX_HEADERS_COMMON = linux-headers-$(KVERSION_SHORT)-common_$(KERNEL_VERSION)-$(KERNEL_SUBVERSION)_all.deb
 $(LINUX_HEADERS_COMMON)_SRC_PATH = $(SRC_PATH)/sonic-linux-kernel
+$(LINUX_HEADERS_COMMON)_CACHE_OVERRIDE = $(SONIC_DPKG_CACHE_METHOD_OVERRIDE)
```
saiarcot895: Can these be moved into the `.dep` files, alongside the existing caching options?

liushilongbuaa: Agree

Makefile.work
```diff
@@ -41,8 +41,9 @@
 #  * ENABLE_DIALOUT: Enable dialout client in telemetry.
 #  *                 Default: unset
 #  *                 Values: y
-#  * SONIC_DPKG_CACHE_METHOD: Specifying method of obtaining the Debian packages from cache: none or cache
+#  * SONIC_DPKG_CACHE_METHOD: Specifying method of obtaining the Debian packages from cache: none, cache, rcache, wcache, rwcache
 #  * SONIC_DPKG_CACHE_SOURCE: Debian package cache location when cache enabled for debian packages
+#  * SONIC_DPKG_CACHE_METHOD_OVERRIDE: Specifying whether to override the method used by SONIC_DPKG_CACHE_METHOD: none, rcache
```
liushilongbuaa: Please update description. Your code support both rcache and wcache and cache

Makefile.cache
liushilongbuaa: Let's update description.

yijingyan2: updated

rules/linux-kernel.mk
yijingyan2: updated

Makefile.work
```diff
@@ -531,6 +532,7 @@ SONIC_BUILD_INSTRUCTION :=  $(MAKE) \
                            KERNEL_PROCURE_METHOD=$(KERNEL_PROCURE_METHOD) \
                            SONIC_DPKG_CACHE_METHOD=$(SONIC_DPKG_CACHE_METHOD) \
                            SONIC_DPKG_CACHE_SOURCE=$(SONIC_DPKG_CACHE_SOURCE) \
+						   SONIC_DPKG_CACHE_METHOD_OVERRIDE=$(SONIC_DPKG_CACHE_METHOD_OVERRIDE) \
```
liushilongbuaa: Please use same tab or space format with previous line.

Makefile.cache
liushilongbuaa: It's better to remove filter-out. The 2 parameters in 'or' function should process in same way.
none is also a kind of cache mode. override should support none type.

yxieca (approved):

PR #24567, opened 2025-11-18
Add prerequisites script and update README

qiluo-msft:
#### Why I did it
Add prerequisites script for automated prerequisites installation and repository cloning. Update README with single script solution.

##### Work item tracking

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

- [ ] 
- [ ]


scripts/prerequisites.sh
```diff
@@ -0,0 +1,102 @@
...
+    sudo gpasswd -a "${USER}" docker
+
+###############################################
+# Clone SONiC Repo
+###############################################
+if [ -d "$SONIC_DIR" ]; then
+    echo "==> Directory $SONIC_DIR already exists, skipping clone."
+else
+    run_step "Cloning SONiC repository" \
+        git clone --recurse-submodules "$SONIC_REPO" "$SONIC_DIR"
```
tahmed-dev, line 94: do we have a hard requirement to clone buildimage into the `$HOME` directory? End users may want to organize their work a bit differently so, it might be a good idea to have basedir be provided as an argument. #Resolved

qiluo-msft: This is not a hard requirement. Improved the script so end user could override if they want.

README.md
```diff
@@ -82,7 +82,22 @@ Any server can be a build image server as long as it has:
 
 A good choice of OS for building SONiC is currently Ubuntu 20.04.
```
nate-nexthop: I have a PR open to update this line to recommend 24.04 instead of the EOL 20.04: https://github.com/sonic-net/sonic-buildimage/pull/22846 

If you want to update the recommended release here in this PR to 24.04, I think that would be reasonable. I'd be happy to close my PR.

scripts/prerequisites.sh
```diff
@@ -0,0 +1,111 @@
...
+###############################################
+
+run_step "Updating apt package index" \
+    sudo apt update
+
+run_step "Installing prerequisites (python3-pip, git)" \
+    sudo apt install -y python3-pip git
+
+run_step "Installing jinjanator (j2)" \
+    pip3 install --user jinjanator
```
saiarcot895: This will need the `--break-system-packages` flag; otherwise, newer versions of Python will fail here with an error message:

```
==> Installing jinjanator (j2)
error: externally-managed-environment

× This environment is externally managed
╰─> To install Python packages system-wide, try apt install
    python3-xyz, where xyz is the package you are trying to
    install.

    If you wish to install a non-Debian-packaged Python package,
    create a virtual environment using python3 -m venv path/to/venv.
    Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
    sure you have python3-full installed.

    If you wish to install a non-Debian packaged Python application,
    it may be easiest to use pipx install xyz, which will manage a
    virtual environment for you. Make sure you have pipx installed.

    See /usr/share/doc/python3.12/README.venv for more information.

note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
hint: See PEP 668 for the detailed specification.
``` #Resolved

File: README.md (hunk 2 above)
Comment (qiluo-msft) [reply README.md]: thanks! commmented on your PR, let's use your PR.

File: scripts/prerequisites.sh (hunk 3 above)
Comment (saiarcot895) [reply scripts/prerequisites.sh]: Alternatively, `sudo apt install j2cli` will work.

File: README.md (hunk 2 above)
Comment (nate-nexthop) [reply README.md]: Sounds good thanks, I've updated my PR pre your comment.

**Review by kperumalbfn — APPROVED** [review]
Please resolve copilot suggestions.

File: scripts/prerequisites.sh [hunk 4]
```diff
@@ -0,0 +1,111 @@
...
+        sudo apt update
+
+    run_step "Installing Docker Engine" \
+        sudo apt install -y docker-ce docker-ce-cli containerd.io
+else
+    echo "==> Docker already installed, skipping."
+fi
+
+run_step "Adding user '${USER}' to docker group" \
+    sudo gpasswd -a "${USER}" docker
```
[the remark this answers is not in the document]
qiluo-msft: since it is idempotent, no harm

README.md
```diff
@@ -82,7 +82,22 @@ Any server can be a build image server as long as it has:
 
 A good choice of OS for building SONiC is currently Ubuntu 22.04.
 
-## Prerequisites
+## Automated prerequisites installation and repository cloning
+
+For convenience, you can use the automated prerequisites script to handle both prerequisites installation and repository cloning:
+
+```shell
+curl -sSL https://raw.githubusercontent.com/sonic-net/sonic-buildimage/master/scripts/prerequisites.sh | bash
```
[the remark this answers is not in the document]
qiluo-msft: It is just an alternative option for user to automate the build environment preparation. User could following README manual steps.

scripts/prerequisites.sh
```diff
@@ -0,0 +1,111 @@
...
+
+###############################################
+# Docker Install Check
+###############################################
+if ! command -v docker >/dev/null 2>&1; then
+    run_step "Installing Docker dependencies" \
+        sudo apt install -y ca-certificates curl gnupg lsb-release
+
+    run_step "Adding Docker GPG key" \
+        bash -c "curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo gpg --dearmor -o /usr/share/keyrings/docker-archive-keyring.gpg"
```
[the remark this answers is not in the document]
qiluo-msft: We assume Ubuntu currently.

PR #24604, opened 2025-11-23
[build] Fix rsync failure in Docker builds when SONIC_CONFIG_USE_NATIVE_DOCKERD_FOR_BUILD is enabled

DavidZagury:
#### Why I did it
When building SONiC with `SONIC_CONFIG_USE_NATIVE_DOCKERD_FOR_BUILD=y`, Docker multi-stage builds fail during rsync operations with the following error:

```
rsync: [generator] failed to set times on "/changes-to-image": Read-only file system (30)
rsync error: some files/attrs were not transferred (see previous errors) (code 23)
```

This issue occurs specifically when:
- Using Docker Engine 20.10.x with BuildKit
- Running builds inside containers (Docker-in-Docker scenario)
- Accessing the host Docker daemon via socket mount
- Using `--mount=type=bind` in multi-stage Dockerfiles

The root cause is that Docker BuildKit creates read-only bind mounts, and rsync with the `-a` flag (which includes `-t` for preserving times) attempts to set timestamps on the mounted directory itself, failing on the read-only mount point.

This blocks builds in environments using native dockerd for build acceleration.

##### Work item tracking

#### How I did it
Added `--omit-dir-times` flag to all rsync commands in Dockerfile.j2 templates that use bind mounts for multi-stage builds.

This flag tells rsync to skip setting directory timestamps (not needed for functionality) while still preserving:
- All file timestamps
- All file content and permissions
- All directory structure

The fix is applied to 37 Dockerfile.j2 files:
- 34 files in `dockers/` directory (common base Docker images used by all platforms)
- 3 files in `platform/mellanox/` directory

#### How to verify it
Build SONiC using SONIC_CONFIG_USE_NATIVE_DOCKERD_FOR_BUILD=y
Verify Docker images build successfully

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

- [ ] 
- [ ]

### DIFF
--- a/platform/vs/docker-gbsyncd-vs/Dockerfile.j2
+++ b/platform/vs/docker-gbsyncd-vs/Dockerfile.j2
@@ -1,4 +1,4 @@
-{% from "dockers/dockerfile-macros.j2" import install_debian_packages, install_python_wheels, copy_files %}
+{% from "dockers/dockerfile-macros.j2" import install_debian_packages, install_python_wheels, copy_files, rsync_from_builder_stage %}
 ARG BASE=docker-config-engine-bookworm-{{DOCKER_USERNAME}}:{{DOCKER_USERTAG}}
 
 FROM $BASE AS base


tirupatihemanth:
@saiarcot895  Please help review/approve/merge this change to fix the prod trixie build as well

bingwang-ms:
@liushilongbuaa , @saiarcot895 Can you please help review?

DavidZagury:
> I see Mellanox platform has been updated, what about Broadcom, VS, and Marvell platforms?

@saiarcot895 as we cannot test it for their platform we cannot do that for other vendors.

volodymyrsamotiy:
@saiarcot895 , could you please approve if no more comments?

dprital:
> @saiarcot895 , could you please approve if no more comments?

@saiarcot895 , can you please ?

bingwang-ms:
@saiarcot895 Can you please review? Thanks

saiarcot895:
VS can absolutely be tested; it's just a VM running on a system with at least 6GB of free RAM and 2 cores that can be spared.

Additionally, because of the specificity of the command, I'd rather not see this diverge between different dockerfiles. This will just mean more maintenance work later when changes need to be made for whatever reason.

saiarcot895:
Actually, one improvement is to make this into a macro in `dockerfile-macros.j2` and use that macro in each Dockerfile.

DavidZagury:
> VS can absolutely be tested; it's just a VM running on a system with at least 6GB of free RAM and 2 cores that can be spared.
> 
> Additionally, because of the specificity of the command, I'd rather not see this diverge between different dockerfiles. This will just mean more maintenance work later when changes need to be made for whatever reason.

@saiarcot895 VS was added

> Actually, one improvement is to make this into a macro in `dockerfile-macros.j2` and use that macro in each Dockerfile.

@saiarcot895 Good point, I have added that macro

liat-grozovik:
@saiarcot895 could you please approve as comments were addressed?

dprital:
@saiarcot895 , Can you please approve this PR ?

kperumalbfn (approved):

dockers/dockerfile-macros.j2
```diff
@@ -43,4 +43,8 @@ COPY \
         RUN pip3 install {{ pkg }}
         RUN pip3 install setuptools==51.0.0
     {%- endfor %}
+{%- endmacro %}
+
+{% macro rsync_from_builder_stage() -%}
+RUN --mount=type=bind,from=base,target=/changes-to-image rsync -axAX --omit-dir-times --no-D --exclude=/sys --exclude=/proc --exclude=/dev --exclude=resolv.conf /changes-to-image/ /
```
benoit-nexthop, line 49: Have you considered doing it this way:

```
FROM scratch
COPY --from=base / /
```

Would that work too?

saiarcot895: @benoit-nexthop That was my initial approach, but because of differences in kernel settings around overlayfs, this could register as all files being changed even if there are no changes. More details in https://github.com/sonic-net/sonic-buildimage/pull/19952/changes/0b85785652269551d4102e9c927c78b0eb8ebf0f.

PR #24605, opened 2025-11-24
Enhance Rust supervisor-proc-exit-listener-rs script to handle redis-communication failure case gracefully

qiluo-msft:
#### Why I did it
Enhance Rust supervisor-proc-exit-listener-rs script to handle redis-communication failure case gracefully similar to python version changes: #24393

Remove risky unwrap().

##### Work item tracking

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

- [ ] 
- [ ]

### DIFF
--- a/src/sonic-supervisord-utilities-rs/src/proc_exit_listener.rs
+++ b/src/sonic-supervisord-utilities-rs/src/proc_exit_listener.rs
@@ -332,11 +332,8 @@ pub fn main_with_parsed_args_and_stdin<S: Read + AsRawFd, P: Poller>(args: Args,
 
         let mut stdin_ready = false;
         for event in events.iter() {
-            match event.token() {
-                STDIN_TOKEN => {
-                    stdin_ready = true;
-                }
-                _ => unreachable!(),
+            if event.token() == STDIN_TOKEN {
+                stdin_ready = true;
             }
         }
 
--- a/src/sonic-supervisord-utilities-rs/src/proc_exit_listener.rs
+++ b/src/sonic-supervisord-utilities-rs/src/proc_exit_listener.rs
@@ -12,7 +12,7 @@ use std::fs::File;
 use std::io::{self, BufRead, BufReader, Read};
 use std::os::unix::io::AsRawFd;
 use std::process;
-use std::sync::{Mutex, OnceLock};
+use std::sync::OnceLock;
 use std::time::{Duration, Instant};
 use swss_common::{ConfigDBConnector, EventPublisher};
 use syslog::Severity;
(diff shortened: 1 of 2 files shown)


src/sonic-supervisord-utilities-rs/src/proc_exit_listener.rs
```diff
@@ -185,75 +181,60 @@ pub fn generate_alerting_message(process_name: &str, status: &str, dead_minutes:
 }
 
 /// Read auto-restart state from ConfigDB
-pub fn get_autorestart_state(container_name: &str, config_db: &dyn ConfigDBTrait) -> Result<String> {
-
-    let features_table = config_db.get_table(FEATURE_TABLE_NAME)
-        .map_err(|e| SupervisorError::Database(format!("Failed to get FEATURE table: {}", e)))?;
+/// Returns empty string on error to allow graceful handling (e.g., when redis-server exits)
+pub fn get_autorestart_state(container_name: &str, config_db: &dyn ConfigDBTrait) -> String {
```
saiarcot895: This should be `Option<String>`,  with `None` being used in place of an empty string. #Resolved

src/sonic-supervisord-utilities-rs/src/proc_exit_listener.rs
```diff
@@ -348,11 +332,8 @@ pub fn main_with_parsed_args_and_stdin<S: Read + AsRawFd, P: Poller>(args: Args,
 
         let mut stdin_ready = false;
         for event in events.iter() {
-            match event.token() {
-                STDIN_TOKEN => {
-                    stdin_ready = true;
-                }
-                _ => unreachable!(),
+            if event.token() == STDIN_TOKEN {
+                stdin_ready = true;
```
saiarcot895, line 336: In this case, I like the older code, because if someone later modifies this code and passes in a new token, then it should loudly complain that this event/token was not handled. #Resolved

qiluo-msft: If (old code) someone later modifies this code and passes in a new token, the loud complain does not 100% happens. And when it happens, the process crash. This is not good for online service. I use log error to complain instead.

src/sonic-supervisord-utilities-rs/src/proc_exit_listener.rs
qiluo-msft: resolved.

src/sonic-supervisord-utilities-rs/src/proc_exit_listener.rs
saiarcot895: That's fair, yeah.

saiarcot895 (approved):

deepak-singhal0408 (approved):

PR #24606, opened 2025-11-24
Remove unwrap() in sonic-ctrmgrd-rs

qiluo-msft:
#### Why I did it
unwrap() function may crash in runtime.

For constant value which is safe, we use expect() instead of unwrap(), so any future programming mistake on changing the constant value could get better crash message. Current value will not trigger crash.

Handle initialization errors gracefully

##### Work item tracking

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

- [ ] 
- [ ]

### DIFF
--- a/src/sonic-ctrmgrd-rs/crates/docker-rs/src/main.rs
+++ b/src/sonic-ctrmgrd-rs/crates/docker-rs/src/main.rs
@@ -32,13 +32,14 @@ struct Cli {
 }
 
 #[tokio::main]
-async fn main() {
-    let identity = CString::new("docker-rs").unwrap();
+async fn main() -> Result<(), container::Error> {
+    let identity = CString::new("docker-rs")
+        .map_err(|e| container::Error::Syslog(format!("invalid identity string: {}", e)))?;
     let syslog = syslog_tracing::Syslog::new(
         identity,
         syslog_tracing::Options::LOG_PID,
         syslog_tracing::Facility::Daemon
-    ).unwrap();
+    ).ok_or_else(|| container::Error::Syslog("failed to initialize syslog".to_string()))?;
     tracing_subscriber::fmt()
         .with_writer(syslog)
         .with_ansi(false)
@@ -74,9 +75,10 @@ async fn main() {
         // Don't exit with failure for wait operations when container was killed (exit code 137)
         if matches!(cli.action, Action::Wait) {
             if let container::Error::Docker(bollard::errors::Error::DockerContainerWaitError { code: 137, .. }) = e {
-                return;
+                return Ok(());
             }
         }
         std::process::exit(FAILURE);
     }
+    Ok(())
 }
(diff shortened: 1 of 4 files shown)


saiarcot895 (approved):

src/sonic-ctrmgrd-rs/crates/docker-rs/src/main.rs
```diff
@@ -32,13 +32,14 @@ struct Cli {
...
-async fn main() {
-    let identity = CString::new("docker-rs").unwrap();
+async fn main() -> Result<(), container::Error> {
+    let identity = CString::new("docker-rs")
+        .map_err(|e| container::Error::Syslog(format!("invalid identity string: {}", e)))?;
     let syslog = syslog_tracing::Syslog::new(
         identity,
         syslog_tracing::Options::LOG_PID,
         syslog_tracing::Facility::Daemon
-    ).unwrap();
```
kperumalbfn, line 41: Do we need to change other places where unwrap is used? https://github.com/sonic-net/sonic-buildimage/blob/27f12abe0fb04f727fd5d1bdd77c90a21285cd60/src/syslog-counter/src/lib.rs#L79 #Resolved

kperumalbfn: Any issues with unwrap_or, unwrap_or_else??

qiluo-msft: syslog-counter will be covered in another future PR.
unwrap_or, unwrap_or_else will not crash, so safe to use.

kperumalbfn (approved):

src/sonic-ctrmgrd-rs/crates/docker-rs/src/main.rs
saiarcot895: For the first comment, it depends on the scenario.

If we were to follow Microsoft's Pragmatic Rust Guidelines and look at [M-PANIC-IS-BUG](https://microsoft.github.io/rust-guidelines/guidelines/universal/index.html#M-PANIC-ON-BUG), then if there is a programming bug, or some situation where recovery is impossible, then we should panic. In the case of `counter.lock().unwrap()`, the [Rust documentation](https://doc.rust-lang.org/std/sync/struct.Mutex.html#method.lock) says this:


> Errors
> 
> If another user of this mutex panicked while holding the mutex, then this call will return an error once the mutex is acquired. The acquired mutex guard will be contained in the returned error.
> Panics
> 
> This function might panic when called if the lock is already held by the current thread.


This means either something else panicked, or the same thread is trying to get the mutex again. Both cases are unrecoverable, and so we should panic. So in this case, that particular usage of `unwrap()` is correct.

PR #24611, opened 2025-11-25
[202505][BGP]Preventing duplicate route reinstall

dgsudharsan:
#### Why I did it
Fixing https://github.com/sonic-net/sonic-buildimage/issues/24290

Ported fix from https://github.com/FRRouting/frr/pull/19788

##### Work item tracking

#### How I did it
Created a patch

#### How to verify it
Manual testing.

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

- [ ] 
- [ ]


dgsudharsan:
Please note that the fix for master/202511 is included in FRR upgrade https://github.com/sonic-net/sonic-buildimage/pull/24330
We are using the same numbering convention across branches and thus i retained 0064 as in master PR

StormLiangMS:
hi @dgsudharsan since it is very late for 202505, I would like to avoid unnecessary regression by this stage, would you update the description with thorough sonic-mgmt full run results? And this fix is more like an enhancement, any justification to take this in besides performance issue with static routes?

dgsudharsan:
@StormLiangMS I checked with the team internally and if MSFT doesn't have a use case for this scenario we will close this backport.

PR #24624, opened 2025-11-26
frr development workflow improvement for sonic

mike-dubrovsky:
#### Why I Did It

The standard build system is inconvenient for developers because:
- Source code hard to browse because it is unpatched after every build.
- Local changes in `src/sonic-frr/frr` are wiped out on each build.

#### How I Did It

Modified `src/sonic-frr/Makefile` to automatically:
- Apply SONiC patches once on the first build.
- Preserve the patched source tree across all subsequent builds.
- Clean only Debian build artifacts between runs (not source edits).
- Auto-detect first-time vs incremental builds.
- Auto-detect changes in sonic frr patches. 

#### How to Verify It

**First build** (creates patched branch `$(FRR_BRANCH)-$(FRR_TAG)-patched`):
```bash
make target/debs/bookworm/frr_10.3-sonic-0_amd64.deb
```

**Edit source code:**
```bash
cd src/sonic-frr/frr
vim zebra/main.c
```

**Rebuild** (preserves your changes):
```bash
make target/debs/bookworm/frr_10.3-sonic-0_amd64.deb
```

#### How It Works

**When building for the first time, or when patch inputs changed:**
The system computes a fingerprint of the patch files.
If no patched branch exists, or the fingerprint differs from the previously recorded one, 
the patched branch `$(FRR_BRANCH)-$(FRR_TAG)-patched` is rebuilt from a clean FRR base.
SONiC patches are reapplied and package metadata is refreshed.
The new fingerprint is saved in a hidden file for future builds.

**When building again with unchanged patch inputs:**
The current fingerprint matches the saved one.
The existing patched branch is reused as-is.

#### Key changes from previous revision:

The build now uses a patchset fingerprint to decide whether the patched FRR branch can be reused.

#### Compatibility

- ✅ Works with bookworm, bullseye, and buster.
- ✅ Compatible with cross-compilation (`CROSS_BUILD_ENVIRON=y`).
- ✅ Output packages are identical to original builds.
- ✅ Applies only to `sonic-frr`, doesn't affect other submodules.

#### Tested Branch

- [x] 202505

#### Description for the Changelog

Preserve sonic-frr source tree across builds for developer workflow.

### DIFF
--- a/src/sonic-frr/Makefile
+++ b/src/sonic-frr/Makefile
@@ -4,42 +4,60 @@ SHELL = /bin/bash
 
 MAIN_TARGET = $(FRR)
 DERIVED_TARGET = $(FRR_PYTHONTOOLS) $(FRR_DBG) $(FRR_SNMP) $(FRR_SNMP_DBG)
-SUFFIX = $(shell date +%Y%m%d\.%H%M%S)
-STG_BRANCH = stg_temp.$(SUFFIX)
+STG_PATCHED_BRANCH = $(FRR_TAG)-patched
+PATCHSET_SHA_FILE = ../.sonic-frr-patch-$(STG_PATCHED_BRANCH).sha1
 DPLANE_FPM_SONIC_MODULE = dplane_fpm_sonic/dplane_fpm_sonic.c
 
 # DEBEMAIL required by gpb dch
 export DEBEMAIL := sonicproject@googlegroups.com
 
 $(addprefix $(DEST)/, $(MAIN_TARGET)): $(DEST)/% :
-	# Build the package
+
+	# Prepare source tree.
+
 	pushd ./frr
-	git fetch origin tag $(FRR_TAG)
-	# Create a branch out of tag as stg needs a branch.
-	# Prefix branch name with "build-" to avoid conflicts with tag name
-	git checkout -B build-$(FRR_TAG) refs/tags/$(FRR_TAG)
+
+	# Rebuild the patched branch only when it is missing or patch inputs changed.
+	current_patch_hash=$$(cat ../patch/series ../patch/*.patch 2>/dev/null | sha1sum | awk '{print $$1}')
+	recorded_patch_hash="$$(cat $(PATCHSET_SHA_FILE) 2>/dev/null || true)"
+	patched_branch_exists=no
+	if git rev-parse --verify --quiet "refs/heads/$(STG_PATCHED_BRANCH)" >/dev/null; then
+		patched_branch_exists=yes
+	fi
+	if [[ "$$patched_branch_exists" == "no" || "$$current_patch_hash" != "$$recorded_patch_hash" ]]; then
+		git fetch origin tag $(FRR_TAG)
+		# Create a branch out of tag as stg needs a branch.
+		# Prefix branch name with "build-" to avoid conflicts with tag name.
+		git checkout -B build-$(FRR_TAG) refs/tags/$(FRR_TAG)
 ifeq ($(CROSS_BUILD_ENVIRON), y)
-	git reset --hard
+		git reset --hard
 endif
-	stg branch --create $(STG_BRANCH) build-$(FRR_TAG)
-	stg import -s ../patch/series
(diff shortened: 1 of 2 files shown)


prsunny:
@StormLiangMS , please review

Makefile.work
```diff
@@ -64,6 +64,10 @@
 #  * ENABLE_MULTIDB: Enable multiple redis database instances.
 #  *                 Default: unset
 #  *                 Values: y
+#  * DEVELOPER_MODE: Preserve source tree across builds.
```
liushilongbuaa: DEVELOPER_MODE only works for sonic-frr.
Other submodules stores patch in other folder. ex: sonic-frr.patch/series. This OPTION don't work.
How about just change src/sonic-frr/Makefile? You can check if it is the first time checkout. If first time, command1; else command2;

mike-dubrovsky: @liushilongbuaa 
Hi Liu, thanks very much for taking a look. I reworked the fix based on your review comments.

mike-dubrovsky:
@liushilongbuaa  - I did UT (direct code modification and changing FRR_TAG in rules/frr.mk) and everything looks good.

liushilongbuaa (dismissed):
LGTM. 
@qiluo-msft , please also help review.

src/sonic-frr/Makefile
```diff
@@ -4,42 +4,43 @@ SHELL = /bin/bash
...
-ifeq ($(CROSS_BUILD_ENVIRON), y)
-	git reset --hard
-endif
-	stg branch --create $(STG_BRANCH) $(FRR_TAG)
-	stg import -s ../patch/series
-	gbp dch --ignore-branch --new-version=$(FRR_VERSION)-sonic-$(FRR_SUBVERSION) --dch-opt="--force-bad-version" --commit --git-author
-	cp ../$(DPLANE_FPM_SONIC_MODULE) zebra/
+
+	# Check if this is the first time checkout.
+	if ! stg branch --list 2>/dev/null | grep -q "$(STG_PATCHED_BRANCH)"; then
```
rminnikanti: @mike-dubrovsky, wondering how this works when a new patch is added. I think it silently ignores applying new patch resulting in issues.

Say,
1. New patch is added to patch directory and patch/series file.
2. Since no change in FRR_TAG and FRR_BRANCH and the previous build created the STG_PATCHED_BRANCH branch, it just checks out the stg branch without importing the new patch.

Let me know if I missed something.

mike-dubrovsky: @rminnikanti you are right ... let me add this logic ... give me a couple of days.

mike-dubrovsky:
@rminnikanti I updated the fix based on your review comments.
New behavior: when the STG branch is created, we store the SHA signatures of all patches in a hidden file. In subsequent builds, we verify that the signatures have not changed before using the branch.

src/sonic-frr/Makefile
```diff
@@ -4,42 +4,55 @@ SHELL = /bin/bash
...
+	if [[ "$$patched_branch_exists" == "no" || "$$current_patch_hash" != "$$recorded_patch_hash" ]]; then
+		git fetch origin $(FRR_BRANCH)
+		git checkout -B $(FRR_BRANCH) origin/$(FRR_BRANCH)
+		if [[ "$$patched_branch_exists" == "yes" ]]; then
+			stg branch --delete --force $(STG_PATCHED_BRANCH)
+		fi
+		stg branch --create $(STG_PATCHED_BRANCH) $(FRR_TAG)
+		stg import -s ../patch/series
+		gbp dch --ignore-branch --new-version=$(FRR_VERSION)-sonic-$(FRR_SUBVERSION) --dch-opt="--force-bad-version" --commit --git-author
+		echo "$$current_patch_hash" > $(PATCHSET_SHA_FILE)
```
rminnikanti, line 41: @mike-dubrovsky this file shows up as an untracked item in Git, which makes the working tree appear dirty. Should we consider adding it to .gitignore?

mike-dubrovsky: @rminnikanti added to .gitignore

mike-dubrovsky:
@StormLiangMS @lipxu @liushilongbuaa Could you please review the latest version ?

PR #24651, opened 2025-11-28
[FPM]: Program SRv6 SID lists with multiple SIDs

cscarpitta:
There is a bug in dplane_fpm_sonic when handling an SRv6 SID List that contains more than one SID.

The issue occurs as follows:

- FRR creates a route with an SRv6 nexthop that includes a list of multiple SIDs.
- dplane_fpm_sonic processes this route and encodes it into a Netlink message, but only includes the first SID from the list; the remaining SIDs are omitted.
- The Netlink message, containing a SID list with only the first SID, is sent to SONiC fpmsyncd.
- SONiC fpmsyncd receives the route information, but the SID list is incomplete
- As a result, the SID list programmed into the ASIC is wrong - it contains only the first SID instead of all the SIDs provided by FRR.

This pull request fixes the issue by ensuring that dplane_fpm_sonic includes the entire SID list in the Netlink message. This guarantees that the correct, complete SID list is installed in the ApplDB and ultimately in the ASIC.

#### Why I did it

There is a bug in dplane_fpm_sonic: when a SID list contains multiple SIDs, dplane_fpm_sonic sends only the first SID to fpmsyncd. As a result, a wrong/incomplete SID list is programmed into the ApplDB and, ultimately, into the ASIC.

#### How I did it

Modify dplane_fpm_sonic to send the complete SID list to SONiC fpmsyncd, instead of sending only the first SID.

#### How to verify it

- Configure a route in FRR with a SID list that contains more than one SID:
```
ipv6 route 2001:db8:2::2::/64 Ethernet104 fcbb:bbbb:1:2:3:4:5:6/fcbb:bbbb:7:8::
```
- Check the route entry in SRV6_SID_LIST_TABLE of ApplDB and ensure it contains the full SID list:
```
sonic-db-cli APPL_DB hgetall SRV6_SID_LIST_TABLE:fcbb:bbbb:1:2:3:4:5:6|fcbb:bbbb:7:8::
{'path': 'fcbb:bbbb:1:2:3:4:5:6|fcbb:bbbb:7:8::'}
```

- [x] master
- [x] 202505

Fix programming of SRv6 SID lists with multiple SIDs.


ahsalam (approved):
LGTM.

GaladrielZhao (approved):
LGTM

src/sonic-frr/dplane_fpm_sonic/dplane_fpm_sonic.c
```diff
@@ -125,13 +125,14 @@ enum custom_rtattr_srv6_localsid {
 
 enum custom_rtattr_encap_srv6 {
 	FPM_ROUTE_ENCAP_SRV6_ENCAP_UNSPEC		= 0,
-	FPM_ROUTE_ENCAP_SRV6_VPN_SID			= 1,
+	FPM_ROUTE_ENCAP_SRV6_SIDS			= 1,
```
selva-nexthop, line 128: FPM_ROUTE_ENCAP_SRV6_ENCAP_SIDLIST -- cannot be used here? 

In the existing SONiC SRv6 design, the “SID list” is modeled as a separate object and table entry (SRV6_SID_LIST_TABLE), with a path field that contains the comma-separated list of SIDs

The controller or FRR typically provides enough information for SONiC to either reference or construct that SID list object, and Srv6Orch uses the SID list to create the SAI SRV6 SIDLIST object (SAI_SRV6_SIDLIST_ATTR_SEGMENT_LIST).

An ENCAP_SIDLIST attribute in the FPM/Netlink encoding fits that model: it is expected to represent “this route is associated with a particular SID-list policy,” not just a flat blob of addresses.

Any backward compatibility issue here?  we are changing the meaning of attribute 1 and might cause issue with older deployments.

src/sonic-frr/dplane_fpm_sonic/dplane_fpm_sonic.c
```diff
@@ -1572,8 +1573,13 @@ static bool netlink_srv6_vpn_route_msg_encode_multipath(int cmd, struct zebra_dp
 			 IPV6_MAX_BYTELEN))
 		return false;
 
-	if (!nl_attr_put(nlmsg, req_size, FPM_ROUTE_ENCAP_SRV6_VPN_SID,
-			 &nexthop->nh_srv6->seg6_segs->seg[0], IPV6_MAX_BYTELEN))
+	if (!nl_attr_put(nlmsg, req_size, FPM_ROUTE_ENCAP_SRV6_SIDS,
```
venkit-nexthop, line 1576: In this code and the one below, should we add a bounds check like below:

```
if (!nexthop->nh_srv6->seg6_segs->num_segs || \
    nexthop->nh_srv6->seg6_segs->num_segs > SRV6_MAX_SEGS)
    return false;
```

selva-nexthop:
@deepak-singhal0408 could you add this PR for routing dashboard for PR tracking.

PR #24657, opened 2025-11-29
Adding the `ttl_mode` attribute for VxLANs

mramezani95:
#### Why I did it
To allow specifying the decap TTL mode used for each VxLAN tunnel.

##### Work item tracking
- Microsoft ADO **(number only)**: 35743756

#### How I did it
Added the new attribtue `ttl_mode` for each VxLAN, which can be set to `uniform` or `pipe`.
sonic-swss PR: https://github.com/sonic-net/sonic-swss/pull/3963

#### How to verify it
Run the tests modified in this PR, or build an image with these changes and see if this attribute can be set without any errors.

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

Tested on the latest master image.

- [master] 

Adding the `ttl_mode` attribute for VxLANs.

https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-yang-models/doc/Configuration.md#vxlan

### DIFF
--- a/src/sonic-yang-models/yang-models/sonic-vxlan.yang
+++ b/src/sonic-yang-models/yang-models/sonic-vxlan.yang
@@ -59,6 +59,13 @@ module sonic-vxlan {
                 leaf dst_ip {
                     type inet:ip-address;
                 }
+
+                leaf ttl_mode {
+                    description "Decap TTL mode";
+                    type string {
+                        pattern "uniform|pipe";
+                    }
+                }
             }
         }
 
--- a/src/sonic-yang-models/doc/Configuration.md
+++ b/src/sonic-yang-models/doc/Configuration.md
@@ -2837,7 +2837,8 @@ Single tunnel example:
     "VXLAN_TUNNEL": {
         "vtep1": {
             "src_ip": "10.10.10.10",
-            "dst_ip": "12.12.12.12"
+            "dst_ip": "12.12.12.12",
+            "ttl_mode": "pipe"
         }
     },
     "VXLAN_TUNNEL_MAP": {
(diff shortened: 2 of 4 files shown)


dgsudharsan (approved):

src/sonic-yang-models/yang-models/sonic-vxlan.yang
```diff
@@ -59,6 +59,13 @@ module sonic-vxlan {
                 leaf dst_ip {
                     type inet:ip-address;
                 }
+
+                leaf ttl_mode {
```
dgsudharsan, line 63: A minor enhancement here is define a custom type in sonic-types.yang for ttl_mode and use it in this yang as well as sonic-tunnel.yang.
You can do it in this PR or in a follow up PR

src/sonic-yang-models/yang-models/sonic-vxlan.yang
```diff
@@ -59,6 +59,13 @@ module sonic-vxlan {
...
+                    description "Decap TTL mode";
+                    type string {
+                        pattern "uniform|pipe";
```
qiluo-msft, line 66: >uniform [](http://example.com/codeflow?start=33&length=7)

What is the default value? Considering the old image has no such field, the default value should align with old image behavior.

prsunny: this is similar to https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-yang-models/yang-models/sonic-tunnel.yang#L86 . If its not set, it defaults to SAI default

PR #24684, opened 2025-12-02
[yang] VNET_ROUTE_TUNNEL yang model support for consistent hashing

NavdhaJindal:
#### Why I did it
To support consistent ECMP for vxlan tunnels, according to specifications in https://github.com/sonic-net/SONiC/pull/2099

##### Work item tracking

#### How I did it
Added multi-value support for endpoint, mac and vni. Added consistent_hashing_buckets. 

#### How to verify it
Run the tests added in this PR or build image with these changes and set the new attributes.

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

- [ ] master
- [ ] 

 add consistent_hashing_buckets and multi-value endpoint, vni and mac for consistent hashing support.

https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-yang-models/doc/Configuration.md#vnet_route_tunnel

### DIFF
--- a/src/sonic-yang-models/yang-models/sonic-vnet.yang
+++ b/src/sonic-yang-models/yang-models/sonic-vnet.yang
@@ -140,7 +140,7 @@ module sonic-vnet {
                 }
 
                 leaf consistent_hashing_buckets {
-                    description "Number of bucket entries received by each nexthop";
+                    description "Number of consistent hashing buckets to use, if consistent hashing is desired";
                     type uint16;
                 }
             }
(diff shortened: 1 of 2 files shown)


src/sonic-yang-models/yang-models/sonic-vnet.yang
```diff
@@ -123,21 +123,26 @@ module sonic-vnet {
                     type stypes:sonic-ip4-prefix;
                 }
                 
-                leaf endpoint {
-                    description "Endpoint/nexthop tunnel IP";
+                leaf-list endpoint {
```
anish-n, line 126: can you double check if this works, endpoint is supposed to be a single string with comma separated set of entries, while this PR introduces it as a list with comma separated string elements within a list, which will make it non-compatible with the schema that swss/orchagent accepts.

NavdhaJindal: Yes, config_db.json will still have a single string of comma-separated values. I added endpoint, vni and macs to the LEAF_LIST_WITH_STRING_VALUE_DICT in sonic_yang_ext.py which takes care of string values in configdb that are defined as leaf-lists in the yang model. Verified via UTs and sample_config_db.json

prsunny: can you confirm its backward compatible?

NavdhaJindal: yes, the change is backward compatible- the modified fields in the config_db.json can still contain a single value, and the newly added field is optional

prsunny:
@qiluo-msft , would you help merge?

PR #24725, opened 2025-12-03
Enable vrf strict in SONiC based on compile flag. Default is off suggested by Prince.

eddieruan-alibaba:
#### Why I did it
Ubuntu VRF strict mode refers to a kernel networking behavior that enforces stricter isolation between VRFs and the main routing table. By default, it is off for backward compatible. But this is not a correct behavior for network VRF. 

For SRv6 VPN, MYSID has DT4 or DT46 action which terminate incoming encap SRv6 packet and trigger VRF lookup for inner packets. This action adds VRF strict mode check in kernel to avoid confusing. 

Discussed in routing WG, Prince suggested to still keep off as default for normal use cases and enable vrf strict only in the case needed.

Also discussed in WG, this flag can't be set via https://github.com/sonic-net/sonic-buildimage/blob/master/files/image_config/sysctl/90-sonic.conf since these flag would be set before network.target, while this vrf strick flag needs to be set after network is enabled. After network is enabled, we will enable vrf first, then set this flag. 

Therefore, we create a compile flag to enable a one time systemd service for this purpose. 

##### Work item tracking

#### How I did it
Set /proc/sys/net/vrf/strict_mode as 1 via system ctl. 

So I add a new systemctl service and enable it during ubuntu boots up.

#### How to verify it
Verify via Phoenixwing test.

admin@PE3:~$ cat /proc/sys/net/vrf/strict_mode
1

admin@PE3:~$ systemctl cat set-vrf-strict-mode
# /etc/systemd/system/set-vrf-strict-mode.service
[Unit]
Description=Set VRF strict mode after VRF module loads
After=network.target

[Service]
Type=oneshot
ExecStartPre=/sbin/modprobe vrf
ExecStart=/bin/sh -c 'echo 1 > /proc/sys/net/vrf/strict_mode'
RemainAfterExit=yes

[Install]
WantedBy=multi-user.target
admin@PE3:~$ systemctl is-active  set-vrf-strict-mode
active

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

- [ ] 
- [ ] 

n/a


cscarpitta:
@eddieruan-alibaba  Please rebase the branch.

Screenshot 2025-12-06 at 19 03 53

files/image_config/set-vrf-strict-mode/set-vrf-strict-mode.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=Set VRF strict mode after VRF module loads
+After=network.target
+
+[Service]
+Type=oneshot
+ExecStartPre=/sbin/modprobe vrf
+ExecStart=/bin/sh -c 'echo 1 > /proc/sys/net/vrf/strict_mode'
+RemainAfterExit=yes
```
venkatmahalingam, line 9: Wondering why this has to be set separately from an oneshot service rather than using the existing sysctl config  file.
https://github.com/sonic-net/sonic-buildimage/blob/master/files/image_config/sysctl/90-sonic.conf

eddieruan-alibaba: @venkatmahalingam look into boot sequence

this 90-sonic.conf would be copied via the following command /usr/lib/sysctl.d/ in file system. 
build_debian.sh:sudo cp files/image_config/sysctl/90-sonic.conf $FILESYSTEM_ROOT/usr/lib/sysctl.d/

It would be loaded via systemd-sysctl.service, before network.target.

======

local-fs.target
  ↓
systemd-sysctl.service  ← loads /usr/lib/sysctl.d/*.conf, /etc/sysctl.d/*.conf, etc.
  ↓
network.target
  ↓
multi-user.target
======

For this service, I need to wait for network.target to be completed,

After=network.target

 then install vrf first via 

ExecStartPre=/sbin/modprobe vrf

Then enable this flag

ExecStart=/bin/sh -c 'echo 1 > /proc/sys/net/vrf/strict_mode'

So I can't put this flag setting in 90-sonic.conf .

So I feel for this case, it would be clean to create a compile flag to enable it.

venkatmahalingam: Looks like if we do CONFIG_NET_VRF=y while building the kernel, you could add vrf.strict_mode=1 in the 90-sonic.conf file itself as the VRF kernel module is loaded/present by default.

If the above one is not working, I'm good with the changes present in this PR.

eddieruan-alibaba: Currently, if I enable vrf after seting this flag. The flag would be reset to zero. I feel loading vrf module would reset the flag, since by default it would set 0 to disable vrf strict as default kernel behavior. 

So even we enable this module in kernel via compile, I feel it is the same logic, since this flag would be set first, then set 0 when vrf is enabled.

venkatmahalingam (approved):

prsunny:
@StormLiangMS , for viz, could you check ms_conflict? I triggered a rerun.

kperumalbfn:
@liushilongbuaa please check the ms_conflict checker

PR #24757, opened 2025-12-05
[SmartSwitch] Add smartswitch bfb checksum file integrity validation

croos12:
Turns the sonic-nvidia-bluefield.bfb file into a tar archive with a bfb, sonic-nvidia-bluefield.bfb-intermediate and a sha256 checksum, sonic-nvidia-bluefield.bfb-intermediate.sha256 so that the contents of the bfb can be validated before shutting down an rshim when running sonic-bfb-installer.sh

#### Why I did it
There is currently a bug that any file can be used as the bfb input for sonic-bfb-installer.sh and the script does not validate if it is a proper bfb before shutting down rshims on smartswitch. This can be a problem for when trying to use the download functionality of the script since if the call to download the bfb fails the script will shutdown the rshims and attempt to install the html response page as a bfb.

#### How I did it
The bfb file is now an archive with the old bfb file, now names bfb-intermediate and a checksum using shasum256. The checksum is used to ensure that the file contents are the bfb that was intended to be installed for the sake of having a method of proving file integrity to the bfb, this is not intended to be a security feature.

#### How to verify it
Build the new bfb using the make file and attempt to install it using sonic-bfb-installer.sh.

Add smartswitch bfb checksum file integrity validation

96F7F12A-FDE2-4C6C-9005-C68D327DB226_1_105_c


bingwang-ms:
@prsunny @prabhataravind Can you help review?

prsunny:
@dgsudharsan , please review/approve. @kperumalbfn to merge

PR #24773, opened 2025-12-08
Build: Fix error printed out by Git when parsing .dep files

### LINKED ISSUE #24614: Bug: Build: "fatal: not a git repository..." errors are printed during a successful build
### Is it platform specific

generic

### Importance or Severity

Medium

### Description of the bug

When building a SONiC image, there are errors printed out to the console from a `git` command (I assume) that say `fatal` which is misleading because it makes users think that the build has failed. See this output from a successful build of the `vs` platform (though I think this affects all platforms):

```
```

This happens at both the `make configure` step of the build as well as the building of the actual image (or probably any target).

### Steps to Reproduce

Git clone, make init, etc... then:
```
make configure PLATFORM=vs
```

There should be no prints to the console that imply a fatal error if the build is not going to fail on those steps. If these build steps really have to run even though they will fail, the failure output should be suppressed (redirect to /dev/null or appropriate)

### Relevant log output

```shell

```

```shell

```

### Attach files (if any)

_No response_

nate-nexthop:
#### Why I did it

During a build or configure, there were 2 prints of this message:

```
```

This is confusing because during a build process if there is a fatal error calling `git` this probably means the build has failed. Accordingly, people were confused about errors being printed even though the rest of the build appeared to succeed.

In this case, I don't believe the build was unsound, unless one was using the DPKG Caching Framework as `Makefile.cache` appears to be the only place where these .dep files are read. I think in both cases here, the DPKG Caching Framework would refuse to cache these packages, thus the build would still be sound, but some caching would be missed.

##### Work item tracking

Fixes #24614 
#### How I did it

The fix is to ensure that we call git ls-files with a real path to the source for these packages. This way the path will be a valid git repository.

For sonic-stp.dep, in sonic-stp.mk the target created is called `STP`, not `SONIC-STP`, therefore we need to change the .dep file to match.

See: https://github.com/nexthop-ai/sonic-buildimage/blob/83e5137fcddd3feda746094d5fa4520e27588e2d/rules/sonic-stp.mk#L6

#### How to verify it

Check your build log and ensure you don't see any messages like this printed:
```
```

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

- [ ] 
- [ ]


arlakshm:
@qiluo-msft can you please help signoff on this PR

arlakshm (approved):

nate-nexthop:
Thanks @arlakshm , is the test issue a known failure? I see the optional mark on one of them, but the kvmtest-t1-lag seems to be failing when trying to download the build. However, the build passed, I'm not sure how this could affect that.
Logs from prepare:
```
 set +x
sonic-mgmt/tests/scripts/getbuild.py --platform vpp --num_asic 1 --buildid 1030349 --url_prefix mssonic/build --build_pipeline_id 2818
Try to get download url from https://dev.azure.com/mssonic/build/_apis/build/builds/1030349/artifacts?artifactName=sonic-buildimage.vpp&api-version=5.0
<snip>
urllib.error.HTTPError: HTTP Error 404: Not Found
```

nate-nexthop:
Tests are passing now, @arlakshm can you help me merge this? Thanks!

nate-nexthop:
@lguohan Could you take a look and help merge this please? Thanks!

securely1g (commented):
Review on behalf of @lguohan:

Simple fix — SONIC-STP to STP variable name correction in sonic-stp.dep. The old name does not match the variable defined in rules/sonic-stp.mk, causing git ls-files to run in wrong directory and printing errors.

LGTM — trivial one-liner fix.

PR #24811, opened 2025-12-11
Add support for xcvrd `dom_temperature_poll_interval` option

Staphylo:
#### Why I did it

Transceiver DOM update is currently too slow to update values in the database.
Especially when it comes to temperature data which is needed to make cooling decisions.
Because this case is rather specific a new thread was added to xcvrd which only polls temperature.
It doesn't do anything else in an effort to reduce the amount of resources taken for every iteration.
This feature  was implemented as part of https://github.com/sonic-net/sonic-platform-daemons/pull/685

##### Work item tracking

#### How I did it

Add the `--dom_temperature_poll_interval` to xcvrd startup  command.
Add this option in `pmon_daemon_control.json` to enable the new xcvrd thread for all Arista devices.

#### How to verify it

Verify that the `TRANSCEIVER_DOM_TEMPERATURE` table is populated in `STATE_DB` when  `xcvrd` is running and cables  support temperature reporting.

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

Enable fast xcvr temperature polling for Arista devices


device/arista/x86_64-arista_common/pmon_daemon_control.json
```diff
@@ -1,5 +1,6 @@
 {
     "skip_fancontrol": true,
-    "enable_xcvrd_sff_mgr": true
```
judyjoseph, line 3: @Staphylo can you add the xcvrd controls under "xcvrd" index eg: https://github.com/nexthop-ai/sonic-buildimage/blob/master/device/nexthop/common/pmon_daemon_control.json

Staphylo: Much better indeed

judyjoseph:
@rlhui @lguohan, please help review and merge. thx

judyjoseph:
@vmittal-msft please cherry-pick to 202511

PR #21889, opened 2025-02-28
[Bgpcfgd]Add correct terminating commands at the end of BGP config command groups

BYGX-wcr:
#### Why I did it

Previously, Bgpcfgd generated groups of BGP configuration commands that did not have proper terminating commands. As a consequence, non-BGP configuration commands following the BGP configuration command may be rejected by the FRR because of FRR's context was not properly switched. You check the details in Issue #21829.

Fix #21829 

##### Work item tracking
- Microsoft ADO **(number only)**:

#### How I did it

Add exit and exit-address-family command at the end of BGP configuration command groups.

#### Which release branch to backport (provide reason below if selected)

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

#### Tested branch (Please provide the tested image version)

- [ ] 
- [ ]

PR #24827, opened 2025-12-12
[bgpcfgd] fix IDF isolation route-map placed inside route bgp context

vishal-nokia:
#### Why I did it
Fixes https://github.com/sonic-net/sonic-buildimage/issues/23616.

[#21889](https://github.com/sonic-net/sonic-buildimage/pull/21889) adds explicit exit commands to properly terminate router bgp context. This broke the `idf_isolation` feature.
In [managers_bgp.py](https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-bgpcfgd/bgpcfgd/managers_bgp.py), the `BGPPeerGroupMgr.update_pg()` method constructs the configuration command as:
``` Python
cmd = ('router bgp %s\n' % kwargs['bgp_asn']) + pg + tsa_rm + idf_isolation_rm + "\nexit"
```
which produces a configuration like this when `idf_isolation` is set to `isolated_no_export`
``` 
router bgp 65100
  neighbor VOQ_CHASSIS_V4_PEER peer-group
  <peer-group configuration>
  <TSA route-map modifications>
route-map CHECK_IDF_ISOLATION permit 1
 match ip address prefix-list PL_LoopbackV4
 set community 12345:12345
...
exit
```
The `route-map` command is a global configuration command and cannot be defined inside `router bgp` context. With the explicit `exit` at the end, FRR treats everything between `router bgp` and `exit` as a single transaction block. The `route-map CHECK_IDF_ISOLATION` definition is silently rejected, and subsequent match commands fail with "Unknown command" because they appear outside any `route-map` context.
```
line 112: % Unknown command[28]: match ip address prefix-list PL_LoopbackV4
line 113: % Unknown command[28]: set community 12345:12345
```
##### Work item tracking

#### How I did it
Move `idf_isolation_rm` after the exit statement so that IDF isolation route-maps are defined at global configuration level.
This produces the correct configuration order.
``` PlainText
router bgp 65100
  neighbor VOQ_CHASSIS_V4_PEER peer-group
  <peer-group configuration>
  <TSA route-map modifications>
exit
route-map CHECK_IDF_ISOLATION permit 1
 match ip address prefix-list PL_LoopbackV4
 set community 12345:12345
```
#### How to verify it
Run
```
sudo idf_isolation isolated_no_export
sudo config save -y
sudo config reload -y
```
All neighbors will come up.

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [x] 202511

- [x] master

 move idf_isolation_rm after exit statement so that IDF isolation route_maps are defined at global configuration level


vishal-nokia:
@tjchadaga for viz.

src/sonic-bgpcfgd/bgpcfgd/managers_bgp.py
```diff
@@ -66,9 +66,9 @@ def update_pg(self, name, **kwargs):
             return False
 
         if kwargs['vrf'] == 'default':
-            cmd = ('router bgp %s\n' % kwargs['bgp_asn']) + pg + tsa_rm + idf_isolation_rm + "\nexit"
+            cmd = ('router bgp %s\n' % kwargs['bgp_asn']) + pg + tsa_rm + "\nexit" + idf_isolation_rm
```
tjchadaga, line 69: @vishal-nokia - This way of adding exits blindly is error prone. Could you please move it to the template instead? Please include exits after route-maps in 

bgpd.tsa.isolate.conf.j2
idf_isolate.conf.j2
idf_unisolate.conf.j2

tjchadaga: + @BYGX-wcr

vishal-nokia: @tjchadaga the exit statements were added by the other PR https://github.com/sonic-net/sonic-buildimage/pull/21889, not by me, which is leading to bad configuration when idf_isolation is set to isolated_no_export.

I've added the detailed RCA in https://github.com/sonic-net/sonic-buildimage/issues/23616.

I agree that adding exits is what led to this issue. The correct fix would be to rework/revert the existing PR and add the exits in the templates itself.
I simply moved the idf_isolate_rm patch to come after the exit, to fix the current issue.

vishal-nokia:
@saksarav-nokia for viz.

src/sonic-bgpcfgd/bgpcfgd/managers_bgp.py
BYGX-wcr: I think we discussed modifying the template at that time but it turned out that the templates are too fragmented and hard to touch

tjchadaga:
@vishal-nokia - are you seeing this issue for isolated_withdraw_all?

vishal-nokia:
@tjchadaga we only see this issue for `isolated_no_export`. Sonic-mgmt tests pertaining to `isolated_withdraw_all` have been passing consistently.

src/sonic-bgpcfgd/bgpcfgd/managers_bgp.py
tjchadaga: @vishal-nokia - the issue in https://github.com/sonic-net/sonic-buildimage/issues/23616 is seen for isolated_no_export alone and not for isolated_withdraw_all or TSA, mainly because the template for isolated_no_export has a "no route-map" at the end, which switches the frr config context out of route-map.

sonic(config)# route-map CHECK_IDF_ISOLATION permit 10
sonic(config-route-map)# set community no-export additive
sonic(config-route-map)# 
sonic(config-route-map)# 
sonic(config-route-map)# no route-map CHECK_IDF_ISOLATION deny 4
sonic(config)# <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<

Now the 'exit' command added by https://github.com/sonic-net/sonic-buildimage/pull/21889 takes it out of the config mode into enable mode where other commands fail.

sonic(config)# exit
sonic#

I think https://github.com/sonic-net/sonic-buildimage/pull/25182 should avoid the issue. Could you please confirm?

vishal-nokia: Confirmed that #25182 resolves the issue. My original fix of moving `idf_isolation_rm` after `exit` also resolved the issue since FRR ends up in config mode, but your approach of reordering the template commands is cleaner as it keeps the context management within the template itself. Thank you for fixing this.

vishal-nokia:
Closing this PR in favor of #25182

PR #24840, opened 2025-12-13
Use snat to set source ip for inbound dpu mgmt traffic

yue-fred-gao:
#### Why I did it
The current implementation uses masquerade that changes source IP automatically based on the egress interface kernel chooses. However, DPUs interfaces are added to a vlan and ip is set on the bridge. kernel is not able to find the proper egress interface. It uses eth0 management IP instead. After we remove the default route over midplane interface in DPU, there is no default route until it is configured (the default route should be over Ethernet interface). Direct ssh to DPU through NAT stops working because there is no route for return traffic in DPU.

Instead of relying on default route, we should set source IP to the address of bridge-midplane so return traffic can follow the same midplane interface.
##### Work item tracking

#### How I did it
Change masquerade to snat and set source ip to bridge-midplane address.
#### How to verify it
Run command "sonic-dpu-mgmt-traffic.sh inbound -e --dpus all --ports 9090,9091,9092,9093,9094,9095,9096,9097" in smartswitch first. Then "ssh admin@<switch-ip> -p 9090" from outside of the switch and verify ssh successfully into DPU0.

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511

- [ ] 
- [ ]


zjswhhh (approved):

files/scripts/sonic-dpu-mgmt-traffic.sh
```diff
@@ -184,7 +184,7 @@ ctrl_dpu_ib_forwarding(){
         dpu_name="${sel_dpu_names[$index]}"
         dpu_midplane_ip="${midplane_ip_dict[$dpu_name]}"
         switch_port="${provided_ports[$index]}"
-        add_rem_valid_iptable $op nat POSTROUTING -p tcp -d $dpu_midplane_ip --dport $dest_port -j MASQUERADE
+        add_rem_valid_iptable $op nat POSTROUTING -p tcp -d $dpu_midplane_ip --dport $dest_port -j SNAT --to-source $midplane_gateway
```
kperumalbfn, line 187: @yue-fred-gao If midplane-gateway IP gets updated through yang or config, does that flow trigger this script to change the source-ip?

yue-fred-gao: Hi @kperumalbfn, no, it doesn't. This script is run manually before starting dash sonic-mgmt tests. It could be integrated in the test case and started during setup. But it is not a daemon process and won't be triggered automatically by config change. And I don't think it is required because as long as the script is run again before the test, the change will be reflected. If there is a test case that changes midplane gateway today (I don't think there is), it needs to call the script to update the rules.

kperumalbfn: Thanks @yue-fred-gao

kperumalbfn (approved):

kperumalbfn:
@vmittal-msft please add to 202511 branch

congh-nvidia:
Hi @prabhataravind , we also need backport to 202511 the following PRs which are related to this one:
https://github.com/sonic-net/sonic-buildimage/pull/24801
https://github.com/sonic-net/sonic-mgmt/pull/21393
Could you please also approve those 2?

PR #24848, opened 2025-12-13
[FPM]: Fix missing SRv6 encapsulation source address

cscarpitta:
In FRR, you can steer IPv4 and IPv6 traffic over an SRv6 SID list.

Here is an example of traffic steering configuration:

```
ipv6 route 2001:db8:2:2::/64 Ethernet16 segments fcbb:bbbb:fe10:fe20::
!
segment-routing
 srv6
  encapsulation
   source-address fcbb:bbbb:1::1
```

With this configuration, the system should encapsulate IPv6 traffic destined for 2001:db8:2:2::/64 with an outer IPv6 header, using fcbb:bbbb:1::1 as the source address. The user expects all encapsulated packets to use this configured source address.

However, there is currently an issue in FPM: when programming the steer route to SONiC fpmsyncd, FPM does not include the configured source address. As a result, the route programmed into ApplDB has an incorrect encapsulation source address, causing data plane traffic to use the wrong source address.

This PR fixes the issue by making sure the configured source address is also sent to fpmsyncd.

#### Why I did it

There is an issue in FPM: FPM programs routes with the wrong encapsulation source address.

#### How I did it

FPM has been extended to send the correct source address to fpmsyncd.

#### How to verify it

A new sonic-mgmt test suite has been added to verify the programming of the IPv4/IPv6 traffic steering: https://github.com/sonic-net/sonic-mgmt/pull/21689


GaladrielZhao:
The code looks good, but it seems the test https://github.com/sonic-net/sonic-mgmt/pull/21689 does not cover the encapsulation source address confirmation?

PR #24849, opened 2025-12-14
[bgpcfgd]: Fix missing SRv6 encapsulation source address

cscarpitta:
Currently, there is no way to configure the source address for SRv6 encapsulation in bgpcfgd. As a result, FRR defaults to using an incorrect source address.

This incorrect source address is then pushed to SONiC fpmsyncd, leading to the data plane using the wrong source address in encapsulated packets.

To resolve this issue, this PR extens bgpcfgd to allow users to configure the SRv6 encapsulation source address.

With this enhancement, the correct source address will be programmed and used throughout the control and data plane.

#### Why I did it

There is an issue in bgpcfgd: bgpcfgd does not allow users to configure the source address for SRv6 encapsulation.
As a result, SONiC may use an invalid source address.

#### How I did it

Extended bgpcfgd to allow users to configure the source address for SRv6 encapsulation.

#### How to verify it

A new sonic-mgmt test suite has been added: https://github.com/sonic-net/sonic-mgmt/pull/21689


GaladrielZhao (approved):
LGTM

PR #24850, opened 2025-12-14
[bgpcfgd]: Add sidlist in static route programming 

cscarpitta:
Currently, bgpcfgd allows you to configure static routes, but it does not provide a way to steer these routes over an SRv6 SID list.

This PR extends bgpcfgd to enable users to steer a route over an SRv6 SID list.

#### Why I did it

bgpcfgd does not provide a way to steer these routes over an SRv6 SID list.

#### How I did it

Extended bgpcfgd to allow users to configure the sidlist.

#### How to verify it

A new sonic-mgmt test suite has been added: https://github.com/sonic-net/sonic-mgmt/pull/21689


GaladrielZhao (approved):
Looks good to me.

PR #24882, opened 2025-12-17
[Mellanox] Updated parse_vpd_data() to use inotify to wait for vpd_data creation

zili11720:
#### Why I did it
This fixes a race condition where chassis.py attempts to access the VPD data file before it's available during system initialization. When the file doesn't exist yet, _parse_vpd_data() returns an empty dictionary, causing the revision field to be populated with "N/A" in Redis instead of the actual hardware revision value.

##### Work item tracking

#### How I did it
Update _parse_vpd_data() to use inotify mechanism to wait for VPD data file creation
before accessing it.

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [x] 202511

- [ ] 
- [ ]


platform/mellanox/mlnx-platform-api/sonic_platform/chassis.py
```diff
@@ -968,6 +974,48 @@ def get_revision(self):
 
         return self.vpd_data.get(REVISION, "N/A")
 
+    def _wait_for_file_creation(self, file_path, timeout):
```
keboliu: Suggest adding this function to this file, https://github.com/sonic-net/sonic-buildimage/blob/master/platform/mellanox/mlnx-platform-api/sonic_platform/utils.py, it will be convinient to reuse this function in other files.

keboliu: is it possible to add some unit tests for the new code?

zili11720: Done.

zili11720:
This PR is replaced by:
https://github.com/sonic-net/sonic-buildimage/pull/25879

PR #24898, opened 2025-12-18
SONiC support for BMC cards based on Aspeed AST2720

chinmoy-nexthop:
#### Why I did it
We need to be able to run SONIC on BMC cards based on AST2720 (ARM64 SOC)

#### How I did it
- New Aspeed platform for ARM64 architecture
- Kernel drivers merged via separate PR to sonic-linux-kernel repo.
- Build infrastructure: platform/aspeed + device/{aspeed, nexthop ...}
- Single sonic-aspeed-arm64.bin target image across all vendors using
  this SOC
- Initial platform infrastructure under platform/aspeed and
  device/{aspeed, nexthop...} directories
- SONiC installer support with single partition and
  multiple directory-based installs along with updating U-Boot environment
  variables for proper boot management
- Ability for vendors to bundle their card specific packages, kernel modules
  etc to the image and install them at run time (first boot) using the existing
  SONiC lazy install infra.
- Removed many unwanted dockers and .debs not relevant to the BMC platform

     

- [ ] 202411
- [ ] 202505

- [ ] 
- [ ]

### DIFF
--- a/device/aspeed/arm64-aspeed_ast2700_evb-r0/99-switchCpu.rules
+++ b/device/aspeed/arm64-aspeed_ast2700_evb-r0/99-switchCpu.rules
@@ -1,0 +1,4 @@
+# Aspeed AST2700 EVB: Dummy placeholder for switch CPU console
+# The EVB does not have a physical switch CPU console connection
+# This file exists to prevent errors in sonic-platform-init.sh
+
--- a/device/aspeed/arm64-aspeed_ast2700_evb-r0/switch_cpu_console.conf
+++ b/device/aspeed/arm64-aspeed_ast2700_evb-r0/switch_cpu_console.conf
@@ -1,0 +1,12 @@
+# Console configuration for Aspeed AST2700 EVB
+# This file defines console port settings for the switch CPU console
+
+# Baud rate for the switch CPU console (default: 115200)
+CONSOLE_BAUD_RATE=115200
+
+# Flow control: 0=disabled, 1=enabled (default: 0)
+CONSOLE_FLOW_CONTROL=0
+
+# Remote device name shown in consutil
+CONSOLE_REMOTE_DEVICE=EVB
+
--- a/device/aspeed/arm64-aspeed_ast2700_evb-r0/udevprefix.conf
+++ b/device/aspeed/arm64-aspeed_ast2700_evb-r0/udevprefix.conf
@@ -1,0 +1,2 @@
+ttySwitchCpu
+
--- a/device/nexthop/arm64-nexthop_b27-r0/99-switchCpu.rules
+++ b/device/nexthop/arm64-nexthop_b27-r0/99-switchCpu.rules
@@ -1,0 +1,3 @@
+# Nexthop B27 BMC: Map host CPU console /dev/ttyS1 to common alias /dev/ttySwitchCpu0
+SUBSYSTEM=="tty", KERNEL=="ttyS1", SYMLINK+="ttySwitchCpu0", MODE="0666"
+
--- a/device/nexthop/arm64-nexthop_b27-r0/switch_cpu_console.conf
+++ b/device/nexthop/arm64-nexthop_b27-r0/switch_cpu_console.conf
@@ -1,0 +1,12 @@
+# Console configuration for Nexthop B27 BMC
+# This file defines console port settings for the switch CPU console
+
+# Baud rate for the switch CPU console (default: 115200)
+CONSOLE_BAUD_RATE=9600
+
+# Flow control: 0=disabled, 1=enabled (default: 0)
+CONSOLE_FLOW_CONTROL=0
+
+# Remote device name shown in consutil
+CONSOLE_REMOTE_DEVICE=SwitchCpu
+
--- a/device/nexthop/arm64-nexthop_b27-r0/udevprefix.conf
+++ b/device/nexthop/arm64-nexthop_b27-r0/udevprefix.conf
@@ -1,0 +1,2 @@
+ttySwitchCpu
+
--- a/platform/aspeed/scripts/sonic-switchcpu-console-init.sh
+++ b/platform/aspeed/scripts/sonic-switchcpu-console-init.sh
@@ -1,0 +1,98 @@
+#!/bin/bash
+# Aspeed Switch CPU Console Initialization Script
+# This script configures the switch CPU console for consutil access
+
+set -e
+
+LOG_FILE="/var/log/aspeed-switchcpu-console-init.log"
+PLATFORM_SYMLINK="/usr/share/sonic/platform"
+
+log() {
+    echo "[$(date '+%Y-%m-%d %H:%M:%S')] $*" | tee -a "$LOG_FILE"
+}
+
+log "Starting switch CPU console initialization..."
+
+# Verify platform symlink exists
+if [ ! -L "$PLATFORM_SYMLINK" ]; then
+    log "ERROR: Platform symlink $PLATFORM_SYMLINK does not exist"
+    exit 1
+fi
+
+RULES_SRC="${PLATFORM_SYMLINK}/99-switchCpu.rules"
+RULES_DEST="/etc/udev/rules.d/99-switchCpu.rules"
+CONSOLE_CONF="${PLATFORM_SYMLINK}/switch_cpu_console.conf"
+
+# Set default console parameters
+CONSOLE_BAUD_RATE=115200
+CONSOLE_FLOW_CONTROL=0
+CONSOLE_REMOTE_DEVICE=SwitchCpu
+
+# Load platform-specific console configuration if available
+if [ -f "$CONSOLE_CONF" ]; then
+    log "Loading console configuration from $CONSOLE_CONF"
+    # shellcheck disable=SC1090
+    . "$CONSOLE_CONF"
+    log "Console settings: baud=$CONSOLE_BAUD_RATE, flow=$CONSOLE_FLOW_CONTROL, device=$CONSOLE_REMOTE_DEVICE"
+else
+    log "No console.conf found, using defaults"
+fi
+
+# Check if platform provides switch CPU console rules
+if [ ! -f "$RULES_SRC" ]; then
+    log "No switch CPU console rules found at $RULES_SRC, skipping console setup"
+    exit 0
+fi
+
+log "Found switch CPU console rules, configuring console..."
+
+# Symlink udev rules
+ln -sf "$RULES_SRC" "$RULES_DEST"
+log "Created symlink: $RULES_DEST -> $RULES_SRC"
+
+# Enable console feature
+log "Enabling console feature..."
+if config console enable; then
+    log "Console feature enabled via config command"
+else
+    log "WARNING: config console enable failed, trying direct DB write..."
+fi
+
+# Verify and ensure it's actually enabled in CONFIG_DB
+if sonic-db-cli CONFIG_DB HSET "CONSOLE_SWITCH|console_mgmt" enabled yes; then
+    log "Console feature enabled in CONFIG_DB"
+else
+    log "ERROR: Failed to enable console feature in CONFIG_DB"
+    exit 1
+fi
+
+# Reload udev to activate the rules
+log "Reloading udev rules..."
+udevadm control --reload-rules
+udevadm trigger
+log "Udev rules reloaded and triggered"
+
+# Configure CONSOLE_PORT|0 with platform settings
+log "Configuring CONSOLE_PORT|0 in CONFIG_DB..."
+
+# Check if entry exists
+if sonic-db-cli CONFIG_DB HGETALL "CONSOLE_PORT|0" >/dev/null 2>&1; then
+    log "CONSOLE_PORT|0 already exists, deleting to ensure correct settings..."
+    sonic-db-cli CONFIG_DB DEL "CONSOLE_PORT|0"
+fi
+
+# Create with platform-specific settings
+log "Creating CONSOLE_PORT|0 with baud=$CONSOLE_BAUD_RATE, flow=$CONSOLE_FLOW_CONTROL, device=$CONSOLE_REMOTE_DEVICE"
+if sonic-db-cli CONFIG_DB HMSET "CONSOLE_PORT|0" \
+    baud_rate "$CONSOLE_BAUD_RATE" \
+    flow_control "$CONSOLE_FLOW_CONTROL" \
+    remote_device "$CONSOLE_REMOTE_DEVICE"; then
+    log "Created CONSOLE_PORT|0 successfully"
+else
+    log "ERROR: Failed to create CONSOLE_PORT|0"
+    exit 1
+fi
+
+log "Switch CPU console initialization complete"
+exit 0
+
--- a/platform/aspeed/systemd/sonic-switchcpu-console-init.service
+++ b/platform/aspeed/systemd/sonic-switchcpu-console-init.service
@@ -1,0 +1,17 @@
+[Unit]
+Description=Aspeed Switch CPU Console Initialization
+Documentation=https://github.com/sonic-net/sonic-buildimage
+After=database.service sonic-platform-init.service config-setup.service
+Requires=database.service
+PartOf=sonic.target
+
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/sonic-switchcpu-console-init.sh
+RemainAfterExit=yes
+StandardOutput=journal
+StandardError=journal
+
+[Install]
+WantedBy=multi-user.target sonic.target
+


files/build_templates/sonic_debian_extension.j2
```diff
@@ -1221,4 +1221,40 @@ sudo touch $FILESYSTEM_ROOT_ETC_SONIC/enable_multidb
...
+# PLATFORM-SPECIFIC SERVICE FILTERING
+# This section MUST remain at the end of the file to ensure all services
+# have been added to GENERATED_SERVICE_FILE before filtering.
+# DO NOT add service registrations (tee -a $GENERATED_SERVICE_FILE) after this point!
+#############################################################################
+
+# Aspeed BMC platform: Filter services for BMC use case
+if [[ $CONFIGURED_PLATFORM == aspeed ]]; then
+    # Show original service count
+    ORIGINAL_COUNT=$(wc -l < $GENERATED_SERVICE_FILE)
```
nats-nokia: Build error is observed.
```
+ sudo rm -f syslog-counter_1.0.0_arm64.deb
+ [[ aspeed == aspeed ]]
++ wc -l
./sonic_debian_extension.sh: line 1288: ./fsroot-aspeed/etc/sonic/generated_services.conf: Permission denied
+ ORIGINAL_COUNT=
+ clean_sys
```


This line requires sudo to resolve
 ORIGINAL_COUNT=$(sudo bash -c "wc -l < $GENERATED_SERVICE_FILE")

chinmoy-nexthop: @nats-nokia , thank you for the feedback. somehow missed notification earlier. We’ll look into this and provide an update soon.

chinmoy-nexthop: Yes, adding sudo would improve consistency and help safeguard against potential umask changes. Thank you for pointing this out.

It was not identified earlier because, the file is created using sudo tee (for example, lines 413 and 418), which results in a root owned file with default permissions of 644 (rw-r--r--). The ubuntu user falls under the “others” category and therefore has read permission .However, if a stricter umask were in effect (for example, 077), the file would be created with permission 600 (owner-only), and the command would fail without sudo. We will push the change.

```
[05:01:53] ubuntu@ip-172-31-7-82:~/private-sonic-buildimage (chinmoy-aspeed)  
$  ls -l fsroot-aspeed/etc/sonic/generated_services.conf
-rw-r--r-- 1 root root 433 Jan 12 15:53 fsroot-aspeed/etc/sonic/generated_services.conf
[05:02:43] ubuntu@ip-172-31-7-82:~/private-sonic-buildimage (chinmoy-aspeed)  
$  wc -l < fsroot-aspeed/etc/sonic/generated_services.conf
19
[05:03:14] ubuntu@ip-172-31-7-82:~/private-sonic-buildimage (chinmoy-aspeed)  
$  sudo bash -c "wc -l < fsroot-aspeed/etc/sonic/generated_services.conf"
19
[05:03:49] ubuntu@ip-172-31-7-82:~/private-sonic-buildimage (chinmoy-aspeed)  
$  whoami
ubuntu
[05:04:36] ubuntu@ip-172-31-7-82:~/private-sonic-buildimage (chinmoy-aspeed)  
$  groups
ubuntu adm dialout cdrom floppy sudo audio dip video plugdev netdev lxd docker
[05:04:40] ubuntu@ip-172-31-7-82:~/private-sonic-buildimage (chinmoy-aspeed)  
$  sudo -n true
[05:04:53] ubuntu@ip-172-31-7-82:~/private-sonic-buildimage (chinmoy-aspeed)  
$  umask
0002
[05:08:53] ubuntu@ip-172-31-7-82:~/private-sonic-buildimage (chinmoy-aspeed) 
$  sudo sh -c 'umask' <===== Created with this umask
0022
```

platform/aspeed/platform_arm64.conf
```diff
@@ -0,0 +1,348 @@
...
+# Global defines
+VAR_LOG=4096  # 4GB for /var/log
+
+# SONiC filesystem definitions (must match installer/install.sh)
+DOCKERFS_DIR=docker
+FILESYSTEM_DOCKERFS=dockerfs.tar.gz
+FILESYSTEM_SQUASHFS=fs.squashfs
+
+# Kernel configuration
+kernel_version=6.12.41+deb13-sonic-arm64
```
kkosciuszkiewicz-arista: We should not hardcode kernel version. Best to remove lines 15-17.

platform/aspeed/platform_arm64.conf
```diff
@@ -0,0 +1,348 @@
...
+prepare_boot_menu() {
+    echo "Sync up cache ..."
+    sync
+    echo "Setting up U-Boot environment for Aspeed AST2700..."
+
+    # Configure U-Boot environment access first
+    # This function detects MTD, device tree, or uses eMMC default
+    configure_uboot_env
+
+    image_name=${image_dir}${kernel_fname}
```
kkosciuszkiewicz-arista: Variables defined here are unused.

platform/aspeed/platform_arm64.conf
```diff
@@ -0,0 +1,348 @@
...
+fi
+
+echo "Installing SONiC from ${install_env:-unknown} on Platform $PLATFORM"
+
+# AST2700 platform configuration
+disk_interface="mmc"
+mmc_bus="mmc0:0001"  # Based on kernel log: mmc0:0001 008GB1
+
+# Device tree and FIT image configuration
+fdt_fname="/usr/lib/linux-image-${kernel_version}/aspeed/${dtb_name}"
```
kkosciuszkiewicz-arista: `fdt_fname` is unused

platform/aspeed/platform_arm64.conf
chander-nexthop: Done. Thanks for pointing out. Ditto for the rest

platform/aspeed/platform_arm64.conf
chander-nexthop: Done.

platform/aspeed/platform_arm64.conf
chander-nexthop: Done.

platform/aspeed/scripts/sonic-uboot-env-init.sh
```diff
@@ -0,0 +1,195 @@
...
+fw_setenv sonic_version_1 "$SONIC_VERSION" || log "ERROR: Failed to set sonic_version_1"
+
+# Old/backup image (none for first installation)
+fw_setenv image_dir_old "" || log "ERROR: Failed to set image_dir_old"
+fw_setenv fit_name_old "" || log "ERROR: Failed to set fit_name_old"
+fw_setenv sonic_version_2 "None" || log "ERROR: Failed to set sonic_version_2"
+fw_setenv linuxargs_old "" || log "ERROR: Failed to set linuxargs_old"
+
+# Kernel command line arguments
+fw_setenv linuxargs "console=ttyS12,115200n8 earlycon=uart8250,mmio32,0x14c33b00 loopfstype=squashfs loop=$IMAGE_DIR/fs.squashfs varlog_size=4096" || log "ERROR: Failed to set linuxargs"
```
oleksandrivantsiv: Console configuration should be provided by the installer.conf (e.g. device/nexthop/arm64-nexthop_b27-r0/installer.conf) file to make this file platform-independent

platform/aspeed/scripts/sonic-uboot-env-init.sh
```diff
@@ -0,0 +1,195 @@
...
+
+# Boot menu with instructions
+fw_setenv print_menu "echo ===================================================; echo SONiC Boot Menu; echo ===================================================; echo To boot \$sonic_version_1; echo   type: run sonic_image_1; echo   at the U-Boot prompt after interrupting U-Boot when it says; echo   \\\"Hit any key to stop autoboot:\\\" during boot; echo; echo To boot \$sonic_version_2; echo   type: run sonic_image_2; echo   at the U-Boot prompt after interrupting U-Boot when it says; echo   \\\"Hit any key to stop autoboot:\\\" during boot; echo; echo ===================================================" || log "ERROR: Failed to set print_menu"
+
+# Boot configuration
+fw_setenv boot_next "run sonic_image_1" || log "ERROR: Failed to set boot_next"
+fw_setenv bootcmd "run print_menu; test -n \"\$boot_once\" && setenv do_boot_once \"\$boot_once\" && setenv boot_once \"\" && saveenv && run do_boot_once; run boot_next" || log "ERROR: Failed to set bootcmd"
+
+# Memory addresses
+fw_setenv loadaddr "0x432000000" || log "ERROR: Failed to set loadaddr"
```
oleksandrivantsiv, line 223: Offsets are hardcoded in two places - here and platform/aspeed/platform_arm64.conf file. Is it possible to have them in one place? How `platform/aspeed/platform_arm64.conf` is used?

platform/aspeed/platform_arm64.conf
```diff
@@ -0,0 +1,384 @@
+#!/bin/sh
+# Implementation of installation behavior for Aspeed AST2700 BMC platform
```
oleksandrivantsiv, line 2: I'm not sure I understand the difference between `platform/aspeed/platform_arm64.conf` and `platform/aspeed/scripts/sonic-uboot-env-init.sh`. From the implementation, it looks like both perform similar tasks, but it’s not clear where `platform_arm64.conf` is actually invoked.

platform/aspeed/sonic-platform-modules-nexthop/common/systemd/usb-network-init.service
```diff
@@ -0,0 +1,14 @@
+[Unit]
```
oleksandrivantsiv, line 1: In which context will this service be started?

platform/aspeed/scripts/sonic-uboot-env-init.sh
chander-nexthop: This is a stop gap, till we have ONIE/SONIE support. The first install on eMMC, done via writing the disk image to eMMC doesn't set the uboot environment for dual sonic image and hence we have this service here. Once ONIE/SONIE support is added, PART-2 of this script can be done away with

platform/aspeed/scripts/sonic-uboot-env-init.sh
chander-nexthop: The function bootloader_menu_config() is invoked from installer/install.sh

platform/aspeed/sonic-platform-modules-nexthop/common/systemd/usb-network-init.service
chander-nexthop: This is a systemd service. I added it here as network connectivity between the BMC card and switch cpu over a USB network device was a requirement

platform/aspeed/platform_arm64.conf
chander-nexthop: platform_arm64.conf is used when
1. sonic-installer install <new_image> is invoked (the new images platform_arm64.conf is used for the function implementations).
2. The script we have to create a disk image (updated to the PR, see build-emmc-installer.sh).
As mentioned earlier, we use PART2 of tje sonic-uboot-env-init.sh to setup uboot environment for 2 sonic image support when we boot for the first time

platform/aspeed/scripts/sonic-uboot-env-init.sh
```diff
@@ -0,0 +1,195 @@
...
+    fi
+else
+    echo "$FW_ENV_CONFIG" > /etc/fw_env.config
+    log "Created /etc/fw_env.config: $FW_ENV_CONFIG"
+fi
+
+#############################################################################
+# PART 2: Set U-Boot environment variables (runs ONCE globally)
+#############################################################################
+
```
nats-nokia, line 100: Part 2 should be skipped for ONIE/SONIE based systems.

if blkid -L "ONIE" >/dev/null 2>&1 || blkid -L "ONIE-BOOT" >/dev/null 2>&1; then
        echo "ONIE parition found. skipping uboot environment configurations."
        exit 0
fi

platform/aspeed/platform_arm64.conf
```diff
@@ -0,0 +1,384 @@
...
+
+    # Standard SONiC approach: Single partition with multiple image directories
+    # sonic_image_1 = current/new image
+    # sonic_image_2 = previous image (for rollback)
+
+    if [ "$install_env" = "onie" ] || [ "$install_env" = "build" ]; then
+        # First installation - initialize boot configuration
+        echo "First installation: Setting up boot configuration for image directory: $image_dir"
+
+        fw_setenv  image_dir_old ""
```
nats-nokia, line 207: we need to add option -f fw_setenv to avoid manual confirmation in ONIE.
if [ "$install_env" = "onie" ] || [ "$install_env" = "build" ]; then
  FW_ARG="-f"
  fw_setenv  ${FW_ARG} image_dir_old ""
......( needed in all fw_setenv lines just like existing sonic platform_arm64.conf)

platform/aspeed/platform_arm64.conf
```diff
@@ -0,0 +1,384 @@
...
+                CURR_IMAGE_DIR=$(echo "$CURR_SONIC_IMAGE" | sed 's/SONiC-OS-/image-/')
+            fi
+
+            if [ -n "$CURR_IMAGE_DIR" ] && [ -d "/host/$CURR_IMAGE_DIR" ]; then
+                image_dir_old="$CURR_IMAGE_DIR"
+                fit_name_old="$CURR_IMAGE_DIR/boot/sonic_arm64.fit"
+                sonic_version_2="$CURR_SONIC_IMAGE"
+
+                # Reconstruct linuxargs from current boot
+                linuxargs_old="console=ttyS12,115200n8 $EARLYCON loopfstype=squashfs loop=$CURR_IMAGE_DIR/$FILESYSTEM_SQUASHFS varlog_size=$VAR_LOG"
```
nats-nokia: We can use the CONSOLE variables which are initialized at the beginning. This will enable platform specific settings later.

linuxargs_old="console=$CONSOLE_PORT,${CONSOLE_SPEED}n8 $EARLYCON loopfstype=squashfs loop=$CURR_IMAGE_DIR/$FILESYSTEM_SQUASHFS varlog_size=$VAR_LOG"

platform/aspeed/platform_arm64.conf
```diff
@@ -0,0 +1,384 @@
...
+    # Boot menu with instructions
+    BORDER='echo "===================================================";'
+    TITLE='echo "SONiC Boot Menu";'
+    BOOT1='echo "To boot $sonic_version_1";echo "  type: run sonic_image_1";echo "  at the U-Boot prompt after interrupting U-Boot when it says";echo "  \"Hit any key to stop autoboot:\" during boot";echo;'
+    BOOT2='echo "To boot $sonic_version_2";echo "  type: run sonic_image_2";echo "  at the U-Boot prompt after interrupting U-Boot when it says";echo "  \"Hit any key to stop autoboot:\" during boot";echo;'
+    FOOTER='echo "===================================================";'
+    fw_setenv  print_menu "$BORDER $TITLE $BORDER $BOOT1 $BOOT2 $FOOTER"
+
+    # Boot arguments for current image (linuxargs contains loop path)
+    fw_setenv  linuxargs "console=ttyS12,115200n8 $EARLYCON loopfstype=squashfs loop=$image_dir/$FILESYSTEM_SQUASHFS varlog_size=$VAR_LOG"
```
nats-nokia: We can use CONSOLE variables same as above.
fw_setenv  ${FW_ARG} linuxargs "console=$CONSOLE_PORT,${CONSOLE_SPEED}n8 $EARLYCON loopfstype=squashfs loop=$image_dir/$FILESYSTEM_SQUASHFS varlog_size=$VAR_LOG"

platform/aspeed/platform_arm64.conf
```diff
@@ -0,0 +1,384 @@
...
+
+    # Boot commands - always load from partition 1 (SONiC-OS)
+    DISK_LOAD='ext4load mmc 0:1 $loadaddr $fit_name'
```
nats-nokia: Instead of hardcoding partition as 1, please use ${demo_part} variable which is initialized at L.No:204 above.

like,
DISK_LOAD='ext4load '$disk_interface' 0:'$demo_part' ${loadaddr} ${fit_name}'
DISK_LOAD_OLD='ext4load '$disk_interface' 0:'$demo_part' ${loadaddr} ${fit_name_old}'

platform/aspeed/scripts/sonic-uboot-env-init.sh
chander-nexthop: Will do.

platform/aspeed/platform_arm64.conf
chander-nexthop: Fixed. Will update the PR. Applies to the comments below as well. regarding the use of -F, I will check and revert

platform/aspeed/scripts/sonic-uboot-env-init.sh
oleksandrivantsiv: While the decision about ONIE/SONiE is not final, and there is no clear understanding of if or when this change will happen, I suggest making the proposed implementation clean and free of duplication. The serial console parameters are currently hardcoded in multiple places. It should be possible to import these console parameters from `installer.conf`, which is available in `$IMAGE_DIR` at this stage of the installation.

platform/aspeed/scripts/sonic-usb-network-init.sh
```diff
@@ -0,0 +1,152 @@
...
+        echo "" > "${GADGET_DIR}/UDC" 2>/dev/null || true
+    fi
+    # Remove symlinks
+    rm -f "${GADGET_DIR}/configs/c.1/${FUNCTION_TYPE}.${INTERFACE_NAME}" 2>/dev/null || true
+    # Remove directories
+    rmdir "${GADGET_DIR}/functions/${FUNCTION_TYPE}.${INTERFACE_NAME}" 2>/dev/null || true
+    rmdir "${GADGET_DIR}/configs/c.1/strings/0x409" 2>/dev/null || true
+    rmdir "${GADGET_DIR}/configs/c.1" 2>/dev/null || true
+    rmdir "${GADGET_DIR}/strings/0x409" 2>/dev/null || true
+    rmdir "${GADGET_DIR}" 2>/dev/null || true
```
oleksandrivantsiv: Is itemized removal really needed here?
Can all of these commands:

```
    # Remove symlinks
    rm -f "${GADGET_DIR}/configs/c.1/${FUNCTION_TYPE}.${INTERFACE_NAME}" 2>/dev/null || true
    # Remove directories
    rmdir "${GADGET_DIR}/functions/${FUNCTION_TYPE}.${INTERFACE_NAME}" 2>/dev/null || true
    rmdir "${GADGET_DIR}/configs/c.1/strings/0x409" 2>/dev/null || true
    rmdir "${GADGET_DIR}/configs/c.1" 2>/dev/null || true
    rmdir "${GADGET_DIR}/strings/0x409" 2>/dev/null || true
    rmdir "${GADGET_DIR}" 2>/dev/null || true
```
be replaced with:
```
rm -rf "${GADGET_DIR}" 
```

platform/aspeed/scripts/sonic-usb-network-init.sh
```diff
@@ -0,0 +1,152 @@
...
+# USB Network Gadget Initialization Script for ASPEED AST2700
+# Sets up USB network device using ConfigFS
+
+set -e
+
+# Configuration
+GADGET_NAME="g1"
+FUNCTION_TYPE="ncm"  # Options: ncm, ecm, eem, rndis
+INTERFACE_NAME="usb0"
+# BMC_IP="192.168.100.1"
```
oleksandrivantsiv: Is this still needed? I assume that once the usb0 interface is created, its state and IP address will be controlled by the SONiC Config DB configuration, which is applied through interfaces-config.service.

platform/aspeed/platform_arm64.conf
oleksandrivantsiv: 1. Is it possible to provide more details on this? sonic-installer should not invoke platform_arm64.conf directly. My assumption that platform_arm64.conf should be part of the image, but I don't see that implementation provided it in this PR uses it in any way.
2. `build-emmc-installer.sh` is not part of this PR.

chander-nexthop: I added the script upon your request via Commit ca44f22d3e4c4530e83e4fc88846761087a59ad6 yesterday. Not sure how you missed it. sonic-installer would invoke the helper functions defined in platform_arm64.conf. Thats how its done for all platforms.

platform/aspeed/scripts/sonic-uboot-env-init.sh
chander-nexthop: Done.

platform/aspeed/scripts/sonic-uboot-env-init.sh
chander-nexthop: Done.

platform/aspeed/platform_arm64.conf
chander-nexthop: As of now, we dont have ONIE, in the eval board with sonic running
root@eval-board-1:/home/admin# fw_setenv --help
fw_setenv (compiled Nov  1 2024)
Usage fw_setenv [OPTION]
 -h, --help                       : print this help
 -c, --config <filename>          : configuration file (by default: /etc/fw_env.config)
 -f, --defenv <filename>          : default environment if no one found (by default: /etc/u-boot-initial-env)
 -m, --namespace <name>           : chose one of sets in the YAML file, default first in YAML
 -V, --version                    : print version and exit
 -s, --script <filename>          : read variables to be set from a script
 
 and after the first install, we use sonic-installer to install new images and here '-f' doesn't work. So I would like to re-visit this after ONIE/SONIE.

platform/aspeed/platform_arm64.conf
chander-nexthop: Suggestions not to hard code has been implemented and now its taken from the installer.conf

platform/aspeed/platform_arm64.conf
chander-nexthop: Done.

platform/aspeed/platform_arm64.conf
chander-nexthop: Done

platform/aspeed/scripts/sonic-usb-network-init.sh
chander-nexthop: Removed.

platform/aspeed/scripts/sonic-usb-network-init.sh
chander-nexthop: Done.

chander-nexthop:
> Hello Chinmoy – we are also building and using the AST2700 eval board to load SONIC build. We would like to leverage your commit but there are several files that are using the FIT implementation rather than a UEFI implementation. We would like the scripts to support a UEFI installation. Are you planning to add UEFI support?
> 
> @lguohan @yxieca

We aren't planning UEFI support as of now, aka immediately, but we should discuss your work in this in our work group meetings. Is there any PR?

dockers/docker-database/docker-database-init.sh
```diff
@@ -41,6 +41,11 @@ then
     fi
 fi
 
+if [[ $IS_BMC_DEVICE == "true" ]]
+then
+    export DATABASE_TYPE="bmcdb"
```
judyjoseph, line 46: We really don't have a use case to have a separate bmcdb ?

files/build_templates/docker_image_ctl.j2
```diff
@@ -864,6 +867,10 @@ if [[ "$DEV" == *"dpu"* ]]; then
     DATABASE_TYPE="dpudb"
 fi
 
+if [[ "$IS_BMC_DEVICE" == "true" ]]; then
+    DATABASE_TYPE="bmcdb"
```
judyjoseph, line 871: Seeing multiple definitions of bmcdb .. is this the DB for BMC in the Switch_host ? if so I think we can use the default database STATE_DB itself for storing BMC state

device/aspeed/arm64-aspeed_ast2700_evb-r0/asic.conf
```diff
@@ -0,0 +1,2 @@
+# aspeed-ast2700 doesn't have a switch asic but SONIC assumes NUM_ASIC to be atleast 1
```
judyjoseph, line 1: Is the error there even if this file "asic.conf" is not present ? or 
Can we fix the place where it imposes the NUM_ASIC = 1 check ?

device/nexthop/arm64-nexthop_b27-r0/platform.json
```diff
@@ -0,0 +1,23 @@
+{
+    "chassis": {
+        "name": "NH-B27-BMC",
+        "thermal_manager": false,
+        "status_led": {
+            "controllable": true,
+            "colors": ["green", "amber", "off"]
```
judyjoseph, line 7: In the real physical device with Switch-Host + BMC, is this led visible outside to user/admin ? can we bind the  led color to the BMC status?

device/nexthop/arm64-nexthop_b27-r0/platform_env.conf
```diff
@@ -0,0 +1,2 @@
+# BMC platform environment
+export BMC_PLATFORM=1
```
judyjoseph: don't need "export"

device/aspeed/arm64-aspeed_ast2700_evb-r0/platform_env.conf
```diff
@@ -0,0 +1,2 @@
+# BMC platform environment
+export BMC_PLATFORM=1
```
judyjoseph: don't need "export"

device/aspeed/arm64-aspeed_ast2700_evb-r0/pmon_daemon_control.json
```diff
@@ -0,0 +1,17 @@
+{
+    "skip_thermalctld": false,
+    "skip_ledd": true,
+    "skip_xcvrd": true,
+    "skip_psud": true,
+    "skip_syseepromd": true,
```
judyjoseph, line 6: We could have syseepromd NOT skipped in BMC to get EEPROM data and store in DB ?

files/build_scripts/aspeed_bmc_filter_services.py
```diff
@@ -0,0 +1,259 @@
...
+    'featured.timer',                 # Timer for featured service
+
+    # Security and monitoring (important for BMC)
+    'caclmgrd.service',               # Firewall for management interfaces
+    'haveged.service',                # Entropy for SSH/SSL/TLS
+    'monit.service',                  # Auto-restart failed services
+    'procdockerstatsd.service',       # Container resource monitoring
+
+    # Time synchronization
+    'ntp.service',                    # NTP time synchronization
```
judyjoseph: Can you check if NTP is replaced with chrony in sonic master

files/build_scripts/aspeed_bmc_filter_services.py
```diff
@@ -0,0 +1,259 @@
...
+    'topology.service',
+
+    # NOS-specific services not needed for BMC
+    'copp-config.service',           # Control Plane Policing - no ASIC on BMC
+    'dhcp_dos_logger.service',       # DHCP DoS detection - no switch ports on BMC
+    'warmboot-finalizer.service',    # Warm boot reconciliation - no SWSS/BGP on BMC
+
+    # SONiC NOS services - not needed for BMC
+    'sonic-hostservice.service',     # SONiC host services (VLAN, LAG, etc.) - NOS-specific
+    'hostcfgd.service',              # SONiC host config daemon - NOS-specific
```
judyjoseph: We would need hostcfgd service .. this is one which handles AAA/PAM

files/build_scripts/aspeed_bmc_filter_services.py
```diff
@@ -0,0 +1,259 @@
...
+"""
+
+import sys
+import os
+import glob
+
+# Services to REMOVE for BMC platform (not needed)
+BMC_EXCLUDED_SERVICES = {
+    # AAA/TACACS - not needed on BMC
+    'tacacs-config.service',
```
judyjoseph: we need AAA/TACACS in BMC - as we can login to BMC directly

files/build_scripts/aspeed_bmc_filter_services.py
```diff
@@ -0,0 +1,259 @@
...
+
+    # UUID daemon - not needed
+    'uuidd.service',                 # UUID daemon - kernel provides UUID generation
+    'uuidd.socket',                  # UUID daemon socket activation
+
+    # Console - not needed if no local display
+    'getty@tty1.service',            # Local console - only if VGA/HDMI available
+
+    # Optional services - can be excluded for minimal BMC
+    'auditd.service',                # Security audit - only needed for compliance
```
judyjoseph: auditd also would be needed @qiluo-msft  could you confirm if we can skip this service

files/build_scripts/aspeed_bmc_filter_services.py
```diff
@@ -0,0 +1,259 @@
...
+    'ztp.service',                    # ZTP for automated device provisioning
+
+    # System health monitoring
+    'system-health.service',          # SONiC system health monitor
+
+    # Platform monitoring
+    'pmon.service',                   # Platform monitor container (sensors, fans, PSU, etc.)
+
+    # Telemetry and monitoring
+    'gnmi.service',                   # gNMI telemetry service
```
judyjoseph, line 74: Can we include sysmgr.service, this is needed alone with gnmi for remote reboot backend implementation.

files/build_scripts/aspeed_bmc_filter_services.py
chinmoy-nexthop: We will look into this and share an update, @judyjoseph . Kindly let us know if addressing this in the next PR works for you.

files/build_scripts/aspeed_bmc_filter_services.py
chinmoy-nexthop: Sure @judyjoseph , will take care.

files/build_scripts/aspeed_bmc_filter_services.py
chinmoy-nexthop: Sure @judyjoseph , we will take care.

files/build_scripts/aspeed_bmc_filter_services.py
chinmoy-nexthop: Sure, @judyjoseph thank you for sharing your comments here . Shall we plan to address the systemd-related feedback in the next PR?

dockers/docker-database/docker-database-init.sh
chander-nexthop: The DATABASE_TYPE bmcdb was added only to avoid BMP_STATE_DB (BGP monitoring protocol database) which doesn't make sense in our platform. Otherwise it behaves in an identical manner to DATABASE_TYPE being empty/not specified. So all the standard databases of SONIC (CONFIG_DB, APPL_DB, STATE_DB etc) would be present. If we don't have this new type, then the condition to avoid BMP_STATE_DB will have to be coded differently. I went with this approach because it was already used by one platform (dpudb).

files/build_templates/docker_image_ctl.j2
chander-nexthop: No, it is to avoid creating BMP_STATE_DB (we dont run routing container here) in case of BMC SONIC.

device/aspeed/arm64-aspeed_ast2700_evb-r0/asic.conf
chander-nexthop: NUM_ASIC being not zero seems to be a fundamental SONIC assumption. Even virtual switch implementations (device/virtual) define it to be 1. So went with this.

device/nexthop/arm64-nexthop_b27-r0/platform.json
chander-nexthop: This is placeholder code and needs to be revisited when the actual hardware platform becomes ready

device/nexthop/arm64-nexthop_b27-r0/platform_env.conf
chander-nexthop: Will remove.

device/aspeed/arm64-aspeed_ast2700_evb-r0/platform_env.conf
chander-nexthop: Will remove.

device/aspeed/arm64-aspeed_ast2700_evb-r0/pmon_daemon_control.json
chander-nexthop: This is placeholder code and needs a revisit when the hardware is fully ready

files/build_scripts/aspeed_bmc_filter_services.py
chander-nexthop: Will do and take up enabling of all the required systemd services (and docker containers) as a follow up PR. Same for all the comments that follow. Hope thats fine?

platform/aspeed/sonic-platform-modules-nexthop/common/scripts/usb-network-init.sh
```diff
@@ -0,0 +1,145 @@
+#!/bin/bash
+# USB Network Gadget Initialization Script for ASPEED AST2700
+# Sets up USB network device using ConfigFS
+
+set -e
+
+# Configuration
+GADGET_NAME="g1"
```
oleksandrivantsiv, line 8: At the moment, I’m not sure whether this script and the service that starts it on system boot can be fully vendor‑independent. It’s possible that initialization of the usb0 network interface varies by vendor, in which case this logic should be moved into the vendor‑specific folder.

platform/aspeed/one-image.mk
```diff
@@ -0,0 +1,23 @@
+# ONIE installer for Aspeed BMC platform with U-Boot
+
+SONIC_ONE_IMAGE = sonic-aspeed-arm64.bin
+$(SONIC_ONE_IMAGE)_ARCH = arm64
+$(SONIC_ONE_IMAGE)_MACHINE = aspeed
+$(SONIC_ONE_IMAGE)_PLATFORM = aspeed
+$(SONIC_ONE_IMAGE)_IMAGE_TYPE = onie
```
oleksandrivantsiv, line 7: Can you confirm whether the generated image is compatible with the sonic-installer that runs on platforms using U-Boot?
Reference: https://github.com/sonic-net/sonic-utilities/blob/master/sonic_installer/bootloader/uboot.py

platform/aspeed/one-image.mk
```diff
@@ -0,0 +1,23 @@
...
+                          $(DOCKER_MUX) $(DOCKER_MACSEC) $(DOCKER_TELEMETRY) \
+                          $(DOCKER_EVENTD) $(DOCKER_DASH_HA) $(DOCKER_STP)
+
+$(info [aspeed] Filtering out packages: $(DISABLED_PACKAGES_LOCAL))
+SONIC_PACKAGES_LOCAL := $(filter-out $(DISABLED_PACKAGES_LOCAL), $(SONIC_PACKAGES_LOCAL))
+
+$(SONIC_ONE_IMAGE)_INSTALLS += $(SYSTEMD_SONIC_GENERATOR)
+$(SONIC_ONE_IMAGE)_LAZY_INSTALLS += $(ASPEED_EVB_AST2700_PLATFORM_MODULE)
+$(SONIC_ONE_IMAGE)_LAZY_INSTALLS += $(ASPEED_NEXTHOP_B27_PLATFORM_MODULE)
+# $(SONIC_ONE_IMAGE)_DOCKERS = $(DOCKER_DATABASE) $(DOCKER_GNMI) $(DOCKER_RESTAPI) $(DOCKER_OBMC_CONSOLE) $(DOCKER_PLATFORM_MONITOR) $(DOCKER_BMCWEB)
```
oleksandrivantsiv: looks like this line is not needed

platform/aspeed/one-image.mk
chander-nexthop: Yes, we use uboot in our hardware and the eval board. Port first burn to eMMC, we use sonic-installer install to install subsequent images.

platform/aspeed/one-image.mk
chander-nexthop: We use it for our internal dev. Will be upstreamed when ready.

platform/aspeed/sonic-platform-modules-nexthop/common/scripts/usb-network-init.sh
chander-nexthop: Can you add more details on what all will be vendor specific? I would prefer to have this in common services, till more details emerge

platform/aspeed/scripts/sonic-uboot-env-init.sh
oleksandrivantsiv: I didn't get this answer. Is it possible to remove code duplication here and in `platform/aspeed/platform_arm64.conf`?

platform/aspeed/sonic-platform-modules-nexthop/common/scripts/usb-network-init.sh
oleksandrivantsiv: I mean that at this moment I'm not sure whether the creation and initialization of the usb0 device will be the same for NVIDIA as it is for Nexhop. For now, let's keep it as is, but we may need to revisit this implementation in the future.

platform/aspeed/scripts/sonic-usb-network-init.sh
```diff
@@ -0,0 +1,145 @@
...
+# USB Network Gadget Initialization Script for ASPEED AST2700
+# Sets up USB network device using ConfigFS
+
+set -e
+
+# Configuration
+GADGET_NAME="g1"
+FUNCTION_TYPE="ncm"  # Options: ncm, ecm, eem, rndis
+INTERFACE_NAME="usb0"
+SUBNET_MASK="24"
```
oleksandrivantsiv: SUBNET_MASK variable is not used

platform/aspeed/scripts/sonic-usb-network-init.sh
```diff
@@ -0,0 +1,145 @@
...
+MANUFACTURER="ASPEED BMC"
+PRODUCT_NAME="USB Network Device"
+DEV_MAC="02:00:00:00:00:01"
+HOST_MAC="02:00:00:00:00:02"
+
+logger -t usb-network "Starting USB network gadget initialization..."
+
+# Step 1: Load kernel modules
+logger -t usb-network "Loading kernel modules..."
+modprobe libcomposite 2>/dev/null || true
```
oleksandrivantsiv: Is there any particular reason why the drivers are loaded from the script rather than by modprobe at system start?

platform/aspeed/scripts/sonic-usb-network-init.sh
```diff
@@ -0,0 +1,145 @@
...
+modprobe usb_f_${FUNCTION_TYPE} 2>/dev/null || true
+modprobe u_ether 2>/dev/null || true
+modprobe aspeed_vhub 2>/dev/null || true
```
oleksandrivantsiv: Please do not redirect errors during driver loading to `/dev/null`. If the driver is critical and required for the USB interface, any error that occurs during loading should be logged along with the return code; otherwise, debugging will be impossible.

If we can safely ignore a failure to load the driver, then is the driver needed at all?

platform/aspeed/scripts/sonic-usb-network-init.sh
chander-nexthop: Will fix. Thanks for pointing out.

platform/aspeed/scripts/sonic-usb-network-init.sh
chander-nexthop: Will fix.

platform/aspeed/sonic-platform-modules-nexthop/common/scripts/usb-network-init.sh
chander-nexthop: Of course! Refactoring should be done when we have more vendors adding support for their platforms.

platform/aspeed/scripts/sonic-usb-network-init.sh
chander-nexthop: These modules pertain to USB gadget configuration for aspeed. So has to be aspeed specific. I will investigate if there is a way to do this in a platform specific way, without having to rely on the platform specific debian package.

platform/aspeed/scripts/sonic-uboot-env-init.sh
chander-nexthop: Part 2 of the uboot env init is a stop gap and hence I thought replication is fine. Will investigate more to see if the replication can be avoided. But does that gate a merge?

platform/aspeed/scripts/sonic-usb-network-init.sh
chander-nexthop: Moved the USB network initialization to the Nexthop vendor platform module

platform/aspeed/sonic-platform-modules-nexthop/common/scripts/usb-network-init.sh
chander-nexthop: Moved the USB creation to Nexthop platform module.

files/build_scripts/aspeed_bmc_filter_services.py
chander-nexthop: Have added all the containers requested for. @judyjoseph please have a look.

files/build_scripts/aspeed_bmc_filter_services.py
chander-nexthop: Done.

files/build_scripts/aspeed_bmc_filter_services.py
chander-nexthop: Done.

device/nexthop/arm64-nexthop_b27-r0/platform_env.conf
chander-nexthop: Done

device/aspeed/arm64-aspeed_ast2700_evb-r0/platform_env.conf
chander-nexthop: Done

chander-nexthop:
@Blueve and @apoorv-arista Please have a look at commit 40f4febf70f3217b844712c7fcea81377fc49fe6 that just got pushed to this PR.

device/aspeed/arm64-aspeed_ast2700_evb-r0/platform.json
```diff
@@ -0,0 +1,23 @@
...
+    "chassis": {
+        "name": "AST2700-EVB-BMC",
+        "thermal_manager": false,
+        "status_led": {
+            "controllable": true,
+            "colors": ["green", "amber", "off"]
+        },
+        "components": [
+            {
+                "name": "BMC"
```
judyjoseph, line 11: Please remove BMC as component. not sure why we need this . Same applies to other platform/sku

device/aspeed/arm64-aspeed_ast2700_evb-r0/platform_components.json
```diff
@@ -0,0 +1,10 @@
+{
+    "chassis": {
+        "AST2700-EVB-BMC": {
+            "component": {
+                "BMC": {},
```
judyjoseph, line 5: BMC as a component here too

device/aspeed/arm64-aspeed_ast2700_evb-r0/platform.json
chander-nexthop: Done via https://github.com/sonic-net/sonic-buildimage/pull/26002

device/aspeed/arm64-aspeed_ast2700_evb-r0/platform_components.json
chander-nexthop: Done via https://github.com/sonic-net/sonic-buildimage/pull/26002

platform/aspeed/sonic-platform-modules-nexthop/common/systemd/usb-network-init.service
```diff
@@ -0,0 +1,12 @@
...
+Description=NextHop USB Network Gadget Service
```
oleksandrivantsiv, line 2: @chinmoy-nexthop, @judyjoseph, please make this service vendor‑specific and move it into a vendor folder. The script it calls has a vendor‑specific implementation and cannot run on our system.

chander-nexthop: Please have a look at the outstanding PR https://github.com/sonic-net/sonic-buildimage/pull/26002/
This PR has done this. You will get it once it merges.

chinmoy-nexthop: Thank you @chander-nexthop

PR #24903, opened 2025-12-18
Add 400g/300m cable length to pg_profile_lookup for Arista-7060X6-64PE-B-P32O64

dakotac-arista:
#### Why I did it
Requested for new F2 topo
##### Work item tracking

#### How I did it

Used same XOFF as 400g/500m 

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [x] 202511
- [x] msft-202503

msft-202503

- [ ] 
- [ ] 

:)

### DIFF
--- a/device/arista/x86_64-arista_7060x6_64pe_b/Arista-7060X6-64PE-B-P32O64/pg_profile_lookup.ini
+++ b/device/arista/x86_64-arista_7060x6_64pe_b/Arista-7060X6-64PE-B-P32O64/pg_profile_lookup.ini
@@ -1,6 +1,7 @@
 # PG lossless profiles.
 # speed cable size    xon  xoff threshold xon_offset
 400000  30m  8636    0    425958  0         3556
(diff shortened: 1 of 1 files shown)


sdszhang (commented):
can you add it for 800g as well? Thanks!

dakotac-arista:
> can you add it for 800g as well? Thanks!

That will require a CSP case for a new bcm config. The values can be reused for a lower speed or shorter cable since it will just waste some memory but we can't increase them. There is only 800g/5m for this hwsku.

sdszhang:
> > can you add it for 800g as well? Thanks!
> 
> That will require a CSP case for a new bcm config. The values can be reused for a lower speed or shorter cable since it will just waste some memory but we can't increase them. There is only 800g/5m for this hwsku.

Thanks. Let me open an Arista issue to track it. Kumaresh mentioned that it can be done using generic buffer config.

sdszhang (approved):

PR #24941, opened 2025-12-23
Unify chassis db address across various platforms chassis/disagg-chassis/smartswitch

judyjoseph:
#### Why I did it
•	To have the Redis CHASSIS_DB instantiated over an IP/subnet, unique across sonic chassis/smart switch platforms. 
•	Keep it independent of the midplane IP/subnet which is defined per platform in vendor files.

##### Work item tracking
- Microsoft ADO **(number only)**: 35696466

#### How I did it
Changes planned with unify approach
1.	Define the chassis DB IP address template file. 
    •	Use the docker IP subnet in case of single ASIC/multi-ASIC 1RU based T2 disaggregated chassis
    •	Else in case of T2 chassis with supervisor and Smart switch use the 10.x.x.x subnet for chassis DB 

     This j2 file is used in the system service “config-chassisdb” to generate the file “/etc/sonic/chassisdb_address”. 

2.	Create a new interface eth-chassisdb in Supervisor in case of T2 chassis and NPU in case of smart switch
    •	This interface is of type dummy, <ip link add eth-chassisdb type dummy>
    •	IP address/subnet is from file: /etc/sonic/chassisdb_address

3.	Create routes in the Linecard (for T2 chassis) and in DPU (for smart switch) to reach the new chassisdb subnet via the midplane IP
    •	The routes are added in the host namespace and asic namespaces (in case of multi asic linecard/dpu) 
         in the format : <ip route add $chassis_db_subnet via $supervisor_npu_midplane_ip>
    •	chassis_db_subnet is the subnet of new “eth-chassisdb” interface in Supervisor/NPU
    •	supervisor_npu_midplane_ip is the midplane IP configured in Supervisor/NPU

4.	During the chassis/smartswitch image upgrade, the supervisor/NPU respectively is upgraded first, followed by the Linecard/DPU. So, there is a state where the Linecard/DPU is running the old code and trying to reach the chassis_db in Supervisor/NPU which is now in new chassis-db subnet ( and not the midplane subnet as earlier) 
    •	To fix this, a DNAT/SNAT entry is added to do a NAT from midplane_ip:6380  chassisdb_address:6380
    •	where 6380 is the port where database-chassis redis server is listening.
    •	This DNAT/DNAT entry will not be cleaned even after the upgrade of Linecard/DPU.
    PR: https://github.com/sonic-net/sonic-host-services/pull/335

#### How to verify it
The following tests were done to validate this feature.
   1.	Bootup chassis with SUP and LC both upgraded at same time    - Done
   2.	Bootup SUP first and make sure it works with Linecard running current 202405 gold version - Done
        a.	Upgrade the LC to this image and check chassis DB connectivity is up - Done
   3.	SUP, LC independent reboots - Done
   4.	Test Bootup and Upgrade sequence with UT2 - Done
   5.	Test Bootup and Upgrade sequence with Smart switch - TODO   
   6.	Test with T2 virtual switch - TODO

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511

- [ ] 
- [ ]


judyjoseph:
Taking a different approach and NOT to introduce a new IP address as done here_( files/image_config/config-chassisdb/chassisdb_address.j2 )_. The idea is to reuse the supervisor IP address derived by vendor/platform code during boot up and creating a common "/etc/sonic/chassisdb_address" file.

PR #24964, opened 2026-01-02
[SmartSwitch] Mask dash-ha service by default

croos12:
#### Why I did it

The dash-ha service is currently always set to be started in SONiC because its service file is copied in sonic_debian_extension.j2. This leads to an issue as when the dash-ha service runs without the dash-ha dpu containers being created, which can happen when starting up smartswitch:

```ERR container: docker get image version for dash-hadpu3 failed with 404 Client Error for http+docker://localhost/v1.50/containers/dash-hadpu3/json: Not Found ("No such container: dash-hadpu3")```

This change sets the dash-ha feature to be disabled by default following the build flag. It will not be enabled on startup.

#### How I did it
I explicitly set to service to be disabled by default
 
#### How to verify it

```root@sonic-npu:/var/log# systemctl status dash-ha
○ dash-ha.service
     Loaded: masked (Reason: Unit dash-ha.service is masked.)
     Active: inactive (dead)

root@sonic-dpu:/var/log# systemctl status dash-ha
○ dash-ha.service
     Loaded: masked (Reason: Unit dash-ha.service is masked.)
     Active: inactive (dead)
```

- [x] 202511_RC


volodymyrsamotiy:
@prabhataravind , could you please help to review?

tirupatihemanth:
Fixes https://github.com/sonic-net/sonic-buildimage/issues/24543

KrisNey-MSFT:
hi @croos12  - should we include Jing Zhang for the HA work?  TY!

prabhataravind:
@zjswhhh please review and sign-off as well.

qiluo-msft:
@croos12 Is it a more concerning issue that "service file is copied in sonic_debian_extension.j2. This leads to an issue as when the dash-ha feature is not explicitly enabled/configured it is not built"? Should we fix it?

croos12:
> @croos12 Is it a more concerning issue that "service file is copied in sonic_debian_extension.j2. This leads to an issue as when the dash-ha feature is not explicitly enabled/configured it is not built"? Should we fix it?

No this should not be concerning. The feature was just not masked correctly when not in use and was explicitly written to run on SONiC startup. This description was confusing and I apologize. In the existing behavior the service has a "condition unmet" if it is not running on an NPU or DPU and should exit without error. The issue here is that while this is running on NPUs before the dash-ha containers on the DPUs have been established. The NPU service will crash and create nuisance error messages.

dgsudharsan:
This doesn't handle the namespace based services and requires an elaborate fix. Request @zjswhhh and @BYGX-wcr  to handle the fix

PR #24989, opened 2026-01-07
[202405] fix upgrateVersion pipeline file

yijingyan2:
#### Why I did it
snapshot versions cannot be correctly fetched, the url for snapshot has changed
https://storage.googleapis.com/golang/go has been deprecated
snmpd version conflict

##### Work item tracking
- Microsoft ADO **(number only)**: 36316816


liushilongbuaa (approved):
LGTM.
Please add ADO to triage snmpd issue. We need find a solution to fix the conflict in future.

PR #25019, opened 2026-01-09
Add metric field to vnet route tunnel yang

Leyza:
#### Why I did it
Add metric field to vnet route tunnel yang models. Metric value can be used to help determine the type of route (e.x. determine if route is static, override, udr, etc.). This field will not affect the route's behavior/has no backend logic attached to it.

Related swss change: https://github.com/sonic-net/sonic-swss/pull/4119

##### Work item tracking

#### How to verify it
Tested manually on testbeds and in UTs.
image

202506 and 202511

- [ ] 
- [ ]


src/sonic-yang-models/yang-models/sonic-vnet.yang
```diff
@@ -143,6 +143,14 @@ module sonic-vnet {
                     description "Number of consistent hashing buckets to use, if consistent hashing is desired";
                     type uint16;
                 }
+
+                leaf metric {
+                    description "Value set to determine type of route";
+                    type uint8 {
+                        range "0..20";
```
anish-n: Could we just allow this to be a unit and don't limit to 0 to 20

src/sonic-yang-models/yang-models/sonic-vnet.yang
```diff
@@ -143,6 +143,14 @@ module sonic-vnet {
...
+                    }
+                    default 0;
```
anish-n: This is an optional field so I think there should be no default?

anish-n (approved):

anish-n (changes requested):
Suggesting changes to make..

src/sonic-yang-models/yang-models/sonic-vnet.yang
Leyza: Removed default

src/sonic-yang-models/yang-models/sonic-vnet.yang
Leyza: Removed range

src/sonic-yang-models/yang-models/sonic-vnet.yang
```diff
@@ -143,6 +143,14 @@ module sonic-vnet {
                     description "Number of consistent hashing buckets to use, if consistent hashing is desired";
                     type uint16;
                 }
+
+                leaf metric {
+                    description "Value set to determine type of route";
```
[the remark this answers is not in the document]
hdwhdw: This is valid. Can you address?

src/sonic-yang-models/yang-models/sonic-vnet.yang
```diff
@@ -143,6 +143,11 @@ module sonic-vnet {
...
```
hdwhdw: Agree with copilot. "Value set to determine type of route" Can be too ambiguous to people not familiar with the context. At least we should have some examples.

Leyza: Updated the description with more info and specified field has no backend behavior.

hdwhdw (approved):

qiluo-msft (dismissed):

qiluo-msft (approved):

anish-n (approved):

prabhataravind:
@yijingyan2 could you please raise a PR for 202511 as there is a conflict? @vmittal-msft  @prsunny  for viz.

yijingyan2:
@Leyza could you please raise a PR for 202511? Thanks

PR #25037, opened 2026-01-13
Reapply "We should check service enabled instead exists (#24754)" (#24794)

qiluo-msft:
#### Why I did it
If teamd feature is disabled,
```
sudo config feature state teamd disabled
```

swss service should not wait on team container at all. Currently there is a process
```
python3 /usr/bin/docker-wait-any -s swss -d syncd teamd
```

However, a thread inside /usr/bin/docker-wait-any is actually crashed, and the behavior is same as ignoring teamd. It still keep the whole system working.

The root fix should check service enabled instead exists, so it will not wait-any on teamd at all. After this PR,

```
python3 /usr/bin/docker-wait-any -s swss -d syncd
```

##### Work item tracking
- Fixing https://github.com/sonic-net/sonic-buildimage/issues/24730

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

- [ ] 
- [ ]

### DIFF
--- a/files/scripts/swss.sh
+++ b/files/scripts/swss.sh
@@ -619,7 +619,7 @@ check_add_bgp_dependency
 check_ports_present
 PORTS_PRESENT=$?
 
-if [[ $PORTS_PRESENT == 0 ]] && [[ $(check_service_enabled "teamd$DEV") == "true" ]]; then
+if [[ $PORTS_PRESENT == 0 ]] && [[ $(check_service_enabled "teamd${DEV:+@$DEV}") == "true" ]]; then
     MULTI_INST_DEPENDENT="teamd"
 fi
 


files/scripts/swss.sh
```diff
@@ -622,7 +619,7 @@ check_add_bgp_dependency
 check_ports_present
 PORTS_PRESENT=$?
 
-if [[ $PORTS_PRESENT == 0 ]] && [[ $(check_service_exists teamd) == "true" ]]; then
+if [[ $PORTS_PRESENT == 0 ]] && [[ $(check_service_enabled teamd) == "true" ]]; then
```
saiarcot895: `teamd` is a per-ASIC container, meaning in a multi-ASIC setup, it will need to check `teamd@0`, `teamd@1`, etc.

saiarcot895 (approved):

PR #25046, opened 2026-01-13
[build] Support override public reproducible build version files

yijingyan2:
#### Why I did it
Public build uses files/build/versions folder to store reproducible build version files. If someone wants to do a small change, it will result code conflict in future.
Now we use a symbol link instead of real folder to provide an easy way to override these files.
##### Work item tracking
- Microsoft ADO **(number only)**: 36315930

#### How I did it
by creating a symbolic link


liushilongbuaa (approved):

yejianquan (approved):
LGTM

PR #25061, opened 2026-01-13
[orchagent]: Enable countersyncd as service in orchagent

Pterosaur:
#### Why I did it
`countersyncd` should be a daemon process that running in the swss container.

##### Work item tracking

#### How I did it
Add a service of countersyncd in the supervisor config file

#### How to verify it
Check it locally.

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511

- [ ] 
- [ ]


dockers/docker-orchagent/supervisord.conf.j2
```diff
@@ -340,3 +340,15 @@ dependent_startup_wait_for=swssconfig:exited
...
+command=/usr/bin/countersyncd --enable-otel
+priority=18
+autostart=false
+autorestart=false
+stdout_logfile=NONE
+stdout_syslog=true
+stderr_logfile=NONE
+stderr_syslog=true
+dependent_startup=true
+dependent_startup_wait_for=swssconfig:exited
```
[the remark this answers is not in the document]
Pterosaur: countersyncd is a rust program, so that's not same.

DavidZagury:
@Pterosaur @r12f 
This change will make countersyncd run regardless to whether or not HFT is even configured.
This will waste resources when HFT is disabled.

Also the way that it is set, countersyncd will always run with only enable-otel flag. The service countersyncd also provides flags such as otel-endpoint or otel-capacity flags - which are not configurable by this implementation.

I suggested to keep it as a service that will only start in response to a CLI command by the user or a DB configuration

Pterosaur:
> @Pterosaur @r12f This change will make countersyncd run regardless to whether or not HFT is even configured. This will waste resources when HFT is disabled.
> 
> Also the way that it is set, countersyncd will always run with only enable-otel flag. The service countersyncd also provides flags such as otel-endpoint or otel-capacity flags - which are not configurable by this implementation.
> 
> I suggested to keep it as a service that will only start in response to a CLI command by the user or a DB configuration

Hi @DavidZagury  Thanks for your suggestion
1. It's impossible to connect the CLI with the countersyncd service. Because we have multiple methods to config HFT. Cli is just one of them. In the most time, we will use GNMI to config it which means the GNMI will directly inject configuration into config db instead of using CLI.
2. In my humble opinion, if there is no data telemetry traffic under the countersyncd which means the HFT feature isn't enabled or configured, the countersyncd will be almost silent. It will only use tens of megabytes of memory and almost zero CPU.
3. To the flag `otel-endpoint`, it's not a standard method to configure the endpoint of otel. The better way is to use the config db that mentioned in: https://github.com/sonic-net/SONiC/blob/master/doc/high-frequency-telemetry/high-frequency-telemetry-hld.md#742-high_frequency_telemetry_profile But we haven't supported that. This flag is exposed right now that is only used for simplifying testing flow.

Pterosaur:
Hi @kperumalbfn , could you please help to merge this PR?

dockers/docker-orchagent/supervisord.conf.j2
```diff
@@ -340,3 +340,15 @@ dependent_startup_wait_for=swssconfig:exited
 environment=ASAN_OPTIONS="log_path=/var/log/asan/fdbsyncd-asan.log{{ asan_extra_options }}"
 {% endif %}
 {%- endif %}
+
+[program:countersyncd]
+command=/usr/bin/countersyncd --enable-otel
+priority=18
+autostart=false
+autorestart=false
```
kperumalbfn, line 348: why autorestart set to false?

Pterosaur: All services in this supervisord are both "autorestart=false". 
According to my understanding, we don't use the default supervisor management mechanism but use the dependent_* plugin of supervisor to manage these processes. So, autostart and autorestart must be set false.

kperumalbfn (approved):

PR #25080, opened 2026-01-15
Remove partition mount for NVME SSDs 

gpunathilell:
#### Why I did it
Since the partitions themselves are not required to be mounted in pmon container, this partition mounts are being removed. This is done so that we do not have failure from test_container_privileged for NVME disks

##### Work item tracking

#### How I did it
Ignore all partitions during pmon container creation (each partition mount will be of the format `/dev/nvme0n1p1`

#### How to verify it
Without change:
```
root@sonic:/home/admin# docker exec -it pmon bash -c 'ls /dev/nvme*'
/dev/nvme0  /dev/nvme0n1  /dev/nvme0n1p1  /dev/nvme0n1p2  /dev/nvme0n1p3  /dev/nvme-fabrics
```
After change:
```
root@sonic:/home/admin# docker exec -it pmon bash -c 'ls /dev/nvme*'
/dev/nvme-fabrics  /dev/nvme0  /dev/nvme0n1
```

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511

- [ ] 
- [ ]


nate-nexthop:
If you still have read-write access to the NVME block device `/dev/nvme0n1`, you still have read-write access to all of the partitions whether or not their individual device files are mounted into the container. This change doesn't seem to make any difference in terms of what the `pmon` container has access to.

To make this change effective, we must update the test so that having the raw block device fail the test in addition to having any partition fail the test. I'm not sure that will be compatible with code in pmon that want to access /dev/sda to check SMART info, for example.

vvolam:
@gpunathilell @dgsudharsan I think this PR is not required any more as a superset is merged by this PR https://github.com/sonic-net/sonic-buildimage/pull/24463/changes. Please review.

gpunathilell:
> @gpunathilell @dgsudharsan I think this PR is not required any more as a superset is merged by this PR https://github.com/sonic-net/sonic-buildimage/pull/24463/changes. Please review.

@vvolam That PR does not solve the issue which causes the test failure. But as @nate-nexthop mentioned we still have read-write access to the block device, which needs to be solved seperately, investigating to see what can be done

liat-grozovik:
@gpunathilell please handle conflicts and re run checkers

gpunathilell:
Closing this PR, as further investigation is needed since the block device (which holds the partition) is still mounted

PR #25143, opened 2026-01-20
[Arista] Improve Removal of __pycache__ Directories

erictang832:
#### Why I did it

The following trace of shell commands was spotted in a build of the Sonic master, in the middle of a sonic-buildimage build:
```
01:21:23  #82 5.951 + find / ```
01:21:23  #82 5.952 + grep -E __pycache__  
01:21:23  #82 5.952 + xargs rm -rf  
01:21:24  #82 6.902 find: '/proc/359/task/359/net': Invalid argument  
01:21:24  #82 6.902 find: '/proc/359/net': Invalid argument
```
There is a more efficient way to do this than listing all files and grepping; for example: `find / -type d -name "__pycache__" -exec rm -rf {}`

#### How I did it

replaced `find / | grep -E "__pycache__" | xargs rm -rf` with `find / -type d -name __pycache__ -exec rm -rf {} + 2>/dev/null || true`. which improves efficiency by avoiding the overhead of piping into grep and ensuring the operation finishes silently even if permission errors occur.

#### How to verify it

We verified that `__pycache__` is successfully removed after the cleanup.

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511

- [ ] 
- [ ]

### DIFF
--- a/src/sonic-build-hooks/scripts/post_run_cleanup
+++ b/src/sonic-build-hooks/scripts/post_run_cleanup
@@ -40,7 +40,7 @@ rm -f $VERSION_DEB_PREFERENCE
 rm -f /etc/apt/preferences.d/01-versions-deb
 
 # Remove all doc files
-find /usr/share/doc -depth \( -type f -o -type l \) ! -name copyright | xargs rm || true
+find /usr/share/doc -depth \( -type f -o -type l \) ! -name copyright -exec rm {} + 2>/dev/null || true
 
 # Clean doc directories that are empty or only contain empty directories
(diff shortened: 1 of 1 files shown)


erictang832:
> Thanks for fixing this!
> 
> While you're in this file, it looks like line 43 could be updated with a similar change, changing | xargs rm to -exec rm like you did here.

Added fix!

nate-nexthop:
> > Thanks for fixing this!
> > While you're in this file, it looks like line 43 could be updated with a similar change, changing | xargs rm to -exec rm like you did here.
> 
> Added fix!

Thanks!

PR #25144, opened 2026-01-20
Dual timer config for mux icmp probe sessions

manamand2020:
#### Why I did it
Adding dual timer configuration for MUX ICMP prober session
https://github.com/manamand2020/SONiC/blob/88c5bf005ba2118fbbcb16ab8e79c8eebb434d45/doc/dualtor/ICMP_Hardware_Offload_and_Protecion.md#config-db

##### Work item tracking

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511

- [ ] 
- [ ]


src/sonic-yang-models/doc/Configuration.md
```diff
@@ -1776,6 +1776,8 @@ The **MUX_CABLE** table is used for dualtor interface configuration. The `cable_
             "server_ipv4": "192.168.0.2/32",
             "server_ipv6": "fc02:1000::30/128",
             "soc_ipv4": "192.168.0.3/32",
+            "self_interval": "3",
```
zjswhhh, line 1779: Should we use "tx" and "rx" instead of "self" and "peer"?

manamand2020: No, because tx and rx are intervals of ICMP probe sessions. Here we wanted to be more specific to indicate the self and peer sessions. Example:
self_interval : 3 --> would mean a normal icmp session with tx interval = 3*positive signal, rx_interval = 3 * negative signal count.
peer_interval : 100 --> would mean a RX only icmp sessions with tx_interval = 0, rx_interval = 100 * negative signal count.

PR #25151, opened 2026-01-21
[Nvidia] [Smartswitch] Added check for file while checking for reboot cause

gpunathilell:
#### Why I did it
Confirm that the PCI tree exists before checking for the watchdog reboot cause, as this reboot cause is dependent on the reading register and the pci device should be available for this to be done, otherwise there are error lgos

##### Work item tracking

#### How I did it
check for the path if it exists using os.path.exists function

#### How to verify it

`get_reboot_cause` execution while the DPU is powered off, we should not get the following error log:
`ERR pmon#chassisd[38]: Failed to check watchdog reason for DPU0! Command '['mlxreg', '-d', '0000:08:00.0', '--reg_name', 'MRSI', '-g', '-indexes', 'device=1']' returned non-zero exit status 1.`

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511

- [ ] 
- [ ]


dgsudharsan (approved):

dgsudharsan:
@vmittal-msft Can you please check why this has not been picked for 202511 branch even though it is approved?

PR #25169, opened 2026-01-23
[ARP_update]: Increase ARP update frequency and add retry for ping and arping

Janetxxx:
#### Why I did it
This PR increases the ARP update frequency by reducing the main loop sleep interval from 300 to 120 seconds, and adds retry logic with 1 attempt (if we do need retry functionality in the future it will be easy to just change this value) and 3-second intervals between retries for ping and arping operations to improve reliability in handling transient network issues.
- For ping – check the result code with ping response (0 for succussed, non-0 for failed) and without ping response, log the result if failed and it will be helpful for debugging other traffic issues.
- For arping - check the result code with arping response and without arping response (0 for succussed, non-0 for failed), log the result if failed (Use arping for ipv4 and use ndisc6 for ipv6).
- Reworked VLAN GARP (arping) to execute per neighbor with retries.

##### Work item tracking

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [x] 202411
- [ ] 202505
- [x] 202511

- [ ] 
- [ ]


files/scripts/arp_update
```diff
@@ -122,10 +147,11 @@ while /bin/true; do
       #   arping -q -w 0 -c 1 -i <VLAN interface> <IP 1>;
       #   arping -q -w 0 -c 1 -i <VLAN interface> <IP 2>;
       #   ...
-      arpingcmd="sed -e 's/ / -i /' -e 's/^/arping -q -w 0 -c 1 /' -e 's/$/;/'"
-      ipcmd="ip -4 neigh show | grep $vlan | cut -d ' ' -f 1,3 | $arpingcmd"
-
-      eval `eval $ipcmd`
+      while read -r ip ifname; do
+          [ -z "$ip" ] && continue
+          arpingcmd="arping -q -w 0 -c 1 -i $ifname $ip >/dev/null"
```
BYGX-wcr: Have we checked the following two things?
1. Does arping work for IPv6 addresses? I think we need to use ndisc for IPv6
2. Does arping/ndisc return non-zero upon failing to get reply?

BYGX-wcr (commented):
LGTM

kperumalbfn:
@theasianpianist could you please review this

r12f:
The failure is ignorable:

image

BYGX-wcr (approved):
LGTM

files/scripts/arp_update
```diff
@@ -14,6 +14,48 @@ logger () {
     command logger -i "$$" -t "arp_update" "$@"
 }
 
+# Function to run a command with retries on failure (3 seconds interval, 10 tries)
+run_with_retry_cmd() {
+    local desc="$1"; shift
+    local max_tries=10
+    local interval=3
```
theasianpianist: How were these values determined?

files/scripts/arp_update
```diff
@@ -162,9 +206,10 @@ while /bin/true; do
         # generates the following command for each FAILED or INCOMPLETE IPv6 neighbor
         # timeout 0.2 ping <neighbor IPv6> -n -q -i 0 -c 1 -W 1 -I <VLAN name> >/dev/null
         if [[ ! -z "$unresolved_kernel_neighbors" ]]; then
-            ping6_template="sed -e 's/^/timeout 0.2 ping /' -e 's/,/ -n -q -i 0 -c 1 -W 1 -I /' -e 's/$/ >\/dev\/null;/'"
-            failed_ip6_neigh_cmd="echo \"$unresolved_kernel_neighbors\" | cut -d ' ' -f 1,3 --output-delimiter=',' | $ping6_template"
-            eval `eval "$failed_ip6_neigh_cmd"`
+            echo "$unresolved_kernel_neighbors" | awk '{print $1, $3}' | while read -r ip ifname; do
```
theasianpianist: Please use `<` redirection to stay consistent with the other loops

theasianpianist (changes requested):
I'm curious why this change is required in the first place, and how the retry values were obtained (10 retries @ 3 second intervals seems excessive). arp_update essentially makes a 'best effort' attempt to resolve any unresolved neighbor entries - this change is essentially forcing the entire script to wait for each individual neighbor which doesn't make any sense to me since we don't have any way of knowing when a neighbor will become reachable again. 

I'm ok with the script running every 2 minutes instead of every 5, but there are potentially hundreds or thousands of neighbors learned on a given device. If we have 100 unresolvable neighbors on a device, the script will take 50+ minutes to run and some neighbors will have to wait longer than if the script made a single attempt to resolve and retried every 2 (or 5) minutes. If we do want to add per-neighbor retries, we should at least batch all the retries and run them together instead of running them serially (but again I'm not sure why retries are necessary).

files/scripts/arp_update
```diff
@@ -14,6 +14,48 @@ logger () {
...
+}
+
+run_ndisc6_with_retry() {
+    local ifname="$1"
+    local ip="$2"
+    run_with_retry_cmd "ndisc6 on ${ifname} to ${ip}" \
+        ndisc6 -q -w 1 -1 "$ip" "$ifname" >/dev/null 2>&1
+}
+
+run_ping_with_retry() {
```
theasianpianist: Seems like this wrapper for ping is not needed, can we just call `run_with_retry_cmd` directly for ping and ping6?

files/scripts/arp_update
Janetxxx: As per our discussion with @prsunny, the 30 seconds is because there could be link flap cases where the ping may not respond transiently.

kperumalbfn:
@theasianpianist could you approve after review

theasianpianist:
@Janetxxx  can you please run the sonic-mgmt test `arp/test_arp_update.py::test_kernel_asic_mac_mismatch` with the changes from this PR and make sure that it's able to pass?

Janetxxx:
> @Janetxxx can you please run the sonic-mgmt test `arp/test_arp_update.py::test_kernel_asic_mac_mismatch` with the changes from this PR and make sure that it's able to pass?

@theasianpianist I've confirmed that it can pass all 5 tests in `arp/test_arp_update.py` include `test_kernel_asic_mac_mismatch`
```
---------------------------------------------------------------- generated xml file: /data/sonic-mgmt-int/tests/logs/arp/test_arp_update.xml -----------------------------------------------------------------
------------------------------------------------------------------------------------------- live log sessionfinish -------------------------------------------------------------------------------------------
01:36:45 __init__.pytest_terminal_summary         L0067 INFO   | Can not get Allure report URL. Please check logs
================================================================================== 5 passed, 1 warning in 211.93s (0:03:31) ==================================================================================
DEBUG:tests.conftest:[log_custom_msg] item: <Function test_dut_ping_learns_mac[str4-sn5640-2-None]>
```

theasianpianist (approved):

files/scripts/arp_update
```diff
@@ -14,6 +14,44 @@ logger () {
     command logger -i "$$" -t "arp_update" "$@"
 }
 
+# Function to run a command with retries on failure
+run_with_retry_cmd() {
+    local desc="$1"; shift
+    local max_tries=1
+    local interval=3
```
r12f: move this too and default to 3 is ok

kperumalbfn (approved):
LGTM, we can push to 202511 and 202412

vmittal-msft:
@Janetxxx please raise a manual pr for 202511 since there are conflicts

dhanasekar-arista:
> I'm curious why this change is required in the first place, and how the retry values were obtained (10 retries @ 3 second intervals seems excessive). arp_update essentially makes a 'best effort' attempt to resolve any unresolved neighbor entries - this change is essentially forcing the entire script to wait for each individual neighbor which doesn't make any sense to me since we don't have any way of knowing when a neighbor will become reachable again.
> 
> I'm ok with the script running every 2 minutes instead of every 5, but there are potentially hundreds or thousands of neighbors learned on a given device. If we have 100 unresolvable neighbors on a device, the script will take 50+ minutes to run and some neighbors will have to wait longer than if the script made a single attempt to resolve and retried every 2 (or 5) minutes. If we do want to add per-neighbor retries, we should at least batch all the retries and run them together instead of running them serially (but again I'm not sure why retries are necessary).

@theasianpianist @Janetxxx 

I've a concern on similar lines, why did we change "-w 0" to "-w 1" thus forcing it to wait for 1 sec on every arping/ndisc6 call.

pdam-arista:
Hi @theasianpianist @Janetxxx,

  We're seeing test failures in arp/test_arp_dualtor.py::test_standby_unsolicited_neigh_learning on the 202511 branch for dual-ToR active-active setups after PR #25169 was cherry-picked via #26815.

  ## Issue

  arp_update now takes 24-30+ seconds (was 2-3 seconds before), causing the test to timeout. The test expects completion within 5 seconds.

  ## Root Cause

  The -w 1 flag added to arping and ndisc6 in PR #25169 makes each neighbor announcement wait 1 second unnecessarily:

  ## Current (slow):
  arping -q -w 1 -c 1 -i "$ifname" "$ip"     # Waits 1s per neighbor
  ndisc6 -q -w 1 -1 "$ip" "$ifname"          # Waits 1s per neighbor

  ## Should be:
  arping -q -w 0 -c 1 -i "$ifname" "$ip"     # No wait
  ndisc6 -q -w 0 -1 "$ip" "$ifname"          # No wait

  Since GARP and unsolicited NAs are announcements (not requests), waiting for a reply is unnecessary.

  ## Request

  Can you update PR #25169 to use -w 0 instead of -w 1? This would restore the original performance while keeping the valuable retry logic you added.

  I've verified locally that setting the wait time to 0 resolves the issue and tests pass.

  I'm reluctant to increase the test timeout as that's just masking the performance regression. The product code fix would be the proper solution.

lolyu:
Hi @prsunny, this PR is currently causing a regression on 202511 dualtor due to the wait introduced. Do you think we need allow waiting for more time?

prsunny:
@theasianpianist , @Janetxxx , can you please check on this?

Janetxxx:
@prsunny @lolyu @pdam-arista @dhanasekar-arista I raised a PR to resolve dualtor performance regression by setting arping/ndisc6 wait timeout to 0: https://github.com/sonic-net/sonic-buildimage/pull/27058

PR #25172, opened 2026-01-23
SONIE

jasonbridges:
#### Why I did it

Initial PR for the SONIE build and installer creation.

##### Work item tracking

#### How to verify it

```
make target/sonie-vs.bin
./test_qemu_installer.sh -6
```

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511

- [ ] 
- [ ] 

Add new targets for build the SONIE image and a QEMU simulation docker container for integration testing.


files/build/versions-public/dockers/docker-sonic-mgmt-framework/versions-deb-bookworm
saiarcot895, line 1: Why do the version control files need to be modified here?

build_image.sh
```diff
@@ -22,152 +22,358 @@ fi
...
                 echo "${d##*/}" >> "$platforms_asic";
             fi;
         fi;
-    done
+    done < <(find -L ./device -maxdepth 2 -mindepth 2 -type d)
 
     # Add kvm to the list
+    if [ "$TARGET_MACHINE" = "broadcom" ] ; then
+      echo "x86_64-kvm_x86_64-r0" >> "$platforms_asic";
+    fi
```
saiarcot895: I'm guessing this is for testing purposes?

recovery_image_build.sh
```diff
@@ -0,0 +1,409 @@
...
+  fi
+
+  initramfs_start=$(date +%s)
+
+  # Resolve compression command before pipeline
+  local compress_cmd
+  case "${COMPRESSION_ALGO}" in
+      zstd)
+          # Use zstd -19 (ultra) for maximum compression to shrink UKI
+          compress_cmd="zstd -19 -T0"
```
saiarcot895: It might be worth it to use `--long`, since the UKI file will be significantly large (compared to a normal initramfs environment).

installer/install.sh
```diff
@@ -6,30 +6,32 @@
...
+fi
+export SCRIPT_DIR
+
+# Source utils
+if [ -r "${SCRIPT_DIR}/utils.sh" ]; then
+    . "${SCRIPT_DIR}/utils.sh"
+else
+    echo "ERROR: utils.sh not found in ${SCRIPT_DIR}" >&2
+    exit 1
+fi
```
aajith-arista, line 24: This causes issues with the old ONIE installer images.

Missing symlink
`installer/utils.sh` -> `sonie-installer/utils.sh`

installer/install.sh
```diff
@@ -45,225 +47,365 @@ read_conf_file() {
...
+        # Remove extra SONiC images if any
+        for f in "$demo_mnt"/image-* ; do
+            if [ -d "$f" ] && [ "$f" != "$demo_mnt/image-$running_sonic_revision" ] && [ "$f" != "$demo_mnt/$image_dir" ]; then
+                log_info "Removing old SONiC installation $f"
+                rm -rf "$f"
+            fi
+        done
+    else
+        demo_mnt="build_raw_image_mnt"
+        demo_dev="$cur_wd/target/sonic-vs-amd64.raw"
```
aajith-arista, line 247: This shouldn't hardcoded to the `vs` platform.

Should be `cur_wd/target/sonic-<CONFIGURED_PLATFORM>.raw`

installer/install.sh
```diff
@@ -45,225 +47,365 @@ read_conf_file() {
...
-if [ -d $demo_mnt/$image_dir ]; then
-    echo "Directory $demo_mnt/$image_dir/ already exists. Cleaning up..."
-    rm -rf $demo_mnt/$image_dir/*
-else
-    mkdir $demo_mnt/$image_dir || {
-        echo "Error: Unable to create SONiC directory"
-        exit 1
-    }
-fi
+    # Ensure XBOOTLDR is mounted at /boot for subsequent installers (sonic-installer)
```
aajith-arista, line 325: Checking `/boot` is probably wrong for `install_env=build`. Here, could the checks could be relative to the loop mounted path ?

installer/install.sh
aajith-arista: Also earlier the code wasn't always building .bin, .raw and .img.gz targets.
It was looking at `IMAGE_TYPE`  to choose the appropriate image to build. Shouldn't this be retained ?

It doesn't make sense to build say kvm images for say `PLATFORM=broadcom`

onie-image.conf
```diff
@@ -13,7 +14,8 @@
 ONIEIMAGE_VERSION=r0
 
 ## Filesystem root
-FILESYSTEM_ROOT=./fsroot-${TARGET_MACHINE}
+: ${FILESYSTEM_ROOT:=./fsroot-${TARGET_MACHINE}}
```
aajith-arista, line 17: Pls see https://github.com/sonic-net/sonic-buildimage/issues/26733

The current code relies on always setting FILESYSTEM_ROOT.
Otherwise sonic-broadcom.bin fails with parallel fsroot builds clobbering each other.

recovery_image_build.sh
```diff
@@ -0,0 +1,679 @@
...
+        -o -path ./mnt \
+        -o -path ./media \
+        -o -path ./var/cache/apt \
+        -o -path ./var/log \
+        -o -path ./usr/share/doc \
+        -o -path ./usr/share/man \
+        -o -path ./usr/share/info \
+        -o -path ./usr/share/locale \
+        -o -path ./usr/share/vim \
+        -o -path ./usr/share/zoneinfo \
```
amit-arista, line 126: We may need this, as /etc/localtime is needed by the database docker to start and it pointed to something in  /usr/share/zoneinfo/

sonie-installer/default_platform.conf
```diff
@@ -0,0 +1,912 @@
...
+  ')
+
+  # Robustly determine block device from partition using sysfs
+  if [ -n "$discovery_dev" ]; then
+    blk_dev=$(find_device_from_partdev "$discovery_dev")
+  fi
+
+  # Fallback detection for empty/corrupted disk if blkid failed
+  if [ -z "$blk_dev" ]; then
+    for dev in /dev/sda /dev/vda /dev/nvme0n1; do
```
aajith-arista: We want to prioritize a nvme drive over a SATA USB stick.
I think the order should be 
```
/dev/nvme0n1   /dev/vda  /dev/sda 
```

onie-image.conf
jasonbridges: good find

PR #25192, opened 2026-01-26
introduce Broadcom legacy image for 7060cx

lipxu:
#### Why I did it
Broadcom would still support TH SAI with current version, but no upgrade plan.

##### Work item tracking
36193015

#### How I did it
Introduce the Broadcom legacy image for 7060cx with the latest SONiC code and old SAI binary.

#### How to verify it
https://elastictest.org/scheduler/testplan/697ecfc0d7617b89998e1669
https://elastictest.org/scheduler/testplan/697ecfc0d7617b89998e1663

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [x] 202511

- [ ] 
- [ ]


platform/broadcom/sai-legacy.mk
```diff
@@ -0,0 +1,15 @@
+# Broadcom LEGACY SAI definitions
+LIBSAIBCM_LEGACY_VERSION = 13.2.1.28
+LIBSAIBCM_LEGACY_BRANCH_NAME = SAI_13.2.1_GA
```
Ryangwaite: For this PR we'll use the current 13.2 latest but in a future PR we'll move to the compat builds is that correct?

Ryangwaite (changes requested):

Ryangwaite:
Hi @lipxu what do you think about adding a `-th` suffix throughout the new "legacy" naming introduced in this PR? For example docker-syncd-brcm-legacy -> docker-syncd-brcm-legacy-th

This should reduce confusion in future when there is a second brcm platform that is EOL which would become another but different legacy version.

StormLiangMS:
hi @lipxu  Would you clarify why we need to change so many files like what we discussed in the description part? Did you test this change with master branch?

platform/broadcom/sai-legacy.mk
lipxu: Yes, that’s correct. Since I built the SONiC legacy image together with the Broadcom-compatible SAI binary, using the previous official SAI version allowed the SONiC image build to remain independent.
Now that the compatible SAI binary is available, I will switch to using it instead.

lipxu:
> Hi @lipxu what do you think about adding a `-th` suffix throughout the new "legacy" naming introduced in this PR? For example docker-syncd-brcm-legacy -> docker-syncd-brcm-legacy-th
> 
> This should reduce confusion in future when there is a second brcm platform that is EOL which would become another but different legacy version.

Thanks for your suggestion. The image was originally named legacy rather than legacy-th because it was intended for a special-case scenario and was expected to be used rarely.
We may consider renaming it to legacy-th if we think a similar solution is needed for other platforms in the future.

lipxu:
> hi @lipxu Would you clarify why we need to change so many files like what we discussed in the description part? Did you test this change with master branch?

Since the legacy ASIC type uses a compatible SAI binary built with a different Broadcom SDK version, we need to include the matching kernel modules and drivers. Thanks.
I haven’t tested it on the master branch yet. I only built a private image based on 202511 and ran the nightly tests.
However, since the PR has passed all the PR checks, I believe it should also work fine on the master branch.

lipxu:
> @lipxu Review comments:
> 
> **Blockers:**
> 
> 1. *_Hardcoded kernel version in \debian/opennsl-modules.install*_ — Install paths reference \lib/modules/6.12.41+deb13-sonic-amd64/extra. This will break on any kernel version change. Should be generated from \KVERSION\ at build time.
> 2. *_\sudo\ in \debian/rules*_ — Package build uses \sudo\ to mutate host paths under /lib/modules\ and /usr/src. This is unsafe in CI and may fail in non-root build containers. Should use \KDIR/\KERNDIR\ to pass kernel header paths instead.
> 
> **Question:**
> 
> 3. **~580 vendored source files** — The entire \saibcm-modules-legacy/\ kernel module source tree is checked into the repo. Is there a submodule or package alternative to avoid carrying this in-tree? If vendoring is the only option, please confirm in the PR description.

Thank you for your detailed review @StormLiangMS ! I'd like to clarify the scope:

### Blockers:

**1. Hardcoded kernel version in `debian/opennsl-modules.install`**
**2. `sudo` in `debian/rules`**

Both of these are **pre-existing issues** that have existed in the codebase since the original implementation and are currently present in:
- `platform/broadcom/saibcm-modules/debian/` (XGS family - in production)
- `platform/broadcom/saibcm-modules-dnx/debian/` (DNX family - in production)

**Key points:**
- These patterns were **not introduced by this PR** 
- These modules have been **building and running successfully in production for a long time** without issues reported
- This PR maintains **consistency** with the established codebase patterns

**Risk Assessment:**
Given that both existing modules work correctly with these patterns, the risk is minimal. 

**Recommendation:**
1. **Merge this PR** - It follows established patterns and adds Legacy ASIC support without introducing new technical debt
2. **Create a follow-up task** - Address these architectural concerns across all three modules (saibcm-modules, saibcm-modules-dnx, and saibcm-modules-legacy) in a unified refactoring effort

This approach ensures we don't block the Legacy ASIC feature while acknowledging the technical debt can be improved across the board.

### Question:

**~665 vendored source files**

The `saibcm-modules-legacy` source tree is currently vendored directly (not a submodule), consistent with the existing `saibcm-modules` module.

**Future plan:**
We plan to follow the `saibcm-modules-dnx` pattern and convert this to a git submodule in the future. However, this refactoring would:
- Require significant effort and a long time to set up a separate repository
- Need thorough validation across all xgs and legacy ASIC platforms

**Recommendation:**
We prefer to merge the current implementation now (matching the established `saibcm-modules` pattern) to unblock Legacy ASIC support, then improve it in a follow-up effort.

Gfrom2016 (approved):
LGTM

lipxu:
> > Hi @lipxu what do you think about adding a `-th` suffix throughout the new "legacy" naming introduced in this PR? For example docker-syncd-brcm-legacy -> docker-syncd-brcm-legacy-th
> > This should reduce confusion in future when there is a second brcm platform that is EOL which would become another but different legacy version.
> 
> Thanks for your suggestion. The image was originally named legacy rather than legacy-th because it was intended for a special-case scenario and was expected to be used rarely. We may consider renaming it to legacy-th if we think a similar solution is needed for other platforms in the future.

Hi, @Ryangwaite , 
I agree with your suggestion and plan to use legacy-th instead of legacy.
However, this change will affect many files and require rebuilding the image and running verification tests, which may take some additional time.
Would it be possible to merge this PR first? I can then create a follow-up PR to handle the renaming.
Thanks a lot for your understanding and support.

Ryangwaite:
> > > Hi @lipxu what do you think about adding a `-th` suffix throughout the new "legacy" naming introduced in this PR? For example docker-syncd-brcm-legacy -> docker-syncd-brcm-legacy-th
> > > This should reduce confusion in future when there is a second brcm platform that is EOL which would become another but different legacy version.
> > 
> > 
> > Thanks for your suggestion. The image was originally named legacy rather than legacy-th because it was intended for a special-case scenario and was expected to be used rarely. We may consider renaming it to legacy-th if we think a similar solution is needed for other platforms in the future.
> 
> Hi, @Ryangwaite , I agree with your suggestion and plan to use legacy-th instead of legacy. However, this change will affect many files and require rebuilding the image and running verification tests, which may take some additional time. Would it be possible to merge this PR first? I can then create a follow-up PR to handle the renaming. Thanks a lot for your understanding and support.

Yeah, that's a good approach

lipxu:
PR tests blocked, all PR test cases had been passed in previous run.
image

PR #24517, opened 2025-11-12
[build] Remove deprecated nodesource.com nodejs source from Bookworm slave container Dockerfile

nate-nexthop:
#### Why I did it
We are attempting to install nodejs 14 from nodesource.com using this line in the Dockerfile:
```
-RUN curl -sL https://deb.nodesource.com/setup_14.x | bash -
```
However, this script is deprecated and because of this the script inserts artificial pauses so the user can heed their warning:
```
================================================================================

                              DEPRECATION WARNING

  Node.js 14.x is no longer actively supported!
  <snip>
Continuing in 20 seconds ...


================================================================================
▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓
================================================================================

                           SCRIPT DEPRECATION WARNING


  This script, located at https://deb.nodesource.com/setup_X, used to
  install Node.js is deprecated now and will eventually be made inactive.
<snip>
TO AVOID THIS WAIT MIGRATE THE SCRIPT
Continuing in 60 seconds (press Ctrl-C to abort) ...
```

We need to stop using this script before it goes away as that will break builds. As it it, it wastes time.

Additionally, Bookworm has a Debian maintained nodejs, v18, which ends up being installed anyway because the sources added from nodesource.com are older.

##### Work item tracking
- Microsoft ADO **(number only)**:

fixes #24516 

#### How I did it

Just remove the line to add the nodesource.com APT sources.

#### How to verify it
Enter the sonic-slave-bash container before and after this change, and verify that the nodejs installed is v18:
Before:
```
Checking sonic-slave-user image: sonic-slave-bookworm-admin:fde86dbe366
admin@3ee74e91ba03:/sonic$ node -v
v18.20.4
```
After:
```
Checking sonic-slave-user image: sonic-slave-bookworm-admin:2d1524f8df4
admin@23cef98ccbd2:/sonic$ node -v
v18.20.4
```

#### Which release branch to backport (provide reason below if selected)

- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505

#### Tested branch (Please provide the tested image version)
master

- [ ] 
- [ ]

PR #25206, opened 2026-01-27
[build] upgrade Node.js to 20.x

yijingyan2:
#### Why I did it
The `make configure` failed while installing Node.js ([example pipeline](https://dev.azure.com/mssonic/build/_build/results?buildId=1023561&view=logs&j=0f6bec3d-a6b0-5abc-8247-29e720dd244c&t=b63e5369-9f8b-5fb6-430b-b609f6805df1)). The public key for node_14.x is not available. Migrating to newest supported version of Node.js
##### Work item tracking
- Microsoft ADO **(number only)**: 36514491

### DIFF
--- a/sonic-slave-bookworm/Dockerfile.j2
+++ b/sonic-slave-bookworm/Dockerfile.j2
@@ -666,7 +666,7 @@ RUN pip3 install git+https://github.com/aristanetworks/swi-tools.git@bead66bf261
 
 {% if CONFIGURED_ARCH != "amd64" -%}
 # Install node.js for azure pipeline
-RUN curl -sL https://deb.nodesource.com/setup_14.x | bash -
+RUN curl -sL https://deb.nodesource.com/setup_20.x | bash -
 RUN eatmydata apt-get install -y nodejs
 
 {%- if CROSS_BUILD_ENVIRON == "y" %}


nate-nexthop:
Out of curiosity, why do we need to install nodejs only on non-AMD64 builds?

Also I had opened https://github.com/sonic-net/sonic-buildimage/pull/24517 a while back to use the nodejs that Debian packages instead of relying on a third party repo. This follows what we have for the Trixie container: https://github.com/sonic-net/sonic-buildimage/blob/87d9f4808ed497574243e572becaad6c30ad7eb6/sonic-slave-trixie/Dockerfile.j2#L674

yijingyan2:
> Out of curiosity, why do we need to install nodejs only on non-AMD64 builds?
> 
> Also I had opened #24517 a while back to use the nodejs that Debian packages instead of relying on a third party repo. This follows what we have for the Trixie container:
> 
> https://github.com/sonic-net/sonic-buildimage/blob/87d9f4808ed497574243e572becaad6c30ad7eb6/sonic-slave-trixie/Dockerfile.j2#L674

Hi Nate, thanks for pointing this out. I'll close this PR. And for why we need to install nodejs only on non-AMD64, I think maybe is because that amd64 images already have nodejs, and for non-amd64 images, they aim to be minimal, so nodejs may not be included? not 100% sure on this.

PR #25220, opened 2026-01-28
[Nexthop] Improve Nexthop watchdog support

### LINKED ISSUE #25222: Enhancement: watchdog API implementation for Nexthop devices
### Is it platform specific

broadcom

### Importance or Severity

Medium

To support arming watchdog by a persistent watchdog puncher as well as manual calls to arm the watchdog.

### Current Behavior

- No support of watchdog puncher by software
- Watchdog API cannot distinguish between arming by watchdog puncher or an explicit, manual arm. This means that a manual arm might be overridden by the watchdog puncher.

As an example, assume device has a watchdog puncher software that arms the watchdog every 10 seconds. User then arms the watchdog `watchdogutil arm -s 20` (20 seconds). User does not call `watchdogutil disarm`. The device does not reboot or power cycle, because watchdog puncher re-arms the watchdog in 10 seconds or less.

### Proposed Behavior

- Separate code paths to allow watchdog puncher to re-arm the watchdog and explicit, manual arm
- Add a watchdog puncher (a systemd service and timer to arm watchdog using `watchdogutil`. Install and enable the timer. Every minute, the FPGA watchdog is armed (punched) with 3 minutes

Create a dedicated API for use by only watchdog puncher. For other explicit arming (by `watchdogutil arm` call, for example), create a watchdog pause file `/var/lock/pddf-locks/watchdog.pause` so that watchdog punching is to have no effect. `/var/lock/pddf-locks/watchdog.pause` is ideal because it is not persistent across reboot. The path is also mounted on pmon so that it is accessible in sonic-mgmt testing. When `disarm` is called , the pause file is removed. Note that the watchdog timer will eventually come in and re-arm the watchdog.

Using the example above, assume device has a watchdog puncher software that arms the watchdog every 10 seconds. User then arms the watchdog `watchdogutil arm -s 20` (20 seconds). User does not call `watchdogutil disarm`. The device eventually reboots or power cycles, in ~20 seconds.

domingo-nexthop:
#### Why I did it

Fixes #25222

- Please ignore commits like `IGNORE THIS COMMIT: Refactor unit tests for sonic-platform-modules-nexthop (PR #25166)`
- To add a persistent software watchdog puncher 
- To support arming the watchdog via a persistent "watchdog puncher" service while also allowing for manual `arm` calls.

##### Work item tracking

#### How I did it
- Add a separate API `arm_from_daemon()` for exclusive use by watchdog puncher
- Create a pause file `/var/lock/pddf-locks/watchdog.pause` when a manual `arm` has been called
- Remove the pause file when `disarm` is called. `/var/lock/pddf-locks/watchdog.pause` is ideal because it is not persistent across reboot. The path is also mounted on pmon so that it is accessible in sonic-mgmt testing. 
- Add a systemd watchdog puncher on Nexthop devices
- Add `override.conf` to disable `watchdog-control.service`, so that FPGA watchdog is enabled permanently 

#### How to verify it

- Unit tests
- [sonic-mgmt test](https://github.com/sonic-net/sonic-mgmt/blob/master/tests/platform_tests/api/test_watchdog.py) 

 ```
$ sudo systemctl list-timers watchdog
NEXT                        LEFT LAST                         PASSED UNIT           ACTIVATES
Wed 2026-01-28 19:33:52 UTC  12s Wed 2026-01-28 19:32:52 UTC 47s ago watchdog.timer watchdog.service

1 timers listed.
Pass --all to see loaded but inactive timers, too.
$ sudo python -c "import sonic_platform; print(sonic_platform.platform.Platform().get_chassis().get_watchdog().get_remaining_time())"
359
```

Wait for the next run.

```
$ sudo systemctl list-timers watchdog
NEXT                        LEFT LAST                         PASSED UNIT           ACTIVATES
Wed 2026-01-28 19:34:52 UTC   6s Wed 2026-01-28 19:33:52 UTC 53s ago watchdog.timer watchdog.service

1 timers listed.
Pass --all to see loaded but inactive timers, too.

$ sudo python -c "import sonic_platform; print(sonic_platform.platform.Platform().get_chassis().get_watchdog().get_remaining_time())"
357
```

Stop the watchdog.timer. System should powercycle eventually

```
$ sudo systemctl stop watchdog.timer
```

Lock file is created after `watchdogutil arm`
```
$ sudo watchdogutil status
Status: Armed
Time remaining: 323 seconds

$ sudo ls -l /var/lock/pddf-locks/watchdog.pause
ls: cannot access '/var/lock/pddf-locks/watchdog.pause': No such file or directory

$ date +%s; sudo watchdogutil arm -s 100
1769629448
Watchdog armed for 100 seconds

$ sudo ls -l /var/lock/pddf-locks/watchdog.pause
-rw-r--r-- 1 root root 10 Jan 28 19:42 /var/lock/pddf-locks/watchdog.pause
```

Watchdog is still armed after 60 seconds (punch frequency)

```
$ date +%s; sudo watchdogutil status
1769629510
Status: Armed
Time remaining: 38 seconds
```

```
$ sudo watchdogutil disarm
Watchdog disarmed successfully

$ date +%s; sudo ls -l /var/run/lock/pddf-locks/watchdog.pause
1769629534
ls: cannot access '/var/run/lock/pddf-locks/watchdog.pause': No such file or directory
```

Watchdog is re-armed by puncher eventually
```
$ date +%s; sudo watchdogutil status
1769629570
Status: Armed
Time remaining: 357 seconds
```

```
$ systemctl status watchdog-control.service
○ watchdog-control.service - watchdog control service
     Loaded: loaded (/usr/lib/systemd/system/watchdog-control.service; enabled-runtime; preset: enabled)
    Drop-In: /run/systemd/generator/watchdog-control.service.d
             └─environment.conf
             /etc/systemd/system/watchdog-control.service.d
             └─override.conf
     Active: inactive (dead) since Wed 2026-01-28 19:42:04 UTC; 4min 14s ago
   Duration: 13ms
 Invocation: 66f49e0f82f84d6e893c6c47f12305b0
   Main PID: 14372 (code=exited, status=0/SUCCESS)
   Mem peak: 2.7M
        CPU: 9ms
```

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [x] 202505
- [x] 202511

- [ ] 
- [ ] 

Support arming watchdog via a persistent "watchdog puncher" while also allowing for manual `arm` calls.


geans-pin (approved):

arlakshm (approved):

yxieca (commented):
## Review: [Nexthop] Improve Watchdog Support

Thanks for the PR @domingo-nexthop! The watchdog puncher design is clean overall. A few concerns:

### ⚠️ Concerns

**1. Pause file timestamp is written but never checked** — `_pause_watchdog_punching()` writes a timestamp (`pause_until_ts`) to the file, but `watchdog.service` only checks `if [[ ! -f "$PAUSE_FILE" ]]` — the timestamp is never read or used for expiry. If `arm(100)` is called and the caller crashes without calling `disarm()`, the pause file persists until reboot and the puncher stays paused indefinitely while the manual arm expires after 100s — so the watchdog fires and power-cycles the system. If that's the intended safety-net behavior, worth documenting it explicitly. Otherwise, consider having the timer check the timestamp and auto-resume after expiry.

**2. Race condition between timer and manual arm** — The timer fires every 60s. If `arm(seconds)` is called right before the timer fires, there's a brief window where the pause file hasn't been created yet but the daemon is about to re-arm to 360s, overriding the user's shorter timeout. This is inherent to the file-based approach and probably acceptable given the use case, but worth a comment in the code.

**3. `arm_from_daemon()` doesn't rate-limit error logging** — If the FPGA is in a bad state, `_do_real_arm(360)` will log an error via syslog every 60 seconds forever. Consider adding rate-limiting or a backoff mechanism for repeated failures.

**4. PR bundles unrelated changes** — The test framework refactoring (~1500 lines of churn), click compatibility fix (`eeprom_utils.py`), and the watchdog feature are all in one PR. The PR description even says to "ignore" the test refactoring commit. Per the project's guidelines ("Single logical change per PR"), I'd suggest splitting the test refactoring and the click compatibility fix into separate PRs.

### 📝 Minor
- `watchdog.service` uses `bash -c` with shell escaping — consider handling the pause check in Python directly to simplify.
- `debian/rules` now copies `*.timer` generically — any future `.timer` file in the service directory will be auto-included. Probably fine but worth noting.
- `OnUnitActiveSec=1min` may drift slightly from last activation; `OnCalendar=*:*:00` would be more precise if exact timing matters (likely doesn't for a 360s window).

### Request

Please address the above concerns with another PR.

PR #25226, opened 2026-01-28
Nexthop platform: Fix platform modules makefiles

nate-nexthop:
#### Why I did it

The targets listed in the .dep file must be the original targets that are first in the .mk file, otherwise the SRC_PATH will be empty. This leads to MPATH being wrong and git printing this error:

The .dep files were also incorrectly using the submodule method for finding files for DEP_FILES but this directory isn't a submodule so we don't need that.

Note, I have split this out from #24773 to keep the changes confined to Nexthop files only.

##### Work item tracking

Fixes #24773 

#### How to verify it

Run make configure and see one less instance of:

In the output.

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511

- [ ] 
- [ ]


nate-nexthop:
@geans-pin @Praveen-Brcm Could you please review and help merge? Thanks!

PR #25260, opened 2026-01-29
[Mellanox] [SED] Change/Reset SED password CLI

benle7:
HLD: https://github.com/sonic-net/SONiC/pull/2171
This PR is dependent on: https://github.com/sonic-net/sonic-platform-common/pull/620

#### Why I did it
Support the SONiC CLIs to change/reset SED password for Mellanox platforms.

##### Work item tracking

#### How I did it
- Install sedutil
- Implement the relevant abstract methods in Mellanox SedMgmt(BaseSedMgmt).

#### How to verify it
config sed change-password
config sed reset-password

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511

- [ ] 
- [ ]


DavidZagury (dismissed):

benle7:
@pavannaregundi @bhouse-nexthop

bhouse-nexthop (dismissed):
1. sedutil should be moved to toplevel, it is not mellanox-specific in any way.  I know of no other utility for this purpose on linux.  This should be installed by default for all vendors.
2. The sed_*.sh scripts are in general very generic.  It doesn't make sense for multiple vendors to need to duplicate these.  These too should be installed by default for all vendors, however there does need to be some vendor-specific "configuration" these scripts would need to read.  It is likely that 99% of vendors will use a TPM for SED unlocking.
    * Instead of naming variables tpm_reg_1, tpm_reg_2, tpm_reg_3, they really should have a purpose.  It looks like reg_3 is some sort of default password, and reg_1 and reg_2 are used for A / B purposes during rotation.
3. I find it _**very**_ strange that Nvidia/Mellanox is using 0x81010001 as that is a reserved handle for the RSA EK on the TPM as per https://trustedcomputinggroup.org/wp-content/uploads/TCG-TPM-v2.0-Provisioning-Guidance-Published-v1r1.pdf ... am I missing something about this?

saiarcot895:
@bhouse-nexthop If I'm reading [this](https://trustedcomputinggroup.org/wp-content/uploads/Registry-of-Reserved-TPM-2.0-Handles-and-Localities-Version-1.2-Revision-1.00_pub.pdf) right, 0x81010000-0x810100ff can be used to store the endorsement primary key (section 2.3.1), so maybe the use of 0x81010001 is fine here?

bhouse-nexthop:
> @bhouse-nexthop If I'm reading [this](https://trustedcomputinggroup.org/wp-content/uploads/Registry-of-Reserved-TPM-2.0-Handles-and-Localities-Version-1.2-Revision-1.00_pub.pdf) right, 0x81010000-0x810100ff can be used to store the endorsement primary key (section 2.3.1), so maybe the use of 0x81010001 is fine here?

I mean it looks pretty clear to me: 
image

Its reserved for endorsement primary keys only.  The reason there are 256 slots are for different key types (RSA vs ECC, etc).  I'm honestly surprised the TPM would even allow you to use it in that manner since this is an auto-generated key handle when you call `tpm2_createek`.

Let me find the reference to the actual slot, I think they started using it in later versions.

bhouse-nexthop:
Not finding anything superseding this:
https://trustedcomputinggroup.org/wp-content/uploads/TCG-TPM-v2.0-Provisioning-Guidance-Published-v1r1.pdf
image

benle7:
@bhouse-nexthop 
I will handle the comments soon. Thanks!

build_debian.sh
```diff
@@ -573,6 +573,14 @@ j2 files/build_templates/bmc_config.json.j2 | sudo tee $FILESYSTEM_ROOT/etc/soni
 sudo LANG=c chroot $FILESYSTEM_ROOT chmod 644 /etc/sonic/bmc_config.json
 sudo LANG=c chroot $FILESYSTEM_ROOT chown root:root /etc/sonic/bmc_config.json
 
+if [[ $CONFIGURED_PLATFORM == mellanox ]]; then
```
DavidZagury, line 591: Add comment about enabling SED that will help other to understand what needs to be added here

DavidZagury: Also - the 3 later lines are for every one - so make it not part of the if

benle7: DONE

DavidZagury (dismissed):

DavidZagury (approved):

liat-grozovik (approved):

liat-grozovik:
@bhouse-nexthop could you please help to approve this merge? just in case it is blocking the merge. as i see all the other dependencies are ok.

PR #25294, opened 2026-02-01
Preserve and restore factory default HWSKU across factory reset operations

gupurush:
#### Why I did it
Factory reset operations were not preserving the original intended default HWSKU value. When a platform's SKU was changed during operation or if there were platform naming convention changes, factory reset would not restore the correct original default SKU. This could lead to mismatches between the intended factory default configuration and the restored configuration.

Additionally, the reset-factory script was trapping the HUP signal which causes premature error cleanup when the terminal session disconnects during factory reset, leaving the system in an inconsistent state.

#### How I did it
- Modified rc.local to save the factory default_sku from /usr/share/sonic/device/$platform/default_sku to /etc/factory_default_sku during first boot
- Added restore_default_hwsku() function in reset-factory script that:
  - Reads the saved /etc/factory_default_sku file
  - Restores the original HWSKU in /usr/share/sonic/device/$platform/default_sku
  - Updates the HWSKU value in /etc/sonic/sonic-environment
  - Executes before config-setup to ensure correct SKU is used during factory reset
- Applied restoration logic only to 'default' and 'only-config' factory reset types
- The saved file persists across reboots in /etc and is backed up via overlay filesystem to /host/image-${SONIC_VERSION}/rw/etc
- Removed HUP signal from trap statement in reset-factory: The factory reset script should continue to completion and reboot even if the terminal session disconnects. HUP (hangup signal) is sent on terminal disconnection, and trapping it could cause premature error cleanup before critical operations complete. The script is designed to reach reboot, so it should not be interrupted by terminal hangups.

#### How to verify it
1. Boot a SONiC image and verify /etc/factory_default_sku is created with the correct default SKU
2. Change the HWSKU to a different value in default_sku file
3. Perform a factory reset (default or only-config mode)
4. Verify that the HWSKU is restored to the original factory default value from /etc/factory_default_sku
5. Check that /usr/share/sonic/device/$platform/default_sku and /etc/sonic/sonic-environment both contain the restored factory default HWSKU
6. Verify factory reset with 'keep-all-config' and 'keep-basic' modes work as expected
7. Verify factory reset completes successfully even if the terminal session disconnects during execution

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [x] 202511

Save factory default_sku on first boot and restore it during factory reset operations


anamehra:
Hi @arlakshm , @qiluo-msft , could you please help with the review? Thanks

Issue #14534, opened 2023-04-05
[yang] Need Yang for VNET table


prsunny:
#### Description

Create a Yang model for VNET table

#### Steps to reproduce the issue:
1.
2.
3.

#### Output of `show version`:

```
(paste your output here)
```
#### Output of `show techsupport`:

```
(paste your output here or download and attach the file here )
```

PR #25312, opened 2026-02-02
[Yang] Add Yang Model and tests for VNET_ROUTE table

alawing:
#### Why I did it
Added YANG model and tests for VNET_ROUTE table to provide schema validation and testing capabilities for VNET route configurations.
##### Work item tracking
- Microsoft ADO **(number only)**:#14534

#### How I did it

- Edit the existing sonic-vnet.yang model to define the VNET_ROUTE table schema

- Added comprehensive test cases covering both valid and invalid configurations

- Implemented validation for:

  - Mandatory nexthop field

  - Mandatory Interface name field

#### How to verify it

Review the YANG model against the config_db schema

Test validated
image

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [x] 202506
- [x] 202511

- [x] 202511
- [ ] 

https://github.com/prsunny/SONiC/blob/gh-pages/doc/vxlan/Vxlan_hld.md#21-config-db

### DIFF
--- a/src/sonic-yang-models/yang-models/sonic-vnet.yang
+++ b/src/sonic-yang-models/yang-models/sonic-vnet.yang
@@ -131,7 +131,7 @@ module sonic-vnet {
 
                 leaf ifname {
                     description "Interface names";
-                    type stypes:ifname-list;
+                    type string;
                     mandatory true;
                 }
             }
--- a/src/sonic-yang-models/yang-templates/sonic-types.yang.j2
+++ b/src/sonic-yang-models/yang-templates/sonic-types.yang.j2
@@ -64,14 +64,6 @@ module sonic-types {
             "Comma-separated IPv4 addresses";
     }
 
-    typedef ifname-list {
-        type string {
-            pattern '([a-zA-Z0-9]+)(,([a-zA-Z0-9]+))*';
-        }
-        description
-            "Comma-separated interface names";
-    }
-
     typedef admin_status {
         type enumeration {
             enum up;
--- a/src/sonic-yang-models/tests/yang_model_tests/tests/vnet.json
+++ b/src/sonic-yang-models/tests/yang_model_tests/tests/vnet.json
@@ -66,10 +66,7 @@
         "desc": "VNET route configuration with invalid IP in comma-separated nexthop list.",
         "eStrKey": "Pattern"
     },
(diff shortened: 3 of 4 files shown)


src/sonic-yang-models/yang-models/sonic-vnet.yang
```diff
@@ -104,6 +104,41 @@ module sonic-vnet {
...
+                }
+
+                leaf prefix {
+                    description "IPv4 prefix in CIDR format";
+                    type stypes:sonic-ip4-prefix;
+                }
+
+                leaf nexthop {
+                    description "Nexthop IP address";
+                    type inet:ipv4-address;
```
Leyza: Will this support multiple nexthops? I believe we have cases where we want multiple nexthops and ifnames (comma separated).

Leyza (changes requested):

src/sonic-yang-models/yang-models/sonic-vnet.yang
```diff
@@ -104,6 +104,41 @@ module sonic-vnet {
...
+                    mandatory true;
+                }
+
+                leaf ifname {
+                    description "Interface name";
+                    type string;
```
NavdhaJindal: nit: should ifname be a leafref to an existing port/portchannel in configdb?

src/sonic-yang-models/yang-models/sonic-vnet.yang
alawing: addressed

src/sonic-yang-models/yang-templates/sonic-types.yang.j2
```diff
@@ -64,6 +64,14 @@ module sonic-types {
             "Comma-separated IPv4 addresses";
     }
 
+    typedef ifname-list {
+        type string {
+            pattern '([a-zA-Z0-9]+)(,([a-zA-Z0-9]+))*';
```
NavdhaJindal: this regex will fail on subinterface names

Leyza: agree think should just keep it as a simple string

alawing: Changed back to string and removed related unit tests.

PR #25327, opened 2026-02-03
Enable PORT PHY attribute collection by default

dhanasekar-arista:
#### Why I did it
These are some useful counters which should be enabled by default.

##### Work item tracking

#### How I did it
Added the config to init_cfg.json.j2

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [x] 202505
- [x] 202511

I have tested it on 202505

- [ ] 
- [ ] 

Enable PORT PHY attribute collection by default

### DIFF
--- a/files/build_templates/init_cfg.json.j2
+++ b/files/build_templates/init_cfg.json.j2
@@ -29,6 +29,9 @@
         "PORT": {
             "FLEX_COUNTER_STATUS": "enable"
         },
+        "PORT_PHY_ATTR": {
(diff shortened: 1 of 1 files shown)


qiluo-msft:
@dhanasekar-arista, @prgeor, @yxieca Please be aware that in prouduction, the content here will be overwriten by golden_config.json, so suggest only adding default value in init_cfg.json, and always keep behavior backward-compatible.

@dhanasekar-arista Could you revert this change?

PR #25328, opened 2026-02-03
[build] Fix rsync failure in otel Docker builds when SONIC_CONFIG_USE_NATIVE_DOCKERD_FOR_BUILD is enabled

DavidZagury:
This is the same fix as in #24604 to the new otel docker

#### Why I did it
When building SONiC with `SONIC_CONFIG_USE_NATIVE_DOCKERD_FOR_BUILD=y`, Docker multi-stage builds fail during rsync operations with the following error:

```
rsync: [generator] failed to set times on "/changes-to-image": Read-only file system (30)
rsync error: some files/attrs were not transferred (see previous errors) (code 23)
```

This issue occurs specifically when:
- Using Docker Engine 20.10.x with BuildKit
- Running builds inside containers (Docker-in-Docker scenario)
- Accessing the host Docker daemon via socket mount
- Using `--mount=type=bind` in multi-stage Dockerfiles

The root cause is that Docker BuildKit creates read-only bind mounts, and rsync with the `-a` flag (which includes `-t` for preserving times) attempts to set timestamps on the mounted directory itself, failing on the read-only mount point.

This blocks builds in environments using native dockerd for build acceleration.

##### Work item tracking

#### How I did it
Use the defined in dockerfile-macros.j2 on the otel docker

#### How to verify it
Build SONiC using SONIC_CONFIG_USE_NATIVE_DOCKERD_FOR_BUILD=y
Verify Docker images build successfully

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511

- [ ] 
- [ ]

### DIFF
--- a/dockers/docker-sonic-otel/Dockerfile.j2
+++ b/dockers/docker-sonic-otel/Dockerfile.j2
@@ -1,4 +1,4 @@
-{% from "dockers/dockerfile-macros.j2" import install_debian_packages, install_python_wheels, copy_files %}
+{% from "dockers/dockerfile-macros.j2" import install_debian_packages, install_python_wheels, copy_files, rsync_from_builder_stage %}
 ARG BASE=docker-config-engine-bookworm-{{DOCKER_USERNAME}}:{{DOCKER_USERTAG}}
 
 FROM $BASE AS base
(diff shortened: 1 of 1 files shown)


Pterosaur (approved):

liat-grozovik:
@DavidZagury is this only for master or needed for other branches?
and before merging it please update the PR description with all the missing data

liat-grozovik (approved):

DavidZagury:
@liat-grozovik this would be needed for 202511 as well
I have updated the PR description

Issue #25279, opened 2026-01-30
Bug:SmartSwitch: Enabling mgmt VRF can crash orchagent/swss due to ZMQ bind address selection


rameshraghupathy:
### Is it platform specific

generic

### Importance or Severity

Critical

### Description of the bug

When mgmt VRF is enabled on SmartSwitch, orchagent.sh may bind the northbound ZMQ server to the eth0 management IP (in the mgmt VRF netns), causing ZMQ bind failure and orchagent/swss containers to crash-loop.

### Steps to Reproduce

- On a SmartSwitch NPU running the failing image, set DEVICE_METADATA|localhost subtype=SmartSwitch and configure a management IP on eth0 (MGMT_INTERFACE).

- Enable management VRF: set MGMT_VRF_CONFIG|vrf_global mgmtVrfEnabled=true.

- Apply the config: sudo config save -y then sudo config reload -y (or reboot).

- Observe swss crash-looping and orchagent exiting; docker logs swss shows ZMQ init/bind failure (binds to the eth0 mgmt IP).

### Actual Behavior and Expected Behavior

**Actual behavior:** After enabling `mgmtVrfEnabled=true` (and applying via `config reload`/reboot), `orchagent` fails to start (ZMQ bind/init fails due to choosing the eth0 mgmt IP), causing the `swss` container to crash-loop and related containers/services may flap.

**Expected behavior:** Enabling management VRF should not impact `orchagent` startup; `swss` should remain stable and `orchagent` should bind ZMQ to a valid address (e.g., loopback or midplane interface in the correct namespace) and continue running normally.

### Relevant log output

```shell
You can see it in the log sequence:
Create ZMQ server with address: tcp://10.3.150.162 (mgmt eth0 IP while mgmt VRF is enabled)

zmq_bind failed on endpoint … zmqerrno: 99 (Cannot assign requested address)

terminate called after throwing an instance of 'std::runtime_error'

orchagent terminated by SIGABRT (core dumped)

process orchagent exited unexpectedly → supervisor restarts/terminates swss (container crash-loop)
```

### Output of `show version`, `show techsupport`

```shell

```

anders-nexthop:
@prsunny @rameshraghupathy this issue is also being discussed here: https://github.com/sonic-net/sonic-swss/pull/4142 There is a solution proposed in that PR as well, please take a look.

vivekkarri:
[...]

PR #25330, opened 2026-02-03
[ZMQ] Avoid bind on management ip

vivekrnv:
#### Why I did it

To avoid starting orchagent on mgmt-ip which serves no purpose and additionally breaks fpmsyncd <-> orchagent communication if enabled

##### Work item tracking

#### How to verify it

Before this change:
```
# ps -aux | grep orch
 /usr/bin/orchagent -d /var/log/swss -b 1024 -s -q tcp://10.245.21.112
```

After this change:
```
# ps -aux | grep orch
/usr/bin/orchagent -d /var/log/swss -b 1024 -s -q tcp://127.0.0.1
```

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511

- [ ] 
- [ ]


dgsudharsan:
@qiluo-msft @prabhataravind  Can you please review?

vivekkarri:
@prsunny , @venkit-nexthop Please review

vivekkarri:
@prabhataravind, @prsunny Can you help signoff?

prabhataravind:
hi @vivekrnv what is the exact scenario where orchagent doesn't listen on eth0-midplane IP on smartswitch? I would think that is not expected. I agree that we don't need to use mgmt IP but that was only a fallback option if eth0-midplane was not up. Can you confirm this? Your change looks good otherwise.

prsunny (approved):
lgtm, @prabhataravind , can you review?

prabhataravind:
> lgtm, @prabhataravind , can you review?

@prsunny reviewed above. Waiting for Vivek's response on the comment above.

vivekkarri:
eth0-midplane is on DPU. If it doesn't have an IP,  the fallback is localhost.

On NPU, we don't have eth0-midplane, so it always would fallback to mgmt-ip which is incorrect INstead on NPU, we should always bind on localhost

prabhataravind:
> eth0-midplane is on DPU. If it doesn't have an IP, the fallback is localhost.
> 
> On NPU, we don't have eth0-midplane, so it always would fallback to mgmt-ip which is incorrect INstead on NPU, we should always bind on localhost

Ok, I was thinking the issue was on the DPU. My bad. Thanks for clarifying.

vivekkarri:
@kperumalbfn, Can you help signoff?

vivekkarri:
> Looks good to me. As discussed in #25279 we need to ensure that binding to mgmt_ip is not necessary any more.

@rustiqly Can you make a change in sonic-swss to remove the option of '-v mgmt'. Added here https://github.com/sonic-net/sonic-swss/pull/3299

PR #25341, opened 2026-02-04
[build] pulling k8s and geneva images from publicmirror

yijingyan2:
#### Why I did it
sometimes the builds get stuck when pulling k8s images
```
+ echo 'pulling universal k8s images ...'
pulling universal k8s images ...
+ sudo https_proxy= LANG=C chroot ./fsroot-vs docker pull k8s.gcr.io/pause:3.5

Login prior to pull:
Log in with your Docker ID or email address to push and pull images from Docker Hub. If you don't have a Docker ID, head over to https://hub.docker.com/ to create one.
```
With this change, we pull k8s images from ACR instead of k8s.gcr.io
##### Work item tracking
- Microsoft ADO **(number only)**: 36578969

#### How to verify
log from test build
image


files/build_templates/sonic_debian_extension.j2
```diff
@@ -921,13 +921,34 @@ sudo LANG=C DOCKER_HOST="$DOCKER_HOST" chroot $FILESYSTEM_ROOT /usr/local/bin/ge
 {% if include_kubernetes == "y" %}
 ## Pull in kubernetes docker images
 echo "pulling universal k8s images ..."
+{% if DEFAULT_CONTAINER_REGISTRY == "publicmirror.azurecr.io/" %}
```
liushilongbuaa: We can't have this if. All kinds of default container registry should support.

yijingyan2: updated

losha228 (approved):

liushilongbuaa (approved):
LGTM

PR #25394, opened 2026-02-09
[build][202511] Fix grpcio 1.15.1 install failure in arm64 env.

liushilongbuaa:
#### Why I did it
arm64 builds start to fail since 2026/02/09.
Root Cause is pypi package setuptools released new version v82.0.0. [LINK](https://setuptools.pypa.io/en/stable/history.html)
pkg_resources is removed from setuptools.

pip3 v23 starts to use virtual env to build packages. grpcio 1.51.1 didn't include pkg_resources in its dependencies. Build will fail.

Why amd64 and armhf build succeed? Because they have manylinux wheel. They don't need to build from source.

Open question:
How can we avoid this kind of build break? 
Disable virtual env for 'pip install'?
##### Work item tracking

#### How I did it
Use host env instead of virtual env when pip install grpcio.

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511

- [ ] 
- [ ]

### DIFF
--- a/dockers/docker-platform-monitor/Dockerfile.j2
+++ b/dockers/docker-platform-monitor/Dockerfile.j2
@@ -47,7 +47,7 @@ RUN apt-get install -y -t bookworm-backports \
 # install any dependencies required by the Arista sonic_platform package.
 # TODO: eliminate the need to install these explicitly.
 RUN pip3 install grpcio==1.51.1 \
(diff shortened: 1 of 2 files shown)


yejianquan:
This PR is approved but the approval is not set as expected

Issue #20353, opened 2024-09-26
j2cli unmaintained and broken on Python 3.12 - Consider alternatives?


cqjjjzr:
#### Description

I'm trying to build SONiC on Ubuntu 24.04, but the j2 tool fails to render the necessary templates. j2cli has been unmaintained for long, and the usage of removed feature is not likely to be fixed by the original author. A migration to alternatives may be necessary.

Possible candidate: https://github.com/kpfleming/jinjanator which is a fork of j2cli.

#### Steps to reproduce the issue:

Execute `make init` on a machine with Python 3.12 as the default Python installation.

#### Describe the results you received:

The environment normally initializes.

#### Describe the results you expected:

The build fails with the following output:

```
Traceback (most recent call last):
  File "/home/charlie/vsonic/venv/bin/j2", line 5, in <module>
    from j2cli import main
  File "/home/charlie/vsonic/venv/lib/python3.12/site-packages/j2cli/__init__.py", line 10, in <module>
    from j2cli.cli import main
  File "/home/charlie/vsonic/venv/lib/python3.12/site-packages/j2cli/cli.py", line 8, in <module>
    import imp, inspect
ModuleNotFoundError: No module named 'imp'
```

The builtin module `imp` was deprecated in Python 3.4 and removed in Python 3.12. However, the j2cli has been unmaintained since January (see https://github.com/kolypto/j2cli/issues/80), so the issue is not likely to be fixed.

#### Workaround

Create the Python venv using version <= 3.11.

judyjoseph:
Please add more details as to which branch you are seeing this issue. @sonic-build please check @xumia

cqjjjzr:
I've see this problem on branch master and 202405, but I expect it to appear on all branches since they are all using `j2cli` to apply the templates.

PR #25402, opened 2026-02-10
[prerequisites] Fix j2 install for Python 3.12+ hosts

rustiqly:
#### Why I did it

On hosts with Python >= 3.12, pip-installed j2cli fails with `ModuleNotFoundError: No module named 'imp'` (see #20353). The prerequisites script already prefers `jinjanator` (which works on Python 3.12+) with `j2cli` as fallback, but has two minor issues.

#### How I did it

- Use `apt-get install -y` instead of `apt install` for the j2cli fallback to ensure non-interactive operation.
- Add a PATH hint when `j2` is not found after install, since `pip3 install --user` puts binaries in `~/.local/bin` which may not be on PATH.

Note: Dockerfile changes for bookworm/bullseye base images are intentionally not included — those containers use Python 3.11/3.9 respectively and are unaffected by the `imp` module removal. The trixie slave container already uses apt-packaged `j2cli`.

#### How to verify it

Run `scripts/prerequisites.sh` on a host with Python >= 3.12 — jinjanator should install successfully via pip. On systems where pip fails, the fallback to `apt-get install -y j2cli` runs non-interactively.

- [ ] 202311
- [ ] 202405
- [ ] 202411

Fix prerequisites.sh j2cli fallback to use non-interactive apt-get -y and add PATH hint for pip user installs.

🦔

### DIFF
--- a/scripts/prerequisites.sh
+++ b/scripts/prerequisites.sh
@@ -36,11 +36,12 @@ run_step "Installing prerequisites (python3-pip, git)" \
     sudo apt install -y python3-pip git
 
 run_step "Installing jinjanator (j2)" \
-    bash -c 'pip3 install --user jinjanator || sudo apt install j2cli'
+    bash -c 'pip3 install --user jinjanator || sudo apt-get install -y j2cli'
 
 echo "==> Testing j2 availability..."
 if ! command -v j2 >/dev/null 2>&1; then
     echo "[ERROR] j2 is not runnable."
+    echo "If installed via pip, ensure ~/.local/bin is in your PATH."
     echo "Please logout and login, then run the script again."
     exit 1
 else


rustiqly:
@lguohan @yxieca Could you review this? Switches j2cli from pip to the Debian-packaged version to fix Python 3.12+ compatibility (fixes #20353).

rustiqly:
Thanks for the review @saiarcot895 — all good points.

Updated the PR:
- **Dropped the bullseye changes** — agreed it's no longer used.
- **Dropped the bookworm Dockerfile change** — Python 3.11 isn't affected by the `imp` removal.
- **Kept prerequisites.sh scoped** — the host may run Python 3.12+. The script already preferred jinjanator with j2cli as fallback, so I just fixed `apt install` → `apt-get install -y` (non-interactive) and added a PATH hint for pip `--user` installs.

Agreed that jinjanator is the right long-term choice over Debian's j2cli packaging.

rustiqly:

scripts/prerequisites.sh
```diff
@@ -36,11 +36,12 @@ run_step "Installing prerequisites (python3-pip, git)" \
     sudo apt install -y python3-pip git
 
 run_step "Installing jinjanator (j2)" \
-    bash -c 'pip3 install --user jinjanator || sudo apt install j2cli'
+    bash -c 'pip3 install --user jinjanator || sudo apt-get install -y j2cli'
```
lguohan, line 39: does this real make difference? the real difference seems to be -y, is that right?

rustiqly: Yes, you're right — the two actual changes are:

1. `apt install` → `apt-get install -y` (ensures non-interactive operation in scripts/CI)
2. Added a PATH hint when `j2` isn't found (`~/.local/bin` may not be on PATH after pip `--user` install)

Pretty minimal. Happy to close this if it's not worth the churn.

PR #25410, opened 2026-02-10
[build] Create all distro python-debs and python-wheels dirs in configure

rustiqly:
#### What I did

The `configure` target in `slave.mk` creates per-distro directories for `target/debs/` and `target/files/` (jessie through trixie), but only creates `target/python-debs/` and `target/python-wheels/` for the current `$(BLDENV)`.

This inconsistency can cause build failures when transitioning between build environments (e.g., bookworm → trixie) after a `make clean` + `make configure`, if the python-debs or python-wheels directory for the next build phase doesn't exist yet.

#### How I did it

- Added explicit per-distro path variables for `python-debs` and `python-wheels` (matching the existing pattern for `DEBS_PATH` and `FILES_PATH`)
- Updated the `configure` target to create all distro-specific `python-debs/` and `python-wheels/` directories, consistent with how `debs/` and `files/` directories are handled

#### How to verify it

1. Run `make configure PLATFORM=vs`
2. Verify that `target/python-debs/` and `target/python-wheels/` contain subdirectories for all supported distros (jessie, stretch, buster, bullseye, bookworm, trixie)

Previously, only the current `$(BLDENV)` subdirectory would be created.

### DIFF
--- a/slave.mk
+++ b/slave.mk
@@ -45,6 +45,18 @@ BOOKWORM_DEBS_PATH = $(TARGET_PATH)/debs/bookworm
 BOOKWORM_FILES_PATH = $(TARGET_PATH)/files/bookworm
 TRIXIE_DEBS_PATH = $(TARGET_PATH)/debs/trixie
 TRIXIE_FILES_PATH = $(TARGET_PATH)/files/trixie
+JESSIE_PYTHON_DEBS_PATH = $(TARGET_PATH)/python-debs/jessie
+JESSIE_PYTHON_WHEELS_PATH = $(TARGET_PATH)/python-wheels/jessie
+STRETCH_PYTHON_DEBS_PATH = $(TARGET_PATH)/python-debs/stretch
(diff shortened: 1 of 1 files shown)


rustiqly:

liushilongbuaa:
@rustiqly , currently, 'make configure' will mkdir for bookworm and trixie. Because bookworm folder will be kept before migration finished. I don't understand why build breaks.
If we really need explicit per-distro path variables, please only keep bookworm and trixie. We don't need keep releases < bookworm, because if we add a bullseye target, much more works need to be done.

rustiqly:
@liushilongbuaa You're right — I re-examined this and the existing `configure` target correctly creates `$(PYTHON_DEBS_PATH)` and `$(PYTHON_WHEELS_PATH)` using `$(BLDENV)`, which handles both the bookworm and trixie phases.

The issue I hit was from manual cleanup (`rm -rf target/python-debs`) between builds without re-running `make configure` — not a bug in the build system itself.

Happy to close this PR since it's not addressing a real issue. Sorry for the noise.

PR #25411, opened 2026-02-10
[build] Add SONIC_BUILD_MEMORY config to limit container memory

rustiqly:
#### What I did

Added `SONIC_BUILD_MEMORY` and `SONIC_BUILD_MEMORY_SWAP` configuration knobs to control the memory limit of the build container.

#### Why I did it

When building with `SONIC_BUILD_JOBS > 1`, parallel C++ compilation (especially packages like sonic-swss, p4lang-p4c) can exhaust system RAM. The kernel OOM killer then silently terminates build processes — or worse, unrelated applications like browsers and IDEs — with no error message or diagnostics. The build just stops, and the developer has no idea why.

By setting a Docker `--memory` limit on the build container, OOM behavior becomes **predictable and contained**: the container gets killed with a clear error instead of the kernel randomly shooting processes across the entire system.

#### How I did it

- Added `SONIC_BUILD_MEMORY` in `rules/config` (commented out by default — no behavior change)
- Added `SONIC_BUILD_MEMORY_SWAP` for optional swap limit control
- Wired both into `DOCKER_RUN` in `Makefile.work` via `--memory` and `--memory-swap` flags

#### How to verify it

1. Set `SONIC_BUILD_MEMORY = 24g` in `rules/config.user`
2. Run a build with `SONIC_BUILD_JOBS=4`
3. Verify with `docker inspect <container>` that MemoryLimit is set
4. If memory is exceeded, the container exits with OOM error (instead of silent kernel kills)

#### Usage

```makefile
# In rules/config.user:
SONIC_BUILD_MEMORY = 24g
SONIC_BUILD_MEMORY_SWAP = 32g
```

Recommended: leave 4-6 GB free for the host OS.

### DIFF
--- a/Makefile.work
+++ b/Makefile.work
@@ -304,10 +304,24 @@ endif
 
 # Consider removing the --ulimit flag once nothing older
 # than Bullseye is being used as a slave container.
+# Build container memory limit (set SONIC_BUILD_MEMORY=none to disable).
+# Constrains OOM kills to the container instead of the kernel silently
+# killing unrelated host processes.
+SONIC_BUILD_MEMORY_FLAGS =
+ifneq ($(SONIC_BUILD_MEMORY),)
+ifneq ($(SONIC_BUILD_MEMORY),none)
+    SONIC_BUILD_MEMORY_FLAGS += --memory=$(SONIC_BUILD_MEMORY)
+    ifneq ($(SONIC_BUILD_MEMORY_SWAP),)
+        SONIC_BUILD_MEMORY_FLAGS += --memory-swap=$(SONIC_BUILD_MEMORY_SWAP)
+    endif
+endif
+endif
+
 DOCKER_RUN := docker run --rm=true --privileged --init \
     -v $(DOCKER_BUILDER_MOUNT) \
     -v "$(DOCKER_LOCKDIR):$(DOCKER_LOCKDIR)" \
     -w $(DOCKER_BUILDER_WORKDIR) \
+    $(SONIC_BUILD_MEMORY_FLAGS) \
     -e "http_proxy=$(http_proxy)" \
     -e "https_proxy=$(https_proxy)" \
     -e "no_proxy=$(no_proxy)" \
--- a/rules/config
+++ b/rules/config
@@ -16,6 +16,21 @@ SONIC_CONFIG_BUILD_JOBS = 1
 # Corresponding -j argument will be passed to make/dpkg commands that build separate packages
 SONIC_CONFIG_MAKE_JOBS = $(shell nproc)
 
+# SONIC_BUILD_MEMORY - set memory limit for the build container.
+# When set, passes --memory and --memory-swap flags to docker run.
+# This prevents the kernel OOM killer from silently killing build processes
+# (or unrelated host processes) when parallel builds exhaust system RAM.
+# The OOM will instead be contained within the build container.
+# Example: SONIC_BUILD_MEMORY = 24g
+# Set to "none" to explicitly disable.
+# SONIC_BUILD_MEMORY =
+
+# SONIC_BUILD_MEMORY_SWAP - set memory+swap limit for the build container.
+# Only takes effect when SONIC_BUILD_MEMORY is set. Defaults to double
+# the value of SONIC_BUILD_MEMORY if not specified (Docker default behavior).
+# Set to -1 for unlimited swap. Set equal to SONIC_BUILD_MEMORY to disable swap.
+# SONIC_BUILD_MEMORY_SWAP =
+
 # DEFAULT_BUILD_LOG_TIMESTAMP - add timestamp in build log
 # Supported format: simple, none
 DEFAULT_BUILD_LOG_TIMESTAMP = none


rustiqly:

Makefile.work
```diff
@@ -315,6 +315,19 @@ DOCKER_RUN := docker run --rm=true --privileged --init \
     --ulimit nofile=524288:524288 \
     $(SONIC_BUILDER_EXTRA_CMDLINE)
 
+# Apply memory limits to the build container when configured.
+# This constrains OOM kills to the container instead of the kernel
+# silently killing unrelated host processes.
+# Set SONIC_BUILD_MEMORY=none to disable.
+ifneq ($(SONIC_BUILD_MEMORY),)
+ifneq ($(SONIC_BUILD_MEMORY),none)
+    DOCKER_RUN += --memory=$(SONIC_BUILD_MEMORY)
```
liushilongbuaa: Adding option in the end may be not strong. When EXTRA_CMDLINE changed, maybe not work.
Can we add before ' -e http_proxy=xx'?

rules/config
```diff
@@ -16,6 +16,21 @@ SONIC_CONFIG_BUILD_JOBS = 1
...
 SONIC_CONFIG_MAKE_JOBS = $(shell nproc)
 
+# SONIC_BUILD_MEMORY - set memory limit for the build container.
+# When set, passes --memory and --memory-swap flags to docker run.
+# This prevents the kernel OOM killer from silently killing build processes
+# (or unrelated desktop applications) when parallel builds exhaust system RAM.
+# The OOM will instead be contained within the build container.
+# Default: auto-detected (total system RAM minus 4 GB reserve for host, minimum 4 GB).
+# Set to a specific value (e.g., 24g) to override, or "none" to disable.
+SONIC_BUILD_MEMORY = $(shell awk '/MemTotal/ {mem=int($$2/1024/1024)-4; if(mem<4) mem=4; printf "%dg", mem}' /proc/meminfo)
```
liushilongbuaa: It's better to comment this line to keep default value.

Makefile.work
rustiqly: Good point. Moved it — memory flags are now computed into `SONIC_BUILD_MEMORY_FLAGS` before the `DOCKER_RUN :=` block and inserted inline before `-e http_proxy`. This way `SONIC_BUILDER_EXTRA_CMDLINE` stays last and can still override everything.

rules/config
rustiqly: Done — default value is now commented out. Users opt in by uncommenting and setting a value.

rustiqly:
## Test Results — SONIC_BUILD_MEMORY_FLAGS

Tested both Makefile logic expansion and Docker runtime behavior.

### Makefile Logic Tests (6/6 pass)

Using a test harness that copies the exact `SONIC_BUILD_MEMORY_FLAGS` logic from `Makefile.work`:

| # | Scenario | Expected | Result |
|---|----------|----------|--------|
| 1 | `SONIC_BUILD_MEMORY` unset (default) | No `--memory` flag | ✅ PASS |
| 2 | `SONIC_BUILD_MEMORY=24g` | `--memory=24g` | ✅ PASS |
| 3 | `SONIC_BUILD_MEMORY=24g` + `SONIC_BUILD_MEMORY_SWAP=48g` | Both flags | ✅ PASS |
| 4 | `SONIC_BUILD_MEMORY=none` (explicit disable) | No `--memory` flag | ✅ PASS |
| 5 | Memory flag placement | Before `-e http_proxy` | ✅ PASS |
| 6 | `SONIC_BUILDER_EXTRA_CMDLINE` stays last | `--cpus=4` at end | ✅ PASS |

### Docker Runtime Tests (4/4 pass)

| # | Scenario | Expected | Result |
|---|----------|----------|--------|
| 7 | `--memory=128m` → cgroup limit | `134217728` (128MB) | ✅ PASS |
| 8 | No `--memory` → unlimited | `max` | ✅ PASS |
| 9 | OOM contained in container | Exit 137 (SIGKILL), host unaffected | ✅ PASS |
| 10 | `--memory=128m --memory-swap=128m` | memory=128MB, swap=0 | ✅ PASS |

All 10 tests pass. The feature works as designed: opt-in memory limit, OOM kills stay inside the container, and `SONIC_BUILDER_EXTRA_CMDLINE` can still override.

rustiqly:
## Integration Test — Build-Like Workload Under Memory Constraint

Simulated a real build scenario: compile + run a memory-hungry process inside a Docker container, with and without `--memory`.

### Test Setup
- Container image: `debian:bookworm-slim` + `build-essential`
- Workload: `make` builds and runs a C program that allocates 10MB chunks until OOM
- Host: 30GB RAM

### Results

**Test A — WITHOUT `--memory` (current behavior):**
```
Allocated 13560MB
Allocated 13570MB
...
Allocated 13630MB
```
Process consumed **13.6GB** of host RAM before being externally killed (timeout). Without the timeout, it would eat all 30GB and trigger the **kernel OOM killer**, which can kill **any** host process — including your desktop, SSH sessions, or other builds.

**Test B — WITH `--memory=64m` (this PR):**
```
Allocated 40MB
Allocated 50MB
...
Allocated 120MB
make: *** [Makefile:3: all] Killed
Exit code: 137 (SIGKILL from cgroup OOM)
```
Process OOM-killed at the container boundary. Host unaffected.

**Test C — Host verification after OOM:**
```
               total        used        free
Mem:            30Gi       7.8Gi        14Gi
```
Host RAM untouched.

### Kernel Log Confirmation
```
oom-kill:constraint=CONSTRAINT_MEMCG,...task=mem_hog,pid=76693
Memory cgroup out of memory: Killed process 76693 (mem_hog) total-vm:135648kB, anon-rss:61248kB
```
The kill was triggered by `CONSTRAINT_MEMCG` (Docker cgroup limit) — **not** the system-wide OOM killer. This is exactly the behavior the PR enables: OOM stays contained inside the build container.

rules/config
```diff
@@ -16,6 +16,21 @@ SONIC_CONFIG_BUILD_JOBS = 1
 # Corresponding -j argument will be passed to make/dpkg commands that build separate packages
 SONIC_CONFIG_MAKE_JOBS = $(shell nproc)
 
+# SONIC_BUILD_MEMORY - set memory limit for the build container.
+# When set, passes --memory and --memory-swap flags to docker run.
+# This prevents the kernel OOM killer from silently killing build processes
+# (or unrelated host processes) when parallel builds exhaust system RAM.
+# The OOM will instead be contained within the build container.
+# Example: SONIC_BUILD_MEMORY = 24g
+# Set to "none" to explicitly disable.
```
saiarcot895, line 25: What's the purpose of having an explicit `none` option for this setting?

PR #25415, opened 2026-02-10
test_seq_idf_isolation.py failure on T2

selva-nexthop:
Root cause:
On T2 device, the TO_BGP_PEER route-map sequence 60 was missing "on-match next", causing it to stop processing and preventing sequence 100 (CHECK_IDF_ISOLATION) from being reached.

Route → Seq 50 (tag contributing routes) → on-match next → Seq 60 (strip local anchor community) → STOP (no on-match next) → Seq 100 (CHECK_IDF_ISOLATION) ← NEVER REACHED!
This caused:
- IDF isolation to not work (no-export community not added)
- test_seq_idf_isolation.py to fail on T2

In disaggregated case, though only LT2 will have the peers with T1, we will see this issue consistently when UT2 is peered with T1, which has discovered the gaps in the templates.  This gap is potential enough cause issues in other cases as well.

Resolution:
Add "on-match next" to sequence 60 in TO_BGP_PEER_V4/V6 route-maps to allow processing to continue to sequence 100 (CHECK_IDF_ISOLATION).

Route → Seq 50 (tag contributing routes) → on-match next → Seq 60 (strip local anchor community) → on-match next → Seq 100 (CHECK_IDF_ISOLATION) ← NOW REACHABLE

Tested the below testcases with this fix and it is passing.
- test_seq_idf_isolation.py
- test_prefix_list.py
- test_ipv6_nlri_over_ipv4.py
- test_bgp_update_timer.py
- test_route_flap.py

##### Work item tracking

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [x] 202511

- [ ] 
- [ ]

### DIFF
--- a/src/grub2/patch/large-uid-skip-cpio-ustar.patch
+++ b/src/grub2/patch/large-uid-skip-cpio-ustar.patch
@@ -1,0 +1,16 @@
+diff --git a/tests/cpio_test.in b/tests/cpio_test.in
+index 5742cf17b..add41ef22 100644
+--- a/tests/cpio_test.in
++++ b/tests/cpio_test.in
+@@ -11,6 +11,10 @@ fi
+ "@builddir@/grub-fs-tester" cpio_odc
+ "@builddir@/grub-fs-tester" cpio_newc
+ "@builddir@/grub-fs-tester" cpio_crc
+-"@builddir@/grub-fs-tester" cpio_ustar
++# If the UID of the user running tests is > 2097151 skip testing ustar as the format does
++# not support large uids.
++if [ `id -u` -le "2097151" ] ; then
++   "@builddir@/grub-fs-tester" cpio_ustar
++fi
+ "@builddir@/grub-fs-tester" cpio_hpbin
+ "@builddir@/grub-fs-tester" cpio_hpodc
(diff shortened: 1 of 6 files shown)


tjchadaga:
@selva-nexthop - currently, test_seq_idf_isolation only runs if there are T1 neighbors, since the feature is intended to add the no-export or withdraw route-maps towards T1.  In your tests, it looks like you have T1 neighbors on UT2, which is causing this issue. By design, UT2 would have only LT2 as the downstream peer and not T1

In a disaggregated T2, these tests would run on LT2 which will not have the anchor route-maps that caused this issue.

The change made here (adding on-match-next) is a good fix, but the title of the PR is misleading. Could you please change this?

selva-nexthop:
> @selva-nexthop - currently, test_seq_idf_isolation only runs if there are T1 neighbors, since the feature is intended to add the no-export or withdraw route-maps towards T1. In your tests, it looks like you have T1 neighbors on UT2, which is causing this issue. By design, UT2 would have only LT2 as the downstream peer and not T1
> 
> In a disaggregated T2, these tests would run on LT2 which will not have the anchor route-maps that caused this issue.
> 
> The change made here (adding on-match-next) is a good fix, but the title of the PR is misleading. Could you please change this?

@tjchadaga, thanks for the insight. I will update the PR title and description accordingly.

PR #25430, opened 2026-02-11
[libyang] Fix hash table resize assertion failure for large configurations

xincunli-sonic:
#### Why I did it

Fix critical minigraph loading failure caused by libyang hash table resize assertion failure.

Root Cause: libyang 1.0.73 has a known bug in hash table resize logic. When the hash table resizes during large config processing, the lydict_val_eq function uses incorrect string length comparison, causing false equality matches and insertion failures.

Fix: Increase LYHT_MIN_SIZE from 8 to 65536. This prevents resize operations for configs with <50,000 unique strings, avoiding the buggy code path entirely. Memory impact is negligible (~2-4MB, <0.05% of device memory).

##### Work item tracking
- Microsoft ADO **(number only)**: 36733961

#### How I did it
1. Added patch file libyang-increase-lyht-min-size.patch
2. Updated series to include the new patch

#### How to verify it
Build verification: Rebuild libyang package

Manual test on device:

Before fix: Crashes at some number of BGP entries with assertion failure
After fix: Successfully loads all BGP entries

Test results:

Full minigraph loading test: Device successfully completes Hardconfig stage with original minigraph that previously failed.

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511

- [ ] 
- [ ]


prasadpkorhale:
Is there a risk of increasing the memory usage due to this change as we will allocate 64K entries always for various parts of the config. Is it possible to do memory profiling for this and check the pre/post results for a particularly large config for a device with high no. of ports?

xincunli-sonic:
> Is there a risk of increasing the memory usage due to this change as we will allocate 64K entries always for various parts of the config. Is it possible to do memory profiling for this and check the pre/post results for a particularly large config for a device with high no. of ports?

I have verified this case, here is the result

### Memory Impact
| Metric | Value |
|--------|-------|
| Additional memory per hash table | ~1.25 MB |
| Number of hash table contexts | 2-3 |
| Total additional memory | ~2.5-4 MB |
| Typical device memory | 8-32 GB |
| Impact | **Negligible (<0.05%)** |

src/libyang/patch/libyang-increase-lyht-min-size.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/src/hash_table.h b/src/hash_table.h
+--- a/src/hash_table.h
++++ b/src/hash_table.h
+@@ -53,7 +53,7 @@
+ #define LYHT_SHRINK_PERCENTAGE 25
+ 
+ /** never shrink beyond this size */
+-#define LYHT_MIN_SIZE 8
++#define LYHT_MIN_SIZE 65536
```
qiluo-msft, line 9: Do you have a testcase? ideally, it fails with old code, and succeed with new code.

xincunli-sonic: I don't see any test case can cover this patch by old and new code, but I did test in lab and verified in prod devices.

yxieca:
Question on libyang patch: LYHT_MIN_SIZE jumps from 8 to 65536. That’s a big baseline memory increase. Could you share the motivation/data (e.g., perf or collision issues) and whether a smaller value would be sufficient?

yxieca:
Thanks for the data on memory impact (2.5–4 MB total, <0.05%). That answers my concern on the large LYHT_MIN_SIZE bump.

vmittal-msft:
@xincunli-sonic is this part of 202511 ? If so,please update label "Included in 202511 branch"

PR #25438, opened 2026-02-11
refactor: Create libyang packages at @libyang* repos

blorente:
This way, several containers can depend on them.


blorente:
I meant to open this on a fork, apologies

Issue #25291, opened 2026-02-01
Bug: healthd crashes on non-service events with multiple dots in unit name


benle7:
### Is it platform specific

generic

### Importance or Severity

Critical

### Description of the bug

healthd / sysmonitor crashes with:
ValueError: too many values to unpack (expected 2)
when receiving events for non-service units (e.g. .device units).

The crash is caused by incorrect parsing in src/system-health/health_checker/sysmonitor.py,
where the code assumes that unit names contain exactly one dot (expected in service units) and performs:
srv_name, last = event.split('.')

However, systemd device and mount unit names may legally contain multiple dots.
When such events are received, event.split('.') returns more than two elements,
causing an unhandled ValueError.

src/system-health/health_checker/sysmonitor.py:Sysmonitor:check_unit_status:
srv_name,last = event.split('.')
            if last == "service":

Since this code is relevant to .service events, we need to verify that the event suffix is '.service',
before the split.

### Steps to Reproduce

- Boot SONiC with healthd running
- Observe systemd events for .device units
- Check syslog

### Actual Behavior and Expected Behavior

Actual Behavior:
"ERR healthd: system_servicetoo many values to unpack (expected 2)"

Expected Behavior:
healthd should not crash when receiving systemd events for non-.service units.

### Relevant log output

```shell
NOTICE healthd#sysmonitor[20198]: Received event:dev-disk-by\x2ddiskseq-12.device from source:sysbus time:2026-01-27 12:14:19
NOTICE healthd#sysmonitor[20198]: Received event:dev-disk-by\x2dloop\x2dinode-259:3\x2d524302.device from source:sysbus time:2026-01-27 12:14:19
NOTICE healthd#sysmonitor[20198]: Received event:dev-disk-by\x2dloop\x2dref-\x5cx2fhost\x5cx2fimage\x2d202412_RC.193\x2dda4488a34_Internal\x5cx2ffs.squashfs.device from source:sysbus time:2026-01-27 12:14:19
ERR healthd: system_servicetoo many values to unpack (expected 2)
NOTICE healthd#sysmonitor[20198]: Received event:dev-loop2.device from source:sysbus time:2026-01-27 12:14:19
NOTICE healthd#sysmonitor[20198]: Received event:sys-devices-virtual-block-loop2.device from source:sysbus time:2026-01-27 12:14:19
[...]

PR #25440, opened 2026-02-11
[healthd] Fix crash on systemd unit names with multiple dots

rustiqly:
## Description

Fixes healthd/sysmonitor crashing with `ValueError: too many values to unpack (expected 2)` when receiving dbus events for systemd units whose names contain multiple dots.

**Root cause:** `event.split(".")` in both `get_unit_status()` and `check_unit_status()` assumes unit names contain exactly one dot. Systemd device and mount unit names can legally contain multiple dots (e.g., `sys-devices-pci0000:00-0000:00:1f.0.device`), causing the unpacking to fail.

**Fix:** Replace `split(".")` with `rsplit(".", 1)` which splits from the right at most once, correctly extracting the suffix (`.service`, `.device`, `.timer`, etc.) regardless of how many dots appear in the unit name.

**Verification:**
```python
# Old behavior (crashes):
>>> "sys-devices-pci0000:00-0000:00:1f.0.device".split(".")
["sys-devices-pci0000:00-0000:00:1f", "0", "device"]  # ValueError on unpack

# New behavior (works):
>>> "sys-devices-pci0000:00-0000:00:1f.0.device".rsplit(".", 1)
["sys-devices-pci0000:00-0000:00:1f.0", "device"]  # Correct
```

Normal service names (single dot) are unaffected:
```python
>>> "mock_bgp.service".rsplit(".", 1)
["mock_bgp", "service"]  # Same as before
```

## Test added
- `test_check_unit_status_multi_dot_unit_name`: verifies that `check_unit_status()` handles multi-dot unit names without crashing.

Fixes: #25291

### DIFF
--- a/src/system-health/health_checker/sysmonitor.py
+++ b/src/system-health/health_checker/sysmonitor.py
@@ -311,7 +311,7 @@ class Sysmonitor(ProcessTaskBase):
         try:
             service_status = "Down"
             service_up_status = "Down"
-            service_name,last_name = event.split('.')
+            service_name,last_name = event.rsplit('.', 1)
 
             sysctl_show = self.run_systemctl_show(event)
 
@@ -454,7 +454,7 @@ class Sysmonitor(ProcessTaskBase):
                 astate = "DOWN"
             self.publish_system_status(astate)
 
-            srv_name,last = event.split('.')
+            srv_name,last = event.rsplit('.', 1)
             # stop on service maybe propagated to timers and in that case,
             # the state_db entry for the service should not be deleted
             if last == "service":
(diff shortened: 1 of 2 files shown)


rustiqly:
## Verified on sonic-vs VM

**Environment:** SONiC master (fcf18f7e0), Python 3.13.5, Debian 13.3

**Bug reproduction (unpatched):**
```
>>> sysmon.check_unit_status("sys-devices-pci0000:00-0000:00:1f.0.device")
ValueError: too many values to unpack (expected 2)
```

**Fix verification (patched `rsplit(".", 1)`):**
```
PASS  multi-dot device   sys-devices-pci0000:00-0000:00:1f.0.device
PASS  multi-dot mount    dev-disk-by\x2did-wwn\x2d0x5001.mount
PASS  single-dot mount   run-user-1000.mount
PASS  normal service     mock_bgp.service
PASS  timer unit         mock_snmp.timer
```

PR #25447, opened 2026-02-11
[Arista][DNX] Workaround for sai.profile being empty when syncd starts

arista-nwolfe:
Original PR https://github.com/Azure/sonic-buildimage-msft/pull/1770 went only to msft-202405 but the issue with `start.sh` running twice hasn't been fixed https://github.com/sonic-net/sonic-buildimage/issues/24399 so creating a PR for master and hopefully cast to 202511.

We're seeing an issue where start.sh is run multiple times on syncd startup and if the 'cp' operation in the 2nd start.sh occurs while the Syncd process is reading the /etc/sai.d/sai.profile it will end up reading an empty file.

Therefor if we skip the 'cp' operation when the file is unchanged the 2nd start.sh run should always be a no-op.

This is a workaround until we find the reason and solution to stop start.sh from running multiple times.

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [x] 202511

### DIFF
--- a/platform/broadcom/docker-syncd-brcm-dnx/start.sh
+++ b/platform/broadcom/docker-syncd-brcm-dnx/start.sh
@@ -26,6 +26,16 @@ else
         sonic-cfggen -d -t $HWSKU_DIR/config.bcm.j2 > /etc/sai.d/config.bcm
     fi
     if [ -f $HWSKU_DIR/sai.profile ]; then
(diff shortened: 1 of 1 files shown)


arlakshm:
@rlhui, can you please help merge this PR

PR #25463, opened 2026-02-12
config-setup: Prefer minigraph.xml over ZTP when config_db.json is absent

yxieca:
### Description of PR

**Summary:**
When `config_db.json` is missing but `minigraph.xml` is present, the config-setup boot sequence previously triggered ZTP (if enabled), which generates a minimal config and runs `config reload`. This removes the device management IP, requiring console access to recover.

Additionally, during warm boot the config initialization path had no guard — if `config_db.json` was absent, ZTP could still be triggered even though warm boot must preserve the existing running configuration.

**Root cause:**
1. `do_config_initialization()` had no awareness of `minigraph.xml`. It only checked for ZTP or factory default, even when a valid minigraph was available on disk.
2. `check_system_warm_boot()` only checked STATE_DB, which may not be available early in the boot sequence. The canonical `SONIC_BOOT_TYPE=warm` in `/proc/cmdline` (used by all other boot-type detection in the codebase) was not checked.
3. `boot_config()` did not skip config initialization during warm boot, allowing ZTP to trigger inappropriately.

**Fix (3 changes in `config-setup`):**

1. **`do_config_initialization()`**: Check for `minigraph.xml` at the top of the function, before ZTP/factory-default logic. If minigraph is available, use `reload_minigraph` and return early. This aligns with the pattern already used in `do_config_migration()`.

2. **`check_system_warm_boot()`**: Enhanced to check `/proc/cmdline` for `SONIC_BOOT_TYPE=warm` first (the authoritative source set by warm-reboot scripts), then fall back to STATE_DB for compatibility. This is consistent with `getBootType()` used in `docker_image_ctl.j2`, `syncd_common.sh`, and `watchdog-control.sh`.

3. **`boot_config()`**: Added warm boot guard after config migration — during warm boot, skip config initialization and ZTP entirely. Also added `minigraph.xml` guard on the ZTP restart block so ZTP erase/restart is skipped when minigraph was used.

This maintains the config priority order consistent with `do_config_migration()`:
`config_db.json` > `minigraph.xml` > ZTP > factory default.

**Addresses:** ADO 36697420 — `[202511.08] Config Reload is Run during warm-boot up`

### Type of change

- [x] Bug fix
- [ ] Configuration change (update something in `files/` or `device/`)

### Back port request
- [ ] 202505
- [x] 202511

With ZTP enabled on 202511 images, upgrade path tests that delete `config_db.json` trigger ZTP's fallback behavior. ZTP generates a new config and runs `config reload`, wiping management IP. The device then requires console access to recover. During warm boot this is especially dangerous as it disrupts the warm restart flow.

#### How did you do it?
1. Added `minigraph.xml` existence check in `do_config_initialization()` before ZTP/factory-default logic
2. Enhanced `check_system_warm_boot()` to check `/proc/cmdline` for `SONIC_BOOT_TYPE=warm` (consistent with all other boot-type detection in the codebase)
3. Added warm boot guard in `boot_config()` to skip config initialization and ZTP during warm boot
4. Added `minigraph.xml` guard on ZTP restart block in `boot_config()` to avoid unnecessary ZTP erase when minigraph was applied

Code review and trace of all boot code paths in `config-setup`. The minigraph fix follows the same pattern already used in `do_config_migration()` which correctly prefers minigraph over factory default. The warm boot detection follows the same `/proc/cmdline` pattern used by `docker_image_ctl.j2`, `syncd_common.sh`, and `watchdog-control.sh`.

### DIFF
--- a/files/image_config/config-setup/config-setup
+++ b/files/image_config/config-setup/config-setup
@@ -180,7 +180,7 @@ check_system_warm_boot()
 
     # Try STATE_DB first — authoritative source that reflects current
     # warm boot state and is cleared after warm boot completes.
-    SYSTEM_WARM_START=$(sonic-db-cli STATE_DB hget "WARM_RESTART_ENABLE_TABLE|system" enable 2>/dev/null)
+    SYSTEM_WARM_START=$(sonic-db-cli STATE_DB hget "WARM_RESTART_ENABLE_TABLE|system" enable)
     DB_RC=$?
 
     if [ $DB_RC -eq 0 ]; then


files/image_config/config-setup/config-setup
```diff
@@ -256,11 +268,24 @@ generate_config()
...
+#  - If ZTP is enabled and no minigraph, ZTP configuration is created
+#  - If ZTP is disabled and no minigraph, factory default configuration
 #    is created
 do_config_initialization()
 {
+    # If minigraph.xml is available, prefer it over ZTP/factory default.
+    # This avoids ZTP triggering a config reload that removes management IP
+    # when the device has a valid minigraph but no config_db.json
+    # (e.g., during upgrade path tests).
+    if [ -r ${MINGRAPH_FILE} ]; then
```
saiarcot895, line 291: In what code path/code flow does this get triggered? Shouldn't this code be in a place where the minigraph is loaded into the system?

yxieca:
@saiarcot895 Great question. `do_config_initialization()` is called from `boot_config()` in two code paths:

1. When `/tmp/pending_config_initialization` or the `CONFIG_SETUP_INITIALIZATION_FLAG` exists (first-time boot after install)
2. When `config_db.json` does not exist at all (e.g., upgrade path tests that deliberately delete it)

In both cases, without this minigraph check, the function falls through to the ZTP/factory-default logic: ZTP generates a minimal config via `ztp-profile.sh` and then runs `config reload`, which wipes the management IP configured by minigraph — requiring console recovery.

This IS the right place for the check because `do_config_initialization()` is the decision point for "config_db.json is missing, what do we do?". The priority order should be:
- `config_db.json` exists -> handled before we even enter this function
- `minigraph.xml` exists -> use `reload_minigraph` (this new check)
- ZTP enabled -> generate ZTP config (existing behavior)
- Otherwise -> generate factory default (existing behavior)

Note that `do_config_migration()` (the image-upgrade path) already has this exact pattern — it checks for minigraph.xml as a fallback when config_db.json is missing (line 404). This change aligns `do_config_initialization()` with the same logic.

saiarcot895:
On images where ZTP is disabled, and a `minigraph.xml` file is already present, the `minigraph.xml` file is used, and a factory-default configuration is not generated. If that was already working without ZTP, then that suggests that this new `reload_minigraph` check is not required for the flow you described.

yxieca:
@saiarcot895 Good point — let me clarify.

When ZTP is **disabled** and `config_db.json` is missing, the existing code takes the factory default path (line 289-293):
```
generate_config factory CONFIG_DB_JSON
reload_configdb CONFIG_DB_JSON
```
This generates a factory default `config_db.json` and loads it via `config reload`. It does **not** use minigraph.xml — the device gets a factory default config, losing whatever was configured via minigraph. So even without ZTP, the device ends up with a factory default config rather than its minigraph-based config.

However, you're right that the **critical** bug is with ZTP **enabled** — that's the scenario from ADO 36697420. When ZTP is enabled, `ztp-profile.sh` creates a minimal ZTP config and `config reload` wipes the management IP, making the device unreachable.

The minigraph check addresses both cases:
1. **ZTP enabled** (the reported bug): Prevents ZTP from generating a config and running `config reload` that wipes mgmt IP
2. **ZTP disabled**: Ensures the device uses its minigraph config instead of falling back to a factory default

Both are consistent with the pattern in `do_config_migration()` (line 404) which already prefers minigraph over factory default when `config_db.json` is missing during image upgrades.

Would you prefer we scope this fix to only the ZTP-enabled case (e.g., guard just the ZTP block at line 295)?

yxieca:
### Suggested Test Plan

**3 code changes, 4 test scenarios:**

#### Scenario A — Primary bug fix (ZTP enabled + minigraph present)
1. Build image with this change, ZTP enabled
2. Place `minigraph.xml` on device, delete `config_db.json`
3. Cold boot
4. **Expected:** Device uses `reload_minigraph`, management IP preserved, no ZTP config reload

#### Scenario B — Regression check (ZTP enabled, no minigraph)
1. ZTP enabled, no `config_db.json`, no `minigraph.xml`
2. Cold boot
3. **Expected:** ZTP triggers as before (existing behavior preserved)

#### Scenario C — Factory default regression (ZTP disabled + minigraph present)
1. ZTP disabled, `minigraph.xml` present, no `config_db.json`
2. Cold boot
3. **Expected:** Device uses `reload_minigraph` instead of factory default

#### Scenario D — Warm boot guard
1. ZTP enabled, `minigraph.xml` present, no `config_db.json`
2. Warm boot (`SONIC_BOOT_TYPE=warm` in `/proc/cmdline`)
3. **Expected:** Config initialization and ZTP skipped entirely, existing running config preserved

#### Verification commands:
- `show runningconfiguration all` — verify config matches minigraph expectations
- `show ip interface` — verify management IP is intact
- `ztp status` — verify ZTP was not triggered (Scenarios A, C, D)
- `grep "minigraph" /var/log/syslog` — confirm `reload_minigraph` was called (Scenarios A, C)
- `grep "Warm boot detected" /var/log/syslog` — confirm warm boot guard (Scenario D)

#### Upgrade path test (end-to-end):
Run the existing upgrade path test suite (which deletes `config_db.json`) on a ZTP-enabled 202511 image. The device should remain reachable via management IP throughout the test — this was the original failure mode reported in ADO 36697420.

yxieca:
### Complete `config-setup boot` workflow after this change

```
boot_config() called
│
├─ 1. check_system_warm_boot()
│     ├─ Check /proc/cmdline for SONIC_BOOT_TYPE=warm  ← NEW
│     │   └─ If found → WARM_BOOT="true", return
│     └─ Fall back: check STATE_DB WARM_RESTART_ENABLE_TABLE
│         └─ If enable=true → WARM_BOOT="true"
│
├─ 2. Pending config migration?
│     (/tmp/pending_config_migration or CONFIG_SETUP_POST_MIGRATION_FLAG)
│     └─ Yes → do_config_migration()
│           ├─ Copy files from old_config (minigraph, config_db, frr, etc.)
│           ├─ Run post-migration hooks
│           ├─ WARM_BOOT=true? → do_db_migration, exit 0 (script ends)
│           ├─ config_db.json present? → reload_configdb, do_db_migration
│           ├─ minigraph.xml present? → reload_minigraph, do_db_migration
│           └─ Neither → log warning
│
├─ 3. WARM_BOOT=true? → "Warm boot detected, skipping", return  ← NEW
│     (No config initialization or ZTP during warm boot)
│
├─ 4. Multi-ASIC (NUM_ASIC > 1)? → return
│     (Config initialization not supported for multi-NPU)
│
├─ 5. Pending config initialization?
│     (/tmp/pending_config_initialization or CONFIG_SETUP_INITIALIZATION_FLAG)
│     └─ Yes → do_config_initialization()
│
├─ 6. config_db.json missing?
│     └─ Yes → do_config_initialization()
│           │
│           │  do_config_initialization():
│           │  ├─ minigraph.xml present?  ← NEW
│           │  │   └─ Yes → reload_minigraph, set CONFIG_DB_INITIALIZED, return
│           │  ├─ ZTP disabled?
│           │  │   └─ generate factory default, reload_configdb
│           │  └─ ZTP enabled?
│           │      └─ generate ZTP config, reload_configdb
│           │
│           └─ After do_config_initialization():
│               └─ minigraph.xml absent AND ZTP enabled?  ← UPDATED guard
│                   ├─ ZTP SUCCESS/FAILED → ztp erase -y
│                   └─ Otherwise → touch /tmp/pending_ztp_restart
│
└─ Done
```

**Key changes (marked with ← NEW / ← UPDATED):**

1. **`check_system_warm_boot()`** — now checks `/proc/cmdline` for `SONIC_BOOT_TYPE=warm` first, consistent with `getBootType()` used in `docker_image_ctl.j2`, `syncd_common.sh`, and `watchdog-control.sh`. Falls back to STATE_DB for compatibility.
2. **Step 3 warm boot guard** — during warm boot, skip all config initialization and ZTP entirely. Existing running config must be preserved.
3. **`do_config_initialization()` minigraph check** — if `minigraph.xml` is present, use `reload_minigraph` and return early. Aligns with the pattern already in `do_config_migration()`.
4. **ZTP restart block** — guarded with `[ ! -e minigraph.xml ]` so ZTP erase/restart is not triggered when minigraph was used.

**Config priority order (consistent across all paths):**
`config_db.json` → `minigraph.xml` → ZTP → factory default

yxieca:
@saiarcot895 Thanks for pushing on this — I want to clarify an important distinction between two code paths.

You're right that when `minigraph.xml` is present, it **is** used during **image upgrades**. That's because the upgrade path goes through `do_config_migration()` (triggered by `pending_config_migration`), which already checks for minigraph as a fallback at line 404:

```bash
elif [ -r ${MINGRAPH_FILE} ]; then
    echo "Use minigraph.xml from old system..."
    reload_minigraph
```

So in the normal upgrade scenario, minigraph works correctly — that path was already fine.

However, `do_config_initialization()` is a **different code path** that is called when:
1. A `pending_config_initialization` flag exists (first-time boot after install)
2. `config_db.json` simply doesn't exist and there's no pending migration

In the **original** `do_config_initialization()`, there is no minigraph check at all:

```bash
# Original code
do_config_initialization()
{
    if ! ztp_is_enabled ; then
        generate_config factory ${CONFIG_DB_JSON}
        reload_configdb ${CONFIG_DB_JSON}      # factory default, ignores minigraph
    fi
    if ztp_is_enabled ; then
        generate_config ztp ...               # ZTP config, ignores minigraph
        reload_configdb ...
    fi
}
```

The reported bug (ADO 36697420) hits this initialization path specifically: the upgrade path tests delete `config_db.json` **without** setting the `pending_config_migration` flag, so `do_config_migration()` is never called. Instead, `boot_config()` sees no `config_db.json` and calls `do_config_initialization()`, which falls through to ZTP (or factory default), ignoring the minigraph.

So to summarize:
- **`do_config_migration()`** (upgrade path) — already handles minigraph ✅
- **`do_config_initialization()`** (no migration, no config_db) — did NOT handle minigraph ❌ ← this PR fixes it

The fix aligns `do_config_initialization()` with the same priority order that `do_config_migration()` already uses.

Ryangwaite (approved):

files/image_config/config-setup/config-setup
```diff
@@ -170,15 +170,27 @@ copy_config_files_and_directories()
...
+    WARM_BOOT="false"
+
+    # Check /proc/cmdline first — this is the authoritative source set by
+    # warm-reboot scripts and is available before database services start.
+    case "$(cat /proc/cmdline)" in
+    *SONIC_BOOT_TYPE=warm*)
+        WARM_BOOT="true"
+        return
+        ;;
+    esac
```
saiarcot895: This might cause issues if the device has been warm-rebooted, but we are reloading everything with a new config/minigraph.

yxieca: Good catch. You're right — \/proc/cmdline\ retains \SONIC_BOOT_TYPE=warm\ for the entire boot session, so if \config-setup\ is invoked again after warm boot completes (e.g., config reload or load_minigraph), the stale cmdline would incorrectly skip config operations.

I've reworked \check_system_warm_boot()\ to prefer STATE_DB as the authoritative source — it reflects the true warm boot state and gets cleared after warm boot completes. \/proc/cmdline\ is now only consulted as a fallback during the \oot\ command, when database services may not be up yet. See commit 9b4cf89.

files/image_config/config-setup/config-setup
```diff
@@ -170,15 +170,37 @@ copy_config_files_and_directories()
...
-    # SYSTEM_WARM_START could be empty, always make WARM_BOOT meaningful.
-    if [[ x"$SYSTEM_WARM_START" == x"true" ]]; then
-        WARM_BOOT="true"
-    else
-        WARM_BOOT="false"
+    WARM_BOOT="false"
+
+    # Try STATE_DB first — authoritative source that reflects current
+    # warm boot state and is cleared after warm boot completes.
+    SYSTEM_WARM_START=$(sonic-db-cli STATE_DB hget "WARM_RESTART_ENABLE_TABLE|system" enable 2>/dev/null)
```
saiarcot895: If there is something printed on stderr by this command, it might be better for later debugging to let it be printed onto the console and syslogs.

yxieca: Good point - removed the 2>/dev/null in commit dab44ba. Now stderr from sonic-db-cli flows through to console/syslog for debugging. We still use the return code to detect DB unavailability and fall back to /proc/cmdline.

saiarcot895 (approved):

PR #25489, opened 2026-02-13
[build] Fix libyang-dev conflict during high-parallelism builds

### LINKED ISSUE #25425: Bug: libyang3-dev and libyang1-dev cannot coexist
### Is it platform specific

generic

### Importance or Severity

High

### Description of the bug

libyang1-dev and libyang3-dev cannot coexist, so build ordering needs to be strictly enforced to ensure the right libyang dev package is installed for what is being built.

As of Trixie, during high parallel builds libyang3-dev will be installed then libswsscommon is built which expects libyang1 and the build fails.

### Steps to Reproduce

Build with parallelism of 16 and notice the build failure.  Re-running make after the failure usually continues due to it recreating the slave and only installing what it needs to continue the build.

It should order builds appropriately.

### Relevant log output

```shell

```

```shell

```

### Attach files (if any)

_No response_

rustiqly:
#### Why I did it

Fixes #25425 — `libyang3-dev` and `libyang1-dev` cannot coexist because both provide the dpkg package `libyang-dev`. With high build parallelism (JOBS≥16), FRR (needs `libyang3-dev`) and packages like `libswsscommon` (needs `libyang1-dev` via `sonic-mgmt-common`) can build concurrently, causing one to replace the other's headers mid-compilation.

The existing `_AFTER` ordering constraint between `LIBYANG` and `FRR` was only applied to the bookworm build environment. In trixie (and any future build environments), no ordering existed, allowing the race condition.

#### How I did it

Removed the `ifeq ($(BLDENV),bookworm)` guard around `$(LIBYANG)_AFTER = $(FRR)` in `rules/libyang.mk` so the constraint applies unconditionally to all build environments.

#### How to verify it

1. Build with high parallelism targeting trixie: `make SONIC_BUILD_JOBS=16 target/sonic-vs.img.gz`
2. Verify no build failures related to libyang header mismatches
3. Verify libyang1 packages are built after FRR completes in both bookworm and trixie phases

202511


bhouse-nexthop:
You didn't like #25427 ?  Any commentary?

rustiqly:
Closing in favor of #25427 which takes a more robust approach — it introduces a phony target gate that ensures all libyang3-dev consumers (FRR + libyang3-py3) are fully installed before libyang1 can build. That handles trixie correctly too (where FRR isn't built). Our one-liner `_AFTER` fix works on bookworm but doesn't properly cover trixie.

@bhouse-nexthop nice work on the phony target mechanism — it's a clean, reusable pattern for future ordering conflicts like this.

PR #25530, opened 2026-02-16
Fix: Skip git commands in container to avoid spurious errors

### LINKED ISSUE #25529: Spurious 'fatal: not a git repository' errors in build logs
## Description
The build log contains repeated 'fatal: not a git repository' errors that appear harmless but clutter the output and may confuse users about build status.

## Error Message
```
```

## Root Cause
In `functions.sh` lines 54-66, the `sonic_get_version()` function runs multiple git commands to determine the version:
```bash
sonic_get_version() {
    local describe=$(git describe --tags 2>/dev/null)
    local latest_tag=$(git describe --tags --abbrev=0 2>/dev/null)
    local branch_name=$(git rev-parse --abbrev-ref HEAD)
    if [ -n "$(git status --untracked-files=no -s --ignore-submodules)" ]; then
        local dirty="-dirty-$BUILD_TIMESTAMP"
    fi
    ...
    echo "${branch_name}.${BUILD_NUMBER}${dirty:--$(git rev-parse --short HEAD)}" | sed 's/\\//_/g'
}
```

When these commands run **inside the docker slave container** during `make all`, the current directory is `/sonic/` which is:
- A bind-mounted directory from the host (not a git repo)
- Does NOT have a `.git` directory
- Git fails to find repository metadata

The error is suppressed by `2>/dev/null` but still captured by `script` logging, appearing in the build log.

## Impact
- **Severity:** Cosmetic (build succeeds regardless)
- **Frequency:** Occurs at start of bookworm and trixie stages
- **User impact:** Confusing error messages, false alarm for monitoring tools
- **Build effect:** None (version fallback works correctly)

## Solution Options
1. **Detect container environment:** Skip git commands if running inside docker, use fallback version
2. **Mount .git into container:** Bind-mount `/data/sonic-buildimage/.git` so git can work
3. **Improve error suppression:** Redirect stderr before git runs instead of relying on 2>/dev/null
4. **Pass version as argument:** Build system could pass pre-computed version to container

## Reproduction
Run any build:
```bash
make SONIC_BUILD_JOBS=4 all
```

Check log for errors:
```bash
grep "fatal: not a git" /tmp/sonic_build_*.log
```

## Build Log Examples
Build logs show these errors appearing after build configuration is printed, at lines ~104 and ~317 in typical bookworm+trixie builds.

securely1g:
Fixes #25529

When sonic_get_version() runs inside the docker slave container, the current directory is /sonic/ which is a bind-mounted directory without a .git folder. This causes spurious 'fatal: not a git repository' errors that clutter build logs and confuse users, even though the build succeeds.

This fix detects if we're running in a container (no .git present) and uses a simple fallback version string instead of running git commands.

### DIFF
--- a/functions.sh
+++ b/functions.sh
@@ -51,6 +51,15 @@ docker_try_rmi() {
 }
 
 sonic_get_version() {
+    # Detect if running in container (no .git in current dir)
+    # Skip git commands inside docker to avoid spurious "fatal: not a git repository" errors
+    if [ ! -d ".git" ]; then
+        # Container fallback: use BUILD_NUMBER as version
+        BUILD_NUMBER=${BUILD_NUMBER:-0}
+        echo "sonic.${BUILD_NUMBER}-container"
+        return 0
+    fi
+
     local describe=$(git describe --tags 2>/dev/null)
     local latest_tag=$(git describe --tags --abbrev=0 2>/dev/null)
     local branch_name=$(git rev-parse --abbrev-ref HEAD)


functions.sh
```diff
@@ -51,6 +51,15 @@ docker_try_rmi() {
 }
 
 sonic_get_version() {
```
lguohan, line 53: where does it get called?

securely1g: `sonic_get_version()` is sourced and called from `slave.mk`:
```
override SONIC_IMAGE_VERSION := $(shell export BUILD_TIMESTAMP=$(BUILD_TIMESTAMP) && export BUILD_NUMBER=$(BUILD_NUMBER) && . functions.sh && sonic_get_version)
```
The issue is that `sonic_get_version` runs `git describe` which fails inside Docker containers where `.git` is not available, producing spurious "fatal: not a git repository" errors in the build log.

securely1g:
Re-authored the commit under my CLA-linked account (`securely1g`) to resolve the EasyCLA "Missing ID on Commit" failure — the previous commit's author email was not linked to a GitHub account with a signed CLA. The code change is unchanged (identical tree); only the commit author and Signed-off-by were corrected. EasyCLA should pass on the re-run.

PR #25557, opened 2026-02-18
Fix dhcp_dos_logger.py to wait for PortInitDone before starting

prajjwal-arista:
Fix dhcp_dos_logger.py to wait for PortInitDone before running tc commands on any interfaces.

#### Why I did it
Addressing the flaky issue mentioned in [Issue](https://github.com/sonic-net/sonic-buildimage/issues/25556)

When config reload happens, dhcp_dos_logger service is restarted and it keeps executing tc command for various interfaces.
During this time, the ports are being removed and reinitialized which causes a the tc commands to fail sometimes when the ports are not in usable states.
The expectation is that dhcp_dos_logger should wait for portInitDone and only then execute tc commands.

#### How I did it
Introduced a wait function in dhcp_dos_logger.py to wait for PortInitDone notification.

#### How to verify it
Can be verified by multiple rounds of config reload and we must not see the following error - 
`ERR dhcp_dos_logger.py: TC command failed for Ethernet\d*: Cannot find device \\"Ethernet\d*\\"`

#### Testing
Manual - ran reload config in repetition to find out no error was seen from dhcp logger.

fix testing - 
Tested this in these two scenarios -
1) When port init phase is happening -
Steps -> do config reload, and see the wait time for the PortInitDone
notification in dhcp dos logger logs.
(Here wait time was found to be around 40 seconds)
logs-
```
Feb 17 03:46:55 ld595 dhcp_dos_logger.py[39636]: Waiting for PortInitDone...
Feb 17 03:47:37 ld595 dhcp_dos_logger.py[39636]: PortInitDone received
```

2) When port init is already done -
Steps -> when system is stable just restart dhcp_dos_logger service, and
see the wait time for the PortInitDone notification in dhcp dos logger logs.
(Here wait time was found to be around 0.1 seconds)
logs-
```
Feb 17 03:48:33 ld595 dhcp_dos_logger.py[46357]: Waiting for PortInitDone...
Feb 17 03:48:33 ld595 dhcp_dos_logger.py[46357]: PortInitDone received
```

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [x] 202511

### DIFF
--- a/files/image_config/dhcp_dos_logger/dhcp_dos_logger.py
+++ b/files/image_config/dhcp_dos_logger/dhcp_dos_logger.py
@@ -56,8 +56,9 @@ def wait_for_port_init_done():
     Wait for PortInitDone event from APP_DB.
 
     Returns:
-        None (blocks until PortInitDone is received)
+        None (blocks until PortInitDone is received or timeout occurs)
     """
+    MAX_WAIT_SECONDS = 300
     appl_db = daemon_base.db_connect("APPL_DB")
 
     sel = swsscommon.Select()
@@ -65,8 +66,14 @@ def wait_for_port_init_done():
     sel.addSelectable(sst)
 
     logger.log_info("Waiting for PortInitDone...")
+    start_time = time.time()
     while True:
         (state, _) = sel.select(1000)
+        elapsed = time.time() - start_time
+
+        if elapsed >= MAX_WAIT_SECONDS:
+            logger.log_warning("Timed out waiting for PortInitDone, proceeding anyway")
+            return
 
         if state == swsscommon.Select.TIMEOUT:
             continue
@@ -75,7 +82,7 @@ def wait_for_port_init_done():
             continue
 
         while True:
-            (key, op, fvp) = sst.pop()
+            (key, _, _) = sst.pop()
             if not key:
                 break
 


StormLiangMS:
@prajjwal-arista 

## Code Review

The fix correctly addresses the root cause: `dhcp_dos_logger` runs `tc` commands on ports that haven't been fully re-initialized after config reload. Waiting for `PortInitDone` from `APPL_DB:PORT_TABLE` before entering the `handler()` loop is the right approach, and is already a well-established pattern in SONiC (e.g., `ports_xcvrd_notify.py`).

---

### :red_circle: Issue: No Maximum Wait Timeout — Risk of Silent Indefinite Block

`wait_for_port_init_done()` can block forever. If `PortInitDone` never arrives (e.g., `portmgrd`/`orchagent` crash, system is in a degraded state), `dhcp_dos_logger` will be silently stuck before ever entering `handler()`, providing zero monitoring. There's no fallback, no max-wait timeout, and no escalation.

Consider adding a max timeout with a warning to allow the daemon to proceed anyway:
`python
MAX_WAIT_SECONDS = 300
waited = 0
while True:
    (state, _) = sel.select(1000)
    waited += 1
    if waited >= MAX_WAIT_SECONDS:
        logger.log_warning("Timed out waiting for PortInitDone, proceeding anyway")
        return
    ...
`

---

### :yellow_circle: Minor: Inner Drain Loop Could Miss `PortInitDone` on Error Path

In the inner `while True` loop, if `sst.pop()` raises an exception (e.g., Redis connection loss), it's not caught and will propagate uncaught, crashing the daemon. The outer `handler()` loop has a `try/except`, but `wait_for_port_init_done()` does not. Worth adding:
`python
try:
    (key, op, fvp) = sst.pop()
except Exception as e:
    logger.log_error(f"Error reading from SubscriberStateTable: {e}")
    break
`

---

### :yellow_circle: Minor: Unused Variables `op` and `fvp`

In the inner loop:
`python
(key, op, fvp) = sst.pop()
`
`op` and `fvp` are never used. Use `_` to signal intent:
`python
(key, _, _) = sst.pop()
`

---

### :white_check_mark: What Looks Good

- **Core logic is correct** and mirrors the established SONiC pattern in `ports_xcvrd_notify.py` exactly. The `SubscriberStateTable` approach correctly handles both cases: waiting for a future `PortInitDone` (during reload) and detecting an already-present `PortInitDone` entry (on service restart), which explains the ~0.1s wait observed in testing.
- **Inner drain loop** (`while True: sst.pop()`) is actually an improvement over the reference — it drains multiple queued events per outer select iteration, ensuring `PortInitDone` isn't missed if other PORT_TABLE entries arrive first.
- **Import refactoring** from `swsscommon.swsscommon.ConfigDBConnector` → `swsscommon.ConfigDBConnector()` is clean and consistent.
- **Testing evidence** is well-documented in the PR description with actual log timestamps.

---

**Overall**: The approach is solid and the core logic is correct. The main concern is the **lack of a timeout guard** in `wait_for_port_init_done()` that could leave the daemon silently blocked in a degraded system state.

prajjwal-arista:
@StormLiangMS 
I've addressed your comment, please review again.

prajjwal-arista:
Please help re-run the failed jobs since the pipeline failures doesn't seem to be related to my changes.

The [failure](https://elastictest.org/scheduler/publictestplan/69a1b9e08fe9299bbb5c0295?testcase=vxlan%2Ftest_vxlan_ecmp.py&type=console) is in vxlan/test_vxlan_ecmp.py::test_vxlan_single_endpoint[v4_in_v4], which is testing VXLAN packet encapsulation and forwarding.
```
  RuntimeError: Didnot get any reply for this destination:150.0.3.1 Its active endpoints:['100.0.1.10']
probably a network connectivity issue.
```

seems to be some connectivity issue

prajjwal-arista:
@StormLiangMS 
Please help by looking into this the job failures are due to following error, which is not related to my changes.
`RuntimeError: Didnot get any reply for this destination:150.0.3.1 Its active endpoints:['100.0.1.10']
`
[Azure.sonic-buildimage (Test kvmtest-t1-lag-vpp by Elastictest [OPTIONAL])](https://github.com/sonic-net/sonic-buildimage/pull/25557/checks?check_run_id=65164787201)

Can you please re-run the failing jobs again?

StormLiangMS:
Friendly ping — this PR needs review for 202511 qualification. cc @lguohan

prajjwal-arista:
@StormLiangMS 
The failed checks are not due to my changes , but memory checkers. Please help with this.

```
        # Check if we stored any memory errors for this test
        if request.node.nodeid in _memory_errors_by_test:
            errors = _memory_errors_by_test.pop(request.node.nodeid)
            if errors:
                failure_message = "\n".join(errors)
                logger.error(f"Memory errors detected in fixture teardown: {failure_message}")
>               pytest.fail(failure_message)
E               Failed: [ALARM]: free:used memory usage increased by 663.0%, exceeds increase threshold 20%% (previous: 2910.0 MB, current: 3573.0 MB)
E               [ALARM]: docker:swss, Current memory usage 9.4% exceeds high threshold 8% (previous: 2.5%, current: 9.4%)
E               [ALARM]: docker:swss memory usage increased by 6.9%, exceeds increase threshold 3% (previous: 2.5%, current: 9.4%)
```

StormLiangMS:
hi @Xichen96 @yxieca  could you help to review?

yxieca:
close and reopen to reset tests.

prajjwal-arista:
@StormLiangMS 
Can you please add label to cherry pick this into 202511 as well.

PR #25571, opened 2026-02-19
Yang: COPP_TRAP_EXCLUDE_PORTS model

### LINKED ISSUE #25570: Enhancement: SONiC cannot be used as a fanout in a testbed topology
### Is it platform specific

generic

### Importance or Severity

Critical

When using SONiC as a fanout in a testbed topology, the CoPP traps are configured on all ports in the system.  This will prevent packets for LACP or LLDP, for instance, from flowing to the testserver through the fanout to the DUT.

A common hack is to use the bcm shell (on broadcom-based switches) to remove the traps manually.  This is clearly a hack and brittle as it doesn't survive any reconfiguration of the switch.

### Current Behavior

Traps are configured on all ports of the switch.

### Proposed Behavior

Have the ability to exclude ports from having traps installed via `SAI_HOSTIF_TRAP_ATTR_EXCLUDE_PORT_LIST`.  We wouldn't want a toggle to remove traps system-wide as there are likely some ports where traps would be desirable to remain enabled (such as if a VXLAN fabric is used in the testbed network)

bhouse-nexthop:
#### Why I did it

Exclude ports from CoPP Traps.  In some cases, such as sonic-mgmt testbeds, LACP and LLDP packets may be trapped
instead of being relayed to the desired endpoint.  This allows configuration to exclude ports from traps to allow
these packets to flow all the way to the desired endpoint.  By default all ports on a switch will have the standard
traps applied.

Implementation in https://github.com/sonic-net/sonic-swss/pull/4234

HLD update: https://github.com/sonic-net/SONiC/pull/2229

#### How I did it

This is just a basic Yang model and validation of the Yang model.  The real implementation is in sonic-swss.

#### How to verify it

Run tests included in this PR.

- [ ] 202505
- [ ] 202511

master as of 2026-01-09

Yang model implementation and tests for CoPP trap exclude ports

https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-yang-models/doc/Configuration.md#copp_trap_exclude_ports

Fixes #25570

### DIFF
--- a/src/sonic-yang-models/yang-models/sonic-copp.yang
+++ b/src/sonic-yang-models/yang-models/sonic-copp.yang
@@ -9,11 +9,18 @@ module sonic-copp {
 		prefix stypes;
 	}
 
+	import sonic-port {
+		prefix port;
+	}
+
 	description "CoPP YANG Module for SONiC OS";
 
 	revision 2021-03-31 {
-		description
-			"First Revision";
+		description "First Revision";
+	}
+
+	revision 2026-01-09 {
+		description "Add COPP_TRAP_EXCLUDE_PORTS";
 	}
 
 	container sonic-copp {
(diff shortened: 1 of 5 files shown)


src/sonic-yang-models/yang-models/sonic-copp.yang
```diff
@@ -168,6 +175,21 @@ module sonic-copp {
 			/* end of list COPP_TRAP_LIST */
 		}
 		/* end of container COPP_TRAP */
+
+		container COPP_TRAP_EXCLUDE_PORTS {
```
rminnikanti, line 179: @bhouse-nexthop , setting SAI_HOSTIF_TRAP_ATTR_EXCLUDE_PORT_LIST is valid only when the trap_action is trap or copy. We can't enforce it from Yang as this is a global configuration. Is my understanding correct?

bhouse-nexthop: That would be true, if someone is directly modifying copp_cfg.j2 we'd have no way to know that.

PR #25576, opened 2026-02-19
[bash]: Preserve shell history across warm-reboot

### LINKED ISSUE #25519: Bug: warm-reboot looses user-history
### Is it platform specific

generic

### Importance or Severity

Medium

### Description of the bug

after warm-reboot applied from Serial-CONSOLE the user's history (in CONSOLE) is empty

NOTE: May be related to the reported https://github.com/sonic-net/sonic-buildimage/issues/24931

### Steps to Reproduce

do some shell-commands
reboot
see the commands are in the history buffer
apply warm-reboot
========> The history buffer is completely empty

Expected: history buffer with previous shell-commands
Obtained: history buffer is completely empty

### Relevant log output

```shell

```

```shell

```

### Attach files (if any)

_No response_

Bojun-Feng:
#### Why I did it

Fix #25519 

During warm-reboot, `kexec` immediately replaces the kernel without sending termination signals to processes. Bash shells are terminated before flushing their in-memory history buffer to disk, so bash history is empty after warm reboot.

#### How I did it

* Modified `files/image_config/bash/bash.bashrc` to persist history after each command
* Added `shopt -s histappend` to append history rather than overwrite on shell exit
* Added `PROMPT_COMMAND="history -a; ${PROMPT_COMMAND}"` to flush history to disk after every command execution
* This ensures history is written in real-time, surviving the abrupt kernel replacement during `kexec`

#### How to verify it

1. Boot a SONiC device and open a serial console session
2. Run several commands (e.g., `ls`, `show version`, `show interfaces status`)
3. Execute `warm-reboot`
4. After the device comes back up, log in and run `history`
5. Verify that the commands from before the warm-reboot are present

Fix warm-reboot losing user shell history by persisting bash history after each command


Bojun-Feng:
Note that the final command (e.g., `warm-reboot` itself) might not be saved. `PROMPT_COMMAND` runs after command completion but `kexec` immediately replaces the kernel before we get to `PROMPT_COMMAND`.
Still, this is a significant improvement over losing all history in buffer.

davidm-arista:
Is this unneccessary extra writes to storage?

Bojun-Feng:
That's fair, I didn't really think it through when making the PR. It would probably be a better idea to just write once at the beginning of the warm reboot command.

Will update PR tonight.

Bojun-Feng:
Unable to reproduce the issue using community SONiC VS image from pipeline build on 20260330 ([Artifact Link](https://sonic-build.azurewebsites.net/ui/sonic/pipelines/142/builds/1074536/artifacts?branchName=master), [Build Link](https://dev.azure.com/mssonic/build/_build/results?buildId=1074536&view=results)), from commit https://github.com/sonic-net/sonic-buildimage/commit/ab081eedf9ea32e769dbc537062ec39d7b40e72d.

Closing this down

PR #25577, opened 2026-02-19
[interfaces-config] retry restart networking

benle7:
#### Why I did it
The command "systemctl restart networking" uses "ifupdown" which has global file lock to prevent concurrent configurations. In some rare cases, we can get a timing conflict between simultaneous processes,
and eth0 remains down.
"error: Another instance of this program is already running"

##### Work item tracking

#### How I did it
Add a retry mechanism for the "systemctl restart networking" command.

#### How to verify it
sudo ip link set eth0 down
sudo flock -x /run/network/.lock sleep 7 &
sudo systemctl restart interfaces-config.service

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511

- [ ] 
- [ ]

### DIFF
--- a/files/image_config/interfaces/interfaces-config.sh
+++ b/files/image_config/interfaces/interfaces-config.sh
@@ -86,7 +86,23 @@ resolvconf_updates_restore
 # Read sysctl conf files again
 sysctl -p /etc/sysctl.d/90-dhcp6-systcl.conf
 
-systemctl restart networking
+MAX_RETRIES=5
+RETRY_DELAY=2
+for ((i=1; i<=MAX_RETRIES; i++)); do
+    LOG_MARK=$(date '+%Y-%m-%d %H:%M:%S')
+    if systemctl restart networking; then
+        if journalctl -u networking --since "$LOG_MARK" | grep -q "error.*already running"; then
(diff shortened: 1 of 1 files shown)


files/image_config/interfaces/interfaces-config.sh
```diff
@@ -86,7 +86,23 @@ resolvconf_updates_restore
 # Read sysctl conf files again
 sysctl -p /etc/sysctl.d/90-dhcp6-systcl.conf
 
-systemctl restart networking
+MAX_RETRIES=5
+RETRY_DELAY=2
+for ((i=1; i<=MAX_RETRIES; i++)); do
+    LOG_MARK=$(date '+%Y-%m-%d %H:%M:%S')
+    if systemctl restart networking; then
+        if journalctl -u networking --since "$LOG_MARK" | grep -q "error.*already running"; then
```
Junchao-Mellanox, line 94: So, even if "systemctl restart networking" return success, it could still hit the issue, right?

benle7: @Junchao-Mellanox  Correct

Junchao-Mellanox (approved):

PR #25600, opened 2026-02-20
Add redis db alias for easy accessing redis db

eddieruan-alibaba:
#### Why I did it

As we enable multi db in redis, it is not easy to use redis-cli to check redis for a targeted db. We need to check codes to get redis port number.

Create a set of aliases which would use swsscommon python script to receive db's port and id. This set of aliases would work for both cases. 

##### Work item tracking

#### How I did it
Use swsscommon python script to get used redis db port and id

#### How to verify it

admin@PE3:~$ alias
alias ls='ls --color=auto'
alias redis-appdb='redis-cli  -n 0 -p 6378'
alias redis-asicdb='redis-cli  -n 1 -p 6377'
alias redis-configdb='redis-cli  -n 4 -p 6379'
alias redis-counterdb='redis-cli  -n 2 -p 6376'
alias redis-flexcounterdb='redis-cli  -n 5 -p 6376'
alias redis-logleveldb='redis-cli  -n 3 -p 6379'
alias redis-statedb='redis-cli  -n 6 -p 6379'
alias redis-sysmondb='redis-cli  -n 10 -p 6379'
admin@PE3:~$ redis-appdb
127.0.0.1:6378> exit
admin@PE3:~$ redis-asicdb
127.0.0.1:6377[1]> exit
admin@PE3:~$ redis-configdb
127.0.0.1:6379[4]> exit
admin@PE3:~$ redis-counterdb
127.0.0.1:6376[2]> exit
admin@PE3:~$ redis-flexcounterdb
127.0.0.1:6376[5]> exit
admin@PE3:~$ redis-logleveldb

127.0.0.1:6379[3]> exit
admin@PE3:~$ redis-statedb
127.0.0.1:6379[6]> exit
admin@PE3:~$ redis-sysmondb
127.0.0.1:6379[10]> exit

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511

- [ ] 
- [ ]

### DIFF
--- a/files/image_config/bash/bash.bashrc
+++ b/files/image_config/bash/bash.bashrc
@@ -93,7 +93,7 @@ generate_sonic_redis_aliases() {
 
     # Run the external Python script, passing DB names as arguments
     local python_output
-    python_output=$(/usr/local/bin/sonic-db-aliases "${db_keys[@]}" 2>&1)
+    python_output=$(/usr/local/bin/sonic-db-aliases.py "${db_keys[@]}" 2>&1)
     local python_exit_code=$?
 
     # Check if Python command failed catastrophically (e.g., module not found)


files/image_config/bash/bash.bashrc
```diff
@@ -73,3 +73,13 @@ if [ -n "$SSH_TARGET_CONSOLE_LINE" ]; then
         exit
     fi
 fi
+
+# Aliases for redis-cli with correct database and port for each SONiC DB
+alias redis-appdb="redis-cli $(python -c 'import swsscommon.swsscommon; print(" -n "+str(swsscommon.swsscommon.SonicDBConfig.getDbId("APPL_DB"))+" -p "+str(swsscommon.swsscommon.SonicDBConfig.getDbPort("APPL_DB")))')"
```
asechoud: use python3 instead

files/image_config/bash/bash.bashrc
```diff
@@ -73,3 +73,13 @@ if [ -n "$SSH_TARGET_CONSOLE_LINE" ]; then
...
+alias redis-asicdb="redis-cli $(python -c 'import swsscommon.swsscommon; print(" -n "+str(swsscommon.swsscommon.SonicDBConfig.getDbId("ASIC_DB"))+" -p "+str(swsscommon.swsscommon.SonicDBConfig.getDbPort("ASIC_DB")))')"
+alias redis-counterdb="redis-cli $(python -c 'import swsscommon.swsscommon; print(" -n "+str(swsscommon.swsscommon.SonicDBConfig.getDbId("COUNTERS_DB"))+" -p "+str(swsscommon.swsscommon.SonicDBConfig.getDbPort("COUNTERS_DB")))')"
+alias redis-logleveldb="redis-cli $(python -c 'import swsscommon.swsscommon; print(" -n "+str(swsscommon.swsscommon.SonicDBConfig.getDbId("LOGLEVEL_DB"))+" -p "+str(swsscommon.swsscommon.SonicDBConfig.getDbPort("LOGLEVEL_DB")))')"
```
asechoud: this is repeatable code.. please write a function and pass the values.

files/image_config/bash/bash.bashrc
```diff
@@ -73,3 +73,13 @@ if [ -n "$SSH_TARGET_CONSOLE_LINE" ]; then
         exit
     fi
 fi
+
```
selva-nexthop, line 76: Good enhancement. A few points to consider;
- Shell Overhead: Each $(python -c ...) forks a process. 16 forks on every shell login is quite heavy. I'd recommend a small helper function that runs Python once and loops through the output to create the aliases.

- Dependency: If Python or swsscommon is broken/missing, bashrc will throw errors on login. Wrapping this in a try/except or a command -v check would be safer.

- Namespace Awareness: On chassis systems, Redis often uses different ports per ASIC. Perhaps we can make the port lookup dynamic based on an environment variable or a namespace check?

files/image_config/bash/bash.bashrc
eddieruan-alibaba: SONiC has already moved from python2 to python3. Python is sym link for python3. 

admin@PE3:~$ ls -l /usr/bin/python*
lrwxrwxrwx 1 root root       7 Jun 17  2024 /usr/bin/python -> python3
lrwxrwxrwx 1 root root      10 Apr  9  2023 /usr/bin/python3 -> python3.11
-rwxr-xr-x 1 root root 6819288 Mar 14  2023 /usr/bin/python3.11

eddieruan-alibaba: Updated it.

files/image_config/bash/bash.bashrc
eddieruan-alibaba: Change to a function

files/image_config/bash/bash.bashrc
eddieruan-alibaba: Take care the first and second bullet. 

The third one I will let Chassis folks to take care it. I don't have any insight on namespace.

files/image_config/bash/bash.bashrc
```diff
@@ -73,3 +73,92 @@ if [ -n "$SSH_TARGET_CONSOLE_LINE" ]; then
...
+    dbs = ['APPL_DB', 'ASIC_DB', 'COUNTERS_DB', 'LOGLEVEL_DB',
+           'CONFIG_DB', 'FLEX_COUNTER_DB', 'STATE_DB', 'APPL_STATE_DB']
+    for db in dbs:
+        try:
+            db_id = SonicDBConfig.getDbId(db)
+            db_port = SonicDBConfig.getDbPort(db)
+            print(f'{db}:{db_id}:{db_port}')
+        except Exception as e:
+            print(f'ERROR:Failed to get config for {db}: {e}', file=sys.stderr)
+            sys.exit(1)
```
asechoud: do you want to continue here instead of sys.exit to create rest of the alias

eddieruan-alibaba: Removed

files/image_config/bash/bash.bashrc
```diff
@@ -73,3 +73,91 @@ if [ -n "$SSH_TARGET_CONSOLE_LINE" ]; then
...
+        ["LOGLEVEL_DB"]="logleveldb"
+        ["CONFIG_DB"]="configdb"
+        ["FLEX_COUNTER_DB"]="flexcounterdb"
+        ["STATE_DB"]="statedb"
+        ["APPL_STATE_DB"]="appstatedb"
+    )
+
+    # Run Python once to get all DB configurations
+    local python_output
+    python_output=$(python3 -c "
```
ashwnsri: Suggestion: would be worth splitting the inline Python block into its own standalone script, e.g. `/usr/local/bin/sonic-db-aliases`? I say so because:

Now the DB-to-alias mapping lives in both the bash associative array (`SONIC_DBS`) and the Python `dbs` list. If the mapping were moved entirely into the Python side as a dict( "DB name" : "alias suffix"), the script could emit the alias suffix directly (e.g. `appdb:0:6379` instead of `APPL_DB:0:6379`). That way the bash caller doesn't need its own lookup table at all, and there's no risk of the two lists drifting out of sync. Additionally, with the python script handling the DB name : alias suffix mapping and error reporting via exit code, the bash function shrinks to just calling the script, checking `$?`, and iterating over the output. 

Of course, the tradeoff would be that we would need to install this additional script to /usr/local/bin/ (or appropriate location) by modifying files/build_templates/sonic_debian_extensions.j2 but it would be a minor change.

eddieruan-alibaba: Updated. Created files/scripts/sonic-db-aliases.py and put it in local/bin/

PR #25611, opened 2026-02-21
[YANG]: Add extensions for SRv6 traffic steering configuration

cscarpitta:
#### Why I did it

To support SRv6 Traffic Steering configuration in bgpcfgd.

#### How I did it

Add "segment" to static-route yang model.

#### How to verify it

Run the unit tests and build image.

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511

#### Tested branch

master

Extend YANG model for SRv6 traffic steering configuration

The config_db schema is defined in the HLD https://github.com/cscarpitta/SONiC/blob/bgpcfgd_srv6_steering_hld/doc/srv6/srv6_steering_bgpcfgd_hld.md


ahsalam (approved):
LGTM

src/sonic-yang-models/yang-models/sonic-static-route.yang
```diff
@@ -130,6 +135,23 @@ module sonic-static-route {
           description
             "blackhole refers to a route that, if matched, discards the message silently.";
         }
+        list sidlist {
```
bhouse-nexthop, line 138: SONiC will error out on this during yang validation.  you can't have a list inside of a list, they collapse on eachother.  You'll actually get errors during a 'config replace' with this logic that loads, serializes, and validates the config.  The error message is confusing.  

It is best practice to also update https://github.com/nexthop-ai/private-sonic-buildimage/blob/master/src/sonic-yang-models/tests/files/sample_config_db.json with all config settings as it does a round-trip yang validation and would have caught this.

bhouse-nexthop (changes requested):

PR #25660, opened 2026-02-25
service_checker should consider k8s rollout containers

qiluo-msft:
#### Why I did it
K8s rollout containers have long names, and we will use container label to identify its original container/service names such as "restapi". This PR will extend service_checker to respect either image native containers or k8s rollout containers for the same service.

##### Work item tracking

#### How I did it
Check all containers labels.

#### How to verify it

Unit test. Manually verified on a sonic device with k8s rollout restapi container.

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511

- [ ] 
- [ ]

### DIFF
--- a/src/system-health/tests/test_system_health.py
+++ b/src/system-health/tests/test_system_health.py
@@ -362,6 +362,7 @@ def test_service_checker_check_by_monit(mock_run):
 @patch('swsscommon.swsscommon.ConfigDBConnector')
 def test_service_checker_k8s_containers(mock_config_db, mock_run, mock_docker_client):
     """Test that service checker recognizes Kubernetes-managed containers by labels"""
+    setup()
     mock_db_data = MagicMock()
     mock_get_table = MagicMock()
     mock_db_data.get_table = mock_get_table
@@ -436,6 +437,7 @@ def test_service_checker_k8s_containers(mock_config_db, mock_run, mock_docker_cl
 @patch('swsscommon.swsscommon.ConfigDBConnector')
 def test_service_checker_mixed_containers(mock_config_db, mock_run, mock_docker_client):
     """Test that service checker handles both regular Docker and Kubernetes containers"""
+    setup()
     mock_db_data = MagicMock()
     mock_get_table = MagicMock()
     mock_db_data.get_table = mock_get_table


src/system-health/health_checker/service_checker.py
```diff
@@ -152,9 +152,23 @@ def get_current_running_containers(self):
...
 
             for ctr in lst:
-                running_containers.add(ctr.name)
-                if ctr.name not in self.container_critical_processes:
-                    self.fill_critical_process_by_container(ctr.name)
+                # Check if this is a Kubernetes-managed container
+                labels = ctr.labels or {}
+                ns = labels.get("io.kubernetes.pod.namespace")
+                dtype = labels.get("io.kubernetes.docker.type")
+                kname = labels.get("io.kubernetes.container.name")
```
make1980, line 159: this is the container name in the form of "telemetry", not the actual container instance name returned by docker ps and can't be used with docker inspect.

we probably should just ignore k8s managed container and let k8s to be responsible for the correctness of the container(like using liveness/readiness probe etc)

qiluo-msft: resolved

make1980 (approved):

src/system-health/health_checker/service_checker.py
```diff
@@ -152,6 +152,20 @@ def get_current_running_containers(self):
...
             for ctr in lst:
+                # Check if this is a Kubernetes-managed container
+                labels = ctr.labels or {}
+                ns = labels.get("io.kubernetes.pod.namespace")
+                dtype = labels.get("io.kubernetes.docker.type")
+                kname = labels.get("io.kubernetes.container.name")
+
+                if ns == "sonic":
+                    # Kubernetes-managed container - add service name to running containers
+                    # but skip critical process checking (k8s has its own health mechanisms)
```
saiarcot895, line 163: Does k8s's health checks verify that required processes are running inside a container (not just the container itself running)?

qiluo-msft: for this specific case, k8s_container still have critical process defined, and if any critical process crash, it will trigger container exit, and k8s get the signal and trigger several retries, if all failed, keep the container exited.

PR #25666, opened 2026-02-25
[grub2] Pin git clone to specific commit hash for reproducibility

rustiqly:
#### What I did
Pin the grub2 `git clone` to a specific commit hash (`f954d68d`) instead of the `debian/2.06-13+deb13u1` tag name.

#### Why I did it
The grub2 build clones from `salsa.debian.org` using a tag name. This has two problems:
1. **Fragility** — salsa.debian.org has been observed returning HTTP 502 errors, causing late build failures (after 400+ packages have already built)
2. **Reproducibility** — tag names can theoretically be moved; commit hashes cannot

grub2 uses `stg` (stacked git) for patch management, which requires a git repository. A tarball approach is not practical here, so we pin to the commit hash instead.

#### How I did it
- Added `GRUB2_COMMIT` variable with the commit hash of the `debian/2.06-13+deb13u1` tag
- Changed `git checkout debian/$(GRUB2_VERSION)` to `git checkout $(GRUB2_COMMIT)`
- No changes to patch application or build logic

#### How to verify it
```bash
make target/debs/trixie/grub2-common_2.06-13+deb13u1_amd64.deb
```

### DIFF
--- a/src/grub2/Makefile
+++ b/src/grub2/Makefile
@@ -18,14 +18,18 @@ endif
 SUFFIX := $(shell date +%Y%m%d\.%H%M%S)
 STG_BRANCH := stg_temp.$(SUFFIX)
 
+# Pinned commit hash for debian/2.06-13+deb13u1 tag.
+# We use a commit hash instead of a tarball because grub2 patches are applied
+# via stg (stacked git), which requires a git repository with history.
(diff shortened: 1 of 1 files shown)


lguohan:
not making sense. please close @rustiqly

rustiqly:
Thanks for the feedback @lguohan — you're right, tags are equally reliable for pinning. Closing this.

Issue #25683, opened 2026-02-26
Bug:[202511/Trixie] pmon container init fails when IMAGE_VERSION is unset (container_startup.py gets empty -v)


rameshraghupathy:
### Is it platform specific

generic

### Importance or Severity

High

### Description of the bug

On 202511/Trixie, the pmon container’s docker_init.sh (from docker-platform-monitor/docker_init.j2) always calls container_startup.py with -v ${IMAGE_VERSION}. When IMAGE_VERSION is not set, this expands to a bare -v and argparse aborts (expected one argument), causing PMON startup/health checks to fail. Setting a default version (e.g., ${IMAGE_VERSION:-0.0.0}) prevents the crash and allows PMON to start normally.

### Steps to Reproduce

Boot a SONiC 202511 / Trixie image on a DUT.

Ensure IMAGE_VERSION is not set in the pmon container init environment (this happens by default in the failing cases).

Restart PMON (or trigger any action that restarts it), for example:

sudo systemctl restart pmon.service
(alternatively: run config load_minigraph -y, which often restarts containers as part of config reload)

Observe pmon startup failure logs:

container_startup.py: error: argument -v/--version: expected one argument

followed by pmon being marked unhealthy / critical service check reporting pmon: False.

### Actual Behavior and Expected Behavior

Actual Behavior:
On 202511/Trixie, pmon startup can fail because docker_init.sh invokes container_startup.py with -v ${IMAGE_VERSION} and when IMAGE_VERSION is unset it expands to a bare -v, causing argparse to error (expected one argument). This leads to PMON failing to initialize properly and health/critical-service checks reporting pmon as down/unhealthy.

Expected Behavior:
pmon should start reliably even when IMAGE_VERSION is not set; container_startup.py should receive a valid version string (or a safe default) and proceed without argument parsing errors, keeping PMON healthy and critical-service checks passing.

### Relevant log output

```shell
usage: container_startup.py [-h] -f FEATURE -o {local,kube} -v VERSION
container_startup.py: error: argument -v/--version: expected one argument
sonic-platform package not installed, attempting to install...
Error: Unable to locate /usr/share/sonic/platform/sonic_platform-1.0-py2-none-any.whl

Status of critical services: {'pmon': False, ...}
```

### Output of `show version`, `show techsupport`

```shell
202511 after Trixie
```

PR #25684, opened 2026-02-26
Fix for #25683  [202511/Trixie] pmon container init fails when IMAGE_VERSION is unset (container_startup.py gets empty -v)

rameshraghupathy:
On 202511/Trixie, the pmon container’s docker_init.sh (from docker-platform-monitor/docker_init.j2) always calls container_startup.py with -v ${IMAGE_VERSION}. When IMAGE_VERSION is not set, this expands to a bare -v and argparse aborts (expected one argument), causing PMON startup/health checks to fail. Setting a default version (e.g., ${IMAGE_VERSION:-0.0.0}) prevents the crash and allows PMON to start normally.

Fixes: #25683

#### Why I did it
The startup script is being called with a required argument (-v/--version), but in 202511/Trixie IMAGE_VERSION is sometimes unset, so the shell expands -v ${IMAGE_VERSION} into a bare -v, and argparse immediately aborts with expected one argument. By changing it to ${IMAGE_VERSION:-0.0.0}, you ensure -v always receives a value, so container_startup.py can run and PMON can complete initialization instead of failing at argument parsing.

##### Work item tracking

#### How I did it

Entered the pmon container:

docker exec -it pmon bash

Edited the init script used to start PMON:

vi /usr/bin/docker_init.sh

Updated the container_startup.py invocation to provide a default version when IMAGE_VERSION is empty:

${CTR_SCRIPT} -f pmon -o ${RUNTIME_OWNER} -v ${IMAGE_VERSION:-0.0.0}

(optionally also default owner: -o ${RUNTIME_OWNER:-local})

Restarted PMON to validate the change:

exit
docker restart pmon
docker logs pmon --tail 100

#### How to verify it
Restart PMON

docker restart pmon

Confirm the previous argparse error is gone

docker logs pmon --tail 200 | grep -E "expected one argument|container_startup.py: error" || echo "OK: no argparse error"

Confirm PMON container is running

docker ps --filter name=pmon

Verify processes inside PMON are running (most definitive)

docker exec -it pmon bash -lc 'supervisorctl status'

All key daemons should show RUNNING.

- [x] 202511

- [x] 202511


rameshraghupathy:
@prabhataravind @vvolam Can you please review this?

prsunny:
@vvolam , please tag if its required for 202511

vvolam:
> @vvolam , please tag if its required for 202511

Yes @prsunny already added 202511 tag.

StormLiangMS:
@saiarcot895 could you help to review also? We do see many pmon can't start with config_reload in 202511 nightly.

yxieca-admin:
@make1980 if we have IMAGE_VERSION missing and default to 0.0.0. would kubesonic always upgrade the docker?

@rameshraghupathy did you find why the variable was empty? Is the docker image the original one came with the sonic image? Is this issue only happening to PMON docker as we know so far? I see all dockers have similar code structure of setting IMAGE_VERSION environment variable.

@saiarcot895 any thoughts?

saiarcot895:
This particular error message (`container_startup.py: error: argument -v/--version: expected one argument`) shouldn't be causing pmon container to fail to start up. I see multiple containers having the same error message in one of our testbeds running 202511:

```
/var/log/syslog.1:2026 Mar  4 00:22:44.963579 sonic INFO telemetry#supervisord: start container_startup.py: error: argument -v/--version: expected one argument
/var/log/syslog.1:2026 Mar  4 00:22:45.592899 sonic INFO gnmi#supervisord: start container_startup.py: error: argument -v/--version: expected one argument
/var/log/syslog.1:2026 Mar  4 00:22:48.087320 sonic INFO lldp#supervisord: start container_startup.py: error: argument -v/--version: expected one argument
/var/log/syslog.1:2026 Mar  4 00:23:03.907570 sonic INFO dhcp_relay#supervisord: start container_startup.py: error: argument -v/--version: expected one argument
/var/log/syslog.1:2026 Mar  4 00:24:59.608530 sonic INFO dhcp_relay#supervisord: start container_startup.py: error: argument -v/--version: expected one argument
/var/log/syslog.1:2026 Mar  4 00:25:19.300900 sonic INFO gnmi#supervisord: start container_startup.py: error: argument -v/--version: expected one argument
/var/log/syslog.1:2026 Mar  4 00:25:26.988104 sonic INFO lldp#supervisord: start container_startup.py: error: argument -v/--version: expected one argument
/var/log/syslog.1:2026 Mar  4 00:25:55.131571 sonic INFO telemetry#supervisord: start container_startup.py: error: argument -v/--version: expected one argument
```

This suggests that the actual error is something else. Additionally, this environment variable should be being set at container build time, so the question is why this variable isn't being set at container build time. Regardless, I don't think this would block pmon container from starting.

rameshraghupathy:
@yxieca-admin @saiarcot895 Thanks for the review comments.

What I’m seeing on 202511/Trixie is specific to `config reload -y` / minigraph deploy: in some runs PMON doesn’t come back up, and logs show:
`container_startup.py ... error: argument -v/--version: expected one argument`

Root cause for that error is the PMON docker init line passing `-v ${IMAGE_VERSION}` even when `IMAGE_VERSION` is unset, which expands to a bare `-v` and argparse aborts. This PR makes it defensive by defaulting the value:
`-v ${IMAGE_VERSION:-0.0.0}`

Also, on our platform with Trixie we no longer have python2, so the old `py2` wheel fallback message is stale/noisy in this environment.

I have not yet root-caused why `IMAGE_VERSION` is empty in the reload path, and I haven’t evaluated kubesonic side-effects yet. If preferred, I can either keep this PR as a narrow PMON mitigation for `config reload -y` or follow up with a broader/root-cause fix for where `IMAGE_VERSION` should be set.

saiarcot895:
Can you attach a `show techsupport` when pmon fails to start? I see the version error in the case of pmon starting up as well, but it doesn't block pmon starting.

```
2026-03-04 00:23:55,486 WARN stopped: ledd (exit status 143)
2026-03-04 00:23:56,487 INFO waiting for supervisor-proc-exit-listener, rsyslogd to die
2026-03-04 00:23:57,497 INFO stopped: rsyslogd (exit status 0)
2026-03-04 00:23:57,499 WARN stopped: supervisor-proc-exit-listener (terminated by SIGTERM)
usage: container_startup.py [-h] -f FEATURE -o {local,kube} -v VERSION
container_startup.py: error: argument -v/--version: expected one argument
sonic-platform package not installed, attempting to install...
Error: Unable to locate /usr/share/sonic/platform/sonic_platform-1.0-py2-none-any.whl
/usr/local/lib/python3.11/dist-packages/supervisor/options.py:474: UserWarning: Supervisord is running as root and it is searching for its configuration file in default locations (including its current working directory); you probably want to specify a "-c" argument specifying an absolute path to a configuration file for improved security.
  self.warnings.warn(
2026-03-04 00:25:34,417 INFO Included extra file "/etc/supervisor/conf.d/supervisord.conf" during parsing
2026-03-04 00:25:34,419 INFO Set uid to user 0 succeeded
2026-03-04 00:25:34,441 INFO RPC interface 'supervisor' initialized
2026-03-04 00:25:34,447 CRIT Server 'unix_http_server' running without any HTTP authentication checking
2026-03-04 00:25:34,448 INFO supervisord started with pid 1
2026-03-04 00:25:34,454 INFO spawned: 'dependent-startup' with pid 26
2026-03-04 00:25:34,468 INFO spawned: 'supervisor-proc-exit-listener' with pid 27
2026-03-04 00:25:34,548 WARN exited: supervisor-proc-exit-listener (exit status 1; not expected)
2026-03-04 00:25:35,488 INFO success: dependent-startup entered RUNNING state, process has stayed up for > than 1 seconds (startsecs)
2026-03-04 00:25:35,553 INFO spawned: 'supervisor-proc-exit-listener' with pid 28
2026-03-04 00:25:35,619 INFO spawned: 'rsyslogd' with pid 29
```

StormLiangMS:
Hi @rameshraghupathy @saiarcot895 @yxieca-admin @make1980,

I've filed #25907 to address the **root cause** of why `IMAGE_VERSION` is empty in all multi-stage Docker builds.

### Root Cause

In Docker multi-stage builds, `ARG` declarations are **scoped to the build stage** they appear in. All 15 affected Dockerfiles declare `ARG image_version` only in the first (builder) stage, but **not** in the final stage. When `rsync_from_builder_stage()` copies files from builder → final, it only copies filesystem content (via `rsync`), not Docker ENV metadata. So in the final stage:

```dockerfile
FROM $BASE           # ← resets ARG scope
# ARG image_version  # ← MISSING — this is the bug

{{ rsync_from_builder_stage() }}   # copies files only, not ENV

ENV IMAGE_VERSION=$image_version   # ← resolves to empty string
```

The fix in #25907 adds `ARG image_version` after the final `FROM $BASE` in all 15 affected Dockerfiles, so `ENV IMAGE_VERSION=$image_version` correctly receives the build-time value.

### Why defaulting to `0.0.0` may cause issues

To @yxieca-admin's question — yes, defaulting `IMAGE_VERSION` to `0.0.0` could affect Kubernetes rollout behavior. In `container_startup.py`, the version is used for:

1. **STATE_DB updates** — `FEATURE|<name>` table gets `container_version` set to this value
2. **Kubernetes label management** — `drop_label()` sets kube labels based on the version
3. **Version comparison** — `instance_higher()` compares running vs. desired versions to decide whether to upgrade

If the version defaults to `0.0.0`, Kubernetes may interpret it as always needing an upgrade (since `0.0.0` would be lower than any real image version), potentially triggering unnecessary container restarts or rollout loops.

#25907 fixes the actual build-time root cause so `IMAGE_VERSION` is correctly set to the real image version in all containers, avoiding both the argparse crash and the version mismatch concern.

rameshraghupathy:
@StormLiangMS @saiarcot895 @yxieca-admin   As https://github.com/sonic-net/sonic-buildimage/pull/25907 fixes the problem I'll close this unless any of you want to keep it open.

@saiarcot895  The PMON container start issue in 202511/Trixie images seem to be related to the general delay introduced by the trixie image.  I see increasing "StartLimitBurst=10"  to 10 from 3 seem to help.  As we know, the test failure may be triggered by the delay in config_setup.service(due to trixie delay), which pmon depends on.  Should I raise a PR for this?

rameshraghupathy:
(1) As https://github.com/sonic-net/sonic-buildimage/pull/25907 fixes the IMAGE_VERSION problem I'm closing this.
(2) For the pmon container startup issue we will raise another PR after the RCA is complete.

PR #25686, opened 2026-02-26
Create profiler container for azure profiler.

yutongzhang-microsoft:
#### Why I did it
This change integrates Azure Profiler into SONiC by adding a dedicated `profiler` container and wiring it into the image build, service management, and logging pipeline.
The goal is to make Azure Profiler available as a system component in SONiC images so it can be started, monitored, and managed consistently with other system containers.

##### Work item tracking

#### How I did it
+ Added a new `docker-sonic-profiler` container
+ Added build rules and config switches for profiler:
+ Added profiler service integration
+ Added profiler log handling:
+ Updated system health logic so platforms that do not include the profiler image will not incorrectly fail container expectation checks.

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511

- [ ] 
- [ ]


r12f:
This PR should be closed in my understanding. The azure profiler is not something in public but internal to ms, hence should not be added here.

PR #25697, opened 2026-02-26
[cache] Reduce cache invalidation scope with SONIC_CACHE_RECIPE_VER

rustiqly:
## Description

**Problem:** `SONIC_COMMON_FILES_LIST` includes `slave.mk`, which means ANY edit to `slave.mk` (1867 lines, changed ~20x per 6 months) invalidates ALL 161 package caches — even when the change is just adding a new target or fixing a typo.

**Analysis:** Of 15 upstream `slave.mk` commits in the last 3 months, most changes (new targets, platform additions, cosmetic edits) do NOT affect existing package build outputs. Per-package `.dep` files already track actual source dependencies.

**Solution:**
- Remove `slave.mk` from `SONIC_COMMON_FILES_LIST`
- Add `SONIC_CACHE_RECIPE_VER` (manual version stamp) to `SONIC_COMMON_FLAGS_LIST`
- Bump `SONIC_CACHE_RECIPE_VER` when a `slave.mk` change genuinely affects package build output (e.g., dpkg recipe changes, compiler flags, install logic)

This follows the same pattern used by Bazel (`config_setting`) and sccache — separate content-tracking from recipe-tracking.

## Local Test Results

```
Build 1 (cold cache): 15 packages cached successfully
Build 2 (warm cache): All 15 loaded from cache (CACHE::LOADED)

Verified:
✅ .dep files no longer reference slave.mk
✅ .flags files include SONIC_CACHE_RECIPE_VER value
✅ Cache keys survive slave.mk edits
✅ Bumping SONIC_CACHE_RECIPE_VER correctly invalidates all caches
```

## When to bump SONIC_CACHE_RECIPE_VER

Bump when a `slave.mk` change affects package build output:
- ✅ Bump: dpkg build recipe changes, compiler flag changes, install logic changes
- ❌ Don't bump: adding new targets, platform additions, cosmetic changes, new docker images

## Impact

- Cache survives ~90% of upstream syncs (vs 0% today)
- Dramatically reduces rebuild time for incremental development
- No change to build correctness — per-package `.dep` files still track all source dependencies

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -124,7 +124,7 @@ $(PLATFORM_PATH):
 configure : $(PLATFORM_PATH)
 	$(call make_work, $@)
 
-clean showtag docker-cleanup sonic-slave-build sonic-slave-bash :
+clean showtag docker-cleanup clean-docker sonic-slave-build sonic-slave-bash :
 	$(call make_work, $@)
 
 # Freeze the versions, see more detail options: scripts/versions_manager.py freeze -h
--- a/slave.mk
+++ b/slave.mk
@@ -1801,6 +1801,37 @@ clean-versions :: .platform
 vclean:: .platform
 	@sudo rm -rf target/vcache/* target/baseimage*
 
+clean-docker :: .platform
+	@echo "=== Cleaning stale Docker build artifacts ==="
+	@# Remove old sonic-slave images (keep only the current tag)
+	@for img in sonic-slave-bookworm sonic-slave-trixie \
+	            sonic-slave-bookworm-$(USERNAME) sonic-slave-trixie-$(USERNAME) \
+	            tmp-sonic-slave-bookworm tmp-sonic-slave-trixie; do \
+		current_id=$$(docker images --format '{{.ID}}' $$img:$(SLAVE_TAG) 2>/dev/null); \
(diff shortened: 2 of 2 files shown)


rustiqly:
Good question @yxieca! You're right that `git hash-object` would fail in a release tarball without `.git`.

In practice, SONiC builds always run from a git checkout (the build system uses `git describe` and submodule operations extensively), so `.git` is always present during cache builds. But it's still good practice to guard it.

I'll add a `.git` presence check — if git isn't available, it'll skip the baseline comparison and just use the full `SONIC_COMMON_FILES_LIST` (no cache optimization, but no noisy warnings either). Pushing shortly.

yxieca (dismissed):
LGTM. The .git guard resolves the baseline hash concern; cache scope/guard logic looks solid.

rustiqly:
Addressed both Copilot review suggestions in `45095b30d`:

1. **Warning message clarity** — Split into explicit If YES / If NO paths so developers know what to do in both cases.
2. **SONIC_COMMON_FILES_LIST documentation** — Added explanatory comment noting slave.mk is intentionally excluded (with pointer to SONIC_CACHE_RECIPE_VER).

Note: Copilot referenced `rules/template.dep` but that file doesn't mention slave.mk — the comment was added directly above `SONIC_COMMON_FILES_LIST` in `Makefile.cache` where it's most relevant.

PR #25730, opened 2026-02-27
Add whitelist for k8s managed container for health_checker

FengPan-Frank:
##### Work item tracking

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511

- [ ] 
- [ ]


src/system-health/health_checker/service_checker.py
```diff
@@ -92,6 +96,10 @@ def get_expected_running_containers(self, feature_table):
 
         container_list = []
         for container_name in feature_table.keys():
+            # Skip containers in the whitelist
+            if container_name in ServiceChecker.CONTAINER_K8S_WHITELIST:
```
qiluo-msft, line 100: it will skip below `if container_name == "telemetry"` branch completely. Is it intentional?

make1980 (approved):

src/system-health/health_checker/service_checker.py
FengPan-Frank: yes intentional for skipping all kubesonic managed container.

src/system-health/health_checker/service_checker.py
```diff
@@ -155,17 +163,13 @@ def get_current_running_containers(self):
                 # Check if this is a Kubernetes-managed container
                 labels = ctr.labels or {}
                 ns = labels.get("io.kubernetes.pod.namespace")
-                dtype = labels.get("io.kubernetes.docker.type")
-                kname = labels.get("io.kubernetes.container.name")
 
                 if ns == "sonic":
```
make1980: I think as long as ns is not empty we can skip the container

qiluo-msft (approved):

src/system-health/health_checker/service_checker.py
FengPan-Frank: updated.

PR #25731, opened 2026-02-27
[Chassis] Database startup: Improve logging of platform calls

arista-nwolfe:
We ran into a crash on one of these platform calls on Arista platform and it was very difficult to debug without the stderr in the syslog.
Unless there is a good reason to redirect the stderr to /dev/null I think we should capture it in syslog.

Don't delete the stderr outputed by the 2 platform calls
Log the result of the platform calls

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [x] msft-202405
- [ ] 202411
- [ ] 202505
- [x] 202511


deepak-singhal0408 (approved):
LGTM

vmittal-msft:
@arista-nwolfe please raise manual pr for 202511 since there are conflicts

PR #25764, opened 2026-03-02
Add bgpcfgd to docker-sonic-vs

securely1g:
## Description
docker-sonic-vs currently lacks bgpcfgd, which means BGP configuration in CONFIG_DB (`BGP_NEIGHBOR` table) is ignored. cSONiC testbed neighbors bypass the standard SONiC `CONFIG_DB → bgpcfgd → FRR` path, requiring direct FRR template files instead.

Adding bgpcfgd enables cSONiC neighbors to use the same configuration flow as production SONiC, where BGP peers are defined in CONFIG_DB and bgpcfgd generates the FRR configuration.

Related: sonic-net/sonic-mgmt#22680 (moves cSONiC BGP config from FRR templates to CONFIG_DB)

### Changes
- `docker-sonic-vs.mk`: Add `SONIC_BGPCFGD` python wheel; add Makefile rule to copy shared FRR templates and `constants.yml` from `dockers/docker-fpm-frr/frr/` and `files/image_config/constants/` into the build context (no file duplication)
- `Dockerfile.j2`: COPY `constants.yml` to `/etc/sonic/` and FRR bgpcfgd templates to `/usr/share/sonic/templates/`
- `supervisord.conf.j2`: Add bgpcfgd program entry
- `start.sh`: Start bgpd and staticd (FRR daemons) before bgpcfgd
- `.gitignore`: Exclude build-time copied files from version control

### Testing
Deployed cSONiC T0 testbed (`vms-kvm-t0-csonic`) with this change:
- bgpcfgd starts successfully, reads CONFIG_DB BGP_NEIGHBOR entries
- Generates correct FRR configuration (router bgp, neighbors, route-maps, peer-groups)
- All 4 BGP sessions Established between DUT and cSONiC neighbors
- CPU ~1% per container, ~150MB memory

### DIFF
--- a/platform/vs/docker-sonic-vs.mk
+++ b/platform/vs/docker-sonic-vs.mk
@@ -12,7 +12,8 @@ $(DOCKER_SONIC_VS)_DEPENDS += $(SYNCD_VS) \
                               $(LIBYANG_PY3) \
                               $(SONIC_UTILITIES_DATA) \
                               $(SONIC_HOST_SERVICES_DATA) \
-                              $(SYSMGR)
+                              $(SYSMGR) \
+                              $(LLDPD)
 
 $(DOCKER_SONIC_VS)_PYTHON_WHEELS += $(SONIC_PY_COMMON_PY3) \
                                     $(SONIC_PLATFORM_COMMON_PY3) \
@@ -20,7 +21,8 @@ $(DOCKER_SONIC_VS)_PYTHON_WHEELS += $(SONIC_PY_COMMON_PY3) \
                                     $(SONIC_YANG_MGMT_PY3) \
                                     $(SONIC_UTILITIES_PY3) \
                                     $(SONIC_HOST_SERVICES_PY3) \
-                                    $(SONIC_BGPCFGD)
+                                    $(SONIC_BGPCFGD) \
+                                    $(DBSYNCD_PY3)
 
 ifeq ($(INSTALL_DEBUG_TOOLS), y)
 $(DOCKER_SONIC_VS)_DEPENDS += $(LIBSWSSCOMMON_DBG) \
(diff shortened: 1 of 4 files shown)


securely1g:
@hdwhdw Could you review? This adds bgpcfgd to docker-sonic-vs so cSONiC testbed neighbors can use CONFIG_DB for BGP configuration (companion to sonic-net/sonic-mgmt#22680).

platform/vs/docker-sonic-vs.mk
```diff
@@ -12,15 +12,17 @@ $(DOCKER_SONIC_VS)_DEPENDS += $(SYNCD_VS) \
...
 
 $(DOCKER_SONIC_VS)_PYTHON_WHEELS += $(SONIC_PY_COMMON_PY3) \
                                     $(SONIC_PLATFORM_COMMON_PY3) \
                                     $(SONIC_YANG_MODELS_PY3) \
                                     $(SONIC_YANG_MGMT_PY3) \
                                     $(SONIC_UTILITIES_PY3) \
                                     $(SONIC_HOST_SERVICES_PY3) \
-                                    $(SONIC_BGPCFGD)
+                                    $(SONIC_BGPCFGD) \
+                                    $(DBSYNCD_PY3)
```
lguohan, line 25: can we have a $(DOCKER_SONIC_VS)_DOCKER += LLDP?

this could automtically include the lldp docker into this vs docker. what it means is that it will add lldp docker's dockerfile into sonic-vs dockerfile. it also merges start.sh and supervisord.conf.j2 into docker_sonic-vs corresponding file. as well as all other neccessary files as part of lldp docker build.

securely1g: I looked into this and I don't see a `_DOCKER +=` mechanism in the current build system that would automatically merge one docker's Dockerfile, start.sh, and supervisord.conf.j2 into another. The only similar pattern is `_LOAD_DOCKERS` which loads a base layer image (e.g., `DOCKER_SWSS_LAYER_BOOKWORM`), and `_DBG_DOCKER` for debug packages.

Currently we manually copy the LLDP files (lldpmgrd, lldpd.conf.j2, lldpdSysDescr.conf.j2, waitfor_lldp_ready.sh) in the `.frr-templates-stamp` target and add the corresponding Dockerfile/supervisord entries. The `$(LLDPD)` package is already in `_DEPENDS`.

If there's a planned build system feature for `_DOCKER +=` that auto-merges sub-docker components, I'd be happy to adopt it — could you point me to an example or the implementation? Otherwise the current approach follows the same manual pattern used for FRR templates.

securely1g: Thanks for the clarification — I understand now, this would be a new build system feature similar to `add_dbg_docker` in `rules/functions`.

Here's my understanding of what `$(DOCKER_SONIC_VS)_DOCKER += LLDP` would do:

**Build system (slave.mk / rules/functions):**
1. Merge the source docker's `_DEPENDS`, `_PYTHON_WHEELS`, `_FILES` into the target
2. Copy the source docker's supervisord program entries into the target's `supervisord.conf.j2`
3. Merge start.sh snippets (or add a hook mechanism)
4. Copy necessary files (lldpmgrd, conf templates, etc.) into the target's build context

**Implementation approach — a new `add_docker_feature` macro in `rules/functions`:**
```makefile
# call: add_docker_feature target_docker.gz, FEATURE_NAME, source_docker.gz
define add_docker_feature
$(1)_DEPENDS += $($(3)_DEPENDS)
$(1)_PYTHON_WHEELS += $($(3)_PYTHON_WHEELS)
$(1)_FILES += $($(3)_FILES)
$(1)_DOCKERS += $(3)
endef
```

Then in the build recipe, for each entry in `_DOCKERS`, automatically:
- Render and append the source's supervisord entries
- Copy source docker files to the build context

I'll prototype this. Should the scope cover LLDP only for now, or should it be generic enough for any docker feature (e.g., SNMP, TEAMD) to be composable into docker-sonic-vs?

PR #25766, opened 2026-03-02
Fix 'show version' crash in docker-sonic-vs: add missing version fields

### LINKED ISSUE #25765: 'show version' crashes in docker-sonic-vs: missing debian_version and kernel_version in sonic_version.yml
## Description
`show version` crashes with `KeyError: 'debian_version'` inside docker-sonic-vs containers because `sonic_version.yml` does not include `debian_version` or `kernel_version` fields.

## Root Cause
`sonic_version.yml.j2` conditionally includes `debian_version` and `kernel_version` only when those variables are defined during the build:

```jinja2
{% if debian_version is defined -%}
debian_version: '{{ debian_version }}'
{% endif -%}
```

For docker-sonic-vs, these variables are not set during the docker build, so the fields are omitted from the generated `sonic_version.yml`.

However, `show/main.py` (in sonic-utilities) accesses them unconditionally:

```python
click.echo("Distribution: Debian {}".format(version_info['debian_version']))
click.echo("Kernel: {}".format(version_info['kernel_version']))
```

## Reproduction
```bash
docker exec <csonic-container> show version
```

Output:
```
/bin/sh: 1: sudo: not found
SONiC Software Version: SONiC.
SONiC OS Version: 13
  ...
KeyError: 'debian_version'
```

## Additional Issue
`build_version` is also empty, producing `SONiC Software Version: SONiC.` — the version string is not populated during docker-sonic-vs builds.

## Proposed Fix
Two possible approaches:
1. **sonic-utilities**: Use `version_info.get()` with fallback defaults (defensive fix)
2. **sonic-buildimage**: Always include `debian_version` and `kernel_version` in `sonic_version.yml.j2` with sensible defaults, and populate `build_version` during docker-sonic-vs builds

Approach 1 is more robust since `sonic_version.yml` contents may vary across versions.

## Environment
- SONiC docker-sonic-vs built from master branch
- sonic-utilities from master branch

securely1g:
## Description
`show version` crashes with `KeyError: 'debian_version'` inside docker-sonic-vs containers because `sonic_version.yml` is missing `debian_version` and `kernel_version` fields.

### Root Cause
`sonic_version.yml.j2` conditionally includes these fields only when the variables are defined:

```jinja2
{% if debian_version is defined -%}
debian_version: '{{ debian_version }}'
{% endif -%}
```

The full image build (`build_debian.sh`) exports them from the filesystem root, but the VS platform `build_sonic_version.sh` never exports these variables, so the j2 template omits them.

`show/main.py` then accesses them unconditionally:
```python
click.echo("Distribution: Debian {}".format(version_info['debian_version']))
click.echo("Kernel: {}".format(version_info['kernel_version']))
```

### Fix
Export `debian_version` and `kernel_version` from the build host in `platform/vs/sonic-version/build_sonic_version.sh`.

### Reproduction
```bash
docker exec <docker-sonic-vs-container> show version
# KeyError: 'debian_version'
```

### Testing
- Built docker-sonic-vs with the fix
- `show version` completes without errors
- `sonic_version.yml` now includes all expected fields

Fixes #25765

### DIFF
--- a/platform/vs/sonic-version/build_sonic_version.sh
+++ b/platform/vs/sonic-version/build_sonic_version.sh
@@ -1,8 +1,11 @@
 export build_version="${sonic_version}"
+export debian_version="$(cat /etc/debian_version 2>/dev/null || echo 'N/A')"
+export kernel_version="$(uname -r)"
 export asic_type="${sonic_asic_platform}"
 export commit_id="$(git rev-parse --short HEAD)"
 export branch="$(git rev-parse --abbrev-ref HEAD)"
 export build_date="$(date -u)"
 export build_number="${BUILD_NUMBER:-0}"
 export built_by="$USER@$BUILD_HOSTNAME"
+export sonic_os_version="${debian_version%%.*}"
 j2 sonic_version.yml.j2 > $1


securely1g:
Closing — the fix belongs in sonic-utilities, not buildimage. Baking debian_version/kernel_version at build time is wrong for containers since they run on any host kernel. Filed sonic-net/sonic-utilities#4324 instead, which uses .get() with runtime fallbacks.

PR #25767, opened 2026-03-02
Fix 'show version' in docker-sonic-vs: populate missing version fields at startup

### LINKED ISSUE #25765: 'show version' crashes in docker-sonic-vs: missing debian_version and kernel_version in sonic_version.yml

securely1g:
## Description
`show version` crashes with `KeyError: 'debian_version'` in docker-sonic-vs containers because `sonic_version.yml` is missing `debian_version` and `kernel_version` fields.

### Root Cause
These fields cannot be set at build time — the container runs on any host with any kernel. The VS platform `build_sonic_version.sh` correctly omits them, but `show version` expects them.

### Fix
Add startup logic in `start.sh` to populate these fields at container start time:
- `kernel_version`: from `uname -r` (actual host kernel)
- `debian_version`: from `/etc/debian_version` (container's Debian version)

Fields are only appended if not already present, so this is safe for images that already include them.

### Companion PR
sonic-net/sonic-utilities#4324 adds `.get()` fallbacks in `show/main.py` as a defensive fix.

### Testing
Verified in docker-sonic-vs container:
```
$ show version
SONiC Software Version: SONiC.master.1-a1b2c3d
SONiC OS Version: 13
Distribution: Debian 13.0
Kernel: 6.17.0-14-generic
...
```

Fixes #25765

### DIFF
--- a/platform/vs/docker-sonic-vs/start.sh
+++ b/platform/vs/docker-sonic-vs/start.sh
@@ -34,6 +34,13 @@ popd
 
 [ -d /etc/sonic ] || mkdir -p /etc/sonic
 
+# Populate runtime fields in sonic_version.yml that are not available at build time
+if [ -f /etc/sonic/sonic_version.yml ]; then
+    if ! grep -q "^kernel_version:" /etc/sonic/sonic_version.yml; then
+        echo "kernel_version: '$(uname -r)'" >> /etc/sonic/sonic_version.yml
+    fi
+fi
+
 # Note: libswsscommon requires a dabase_config file in /var/run/redis/sonic-db/
 # Prepare this file before any dependent application, such as sonic-cfggen
 mkdir -p /var/run/redis/sonic-db
--- a/platform/vs/sonic-version.mk
+++ b/platform/vs/sonic-version.mk
@@ -1,6 +1,6 @@
 # sonic version yml file
 
-sonic_version=$(SONIC_GET_VERSION)
+sonic_version=$(SONIC_IMAGE_VERSION)
 sonic_asic_platform=$(CONFIGURED_PLATFORM)
 sonic_os_version=$(SONIC_OS_VERSION)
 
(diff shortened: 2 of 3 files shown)


securely1g:
Note:  (sonic-platform-vpp submodule) has the same issue in its `build_sonic_version.sh` — it exports `kernel_version=$(uname -r)` at build time, which captures the build host kernel, not the runtime kernel. That needs a separate PR against `sonic-net/sonic-platform-vpp`.

securely1g:
Closing — superseded by #25949 which includes LLDP + `_INCLUDE_DOCKER` feature composition.

PR #25850, opened 2026-03-03
Update port_config.ini header to correctly display 2024-2026

croos12:
#### Why I did it
port_config.ini header has become outdated due to the new year and needs to be updated to 2026.


croos12:
@rustiqly can you take a look?

PR #25875, opened 2026-03-04
[PDDF] Add IIO, SMBus, scaling and threshold support to voltage sensor base class

pbdspec:
#### Why I did it

PDDF voltage sensors previously relied on the common base class to read hwmon attributes. Platforms that expose voltage rails through the Linux IIO subsystem, such as ADC-backed rails, could not use the common PddfVoltageSensor path without adding platform-specific Python code.

An earlier version of this change also added raw userspace SMBus/PMBus access to the generic PDDF voltage sensor path. Based on review feedback, that is the wrong abstraction boundary: generic PDDF should not perform board-specific I2C transactions, mux selection, or PMBus VOUT decoding. Those responsibilities belong in kernel drivers and platform init code, with PDDF consuming the resulting sysfs interfaces.

#### How I did it

Updated the generic PDDF voltage sensor flow so PddfVoltageSensor reads voltage values through PddfApi-resolved sysfs paths only:

- hwmon sensors continue to use the existing hwmon attribute path
- IIO sensors use sensor_type="iio" and channel_id metadata to read in_voltageN_raw
- scaling remains driven by multi_factor and decimate_factor in pddf-device.json
- threshold APIs continue to prefer hwmon sysfs limits and fall back to JSON thresholds

Extended PddfApi with generic sysfs resolution helpers for sensor devices:

- resolve static and dynamic sensor paths
- locate I2C devices and backing devices from topology metadata
- locate hwmon and IIO sysfs directories under the bound kernel device
- support dynamic mux child bus lookup for platform-described topologies
- synthesize the default IIO voltage attribute from channel_id when attr_list is empty

This keeps generic PDDF hardware-independent. Platform-specific code is expected to load/bind the appropriate kernel drivers so voltage rails appear through hwmon or IIO sysfs before PDDF reads them.

#### How to verify it

- python3 -m py_compile platform/pddf/platform-api-pddf-base/sonic_platform_pddf_base/pddf_voltage_sensor.py platform/pddf/platform-api-pddf-base/sonic_platform_pddf_base/pddfapi.py
- git diff --check -- platform/pddf/platform-api-pddf-base/sonic_platform_pddf_base/pddf_voltage_sensor.py platform/pddf/platform-api-pddf-base/sonic_platform_pddf_base/pddfapi.py
- Verified on Juniper PTX10002 hardware with platform init binding the required kernel drivers: MAX1139 IIO rails, PMBus/hwmon VRM rails, CPU rails, and hotswap rails all report through show platform voltage.

N/A

Enhance the PDDF common voltage sensor path to support Linux IIO-backed voltage sensors through sysfs while preserving existing hwmon behavior. Raw SMBus/PMBus access is intentionally kept out of generic PDDF; platforms should expose such rails via kernel hwmon/IIO drivers.

N/A

Signed-off-by: Satish Gopalakrishnan <satishg@juniper.net>

### DIFF
--- a/platform/pddf/platform-api-pddf-base/sonic_platform_pddf_base/pddf_voltage_sensor.py
+++ b/platform/pddf/platform-api-pddf-base/sonic_platform_pddf_base/pddf_voltage_sensor.py
@@ -64,37 +64,13 @@ class PddfVoltageSensor(VoltageSensorBase):
     # ------------------------------------------------------------------ #
     def get_value(self):
         """Read voltage value in millivolts.  Returns float or None."""
-        if self._sensor_type == 'iio':
-            return self._read_iio()
-        return self._read_hwmon()
-
-    def _read_hwmon(self):
-        """Read via standard PDDF hwmon sysfs path."""
         output = self.pddf_obj.get_attr_name_output(self.sensor_obj_name, "volt1_input")
         if not output:
-            return None
-
-        if output['status'].isalpha():
-            return None
-
-        raw = float(output['status'])
-        return self._scale(raw)
-
-    # ------------------------------------------------------------------ #
-    #  IIO read path
-    # ------------------------------------------------------------------ #
-    def _read_iio(self):
-        """Read voltage via IIO sysfs through the standard PDDF accessor."""
-        output = self.pddf_obj.get_attr_name_output(self.sensor_obj_name, "volt1_input")
-        if not output:
-            return None
-
-        if output['status'].isalpha():
             return None
 
         try:
             return self._scale(float(output['status']))
-        except ValueError:
+        except (KeyError, TypeError, ValueError):
             return None
 
     # ------------------------------------------------------------------ #
@@ -120,9 +96,10 @@ class PddfVoltageSensor(VoltageSensorBase):
         output = self.pddf_obj.get_attr_name_output(self.sensor_obj_name, attr_name)
         if not output:
             return None
-        if output['status'].isalpha():
+        try:
+            return float(output['status'])
+        except (KeyError, TypeError, ValueError):
             return None
-        return float(output['status'])
 
     def _get_threshold_from_json(self, key):
         """Fallback: look up static threshold from dev_attr in JSON.
--- a/platform/pddf/platform-api-pddf-base/sonic_platform_pddf_base/pddf_voltage_sensor.py
+++ b/platform/pddf/platform-api-pddf-base/sonic_platform_pddf_base/pddf_voltage_sensor.py
@@ -6,7 +6,15 @@ except ImportError as e:
     raise ImportError(str(e) + "- required module not found")
 
 class PddfVoltageSensor(VoltageSensorBase):
-    """PDDF generic Voltage Sensor class"""
+    """PDDF generic Voltage Sensor class.
+
+    Supports sysfs-backed sensor read methods based on sensor_type in dev_attr:
+      - hwmon (default): reads hwmon sysfs via standard PDDF path
+      - iio: reads IIO sysfs (iio:deviceX/in_voltageN_raw)
+
+    Scaling (multi_factor / decimate_factor) is applied to all reads.
+    Thresholds are read from JSON dev_attr if present, falling back to sysfs.
+    """
     pddf_obj = {}
     plugin_data = {}
 
@@ -23,73 +31,137 @@ class PddfVoltageSensor(VoltageSensorBase):
             raise ValueError(f"Voltage sensor {self.sensor_obj_name} not found in PDDF data")
 
         self.sensor_obj = self.pddf_obj.data[self.sensor_obj_name]
-        self.display_name = self.sensor_obj.get('dev_attr', {}).get('display_name', self.sensor_obj_name)
-        
+
+        # dev_attr may be at top level or under i2c
+        self._top_dev_attr = self.sensor_obj.get('dev_attr', {})
+        self._i2c_dev_attr = self.pddf_obj.get_sensor_i2c_dev_attr(self.sensor_obj)
+
+        self.display_name = (self._top_dev_attr.get('display_name') or
+                             self._i2c_dev_attr.get('display_name', self.sensor_obj_name))
+
+        # Sensor type: 'iio' or '' (hwmon default)
+        self._sensor_type = self._i2c_dev_attr.get('sensor_type', '')
+
+        # Scaling factors (default 1:1 if missing or invalid in JSON)
+        try:
+            self._multi_factor = float(
+                self._top_dev_attr.get('multi_factor',
+                    self._i2c_dev_attr.get('multi_factor', '1')))
+        except (ValueError, TypeError):
+            self._multi_factor = 1.0
+        try:
+            self._decimate_factor = float(
+                self._top_dev_attr.get('decimate_factor',
+                    self._i2c_dev_attr.get('decimate_factor', '1')))
+        except (ValueError, TypeError):
+            self._decimate_factor = 1.0
 
     def get_name(self):
-        if 'dev_attr' in self.sensor_obj.keys():
-            if 'display_name' in self.sensor_obj['dev_attr']:
-                return str(self.sensor_obj['dev_attr']['display_name'])
-        # In case of errors
-        return (self.sensor_obj_name)
-    
+        return self.display_name
 
+    # ------------------------------------------------------------------ #
+    #  Value read — dispatches by sensor_type
+    # ------------------------------------------------------------------ #
     def get_value(self):
+        """Read voltage value in millivolts.  Returns float or None."""
+        if self._sensor_type == 'iio':
+            return self._read_iio()
+        return self._read_hwmon()
+
+    def _read_hwmon(self):
+        """Read via standard PDDF hwmon sysfs path."""
         output = self.pddf_obj.get_attr_name_output(self.sensor_obj_name, "volt1_input")
         if not output:
             return None
 
         if output['status'].isalpha():
-            attr_value = None
-        else:
-            attr_value = float(output['status'])
-        
-        return attr_value
+            return None
+
+        raw = float(output['status'])
+        return self._scale(raw)
+
+    # ------------------------------------------------------------------ #
+    #  IIO read path
+    # ------------------------------------------------------------------ #
+    def _read_iio(self):
+        """Read voltage via IIO sysfs through the standard PDDF accessor."""
+        output = self.pddf_obj.get_attr_name_output(self.sensor_obj_name, "volt1_input")
+        if not output:
+            return None
+
+        if output['status'].isalpha():
+            return None
+
+        try:
+            return self._scale(float(output['status']))
+        except ValueError:
+            return None
+
+    # ------------------------------------------------------------------ #
+    #  Scaling
+    # ------------------------------------------------------------------ #
+    def _scale(self, raw_value):
+        """Apply multi/decimate scaling.  Returns millivolts (whole number)."""
+        if raw_value is None or self._decimate_factor == 0:
+            return None
+        # Truncate to integer millivolts, return as float for API consistency
+        return float(int(raw_value * self._multi_factor / self._decimate_factor))
+
+    # ------------------------------------------------------------------ #
+    #  Thresholds — sysfs first (chip hw), JSON dev_attr fallback
+    # ------------------------------------------------------------------ #
+    def _get_threshold_from_sysfs(self, attr_name):
+        """Read threshold from hwmon sysfs via PDDF attr_list mapping.
+
+        Works when the chip supports the PMBus warn/fault registers
+        (e.g. inN_max, inN_min, inN_crit, inN_lcrit) and the JSON
+        attr_list maps them via drv_attr_name.
+        """
+        output = self.pddf_obj.get_attr_name_output(self.sensor_obj_name, attr_name)
+        if not output:
+            return None
+        if output['status'].isalpha():
+            return None
+        return float(output['status'])
+
+    def _get_threshold_from_json(self, key):
+        """Fallback: look up static threshold from dev_attr in JSON.
+
+        Used for chips that don't support warn limit registers
+        (e.g. TPS53659/TPS53622 lack VOUT_OV_WARN / VOUT_UV_WARN).
+        """
+        val = self._top_dev_attr.get(key)
+        if val is None:
+            val = self._i2c_dev_attr.get(key)
+        if val is None:
+            return None
+        try:
+            return float(val)
+        except (ValueError, TypeError):
+            return None
 
     def get_low_threshold(self):
-        output = self.pddf_obj.get_attr_name_output(self.sensor_obj_name, "volt1_low_threshold")
-        if not output:
-            return None
-
-        if output['status'].isalpha():
-            attr_value = None
-        else:
-            attr_value = float(output['status'])
-
-        return attr_value
+        # Try sysfs first (chip hw warn limit)
+        val = self._get_threshold_from_sysfs("volt1_low_threshold")
+        if val is not None:
+            return val
+        # Fallback: static JSON threshold
+        return self._get_threshold_from_json('low_threshold')
 
     def get_high_threshold(self):
-        output = self.pddf_obj.get_attr_name_output(self.sensor_obj_name, "volt1_high_threshold")
-        if not output:
-            return None
+        val = self._get_threshold_from_sysfs("volt1_high_threshold")
+        if val is not None:
+            return val
+        return self._get_threshold_from_json('high_threshold')
 
-        if output['status'].isalpha():
-            attr_value = None
-        else:
-            attr_value = float(output['status'])
-
-        return attr_value
-    
     def get_high_critical_threshold(self):
-        output = self.pddf_obj.get_attr_name_output(self.sensor_obj_name, "volt1_crit_high_threshold")
-        if not output:
-            return None
+        val = self._get_threshold_from_sysfs("volt1_crit_high_threshold")
+        if val is not None:
+            return val
+        return self._get_threshold_from_json('crit_high_threshold')
 
-        if output['status'].isalpha():
-            attr_value = None
-        else:
-            attr_value = float(output['status'])
-
-        return attr_value
-    
     def get_low_critical_threshold(self):
-        output = self.pddf_obj.get_attr_name_output(self.sensor_obj_name, "volt1_crit_low_threshold")
-        if not output:
-            return None
-
-        if output['status'].isalpha():
-            attr_value = None
-        else:
-            attr_value = float(output['status'])
-
-        return attr_value
+        val = self._get_threshold_from_sysfs("volt1_crit_low_threshold")
+        if val is not None:
+            return val
+        return self._get_threshold_from_json('crit_low_threshold')
--- a/platform/pddf/platform-api-pddf-base/sonic_platform_pddf_base/pddfapi.py
+++ b/platform/pddf/platform-api-pddf-base/sonic_platform_pddf_base/pddfapi.py
@@ -16,6 +16,8 @@ SONIC_CFGGEN_PATH = '/usr/local/bin/sonic-cfggen'
 LED_CTRL_LOCK_PATH = '/var/lock/pddf-locks/pddf-api-led.lock'
 HWSKU_KEY = 'DEVICE_METADATA.localhost.hwsku'
 PLATFORM_KEY = 'DEVICE_METADATA.localhost.platform'
+SENSOR_TYPE_HWMON = 'hwmon'
+SENSOR_TYPE_IIO = 'iio'
 
 dirname = os.path.dirname(os.path.realpath(__file__))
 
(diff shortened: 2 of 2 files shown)


pbdspec:
azpw/ run

pbdspec:
=== Voltage Sensor Readings (39 sensors) — Juniper Platform ===

Sensor Name                      Value(mV)  HighThresh  LowThresh  Status
--------------------------------------------------------------------------------

--- IIO sensors (max1139 ADC, channel 0-10) ---
MAX1139_BLRE_SD1V1                   1108.0      1160.0     1040.0  OK
MAX1139_BLRE_VCC1V1                  1104.0      1160.0     1040.0  OK
MAX1139_CPU0_VCC1P8_AUX_NAC          1806.0      1890.0     1710.0  OK
MAX1139_PTP_VCCO1V8                  1808.0      1890.0     1710.0  OK
MAX1139_PTP_VCCO3V3                  3320.0      3470.0     3130.0  OK
MAX1139_PTP_VINT1V0                  1006.0      1050.0      950.0  OK
MAX1139_PTP_VMGT1V2                  1232.0      1260.0     1140.0  OK
MAX1139_SSD0_BIAS3P3                 3316.0      3470.0     3130.0  OK
MAX1139_SSD1_BIAS3P3                 3324.0      3470.0     3130.0  OK
MAX1139_USB_VCC5P0                   5028.0      5250.0     4750.0  OK
MAX1139_VDD5V0                       5028.0      5250.0     4750.0  OK

--- hwmon sensors (tps53659, max16602, max16545 via DCDC+virt_parent) ---
CPU0_VCC1P2_VDDQ01                   1230.0      1260.0     1140.0  OK
CPU0_VCC2P5_VPP01                    2500.0      2560.0     2440.0  OK
CPU0_VCC1P8_AUX                      1800.0      1860.0     1740.0  OK
CPU0_VCCIN_CPU                       1790.0      1860.0     1740.0  OK
CPU0_VCC1P05                         1060.0      1100.0     1000.0  OK
CPU0_VCCVNN_PCH                       950.0      1013.0      917.0  OK
CPU0_VCC1P0_PCIE4                    1000.0      1060.0      940.0  OK
CPU0_VCCVNN_NAC                       855.0       893.0      807.0  OK
CPU0_VCC3P3_AUX                      3300.0      3465.0     3135.0  OK
CPU_HOTSWAP                          1192.0      1260.0     1140.0  OK
BXX_HOTSWAP_0                        1198.0      1260.0     1140.0  OK
BXX_HOTSWAP_1                        1191.0      1260.0     1140.0  OK
BXX_HOTSWAP_2                        1194.0      1260.0     1140.0  OK
BXX_HOTSWAP_3                        1199.0      1260.0     1140.0  OK
BXX_HOTSWAP_4                        1196.0      1260.0     1140.0  OK

--- SMBus sensors (max20796 via raw PMBus reads behind PCA9548 MUX) ---
BXX_FAB_0V85                          847.0       893.0      807.0  OK
BXX_WAN_0V85                          849.0       893.0      807.0  OK
BXX_FAB_1V1                          1097.0      1155.0     1045.0  OK
BX0_CORE_0V75                         750.0       800.0      700.0  OK
BX1_CORE_0V75                         750.0       800.0      700.0  OK
BXX_WAN_1V1                          1100.0      1260.0     1140.0  Warning
BXX_HBM_1V2                          1199.0      1260.0     1140.0  OK
BX0_DIE_CRACK_MON                     702.0       800.0      550.0  OK
BX1_DIE_CRACK_MON                     702.0       800.0      550.0  OK
BX0_RING_H                           1822.0      2500.0        0.0  OK
BX0_RING_T                              2.0      2500.0        0.0  OK
BX1_RING_H                           1822.0      2500.0        0.0  OK
BX1_RING_T                              2.0      2500.0        0.0  OK

pbdspec:
@gordon-nexthop can you please review this change ?

ashwnsri:
@judyjoseph @yxieca is there anybody else who can review this PR? TIA.

platform/pddf/platform-api-pddf-base/sonic_platform_pddf_base/pddf_voltage_sensor.py
```diff
@@ -23,73 +76,390 @@ def __init__(self, index, pddf_data=None, pddf_plugin_data=None):
...
+    def get_low_critical_threshold(self):
+        val = self._get_threshold_from_sysfs("volt1_crit_low_threshold")
+        if val is not None:
+            return val
+        return self._get_threshold_from_json('crit_low_threshold')
+
+    # ------------------------------------------------------------------ #
+    #  I2C device path helpers
+    # ------------------------------------------------------------------ #
+    def _find_i2c_device_path(self):
```
gordon-nexthop: Again I think all of this should belong in `pddfapi.py`, this is meant to be a standard interface?

platform/pddf/platform-api-pddf-base/sonic_platform_pddf_base/pddfapi.py
```diff
@@ -408,8 +408,75 @@ def _path_expand(*path):
...
         return self.show_attr_hwmon_device(dev, ops, "voltage-sensors")
 
+    def show_attr_iio_device(self, dev, ops, data_sysfs_key):
+        """Resolve IIO sysfs path for voltage/current sensors on ADC chips.
+
+        IIO ADCs (e.g. max1139, max11613) expose voltage readings at:
+            /sys/bus/i2c/devices/{bus}-{addr}/iio:device*/in_voltageN_raw
+        where N is the ADC channel_id from dev_attr.
+        """
+        def _path_expand(*path):
```
gordon-nexthop: nit: commonize this since its used by `hwmon` too.

platform/pddf/platform-api-pddf-base/sonic_platform_pddf_base/pddf_voltage_sensor.py
```diff
@@ -23,73 +76,390 @@ def __init__(self, index, pddf_data=None, pddf_plugin_data=None):
...
-    def get_low_threshold(self):
-        output = self.pddf_obj.get_attr_name_output(self.sensor_obj_name, "volt1_low_threshold")
-        if not output:
+        raw = float(output['status'])
+        return self._scale(raw)
+
+    # ------------------------------------------------------------------ #
+    #  IIO read path
+    # ------------------------------------------------------------------ #
+    def _read_iio(self):
```
gordon-nexthop: Why can we not just use `get_attr_name_output`, you’ve implemented this already in `pddfapi.py` via `show_attr_iio_device`?

platform/pddf/platform-api-pddf-base/sonic_platform_pddf_base/pddf_voltage_sensor.py
gordon-nexthop, line 1: I do not think a lot of these changes don’t belong in this files:
1) `sonic_platform_pddf_base` classes are meant to be meant to be a simple wrappers around `sonic_platform_base` that add in use of the `PddfApi` to grab information.
2) A lot of this is not specific to voltage sensors - it applies to current and probably temperature sensors too
3) Even in pddf API the idea is that we’re interacting with SYSFS interfaces / commands encoded inside of the `pddf-device.json`. 

Putting in this driver level code is not the idea (it should be abstracted behind a sysfs interface) and only represents quite a small subset, of relevant information within the pmbus spec. It still can be common across vendors by living in `platform/pddf/i2c/modules`?

platform/pddf/platform-api-pddf-base/sonic_platform_pddf_base/pddfapi.py
```diff
@@ -408,8 +408,75 @@ def _path_expand(*path):
         return ret
 
     def show_attr_voltage_sensor_device(self, dev, ops):
+        sensor_type = dev.get('i2c', {}).get('dev_attr', {}).get('sensor_type', '')
+        if sensor_type == 'iio':
```
gordon-nexthop: Perhaps abstract this logic into a common method that can be used by other sensor types e.g. this function is just a call to  `show_attr_sensor_device` that just passes in `voltage_sensors` in addition to the `dev` and `ops`. Also maybe enum for sensor type?

adyeung:
@geans-pin please help review

geans-pin (commented):
Some comments for this PR, please also provide the UT log for the PR change to insure the change works fine. 

[High] JSON field names in the PR description do not match the code

The sample JSON for SMBus shows:

"parent_mux_bus": "0x3d",
"parent_mux_addr": "0x70",
"parent_mux_ch": "0x08"
But _read_smbus() dispatches based on mux2_addr / mux2_chan, and _read_smbus_via_parent() reads adapter_name, mux_chan, mux2_addr, and mux2_chan from dev_attr. None of these match the documented JSON keys. Platforms that follow the PR description JSON will silently take the non-MUX path and read the wrong bus. The author needs to reconcile the sample JSON with the code (or vice versa).

2. [High] MUX channel vs. bitmask ambiguity in _read_smbus_via_parent()

mux2_chan = int(dev_attr.get('mux2_chan', '0'))
...
mux_select = 1 << mux2_chan  # PCA9548: one-hot channel bitmask
The comment says mux2_chan is the channel number (0–7), so 1 << 4 → 0x10 for channel 4. But the PR description shows "parent_mux_ch": "0x08" which is decimal 8 — an invalid channel number on a PCA9548 (channels are 0–7). If vendors follow the documented JSON and use 0x08 as a channel mask value (meaning channel 3), 1 << 8 = 256 would be written to the MUX, corrupting its state. The intent needs to be clarified: is the JSON value a channel index or a bitmask?

3. [High] virt_parent lookup targets the wrong JSON key

In _find_i2c_device_path():

if "virt_parent" in self.sensor_obj.get('dev_info', {}):
But the sample JSON places virt_parent under the i2c key, not dev_info:

"i2c": { "virt_parent": "VRM1", ... }
This check will always be False for any platform using the documented schema, silently falling back to the sensor's own topo_info instead of the parent VRM.

pbdspec:
> Some comments for this PR, please also provide the UT log for the PR change to insure the change works fine.
> 
> [High] JSON field names in the PR description do not match the code
> 
> The sample JSON for SMBus shows:
> 
> "parent_mux_bus": "0x3d", "parent_mux_addr": "0x70", "parent_mux_ch": "0x08" But _read_smbus() dispatches based on mux2_addr / mux2_chan, and _read_smbus_via_parent() reads adapter_name, mux_chan, mux2_addr, and mux2_chan from dev_attr. None of these match the documented JSON keys. Platforms that follow the PR description JSON will silently take the non-MUX path and read the wrong bus. The author needs to reconcile the sample JSON with the code (or vice versa).
> 
> 2. [High] MUX channel vs. bitmask ambiguity in _read_smbus_via_parent()
> 
> mux2_chan = int(dev_attr.get('mux2_chan', '0')) ... mux_select = 1 << mux2_chan # PCA9548: one-hot channel bitmask The comment says mux2_chan is the channel number (0–7), so 1 << 4 → 0x10 for channel 4. But the PR description shows "parent_mux_ch": "0x08" which is decimal 8 — an invalid channel number on a PCA9548 (channels are 0–7). If vendors follow the documented JSON and use 0x08 as a channel mask value (meaning channel 3), 1 << 8 = 256 would be written to the MUX, corrupting its state. The intent needs to be clarified: is the JSON value a channel index or a bitmask?
> 
> 3. [High] virt_parent lookup targets the wrong JSON key
> 
> In _find_i2c_device_path():
> 
> if "virt_parent" in self.sensor_obj.get('dev_info', {}): But the sample JSON places virt_parent under the i2c key, not dev_info:
> 
> "i2c": { "virt_parent": "VRM1", ... } This check will always be False for any platform using the documented schema, silently falling back to the sensor's own topo_info instead of the parent VRM.

@geans-pin 

Thanks for the detailed review.

I reworked the PR to remove the raw SMBus/PMBus implementation from generic PDDF. The generic voltage sensor path now reads only kernel-exposed sysfs data: existing hwmon paths and IIO paths.

As a result, the SMBus sample JSON from the earlier PR description is no longer applicable and has been removed. Fields such as parent_mux_bus, parent_mux_addr, parent_mux_ch, mux2_addr, and mux2_chan are not part of the updated generic PDDF voltage sensor behavior. This resolves the JSON/code mismatch and the mux channel vs. bitmask ambiguity called out in comments 1 and 2.

For the virt_parent comment, agreed that the previous description/code was confusing. The updated code follows the existing PDDF i2c object/sysfs resolution model and avoids the removed raw-SMBus helper path.

Regarding UT logs: there is no existing UT target for the PDDF platform-api-pddf-base voltage sensor path in this repo. For this revision, I validated with:
- python3 -m py_compile on the changed PDDF Python files
- git diff --check for the changed files
- check that no raw SMBus/PMBus access remains in generic PDDF
- runtime validation on Juniper/HPE PTX10002 hardware using `show platform voltage`, covering hwmon, IIO ADC, PMBus-through-hwmon, CPU, and hotswap voltage sensors

I also updated the PR description/commit message to match the current sysfs-based implementation.

platform/pddf/platform-api-pddf-base/sonic_platform_pddf_base/pddf_voltage_sensor.py
pbdspec: Thanks Gordon, agreed.

I reworked the change based on this feedback. The updated revision removes the driver-level raw SMBus/PMBus handling from `sonic_platform_pddf_base`. Generic PDDF no longer performs PMBus register reads, VOUT decode, or mux manipulation.

The voltage sensor class is back to being a thin platform API wrapper: it asks `PddfApi` for the configured voltage attribute and applies the existing scaling/threshold handling. The hardware access path is now through kernel-exposed sysfs interfaces only:
- existing hwmon attributes
- IIO attributes for ADC-backed voltage sensors

For PMBus regulators, the platform is expected to bind the appropriate kernel driver and expose the voltage through hwmon. Any platform-specific device binding/topology belongs in platform init / platform data, not in the generic PDDF voltage sensor class.

This also keeps the generic sysfs lookup behavior in `PddfApi`, where it can be shared by other sensor classes if needed, instead of embedding driver-level logic in `PddfVoltageSensor`.

platform/pddf/platform-api-pddf-base/sonic_platform_pddf_base/pddf_voltage_sensor.py
pbdspec: Agreed. This was a good catch.

I updated the implementation so `PddfVoltageSensor` now uses `get_attr_name_output()` for both hwmon and IIO voltage reads instead of duplicating IIO path/read logic inside the voltage sensor class.

The IIO-specific path resolution remains inside `PddfApi` through `show_attr_iio_device()`, and the voltage sensor class simply asks `PddfApi` for the configured `volt1_input` value. This keeps `PddfVoltageSensor` as a thin wrapper around `PddfApi`, consistent with the existing PDDF platform API structure.

platform/pddf/platform-api-pddf-base/sonic_platform_pddf_base/pddfapi.py
pbdspec: Agreed, fixed.

I moved the path expansion logic into a shared helper in `PddfApi` and updated the IIO path resolution to use it. The existing hwmon/static path handling can use the same helper, so this avoids duplicating the glob/path expansion logic between hwmon and IIO.

platform/pddf/platform-api-pddf-base/sonic_platform_pddf_base/pddf_voltage_sensor.py
pbdspec: Agreed, fixed.

I moved the I2C/sysfs path discovery into `PddfApi`, where the existing PDDF sysfs resolution logic already lives. `PddfVoltageSensor` no longer has its own I2C path helper; it now calls `get_attr_name_output()` for both hwmon and IIO voltage sensors and only handles voltage-specific scaling/threshold behavior.

This keeps `PddfVoltageSensor` as the standard platform API wrapper and keeps sysfs/device discovery centralized in `PddfApi`.

platform/pddf/platform-api-pddf-base/sonic_platform_pddf_base/pddfapi.py
pbdspec: Agreed, addressed in the updated revision.

I added named sensor type constants and moved the common hwmon/IIO sysfs path resolution into `show_attr_sensor_device()`. Both `show_attr_hwmon_device()` and `show_attr_iio_device()` now call that common helper, and current/temp sensors continue to go through the hwmon wrapper.

`show_attr_voltage_sensor_device()` remains as the voltage-specific entry point for compatibility with the existing PDDF dispatch model and to pass the `"voltage-sensors"` data_sysfs key, but it no longer owns the actual sysfs path construction. It only selects the IIO vs. hwmon wrapper based on `sensor_type`.

I did not introduce a Python enum class because this codebase currently uses simple module-level constants for this style of dispatch, so I used `SENSOR_TYPE_HWMON` and `SENSOR_TYPE_IIO` to keep the change consistent with the existing style.

platform/pddf/platform-api-pddf-base/sonic_platform_pddf_base/pddfapi.py
```diff
@@ -358,44 +474,138 @@ def show_attr_mux_device(self, dev, ops):
...
+            return None
+        return full_path[0]
+
+    def _get_i2c_backing_device(self, dev):
+        virt_parent = dev.get('dev_info', {}).get('virt_parent')
+        if virt_parent is None:
+            virt_parent = dev.get('i2c', {}).get('virt_parent')
+        return self.data.get(virt_parent, dev) if virt_parent else dev
+
+    def _get_attr_list(self, dev):
```
gordon-nexthop: nit: `get_i2c_attr_list`

platform/pddf/platform-api-pddf-base/sonic_platform_pddf_base/pddfapi.py
```diff
@@ -358,44 +474,138 @@ def show_attr_mux_device(self, dev, ops):
...
-                    real_name = attr['attr_name']
-
-                if 'topo_info' in i2c_dev['i2c']:
-                    path = self.show_device_sysfs(i2c_dev, ops)+"/%d-00%02x/"%(int(i2c_dev['i2c']['topo_info']['parent_bus'], 0),
-                            int(i2c_dev['i2c']['topo_info']['dev_addr'], 0))
-                    full_path = _path_expand(path, 'hwmon', 'hwmon*', real_name)
-                elif 'path_info' in i2c_dev['i2c']:
-                    path = i2c_dev['i2c']['path_info']['sysfs_base_path']
-                    full_path = _path_expand(path, real_name)
+                real_name = self._get_sensor_attr_real_name(
```
gordon-nexthop: This `real_name` derivation refactoring is quite confusing, why can we not use the previous contract where you must define an `attr_list` underneath `i2c` for a device, and if you’d like to have a `real_name` you must provide a `drv_attr_name`? Hardcoding defaults seems undesirable unless it is the case that these truly are IIO defaults for voltage/current sensors?

gordon-nexthop (commented):
I might have missed this but I couldn’t see an example `pddf-device.json` that allows for this, would be helpful for review.

pbdspec:
> I might have missed this but I couldn’t see an example `pddf-device.json` that allows for this, would be helpful for review.

@gordon-nexthop here is the expected pddf-device.json shape for an IIO-backed voltage sensor:

{
    "VOLTAGE1": {
        "dev_info": {
            "device_type": "VOLTAGE_SENSOR",
            "device_name": "VOLTAGE1"
        },
        "i2c": {
            "topo_info": {
                "parent_bus": "0x5D",
                "dev_addr": "0x35",
                "dev_type": "max1139"
            },
            "dev_attr": {
                "display_name": "MAX1139_BLRE_SD1V1",
                "sensor_type": "iio",
                "multi_factor": "2",
                "decimate_factor": "1"
            },
            "attr_list": [
                {
                    "attr_name": "volt1_input",
                    "drv_attr_name": "in_voltage2_raw"
                }
            ]
        }
    }
}

Here attr_name is the logical PDDF attribute requested by the voltage sensor API, and drv_attr_name is the actual sysfs IIO attribute exposed by the kernel driver.

platform/pddf/platform-api-pddf-base/sonic_platform_pddf_base/pddfapi.py
pbdspec: Addressed. I removed the default real_name derivation and kept the previous explicit attr_list contract. The code now uses attr['drv_attr_name'] when provided, otherwise attr['attr_name']; platform data must explicitly map logical attrs to sysfs names through drv_attr_name.

platform/pddf/platform-api-pddf-base/sonic_platform_pddf_base/pddfapi.py
pbdspec: Addressed. Renamed the helper to _get_i2c_attr_list() since it specifically reads i2c.attr_list from pddf-device.json.

geans-pin:
Some comments:

**_find_mux_child_bus — chan_id hex-string format mismatch**
pddfapi.py
Lines 585-613
def _find_mux_child_bus(self, parent_bus, chan_id, mux_addr=None):
    parent_path = '/sys/bus/i2c/devices/i2c-%d' % parent_bus
    expected_name = 'i2c-%d-mux (chan_id %s)' % (parent_bus, chan_id)
The Linux i2c-mux driver names child buses as "i2c-N-mux (chan_id C)" where C is always a decimal integer (e.g., chan_id 2). If mux_chan is stored in pddf-device.json as a hex string ("0x2") — which is common in PDDF JSON — then expected_name becomes "i2c-5-mux (chan_id 0x2)" and the lookup silently fails, returning None. The channel ID should be normalised to a decimal string:
expected_name = 'i2c-%d-mux (chan_id %d)' % (parent_bus, int(chan_id, 0))
The same issue exists in _get_dynamic_i2c_bus which passes dev_attr[mux_key] directly.

**_get_i2c_backing_device — silent fallback may hide misconfigured JSON**
pddfapi.py
Lines 483-487
def _get_i2c_backing_device(self, dev):
    virt_parent = dev.get('dev_info', {}).get('virt_parent')
    if virt_parent is None:
        virt_parent = dev.get('i2c', {}).get('virt_parent')
    return self.data.get(virt_parent, dev) if virt_parent else dev
The original code did self.data[dev['dev_info']['virt_parent']] — a KeyError if the parent name was wrong. The new self.data.get(virt_parent, dev) silently falls back to the device itself if the named parent is not found in self.data. A platform with a typo in virt_parent would read from the wrong sysfs path with no error. At minimum a warning log is warranted:
if virt_parent and virt_parent not in self.data:
    print("PDDF: virt_parent '{}' not found in data, using self".format(virt_parent))


**_read_hwmon and _read_iio are functionally identical**
def _read_hwmon(self):
    output = self.pddf_obj.get_attr_name_output(self.sensor_obj_name, "volt1_input")
    ...

def _read_iio(self):
    output = self.pddf_obj.get_attr_name_output(self.sensor_obj_name, "volt1_input")
    ...

Both methods call the exact same get_attr_name_output call and apply the same scaling. The sensor-type dispatch is already fully encapsulated in PddfApi.show_attr_voltage_sensor_device. Having two separate methods that do the same thing is confusing to future maintainers and suggests the distinction is meaningful when it is not. Consider collapsing into a single _read_voltage():
def get_value(self):
    output = self.pddf_obj.get_attr_name_output(self.sensor_obj_name, "volt1_input")
    if not output or output['status'].isalpha():
        return None
    try:
        return self._scale(float(output['status']))
    except ValueError:
        return None

pbdspec:
> Some comments:
> 
> **_find_mux_child_bus — chan_id hex-string format mismatch** pddfapi.py Lines 585-613 def _find_mux_child_bus(self, parent_bus, chan_id, mux_addr=None): parent_path = '/sys/bus/i2c/devices/i2c-%d' % parent_bus expected_name = 'i2c-%d-mux (chan_id %s)' % (parent_bus, chan_id) The Linux i2c-mux driver names child buses as "i2c-N-mux (chan_id C)" where C is always a decimal integer (e.g., chan_id 2). If mux_chan is stored in pddf-device.json as a hex string ("0x2") — which is common in PDDF JSON — then expected_name becomes "i2c-5-mux (chan_id 0x2)" and the lookup silently fails, returning None. The channel ID should be normalised to a decimal string: expected_name = 'i2c-%d-mux (chan_id %d)' % (parent_bus, int(chan_id, 0)) The same issue exists in _get_dynamic_i2c_bus which passes dev_attr[mux_key] directly.
> 
> **_get_i2c_backing_device — silent fallback may hide misconfigured JSON** pddfapi.py Lines 483-487 def _get_i2c_backing_device(self, dev): virt_parent = dev.get('dev_info', {}).get('virt_parent') if virt_parent is None: virt_parent = dev.get('i2c', {}).get('virt_parent') return self.data.get(virt_parent, dev) if virt_parent else dev The original code did self.data[dev['dev_info']['virt_parent']] — a KeyError if the parent name was wrong. The new self.data.get(virt_parent, dev) silently falls back to the device itself if the named parent is not found in self.data. A platform with a typo in virt_parent would read from the wrong sysfs path with no error. At minimum a warning log is warranted: if virt_parent and virt_parent not in self.data: print("PDDF: virt_parent '{}' not found in data, using self".format(virt_parent))
> 
> **_read_hwmon and _read_iio are functionally identical** def _read_hwmon(self): output = self.pddf_obj.get_attr_name_output(self.sensor_obj_name, "volt1_input") ...
> 
> def _read_iio(self): output = self.pddf_obj.get_attr_name_output(self.sensor_obj_name, "volt1_input") ...
> 
> Both methods call the exact same get_attr_name_output call and apply the same scaling. The sensor-type dispatch is already fully encapsulated in PddfApi.show_attr_voltage_sensor_device. Having two separate methods that do the same thing is confusing to future maintainers and suggests the distinction is meaningful when it is not. Consider collapsing into a single _read_voltage(): def get_value(self): output = self.pddf_obj.get_attr_name_output(self.sensor_obj_name, "volt1_input") if not output or output['status'].isalpha(): return None try: return self._scale(float(output['status'])) except ValueError: return None

Addressed.

1. Normalized mux channel values with int(chan_id, 0), so both decimal and hex JSON values match the kernel child-bus name.

2. Updated virt_parent handling so a configured but missing parent no longer silently falls back to the child device. It now logs the missing parent and returns no path.

3. Collapsed the duplicate _read_hwmon() / _read_iio() paths into a single get_value() implementation, since PddfApi already handles hwmon vs IIO path resolution.

I also added boundary checks for hex/decimal mux channels, missing virt_parent, and nonnumeric voltage/threshold reads.

Thanks!

gordon-nexthop (approved):
LGTM based on my feedback, thank you!

pbdspec:
@geans-pin can you please let me know if the latest changes look good ?

ashwnsri:
@adyeung @geans-pin please help review/approve (as appropriate), TIA

ashwnsri:
@yxieca Please help approve this PR. TIA. @geans-pin @adyeung

geans-pin (approved):

pbdspec:
@geans-pin can you please help in merging this change also ? I dont have permissions to merge , please let me know if you can help in this as well. Thanks!

PR #25876, opened 2026-03-04
ci: fix s360 security vulnerability

auspham:
#### Why I did it

Due to the recent s360 vulnerability reports. This PR is needed to address the problems. We have vulnerability in the follow image:
1. docker-ptf
2. debian
3. docker-sonic-mgmt

##### Work item tracking
- Microsoft ADO **(number only)**: 36979761

#### How I did it
Apply the recommended version from security report

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511

- [ ] 
- [ ]

### DIFF
--- a/dockers/docker-base-bookworm/Dockerfile.j2
+++ b/dockers/docker-base-bookworm/Dockerfile.j2
@@ -56,6 +56,10 @@ RUN apt update &&            \
         libwrap0            \
         libatomic1
 
+# Security fixes: upgrade vulnerable base packages (S360 scan remediation)
+RUN apt-get update && apt-get upgrade -y \
+    && rm -rf /var/lib/apt/lists/*
+
 # Add a config file to allow pip to install packages outside of apt/the Debian repos
 COPY ["pip.conf", "/etc/pip.conf"]
 
--- a/dockers/docker-ptf/Dockerfile.j2
+++ b/dockers/docker-ptf/Dockerfile.j2
@@ -106,7 +106,7 @@ RUN apt-get update          \
     {% else %}
     && GRPCURL_ARCH=linux_x86_64 \
     {% endif %}
-    && GRPCURL_VERSION=1.9.1 && \
+    && GRPCURL_VERSION=1.9.3 && \
     curl -L "https://github.com/fullstorydev/grpcurl/releases/download/v${GRPCURL_VERSION}/grpcurl_${GRPCURL_VERSION}_${GRPCURL_ARCH}.tar.gz" \
       -o /tmp/grpcurl_${GRPCURL_VERSION}_${GRPCURL_ARCH}.tar.gz && \
     cd /tmp && \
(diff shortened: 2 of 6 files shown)


files/build/versions-public/dockers/docker-ptf/versions-py3
```diff
@@ -50,7 +50,7 @@ rpyc==6.0.2
 saithrift==0.9
 scapy==2.7.0
 serpent==1.42
-setuptools==51.0.0
+setuptools==65.5.1
```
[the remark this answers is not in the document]
auspham: The observation about `install_offending_packages` re-pinning `setuptools==51.0.0` is correct — the macro in `dockers/dockerfile-macros.j2` does run `pip3 install setuptools==51.0.0` after each offending package. However, the suggestion to revert versions-py3 back to `setuptools==51.0.0` would leave the security vulnerability unfixed. Instead, the `install_offending_packages` macro in `dockerfile-macros.j2` should also be updated to pin to `setuptools==65.5.1` (or higher) so the fix actually takes effect in the built image.

files/build/versions-public/dockers/docker-ptf/versions-py3
```diff
@@ -50,7 +50,7 @@ rpyc==6.0.2
...
 simplegeneric==0.8.1
 six==1.17.0
 stack-data==0.6.3
```
wangxin, line 56: Is this file maintained by reproducible build? @liushilongbuaa Can you help confirm?
If yes, then it should not be manually updated.

dockers/docker-ptf/Dockerfile.j2
```diff
@@ -339,7 +341,8 @@ COPY gnxi-patches/ gnxi/patches/
 RUN cd gnxi \
     && quilt push -a \
     && cd gnmi_cli_py \
-    && pip install -r requirements.txt
+    && pip install -r requirements.txt \
```
wangxin, line 350: What about the packages specified in the `requirements.txt`? Those packages may need upgrade too.

dockers/docker-ptf/Dockerfile.j2
```diff
@@ -155,7 +155,7 @@ RUN rm -rf /debs \
     && rm -f get-pip.py \
     && pip install setuptools \
     && pip install supervisor \
-    && pip install ipython==5.4.1 \
+    && pip install ipython==5.11 \
```
wangxin: For general purpose packages, maybe the better idea is not to fix the package version. Reproducible build will always try to get the latest package.
If pinned package version, there is high chance that some vulnerability issue is identified in this version in the future. To reduce vulnerability risk, it would be better to keep package versions up to date.

files/build/versions-public/dockers/docker-ptf/versions-py3
liushilongbuaa: It can't manually update.

auspham: removed this one

dockers/docker-ptf/Dockerfile.j2
auspham: Since this is other repository. We updated the patch 0005-Enhance-gnmi_cli_py. There will be an upgrade there.

dockers/docker-ptf/Dockerfile.j2
auspham: Thanks. Removed the pinned version

dockers/docker-base-bookworm/Dockerfile.j2
```diff
@@ -56,6 +56,28 @@ RUN apt update &&            \
...
+        libpam-modules-bin   \
+        libpam-runtime       \
+        libpam0g             \
+        libstdc++6           \
+        libsystemd0          \
+        libudev1             \
+        login                \
+        passwd               \
+        perl-base            \
+    && rm -rf /var/lib/apt/lists/*
```
wangxin, line 61: If just run `apt-get update` and then `apt-get upgrade`, can all the packages be upgraded? Then no need to hard code the package names one by one. What's more, it is not scalable. When some extra packages have new version, need to update the list again.

wangxin (approved):

PR #25911, opened 2026-03-05
[cherry-pick][202511] Fix IMAGE_VERSION unset in final stage of multi-stage Docker builds

StormLiangMS:
## [cherry-pick] Fix IMAGE_VERSION unset in final stage of multi-stage Docker builds

Cherry-pick of #25907 to 202511 branch.

### Description

In Docker multi-stage builds, `ARG` declarations are scoped to the build stage they appear in. All affected Dockerfiles declare `ARG image_version` only in the first (builder) stage but not in the final stage. Since `rsync_from_builder_stage()` copies only filesystem content (not Docker ENV metadata), `ENV IMAGE_VERSION=$image_version` in the final stage resolves to an empty string.

This causes `container_startup.py` to fail with:
```
error: argument -v/--version: expected one argument
```

### Fix

Add `ARG image_version` after the final `FROM $BASE` in all affected Dockerfiles (15 files on 202511), and add missing `ENV IMAGE_VERSION=$image_version` in `docker-sonic-bmp` where it was absent from the final stage.

Note: `docker-restapi-sidecar` was removed in 202511 branch, so only 15 files are affected (vs 16 on master).

#### How I did it
Added `ARG image_version` in the final build stage of each multi-stage Dockerfile so that `ENV IMAGE_VERSION=$image_version` resolves correctly.

#### How to verify it
```bash
docker inspect <container> --format '{{.Config.Env}}' | grep IMAGE_VERSION
```
Should show the actual image version instead of empty string.

- [x] 202511

### DIFF
--- a/dockers/docker-auditd/Dockerfile.j2
+++ b/dockers/docker-auditd/Dockerfile.j2
@@ -21,6 +21,8 @@ RUN chmod +x /usr/bin/config_checker.py
 
 FROM $BASE
 
+ARG image_version
+
 {{ rsync_from_builder_stage() }}
 
 # Make apt-get non-interactive
--- a/dockers/docker-bmp-watchdog/Dockerfile.j2
+++ b/dockers/docker-bmp-watchdog/Dockerfile.j2
@@ -39,6 +39,8 @@ RUN chmod +x /usr/bin/bmp_watchdog
 
 FROM $BASE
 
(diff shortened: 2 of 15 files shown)


StormLiangMS:
hi @saiarcot895 @yxieca @saiarcot895  could you help to review?

PR #25912, opened 2026-03-05
Always install bash-completion in image

saiarcot895:
#### Why I did it

With sonic-net/sonic-utilities#4163, the bash-completion package was moved from Requires to Suggests. This means that it won't be installed by default.

##### Work item tracking

#### How I did it

Make sure we install it always in `build_debian.sh`, so that completions work.

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511

- [ ] 
- [ ]


tjchadaga (approved):

yxieca:
@liat-grozovik I thought there was some issue with installing this by default? On wimpy CPU platform like 2700 we had issue before?

Issue #14899, opened 2023-05-01
memory_checker log spam when feature is disabled


bluecmd:
#### Steps to reproduce the issue:
1. `sudo config feature state telemetry disabled`
2. `sudo config feature state snmp disabled`

#### Describe the results you received:

The log contains the following every minute:
```
May  1 06:10:59.311151 svc-kg-sw1 INFO memory_checker: [memory_checker] Exits without checking memory usage since container 'telemetry' is not running!
May  1 06:11:59.270778 svc-kg-sw1 INFO memory_checker: [memory_checker] Exits without checking memory usage since container 'telemetry' is not running!
May  1 06:12:59.345380 svc-kg-sw1 INFO memory_checker: [memory_checker] Exits without checking memory usage since container 'telemetry' is not running!
May  1 06:13:59.349895 svc-kg-sw1 INFO memory_checker: [memory_checker] Exits without checking memory usage since container 'telemetry' is not running!
May  1 06:14:59.440201 svc-kg-sw1 INFO memory_checker: [memory_checker] Exits without checking memory usage since container 'telemetry' is not running!
May  1 06:15:59.595638 svc-kg-sw1 INFO memory_checker: [memory_checker] Exits without checking memory usage since container 'telemetry' is not running!
May  1 06:16:59.494024 svc-kg-sw1 INFO memory_checker: [memory_checker] Exits without checking memory usage since container 'telemetry' is not running!
May  1 06:17:59.549450 svc-kg-sw1 INFO memory_checker: [memory_checker] Exits without checking memory usage since container 'telemetry' is not running!
May  1 06:18:59.608446 svc-kg-sw1 INFO memory_checker: [memory_checker] Exits without checking memory usage since container 'telemetry' is not running!
May  1 06:19:59.637789 svc-kg-sw1 INFO memory_checker: [memory_checker] Exits without checking memory usage since container 'telemetry' is not running!
May  1 06:20:59.940657 svc-kg-sw1 INFO memory_checker: [memory_checker] Exits without checking memory usage since container 'telemetry' is not running!
May  1 06:21:59.718121 svc-kg-sw1 INFO memory_checker: [memory_checker] Exits without checking memory usage since container 'telemetry' is not running!
May  1 06:22:59.805145 svc-kg-sw1 INFO memory_checker: [memory_checker] Exits without checking memory usage since container 'telemetry' is not running!
[...]

PR #25915, opened 2026-03-05
Fix memory_checker log spam when feature is disabled

Sumalatha-G-ML:
#### Why I did it

Fix memory_checker log spam when feature is disabled.

##### Work item tracking
- Microsoft ADO **(number only)**: N/A
- GitHub Issue: #14899

The memory_checker script was generating spam log messages every minute when containers were not running, even when the corresponding features (like telemetry, snmp, otel) were intentionally disabled in CONFIG_DB. This made it difficult to identify real issues in the logs.

#### How I did it

Added a check to determine if a feature is disabled in the CONFIG_DB FEATURE table before logging the 'container not running' message.

**Changes:**
- Added `is_feature_enabled()` function to check feature state in CONFIG_DB
- Modified `main()` to only log 'container not running' if feature is enabled
- Imported `ConfigDBConnector` from swsscommon for database access

**Behavior:**
- If feature is disabled: Script exits silently without logging
- If feature is enabled but container not running: Script logs the message (existing behavior preserved)
- If feature state cannot be determined: Script logs the message (fail-safe to preserve existing behavior)

#### How to verify it

**Test Case 1: Disabled Feature (No Log Spam)**
```bash
config feature state telemetry disabled
# Wait for monit to run memory_checker
grep "memory_checker.*telemetry" /var/log/syslog
# Expected: No new log entries

Test Case 2: Enabled Feature with Stopped Container (Existing Behavior)
config feature state snmp enabled
docker stop snmp
# Wait for monit to run memory_checker
grep "memory_checker.*snmp" /var/log/syslog
# Expected: Log message appears: "Exits without checking memory usage since container 'snmp' is not running!"

Test Case 3: Enabled Feature with Running Container (Normal Operation)
config feature state telemetry enabled
docker ps | grep telemetry
# Expected: No error logs (normal operation)

Which release branch to backport (provide reason below if selected)
202305
202311
202405
202411
202505
202511
Reason: This is a bug fix that reduces log spam. Can be backported to active release branches if needed.

Description for the changelog
Fix memory_checker log spam for disabled features by checking CONFIG_DB FEATURE table state before logging

Link to config_db schema for YANG module changes
N/A - No YANG model changes

Reason: This is a bug fix that reduces log spam. Can be backported to active release branches if needed.


ramanan-net (approved):
LGTM

PR #25927, opened 2026-03-05
[Workaround][macsec][q3d] Disable bad svtag interrupt on macsec core

saravanan-nexthop:
MACsec cores 13 and 16 are seeing continuous interrupt storm for interrupt 136 on init. These are raised for bad SvTag packets, but the system is not seeing these packets. While the root cause is still being worked, temporarily mask the interrupts.
We need this PR's diff until BCM CSP CS00012446293 is fixed

##### Work item tracking

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [x] 202511

- [ ] 
- [ ]


arlakshm:
@saravanan-nexthop is the PR ready for review?

PR #25933, opened 2026-03-06
Enhance FRR daemon readiness monitoring with detailed diagnostics

ypcisco:
#### Why I did it

In scale scenarios where zebra daemon takes longer than 20 seconds to start, bgpcfgd was timing out and exiting. When zebra eventually came up, bgpcfgd was no longer running, resulting in BGP configurations not being pushed. This change increases the timeout and adds detailed logging to diagnose such startup timing issues.

##### Work item tracking

#### How I did it

- Added 120-second timeout enforcement
- Track retry attempts and timing (elapsed/remaining time)
- Log which daemons are found vs missing on each retry
- Increased polling interval from 100ms to 1s
- Enhanced success/failure messages with timing details

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511

- [ ] 
- [ ]

### DIFF
--- a/src/sonic-bgpcfgd/bgpcfgd/frr.py
+++ b/src/sonic-bgpcfgd/bgpcfgd/frr.py
@@ -19,15 +19,39 @@ class FRR(object):
         :param seconds: number of seconds to wait, until raise an error
         """
         stop_time = datetime.datetime.now() + datetime.timedelta(seconds=seconds)
-        log_info("Start waiting for FRR daemons: %s" % str(datetime.datetime.now()))
+        start_time = datetime.datetime.now()
+
+        log_info("Start waiting for FRR daemons (timeout=%ds): %s" % (seconds, str(start_time)))
+        log_info("Required daemons: %s" % str(self.daemons))
+
+        retry_count = 0
+
         while datetime.datetime.now() < stop_time:
+            retry_count += 1
             ret_code, out, err = run_command(["vtysh", "-c", "show daemons"], hide_errors=True)
+
             if ret_code == 0 and all(daemon in out for daemon in self.daemons):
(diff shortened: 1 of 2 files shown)


src/sonic-bgpcfgd/bgpcfgd/frr.py
```diff
@@ -19,15 +19,39 @@ def wait_for_daemons(self, seconds):
...
+                missing_daemons = [d for d in self.daemons if d not in out]
+                log_warn("Waiting for daemons (%.1fs elapsed, %.1fs remaining, attempt %d): found=%s missing=%s" %
+                        (elapsed, remaining, retry_count, found_daemons, missing_daemons))
             else:
-                log_warn("Can't read daemon status from FRR: %s" % str(err))
-            time.sleep(0.1)  # sleep 100 ms
+                log_warn("Can't read daemon status from FRR (%.1fs elapsed, %.1fs remaining, attempt %d): %s" %
+                        (elapsed, remaining, retry_count, str(err)))
+
+            time.sleep(1.0)
```
anish-n: I wonder if its better to retry aggressively so that we can bring up bgp as soon as possible, so we keep it 0.1s instead of increasing sleep time to 1s. If the reason why the polling has been increased to 1s is that we will log too much, we can add additional logic to log only once every second

ypcisco: Updated

deepak-singhal0408 (approved):
Looks good..

anish-n (approved):

deepak-singhal0408:
Hi @lguohan @yxieca @rlhui, could you help merge this PR.

PR #25947, opened 2026-03-07
Fix dash-ha container after upgrading to trixie

yue-fred-gao:
#### Why I did it
dash-ha container failed to start after SONiC upgrades to trixie. 
1. A bug in  docker_image_ctl.j2 renders an invalid argument to docker create. The invalid argument was ignored by docker in bookworm but rejected by trixie
2. /etc/timezone has deprecated in trixie and needs to change to /etc/localtime
##### Work item tracking

#### How I did it
Fix the logic error that renders invalid argument in docker_image_ctl.j2 
Change /etc/timezone to /etc/localtime in docker-dash-ha.mk

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511

- [ ] 
- [ ]

### DIFF
--- a/files/build_templates/docker_image_ctl.j2
+++ b/files/build_templates/docker_image_ctl.j2
@@ -611,6 +611,8 @@ start() {
     else
         DB_OPT=$DB_OPT" --add-host=redis_chassis.server:$chassis_db_address "
     fi
+    {%- elif docker_container_name == "dash-ha" %}
+    REDIS_MNT="-v /var/run/redis:/var/run/redis:rw"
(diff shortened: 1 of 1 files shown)


prsunny:
@yue-fred-gao , could you create a separate PR for 202511?

yue-fred-gao:
> @yue-fred-gao , could you create a separate PR for 202511?

done: https://github.com/sonic-net/sonic-buildimage/pull/25986

PR #25950, opened 2026-03-08
[snmpd] Fix Trixie build: t64 package names and missing patch dir

rustiqly:
## Description
[agent]
Fix snmpd build under Trixie slave container.

### What I did
1. **t64 package renames**: Debian Trixie renames shared library packages as part of the 64-bit time_t transition. Make `DERIVED_TARGETS` in `src/snmpd/Makefile` conditional on `BLDENV`:
   - `libsnmp40` → `libsnmp40t64` (trixie)
   - `libsnmp40-dbgsym` → `libsnmp40t64-dbgsym` (trixie)
   - `libnetsnmptrapd40` dropped in trixie (merged into libsnmp40t64)

2. **Missing patch directory**: Add `patch-5.9.4+dfsg/` for the Trixie snmpd version (5.9.4). Without this, `stg import -s ../patch-$(SNMPD_VERSION)/series` fails. Patches are copied from 5.9.3 (apply cleanly to 5.9.4).

### Why I did it
`rules/snmpd.mk` already has per-BLDENV t64 conditional names, but `src/snmpd/Makefile` did not — causing the `mv $(DERIVED_TARGETS)` command to fail when the actual .deb files have t64 suffixed names.

The missing `patch-5.9.4+dfsg/` directory causes an immediate build failure in the Trixie phase.

### How I verified it
Built snmpd successfully under the trixie-only build branch.
Bookworm path is unchanged (hits the `else` branch).

## Type of change
- Bug fix

### DIFF
--- a/src/snmpd/patch-5.9.4+dfsg/0008-Enable-macro-DEB_BUILD_ARCH_OS-in-order-to-build-ipv.patch
+++ b/src/snmpd/patch-5.9.4+dfsg/0008-Enable-macro-DEB_BUILD_ARCH_OS-in-order-to-build-ipv.patch
@@ -1,0 +1,25 @@
+From bd7d6d79217da6b5a2eb3a3ee1d409c175a61e41 Mon Sep 17 00:00:00 2001
+From: Qi Luo <qiluo-msft@users.noreply.github.com>
+Date: Fri, 20 Sep 2019 00:42:19 +0000
+Subject: [PATCH] Enable macro DEB_BUILD_ARCH_OS in order to build ipv6 feature
+
+---
+ debian/rules | 2 ++
+ 1 file changed, 2 insertions(+)
+
+diff --git a/debian/rules b/debian/rules
+index 34d8509..496be59 100755
+--- a/debian/rules
++++ b/debian/rules
+@@ -4,6 +4,8 @@
+ export DEB_BUILD_MAINT_OPTIONS := hardening=+all
+ include /usr/share/dpkg/architecture.mk
+ include /usr/share/dpkg/buildtools.mk
++DEB_HOST_MULTIARCH ?= $(shell dpkg-architecture -qDEB_HOST_MULTIARCH)
++DEB_BUILD_ARCH_OS  ?= $(shell dpkg-architecture -qDEB_BUILD_ARCH_OS)
+ 
+ LIB_VERSION = 40t64
+ 
+-- 
+2.18.0
+
--- a/src/snmpd/Makefile
+++ b/src/snmpd/Makefile
@@ -4,19 +4,26 @@ SHELL = /bin/bash
 
 MAIN_TARGET = libsnmp-base_$(SNMPD_VERSION_FULL)_all.deb
 ifneq (,$(findstring 5.9,$(SNMPD_VERSION)))
+# Common targets for all 5.9.x versions
 DERIVED_TARGETS = snmptrapd_$(SNMPD_VERSION_FULL)_$(CONFIGURED_ARCH).deb \
 		  snmptrapd-dbgsym_$(SNMPD_VERSION_FULL)_$(CONFIGURED_ARCH).deb \
 		  snmp_$(SNMPD_VERSION_FULL)_$(CONFIGURED_ARCH).deb \
 		  snmpd_$(SNMPD_VERSION_FULL)_$(CONFIGURED_ARCH).deb \
 		  snmp-dbgsym_$(SNMPD_VERSION_FULL)_$(CONFIGURED_ARCH).deb \
 		  snmpd-dbgsym_$(SNMPD_VERSION_FULL)_$(CONFIGURED_ARCH).deb \
-		  libsnmp40_$(SNMPD_VERSION_FULL)_$(CONFIGURED_ARCH).deb \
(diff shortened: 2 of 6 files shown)


lguohan:
@rustiqly , please provide sonic-mgmt test results

StormLiangMS:
## Review

Good fix for the Trixie build. The t64 package rename handling and patch directory addition look correct. A few observations:

**1. Potential issue: `rules/snmpd.mk` still defines `LIBNETSNMPTRAPD40` for trixie**

The PR description says *libnetsnmptrapd40 dropped in trixie (merged into libsnmp40t64)*, and the Makefile's `DERIVED_TARGETS` correctly omits it for trixie. However, `rules/snmpd.mk` still registers it:

`makefile
ifeq ($(BLDENV),trixie)
LIBNETSNMPTRAPD40 = libnetsnmptrapd40t64_$(SNMPD_VERSION_FULL)_$(CONFIGURED_ARCH).deb
...
$(eval $(call add_derived_package,$(LIBSNMP_BASE),$(LIBNETSNMPTRAPD40)))
`

If this package is truly dropped in Trixie's snmpd 5.9.4, then `rules/snmpd.mk` should also be updated to skip the `LIBNETSNMPTRAPD40` definition and `add_derived_package` call for trixie. Otherwise the build system may expect a .deb that never gets produced/moved. The current CI builds appear to be Bookworm-only — this could fail in an actual Trixie build.

**2. Patches look good** — the 5 patches in `patch-5.9.4+dfsg/` match `patch-5.9.3+dfsg/series` exactly. Same patches, same order. Since the PR states they apply cleanly to 5.9.4, this is fine.

**3. Makefile conditionals are correct** — the `ifeq ($(BLDENV),trixie)` pattern mirrors what's already in `rules/snmpd.mk`. Clean approach.

**4. Failed CI checks** appear to be ElasticTest KVM test infra issues (dpu, t0-sonic, multi-asic-t1, t1-lag-vpp) + `ms_conflict` — likely unrelated to this change. Retriggering.

StormLiangMS:
@liushilongbuaa same here, could you help to take a look at ms_conflict?

src/snmpd/patch-5.9.4+dfsg/0012-agent-Makefile.in-Unbreak-the-enable-minimalist-buil.patch
```diff
@@ -0,0 +1,30 @@
+From 0d3e33ff197d1d15c0036b69cca5198e54781efe Mon Sep 17 00:00:00 2001
```
rminnikanti: @rustiqly, @lihuay  This patch is no longer required in 5.9.4+dfsg version. Its already present in upstream.

rminnikanti:
@saiarcot895 for viz.
This PR is required for FRR trixie upgrade.

src/snmpd/patch-5.9.4+dfsg/0008-Enable-macro-DEB_BUILD_ARCH_OS-in-order-to-build-ipv.patch
```diff
@@ -0,0 +1,25 @@
...
+--- a/debian/rules
++++ b/debian/rules
+@@ -4,6 +4,8 @@
+ export DEB_BUILD_MAINT_OPTIONS := hardening=+all
+ include /usr/share/dpkg/architecture.mk
+ include /usr/share/dpkg/buildtools.mk
++DEB_HOST_MULTIARCH ?= $(shell dpkg-architecture -qDEB_HOST_MULTIARCH)
++DEB_BUILD_ARCH_OS  ?= $(shell dpkg-architecture -qDEB_BUILD_ARCH_OS)
+ 
+ LIB_VERSION = 40
```
rminnikanti: @rustiqly @lihuay LIB_VERSION is 40t64 in 5.9.4+dfsg. Unsure how this patch still showing "40" after patch porting.

saiarcot895:
@rminnikanti can you instead look at removing the local snmpd build for Trixie entirely? The one non-build patch that we have there doesn't need to be there anymore, I think.

src/snmpd/patch-5.9.4+dfsg/0012-agent-Makefile.in-Unbreak-the-enable-minimalist-buil.patch
rustiqly: You're right — patch 0012 (minimalist build fix) is already applied upstream in 5.9.4+dfsg. Removed it from the series. Thanks for catching this!

src/snmpd/patch-5.9.4+dfsg/0008-Enable-macro-DEB_BUILD_ARCH_OS-in-order-to-build-ipv.patch
rustiqly: Good catch — updated `LIB_VERSION` from `40` to `40t64` in both patch 0008 and the cross-compile patch to match the upstream 5.9.4+dfsg value. Fixed in 42a7cbc.

rustiqly:
Closing per @saiarcot895's feedback — the local snmpd build for Trixie is being removed entirely, making this patch directory and t64 renaming unnecessary. Thanks for the guidance!

PR #25955, opened 2026-03-08
Optimize k8s_pod_control to avoid extra apiServer call

FengPan-Frank:
#### Why I did it
Optimize k8s_pod_control to avoid extra APIServer call

#### How I did it
1. Use the local kubelet API instead of kubectl for getting pods, 
2. Update cmd_wait logic as below:
Loop up to WAIT_MAX_ROUNDS rounds (5 minutes at 20s per round), Return 0 immediately when a Running pod is found, Exit 1 after exhausting all rounds, which causes the systemd service to enter a failed state.

#### How to verify it
image

before fix
image

after fix
image

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511

- [ ] 
- [ ]

### DIFF
--- a/files/scripts/k8s_pod_control.sh
+++ b/files/scripts/k8s_pod_control.sh
@@ -169,7 +169,7 @@ cmd_start() {
 
 # stop is a no-op: K8s controls the pod lifecycle; deleting the pod here
 # would just cause the controller to recreate it immediately.
-cmd_stop()    { log "stop: no-op (pod lifecycle is managed by K8s)"; }
+cmd_stop()    { kill_pods; }
 cmd_restart() { kill_pods; }
 
 cmd_status() {
@@ -190,16 +190,27 @@ cmd_status() {
 }
 
 cmd_wait() {
-  log "Waiting on pods (ns=${NS}, selector=${POD_SELECTOR}) on node ${NODE_NAME}…"
+  local max_rounds="${WAIT_MAX_ROUNDS:-15}"
+  local round=0
+  log "Waiting on pods (ns=${NS}, selector=${POD_SELECTOR}) on node ${NODE_NAME} (max ${max_rounds} rounds)…"
   while true; do
     local out=""; out="$(pods_on_node)"
     if [[ -z "$out" ]]; then
-      sleep 5; continue
+      if (( ++round > max_rounds )); then
+        log "ERROR: timed out after ${max_rounds} rounds waiting for pods (ns=${NS}, selector=${POD_SELECTOR}) on ${NODE_NAME}"
+        exit 1
+      fi
+      sleep 20; continue
     fi
     if awk '$2=="Running"{found=1} END{exit found?0:1}' <<<"$out"; then
-      sleep 60
+      round=0
+      sleep 20
     else
-      sleep 5
+      if (( ++round > max_rounds )); then
+        log "ERROR: timed out after ${max_rounds} rounds waiting for pods (ns=${NS}, selector=${POD_SELECTOR}) on ${NODE_NAME}"
+        exit 1
+      fi
+      sleep 20
     fi
   done
 }


files/scripts/k8s_pod_control.sh
```diff
@@ -124,7 +155,9 @@ cmd_start() {
   exit 0
 }
 
-cmd_stop()    { kill_pods; }
+# stop is a no-op: K8s controls the pod lifecycle; deleting the pod here
+# would just cause the controller to recreate it immediately.
+cmd_stop()    { log "stop: no-op (pod lifecycle is managed by K8s)"; }
```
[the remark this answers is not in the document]
FengPan-Frank: stop is not supported in k8s rollout container now.

files/scripts/k8s_pod_control.sh
```diff
@@ -30,6 +30,11 @@ MAX_ATTEMPTS=10
 BACKOFF_START=1
 BACKOFF_MAX=8
 
+# Kubelet local API – queries stay on-node; avoids API server load
+KUBELET_URL="https://localhost:10250"
+KUBELET_CERT="/etc/kubernetes/pki/apiserver-kubelet-client.crt"
+KUBELET_KEY="/etc/kubernetes/pki/apiserver-kubelet-client.key"
+
```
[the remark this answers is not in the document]
FengPan-Frank: already validated now.

files/scripts/k8s_pod_control.sh
```diff
@@ -56,18 +61,44 @@ kubectl_retry() {
...
+      --max-time 5 "${KUBELET_URL}/pods" 2>&1); rc=$?
+    if (( rc == 0 )); then
+      printf '%s' "$out"
+      return 0
+    fi
+    if (( attempt >= MAX_ATTEMPTS )); then
+      echo "$out" >&2
+      return "$rc"
+    fi
+    log "kubelet curl retry ${attempt}/${MAX_ATTEMPTS}"
```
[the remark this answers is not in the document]
FengPan-Frank: uses -w '%{http_code}' to capture and validate HTTP status codes, with descriptive error messages for 401/403 vs other failures. Non-2xx responses trigger retries instead of silently returning bad data.

files/scripts/k8s_pod_control.sh
```diff
@@ -56,18 +61,44 @@ kubectl_retry() {
   done
 }
 
+kubelet_get_pods() {
+  local attempt=1 backoff=${BACKOFF_START} out rc
+  while true; do
+    out=$(curl -sk --cert "${KUBELET_CERT}" --key "${KUBELET_KEY}" \
+      --max-time 5 "${KUBELET_URL}/pods" 2>&1); rc=$?
```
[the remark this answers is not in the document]
FengPan-Frank: Kept -k since the kubelet's serving cert is self-signed (not signed by cluster CA). Added a clear comment explaining why this is safe — the connection is localhost-only, no network MITM path, and client auth is still enforced via certs

files/scripts/k8s_pod_control.sh
```diff
@@ -56,18 +65,59 @@ kubectl_retry() {
...
+    (( attempt++ ))
+  done
+}
+
 pods_on_node() {
-  kubectl_retry -n "${NS}" get pods \
-    --field-selector "spec.nodeName=${NODE_NAME}" \
-    -l "${POD_SELECTOR}" \
-    -o jsonpath='{range .items[*]}{.metadata.name}{" "}{.status.phase}{"\n"}{end}' || true
+  local sel_key="${POD_SELECTOR%%=*}"
```
make1980: nit: let's try to avoid uncommon acronym. so use selector_key or just key here. same for sel_val

files/scripts/k8s_pod_control.sh
```diff
@@ -56,18 +65,59 @@ kubectl_retry() {
...
+     select(.metadata.labels[$key] == $val) |
+     "\(.metadata.name) \(.status.phase)"' || true
 }
 
 pod_names_on_node() {
-  kubectl_retry -n "${NS}" get pods \
-    --field-selector "spec.nodeName=${NODE_NAME}" \
-    -l "${POD_SELECTOR}" \
-    -o jsonpath='{range .items[*]}{.metadata.name}{"\n"}{end}' || true
+  local sel_key="${POD_SELECTOR%%=*}"
```
make1980: can we refactor this with the function above? maybe split line 116 to first get pod json based on key/value and then parse the json further with different fields?

also we should add UT for this - i understand it's non-trivial but we can mock the output from curl command and verify the jq part only.

files/scripts/k8s_pod_control.sh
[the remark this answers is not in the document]
make1980: why is that? isn't the plan to have container image startup script to check feature status and stop?

files/scripts/k8s_pod_control.sh
```diff
@@ -109,22 +159,17 @@ kill_pods() {
...
-      2> >(systemd-cat -t "${SERVICE_NAME}-start" -p err)
-  else
-    # background + pipe to syslog via logger in case systemd-journald is masked/disabled
-    ( kill_pods ) \
-      > >(logger -t "${SERVICE_NAME}-start" -p user.info) \
-      2> >(logger -t "${SERVICE_NAME}-start" -p user.err)
-  fi &
-  disown
-  exit 0
+  # Re-invoke ourselves with the "restart" action under a hard 20s cap.
```
make1980, line 162: why do we change this?

make1980 (commented):
added comments

files/scripts/k8s_pod_control.sh
FengPan-Frank: Renamed sel_key/sel_val to key/val in both pods_on_node() and pod_names_on_node().

files/scripts/k8s_pod_control.sh
FengPan-Frank: Just identified below issue during test:
1. When parent exists (exit 0), those orphaned processes stayed in the cgroup. so on the next systemctl restart, systemd warned: "Found left-over process ... in control group while starting unit.".
2. Timeout risk, when we first made it synchronous (kill_pods directly), the retry logic (10 attempts × exponential backoff) could exceed the service's TimeoutStartSec=30s, causing ExecStartPre to be killed by systemd.

losha228 (dismissed):
LGTM

files/scripts/k8s_pod_control.sh
make1980: sorry I don't understand how restart will solve this - restart right now does synchronous kill so why don't we call synchronous kill directly here with some timeout? timeout command seems to work with bash function as well.

FengPan-Frank: timeout is an external binary which can only execute external commands, not shell functions. like below:
image

files/scripts/k8s_pod_control.sh
FengPan-Frank: added

make1980 (dismissed):

files/scripts/k8s_pod_control.sh
```diff
@@ -145,16 +190,27 @@ cmd_status() {
...
+      if (( ++round > max_rounds )); then
+        log "ERROR: timed out after ${max_rounds} rounds waiting for pods (ns=${NS}, selector=${POD_SELECTOR}) on ${NODE_NAME}"
+        exit 1
+      fi
+      sleep 20; continue
     fi
     if awk '$2=="Running"{found=1} END{exit found?0:1}' <<<"$out"; then
-      sleep 60
+      round=0
+      sleep 20
```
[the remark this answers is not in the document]
FengPan-Frank: keep cmd_wait looping

files/scripts/k8s_pod_control.sh
```diff
@@ -109,21 +159,16 @@ kill_pods() {
...
+  # If the API server is unreachable, timeout(1) kills the child so we
+  # stay well within the service's TimeoutStartSec (30s).
+  timeout 20 "${BASH_SOURCE[0]}" "${SERVICE_NAME}" restart 2>&1 \
+    | logger -t "${SERVICE_NAME}-start" || true
 }
 
+# stop is a no-op: K8s controls the pod lifecycle; deleting the pod here
+# would just cause the controller to recreate it immediately.
 cmd_stop()    { kill_pods; }
 cmd_restart() { kill_pods; }
```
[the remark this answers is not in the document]
FengPan-Frank: not stop actually, just make delete pod as simple operation.

make1980 (approved):

qiluo-msft (approved):

PR #25991, opened 2026-03-10
Fix DNS in containers by preserving resolv.conf symlink during image …

william8545:
#### Dependency
This PR depends on https://github.com/sonic-net/sonic-utilities/pull/4365. The other PR should be merged first before this one can be merged.

#### Why I did it

After installing SONiC 202511 from ONIE, 10 out of 15 docker containers have empty `/etc/resolv.conf` and no DNS resolution. This is a regression from 202412.

The Trixie base image upgrade introduced two lines in `build_debian.sh` that destroy the `/etc/resolv.conf` symlink (created by the `resolvconf` package) and replace it with a regular empty file:

```bash
sudo rm -f $FILESYSTEM_ROOT/etc/resolv.conf
sudo touch $FILESYSTEM_ROOT/etc/resolv.conf
```

This breaks the DNS propagation chain to docker containers because `/etc/resolvconf/update.d/libc` checks whether `/etc/resolv.conf` is a symlink to `/run/resolvconf/resolv.conf` before notifying downstream consumers (including `update-libc.d/update-containers`). When the symlink is missing, DHCP-obtained DNS is never propagated to containers.

##### Work item tracking

#### How I did it

Replaced `sudo touch` with `sudo ln -sf /run/resolvconf/resolv.conf` to preserve the symlink that the `resolvconf` package expects:

```bash
sudo rm -f $FILESYSTEM_ROOT/etc/resolv.conf
sudo ln -sf /run/resolvconf/resolv.conf $FILESYSTEM_ROOT/etc/resolv.conf
```

This is consistent with what `resolv-config.sh` does at runtime (`ln -sf /run/resolvconf/resolv.conf /etc/resolv.conf`) and matches the behavior of all SONiC releases prior to 202511.

#### How to verify it

1. Install from ONIE on a switch
2. After boot, verify:
   ```bash
   # Host resolv.conf should be a symlink
   ls -la /etc/resolv.conf
   # Expected: /etc/resolv.conf -> /run/resolvconf/resolv.conf

   # All containers should have DNS
   for c in $(docker ps --format '{{.Names}}'); do
     echo "=== $c ==="
     docker exec $c cat /etc/resolv.conf
   done
   ```

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [x] 202511

- [ ] 
- [ ]

### DIFF
--- a/build_debian.sh
+++ b/build_debian.sh
@@ -891,8 +891,6 @@ sudo mkdir -p $FILESYSTEM_ROOT/var/lib/docker
 ## Clear DNS configuration inherited from the build server
 sudo rm -f $FILESYSTEM_ROOT/etc/resolvconf/resolv.conf.d/original
 sudo cp files/image_config/resolv-config/resolv.conf.head $FILESYSTEM_ROOT/etc/resolvconf/resolv.conf.d/head
-sudo rm -f $FILESYSTEM_ROOT/etc/resolv.conf
-sudo touch $FILESYSTEM_ROOT/etc/resolv.conf
 
 ## Optimize filesystem size
 if [ "$BUILD_REDUCE_IMAGE_SIZE" = "y" ]; then


saiarcot895:
I think I'm missing something: if, at runtime, `resolv-config.sh` is recreating the symlink, and is running `resolvconf -u`, then why do some containers still not have a properly-populated `/etc/resolv.conf`?

william8545:
Because `update.d/libc` has two sequential gate checks with a file write between them.
The two gates are satisfied at different times but never in the same invocation — and the intermediate write permanently prevents the second invocation from reaching Gate 2.

The relevant code is in [`/etc/resolvconf/update.d/libc`](https://sources.debian.org/src/resolvconf/1.94/etc/resolvconf/update.d/libc) (resolvconf 1.94, Trixie). The execution order of the three critical sections:

1. **[L142-146](https://sources.debian.org/src/resolvconf/1.94/etc/resolvconf/update.d/libc#L142-146) — Gate 1: content-unchanged check**
   ```sh
   if [ -f "$DYNAMICRSLVCNFFILE" ] && [ "$(cat $TMPFILE)" = "$(cat $DYNAMICRSLVCNFFILE)" ] ; then
       rm -f "$TMPFILE"
       exit 0
   fi
   ```
   Compares newly generated content (`$TMPFILE`) with the existing `/run/resolvconf/resolv.conf`.
   If identical, exits.

2. **[L149](https://sources.debian.org/src/resolvconf/1.94/etc/resolvconf/update.d/libc#L149) — File write (between the two gates)**
   ```sh
   mv -f "$TMPFILE" "$DYNAMICRSLVCNFFILE"
   ```
   Persists the new content to `/run/resolvconf/resolv.conf` **before** the symlink check.
   Once written, all future invocations will see "unchanged" at Gate 1.

3. **[L153](https://sources.debian.org/src/resolvconf/1.94/etc/resolvconf/update.d/libc#L153) — Gate 2: symlink check**
   ```sh
   resolv_conf_is_symlinked_to_dynamic_file || exit 0
   ```
   Where [`resolv_conf_is_symlinked_to_dynamic_file`](https://sources.debian.org/src/resolvconf/1.94/etc/resolvconf/update.d/libc#L60) is:
   ```sh
   [ -L ${ETC}/resolv.conf ] && [ "$(readlink -m ${ETC}/resolv.conf)" = "$DYNAMICRSLVCNFFILE" ]
   ```
   If `/etc/resolv.conf` is not a symlink to `/run/resolvconf/resolv.conf`, exits without
   running `update-libc.d/` (which contains `update-containers`).

The ordering of L149 before L153 is the root of the deadlock: a Gate 2 failure commits the
content to disk, making any subsequent invocation fail at Gate 1 regardless of whether the
symlink has since been created.


### Containers start before DHCP — all read empty `/etc/resolv.conf`

```
21:16:31.356832  resolv-config: [update-containers:main] Updating resolv.conf for container database
21:16:37.943833  resolv-config: [update-containers:main] Updating resolv.conf for container restapi
21:16:39.007647  resolv-config: [update-containers:main] Updating resolv.conf for container swss
......
```

Each of these is `postStartAction` calling `update-containers $DOCKERNAME`. The script does
`RESOLV_CONTENT=$(cat /etc/resolv.conf)` — but `/etc/resolv.conf` is a regular empty file
(from `build_debian.sh`'s `touch`), so every container gets empty DNS written into it.

10 containers running with empty DNS. DHCP hasn't completed yet. The
`libc` → `update-containers` bulk update is supposed to fix them once DHCP finishes.


### DHCP completes at 21:16:53 — bulk update blocked by Gate 2

```
21:16:53.379431  dhclient[9279]: DHCPDISCOVER on eth0 to 255.255.255.255 port 67 interval 8
21:16:53.381622  dhclient[9279]: DHCPOFFER of 10.210.25.14 from 10.222.36.70
21:16:53.381658  dhclient[9279]: DHCPREQUEST for 10.210.25.14 on eth0 to 255.255.255.255 port 67
21:16:53.384519  dhclient[9279]: DHCPACK of 10.210.25.14 from 10.222.36.70
21:16:53.453073  dhclient[9279]: bound to 10.210.25.14 -- renewal in 10641 seconds.
```

dhclient hook calls `resolvconf -a eth0.dhclient` → triggers `update.d/libc`:

| Step | What happens | State |
|---|---|---|
| 1 | Generate new content: `nameserver 10.211.0.163`, `10.211.0.164`, `10.222.36.70` | |
| 2 (Gate 1) | Compare with `/run/resolvconf/resolv.conf` (empty at boot) → **different → PASS** | |
| 3 (write) | `mv $TMPFILE /run/resolvconf/resolv.conf` — **file now has DNS content** | `/run/.../resolv.conf` = DNS |
| 4 (Gate 2) | `[ -L /etc/resolv.conf ]` → **FALSE** (regular file from `touch`) → **FAIL → exit** | `/etc/resolv.conf` = empty regular file |
| 5 | never reached — `update-containers` not called | |

Evidence the bulk update did NOT fire: grepping `raw_logs/syslog.log` for `"Starting resolv.conf update for running containers"` returns **zero matches** between 21:16:53 and 21:18:01. Only individual `"Updating resolv.conf for container"` entries exist.

10 running containers still have empty DNS. `/run/resolvconf/resolv.conf` now has the
correct content, but it will never be different again — this is the latch that blocks all
future invocations at Gate 1.


### resolv-config.service runs at 21:18:01 — blocked by Gate 1

```
syslog.log:461    2026-03-10T21:16:26.581866  systemd[1]: Started hostcfgd.timer - Delays hostcfgd daemon until SONiC has started.
syslog.log:41100  2026-03-10T21:17:54.960809  systemd[1]: Started hostcfgd.service - Host config enforcer daemon.
syslog.log:41263  2026-03-10T21:18:01.062910  systemd[1]: Starting resolv-config.service - Update DNS configuration...
syslog.log:41274  2026-03-10T21:18:02.164828  systemd[1]: Finished resolv-config.service - Update DNS configuration.
```

`resolv-config.sh` executes:

```sh
# resolv-config.sh line 10 — creates the symlink
ln -sf /run/resolvconf/resolv.conf /etc/resolv.conf

# resolv-config.sh line 51 — triggers update.d/libc
resolvconf -u
```

`update.d/libc` runs again:

| Step | What happens |
|---|---|
| 1 | Generate new content: same `nameserver 10.211.0.163...` |
| 2 (Gate 1) | Compare with `/run/resolvconf/resolv.conf` — **identical** (written at 21:16:53) → **FAIL → exit** |
| 3 | never reached |
| 4 (Gate 2) | never reached (would PASS now — symlink exists) |
| 5 | never reached — `update-containers` not called |

10 containers still have empty DNS. The symlink now exists, but Gate 1's content-unchanged optimization exits before Gate 2 is ever evaluated.


### Late containers start after 21:18:02 — these get correct DNS

```
syslog.log:41742  2026-03-10T21:18:36.445352  resolv-config: [update-containers:main] Updating resolv.conf for container gnmi
syslog.log:41842  2026-03-10T21:18:39.153862  resolv-config: [update-containers:main] Updating resolv.conf for container lldp
syslog.log:42056  2026-03-10T21:18:41.989390  resolv-config: [update-containers:main] Updating resolv.conf for container mgmt-framework
syslog.log:42550  2026-03-10T21:18:48.188108  resolv-config: [update-containers:main] Updating resolv.conf for container snmp
syslog.log:44559  2026-03-10T21:18:51.034053  resolv-config: [update-containers:main] Updating resolv.conf for container what-just-happened
```

These start after 21:18:02 (when `resolv-config.service` finished and the symlink exists).
`cat /etc/resolv.conf` now follows the symlink to `/run/resolvconf/resolv.conf` and gets DNS
content. This is just per-container `postStartAction`, not the bulk update — that never fired.

saiarcot895:
So IIRC, the original intention of this change was to prevent the build server's DNS getting injected into the image. However, it looks like my initial attempt just removed the symlink itself, but didn't remove the actual file with the build server's DNS. Despite that, since the file in question is under `/run`, it would get hidden by a tmpfs mount.

This change is re-adding the symlink that was there before the `rm -f $FILESYSTEM_ROOT/etc/resolv.conf`. Given that, it might be better to just remove the `rm -f` and `touch` lines. The symlink will be there anyways.

william8545:
> So IIRC, the original intention of this change was to prevent the build server's DNS getting injected into the image. However, it looks like my initial attempt just removed the symlink itself, but didn't remove the actual file with the build server's DNS. Despite that, since the file in question is under `/run`, it would get hidden by a tmpfs mount.
> 
> This change is re-adding the symlink that was there before the `rm -f $FILESYSTEM_ROOT/etc/resolv.conf`. Given that, it might be better to just remove the `rm -f` and `touch` lines. The symlink will be there anyways.

Agree, those two lines are redundant now
It works properly without them.

william8545:
@saiarcot895 Please check this PR - https://github.com/sonic-net/sonic-utilities/pull/4365 too, thanks.

vmittal-msft:
@yxieca @rlhui please help merge

PR #26002, opened 2026-03-10
Support for BMC cards based on Aspeed 2720 - Phase 2

chander-nexthop:
- Address review comments to remove BMC component from platform json
      files
- We dont intend to use obmc-console, so remove the same from platform packages
- Removed the remnant setup.py
- Fixed bugs in watchdog code
- Moved aspeed services to a debian package instead of platform check + install in common code (build_debian.sh)
- Added helper scripts to create and write a disk image into eMMC
- show version now printed the chassis serial number too
- Usb interface renamed to bmc0
- bmc.json to specify BMC IP and Switch Host IP address
- eepromd runs, populates DB. show platform syseeprom would work (tested by populating some data into eeprom)

#### Why I did it
Changes based on review comments received and to cater to the platform HLD https://github.com/sonic-net/SONiC/pull/2215/

##### Work item tracking

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511

- [ ] 
- [ ]


yxieca (commented):
## Code Review

This PR is a cleanup/refactor for the Aspeed BMC platform in SONiC. Here's a breakdown:

### Changes Summary

**1. Platform services → proper Debian package** ✅ Good
- Moves scripts/services from ad-hoc `build_debian.sh` installation into a proper `aspeed-platform-services` Debian package (`.mk`, `debian/` scaffolding, `rules.mk`, `one-image.mk`)
- This is the right pattern — proper packaging vs. raw file copies in build scripts

**2. New watchdog keepalive daemon** (`watchdog-keepalive.sh`)
- 180s timeout, 60s keepalive interval
- Custom log rotation (5 files max)
- Uses `watchdogutil arm/disarm` for setup/teardown

**3. Watchdog platform API refactor** (both AST2700 EVB and NextHop)
- Removes `self.armed` tracking in favor of sysfs state (`is_armed()` reads sysfs `state`)
- Removes `_close_watchdog()` (magic 'V' close) — replaces with `_disablewatchdog()` using ioctl `WDIOS_DISABLECARD`
- Adds proper ioctl methods: `_enablewatchdog`, `_keepalive`, `_settimeout`, `_gettimeout`, `_gettimeleft`
- `arm()` now caps at 300s max — is this intentional? Some use cases may need longer timeouts

**4. Reboot cause logic improvement** (both chassis.py files)
- `WDIOF_CARDRESET` now checks for software reboot cause file before defaulting to watchdog
- Previously always returned `NON_HARDWARE`, now returns `WATCHDOG` when no software cause found — good fix

**5. Removed BMC component from platform.json** (both EVB and NextHop)
- Removes "BMC" from components list — why? Is BMC FW update not supported via `fwutil`?

**6. Removed obmc-console configs** (both EVB and NextHop)
- Deleted `server.tty.conf` files and related install logic — are these now managed elsewhere (OpenBMC side)?

**7. USB network init fix** (NextHop)
- Hardcodes UDC name to `12021000.usb-vhub:p1` instead of auto-detecting first available — could this break on different AST2700 boards?

**8. New eMMC installer script** (`install-sonic-to-emmc.sh`)
- Installs SONiC from OpenBMC to eMMC with U-Boot env setup
- Has commented-out alternatives and hardcoded `BOOTCONF="nexthop-b27-r0"` — should these be configurable rather than requiring source edits?

### Issues

**1. Duplicate code across EVB and NextHop**
- `chassis.py` reboot cause changes are identical in both `ast2700/` and `nexthop/common/`
- `watchdog.py` changes are identical in both
- Should this be a shared base class?

**2. Commented-out code left in**
- `build-emmc-image-installer.sh`: `#EMMC_SIZE_MB=7168`
- `install-sonic-to-emmc.sh`: `#BOOTCONF="ast2700-evb"`
- These should be removed or made configurable

**3. `watchdog-keepalive.sh` logs every 60s**
- `echo "Watchdog keepalive sent"` on every iteration generates ~1440 log lines/day — consider removing or making it periodic (every Nth keepalive)

**4. Deleted `ast2700/setup.py`** — is the platform wheel now built differently, or was this unused?

**5. PR title is vague** — "Address pending review comments and some cleanup" doesn't convey the scope. This has significant changes (new Debian package, watchdog refactor, reboot cause fix, eMMC installer).

chander-nexthop:
> ## Code Review
> This PR is a cleanup/refactor for the Aspeed BMC platform in SONiC. Here's a breakdown:
> 
> ### Changes Summary
> **1. Platform services → proper Debian package** ✅ Good
> 
> * Moves scripts/services from ad-hoc `build_debian.sh` installation into a proper `aspeed-platform-services` Debian package (`.mk`, `debian/` scaffolding, `rules.mk`, `one-image.mk`)
> * This is the right pattern — proper packaging vs. raw file copies in build scripts
> 
> **2. New watchdog keepalive daemon** (`watchdog-keepalive.sh`)
> 
> * 180s timeout, 60s keepalive interval
> * Custom log rotation (5 files max)
> * Uses `watchdogutil arm/disarm` for setup/teardown
> 
> **3. Watchdog platform API refactor** (both AST2700 EVB and NextHop)
> 
> * Removes `self.armed` tracking in favor of sysfs state (`is_armed()` reads sysfs `state`)
> * Removes `_close_watchdog()` (magic 'V' close) — replaces with `_disablewatchdog()` using ioctl `WDIOS_DISABLECARD`
> * Adds proper ioctl methods: `_enablewatchdog`, `_keepalive`, `_settimeout`, `_gettimeout`, `_gettimeleft`
> * `arm()` now caps at 300s max — is this intentional? Some use cases may need longer timeouts
> 
> **4. Reboot cause logic improvement** (both chassis.py files)
> 
> * `WDIOF_CARDRESET` now checks for software reboot cause file before defaulting to watchdog
> * Previously always returned `NON_HARDWARE`, now returns `WATCHDOG` when no software cause found — good fix
> 
> **5. Removed BMC component from platform.json** (both EVB and NextHop)
> 
> * Removes "BMC" from components list — why? Is BMC FW update not supported via `fwutil`?

This platform itself is a BMC and makes little sense to have a BMC component within it was the point Judy made in the code review. That made sense and hence removed it.

> 
> **6. Removed obmc-console configs** (both EVB and NextHop)
> 
> * Deleted `server.tty.conf` files and related install logic — are these now managed elsewhere (OpenBMC side)?

We no longer use obmc-console, instead we have moved to use SONiC's native consutil service and hence removed this.

> 
> **7. USB network init fix** (NextHop)
> 
> * Hardcodes UDC name to `12021000.usb-vhub:p1` instead of auto-detecting first available — could this break on different AST2700 boards?

This is under the Nexthop platform module and is specific to our card. Will not affect the rest.

> 
> **8. New eMMC installer script** (`install-sonic-to-emmc.sh`)
> 
> * Installs SONiC from OpenBMC to eMMC with U-Boot env setup
> * Has commented-out alternatives and hardcoded `BOOTCONF="nexthop-b27-r0"` — should these be configurable rather than requiring source edits?

Fair point. As these scripts are helper scripts (not part of the image build) I didn't make them configurable. If we really need that I can make it as such. Do let me know.

> 
> ### Issues
> **1. Duplicate code across EVB and NextHop**
> 
> * `chassis.py` reboot cause changes are identical in both `ast2700/` and `nexthop/common/`
> * `watchdog.py` changes are identical in both
> * Should this be a shared base class?

If we make this a shared base class, then we will need a new intermediate python wheel package. Also, chassis.py will change for nexthop card to accommodate eeprom and other features. As of today there are same, but expected to diverge.

> 
> **2. Commented-out code left in**
> 
> * `build-emmc-image-installer.sh`: `#EMMC_SIZE_MB=7168`
> * `install-sonic-to-emmc.sh`: `#BOOTCONF="ast2700-evb"`
> * These should be removed or made configurable

Eval board ships with a 8G eMMC whereas Nexthop card will ship with a 32G (but 10G due to pSLC). To catrer
to both these cards, I just added lines which can be uncommented depending on which cards these scripts target.
I thought as these are helper scripts, there might be more additions depending vendor requirements. So instead of making this configurable I just added comments. I can fix them if you insist.
> 
> **3. `watchdog-keepalive.sh` logs every 60s**
> 
> * `echo "Watchdog keepalive sent"` on every iteration generates ~1440 log lines/day — consider removing or making it periodic (every Nth keepalive)
> 
Thanks for pointing out, Will fix.

> **4. Deleted `ast2700/setup.py`** — is the platform wheel now built differently, or was this unused?

yes it was unused.

> 
> **5. PR title is vague** — "Address pending review comments and some cleanup" doesn't convey the scope. This has significant changes (new Debian package, watchdog refactor, reboot cause fix, eMMC installer).

Will fix the title.

platform/aspeed/aspeed-platform-services/systemd/watchdog-keepalive.service
```diff
@@ -0,0 +1,17 @@
+[Unit]
+Description=Watchdog Keepalive Daemon
+Documentation=man:watchdog(8)
+After=multi-user.target
+Before=shutdown.target
+
+[Service]
+Type=simple
+ExecStart=/usr/bin/watchdog-keepalive.sh
```
judyjoseph, line 9: do we need infra change to copy the watchdog script from platform/aspeed/aspeed-platform-services/scripts/watchdog-keepalive.sh to /usr/bin -- as I see some line relating to this removed in build_debian.sh‎ ?

chander-nexthop: No, we now have a separate debian package, aspeed-platform-services. Its install rules takes care of this. Earlier we were doing it from build_debian.sh directly with a platform check. That was a stop gap.

device/nexthop/arm64-nexthop_b27-r0/bmc.json
```diff
@@ -0,0 +1,7 @@
+{
```
judyjoseph, line 1: we would not need this file .. Can you check this PR https://github.com/sonic-net/sonic-buildimage/pull/26544,  we define this file globally in sonic land and trying to use it : **files/image_config/constants/bmc.json**

platform/aspeed/aspeed-platform-services/scripts/watchdog-keepalive.sh
```diff
@@ -0,0 +1,80 @@
...
+# Watchdog keepalive daemon for Aspeed platforms.
+#
+# This daemon keeps the hardware watchdog alive by sending periodic keepalive
+# signals. It opens /dev/watchdog and writes 'w' character every 60 seconds
+# (with a 180 second watchdog timeout).
+#
+
+WATCHDOG_DEVICE="/dev/watchdog0"
+KEEPALIVE_INTERVAL=60  # seconds
+LOG_FILE="/var/log/aspeed-watchdog.log"
```
judyjoseph: In BMC we are planning to keep /var/log in tmpfs .. please again refer : https://github.com/sonic-net/sonic-buildimage/pull/26544 ... is that fine if we loose these watchdog logs ? else we need to move to a persistent storage OR put any critical failure/restart logs into /host/bmc/event.log itself ?

platform/aspeed/sonic-platform-modules-nexthop/common/scripts/usb-network-init.sh
```diff
@@ -109,6 +109,39 @@ ip link set "${INTERFACE_NAME}" up
 # Enable IPv6 on the interface
 sysctl -w net.ipv6.conf.${INTERFACE_NAME}.disable_ipv6=0 2>/dev/null || true
 
+# Step 6b: Add static IPv4 for gNOI communication with switch host
+# Read configuration from bmc.json
```
judyjoseph: The ip address config can we let it be done in interface-config systemd service ? https://github.com/sonic-net/sonic-buildimage/blob/4c3e98903cef7cee91022c363dae364219309e44/files/image_config/interfaces/interfaces.j2#L33

platform/aspeed/sonic-platform-modules-nexthop/common/systemd/sonic-usb-network-udev-init.service
```diff
@@ -0,0 +1,16 @@
+[Unit]
+Description=Aspeed USB Network Gadget udev Rules Initialization
+Documentation=https://github.com/sonic-net/sonic-buildimage
+After=local-fs.target
+Before=usb-network-init.service
+
+[Service]
+Type=oneshot
+ExecStart=/usr/local/bin/sonic-usb-network-udev-init.sh
```
judyjoseph, line 9: Assume this systemd service is run much before interfaces-config service where we configure ip address on usb interface

platform/aspeed/sonic-platform-modules-nexthop/common/sonic_platform/chassis.py
```diff
@@ -56,6 +58,14 @@ def __init__(self):
         # Initialize watchdog (same as base class)
         self._watchdog = Watchdog()
 
+        # Initialize eeprom
+        self._eeprom = Eeprom()
+
+        # Initialize Switch Host Module (x86 CPU managed by BMC)
+        self._module_list = []
+        switch_host = SwitchHostModule(module_index=0)
```
judyjoseph, line 66: yes Either we can just have the SwitchHost alone as module ... Do you think will be good to have BMC also modelled as another module ? Refer in pmon desgin doc : https://github.com/judyjoseph/SONiC/blob/ed7cae46d77df16b3649d415a21ab4149cb03d73/doc/bmc/sonicBMC/pmon-bmc-design.md#224-platform-apis

More details here : https://github.com/sonic-net/sonic-platform-common/pull/634

This was we can implement API's like admin_state(up/down) .. get_serial_number() etc for BMC as a module .. is.o doing it in chassis.py -- which is for the device as a whole  -- let me know your feedback

chander-nexthop: I think the BMC access methods should lie within chassis.py like in other switch platforms. From SONiC running on the BMC, treating "self" as chassis makes sense and follows existing conventions imho, instead of treating "self" as another component.

device/nexthop/arm64-nexthop_b27-r0/bmc.json
chander-nexthop: I was under the impression the platform monitoring HLD allows for this bmc.json and hence took this approach.

platform/aspeed/aspeed-platform-services/scripts/watchdog-keepalive.sh
chander-nexthop: Will use tmpfs for /var/log and move the watchdog to /host/bmc/watchdog.log

device/nexthop/arm64-nexthop_b27-r0/bmc.json
chander-nexthop: I am still using it as src/sonic-config-engine/sonic-cfggen, get_bmc_data, reads this to  update DEVICE_METADATA|bmc and once thats done, files/image_config/interfaces/interfaces.j2 and files/image_config/interfaces/interfaces-config.sh use that to create /etc/network/interfaces.

platform/aspeed/aspeed-platform-services/scripts/watchdog-keepalive.sh
chander-nexthop: Done.

platform/aspeed/sonic-platform-modules-nexthop/common/scripts/usb-network-init.sh
chander-nexthop: Done.

platform/aspeed/sonic-platform-modules-nexthop/common/systemd/sonic-usb-network-udev-init.service
chander-nexthop: Yes, checked and verfied.

device/nexthop/arm64-nexthop_b27-r0/bmc.json
chander-nexthop: Moved to this approach in a20c00989acdf800a6749189416e1c7dbb5084ad

platform/aspeed/aspeed-platform-services/scripts/watchdog-keepalive.sh
```diff
@@ -0,0 +1,98 @@
...
+
+trap 'cleanup' SIGTERM SIGINT
+
+# Rotate logs
+rotate_logs
+
+# Start logging
+echo "$(date -u): Watchdog keepalive daemon starting" > "$LOG_FILE"
+
+# Set watchdog interval to 180 seconds
```
roger-nexthop, line 62: Delete this line, timeout is also mentioned below

platform/aspeed/sonic-platform-modules-ast-evb/ast2700/sonic_platform/watchdog.py
```diff
@@ -167,54 +147,65 @@ def is_armed(self):
...
+
+    def get_remaining_time(self):
+        """
+        Get the number of seconds remaining on the watchdog timer
+
+        Returns:
+            An integer specifying the number of seconds remaining on the
+            watchdog timer. If the watchdog is not armed, returns -1.
+        """
+        return -1
```
roger-nexthop, line 210: Are we able to get the value if watchdog is armed here?

device/nexthop/arm64-nexthop_b27-r0/pmon_daemon_control.json
```diff
@@ -3,9 +3,10 @@
...
     "skip_xcvrd": true,
     "skip_psud": true,
-    "skip_syseepromd": true,
+    "skip_syseepromd": false,
     "skip_pcied": true,
     "skip_chassisd": true,
+    "skip_chassis_db_init": false,
     "skip_fancontrol": true,
     "include_sensormond": false,
     "thermalctld": {
```
judyjoseph, line 12: @chander-nexthop  can you add this timer "liquid_cooling_update_interval" in this file for your platform 

https://github.com/sonic-net/sonic-buildimage/blob/600ee33467f5d2c06cd00ba61cbd95b0ef8e2408/dockers/docker-platform-monitor/docker-pmon.supervisord.conf.j2#L234

roger-nexthop: We will take this up in a follow-up PR.

judyjoseph (approved):
Few minor comments as below  -- But can be taken up in a follow up PR 
> Need not define a separate device/nexthop/arm64-nexthop_b27-r0/bmc.json in platform/hwsku 
> Add "liquid_cooling_update_interval" in this pmon_daemon.json file

device/nexthop/arm64-nexthop_b27-r0/bmc.json
```diff
@@ -0,0 +1,5 @@
...
+    "bmc_if_name": "bmc0",
+    "bmc_if_addr": "192.168.100.1",
```
Blueve, line 3: I have concern about using this IP range as it could conflicting with real IP that used by mgmt port in lab/si/bsl/prod scenario.

Blueve: I recommend to pick a subnet from IPv4 Link-Local — 169.254.0.0/16 (RFC 3927)

(I remember docker, dhcp server, dpu uses above range, so we just need to make sure not conflicting with these SONiC inside scenarios)

Blueve:
@roger-nexthop can you help check the IP assignment?
According to doc: https://github.com/sonic-net/SONiC/blob/master/doc/bmc/sonicBMC/pmon-bmc-design.md#213-host-bmc-link

We should use 169.254.**100**.0/30

chander-nexthop:
> @roger-nexthop can you help check the IP assignment? According to doc: https://github.com/sonic-net/SONiC/blob/master/doc/bmc/sonicBMC/pmon-bmc-design.md#213-host-bmc-link
> 
> We should use 169.254.**100**.0/30

Yes this is being fixed locally in our repo. Will raise a PR today for upstream.

chander-nexthop:
@Blueve Fixed via https://github.com/sonic-net/sonic-buildimage/pull/27227
Can you take a look, approve and merge? Many thanks in advance,

PR #26137, opened 2026-03-12
[arp_update] Remove PID from arp_update logger output

Janetxxx:
#### Why I did it
arp_update logs for static routes showed entries like:
    98 static route nexthop not resolved ...
which could be misinterpreted as 98 unreachable servers. 

This change modifies the logger() function to remove PID argument.

##### Work item tracking

#### How I did it
- PID in `arp_update` logs is not needed.
- Keeping only tag makes logs cleaner while preserving filterability (`-t arp_update`).

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [x] 202511

- [ ] 
- [ ]

### DIFF
--- a/files/scripts/arp_update
+++ b/files/scripts/arp_update
@@ -248,9 +248,11 @@ while /bin/true; do
       ip="$( cut -d ':' -f 3- <<< "$neigh" )"
       if [[ $intf == *"Vlan"* ]]; then
           if [[ $ip == *"."* ]] && [[ ! $KERNEIGH4 =~ "${ip},${intf}" ]]; then
+              logger "mismatch arp entry, pinging ${ip} on ${intf}"
               run_with_retry_cmd "db/kernel mismatch ping to ${ip} on ${intf}" \
                 timeout 0.2 ping -I "$intf" -n -q -i 0 -c 1 -W 1 "$ip" >/dev/null 2>&1
           elif [[ $ip == *":"* ]] && [[ ! $KERNEIGH6 =~ "${ip},${intf}" ]]; then
+              logger "mismatch v6 nbr entry, pinging ${ip} on ${intf}"
               run_with_retry_cmd "db/kernel mismatch ping6 to ${ip} on ${intf}" \
                 timeout 0.2 ping6 -I "$intf" -n -q -i 0 -c 1 -W 1 "$ip" >/dev/null 2>&1
           fi


prsunny:
@Janetxxx , we had some logging issues with arp_update in the past. @theasianpianist , please review.

files/scripts/arp_update
```diff
@@ -11,7 +11,8 @@ ARP_UPDATE_VARS_FILE="/usr/share/sonic/templates/arp_update_vars.j2"
 
 # Overload `logger` command to include arp_update tag
 logger () {
-    command logger -i "$$" -t "arp_update" "$@"
+    local pid="$$"
+    command logger -i -t "arp_update" "[$pid]" "$@"
```
banidoru: **Suggestion (non-blocking):** Consider using `--id=$$` instead of `-i` to place the *script's* PID in the syslog PID field:
```suggestion
    command logger --id=$$ -t "arp_update" "$@"
```
`-i` logs the PID of the transient `logger` child process (a different value each call), which isn't useful for correlation. `--id=$$` would put the script's PID in the standard syslog PID field (e.g. `arp_update[98]: static route nexthop not resolved ...`), making the `[$pid]` prefix in the message body redundant.

This is entirely optional — the current code works correctly.

r12f (approved):

files/scripts/arp_update
theasianpianist: @Janetxxx I think we can actually remove the `-i`/`--id` argument entirely, there's no benefit to having the script PID included with the log message.

banidoru (approved):
Re-review (iteration 2): All prior review threads have been resolved. The updated change cleanly removes `-i "$$"` from the logger wrapper, eliminating the spurious PID from log messages while preserving `$@` passthrough for caller options (e.g., `-p priority`). No new issues found. LGTM.

banidoru (approved):
Clean fix. Removing `-i "$$"` eliminates the spurious PID prefix in the log message while preserving `$@` passthrough for caller options like `-p`. All prior review threads have been addressed and resolved. No new issues found.

files/scripts/arp_update
```diff
@@ -251,11 +248,9 @@ while /bin/true; do
       ip="$( cut -d ':' -f 3- <<< "$neigh" )"
       if [[ $intf == *"Vlan"* ]]; then
           if [[ $ip == *"."* ]] && [[ ! $KERNEIGH4 =~ "${ip},${intf}" ]]; then
-              logger "mismatch arp entry, pinging ${ip} on ${intf}"
```
prsunny, line 254: @Janetxxx , i think we can keep these logs since its not frequent and we would need this for some debugging. @theasianpianist

files/scripts/arp_update
```diff
@@ -97,9 +97,6 @@ while /bin/true; do
                   fi
                   intf_up=$(ip link show $interface | grep "state UP")
                   if [[ -n "$intf_up" ]]; then
-                      if [[ -z $(echo ${neigh_state} | grep 'STALE') ]]; then
```
prsunny, line 100: lgtm

prsunny (approved):
lgtm, @theasianpianist please re-review and signoff

prsunny:
> ## Review
> **Change 1 — PID removal from logger**: ✅ LGTM. The original `logger -i "$$"` caused the shell PID to appear as the first word of every log message (e.g., `98 static route nexthop not resolved...`), which is confusing. Removing both `-i` and `"$$"` is the right call — the `-t arp_update` tag is sufficient for filtering.
> 
> **Change 2 — Removing the diagnostic log line**: ⚠️ This contradicts @prsunny's feedback to keep the logs. The removed line:
> 
> ```shell
> if [[ -z $(echo ${neigh_state} | grep 'STALE') ]]; then
>     logger "static route nexthop not resolved ($neigh_state), pinging $nexthop on $interface"
> fi
> ```
> 
> provides useful diagnostic info — it tells you the neighbor state (INCOMPLETE, REACHABLE, DELAY, etc.) when a static route nexthop needs resolution. The `run_with_retry_cmd` below only logs on retry/failure, so without this line you lose visibility into _why_ pings are being sent.
> 
> **Suggestion**: Keep the log line but without the PID confusion (which is already fixed by Change 1). The log will now cleanly show:
> 
> ```
> arp_update: static route nexthop not resolved (INCOMPLETE), pinging 10.0.0.1 on Ethernet0
> ```
> 
> instead of the old confusing:
> 
> ```
> arp_update: 98 static route nexthop not resolved (INCOMPLETE), pinging 10.0.0.1 on Ethernet0
> ```
> 
> **TL;DR**: Please restore the 3 removed lines (the diagnostic log) per @prsunny's request. The PID fix alone already addresses the original issue.

@StormLiangMS , looks like this comment was for the other previous commit that had other log messages.  Removing the "static route nexthop not resolved" is expected as this is causing unnecessary logs even in a normal scenario.

PR #26151, opened 2026-03-12
Pass FIB suppression flag and to orchagent to control APPL_STATE_DB and pub/sub

mike-dubrovsky:
#### Why I did it

Route download speed degraded after BGP prefix suppression was introduced. RouteOrch::publishRouteState() unconditionally writes to APPL_STATE_DB and sends a pub/sub notification on every route add/remove, even when FIB suppression is disabled and no consumer needs the data.

##### Work item tracking

#### How I did it
This PR is part of a multi-repo change to make this a startup-time flag.

orchagent.sh now:

Reads suppress-fib-pending from CONFIG_DB and passes -F to orchagent when enabled (same pattern as synchronous_mode)

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511

- [ ] 
- [ ] 

This change needs to go together with 
https://github.com/sonic-net/sonic-swss/pull/4333
https://github.com/sonic-net/sonic-utilities/pull/4361
https://github.com/sonic-net/sonic-mgmt/pull/22916
https://github.com/sonic-net/SONiC/pull/2335

### DIFF
--- a/dockers/docker-orchagent/orchagent.sh
+++ b/dockers/docker-orchagent/orchagent.sh
@@ -108,6 +108,11 @@ if [ "$ASYNC_SWSS_REC" == "enabled" ]; then
     ORCHAGENT_ARGS+="-A "
 fi
 
+SUPPRESS_FIB_CONFIG=`sonic-db-cli CONFIG_DB hget "DEVICE_METADATA|localhost" "suppress-fib-pending"`
+if [ "$SUPPRESS_FIB_CONFIG" == "enabled" ]; then
+    ORCHAGENT_ARGS+="-F "
+fi
+
 # Add platform specific arguments if necessary
 if [ "$platform" == "broadcom" ]; then
     ORCHAGENT_ARGS+="-m $MAC_ADDRESS"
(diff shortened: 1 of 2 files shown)


dockers/docker-orchagent/orchagent.sh
```diff
@@ -71,6 +71,11 @@ if [ "$ASYNC_SWSS_REC" == "enabled" ]; then
     ORCHAGENT_ARGS+="-A "
 fi
 
+SUPPRESS_FIB_CONFIG=`sonic-db-cli CONFIG_DB hget "DEVICE_METADATA|localhost" "suppress-fib-pending"`
```
amitg90, line 111: Can this flag suppress-fib-pending change at run time, 
if yes what is the expectation here, Do we need to restart swss to pick up this new change in config DB ?

amitg90 (commented):
Please explain the expected behavior when this flag changes in CONFIG_DB.

dockers/docker-orchagent/orchagent.sh
mike-dubrovsky: Our current design requires a switch reload before the flag becomes operational. The companion sonic-utilities PR #[4361](https://github.com/sonic-net/sonic-utilities/pull/4361)  adds a CLI message informing the user of this.

deepak-singhal0408:
> Please explain the expected behavior when this flag changes in CONFIG_DB.

this was discussed in routing-wg and conclusion on behavior is captured here https://github.com/sonic-net/sonic-swss/pull/4333#issuecomment-4985864254  and in HLD PR.

deepak-singhal0408 (dismissed):

deepak-singhal0408 (approved):

PR #26157, opened 2026-03-12
[Arista] Update platform library submodules for 202505

ccroy-arista:
#### Why I did it
Introduce several fixes/improvements for Arista fan and temperature drivers, primarily for the Arista-7060X6 platforms.

##### Work item tracking

#### How I did it
Updated reference to arista drivers repository.

#### How to verify it
Flash a device with the newer image and confirm fan and temperature readings/behavior are as expected.

202505

Includes 6 changes:
- minke-fan-cpld: Avoid logging 'unknown fan id' when no fan is present
- moby: change target temp of TH5 diodes
- minke-fan-cpld: separate slot and fan initialization in cpld_init
- minke-fan-cpld: support partial fan insertion
- temp: Update TH5 remote diode ideality register
- quicksilver: enable TH5 external diodes


davidm-arista:
Removing 202511 request.  This PR is based on the 202505 branch.

ccroy-arista:
> LGTM.
> 
> Also @ccroy-arista can you confirm if these changes #24631 are already in 202505 image which we use for Moby ? esp the change to start using the incPID algorithm from Samuel

@judyjoseph are you referring to the 202505 image that MSFT is using? Arista does not have that information; that would have to come from MSFT.

PR #26165, opened 2026-03-13
[build] Fix sonic-mgmt-common build failure due to libyang-dev conflict

xq9mend:
#### Why I did it

`sonic-mgmt-common` depends on `libyang1-dev` but was building with `libyang3-dev` headers installed, causing a compile error:

```
error: too many arguments to function 'ly_set_log_clb'
```

`libyang1-dev` and `libyang3-dev` both provide the `libyang-dev` package and overwrite each other. PR #25427 introduced `LIBYANG3-DEV-DONE` as a phony gate to serialize this ordering, but `sonic-mgmt-common` was not wired to it, allowing it to build before `libyang1-dev` was properly installed.

#### How I did it

Added `$(LIBYANG3-DEV-DONE)` as a phony dependency of `sonic-mgmt-common`, ensuring it only builds after the libyang3 phase completes and libyang1-dev takes over — consistent with the approach introduced in PR #25427.

#### How to verify it

Build `target/sonic-vs.img.gz` from a clean state without build cache. Previously failed at `sonic-mgmt-common` with a libyang API mismatch; now passes.

None (master only)


xq9mend:
@bhouse-nexthop Could you review this? It follows the same pattern introduced in your PR #25427 — wiring `sonic-mgmt-common` to `LIBYANG3-DEV-DONE` to fix a libyang-dev conflict on clean builds.

bhouse-nexthop (commented):
In theory this shouldn't be needed.  `$(SONIC_MGMT_COMMON)` depends on `$(LIBYANG_DEV) $(LIBYANG)`, and `$(LIBYANG)` has `$(LIBYANG)_PHONIES += $(LIBYANG3-DEV-DONE)` so I don't see how `$(LIBYANG)` could be satisfied if `$(LIBYANG3-DEV-DONE)` isn't.

Can you tell me how to reliably reproduce this?  

The only way I've seen it reproduce is if there is a build failure in a package later down the chain.  Then the user issues a rebuild.  When this occurs it rebuilds the containers and the dependency tracking for this is utterly broken (and its not just libyang related).

I'm working on killing libyang1 with a bunch of PRs.

PR #26208, opened 2026-03-16
[202511] Flush ASIC DB on reboot

DavidZagury:
#### Why I did it
To fix https://github.com/sonic-net/sonic-buildimage/issues/25193

##### Work item tracking

#### How I did it
Flush ASIC_DB to prevent stale data issues on syncd restart
This ensures that if syncd restarts before swss, it won't read stale lane maps

#### How to verify it
Cause crash of syncd by killing the syncd process.
Verify no errors are raised in the log.

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511

- [ ] 
- [ ]

### DIFF
--- a/files/scripts/syncd_common.sh
+++ b/files/scripts/syncd_common.sh
@@ -173,5 +173,14 @@ stop() {
 
     stopplatform2
 
+    # Flush ASIC_DB to prevent stale data issues on syncd restart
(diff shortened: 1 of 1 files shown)


liat-grozovik:
some comments were added to the master PR. once they are solved and agreed we should update this PR and then merge it

PR #26209, opened 2026-03-16
Support empty DEFAULT_PASSWORD for SONiC installer

xq9mend:
## Summary

Sets the default password for the `admin` account to empty string (no password) for installer builds.

## Behavior

- **Console login**: works without a password — type `admin` and press Enter to get a shell
- **SSH login**: blocked by default (`PermitEmptyPasswords no` in sshd config) — user must set a password first
- **`/etc/shadow`**: admin entry has an empty password field (`admin::...`)

## Motivation

Previously the default password `YourPaSsWoRd` was a placeholder that users had to know and change. An empty password is more explicit — it signals "not set yet" and forces the operator to set a real password before SSH becomes usable. This improves out-of-box experience for lab/test setups where console access is the first point of entry.

## Testing

Tested on KVM VS image (`x86_64-kvm_x86_64-r0`):

- Console login with empty password: works
- `/etc/shadow` confirms empty password field
- SSH login blocked as expected

### DIFF
--- a/src/sonic-yang-models/yang-models/sonic-kdump.yang
+++ b/src/sonic-yang-models/yang-models/sonic-kdump.yang
@@ -62,7 +62,7 @@ module sonic-kdump {
 
                 leaf ssh_string {
                     type string {
-                        pattern "([a-zA-Z_][a-zA-Z0-9._%+-]*@([a-zA-Z0-9.-]+|[0-9]{1,3}(\\.[0-9]{1,3}){3}))";
+                        pattern "([a-zA-Z0-9._%+-]+@([a-zA-Z0-9.-]+|[0-9]{1,3}(\\.[0-9]{1,3}){3}))";
                     }
                     description  "Remote ssh connection string";
 


xq9mend:
@Yarden-Z could you please review this PR? Thanks!

check_install.py
```diff
@@ -44,6 +44,15 @@ def main():
         if i == 0:
             # send user name
             p.sendline(args.u)
+            # If password is empty, the shell prompt may appear directly (no Password: prompt)
```
saiarcot895: Even if the shell prompt appears after entering the username, this should be handled by line 43, and shouldn't need any additional code.

xq9mend: Good catch, you are right. The outer loop already handles `cmd_prompt` (i==3) on the next iteration. Removed the redundant block.

rules/config
```diff
@@ -75,7 +75,9 @@ CHANGE_DEFAULT_PASSWORD ?= n
 DEFAULT_USERNAME = admin
 
 # DEFAULT_PASSWORD - default password for installer build
-DEFAULT_PASSWORD = YourPaSsWoRd
+# Empty string means admin account has no password; console login works,
+# SSH login is blocked (PermitEmptyPasswords no). Set a real password on first login.
+DEFAULT_PASSWORD =
```
Yarden-Z: Can we leave the default password with YourPaSsWoRd?
Most automatons today rely on the ssh connection with the default password, changing this basically changes the default behavior of SONiC enforcing everyone to adjust the default compile config for the Debian config causing more fork outs than currently today.
Leaving the default password but allowing also an empty password allows to accommodate this request while maintaining backwards compatibility

rules/config
```diff
@@ -127,7 +130,7 @@ ENABLE_ORGANIZATION_EXTENSIONS = y
 # DEFAULT_KERNEL_PROCURE_METHOD - default method for obtaining kernel
 #   build:    build kernel from source
 #   download: download pre-built kernel from Azure storage.
-DEFAULT_KERNEL_PROCURE_METHOD = build
+DEFAULT_KERNEL_PROCURE_METHOD = download
```
saiarcot895: Why did this get changed here?

rules/config
xq9mend: Good point. Restored `DEFAULT_PASSWORD = YourPaSsWoRd` to maintain backwards compatibility. The infrastructure to handle empty `DEFAULT_PASSWORD` is still in place for users who explicitly opt in. A separate PR (#26487) proposes actually changing the default for those who want the improved security behavior.

rules/config
xq9mend: It was an unintentional leak of a local build optimization (`DEFAULT_KERNEL_PROCURE_METHOD = download`). Reverted — the value is now correctly `build` matching the upstream default.

rules/config
maxtara: I'd prefer we change the default behavior, and create a secure-default in this project. 
People should be able to opt in to insecure behavior, not opt out.

PR #26244, opened 2026-03-18
Fix sidecar rollout issue

FengPan-Frank:
#### Why I did it
Fix sidecar rollout issue

##### Work item tracking

#### How I did it
1. Place each k8s_pod_control.sh to dedicated repo for each sidecar, to avoid potential overwrite issue when multiple sidecar co-exist.
2. Update cmd_wait with no-op but sleep, so that systemd service is still active but not triggering kubelet to get pod status.

#### How to verify it
Verified by telemetry service.

XXXXXX@XXXXXXXX:~ sudo ls /usr/share/sonic/scripts/docker-telemetry-sidecar/
k8s_pod_control.sh
XXXXXX@XXXXXXXX:~ sudo systemctl status telemetry.service 
● telemetry.service - Telemetry container
     Loaded: loaded (/lib/systemd/system/telemetry.service; static)
    Drop-In: /etc/systemd/system/telemetry.service.d
             └─auto_restart.conf
     Active: active (running) since Wed 2026-03-18 12:43:12 UTC; 23s ago
    Process: 4028247 ExecStartPre=/usr/local/bin/telemetry.sh start (code=exited, status=0/SUCCESS)
   Main PID: 4028267 (k8s_pod_control)
      Tasks: 2 (limit: 38299)
     Memory: ~2M
     CGroup: /system.slice/telemetry.service
             ├─4028267 /bin/bash /usr/share/sonic/scripts/docker-telemetry-sidecar/k8s_pod_control.sh telemetry wait
             └─4028xxx sleep 300

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511

- [ ] 
- [ ]

### DIFF
--- a/dockers/docker-restapi-sidecar/systemd_stub.py
+++ b/dockers/docker-restapi-sidecar/systemd_stub.py
@@ -108,6 +108,10 @@ POST_COPY_ACTIONS = {
         ["sudo", "systemctl", "daemon-reload"],
         ["sudo", "systemctl", "restart", "monit"],
     ],
+    "/usr/share/sonic/scripts/docker-restapi-sidecar/k8s_pod_control.sh": [
+        ["sudo", "systemctl", "daemon-reload"],
+        ["sudo", "systemctl", "restart", "restapi"],
+    ],
 }
 
 
@@ -166,7 +170,7 @@ def ensure_sync() -> bool:
     sync_items_list: List[SyncItem] = [
         SyncItem(restapi_src, "/usr/bin/restapi.sh", mode=0o755),
         SyncItem(container_checker_src, "/bin/container_checker", mode=0o755),
-        SyncItem("/usr/share/sonic/scripts/k8s_pod_control.sh", "/usr/share/sonic/scripts/k8s_pod_control.sh", mode=0o755),
+        SyncItem("/usr/share/sonic/scripts/k8s_pod_control.sh", "/usr/share/sonic/scripts/docker-restapi-sidecar/k8s_pod_control.sh", mode=0o755),
         SyncItem(container_restapi_service, HOST_RESTAPI_SERVICE, mode=0o644),
     ]
     return sync_items(sync_items_list, POST_COPY_ACTIONS)
--- a/dockers/docker-restapi-sidecar/cli-plugin-tests/test_systemd_stub.py
+++ b/dockers/docker-restapi-sidecar/cli-plugin-tests/test_systemd_stub.py
@@ -212,7 +212,7 @@ def test_sync_no_change_fast_path(ss):
     # Put same files on host
     host_fs["/usr/bin/restapi.sh"] = b"same"
     host_fs["/bin/container_checker"] = b"same"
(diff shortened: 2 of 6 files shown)


dockers/docker-restapi-sidecar/systemd_stub.py
```diff
@@ -108,6 +108,10 @@ def _get_branch_name() -> str:
         ["sudo", "systemctl", "daemon-reload"],
         ["sudo", "systemctl", "restart", "monit"],
     ],
+    "/usr/share/sonic/scripts/docker-restapi-sidecar/k8s_pod_control.sh": [
+        ["sudo", "systemctl", "daemon-reload"],
+        ["sudo", "systemctl", "restart", "restapi"],
```
make1980, line 113: maybe we shouldn't restart restapi - when k8s deployment happens this code is executed and at that time there is no need to invoke systemctl restart

dockers/docker-restapi-sidecar/systemd_stub.py
```diff
@@ -166,7 +170,7 @@ def ensure_sync() -> bool:
     sync_items_list: List[SyncItem] = [
         SyncItem(restapi_src, "/usr/bin/restapi.sh", mode=0o755),
         SyncItem(container_checker_src, "/bin/container_checker", mode=0o755),
-        SyncItem("/usr/share/sonic/scripts/k8s_pod_control.sh", "/usr/share/sonic/scripts/k8s_pod_control.sh", mode=0o755),
+        SyncItem("/usr/share/sonic/scripts/k8s_pod_control.sh", "/usr/share/sonic/scripts/docker-restapi-sidecar/k8s_pod_control.sh", mode=0o755),
```
make1980, line 173: we should make this change for acms as well for consistency. in case telemetry/restapi update fails for whatever reason as long as we update acms it won't be affected by telemetry/restapi.

FengPan-Frank: acms is in internal repo only.

dockers/docker-restapi-sidecar/systemd_stub.py
FengPan-Frank: what if there's change made in k8s_pod_control.sh?

make1980: that only requires daemon-reload, right? why do we need to use systemctl to restart the service? the reload will take effect next time when systemctl is used.

do we have anything that's needs to be executed in k8s_pod_control.sh or any systemd script? if yes we should move them to be within the container so that k8s already rerun them.

dockers/docker-restapi-sidecar/systemd_stub.py
make1980: how does that affect this comment?

dockers/docker-restapi-sidecar/systemd_stub.py
qiluo-msft: Is it possible that k8s_restapi get rollout before k8s_restapi-sidecar? I guess we need to restart restapi service in this conidtion.

PR #26317, opened 2026-03-22
sonic-py-common: remove unused variables in fabric config lookups

Sukuna0007Abhi:
Remove unused hwsku_json_file assignments in get_path_to_fabric_monitor_config_file() and get_path_to_fabric_port_config_file(). The variables were computed but never referenced.


Sukuna0007Abhi:
Pls review @lguohan @rck-innovium

PR #26318, opened 2026-03-22
sonic-py-common: use yaml.safe_load unconditionally in get_sonic_version_info

Sukuna0007Abhi:
#### How I did it:

Replaced the if/else version check with a single yaml.safe_load() call.

#### How to verify it:

The existing test_get_sonic_version in device_info_test.py
covers this function. yaml.safe_load() correctly parses the simple key-value format of sonic_version.yml.


Sukuna0007Abhi:
Pls review @lguohan @rck-innovium

PR #26322, opened 2026-03-22
mgmt interface: Fix incorrect management route addition order

rminnikanti:
#### Why I did it
The connected (subnet) route must be added before the default gateway route. Without the connected route in the routing table, the gateway route addition intermittently fails with:

```bash
networking[632]: warning: eth0: up cmd 'ip -6 route add default via 2004::1 dev eth0 table default metric 201' failed: returned 2 (RTNETLINK answers: No route to host)
```

##### Work item tracking

#### How I did it
swap the order of the two 'up' route commands in interfaces.j2 so the connected route is installed first, then the default gateway.

#### How to verify it
Sonic-mgmt:
1. deploy-mg and verify default route exists
2. Test cases - ip/test_mgmt_ipv6_only.py::test_syslog_ipv6_only and syslog/test_syslog.py::test_syslog

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [X] 202511

- [X] 202511

[interfaces-config]: Fix incorrect management route addition order

Signed-off-by: Ravi Minnikanti <rminnikanti@marvell.com>

### DIFF
--- a/files/image_config/interfaces/interfaces.j2
+++ b/files/image_config/interfaces/interfaces.j2
@@ -92,8 +92,8 @@ iface {{ name }} {{ 'inet' if prefix | ipv4 else 'inet6' }} static
 {% set force_mgmt_route_priority = 32764 %}
     ########## management network policy routing rules
     # management port up rules
-    up ip {{ '-4' if prefix | ipv4 else '-6' }} route add default via {{ MGMT_INTERFACE[(name, prefix)]['gwaddr'] }} dev {{ name }} table {{ vrf_table }} metric 201
     up ip {{ '-4' if prefix | ipv4 else '-6' }} route add {{ prefix | network }}/{{ prefix | prefixlen }} dev {{ name }} table {{ vrf_table }}
+    up ip {{ '-4' if prefix | ipv4 else '-6' }} route add default via {{ MGMT_INTERFACE[(name, prefix)]['gwaddr'] }} dev {{ name }} table {{ vrf_table }} metric 201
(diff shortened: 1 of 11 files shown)


files/image_config/interfaces/interfaces.j2
```diff
@@ -92,8 +92,8 @@ iface {{ name }} {{ 'inet' if prefix | ipv4 else 'inet6' }} static
 {% set force_mgmt_route_priority = 32764 %}
     ########## management network policy routing rules
     # management port up rules
-    up ip {{ '-4' if prefix | ipv4 else '-6' }} route add default via {{ MGMT_INTERFACE[(name, prefix)]['gwaddr'] }} dev {{ name }} table {{ vrf_table }} metric 201
     up ip {{ '-4' if prefix | ipv4 else '-6' }} route add {{ prefix | network }}/{{ prefix | prefixlen }} dev {{ name }} table {{ vrf_table }}
```
saiarcot895, line 95: Should this rule (`ip route add <prefix>/<prefixlen> dev <name>`) be already added by the kernel because of the address/netmask specified above? Or is this a case of DHCPv6 using /128 subnet?

rminnikanti: @saiarcot895 Thanks for the review.
kernel adds connected route implicitly to "main" routing table.
/etc/network/interfaces adds to "default" routing table and mgmt vrf table (when configured)

This PR fixes the order of route addition.

PR #26330, opened 2026-03-23
Fix DHCP server failure when secondary IP is configured on VLAN interface

cshivashgit:
#### Why I did it

Kea DHCP server crashes when a VLAN interface has a secondary IPv4 address
(`"secondary": "true"` in CONFIG_DB).

##### Root Cause

When a secondary IP (e.g. `192.169.0.1/22`) is added to a VLAN interface that
already has a primary IP (`192.168.0.1/21`), `_get_vlan_ipv4_interface()`
generates Kea subnets for **both** addresses. These subnets overlap, causing
Kea to fail on startup/reload.
Additionally, every secondary IP change in CONFIG_DB triggers a Kea reload —
even though secondary IPs aren't served by DHCP — leading to unnecessary
service disruptions.

##### Work item tracking

###### `dhcp_cfggen.py`
- Pass the full `vlan_interface` dict (not just keys) to
  `_get_vlan_ipv4_interface()` so it can inspect entry fields.
- Skip entries where `"secondary"` is `"true"`, preventing overlapping Kea
  subnet generation.

###### `dhcp_db_monitor.py`
- Skip Kea reload for secondary IP `SET` events (SET-only skip by design;
  `DEL` of a secondary IP still triggers one reload, which is acceptable since
  secondary-IP deletion is rare).
- Add a 0.5s debounce to coalesce rapid paired CONFIG_DB events (e.g. add-IP
  immediately followed by remove-IP) into a single Kea SIGHUP reload.

###### Test data
- `mock_config_db.json`: Add secondary IP entry for `Vlan1000`.
- `dhcp_db_monitor_test_data.json`: Add SET/DEL test cases for secondary IP
  events.

#### How to verify it

1. Configure a VLAN interface (e.g. `Vlan1000`) with a primary IPv4 address
   (`192.168.0.1/21`).
2. Add a secondary IPv4 address (`192.169.0.1/22`, `"secondary": "true"`) to
   the same VLAN.
3. Confirm the Kea DHCP server starts/reloads successfully (no overlapping
   subnet failure) and that the secondary subnet does **not** appear in the
   rendered Kea config.

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511

- [ ] 
- [ ] 

Fix Kea DHCP server failure when a secondary IPv4 address is configured on a
VLAN interface, and avoid unnecessary Kea reloads on secondary IP changes.


vrajeshe:
Following

cshivashgit:
@lguohan Can you review these changes?

vrajeshe:
Hi @yxieca can you please check and review the same
@ashutosh-agrawal  FYI

yxieca:
@cshivashgit please update the  PR template, you have good information already, please spread them in the template regarding what why and how sections.

yxieca:
Reviewed by: AI agent on behalf of Ying.

Decision: Comment (one clarification + nits)

Summary
- Filtering secondary IPs out of Kea subnet generation looks correct and is well-targeted at the reported crash. The debounce in `_process_check()` is also a reasonable approach to coalesce paired DEL+SET events into a single SIGHUP. One small clarification on DEL handling before approving.

Clarification request
- `_process_check()` in `dhcp_db_monitor.py` skips Kea reload for secondary-IP **SET** events by inspecting `entry` for `secondary == "true"`. On **DEL** events, however, `entry` is empty, so the secondary check can't match and the function falls through to `clear_event(); return True`. The new test case `["Vlan1000|192.169.0.1/22", "DEL", []] -> exp_res: true` confirms this — deleting a secondary IP still triggers a full Kea reload.

  Could you confirm whether this asymmetry is intentional (DEL-still-reloads is acceptable because deleting a secondary IP is rare and the cost is just one extra reload), or whether DEL of a secondary IP should also be skipped (e.g. by tracking secondary keys on SET and consulting that set on DEL, or by querying CONFIG_DB on DEL)? Either answer is fine — if intentional, a comment in the code and/or the PR description noting "SET-only by design" would help future readers.

Grounded nits
- `dhcp_db_monitor.py` — The `time.sleep(0.5)` is well-justified by the paired DEL+SET coalescing rationale, but the 0.5s value is a magic number. Consider extracting it as a named module constant (e.g. `KEA_RELOAD_DEBOUNCE_SECONDS = 0.5`) and noting in the docstring that future tuning should be data-driven.
- `dhcp_cfggen.py` `_get_vlan_ipv4_interface()` — the `entry.get("secondary") == "true"` check runs before the `network.version != 4` check. Harmless today, but it means an IPv6 entry marked `secondary=true` would also be skipped by this IPv4-focused helper. Either reorder (IPv4 check first) or rename to clarify the helper now filters secondary IPs of any family.
- Docstring/example for `_get_vlan_ipv4_interface()` shows the non-secondary entry as `{}`, but the mock data uses `{"NULL": "NULL"}` for those rows. Aligning the example with the actual schema (or vice versa) will avoid confusion.
- Consider an end-to-end test asserting that secondary IPs do not appear in the rendered Kea subnet output (not just monitor behavior). The current tests cover monitor-skip and cfggen filtering separately; a single integration check would catch regressions in either path.

PR #26349, opened 2026-03-24
[sonic-bgpcfgd] Add ability to wait on rehydration for additional Loopback IPs during reloads

anish-n:
#### Why I did it
If there are additional Loopback IPs defined in config db that are being advertised out via frr template files(using bgp network statements) and their config has not replayed in sonic-bgpcfgd, then these loopback ips transiently fail to be advertised out by the templates.

##### Work item tracking

#### How I did it
We introduce the ability to check for and fail peer addition requests if a additional loopbacks, as specified via a per-peer J2 file(additional_loopbacks.j2), are not replayed to bgpcfgd yet.

#### How to verify it
bgpcfgd unit test, and real device test in lab

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [x] 202511

- [ ] 
- [ ]

### DIFF
--- a/src/sonic-bgpcfgd/bgpcfgd/managers_bgp.py
+++ b/src/sonic-bgpcfgd/bgpcfgd/managers_bgp.py
@@ -243,6 +243,8 @@ class BGPPeerMgrBase(Manager):
         return True
 
     def post_dependencies_init(self):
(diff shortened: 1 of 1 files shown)


prsunny:
@StormLiangMS , @deepak-singhal0408 , please check the labels for cherry-pick.

PR #26374, opened 2026-03-25
fix: golang security

auspham:
#### Why I did it
Golang security fix for docker-ptf:latest

Currently, we're flagged for (GHSA-vvgc-356p-c3xw) and (GHSA-j5w8-q4qc-rx2x).

Instead of pinning the versioning. We should just upgrade them to latest.

##### Work item tracking
- Microsoft ADO **(number only)**: 36979761

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511

- [ ] 
- [ ]

### DIFF
--- a/dockers/docker-ptf/Dockerfile.j2
+++ b/dockers/docker-ptf/Dockerfile.j2
@@ -117,13 +117,14 @@ RUN GO_ARCH=amd64 \
 
 ENV PATH="/usr/local/go/bin:$HOME/go/bin:$PATH"
 
-# Build grpcurl from source with patched Go and patched golang.org/x/net
-# to address GHSA-qxp5-gwg8-xv66.
+# Build grpcurl from source with patched Go and golang.org/x/* deps
+# upgraded to latest to address current and future golang.org/x/* CVEs.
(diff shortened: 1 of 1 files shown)


wangxin (approved):

ronan-nexthop:
Is this actually required? Isn't the docker-ptf container just used for testing, and now we lose reproducable builds as it will pull a new master each time. 

For example, let's say in 202505, things remain stable for a while and an unrelated change to the docker file comes in, we need to rebuild the docker image and now we are pulling from master which an older branch definitely does not want.

ronan-nexthop:
@auspham @wangxin can you respond here please

auspham:
Hi @ronan-nexthop, while the docker-ptf is only applicable for testing only and in an isolated environment, its vulnerability is still a requirement for us to address. When addressing the vulnerability, our team will make sure there is no breaking changes that might affect the community. 

To your question of reproducible build, we have a mechanism to lock in all the dependency version internally to last known working version if you're using docker-ptf image from our repository. If there is a breaking change from upstream, our team will be notified and will deliver a fix as soon as possible.

PR #4394, opened 2020-04-08
[Juniper QFX5210] Fix Python errors

jleveque:
- `__init__` method cannot return a value
- Also remove unnecessary 'pass' statement

PR #26423, opened 2026-03-26
[src/libraries/sonic-fib]: Add sonic-fib library to build libnexthopgroup deb for RIB/FIB

GaladrielZhao:
#### Why I did it
This is the first PR of the RIB/FIB project.

Currently, fpmsyncd receives only partial nexthop information from FRR. To support nexthop object with full information
(called as NextHopGroupFull) passed from FRR (zebra) to fpmsyncd, a shared serialization library is needed so both processes can encode or decode NHG data consistently.

##### Work item tracking

#### How I did it
Introduce sonic-fib, a C++ library that provides encoding and decoding capabilities for NextHopGroupFull objects:
-   FRR uses the encoding function to generate a JSON string embedded in messages sent to fpmsyncd
-   fpmsyncd uses the decoding function to reconstruct the NHG object from the incoming JSON string

To minimize human error, the library uses a JSON Schema as the single source of truth and Jinja2 templates to auto-generate the required code.

(1) Key components:
- **Data** **schema**: NextHopGroupFull.json drives auto-generation of C++ classes, JSON codecs, and C struct wrappers via render_schema.py + Jinja2.
- **C-API**: C-callable interface for FRR to encode NHG objects into JSON.
- **Logging** **callback**: pluggable FIB_LOG macro with per-process callback registration (fib_frr_register_callback for FRR, registerLogCallback for SWSS).
- **Unit** **tests**: covers C++ class behavior, JSON round-trip, C-API conversion, and logging callback using Google Test.

(2) Architecture
image

(3) Directory Structure
```
sonic-fib/
├── schema/
│   └── NextHopGroupFull.json      # Single source of truth: data schema （source)
├── scripts/
│   └── render_schema.py           # Code generation engine (source)
├── templates/
│   ├── *.j2                       # Jinja2 templates for code generation (source)
│   │   ├── nexthopgroupfull.h.j2  
│   │   ├── nexthopgroupfull.cpp.j2
│   │   ├── nexthopgroupfull_json.h.j2
│   │   └── c_nexthopgroupfull.h.j2
│   └── references/                # Generated code (reference only, for review)
│       ├── nexthopgroupfull.h         # NextHopGroupFull definitions
│       ├── nexthopgroupfull.cpp       # NextHopGroupFull function implementation
│       ├── nexthopgroupfull_json.h    # NextHopGroupFull to/from_json functions
│       └── c_nexthopgroupfull.h       # C_NextHopGroupFull definitions
├── src/
│   ├── c-api/
│   │   ├── nexthopgroup_capi.h    # C API header
│   │   └── nexthopgroup_capi.cpp  # C→C++ conversion & JSON wrapper
│   ├── nexthopgroup_debug.h       # Logging callback interface
│   └── nexthopgroup_debug.cpp     # Logging implementation
└── tests/
    ├── nexthopgroupfull_ut.cpp         # C++ class unit tests
    ├── nexthopgroupfull_json_ut.cpp    # JSON serialization tests
    ├── nexthopgroup_debug_ut.cpp       # Logging callback tests
    ├── c_api_ut.cpp                    # C-API round-trip tests
    └── main.cpp                        # gtest entry point
```

***Note***: The generated files in templates/references are only kept for review and will not be committed later.

#### How to verify it
There is a UT module in sonic-fib, containing GTests for all functions.

image


src/libraries/sonic-fib/schema/NextHopGroupFull.json
```diff
@@ -0,0 +1,271 @@
...
+      "properties": {
+        "encap_behavior": { "$ref": "#/$defs/srv6_headend_behavior" },
+        "num_segs": { "$ref": "#/$defs/uint8" },
+        "seg": {
+          "type": "array",
+          "items": { "$ref": "#/$defs/in6_address" },
+          "minItems": 0
+        }
+      },
+      "required": ["encap_behavior", "segs"],
```
selva-nexthop: The line 245 has this "seg" but the required field is referring it as "segs".  This might lead to validation errors.

src/libraries/sonic-fib/src/c-api/nexthopgroup_capi.cpp
```diff
@@ -0,0 +1,221 @@
...
+    if (!c_nhg) {
+        FIB_LOG(fib::LogLevel::ERROR, "Do NOT pass in an empty C_NextHopGroupFull *");
+        return nullptr;
+    }
+
+    try {
+        FIB_LOG(fib::LogLevel::DEBUG, "multipaths %d", multipaths);
+        /* Convert C array to C++ vector */
+        vector<fib::nh_grp_full> cpp_nh_grp_full_list;
+        for (int i = 0; i < (MULTIPATH_NUM * MAX_NHG_RECURSION) + 1; i++) {
```
selva-nexthop: The schema defines id with "minimum": 0 and "default_value": "0", meaning id = 0 is a valid value. In FRR, NHG IDs are kernel-assigned and zero is used for unresolved or default entries. If any nexthop in the middle of the array has id == 0, all subsequent nexthops are silently dropped. The same sentinel pattern applies to depends[] and dependents[].

src/libraries/sonic-fib/templates/references/nexthopgroupfull.cpp
```diff
@@ -0,0 +1,345 @@
...
+                                    nh_seg6_segs_in->num_segs * sizeof(struct in6_addr);
+        nh_srv6->seg6_segs = (struct seg6_seg_stack *)malloc(total_size);
+        if (!nh_srv6->seg6_segs) {
+            FIB_LOG(fib::LogLevel::ERROR, "seg6_segs allocation failed in NextHopGroupFull constructor, abort");
+            free(nh_srv6);
+            nh_srv6 = nullptr;
+            return;
+        }
+
+        memcpy(nh_srv6->seg6_segs, nh_seg6_segs_in, sizeof(struct seg6_seg_stack));
```
selva-nexthop: sizeof(struct seg6_seg_stack) -- might not include the seg[] array data size and the following code then re-copies individual entries via a loop, making the initial memcpy both redundant and incorrect (it copies stale bytes from the destination's uninitialized memory into the beginning of the new allocation). In contrast, the copy constructor correctly computes sizeof(seg6_seg_stack) + num_segs * sizeof(in6_addr) for the full copy.

src/libraries/sonic-fib/src/Makefile.am
```diff
@@ -0,0 +1,90 @@
...
+
+# Install C-API headers
+nexthopgroup_capi_headerdir = $(includedir)/nexthopgroup/c-api
+nexthopgroup_capi_header_HEADERS = \
+    src/c-api/nexthopgroup_capi.h
+
+src_libnexthopgroup_la_CXXFLAGS = $(DBGFLAGS) $(AM_CFLAGS) $(CFLAGS_COMMON) $(CODE_COVERAGE_CXXFLAGS)
+src_libnexthopgroup_la_CPPFLAGS = $(DBGFLAGS) $(AM_CFLAGS) $(CFLAGS_COMMON) $(CODE_COVERAGE_CPPFLAGS)
+src_libnexthopgroup_la_LIBADD = -lpthread
+src_libnexthopgroup_la_LDFLAGS = -Wl,-z,now $(LDFLAGS)
```
selva-nexthop, line 89: Is it better to have any versioning info for this library?  When the schema evolves (a new field added, a field renamed), there is no mechanism to detect version mismatch between FRR and SWSS packages built at different times.

src/libraries/sonic-fib/src/c-api/nexthopgroup_capi.h
```diff
@@ -0,0 +1,37 @@
...
+#endif
+
+typedef struct NextHopGroupFull NextHopGroupFull;
+
+// Generate C++ NextHopGroupFull from C_NextHopGroupFull
+char* nexthopgroupfull_json_from_c_nhg_multi(const struct C_NextHopGroupFull* c_nhg, uint16_t multipaths);
+char* nexthopgroupfull_json_from_c_nhg_singleton(const struct C_NextHopGroupFull* c_nhg, uint16_t multipaths);
+
+// Free the object
+void nexthopgroup_free(NextHopGroupFull* obj);
```
selva-nexthop: nexthopgroupfull_json_from_c_nhg_multi and nexthopgroupfull_json_from_c_nhg_singleton returns char*, but the companion nexthopgroup_free function frees NextHopGroupFull* C++ object.

rules/sonic-fib.mk
```diff
@@ -0,0 +1,28 @@
+# libFIB package
+
+LIBFIB_VERSION = 1.0.0
+LIBFIB_NAME = libnexthopgroup
+
+LIBFIB = $(LIBFIB_NAME)_$(LIBFIB_VERSION)_$(CONFIGURED_ARCH).deb
+$(LIBFIB)_SRC_PATH = $(SRC_PATH)/libraries/sonic-fib
+$(LIBFIB)_VERSION = $(LIBFIB_VERSION)
+$(LIBFIB)_NAME = $(LIBFIB_NAME)
+$(LIBFIB)_DEPENDS += $(LIBNL3_DEV) $(LIBNL_GENL3_DEV) \
```
selva-nexthop: why do we need LIBNL3, LIBNL_GENL3, LIBNL_ROUTE3, LIBNL_NF3, and LIBNL_CLI as build-time and runtime dependencies?  libnexthopgroup is a pure JSON serialization library — it performs no netlink socket operations, parses no kernel netlink messages, and has no network interface queries.

src/libraries/sonic-fib/templates/references/c_nexthopgroupfull.h
```diff
@@ -0,0 +1,149 @@
...
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/* Manual definitions */
+typedef uint32_t vrf_id_t;
+typedef signed int ifindex_t;
+
+#define MULTIPATH_NUM 256
```
selva-nexthop: FRR's MULTIPATH_NUM is a compile-time ./configure option that can be set to 8, 32, 64, 256, or higher. If FRR is compiled with --enable-multipath=64, its internal nexthop array size diverges from the size declared here

src/libraries/sonic-fib/schema/NextHopGroupFull.json
```diff
@@ -0,0 +1,271 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "NextHopGroupFull.json",
```
selva-nexthop, line 3: The NextHopGroupFull.json schema has no version field. During a rolling upgrade of a SONiC system where FRR and SWSS packages are updated independently, the FRR instance may encode NHGFULL objects using a newer schema (with additional fields) while the fpmsyncd instance still uses the older decoder.

src/libraries/sonic-fib/src/c-api/nexthopgroup_capi.cpp
```diff
@@ -0,0 +1,221 @@
...
+
+        /* Convert C++ Obj to JSON stirng */
+        char* json_str = nexthopgroup_to_json(cpp_nhg);
+        FIB_LOG(fib::LogLevel::DEBUG, "json_str length %zu, str: %s", 
+            json_str ? strlen(json_str) : 0, json_str ? json_str : "null");
+
+        nexthopgroup_free(cpp_nhg);
+        return json_str;
+
+    } catch (...) {
```
selva-nexthop, line 88: this is applicable for all catch blocks.   The catch blocks discards the exception type and message entirely. In production, if nlohmann::json throws parse_error or type_error, the only log is "Converting failed" with no indication of which field caused the failure, what the actual JSON string was, or what type was expected. The standard pattern for SONiC C++ code is to catch std::exception& first for its .what() message, then catch (...) as a final fallback. Without this, debugging production route programming failures becomes extremely difficult.

src/libraries/sonic-fib/templates/references/c_nexthopgroupfull.h
```diff
@@ -0,0 +1,149 @@
...
+
+/* --- Root Struct: C_NextHopGroupFull --- */
+struct C_NextHopGroupFull {
+    uint32_t id;
+    uint32_t key;
+    uint8_t weight;
+    uint8_t flags;
+    uint32_t nhg_flags;
+
+#define NEXTHOP_FLAG_ONLINK (1 << 3)
```
selva-nexthop: #define NEXTHOP_FLAG_ONLINK (1 << 3) appears in both nexthopgroupfull.h (C++) and c_nexthopgroupfull.h (C). When a translation unit includes both headers (which happens in nexthopgroup_capi.cpp), this might cause warning and, depending on compiler settings, an error — especially given the -Wall -Werror flags in configure.ac. The constant should live in a single shared header included by both.

src/libraries/sonic-fib/templates/references/c_nexthopgroupfull.h
selva-nexthop, line 1: templates/references/ directory contains pre-generated output from the Jinja2 templates, committed to the repo "for review" with the stated intent that they "will not be committed later" -- we need to generate when there is a change or update in the schema and commit the generated files too?

src/libraries/sonic-fib/debian/control
```diff
@@ -0,0 +1,24 @@
+Source: sonic-fib
+Section: libs
+Maintainer: Eddie Ruan <your-email@example.com>
+Priority: optional
+Build-Depends: debhelper (>= 12),
+               autotools-dev,
+               python3,
+               nlohmann-json3-dev,
```
selva-nexthop, line 8: https://github.com/nlohmann/json/discussions/3372

AI datacenter fabric with BGP-learned routes at scale — e.g., 1M prefixes with 64-way ECMP NHG updates after a spine failure — the encoding/decoding cost in fpmsyncd's critical path will be non-trivial. The community has noted that nlohmann::json with ordered_json can be 8–10× slower than standard JSON in some configurations. At the scale targets expected for SONiC AI fabric deployments, this serialization bottleneck should be benchmarked against a binary format (e.g., Protocol Buffers or a custom compact encoding) before committing to JSON as the wire format.

selva-nexthop (commented):
/

src/libraries/sonic-fib/schema/NextHopGroupFull.json
selva-nexthop, line 1: Which part of  NextHopGroupFull.json schema cannot be represented by protobuff?

src/libraries/sonic-fib/schema/NextHopGroupFull.json
GaladrielZhao: Thanks very much for your review. I've fixed it.

src/libraries/sonic-fib/src/c-api/nexthopgroup_capi.cpp
GaladrielZhao: Yes, exactly. I've changed the C-API functions to accept explicit count parameters (nh_grp_full_count, depends_count, dependents_count) instead of relying on sentinel values. The iteration now uses these counts directly, ensuring all elements are processed correctly regardless of their values

src/libraries/sonic-fib/templates/references/nexthopgroupfull.cpp
GaladrielZhao: Thanks, I've changed it as in copy constructor.

src/libraries/sonic-fib/src/Makefile.am
GaladrielZhao: I would like to use Standards-Version: 1.0.0 in debian/control as the versioning mechanism for this library. Since both FRR and SWSS depend on libnexthopgroup (= ${binary:Version}), dpkg will enforce that the installed package version matches exactly at install time.
To make the version also visible at compile time and runtime, I add LIBNEXTHOPGROUP_VERSION "1.0.0" to the C-API header and a nexthopgroup_version() function. All four places (rules/sonic-fib.mk, debian/changelog, configure.ac, and the header macro) are aligned and should be bumped together when the schema evolves.

src/libraries/sonic-fib/src/c-api/nexthopgroup_capi.h
GaladrielZhao: The nexthopgroup_free() function in nexthopgroupfull_json_from_c_nhg_multi/singleton() free the C++ NextHopGroupFull object constructed by c-api in sonic-fib. As for the json string and the C_NextHopGroupFull object built and passed in c-api, they're freed in dplane_fpm_sonic.c after encoding, which is committed in another PR(https://github.com/sonic-net/sonic-buildimage/pull/26486):
```
static ssize_t netlink_nexthopgroupfull_msg_encode(...){
    ...
cleanup:
    free(json_str);
	free_c_nexthopgroupfull(&c_nhg);
    ...
}
```
The complete logic between dplane_fpm_sonic.c and sonic-fib is:
```
#1 Build C_NextHopGroupFull   >>>>  in dplane_fpm_sonic.c
#2 Call nexthopgroupfull_json_from_c_nhg_multi/singleton(C_NextHopGroupFull) >>>> in sonic-fib
    #2.1 Allocate NextHopGroupFull (C++) object and JSON string
    #2.2 return JSON
    #2.3 Free NextHopGroupFull (C++) object
#3 Encode the json_str as message, then free json_str and C_NextHopGroupFull Object  >>>> dplane_fpm_sonic.c
```

rules/sonic-fib.mk
GaladrielZhao: You're right. These dependencies are no use here. I've removed them.

src/libraries/sonic-fib/templates/references/c_nexthopgroupfull.h
GaladrielZhao: Sonic-fib cannot determine FRR's compile-time MULTIPATH_NUM at build time. However, SONiC already locks FRR's MULTIPATH_NUM to 514 via the build patch 0016-SONiC-ONLY-Set-multipath-to-514-and-disable-bgp-vnc-for-optimiz.patch. I've changed sonic-fib's MULTIPATH_NUM to 514 as well. Now sonic-fib uses the count parameter to limit iteration rather than the array size, and it can correctly handle all elements without accessing out-of-bounds memory.

src/libraries/sonic-fib/schema/NextHopGroupFull.json
GaladrielZhao: SONiC is a tight coupled system, the impact changes would be in buildimage and swss two repo. We don't want to support different version of encoding / decoding. Currently it is not feasible in SONiC build environment. Any update in sonic-fib needs to to ensure backward compatible to existing SWSS codes.

src/libraries/sonic-fib/src/c-api/nexthopgroup_capi.cpp
GaladrielZhao: Thanks. I change to update all catch blocks to first catch (const std::exception& e) to log e.what(), then catch (...) as a fallback.

src/libraries/sonic-fib/templates/references/c_nexthopgroupfull.h
GaladrielZhao: Removed the NEXTHOP_FLAG_ONLINK macro definition from sonic-fib since we don't actually use it. FRR has its own definition in the zebra codebase and uses that instead.

src/libraries/sonic-fib/templates/references/c_nexthopgroupfull.h
GaladrielZhao: These files are only for reference during review. I will remove them from this PR before merging. They will be generated during each build process and **none of them will be committed as source code**.

src/libraries/sonic-fib/debian/control
GaladrielZhao: This change applies to NHG only. Compared to 1M prefixes, the number of NHGs is very limited, typically in the thousands. During convergence, the volume of NHG messages is limited since route messages use NHG IDs instead of this structure.
We understand that nlohmann::json serialization might be slow. However, since the number of messages is relatively small, the performance penalty is minimal. Currently, we prefer nlohmann::json because the performance impact is negligible and it improves debuggability. If this becomes a concern in the future, we can adopt a binary JSON format.

src/libraries/sonic-fib/schema/NextHopGroupFull.json
GaladrielZhao: Several types in NextHopGroupFull cannot be directly represented by protobuf.
- C unions (e.g., gate/bh_type): protobuf has no equivalent, requires oneof with different semantics.
- Zero-length hash markers (_hash_begin[0], _hash_end[0]): used for continuous memory hashing, protobuf cannot guarantee memory layout.

There are several types can be represented while incompatible with existing FRR structures:
- Flexible arrays (seg6_seg_stack with seg[0]): protobuf uses repeated fields with different memory layout
- Nested C structures from FRR (struct nexthop_srv6): protobuf generates its own message types

We choose JSON Schema over protobuf for three key reasons.

The first one is safety. JSON Schema serves as the single source of truth for our data model. From this schema, we simultaneously generate C++ classes and their serialization/deserialization logic. In practice, FRR's C structures are converted to C++ objects through our C API layer, which are then serialized using the auto-generated code. This approach ensures that zebra and fpmsyncd always share identical data models, completely eliminating the manual coding errors that would occur if we maintained separate definitions.

Then is flexibility. The Jinja2 template logic gives us complete control over how C-specific constructs are handled during serialization. For example, we can precisely define how C unions like gate/src/rmap_src are converted to strings via gaddr_to_string(), how flexible arrays perform complete memory copies, how memory hash markers are preserved, and how conditional fields like bh_type are only serialized for blackhole nexthops. With protobuf, the code generation is fixed and cannot accommodate these C-specific requirements without adding manual conversion layers that would break away from the generated code.

The third is Debuggability. JSON is inherently human-readable and can be directly inspected in logs and debug outputs. We can read the exact value of the complete encoded/decoded message instead of separate values of each field.

GaladrielZhao:
Hi @selva-nexthop , I've addressed all the previous comments. Could you please take another look when you have a chance?  Thanks very much.

eddieruan-alibaba:
@selva-nexthop  can you help to review it again, @GaladrielZhao has already taken care of your previous comments.

GaladrielZhao:
Removed the files in sonic-fib/templates/references/ as they are generated during compilation and should not be committed. However, for review purposes, I’ve created a backup branch where you can find these files: https://github.com/GaladrielZhao/sonic-buildimage/tree/ribfib/sonic-fib_backup/src/libraries/sonic-fib/templates/references

Also squashed the commits into two for clearer commit messages.

eddieruan-alibaba:
@deepak-singhal0408  @selva-nexthop  can you help to finish reviewing this PR?

Also @deepak-singhal0408  can you help to check this sanity check failure? I saw it happened in multiple PRs. I don't understand why we need to check docker-ptf in all PRs. It is a helper docker anyway. 

[Azure.sonic-buildimage (Test Trivy vulnerability scan (docker-ptf))](https://github.com/sonic-net/sonic-buildimage/pull/26423/checks?check_run_id=75387640885)Failing after 29m

selva-nexthop (approved):

eddieruan-alibaba (dismissed):

src/libraries/sonic-fib/templates/nexthopgroupfull.cpp.j2
```diff
@@ -0,0 +1,359 @@
...
+    nh_label_type = other.nh_label_type;
+    bh_type = other.bh_type;
+
+    memcpy(&gate, &other.gate, sizeof(g_addr));
+    memcpy(&src, &other.src, sizeof(g_addr));
+    memcpy(&rmap_src, &other.rmap_src, sizeof(g_addr));
+
+    /* Deep copy for nh_srv6 information */
+    if (nh_srv6 != nullptr) {
+        delete nh_srv6;
```
deepak-singhal0408: ⚠️ **Undefined behavior: mixing `malloc`/`free` and `new`/`delete` for `nh_srv6`**

The copy assignment uses `delete` here (L207) and `new` at L210, but:
- The singleton constructor (L95) and copy constructor (L143) allocate with `malloc`
- The destructor (L355) frees with `free()`
- `from_json` in `nexthopgroupfull_json.h.j2:152` allocates with `new`

This means:
- If a `malloc`'d `nh_srv6` is copy-assigned → `delete` on `malloc`'d memory (undefined behavior)
- If a `new`'d `nh_srv6` (from `from_json`) is destructed → `free()` on `new`'d memory (undefined behavior)

Both cause heap corruption. Since `seg6_segs` uses flexible arrays requiring `malloc`, I'd suggest standardizing on `malloc`/`free` everywhere:
```cpp
// Line 207: replace delete with free
free(nh_srv6);

// Line 210: replace new with malloc+memset
nh_srv6 = static_cast<nexthop_srv6*>(malloc(sizeof(nexthop_srv6)));
if (!nh_srv6) { /* handle error */ }
memset(nh_srv6, 0, sizeof(nexthop_srv6));
```
Also fix `from_json` in `nexthopgroupfull_json.h.j2:152` (`new nexthop_srv6{}` → `malloc`).

src/libraries/sonic-fib/src/nexthopgroup_debug.cpp
```diff
@@ -0,0 +1,101 @@
...
+void fib::internalLog(LogLevel level, const char* file, int line,
+                        const char* func, const char* format, ...) {
+    auto& state = getState();
+    fib::LogCallback cb;
+    va_list args;
+    va_start(args, format);
+    {
+        std::lock_guard<std::mutex> lock(state.mutex);
+        if (static_cast<int>(level) < static_cast<int>(state.level) )
+            return;
```
deepak-singhal0408, line 89: va_start(args, format) is called at line 87, but this return at line 91 exits without calling va_end(args). Per C standard §7.16, every va_start must have a matching va_end — skipping it is undefined behavior.
 
 In practice, the FIB_LOG macro pre-filters by level before calling internalLog, so this path only fires in the rare race where level is raised between the macro check and this re-check. Still, consider adding this fix?:
 
  if (static_cast<int>(level) < static_cast<int>(state.level)) {
      va_end(args);
      return;
  }

src/libraries/sonic-fib/src/nexthopgroup_debug.cpp
```diff
@@ -0,0 +1,101 @@
...
+    fib::LogLevel level = fib::LogLevel::DEBUG;
+};
+LoggerState& getState() {
+    static LoggerState state;
+    return state;
+}
+
+// Default fallback: print to stderr
+void defaultLog(LogLevel level, const char* file, int line,
+                const char* func, const char* format, ...) {
```
deepak-singhal0408: ⚠️ **`defaultLog` signature mismatch with `LogCallback` — formatted arguments are lost**

`defaultLog` is declared as variadic (`const char* format, ...`) but is assigned to `LogCallback` which expects `va_list` as the last parameter. At line 99, it's called as `cb(level, file, line, func, format, args)` — passing a `va_list`.

`defaultLog` then does its own `va_start(args, format)` (line 35) which creates a new `va_list` reading from its own stack frame — not from the caller's `args`. The result: all format specifiers (`%d`, `%s`, etc.) print garbage or crash.

Fix — change signature to match `LogCallback`:
```cpp
void defaultLog(LogLevel level, const char* file, int line,
                const char* func, const char* format, va_list args) {
    // Remove va_start/va_end — use the passed args directly
    int written = std::vsnprintf(buf.data(), buf.size(), format, args);
    // ...
}
```

src/libraries/sonic-fib/templates/nexthopgroupfull.cpp.j2
```diff
@@ -0,0 +1,359 @@
...
+        }
+
+    FIB_LOG(fib::LogLevel::DEBUG, "NextHopGroupFull copy constructor finished");
+    }
+}
+
+/* Copy Assignment Operator */
+{{ root_struct_name }}& {{ root_struct_name }}::operator = (const NextHopGroupFull &other)
+{
+    FIB_LOG(fib::LogLevel::DEBUG,  "[CPP DEBUG] NextHopGroupFull copy assignment operator started for id: %d ", other.id,
```
deepak-singhal0408: **Broken `FIB_LOG` format strings — multiple format+value pairs passed as separate args**

```cpp
FIB_LOG(fib::LogLevel::DEBUG, "[CPP DEBUG] ... id: %d ", other.id,
        ", key: %x", other.key, ", nhg_flags: %x ", other.nhg_flags);
```

 vsnprintf only uses the first format string ("[CPP DEBUG] ... id: %d "). The remaining arguments are ignored — resulting in truncated log output (only id prints; key and nhg_flags are silently dropped). Same issue at lines 237 and 338-339.
 
Consider using a single format string:
```cpp
FIB_LOG(fib::LogLevel::DEBUG,
        "[CPP DEBUG] copy assignment started for id: %d, key: 0x%x, nhg_flags: 0x%x",
        other.id, other.key, other.nhg_flags);
```

src/libraries/sonic-fib/src/c-api/nexthopgroup_capi.cpp
```diff
@@ -0,0 +1,223 @@
...
+        FIB_LOG(fib::LogLevel::ERROR, "Do NOT pass in an empty C_NextHopGroupFull *");
+        return nullptr;
+    }
+
+    try {
+        FIB_LOG(fib::LogLevel::DEBUG, "nh_grp_full_count %d, depends_count %d, dependents_count %d, is_recurisve 0x%x",
+            nh_grp_full_count, depends_count, dependents_count, (uint8_t)is_recurisve);
+        /* Convert C array to C++ vector */
+        vector<fib::nh_grp_full> cpp_nh_grp_full_list;
+        for (uint32_t i = 0; i < nh_grp_full_count; i++) {
```
deepak-singhal0408, line 45: **Consider adding bounds checks on array count parameters**

The counts `nh_grp_full_count`, `depends_count`, and `dependents_count` are used directly in loops (lines 35, 45, 49). While the current caller (dplane_fpm_sonic) is trusted and FRR enforces its own limits, a defensive check here would prevent out-of-bounds reads if the API is ever called from a different context:

```cpp
if (nh_grp_full_count > (MULTIPATH_NUM * MAX_NHG_RECURSION) + 1 ||
    depends_count > MULTIPATH_NUM + 1 ||
    dependents_count > MULTIPATH_NUM + 1) {
    FIB_LOG(fib::LogLevel::ERROR, "Count exceeds array bounds");
    return nullptr;
}
```

src/libraries/sonic-fib/templates/nexthopgroupfull_json.h.j2
```diff
@@ -0,0 +1,224 @@
...
+        return;
+    }
+    nlohmann::ordered_json segs = nlohmann::ordered_json::array();
+    for (int i = 0; i < stack->num_segs; ++i) {
+        segs.push_back(ipv6_to_string(stack->seg[i]));
+    }
+    j = nlohmann::ordered_json{
+        {"encap_behavior", stack->encap_behavior},
+        {"num_segs", stack->num_segs},
+        {"segs", segs}
```
deepak-singhal0408: **Schema/wire-format key mismatch: `"seg"` vs `"segs"`**

The JSON Schema (`NextHopGroupFull.json:245`) defines the field as `"seg"`, but the serialization here uses `"segs"` and `from_json` (line 123) also reads `"segs"`. Encode/decode agree so it works at runtime, but the schema — which is documented as the "single source of truth" — is now inaccurate.

If anyone generates a validator or alternate parser from the schema, it will fail to match the actual wire format.

Suggest renaming the schema field from `"seg"` to `"segs"` to match what's actually produced.

src/libraries/sonic-fib/templates/nexthopgroupfull.cpp.j2
GaladrielZhao: Thanks for  your review. 
Fixed, using malloc/free for nh_srv6. However, we use `*nh_srv6 = nexthop_srv6()` instead of `memset(nh_srv6, 0, sizeof(nexthop_srv6))`, because nexthop_srv6 contains non-trivial members (seg6local_ctx) and memset on it triggers -Werror=class-memaccess.

src/libraries/sonic-fib/src/nexthopgroup_debug.cpp
GaladrielZhao: Fixed.

src/libraries/sonic-fib/src/nexthopgroup_debug.cpp
GaladrielZhao: Fixed.

src/libraries/sonic-fib/templates/nexthopgroupfull.cpp.j2
GaladrielZhao: Fixed.

src/libraries/sonic-fib/src/c-api/nexthopgroup_capi.cpp
GaladrielZhao: Added.

src/libraries/sonic-fib/templates/nexthopgroupfull_json.h.j2
GaladrielZhao: Fixed. We use "seg" to match the FRR struct field name :
```
struct seg6_seg_stack {
	enum srv6_headend_behavior encap_behavior;
	uint8_t num_segs;
	struct in6_addr seg[0]; /* 1 or more segs */
};
```
The serialization code (to_json/from_json) has been updated to use "seg" as well, so the schema and wire format are now consistent.

GaladrielZhao:
Hi @deepak-singhal0408 , thanks a lot for your review. I've addressed all the comments above.

deepak-singhal0408 (approved):
Thanks for addressing the review comments @GaladrielZhao .. LGTM.

radha-danda:
@GaladrielZhao, sairedis PR CI checks are failing with below linking errors. Can you please check if it is related to your PR and kindly fix? 

```
2026-05-25T05:56:13.6377525Z /bin/bash ../libtool  --tag=CXX   --mode=link g++  -g -O2 -ffile-prefix-map=/__w/1/s=. -fstack-protector-strong -Wformat -Werror=format-security  -Wl,-z,relro -o fpmsyncd fpmsyncd-fpmsyncd.o fpmsyncd-fpmlink.o fpmsyncd-routesync.o fpmsyncd-warmRestartHelper.o fpmsyncd-orch_zmq_config.o    -lnl-3 -lnl-route-3 -lswsscommon -lnexthopgroup -lnl-nf-3 -lnl-route-3 -lnl-genl-3 -lnl-3 -lhiredis 
2026-05-25T05:56:13.7125550Z libtool: link: g++ -g -O2 -ffile-prefix-map=/__w/1/s=. -fstack-protector-strong -Wformat -Werror=format-security -Wl,-z -Wl,relro -o fpmsyncd fpmsyncd-fpmsyncd.o fpmsyncd-fpmlink.o fpmsyncd-routesync.o fpmsyncd-warmRestartHelper.o fpmsyncd-orch_zmq_config.o  -lswsscommon -lnexthopgroup -lnl-nf-3 -lnl-route-3 -lnl-genl-3 -lnl-3 -lhiredis
2026-05-25T05:56:13.7258134Z /usr/bin/ld: cannot find -lnexthopgroup: No such file or directory
2026-05-25T05:56:13.7465330Z collect2: error: ld returned 1 exit status
2026-05-25T05:56:13.7477116Z make[4]: *** [Makefile:419: fpmsyncd] Error 1
2026-05-25T05:56:13.7478869Z make[4]: Leaving directory '/__w/1/s/fpmsyncd'
2026-05-25T05:56:13.7480812Z make[3]: *** [Makefile:420: all-recursive] Error 1
```

GaladrielZhao:
> @GaladrielZhao, sairedis PR CI checks are failing with below linking errors. Can you please check if it is related to your PR and kindly fix?
> 
> ```
> 2026-05-25T05:56:13.6377525Z /bin/bash ../libtool  --tag=CXX   --mode=link g++  -g -O2 -ffile-prefix-map=/__w/1/s=. -fstack-protector-strong -Wformat -Werror=format-security  -Wl,-z,relro -o fpmsyncd fpmsyncd-fpmsyncd.o fpmsyncd-fpmlink.o fpmsyncd-routesync.o fpmsyncd-warmRestartHelper.o fpmsyncd-orch_zmq_config.o    -lnl-3 -lnl-route-3 -lswsscommon -lnexthopgroup -lnl-nf-3 -lnl-route-3 -lnl-genl-3 -lnl-3 -lhiredis 
> 2026-05-25T05:56:13.7125550Z libtool: link: g++ -g -O2 -ffile-prefix-map=/__w/1/s=. -fstack-protector-strong -Wformat -Werror=format-security -Wl,-z -Wl,relro -o fpmsyncd fpmsyncd-fpmsyncd.o fpmsyncd-fpmlink.o fpmsyncd-routesync.o fpmsyncd-warmRestartHelper.o fpmsyncd-orch_zmq_config.o  -lswsscommon -lnexthopgroup -lnl-nf-3 -lnl-route-3 -lnl-genl-3 -lnl-3 -lhiredis
> 2026-05-25T05:56:13.7258134Z /usr/bin/ld: cannot find -lnexthopgroup: No such file or directory
> 2026-05-25T05:56:13.7465330Z collect2: error: ld returned 1 exit status
> 2026-05-25T05:56:13.7477116Z make[4]: *** [Makefile:419: fpmsyncd] Error 1
> 2026-05-25T05:56:13.7478869Z make[4]: Leaving directory '/__w/1/s/fpmsyncd'
> 2026-05-25T05:56:13.7480812Z make[3]: *** [Makefile:420: all-recursive] Error 1
> ```

Hi @radha-danda , thanks for your update.
I believe this is caused by [sonic-swss PR #4394](https://github.com/sonic-net/sonic-swss/pull/4394) which added -lnexthopgroup to fpmsyncd's link dependencies and updated swss's own CI pipeline to download the libnexthopgroup deb, but did not update the sairedis CI pipeline accordingly. So when sairedis CI compiles swss from master, it cannot find the library.

I've opened a PR https://github.com/sonic-net/sonic-sairedis/pull/1908 to fix this. It adds the libnexthopgroup deb download and installation to sairedis CI pipeline files.

PR #26446, opened 2026-03-28
Fix system-health crash on VS platform due to missing metadata file

### LINKED ISSUE #26429: system-health service crashes on trixie VS image due to missing vs_chassis_metadata.json
## Description

On trixie-based VS (virtual switch) images, `system-health.service` crashes with:

```
FileNotFoundError: Metadata file /etc/sonic/vs_chassis_metadata.json not found
```

The service fails repeatedly and gives up restarting. This prevents `SYSTEM_READY|SYSTEM_STATE` from being set in STATE_DB, which blocks any daemon that waits for system-ready (e.g., hsflowd in the sflow container).

## Impact

- **hsflowd** loops in `waitConfig` / `getSystemReady()` forever, never reads CONFIG_DB, never generates `/etc/hsflowd.auto`
- All sflow KVM tests fail with `hsflowd failed to initialize collector(s) within 240 seconds`
- Potentially affects any other service that depends on `SYSTEM_READY` in STATE_DB
- Issue is masked in CI because sflow tests are blanket-skipped via `tests_mark_conditions.yaml` (sonic-mgmt#21701)

## Reproduction

1. Boot a trixie-based sonic-vs image on KVM
2. Check `systemctl status system-health` — it will be in failed state
3. Check `redis-cli -n 6 HGETALL 'SYSTEM_READY|SYSTEM_STATE'` — empty
4. Enable sflow and add a collector — `/etc/hsflowd.auto` stays empty

## Workaround

Manually set system ready in STATE_DB:
```
redis-cli -n 6 HSET 'SYSTEM_READY|SYSTEM_STATE' Status UP
```
This immediately unblocks hsflowd and all dependent services.

## Root Cause

The VS platform is missing `/etc/sonic/vs_chassis_metadata.json`. The system-health service expects this file but doesn't handle its absence gracefully on VS platform.

## Proposed Fix

Working on a PR — either create the missing metadata file for VS platform or make system-health handle its absence gracefully.

## Environment

- Platform: `x86_64-kvm_x86_64-r0`
- SONiC branch: trixie (Debian 13)
- systemd: 257
- Discovered during: PR #26421 (sflow trixie container upgrade)

yxieca:
## Description

The VS chassis platform module (`sonic_platform/chassis.py`) raises `FileNotFoundError` when `/etc/sonic/vs_chassis_metadata.json` does not exist. This file is only present on VS chassis (T2-VOQ) setups introduced in #18512, not on standalone VS platforms (e.g., ToR used in KVM testbeds).

## Impact

The crash prevents `system-health.service` from running, which means `SYSTEM_READY|SYSTEM_STATE` is never set in STATE_DB. This blocks any daemon that waits for system-ready — specifically `hsflowd` in the sflow container, which loops in `waitConfig`/`getSystemReady()` forever, never reads CONFIG_DB, and never generates `/etc/hsflowd.auto`.

This causes all sflow KVM tests to fail on trixie images with:
```
Failed: hsflowd failed to initialize collector(s) ['20.1.1.2'] within 240 seconds
```

The issue is currently masked in CI because sflow tests are blanket-skipped via `tests_mark_conditions.yaml` (sonic-mgmt#21701).

## Fix

Return an empty metadata dict when the file doesn't exist instead of raising an exception. The chassis methods that consume metadata (`get_supervisor_slot`, `get_linecard_slot`, `get_my_slot`) already raise `KeyError`/`ValueError` for missing fields, so the behavior is correct for chassis setups.

## Testing

- **15/15 sflow tests passed** on KVM T0 testbed (vlab-01, P520-2) with this fix
- All reboot types tested (cold, fast, warm) — all passed
- `system-health.service` starts successfully on standalone VS, setting `SYSTEM_READY` automatically

Fixes #26429
Supersedes #26430

### DIFF
--- a/device/virtual/x86_64-kvm_x86_64-r0/system_health_monitoring_config.json
+++ b/device/virtual/x86_64-kvm_x86_64-r0/system_health_monitoring_config.json
@@ -1,0 +1,11 @@
+{
+    "services_to_ignore": ["smartmontools.service"],
+    "devices_to_ignore": [],
+    "user_defined_checkers": [],
+    "polling_interval": 60,
+    "led_color": {
+        "fault": "red",
+        "normal": "green",
+        "booting": "orange_blink"
+    }
+}
--- a/platform/vs/sonic-platform-modules-vs/sonic_platform/chassis.py
+++ b/platform/vs/sonic-platform-modules-vs/sonic_platform/chassis.py
@@ -30,21 +30,28 @@ class Chassis(ChassisBase):
         if os.path.exists(self.metadata_file):
             with open(self.metadata_file, 'r') as f:
                 metadata = json.load(f)
-        else:
-            raise FileNotFoundError("Metadata file {} not found".format(self.metadata_file))
         return metadata
 
+    def _require_metadata(self):
+        if not self.metadata:
+            raise FileNotFoundError(
+                "Metadata file {} is required for chassis operations".format(
+                    self.metadata_file))
+
     def get_supervisor_slot(self):
+        self._require_metadata()
         if 'sup_slot_num' not in self.metadata:
             raise KeyError("sup_slot_num not found in Metadata file {}".format(self.metadata_file))
         return self.metadata['sup_slot_num']
 
     def get_linecard_slot(self):
+        self._require_metadata()
         if 'lc_slot_num' not in self.metadata:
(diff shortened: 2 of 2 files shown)


arlakshm (approved):

ryanzhu706:
Hi @yxieca Can we merge this PR which is needed for upgrade test on KVM? Thank you!

ryanzhu706 (approved):

PR #26452, opened 2026-03-29
Build gperftools (tcmalloc) from source for docker-fpm-frr

StormLiangMS:
#### Why I did it

Install tcmalloc into the docker-fpm-frr container to prepare for FRR memory optimization. tcmalloc provides a more efficient memory allocator that reduces heap fragmentation compared to glibc malloc, which is a known issue with long-running FRR processes.

The stock Debian `libgoogle-perftools4` package already has working `MallocExtension_ReleaseToSystem()` via the `MADV_DONTNEED` code path (verified against gperftools 2.10 and 2.16 source — thanks @saiarcot895 for the thorough analysis), so no custom build or patches are needed.

This PR only installs the library — it does **not** enable or load tcmalloc. Activation (via `LD_PRELOAD` or FRR's native `--enable-gperf-tcmalloc`) will be handled in a separate follow-up PR.

Reference: [FRRouting/frr#19377](https://github.com/FRRouting/frr/pull/19377)

##### Work item tracking
- Microsoft ADO:

#### How I did it

Added `libgoogle-perftools4` (which provides `libtcmalloc.so.4`) to the apt-get install list in the docker-fpm-frr Dockerfile.

#### How to verify it

1. Build the image
2. Verify the library is present: `docker exec -it bgp ls -la /usr/lib/x86_64-linux-gnu/libtcmalloc*`
3. Verify the package is installed: `docker exec -it bgp dpkg -l | grep perftools`

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511

- [x] master (CI passed)

Install tcmalloc (libgoogle-perftools4) into docker-fpm-frr container from Debian repos

N/A — no config_db or YANG changes

🐆

### DIFF
--- a/dockers/docker-fpm-frr/Dockerfile.j2
+++ b/dockers/docker-fpm-frr/Dockerfile.j2
@@ -16,6 +16,7 @@ ENV DEBIAN_FRONTEND=noninteractive
 RUN apt-get update   && \
     apt-get install -y  \
         iproute2        \
+        libgoogle-perftools4t64 \
         logrotate
 
 RUN groupadd -g ${frr_user_gid} frr


src/gperftools/patch/0001-Enable-FREE_MMAP_PROT_NONE-for-Linux.patch
```diff
@@ -0,0 +1,25 @@
...
+
+diff --git a/configure.ac b/configure.ac
+--- a/configure.ac
++++ b/configure.ac
+@@ -53,4 +53,5 @@ case "$host" in
+    *-cygwin*) default_enable_heap_checker=no; default_enable_cpu_profiler=no;;
+    *-freebsd*) default_enable_heap_checker=no;;
+    *-darwin*) default_enable_heap_checker=no;;
++   *-linux*) AC_DEFINE(FREE_MMAP_PROT_NONE, 1, [Use mmap PROT_NONE to release memory on Linux]);;
+ esac
```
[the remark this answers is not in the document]
StormLiangMS: False positive. \*-linux*\ is a shell glob (autoconf \case\ pattern), not a regex — it matches any string containing \-linux\, so \x86_64-pc-linux-gnu\ matches fine. This is the same pattern style used by the existing upstream cases (\*-cygwin*\, \*-freebsd*\, \*-darwin*\).

rules/docker-fpm-frr.mk
```diff
@@ -7,7 +7,7 @@ DOCKER_FPM_FRR_DBG = $(DOCKER_FPM_FRR_STEM)-$(DBG_IMAGE_MARK).gz
 $(DOCKER_FPM_FRR)_PATH = $(DOCKERS_PATH)/$(DOCKER_FPM_FRR_STEM)
 $(DOCKER_FPM_FRR)_PYTHON_WHEELS += $(SONIC_BGPCFGD) $(SONIC_FRR_MGMT_FRAMEWORK)
 
-$(DOCKER_FPM_FRR)_DEPENDS += $(FRR) $(FRR_SNMP) $(SWSS) $(LIBYANG3)
+$(DOCKER_FPM_FRR)_DEPENDS += $(FRR) $(FRR_SNMP) $(SWSS) $(LIBYANG3) $(LIBGOOGLE_PERFTOOLS4)
```
rminnikanti: Hi @StormLiangMS,

This PR only adding LIBGOOGLE_PERFTOOLS4 as an FRR dependency. I don't see its adding LD_PRELOAD anywhere to intercept memory allocation functions of FRR daemons. 

Let me know if I am missing something.

vganesan-nokia:
@StormLiangMS, the PR description mentions change in docker-fpm-frr: docker_init.sh. But there is no docker_init.sh (docker-fpm-frr) in this PR.

rules/docker-fpm-frr.mk
StormLiangMS: Hi @rminnikanti, You’re correct — this PR intentionally only builds gperftools from source and installs libgoogle-perftools4 (which provides libtcmalloc.so.4) into the FRR container. It does not enable or load tcmalloc yet.

The purpose of this PR is solely to support the build of tcmalloc. How to enable it would fall under the scope of FRR upgrade work, either by using the same method as FRR native (--enable-gperf-tcmalloc) or other approaches. I believe that is outside the scope of this PR, JFYR. I filed a separate PR to enable the tcmalloc, (https://github.com/sonic-net/sonic-buildimage/pull/26697)

StormLiangMS:
> @StormLiangMS, the PR description mentions change in docker-fpm-frr: docker_init.sh. But there is no docker_init.sh (docker-fpm-frr) in this PR.

hi @vganesan-nokia That’s correct — I changed the implementation but forgot to update the description. I’ve just updated it. Thanks for pointing that out!

rminnikanti:
FRR migration to trixie PR is open since few weeks now. This may need re-verification if it gets merged.

https://github.com/sonic-net/sonic-buildimage/pull/26384 
@saiarcot895 may I know if the FRR trixie migration PR can be merged?

StormLiangMS:
> LGTM — building gperftools from source with the FREE_MMAP_PROT_NONE patch is the FRR-recommended approach. Stock Debian packages don't enable this, and that hasn't changed even in gperftools 2.18.1 (latest upstream).
> 
> One thing to keep in mind for the future: when docker-fpm-frr moves to Trixie, this recipe will need updating:
> 
> Package names gain the t64 suffix (libgoogle-perftools4 → libgoogle-perftools4t64, libtcmalloc-minimal4 → libtcmalloc-minimal4t64) Trixie ships gperftools 2.16-1 (vs bookworm's 2.10-1), so the source build version and patch offsets may need adjustment Not a blocker — just a heads-up for Trixie migration.
> 
> Also, @StormLiangMS, could you update the title a bit, something like, "Build gperftools (tcmalloc) from source for docker-fpm-frr" ?

thanks, updated. @deepak-singhal0408

rules/gperftools.mk
```diff
@@ -0,0 +1,19 @@
+# gperftools (google-perftools) - tcmalloc
+
+GPERFTOOLS_VERSION = 2.10
```
saiarcot895: Why is an old version of gperftools being used?

saiarcot895: Never mind, I see Bookworm is on 2.10.

src/gperftools/patch/0001-Enable-FREE_MMAP_PROT_NONE-for-Linux.patch
```diff
@@ -0,0 +1,25 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: StormLiangMS <stormliang@microsoft.com>
+Date: Sun, 29 Mar 2026 17:00:00 +0800
+Subject: [PATCH] Enable FREE_MMAP_PROT_NONE for Linux
+
+Without this define, MallocExtension_ReleaseToSystem() does not
+actually release memory back to the OS on Linux. This is required
+for the FRR tcmalloc memory release timer to function correctly.
```
saiarcot895: This change doesn't make sense.

The only places `FREE_MMAP_PROT_NONE` is defined is within `TCMalloc_SystemRelease` and `TCMalloc_SystemCommit`. This comment is talking about releasing memory back to the OS, so looking further into `TCMalloc_SystemRelease`, the entire function is under an if-check:

```
#if defined(FREE_MMAP_PROT_NONE) && defined(HAVE_MMAP) || defined(MADV_FREE)
```

`FREE_MMAP_PROT_NONE` isn't defined by default, and `MADV_FREE` ends up being defined:

```
// Linux added support for MADV_FREE in 4.5 but we aren't ready to use it
// yet. Among other things, using compile-time detection leads to poor
// results when compiling on a system with MADV_FREE and running on a
// system without it. See https://github.com/gperftools/gperftools/issues/780.
#if defined(__linux__) && defined(MADV_FREE) && !defined(TCMALLOC_USE_MADV_FREE)
# undef MADV_FREE
#endif

// MADV_FREE is specifically designed for use by malloc(), but only
// FreeBSD supports it; in linux we fall back to the somewhat inferior
// MADV_DONTNEED.
#if !defined(MADV_FREE) && defined(MADV_DONTNEED)
# define MADV_FREE  MADV_DONTNEED
#endif
```

While the linux userspace headers in both Bookworm and Trixie define `MADV_FREE`, because of potentially unknown behavior around `MADV_FREE`, it gets undefined. However, because `MADV_DONTNEED` is defined, and `MADV_FREE` is no longer defined, `MADV_FREE` gets defined to the same value as `MADV_DONTNEED`. This means that the original if-check pasted at the top becomes true, because `MADV_FREE` is defined.

In `TCMalloc_SystemRelease` then, the `madvise` code path is taken. This can be verified to be in the compiled build from Debian (taken from tcmalloc 2.16 as compiled for Debian Trixie):

```
00000000000131b0 <_Z22TCMalloc_SystemReleasePvm@@Base>:
   131b0:       f3 0f 1e fa             endbr64
   131b4:       48 8b 05 8d 4c 01 00    mov    0x14c8d(%rip),%rax        # 27e48 <_ZN60FLAG__namespace_do_not_use_directly_use_DECLARE_bool_instead35FLAGS_malloc_disable_memory_releaseE@@Base-0xc39>
   131bb:       80 38 00                cmpb   $0x0,(%rax)
   131be:       0f 85 94 00 00 00       jne    13258 <_Z22TCMalloc_SystemReleasePvm@@Base+0xa8>
   131c4:       55                      push   %rbp
   131c5:       48 89 e5                mov    %rsp,%rbp
   131c8:       41 55                   push   %r13
   131ca:       49 89 f5                mov    %rsi,%r13
   131cd:       41 54                   push   %r12
   131cf:       53                      push   %rbx
   131d0:       48 89 fb                mov    %rdi,%rbx
   131d3:       48 83 ec 08             sub    $0x8,%rsp
   131d7:       48 8b 05 0a 59 01 00    mov    0x1590a(%rip),%rax        # 28ae8 <tcmalloc_sys_alloc@@Base+0x58>
   131de:       48 85 c0                test   %rax,%rax
   131e1:       75 0e                   jne    131f1 <_Z22TCMalloc_SystemReleasePvm@@Base+0x41>
   131e3:       e8 88 aa ff ff          call   dc70 <getpagesize@plt>
   131e8:       48 98                   cltq
   131ea:       48 89 05 f7 58 01 00    mov    %rax,0x158f7(%rip)        # 28ae8 <tcmalloc_sys_alloc@@Base+0x58>
   131f1:       4c 8d 64 03 ff          lea    -0x1(%rbx,%rax,1),%r12
   131f6:       48 f7 d8                neg    %rax
   131f9:       4c 01 eb                add    %r13,%rbx
   131fc:       49 21 c4                and    %rax,%r12
   131ff:       48 21 c3                and    %rax,%rbx
   13202:       31 c0                   xor    %eax,%eax
   13204:       49 39 dc                cmp    %rbx,%r12
   13207:       72 0f                   jb     13218 <_Z22TCMalloc_SystemReleasePvm@@Base+0x68>
   13209:       48 83 c4 08             add    $0x8,%rsp
   1320d:       5b                      pop    %rbx
   1320e:       41 5c                   pop    %r12
   13210:       41 5d                   pop    %r13
   13212:       5d                      pop    %rbp
   13213:       c3                      ret
   13214:       0f 1f 40 00             nopl   0x0(%rax)
   13218:       4c 29 e3                sub    %r12,%rbx
   1321b:       0f 1f 44 00 00          nopl   0x0(%rax,%rax,1)
   13220:       ba 04 00 00 00          mov    $0x4,%edx
   13225:       48 89 de                mov    %rbx,%rsi
   13228:       4c 89 e7                mov    %r12,%rdi
   1322b:       e8 10 a6 ff ff          call   d840 <madvise@plt>
   13230:       41 89 c5                mov    %eax,%r13d
   13233:       e8 08 9e ff ff          call   d040 <__errno_location@plt>
   13238:       83 38 0b                cmpl   $0xb,(%rax)
   1323b:       75 06                   jne    13243 <_Z22TCMalloc_SystemReleasePvm@@Base+0x93>
   1323d:       41 83 fd ff             cmp    $0xffffffff,%r13d
   13241:       74 dd                   je     13220 <_Z22TCMalloc_SystemReleasePvm@@Base+0x70>
   13243:       41 83 fd ff             cmp    $0xffffffff,%r13d
   13247:       0f 95 c0                setne  %al
   1324a:       48 83 c4 08             add    $0x8,%rsp
   1324e:       5b                      pop    %rbx
   1324f:       41 5c                   pop    %r12
   13251:       41 5d                   pop    %r13
   13253:       5d                      pop    %rbp
   13254:       c3                      ret
   13255:       0f 1f 00                nopl   (%rax)
   13258:       31 c0                   xor    %eax,%eax
   1325a:       c3                      ret
   1325b:       0f 1f 44 00 00          nopl   0x0(%rax,%rax,1)
```

Note the `madvise` call at 0x1322b, and `0x4` (the value of `MADV_DONTNEED`) being written to `%edx` (where the 3rd argument is written to for amd64 architectures).

This means that `MallocExtension_ReleaseToSystem` is actually releasing memory back to the system. This also means that there should be no need to build gperftools locally, or with this patch.

saiarcot895: I will also note that `malloc_trim` is just calling `madvise` with `MADV_DONTNEED`, and we have seen that to release memory back to the system (in the form of RSS for the process going down and `free -m` reporting higher values for free and available system memory after `malloc_trim`).

StormLiangMS: @saiarcot895 Thanks for the thorough analysis — you're absolutely right. We verified against both the 2.10 and 2.16 source: the MADV_FREE -> MADV_DONTNEED redefine means TCMalloc_SystemRelease already uses madvise(MADV_DONTNEED) in the stock Debian build, so the FREE_MMAP_PROT_NONE patch and custom build are unnecessary.

Simplifying this PR to just apt-get install libgoogle-perftools4 in the FRR container Dockerfile. No more building from source.

saiarcot895 (changes requested):
Given that this PR is now just a library package installation, is this PR needed? Why not just have a single PR that adds the development header package into a slave container and enables `--enable-gperf-tcmalloc` for the FRR build?

StormLiangMS:
> Given that this PR is now just a library package installation, is this PR needed? Why not just have a single PR that adds the development header package into a slave container and enables `--enable-gperf-tcmalloc` for the FRR build?

make sense to me. @rminnikanti would you help with PR to enable it --enable-gperf-tcmalloc and with the apt-get? let me know, then I can close this PR.

src/gperftools/patch/0001-Enable-FREE_MMAP_PROT_NONE-for-Linux.patch
rminnikanti: @saiarcot895 should we pass this information to FRR PR? FRR PR recommends building libgoogle-perftools4 with FREE_MMAP_PROT_NONE. 
PR: https://github.com/FRRouting/frr/pull/19377 

If you agree, would you like to comment on the PR?

PR #26454, opened 2026-03-29
[supervisord-utilities-rs] Wait for syslog connection to be established

MavChtz:
#### Why I did it
The new Rust listener fails if it starts before the syslog daemon is initialized. This occurs because it attempts to connect to the socket without retrying on failure.

This wasn't an issue with the Python version since the Python syslog library would only establish the connection to the syslog daemon once the first message was being sent.

#### How I did it
The connection to the syslog daemon is repeatedly attempted on failure every second.

#### How to verify it
1. Run `sudo config reload`
2. Run `supervisorctl status` for any container (e.g swss)
3. Observe `RUNNING` instead of `FATAL`
4. Repeat at least ten times due the unpredictable nature of this issue

- [x] 202511
This was introduced with https://github.com/sonic-net/sonic-buildimage/pull/24698.

-  [x]  202511

Wait for syslog connection to be established in supervisord-utilities-rs.

🐈🐱

### DIFF
--- a/src/sonic-supervisord-utilities-rs/src/proc_exit_listener.rs
+++ b/src/sonic-supervisord-utilities-rs/src/proc_exit_listener.rs
@@ -11,6 +11,7 @@ use std::collections::HashMap;
 use std::fs::File;
 use std::io::{self, BufRead, BufReader, Read};
 use std::os::unix::io::AsRawFd;
+use std::thread;
 use std::process;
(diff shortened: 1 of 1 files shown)


MavChtz:
Closing in favour of https://github.com/sonic-net/sonic-buildimage/pull/26840/

PR #26486, opened 2026-03-31
[FRR] Support NHG information encoding and send to fpmsyncd for RIB/FIB

GaladrielZhao:
#### Why I did it
This is one of the RIB/FIB PRs, to support sending enough NHG information to construct FIB in fpmsyncd.

#### How I did it
Use sonic-fib's Data Schema to encode zebra nexthop information as JSON string and send it to fpmsyncd as part of the RIB/FIB project. The modifications are in dplane_fpm_sonic.c and supervisord.conf.j2, and can be divided into four logical parts.

1 zebra info -> C_NextHopGroupFull -> C++ NextHopGroupFull -> JSON string
    Add feature to build C_NextHopGroupFull (defined in sonic-fib),
    and call sonic-fib's C-API to construct C++ NextHopGroupFull,
    then return the JSON string back to zebra.

2 Encode JSON string as message and send to fpmsyncd:
    Implement the encoding function and modify the fpm_nl_enqueue's
    DPLANE_OP_NH_XXXX case to call the new one.

3 Add DEVICE_METADATA config to enable RIB/FIB feature
    Read DEVICE_METADATA.localhost.nhg_fib from ConfigDB to conditionally
    pass the --nhg-fib flag to zebra at startup via supervisord.conf.j2,
    enabling the RIB/FIB NHG full encoding.

4 vtysh command to control FIB log level at runtime
    Add `fpm fib-log-level <debug|info|warn|error>` command to dynamically
    adjust the FIB library log level, bridging fib::LogLevel in sonic-fib
    with zebra's native logging system.

#### How to verify it
- Phoenixwing test
- PR sanity

image

### DIFF
--- a/src/sonic-frr/dplane_fpm_sonic/dplane_fpm_sonic.c
+++ b/src/sonic-frr/dplane_fpm_sonic/dplane_fpm_sonic.c
@@ -2749,6 +2749,22 @@ static ssize_t netlink_nexthopgroupfull_msg_encode(uint16_t cmd,
 		struct C_NextHopGroupFull c_nhg;
 		const struct C_NextHopGroupFull *c_nhg_ptr = &c_nhg;
 
+		memset(&c_nhg, 0, sizeof(c_nhg));
+
+		/* Defensive bounds check on array count */
+		if (dplane_ctx_get_nhe_nh_grp_full_count(ctx) >
+			    (MULTIPATH_NUM * MAX_NHG_RECURSION) + 1 ||
+		    dplane_ctx_get_nhe_depends_count(ctx) > MULTIPATH_NUM + 1 ||
+		    dplane_ctx_get_nhe_dependents_count(ctx) > MULTIPATH_NUM + 1) {
+			zlog_err(
+				"%s: C_NextHopGroupFull array count exceeds bounds: nh_grp_full_count=%u, depends_count=%u, dependents_count=%u",
+				__func__, dplane_ctx_get_nhe_nh_grp_full_count(ctx),
+				dplane_ctx_get_nhe_depends_count(ctx),
+				dplane_ctx_get_nhe_dependents_count(ctx));
+			free_c_nexthopgroupfull(&c_nhg);
+			return -1;
+		}
+
 		/*
 		 * We also distinguish between a "group" and a singleton similar
 		 * as what is done in netlink_nexthop_msg_encode.
--- a/src/sonic-frr/dplane_fpm_sonic/dplane_fpm_sonic.c
+++ b/src/sonic-frr/dplane_fpm_sonic/dplane_fpm_sonic.c
@@ -59,7 +59,19 @@
 #include "fpm/fpm.h"
 #include "lib/srv6.h"
 #include "lib/vrf.h"
+#include <nexthopgroup/c_nexthopgroupfull.h>
 #include <nexthopgroup/c-api/nexthopgroup_capi.h>
+
+/* Global flag set by zebra --nhg-fib command-line option. */
+extern bool zebra_nhg_fib_enabled;
+
+/* FIB log level constants, aligned with fib::LogLevel in nexthopgroup_debug.h */
+enum fib_log_level {
+	FIB_LOG_LEVEL_DEBUG = 0,
+	FIB_LOG_LEVEL_INFO  = 1,
+	FIB_LOG_LEVEL_WARN  = 2,
+	FIB_LOG_LEVEL_ERROR = 3,
+};
 
 #define SOUTHBOUND_DEFAULT_ADDR INADDR_LOOPBACK
 #define SOUTHBOUND_DEFAULT_PORT 2620
@@ -103,6 +115,8 @@ enum custom_nlmsg_types {
 	RTM_DELSRV6VPNROUTE		= 3001,
 	RTM_NEWSIDLIST			= 4000,
 	RTM_DELSIDLIST			= 4001,
+	RTM_NEWNHGFIB			= 5000,
+	RTM_DELNHGFIB			= 5001,
 };
 
 /* Custom Netlink attribute types */
@@ -130,7 +144,7 @@ enum custom_rtattr_encap_srv6 {
 	FPM_ROUTE_ENCAP_SRV6_ENCAP_UNSPEC		= 0,
 	FPM_ROUTE_ENCAP_SRV6_VPN_SID			= 1,
 	FPM_ROUTE_ENCAP_SRV6_ENCAP_SRC_ADDR		= 2,
-	FPM_ROUTE_ENCAP_SRV6_PIC_ID			= 3,
+	FPM_ROUTE_ENCAP_SRV6_NH_RECEIVED_ID		= 3,
 	FPM_ROUTE_ENCAP_SRV6_NH_ID  			= 4,
 	FPM_ROUTE_ENCAP_SRV6_ENCAP_SIDLIST_NAME		= 5,
 	FPM_ROUTE_ENCAP_SRV6_ENCAP_SIDLIST_LEN		= 6,
@@ -170,6 +184,10 @@ enum custom_rtattr_srv6_localsid_action {
 	FPM_SRV6_LOCALSID_ACTION_UDT46				= 21,
 };
 
+enum custoem_rtattr_nexthop_group {
+	FPM_NHA_JSON_STR				= 2,
+};
+
 static const char *prov_name = "dplane_fpm_sonic";
 
 static atomic_bool fpm_cleaning_up;
@@ -181,6 +199,8 @@ struct fpm_nl_ctx {
 	bool connecting;
 	bool use_nhg;
 	bool use_route_replace;
+	bool use_nhg_fib;
+	enum fib_log_level fib_log_level;
 	struct sockaddr_storage addr;
 
 	/* data plane buffers. */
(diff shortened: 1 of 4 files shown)


eddieruan-alibaba (dismissed):

eddieruan-alibaba (dismissed):

src/sonic-frr/dplane_fpm_sonic/dplane_fpm_sonic.c
```diff
@@ -130,7 +144,7 @@ enum custom_rtattr_encap_srv6 {
 	FPM_ROUTE_ENCAP_SRV6_ENCAP_UNSPEC		= 0,
 	FPM_ROUTE_ENCAP_SRV6_VPN_SID			= 1,
 	FPM_ROUTE_ENCAP_SRV6_ENCAP_SRC_ADDR		= 2,
-	FPM_ROUTE_ENCAP_SRV6_PIC_ID			= 3,
+	FPM_ROUTE_ENCAP_SRV6_NH_RECEIVED_ID		= 3,
```
venkit-nexthop, line 147: Is RECEIVED same as ORIGINAL described in the HLD?
If yes can we please keep names consistent with the documentation. Can you either change this, or the document (all locations, so that it is not confusing).

src/sonic-frr/dplane_fpm_sonic/dplane_fpm_sonic.c
```diff
@@ -2582,26 +3014,44 @@ static int fpm_nl_enqueue(struct fpm_nl_ctx *fnc, struct zebra_dplane_ctx *ctx)
...
+					 __func__);
+				return 0;
+			}
+		} else {
+			rv = netlink_nexthop_msg_encode(RTM_DELNEXTHOP, ctx, nl_buf,
+							sizeof(nl_buf), true);
+			if (rv <= 0) {
+				zlog_err("%s: netlink_nexthop_msg_encode failed",
+					 __func__);
+				return 0;
```
venkit-nexthop, line 3222: This is missing the `dplane_ctx_set_status(ctx, ZEBRA_DPLANE_REQUEST_FAILURE);` which was present earlier. This is also true later below. Why is that?

src/sonic-frr/dplane_fpm_sonic/dplane_fpm_sonic.c
```diff
@@ -1106,6 +1199,192 @@ static bool has_srv6_localsid_nexthop(struct zebra_dplane_ctx *ctx)
...
+		/* set nexthop type */
+		c_nhg->type = nh->type;
+
+		is_recurisve = true;
+	}
+
+	/* set nh_grp_full_list */
+	const struct nh_grp_full *nh_grp_full_list = dplane_ctx_get_nhe_nh_grp_full(ctx);
+	for (uint32_t i = 0; i < dplane_ctx_get_nhe_nh_grp_full_count(ctx); i++) {
+		c_nhg->nh_grp_full_list[i].id = nh_grp_full_list[i].id;
```
venkit-nexthop, line 1242: bounds check for c_nhg ?

src/sonic-frr/dplane_fpm_sonic/dplane_fpm_sonic.c
```diff
@@ -1517,7 +1796,7 @@ static ssize_t netlink_vpn_route_msg_encode(int cmd,
 	if (!nest)
 		return false;
 
-    if (!nl_attr_put32(&req->n, datalen, FPM_ROUTE_ENCAP_SRV6_PIC_ID, pic_id)){
+    if (!nl_attr_put32(&req->n, datalen, FPM_ROUTE_ENCAP_SRV6_NH_RECEIVED_ID, nhg_received_id)){
```
venkit-nexthop, line 1799: Does this change break existing fpmsyncd code?

src/sonic-frr/dplane_fpm_sonic/dplane_fpm_sonic.c
GaladrielZhao: Many thanks for your review. Yes, RECEIVED is the same as ORIGINAL in the HLD. I've updated every location in HLD (doc/ribfib/ribfib.md and the corresponding diagram filename) (https://github.com/sonic-net/SONiC/pull/2060) to use RECEIVED for consistency.

src/sonic-frr/dplane_fpm_sonic/dplane_fpm_sonic.c
GaladrielZhao: It was deleted by mistake. Fixed, thanks.

src/sonic-frr/dplane_fpm_sonic/dplane_fpm_sonic.c
GaladrielZhao: Thanks, added.

src/sonic-frr/dplane_fpm_sonic/dplane_fpm_sonic.c
GaladrielZhao: This won't break fpmsyncd. This PR only updates the SRv6 VPN route encoding path adapting to the RIB/FIB. The message structure itself remains unchanged and FRR upstream does not yet enable sending this message, so with this change fpmsyncd won't see any new behavior. The downstream handling function in fpmsyncd will be updated in the follow-up PRs to complete the RIB/FIB logic implementation.

eddieruan-alibaba:
@venkit-nexthop  can you help to review it again? @GaladrielZhao  has already addressed all your comments.

src/sonic-frr/dplane_fpm_sonic/dplane_fpm_sonic.c
```diff
@@ -1106,6 +1199,192 @@ static bool has_srv6_localsid_nexthop(struct zebra_dplane_ctx *ctx)
...
+
+		/* set nexthop type */
+		c_nhg->type = nh->type;
+
+		is_recurisve = true;
+	}
+
+	/* set nh_grp_full_list */
+	const struct nh_grp_full *nh_grp_full_list = dplane_ctx_get_nhe_nh_grp_full(ctx);
+	for (uint32_t i = 0; i < dplane_ctx_get_nhe_nh_grp_full_count(ctx); i++) {
```
selva-nexthop, line 1241: Here we copy `nh_grp_full_list` / depends / dependents without checking against the array sizes on the `C_NextHopGroupFull` side. Zebra already has overflow checks — we should probably do the same here so we don't write past the end of the struct.

src/sonic-frr/dplane_fpm_sonic/dplane_fpm_sonic.c
```diff
@@ -1106,6 +1199,192 @@ static bool has_srv6_localsid_nexthop(struct zebra_dplane_ctx *ctx)
...
+	c_nhg->key = key;
+
+	/* set nhg_flags */
+	c_nhg->nhg_flags = dplane_ctx_get_nhe_nhg_flags(ctx);
+	bool is_recurisve = false;
+	/*
+	 * For recursive NH, we keep the nexthop information for convergence handling
+	 */
+	if (CHECK_FLAG(c_nhg->nhg_flags, NEXTHOP_GROUP_RECURSIVE)) {
+		const struct nexthop *nh = nhg->nexthop;
```
selva-nexthop, line 1230: For recursive NHGs we use `nhg->nexthop` without a NULL check. If that group ever comes through empty, zebra will crash. A quick guard would be good.

src/sonic-frr/dplane_fpm_sonic/dplane_fpm_sonic.c
```diff
@@ -3645,20 +4121,67 @@ static int fpm_nl_new(struct event_loop *tm)
...
+    if (level < current_log_level &&  !IS_ZEBRA_DEBUG_FPM) {
+        return;
+    }
+
+    /*
+     * Direct stderr — FRR convention for FIB path debugging
+     * We can't use FRR's zlog here because no API for forwarding va_list, and
+     * we want to preserve the original log level and formatting as much as possible.
+     */
+    fprintf(stderr, "[ZEBRA:FIB] %s:%d (%s) ", file, line, func);
```
selva-nexthop, line 4163: The FIB log path is writing straight to `fprintf(stderr)`. I understand the va_list limitation with zlog, but this will bypass the normal FRR log file / syslog setup and be annoying to operate. Could we format into a buffer and then call `zlog_debug`?

src/sonic-frr/dplane_fpm_sonic/dplane_fpm_sonic.c
```diff
@@ -1106,6 +1199,192 @@ static bool has_srv6_localsid_nexthop(struct zebra_dplane_ctx *ctx)
...
+	for (uint32_t i = 0; i < dplane_ctx_get_nhe_dependents_count(ctx); i++) {
+		c_nhg->dependents[i] = dependents[i];
+	}
+
+	/* set nexthop srv6 information if present */
+	if (nh->nh_srv6 != NULL) {
+		/* Get the default SRv6 context which contains seg6's src IP */
+		struct zebra_srv6 *srv6 = zebra_srv6_get_default();
+
+		c_nhg->nh_srv6 = malloc(sizeof(struct C_nexthop_srv6));
```
selva-nexthop, line 1334: If `malloc` for `nh_srv6` / `seg6_segs` fails, we keep going with a partial object and still try to encode JSON. Better to fail the encode and mark the dplane request as failed in that case.

src/sonic-frr/patch/series
```diff
@@ -85,3 +85,4 @@
 0114-Provide-interface-when-installing-uninstalling-SRv6-uA-SIDs.patch
 0115-zebra-Update-promiscuity-flag-silently-without-route.patch
 0116-zebra-keep-route_node-lock-in-NB-RIB-route-lookup_next.patch
+0117-Add-Dplane-definitions-and-functions-for-ribfib.patch
```
selva-nexthop, line 88: The embedded 0117 patch is only a subset of FRR#21415, and it already diverges a bit (`OPTION_NHG_FIB` is 2005 here vs 2004 upstream because of `kernel-mac-ext-learn`). How are we planning to keep these in sync as 21415 moves?

src/sonic-frr/dplane_fpm_sonic/dplane_fpm_sonic.c
```diff
@@ -170,6 +184,10 @@ enum custom_rtattr_srv6_localsid_action {
 	FPM_SRV6_LOCALSID_ACTION_UDT46				= 21,
 };
 
+enum custoem_rtattr_nexthop_group {
```
selva-nexthop, line 187: Small nits while I'm here: `custoem_rtattr_nexthop_group` and `is_recurisve` look like typos. Also `FPM_NHA_JSON_STR = 2` with nothing at 1 — intentional so we leave room for `NHA_ID`?

PR #26487, opened 2026-03-31
rules/config: set DEFAULT_PASSWORD to empty for improved security

xq9mend:
#### Why I did it

The hardcoded default password `YourPaSsWoRd` in `rules/config` is a well-known credential. Leaving it unchanged after deployment is a security risk. Setting it to empty requires operators to configure a real password before SSH password login works.

##### Work item tracking
- Microsoft ADO:

#### How I did it

Set `DEFAULT_PASSWORD =` (empty) in `rules/config`. With an empty password: console login works without a password, SSH password login is blocked (`PermitEmptyPasswords no`), and users must set a real password before remote access is possible.

#### How to verify it

Build a VS image and verify the admin account has no password on fresh install. Verify SSH password login is rejected until a password is set with `passwd`.

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511

- [ ] master

Set DEFAULT_PASSWORD to empty to require explicit password configuration on new deployments.

N/A


maxtara (approved):

lguohan (commented):
There are lots of downstream dependency on this. Suggest not to change this behavior.

lguohan (changes requested):

PR #26498, opened 2026-04-01
[slave.mk]: Add linux-kbuild install dependency to installer target

### LINKED ISSUE #26483: Regression: Incremental rebuild broken for signed images
### Is it platform specific

generic

### Importance or Severity

Medium

### Previous Working Version

202505

### Steps to Reproduce

In 202505 and earlier, incremental rebuild of the final sonic.bin file was possible, which enabled very fast build of multiple images with slightly different build flags. For example, building ZTP and non-ZTP images in the same workspace was possible with only a few extra minutes of build time:

```
make init
make configure
make ENABLE_ZTP=y all
mv target/sonic.bin target/sonic.ztp.bin
make ENABLE_ZTP=n all # This step only takes 10 minutes because only .bin repackaging happens
```

From 202511 onwards, this functionality is broken when image signing is also performed, as the `sign-file` kernel script is missing from the sonic-slave docker _container_ on the second bin build only. 

This appears to be due to the lack of the linux-kbuild package in the sonic-slave-trixie docker _image_.

Prior to trixie, the linux-kbuild package, which provides the `sign-file` utility, was installed into the sonic-slave docker _image_ directly by the `sonic-slave-*/Dockerfile.j2` files:

https://github.com/sonic-net/sonic-buildimage/blob/8e65b0854512fde4be8afd4737bf69a48b05565e/sonic-slave-bookworm/Dockerfile.j2#L238

However, the linux-kbuild install is missing from `sonic-slave-trixie/Dockerfile.j2`.

Instead, the linux-kbuild package is installed into the sonic-slave-trixie _container_ during the build process, specifically in by the build_debian.sh script during the creation of the `sonic-<PLATFORM>.bin__<PLATFORM>__rfs.squashfs` target:

https://github.com/sonic-net/sonic-buildimage/blob/8e65b0854512fde4be8afd4737bf69a48b05565e/build_debian.sh#L168-L169

This works fine for the first build. But, when building a second .bin file with slightly different flags (e.g. disabling ZTP), the rfs.squashfs target is not rebuilt, so the linux-kbuild package is never installed into the sonic-slave-trixie container. 

Then, when the image signing logic is invoked during the second bin generation, the requisite `sign-file` binary is not present in the container, causing the build to fail:

```
ERROR: LOCAL_SIGN_FILE=/usr/lib/linux-kbuild-6.12.41+deb13/scripts/sign-file file does not exist
```

I see a few potential ways to resolve this:

1. Install the linux-kbuild package in the sonic-slave-trixie docker _image_, if possible
2. Add logic in the .bin generation step to install linux-kbuild into the sonic-slave container
3. Force the rfs.squashfs target to rebuild when a new bin file is generated, or at least confirm the lack of squashfs rebuild is intentional (rather than due to a missing dependency)

### Impact of this regression

Incremental rebuild of the .bin file is not possible when image signing is needed as well. This substantially increases build time when producing, for example, ZTP and non-ZTP images from the same sources

### Relevant log output

```shell

```

```shell

```

### Attach files (if any)

_No response_

Bojun-Feng:
#### Why I did it

Fix https://github.com/sonic-net/sonic-buildimage/issues/26483

Currently, the `linux-kbuild` package (which provides the `sign-file` utility) is only installed into the sonic-slave container during the `rfs.squashfs` build step in `build_debian.sh`.

On incremental rebuilds, the `rfs.squashfs` target is not rebuilt, so `linux-kbuild` is never re-installed into the container, causing the signing step to fail with `ERROR: LOCAL_SIGN_FILE=... file does not exist`.

#### How I did it
* Added `$(LINUX_KBUILD)-install` as a dependency to the `$(SONIC_INSTALLERS)` target in `slave.mk`
* This ensures the `linux-kbuild` package is installed into the sonic-slave container before `.bin` generation, regardless of whether the `rfs.squashfs` target was rebuilt
* This approach was suggested by @saiarcot895 as the preferred fix, since it records the dependency explicitly in make rather than relying on a side effect of the squashfs build step

#### How to verify it
1. Build a signed image: `make ENABLE_ZTP=y all`
2. Move the output: `mv target/sonic.bin target/sonic.ztp.bin`
3. Build again with different flags: `make ENABLE_ZTP=n all`
4. Verify the second build completes without `sign-file` errors

* Fixed incremental rebuild failure for signed images by adding linux-kbuild install dependency to the installer target

### DIFF
--- a/build_debian.sh
+++ b/build_debian.sh
@@ -165,9 +165,6 @@ fi
 ## Update initramfs for booting with squashfs+overlay
 cat files/initramfs-tools/modules | sudo tee -a $FILESYSTEM_ROOT/etc/initramfs-tools/modules > /dev/null
 
-## Install kbuild for sign-file into docker image (not fsroot)
-sudo LANG=C DEBIAN_FRONTEND=noninteractive apt -y --allow-downgrades install ./$debs_path/linux-kbuild-${LINUX_KERNEL_VERSION}*_${CONFIGURED_ARCH}.deb
-
 ## Hook into initramfs: change fs type from vfat to ext4 on arista switches
 sudo mkdir -p $FILESYSTEM_ROOT/etc/initramfs-tools/scripts/init-premount/
 sudo cp files/initramfs-tools/arista-convertfs $FILESYSTEM_ROOT/etc/initramfs-tools/scripts/init-premount/arista-convertfs
--- a/slave.mk
+++ b/slave.mk
@@ -1520,7 +1520,8 @@ $(addprefix $(TARGET_PATH)/, $(SONIC_INSTALLERS)) : $(TARGET_PATH)/% : \
         $(addprefix $(PYTHON_WHEELS_PATH)/,$(SONIC_YANG_MGMT_PY3)) \
         $(addprefix $(PYTHON_WHEELS_PATH)/,$(SYSTEM_HEALTH)) \
         $(addprefix $(PYTHON_WHEELS_PATH)/,$(SONIC_HOST_SERVICES_PY3)) \
-        $$(addprefix $(TARGET_PATH)/,$$($$*_RFS_DEPENDS))
+        $$(addprefix $(TARGET_PATH)/,$$($$*_RFS_DEPENDS)) \
+        $(addprefix $(IMAGE_DISTRO_DEBS_PATH)/,$(LINUX_KBUILD)-install)
 
 	$(HEADER)
 	# Pass initramfs and linux kernel explicitly. They are used for all platforms


Bojun-Feng:
Update: Re-run build locally, verified that the `kbuild` package is indeed rebuilt on incremental rebuilds. Unfortunately I made new clones and forgot to enable signing on them, but now we know the fix works at the dependency level.

<details>
<summary>More Details</summary>

### Setup

Run these in two different directories
```
# Master head on 03/31/2026
git clone https://github.com/sonic-net/sonic-buildimage
git checkout b9d726a3710da24b605353359fa891b3b6375e1e 

# This PR, exactly one commit ahead of commit above
git clone https://github.com/Bojun-Feng/sonic-buildimage
git checkout bug/fix-incremental-rebuild-kbuild
git checkout 8c4302591ea58c256e2219fbfccd21d628292481
```

### Commands

Run the same commands in both clones
```
git submodule update --init --recursive

make init &&  \
make configure \
  PLATFORM=vs

# By default ENABLE_ZTP=n
make \
  SONIC_BUILD_JOBS=4 \
  SONIC_BUILD_RETRY_COUNT=1 \
  target/sonic-vs.img.gz

# Remove artifact, rebuild with ENABLE_ZTP=y
mv ./target/sonic-vs.img.gz ./target/sonic-vs.noztp.img.gz

make \
  SONIC_BUILD_JOBS=4 \
  SONIC_BUILD_RETRY_COUNT=1 \
  ENABLE_ZTP=y \
  target/sonic-vs.img.gz
```

In master on rebuild:
```
[ 01 ] [ target/debs/trixie/sonic-ztp_1.0.0_all.deb ]
[ 01 ] [ target/sonic-vs.img.gz ]
```

In PR on rebuild:
```
[ 01 ] [ target/debs/trixie/sonic-ztp_1.0.0_all.deb ]
[ 02 ] [ target/debs/trixie/linux-kbuild-6.12.41+deb13_6.12.41-1_amd64.deb-install ]
[ 01 ] [ target/sonic-vs.img.gz ]
```
</details>

I'm removing the `apt install linux-kbuild` now. Will make a new clone and build again with signing enabled to ensure this PR fixes the `sign-file` issue.

Bojun-Feng:
> Thanks @Bojun-Feng! Weirdly, I'm no longer able to reproduce this issue when testing with the baseline `master` code. Not sure what happened... Did you experience this at all in your testing?

I've been using commit https://github.com/sonic-net/sonic-buildimage/commit/b9d726a3710da24b605353359fa891b3b6375e1e, which was the master head when this PR was first created:

```git checkout b9d726a3710da24b605353359fa891b3b6375e1e```

When building on that commit without the PR fix, I ran into a different docker timeout issue that resolved itself after waiting and building again with auto-retry. On the second build I was able to replicate the issue.

Additionally, there is an issue with Rust blocking builds with `docker-restapi-watchdog` (it's mentioned above). My virtual switch builds don't include them, but they are blocking the CI/CD pipelines.

I haven't tested on the current master or 202505/202511 branches. Hope this helps.

jusherma (approved):

saiarcot895 (approved):

sonic-ci-gen:
Seem all checked have passed. Could this be merged?

jusherma:
@qiluo-msft @xumia @lguohan Code has been approved and checks have passed. Can you please merge?

jusherma:
> @qiluo-msft @xumia @lguohan Code has been approved and checks have passed. Can you please merge?

@saiarcot895 Can you help?

liushilongbuaa (approved):
Approved.

liushilongbuaa:
Building squashfs only takes 5 minutes. You can also try 'rm -rf target/*.squashfs' to disable cache when configuration changed.

PR #26506, opened 2026-04-01
[snmp] Set v1trapaddress to avoid expensive get_myaddr() on every trap

mike-dubrovsky:
#### Why I did it
During AZD testing on a device with 2K interfaces and 2K BGP peers, BGP was spending the majority of its CPU time sending SNMP traps.
image

#### How I did it

When net-snmp lib sends a trap notification, it unconditionally discovers a local IP address to populate the agent_addr field in the SNMPv1 TRAP PDU. This calls [get_myaddr()](https://github.com/net-snmp/net-snmp/blob/e8a9e5232842b793c6ae51663f6bd3ebfac3fdf5/agent/agent_trap.c#L978
), which opens a socket and enumerates all network interfaces via multiple ioctl(SIOCGIFCONF/SIOCGIFFLAGS) on every trap. 
The agent_addr field only exists in SNMPv1 TRAP PDUs; v2c/v3 does not have this field.

This is a significant bottleneck when the router has thousands of interfaces.

Hardcoded  v1trapaddress to a link-local address (169.254.0.1) so net-snmp does not need to look up a local address on every trap.

- docker-fpm-frr/snmp.conf: unconditional, since AgentX discards the agent_addr field before forwarding to the snmp container.
- docker-snmp/snmpd.conf.j2: conditional, only when no v1 trap receiver is configured. When a v1 receiver exists, snmpd resolves the real address.

#### How to verify it
Configured a v1 trap receiver:
  config snmptrap modify 1 10.0.0.1
Started
  tcpdump -i any port 162 -v
Executed 
 clear ip bgp *
Verified that:
1) No change in the trap agent_addr field after changes to the FRR docker only.
2) No change in the trap agent_addr field after changes to both the FRR and SNMP dockers.

AZD deployment 

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511

- [x] master

bgp_interface_profile_lbr

### DIFF
--- a/dockers/docker-fpm-frr/snmp.conf
+++ b/dockers/docker-fpm-frr/snmp.conf
@@ -5,3 +5,7 @@
 # Check that a snmpwalk to 1.3.6.1.2.1.15 gives an output
 # Further verification: 1.3.6.1.2.1.15.2.0 = INTEGER: 65000 the returned value should be the confiugred ASN
 agentXSocket    tcp:localhost:3161
+# Set v1trapaddress to avoid the expensive net-snmp get_myaddr() ioctl
+# enumeration on every trap. The agent_addr field is only used in SNMPv1
+# TRAP PDUs and is discarded by AgentX before reaching the snmp container.
+v1trapaddress   169.254.0.1
--- a/dockers/docker-snmp/snmpd.conf.j2
+++ b/dockers/docker-snmp/snmpd.conf.j2
@@ -141,6 +141,14 @@ load   12 10 5
 #
 # Note: disabled snmp traps due to side effect of causing snmpd to listen on all ports (0.0.0.0)
 #
+# Set v1trapaddress to avoid the expensive net-snmp get_myaddr() ioctl
+# enumeration on every trap. The agent_addr field is only used in SNMPv1
+# TRAP PDUs; v2c/v3 does not have this field.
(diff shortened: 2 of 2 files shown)


mike-dubrovsky:
> Is 169.254.0.1 going to be always available/defined? I don't see it on a physical T0 device. Would 127.0.0.1 be better?

@saiarcot895
The address just needs to be set to something to keep the net-snmp library from calling get_myaddr(). It is never actually sent on the wire:

In the FRR docker, AgentX discards the agent_addr field before forwarding to snmpd.
In the snmpd docker, v1trapaddress is only set when no v1 trap receiver is configured, so only v2c/v3 traps are sent, which do not carry the agent_addr field at all.

169.254.0.1 is from the link-local range (169.254.0.0/16, RFC 3927), but since the address is never transmitted, 127.0.0.1 would work equally well. Happy to switch to 127.0.0.1 if you prefer — it's arguably more conventional for a placeholder.

saiarcot895 (approved):

anish-n (approved):

qiluo-msft (approved):

mike-dubrovsky:
sorry for delay ... wonder if we can merge it ?

mike-dubrovsky:
@qiluo-msft @saiarcot895 
hi Qi and Saikrishna, how can we merge it ?

PR #26507, opened 2026-04-01
Fix QoS scheduler config for VOQ chassis active system ports

arlakshm:
#### Why I did it
Update qos_config.j2 to apply scheduler configuration only to active system ports (those with device neighbors) instead of all system ports. Add corresponding test minigraph XMLs and expected output JSON files for no-config and pch-config LC T2 scenarios.

Changes from Azure/sonic-buildimage-msft#1775.

##### Work item tracking
Fixes https://github.com/sonic-net/sonic-buildimage/issues/24443

#### How I did it
Add the schedule configuration only on system port when the port is active

#### How to verify it
UT

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511

- [ ] 
- [ ]


saksarav-nokia (approved):
LGTM

PR #26514, opened 2026-04-01
[watchdog]: Add Cargo.lock files for restapi and gnmi watchdogs

Bojun-Feng:
#### Why I did it

Fix https://github.com/sonic-net/sonic-buildimage/issues/26513

`docker-restapi-watchdog` build fails on current master because the `icu_*` crates released v2.2.0 with MSRV 1.86, while the Dockerfile pins `rustc` to 1.83.0. We need a `Cargo.lock` to pin the `icu_*` version as well.

`docker-gnmi-watchdog` has the same missing lockfile and is vulnerable to identical breakage from any future MSRV bump in its dependency tree.

Note `docker-bmp-watchdog` and `docker-telemetry-watchdog` already commit `Cargo.lock` files and are unaffected — this PR brings the remaining two watchdogs in line with that convention.

#### How I did it
* Generated `Cargo.lock` for `dockers/docker-restapi-watchdog/watchdog/` and downgraded `icu_*` crates from v2.2.0 to v2.1.x (MSRV 1.83) to match the Dockerfile-pinned `rustc` version.
* Generated `Cargo.lock` for `dockers/docker-gnmi-watchdog/watchdog/`.

#### How to verify it
CI/CD builds for Broadcom and marvell_prestera_armhf targets should pass.

Add Cargo.lock files for docker-restapi-watchdog and docker-gnmi-watchdog to prevent build failures from transitive MSRV bumps.


Bojun-Feng:
Seems like all builds in pipeline passed. Opening this up for review

PR #26544, opened 2026-04-03
Generic changes and new py-common API's for BMC based platforms

judyjoseph:
#### Why I did it
Updates in sonic-buildimage changes based on : [pmon-bmc-design](https://github.com/sonic-net/SONiC/blob/master/doc/bmc/sonicBMC/pmon-bmc-design.md)

##### Work item tracking

#### How I did it

Some of the changes 

*   Extend `sonic-py-common` with platform APIs
    *   Add switch/BMC role and cooling type helpers
    *   Refactor BMC address resolution via `bmc.json`
    *   Support direct TCP Redis access via `db_connect_remote`
    *   Add unit test coverage for all new APIs
   
*   Define global BMC network config in `bmc.json`

*   Add BMC event logging and tmpfs `/var/log`
    *   Persist BMC event logs on eMMC-backed `/host`
    *   Enable log rotation and force logs-in-RAM on BMC
    
*   Remove `DATABASE_TYPE="bmcdb"` and related guards
    *   Eliminate bmcdb-specific logic from database init and templates
    *   Allow `redis_bmp` and `BMP_STATE_DB` to follow standard device flow
    
*   Detect Switch Host vs Switch BMC via `platform_env.conf`
    *   Derive `IS_SWITCH_HOST` / `IS_SWITCH_BMC`
    *   Reuse existing database init logic without new env vars
    
*   Update pmon for BMC role
    *   Enable `bmcctld` and suppress `chassisd` on Switch BMC
    *   Propagate `IS_SWITCH_BMC` into supervisor configuration

*   Adjust pmon daemon control for arm64 BMC platforms
    *   Enable minimal required daemons
    *   Disable chassis-, optics-, PSU-, fan-, and sensor-related services

#### How to verify it
Verifyed bringing up the bmc eval board and make sure the pmon docker stays up

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511

- [ ] 
- [ ]


judyjoseph:
@chander-nexthop a related PR to your changes please review

files/initramfs-tools/union-mount.j2
```diff
@@ -54,6 +54,15 @@ for x in $(cat /proc/cmdline); do
     esac
 done
 
+# On Switch-BMC systems /var/log is always on tmpfs to avoid wearing out
+# the eMMC flash with high-frequency syslog writes.
+_platform=$(grep '^onie_platform=' ${rootmnt}/host/machine.conf 2>/dev/null | cut -d= -f2)
+_penv="${rootmnt}/usr/share/sonic/device/${_platform}/platform_env.conf"
+if [ -f "$_penv" ] && grep -q '^switch_bmc=1' "$_penv" 2>/dev/null; then
```
judyjoseph, line 61: is there any other construct I can use to check if this is is_bmc()  ? not sure if platform_env file is available at this point with intiramfs

dockers/docker-platform-monitor/docker-pmon.supervisord.conf.j2
```diff
@@ -40,7 +57,7 @@ startsecs=0
 dependent_startup=true
 {% endif %}
 
-{% if not skip_chassisd and (IS_MODULAR_CHASSIS == 1 or is_smartswitch) %}
+{% if not skip_chassisd and (IS_MODULAR_CHASSIS == 1 or is_smartswitch) and not IS_SWITCH_BMC %}
```
chander-nexthop: Do we really need the IS_SWICTH_BMC check here? We disable chasssid in pmon_daemon_control.json?

dockers/docker-platform-monitor/docker_init.j2
```diff
@@ -140,7 +140,17 @@ if [ -e $MODULAR_CHASSISDB_CONF_FILE ] && [[ $disaggregated_chassis -ne 1 ]]; th
     IS_MODULAR_CHASSIS=1
 fi
 
-confvar="{\"HAVE_SENSORS_CONF\":$HAVE_SENSORS_CONF, \"HAVE_FANCONTROL_CONF\":$HAVE_FANCONTROL_CONF, \"API_VERSION\":$SONIC_PLATFORM_API_PYTHON_VERSION, \"IS_MODULAR_CHASSIS\":$IS_MODULAR_CHASSIS}"
+# Determine if this pmon instance is running on the Switch BMC.
+# switch_bmc=1 is set in platform_env.conf on BMC systems.
```
chander-nexthop, line 152: Are you planning to add switch_bmc=1 to device/nexthop/arm64-nexthop_b27-r0/platform_env.conf. I dont see that change in this PR?

files/image_config/constants/bmc.json
```diff
@@ -0,0 +1,6 @@
+{
+    "bmc_if_name": "bmc0",
+    "bmc_if_addr": "169.254.100.2",
+    "bmc_addr": "169.254.100.1",
+    "bmc_net_mask": "255.255.255.252"
+}
```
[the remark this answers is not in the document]
chander-nexthop: Should we just have platform specific json file or are you planning to take care of this point above?

files/initramfs-tools/union-mount.j2
chander-nexthop: /host/machine.conf will have MACHINE="aspeed_ast2700". This can be used?

judyjoseph: @chander-nexthop I feel we should not bind to aspeed .. if any vendor/partner starts to use a different chip i.so aspeed.

files/image_config/constants/bmc.json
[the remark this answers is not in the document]
judyjoseph: that was missed .. adding it - thanks for pointing out

files/build_templates/docker_image_ctl.j2
```diff
@@ -433,6 +431,17 @@ start() {
...
+    # Normalise to 0 when absent so downstream env vars are always defined.
+    IS_SWITCH_HOST=${switch_host:-0}
+    IS_SWITCH_BMC=${switch_bmc:-0}
+
+    # Derive IS_BMC_DEVICE from switch_bmc so docker-database-init.sh can use
+    # the existing IS_BMC_DEVICE mechanism without extra env vars.
+    if [[ "$IS_SWITCH_BMC" == "1" ]]; then
+        IS_BMC_DEVICE=true
+    fi
+
```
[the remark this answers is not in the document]
judyjoseph: @Copilot can you check again it gets passed when we create docker

src/sonic-yang-models/yang-models/sonic-chassis-module.yang
```diff
@@ -20,16 +20,26 @@ module sonic-chassis-module {
                 key "name";
                 leaf name {
                     type string {
-                        pattern "LINE-CARD[0-9]+|FABRIC-CARD[0-9]+|DPU[0-9]+";
+                        pattern "LINE-CARD[0-9]+|FABRIC-CARD[0-9]+|DPU[0-9]+|SWITCH-HOST";
                     }
-                    description "Line-card or fabric-card or dpu module name";
+                    description "Line-card, fabric-card, dpu, or switch-host module name";
```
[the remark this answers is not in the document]
judyjoseph: @copilot apply changes based on [this feedback](https://github.com/sonic-net/sonic-buildimage/pull/26544#discussion_r3102870006)

files/initramfs-tools/union-mount.j2
chander-nexthop: That time, the condition can be augmented. What I meant to say was, if we see aspeed we can assume BMC

judyjoseph:
@qiluo-msft  could you review the Yang model changes please thx

dockers/docker-platform-monitor/docker-pmon.supervisord.conf.j2
```diff
@@ -31,6 +31,23 @@ stderr_syslog=true
...
+autostart=false
+autorestart=unexpected
+stdout_logfile=NONE
+stdout_syslog=true
+stderr_logfile=NONE
+stderr_syslog=true
+startsecs=10
+startretries=50
+dependent_startup=true
+dependent_startup_wait_for=rsyslogd:running
```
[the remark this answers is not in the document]
judyjoseph: @copilot apply changes based on [this feedback](https://github.com/sonic-net/sonic-buildimage/pull/26544#discussion_r3140482358)

judyjoseph:
@copilot apply changes based on the comments in [this thread](https://github.com/sonic-net/sonic-buildimage/pull/26544#pullrequestreview-4173614591)

rules/sonic-bmcctld.dep
```diff
@@ -0,0 +1,16 @@
+SPATH       := $($(SONIC_BMCCTLD_PY2)_SRC_PATH)
+DEP_FILES   := $(SONIC_COMMON_FILES_LIST) rules/sonic-bmcctld.mk rules/sonic-bmcctld.dep
+DEP_FILES   += $(SONIC_COMMON_BASE_FILES_LIST)
+SMDEP_FILES := $(addprefix $(SPATH)/,$(shell cd $(SPATH) && git ls-files))
+
+$(SONIC_BMCCTLD_PY2)_CACHE_MODE  := GIT_CONTENT_SHA
```
nate-nexthop: Should we remove all the PY2 references? Looks like it's dead code.

rules/sonic-bmcctld.mk
```diff
@@ -0,0 +1,19 @@
+# sonic-bmcctld (SONiC BMC control daemon) Debian package
+
+# SONIC_BMCCTLD_PY2 package
+
+SONIC_BMCCTLD_PY2 = sonic_bmcctld-1.0-py2-none-any.whl
```
nate-nexthop: And here can we also remove PY2 references?

src/sonic-py-common/sonic_py_common/device_info.py
```diff
@@ -693,6 +694,51 @@ def is_dpu():
     return False
 
 
+def is_switch_host():
```
nate-nexthop: These 2 functions appear to be identical except for a string constant. Could the parsing logic be separated into a separate function that takes the desired string constant as an argument?

files/initramfs-tools/union-mount.j2
```diff
@@ -54,6 +54,15 @@ for x in $(cat /proc/cmdline); do
...
+    logs_inram=true
```
judyjoseph, line 62: Do we need this change ?

nate-nexthop:
Thanks for addressing the comments!

judyjoseph:
There are failures in this PR as we need to get the **new bmcctld** daemon https://github.com/sonic-net/sonic-platform-daemons/pull/782 first in sonic-platform-daemons before restarting this build run again.

files/image_config/interfaces/interfaces.j2
```diff
@@ -30,12 +30,12 @@ iface lo inet loopback
 {% endblock loopback %}
 {% block mgmt_interface %}
 
-{%- if DEVICE_METADATA and 'bmc' in DEVICE_METADATA.keys() and ('bmc_if_name' in DEVICE_METADATA['bmc']) and ('bmc_if_addr' in DEVICE_METADATA['bmc']) and ('bmc_net_mask' in DEVICE_METADATA['bmc']) %}
+{%- if bmc_if_name is defined and bmc_if_addr is defined and bmc_net_mask is defined %}
```
judyjoseph: @benle7 could you review this change and make sure it is ok ? I have replaced the BMC IP getting filled into DEVICE_METADATA ( not a usual practice in SONiC ), instead parsing it in files/image_config/interfaces/interfaces-config.sh

roger-nexthop (commented):
Thanks Judy, looks good to me.

files/initramfs-tools/union-mount.j2
chander-nexthop: As we are setting in the command line the expliciy check here is not needed in the common case. Need to think a little bit about our first boot. The script we use internally to flash the 1st image into emmc, also setups the uboot environment such that the sonic kernel is booted with command line argument "logs_inram=on". Not sure about other vendors. If everyone can ensure this, then this hunk is not needed

judyjoseph: Ok, I would keep it here to make it generic across all platforms.

yxieca (dismissed):

dockers/docker-platform-monitor/docker_init.j2
judyjoseph: already present

files/build_templates/sonic_debian_extension.j2
```diff
@@ -782,6 +782,11 @@ sudo cp $IMAGE_CONFIGS/snmp/snmp.yml $FILESYSTEM_ROOT/etc/sonic/
 # Copy ASN configuration files
 sudo cp $IMAGE_CONFIGS/constants/constants.yml $FILESYSTEM_ROOT/etc/sonic/
 
+# Copy BMC network configuration file (aspeed BMC platform only)
+if [[ $CONFIGURED_PLATFORM == aspeed ]]; then
+    sudo cp $IMAGE_CONFIGS/constants/bmc.json $FILESYSTEM_ROOT/etc/sonic/
+fi
```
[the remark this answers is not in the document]
judyjoseph: ignoring this .. currently we are checking for aspeed to identify and copy bmc.json file

judyjoseph:
@benle7 @oleksandrivantsiv Could you reveiew this PR, just wanted to make sure it doesn't break any existing bmc infra already present , especially changes relating to bmc.json file

files/build_templates/sonic_debian_extension.j2
```diff
@@ -782,6 +782,11 @@ sudo cp $IMAGE_CONFIGS/snmp/snmp.yml $FILESYSTEM_ROOT/etc/sonic/
 # Copy ASN configuration files
 sudo cp $IMAGE_CONFIGS/constants/constants.yml $FILESYSTEM_ROOT/etc/sonic/
 
+# Copy BMC network configuration file (aspeed BMC platform only)
+if [[ $CONFIGURED_PLATFORM == aspeed ]]; then
```
benle7: A few questions regarding this global bmc.json:
- Why do we copy it only for the ASPEED case?
- Does it work for all ASPEED BMC vendors (the folders under platform/aspeed/)?
- If a vendor wants to override it, do they need to create their own bmc.json inside the internal BMC platform folder (platform/aspeed/sonic-platform-modules-<vendor>)?

src/sonic-py-common/sonic_py_common/device_info.py
```diff
@@ -985,18 +1012,65 @@ def is_warm_restart_enabled(container_name):
...
+    if absent, falls back to the SONiC-wide global bmc.json (/etc/sonic/bmc.json).
+    This allows a platform to override the global defaults with platform-specific
+    IP addresses.
+
+    Returns:
+        A dict with bmc_if_name, bmc_if_addr, bmc_addr and bmc_net_mask,
+        or None if no bmc.json is found.
+    """
     try:
         platform_path = get_path_to_platform_dir()
```
benle7: Was `get_path_to_platform_dir()` updated to return the internal BMC ASPEED platform folder (where the global `bmc.json` should be overridden) when running on the BMC side?

files/image_config/interfaces/interfaces.j2
```diff
@@ -30,12 +30,20 @@ iface lo inet loopback
...
 {% block mgmt_interface %}
 
-{%- if DEVICE_METADATA and 'bmc' in DEVICE_METADATA.keys() and ('bmc_if_name' in DEVICE_METADATA['bmc']) and ('bmc_if_addr' in DEVICE_METADATA['bmc']) and ('bmc_net_mask' in DEVICE_METADATA['bmc']) %}
-# BMC interface
-auto {{ DEVICE_METADATA['bmc']['bmc_if_name'] }}
-iface {{ DEVICE_METADATA['bmc']['bmc_if_name'] }} inet static
-    address {{ DEVICE_METADATA['bmc']['bmc_if_addr'] }}
-    netmask {{ DEVICE_METADATA['bmc']['bmc_net_mask'] }}
+{%- if bmc_if_name is defined and bmc_net_mask is defined %}
+{%- if IS_SWITCH_BMC == 1 and bmc_addr is defined %}
```
benle7, line 34: Is it guaranteed that either `IS_SWITCH_HOST` or `IS_SWITCH_BMC` will be `1`?
Is there a case where both are `0`?
What happens when running on the host side of a platform without a BMC?

files/image_config/constants/bmc.json
benle7, line 6: Should this file be used only on the BMC side, or on both the host and BMC sides?

I'm asking because there are places that rely on the existence of `platform/bmc.json` to determine whether BMC is supported/existing.

For example:
`src/sonic-utilities/scripts/generate_dump:is_bmc_supported`
(It searches for `/usr/share/sonic/device/$platform/bmc.json`.)

files/image_config/interfaces/interfaces.j2
```diff
@@ -30,12 +30,20 @@ iface lo inet loopback
...
+# BMC interface (Switch-BMC side)
+auto {{ bmc_if_name }}
+iface {{ bmc_if_name }} inet static
+    address {{ bmc_addr }}
```
benle7, line 38: Just to confirm, is the plan for configuring the `usb0` (`bmc0`) interface on the BMC side
the same as on the host side?
That is, write the interface configuration to `/etc/network/interfaces`
and let `networking.service` bring it up?

Does this force all vendors (on the BMC side) to configure the interface this way?
If they do not add a platform-specific bmc.json, will it still use the global bmc.json?

files/initramfs-tools/union-mount.j2
oleksandrivantsiv: @judyjoseph, we can define a new variable in '/host/machine.conf' that all vendors can set. We have a single script that generates 'machine.conf': 'platform/aspeed/scripts/sonic-machine-conf-init.sh'.

benle7 (commented):
Thanks @judyjoseph 
I added a few questions.

files/build_templates/sonic_debian_extension.j2
judyjoseph: I am seeing aspeed used multiple places in this file - as we have only aspeed as supported BMC. Plan is to change it globally when we have multiple vendors.

judyjoseph: To override we need to define it in /device/platform/sku dir : eg: https://github.com/sonic-net/sonic-buildimage/blob/af8111425c0b2fd0d620d5d3a470559631cbda52/device/nexthop/arm64-nexthop_b27-r0/bmc.json#L3

files/image_config/constants/bmc.json
judyjoseph: yes idea was to use this file on both ends in SWITCH-HOST and in BMC . I see this is the only place we have a check. Will update is_bmc_supported()

files/image_config/interfaces/interfaces.j2
judyjoseph: On a platform without BMC.. we need not configure this bmc0 interface anyways ?

files/image_config/interfaces/interfaces.j2
judyjoseph: current approach is to check and use the platform-specific bmc.json (not recommended though - if there is a real need for this in that platform ) -- else fallback to global bmc.json

src/sonic-py-common/sonic_py_common/device_info.py
judyjoseph: no as shared in earlier comment .. it gets from /device/platform/sku directory

files/image_config/interfaces/interfaces-config.sh
```diff
@@ -70,7 +70,32 @@ CFGGEN_PARAMS=" \
     -t /usr/share/sonic/templates/90-dhcp6-systcl.conf.j2,/etc/sysctl.d/90-dhcp6-systcl.conf \
     -t /usr/share/sonic/templates/dhclient.conf.j2,/etc/dhcp/dhclient.conf \
 "
-sonic-cfggen $CFGGEN_PARAMS
+
+# On BMC/Switch-Host platforms, pass bmc.json and the role to sonic-cfggen
+# so interfaces.j2 can render the BMC interface stanza with the correct IP.
+#   switch_bmc=1  -> use bmc_addr  (BMC's own IP on the link)
+#   switch_host=1 -> use bmc_if_addr (Switch-Host's IP on the BMC link)
+PLATFORM_ENV_CONF="/usr/share/sonic/platform/platform_env.conf"
```
benle7: Do you mean to "device/$platform/platform_env.conf"?
Can you explain this file and what the expectations are for each platform here?
Should this file provide the indication whether the platform is BMC side or Host side (with BMC)?
Should we update the existing platform folders (host side) to include this flag?

files/build_templates/sonic_debian_extension.j2
benle7: OK so it seems "aspeed" is the CONFIGURED_PLATFORM for all platforms (in the BMC side).

files/image_config/interfaces/interfaces.j2
benle7: From the code it seems we configure the interface only when either IS_SWITCH_BMC or IS_SWITCH_HOST is 1. So for Host side of platform without BMC we will not configure the interface. Right?

files/image_config/constants/bmc.json
benle7: It seems you install the global bmc.json file only when $CONFIGURED_PLATFORM == aspeed, so it will be used only for the BMC side, right?

As I see it now, for the host side, we still need bmc.json under the platform folder.
As we have here for example:
device/mellanox/x86_64-nvidia_sn6600_ld-r0/bmc.json

For the BMC side, we can either:
Add bmc.json under the platform folder at /aspeed/nvidia/, or
Use the global /etc/sonic/bmc.json.

That's correct?

Regarding the generate_dump:is_bmc_supported, it is used for the host side when executing "show techsupport", so if on the host side we still force to have bmc.json in the platform folder, we are good.
But if we plan to use the global bmc.json also for the host side, so it should be upadted.

Yakiv-Huryk (dismissed):

files/build_templates/sonic_debian_extension.j2
benle7: @judyjoseph 
The global bmc.json file should be identical for both sides BMC and host.
In the current code, we install it only for the BMC side (CONFIGURED_PLATFORM == "aspeed").
So it required us to create an additional file for the host side and maintain 2 identical files.
Maybe we can leverage the new global file also for the host side.
WDYT?

files/image_config/constants/bmc.json
```diff
@@ -0,0 +1,6 @@
+{
+    "bmc_if_name": "bmc0",
+    "bmc_if_addr": "169.254.100.2",
```
benle7, line 3: In the original infra (and our bmc.json file) we used 169.254.0.2 169.254.0.1.
Can you apply it to this new global json?

files/image_config/constants/bmc.json
judyjoseph: Thanks for bring this into notice ! The actual idea is to use this same file in both ends in SWITCH-HOST and in BMC.  I will remove this check $CONFIGURED_PLATFORM == aspeed so that this file will be copied on both ends - it will be used if we have "Switch_host=1" defined in the platform_env.conf file to configure the host -- bmc interconnect interfaces.

We need to update the generate_dump:is_bmc_supported too.

files/image_config/interfaces/interfaces-config.sh
judyjoseph: @benle7  Updated the logic in config-setup service so that we have /etc/sonic/bmc.json in Host and BMC , and it will be based on switch_host=1/switch_bmc=1 defenitions in platform_env.conf files

benle7: Thanks!

files/image_config/constants/bmc.json
benle7: Thank you!

files/image_config/constants/bmc.json
judyjoseph: I followed similar style as in smart switch platforms, Let me know any real issue you face - else prefer to keep same - thanks

benle7: Hi @judyjoseph 

This change breaks the integration we implemented with OpenBMC, because we currently rely on these specific IP addresses and configured them on both sides - the host side and the OpenBMC side.

We need to keep the same IP addressing scheme between OpenBMC and SONiC BMC to maintain compatibility and avoid additional platform-specific handling.

So we would really appreciate it if you could align on using the same addresses as before (169.254.0.1 / 169.254.0.2).

Otherwise, we can discuss it further in the weekly meeting.

Thanks!

FYI @oleksandrivantsiv

judyjoseph: Ok the major reason going for this address was a comment from https://www.rfc-editor.org/rfc/rfc3927.html. 169.254.0.1. 169.254.0.2 falls in this range of 256 addresses which is reserved . Can you check if it is possible to change on your end ?

```
[2.1](https://www.rfc-editor.org/rfc/rfc3927.html#section-2.1).  Link-Local Address Selection

   When a host wishes to configure an IPv4 Link-Local address, it
   selects an address using a pseudo-random number generator with a
   uniform distribution in the range from 169.254.1.0 to 169.254.254.255
   inclusive.

   The IPv4 prefix 169.254/16 is registered with the IANA for this
   purpose.  The first 256 and last 256 addresses in the 169.254/16
   prefix are reserved for future use and MUST NOT be selected by a host
   using this dynamic configuration mechanism.
```

yxieca (approved):

judyjoseph:
This PR is already in 202607 branch in sonic-buildimage-msft

PR #26499, opened 2026-04-01
[gnmi] Migrate gnmi, telemetry, and watchdog containers to Debian Trixie

hdwhdw:
Part of #25959

## Summary

Migrate the gnmi, telemetry, and watchdog containers from Debian Bookworm to Debian Trixie.

Update base image references from `docker-config-engine-bookworm` to `docker-config-engine-trixie` for all gnmi-related containers. Move them from `SONIC_BOOKWORM_DOCKERS` to `SONIC_TRIXIE_DOCKERS` to align with the ongoing container migration to Debian 13 (Trixie).

`docker-telemetry-watchdog` and `docker-sonic-telemetry` are both `FROM docker-sonic-gnmi`, so they must follow it to Trixie.

This follows the same pattern as PR #25942 (Brcm containers migrate to trixie) and builds on the Trixie base layers added in PR #25184.

### Containers migrated

- `docker-sonic-gnmi`
- `docker-gnmi-sidecar`
- `docker-gnmi-watchdog`
- `docker-telemetry-watchdog`
- `docker-sonic-telemetry`

### FIPS fixes

- Updated trixie FIPS version from `1.8.0-24-gd744cf2` to `1.8.0-24-gd744cf2-2` to include the Golang FIPS fix with sonic_fips detection and symcryptprovider patches.
- Fixed trixie Go FIPS package version from `1.24.4-1` to `1.24.4-1+fips` (missing `+fips` suffix caused armhf build failure).

## Testing

- Built full VS image locally and deployed to KVM testbed (vlab-01)
- Verified gnmi container runs Debian 13 (Trixie)
- `gnmi/test_gnmi_stress.py`: **1 passed**
- `telemetry/test_telemetry_cert_rotation.py`: **4 passed**

Signed-off-by: Dawei Huang

PR #26548, opened 2026-04-03
Migrate sonic-mgmt-framework container to Debian Trixie

amrutasali:
#### How to verify it - The sonic-mgmt-framework container  builds successfully under BLDENV=trixie. Installed it on a recent SONiC image and verified that the docker os distribution shows Trixie and basic curl test on REST server hosted within the docker works fine.

**Note:** The sonic-mgmt-framework build for trixie has a dependency on below PRs  and then a submodule update for the 2 repos :
https://github.com/sonic-net/sonic-mgmt-common/pull/211 
https://github.com/sonic-net/sonic-mgmt-framework/pull/158


amrutasali:
@saiarcot895 Please add this to tracker - https://github.com/sonic-net/SONiC/issues/2169  .
Note : sonic-gnmi(bump Golang to 1.24.4) repo PR will be opened next week

amrutasali:
> > @saiarcot895 Please add this to tracker - [sonic-net/SONiC#2169](https://github.com/sonic-net/SONiC/issues/2169) . Note : sonic-gnmi(bump Golang to 1.24.4) repo PR will be opened next week
> 
> @amrutasali Please note #26499.

@saiarcot895  yes I'm aware and got it confirmed - https://github.com/sonic-net/sonic-buildimage/issues/25959#issuecomment-4174392108 .So will  just bump Golang version for soni-gnmi repository.

rules/docker-sonic-mgmt-framework.mk
```diff
@@ -38,5 +38,5 @@ $(DOCKER_MGMT_FRAMEWORK)_RUN_OPT += --mount type=bind,source="/var/platform/",ta
 
 $(DOCKER_MGMT_FRAMEWORK)_BASE_IMAGE_FILES += sonic-cli:/usr/bin/sonic-cli
```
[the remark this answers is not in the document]
amrutasali: Why the current code is correct

  • The image does build under trixie via the SONIC_DOCKER_IMAGES + filter-out mechanism
  • The image intentionally doesn't build under bookworm - it depends on DOCKER_CONFIG_ENGINE_TRIXIE (line 15), which doesn't exist in bookworm
  builds
  • SONIC_TRIXIE_DOCKERS is unused in slave.mk, but this is the same pattern used by all other trixie-migrated dockers (database, dhcp-relay,
  teamd, router-advertiser, config-engine-trixie, base-trixie, etc. - 8 other .mk files do the same thing). It's a forward-looking categorization
  for when slave.mk adds a BLDENV=trixie block.

Why the suggested change would be harmful
Putting it back in SONIC_BOOKWORM_DOCKERS would:

  • Build mgmt-framework under BLDENV=bookworm where DOCKER_CONFIG_ENGINE_TRIXIE is unavailable, breaking the bookworm build
  • Exclude it from the trixie build (the filter-out on line 1187 would remove it)

This is exactly backwards from what's needed. The Copilot missed that SONIC_DOCKER_IMAGES (line 14) is what actually controls inclusion in the
default/trixie build path, not SONIC_TRIXIE_DOCKERS. No changes needed.

amrutasali:
@saiarcot895 this is ready to be merged

rules/sonic-mgmt-common.mk
```diff
@@ -6,7 +6,9 @@ $(SONIC_MGMT_COMMON)_SRC_PATH = $(SRC_PATH)/sonic-mgmt-common
 $(SONIC_MGMT_COMMON)_DEPENDS  = $(LIBYANG_DEV) $(LIBYANG)
 $(SONIC_MGMT_COMMON)_RDEPENDS = $(LIBYANG)
 $(SONIC_MGMT_COMMON)_WHEEL_DEPENDS = $(SONIC_YANG_MODELS_PY3)
+ifeq ($(BLDENV),trixie)
```
saiarcot895: Why does `ifeq ($(BLDENV),trixie)` need to be added to these files?

amrutasali: From this submodule [PR](https://github.com/sonic-net/sonic-buildimage/pull/26780) pipeline failure it seems like the CI pipeline build has two separate phases:

  - Package build phase: Builds debian packages (in both bookworm and trixie)
  - Docker build phase: Builds docker images using packages from their respective DEBS_PATH

So the build system builds packages in both bookworm and trixie environments. The sonic-mgmt-common package with go 1.24.4 requirement cannot be built in bookworm (Go 1.19.8), but can be built in trixie (Go 1.24.4).
This PR #26548 migrates docker-sonic-mgmt-framework to use trixie base, which means the framework will use the trixie-built sonic-mgmt-common package instead of the bookworm one.
However, the build still tries to build sonic-mgmt-common in bookworm first and fails.

Hence the conditional.

amrutasali: @saiarcot895 let me know if the understanding of CI pipeline is correct, else i can remove the  the conditional .

hdwhdw:
@yxieca ping for merge. This blocks sonic-mgmt-common submodule update, which in turn blocks sonic-gnmi submodule (already 3 weeks old). @saiarcot895 is this PR good for merge?

rules/sonic-mgmt-common.mk
saiarcot895: The build infra will build a package for a particular version of Debian only if something has a dependency on it. With this change, there shouldn't be anything depending on sonic-mgmt-common in Bookworm. The `ifeq` lines can be removed.

saiarcot895 (approved):

rules/sonic-mgmt-common.mk
amrutasali: Have removed those lines

PR #26561, opened 2026-04-04
Upgrade nat docker container from bookworm to trixie

yxieca:
#### What I did
Upgrade docker-nat base image from bookworm (docker-swss-layer-bookworm) to trixie (docker-swss-layer-trixie).

Part of the Debian Trixie upgrade effort tracked in https://github.com/sonic-net/SONiC/issues/2169.

#### How I did it
- Updated `Dockerfile.j2` base image reference from bookworm to trixie
- Updated `rules/docker-nat.mk` to use trixie layer variables and register as trixie dockers

#### How to verify it
- Build sonic-vs image with this change and verify docker-nat starts correctly
- VS image build tested successfully on local build host


lihuay (dismissed):

rules/docker-nat.mk
```diff
@@ -6,12 +6,16 @@ DOCKER_NAT_DBG = $(DOCKER_NAT_STEM)-$(DBG_IMAGE_MARK).gz
 
 $(DOCKER_NAT)_PATH = $(DOCKERS_PATH)/$(DOCKER_NAT_STEM)
 
-$(DOCKER_NAT)_DEPENDS += $(SWSS) $(IPTABLESIP4TC) $(IPTABLESIP6TC) $(IPTABLESIPTC) $(IPXTABLES12) $(IPTABLES)
-$(DOCKER_NAT)_DBG_DEPENDS = $($(DOCKER_SWSS_LAYER_BOOKWORM)_DBG_DEPENDS)
+$(DOCKER_NAT)_DEPENDS += $(SWSS)
+# On trixie, iptables comes from Debian repos (not built locally)
+ifeq ($(BLDENV),bookworm)
+$(DOCKER_NAT)_DEPENDS += $(IPTABLESIP4TC) $(IPTABLESIP6TC) $(IPTABLESIPTC) $(IPXTABLES12) $(IPTABLES)
+endif
```
saiarcot895: Why is a bookworm section being added?

yxieca: Good catch — removed in 5a1efa35f. Since docker-nat is now trixie-only, the bookworm conditional was dead code. On trixie, iptables comes from Debian repos (installed via apt in the Dockerfile).

saiarcot895 (approved):

PR #26564, opened 2026-04-05
Add static analysis preparation pipeline

xq9mend:
#### Why I did it

Static analysis tools (Semgrep, Bandit, etc.) require a fully prepared source tree — with all externally-fetched sources present and patches applied — before they can run. This pipeline performs that preparation without triggering a full SONiC build.

##### Work item tracking
- Microsoft ADO:

#### How I did it

A Python script (`.azure-pipelines/scripts/prepare_source.py`) handles all source acquisition:
- Runs `dget` to fetch and unpack Debian source packages
- Runs `git clone` to fetch external git repositories (including SHA-pinned and tag-pinned refs)
- Applies any in-tree SONiC patches on top
- Validates that all expected source directories are present

The script is invoked by both an Azure Pipelines YAML and a GitHub Actions workflow that share the same logic. The GitHub Actions workflow also uploads the prepared source tree as an artifact and compares it against a fresh recursive clone to verify completeness.

#### How to verify it

GitHub Actions CI run passes and the artifact diff shows exactly the expected externally-fetched sources with no unintended omissions:

**In artifact but not in fresh clone (externally fetched — 27 sources):**
```
platform/mellanox/iproute2/iproute2-6.1.0
src/bash/bash-5.2.37
src/flashrom/flashrom-0.9.7
src/grub2/grub2
src/ifupdown2/ifupdown2-3.0.0-1
src/iptables/iptables-1.8.9
src/isc-dhcp/isc-dhcp-4.4.3-P1
src/kdump-tools/kdump-tools-1.10.7
src/libnl3/libnl3-3.7.0
src/libteam/libteam-1.31
src/libyang/libyang-1.0.73
src/libyang3-py3/libyang-python
src/libyang3/libyang-3.12.2
src/lldpd/lldpd-1.0.16
src/lm-sensors/lm-sensors-3.6.0
src/monit/monit
src/openssh/openssh-10.0p1
src/radius/pam/freeradius/freeradius-server
src/rdb-cli/librdb
src/sflow/hsflowd/host-sflow
src/sflow/libpsample
src/snmpd/net-snmp-5.9.3+dfsg
src/socat/socat-1.7.4.1
src/sonic-mgmt-common/models/yang/oc-pyang
src/tacacs/audisp/audisp-tacplus
src/tacacs/nss/libnss-tacplus
src/thrift/thrift-0.11.0
```

**In fresh clone but not in artifact (intentionally excluded — 4 entries):**
```
src/p4lang
src/sonic-device-data
src/sonic-gnmi/testdata
src/sonic-linux-kernel
```

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511

- [ ] 

Add static analysis preparation pipeline that fetches all external sources and applies patches without performing a full build, producing an artifact ready for static analysis tools.

N/A


.github/workflows/static-analysis-prep.yml
saiarcot895, line 1: Why is this being duplicated? What's the benefit here? This should ideally be only in the Azure Pipeline or Github Actions CI.

.azure-pipelines/scripts/prepare_source.py
saiarcot895, line 1: This doesn't feel maintainable to me, and assumes specific syntaxes used for cloning repos or checking out a specific hash, in addition to having workarounds for specific repos. What about using the metadata that we already have in our build system? `SONIC_MAKE_DEPS` involves calling `Makefile` to run some set of commands, which implies there's some external repo/files/package that might get downloaded. You can have `make` and `dpkg-buildpackage` commands be a no-op, and have everything else be executed (including preparing the source repo and applying patches.

It's not perfectly clean (SONiC build infra doesn't have a separate step of getting all sources), but it would be more robust.

saiarcot895 (changes requested):
See above comment on `prepare_source.py`.

.github/workflows/static-analysis-prep.yml
xq9mend: The GitHub Actions workflow covers public PRs and is the preferred path for cost reasons. I included the Azure Pipeline version for compatibility with internal workflows (no GitHub Actions)

xq9mend:
The maintainability concern is addressed in the latest commits. Rather than relying on heuristic regex to parse Makefile syntax, the script now uses `make -n` (dry-run) with the package's own Makefile to fully expand Make variables and extract git clone info. This handles all packages correctly:

- Variables defined in the package's own Makefile (e.g. `RASDAEMON_COMMIT`) resolve correctly
- Packages with `pushd <subdir>` before clone (e.g. radius/pam) get the correct dest path
- `git checkout -b branch [-f] commit` patterns handled correctly

The `_find_deb_var_for_pkg()` function uses `_SRC_PATH` assignments to identify the correct deb variable per package, even in shared rules files (e.g. tacacs.mk covering multiple packages). The heuristic regex fallback has been removed — make-n handles all 9 git clone packages (100% coverage).

xq9mend:
Closing in favor of #26892 which has been merged.

PR #26567, opened 2026-04-05
[iccpd]: Add bounds check for num_of_entry in mLACP TLV handlers

securely1g:
### What I did

Add bounds validation for peer-supplied `num_of_entry` / `num_of_vlan_id` fields in iccpd mLACP TLV receive handlers to prevent out-of-bounds heap reads.

### Why I did it

The MAC, ARP, NDISC, and PortChannel TLV handlers (`mlacp_sync_recv_macInfo`, `mlacp_sync_recv_arpInfo`, `mlacp_sync_recv_ndiscInfo`, `mlacp_sync_recv_portChanInfo`) trust the peer-supplied count field without checking it against the actual message length (`msg->len`). A peer that sends a TLV with a large `num_of_entry` but a short body causes the handler to iterate past the heap-allocated `msg->buf`, resulting in:

- **Out-of-bounds heap read** — reading up to ~1.9 MB past the allocation (count up to 65535)
- **iccpd crash** (SIGSEGV) — DoS of the MCLAG control plane, risking split-brain
- **Heap data leakage** — stale heap bytes logged via `ICCPD_LOG_INFO` and potentially programmed into FDB/ARP entries

### How I did it

Added validation in each `mlacp_sync_recv_*()` function to check that `msg->len >= sizeof(ICCHdr) + sizeof(TLV header) + count * sizeof(entry)` before passing the TLV to the update function. Messages that fail validation are dropped with a `ICCPD_LOG_WARN` log message.

### How to verify it

1. With an MCLAG testbed, send a crafted ICCP message with `TLV_T_MLACP_MAC_INFO` containing `num_of_entry=0xFFFF` and an empty MacEntry body
2. Before fix: iccpd crashes with SIGSEGV
3. After fix: iccpd logs a warning and drops the message; daemon remains stable

### Affected handlers

| Handler | Count field | Entry type |
|---|---|---|
| `mlacp_sync_recv_macInfo` | `num_of_entry` | `struct mLACPMACData` |
| `mlacp_sync_recv_arpInfo` | `num_of_entry` | `struct ARPMsg` |
| `mlacp_sync_recv_ndiscInfo` | `num_of_entry` | `struct NDISCMsg` |
| `mlacp_sync_recv_portChanInfo` | `num_of_vlan_id` | `struct mLACPVLANData` |

### DIFF
--- a/src/iccpd/src/mlacp_fsm.c
+++ b/src/iccpd/src/mlacp_fsm.c
@@ -575,8 +575,30 @@ static void mlacp_sync_recv_syncReq(struct CSM* csm, struct Msg* msg)
 static void mlacp_sync_recv_portChanInfo(struct CSM* csm, struct Msg* msg)
 {
     mLACPPortChannelInfoTLV* portconf = NULL;
+    size_t tlv_len;
+    int count;
 
     portconf = (mLACPPortChannelInfoTLV*)&(msg->buf[sizeof(ICCHdr)]);
+
+    /* Validate that msg is large enough to hold the TLV header */
+    if (msg->len < sizeof(ICCHdr) + sizeof(mLACPPortChannelInfoTLV))
+    {
+        ICCPD_LOG_WARN(__FUNCTION__, "Received PortChannel Info TLV too short for header: %zu", msg->len);
+        return;
+    }
+
+    /* Validate num_of_vlan_id against actual message length */
+    count = ntohs(portconf->num_of_vlan_id);
+    tlv_len = sizeof(ICCHdr) + sizeof(mLACPPortChannelInfoTLV)
+              + (size_t)count * sizeof(struct mLACPVLANData);
+    if (tlv_len > msg->len)
+    {
+        ICCPD_LOG_WARN(__FUNCTION__,
+            "PortChannel Info num_of_vlan_id %d exceeds msg len %zu, dropping",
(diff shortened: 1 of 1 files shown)


src/iccpd/src/mlacp_fsm.c
```diff
@@ -575,8 +575,30 @@ static void mlacp_sync_recv_syncReq(struct CSM* csm, struct Msg* msg)
...
     mLACPPortChannelInfoTLV* portconf = NULL;
+    size_t tlv_len;
+    int count;
 
     portconf = (mLACPPortChannelInfoTLV*)&(msg->buf[sizeof(ICCHdr)]);
+
+    /* Validate that msg is large enough to hold the TLV header */
+    if (msg->len < sizeof(ICCHdr) + sizeof(mLACPPortChannelInfoTLV))
+    {
+        ICCPD_LOG_WARN(__FUNCTION__, "Received PortChannel Info TLV too short for header: %zd", msg->len);
```
qiluo-msft: Should be %zu. Appears 8x in the diff

securely1g:
@qiluo-msft Good catch — fixed all 8 occurrences to use `%zu` since `msg->len` is `size_t` (unsigned). Force-pushed.

qiluo-msft (approved):

PR #26570, opened 2026-04-05
[functions.sh] Quote variable expansions in docker_try_rmi

yxieca:
#### Why I did it

`docker_try_rmi()` in `functions.sh` passes `$image_name` to `docker inspect` and `docker rmi` without quotes. Unquoted variable expansion in bash is subject to word splitting (on `$IFS`) and pathname/glob expansion before the resulting words are handed to the command.

This is a defensive shell-correctness fix: the function should pass its argument to `docker` as a single token regardless of what the caller provides. Quoting closes a class of unintended behavior where the shell could interpret part of the value rather than passing it through opaquely.

There is no functional change for well-formed image names.

##### Work item tracking

#### How I did it

Wrapped the two `$image_name` expansions in double quotes:

- `docker inspect --format="{{json .Id}}" "$image_name"`
- `docker rmi "$image_name"`

The local declaration (`local image_name="$1"`) and the `-f "ancestor=$1"` filter were already quoted and are unchanged.

#### How to verify it

This is a **build-time-only** change — `docker_try_rmi()` is invoked from `build_docker.sh` and `get_docker-base.sh` during `make`, on the build host. It is not installed into the image. A clean image build is therefore the complete validation.

1. Full VS image build (`make target/sonic-vs.img.gz`) — completes cleanly; `docker_try_rmi` runs during the docker image cleanup path with quoted args, no behavior change.
2. `bash -n functions.sh` — syntax check passes.
3. Optional: `shellcheck functions.sh` — SC2086 warnings on these two lines are cleared.

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511

Low-risk two-line shell-hardening change. Backport at maintainer discretion — applies cleanly to all listed branches but is not urgent for any specific release.

- [ ] 
- [ ] 

Quote `$image_name` in `docker_try_rmi` to prevent shell word-splitting and glob expansion.

N/A — no YANG or config_db changes.

🦔

### DIFF
--- a/functions.sh
+++ b/functions.sh
@@ -40,13 +40,13 @@ die() {
 docker_try_rmi() {
     local image_name="$1"
     ## Note: inspect output has quotation characters, so sed to remove it as an argument
-    local image_id=$(docker inspect --format="{{json .Id}}" $image_name | sed -e 's/^"//' -e 's/"$//')
+    local image_id=$(docker inspect --format="{{json .Id}}" "$image_name" | sed -e 's/^"//' -e 's/"$//')
     [ -z "$image_id" ] || {
         ## Remove all the exited containers from this image
-        docker ps -a -q -f "status=exited" -f "ancestor=$1" | xargs --no-run-if-empty docker rm
+        docker ps -a -q -f "status=exited" -f "ancestor=$image_name" | xargs --no-run-if-empty docker rm
         ## Note: If there are running containers from this image, the build system is in an
         ##   unexpected state. The 'rmi' will fail and we need investigate the build environment.
-        docker rmi $image_name
+        docker rmi "$image_name"
     }
 }
 


functions.sh
```diff
@@ -40,13 +40,13 @@ die() {
 docker_try_rmi() {
     local image_name="$1"
     ## Note: inspect output has quotation characters, so sed to remove it as an argument
-    local image_id=$(docker inspect --format="{{json .Id}}" $image_name | sed -e 's/^"//' -e 's/"$//')
+    local image_id=$(docker inspect --format="{{json .Id}}" "$image_name" | sed -e 's/^"//' -e 's/"$//')
     [ -z "$image_id" ] || {
         ## Remove all the exited containers from this image
         docker ps -a -q -f "status=exited" -f "ancestor=$1" | xargs --no-run-if-empty docker rm
```
qiluo-msft: nitpick: for consistency the PR could also use $image_name instead of $1 there since the function already aliased it on line 41
```suggestion
        docker ps -a -q -f "status=exited" -f "ancestor=$image_name" | xargs --no-run-if-empty docker rm
```

yxieca: Done — applied, thanks.

lguohan (approved):

qiluo-msft (approved):

PR #26579, opened 2026-04-06
[supervisord-utilities-rs] Replace syslog crate with libc

MavChtz:
The new Rust listener fails if it starts before the syslog daemon is initialized. This occurs because it attempts to connect to the socket without retrying on failure.

The syslog crate doesn't attempt to reconnect when its connection to the socket is lost nor does it have a way to check if the connection was broken.

The libc syslog will behave the same as the Python version as that is what Python calls internally.

##### Work item tracking

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511

- [ ] 
- [ ]


MavChtz:
Closing in favour of https://github.com/sonic-net/sonic-buildimage/pull/26840

Issue #22385, opened 2025-04-20
Enhancement: Upgrade to libyang3


bradh352:
### Is it platform specific

generic

### Importance or Severity

High

### Description of the enhancement

SONiC currently uses 3 versions of libyang: 1.0.73, 1.0.184, and 2.1.148. All are end of life and potentially have security issues. The current release is 3.12.2 and that is what we are targeting here.

There are also numerous libyang related bugs that have been "worked around" in SONiC due to the shortcomings of libyang1. Hopefully upgrading to a modern version will resolve such issues.

### Current Behavior

3 libyang versions

### Proposed Behavior

single libyang version

### PR Tracking Table
| PR | description | dependencies |
|---------|--------|--------|
| #21679 | Remove libyang 1.0.184. build libyang 3.12.2, libyang-python 3.1.0.  Migrate FRR to libyang 3.  Update sonic-yang-models test cases specifically to make it easier to see libyang3-specific changes once those land. **THIS CAN BE MERGED WITHOUT ANY POSSIBLE IMPACT** | None |
| #21719 | Port sonic-yang-models test cases **THIS CAN BE MERGED WITHOUT ANY POSSIBLE IMPACT** | #21679 |
| #21716 | Port sonic-yang-mgmt. | #21679, #21719 |
| #21718 | Update all subprojects that use libyang1 to depend on libyang3 | #21679, #21719, #21716 |
| https://github.com/sonic-net/sonic-utilities/pull/3766 | sonic-utilities | #21679, #21719, #21716, #21718 |
| #21752, https://github.com/sonic-net/sonic-swss-common/pull/973 | swss-common | #21679, #21719, #21716, #21718 |
| #21789, https://github.com/sonic-net/sonic-mgmt-common/pull/157 | sonic-mgmt-common (GOlang) | #21679, #21718 |
| https://github.com/sonic-net/sonic-platform-daemons/pull/589 , https://github.com/sonic-net/sonic-swss/pull/3518 , https://github.com/sonic-net/sonic-dbsyncd/pull/74 , https://github.com/sonic-net/sonic-restapi/pull/157 , https://github.com/sonic-net/sonic-snmpagent/pull/348 , https://github.com/sonic-net/sonic-platform-common/pull/544 | azure pipelines: update various external repos to pull in the libyang3 debs that are otherwise not in need of modification | all of the above |
| #21790 | cleanup / remove libyang1 | all of the above |

prgeor:
[...]

PR #21716, opened 2025-02-12
libyang: upgrade to v3 step 2 -- port sonic-yang-mgmt

bradh352:
#### Why I did it

libyang3 porting.

Port of sonic-yang-mgmt

Depends PR #21679

##### Work item tracking
- Microsoft ADO **(number only)**:

#### How to verify it

All PRs listed on #21679 must be merged into a local build tree in order to test due to cross-repository dependencies.

#### Which release branch to backport (provide reason below if selected)

None

#### Tested branch (Please provide the tested image version)

master as of 20250212

#### Description for the changelog

libyang: upgrade to v3 step 2 -- port sonic-yang-mgmt

#### Link to config_db schema for YANG module changes

N/A

#### A picture of a cute animal (not mandatory but encouraged)
Signed-off-by: Brad House (@bradh352)

PR #26584, opened 2026-04-06
[yang-mgmt] Port sonic-yang-mgmt from libyang1 to libyang3 Python bindings

bhouse-nexthop:
#### Why I did it

Port `sonic-yang-mgmt` from the libyang1 Python bindings (`import yang`) to the libyang3 Python bindings (`import libyang`) as part of the SONiC-wide libyang3 upgrade tracked in #22385.

Replaces the earlier PR #21716, which was written against an older codebase. The current branch has diverged significantly (the `sonic_yang_path.py` mixin, an enhanced `find_data_dependencies` with DFS + value-matching, schema-dependency caching, etc.), so a fresh port was needed.

A second pass on this branch (commit `045f74b8e`) goes further than a straight 1:1 port: it removes the manual `uses`/`grouping` inlining the old code did over the YIN/`xmltodict` representation and walks libyang3's compiled schema tree directly. libyang2/3 introduced a parsed-vs-compiled split in the C library — `lysp_*` preserves `uses`/`grouping`/`augment`/`deviation` as discrete nodes, while `lysc_*` has them already resolved. The libyang3-py SNode iteration (`Module.children`, `SContainer.children`, `lys_getnext`) walks the compiled tree, so consuming SNodes directly gives us the post-resolution view "for free" and the manual inlining (`_compileUsesClauseModel` / `_preProcessYangGrouping` / `_findYangModuleFromPrefix` / `_loadJsonYangModel` / `print_mem("yin")` + `xmltodict.parse`) can be deleted entirely.

**Dependencies:**
- Depends on sonic-net/sonic-buildimage#26563 merging first (yang-mgmt prerequisite fixes: `find_data_dependencies` rewrite, uses compilation, test isolation)
- Depends on sonic-net/sonic-utilities#4118 merging first (removes direct libyang dependency from sonic-utilities)

##### Work item tracking

#### How I did it

The branch is structured as six commits with logical separation:

1. **`sonic-yang-mgmt: Depend on libyang3 instead of libyang1`** — switches build/runtime dependencies from `LIBYANG`/`LIBYANG_CPP`/`LIBYANG_PY3` to `LIBYANG3`/`LIBYANG3_PY3` across Dockerfiles and `rules/*.mk`. `libyang_cpp` is dropped — `libyang3-py3` binds directly to the C library.
2. **`sonic-yang-mgmt: Port code to libyang3 Python bindings`** — the core API migration in `sonic_yang.py` and `sonic_yang_ext.py`. See API mapping table below.
3. **`sonic-yang-mgmt: Update xpath_join() for libyang3 prefix convention`** — libyang3 uses RFC 7951 JSON-style prefixes for schema xpaths; only the first token gets a module prefix, child tokens inherit from parent.
4. **`sonic-yang-mgmt: Update test cases for libyang3`** — test data updates (see "How to verify" below). No production-code change.
5. **`sonic-yang-mgmt: Fix pyright type errors across all modules`** — resolves all pyright errors (82 in `sonic_yang_ext.py`, 17 in `sonic_yang.py`, 16 in `sonic_yang_path.py` → 0 in all three) via `TYPE_CHECKING` mixin stubs, `Optional` annotations, and `None` guards.
6. **`sonic-yang-mgmt: Walk libyang3 compiled schema directly; drop manual uses/grouping inlining`** — the second-pass refactor. Replaces every `model.get('list')` / `model['key']['@value']` / `leaf['type']['@name']` access with the corresponding SNode call, deletes the `xmltodict`-based pathway, and drops the `xmltodict` dependency. Net ~600 lines removed.

##### Core API migration (commit 2)

| Operation | libyang1 | libyang3 |
|---|---|---|
| Context creation | `ly.Context(yang_dir, options)` | `ly.Context(yang_dir)` |
| Module loading | `ctx.parse_module_path(f, ly.LYS_IN_YANG)` | `ctx.parse_module_file(fh, "yang")` |
| Data parsing | `ctx.parse_data_path(f, ly.LYD_JSON, flags)` | `ctx.parse_data_file(fh, "json", no_state=True, strict=True, json_string_datatypes=True)` |
| Data node create | `root.new_path(ctx, xpath, val, 0, 0)` | `ctx.create_data_path(xpath, parent=root, value=val)` |
| Data node search | `root.find_path(xpath).data()` | `root.find_all(xpath)` |
| Validation | `node.validate(ly.LYD_OPT_CONFIG, ctx)` | `node.validate(no_state=True)` |
| Serialization | format constants (`ly.LYD_JSON`) | string names (`"json"`, `"yang"`, `"yin"`) |
| Value access | `data_node.subtype().value_str()` | `data_node.value()` (with `isinstance` `DLeaf`/`DLeafList` guard) |
| Schema path | `schema_node.path()` | `schema_node.schema_path()` |
| Type info | `subtype().type().base()` (int constant) | `type().basename()` (string: `"string"`, `"uint16"`, `"leafref"`, …) |
| Leafref path | `subtype().type().info().lref().path()` | `schema_node.type().leafref_path()` |
| Leafref type tag | `ly.LY_TYPE_LEAFREF` | `ly.Type.LEAFREF` |
| Backlinks | recursive per-node lookup | `ctx.find_backlinks_paths()` + `ctx.find_leafref_path_target_paths()`, cached in `backlinkMap` |

Also added `__del__()` for explicit `root.free()` / `ctx.destroy()` cleanup; removed `_load_schema_modules_ctx()`, `_str_to_type()`, and the `sonic_yang_options` constructor parameter.

##### Compiled-schema refactor (commit 6)

The original code called `module.print_mem("yin")` on each loaded module, parsed the resulting XML with `xmltodict`, and stored the dict tree as `self.yJson`. Because libyang3's YIN printer emits the *parsed* tree (`printer_schema.c:56-64` → `yin_print_parsed_module`), `<uses>` and `<grouping>` survive into the dict tree — so `sonic_yang_ext` re-implemented the inlining itself: `_preProcessYang` + `_compileUsesClauseModel` + `_compileUsesClauseList` + `_compileUsesClauseRefine` + `_findYangModuleFromPrefix`.

Walking the compiled SNode tree directly lets us delete that whole pathway:

- `self.confDbYangMap[<table>]` now holds an `SContainer` instead of an `xmltodict` dict, plus the module name and top-level container name as plain strings.
- `_createDBTableToModuleMap` walks `module.children(types=(SNode.CONTAINER,))` for each loaded module, registers each child container as a ConfigDB table.
- Every `_xlate*` / `_revXlate*` function takes an SNode and dispatches via `snode.children(types=...)`. Dict-vs-list `isinstance` branching is gone — iterators are always iterators.
- `_createLeafDict` shrinks from a 30-line walker that explicitly unwrapped choice/case to a 4-line generator: `iter_children` with default options already descends into `choice`/`case` and yields the leaves flat, so `model_snode.children(types=(LEAF, LEAFLIST))` covers leaves under choice/case as well as direct leaves.
- The path module (`sonic_yang_path.py`) gets the same conversion: every helper that walked the YIN dict (`__get_xpath_tokens_from_*`, `__get_configdb_path_tokens_from_*`, `__get_list_model`, `__type1_get_model`, `__type1_get_xpath_token`, etc.) now walks SNode children with type filtering.
- The `sonic-cfg-help` CLI is rewritten the same way: `snode.description()`, `snode.mandatory()`, `SLeaf.default()`, `type().basename()`, `type().leafref_path()`, `type().union_types()`, `slist.keys()`, `case.when_conditions()`. `print_documentation` drives off `confDbYangMap` directly instead of re-iterating `yJson`.

Removed: `self.yJson`, `self.preProcessedYang`, `_loadJsonYangModel`, `_preProcessYang`, `_preProcessYangGrouping`, `_compileUsesClause`, `_compileUsesClauseModel`, `_compileUsesClauseList`, `_compileUsesClauseRefine`, `_compileUsesClauseRefineApply`, `_findYangModuleFromPrefix`, the `from xmltodict import parse` import, and the `xmltodict==0.12.0` entry from `setup.py` `install_requires` and `tests_require`.

##### Test changes

Two white-box test classes are deleted because they exercised structures that no longer exist:

- `Test_SonicYang_UsesCompilation` (8 tests) asserted on `self.yJson` shape and `self.preProcessedYang['grouping']` content — both are internal state of the deleted machinery. Replaced by a single black-box test, `Test_SonicYang_UsesResolution::test_grouping_supplied_nodes_in_compiled_tree`, that loads the existing `test-grouping.yang` fixture and asks the libyang context to resolve `TEST_GROUPING_LIST` via `ctx.find_path()`. The SNode children must include `settings` (from `group-with-container`), `member` (from `group-with-list`), `description` (from nested-uses chain `nested-uses-group` → `simple-fields`), and `extra` (native leaf in the nested grouping) — proving the same coverage at the user-visible schema level.
- `TestCfgHelpUnionType` (3 tests) constructed `xmltodict`-shaped dicts (`{'@name': ..., 'type': {'@name': 'union', 'type': ...}}`) and called `parse_leaf` directly to verify the dict-vs-list ambiguity that `xmltodict` produced for unions with a single member type — that whole class of bug doesn't exist under the SNode API.

The remaining `TestCfgHelp` class (7 black-box tests that subprocess `sonic-cfg-help` and assert on stdout) is **untouched** and continues to validate the user-facing CLI for table, field, leaf-list, type-1 map, when-condition, and nested-container patterns.

Real SONiC YANG models use `grouping`/`uses` extensively (sonic-types typedefs, sonic-bgp-common, qos/buffer table groupings), so `test_validate_yang_models`, `test_xlate_rev_xlate`, `test_find_data_dependencies`, `test_find_schema_dependencies`, `test_get_data_type`, `test_get_leafref_*`, and `test_find_data_node_schema_xpath` continue to exercise the resolution path against the real model set.

#### How to verify it

```
make target/python-wheels/bookworm/sonic_yang_mgmt-1.0-py3-none-any.whl
grep passed target/python-wheels/bookworm/sonic_yang_mgmt-1.0-py3-none-any.whl.log
```

→ `42 passed in 1.56s` (build elapsed: ~12 s).

Pyright is clean on all six changed files — only the host-only `Import "libyang" could not be resolved` remains, which is expected (libyang3-py3 is installed inside the build container, not on the build host).

Test data updates (commit 4) required for libyang3 semantics:

- `test_SonicYang.json`: `data_type` strings (`"LY_TYPE_STRING"` → `"string"`, `"LY_TYPE_LEAFREF"` → `"leafref"`, …), schema xpaths updated to v3 prefix convention (only first token prefixed), `leafref_path` values use YANG prefix instead of module name, `delete_nodes.valid` field changed from string to boolean.
- `config_data.json`: pattern-validated leaf values changed (`true`/`false` → `"up"`/`"down"`) since libyang3 enforces string patterns strictly.
- `test-yang-structure.yang`: pattern updated to `"down|up"` to match config data.
- `test_sonic_yang.py`: removed `_str_to_type()` calls, updated `_merge_data()` signature, added `_deleteNode` return-value assertion.

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511

- [x] master (as of 2026-04-28)

[yang-mgmt] Port sonic-yang-mgmt from libyang1 to libyang3 Python bindings; walk compiled schema tree to drop manual `uses`/`grouping` inlining and the `xmltodict` dependency.

N/A — no YANG model schema changes, only the management library that processes them.


src/sonic-yang-mgmt/sonic_yang_ext.py
```diff
@@ -375,141 +199,71 @@ def _extractKey(self, tableKey, keys):
...
-        if leafDict[key]['__isleafList']:
+        if is_leaf_list:
             vValue = list()
             if isinstance(value, str) and (self.elementPath[0], self.elementPath[-1]) in LEAF_LIST_WITH_STRING_VALUE_DICT:
                 # For field defined as leaf-list but has string value in CONFIG DB, need do special handling here. For exampe:
                 # port.adv_speeds in CONFIG DB has value "100,1000,10000", it shall be transferred to [100,1000,10000] as YANG value here to
-                # make it align with its YANG definition.
-                value = (x.strip() for x in value.split(LEAF_LIST_WITH_STRING_VALUE_DICT[(self.elementPath[0], self.elementPath[-1])]))
+                # be compliant with yang model.
+                value = value.split(LEAF_LIST_WITH_STRING_VALUE_DICT[(self.elementPath[0], self.elementPath[-1])])
```
rimunagala: Nit (non-blocking): The .strip() on each split element was dropped here. A few entries in LEAF_LIST_WITH_STRING_VALUE_DICT are string-typed (e.g. MIRROR_SESSION.src_ip, BGP_ALLOWED_PREFIXES.prefixes_v4/v6, NTP.src_intf), so a value like "10.0.0.1, 10.0.0.2" would now keep the leading space and could fail YANG pattern validation. Risk is very low since ConfigDB doesn't store spaces around delimiters, but
consider keeping the .strip() as a defensive measure:

 value = [x.strip() for x in value.split(LEAF_LIST_WITH_STRING_VALUE_DICT[(self.elementPath[0], self.elementPath[-1])])]

rimunagala (dismissed):

src/sonic-yang-mgmt/sonic_yang_ext.py
bhouse-nexthop: Accepted — restored `.strip()` via list comprehension in 9893b4520. Thanks for catching the regression from the libyang1 code path.

rimunagala (approved):

rookie-who (approved):

rookie-who (approved):

vaibhavhd (approved):

yutongzhang-microsoft:
Hi, @bhouse-nexthop , this PR will cause the issue https://dev.azure.com/mssonic/build/_build/results?buildId=1119863&view=logs&j=5562265b-363f-5c64-5cff-7b508c285721&t=2a766db5-c3a4-56e8-abe6-434d5e07f32d, which block the CI, can you please check?

bradh352:
> Hi, @bhouse-nexthop , this PR will cause the issue https://dev.azure.com/mssonic/build/_build/results?buildId=1119863&view=logs&j=5562265b-363f-5c64-5cff-7b508c285721&t=2a766db5-c3a4-56e8-abe6-434d5e07f32d, which block the CI, can you please check?

This PR is the proper fix:
https://github.com/sonic-net/sonic-swss-common/pull/973

PR #26617, opened 2026-04-07
[bgpcfgd]: Validate aggregate address prefix and reject prefixes with host bits set

Guoyuan001:
#### Why I did it

When an invalid aggregate address prefix (e.g., 10.100.0.1/24 with host bits set) is configured in CONFIG_DB, bgpcfgd silently accepts it and pushes invalid commands to FRR. This can cause unexpected BGP behavior. Adding prefix validation with a clear error log makes it easier to diagnose misconfiguration.

##### Work item tracking
- Microsoft ADO **(number only)**: 37506197

#### How I did it

Added a validate_prefix() function in managers_aggregate_address.py that uses Python's ipaddress.ip_network(prefix, strict=True) to verify the prefix is a valid network address (no host bits set).
Called validate_prefix() at the beginning of address_set_handler(). If validation fails, the handler logs an error (log_err) with the prefix and reason, and returns False to skip the invalid entry.
Added unit tests for validate_prefix() covering valid prefixes (e.g., 10.100.0.0/16, 2001:db8::/32, 0.0.0.0/0) and invalid prefixes with host bits set.
Added integration-style test test_host_bits_set_rejected verifying that address_set_handler rejects bad prefixes, does not push commands, and marks state as inactive.

#### How to verify it
All tests including test_validate_prefix and test_host_bits_set_rejected should pass.

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511
- [x] 202603

- [x]  SONiC.20251110.19
- [ ] 

Add prefix validation in bgpcfgd aggregate address manager to reject prefixes with host bits set and log an error.

### DIFF
--- a/src/sonic-bgpcfgd/bgpcfgd/managers_aggregate_address.py
+++ b/src/sonic-bgpcfgd/bgpcfgd/managers_aggregate_address.py
@@ -127,11 +127,11 @@ class AggregateAddressMgr(Manager):
 
     def del_handler(self, key):
         address_state = self.get_address_from_state_db(key)
-        if address_state.get(ADDRESS_STATE_KEY) == ADDRESS_ACTIVE_STATE:
+        if address_state.get(ADDRESS_STATE_KEY) == ADDRESS_INACTIVE_STATE:
+            log_info("AggregateAddressMgr::address %s is inactive, skip FRR removal" % key2prefix(key))
+        else:
             if self.address_del_handler(key, address_state):
                 log_info("AggregateAddressMgr::delete address %s success" % key)
-        else:
-            log_info("AggregateAddressMgr::address %s is not active, skip FRR removal" % key2prefix(key))
         self.del_address_state(key)
         return True
 
--- a/src/sonic-bgpcfgd/bgpcfgd/managers_aggregate_address.py
+++ b/src/sonic-bgpcfgd/bgpcfgd/managers_aggregate_address.py
@@ -1,3 +1,5 @@
+import ipaddress
+
 from swsscommon import swsscommon
 
 from .log import log_info, log_err
@@ -77,6 +79,12 @@ class AggregateAddressMgr(Manager):
     def address_set_handler(self, key, data):
         bgp_asn = self.directory.get_slot(CONFIG_DB_NAME, swsscommon.CFG_DEVICE_METADATA_TABLE_NAME)["localhost"]["bgp_asn"]
         prefix = key2prefix(key)
+
+        valid, reason = validate_prefix(prefix)
+        if not valid:
+            log_err("AggregateAddressMgr::invalid aggregate prefix %s: %s" % (prefix, reason))
+            return False
+
         is_v4 = '.' in prefix
         cmd_list = []
 
@@ -198,6 +206,15 @@ class AggregateAddressMgr(Manager):
 def key2prefix(key):
     prefix = key.split("|")[-1]
     return prefix
+
+
+def validate_prefix(prefix):
+    """Return (True, None) if prefix is a valid network address, or (False, reason) otherwise."""
+    try:
+        ipaddress.ip_network(prefix, strict=True)
+    except ValueError as e:
+        return False, str(e)
+    return True, None
 
 
 def generate_aggregate_address_commands(asn, prefix, is_v4, is_remove, summary_only=COMMON_FALSE_STRING, as_set=COMMON_FALSE_STRING):
--- a/src/sonic-bgpcfgd/bgpcfgd/managers_aggregate_address.py
+++ b/src/sonic-bgpcfgd/bgpcfgd/managers_aggregate_address.py
@@ -88,7 +88,7 @@ class AggregateAddressMgr(Manager):
             log_err("AggregateAddressMgr::invalid aggregate prefix %s: %s" % (prefix, reason))
             return False
 
-        is_v4 = '.' in prefix
+        is_v4 = ipaddress.ip_network(prefix, strict=True).version == 4
         cmd_list = []
 
         aggregates_cmds = generate_aggregate_address_commands(
@@ -138,7 +138,7 @@ class AggregateAddressMgr(Manager):
     def address_del_handler(self, key, data):
         bgp_asn = self.directory.get_slot(CONFIG_DB_NAME, swsscommon.CFG_DEVICE_METADATA_TABLE_NAME)["localhost"]["bgp_asn"]
         prefix = key2prefix(key)
-        is_v4 = '.' in prefix
+        is_v4 = ipaddress.ip_network(prefix, strict=False).version == 4
         cmd_list = []
 
         aggregates_cmds = generate_aggregate_address_commands(
--- a/src/sonic-bgpcfgd/bgpcfgd/managers_aggregate_address.py
+++ b/src/sonic-bgpcfgd/bgpcfgd/managers_aggregate_address.py
@@ -52,7 +52,7 @@ class AggregateAddressMgr(Manager):
                 if self.address_set_handler(address[0], address[1]):
                     self.set_address_state(address[0], address[1], ADDRESS_ACTIVE_STATE)
                 else:
-                    log_err("AggregateAddressMgr::set address %s failed during BBR change" % key2prefix(address[0]))
+                    log_err("AggregateAddressMgr::set address %s failed during BBR change (validation or FRR push error)" % key2prefix(address[0]))
                     self.set_address_state(address[0], address[1], ADDRESS_INACTIVE_STATE)
         elif bbr_status == BGP_BBR_STATUS_DISABLED:
             log_info("AggregateAddressMgr::BBR state changed to %s with bbr_required addresses %s" % (bbr_status, addresses))
@@ -75,7 +75,7 @@ class AggregateAddressMgr(Manager):
             if self.address_set_handler(key, data):
                 self.set_address_state(key, data, ADDRESS_ACTIVE_STATE)
             else:
-                log_err("AggregateAddressMgr::set address %s failed" % key2prefix(key))
+                log_err("AggregateAddressMgr::set address %s failed (validation or FRR push error)" % key2prefix(key))
                 self.set_address_state(key, data, ADDRESS_INACTIVE_STATE)
         return True
 
@@ -83,12 +83,12 @@ class AggregateAddressMgr(Manager):
         bgp_asn = self.directory.get_slot(CONFIG_DB_NAME, swsscommon.CFG_DEVICE_METADATA_TABLE_NAME)["localhost"]["bgp_asn"]
         prefix = key2prefix(key)
 
-        valid, reason = validate_prefix(prefix)
-        if not valid:
+        net, reason = validate_prefix(prefix)
+        if net is None:
             log_err("AggregateAddressMgr::invalid aggregate prefix %s: %s" % (prefix, reason))
             return False
 
-        is_v4 = ipaddress.ip_network(prefix, strict=True).version == 4
+        is_v4 = net.version == 4
         cmd_list = []
 
         aggregates_cmds = generate_aggregate_address_commands(
@@ -138,7 +138,8 @@ class AggregateAddressMgr(Manager):
     def address_del_handler(self, key, data):
         bgp_asn = self.directory.get_slot(CONFIG_DB_NAME, swsscommon.CFG_DEVICE_METADATA_TABLE_NAME)["localhost"]["bgp_asn"]
         prefix = key2prefix(key)
-        is_v4 = ipaddress.ip_network(prefix, strict=False).version == 4
+        net, _ = validate_prefix(prefix)
+        is_v4 = net.version == 4 if net else False
         cmd_list = []
 
         aggregates_cmds = generate_aggregate_address_commands(
@@ -215,14 +216,14 @@ def key2prefix(key):
 
 
 def validate_prefix(prefix):
-    """Return (True, None) if prefix is a valid network address, or (False, reason) otherwise."""
+    """Return (network, None) if prefix is valid, or (None, reason) otherwise."""
     if '/' not in prefix:
-        return False, "missing prefix length"
+        return None, "missing prefix length"
     try:
-        ipaddress.ip_network(prefix, strict=True)
+        net = ipaddress.ip_network(prefix, strict=True)
     except ValueError as e:
-        return False, str(e)
-    return True, None
+        return None, str(e)
+    return net, None
 
 
 def generate_aggregate_address_commands(asn, prefix, is_v4, is_remove, summary_only=COMMON_FALSE_STRING, as_set=COMMON_FALSE_STRING):
--- a/src/sonic-bgpcfgd/bgpcfgd/managers_aggregate_address.py
+++ b/src/sonic-bgpcfgd/bgpcfgd/managers_aggregate_address.py
@@ -52,7 +52,7 @@ class AggregateAddressMgr(Manager):
                 if self.address_set_handler(address[0], address[1]):
                     self.set_address_state(address[0], address[1], ADDRESS_ACTIVE_STATE)
                 else:
-                    log_err("AggregateAddressMgr::set address %s failed during BBR change (validation or FRR push error)" % key2prefix(address[0]))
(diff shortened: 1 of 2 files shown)


src/sonic-bgpcfgd/bgpcfgd/managers_aggregate_address.py
```diff
@@ -200,6 +208,15 @@ def key2prefix(key):
     return prefix
 
 
+def validate_prefix(prefix):
+    """Return (True, None) if prefix is a valid network address, or (False, reason) otherwise."""
+    try:
+        ipaddress.ip_network(prefix, strict=True)
```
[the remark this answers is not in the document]
Guoyuan001: strict=True is expected

src/sonic-bgpcfgd/bgpcfgd/managers_aggregate_address.py
```diff
@@ -77,6 +79,12 @@ def set_handler(self, key, data):
     def address_set_handler(self, key, data):
         bgp_asn = self.directory.get_slot(CONFIG_DB_NAME, swsscommon.CFG_DEVICE_METADATA_TABLE_NAME)["localhost"]["bgp_asn"]
         prefix = key2prefix(key)
+
+        valid, reason = validate_prefix(prefix)
+        if not valid:
+            log_err("AggregateAddressMgr::invalid aggregate prefix %s: %s" % (prefix, reason))
```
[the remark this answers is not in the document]
Guoyuan001: fixed in line 56

src/sonic-bgpcfgd/bgpcfgd/managers_aggregate_address.py
```diff
@@ -77,6 +82,12 @@ def set_handler(self, key, data):
...
+            return False
+
```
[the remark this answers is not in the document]
Guoyuan001: fix inactive delete senario

src/sonic-bgpcfgd/bgpcfgd/managers_aggregate_address.py
```diff
@@ -200,6 +211,15 @@ def key2prefix(key):
     return prefix
 
 
+def validate_prefix(prefix):
+    """Return (True, None) if prefix is a valid network address, or (False, reason) otherwise."""
+    try:
+        ipaddress.ip_network(prefix, strict=True)
+    except ValueError as e:
+        return False, str(e)
+    return True, None
```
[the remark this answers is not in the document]
Guoyuan001: fixed

src/sonic-bgpcfgd/bgpcfgd/managers_aggregate_address.py
```diff
@@ -116,9 +127,12 @@ def address_set_handler(self, key, data):
...
     def del_handler(self, key):
         address_state = self.get_address_from_state_db(key)
-        if self.address_del_handler(key, address_state):
-            log_info("AggregateAddressMgr::delete address %s success" % key)
-            self.del_address_state(key)
+        if address_state.get(ADDRESS_STATE_KEY) == ADDRESS_ACTIVE_STATE:
+            if self.address_del_handler(key, address_state):
+                log_info("AggregateAddressMgr::delete address %s success" % key)
+        else:
+            log_info("AggregateAddressMgr::address %s is not active, skip FRR removal" % key2prefix(key))
```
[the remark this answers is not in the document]
Guoyuan001: fixed

src/sonic-bgpcfgd/bgpcfgd/managers_aggregate_address.py
```diff
@@ -77,6 +82,12 @@ def set_handler(self, key, data):
     def address_set_handler(self, key, data):
         bgp_asn = self.directory.get_slot(CONFIG_DB_NAME, swsscommon.CFG_DEVICE_METADATA_TABLE_NAME)["localhost"]["bgp_asn"]
         prefix = key2prefix(key)
+
+        valid, reason = validate_prefix(prefix)
+        if not valid:
+            log_err("AggregateAddressMgr::invalid aggregate prefix %s: %s" % (prefix, reason))
+            return False
+
         is_v4 = '.' in prefix
```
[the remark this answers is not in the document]
Guoyuan001: fixed

shixizhang (commented):
Thanks for adding this validation -- catching host-bits-set prefixes before they reach FRR is a good defensive measure.

src/sonic-bgpcfgd/bgpcfgd/managers_aggregate_address.py
```diff
@@ -70,14 +75,20 @@ def set_handler(self, key, data):
...
         bgp_asn = self.directory.get_slot(CONFIG_DB_NAME, swsscommon.CFG_DEVICE_METADATA_TABLE_NAME)["localhost"]["bgp_asn"]
         prefix = key2prefix(key)
-        is_v4 = '.' in prefix
+
+        valid, reason = validate_prefix(prefix)
+        if not valid:
+            log_err("AggregateAddressMgr::invalid aggregate prefix %s: %s" % (prefix, reason))
+            return False
+
+        is_v4 = ipaddress.ip_network(prefix, strict=True).version == 4
```
shixizhang: **Suggestion: avoid double-parsing the prefix**

`validate_prefix()` already calls `ipaddress.ip_network(prefix, strict=True)` a few lines above. This line parses the exact same string again. Consider returning the parsed network object from `validate_prefix()` so you can reuse it here:

```python
def validate_prefix(prefix):
    if '/' not in prefix:
        return None, "missing prefix length"
    try:
        net = ipaddress.ip_network(prefix, strict=True)
    except ValueError as e:
        return None, str(e)
    return net, None
```

Then here:
```python
net, reason = validate_prefix(prefix)
if net is None:
    log_err(...)
    return False
is_v4 = net.version == 4
```

shixizhang (commented):
Thanks for adding this validation -- catching host-bits-set prefixes before they reach FRR is a good improvement. Left a few inline suggestions. Overall looks solid!

src/sonic-bgpcfgd/bgpcfgd/managers_aggregate_address.py
```diff
@@ -116,15 +127,18 @@ def address_set_handler(self, key, data):
...
+            if self.address_del_handler(key, address_state):
+                log_info("AggregateAddressMgr::delete address %s success" % key)
+        self.del_address_state(key)
         return True
 
     def address_del_handler(self, key, data):
         bgp_asn = self.directory.get_slot(CONFIG_DB_NAME, swsscommon.CFG_DEVICE_METADATA_TABLE_NAME)["localhost"]["bgp_asn"]
         prefix = key2prefix(key)
-        is_v4 = '.' in prefix
+        is_v4 = ipaddress.ip_network(prefix, strict=False).version == 4
```
shixizhang: **Nit: add a comment explaining why `strict=False` here**

In `address_set_handler` you use `strict=True` (via `validate_prefix`), but here in `address_del_handler` you intentionally use `strict=False`. This is correct -- the delete path needs to handle removing previously-accepted bad prefixes. But without a comment, a future reader might think this is a bug and "fix" it to `strict=True` to match the set path.

```python
# strict=False: delete path must handle prefixes that were accepted before validation was added
is_v4 = ipaddress.ip_network(prefix, strict=False).version == 4
```

src/sonic-bgpcfgd/bgpcfgd/managers_aggregate_address.py
```diff
@@ -49,6 +51,9 @@ def on_bbr_change(self):
             for address in addresses:
                 if self.address_set_handler(address[0], address[1]):
                     self.set_address_state(address[0], address[1], ADDRESS_ACTIVE_STATE)
+                else:
+                    log_err("AggregateAddressMgr::set address %s failed during BBR change" % key2prefix(address[0]))
```
shixizhang: **Consider logging more context here**

Since `on_bbr_change` re-activates previously configured addresses, a prefix validation failure here would mean the prefix was somehow accepted earlier but is now rejected -- a surprising condition. The current message ("set address ... failed during BBR change") doesn't distinguish prefix validation failure from FRR push failure. Adding the reason would make debugging easier:

```python
log_err("AggregateAddressMgr::set address %s failed during BBR change" % key2prefix(address[0]))
```

versus something like:
```python
log_err("AggregateAddressMgr::address %s failed during BBR change (validation or FRR push error)" % key2prefix(address[0]))
```

Or better yet, have `address_set_handler` return a reason string on failure.

src/sonic-bgpcfgd/bgpcfgd/managers_aggregate_address.py
```diff
@@ -116,15 +127,18 @@ def address_set_handler(self, key, data):
 
     def del_handler(self, key):
         address_state = self.get_address_from_state_db(key)
-        if self.address_del_handler(key, address_state):
-            log_info("AggregateAddressMgr::delete address %s success" % key)
-            self.del_address_state(key)
+        if address_state.get(ADDRESS_STATE_KEY) == ADDRESS_INACTIVE_STATE:
```
shixizhang, line 140: **Suggestion: add a test for this new inactive-skip path**

This is a meaningful behavioral change -- inactive entries now skip FRR removal entirely. But there's no dedicated test covering this path. Consider adding one that:
1. Sets up an aggregate with an invalid prefix (state = inactive)
2. Calls `del_handler` on that prefix
3. Asserts `cfg_mgr.push_list` was NOT called
4. Asserts the STATE_DB entry is cleaned up

This would also serve as a regression test if the del_handler logic changes later.

src/sonic-bgpcfgd/bgpcfgd/managers_aggregate_address.py
```diff
@@ -200,6 +214,17 @@ def key2prefix(key):
     return prefix
 
 
+def validate_prefix(prefix):
```
shixizhang, line 228: **Nice clean validation function!** One minor thought: if you return the parsed `ip_network` object instead of `True` (as suggested on the `is_v4` comment above), callers get both validation and the parsed result in one call, avoiding the redundant parse in `address_set_handler`.

Guoyuan001: fixed

src/sonic-bgpcfgd/bgpcfgd/managers_aggregate_address.py
Guoyuan001: fixed

src/sonic-bgpcfgd/bgpcfgd/managers_aggregate_address.py
Guoyuan001: fixed

src/sonic-bgpcfgd/bgpcfgd/managers_aggregate_address.py
Guoyuan001: fixed

src/sonic-bgpcfgd/bgpcfgd/managers_aggregate_address.py
Guoyuan001: fixed

shixizhang (dismissed):
All previous comments have been addressed in the latest revision. Specifically:

1. **Redundant parsing** -- `validate_prefix()` now returns `(net, reason)` and callers reuse the parsed network object. No double-parsing. 
2. **strict=True/False consistency** -- `address_del_handler` now also uses `validate_prefix()` with a safe fallback (`net.version == 4 if net else False`). Clean.
3. **on_bbr_change error log** -- Message now includes "(validation or FRR push error)" for better diagnostics. 
4. **del_handler inactive-skip test** -- New `test_inactive_entry_skips_frr_removal` test added, covering the exact scenario. 
5. **validate_prefix return type** -- Now returns the parsed `ip_network` object, consistent with how callers use it. 
6. **del_handler existing test** -- `__del_handler_validate` updated to check active state before expecting FRR commands. 

One minor note for follow-up (non-blocking): the 10 incremental commits ("fix bug", "fix ut case data", etc.) could be squashed before merge for a cleaner git history.

LGTM!

src/sonic-bgpcfgd/bgpcfgd/managers_aggregate_address.py
```diff
@@ -49,6 +51,9 @@ def on_bbr_change(self):
             for address in addresses:
                 if self.address_set_handler(address[0], address[1]):
                     self.set_address_state(address[0], address[1], ADDRESS_ACTIVE_STATE)
+                else:
```
shixizhang, line 54: Minor log-level inconsistency: when `address_set_handler` rejects an invalid prefix here (called from `on_bbr_change`), the failure is logged at `log_info`. But the same invalid prefix going through `set_handler` would be caught earlier and logged at `log_err`. Consider using `log_err` here too for consistency, or at least mention in the message that this path is the BBR-change retry — helps operators searching logs for all validation failures with a single severity filter.

Non-blocking, just a suggestion for consistency.

src/sonic-bgpcfgd/bgpcfgd/managers_aggregate_address.py
```diff
@@ -59,25 +64,41 @@ def on_bbr_change(self):
...
-        if bbr_status not in (BGP_BBR_STATUS_ENABLED, BGP_BBR_STATUS_DISABLED):
-            log_info("AggregateAddressMgr::BBR state is unknown. Skip the address %s" % key2prefix(key))
+        prefix = key2prefix(key)
+        net, reason = validate_prefix(prefix)
+        if net is None:
+            log_err("AggregateAddressMgr::invalid aggregate prefix %s: %s" % (prefix, reason))
+            self.set_address_state(key, data, ADDRESS_INACTIVE_STATE)
+            return True
+        if self.directory.path_exist(CONFIG_DB_NAME, BGP_BBR_TABLE_NAME, BGP_BBR_STATUS_KEY):
+            bbr_status = self.directory.get(CONFIG_DB_NAME, BGP_BBR_TABLE_NAME, BGP_BBR_STATUS_KEY)
```
shixizhang, line 74: Good improvement — `path_exist()` guard avoids crashing when the BBR table doesn't exist yet. The refined logic (only skip on unknown BBR when `bbr_required=true`) also makes sense: a non-BBR aggregate address shouldn't be blocked by missing BBR config.

One thought: now that `set_handler` validates prefix first and returns `True` on invalid prefix — meaning the manager considers the entry "handled" even though it's invalid — just want to confirm this won't cause the framework to skip retries when CONFIG_DB gets corrected later. If CONFIG_DB sends a new SET for the corrected prefix, it would be a different key, so it should be fine. But if the user fixes the prefix in-place (same key, corrected value), does the framework re-call `set_handler`? Worth verifying.

shixizhang (commented):
Re-reviewing the 3 commits since my prior LGTM (d46654a6, 39383d18, 07ad5a42). The changes look reasonable — moving validation up to `set_handler` as an early gate is cleaner, and guarding `directory.get()` with `path_exist()` prevents a crash when BBR table hasn't been initialized. Two minor inline notes below, neither blocking.

src/sonic-bgpcfgd/bgpcfgd/managers_aggregate_address.py
Guoyuan001: Any err syslog during tests requires adding loganalyzer.expect_regex entries. To be simple, I keep it as info

src/sonic-bgpcfgd/bgpcfgd/managers_aggregate_address.py
Guoyuan001: Returning True on invalid prefix is correct and safe. The Manager framework has no "already handled" cache — it does not track which keys returned True. Every CONFIG_DB SET event (including in-place updates to the same key) triggers a fresh call to set_handler() with the new data.

So both scenarios work:

User fixes prefix with a new key → new SET event → set_handler() called → works
User fixes prefix in-place (same key) → SubscriberStateTable fires a new SET event → set_handler() called again with corrected data → works

shixizhang (approved):
Re-reviewed the 3 commits since my prior LGTM (d46654a6, 39383d18, 07ad5a42). Changes are sound — early validation gate in set_handler, path_exist() guard for BBR table, and refined BBR-unknown logic are all improvements. Two minor inline notes posted (log-level consistency and framework retry behavior) are non-blocking. LGTM!

Blueve (approved):

PR #26638, opened 2026-04-07
Runtime knob for Southbound ZMQ

inder-nexthop:
#### Why I did it
Enable runtime control of ZMQ communication between orchagent and syncd (southbound) via a CONFIG_DB knob.

##### Work item tracking

#### How I did it

1. Added orch_southbound_zmq_enabled leaf to sonic-device_metadata.yang with default false.
2. Added orch_southbound_zmq_enabled to swss_vars.j2 so the value is rendered from CONFIG_DB's DEVICE_METADATA into SWSS_VARS available to container startup scripts.
3. Modified orchagent.sh to read the knob from SWSS_VARS (via jq) and start orchagent with -z zmq_sync when enabled.

#### How to verify it

1. Enable the knob: sonic-db-cli CONFIG_DB hset "DEVICE_METADATA|localhost" "orch_southbound_zmq_enabled" "true"
2. Restart swss: sudo systemctl restart swss
3. Verify orchagent started with ZMQ flags: ps aux | grep orchagent should show -z zmq_sync

Add runtime CONFIG_DB knob (orch_southbound_zmq_enabled) to enable ZMQ communication between orchagent and syncd for southbound route programming.

#### Related PRs:
https://github.com/sonic-net/sonic-sairedis/pull/1835
https://github.com/sonic-net/sonic-swss/pull/4556

### DIFF
--- a/dockers/docker-orchagent/orchagent.sh
+++ b/dockers/docker-orchagent/orchagent.sh
@@ -35,8 +35,11 @@ fi
 # Set zmq mode by default for smartswitch DPU and increase the max bulk limit
 # Otherwise, set synchronous mode if it is enabled in CONFIG_DB
 SYNC_MODE=$(echo $SWSS_VARS | jq -r '.synchronous_mode')
+SOUTHBOUND_ZMQ=$(echo $SWSS_VARS | jq -r '.orch_southbound_zmq_enabled')
 if [ "$LOCALHOST_SWITCHTYPE" == "dpu" ]; then
     ORCHAGENT_ARGS+="-z zmq_sync -k 65536 "
+elif [ "$SOUTHBOUND_ZMQ" == "true" ]; then
+    ORCHAGENT_ARGS+="-z zmq_sync "
 elif [ "$SYNC_MODE" == "enable" ]; then
     ORCHAGENT_ARGS+="-s "
 fi
--- a/files/build_templates/swss_vars.j2
+++ b/files/build_templates/swss_vars.j2
@@ -12,5 +12,6 @@
     {% endif -%}
     "dual_tor": {% if DEVICE_METADATA.localhost.type == "ToRRouter" and DEVICE_METADATA.localhost.subtype == "DualToR" %}"enable"{% else %}"disable"{% endif %},
     "dscp_remapping": {% if SYSTEM_DEFAULTS is defined and SYSTEM_DEFAULTS.tunnel_qos_remap is defined and SYSTEM_DEFAULTS.tunnel_qos_remap.status == "enabled" %}"enable"{% else %}"disable"{% endif %},
+    "orch_southbound_zmq_enabled": {% if DEVICE_METADATA.localhost.orch_southbound_zmq_enabled is defined and DEVICE_METADATA.localhost.orch_southbound_zmq_enabled == "true" %}"true"{% else %}"false"{% endif %},
     "switch_type": "{{ DEVICE_METADATA.localhost.switch_type }}"
 }
--- a/platform/vs/docker-sonic-vs/orchagent.sh
+++ b/platform/vs/docker-sonic-vs/orchagent.sh
@@ -49,9 +49,12 @@ fi
 # Set zmq mode by default for DPU vs
 # Otherwise, set synchronous mode if it is enabled in CONFIG_DB
 SYNC_MODE=$(echo $SWSS_VARS | jq -r '.synchronous_mode')
+SOUTHBOUND_ZMQ=$(echo $SWSS_VARS | jq -r '.orch_southbound_zmq_enabled')
 
 if [ "$SWITCH_TYPE" == "dpu" ]; then
     ORCHAGENT_ARGS+="-z zmq_sync -k 65536 "
+elif [ "$SOUTHBOUND_ZMQ" == "true" ]; then
+    ORCHAGENT_ARGS+="-z zmq_sync "
 elif [ "$SYNC_MODE" == "enable" ]; then
     ORCHAGENT_ARGS+="-s "
 fi
--- a/src/sonic-yang-models/yang-models/sonic-device_metadata.yang
+++ b/src/sonic-yang-models/yang-models/sonic-device_metadata.yang
@@ -318,6 +318,12 @@ module sonic-device_metadata {
                     default "false";
                 }
 
+                leaf orch_southbound_zmq_enabled {
+                    type boolean;
+                    description "Enable ZMQ feature between orchagent and syncd (southbound).";
+                    default "false";
+                }
+
                 leaf syslog_with_osversion {
                     type boolean;
                     description "Enable syslog with OS version feature.";


dockers/docker-orchagent/orchagent.sh
```diff
@@ -35,8 +35,11 @@ fi
 # Set zmq mode by default for smartswitch DPU and increase the max bulk limit
 # Otherwise, set synchronous mode if it is enabled in CONFIG_DB
 SYNC_MODE=$(echo $SWSS_VARS | jq -r '.synchronous_mode')
+SOUTHBOUND_ZMQ=$(echo $SWSS_VARS | jq -r '.orch_southbound_zmq_enabled')
 if [ "$LOCALHOST_SWITCHTYPE" == "dpu" ]; then
     ORCHAGENT_ARGS+="-z zmq_sync -k 65536 "
+elif [ "$SOUTHBOUND_ZMQ" == "true" ]; then
+    ORCHAGENT_ARGS+="-z zmq_sync -k 10000 "
```
deepak-singhal0408: -k 10000 is clubbed with -z zmq_sync here — bulk size and transport mode are independent tuning knobs. An operator may want ZMQ for lower latency but keep the default bulk size (1000). Keeping higher bulk size has other side effects as well

Can we not club them together please.

inder-nexthop: Yes I will remove that.

deepak-singhal0408 (approved):
LGTM. thanks,

dockers/docker-orchagent/orchagent.sh
```diff
@@ -35,8 +35,11 @@ fi
 # Set zmq mode by default for smartswitch DPU and increase the max bulk limit
 # Otherwise, set synchronous mode if it is enabled in CONFIG_DB
 SYNC_MODE=$(echo $SWSS_VARS | jq -r '.synchronous_mode')
+SOUTHBOUND_ZMQ=$(echo $SWSS_VARS | jq -r '.orch_southbound_zmq_enabled')
 if [ "$LOCALHOST_SWITCHTYPE" == "dpu" ]; then
     ORCHAGENT_ARGS+="-z zmq_sync -k 65536 "
+elif [ "$SOUTHBOUND_ZMQ" == "true" ]; then
```
deepak-singhal0408, line 41: This same `SOUTHBOUND_ZMQ` elif block also needs to be added to `platform/vs/docker-sonic-vs/orchagent.sh` (the all-in-one VS container script). That file currently only has DPU and sync-mode branches (lines 53-57):

```bash
if [ "$SWITCH_TYPE" == "dpu" ]; then
    ORCHAGENT_ARGS+="-z zmq_sync -k 65536 "
elif [ "$SYNC_MODE" == "enable" ]; then
    ORCHAGENT_ARGS+="-s "
fi
```

Without this fix, the ZMQ knob won't work in `docker-sonic-vs` (used by sairedis vstest CI), which blocks vstest coverage for this feature.

deepak-singhal0408 (commented):
The same `SOUTHBOUND_ZMQ` elif needs to be added to `platform/vs/docker-sonic-vs/orchagent.sh` as well — see inline comment.

deepak-singhal0408 (changes requested):

deepak-singhal0408:
>Modified orchagent.sh to read the knob from SWSS_VARS (via jq) and start orchagent with -z zmq_sync -k 10000 when enabled.
How to verify it
Enable the knob: sonic-db-cli CONFIG_DB hset "DEVICE_METADATA|localhost" "orch_southbound_zmq_enabled" "true"
Restart swss: sudo systemctl restart swss
Verify orchagent started with ZMQ flags: ps aux | grep orchagent should show -z zmq_sync -k 10000.

Also, update the description to remove -k 10000?

deepak-singhal0408 (approved):

deepak-singhal0408:
Since https://github.com/sonic-net/sonic-swss/pull/4451 got closed, pasting the design proposal here for later reference..

>>>>>
Fundamentally, I consider zmq as Infra/PI piece, and coupling this with platforms/HWSKUs creates duplication/complex code.

I was going through of this PR set and the ASIC_DB persistence PRs (https://github.com/sonic-net/sonic-sairedis/pull/1801, https://github.com/sonic-net/sonic-sairedis/pull/1819). I think there's an opportunity to significantly simplify the design. Wanted to share my thoughts.. up for discussion.

Core observation: The current design couples three independent decisions into one trigger:

ZMQ transport ON/OFF
ASIC_DB persistence ON/OFF
context_config.json required
On DPU, coupling these made sense — DPUs don't need ASIC_DB, and they already ship context_config.json. But for non-DPU switches, this coupling creates the problems we're solving with -g, ASIC_DB restore PRs, and per-hwsku JSON files.. we have bunch of platforms which dont ship with context_config.json.. Also this PR set had a need to add context_config.json for vs..(JFYI, in the past, we had built some similar platform dependency in Chassis for Midplane-IPs and later had to move it out of platform specific code)..

Proposed simplification (4 changes):

ContextConfig.cpp — Change defaults from ipc:///tmp/zmq_ep to tcp://127.0.0.1:5555. Every platform
that ships context_config.json already uses TCP localhost. The IPC default looks like dead code (please correct me if I am wrong). This eliminates the need for context_config.json just to get ZMQ transport working.

Syncd.cpp — Tighten the DisabledRedisClient guard from if (zmq_enable && !VS) to if
(zmq_enable && !VS && isDpu). Non-DPU switches get ZMQ transport with ASIC_DB preserved — tooling
(routeCheck, SNMP, show commands) works immediately without any ASIC_DB restore mechanism. To also
remove syncd's Redis write from the hot path, above PRs for async persistence is still valuable but
becomes a pure performance optimization, not a correctness requirement. Much cleaner scope for those PRs.

orchagent.sh / syncd_init_common.sh — Read a CONFIG_DB knob (zmq_southbound or similar). If enabled, pass -z zmq_sync. Keep the existing elif switch_type == dpu as fallback for backward compatibility — DPU vendors can migrate to the CONFIG_DB knob at their own pace.

No -g flag needed — I assume this is to keep gearbox communication channel via redis? if so, Gearbox is naturally handled by process separation (syncd gets -z, gbsyncd doesn't — they're separate processes with independent command lines). Orchagent still reads context_config.json for gearbox topology (guids), and selectively enables zmq for the required context.

Result: Any switch — can enable ZMQ with one CONFIG_DB key. No context_config.json needed. No ASIC_DB breakage. DPUs continue working unchanged. And https://github.com/sonic-net/sonic-sairedis/pull/1801, https://github.com/sonic-net/sonic-sairedis/pull/1819 can focus purely on async perf optimization rather than correctness restoration.

deepak-singhal0408:
@prsunny @prabhataravind, please also help review/sign-off? thanks,

files/build_templates/swss_vars.j2
```diff
@@ -12,5 +12,6 @@
     {% endif -%}
     "dual_tor": {% if DEVICE_METADATA.localhost.type == "ToRRouter" and DEVICE_METADATA.localhost.subtype == "DualToR" %}"enable"{% else %}"disable"{% endif %},
     "dscp_remapping": {% if SYSTEM_DEFAULTS is defined and SYSTEM_DEFAULTS.tunnel_qos_remap is defined and SYSTEM_DEFAULTS.tunnel_qos_remap.status == "enabled" %}"enable"{% else %}"disable"{% endif %},
+    "orch_southbound_zmq_enabled": {% if DEVICE_METADATA.localhost.orch_southbound_zmq_enabled is defined and DEVICE_METADATA.localhost.orch_southbound_zmq_enabled == "true" %}"true"{% else %}"false"{% endif %},
```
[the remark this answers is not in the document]
inder-nexthop: swss_vars.j2 just renders the value from CONFIG_DB into SWSS_VARS. The P4 orchagent.sh never reads orch_southbound_zmq_enabled from it, so the value is there but ignored.

prabhataravind (approved):

vganesan-nokia:
@inder-nexthop, It would be good to use this same config knob to enable southbound zmq counterpat in syncd, Take value of this knob (aready avaiable from swss_vars) update command line option to enable zmq in syncd (modify script syncd_init_common.sh) similar to done in orchagent.sh for swss.

inder-nexthop:
> @inder-nexthop, It would be good to use this same config knob to enable southbound zmq counterpat in syncd, Take value of this knob (aready avaiable from swss_vars) update command line option to enable zmq in syncd (modify script syncd_init_common.sh) similar to done in orchagent.sh for swss.

@vganesan-nokia this is taken care of in https://github.com/sonic-net/sonic-sairedis/pull/1835

deepak-singhal0408:
@kperumalbfn @lguohan could you please help merge this PR? thanks,

Issue #25745, opened 2026-02-28
Bug: gNMI authentication fails with password


jasper-hou-micas:
### Is it platform specific

generic

### Importance or Severity

Medium

### Description of the bug

When password authentication is used, the **admin** user fails to authenticate while running the `gnmi_get` command, returning the error:
`unknown user admin`.

The root cause is that the host’s `/etc/passwd` is mounted into the container as a directory, which prevents the gNMI docker from accessing the user information.

### Steps to Reproduce

```
gnmi_get \
  -xpath_target SONiC-YANG \
  -xpath /sonic-interface:sonic-interface/ \
  -target_addr 127.0.0.1:8090 \
  -username admin \
  -password YourPaSsWoRd \
  -insecure
```

#### Expected:

The interface information is returned successfully.

#### Actual
Authentication fails with the error: `unknown user admin`.

### Relevant log output

```shell
root@sonic:/# ps -ef |grep 8090
root      419229  416428  0 17:06 pts/3    00:00:01 /usr/sbin/telemetry -logtostderr --noTLS --por  t 8090 --allow_no_client_auth -v=7 --threshold 100 --idle_conn_duration 5 --client_auth password
root      449282  448538  0 17:14 pts/4    00:00:00 grep 8090
root@sonic:/#
gnmi_get   -xpath_target SONiC-YANG   -xpath /sonic-interface:sonic-interface/   -taet_addr 127.0.0.1:8090   -username admin   -password YourPaSsWoRd -insecure


[9] false} for the node name: sonic-interface
I0227 17:15:49.052458  451646 path_validator.go:179] validatePath: module name: sonic-interface fo  und for the node sonic-interface:
I0227 17:15:49.052470  451646 path_validator.go:189] validatePath: appeneded the module prefix nam  e for the path node: name:"sonic-interface:sonic-interface"
I0227 17:15:49.052488  451646 path_validator.go:296] updateStructFieldVal: struct field: &{SonicIn  terface  *ocbinds.SonicInterface_SonicInterface path:"sonic-interface" module:"sonic-interface" 72   [9] false}; struct field type: *ocbinds.SonicInterface_SonicInterface
I0227 17:15:49.052504  451646 path_validator.go:307] updateStructFieldVal: field type is pointer
I0227 17:15:49.052514  451646 path_validator.go:314] updateStructFieldVal: sVal&{<nil>}
[...]

PR #26656, opened 2026-04-08
mgmt-framework: set secure default for REST server authentication

xq9mend:
#### Why I did it

When the `REST_SERVER` table is absent from CONFIG_DB (the default on a fresh image), `rest-server.sh` left `CLIENT_AUTH` unset, causing the REST server to start without authentication.

This change defaults `CLIENT_AUTH` to `user` when it is not explicitly configured, so the REST server requires authentication on every request unless the operator explicitly sets a different mode in CONFIG_DB.

#### How I did it

Added a fallback in `rest-server.sh` to set `CLIENT_AUTH=user` when it is not populated from CONFIG_DB.

#### How to verify it

On a default image (no `REST_SERVER` table in CONFIG_DB), verify the REST server requires authentication:
```
curl -sk https://<switch>:443/restconf/data/openconfig-system:system -o /dev/null -w '%{http_code}\n'
```
Expected: `401` (previously `200` on default images).

Authenticated requests continue to work normally.

---

> **⚠️ Breaking Change**
>
> Deployments that rely on unauthenticated REST access (no `REST_SERVER` entry in CONFIG_DB) will require authentication after upgrading to an image containing this change.
>
> **Migration:** To explicitly preserve unauthenticated access, set `client_auth = none` in `REST_SERVER|default` in CONFIG_DB before upgrading:
> ```
> sonic-db-cli CONFIG_DB hset "REST_SERVER|default" client_auth none
> ```


xq9mend:
@sachinholla @tahmed-dev @dutta-partha

dockers/docker-sonic-mgmt-framework/rest-server.sh
```diff
@@ -25,6 +25,11 @@ if [ -n "$REST_SERVER" ]; then
     CA_CRT=$(echo $REST_SERVER | jq -r '.ca_crt // empty')
 fi
 
+# Default to user authentication when not explicitly configured
+if [ -z "$CLIENT_AUTH" ]; then
+    CLIENT_AUTH="user"
```
tahmed-dev, line 30: When `REST_SERVER|default` exists in CONFIG_DB but `client_auth` is deliberately omitted (as opposed to the entire table being absent), `jq -r '.client_auth // empty'` still returns empty — so this fallback would also override that case. Is that the intended behavior? If an operator explicitly configures a `REST_SERVER` entry but leaves `client_auth` unset, should that be treated the same as "no `REST_SERVER` table at all"?

dockers/docker-sonic-mgmt-framework/rest-server.sh
```diff
@@ -25,6 +25,11 @@ if [ -n "$REST_SERVER" ]; then
     CA_CRT=$(echo $REST_SERVER | jq -r '.ca_crt // empty')
 fi
 
+# Default to user authentication when not explicitly configured
+if [ -z "$CLIENT_AUTH" ]; then
```
tahmed-dev, line 29: The Go binary defines three modes: `none`, `cert`, and `user`. The `cert` mode requires `--cacert` (CA certificate) and enforces mutual TLS. Given that this default activates `user` mode (password auth over TLS with a self-signed cert on fresh images), have we considered whether `cert` would be a stronger secure default? Or would that break too many existing deployments that rely on password-based REST access?

dockers/docker-sonic-mgmt-framework/rest-server.sh
```diff
@@ -25,6 +25,11 @@ if [ -n "$REST_SERVER" ]; then
     CA_CRT=$(echo $REST_SERVER | jq -r '.ca_crt // empty')
 fi
 
+# Default to user authentication when not explicitly configured
```
tahmed-dev, line 28: The fallback is placed after the `REST_SERVER` parsing block but before the cert/X509 handling. This means it also fires when `REST_SERVER` exists but `client_auth` is missing. Would it be cleaner to set the default inside the `if [ -n \"$REST_SERVER\" ]` block as an `else` on the jq extraction, e.g.,

```bash
CLIENT_AUTH=$(echo $REST_SERVER | jq -r '.client_auth // "user"')
```

That way the fallback only applies when the table exists but the field is missing, and a separate default handles the "no table at all" case — making the two scenarios explicitly distinct.

tahmed-dev (commented):
Thanks for the security fix — the intent is solid. A few questions below.

tahmed-dev:
One more general question: for existing deployments that have been running without `REST_SERVER` in CONFIG_DB (and relying on unauthenticated REST access), this change will silently require authentication after an image upgrade. Is there a release note or migration note planned? Operators running automation against the REST API without credentials would break on upgrade.

dockers/docker-sonic-mgmt-framework/rest-server.sh
xq9mend: Good suggestion. Applied inline `// "user"` default in the jq. Kept the outer fallback for fresh installs where REST_SERVER table doesnt exist.

dockers/docker-sonic-mgmt-framework/rest-server.sh
xq9mend: cert mode requires client certificates which most existing deployments dont have set up. user mode (TLS + password) is a meaningful improvement over none without breaking existing REST API consumers.

dockers/docker-sonic-mgmt-framework/rest-server.sh
xq9mend: Intentional. If REST_SERVER table exists but client_auth is omitted, we treat that as not configured and apply the secure default. Operator who wants different auth should set client_auth explicitly.

sachinholla:
hi @xq9mend, is password authentication working now? The gNMI server authentication issue #25745 was applicable for REST server too.

xq9mend:
This PR focuses on the security concern only — ensuring REST server does not start with no authentication by default. It does not intend to fix the underlying password authentication issue (#25745). That is tracked separately.

xq9mend:
Valid concern. This is an intentional security improvement — silent unauthenticated REST access is being closed. A note will be added to the PR description: operators who need to temporarily keep unauthenticated access can explicitly set `client_auth = none` in `REST_SERVER|default` in CONFIG_DB.

tahmed-dev (approved):

yxieca (approved):

PR #26679, opened 2026-04-09
[SmartSwitch] Support maximum dash configuration size

croos12:
Depends on: https://github.com/sonic-net/sonic-gnmi/pull/643

#### Why I did it
At max DASH scale on the DPU, the current pipeline can't deliver the full config:
- orchagent ZMQ bulk limit is too small for the number of items pushed.
- sonic-gnmi max send/recv message sizes are too small for the payloads.

### How I did it
| Component  | Parameter                                   | Before  | After         |
| ---------- | ------------------------------------------- | ------- | ------------- |
| orchagent  | `-k` bulk limit (DPU, via `DPU_BATCH_SIZE`) | 65,536  | 125,000 items |
| sonic-gnmi | `-max_recv_msg_size` / `-max_send_msg_size` | default | 32 MB         |

Individual gNMI messages stay below 32 MB even at max scale, so 32 MB is sufficient. 

### How to verify it

Push the maximum DASH scale config through gNMI on a SmartSwitch + DPU and confirm the full config is applied.


vivekrnv (dismissed):

vivekrnv:
@prsunny FYI

prsunny:
@croos12 , please address copilot comments

dockers/docker-sonic-gnmi/gnmi-native.sh
```diff
@@ -88,7 +88,7 @@ fi
 # Enable ZMQ for SmartSwitch
 LOCALHOST_SUBTYPE=`sonic-db-cli CONFIG_DB hget "DEVICE_METADATA|localhost" "subtype"`
 if [[ x"${LOCALHOST_SUBTYPE}" == x"SmartSwitch" ]]; then
-    TELEMETRY_ARGS+=" -zmq_port=8100"
+    TELEMETRY_ARGS+=" -zmq_port=8100 -max_recv_msg_size=$((32*1024*1024)) -max_send_msg_size=$((32*1024*1024))"
```
prabhataravind, line 91: @croos12 with a batch size of 125k and gnmi send/recv buffer sizes increased to 32MB, you might also need to increase MQ_RESPONSE_MAX_COUNT (in swss-common) that determines the zmqclient send buffer and zmqserver recv buffer sizes. Currently it is set to 16MB.

croos12: I'm going to need to see why tests are failing, but I have a PR here: https://github.com/sonic-net/sonic-swss-common/pull/1178

vivekkarri: @prabhataravind, we need not increase MQ_RESPONSE_MAX_COUNT (for now). Let me explain. 

Currently, AFAIK and correct me if i'm wrong, sonic-gnmi makes one set call on ZmqProducerStateTable per update. So 100k updated would be equivalent in making 100k Set calls. So, the buffer size will not have to be 32 MB. 

ZmqProducerStateTable supports a batch API

`
void ZmqProducerStateTable::set(const std::vector<KeyOpFieldsValuesTuple> &values)
`

So, firstly sonic-gnmi needs to be updated and then we should evaluate if an increase in MQ_RESPONSE_MAX_COUNT is warranted and if so, if 32 MB is enough

but for now, increase in MQ_RESPONSE_MAX_COUNT is not in scope of this change

prabhataravind: Agree. We can revisit this when gnmi server uses batching.

@hdwhdw is this something we plan to do? Seems like this would be a useful enhancement for operations at scale.

prabhataravind: Created an enhancement request to do batching in gnmi - https://github.com/sonic-net/sonic-buildimage/issues/27250

croos12:
@prabhataravind it is only failing the optional trivy job

prabhataravind (approved):

prabhataravind:
@kperumalbfn could you help merge?

PR #26421, opened 2026-03-26
Upgrade sflow container to Trixie

yxieca:
> **Note**: This PR was created with the assistance of an AI agent (OpenClaw/Claude).

#### Why I did it

As part of the incremental Trixie migration effort, upgrade the sflow Docker container base image from Bookworm (docker-swss-layer-bookworm) to Trixie (docker-swss-layer-trixie).

##### Work item tracking
- Microsoft ADO:

#### How I did it

- Updated `Dockerfile.j2` to use `docker-swss-layer-trixie` as the base image
- Updated `docker-sflow.mk` to reference TRIXIE instead of BOOKWORM for:
  - Debug dependencies
  - Debug image packages
  - Load dockers
  - Docker category lists (SONIC_TRIXIE_DOCKERS / SONIC_TRIXIE_DBG_DOCKERS)

#### How to verify it

- Build sonic-vs image with this change
- Run sflow test suite against KVM testbed

**Local KVM test results** (`sflow/test_sflow.py` with `--device_type=vs`, sonic-mgmt at `1577c00c2` including PR #22186 sflow fixes):

| | Trixie |
|---|---|
| Passed | 14 |
| Failed | 1 |

- 1 failure (`TestAgentId::testDelAgent`): After deleting sflow agent-id, hsflowd does not write `agentIP` to `hsflowd.auto` within the 30s timeout. Retest with latest sonic-mgmt (including #22186) in progress.
- Note: CI skips all sflow tests due to sonic-mgmt#21701 (blanket conditional_mark skip). Local testing is the only validation.

#### Which release branch to backport (provide reason below if selected)

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511

#### Tested branch (Please provide the tested image version)

- [x] master (locally built sonic-vs from commit 2e2cb14fe)

#### Description for the changelog

Upgrade sflow container from Bookworm to Trixie

#### Link to config_db schema for YANG module changes

N/A — no YANG model changes

#### A picture of a cute animal (not mandatory but encouraged)

🦊

PR #26699, opened 2026-04-10
hsflowd: defer redisAsyncCommand calls from within callbacks

yxieca:
## Description
libhiredis 1.2.0 (Debian trixie) added a strict assertion (`assert(!(c->flags & REDIS_IN_CALLBACK))`) that prevents calling `redisAsyncCommand()` from within a Redis async callback. hsflowd's `mod_sonic.c` does exactly this in several callback functions, causing SIGABRT on trixie.

This patch defers all Redis async command calls from callbacks to the next `evt_tick` iteration using boolean flags. The evt_tick state machine already runs periodically and now checks these flags to process deferred requests.

### Changes
- Add deferred-request flags to `HSP_mod_SONIC` struct
- Replace direct `redisAsyncCommand()` calls in callbacks with flag-setting
- Process deferred requests in `evt_tick` handler

### Testing
- 15/15 sflow tests pass on VS KVM testbed with trixie sflow container
- Bookworm compatibility maintained (flags are simply never set if callbacks work directly)

### Related
- Needed for sflow trixie upgrade (#26421)
- See also: #26446 (system-health fix)

### DIFF
--- a/src/sflow/hsflowd/patch/0006-mod_sonic-defer-redis-commands.patch
+++ b/src/sflow/hsflowd/patch/0006-mod_sonic-defer-redis-commands.patch
@@ -1,0 +1,105 @@
+--- a/src/Linux/mod_sonic.c
++++ b/src/Linux/mod_sonic.c
+@@ -146,6 +146,10 @@
+     char *sflow_agent;
+     UTHash *collectors;
+     UTArray *newCollectors;
++    bool reqLagRefresh:1;
++    bool reqSflowRefresh:1;
++    bool reqCollectorRefresh:1;
++    bool collectorsNeedSync:1;
+     EVEvent *configStartEvent;
+     EVEvent *configEvent;
+     EVEvent *configEndEvent;
+@@ -1011,7 +1015,10 @@
+       }
+     }
+     // there may be more to map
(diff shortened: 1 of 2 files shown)


src/sflow/hsflowd/patch/0006-mod_sonic-defer-redis-commands.patch
```diff
@@ -0,0 +1,105 @@
...
++    mdata->reqCollectorRefresh = YES;
+   }
+ 
+   static void dbEvt_subscribeCB(redisAsyncContext *ctx, void *magic, void *req_magic)
+@@ -1924,7 +1940,23 @@
+       if(!discoverNewPorts(mod))
+ 	mapPorts(mod);
+       syncSwitchPorts(mod);
+-      discoverNewCollectors(mod);
++      if(!discoverNewCollectors(mod) && mdata->collectorsNeedSync) {
```
saiarcot895, line 86: Minor stylistic choice, but use `discoverNewCollectors(mod) == NO` instead of `!discoverNewCollectors(mod)` to match the pattern used in some other places of the file.

saiarcot895 (approved):

PR #26700, opened 2026-04-10
hsflowd: stop overriding SIGABRT signal handler

yxieca:
## Description
hsflowd registers a custom signal handler for SIGABRT which catches assertion failures and prevents proper core dump generation. This makes debugging runtime crashes very difficult — the process continues in a broken state instead of terminating with a useful core file.

### Changes
- Remove `sigaction(SIGABRT, &sa, NULL)` from hsflowd main() signal handler setup
- Default SIGABRT handler will now produce proper core dumps

### Testing
- 15/15 sflow tests pass on VS KVM testbed
- No behavioral change for normal operation (SIGABRT only fires on assertion failures)

### Related
- Helps debug sflow trixie issues (#26421)
- libhiredis 1.2.0 assertion triggers SIGABRT when redisAsyncCommand called from callbacks

### DIFF
--- a/src/sflow/hsflowd/patch/0007-hsflowd-stop-overriding-sigabrt.patch
+++ b/src/sflow/hsflowd/patch/0007-hsflowd-stop-overriding-sigabrt.patch
@@ -1,11 +1,16 @@
 --- a/src/Linux/hsflowd.c
 +++ b/src/Linux/hsflowd.c
-@@ -1849,7 +1849,7 @@
-     sigaction(SIGILL, &sa, NULL);
-     sigaction(SIGBUS, &sa, NULL);
-     sigaction(SIGXFSZ, &sa, NULL);
+@@ -1844,14 +1844,9 @@
+     sa.sa_sigaction = signal_handler;
+     sigemptyset(&sa.sa_mask);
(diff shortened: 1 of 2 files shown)


yxieca:
Note: only the crash signal overrides (SIGSEGV, SIGILL, SIGBUS, SIGXFSZ, SIGABRT) are removed. The remaining signals are retained because they have useful application-level handlers:

- **SIGTERM** → graceful shutdown (EVStop)
- **SIGINT** → abrupt exit with cleanup
- **SIGUSR1** → decrease debug log level
- **SIGUSR2** → increase debug log level
- **SIGHUP** → reopen log files (for logrotate)

The crash signal handler was catching fatal signals, logging a text backtrace, then calling `exit(sig)` — which suppressed core dump generation. With the default kernel handlers restored, crashes will now produce proper core dumps for post-mortem debugging with GDB.

yxieca:
This PR includes the 0006 defer-redis patch from #26699 and should be merged after #26699.

yxieca:
Close and reopen to retrigger the CI checkers.

saiarcot895 (approved):

PR #26704, opened 2026-04-10
[restapi-sidecar]: Switch restapi.service User from root to admin and add stale unit cleanup

maipbui:
[restapi-sidecar]: Switch restapi.service User from root to admin and add stale unit cleanup

#### Why I did it
Port the changes from telemetry-sidecar PR #26267 to restapi-sidecar.

Previous sidecar versions overwrote `/lib/systemd/system/restapi.service` with a variant containing `User=root` (needed for kubectl). Now that kubectl is replaced by docker commands, we no longer need root and should restore `User=admin`. Additionally, syncing the service file every cycle caused endless overwrites due to systemd generators ingesting data into the unit file.
##### Work item tracking
- Microsoft ADO **(number only)**: 37486325

#### How I did it
1. Updated all per-branch `restapi.service` files (`202311`–`202511`) to use `User=admin` instead of `User=root`.
2. Stopped syncing `restapi.service` every cycle — removed it from `SYNC_ITEMS` and `POST_COPY_ACTIONS` in `systemd_stub.py`.
3. Added a one-shot stale unit cleanup (`_cleanup_stale_service_unit()`) that detects `User=root` on the host and restores the clean packed per-branch version. The cleanup is branch-aware (calls `_get_branch_name()` + `_resolve_branch()` to select the correct service file).
4. Updated tests
#### How to verify it
https://elastictest.org/scheduler/testplan/69ea34720aae266263d6520e

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511

- [ ] 
- [ ]


FengPan-Frank (approved):

dockers/docker-restapi-sidecar/systemd_stub.py
```diff
@@ -94,10 +95,6 @@ def _get_branch_name() -> str:
         return "private"
 
 POST_COPY_ACTIONS = {
-    "/lib/systemd/system/restapi.service": [
-        ["sudo", "systemctl", "daemon-reload"],
```
qiluo-msft, line 98: Do we still need systemctl daemon-reload ?

qiluo-msft: @make1980 to check

make1980: yes - that's still needed if we ever change the systemd config file

maipbui: If `User=root` is found in `restapi.service`, `_cleanup_stale_service_unit()` already handles the User=root → `User=admin` fix including daemon-reload + restart restapi. Otherwise, the service file is already correct and no reload/restart is needed.

So I don't think we need adding reload and restart to POST_COPY_ACTIONS. @FengPan-Frank can you confirm?

qiluo-msft (approved):

Pterosaur:
Hi @maipbui , should we backport this commit to the branches 202311–202511?

PR #26726, opened 2026-04-12
[orchagent] Add async swss.rec config knob support

vpandian-nokia:
#### Why I did it

This change is the companion `sonic-buildimage` update for the async `swss.rec` support added in [`sonic-net/sonic-swss#4400`](https://github.com/sonic-net/sonic-swss/pull/4400).

It adds operational control for async `swss.rec` enablement through `DEVICE_METADATA|localhost|async_swss_rec`.

The `orchagent` startup script now reads `async_swss_rec` from `CONFIG_DB` and appends `-A` only when the knob is explicitly set to `enabled`.
This change also updates `sonic-device_metadata.yang` so `async_swss_rec` is modeled as part of `DEVICE_METADATA`, with:
- allowed values: `enabled` / `disabled`
- default: `disabled`

##### Work item tracking

#### How I did it

- Updated `dockers/docker-orchagent/orchagent.sh` to read `DEVICE_METADATA|localhost|async_swss_rec` from `CONFIG_DB` and append `-A` only when the knob is explicitly set to `enabled`.
- Updated `src/sonic-yang-models/yang-models/sonic-device_metadata.yang` to model `async_swss_rec` as part of `DEVICE_METADATA`, with allowed values `enabled` / `disabled` and default `disabled`.

#### How to verify it

- Confirmed that with no `async_swss_rec` setting, `orchagent` starts without `-A`.
- Confirmed that with `DEVICE_METADATA|localhost|async_swss_rec=enabled`, `orchagent.sh` appends `-A`.
- Confirmed on DUT that `orchagent` starts with `-A` when the knob is enabled.
- Confirmed the YANG model now includes `async_swss_rec` with `enabled` / `disabled` values and default `disabled`.

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511

- [ ] 
- [ ] 

Add `DEVICE_METADATA|localhost|async_swss_rec` support to control async `swss.rec` enablement in `orchagent`.

`DEVICE_METADATA` is modeled in `src/sonic-yang-models/yang-models/sonic-device_metadata.yang`; this change adds the `async_swss_rec` leaf there.

### DIFF
--- a/dockers/docker-orchagent/orchagent.sh
+++ b/dockers/docker-orchagent/orchagent.sh
@@ -62,6 +62,12 @@ if [[ "$NAMESPACE_ID" ]]; then
     ORCHAGENT_ARGS+="-f swss.asic$NAMESPACE_ID.rec -j sairedis.asic$NAMESPACE_ID.rec "
 fi
 
+# Enable async swss recorder when explicitly configured
+ASYNC_SWSS_REC=$(sonic-db-cli CONFIG_DB hget "DEVICE_METADATA|localhost" "async_swss_rec")
+if [ "$ASYNC_SWSS_REC" == "enabled" ]; then
+    ORCHAGENT_ARGS+="-A "
+fi
+
 # Add platform specific arguments if necessary
 if [ "$platform" == "broadcom" ]; then
     ORCHAGENT_ARGS+="-m $MAC_ADDRESS"
--- a/src/sonic-yang-models/yang-models/sonic-device_metadata.yang
+++ b/src/sonic-yang-models/yang-models/sonic-device_metadata.yang
@@ -19,6 +19,10 @@ module sonic-device_metadata {
 
     description "DEVICE_METADATA YANG Module for SONiC OS";
 
+    revision 2026-04-10 {
+        description "Added async_swss_rec field to control async swss recorder.";
+    }
+
     revision 2021-02-27 {
         description "Added frr_mgmt_framework_config field to handle BGP
             config DB schema events to configure FRR protocols.";
@@ -228,6 +232,15 @@ module sonic-device_metadata {
                     must "((current() = 'disabled') or (current() = 'enabled' and ../synchronous_mode = 'enable'))" {
                         error-message "ASIC synchronous mode must be enabled in order to enable suppress FIB pending feature";
                     }
+                }
+
+                leaf async_swss_rec {
+                    description "Enable asynchronous swss.rec recording in orchagent.";
+                    type enumeration {
+                        enum enabled;
+                        enum disabled;
+                    }
+                    default disabled;
                 }
                 leaf rack_mgmt_map {
                     type string {


deepak-singhal0408 (approved):
LGTM.

deepak-singhal0408:
@prsunny @prabhataravind could you please also help review?

prabhataravind (approved):

dockers/docker-orchagent/orchagent.sh
```diff
@@ -62,6 +62,12 @@ if [[ "$NAMESPACE_ID" ]]; then
     ORCHAGENT_ARGS+="-f swss.asic$NAMESPACE_ID.rec -j sairedis.asic$NAMESPACE_ID.rec "
 fi
 
+# Enable async swss recorder when explicitly configured
+ASYNC_SWSS_REC=$(sonic-db-cli CONFIG_DB hget "DEVICE_METADATA|localhost" "async_swss_rec")
```
vganesan-nokia, line 66: Suggest get async_swss_rec value from swss_vars.j2 instead of reading conifg db. This is similar to getting other config db variables such as asic_id, synchronous_mode and so on.

vpandian-nokia: Thanks for the suggestion. Using `swss_vars.j2` would be a reasonable consistency cleanup, but this knob is only read once during startup, and direct `CONFIG_DB` reads are already used for other startup parameters as well. So functionally the current `orchagent.sh` approach is sufficient for this PR. All 23 checks have already passed, and there does not appear to be a technical issue with the current implementation, so my preference would be to avoid expanding scope here unless there is a strong preference to include that alignment in this PR.

vpandian-nokia: Will track `swss_vars.j2` alignment as a follow-up cleanup if there is broader consensus. Resolving this thread to keep this PR scoped to the knob itself, but happy to revisit it in a dedicated PR.

deepak-singhal0408:
@lguohan @rlhui could you please help merge this PR? Thanks,

dockers/docker-orchagent/orchagent.sh
```diff
@@ -62,6 +62,12 @@ if [[ "$NAMESPACE_ID" ]]; then
...
+if [ "$ASYNC_SWSS_REC" == "enabled" ]; then
```
kperumalbfn, line 67: @vpandian-nokia why do we need this from DEVICE_METADATA? Can we update the mode based using other orchagent init files? This is only for the initialization and cannot be updated dynamically.

vpandian-nokia: @kperumalbfn Thanks for the review.

The intent here was to make async `swss.rec` operationally manageable through persistent configuration, with explicit modeling in `DEVICE_METADATA` and default-off behavior.

Although this knob is only used during `orchagent` initialization today, that is consistent with other `DEVICE_METADATA|localhost` settings that act as boot-time / init-time behavior controls rather than dynamically mutable runtime toggles.

If there is a preferred alternative location for init-time `orchagent` behavior knobs, I’m open to aligning with that. For this change, the main goal was to keep it as a persistent, explicitly modeled, default-off knob.

kperumalbfn: Ok, thanks. LGTM

vpandian-nokia: Thanks for the review.

PR #26734, opened 2026-04-13
Adding resource limit to Telemetry

sharmavijay-ms:
##### Work item tracking

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511

- [ ] 
- [ ]

### DIFF
--- a/files/build_templates/telemetry.service.j2
+++ b/files/build_templates/telemetry.service.j2
@@ -13,10 +13,6 @@ ExecStartPre=/usr/local/bin/{{docker_container_name}}.sh start
 ExecStart=/usr/local/bin/{{docker_container_name}}.sh wait
 ExecStop=/usr/local/bin/{{docker_container_name}}.sh stop
 RestartSec=30
-MemoryHigh=300M
-MemoryMax=600M
-CPUQuota=600m
-CPUWeight=300
 
 [Install]
 WantedBy=sonic.target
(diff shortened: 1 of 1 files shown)


files/build_templates/telemetry.service.j2
```diff
@@ -13,6 +13,10 @@ ExecStartPre=/usr/local/bin/{{docker_container_name}}.sh start
 ExecStart=/usr/local/bin/{{docker_container_name}}.sh wait
 ExecStop=/usr/local/bin/{{docker_container_name}}.sh stop
 RestartSec=30
+MemoryHigh=300M
```
zbud-msft, line 16: Setting limits here will not impact telemetry process resource limits. Only telemetry.sh as telemetry.service cgroup only owns the telemetry.sh startup script.

telemetry.service is a thin systemd wrapper whose ExecStart= runs telemetry.sh, which calls docker start telemetry and then docker wait. The only process that ends up in the telemetry.service cgroup is that shell wrapper (a few MB of bash). The actual telemetry container processes are forked by the docker daemon and placed in a separate cgroup — /sys/fs/cgroup/system.slice/docker-<containerid>.scope — which is a sibling of telemetry.service under system.slice/, not a child of it.

You can confirm on any running device:


cat /proc/$(pidof telemetry)/cgroup

zbud-msft (changes requested):

zbud-msft:
We would need to add the resource limits to the actual docker create command...

PR #26735, opened 2026-04-13
Auto generate system port configs for nexthop platforms

lakshmi-nexthop:
#### Why I did it
On a reset, system port configuration can be auto generated (just like regular port configs) for fixed systems (upper t2).

##### Work item tracking
- Microsoft ADO **(number only)**: https://github.com/sonic-net/sonic-buildimage/issues/26364

#### How I did it
- When sonic-cfggen is invoked for factory reset, system ports are also generated based on port_config.ini file along with port configs. This behavior is controlled thru a flag from paltform_config.env.
- Flag is currently set for nexthop ut2 platform only, but can also be used for multi asic voq platforms.

#### How to verify it
- Onie reset and verify that ports come up
- factory reset and verify that ports come up

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511

- [ ] 
- [ ]


device/nexthop/x86_64-nexthop_5010-r0/asic.conf
```diff
@@ -0,0 +1,2 @@
+NUM_ASIC=1
```
arlakshm, line 1: Why do we need this change. The asic.conf is defined only for multi-asic devices. We can get this info from port-config.ini

src/sonic-config-engine/sonic-cfggen
```diff
@@ -403,6 +422,17 @@ def main():
         if  brkout_table is not None:
             deep_update(data, {'BREAKOUT_CFG': brkout_table})
 
+        if device_info.is_generate_voq():
```
arlakshm, line 425: Can we add this when the preset configuration needs to be generated? Otherwise the configuration should be generated from minigraphs or golden_config_db?

arlakshm (commented):
Please add unit tests for this change

device/nexthop/x86_64-nexthop_5010-r0/asic.conf
lakshmi-nexthop: System port config for cpu port needs num cores. And this information for cpu is not present in port_config.ini right now. Even if we take num cores from other front panel ports, theres also cpu port speed that is currently hardcoded.
May be asic.conf can be used to for this purpose or port_config.ini can also include cpu port info. But then the port config generation will need to skip cpu port. 
I am fine to go with either way, but thought asic.conf is a better option.

src/sonic-config-engine/sonic-cfggen
lakshmi-nexthop: Are you suggesting to check for preset in addition to is_generate_voq?

arpit-nexthop (approved):

src/sonic-config-engine/sonic-cfggen
lakshmi-nexthop: ok, I think I understand this comment. will put in a check for preset.

lakshmi-nexthop:
> Please add unit tests for this change

I missed this comment. Added UT also finally.

lakshmi-nexthop:
@arlakshm we have had enough soak in time for this diff. Can you please check?

lakshmi-nexthop:
@arlakshm Can you please review this PR? 
@saravanan-nexthop FYI.

device/nexthop/x86_64-nexthop_5010-r0/asic.conf
lakshmi-nexthop: @arlakshm Can you check this comment and response? I will go with whichever option you choose.

deepak-singhal0408:
@arlakshm could you please re-review this PR. thanks,

saravanan-nexthop:
@tjchadaga could you please review this change

device/nexthop/x86_64-nexthop_5010-r0/asic.conf
vmittal-msft: **[Open thread] `asic.conf` on a single-ASIC device.** This is the crux of @arlakshm's unresolved Apr-19 comment. `asic.conf` is conventionally present only on multi-ASIC platforms, and its presence/`NUM_ASIC` is consulted by multi-ASIC detection and namespace/DB-config paths. Please confirm with maintainers that shipping it here has no side effects on single-ASIC behavior. If it's only needed to source `NUM_CORES`/CPU speed, consider putting that in `platform_env.conf` (or the hwsku dir) rather than introducing `asic.conf`.

src/sonic-config-engine/tests/voq-test-port-config.ini
```diff
@@ -0,0 +1,5 @@
+# name          lanes               alias       index    speed    fec
```
vmittal-msft: **Test config isn't representative of the feature.** The header (`name lanes alias index speed fec`) has none of the VOQ columns (`core_id`, `core_port_id`, `num_voq`) that `parse_system_port_config_file()` maps. As a result the front-panel `SYSTEM_PORT` entries produced in these tests contain only `switch_id/speed/system_port_id`, and `test_system_port_structure_correctness` still passes despite the missing fields. Please add a port-config that includes the VOQ columns and assert `core_index/core_port_index/num_voq` are populated, so the tests cover the real output.

src/sonic-config-engine/portconfig.py
```diff
@@ -211,6 +211,41 @@ def get_port_config(hwsku=None, platform=None, port_config_file=None, hwsku_conf
     else:
         return parse_port_config_file(port_config_file)
 
+def get_system_port_config(hwsku=None, hostname=None, asic_name=None, port_config_file=None):
+    config_db = db_connect_configdb(asic_name)
+    config_db_hwsku = device_info.get_localhost_info('hwsku', config_db=config_db)
+    # Skip CONFIG_DB when caller requested a specific hwsku that doesn't match
+    # the running one — otherwise stale SYSTEM_PORT rows from the live SKU
+    # leak into the dump for an unrelated SKU.
+    if config_db is not None and (hwsku is None or config_db_hwsku == hwsku):
```
vmittal-msft: **Missing `port_config_file is None` guard.** Unlike `get_port_config()` (which reads CONFIG_DB only `if ... and port_config_file is None and ...`), this reads CONFIG_DB whenever the hwsku matches. On a live device a caller that explicitly passes `port_config_file` (e.g. `sonic-cfggen -p ...`) is silently overridden by DB rows. Suggest:
```python
if config_db is not None and port_config_file is None and (hwsku is None or config_db_hwsku == hwsku):
```

src/sonic-config-engine/portconfig.py
```diff
@@ -211,6 +211,41 @@ def get_port_config(hwsku=None, platform=None, port_config_file=None, hwsku_conf
...
+        port_data = config_db.get_table("SYSTEM_PORT")
+        if bool(port_data):
+            port_data_str_keys = {
+                f"{k[0]}|{k[1]}|{k[2]}" if isinstance(k, tuple) else k: v
+                for k, v in port_data.items()
+            }
+            sys_ports = ast.literal_eval(json.dumps(port_data_str_keys))
```
vmittal-msft, line 227: Nit: `ast.literal_eval(json.dumps(...))` round-trips through JSON just to deep-copy. It mirrors the existing `get_fabric_port_config` pattern so it's not blocking, but it will raise if any value serializes to JSON `true/false/null` (invalid Python literals). `port_data_str_keys` is already the shape you want — returning it directly, or `copy.deepcopy()`, is clearer and safer.

src/sonic-config-engine/portconfig.py
```diff
@@ -246,6 +281,57 @@ def parse_port_config_file(port_config_file):
                 port_alias_asic_map[data['alias']] = data['asic_port_name'].strip()
     return (ports, port_alias_map, port_alias_asic_map)
 
+def parse_system_port_config_file(port_config_file, hostname, asic_name, asic_id):
+    ports = {}
+    files = [port_config_file]
```
vmittal-msft, line 291: Nit: `files = [port_config_file]` + `for file in files:` is a single-element loop — dead complexity. Iterate the one file directly (and `file` shadows a builtin).

src/sonic-config-engine/portconfig.py
```diff
@@ -246,6 +281,57 @@ def parse_port_config_file(port_config_file):
...
+                    continue
+
+                tokens = line.split()
+                if len(tokens) < 2:
+                    continue
+
+                name_index = titles.index('name')
+                name = tokens[name_index]
+                # Build a filtered and renamed dictionary
+                field_map = {
```
vmittal-msft, line 325: **Required VOQ fields are effectively optional here.** `core_index`, `core_port_index`, and `num_voq` are added only when the column exists, and `except ValueError: continue` silently skips them otherwise. A standard `port_config.ini` (and the fixture in this PR) has none of these columns, so the resulting `SYSTEM_PORT` rows are incomplete for VOQ orchestration. Please validate these columns are present (or fail clearly) when `generate_system_port_config` is enabled.

vmittal-msft (commented):
Thanks for the change — auto-generating `SYSTEM_PORT` on reset for fixed VOQ (upper-T2) platforms is a useful direction, and gating on `--preset` addresses the earlier feedback. I left inline comments; a few are worth resolving before merge:

**Higher priority**
- Generated `SYSTEM_PORT` rows omit `core_index/core_port_index/num_voq` unless the port_config has those columns — and the new test fixture doesn't, so the tests never exercise a complete entry.
- `hostname` is read from `/etc/hostname` with an empty fallback and can mismatch the preset-set `DEVICE_METADATA` hostname.
- The `asic.conf`-on-single-ASIC question from @arlakshm is still open.

**Medium**
- `get_system_port_config` ignores an explicitly-passed port_config when CONFIG_DB is populated (missing the `port_config_file is None` guard that `get_port_config` has).
- CPU port values and `system_port_id` derivation are hardcoded / single-ASIC-only, though the description mentions multi-ASIC reuse.
- Injection happens before `generate_sample_config`, which is order-fragile.

Plus a few nits (stale comment, dead loop, test assertions). Details inline. Submitting as comments, not blocking.

src/sonic-config-engine/portconfig.py
```diff
@@ -246,6 +281,57 @@ def parse_port_config_file(port_config_file):
...
+                system_port_data = {}
+                system_port_data['switch_id'] = asic_id
+                for original_key, new_key in field_map.items():
+                    try:
+                        idx = titles.index(original_key)
+                        system_port_data[new_key] = tokens[idx]
+                    except ValueError:
+                        continue
+                index = tokens[titles.index('index')]
+                system_port_data['system_port_id'] = int(index)
```
vmittal-msft: `system_port_id` is `int(index)` here but the CPU entry below uses the string `"0"` — inconsistent types for the same `uint32` field; please pick one (int recommended). Also, `index` is the 1-based front-panel index, which is only a valid system-port-id for a self-contained single-ASIC device — it won't be globally unique for the multi-ASIC/chassis reuse the description mentions. A comment documenting the single-ASIC scope would help.

src/sonic-config-engine/portconfig.py
```diff
@@ -246,6 +281,57 @@ def parse_port_config_file(port_config_file):
...
+                system_port_name = hostname + "|" + asic_name + "|" + name
+                ports[system_port_name] = system_port_data
+
+    # add system port for CPU
```
vmittal-msft: The CPU system port is fully hardcoded (`num_voq=8`, `speed=10000`, `core_index/core_port_index/system_port_id=0`). This won't generalize across platforms (or the multi-ASIC reuse mentioned in the description), and `num_voq=8`/`speed=10000` are silent assumptions. Please derive these from `asic.conf`/platform metadata, or at minimum document why they're constant.

src/sonic-config-engine/sonic-cfggen
```diff
@@ -397,6 +417,17 @@ def main():
         if  brkout_table is not None:
             deep_update(data, {'BREAKOUT_CFG': brkout_table})
 
+        if device_info.is_generate_voq() and args.preset is not None:
```
vmittal-msft: **Ordering is fragile.** `SYSTEM_PORT`/`DEVICE_METADATA` are injected here, but `--preset` is applied later via `generate_sample_config()` (~line 576). It only survives because `l1/l2/l3/t1` mutate `data` in place; `generate_empty_config()` rebuilds `data` and would drop `SYSTEM_PORT`, and `generate_t1_sample_config()` overwrites `hostname` afterwards (see next comment). Safer to build the VOQ config *after* `generate_sample_config()` so it's keyed off the final metadata.

src/sonic-config-engine/sonic-cfggen
```diff
@@ -397,6 +417,17 @@ def main():
...
+            hostname = ""
+            try:
+                with open('/etc/hostname', 'r') as f:
```
vmittal-msft: **Hostname sourcing.** Reading `/etc/hostname` with a `""` fallback is risky: an empty hostname yields invalid `SYSTEM_PORT` keys like `|Asic0|Ethernet0` (fails the `hostname` type in `sonic-system-port.yang`), and a `SYSTEM_PORT` hostname must equal `DEVICE_METADATA.localhost.hostname`. Since presets run afterward, `--preset t1` sets `hostname='sonic'`, causing a mismatch with the `/etc/hostname`-keyed system ports. Prefer `device_info.get_hostname()` and/or key the system ports off `data['DEVICE_METADATA']['localhost']['hostname']` after the preset is applied.

src/sonic-py-common/sonic_py_common/device_info.py
```diff
@@ -881,6 +922,42 @@ def is_macsec_supported():
...
 
+# Check if this platform needs to generate voq configs.
+def is_generate_voq():
+    platform_env_conf_file_path = get_platform_env_conf_file_path()
+
+    # platform_env.conf file not present for platform
+    if platform_env_conf_file_path is None:
+        return False
+
+    # Else open the file check for keyword - macsec_enabled -
```
vmittal-msft, line 933: Nit: stale copy-paste comment — this checks `generate_system_port_config`, not `macsec_enabled`.

src/sonic-py-common/sonic_py_common/device_info.py
```diff
@@ -881,6 +922,42 @@ def is_macsec_supported():
...
+               continue
+            if tokens[0].lower() == 'generate_system_port_config':
+                val = tokens[1].strip()
+                if val == '1':
+                    return True
+
+    return False
+
+# Get voq cofig data - number of cores
+def get_num_asic_cores():
```
vmittal-msft, line 947: Minor: `int(num_cores)` will raise if the `NUM_CORES` line is malformed, and the function mixes an `int` early-return (`0`) with a `str`-derived `int()` result. Consider a try/except with a consistent default. (Naming is fine, just confirm `NUM_CORES` maps to the `max_cores` DEVICE_METADATA semantics.)

lakshmi-nexthop: yes, num_cores is max_cores in DEVICE_METADATA

src/sonic-config-engine/sonic-cfggen
lakshmi-nexthop: moved the sysport generation to last after generate_sample_config. 
Also, there are cases like factory reset when config db is not present at all. And for a preset like l1, hostname is not set. In those cases, /etc/hostname has the hostname.
How about following the below order.
 check for hostname in data first, config db next, and then /etc/hostname?

src/sonic-config-engine/sonic-cfggen
lakshmi-nexthop: same as above.

src/sonic-config-engine/portconfig.py
lakshmi-nexthop: for multi asic cases, we need a way to configure these params for CPU. I though asic.conf is the place, but given asic.conf is only present for multi asic, I have deleted asic.conf for now. I think we need to solve where we want to add all params for cpu port.
For now, adding comments.

src/sonic-config-engine/portconfig.py
lakshmi-nexthop: done.

src/sonic-config-engine/tests/voq-test-port-config.ini
lakshmi-nexthop: fixed.

src/sonic-config-engine/portconfig.py
lakshmi-nexthop: done.

device/nexthop/x86_64-nexthop_5010-r0/asic.conf
lakshmi-nexthop: I was missing the fact that asic.conf is multi asic as of now and creating it for single asic might have repurcussions.
So, I removed asic.conf and moved the num cores to platform.env. 
But in the future, we need to solve where do we want to put CPU info considering single and multi asic cases.

PR #26738, opened 2026-04-13
[FRR] Enable building and running sharpd

kalash-nexthop:
#### Why I did it
We need `sharpd` support enabled by default in SONiC so that we can easily do manual testing of various frr features, e.g. creating a route with primary and backup nexthop is very straightforward with sharpd. It also allows us to write a route programming benchmarking test that can inject hundreds of thousands of routes into zebra very easily. The overhead of having it run all the time is minimal, but it comes very handy for various testing related use cases.

##### Work item tracking

#### How I did it
Created a patch in sonic-frr to let frr be configured with `--enable-sharpd`. Plus added an entry for sharpd in `dockers/docker-fpm-frr/frr/supervisord/supervisord.conf.j2` so that it gets started as part of bgp container.

#### How to verify it
Manually tested and found that sharpd is built and is running as part of bgp container.

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511

- [ ] 
- [ ]


venkit-nexthop:
@deepak-singhal0408 for viz

src/sonic-frr/patch/series
```diff
@@ -21,3 +21,4 @@
 0064-bgpd-Prevent-unnecessary-re-install-of-routes.patch
 0065-SONiC-ONLY-bgpd-reduce-suppress-fib-advertisement-delay-to-50ms.patch
 0066-SONiC-ONLY-zebra-skip-if-add-update-in-speed-timer-for-unready-ifp.patch
+0070-Enable-sharpd-daemon-for-testing.patch
```
deepak-singhal0408: it will be a SONiC only patch? or are you planning to upstream as well?

deepak-singhal0408:
Hi @dgsudharsan @eddieruan-alibaba , could you help review as well.

src/sonic-frr/patch/series
kalash-nexthop: Thanks for looking @deepak-singhal0408 . This will be a SONiC only patch. I don't plan to enable this flag by default in upstream FRR.

We are using this heavily in our internal SONiC workflows to tune route programming performance, and find regressions etc. I don't think all FRR deployments have the same use cases.

deepak-singhal0408: then rename the patch? 0070-SONiC-ONLY-*****

kalash-nexthop: Fair enough. Done. Thanks again.

dgsudharsan (dismissed):

kalash-nexthop:
Hi @deepak-singhal0408 @dgsudharsan , I had to rebase this branch to retrigger the failing checks. Could you please approve again?

dockers/docker-fpm-frr/frr/supervisord/supervisord.conf.j2
```diff
@@ -263,3 +263,15 @@ stderr_logfile=syslog
...
+
+[program:sharpd]
+command=/usr/lib/frr/sharpd -A 127.0.0.1 -P 0
+priority=5
+stopsignal=KILL
+autostart=false
+autorestart=false
+startsecs=0
+stdout_logfile=syslog
+stderr_logfile=syslog
```
deepak-singhal0408: Please fix `stdout_logfile=syslog` → `stdout_logfile=NONE` + `stdout_syslog=true` (and same for stderr) to match other daemons in this file (zebra, bgpd, staticd, ospfd, etc.).

deepak-singhal0408 (commented):
The t0/t1-lag Elastictest failures are PR-induced. sharpd is starting (despite `autostart=false`) because `dependent_startup=true` + `dependent_startup_wait_for=zebra:running` causes the dependent-startup plugin to launch it once zebra is up. sharpd then logs ERR about VRF interfaces it doesn't recognize, which `analyze_logs` catches:

```
ERR bgp#sharpd[55]: INTERFACE_STATE: Cannot find IF Vrf_01 in VRF 115
```

To keep sharpd truly disabled by default, remove `dependent_startup=true` and `dependent_startup_wait_for=zebra:running`.

dockers/docker-fpm-frr/frr/supervisord/supervisord.conf.j2
kalash-nexthop: I am so sorry about this @deepak-singhal0408 . I misremembered that the CI was passing earlier. We (at nexthop internally) keep sharpd running and ignore these failures in tests, so I totally forgot about these logs. I will take care of these and update the diff.

And thanks for checking this.

deepak-singhal0408 (approved):

dgsudharsan (approved):

kalash-nexthop:
Hi @deepak-singhal0408 . How can we merge this?

deepak-singhal0408:
> Hi @deepak-singhal0408 . How can we merge this?

@eddieruan-alibaba @kperumalbfn @yxieca please help with merge.. thanks,

eddieruan-alibaba:
Should this sharpd be enabled for testing purpose only, instead of default in SONiC? 

In deployment, an extra unused and unmonitored process could be a headache. 

If we want to enable sharpd in deployment, we need to add monitoring to this process to avoid this guy behave bad, for example leaking memory.

eddieruan-alibaba (changes requested):
I am not a big fan to enable this process in production. 

Can we enable it for testing only.

kalash-nexthop:
> I am not a big fan to enable this process in production.
> 
> Can we enable it for testing only.

Hi @eddieruan-alibaba , I understand your concern, and we want to use it mainly for testing. But is there a way in SONiC today to enable some compile and run time knob only for testing? Pardon my ignorance, but do we have test builds in the build system, or it is left entirely up to vendors that use SONiC?

eddieruan-alibaba:
Discussed in WG meeting, since the following two flags are off, sharpd would not be launched in production.

autostart=false
autorestart=false

So it is ok.

eddieruan-alibaba (approved):

kalash-nexthop:
Thanks @eddieruan-alibaba

PR #26756, opened 2026-04-14
[yang] Add sample_rate and truncate_size to sonic-mirror-session

Janetxxx:
Add two new optional leaves to MIRROR_SESSION_LIST in sonic-mirror-session.yang for sampled port mirroring with truncation support:

 - sample_rate: optional, range
   2..max. When not configured, full mirroring is used (no sampling).
  - truncate_size: optional, range
   64..9216. When not configured, no truncation is applied.

#### Why I did it
 Add YANG model support for the sampled port mirroring with truncation feature. This feature allows users to configure a sampling rate and truncation size on ERSPAN mirror sessions, reducing mirror 
bandwidth.

##### Work item tracking

#### How I did it
 Added two new leaf definitions to the `MIRROR_SESSION_LIST` in `sonic-mirror-session.yang`, after the existing `policer` leaf. 

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511

- [ ] 
- [ ] 

 Add `sample_rate` and `truncate_size` leaves to sonic-mirror-session YANG model for sampled port mirroring with truncation support.

### DIFF
--- a/src/sonic-yang-models/tests/yang_model_tests/tests/mirror_session.json
+++ b/src/sonic-yang-models/tests/yang_model_tests/tests/mirror_session.json
@@ -125,6 +125,14 @@
         "desc": "Configuring ERSPAN entry with sample_rate below minimum (100).",
         "eStr": "Sample rate must be 0 (no sampling/full mirroring) or 256-8388608"
     },
+    "MIRROR_ERSPAN_ENTRY_WRONG_SAMPLE_RATE_ONE": {
+        "desc": "Configuring ERSPAN entry with sample_rate=1 (in the invalid 1-255 gap).",
+        "eStr": "Sample rate must be 0 (no sampling/full mirroring) or 256-8388608"
+    },
+    "MIRROR_ERSPAN_ENTRY_WRONG_SAMPLE_RATE_255": {
+        "desc": "Configuring ERSPAN entry with sample_rate=255 (boundary of invalid 1-255 gap).",
+        "eStr": "Sample rate must be 0 (no sampling/full mirroring) or 256-8388608"
+    },
     "MIRROR_ERSPAN_ENTRY_WRONG_SAMPLE_RATE_TOO_HIGH": {
         "desc": "Configuring ERSPAN entry with sample_rate above maximum (8388609).",
         "eStr": "Sample rate must be 0 (no sampling/full mirroring) or 256-8388608"
--- a/src/sonic-yang-models/tests/yang_model_tests/tests/mirror_session.json
+++ b/src/sonic-yang-models/tests/yang_model_tests/tests/mirror_session.json
@@ -123,19 +123,19 @@
     },
     "MIRROR_ERSPAN_ENTRY_WRONG_SAMPLE_RATE_TOO_LOW": {
         "desc": "Configuring ERSPAN entry with sample_rate below minimum (100).",
-        "eStr": "Sample rate must be 0 (disabled) or 256-8388608"
+        "eStr": "Sample rate must be 0 (no sampling/full mirroring) or 256-8388608"
     },
     "MIRROR_ERSPAN_ENTRY_WRONG_SAMPLE_RATE_TOO_HIGH": {
         "desc": "Configuring ERSPAN entry with sample_rate above maximum (8388609).",
-        "eStr": "Sample rate must be 0 (disabled) or 256-8388608"
+        "eStr": "Sample rate must be 0 (no sampling/full mirroring) or 256-8388608"
     },
     "MIRROR_ERSPAN_ENTRY_WRONG_TRUNCATE_SIZE_TOO_LOW": {
         "desc": "Configuring ERSPAN entry with truncate_size below minimum (32).",
-        "eStr": "Truncate size must be 0 (disabled) or 64-9216 bytes"
+        "eStr": "Truncate size must be 0 (no truncation) or 64-9216 bytes"
     },
     "MIRROR_ERSPAN_ENTRY_WRONG_TRUNCATE_SIZE_TOO_HIGH": {
         "desc": "Configuring ERSPAN entry with truncate_size above maximum (9217).",
-        "eStr": "Truncate size must be 0 (disabled) or 64-9216 bytes"
+        "eStr": "Truncate size must be 0 (no truncation) or 64-9216 bytes"
     },
     "MIRROR_ERSPAN_ENTRY_WRONG_SAMPLE_RATE_TYPE": {
         "desc": "Configuring SPAN entry with sample_rate (ERSPAN only).",
--- a/src/sonic-yang-models/tests/yang_model_tests/tests/mirror_session.json
+++ b/src/sonic-yang-models/tests/yang_model_tests/tests/mirror_session.json
@@ -152,5 +152,8 @@
     "MIRROR_ERSPAN_ENTRY_WRONG_TRUNCATE_SIZE_TYPE": {
         "desc": "Configuring SPAN entry with truncate_size (ERSPAN only).",
         "eStrKey": "When"
+    },
+    "MIRROR_ERSPAN_ENTRY_WITH_ZERO_SAMPLE_RATE_AND_TRUNCATE": {
+        "desc": "Configuring ERSPAN entry with sample_rate=0 and truncate_size=0 (disabled/full mirroring)."
     }
 }
--- a/src/sonic-yang-models/tests/yang_model_tests/tests/mirror_session.json
+++ b/src/sonic-yang-models/tests/yang_model_tests/tests/mirror_session.json
@@ -99,5 +99,50 @@
     "MIRROR_ERSPAN_ENTRY_WITH_INVALID_POLICER": {
         "desc": "Configuring SPAN entry with invalid policer",
         "eStrKey": "LeafRef"
+    },
+    "MIRROR_ERSPAN_ENTRY_WITH_VALID_SAMPLE_RATE": {
+        "desc": "Configuring ERSPAN entry with valid sample_rate (50000)."
+    },
+    "MIRROR_ERSPAN_ENTRY_WITH_VALID_SAMPLE_RATE_MIN": {
+        "desc": "Configuring ERSPAN entry with minimum valid sample_rate (256)."
+    },
+    "MIRROR_ERSPAN_ENTRY_WITH_VALID_SAMPLE_RATE_MAX": {
+        "desc": "Configuring ERSPAN entry with maximum valid sample_rate (8388608)."
+    },
+    "MIRROR_ERSPAN_ENTRY_WITH_VALID_TRUNCATE_SIZE": {
+        "desc": "Configuring ERSPAN entry with valid truncate_size (128)."
+    },
+    "MIRROR_ERSPAN_ENTRY_WITH_VALID_TRUNCATE_SIZE_MIN": {
+        "desc": "Configuring ERSPAN entry with minimum valid truncate_size (64)."
+    },
+    "MIRROR_ERSPAN_ENTRY_WITH_VALID_TRUNCATE_SIZE_MAX": {
+        "desc": "Configuring ERSPAN entry with maximum valid truncate_size (9216)."
+    },
+    "MIRROR_ERSPAN_ENTRY_WITH_VALID_SAMPLE_RATE_AND_TRUNCATE": {
+        "desc": "Configuring ERSPAN entry with both sample_rate and truncate_size."
+    },
+    "MIRROR_ERSPAN_ENTRY_WRONG_SAMPLE_RATE_TOO_LOW": {
+        "desc": "Configuring ERSPAN entry with sample_rate below minimum (100).",
+        "eStr": "Sample rate must be 0 (disabled) or 256-8388608"
+    },
+    "MIRROR_ERSPAN_ENTRY_WRONG_SAMPLE_RATE_TOO_HIGH": {
+        "desc": "Configuring ERSPAN entry with sample_rate above maximum (8388609).",
+        "eStr": "Sample rate must be 0 (disabled) or 256-8388608"
+    },
+    "MIRROR_ERSPAN_ENTRY_WRONG_TRUNCATE_SIZE_TOO_LOW": {
+        "desc": "Configuring ERSPAN entry with truncate_size below minimum (32).",
+        "eStr": "Truncate size must be 0 (disabled) or 64-9216 bytes"
+    },
+    "MIRROR_ERSPAN_ENTRY_WRONG_TRUNCATE_SIZE_TOO_HIGH": {
+        "desc": "Configuring ERSPAN entry with truncate_size above maximum (9217).",
+        "eStr": "Truncate size must be 0 (disabled) or 64-9216 bytes"
+    },
+    "MIRROR_ERSPAN_ENTRY_WRONG_SAMPLE_RATE_TYPE": {
+        "desc": "Configuring SPAN entry with sample_rate (ERSPAN only).",
+        "eStrKey": "When"
+    },
+    "MIRROR_ERSPAN_ENTRY_WRONG_TRUNCATE_SIZE_TYPE": {
+        "desc": "Configuring SPAN entry with truncate_size (ERSPAN only).",
+        "eStrKey": "When"
     }
 }
--- a/src/sonic-yang-models/tests/yang_model_tests/tests/mirror_session.json
+++ b/src/sonic-yang-models/tests/yang_model_tests/tests/mirror_session.json
@@ -123,27 +123,19 @@
     },
     "MIRROR_ERSPAN_ENTRY_WRONG_SAMPLE_RATE_TOO_LOW": {
         "desc": "Configuring ERSPAN entry with sample_rate below minimum (100).",
-        "eStr": "Sample rate must be 0 (no sampling/full mirroring) or 256-8388608"
-    },
-    "MIRROR_ERSPAN_ENTRY_WRONG_SAMPLE_RATE_ONE": {
-        "desc": "Configuring ERSPAN entry with sample_rate=1 (in the invalid 1-255 gap).",
-        "eStr": "Sample rate must be 0 (no sampling/full mirroring) or 256-8388608"
-    },
-    "MIRROR_ERSPAN_ENTRY_WRONG_SAMPLE_RATE_255": {
-        "desc": "Configuring ERSPAN entry with sample_rate=255 (boundary of invalid 1-255 gap).",
-        "eStr": "Sample rate must be 0 (no sampling/full mirroring) or 256-8388608"
+        "eStr": "Sample rate must be 256..8388608"
     },
     "MIRROR_ERSPAN_ENTRY_WRONG_SAMPLE_RATE_TOO_HIGH": {
         "desc": "Configuring ERSPAN entry with sample_rate above maximum (8388609).",
-        "eStr": "Sample rate must be 0 (no sampling/full mirroring) or 256-8388608"
+        "eStr": "Sample rate must be 256..8388608"
     },
     "MIRROR_ERSPAN_ENTRY_WRONG_TRUNCATE_SIZE_TOO_LOW": {
         "desc": "Configuring ERSPAN entry with truncate_size below minimum (32).",
-        "eStr": "Truncate size must be 0 (no truncation) or 64-9216 bytes"
+        "eStr": "Truncate size must be 64..9216"
     },
     "MIRROR_ERSPAN_ENTRY_WRONG_TRUNCATE_SIZE_TOO_HIGH": {
         "desc": "Configuring ERSPAN entry with truncate_size above maximum (9217).",
-        "eStr": "Truncate size must be 0 (no truncation) or 64-9216 bytes"
+        "eStr": "Truncate size must be 64..9216"
     },
     "MIRROR_ERSPAN_ENTRY_WRONG_SAMPLE_RATE_TYPE": {
         "desc": "Configuring SPAN entry with sample_rate (ERSPAN only).",
@@ -152,8 +144,5 @@
     "MIRROR_ERSPAN_ENTRY_WRONG_TRUNCATE_SIZE_TYPE": {
         "desc": "Configuring SPAN entry with truncate_size (ERSPAN only).",
         "eStrKey": "When"
-    },
-    "MIRROR_ERSPAN_ENTRY_WITH_ZERO_SAMPLE_RATE_AND_TRUNCATE": {
-        "desc": "Configuring ERSPAN entry with sample_rate=0 and truncate_size=0 (disabled/full mirroring)."
     }
 }
--- a/src/sonic-yang-models/tests/yang_model_tests/tests/mirror_session.json
+++ b/src/sonic-yang-models/tests/yang_model_tests/tests/mirror_session.json
@@ -101,40 +101,40 @@
         "eStrKey": "LeafRef"
     },
     "MIRROR_ERSPAN_ENTRY_WITH_VALID_SAMPLE_RATE": {
-        "desc": "Configuring ERSPAN entry with valid sample_rate (50000)."
+        "desc": "Configuring ERSPAN entry with a valid sample_rate."
     },
     "MIRROR_ERSPAN_ENTRY_WITH_VALID_SAMPLE_RATE_MIN": {
-        "desc": "Configuring ERSPAN entry with minimum valid sample_rate (256)."
+        "desc": "Configuring ERSPAN entry with the minimum valid sample_rate."
     },
     "MIRROR_ERSPAN_ENTRY_WITH_VALID_SAMPLE_RATE_MAX": {
-        "desc": "Configuring ERSPAN entry with maximum valid sample_rate (8388608)."
+        "desc": "Configuring ERSPAN entry with the maximum valid sample_rate."
     },
     "MIRROR_ERSPAN_ENTRY_WITH_VALID_TRUNCATE_SIZE": {
-        "desc": "Configuring ERSPAN entry with valid truncate_size (128)."
+        "desc": "Configuring ERSPAN entry with a valid truncate_size."
     },
     "MIRROR_ERSPAN_ENTRY_WITH_VALID_TRUNCATE_SIZE_MIN": {
-        "desc": "Configuring ERSPAN entry with minimum valid truncate_size (64)."
+        "desc": "Configuring ERSPAN entry with the minimum valid truncate_size."
     },
     "MIRROR_ERSPAN_ENTRY_WITH_VALID_TRUNCATE_SIZE_MAX": {
-        "desc": "Configuring ERSPAN entry with maximum valid truncate_size (9216)."
+        "desc": "Configuring ERSPAN entry with the maximum valid truncate_size."
     },
     "MIRROR_ERSPAN_ENTRY_WITH_VALID_SAMPLE_RATE_AND_TRUNCATE": {
         "desc": "Configuring ERSPAN entry with both sample_rate and truncate_size."
     },
     "MIRROR_ERSPAN_ENTRY_WRONG_SAMPLE_RATE_TOO_LOW": {
-        "desc": "Configuring ERSPAN entry with sample_rate below minimum (100).",
+        "desc": "Configuring ERSPAN entry with a sample_rate below the allowed range.",
         "eStr": "Sample rate must be 256..8388608"
     },
     "MIRROR_ERSPAN_ENTRY_WRONG_SAMPLE_RATE_TOO_HIGH": {
-        "desc": "Configuring ERSPAN entry with sample_rate above maximum (8388609).",
+        "desc": "Configuring ERSPAN entry with a sample_rate above the allowed range.",
         "eStr": "Sample rate must be 256..8388608"
     },
     "MIRROR_ERSPAN_ENTRY_WRONG_TRUNCATE_SIZE_TOO_LOW": {
-        "desc": "Configuring ERSPAN entry with truncate_size below minimum (32).",
+        "desc": "Configuring ERSPAN entry with a truncate_size below the allowed range.",
         "eStr": "Truncate size must be 64..9216"
     },
     "MIRROR_ERSPAN_ENTRY_WRONG_TRUNCATE_SIZE_TOO_HIGH": {
-        "desc": "Configuring ERSPAN entry with truncate_size above maximum (9217).",
+        "desc": "Configuring ERSPAN entry with a truncate_size above the allowed range.",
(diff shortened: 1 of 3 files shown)


src/sonic-yang-models/tests/yang_model_tests/tests_config/mirror_session.json
```diff
@@ -550,5 +550,219 @@
...
+                    {
+                        "name": "erspan",
+                        "type": "ERSPAN",
+                        "dst_ip": "11.1.1.1",
+                        "src_ip": "10.1.1.1",
+                        "gre_type": "0x1234",
+                        "dscp": "10",
+                        "sample_rate": "50000",
+                        "truncate_size": "128"
+                    }
```
banidoru, line 669: The test matrix covers min/mid/max for both valid and invalid cases — nice. However, there are no tests for `sample_rate=0` or `truncate_size=0`, which are the explicitly-valid "disabled" default values per the YANG range `0|256..8388608` and `0|64..9216`. These are important boundary cases to validate — please add positive test configs for both `"sample_rate": "0"` and `"truncate_size": "0"`.

banidoru (commented):
- Good: `when` guards correctly restrict both leaves to ERSPAN type, revision history updated, ranges well-defined with clear error messages
- Good: Comprehensive test matrix covering valid min/mid/max, invalid boundaries, and SPAN-type rejection
- Missing tests for `sample_rate=0` and `truncate_size=0` — the default "disabled" values are valid per the range but have no explicit positive test
- Trailing whitespace on line 192 of the YANG file (blank line after `policer` leaf) — minor lint issue

src/sonic-yang-models/tests/yang_model_tests/tests_config/mirror_session.json
```diff
@@ -550,5 +550,219 @@
...
+                    {
+                        "name": "span",
+                        "type": "SPAN",
+                        "truncate_size": "128"
+                    }
+                ]
+            }
+        }
     }
 }
```
banidoru, line 768: There are no test configs for `sample_rate=0` or `truncate_size=0`. Since 0 is the default and means "disabled", it should be explicitly validated as accepted. Add positive test cases like `MIRROR_ERSPAN_ENTRY_WITH_SAMPLE_RATE_ZERO` and `MIRROR_ERSPAN_ENTRY_WITH_TRUNCATE_SIZE_ZERO`.

src/sonic-yang-models/tests/yang_model_tests/tests/mirror_session.json
```diff
@@ -99,5 +99,50 @@
...
+        "desc": "Configuring ERSPAN entry with minimum valid truncate_size (64)."
+    },
+    "MIRROR_ERSPAN_ENTRY_WITH_VALID_TRUNCATE_SIZE_MAX": {
+        "desc": "Configuring ERSPAN entry with maximum valid truncate_size (9216)."
+    },
+    "MIRROR_ERSPAN_ENTRY_WITH_VALID_SAMPLE_RATE_AND_TRUNCATE": {
+        "desc": "Configuring ERSPAN entry with both sample_rate and truncate_size."
+    },
+    "MIRROR_ERSPAN_ENTRY_WRONG_SAMPLE_RATE_TOO_LOW": {
+        "desc": "Configuring ERSPAN entry with sample_rate below minimum (100).",
```
banidoru, line 125: The invalid `sample_rate` tests use 100 (too low) and 8388609 (too high), but there's no test for values in the 1–255 gap (e.g., `sample_rate=1` or `sample_rate=255`). These boundary values would catch off-by-one errors in the `0|256..8388608` range expression more precisely.

src/sonic-yang-models/yang-models/sonic-mirror-session.yang
```diff
@@ -183,6 +189,35 @@ module sonic-mirror-session {
                     description
                         "Policer to be applied for the mirrored traffic.";
                 }
+                
+                leaf sample_rate {
+                    when "current()/../type = 'ERSPAN'";
+                    type uint32 {
+                        range "0|256..8388608" {
```
banidoru, line 196: The range lower bound of 256 for `sample_rate` presumably aligns with sFlow's `sample_rate` range, but there's no comment or reference here explaining this choice. Consider adding a brief comment (e.g., `/* consistent with sFlow sampling range */`) so future maintainers understand why 1–255 is excluded.

src/sonic-yang-models/yang-models/sonic-mirror-session.yang
```diff
@@ -183,6 +189,35 @@ module sonic-mirror-session {
                     description
```
banidoru, line 189: The `policer` leaf immediately above has no `when` guard on session type, meaning it applies to both SPAN and ERSPAN. The new `sample_rate` and `truncate_size` leaves are correctly gated to ERSPAN only. Is this asymmetry intentional? If policer is also ERSPAN-only in practice, it should get a `when` guard too (separate PR is fine).

banidoru (commented):
- Tests for `sample_rate=0` and `truncate_size=0` are still missing — these are the default/disabled values and should have explicit positive test coverage.
- The `sample_rate` range lower bound of 256 appears to mirror sFlow conventions, but there's no reference or comment in the YANG model linking to the hardware/spec rationale. Future maintainers will wonder why 1–255 is invalid.
- `sample_rate` and `truncate_size` are gated with `when` on ERSPAN, but the `policer` leaf right above them applies to all session types with no guard — worth confirming this asymmetry is intentional.
- No negative test for `sample_rate=1` (in the 1–255 gap) — only 100 and 8388609 are tested. A value of 1 or 255 would exercise the boundary more precisely.

src/sonic-yang-models/yang-models/sonic-mirror-session.yang
Janetxxx: The when guard on sample_rate and truncate_size is because sampling/truncation is only supported on ERSPAN sessions.

banidoru (approved):
All prior feedback addressed. LGTM.

- Revision entry added for schema versioning
- `when` guards correctly gate both leaves to ERSPAN only
- Error messages are clear and semantically accurate
- Test coverage is thorough: boundary values (0, 1, 255, 256, max, max+1), gap values, type-mismatch, and combined configs
- Values properly string-typed to match CONFIG_DB convention

banidoru (approved):
LGTM

src/sonic-yang-models/yang-models/sonic-mirror-session.yang
```diff
@@ -183,6 +189,35 @@ module sonic-mirror-session {
...
                         "Policer to be applied for the mirrored traffic.";
                 }
+
```
banidoru: Trailing whitespace is still present on this blank line. The previous review thread was resolved but the fix wasn't applied. Please remove the trailing spaces.

src/sonic-yang-models/tests/yang_model_tests/tests/mirror_session.json
```diff
@@ -99,5 +99,58 @@
...
+        "eStr": "Truncate size must be 0 (no truncation) or 64-9216 bytes"
+    },
+    "MIRROR_ERSPAN_ENTRY_WRONG_TRUNCATE_SIZE_TOO_HIGH": {
+        "desc": "Configuring ERSPAN entry with truncate_size above maximum (9217).",
+        "eStr": "Truncate size must be 0 (no truncation) or 64-9216 bytes"
+    },
+    "MIRROR_ERSPAN_ENTRY_WRONG_SAMPLE_RATE_TYPE": {
+        "desc": "Configuring SPAN entry with sample_rate (ERSPAN only).",
+        "eStrKey": "When"
+    },
```
banidoru, line 151: Still missing test cases for `sample_rate=0` and `truncate_size=0` (the default/disabled values). These should be added as valid-case tests to confirm the `0|256..8388608` and `0|64..9216` ranges accept 0 correctly.

banidoru (commented):
Re-review (iteration 2):

- All prior review threads resolved; `when` guards, error messages, string types in test configs, revision history, and gap boundary tests (1, 255) all addressed — good work.
- **Trailing whitespace still present** on the blank line between `policer` and `sample_rate` (line 192 in the YANG file). Thread was resolved but the actual whitespace wasn't removed.
- **Missing `sample_rate=0` and `truncate_size=0` test cases.** These are the default/disabled values and should have explicit valid-case tests. Previously flagged, thread resolved, but tests not added.
- YANG model logic is correct — ranges, `when` guards, error-app-tags all look solid.
- Overall: minor nits remaining, no functional issues. LGTM with the above addressed.

banidoru (approved):
All reviewers approved. LGTM.

banidoru (approved):
Iteration 3 re-review — all prior concerns addressed. LGTM.

- Error messages now correctly distinguish sampling vs truncation semantics
- Zero-value test cases added for both `sample_rate=0` and `truncate_size=0`
- Gap boundary tests (1, 255) cover the `0|256..8388608` range precisely
- Trailing whitespace fixed
- Test values properly quoted as strings matching CONFIG_DB convention
- Revision entry added, `when` guards correctly scope leaves to ERSPAN

src/sonic-yang-models/tests/yang_model_tests/tests/mirror_session.json
```diff
@@ -99,5 +99,61 @@
...
+        "desc": "Configuring ERSPAN entry with sample_rate above maximum (8388609).",
+        "eStr": "Sample rate must be 0 (no sampling/full mirroring) or 256-8388608"
+    },
+    "MIRROR_ERSPAN_ENTRY_WRONG_TRUNCATE_SIZE_TOO_LOW": {
+        "desc": "Configuring ERSPAN entry with truncate_size below minimum (32).",
+        "eStr": "Truncate size must be 0 (no truncation) or 64-9216 bytes"
+    },
+    "MIRROR_ERSPAN_ENTRY_WRONG_TRUNCATE_SIZE_TOO_HIGH": {
+        "desc": "Configuring ERSPAN entry with truncate_size above maximum (9217).",
+        "eStr": "Truncate size must be 0 (no truncation) or 64-9216 bytes"
```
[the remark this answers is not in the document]
Janetxxx: Existing test cases in this file (e.g., MIRROR_ERSPAN_ENTRY_WRONG_DST_IP, MIRROR_ERSPAN_ENTRY_WRONG_DSCP) also use string format for eStr. Keeping consistent with the existing pattern.

banidoru (approved):
- YANG model changes look correct: `when` guards, proper ranges, error messages, revision entry all present
- Test coverage is thorough: boundary values (0, 256, 8388608, 64, 9216), gap values (1, 255), out-of-range, wrong-type (SPAN), and combined sample_rate+truncate_size
- All prior review feedback has been addressed
- Note: `eStr` uses string format (not array), but this is consistent with existing test patterns in the file — the test harness behavior is a pre-existing concern, not introduced here

banidoru (approved):
- All prior review feedback addressed: revision entry, `when` guards, error messages, trailing whitespace fixed
- Comprehensive test coverage now includes boundary values (0, 256, 8388608, 64, 9216), gap values (1, 255), and `when`-guard negative tests for SPAN type
- Test configs correctly use string-typed values consistent with CONFIG_DB conventions
- `eStr` uses string format matching existing patterns in the file (dns.json, mirror_session.json) — technically weaker than array format but consistent with codebase convention
- LGTM, no new issues found

banidoru (approved):
Iteration 2 re-review — all prior feedback addressed. LGTM.

- Revision history added ✓
- `when` guards properly gate both leaves to ERSPAN-only ✓
- Error messages are correct and distinct for each leaf ✓
- Test coverage is thorough: valid (min/mid/max), invalid (below/above/gap), `when` enforcement on SPAN, and zero-value defaults ✓
- Values use string format consistent with existing test patterns ✓
- No trailing whitespace issues ✓

banidoru (approved):
All reviewers approved. LGTM.

src/sonic-yang-models/tests/yang_model_tests/tests/mirror_session.json
```diff
@@ -99,5 +99,61 @@
...
+    "MIRROR_ERSPAN_ENTRY_WRONG_SAMPLE_RATE_255": {
+        "desc": "Configuring ERSPAN entry with sample_rate=255 (boundary of invalid 1-255 gap).",
+        "eStr": "Sample rate must be 0 (no sampling/full mirroring) or 256-8388608"
+    },
+    "MIRROR_ERSPAN_ENTRY_WRONG_SAMPLE_RATE_TOO_HIGH": {
+        "desc": "Configuring ERSPAN entry with sample_rate above maximum (8388609).",
+        "eStr": "Sample rate must be 0 (no sampling/full mirroring) or 256-8388608"
+    },
+    "MIRROR_ERSPAN_ENTRY_WRONG_TRUNCATE_SIZE_TOO_LOW": {
+        "desc": "Configuring ERSPAN entry with truncate_size below minimum (32).",
```
r12f: This looks inconsistent with the error message below.

src/sonic-yang-models/tests/yang_model_tests/tests/mirror_session.json
```diff
@@ -99,5 +99,61 @@
...
         "eStrKey": "LeafRef"
+    },
+    "MIRROR_ERSPAN_ENTRY_WITH_VALID_SAMPLE_RATE": {
+        "desc": "Configuring ERSPAN entry with valid sample_rate (50000)."
+    },
+    "MIRROR_ERSPAN_ENTRY_WITH_VALID_SAMPLE_RATE_MIN": {
+        "desc": "Configuring ERSPAN entry with minimum valid sample_rate (256)."
+    },
+    "MIRROR_ERSPAN_ENTRY_WITH_VALID_SAMPLE_RATE_MAX": {
+        "desc": "Configuring ERSPAN entry with maximum valid sample_rate (8388608)."
```
r12f: Do we need a max here?

src/sonic-yang-models/tests/yang_model_tests/tests/mirror_session.json
```diff
@@ -99,5 +99,61 @@
...
+        "desc": "Configuring ERSPAN entry with maximum valid truncate_size (9216)."
+    },
+    "MIRROR_ERSPAN_ENTRY_WITH_VALID_SAMPLE_RATE_AND_TRUNCATE": {
+        "desc": "Configuring ERSPAN entry with both sample_rate and truncate_size."
+    },
+    "MIRROR_ERSPAN_ENTRY_WRONG_SAMPLE_RATE_TOO_LOW": {
+        "desc": "Configuring ERSPAN entry with sample_rate below minimum (100).",
+        "eStr": "Sample rate must be 0 (no sampling/full mirroring) or 256-8388608"
+    },
+    "MIRROR_ERSPAN_ENTRY_WRONG_SAMPLE_RATE_ONE": {
```
r12f: Do we still need check explicitly against 1?

src/sonic-yang-models/tests/yang_model_tests/tests/mirror_session.json
```diff
@@ -99,5 +99,61 @@
...
+        "desc": "Configuring ERSPAN entry with sample_rate=1 (in the invalid 1-255 gap).",
+        "eStr": "Sample rate must be 0 (no sampling/full mirroring) or 256-8388608"
+    },
+    "MIRROR_ERSPAN_ENTRY_WRONG_SAMPLE_RATE_255": {
```
r12f: This looks like min check?

src/sonic-yang-models/tests/yang_model_tests/tests/mirror_session.json
```diff
@@ -99,5 +99,61 @@
...
+        "desc": "Configuring ERSPAN entry with sample_rate=255 (boundary of invalid 1-255 gap).",
+        "eStr": "Sample rate must be 0 (no sampling/full mirroring) or 256-8388608"
+    },
+    "MIRROR_ERSPAN_ENTRY_WRONG_SAMPLE_RATE_TOO_HIGH": {
```
r12f: Is this dup to max check?

src/sonic-yang-models/tests/yang_model_tests/tests/mirror_session.json
Janetxxx: The upper bound 8388608 was aligned with sFlow's sample_rate range in [sonic-sflow.yang](https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-yang-models/yang-models/sonic-sflow.yang#L128). However, since sampled port mirroring is hardware-offloaded, there's no technical reason to cap the max. Happy to remove the upper bound if you prefer, I can change the range to `256..max` which is equal to `256..4294967295`

src/sonic-yang-models/tests/yang_model_tests/tests/mirror_session.json
banidoru: Addressed in latest commit — the range was simplified to `256..8388608` without the 0 value, and the eStr values now consistently use `"Sample rate must be 256..8388608"` format matching the YANG error-message.

src/sonic-yang-models/tests/yang_model_tests/tests/mirror_session.json
banidoru: With the range simplified to `256..8388608` (0 removed), testing against 1 and 255 is still useful as boundary tests for the gap below the valid range. However, the existing TOO_LOW test (100) already covers the invalid-below-min case. These are redundant but harmless — deferring to PR author on whether to keep them for thoroughness or trim.

src/sonic-yang-models/tests/yang_model_tests/tests/mirror_session.json
banidoru: Yes, sample_rate=255 is essentially testing the boundary just below the valid minimum (256). It's a valid negative boundary test but overlaps with the TOO_LOW (100) test in purpose.

src/sonic-yang-models/tests/yang_model_tests/tests/mirror_session.json
banidoru: WRONG_SAMPLE_RATE_TOO_HIGH tests 8388609 (max+1), which is indeed the upper boundary negative test. It's not a duplicate of the valid max test (8388608) — it's the invalid counterpart. But it does overlap with 255 in being 'just outside range' tests.

banidoru (approved):
- YANG model is clean: `when` guards, ranges, error-messages, revision entry all look correct
- Test coverage is thorough: valid min/mid/max, invalid boundaries, `when`-guard type tests
- Values properly string-typed in test configs, consistent with existing patterns
- Previous review concerns (trailing whitespace, error-message wording, missing 0-value tests, revision entry) all addressed
- Minor: the sample_rate=1 and sample_rate=255 negative tests overlap with TOO_LOW (100) — harmless but could be trimmed

src/sonic-yang-models/tests/yang_model_tests/tests/mirror_session.json
banidoru: Confirmed — the eStr values now consistently use `"Sample rate must be 256..8388608"` and `"Truncate size must be 64..9216"` matching the YANG error-messages. Fixed.

src/sonic-yang-models/tests/yang_model_tests/tests/mirror_session.json
banidoru: Agree these are redundant with the TOO_LOW (100) test. The author could consolidate — keeping just TOO_LOW (100) and adding a boundary test at 255 would be sufficient. Up to the author whether to trim.

src/sonic-yang-models/tests/yang_model_tests/tests/mirror_session.json
banidoru: Yes, 255 is the boundary just below valid min (256). Useful as a precise boundary test but overlaps with TOO_LOW. Could be kept or trimmed per author preference.

src/sonic-yang-models/tests/yang_model_tests/tests/mirror_session.json
banidoru: Not a duplicate — TOO_HIGH (8388609) is the invalid counterpart to the valid MAX (8388608) test. This one should stay as it tests the exact upper boundary.

banidoru (approved):
- All prior review concerns addressed: `when` guards, revision entry, error messages, string-typed test values, boundary tests all look correct
- Range simplification to `256..8388608` and `64..9216` (removing 0) is cleaner — absence-means-disabled is the right semantic
- Error messages now consistent between YANG model and test expectations
- Test coverage is thorough: valid min/mid/max, invalid boundary, wrong-type (SPAN) cases all present
- Minor: consider trimming redundant negative tests (1, 255 overlap with TOO_LOW) per maintainer feedback, but not blocking

src/sonic-yang-models/tests/yang_model_tests/tests/mirror_session.json
banidoru: Verified — all eStr values now consistently use `"Sample rate must be 256..8388608"` matching the YANG error-message. The inconsistency is resolved.

src/sonic-yang-models/tests/yang_model_tests/tests/mirror_session.json
banidoru: Confirmed — WRONG_SAMPLE_RATE_TOO_HIGH (8388609) tests the invalid upper boundary, distinct from the valid MAX (8388608) test. Not a duplicate.

banidoru (approved):
- YANG model additions are correct: `sample_rate` and `truncate_size` with proper `when` guards restricting to ERSPAN only
- Range constraints and error messages are consistent across YANG model and test expectations
- Test coverage is thorough: valid min/mid/max, invalid boundaries, wrong session type — all covered
- Revision history updated appropriately
- Previous review concerns (inconsistent eStr, dup test questions) have been addressed

banidoru (approved):
All reviewers approved. LGTM.

banidoru (approved):
LGTM. Clean iteration:
- Range simplified to `256..8388608` / `64..9216` — removes the `0` special case and `default 0`, cleaner YANG modeling
- Descriptions correctly updated to "when not configured" semantics
- Error messages and test eStr values now consistent
- Redundant boundary tests trimmed per maintainer feedback
- All prior review threads addressed and resolved

banidoru (approved):
- All prior review threads addressed and resolved
- YANG model clean: `when` guards, revision entry, range constraints, error-messages all correct
- Test coverage solid: valid boundary (min/mid/max), invalid boundary (too-low/too-high), `when` guard rejection for SPAN type
- String-typed values in test configs consistent with existing patterns
- No new issues found in this iteration

banidoru (approved):
All reviewers approved. LGTM.

tirupatihemanth (dismissed):

src/sonic-yang-models/tests/yang_model_tests/tests/mirror_session.json
```diff
@@ -99,5 +99,50 @@
...
+        "desc": "Configuring ERSPAN entry with sample_rate below minimum (100).",
+        "eStr": "Sample rate must be 256..8388608"
+    },
+    "MIRROR_ERSPAN_ENTRY_WRONG_SAMPLE_RATE_TOO_HIGH": {
+        "desc": "Configuring ERSPAN entry with sample_rate above maximum (8388609).",
+        "eStr": "Sample rate must be 256..8388608"
+    },
+    "MIRROR_ERSPAN_ENTRY_WRONG_TRUNCATE_SIZE_TOO_LOW": {
+        "desc": "Configuring ERSPAN entry with truncate_size below minimum (32).",
+        "eStr": "Truncate size must be 64..9216"
```
r12f, line 127: Why 64 here not 256?

src/sonic-yang-models/tests/yang_model_tests/tests/mirror_session.json
```diff
@@ -99,5 +99,50 @@
...
+    },
+    "MIRROR_ERSPAN_ENTRY_WITH_VALID_TRUNCATE_SIZE_MAX": {
+        "desc": "Configuring ERSPAN entry with maximum valid truncate_size (9216)."
+    },
+    "MIRROR_ERSPAN_ENTRY_WITH_VALID_SAMPLE_RATE_AND_TRUNCATE": {
+        "desc": "Configuring ERSPAN entry with both sample_rate and truncate_size."
+    },
+    "MIRROR_ERSPAN_ENTRY_WRONG_SAMPLE_RATE_TOO_LOW": {
+        "desc": "Configuring ERSPAN entry with sample_rate below minimum (100).",
+        "eStr": "Sample rate must be 256..8388608"
```
r12f: Why 256 here not 100?

Janetxxx: I went with `256..8388608` to follow the existing sFlow YANG convention:
https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-yang-models/yang-models/sonic-sflow.yang#L125-L132
Both ERSPAN sampling and sFlow rely on the same underlying SAI sampling mechanism, so they share the same range.

src/sonic-yang-models/tests/yang_model_tests/tests/mirror_session.json
Janetxxx: This is based on protocol-level limits. I’ll sync with you offline :)

src/sonic-yang-models/yang-models/sonic-mirror-session.yang
```diff
@@ -183,6 +189,34 @@ module sonic-mirror-session {
...
+                leaf sample_rate {
+                    when "current()/../type = 'ERSPAN'";
+                    type uint32 {
+                        range "256..8388608" {
+                            error-message "Sample rate must be 256..8388608";
```
r12f: we should confirm this number. I don't think it is blocking, but my instinct is telling me we should not use this number.

r12f (dismissed):
signed off with a minor comment. non-blocking, but need to confirm and come back.

src/sonic-yang-models/yang-models/sonic-mirror-session.yang
```diff
@@ -183,6 +189,34 @@ module sonic-mirror-session {
                     description
                         "Policer to be applied for the mirrored traffic.";
                 }
+
+                leaf sample_rate {
+                    when "current()/../type = 'ERSPAN'";
+                    type uint32 {
+                        range "256..8388608" {
```
kperumalbfn: @Janetxxx Is this PPS?

kperumalbfn:
@Janetxxx could you check if we have sonic-mgmt tests for truncation?

kperumalbfn:
@Janetxxx pls resolve all the comments

src/sonic-yang-models/yang-models/sonic-mirror-session.yang
Janetxxx: Have sent email for confirming SAI numbers, will come back to this.

src/sonic-yang-models/yang-models/sonic-mirror-session.yang
r12f: yes.

vaibhavhd (dismissed):
LGTM. Please be careful to not enable these optional knobs to limit packets being mirrored as a default SONiC behavior. In warm-reboot cases, we need all traffic to be mirrored to destination server, serving ARP/ND responses.

src/sonic-yang-models/yang-models/sonic-mirror-session.yang
```diff
@@ -183,6 +189,34 @@ module sonic-mirror-session {
                     description
                         "Policer to be applied for the mirrored traffic.";
                 }
+
+                leaf sample_rate {
+                    when "current()/../type = 'ERSPAN'";
+                    type uint32 {
```
dgsudharsan, line 195: A minor enhancement. Can we define a new type called sample-rate in sonic-types.yang and use it? Given we use it for both sflow and now here, we need to have a common type which can be updated at a single place if required (Generally gets updated when we add higher port speeds)

dgsudharsan (dismissed):

src/sonic-yang-models/yang-models/sonic-mirror-session.yang
r12f: Hi Sud, may I know the reason?

dgsudharsan:
@Janetxxx It would be fine if you can address my comment in a separate PR later as well.

src/sonic-yang-models/yang-models/sonic-mirror-session.yang
dgsudharsan: @r12f We have sample rate here as well https://github.com/sonic-net/sonic-buildimage/blob/fab4a1875dcbd652c82033af5472de363f6d3925/src/sonic-yang-models/yang-models/sonic-sflow.yang#L125
In such scenarios when we have a common type used across yang models we define it in a custom type here
https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-yang-models/yang-templates/sonic-types.yang.j2

In this way when we need to update the range in the future it will be just in one place.

src/sonic-yang-models/yang-models/sonic-mirror-session.yang
r12f: sorry, to correct - it is packets, not pps.

src/sonic-yang-models/yang-models/sonic-mirror-session.yang
r12f: got it. I think we should also change the range to >1 instead of having this. It doesn't make sense to limit max, because more powerful ASIC can come out and exceed this number (which already have).

Janetxxx:
> @Janetxxx could you check if we have sonic-mgmt tests for truncation?

@kperumalbfn Yes I prepared a sonic-mgmt PR covering truncation size https://github.com/sonic-net/sonic-mgmt/pull/24776

src/sonic-yang-models/yang-models/sonic-mirror-session.yang
Janetxxx: @dgsudharsan In this PR I changed the mirror sample_rate range to 2..max, while sflow still uses 256..8388608. Since the two ranges differ, adding a shared typedef in sonic-types.yang wouldn't actually be shared, so I'll skip it for now.

Janetxxx:
@dgsudharsan Can you help re-review this PR? Thanks

kperumalbfn (approved):

dgsudharsan (approved):

r12f (approved):

PR #26764, opened 2026-04-14
[smartswitch] Start pmon after databasedpu instance services on SmartSwitch host (Fixes #26760)

### LINKED ISSUE #26760: Bug:[Smartswitch] databasedpuX containers fail to start after image install due to Docker startup race with bridge-midplane readiness
### Is it platform specific

generic

### Importance or Severity

Critical

### Description of the bug

After installing a new SONiC 202511 image over an existing 202511 image using sonic-installer install, the SmartSwitch DPU database containers (databasedpu0..7) do not start automatically on first boot.

Observed behavior:

On first boot after image install:
database@dpu0..7.service remain inactive initially
databasedpu0..7 containers are not present

Manually starting:

systemctl start database@dpu0.service

(and similarly for other DPUs) brings them up successfully.

Subsequent reload / reboot works fine once containers have started once.

Root cause:
Docker starts too early during first boot before SmartSwitch bridge-midplane network readiness is fully complete.

During first boot:

docker.service starts before:
network-online.target
systemd-networkd-wait-online.service
systemd-networkd-wait-online@bridge-midplane.service

This causes transient Docker startup failures related to IPv6 / bridge initialization:

failed to create NAT chain DOCKER
failed to create DOCKER-USER ipv6 chain
failed to create default bridge network

Since database@dpuX.service depends on Docker, DPU DB containers never start.

Evidence:
Initial failing boot showed Docker startup failure:

ip6tables ... Could not fetch rule set generation id: Invalid argument
failed to create NAT chain DOCKER

After adding systemd Docker ordering + retry override:

Docker starts cleanly during boot
all databasedpu0..7 containers come up automatically within ~2 minutes of boot

Commit a7f4bfa96d58e367415d8169c965613775e7c0d6 enabled Docker IPv6 support (--ipv6=true --fixed-cidr-v6=fd00::/80). On SmartSwitch first boot after image install, Docker may start before network / bridge-midplane / ip6tables readiness is complete, causing Docker bridge network initialization failure and preventing database@dpuX startup. The fix is to delay Docker until network readiness and increase restart tolerance, rather than reverting IPv6 support.

PR https://github.com/sonic-net/sonic-buildimage/pull/26764

### Steps to Reproduce

Install a new SONiC 202511 image on a SmartSwitch system that already has an existing 202511 image:

sonic-installer install <new-sonic-image>
reboot

After first boot into the new image, check Docker and DPU database container status:

systemctl status docker.service --no-pager
systemctl status 'database@dpu*' --no-pager
docker ps -a | grep databasedpu
Observe that during early boot:
docker.service may fail transiently with IPv6 / bridge initialization errors
database@dpu0..7.service remain inactive
databasedpu0..7 containers are not started automatically

Manually start DPU DB services:

for i in {0..7}; do systemctl start database@dpu${i}.service; done

Verify that all DPU DB containers now start successfully:

docker ps -a | grep databasedpu

Reboot again:

reboot

Verify that after the manual recovery, subsequent reboots typically work fine:

docker ps -a | grep databasedpu

This reproduces the first-boot startup race where Docker starts before bridge-midplane network readiness and prevents automatic DPU DB container startup

Actual Behavior

After sonic-installer install of a new SONiC 202511 image on a SmartSwitch and rebooting into the new image, docker.service starts too early during first boot and can fail transiently during Docker bridge / IPv6 initialization. As a result, database@dpu0..7.service remain inactive initially and databasedpu0..7 containers do not come up automatically. Later in the boot, Docker may recover and the rest of the system eventually comes up, or the DPU DB services can be started manually.

Expected Behavior

After sonic-installer install and the first reboot into the new image, docker.service should wait until the SmartSwitch bridge-midplane network is ready, start successfully on its own, and all database@dpu0..7.service instances should automatically start without manual intervention. The corresponding databasedpu0..7 containers should be present and running as part of normal first-boot initialization.

### Relevant log output

```shell
failed to start daemon: Error initializing network controller: error creating default "bridge" network:
Failed to Setup IP tables: Unable to enable ACCEPT OUTGOING rule:
iptables failed: ip6tables --wait -t filter -A DOCKER-FORWARD -i docker0 -j ACCEPT:
ip6tables v1.8.11 (nf_tables): Could not fetch rule set generation id: Invalid argument
```

```shell
SONiC Software Version: SONiC.202511.<build_version>
SONiC OS Version: 13
Distribution: Debian 13.1 (trixie)
Kernel: 6.12.x
Build commit: <commit_sha>
Build date: <build_date>
Platform: x86_64-8102_28fh_dpu_o-r0
HwSKU: Cisco-8102-28FH-DPU-O
ASIC: Cisco-8000
ASIC Count: 1
Platform subtype: SmartSwitch
Serial Number: <serial_number>
Model Number: Cisco 8102-28FH-DPU-O
```

### Attach files (if any)

_No response_

rameshraghupathy:
This PR fixes a SmartSwitch boot ordering issue where pmon may start before database@dpuX.service instances are brought up, leading to missing or delayed DPU database availability during system initialization.

The fix introduces a SmartSwitch-specific systemd ordering constraint generated at boot time. It ensures that pmon.service is started only after all database@dpuX.service instances are started, where the number of DPU instances is derived dynamically from platform configuration.

This change is implemented in the systemd generator and applies only to SmartSwitch NPU systems.

Validated on SmartSwitch:

fresh image install and reboot into new image
all database@dpu0..N.service instances start automatically
pmon starts only after DPU database services are initialized
no manual recovery (e.g., systemctl start database@dpuX.service) is required

Fixes #26760

#### Why I did it
On SmartSwitch systems, pmon can start before database@dpuX.service instances are initialized. This results in missing or incomplete DPU state during early boot and can require manual recovery to start DPU database services.

The goal was to ensure that:

all DPU database services are started during boot
pmon consumes them only after they are ready
the solution is platform-agnostic (based on number of DPUs)
the solution does not introduce runtime coupling between DPU failures and NPU operation

##### Work item tracking

#### How I did it
A SmartSwitch-specific systemd generator update was introduced:

For SmartSwitch NPU systems, a drop-in file is generated:

pmon.service.d/dpu-database-ordering.conf

This file adds:

Wants=database@dpuX.service
After=database@dpuX.service

for all DPU instances (X derived from num_dpus)

This ensures:

database@dpuX.service instances are started as part of the pmon start transaction
pmon starts only after those services
no hardcoded DPU count (fully platform-agnostic)
no runtime systemctl calls from the generator

#### How to verify it
Install a new SONiC image on a SmartSwitch system:

sonic-installer install <image>
reboot
After boot:

Verify all DPU database services:

systemctl status database@dpu*

Verify containers:

docker ps | grep databasedpu

Verify pmon ordering:

systemctl status pmon
systemctl cat pmon

Expected result:

all database@dpuX.service instances are active
all databasedpuX containers are running
pmon starts after DPU DB services
no manual intervention required

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ x] 202511

Fix SmartSwitch boot ordering to ensure DPU database services (database@dpuX) are started before pmon, enabling reliable automatic bring-up of DPU database containers after image install.


arlakshm:
@prsunny @vvolam, can you review this PR?

vvolam:
@rameshraghupathy Is this a regression or behavior change?

rameshraghupathy:
> @rameshraghupathy Is this a regression or behavior change?

@vvolam This is a regression fix, not an intended behavior change. On SmartSwitch lit-mode, databasedpuX instances are expected to be started before PMON because PMON depends on DPU DB availability when DPUs are present. Previously they were effectively coupled with the main database startup path. After the recent decoupling work, host services/containers no longer bounce when databasedpuX bounces, but that also means we need an explicit boot-time mechanism to start/order database@dpuX. This PR adds that missing startup ordering by making PMON Wants= and After= all database@dpuX.service instances. So the intent is to restore the expected SmartSwitch lit-mode boot behavior without reintroducing runtime coupling.

KrisNey-MSFT:
@vvolam we discussed this today in our call; could we approve and move forward to get it merged pls?

vvolam:
@rameshraghupathy could you resolve build failures?

@saiarcot895 @judyjoseph could you please review and make sure it does not impact other SKUs?

vvolam (commented):
Overall the logic is correct and the systemd semantics (Wants + After) are appropriate. A few suggestions below.

src/systemd-sonic-generator/systemd-sonic-generator.cpp
```diff
@@ -982,20 +982,31 @@ static int render_network_service_for_smart_switch(const std::filesystem::path&
...
 
         auto unit_ordering_file_path = unit_override_dir / "ordering.conf";
 
-        std::ofstream unit_ordering_file;
-        unit_ordering_file.open(unit_ordering_file_path);
+        std::ofstream unit_ordering_file(unit_ordering_file_path);
         unit_ordering_file << "[Unit]\n";
         unit_ordering_file << "Requires=systemd-networkd-wait-online@bridge-midplane.service\n";
         unit_ordering_file << "After=systemd-networkd-wait-online@bridge-midplane.service\n";
     }
```
vvolam, line 1005: `create_directory` can throw if the path already exists as a file or on permission errors. Consider using the `std::error_code` overload to handle this gracefully, e.g.:

```cpp
std::error_code ec;
std::filesystem::create_directory(pmon_override_dir, ec);
if (ec) {
    // log and return error
    return -1;
}
```

The existing loop above has the same pattern, so this could be addressed holistically.

src/systemd-sonic-generator/systemd-sonic-generator.cpp
```diff
@@ -982,20 +982,31 @@ static int render_network_service_for_smart_switch(const std::filesystem::path&
         return 0;
     }
 
-    // Render Before instruction for midplane network with database service
+    // Render database@dpuX ordering with midplane network
     for (int i = 0; i < num_dpus; i++) {
         auto unit_override_dir = install_dir / std::format("database@dpu{}.service.d", i);
         std::filesystem::create_directory(unit_override_dir);
 
         auto unit_ordering_file_path = unit_override_dir / "ordering.conf";
```
vvolam, line 999: Nit: Refactoring the existing `ofstream` from `.open()` to constructor-based init is a nice cleanup, but mixing it with the new feature change makes the diff harder to review. Consider separating this into its own commit.

rameshraghupathy: @vvolam I will update the patch to use the std::error_code overload for create_directory() and handle failures consistently for both the existing database@dpuX.service.d override directory and the new pmon.service.d override directory.

src/systemd-sonic-generator/systemd-sonic-generator.cpp
rameshraghupathy: @vvolam I will also keep the unrelated ofstream cleanup out of this feature change so the diff stays focused on the SmartSwitch pmon/database ordering fix.

rameshraghupathy (commented):
@vvolam Thanks, updated the PR.

saiarcot895 (approved):

vvolam (approved):
LGTM

vvolam:
@kperumalbfn could you help merge?

PR #26833, opened 2026-04-16
[cisco-8000]: Update platform ref to 202511.1.0.7

wsycqyz:
#### Why I did it
Update Cisco 8000 platform checkout ref from 202511.1.0.3 to 202511.1.0.7.

#### How I did it
Updated ref in platform/checkout/cisco-8000.ini.

#### How to verify it
Build with Cisco 8000 platform and verify the correct platform submodule is checked out.


StormLiangMS (approved):
LGTM

PR #26838, opened 2026-04-16
Set rsyslogd autorestart to unexpected

MavChtz:
#### Why I did it
Applications within containers rely on rsyslogd to forward logs to the host's syslog. If rsyslogd dies and is not automatically restarted, subsequent errors go unnoticed.

#### How I did it
Use `autorestart=unexpected` on rsyslogd entries in supervisord configurations. When set, the process is restarted if it exits with an exit code not listed in exitcodes (see [supervisord documentation](https://supervisord.org/configuration.html#program-x-section-values)).

#### How to verify it
Before the change (e.g., on the LLDP container):

1. Note the rsyslogd PID: 
`sudo docker exec -it lldp bash -c 'pidof rsyslogd'`
2. Kill the process with -9: 
`sudo docker exec -it lldp bash -c 'kill -9 $(pidof rsyslogd)'`
3. Check the rsyslogd PID again: 
`sudo docker exec -it lldp bash -c 'pidof rsyslogd'`
rsyslogd does not come back.

After the change, following the same steps results in rsyslogd being restarted automatically (observable, for example, on the eventd container today).

- [x] 202511

Ensure rsyslogd is restarted if it exits unexpectedly so that events continue to be logged. 

Richard the cat


nimicho:
@lguohan ,I am proposing PR https://github.com/sonic-net/sonic-buildimage/pull/28217 to replace this one due to owner change. I will be responsible to tackle any comments or issues from now on.

PR #26865, opened 2026-04-17
[build] Add timeout for apt-get build hook to prevent long hangs

yijingyan2:
#### Why I did it
Some apt-get operations can hang for hours and result in build failure due to timeout.
This change adds timeout and retries for apt-get commands, so builds fail fast and can be retried again instead of hanging for hours.
##### Work item tracking

#### How I did it
Apply timeout wrapper in apt-get hook
#### How to verify it
tested with debug messages `apt-config dump`

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511

- [ ] 
- [ ]


rules/config
```diff
@@ -406,3 +406,6 @@ BUILD_SKIP_TEST ?= n
 # target/ccache/ and persists across builds.
 # Default: disabled (no impact on clean builds without a warm cache)
 SONIC_CONFIG_USE_CCACHE ?= n
+
+# apt-get timeout for HTTP and HTTPS connections, in seconds
+BUILD_APT_GET_HTTP_TIMEOUT ?= 60
```
liushilongbuaa: We already have such features. Please confirm if these files work and combine them.
files/apt/apt.conf.d/apt-multiple-retries
files/apt/apt-retries-count.j2
Copilot recommends config. We need validate if it works or not.
```
Acquire::http::Timeout "20";
Acquire::https::Timeout "20";
Acquire::ftp::Timeout "20";
Acquire::Retries "3";
Acquire::http::Dl-Limit "10";
Acquire::https::Dl-Limit "10";
```

dockers/docker-base-bookworm/apt-timeout-n-retries
```diff
@@ -0,0 +1,7 @@
+# Instruct apt-get to timeout after 20 seconds and retry 3 times before giving up. 
+# This can help mitigate issues with slow or unreliable network connections when using apt-get in Docker builds.
+
+Acquire::http::Timeout "20";
+Acquire::https::Timeout "20";
+Acquire::ftp::Timeout "20";
+Acquire::Retries "3";
```
liushilongbuaa: It's better to check references to files/apt/apt.conf.d/*, and use this folder as global settings for apt.

liushilongbuaa (approved):
LGTM

PR #26144, opened 2026-03-12
[lldp] Fix transient MAC address Port ID during LLDP daemon startup

ZhaohuiS:
#### Why I did it

When lldpd starts (or restarts), it sends the first LLDP frames with **MAC addresses as Port IDs** instead of the configured interface aliases (e.g., `Ethernet1/1`). This is because:

1. lldpd starts in paused state and loads its config file (`/etc/lldpd.conf`)
2. The config file only configures the management port (`eth0`) portidsubtype — no front-panel port configs exist
3. After processing all config lines, lldpd internally **auto-resumes** (hardcoded behavior in lldpd's internal lldpcli)
4. The first LLDP frames are sent with default MAC-based Port IDs
5. lldpmgrd starts 2-3 seconds later and reconfigures each port with the correct alias via `lldpcli`
6. This triggers an **MSAP change** (shutdown frame + new frame with correct Port ID) on every port

Peers see transient neighbor flapping: a MAC-based entry appears briefly, then gets replaced by the correct interface name entry. This can trigger monitoring alerts and confuse network management systems (e.g., LLDP-based topology discovery, automated cabling validation).

Related issues:
- Fixes #1488
- Fixes #1457

##### Work item tracking
- Microsoft ADO **37084792**:

#### How I did it

Added port ID subtype configuration for **all front-panel ports** directly in the `lldpd.conf.j2` Jinja2 template. The template iterates over the `PORT` table from ConfigDB and generates `configure ports <name> lldp portidsubtype local <alias>` lines for every port that has an alias defined.

These configuration lines are processed by lldpd during startup config loading, **before** the internal auto-resume fires. This ensures the very first LLDP frame already carries the correct interface alias as Port ID, eliminating the transient MAC-based Port ID window entirely.

The change is additive — lldpmgrd continues to handle dynamic port configuration changes at runtime. When lldpmgrd later processes the same ports, the portidsubtype is already correct, so no MSAP change occurs (only the port description gets added, which is expected).

**Key technical findings from investigation:**
- lldpd starts paused by default (the `pause` directive in config is actually redundant)
- lldpd's internal lldpcli auto-resumes after ALL config file lines are processed — this cannot be prevented via config
- The `resume` call in `waitfor_lldp_ready.sh` (from PR #5493) fires after auto-resume, so it's also redundant
- The `PORT` variable is available in sonic-cfggen template context from ConfigDB's PORT table

#### How to verify it

[...]

PR #26873, opened 2026-04-17
[lldp] Replace per-port portidsubtype loop with global ifname directive

ZhaohuiS:
### Description of PR

Summary:
Replace the per-port portidsubtype configuration loop from PR #26144 with a single global `configure lldp portidsubtype ifname` directive.

Fixes https://github.com/sonic-net/sonic-buildimage/issues/26568

#### Background: Why portidsubtype matters

Without any portidsubtype configuration, lldpd uses MAC address as the default Port ID. When lldpd auto-resumes after config processing, the first LLDP frame carries MAC-based Port IDs. Peers then see a transient MSAP change when `lldpmgrd` later reconfigures portidsubtype to `local` (alias), causing unnecessary neighbor flaps.

PR #26144 solved this by adding a Jinja2 loop generating per-port `configure ports <name> lldp portidsubtype local <alias>` lines in `lldpd.conf`. This worked fine on low-port-count platforms.

#### The problem on high-port-count platforms

On platforms with many ports (e.g., Mellanox SN5640 with 512 ports), PR #26144 generates 512+ config lines. Processing these lines blocks lldpd for **11+ seconds** at startup. During this blackout:
- Neighbors with 30-second TTL begin timing out
- When lldpd resumes, it sends frames with new MSAPs, causing neighbor table churn
- The churn cascades through lldp_syncd, creating a feedback loop that never stabilizes
- Result: LLDP neighbor count oscillates indefinitely (observed 161→166→163→171 after 4+ minutes)

#### The fix: global `portidsubtype ifname`

The single global directive `configure lldp portidsubtype ifname` tells lldpd to use the Linux interface name (e.g., `Ethernet0`) as Port ID for **all** ports, instead of MAC address. This achieves the same goal as the per-port loop (no MAC-as-Port-ID) with O(1) config processing time.

`lldpmgrd` still runs later to set the final `portidsubtype` to `local` + alias (e.g., `etp1a`) per port — this is the existing designed behavior and works correctly.

**Key trade-off:** The first LLDP frame after restart uses the interface name (`Ethernet0`) instead of the alias (`etp9a`). When lldpmgrd starts and reconfigures, there is a one-time MSAP transition per port. This is acceptable because:
1. LLDP is an informational protocol — no data plane impact
2. The transition is atomic and completes within seconds
3. The alternative (batch per-port config) causes the same 512-line processing blackout

### Type of change

- [x] Bug fix

### Back port request
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511

On high-port-count platforms (SN5640, 512 ports), the per-port portidsubtype loop from PR #26144 blocks lldpd startup for 11+ seconds, causing LLDP neighbor churn and cascading instability that never converges.

#### How did you do it?
Replaced the per-port Jinja2 loop (N lines) with a single global directive (1 line):
```
configure lldp portidsubtype ifname
```

A/B tested on Mellanox SN5640 (str5-sn5640-6, topology t0-isolated-d256u256s2, 259 VM neighbors):

| Configuration | Config Lines | Convergence | Stable at 259? | Deletes |
|--------------|-------------|-------------|----------------|---------|
| PR #26144 (per-port loop) | 519 | never stable | ❌ | 74+ ongoing |
| No PR #26144 (baseline) | 5 | 14s | ✅ | 0 |
| **This PR (global ifname)** | **6** | **~30s** | **✅** | **0** |
| Batch per-port in waitfor_lldp_ready.sh | 6 + 512 batch | never stable | ❌ | oscillating |

The global ifname approach was tested across multiple restart cycles with zero neighbor drops.

#### Any platform specific information?
Primarily affects high-port-count platforms (SN5640 with 512 ports). Lower port count platforms were not affected by the original issue but benefit from the simpler config.

#### Supported testbed topology if it's a new test case?
N/A (not a new test case)

### Documentation
N/A

### DIFF
--- a/dockers/docker-lldp/lldpd.conf.j2
+++ b/dockers/docker-lldp/lldpd.conf.j2
@@ -27,24 +27,15 @@ configure system ip management pattern {{ mgmt_if.ipv6 }}
 {% endif %}
 {% endif %}
 configure system hostname {{ DEVICE_METADATA['localhost']['hostname'] }}
-{# Configure port ID subtype for front-panel ports at startup.
-   Without this, lldpd uses MAC address as default Port ID. When lldpd
-   auto-resumes after config processing, the first LLDP frame would carry
-   MAC-based Port IDs. Peers see a transient MSAP change when lldpmgrd
-   later reconfigures portidsubtype, causing unnecessary neighbor flaps.
-   Special ports (inband/recirculation/backplane) are excluded to match
-   lldpmgrd behavior which skips these port types. #}
-{% if PORT %}
-{% for port_name in PORT|sort %}
-{% if not port_name.startswith('Ethernet-IB') and not port_name.startswith('Ethernet-Rec') and not port_name.startswith('Ethernet-BP') %}
-{% set port_alias = PORT[port_name].alias|default('') %}
-{% if port_alias %}
-configure ports {{ port_name }} lldp portidsubtype local {{ port_alias }}
-{% else %}
-configure ports {{ port_name }} lldp portidsubtype local {{ port_name }}
-{% endif %}
-{% endif %}
-{% endfor %}
-{% endif %}
+{# Set global portidsubtype to ifname so the first LLDP frame after startup
+   uses the Linux interface name (e.g. Ethernet0) instead of MAC address.
+   This avoids the per-port loop from PR #26144 which added N config lines
+   (one per front-panel port). On high-port-count platforms (e.g. 512 ports),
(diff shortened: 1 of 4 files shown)


ZhaohuiS:
close it for retrigger PR test.

ZhaohuiS:
open it for retrigger PR test.

ZhaohuiS:
@StormLiangMS @r12f  could you pls help review?

StormLiangMS (approved):
LGTM

PR #26892, opened 2026-04-19
slave.mk: archive patched source tree for static analysis

xq9mend:
#### Why I did it

Downstream static analysis jobs need access to the patched source tree (after `quilt push` has applied patches, before `quilt pop` removes them). Without this, a separate source preparation step is required, duplicating build work.

##### Work item tracking
- Microsoft ADO:

#### How I did it

Added `ARCHIVE_PATCHED_SOURCE` macro to `slave.mk` that rsyncs each package's working directory to `target/sonic-buildimage-source/` after build, before quilt pop. Controlled by `ENABLE_SOURCE_ARCHIVE ?= y`. Excludes compiled artifacts (`.o .a .so`, Rust/Go/Python build output). Added `configure` and `distclean` targets for the archive dir.

#### How to verify it

Build with `ENABLE_SOURCE_ARCHIVE=y` (default). Verify `target/sonic-buildimage-source/` is populated with patched source trees after the build. Verify `.o`/`.so` files are excluded.

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511

- [ ] master

Archive patched source trees to target/sonic-buildimage-source/ during build for downstream static analysis.

N/A


slave.mk
```diff
@@ -51,6 +51,8 @@ TRIXIE_PHONY_PATH = $(TARGET_PATH)/phony/trixie
 
 DBG_IMAGE_MARK = dbg
 DBG_SRC_ARCHIVE_FILE = $(TARGET_PATH)/sonic_src.tar.gz
+SOURCE_ARCHIVE_PATH = $(TARGET_PATH)/sonic-buildimage-source
+ENABLE_SOURCE_ARCHIVE ?= y
```
saiarcot895: Should it default to yes? This might increase disk space significantly on regular developer builds.

slave.mk
```diff
@@ -860,6 +869,49 @@ SONIC_TARGET_LIST += $(addprefix $(PHONY_PATH)/, $(SONIC_PHONIES))
...
+			--exclude='debian/debhelper-build-stamp' \
+			--exclude='debian/files' \
+			--exclude='debian/*.substvars' \
+			--exclude='debian/*.log' \
+			--exclude='debian/*.debhelper' --exclude='debian/*.debhelper.log' \
+			--filter='exclude debian/*/DEBIAN/' \
+			--filter='exclude debian/*/usr/' \
+			--filter='exclude debian/*/lib/' \
+			--filter='exclude debian/*/etc/' \
+			--filter='exclude debian/*/var/' \
```
saiarcot895, line 905: Why the change to use `--filter` here?

slave.mk
```diff
@@ -889,6 +941,11 @@ $(addprefix $(DEBS_PATH)/, $(SONIC_MAKE_DEBS)) : $(DEBS_PATH)/% : .platform $$(a
 
 	fi
 
+	# Archive patched source for static analysis (always, even on cache hit)
+	if [ -f $($*_SRC_PATH).patch/series ]; then pushd $($*_SRC_PATH) && ( quilt pop -a -f 1>/dev/null 2>&1 || true ) && QUILT_PATCHES=../$(notdir $($*_SRC_PATH)).patch quilt push -a; popd; fi $(LOG)
+	$(call ARCHIVE_PATCHED_SOURCE,$*)
+	if [ -f $($*_SRC_PATH).patch/series ]; then pushd $($*_SRC_PATH) && quilt pop -a -f; [ -d .pc ] && rm -rf .pc; popd; fi $(LOG)
```
saiarcot895: Instead of applying and removing the patch just for this, maybe rework the function so that the source will be archived, then patches will be removed.

slave.mk
saiarcot895: Please also add a `rules/config` entry for this.

slave.mk
xq9mend: `--exclude` only matches against the filename component, not the full path. To exclude a glob like `debian/*/DEBIAN/`, which needs to match on a path with wildcards in an intermediate directory, `--filter` with an `exclude` rule is required. `--exclude` would need to be applied to `DEBIAN/` alone, which would be too broad.

slave.mk
xq9mend: Good point. Changed default to `n`. CI pipelines (azure-pipelines-build.yml and build-template.yml) now explicitly pass `ENABLE_SOURCE_ARCHIVE=y`, so developer builds are unaffected.

slave.mk
xq9mend: Done. Moved `ARCHIVE_PATCHED_SOURCE` inside the cache-miss guard, before the quilt pop, so patches are still applied when archiving. Removed the outside-guard push/archive/pop blocks entirely.

slave.mk
xq9mend: Added.

saiarcot895 (approved):

PR #26893, opened 2026-04-19
gnmi: use --insecure mode instead of --noTLS when no certs configured

xq9mend:
#### Why I did it

When no TLS certificates are configured, gNMI was started with `--noTLS` which runs entirely cleartext (no encryption, no TLS handshake). This unnecessarily weakens transport security since `--insecure` mode is available and provides TLS encryption using a self-signed test certificate.

`--insecure` uses `testcert.NewCert()` to generate a self-signed certificate, so gRPC connections are TLS-encrypted. Remote clients can connect without verifying the certificate (they skip cert validation), but the transport itself is no longer cleartext.

##### Work item tracking
- Microsoft ADO:

#### How I did it

1. Replace `--noTLS` with `--insecure` in the `else` branch (no certs configured) of `gnmi-native.sh` and `telemetry.sh`.
2. Bump `src/sonic-gnmi` to commit `c6f6e71` (sonic-net/sonic-gnmi#709) which adds the optional `--bind_address` flag.

#### How to verify it

On a SONiC device with no gNMI certs configured:
- gNMI service starts without `--noTLS` flag
- gNMI clients can still connect using InsecureSkipVerify
- Wireshark/tcpdump on port 8080 shows TLS handshake (not plaintext)

- [ ] 202411
- [ ] 202505
- [ ] 202605

`gnmi`: use `--insecure` (TLS with self-signed cert) instead of `--noTLS` (cleartext) when no certificates are configured

N/A

### DIFF
--- a/dockers/docker-sonic-gnmi/gnmi-native.sh
+++ b/dockers/docker-sonic-gnmi/gnmi-native.sh
@@ -57,7 +57,7 @@ elif [ -n "$X509" ]; then
         TELEMETRY_ARGS+=" --ca_crt $CA_CRT"
     fi
 else
-    TELEMETRY_ARGS+=" --noTLS"
+    TELEMETRY_ARGS+=" --noTLS --bind_address 127.0.0.1"
 fi
 
 # If no configuration entry exists for TELEMETRY, create one default port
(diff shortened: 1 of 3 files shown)


hdwhdw (dismissed):
Approved but probably need to update the sonic-gnmi submodule before PR pass.

hdwhdw (dismissed):

PR #26912, opened 2026-04-21
chore: pin ptf_nn_agent.py to use nnpy

auspham:
#### Why I did it

With this commit here, PTF replace nnpy with pynng [p4lang/ptf](https://github.com/p4lang/ptf/commit/ebc00f94fe8a3763cf90339d7c69be7976cf0767) 

However, we don't have pynng. To unblock us now, I'll pin ptf_nn_agent.py to a previous version that doesn't require pynng

##### Work item tracking

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511

- [ ] 
- [ ]

### DIFF
--- a/dockers/docker-ptf/Dockerfile.j2
+++ b/dockers/docker-ptf/Dockerfile.j2
@@ -253,7 +253,7 @@ RUN rm -rf /debs \
 {% endif %}
     && mkdir -p /opt       \
     && cd /opt             \
-    && wget https://raw.githubusercontent.com/p4lang/ptf/master/ptf_nn/ptf_nn_agent.py
+    && wget https://raw.githubusercontent.com/p4lang/ptf/23ebe7237f3c284032bda02fbd1f4a98f1bc12f4/ptf_nn/ptf_nn_agent.py
 
 {% if PTF_ENV_PY_VER == "py3" %}
 RUN curl -L -o tacacs.tar.gz https://shrubbery.net/pub/tac_plus/tacacs-F4.0.4.31.tar.gz\


dockers/docker-ptf/Dockerfile.j2
```diff
@@ -253,7 +253,7 @@ RUN rm -rf /debs \
 {% endif %}
     && mkdir -p /opt       \
     && cd /opt             \
-    && wget https://raw.githubusercontent.com/p4lang/ptf/master/ptf_nn/ptf_nn_agent.py
+    && wget https://raw.githubusercontent.com/p4lang/ptf/23ebe7237f3c284032bda02fbd1f4a98f1bc12f4/ptf_nn/ptf_nn_agent.py
```
ZhaohuiS, line 256: This pin works as a short-term unblock, but it freezes ptf_nn_agent at a 2024 snapshot and misses any future bug fixes from upstream. The forward fix would be to replace `nnpy` with `pynng` in this Dockerfile (and drop the nanomsg-from-source build ~20 lines above), then point back at `master`. There is already a similar migration in sonic-mgmt PR #23719 for the copp helper path. Is there a tracking issue to do the full pynng migration in docker-ptf so this pin doesn't become permanent?

auspham: Hi @ZhaohuiS. No there is no tracking issue. I didn't upgrade to pynng as i'm not sure if it will introduce any new failures or vulnerability security

To avoid additional work, I pin down the ptf_nn_agent. I agree that we should have an issue and assign to relevant owner to track this.

opcoder0, line 256: Yeah installing pynng may be better. I didn't see a failure in the latest docker-ptf image. Is this failing CI / PR checkers already ?

auspham: Bug filed for track: https://github.com/sonic-net/sonic-buildimage/issues/26925

auspham: It's failing on PR check yes

xwjiang-ms (approved):
Approve for temporary unblock PR test

PR #26922, opened 2026-04-21
installer: fix NVMe partition umount and partprobe retry in create_de…

DavidZagury:
#### Why I did it
SONiC installation failed on NVMe devices when reinstalling in the same ONIE session due to two bugs in create_demo_gpt_partition in installer/default_platform.conf:

1. The partition delete loop used $blk_dev$part_index (e.g. /dev/nvme0n13) instead of ${blk_dev}${blk_suffix}$part_index (e.g. /dev/nvme0n1p3). The device /dev/nvme0n13 does not exist, so df exits non-zero, the && never fires, and umount is silently skipped on every NVMe partition - leaving it referenced and causing partprobe to fail.
2. After rebooting from SONiC into ONIE, udev or blkid may transiently hold a SONiC partition open during ONIE boot-time scanning, causing partprobe to fail with EBUSY. The hard exit on failure aborted the install immediately.

##### Work item tracking

#### How I did it

- Fixed the device path in the partition delete loop and cur_part skip check to use ${blk_dev}${blk_suffix}$part_index, consistent with the rest of the script. blk_suffix is "p" for NVMe and empty for all other block devices, so non-NVMe behavior is unchanged.
- Replaced the single partprobe call with a retry loop (up to 10 attempts, 1-second sleep between retries) so the installer waits for transient partition references to be released before giving up.

#### How to verify it

- NVMe: Boot into ONIE on an NVMe-based device, install SONiC, reboot back into ONIE, and reinstall SONiC in the same session. Verify the installation completes successfully without umount or partprobe errors.
- Non-NVMe (SATA/SSD): Perform the same reinstall flow on a non-NVMe device and confirm no regression in partition deletion or partprobe behavior.

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511

- [ ] 
- [ ]

### DIFF
--- a/installer/default_platform.conf
+++ b/installer/default_platform.conf
@@ -111,6 +111,8 @@ legacy_volume_label="ACS-OS"
 create_demo_gpt_partition()
 {
     blk_dev="$1"
+    blk_suffix=
+    echo $blk_dev | grep -q nvme0 && blk_suffix="p"
 
     # Create a temp fifo and store string in variable
     tmpfifo=$(mktemp -u)
@@ -147,16 +149,18 @@ create_demo_gpt_partition()
             # that event and return EBUSY.  udevadm settle would be ideal but is
             # not available in ONIE, so retry until udev finishes.
             partprobe_ok=0
-            for i in 1 2 3 4 5 6 7 8 9 10; do
-                if partprobe 2>/dev/null; then
+            i=1
+            while [ $i -le 10 ]; do
+                if partprobe_err=$(partprobe 2>&1); then
                     partprobe_ok=1
                     break
                 fi
-                echo "partprobe attempt $i failed, retrying in 1s..."
+                echo "partprobe attempt $i failed: $partprobe_err (retrying in 1s...)"
                 sleep 1
+                i=$((i + 1))
             done
             [ $partprobe_ok -eq 1 ] || {
-                echo "Error: Unable to partprobe"
+                echo "Error: partprobe failed after 10 attempts, last error: $partprobe_err"
                 exit 1
             }
         done < $tmpfifo
--- a/installer/default_platform.conf
+++ b/installer/default_platform.conf
@@ -125,12 +125,16 @@ create_demo_gpt_partition()
         # Note: You can use any character as a separator for sed, not just '/'
         echo "$demo_part" > $tmpfifo &
         while read -r part_index; do
-            if [ "$blk_dev$part_index" = "$cur_part" ]; then continue; fi
+            if [ "${blk_dev}${blk_suffix}$part_index" = "$cur_part" ]; then continue; fi
             echo "deleting partition $part_index ..."
             # if the partition is already mounted, umount first
-            df $blk_dev$part_index 2>/dev/null && {
-                umount $blk_dev$part_index || {
-                    echo "Error: Unable to umount $blk_dev$part_index"
+            # Note: NVMe partitions require the 'p' suffix (e.g. /dev/nvme0n1p3),
+            # so use ${blk_dev}${blk_suffix}$part_index instead of $blk_dev$part_index.
+            # ONIE does not mount the SONiC partition normally, but if a previous
+            # install was interrupted this ensures umount is not silently skipped.
+            df ${blk_dev}${blk_suffix}$part_index 2>/dev/null && {
+                umount ${blk_dev}${blk_suffix}$part_index || {
+                    echo "Error: Unable to umount ${blk_dev}${blk_suffix}$part_index"
                     exit 1
                 }
             }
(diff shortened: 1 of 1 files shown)


nazariig (dismissed):

installer/default_platform.conf
```diff
@@ -125,20 +125,37 @@ create_demo_gpt_partition()
...
                 exit 1
             }
-            partprobe || {
+            # Retry partprobe for up to 10 seconds.  sgdisk -d triggers a udev
+            # event; on fast NVMe devices partprobe may race with udev processing
+            # that event and return EBUSY.  udevadm settle would be ideal but is
+            # not available in ONIE, so retry until udev finishes.
+            partprobe_ok=0
+            for i in 1 2 3 4 5 6 7 8 9 10; do
+                if partprobe 2>/dev/null; then
```
liat-grozovik: partprobe 2>/dev/null discards all error output during retries.
should you review what are the info returned and maybe use it in the log itself for better debugability?

installer/default_platform.conf
```diff
@@ -125,20 +125,37 @@ create_demo_gpt_partition()
...
+                    partprobe_ok=1
+                    break
+                fi
+                echo "partprobe attempt $i failed, retrying in 1s..."
+                sleep 1
+            done
+            [ $partprobe_ok -eq 1 ] || {
```
liat-grozovik, line 162: is this the case when all attempts fails? if so, is it correct to assume this error message as not able to do at all?

installer/default_platform.conf
```diff
@@ -125,20 +125,37 @@ create_demo_gpt_partition()
         # Note: You can use any character as a separator for sed, not just '/'
         echo "$demo_part" > $tmpfifo &
         while read -r part_index; do
-            if [ "$blk_dev$part_index" = "$cur_part" ]; then continue; fi
+            if [ "${blk_dev}${blk_suffix}$part_index" = "$cur_part" ]; then continue; fi
```
liat-grozovik, line 130: should u check if it is set before the function is called? i assume if not u will use the broken path

installer/default_platform.conf
```diff
@@ -125,20 +125,37 @@ create_demo_gpt_partition()
...
                 echo "Error: Unable to delete partition $part_index on $blk_dev"
                 exit 1
             }
-            partprobe || {
+            # Retry partprobe for up to 10 seconds.  sgdisk -d triggers a udev
+            # event; on fast NVMe devices partprobe may race with udev processing
+            # that event and return EBUSY.  udevadm settle would be ideal but is
+            # not available in ONIE, so retry until udev finishes.
+            partprobe_ok=0
+            for i in 1 2 3 4 5 6 7 8 9 10; do
```
liat-grozovik: should u consider this way of loop for better future updates?

  i=0
  while [ $i -lt 10 ]; do
      ...
      i=$((i + 1))
  done

liat-grozovik (changes requested):

installer/default_platform.conf
DavidZagury: Yes, this branch is hit only after all 10 attempts have failed. 
Updated the message to include both the retry count and the last captured error from partprobe, so it's clear we retry and what the kernel kept reporting.                                                                                                                        
After that many attempt I don't expect the issue will be solved without actively  understanding and fixing the issue

installer/default_platform.conf
DavidZagury: blk_suffix should always be set at this point because we first run through create_partition, but this is a good practice to have this set on this function and not rely on other calls.

installer/default_platform.conf
DavidZagury: agreed, silently dropping partprobe's stderr can make debugging harder. Updated to capture the error output

liat-grozovik (dismissed):

liat-grozovik:
@nazariig as pre you had comments and req for changes, can you please review and approve it from your end so i can merge it?

nazariig (dismissed):

nazariig:
> @nazariig as pre you had comments and req for changes, can you please review and approve it from your end so i can merge it?

@liat-grozovik done

installer/default_platform.conf
```diff
@@ -111,6 +111,8 @@ legacy_volume_label="ACS-OS"
 create_demo_gpt_partition()
 {
     blk_dev="$1"
+    blk_suffix=
+    echo $blk_dev | grep -q nvme0 && blk_suffix="p"
```
qiluo-msft: 0 is a magic number. how about one platform with multiple NVMe controllers where ONIE lives on a non-zero one.

vmittal-msft:
@DavidZagury is this must fix for 202511 ?

DavidZagury:
> @DavidZagury is this must fix for 202511 ?

@vmittal-msft  The issue is in 202511, but it is a rare issue. Still better to have it I believe

Sourabh-Kumar7:
@qiluo-msft could you please help re-review and approve this change?

qiluo-msft (approved):

PR #26977, opened 2026-04-24
Limit CPU and memory usage for SONiC containers

sharmavijay-ms:
#### Why I did it
We observed that some control‑plane containers can consume excessive CPU or memory under load, which may impact other critical SONiC services and overall system stability.
We need to restrict the memory consumption for telemetry. Instead of doing it from monit, OR no-resource limit we are implementing limits.

##### Work item tracking
- Microsoft ADO **37673592**:

#### How I did it
I added Docker runtime resource limits directly to the docker create command.
--memory=600M
Enforces a hard memory limit equivalent to MemoryMax=600M.
--memory-swap=600M
Prevents the container from using swap beyond its memory limit (no silent spillover).
--cpus=0.6
Limits the container to 60% of a single CPU, functionally equivalent to CPUQuota=600m.

#### How to verify it
We checked memory consumption in our environment.
Also tested by manually changing (lower values for easy and fast repro) the configs and hitting limit.

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511

- [ ] 
- [ ] 

puppy


zbud-msft (approved):
lgtm

prabhataravind:
@judyjoseph for viz.

PR #26990, opened 2026-04-25
Fan support for multiple duty cycles in one word

caleb-nexthop:
#### Why I did it
On some PDDF platforms, multiple fans share a single PWM/duty-cycle register word, with each fan controlled by a bitfield in that word. The existing multifpgapci fan handlers only apply the mask without shifting based on the mask’s bit position, which leads to incorrect values when the fan’s bits are not at bit 0. This causes `show platform fan` to report incorrect or zero speeds, and control operations can affect the wrong bits. I fixed the mask/shift handling so that fan PWM/duty-cycle values are read and written correctly for these platforms.

##### Work item tracking
- https://github.com/sonic-net/sonic-buildimage/issues/26989 

#### How I did it
- Updated the PDDF multifpgapci fan PWM and duty-cycle get/set handlers so that when a mask is present, the code:
  - Masks the raw register value and then right-shifts based on the number of trailing zeros in the mask when reading, so the fan’s field is brought down to bit 0 before use.
  - Left-shifts the value based on the same mask when writing, then re-applies the mask before updating the register, so only the intended bitfield is modified.
- Introduced small control flags in the handlers so that negative-value checks and subsequent masking are skipped in the few paths where the value has already been normalized and validated.
- Kept the behavior for other device types unchanged and scoped the new logic specifically to the multifpgapci case so platforms that already rely on single-byte/word registers without bitfield packing continue to work as before.

#### How to verify it

1. Use a platform where the PDDF multifpgapci fan driver is configured so that multiple fans’ PWM/duty-cycle values are packed into a single register word with non-overlapping masks.
2. Boot an image that contains this change.
3. Run `show platform fan` and confirm that:
   - Fan speeds / duty cycles are non-zero and consistent with hardware behavior.
   - All fans that share the register report distinct, sensible values.
4. Run `sudo pddf_fanutil getspeed ` and read the value.
5. Set fan speed using `sudo pddf_fanutil setspeed 80`
6. Run `sudo pddf_fanutil getspeed ` and read the value, confirming it has changed accordingly.
   - The reported PWM/duty-cycle matches the requested setting for each affected fan.
7. Repeat the test with another system with duty cycle values on separate register words and confirm functionality is not regressed by this change.

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511

- [x] SONiC Software Version: SONiC.master-26990.1107932-ae3647de4
Tested on NH-4010 system to ensure change did not regress existing fan functionality

Tested change on NH-4210 with a separate branch including the changes from https://github.com/sonic-net/sonic-buildimage/pull/27154
To test functionality on a system with multiple duty cycles in one register word.
Confirm change resolved issue of only the first duty cycle being written and read by following the above test plan.

Fix PDDF multifpgapci fan PWM/duty-cycle mask handling when multiple fans share one register word.

N/A – no config_db or YANG model changes.

### DIFF
--- a/platform/pddf/i2c/modules/fan/driver/pddf_fan_api.c
+++ b/platform/pddf/i2c/modules/fan/driver/pddf_fan_api.c
@@ -687,9 +687,7 @@ int sonic_i2c_get_fan_rpm_default(void *client, FAN_DATA_ATTR *udata, void *info
             goto ret;
 
         if (udata->mask != 0) {
-            val &= udata->mask;
-            int bit_shift = count_trailing_zeros(udata->mask);
-            val >>= bit_shift;
+            val = (int)(((uint32_t)val & udata->mask) >> count_trailing_zeros(udata->mask));
         }
         skip_neg_check = true;
     }
@@ -702,7 +700,6 @@ int sonic_i2c_get_fan_rpm_default(void *client, FAN_DATA_ATTR *udata, void *info
         else if (udata->len ==2)
         {
             val = i2c_smbus_read_word_swapped((struct i2c_client *)client, udata->offset);
-            
         }
     }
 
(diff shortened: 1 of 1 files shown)


prabhataravind:
@caleb-nexthop please move the PR to ready once your testing is complete.

caleb-nexthop:
Will transition out of draft once I have gathered test data. Tested on another branch and need to test again with a build based on sonic-net:master

caleb-nexthop:
Testing has been completed.

geans-pin (approved):

Praveen-Brcm (approved):

PR #27008, opened 2026-04-27
[gnmi]: RELP+UDP failover rsyslog config for gnmi/telemetry container family

hdwhdw:
#### Why I did it

Upstream commit f991251ab4 (#18113) switched container→host syslog from UDP/514 to RELP/2514 in `docker-base-trixie`. K8s-deployed gnmi images must keep working on older hosts that still only listen on UDP/514, while using RELP on newer hosts.

##### Work item tracking

#### How I did it

Override `/etc/rsyslog.conf` in `docker-gnmi-sidecar`, `docker-sonic-gnmi`, and `docker-gnmi-watchdog` with a RELP-first / UDP-fallback config. The override is a minimal diff against the current `docker-base-trixie/etc/rsyslog.conf` template — one modified line plus the new fallback action and an explanatory comment:

```
module(load="omrelp")
*.* action(type="omrelp" target=... port="2514"
           action.resumeRetryCount="0" action.resumeInterval="30"
           Template="ForwardFormatInContainer")
*.* action(type="omfwd"  target=... port="514" protocol="udp"
           action.execOnlyWhenPreviousIsSuspended="on"
           Template="ForwardFormatInContainer")
```

`docker-sonic-gnmi` is the base image for `docker-sonic-telemetry` and `docker-telemetry-watchdog`, so those two images also pick up the new RELP→UDP failover behavior via inheritance. This is intentional — they are the same gnmi/telemetry family of containers and have the same compatibility requirement against UDP-only host images.

Two non-obvious requirements (validated on rsyslog 8.2504.0):

1. `resumeRetryCount="0"` — `-1` retries forever and prevents fallback; `>0` only delays it.
2. No per-action queue on the omrelp action — a queue defers send to a worker thread, so `execOnlyWhenPreviousIsSuspended` on the next action sees stale state and UDP fallback never fires.

The `COPY` runs before the existing syslogtag `sed` so per-container tagging is preserved.

#### How to verify it

Tested on rsyslog 8.2504.0 with a receiver running both `imrelp` and `imudp`, toggling RELP availability via iptables: cold start with RELP down, mid-stream RELP failure, and recovery all transitioned within one `resumeInterval` with no loss and no duplicates. The same behavior applies to `docker-sonic-telemetry` and `docker-telemetry-watchdog` since they inherit the override from `docker-sonic-gnmi`.

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511

- [x] master

[gnmi]: RELP→UDP failover in `/etc/rsyslog.conf` for the gnmi/telemetry container family (`docker-sonic-gnmi`, `docker-gnmi-sidecar`, `docker-gnmi-watchdog`, and the `docker-sonic-telemetry` / `docker-telemetry-watchdog` images that inherit from `docker-sonic-gnmi`), to stay compatible with UDP-only host images.

N/A


dockers/docker-gnmi-sidecar/etc/rsyslog.conf
```diff
@@ -0,0 +1,106 @@
...
+#### MODULES ####
+#################
+
+$ModLoad imuxsock # provides support for local system logging
+module(load="omrelp")
+
+#
+# Set a rate limit on messages from the container
+#
+$SystemLogRateLimitInterval 300
```
saiarcot895: Please use the new syntax for these. The current templates should be using it.

dockers/docker-gnmi-sidecar/etc/rsyslog.conf
```diff
@@ -0,0 +1,103 @@
...
+#
+module(load="imuxsock" SysSock.RateLimit.Interval="300" SysSock.RateLimit.Burst="20000") # provides support for local system logging
+module(load="omrelp")
+
+#$ModLoad imklog  # provides kernel logging support
+#$ModLoad immark  # provides --MARK-- message capability
+
+# provides UDP syslog reception
+#$ModLoad imudp
+#$UDPServerRun 514
```
saiarcot895: For the full file, can you take the current rsyslog template, and modify that instead? If I were to do a diff, there are going to be more lines changed than just the two changes at the top.

hdwhdw:
Addressed review feedback in 295112407:

- @saiarcot895 — rewrote `etc/rsyslog.conf` for `docker-gnmi-sidecar`, `docker-gnmi-watchdog`, and `docker-sonic-gnmi` as a minimal diff against the current `docker-base-trixie/etc/rsyslog.conf` template instead of a full-file replacement. The diff is now one modified action line + the new omfwd fallback action + an explanatory comment. New-style `module()` / `action()` syntax throughout.
- @Copilot — confirmed intent: `docker-sonic-telemetry` and `docker-telemetry-watchdog` inherit `/etc/rsyslog.conf` from `docker-sonic-gnmi` and should keep the same RELP→UDP failover behavior. Updated the PR title and description to call this out.

cc @zbud-msft — FYI, this changes syslog forwarding behavior in `docker-sonic-telemetry` and `docker-telemetry-watchdog` (they inherit the override from `docker-sonic-gnmi`). Please flag if that's not what you want.

hdwhdw:
Follow-up after offline thinking on @saiarcot895's "use the current template" comment.

The canonical container rsyslog config in the tree is `files/image_config/rsyslog/rsyslog-container.conf.j2`, used in two places:

* `files/build_templates/docker_image_ctl.j2` renders it on the host and `docker cp`s the result into the container's `/etc/rsyslog.conf` on multi-ASIC platforms (`NUM_ASIC -gt 1`), overwriting whatever was baked into the image.
* `src/sonic-containercfgd/containercfgd/containercfgd.py` re-renders the same template inside the container whenever the `SYSLOG_CONFIG_FEATURE` table changes.

The static `/etc/rsyslog.conf` baked into `docker-base-trixie` (and inherited by these images) only wins on:

* K8s-deployed `docker-gnmi-sidecar` / `docker-gnmi-watchdog` (no `docker_image_ctl.j2` involvement, no `containercfgd` in their supervisord configs).
* Single-ASIC native `docker-sonic-gnmi` / `docker-sonic-telemetry` / `docker-telemetry-watchdog` (`updateSyslogConf()` is gated on multi-ASIC).

So if we only patched `rsyslog-container.conf.j2`, the K8s + single-ASIC native paths would silently revert to RELP-only and stop working against UDP-only hosts. If we only patched the static files, the multi-ASIC native path would silently revert to RELP-only when `docker_image_ctl.j2` does its `docker cp`. Both paths need to be covered, hence belt-and-suspenders.

61c7f099 adds the failover into `rsyslog-container.conf.j2`, scoped to the three gnmi-family container names (`gnmi`, `telemetry`, `telemetry_watchdog`) so it does not change behavior for any other SONiC container. Verified by rendering the template with each of those names and with `swss`/`bgp` for the negative case.

295112407 (earlier in the day) shrinks the static-override files to be a minimal diff against `dockers/docker-base-trixie/etc/rsyslog.conf`: one modified action line, one new fallback action, plus an explanatory comment.

PTAL.

hdwhdw:
After thinking more about this, eab0a876 reverts 61c7f099 \u2014 the `rsyslog-container.conf.j2` change is not actually needed.

The j2 template only renders into `/etc/rsyslog.conf` in two situations:

1. `docker_image_ctl.j2:updateSyslogConf()` runs it on the host and `docker cp`s the result into the container, but only when `NUM_ASIC > 1`.
2. `sonic-containercfgd` re-renders it inside the container when `SYSLOG_CONFIG_FEATURE` changes \u2014 and `docker-sonic-gnmi/supervisord.conf` does not start `containercfgd`.

So the j2 path applies only to the multi-ASIC native deployment of `docker-sonic-gnmi` / `docker-sonic-telemetry` / `docker-telemetry-watchdog`. In that deployment shape the host and the container are built from the same buildimage, so the host's `imrelp` listener on 2514 is always present and the UDP fallback can never trigger. Patching the j2 template would just add dead code to a shared file.

The deployment shapes that actually need the failover are:

* K8s-deployed `docker-gnmi-sidecar` and `docker-gnmi-watchdog` running against older host images (the original motivation for this PR). These do not go through `docker_image_ctl.j2` and do not run `containercfgd`, so they pick up `/etc/rsyslog.conf` from the image at build time.
* Single-ASIC native `docker-sonic-gnmi` / `docker-sonic-telemetry` / `docker-telemetry-watchdog` (`updateSyslogConf` is gated on multi-ASIC, so the build-time file wins).

Both of these are covered by the per-container static override in commit 295112407, which is now a minimal diff against `dockers/docker-base-trixie/etc/rsyslog.conf` (one modified action line + one new fallback action + an explanatory comment), as @saiarcot895 asked.

Net state of the PR on top of `655da167` is therefore:

* `docker-{gnmi-sidecar, gnmi-watchdog, sonic-gnmi}/etc/rsyslog.conf` \u2014 small diff vs the base template, RELP-first with UDP fallback.
* Corresponding Dockerfile.j2 changes to `COPY` it over the inherited base config.
* No changes to shared infra (j2 template, base images, other containers).

hdwhdw:
Further simplification in 21260db7.

Replaced the three duplicated `etc/rsyslog.conf` files (which were each a ~80-line copy of `docker-base-trixie/etc/rsyslog.conf` with two small edits) with one reusable Jinja macro `apply_relp_udp_failover` in `dockers/dockerfile-macros.j2`. The macro emits a `RUN sed` against the inherited `/etc/rsyslog.conf` that rewrites the existing `omrelp` action in place and appends the `omfwd` UDP fallback action.

```
{% from "dockers/dockerfile-macros.j2" import apply_relp_udp_failover %}
...
{{ apply_relp_udp_failover() }}
```

Properties:

* Diff against the base template is now structural rather than textual \u2014 the transform stays correct as long as the base template keeps a single line matching `^\\*\\.\\* action\(type="omrelp"`. If the base template gains other config (rate limits, formats, includes, etc.), this image inherits those changes automatically.
* Single-quoting the sed expression keeps the backtick subshell literal so rsyslog (not the docker-build shell) expands `$SYSLOG_TARGET_IP` at container runtime. Verified end-to-end through `sh -c` against the actual `docker-base-trixie/etc/rsyslog.conf`.
* Future containers that need the same compatibility shim can opt in with one import line + one macro call. No `/etc/rsyslog.conf` file under `dockers/<image>/etc/` required.

Total PR diff is now 4 files, +28/-2 lines.

PTAL.

hdwhdw:
Superseded by #28169, which replaces the build-time `apply_relp_udp_failover` sed macro with a single owned config file (`files/image_config/rsyslog/rsyslog-container-gnmi.conf`) applied to the gNMI/telemetry container family via the standard `_FILES` + `COPY` mechanism. Closing in favor of that approach.

PR #27041, opened 2026-04-28
Fix docker-start for DooD with native dockerd enabled

sertem-ciena:
When SONIC_CONFIG_USE_NATIVE_DOCKERD_FOR_BUILD=y, skip docker service management and only test docker responsiveness. In DooD, the docker daemon is owned by the host and cannot be restarted from the container.

#### Why I did it
Build fails in Docker-out-of-Docker (DooD) when SONIC_CONFIG_USE_NATIVE_DOCKERD_FOR_BUILD=y because the logic in the docker-start target incorrectly attempts to start the docker service inside the container.

The Problem:
In DooD with native dockerd enabled, the host's dockerd is mounted into the slave container via /var/run/docker.sock dockerd is owned and managed by the host init system, NOT the container. 
The current logic: 
`test x$(SONIC_CONFIG_USE_NATIVE_DOCKERD_FOR_BUILD) != x"y" && sudo service docker status &> /dev/null || ( sudo service docker start &> /dev/null && ./scripts/wait_for_docker.sh 60 )` 
When SONIC_CONFIG_USE_NATIVE_DOCKERD_FOR_BUILD=y, the condition x$(SONIC_CONFIG_USE_NATIVE_DOCKERD_FOR_BUILD) != x"y" evaluates to FALSE The || operator then forces evaluation of the right side: sudo service docker start ... This fails because the container cannot manage the host's docker service.

##### Work item tracking

#### How I did it
Changed the docker-start target logic from a boolean expression to an explicit if/else.

#### How to verify it
- Test with native dockerd enabled (DooD scenario) - should succeed without attempting to manage the docker service `SONIC_CONFIG_USE_NATIVE_DOCKERD_FOR_BUILD=y make docker-start`

- Test without native dockerd (standard build) - should proceed with normal docker service checks 
`make docker-start
`

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511

- [ ] 
- [ ] 

Fixed docker-start target to skip service management when using native dockerd in DooD mode

### DIFF
--- a/slave.mk
+++ b/slave.mk
@@ -1092,7 +1092,11 @@ docker-start :
 	$(Q)sudo bash -c "{ echo \"export http_proxy=$$http_proxy\"; \
 	            echo \"export https_proxy=$$https_proxy\"; \
 	            echo \"export no_proxy=$$no_proxy\"; } >> /etc/default/docker"
-	$(Q)test x$(SONIC_CONFIG_USE_NATIVE_DOCKERD_FOR_BUILD) != x"y" && sudo service docker status &> /dev/null || ( sudo service docker start &> /dev/null && ./scripts/wait_for_docker.sh 60 )
+	$(Q)if [ x$(SONIC_CONFIG_USE_NATIVE_DOCKERD_FOR_BUILD) = x"y" ]; then \
+		test -S /var/run/docker.sock && sudo docker ps &> /dev/null || ./scripts/wait_for_docker.sh 60; \
+	else \
+		sudo service docker status &> /dev/null || ( sudo service docker start &> /dev/null && ./scripts/wait_for_docker.sh 60 ); \
+	fi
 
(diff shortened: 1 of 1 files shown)


sertem-ciena:
Contributing a picture of a cute animal: my bird :) 
image
@qiluo-msft, @xumia,  @lguohan, please review, thanks!

securely1g (commented):
Review on behalf of @lguohan:

Clean fix for DooD (Docker-outside-of-Docker) builds. The logic change is correct — when `SONIC_CONFIG_USE_NATIVE_DOCKERD_FOR_BUILD=y`, the container shouldn't try to restart the host's docker service.

The new approach (check socket exists + `docker ps` works, else wait) is more robust. LGTM.

lguohan (approved):

